Issue #10910, opened 2015-10-20
App banners causes flash of unstyled content


OliverJAsh:
This is because the CSS isn't inlined. Instead, the CSS arrives later and the styles are applied in a subsequent paint.

PR #10917, opened 2015-10-20
Prevent app banners from jumping the page

### LINKED ISSUE #10882: iOS banner causes page jump
When the iOS banner loads, it causes the whole page to jump.

Do we have tracking on how many users are seeing the iOS banner (iOS users who have *not* dismissed the banner)? There must be a large number of users who are receiving this poor experience.

OliverJAsh:
I think we need to improve the quality of our experience for first time users, as this is crucial to their impression of our site.

The app banners cause the whole page to jump when they load, causing the user to lose their reading position. This results in a poor user experience for all our first time users who are using an iOS or Android device (tablet only, because the sticky nav is disabled on mobile). Not only that, we continue to show the app banner until after the fourth impression, plus many users who browse in incognito mode will get this experience for every page view. 

The demos below demonstrate the following: load the page, scroll down to content, app banner loads, scroll up to down of page

Fixes #10882

To do:
- [x] Fix https://github.com/guardian/frontend/issues/10910 first to avoid double jump

/cc @johnduffell 
## Before

When the app banner loads, the page jumps.

## After

There is no longer any page jump when the app banner loads.

To allow the user a chance of seeing the app banner, we let the page jump if the user has not scrolled:

### DIFF
--- a/static/src/javascripts/projects/common/modules/ui/smartAppBanner.js
+++ b/static/src/javascripts/projects/common/modules/ui/smartAppBanner.js
@@ -1,4 +1,5 @@
 define([
+    'fastdom',
     'common/utils/$',
     'common/utils/cookies',
     'common/utils/detect',
@@ -7,6 +8,7 @@ define([
     'common/modules/user-prefs',
     'common/modules/ui/message'
 ], function (
+    fastdom,
     $,
     cookies,
     detect,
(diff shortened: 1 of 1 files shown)


static/src/javascripts/projects/common/modules/ui/smartAppBanner.js
```diff
@@ -58,7 +60,16 @@ define([
...
+
         cookies.add(COOKIE_IMPRESSION_KEY, impressions + 1);
+
+        fastdom.read(function () {
+            var $banner = $('.site-message--ios, .site-message--android');
+            var bannerHeight = $banner.dim().height;
+            if (window.scrollY !== 0) {
+                window.scrollTo(window.scrollX, window.scrollY + bannerHeight);
+            }
+        });
```
OliverJAsh: I wasn't sure if deferring the scroll update to the next animation frame would result in a flicker, but in my testing it seems fine.

johnduffell:
nice :+1:

gtrufitt:
:+1:  :fireworks:

PR #10918, opened 2015-10-20
Smaller header for immersive template on Safari mobile

sammorrisdesign-zz:
My terrible solution to a problem with Safari's implementation of `vh`. [Read More](http://nicolas-hoizey.com/2015/02/viewport-height-is-taller-than-the-visible-part-of-the-document-in-some-mobile-browsers.html) about the issue. Basically I'm having to shave some pixels off just for webkit on mobile, hence the `-webkit-calc`.

Before:
img_1620

After:
img_1619

### DIFF
--- a/static/src/stylesheets/module/content/_article-immersive.scss
+++ b/static/src/stylesheets/module/content/_article-immersive.scss
@@ -102,10 +102,14 @@
         .content__head--mobile {
             position: relative;
             height: 100vh;
-            height: -webkit-calc(100vh - 48px); // To compensate for mobile Safari viewport height issue.
             max-height: 800px;
             background-position: center center;
             background-size: cover;
+
+            // Query is to target Safari only
+            @media screen and (min-color-index:0) and (-webkit-min-device-pixel-ratio:0) {
+                height: -webkit-calc(100vh - 48px); // To compensate for mobile Safari viewport height issue.
+            }
         }
 
         .content__head--desktop {
--- a/static/src/stylesheets/module/content/_article-immersive.scss
+++ b/static/src/stylesheets/module/content/_article-immersive.scss
@@ -102,6 +102,7 @@
         .content__head--mobile {
             position: relative;
             height: 100vh;
+            height: -webkit-calc(100vh - 48px); // To compensate for mobile Safari viewport height issue.
             max-height: 800px;
             background-position: center center;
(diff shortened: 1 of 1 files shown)


static/src/stylesheets/module/content/_article-immersive.scss
```diff
@@ -102,6 +102,7 @@
         .content__head--mobile {
             position: relative;
             height: 100vh;
+            height: -webkit-calc(100vh - 48px); // To compensate for mobile Safari viewport height issue.
```
OliverJAsh: Won't this match in Blink, too?

sammorrisdesign: No real alternative that I can think of

OliverJAsh: So will this appear broken in Blink? Not sure if there's an alternative, can't think of one

sammorrisdesign: It means you can see a bit of the standfirst on mobile Chrome. Unfortunately...

NataliaLKB: There must be a way to only target safari, and not affect chrome. Did you take a look at [this stack overflow question](http://stackoverflow.com/questions/16348489/is-there-a-css-hack-for-safari-only-not-chrome) This seems like a possible solution:

> Now we can target iOS devices only. As mentioned above, since Chrome on iOS is rooted in Safari, it of course hits that one as well.

```css
/* Safari 9.0+ (iOS Only) */

@supports (-webkit-text-size-adjust:none) and (not (-ms-accelerator:true))
{

  .safari_only {
    color:#0000FF; 
    background-color:#CCCCCC; 
  }

}
```

There are only "hacks" mentioned which may be worth trying.

sammorrisdesign:
@OliverJAsh I've added the hack and it appears to work. Although is gross. I've added a notice to explain

static/src/stylesheets/module/content/_article-immersive.scss
OliverJAsh: There's no guarantee that browsers will not implement/remove these declarations in the future, thus changing the result of this. But I don't have a better suggestion, so :+1:

OliverJAsh:
:+1:

PR #10919, opened 2015-10-20
Allow the letter I to be used as a drop cap

sammorrisdesign-zz:
We make sure the word that the cap is applied to is 3 or more characters long, so there's no need to remove `I` from the drop-cappable characters list.

Before:
screen shot 2015-10-20 at 14 54 13

After:
screen shot 2015-10-20 at 14 54 20


Jholder112233:
:+1:

PR #10908, opened 2015-10-19
Layout fixed/small/slow-V-mpu Slice Without MPU

kelvin-chappell:
Now container has different layouts according to whether or not it should show an MPU.

/cc @kenlim @janua

PR #10920, opened 2015-10-20
Layout news/most-popular Container Without MPU

kelvin-chappell:
This is a slightly different solution to #10908 because it's not a fixed container type.
Others to come should all be slight variations of #10908.

Popular container with MPU:
image

Popular container without MPU:
image

/cc @kenlim @janua @sammorrisdesign

### DIFF
--- a/common/app/layout/Front.scala
+++ b/common/app/layout/Front.scala
@@ -106,7 +106,8 @@ object ContainerCommercialOptions {
     DfpAgent.isAdvertisementFeature(config),
     DfpAgent.isFoundationSupported(config),
     DfpAgent.sponsorshipTag(config),
-    DfpAgent.sponsorshipType(config)
+    DfpAgent.sponsorshipType(config),
+    omitMPU = false
   )
 
   val empty = ContainerCommercialOptions(
@@ -114,7 +115,8 @@ object ContainerCommercialOptions {
     isAdvertisementFeature = false,
     isFoundationSupported = false,
     sponsorshipTag = None,
-    sponsorshipType = None
+    sponsorshipType = None,
+    omitMPU = false
   )
 
   def mostPopular(omitMPU: Boolean) = empty.copy(omitMPU = omitMPU)
@@ -126,7 +128,7 @@ case class ContainerCommercialOptions(
   isFoundationSupported: Boolean,
   sponsorshipTag: Option[SponsorshipTag],
   sponsorshipType: Option[String],
-  omitMPU: Boolean = false
+  omitMPU: Boolean
 ) {
   val isPaidFor = isSponsored || isAdvertisementFeature || isFoundationSupported
 }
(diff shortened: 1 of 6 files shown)


common/app/layout/Front.scala
```diff
@@ -122,7 +124,8 @@ case class ContainerCommercialOptions(
   isAdvertisementFeature: Boolean,
   isFoundationSupported: Boolean,
   sponsorshipTag: Option[SponsorshipTag],
-  sponsorshipType: Option[String]
+  sponsorshipType: Option[String],
+  omitMPU: Boolean = false
```
kelvin-chappell: Adding an `omitMPU` flag to `CommercialOptions` so that it's easily available in templates.

common/app/layout/Front.scala
```diff
@@ -184,7 +188,7 @@ object FaciaContainer {
     config.config.showLatestUpdate,
     // popular containers should never be sponsored
     container match {
-      case MostPopular => ContainerCommercialOptions.empty
+      case MostPopular => ContainerCommercialOptions.mostPopular(omitMPU)
```
kelvin-chappell: It gets populated here.

common/app/layout/Front.scala
```diff
@@ -170,7 +173,8 @@ object FaciaContainer {
     config: CollectionConfigWithId,
     collectionEssentials: CollectionEssentials,
     containerLayout: Option[ContainerLayout],
-    componentId: Option[String]
+    componentId: Option[String],
+    omitMPU: Boolean = false
```
janua: Could we get rid of this default and instead change the callers? I don't think there's that many!

common/app/views/support/MostPopular.scala
```diff
@@ -0,0 +1,21 @@
+package views.support
+
+import layout.FaciaContainer
+import play.api.mvc.RequestHeader
+
+object MostPopular extends implicits.Requests {
+
+  def numItemsToShow(container: FaciaContainer)(implicit request: RequestHeader): Int = {
```
janua: What's using this `implicit` `RequestHeader`?

janua:
:+1:

common/app/views/support/MostPopular.scala
kelvin-chappell: Good point.  I was going to use it but ended up not.  I'll get rid of it.

common/app/layout/Front.scala
kelvin-chappell: I was afraid of making too many changes around this code, but I think you're right.  I'll change it.

janua: Yeah, I've had too many run ins with defaults to use them anymore, especially in case classes.
Don't mind them too much on methods!

kelvin-chappell: I've removed the default value from the case class now. This line is actually a default parameter value.

PR #10922, opened 2015-10-20
Badge for pages with page skin position fix

uplne:
Before:
screen shot 2015-10-20 at 14 50 23

After:
screen shot 2015-10-20 at 17 19 57


jbreckmckye:
So long as there's no conflict with anything else on the page, :+1:

PR #10923, opened 2015-10-20
Add byline to Immersive Header

sammorrisdesign-zz:
This PR adjusts spacing for the headline and standfirst and adds some new logic for standfirsts in the immersive template.

So, if the standfirst includes the contributor:
screen shot 2015-10-20 at 17 33 26

If the standfirst doesn't include the contributor:
screen shot 2015-10-20 at 17 33 30

Normal standfirsts on non-immersive pages get left alone.

This was the solution I've arrived but there might be a nicer way of getting there.

### DIFF
--- a/common/app/views/fragments/standfirst.scala.html
+++ b/common/app/views/fragments/standfirst.scala.html
@@ -15,7 +15,17 @@
     }
     @content.standfirst.map { s =>
         @defining(Edition(request)) { edition =>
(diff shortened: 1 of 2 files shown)


sndrs:
:+1: outstanding work :100: :rabbit:

johnduffell:
:+1: nice

PR #10924, opened 2015-10-20
Prevent app banners from appearing unstyled and FOUC

### LINKED ISSUE #10910: App banners causes flash of unstyled content

OliverJAsh:
Fixes #10910

This is a race condition: if the CSS loads before the JS executes, you don't get a flash, otherwise you do.

One solution would be to add this CSS to the inline CSS, but that would increase the inefficiency and bloat of our inline CSS, as we hardly ever show these banners.

This solution will force the JS to wait until the CSS has loaded before injecting the DOM element.

I wonder if we could move more CSS out of inline and towards this model of waiting to execute JS.

/cc @sndrs

### DIFF
--- a/common/app/templates/headerInlineJS/config.scala.js
+++ b/common/app/templates/headerInlineJS/config.scala.js
@@ -18,7 +18,8 @@ window.guardian = {
             )
     ),
     css: {
-        loaded: false
+        loaded: false,
+        loadedListeners: []
     },
     config: @defining(Edition(request)) { edition => {
         "page": @JavaScript(StringEncodings.jsonToJS(Json.stringify(JavaScriptPage(item).get))),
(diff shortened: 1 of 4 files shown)


common/app/templates/headerInlineJS/config.scala.js
```diff
@@ -18,7 +18,8 @@ window.guardian = {
             )
     ),
     css: {
-        loaded: false
```
OliverJAsh: This wasn't used, but now it is :smile:

common/app/templates/headerInlineJS/config.scala.js
```diff
@@ -18,7 +18,8 @@ window.guardian = {
...
+        loaded: false,
+        loadedListeners: []
```
sndrs: I'd call this `loadQueue` or something similar, since they're not listening, they're called

OliverJAsh: Done

sndrs:
:+1:

PR #10925, opened 2015-10-21
Fix syncing issues between service worker and offline page

### LINKED ISSUE #10667: Offline page and ServiceWorker out of sync
* SW caches /app-1.js
* /offline-page requests /app-2.js

This happens because, since the last deploy, /offline-page has been revalidated but /service-worker.js has not.

We need to find a way to lockstep these two files, to guarantee users will be caching the same version of assets requested by /offline-page.

Here is our current cache setup:

```
/service-worker.js
< Cache-Control: max-age=3600, stale-while-revalidate=360, stale-if-error=864000, private

/offline-page
< Cache-Control: max-age=60, stale-while-revalidate=6, stale-if-error=864000, private
```

OliverJAsh:
Fixes #10667

To do:
- [ ] Update nginx config to route `/offline-page.json` instead of `/offline-page`: https://github.com/guardian/platform/pull/844
- [ ] Update fastly SSL config to force HTTPS for `/offline-page.json` instead of `/offline-page`: https://github.com/guardian/fastly-edge-cache/pull/493

### DIFF
--- a/applications/app/templates/serviceWorker.scala.js
+++ b/applications/app/templates/serviceWorker.scala.js
@@ -19,18 +19,44 @@
         return new Date().toISOString().split('T')[0];
     };
 
+    var fetchAll = function (inputs) {
+        return Promise.all(inputs.map(function (input) {
+            return fetch(input);
+        }));
+    };
+
+    var cachePageAndAssetResponses = function (jsonResponse, assetResponses) {
+        var cacheName = [getISODate(), staticCacheName].join('-');
+        return caches.open(cacheName).then(function (cache) {
+            return jsonResponse.clone().json().then(function (jsonResponseJson) {
+                var pageRequest = new Request('/offline-page');
+                var pageResponse = new Response(jsonResponseJson.html, { headers: { 'Content-Type': 'text/html' } });
+                return Promise.all([
+                    cache.put(pageRequest, pageResponse)
+                ].concat(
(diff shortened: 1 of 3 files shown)


applications/app/templates/serviceWorker.scala.js
```diff
@@ -19,18 +19,44 @@
...
-                '/offline-page',
-                '@Static("stylesheets/head.content.css")',
-                '@Static("stylesheets/content.css")',
-                '@Static("stylesheets/print.css")',
-                // Crossword pages use jspm
-                '@StaticJspm("javascripts/core.js")',
-                '@StaticJspm("javascripts/bootstraps/app.js")',
-                '@StaticJspm("javascripts/es6/bootstraps/crosswords.js")'
-            ]);
+        // Fetch page and all assets. Iff all responses are OK then cache all assets and page.
```
gtrufitt: minor sp - iff

gtrufitt:
:+1:

applications/app/templates/serviceWorker.scala.js
OliverJAsh: it means if and only if: https://en.wikipedia.org/wiki/If_and_only_if

gtrufitt: oooooh right.

Issue #9482, opened 2015-06-15
Images in liveblogs requested at too low a resolution


paperboyo:
Hello,

Seen on FF, Chrome, IE11 on Win 8.1. On two breakpoints, liveblogs pages request images at lower resolution than they are shown (at 100% zoom).

On BP 740px-980px - they are shown at 700px but are requested at 620px.

On BP 1140px-1300px - they are shown at 780px but are requested at 620px!

This results in unnecessarily blurry images.

Loosely related to https://github.com/guardian/frontend/issues/9278.

BTW, since this problem manifests in different places, is there a _simple_ way to test the whole site for occurrences of imagery being requested at lower resolution than they are shown at 100% zoom? Then, we could catch them all.

And this is needed before we start to think about showing them at higher resolutions if the (desktop) browsers are zoomed in ~~(maybe by sniffing ``window.devicePixelRatio``?) :wink:~~

@robertberry 

Regards
Mateusz

ScottPainterGNM:
Is this still an issue or sorted by #10751 ??

paperboyo:
Yep, @ScottPainterGNM.

JonnyWeeks:
@OliverJAsh @paperboyo hey oliver and mateusz, it looks to my eyes as if the liveblog break-point issue is now fixed.... is that the case?? If so, thank you v much. jonny

OliverJAsh:
Yes, this is fixed now. Thanks

JonnyWeeks:
@OliverJAsh @paperboyo thanks for fixing the liveblog issue guys, hugely appreciated.

You know how i also mentioned that sometimes entire containers appear to be low res, well that issue has popped up again. I screengrabbed it for you and immediately tried to inspect element to get the specific deets for you, but as soon as i went to inspect element the site rejigged itself, an advert appeared alongside it and the content in the container suddenly appeared sharp. I don'y really understand it but to my ill-educated eye it does seem to be some kind of responsive issue. I've seen it over recent months and it always afflicts all of the images in one container at a time (somehow the images in neighbouring containers are fine). Here's the only screengrab I have:
![picture 604](https://cloud.githubusercontent.com/assets/11023438/10679851/e70b0064-7966-11e5-8e38-fc6b788d852f.png)

lemme know what you think
cheers
jonny
p.s. my browser is chrome and i was deffo viewing it at 100% when it looked wrong.

paperboyo:
@JonnyWeeks Thanks should go entirely to Oliver!

[...]

Issue #10493, opened 2015-09-13
Day view tag pages request images at absurdly low resolution


paperboyo:
Hello,

Steps:
1. Go to any Tag page (e.g. http://www.theguardian.com/commentisfree/series/observer-comment-cartoon or http://www.theguardian.com/world/series/eyewitness)
2. Click on any day link (e.g. http://www.theguardian.com/commentisfree/series/observer-comment-cartoon/2015/sep/05/all or http://www.theguardian.com/world/series/eyewitness/2015/sep/12/all)
3. Observe how images display at 940px wide (if there is just one per day) but are requested at 220px!
4. Weep.

Related to: https://github.com/guardian/frontend/issues/10467, https://github.com/guardian/frontend/issues/9482.

Also, would we know why the last cartoons on http://www.theguardian.com/cartoons/archive/all are from  11 April 2015?! (cc/ @na7hangood).

Regards
Mateusz

paperboyo:
Fixed by https://github.com/guardian/frontend/pull/10751. Thanks!

Issue #10467, opened 2015-09-10
Images on cartoon/picture type pages requested at too low a resolution


paperboyo:
Hello,

Example URL: http://www.theguardian.com/commentisfree/picture/2015/sep/09/happy-twonyversary-prime-minister

The images on Cartoon/Picture type pages are requested at 700px max while they are shown at 940px wide on widest breakpoint resulting in crappy quality highlighted by Autrialia's picture editors. There are higher res mastercrops available, we just need to add a correct `src` size.

cc/ @JonnyWeeks

Related to https://github.com/guardian/frontend/issues/9482 (and, possibly, more...).

Regards
Mateusz

ScottPainterGNM:
Is this still an issue or sorted by #10751 ??

paperboyo:
Still very much an issue, @ScottPainterGNM.

JonnyWeeks:
Hey Scott, yep as Mateusz says it's still very much an issue on
Conposer Picture/Cartoon pages. And the identical break-point issue still
exists for Composer Liveblogs as well, so the lion's share of our
photography looks pretty bad when viewed on desktop.

Just to clarify, the problems are only visible when you view the site using
a large browser. When you shrink your browser down to tablet size the
responsive nature of the site obviously kicks in and consequently the
photos all look fine at smaller sizes. Moreover the issue seems to
be specific only to images within the body of Composer Liveblog pages and
Composer Picture/Cartoon pages; by that I mean that expanded views are
unaffected, as are Composer Article pages.

Cheers
Jonny


On Tuesday, 13 October 2015, Mateusz <notifications@github.com> wrote:

> Still very much an issue, @ScottPainterGNM
> <https://github.com/ScottPainterGNM>.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/guardian/frontend/issues/10467#issuecomment-147410011>
> .
>


-- 



*Jonny Weeks*
*picture editor, **Guardian Australia*
*email: jonny.weeks@theguardian.com <jonny.weeks@theguardian.com>*
*mobile: +61(0)448018833*
*landline: +61(0)280768545*
*twitter: @WeeksJonny*
*website: **www.theguardian.com/au <http://www.theguardian.com/au>*
*address: **Level 2, 19 Foster St, Sydney, NSW 2010*

• How to supply photos to Guardian Australia <http://tinyurl.com/oqu38ap>
• How to use The Grid (for Australia staff only)
<http://tinyurl.com/nug8uor>

[...]

Issue #9278, opened 2015-05-27
Images on fronts requested at too low a resolution


paperboyo:
Hello,

It's hard to find all instances. The most offending occur in this new grey election special container that is being used sporadically for special events. Take a look, please, at the example from UK network front just now. Image resolution on widest BP is 700px at 100% zoom. Chrome 43 loads a low resolution 605px (sic!) image:

``
<img class="responsive-img" srcset="//i.guim.co.uk/static/w-445/h--/q-95/sys-images/Guardian/Pix/pictures/2015/5/27/1432712203406/5f19111d-2bd1-4c9b-a601-e5f394ebc72d-620x372.jpeg 445w, //i.guim.co.uk/static/w-605/h--/q-95/sys-images/Guardian/Pix/pictures/2015/5/27/1432712203406/5f19111d-2bd1-4c9b-a601-e5f394ebc72d-620x372.jpeg 605w" sizes="100%" src="//i.guim.co.uk/static/w-300/h--/q-95/sys-images/Guardian/Pix/pictures/2015/5/27/1432712203406/5f19111d-2bd1-4c9b-a601-e5f394ebc72d-620x372.jpeg">
``

But Firefox 38.0.1 loads a 445px wide image!

``
<img class="responsive-img" srcset="//i.guim.co.uk/static/w-445/h--/q-95/sys-images/Guardian/Pix/pictures/2015/5/27/1432712203406/5f19111d-2bd1-4c9b-a601-e5f394ebc72d-620x372.jpeg 445w, //i.guim.co.uk/static/w-605/h--/q-95/sys-images/Guardian/Pix/pictures/2015/5/27/1432712203406/5f19111d-2bd1-4c9b-a601-e5f394ebc72d-620x372.jpeg 605w" sizes="100%" src="//i.guim.co.uk/static/w-300/h--/q-95/sys-images/Guardian/Pix/pictures/2015/5/27/1432712203406/5f19111d-2bd1-4c9b-a601-e5f394ebc72d-620x372.jpeg">
``

Another instance was visible in Cannes uncovered series container (visible here http://www.theguardian.com/film/filmblog/2015/may/13/cannes-uncovered-day-1), when there wasn't much stories and the tiles were full-width. Now, they are fine...

Ideally, there would be a way to automatically look for images that are requested at the lower resolution than they are shown at 100% zoom.

[...]

PR #10928, opened 2015-10-21
Use different image sizes for immersive header

sammorrisdesign-zz:
Use appropriately sized images on the immersive header. Had to break the seal on the `Breakpoint` trait, not sure if there's a better way to do it.

### DIFF
--- a/common/app/layout/Breakpoint.scala
+++ b/common/app/layout/Breakpoint.scala
@@ -2,7 +2,7 @@ package layout
 
 import scalaz.syntax.std.option._
 
-sealed trait Breakpoint {
+trait Breakpoint {
   val minWidth: Option[Int]
 }
 
--- a/common/app/layout/Breakpoint.scala
+++ b/common/app/layout/Breakpoint.scala
@@ -2,7 +2,7 @@ package layout
 
 import scalaz.syntax.std.option._
 
-trait Breakpoint {
+sealed trait Breakpoint {
   val minWidth: Option[Int]
 }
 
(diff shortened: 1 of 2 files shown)


paperboyo:
Woohoo! 2000px sounds (looks!) much better than 1300px. Even though, I'm on 2560px ;) Thanks, @sammorrisdesign!

paperboyo:
@sammorrisdesign Noticed something... On all breakpoints showcased images are requested at 20px less that they are shown at 100% zoom - have the text column got wider? This results in unnecessarily blurry images. Similar problem to recently fixed: https://github.com/guardian/frontend/issues/9482, https://github.com/guardian/frontend/issues/9278, https://github.com/guardian/frontend/issues/10467, https://github.com/guardian/frontend/issues/10493. Do you want me to open a separate issue?

sammorrisdesign:
@paperboyo Yeah those are mid-article images, not this. So separate issue.

paperboyo:
New issue opened: https://github.com/guardian/frontend/issues/10947.

common/app/layout/Breakpoint.scala
```diff
@@ -2,7 +2,7 @@ package layout
 
 import scalaz.syntax.std.option._
 
-sealed trait Breakpoint {
```
sndrs: i have no idea if this is a good thing or a bad thing

sndrs:
great apart from comment - might want to get @jennysivapalan or @janua to check that.

:+1: apart form that

common/app/layout/Breakpoint.scala
johnduffell: what happens if you don't break it? I thought sealed just meant you couldn't define Breakpoint instances in any other file.

sammorrisdesign: Yup you're right, I was being an idiot. Reverted it now.

johnduffell: :stuck_out_tongue_winking_eye:

PR #10934, opened 2015-10-21
Force hide the fullscreen image icon AMP

zeftilldeath:
Temporary fix until the feature is added. Also bumped down the weight of the section label for good measure.


sammorrisdesign:
:+1: For fix
:-1: for sneaky font weight change

PR #10935, opened 2015-10-22
shrink the buckets and run the test for twice as long

johnduffell:
this leaves some space for other tests @Jholder112233


Jholder112233:
:+1:

PR #10940, opened 2015-10-22
Re-use AssetMap

OliverJAsh:
I made this change because AssetMap uses a different path for the asset map file in dev, and I wanted to re-use this for `systemJs.jspmAssetMap`.


janua:
:+1:

PR #10941, opened 2015-10-22
AB test: your morning briefing container

holder-josh:
An AB test to replace the related content container with the headlines container from the network front and label it 'Your morning briefing'.

The switch will happen in the following circumstances:
1. You're on the UK edition
2. Your machine's time is between 6am and 11am
3. You're reading an article in the world, uk-news or politics section

It's a 20% audience test, so 10% in the control, 10% in the variant.

**This creates a new endpoint which allows you to request any container in the layout it has on fronts**

screen shot 2015-10-22 at 12 44 26


crifmulholland:
This looks great Josh. Can you please make sure the campaign code on the daily email signup treat is unique to this test? We're doing a lot of work with email sign up, and so this test will be interesting

Jholder112233:
@crifmulholland good shout, I've added an extra data-link-name to the treat called 'morning-briefing-ab-treat'. That should be enough to identify it.

common/app/conf/switches/ABTestSwitches.scala
```diff
@@ -58,4 +58,13 @@ trait ABTestSwitches {
     exposeClientSide = true
   )
 
+  val ABInjectHeadlinesTest = Switch(
+    "A/B Tests",
+    "ab-inject-headlines-test",
+    "Switch to inject the headlines container instead of related content in the world, uk-news and politics sections between the hours of 6am-11am on the UK edition",
```
johnduffell: good description!

johnduffell:
looks good :+1: can't wait to see the results

OliverJAsh:
Awesome, can't wait to see the results. :+1:

PR #10943, opened 2015-10-22
Update A/B test to track correctly

gtrufitt:
I wasn't tracking the message showing in omniture so needed to ensure the correct vars were set on the event.

Other minor updates like not showing the message on email sign-up page and not opening the link in a new page.


OliverJAsh:
:+1:

PR #10909, opened 2015-10-20
Fetch user features from membership API

jbreckmckye:
This is a change for the ad-free experience product.

In it, we
- get the user's available features for their membership level, from a web service
- store that data and an expiry date together as cookies
- in the head of the page, read the data and use it to decide whether to apply an 'adfree' class to the page (which will hide ad elements before they render, once some CSS is added)
- in the post-render scripts, provide a proxy to the state those pre-render scripts choose
- in the post-render scripts, decide whether the data is trustworthy and dispatch requests for new data if needs be
- clean up lingering historical user data
- hinge this behaviour upon a switch, and clean up unwanted user data when the switch is disabled

**Some shortcomings**
- I haven't yet been able to unit test this code, but I know that others (@Calanthe) needed this code in place sooner rather than later

**Queries**
- I know that dotcom have been thinking about testing preflight scripts, or bundling them from external static JS files which Karma can read from. But my understanding is that not much work has been done on this yet. Does anyone know any differently? As is, my pre-flight code is difficult to test.

@Calanthe, @uplne

PR #10944, opened 2015-10-22
Unit test user-features.js

jbreckmckye:
This pull request adds a unit test for the user-features.js file added in #10909.

In it, we verify that:
- [x] the isAdfree method will either defer to what's set by the preflight JavaScript
- [x] we'll only make AJAX requests for user feature data when we actually need them
- [x] we accompany our persisted feature data with a future expiry date


jbreckmckye:
This has been invalidated by changes in #10945; I'm going to open a new PR once is resolve those differences.

PR #10945, opened 2015-10-22
Do not show adblock message for paying members

Calanthe:
This PR uses logic from the https://github.com/guardian/frontend/pull/10909 to create and check the cookie which tells if user is a paying member.


static/src/javascripts/projects/common/modules/commercial/user-features.js
```diff
@@ -39,8 +40,16 @@ define([
         }
     }
 
+    function isPayingMember() {
+        if (cookies.get(PERSISTENCE_KEYS.PAYING_MEMBER_COOKIE) !== undefined && cookies.get(PERSISTENCE_KEYS.PAYING_MEMBER_COOKIE) !== 'false') {
```
jbreckmckye: I would just write

    return !featuresDataIsOld() && cookies.get(PERSISTENCE_KEYS.PAYING_MEMBER_COOKIE) === 'true'

jbreckmckye:
:+1:

jbreckmckye:
I've added another commit to @Calanthe's PR - I needed to make a change in order for the message to never appear to logged in users without cookies (but still appear to logged out users). Could someone take another look at this before I merge it?

kelvin-chappell:
:+1: 
but would it be possible to create a membership module similar to the identity module to query for `isMember()` and `shouldHaveAdFree()` or something like that to hide the implementation details?

jbreckmckye:
@kelvin-chappell - I guess that's what user-features.js should be doing, but it's also responsible for updating the persisted data right now. Perhaps naming it after the web service would be more idiomatic?

PR #10946, opened 2015-10-22
Update omnitureMedia.js

mkopka:
```
     Updated line of code as we only need to pass value into eVar11.
```

   s.eVar11 = s.prop11 = isEmbed ? 'Embedded' : config.page.sectionName || '';


rich-nguyen:
:+1:

PR #10949, opened 2015-10-22
Add shrinkwrap cleaner and npm install docs

OliverJAsh:
`npm shrinkwrap` spits out different metadata depending on the npm version, which makes our diffs much too difficult to read properly.

Browsing [the issue](https://github.com/npm/npm/issues/3581) I found [a little script](https://github.com/angular/angular.js/blob/master/scripts/npm/clean-shrinkwrap.js) we can use to tidy up the shrinkwrap and keep it consistent.

I've also tried to document the npm install process.


sndrs:
great :+1: 

might be nice to make a `make install` command that does all the installation steps for you?

Issue #10936, opened 2015-10-22
When SW is installed, HTTPS->HTTP redirects result in offline page


OliverJAsh:
The [fetch event in the Service Worker](https://github.com/guardian/frontend/blob/c54f36f673e647e2f24933ccf0d7019728b48b53/applications/app/templates/serviceWorker.scala.js#L95) will throw when: you make a navigation request to a HTTPS page that results in HTTP redirect (302 response with a HTTP `Location`), i.e. https://www.theguardian.com/crosswords. This means you will see the offline page.

`fetch` will follow the redirect but throw a mixed content error when it makes a HTTP request.

According to @jakearchibald, this is a bug in Chrome:

>  The request redirect mode is "manual" because it's a navigation that means fetch() shouldn't follow redirects, it should return them. So the response it gives back would be the HTTPS same-origin response containing the 302 or whatever

I don't think any of our users will be effected by this, because, as far as I know, none of our links are protocol relative.

Demo: https://jakearchibald.github.io/isserviceworkerready/demos/http-redirect/
Chrome bug: https://code.google.com/p/chromium/issues/detail?id=546495

OliverJAsh:
This is a problem for custom redirects, i.e.:

http://www.theguardian.com/info/2015/apr/27/removed-embargoed-article

which automatically redirects to:

https://www.theguardian.com/info/2015/apr/27/removed-embargoed-article

and then redirects (via our custom redirects) to:

http://www.theguardian.com/environment/2015/apr/27/east-australia-one-of-11-areas-account-for-80-of-world-forest-loss-by-2030

This throws a mixed content error.

OliverJAsh:
We have worked around this issue in https://github.com/guardian/frontend/pull/10950

OliverJAsh:
This browser bug has been solved and will be fixed in Chrome 49. https://code.google.com/p/chromium/issues/detail?id=546495

ScottPainterGNM:
@OliverJAsh safe to close in which case?

OliverJAsh:
Yup, workaround is in place which I can revert now anyway :+1:

PR #10950, opened 2015-10-22
Workaround bug in Chrome which results in broken HTTPS->HTTP redirects

OliverJAsh:
Workaround for #10936


OliverJAsh:
Unfortunately this doesn't fix the real world issue with custom redirects we have described here: https://github.com/guardian/frontend/issues/10936#issuecomment-151490371

jennysivapalan:
:+1:

PR #10951, opened 2015-10-22
footer share buttons

desbo:
As well as promoting the sharing links to buttons, this splits out the AMP/non-amp footer templates. 

I'm also working on refactoring the footer to make it much easier to work with, once that's done we'll be able to simplify it and maybe bring the AMP one back in.

screen shot 2015-10-22 at 19 28 05

screen shot 2015-10-22 at 19 41 16

screen shot 2015-10-22 at 19 41 28

### DIFF
--- a/static/src/stylesheets/layout/_footer.scss
+++ b/static/src/stylesheets/layout/_footer.scss
@@ -78,7 +78,6 @@ $c-primary-footer-background-side-bar: mix($c-primary-footer-background, #ffffff
     &,
     a {
         color: colour(neutral-4);
-        text-transform: lowercase;
     }
 }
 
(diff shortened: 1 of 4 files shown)


paperboyo:
Oh, there is a place for RSS finally! And here! Oh, there isn’t :(

static/src/stylesheets/layout/_footer.scss
```diff
@@ -103,12 +108,63 @@ $c-primary-footer-background-side-bar: mix($c-primary-footer-background, #ffffff
...
+
+    @include mq(tablet) {
+        display: inline;
+    }
+}
+
+.footer__follow-us__title {
+    @include fs-header(3);
+    color: colour(neutral-7);
+    text-transform: initial;
```
gtrufitt: initial isn't in < IE10, does this need to be specific?

gtrufitt:
:+1:  All good other than minor comment.

PR #10952, opened 2015-10-23
Layout dynamic/slow-mpu Container Without MPU

kelvin-chappell:
Now this container has different layouts according to whether or not it should show an MPU.

Single slice with MPU:
image

Single slice without MPU:
image

Multiple slice with MPU:
image

Multiple slice without MPU:
image

/cc @Pearson82 @janua

### DIFF
--- a/common/app/views/support/GetClasses.scala
+++ b/common/app/views/support/GetClasses.scala
@@ -1,9 +1,9 @@
 package views.support
 
-import com.gu.facia.api.utils.{Audio, Video, Gallery}
-import layout._
+import com.gu.facia.api.utils.{Audio, Gallery, Video}
 import conf.switches.Switches.SaveForLaterSwitch
-import slices.{DynamicSlowMPU, Dynamic}
+import layout._
+import slices.{Dynamic, DynamicSlowMPU}
 
 object GetClasses {
   def forHtmlBlob(item: HtmlBlob) = {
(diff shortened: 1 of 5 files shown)


Pearson82:
Layout looks good to me :+1:

common/app/views/support/GetClasses.scala
```diff
@@ -58,7 +58,7 @@ object GetClasses {
       containerDefinition.showLatestUpdate,
       containerDefinition.index == 0 && containerDefinition.customHeader.isEmpty,
       containerDefinition.displayName.isDefined,
-      containerDefinition.displayName == Some("headlines"),
+      containerDefinition.displayName.contains("headlines"),
```
janua: Won't this fail on `scala 2.10`?
Locally I think this won't compile for me, as `Option` doesn't have a `contain` method before `2.10`.

janua:
:+1:

common/app/views/support/GetClasses.scala
kelvin-chappell: See https://github.com/guardian/frontend/blob/master/project/Prototypes.scala#L26

You're behind the times!

janua: I'm not sure why, but I have been bitten by this about 3 times now and I **thought** I was on 2.11 but it then fails to compile and had to revert it.

PR #10953, opened 2015-10-23
Show the EU cookies message to UK and INT edition readers.

JustinPinner:
This variant doesn't depend on Fastly and is behind a switch.

### DIFF
--- a/static/src/javascripts/projects/common/modules/ui/cookiesBanner.js
+++ b/static/src/javascripts/projects/common/modules/ui/cookiesBanner.js
@@ -20,24 +20,27 @@ define([
     /**
      * Rules:
      *
-     * EU visitors only
+     * UK / INT edition readers only
      * Never seen the cookie message before
      * Show once only
      * Show only on FIRST page view
      * Persist close state
      */
     function init() {
-        var EU_COOKIE_MSG = 'GU_EU_MSG',
-            euMessageCookie = cookies.get(EU_COOKIE_MSG);
-
-        if (euMessageCookie && euMessageCookie === 'unseen') {
-            var link = 'https://www.theguardian.com/info/cookies',
-                txt = 'Welcome to the Guardian. This site uses cookies, read our policy <a href="' + link + '" class="cookie-message__link">here</a>',
-                opts = {important: true},
-                cookieLifeDays = 365,
-                msg = new Message('cookies');
-            msg.show(txt, opts);
-            cookies.add(EU_COOKIE_MSG, 'seen', cookieLifeDays);
+        if (config.switches.euCookieMsg) {
+            if (config.page.edition === 'UK' || config.page.edition === 'INT') {
+                var EU_COOKIE_MSG = 'GU_EU_MSG',
+                    euMessageCookie = cookies.get(EU_COOKIE_MSG);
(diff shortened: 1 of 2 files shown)


static/src/javascripts/projects/common/modules/ui/cookiesBanner.js
```diff
@@ -20,24 +20,27 @@ define([
...
-                cookieLifeDays = 365,
-                msg = new Message('cookies');
-            msg.show(txt, opts);
-            cookies.add(EU_COOKIE_MSG, 'seen', cookieLifeDays);
+        if (config.switches.euCookieMsg) {
+            if (config.page.edition === 'UK' || config.page.edition === 'INT') {
+                var EU_COOKIE_MSG = 'GU_EU_MSG',
+                    euMessageCookie = cookies.get(EU_COOKIE_MSG);
+                if (!euMessageCookie || euMessageCookie != 'seen') {
+                    var link = 'https://www.theguardian.com/info/cookies',
```
jbreckmckye: I don't really understand why we have this 'one var' rule in our codebase.

static/src/javascripts/projects/common/modules/ui/cookiesBanner.js
```diff
@@ -20,24 +20,27 @@ define([
...
+                        txt = 'Welcome to the Guardian. This site uses cookies, read our policy <a href="' + link + '" class="cookie-message__link">here</a>',
+                        opts = {important: true},
+                        cookieLifeDays = 365,
```
johnduffell: so it shows it once a year?

static/src/javascripts/projects/common/modules/ui/cookiesBanner.js
johnduffell: this has been done a couple of times, and from my memory I think the summary was that it's mitigated by functions being kept fairly focussed (massive functions have other problems not just scoping issues), and it's also working around a surprising for some effect of javascript where the vars are declared at that point even if you write them in the code later.
However I'm not in either camp (other than the keep functions short camp)

johnduffell:
code looks fine but I don't think it should be a big problem to rely on fastly geoip - after all it knows where you are!  weather works quite well that way.

JustinPinner:
So is that a :+1: ?

JustinPinner:
Also weather is an additional async request which we're trying to avoid here.

static/src/javascripts/projects/common/modules/ui/cookiesBanner.js
jbreckmckye: I tend to think that most of the code that would hinge on the vagaries of hoisting behaviour would be weird in other ways too, e.g. using the same name on an inner and outer variable (which would be confusing). As you allude, I tend to think these ambiguities are mitigated by other good practices like keeping a function small and focused!

Once we're in ES6 land perhaps we can just put this issue to bed by banning `var` and using `const` as much as possible. (Unrelated fact: according to the spec, consts aren't supposed to be hoisted... which is actually pretty weird)

johnduffell:
:+1: code looks fine, in the next PR need to revisit to make sure non-EU users don't see this

PR #10954, opened 2015-10-23
Remove most popular container location AB test

holder-josh:
Test has finished, results incoming soon. Either way, this code won't be used and is easy to reintroduce in a separate test anyway, so let's remove it.


johnduffell:
:+1:

PR #10957, opened 2015-10-23
footer grid fixes

desbo:
A few tweaks to the footer: line it up with the grid, remove the "follow" header and make the buttons small so the text fits in everywhere.

screen shot 2015-10-23 at 17 04 58
screen shot 2015-10-23 at 17 08 40
screen shot 2015-10-23 at 17 08 21

@superfrank


superfrank:
(Y)

PR #10958, opened 2015-10-23
Revert Immersive header fix

sammorrisdesign-zz:
This actually breaks stuff, I will investigate more next week, but don't want this in current state over weekend


sndrs:
:+1:

PR #10960, opened 2015-10-26
Adds more logging for most popular video 

jennysivapalan:
So that we can identify why we occasionally drop the most popular video component.


sndrs:
:+1:

PR #10961, opened 2015-10-26
Send GU_U and SC_GU_U cookies on the membership AJAX request

jbreckmckye:
We're trying to call the membership service to know if a user is paying us money, and stop sending them anti-adblock messages if they are, but the request to the server doesn't actually send request cookies unless we explicitly demand that behaviour.

Needed to fix an issue in production related to #10945


uplne:
:+1:

PR #10962, opened 2015-10-26
Right mpu space fix

uplne:
Before: Match stats are pushed down by MPU
screen shot 2015-10-26 at 11 22 21

After: We are not loading MPU for 'Match Reports' articles.
screen shot 2015-10-23 at 17 28 06

@jennysivapalan @mchv We might need to look at this a little bit further as for the very short articles even removing MPU isn't enough and match stats are going below the comments. It may need some design decisions.


static/src/javascripts/projects/common/modules/commercial/article-aside-adverts.js
```diff
@@ -32,7 +32,7 @@ define([
             $adSlotContainer;
 
         // are article aside ads disabled, or secondary column hidden?
-        if (!commercialFeatures.articleAsideAdverts || colIsHidden) {
+        if (!commercialFeatures.articleAsideAdverts || colIsHidden || config.hasTone('Match reports')) {
```
jbreckmckye: Is this going to be here permanently? Perhaps there's a way we can handle it in commercialFeaturePolicies. Perhaps a tonePolicy that applies all of these little switches.

uplne:
@jbreckmckye great idea. I've updated the code.

jbreckmckye:
:+1:

PR #10966, opened 2015-10-26
Unit test membership JS

jbreckmckye:
Let's cover the functionality of `user-features.js` with some unit tests.

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/user-features.js
+++ b/static/src/javascripts/projects/common/modules/commercial/user-features.js
@@ -13,11 +13,33 @@ define([
     storage,
     identity
 ) {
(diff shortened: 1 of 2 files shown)


kelvin-chappell:
:+1: 

Though still think it would be nice to separate out user features that are membership features from those that are features from being logged in or a subscriber, for example.  I know there aren't any at the moment.

jbreckmckye:
@kelvin-chappell - so, you mean making `user-features` just `membership-features`? I'm not sure I understand the difference between being a member and a subscriber.

kelvin-chappell:
Or just `membership`, but it doesn't need to be done in this PR.  This is fine to go.

piuccio:
:heart_eyes:

PR #10969, opened 2015-10-27
Normalise short urls before archive dynamo lookup

rich-nguyen:
This change means that when redirects are added for short urls like `www.theguardian.com/p/4djw9`,  requests will still be redirected even when they have campaign codes, like `www.theguardian.com/p/4djw9/stw`. Normal rules apply (the normalised url must be a 404 to trigger the redirect mechanism)


JustinPinner:
Looks :+1: to me.

jennysivapalan:
:+1:

PR #10970, opened 2015-10-27
Shrinkwrap cleaner fix

OliverJAsh:
Fixes a bug introduced by #10949 whereby the shrinkwrap cleaner incorrectly removes the `resolved` property if the dependency is coming from GitHub via HTTP(S). See https://github.com/guardian/frontend/commit/382fff0681b70431fe885e5fb72232215af855de


gtrufitt:
:+1:

PR #10968, opened 2015-10-27
Revert: use systemjs for javascript tests

OliverJAsh:
This is mostly a revert of #9103, but I had to manually downgrade the ES6 to ES5 and move any new tests/delete any old tests.

Revert "Merge pull request #9103 from guardian/systemjs-tests"

This reverts commit c51be021f31d15492bbf9b953ca81a723c6a95e5, reversing
changes made to 3803238b54be7e1f8b7d6e22f7958b9a400b6b34.

PR #10971, opened 2015-10-27
Test fixes

OliverJAsh:
Fixes for #10968

A test was edited in master, so when I merged #10968, the [build no longer passed](http://teamcity.gu-web.net:8111/viewLog.html?buildId=1569&buildTypeId=dotcom_master&tab=buildLog#_focus=1567). This edited test is migrated in e426d63e2880d5d4d7fc4481f0ee82930f7a4c82.

I also noticed the DFP tests were [failing intermittently](http://teamcity.gu-web.net:8111/viewLog.html?buildId=1573&buildTypeId=dotcom_pullrequests&tab=buildLog#_focus=1241). I have worked around this in 5c50e9b6f6d4618092d92771210ebba5a544db26. I think `addLabel` should return a promise, instead of crossing our fingers and hoping the timers run in the desired order? /cc @jbreckmckye and team


gtrufitt:
:+1:

jbreckmckye:
@OliverJAsh - should production code return promises just for the sake of testing its asynchronous side effects? I understand why it's necessary but it seems like an antipattern. It might make more sense to extract the logic that's actually under question and validate what it writes to fastdom rather than threading it into a fixture and waiting for the operation to complete. 

We've talked before about splitting dfp.js up into smaller modules. @uplne and @Calanthe may have some opinions on this.

Otherwise, though, :+1:

OliverJAsh:
Not sure about that to be honest @jbreckmckye. If a function is asynchronous I would expect an interface for that, simply out of correctness. Useful for testing and potentially future consumers of the function. I don't think it's much of an overhead? Interesting thoughts though, open minded :smile:

jbreckmckye:
@OliverJAsh - it's not much of an overhead, but it does imply to readers that there is a downstream dependency on that interface in production code. In the case of a continuation-passing style like this one, the actual dispatch isn't all that important to verifying the code under test. All we _really_ want to know is that the right HTML is going to be rendered, and we mightn't actually have to do any DOM manipulation to verify that.

OliverJAsh:
I think I see what you're saying. I don't mind, as long as we don't have race conditions like we do today.

PR #10972, opened 2015-10-27
Check whether a message is shown before firing the tracking hit

gtrufitt:
Return a bool from the message module show function, so that we can ensure we're tracking only when the message is definitely shown.


OliverJAsh:
:+1:

PR #10974, opened 2015-10-27
Only attempt ServiceWorker install on HTTPS pages

OliverJAsh:
This will prevent us from seeing `Uncaught (in promise) DOMException: Only secure origins are allowed` in the console on non-HTTP pages.


gtrufitt:
:+1:

PR #10975, opened 2015-10-27
Use the geoip.continent_code from fastly to determine EU readers

JustinPinner:
This is instead of showing to anyone reading the uk or int editions.

### DIFF
--- a/static/src/javascripts/projects/common/modules/ui/cookiesBanner.js
+++ b/static/src/javascripts/projects/common/modules/ui/cookiesBanner.js
@@ -28,7 +28,8 @@ define([
      */
     function init() {
         if (config.switches.euCookieMsg) {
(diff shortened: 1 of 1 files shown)


kelvin-chappell:
:+1:

static/src/javascripts/projects/common/modules/ui/cookiesBanner.js
```diff
@@ -28,7 +28,8 @@ define([
      */
     function init() {
         if (config.switches.euCookieMsg) {
-            if (config.page.edition === 'UK' || config.page.edition === 'INT') {
+            var geoContinentCookie = cookies.get('GU_geo_continent');
```
johnduffell: I'm surprised this lints, normally vars have to be at the start of the function

johnduffell:
:+1: great ship it! =)

PR #10976, opened 2015-10-27
AMP removed !important 

zeftilldeath:
Removed !important as it goes against Amp spec, and re-jigged ordering of the css to make sure it still works.


jbreckmckye:
!important isn't allowed on Amp? Why is that - does it let them parse the CSS faster somehow?

stephanfowler:
:+1:

stephanfowler:
@jbreckmckye the spec says "Usage of the !important qualifier is not allowed. This is a necessary requirement to enable AMP to enforce its element sizing invariants."

PR #10978, opened 2015-10-27
Update email message and cookie banner to prevent double shown message

gtrufitt:
This PR updates the cookie message with the important var, so that it always shows (overriding currently shown messages), however this meant that you were shown the email banner briefly  before cookie kicked in, so have added a listener to ensure that we only shown the email banner once the cookie banner fails any of its 'show' `if` statements.

### DIFF
--- a/static/src/javascripts/projects/common/modules/experiments/tests/rtrt-email-message.js
+++ b/static/src/javascripts/projects/common/modules/experiments/tests/rtrt-email-message.js
@@ -6,7 +6,8 @@ define([
     'common/views/svgs',
     'common/utils/_',
     'common/modules/commercial/third-party-tags/krux',
-    'common/modules/identity/api'
+    'common/modules/identity/api',
+    'common/utils/mediator'
 ], function (
     config,
     template,
@@ -15,7 +16,8 @@ define([
     svgs,
     _,
     krux,
-    Id
+    Id,
+    mediator
 ) {
 
     var messageId = 'rtrt-email-message';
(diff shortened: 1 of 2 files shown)


static/src/javascripts/projects/common/modules/ui/cookiesBanner.js
```diff
@@ -36,12 +38,15 @@ define([
                         txt = 'Welcome to the Guardian. This site uses cookies, read our policy <a href="' + link + '" class="cookie-message__link">here</a>',
                         opts = {important: true},
                         cookieLifeDays = 365,
-                        msg = new Message('cookies');
-                    msg.show(txt, opts);
+                        msg = new Message('cookies', opts);
```
gtrufitt: Ping @JustinPinner - I move the opts object as important is on the constructor.

static/src/javascripts/projects/common/modules/experiments/tests/rtrt-email-message.js
```diff
@@ -76,18 +79,22 @@ define([
             {
                 id: 'targeted-loyal-A',
                 test: function () {
-                    createMessage('p2lq8cs6r'); // 10 visits or more to the Guardian
+                    kruxSegmentId = 'p2lq8cs6r';
```
desbo: I think it'd be better to keep the segment id as a parameter to `createMessage`. Maybe it should return a function so you can just pass `createMessage(id)` to mediator.

gtrufitt: Good shout, have updated

desbo:
:+1:

PR #10981, opened 2015-10-27
Simplify drop cap rules

sammorrisdesign-zz:
Removes the two sizes of drop caps (now that we can get two in the same article) and it allows two character words to be drop capable (as per requests from editorial).

Note: tests are failing, have been fixed in [this PR](https://github.com/guardian/frontend/pull/10979) which needs to go through first


rich-nguyen:
:+1:

PR #10983, opened 2015-10-27
Remove minify more switch

janua:
This removes the `MoreMinify` switch which was used to switch off the closure compiler on sensitive parts of the `head`.

Everything seems to be fine, and this expires this friday.

### DIFF
--- a/common/app/conf/switches/PerformanceSwitches.scala
+++ b/common/app/conf/switches/PerformanceSwitches.scala
@@ -5,10 +5,10 @@ import org.joda.time.LocalDate
 
 trait PerformanceSwitches {
 
-  val ClosureCompilerStandardOptimisation = Switch(
+  val InlineJSStandardOptimisation = Switch(
     "Performance",
-    "closure-compiler-standard",
-    "If this switch is on, the closure compiler will use standard optimisation instead of whitespace only",
+    "inline-standard-optimisation",
+    "If this switch is on, the inline JS will be compressed using closure compiler's standard optimisation instead of whitespace only",
     safeState = On,
     sellByDate = never,
     exposeClientSide = false
--- a/common/app/conf/switches/PerformanceSwitches.scala
+++ b/common/app/conf/switches/PerformanceSwitches.scala
@@ -14,15 +14,6 @@ trait PerformanceSwitches {
     exposeClientSide = false
   )
 
-  val MoreInlineHead = Switch(
-    "Performance",
-    "minify-more-inline-js",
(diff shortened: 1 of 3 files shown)


rich-nguyen:
:+1: remove `MinifyInlineJsSwitch` too

janua:
@rich-nguyen @sndrs 
So after looking at the switch, I think it is still handy to keep. The problem with it is that the name was misleading. Now it's called `ClosureCompilerStandardOptimisation`. This means we can drop to `WHITESPACE_ONLY` if we need to.

You can't actually turn off the closure compiler in the head.

common/app/conf/switches/PerformanceSwitches.scala
```diff
@@ -5,24 +5,15 @@ import org.joda.time.LocalDate
 
 trait PerformanceSwitches {
 
-  val MinifyInlineJsSwitch = Switch(
+  val ClosureCompilerStandardOptimisation = Switch(
```
sndrs: maybe `InlineJSStandardOptimisation`? It's not closure that's the thing here, it's which JS is compressed

common/app/conf/switches/PerformanceSwitches.scala
```diff
@@ -5,24 +5,15 @@ import org.joda.time.LocalDate
...
     "Performance",
-    "minify-inline-js",
-    "If this switch is on, InlineJs output will be minified by closure compiler",
-    safeState = Off,
+    "closure-compiler-standard",
+    "If this switch is on, the closure compiler will use standard optimisation instead of whitespace only",
```
sndrs: "If this switch is on, the inline JS will be compressed using closure compiler's standard optimisation instead of whitespace only"

common/app/conf/switches/PerformanceSwitches.scala
```diff
@@ -5,24 +5,15 @@ import org.joda.time.LocalDate
 
 trait PerformanceSwitches {
 
-  val MinifyInlineJsSwitch = Switch(
+  val ClosureCompilerStandardOptimisation = Switch(
     "Performance",
-    "minify-inline-js",
-    "If this switch is on, InlineJs output will be minified by closure compiler",
-    safeState = Off,
+    "closure-compiler-standard",
```
sndrs: "inline-js-standard-optimisation"

sndrs:
:+1: apart from comments (which are only suggestions)

janua:
@sndrs Yes, english master.

PR #10985, opened 2015-10-28
always send omniture pixels over https

johnduffell:
we used to check a header to decide whether to send an https or http omniture pixel (for no-js)
However this header is set by the ELB and the router sends all requests over http.   So in amp (https) the pixel tracking wasn't working.

I considered making things https after the router, but it seemed overkill - we are not advertising end to end encryption, and most of the https benefits are service worker and ability to access our site where things are blocked or random ISPs inject stuff into the page.
I also considered putting in a new header at fastly to say if it's secure but it turned out there's only one bit of code that uses it anyway.

Another option suggested by @rich-nguyen was to put everything in a VPC, but maybe this is a job for the future.

@mkopka would that cause any issues at your end?

### DIFF
--- a/common/app/common/LinkTo.scala
+++ b/common/app/common/LinkTo.scala
@@ -2,6 +2,7 @@ package common
 
 import common.editions.{Au, Us, Uk}
 import conf.Configuration
+import conf.switches.Switches
 import implicits.Requests
 import layout.ContentCard
 import model.Trail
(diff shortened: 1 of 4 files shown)


johnduffell:
updated it to be behind a switch in case we have a disaster.  I'll need to update https://github.com/guardian/frontend/blob/master/article/test/AnalyticsFeatureTest.scala#L32 when we change the default to secure

stephanfowler:
:1

PR #10986, opened 2015-10-28
Make sticky nav defer to the sensitive content commercial policy

jbreckmckye:
This is a prerequisite to hiding the upper banner in adfree mode.

@Calanthe - I guess you're our sticky navigation 'guru'!


Calanthe:
:+1:

PR #10987, opened 2015-10-28
Improve display on immersive header for mobile

sammorrisdesign-zz:
Using Tables to fix everybody's problems..

giphy

Before:
unnamed

After:
unnamed-1

### DIFF
--- a/static/src/stylesheets/module/content/_article-immersive.scss
+++ b/static/src/stylesheets/module/content/_article-immersive.scss
@@ -98,11 +98,27 @@
         }
     }
 
+    .content__head--desktop {
+        height: 100vh;
+        max-height: 800px;
+
+        @include mq(desktop) {
+            max-height: auto;
+        }
+    }
+
+    .content__header {
+        position: absolute;
+        bottom: 0;
+        left: 0;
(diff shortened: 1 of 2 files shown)


static/src/stylesheets/module/content/_article-immersive.scss
```diff
@@ -98,11 +98,23 @@
         }
     }
 
+    .content__head--desktop {
+        height: 100vh;
```
jbreckmckye: Do we care about IE8? ('No' is a legitimate answer)

static/src/stylesheets/module/content/_article-immersive.scss
```diff
@@ -98,11 +98,23 @@
...
+        right: 0;
+    }
+
     @include mq($until: desktop) {
         .content__head--mobile {
             position: relative;
-            height: 100vh;
-            max-height: 800px;
+            display: table-cell;
+            vertical-align: bottom;
```
jbreckmckye: Do we need the parent to have display:table for this to work everywhere, or is this enough?

rich-nguyen:
:+1:

PR #10989, opened 2015-10-28
Fail grunt early if dependencies are invalid

OliverJAsh:
When we change update dependencies, we notify users on merge:

image

But it's still possible to run Grunt, and eventually Grunt will probably fail because it is expecting a different version of a certain dependency. It's easy to get confused when this happens, because there's no obvious cause, e.g. this happened to @jbreckmckye and others when I updated the jscs package yesterday.

We can catch these problems early by checking that our dependencies are in sync with the package manifest and shrinkwrap, using this tool: https://github.com/mzgol/check-dependencies

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,5 +1,13 @@
 'use strict';
 /* global module: false, process: false */
+
+var dependencyTest = require('check-dependencies').sync();
+
+if (dependencyTest.status !== 0) {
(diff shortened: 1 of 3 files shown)


OliverJAsh:
Here's a preview:

image

This is almost instant.

jbreckmckye:
Wow, really nice! :+1:

OliverJAsh:
Don't forget to run `npm install` when you pull master!

johnduffell:
awesome!

PR #10990, opened 2015-10-28
Remove bootstraps/dev

OliverJAsh:
I asked around and it seems like nobody uses this little dev utility which includes live updating CSS. I've had it commented out locally since I started because the web sockets seem to drive my CPU crazy.


piuccio:
:+1: :heart:

PR #10992, opened 2015-10-28
Update the fallback image to the new branding

jennysivapalan:
A request came in to update the fallback image for Facebook to match the new branding.
The old Facebook fallback image was on `static.guim` which we are not sure a) how to update b) think this will change with all the migration work. The Grid does not support pngs. So we thought we'll just add the fallback image to our repo.


OliverJAsh:
:+1:

PR #10996, opened 2015-10-28
Remove Ad-Feature Expiry Code

kelvin-chappell:
As it's being done further upstream now, and 410 status is coming through in Capi.


rich-nguyen:
:+1:

PR #9789, opened 2015-07-15
Sticky nav scenario with adblock

Calanthe:
Sticky nav scenario with adblock turned on.

stickynavadblock

PR #10997, opened 2015-10-28
Hide top ad banners in adfree mode

jbreckmckye:
- In adfree mode, hide the top banner with CSS before they even render (to avoid any flashing). We append a `:not(.is-adfree)` to the selectors which turn on slots for different breakpoints. IE8 doesn't get adverts, so its lack of :not support isn't a problem.
- In DFP.js, strip the ad slot from the DOM so that DFP doesn't make any redundant ad calls (which would waste network resources and skew our view measurements)


static/src/stylesheets/module/_adslot.scss
```diff
@@ -120,14 +120,14 @@
 }
 .top-banner-ad-container--above-nav {
     @include mq(tablet) {
-        .js-on & {
+        .js-on:not(.is-adfree) & {
```
Calanthe: Where do you apply that `is-adfree` class?

jbreckmckye: It's from the preflight scripts. See https://github.com/guardian/frontend/pull/10909

Calanthe:
It seems fine :+1:

paperboyo:
Oh, so we do have a way to fix huge-white-space-above-nav-in-ff-with-adblock? (https://github.com/guardian/frontend/pull/9789#issuecomment-133370552)

PR #10998, opened 2015-10-28
Throw Useful Exception when Asset Not Found

kelvin-chappell:


JustinPinner:
Nice :+1:

johnduffell:
yes yes yes :+1: ship it!

PR #11000, opened 2015-10-28
Redirects for short urls should pass through campaign code

rich-nguyen:
our archive redirect system needs to pass through the campaign code from the original request

### DIFF
--- a/applications/app/controllers/ShortUrlsCampaignController.scala
+++ b/applications/app/controllers/ShortUrlsCampaignController.scala
@@ -1,33 +1,13 @@
 package controllers
 
-import common.{LinkTo, Logging}
+import common.{ExecutionContexts, LinkTo, Logging}
 import common.`package`._
+import campaigns.ShortCampaignCodes
 import conf.LiveContentApi
 import model.Cached
 import play.api.mvc.{RequestHeader, Action, Controller}
 
-import scala.concurrent.ExecutionContext.Implicits.global
-
-object ShortUrlsCampaignController extends Controller with Logging {
-
-  val campaigns = Map (
-    "iw" -> "twt_ipd",
-    "wc" -> "twt_wc",
-    "tf" -> "twt_fd",
-    "fb" -> "fb_gu",
-    "fo" -> "fb_ot",
-    "us" -> "fb_us",
-    "au" -> "soc_567",
-    "tw" -> "twt_gu",
-    "at" -> "twt_atn",
-    "sfb"-> "share_btn_fb",
-    "ip" -> "twt_iph",
-    "stw" -> "share_btn_tw",
-    "swa" -> "share_btn_wa",
-    "em" -> "email",
-    "sgp" -> "share_btn_gp",
-    "sbl" -> "share_btn_link"
-  )
+object ShortUrlsCampaignController extends Controller with Logging with ExecutionContexts {
 
   def redirectShortUrl(shortUrl: String) = Action.async { implicit request =>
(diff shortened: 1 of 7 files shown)


applications/app/controllers/ShortUrlsCampaignController.scala
```diff
@@ -1,33 +1,13 @@
...
-    "at" -> "twt_atn",
-    "sfb"-> "share_btn_fb",
-    "ip" -> "twt_iph",
-    "stw" -> "share_btn_tw",
-    "swa" -> "share_btn_wa",
-    "em" -> "email",
-    "sgp" -> "share_btn_gp",
-    "sbl" -> "share_btn_link"
-  )
+object ShortUrlsCampaignController extends Controller with Logging with ExecutionContexts {
```
jennysivapalan: I meant to rename this as it fetches the full path for short urls with and without campaign codes. Shall we do it now?

rich-nguyen: yep, so `ShortUrlsController`?

jennysivapalan: :+1:

archive/app/controllers/ArchiveController.scala
```diff
@@ -59,6 +60,7 @@ object ArchiveController extends Controller with Logging with ExecutionContexts
         Some(path)
       case _ => None
     }
+    s"http://${normalised.getOrElse(path)}"
   }
```
jennysivapalan: Why has this changed?

archive/app/controllers/ArchiveController.scala
```diff
@@ -66,9 +68,31 @@ object ArchiveController extends Controller with Logging with ExecutionContexts
...
+
+    val destinationHasCampaign = destination.location match {
+      case shortUrlWithCampaign(_) => true
+      case urlWithCampaignParam() => true
+      case _ => false
+    }
+
+    path match {
+      case shortUrlWithCampaign(campaign) if !destinationHasCampaign => {
+        val uri = javax.ws.rs.core.UriBuilder.fromPath(destination.location)
```
jennysivapalan: Out of curiosity what made you choose this library? I just did a quick scan of the codebase (as I didn't know what library we used) and it seems `java.net.URI` is preferred?

archive/app/controllers/ArchiveController.scala
jennysivapalan: Ignore me - just seen the update below

jennysivapalan:
:+1:

archive/app/controllers/ArchiveController.scala
rich-nguyen: It's the same as before, another part of the controller was prefixing `http` to the normalised path. I decided to move all the string manipulation into `normalise`

archive/app/controllers/ArchiveController.scala
rich-nguyen: this library does actually build `java.net.URI` objects, they are fine. The thing I have bad history with is [`URLEncoder`](http://docs.oracle.com/javase/7/docs/api/java/net/URLEncoder.html), which looks like it can build sensible `java.net.URI` objects, but is specifically for encoding HTML forms urls (`application/x-www-form-urlencoded`). Despite its name, `URLEncoder` isn't actually rfc2396 compliant. This has bitten us regularly, so I recently added this Jersey library to help with the issue. I explicitly do not condone url building, it's well dodgy.

PR #11001, opened 2015-10-28
Don't show adfree mode until we have an unambiguous cookie value

jbreckmckye:
This is a change on the request of the business. It means that when a user first logs in, they always have at least one pageview with adverts, even if they are an adfree user.

This prevents ad impressions being lost when logged-in users without adfree first sign in.


johnduffell:
:+1: fair enough

PR #11002, opened 2015-10-28
Remove jspm

OliverJAsh:
Depends on https://github.com/guardian/frontend/pull/10999

This is a sad day. We're going back to a dev workflow with ~1 minute compile times _each time you want to make a change to your JavaScript_.

This test has been running for a long time and needed to be brought to an end as a prerequisite to the universal/core work we're doing. But we still have good reasons to invest in a new module loader:
- refresh to reload OR incremental bundling (instead of the 1 minute compile times we have today). This is significant to developer happiness/sanity
- ES6 (for our own sanity and to help with recruitment)
- depend on CSS from our JS, to help fix the inefficiency of our inline CSS

Not so important but nice to have:
- import any module format, not just AMD. This will be increasingly important as the world of open source moves away from distributing AMD

To do:
- [x] Delete `static/src/jspm_packages` on the TeamCity agents to prevent linting from hanging http://teamcity.gu-web.net:8111/viewLog.html?buildId=1667&buildTypeId=dotcom_pullrequests&tab=buildLog&state=&expand=all#_focus=288
- [ ] Remove `JSPM_GITHUB_AUTH` from TeamCity

Here is the test report https://docs.google.com/document/d/1CtrWiHPkCXSKl5afbQxY9hH5QELgYrt-9E8fMCMa7Pk/edit


theefer:
Shame. 

Also:
* simple but powerful bundling
* optional typescript support
* automated configuration when installing packages 
* deep dependency tree version locking (repeatable builds) 
Etc

OliverJAsh:
@desbo The diff is now clean :smile:

desbo:
:+1: :sob:

PR #11003, opened 2015-10-28
All retina iPads see universal fronts all the time.

JustinPinner:
The core fronts test for retina iPads consistently showed an improvement in the length of time those readers were on the site, and we had a marked drop in emails to user help, so we're making the change permanent.

@michaelwmcnamara are you happy with these new beacons?


johnduffell:
great :+1: happy ipad users! :ship:

stephanfowler:
:+1: - though this increases the pressure to fix ad loading (think only the first one gets loaded?)

michaelwmcnamara:
+1 does exactly what it says on the tin :-)

JustinPinner:
We're waiting for some comms to happen before merging. Shouldn't be too long.

JustinPinner:
@michaelwmcnamara you mentioned this had been given the go-ahead. Could you confirm that here? Thanks.

michaelwmcnamara:
Confirmed - we are good to go.

Issue #10034, opened 2015-08-05
Prerolls do not play when using multiple players in one article


Calanthe:
On pages with multiple players (like here http://www.theguardian.com/us-news/2015/jul/21/sandra-bland-dashcam-video-arrest-released?adtest=12 - if after few refreshes preroll does not work at all, ask ad-opts to turn the test campaign again) after playing the first video, an ad starts to play as well in the background and it seems to be attached to the second video. It is impossible to turn it off.
![screen shot 2015-08-05 at 12 27 09](https://cloud.githubusercontent.com/assets/489567/9084678/df079c42-3b6d-11e5-8a2c-015425121188.png)
(notice the 'You may skip this advert (...)' button which does not work)

After investigation I noticed that there is something wrong with the `player` object sent by the video.js.ima (https://github.com/guardian/frontend/blob/157d7230cd1dbc1837ddb8d3c2b55de50862234d/static/src/javascripts/components/videojs-ima/videojs.ima.js#L188). After clicking and running the first video on page, the returned player has always wrong scope - second video tag instead of the first one (https://github.com/guardian/frontend/blob/5fa70b691c239b63fdafae290d0cbf2080460f8b/static/src/javascripts/projects/common/modules/video/events.js#L91). The state diagram from https://github.com/videojs/videojs-contrib-ads might be useful as well.

There is a similar (I think) issue raised in the videojs-ima repository - https://github.com/googleads/videojs-ima/issues/89. This needed to be discussed in more details with someone from the sustain team.

There is also a console error while playing the ad:
![screen shot 2015-08-05 at 12 30 23](https://cloud.githubusercontent.com/assets/489567/9084664/c619c03e-3b6d-11e5-8d83-1d9c19984865.png)
But even after fixing it, the same problem occurs.

Calanthe:
This is definitely google ima's issue https://github.com/googleads/videojs-ima/issues/116. Let's wait for their fix.

JonNorman:
Looks the guy who went to fix this ended up not needing it and so abandoned it. Fortunately this was picked up by someone else, so when [this PR](https://github.com/googleads/videojs-ima/pull/215/files) gets accepted, we can have another look.

ScottPainterGNM:
Looks like this is now in [this PR](https://github.com/googleads/videojs-ima/pull/255) instead, fingers crossed!

ScottPainterGNM:
That PR was now accepted so hopefully this might work

SiAdcock:
[...]

PR #11004, opened 2015-10-28
Multiple videos preroll disabled

Calanthe:
In order to temporary obey the bug with multiple pre-rolls (https://github.com/guardian/frontend/issues/10034) we disabled video pre-rolls at all everywhere, where there is more than one video.

This needs to be tested especially on fronts and article pages.

CC @kenlim @rich-nguyen

### DIFF
--- a/static/src/javascripts/bootstraps/media.js
+++ b/static/src/javascripts/bootstraps/media.js
@@ -230,6 +230,8 @@ define([
 
             player.persistvolume({namespace: 'gu.vjs'});
 
+            console.log('player');
+
             deferToAnalytics(function () {
 
                 events.initOmnitureTracking(player);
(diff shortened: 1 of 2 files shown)


static/src/javascripts/bootstraps/media.js
```diff
@@ -240,7 +240,7 @@ define([
 
                     player.fullscreener();
 
-                    if (commercialFeatures.videoPreRolls && !blockVideoAds) {
+                    if (commercialFeatures.videoPreRolls && !blockVideoAds && !config.page.hasMultipleVideosInPage) {
```
jbreckmckye: It might be worth adding a comment describing the pitfalls of running multiple prerolls on a page, and how a future developer can validate whether the problem is still an issue.

rich-nguyen:
Looks fine if prerolls should be disabled for body content containing multiple videos. Did we ever understand why this isn't affecting containers of videos? If it does cause problems within containers, almost every video page will be an issue, because of the videos containers below the body.

Calanthe:
To be honest, no. We've been investigating that with @kenlim and video pre-rolls seem to be working fine on video pages. The bug affects only articles with more than one video in it.

michaelwmcnamara:
+1 - change prevents video pre-roll ads from running when more than 1 video on the page - this stops the bug from occuring. Pages with 1 video and video pages still get ads. 
A good stop gap while we work on getting a full fix together.

PR #11007, opened 2015-10-29
Extend the switch until Grant is back.

jennysivapalan:
Not sure of the consequence of this switch turning off so extending it until @gklopper  is back.
:ghost: :jack_o_lantern:


johnduffell:
actually it's already off, I don't think grant had time to do anything with it.  It worked quite well but takedowns weren't showing up in the updated list so they were staying in the cache.
Not sure if anyone's got time to look at fixing it up as it'd be a good thing to improve our hit ratio.

jennysivapalan:
Thanks will check in with Grant when he is back and work out the next steps.

PR #11008, opened 2015-10-29
Remove adfree cookies on signout

jbreckmckye:
As part of the work for ad-free web, we're storing the user's membership features in the form of cookies. These values only make sense attached to a particular user, and we scrub them with JavaScript when the user is signed out. However, because that script runs later on in the page, it's not really something we should probably rely on.

This pull request makes us nix the cookies when the user hits the signout controller.


johnduffell:
good idea :+1:

PR #11009, opened 2015-10-29
curl: load source assets in dev to avoid compiling

OliverJAsh:
This change means we can have a "refresh to reload" workflow using our current module loader, just like what we had with jspm. Still stuck on ES5 for now, but one thing at a time.


sndrs:
magic :gem: :ghost: :+1:

Gruntfile.js
```diff
@@ -99,7 +99,11 @@ module.exports = function (grunt) {
     grunt.registerTask('compile:js', function (fullCompile) {
         grunt.task.run(['clean:js', 'compile:inlineSvgs']);
 
-        grunt.task.run(['concurrent:requireJS', 'copy:javascript']);
+        if (!options.isDev) {
```
rich-nguyen: i think if a dev is running this task, then it should do the `concurrent:requireJS` task. Why do we need the guard? is it for a watch?

sndrs: it's just not used at all. we can still build them, but you'll never load them in dev

grunt-configs/copy.js
```diff
@@ -31,11 +31,12 @@ module.exports = function (grunt, options) {
                         '**/*'
                     ],
                     dest: options.staticHashDir + 'javascripts/vendor'
-                },
-                {
+                },{
                     expand: true,
-                    cwd: options.requirejsDir,
-                    src: [
+                    cwd: options.isDev ? options.staticSrcDir + 'javascripts' : options.requirejsDir,
```
rich-nguyen: Why is a verbatim copy needed from `static/src` to `static/target`? Aren't we loading directly from `static/src`?

sndrs: `AssetMap` was expecting them in the map. this puts them in place so asset_hash includes them

Gruntfile.js
rich-nguyen: that's fine. it makes more sense to compile the js if you've run this task.

grunt-configs/copy.js
rich-nguyen: We shouldn't be using `AssetMap`, because we don't need to resolve paths to hashed paths.

Gruntfile.js
sndrs: i'm not sure why? if you still compile the JS in `--dev`, you'll have to wait for it to compile but never download it. not sure what the benefit is? it was the same with JSPM, we never bundled in dev. same with uglifying etc?

sndrs: hmmm, do you mean because you've run `compile:js`? i see your point...

OliverJAsh:
Thanks for the changes @rich-nguyen and @sndrs :+1:

sndrs:
:+1: from me. love this.

PR #11010, opened 2015-10-29
Soulmates component layout change

tijanamiletic:
Swapping around the search box and the member list.

Before:
screen shot 2015-10-29 at 13 09 45

After:
screen shot 2015-10-29 at 13 09 57


Calanthe:
:+1:

PR #11014, opened 2015-10-29
Improve code element style

OliverJAsh:
## Before

image
## After

image

/cc @sammorrisdesign

### DIFF
--- a/static/src/stylesheets/module/external/_from-content-api.scss
+++ b/static/src/stylesheets/module/external/_from-content-api.scss
@@ -116,6 +116,10 @@
 .from-content-api pre {
     // Unset from `.from-content-api`
     word-wrap: normal;
+    background-color: #f6f9fb;
+    border: 1px solid #d3e1ea;
+    padding: $gs-baseline*.75 $gs-gutter*.75;
+    border-radius: 4px;
 }
 
 
--- a/static/src/stylesheets/module/external/_from-content-api.scss
+++ b/static/src/stylesheets/module/external/_from-content-api.scss
@@ -116,8 +116,9 @@
 .from-content-api pre {
     // Unset from `.from-content-api`
     word-wrap: normal;
-    background-color: #f6f9fb;
-    border: 1px solid #d3e1ea;
+    $x: lighten(colour(guardian-brand-light), 30%);
+    background-color: $x;
+    border: 1px solid darken($x, 10%);
     padding: $gs-baseline*.75 $gs-gutter*.75;
     border-radius: 4px;
 }


static/src/stylesheets/module/external/_from-content-api.scss
```diff
@@ -116,6 +116,10 @@
 .from-content-api pre {
     // Unset from `.from-content-api`
     word-wrap: normal;
+    background-color: #f6f9fb;
```
OliverJAsh: No colours for this in http://guardian.github.io/pasteup-palette/demo/

sammorrisdesign: Could maybe achieve them using existing colours and `lighten()` sass colour function.

OliverJAsh: See 59cd227cef25faacd644c3bf8cdee8baae887ebb

OliverJAsh:
After updating colours to use palette:
image

static/src/stylesheets/module/external/_from-content-api.scss
```diff
@@ -116,6 +116,11 @@
 .from-content-api pre {
```
piuccio: the classname is quite misleading. What else does this change affect?

paperboyo:
This here journalist is a programmer! :scream_cat:

static/src/stylesheets/module/external/_from-content-api.scss
OliverJAsh: Not sure I follow. Just ` .from-content-api pre`. pre elements are for pre-formatted text. Code elements are the only usage of those that I know of.

static/src/stylesheets/module/external/_from-content-api.scss
```diff
@@ -116,6 +116,11 @@
...
     // Unset from `.from-content-api`
     word-wrap: normal;
+    $x: lighten(colour(guardian-brand-light), 30%);
```
sndrs: `$x` :eyes: ?

sammorrisdesign:
:+1:

static/src/stylesheets/module/external/_from-content-api.scss
johnduffell: have you chatted with the composer guys about this?  I think they would have an opinion - on first glance it does seem a bit broad, but if pre always wants to look like code then it's fine

static/src/stylesheets/module/external/_from-content-api.scss
OliverJAsh: I'm not worried about it because the scope is small and the use is clear, but I can give it a better name if you like?

static/src/stylesheets/module/external/_from-content-api.scss
OliverJAsh: Sorry @johnduffell, saw this comment after merge. I'm talking to James now to find out.

OliverJAsh: They don't know if it's used anywhere else. :disappointed: `pre` already had monospace font styling—I'm only adding a background, border, and padding here, so it shouldn't hurt anything.

johnduffell: fair enough I'm pretty sure the benefit outweighs the risk :+1:

PR #11016, opened 2015-10-29
add the referrer to amp tracking

johnduffell:
This doesn't actually work yet because google haven't shipped the referrer feature, but once they do we should get the referrer in omniture for amp.  We will ship once it's released by google.

The feature is this: https://github.com/ampproject/amphtml/blob/master/builtins/amp-pixel.md
and scroll down to $DOCUMENT_REFERRER
@stephanfowler

### DIFF
--- a/common/app/views/mainAMP.scala.html
+++ b/common/app/views/mainAMP.scala.html
@@ -21,7 +21,7 @@
     <body>
         @defining(s"${request.host}${request.path}") { path =>
             @defining({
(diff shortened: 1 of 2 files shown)


rich-nguyen:
:+1:

johnduffell:
This doesn't work yet because of https://github.com/ampproject/amphtml/issues/832 so I will wait for a response and close this if it doesn't look like a quick resolution

johnduffell:
this has changed to not need the dollar so reopening

PR #11018, opened 2015-10-29
Style the Chapter Headings

sammorrisdesign-zz:
I've also increased the time on the switch as it's not ready to go live while Composer finishes their work on the feature. Cool.

It's a toggle until Desktop
screen shot 2015-10-29 at 17 16 03

Columnised on tablets
screen shot 2015-10-29 at 17 15 17

Right hand Columned on Desktop
screen shot 2015-10-29 at 17 15 09


Jholder112233:
How's this play with the most popular container and the RH ad?

sammorrisdesign:
@Jholder112233 Must have taken the screenshots after resizing from mobile. It goes in above those on desktop.

common/app/views/fragments/chapterHeadings.scala.html
```diff
@@ -4,10 +4,20 @@
 
 @if(ChapterHeadingsSwitch.isSwitchedOn && content.chapterHeadings.size > 2) {
     <div class="chapters">
-        <span class="chapters__heading">Contents</span>
-        <ol>
+        <h3 class="chapters__heading">Chapter guide</h3>
+        <!-- THIS NEEDS TO BE FINISHED -->
+        <button class="chapters__button js-toggle-chapters meta-button hide-on-desktop" data-link-name="weather-toggle-chapters" href="#toggle-chapters">
+            <span class="inline-arrow inline-icon chapters__toggle-icon">
+                <svg width="24" height="10" viewBox="0 0 24 10"><path d="M0 .7l10.5 8.4 1.1.9 1.1-.9L23.2.7l-.5-.7-11.1 6.3L.6 0 0 .7z"></path></svg>
```
Jholder112233: This SVG should be included using the standard inline SVG helper

common/app/views/fragments/chapterHeadings.scala.html
```diff
@@ -4,10 +4,20 @@
 
 @if(ChapterHeadingsSwitch.isSwitchedOn && content.chapterHeadings.size > 2) {
     <div class="chapters">
-        <span class="chapters__heading">Contents</span>
-        <ol>
+        <h3 class="chapters__heading">Chapter guide</h3>
+        <!-- THIS NEEDS TO BE FINISHED -->
+        <button class="chapters__button js-toggle-chapters meta-button hide-on-desktop" data-link-name="weather-toggle-chapters" href="#toggle-chapters">
```
Jholder112233: This arrow is still going to be visible even on core when the JavaScript isn't running. You can use the "modern-visible" class to make it only show on modern browsers (not in core).

static/src/javascripts/projects/common/modules/article/chapters.js
```diff
@@ -0,0 +1,24 @@
...
+    $
+) {
+    return {
+        init: function () {
+            this.bindEvents();
+        },
+
+        bindEvents: function () {
+            bean.on(document.body, 'click', '.js-toggle-chapters', function (e) {
+                e.preventDefault();
```
Jholder112233: Do we need to preventDefault here? Can't we just remove the href from the button?

sammorrisdesign: Here I was following what we did for the weather toggle. Happy to remove. Maybe @uplne can shed light on it.

uplne: Because I did it in the past it doesn't mean its good solution :-) It depends on the context of the link. Do we need it to behave like a link without JS (should it take user somewhere without JS?) or should it behave like a button. Maybe the use of button element will be more semantic if it should behave as button.

static/src/stylesheets/module/_chapters.scss
```diff
@@ -1,19 +1,85 @@
...
+    }
+
+    .content__main-column & {
+        @include mq(desktop) {
+            display: none;
+        }
 
-        @include mq(tablet) {
-            column-count: 2;
+        .is-not-modern & {
```
Jholder112233: Didn't you just want to scope this to mobile? In this current form, you could use the modern-visible helper class and not have to repeat the css.

sammorrisdesign: I currently add the chapters twice, so adding `modern-visible` to the template would hide both. Whereas I only want to hide it on mobile because of the toggle.

Jholder112233: I was imagining you'd just add modern-visible to  <button class="chapters__button js-toggle-chapters meta-button hide-on-desktop" data-link-name="weather-toggle-chapters">

as it functionally it works without JavaScript apart from the toggle?

sammorrisdesign: That's an option but the chapters are collapsed by default on mobile and seeing as they could theoretically get quite big, I'd rather just not give the option to them.

Jholder112233:
:+1:

common/app/views/fragments/chapterHeadings.scala.html
```diff
@@ -4,10 +4,17 @@
 
 @if(ChapterHeadingsSwitch.isSwitchedOn && content.chapterHeadings.size > 2) {
     <div class="chapters">
-        <span class="chapters__heading">Contents</span>
-        <ol>
+        <h3 class="chapters__heading">Chapter guide</h3>
+        <button class="chapters__button js-toggle-chapters meta-button hide-on-desktop" data-link-name="weather-toggle-chapters">
```
Jholder112233: Update this data-link-name before merge!

PR #11019, opened 2015-10-29
Rename admin link to give more context

jennysivapalan:
I recently learnt about the trouble shooter page on Admin. It's pretty useful for checking the existence of a path. I think renaming the link on the index page might help others find and use it too.


rich-nguyen:
:+1:

PR #11020, opened 2015-10-30
Omniture track external links

rich-nguyen:
We would like to track external links.

https://marketing.adobe.com/resources/help/en_US/sc/implement/trackExternalLinks.html


gtrufitt:
:+1:

PR #11022, opened 2015-10-30
Move to `make`

sndrs:
This is a proposal to elevate `make` to become the default way to manage the assets from the command line, retiring `grunt` (though not behind the scenes). 

It should cover all the day-to-day things for which we currently use `grunt`, and improve upon some of them (this is useful, for example, in the instance `grunt install` doesn't work because you haven't `npm install`ed `grunt` already; it also means no API changes to run the site whatever we use in the future).
- replaces `grunt install` and `grunt compile`
  screen shot 2015-10-29 at 16 59 51
- adds a `make watch` command that fires up a browser-sync instance and live-reloads `*.js`/`*.scss` changes (to take advantage of #11009)
  screen shot 2015-10-29 at 17 10 09
- adds a `make reinstall` to nuke 3rd party dependencies and reinstall them in one go
- adds a `make shrinkwrap` to easily lock down NPM packages
  screen shot 2015-10-29 at 18 56 43
- displays an options menu using just `make` or `make help`
  screen shot 2015-10-29 at 19 05 08

thoughts?


Gruntfile.js
```diff
@@ -57,9 +57,9 @@ module.exports = function (grunt) {
         grunt.log.subhead('Running Grunt in DEV mode');
     }
 
-    // Default task
+    // Default task - used by grunt-tc
```
OliverJAsh: Could TeamCity use make instead of grunt too? Not sure if that makes sense

dev/bs-config.js
```diff
@@ -27,6 +27,7 @@ module.exports = {
         "cwd": "../"
     },
     "files": [
+        "static/src/javascripts/*.js",
         "static/src/javascripts/**/*.js",
```
OliverJAsh: Won't this glob match everything caught by the glob you added?

Gruntfile.js
sndrs: yeah potentially. i wanted to limit this to (hopefully) non-breaking stuff. assume `make` is available there etc, but was't sure. happy to change if anyone (@rich-nguyen?) can confirm it should be ok?

dev/bs-config.js
sndrs: you'd think so! but it wasn't picking up top level (`core.js` etc) file changes without it...

makefile
```diff
@@ -1,27 +1,56 @@
...
-	grunt test --dev
+	@grunt test --dev
 
 validate:
-	grunt validate
+	@grunt validate
+
+shrinkwrap:
+	@npm shrinkwrap && node dev/clean-shrinkwrap.js
+	@node dev/message.js shrinkwrap
```
OliverJAsh: Am I reading this wrong? When we run `make shrinkwrap`, we shrinkwrap and then show a warning message?

dev/bs-config.js
OliverJAsh: Oh, must be a bug!

makefile
sndrs: should just be a confirmation message?

OliverJAsh:
I especially like `make help` :smile: Nice!

sndrs:
Made the `grunt compile` message a bit more useful:

screen shot 2015-10-30 at 11 26 23

git-hooks/message.js
```diff
@@ -5,10 +5,10 @@ var megalog = require('megalog');
  */
 switch (process.argv[2]) {
   case 'update':
-    megalog.info('Dependencies have been updated.\n\nRun `grunt install`.');
+    megalog.info('Dependencies have been updated.\n\nRun `make install`.');
     break;
 
   case 'shrinkwrap':
-    megalog.error('This commit changes `package.json`, but doesn\'t include `npm-shrinkwrap.json`.\n\nIf you\'re updating npm packages, please run `npm shrinkwrap && node dev/clean-shrinkwrap.js` to write the new dependencies to `npm-shrinkwrap.json` and add that file before committing.');
+    megalog.error('You have changes to `package.json` but havn’t updated `npm-shrinkwrap.json`.\n\nPlease run `make shrinkwrap` and commit the changes to `npm-shrinkwrap.json`.');
```
gtrufitt: Minor spelling on haven't

makefile
```diff
@@ -1,27 +1,56 @@
...
+uninstall: uninstall-application uninstall-dev
+	@echo 'All 3rd party dependencies have been uninstalled.'
 
 test:
-	grunt test --dev
+	@grunt test --dev
 
 validate:
-	grunt validate
+	@grunt validate
```
gtrufitt: Would we want to be able to validate css and js separately via make like running ``grunt validate:css``? Same with compile? Not sure if it's needed at the higher level (or whether the watcher on compile compiles specific files?)

rich-nguyen:
I'm happy to use `make`, `grunt`, `gulp`, etc..., whatever the wider team chooses they are most comfortable with maintaining. Be sure to communicate the rationale, benefits and the workflow to everyone who needs to be informed, especially the browsersync stuff.

I agree with @OliverJAsh, team city should build things a similar way to how we do it locally. It may seem less risky to skip this from a Pull Request perspective, but it's more risky in the long run to support several different ways of building.

Also if `make` is the principal build system, can the `dev/package.json` be moved into the top level `package.json` under `devDependencies`, and move all of  `dev/makefile` into it's parent `makefile`?

:+1:

makefile
sndrs: yeah good thinking. not sure why github hasn't picked this up, but change for this is here: https://github.com/guardian/frontend/commit/46cd127f950bc21fd1e120bb540fbe563f05553e

gtrufitt: Aces 
yiss

PR #11023, opened 2015-10-30
Large top ad AB test

uplne:
To test user behaviour when large top ad format is active.
1. Only US edition
2. 0.5% of the users won't see the ad format
3. other 0.5% is just a control group

@kenlim @Calanthe


Calanthe:
:+1:

PR #11026, opened 2015-10-30
AB test: inject network front instead of most popular on content pages

holder-josh:
This test injects the entire network front underneath an article in the 'most popular' slot. The network front represents the breadth of content that the Guardian offers, so why not also display that on article pages?

frontatbottom

Sample size needs updating still.


NataliaLKB:
Looking forward to seeing the results from this test! :+1:

PR #11028, opened 2015-10-30
Use 'the Guardian', not  'The Guardian'

rtyley:
Jenny Heeley & Lauren Babcock pointed out this should be lower case.

The [Guardian Style Guide](http://www.theguardian.com/guardian-observer-style-guide-t) says of 'the' : `lc for newspapers (the Guardian)`

cc @chrisjowen 

```
---------- Forwarded message ----------
From: Lauren Babcock
Date: 29 October 2015 at 16:52
Subject: Fwd: Screengrab
To: Nick M Miller

Hello there - do you know who builds these? Should be a lower case t. Thanks so much
```


rich-nguyen:
:+1:

PR #11029, opened 2015-10-30
Only consider paths as matching if the destination ends with the r1 path

rich-nguyen:
This strengthens the conditions needed to consider a path as one which 'links to itself'. This should only be true if the candidate redirect path string ends with the input string (rather than contains).


jennysivapalan:
:four_leaf_clover: :+1:

PR #11030, opened 2015-10-30
Rename X-GU-ID-REMOTE-IP to X-Forwarded-For

niklasvincent:
We don't check `X-GU-ID-REMOTE-IP` on our side, but having `X-Forwarded-For`[1] would mean it gets automatically picked up on our side. Right now this information is not processed.
1. http://docs.aws.amazon.com/ElasticLoadBalancing/latest/DeveloperGuide/x-forwarded-headers.html#x-forwarded-for


danmassie:
:+1:

PR #11031, opened 2015-11-02
Separate r1 path check from general redirect regex pattern

rich-nguyen:
This should fix problems with redirects that were being checked for loops and then 404ing


jennysivapalan:
:+1: :weary:

rich-nguyen:
:santa:

PR #11032, opened 2015-11-02
Allow for embeds in MainMedia on the Immersive Template

sammorrisdesign-zz:
One of the main reasons this template exists. We now can smoothly transition in embeds when they're added as MainMedia elements. Example gif....

immersiveembed

Smooth.


sndrs:
lush :+1:

PR #11037, opened 2015-11-02
Provide dev mode for video player

rich-nguyen:
Updates the dev config to load video modules. The plugins are valid require modules, but they do not load in curl because they `define` with an ID. I think the simplest solution is to provide a dev mode video player.


rich-nguyen:
@Calanthe

rich-nguyen:
Hope this helps, @Calanthe

Calanthe:
Thanks! :+1:

PR #11038, opened 2015-11-02
Allow slow mpu to boost 2nd item 

sndrs:
before:

screen shot 2015-11-02 at 15 52 00

after:

screen shot 2015-11-02 at 15 51 04

@jennysivapalan `lift(2)` didn't work, but `tail.headOption` does. you think it's ok?


rich-nguyen:
:+1:

PR #11040, opened 2015-11-02
Rename 'dfp.js' to something adblock-friendly

jbreckmckye:
This pull request fixes an issue on dev, where AdblockPlus could prevent application bootstraps from running.

Unlike in production, where dfp.js is pulled in as part of the commercial.js bundle, on dev, dfp.js is pulled in as a separate file with Curl. The problem, though, is that the filename 'dfp' triggers an adblock rule, and the request gets bounced by the browser. Curl can't handle this, it seems, and lots of modules fail to load - including some non-commercial ones.

Although this only affects dev, it's still a problem because it's helpful to see how new features render under an adblocker, and we often write features or code paths specifically for these users.

As such, I have renamed 'dfp.js' to 'dfp-api.js', and added a note explaining the sensitivity of the filename.

I'm happy to consider other filenames; this just seemed like the most conservative change that would work. I've tested it under AdblockPlus, and checked there's no similar entries in the EasyList open source ad content blacklist.


sndrs:
nice :+1:

PR #11041, opened 2015-11-02
Fix Slices at Mobile Breakpoint

kelvin-chappell:
This fixes a layout issue on mobile.
I don't think I've broken anything anywhere else with this fix.

Before:

image

After:

image

/cc @jbreckmckye @stephanfowler


sndrs:
:+1:

Issue #10947, opened 2015-10-22
Showcased images in the new Immersive template are being requested at too low a resolution


paperboyo:
Hello,

In the new Immersive template the showcased in-article images are requested at smaller resolution that they are shown at (at 100% zoom). This results in unnecessarily blurry images. As far as I can tell, this affects all breakpoints (they are requested 20px smaller in width, maybe because Immersive template has a wider text column?).

Related to (all fixed): https://github.com/guardian/frontend/issues/9482, https://github.com/guardian/frontend/issues/9278, https://github.com/guardian/frontend/issues/10467, https://github.com/guardian/frontend/issues/10493.

Regards
Mateusz

sammorrisdesign:
This should be fixed in https://github.com/guardian/frontend/pull/11044

paperboyo:
Thanks, @sammorrisdesign!

PR #11044, opened 2015-11-02
Add new Immersive figure weighting

sammorrisdesign-zz:
It's time to go BIG with a new image weighting. This is scoped to the new `immersive` template. The name suggests this and going bigger than `showcase` on regular articles would be strange and break stuff.

Mobile (flush to edge of device)
screen shot 2015-11-02 at 17 45 47

Desktop (expands into RHC)
screen shot 2015-11-02 at 17 45 09

### DIFF
--- a/common/app/views/support/HtmlCleaner.scala
+++ b/common/app/views/support/HtmlCleaner.scala
@@ -103,7 +103,7 @@ case class PictureCleaner(article: Article, amp: Boolean)(implicit request: Requ
       container <- findContainerFromId(figure.attr("data-media-id"), image.attr("src"))
       image <- container.largestImage
     }{
-      val hinting = findBreakpointWidths(figure)
+      val hinting = findBreakpointWidths(figure, article.isImmersive)
       val relation = if (article.isLiveBlog) { LiveBlogMedia } else { BodyMedia }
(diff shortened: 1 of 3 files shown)


common/app/views/support/HtmlCleaner.scala
```diff
@@ -103,7 +103,7 @@ case class PictureCleaner(article: Article, amp: Boolean)(implicit request: Requ
       container <- findContainerFromId(figure.attr("data-media-id"), image.attr("src"))
       image <- container.largestImage
     }{
-      val hinting = findBreakpointWidths(figure)
+      val hinting = findBreakpointWidths(figure, article.isImmersive)
```
rich-nguyen: Hmmn, this is nearly consistent, but look below at `LiveBlogMedia`. We choose the `ContentRelation` object based on the content type. So if we had a `ImmersiveMedia`, alongside `LiveBlogMedia` and `BodyMedia`, then there would be no `isImmersive` flags needed. It's also more scalable, when we inevitably introduce `SuperDuperImmersiveMedia`

rich-nguyen:
:+1: a tweak to the object structure as commented would be great

paperboyo:
> `SuperDuperImmersiveMedia`

:laughing:

Also, should magically fix https://github.com/guardian/frontend/issues/10947, right?

PR #11046, opened 2015-11-03
Create requestIdleCallback preference for throttleScroll and ad-loading

sndrs:
Allows you to opt-in to a version of the site that delays `throttledScroll` and all commercial dom manipulation until an `idleCallback` is available (if supported).

This is to test the effect of `requestIdleCallback` on the FPS/interaction hit we have currently at load.

cc @uplne @Calanthe @jbreckmckye @paullewis

Opt-in would be available via http://www.theguardian.com/uk#gu.prefs.use-idle-callback=true


static/src/javascripts/bootstraps/common.js
```diff
@@ -238,18 +238,19 @@ define([
             // However, this means it's VITAL that all writes in callbacks are delegated to fastdom
             throttledScrollEvent: function () {
                 var running = false;
-                window.addEventListener('scroll', function () {
+                var emit = function () {
                     if (!running) {
                         running = true;
-                        // Use fastdom to avoid forced synchronous layout:
```
jbreckmckye: Isn't this comment still true if we're not using RIC? (Look, I already created a cool new acronym for it)

sndrs: yeah it just seemed unnecessary

static/src/javascripts/projects/common/modules/commercial/article-body-adverts.js
```diff
@@ -75,7 +75,7 @@ define([
 
                 ads.push($ad);
                 return new Promise(function (resolve) {
-                    fastdom.write(function () {
+                    idleFastdom.write(function () {
```
jbreckmckye: Just a heads up - I've actually been looking at this very write today. It seems the way we're finding spaces and writing elements asynchronously can create some race conditions. That means we might have to either do some reads and writes synchronously, or do some careful scheduling that prevents ads being injected into spaces that are no longer empty. 

If you're doing some performance investigation, though, I'm wary of doing anything that might add another variable to your testing.

jbreckmckye:
Very exciting stuff.

:+1:

PR #11047, opened 2015-11-03
Point the azed crosswords nav at their new landing page

JustinPinner:
We now have a new location for azed crosswords. Update the navigation to use the new target.


jbreckmckye:
:+1:

PR #11048, opened 2015-11-03
Fixed badge position

uplne:
Before:
screen shot 2015-11-03 at 15 37 56

After:
screen shot 2015-11-03 at 15 38 45


Calanthe:
The `commercial-dfp-multi` class is used in `/commercial/app/views/contentapi/items.scala.html`. Are you sure it is not needed?

uplne:
@Calanthe I've just checked it and this fix is exactly for component served by that template so I think it is safe

Calanthe:
ok :+1:

PR #11049, opened 2015-11-03
Change order of loading indicator and embed in immersive templates

sammorrisdesign-zz:
For some reason `z-index: -1` was doing what I wanted too locally but `PROD` decided it didn't fancy any of it. So I'm just doing this instead. Good old natural z-index.


johnduffell:
:+1:

PR #11050, opened 2015-11-03
Get's rid of path pattern matcher for redirects

jennysivapalan:
The path pattern matcher has stopped /masterclasses redirects /guardian-masterclasses.

cc @rich-nguyen


JustinPinner:
This makes sense from the point of view of the described problem, so :+1: from me

rich-nguyen:
:+1:

PR #11051, opened 2015-11-03
Adblock banner fix for paying members

Calanthe:
From now on, the adblock banner (together with the adblock messsage) will be shown only to non members who also use adblockers.

screen shot 2015-11-03 at 17 30 31

I also had to rename the `large-top-ad` to the `large-top-slot` file because adblock plus seems to block that file on code.


static/src/javascripts/projects/common/modules/commercial/commercial-feature-policies.js
```diff
@@ -67,6 +69,31 @@ define([
         }
     };
 
+    policies.showAdblockMessages = function () {
```
jbreckmckye: I am not sure I am entirely comfortable with this.

Firstly, the idea is that these policies each return a map of flags that refer to specific _features_. They are named 'feature policies', after all. But neither _showAdblockMessages_ or _memberWithAdblock_ are what I'd call features, so this doesn't really seem consistent with the way the rest of these work.

Secondly, 'policies' have so far only referred to scenarios, rather than features. The idea has been that rather than a policy like 'videoPreRolls' full of switches, you'd have policies named 'sensitiveContent' and 'adFree' that establish what feature changes those abstract concepts comprise. A policy that just turns on or off a feature is really just a transplanted if statement. I know there is a 'membershipMessages' policy that does just that above, but we didn't add it, and, I'm not 100% sure about that one, either.

Thirdly, there's a constructor for 'commercialFeatureSwitches' that mostly comprises all the switches available, but not these. This makes sense - if we put these switches on that constructor, then the sensitive content policy would turn them off - but it does mean the original class kinda no longer makes sense, and is actually a little misleading (because it looks like definition of the aggregated map of switches eventually returned), and I think we need to replace it.

There are some good questions about how exactly this policy file should work and what's the most scalable and expressive way to map domain concepts to these flags. I kinda expected us to have to rejig things around as we discovered more requirements. But I'd really like it if we can start thinking about those problems before this file becomes too overgrown.

static/src/javascripts/projects/common/modules/commercial/commercial-feature-policies.js
```diff
@@ -3,6 +3,7 @@ define([
     'common/utils/location',
```
jbreckmckye: We should ideally marry any changes in this file with some tests in commercial-feature-policies.spec.js, too.

static/src/javascripts/projects/common/modules/commercial/commercial-feature-policies.js
```diff
@@ -3,6 +3,7 @@ define([
...
     'common/utils/config',
     'common/utils/detect',
+    'common/utils/storage',
```
jbreckmckye: Should we unit test these changes?

static/src/javascripts/projects/common/modules/commercial/commercial-feature-policies.js
```diff
@@ -10,6 +11,7 @@ define([
     location,
     config,
     detect,
+    storage,
```
jbreckmckye: This dependency is redundant.

static/src/javascripts/projects/common/modules/commercial/adblock-rules.js
```diff
@@ -0,0 +1,30 @@
...
+        return sharedRules() && userFeatures.isPayingMember();
+    }
+
+    function isNonMemberWithAdblock() {
+        return sharedRules() && !userFeatures.isPayingMember();
+    }
+
+    return {
+        isMemberWithAdblock: isMemberWithAdblock,
+        isNonMemberWithAdblock: isNonMemberWithAdblock
```
jbreckmckye: I don't really think membership is a responsibility of this file. Why not just check the membership status in donot-use-adblock.js?

static/src/javascripts/projects/common/modules/commercial/adblock-rules.js
```diff
@@ -0,0 +1,30 @@
+define([
```
jbreckmckye: Perhaps this should just concern whether the message is displayed? It could just return a function,

    hasAdblockMessages() {Boolean}

You could call that in donot-use-adblock and then your test in sticky.js would be -

    if (hasAdblockMessages()) {
        // do DOM manipulation...
    }


    getUpdateMethod = function() {
        if (detect.hasAdblock() && !hasAdblockMessages(){ {
              // this code doesn't actually care if someone is a member. It just cares whether something
              // is going to replace the upper banner.
        }
    }

jbreckmckye:
Groovy :+1:

michaelwmcnamara:
QA +1 this looks good!

PR #11052, opened 2015-11-04
Extend Fixed Top-Above-Nav Switch

kelvin-chappell:
This switch is either going to be expanded to be used on all fronts or it will be removed altogether, depending on future of top slot.  But need more time to test and decide.


jbreckmckye:
:+1:

PR #11056, opened 2015-11-04
Don't display the supporter banner to paid members

joelochlann:
The supporter banner is this one:

picture 37

We don't want to display it to people who are already paid members. This work has [already been done for the ad blocker banner](https://github.com/guardian/frontend/pull/10945) so I'm using the `userFeatures.isPayingMember()` function added in that PR.

@Calanthe @jbreckmckye


Calanthe:
:+1: It looks ok. If you have any problems with testing locally, we can do it together on my machine.

PR #11057, opened 2015-11-04
Design the crap out of the pullquotes

sammorrisdesign-zz:
We've got (are getting) weightings for pull quotes to make use of in the `immersive` template. This makes use of them. Quotable.

Everything until tablet and the `inline` weighting
screen shot 2015-11-04 at 17 03 11

`supporting`, this is the default
screen shot 2015-11-04 at 17 03 27

`showcase`
screen shot 2015-11-04 at 17 03 02


johnduffell:
:+1: pink's not my colour though :stuck_out_tongue:

PR #11058, opened 2015-11-04
Rugby for life

jennysivapalan:
We're allowed to keep the rugby world cup data forever! 
- Remove table and match switches. All rugby data is now controlled by just one switch `RugbyScoresSwitch`
- Removes friendlies for pre-prod
- Removes live score / match fetching as the competition is complete and reduces the poll time of the remaining feeds. If we get this data for future competitions we can just revert https://github.com/guardian/frontend/commit/86c06cad1b5011b6b83cc8401e2c5953f85a7389
- Updates the nav bar as requested by Katie. 

:rugby_football: :trophy: :earth_asia: :new: :bird:


common/app/conf/switches/FeatureSwitches.scala
```diff
@@ -332,6 +296,15 @@ trait FeatureSwitches {
     exposeClientSide = false
   )
 
+  val RugbyScoresSwitch = Switch(
+    "Feature",
+    "rugby-world-cup",
+    "If this switch is on rugby world cup scores will be loaded in to rugby match reports and liveblogs",
+    safeState = Off,
+    sellByDate = never,
+    exposeClientSide = true
```
kelvin-chappell: Is this being used on the clientside?

kelvin-chappell:
:+1:

common/app/conf/switches/FeatureSwitches.scala
jennysivapalan: Good spot - I moved it to be checked in the Scala code so will update this.

PR #11061, opened 2015-11-05
Remove switches from user-features.js

jbreckmckye:
We used to only perform requests for the membership API if switches for the adfree view and adblock messages were on. However, `membership-messages.js` now relies upon this data, and it doesn't have a switch. So the membership requests should always be firing.

The code removed is actually something of a timebomb, because it would be quite easy to break the membership messages just by disabling the adblock-messaging switch (which is the only one on in prod right now). If we turned that off, clients would immediately start deleting their membership data.


Calanthe:
:+1:

PR #11064, opened 2015-11-05
Crosswords tracking on buttons

gtrufitt:
Add the data link name attribute to crossword buttons to ensure proper tracking in omniture.

Also fixes a but where the confirm all buttons weren't working.

### DIFF
--- a/static/src/javascripts/projects/common/modules/crosswords/controls.js
+++ b/static/src/javascripts/projects/common/modules/crosswords/controls.js
@@ -23,6 +23,7 @@ define([
                 className: buttonClassName + ' ' + buttonGenericClassName,
                 onClick: this.props.onClearAll,
                 key: 'clear',
+                'data-link-name': 'Crosswords | Clear all',
                 text: 'Clear all'
             }));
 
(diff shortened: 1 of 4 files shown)


crifmulholland:
The data link names are usually a combination of the containing div, and then the individual buttons. Rather than the full string being on the individual data link name. 

However, the end result in omniture is the same, so :+1: from me on the names!

static/src/javascripts/projects/common/modules/crosswords/controls.js
```diff
@@ -71,7 +77,8 @@ define([
                     'button', {
                         className: buttonClassName + ' ' + buttonCurrentClassName,
                         onClick: this.props.onToggleAnagramHelper,
-                        key: 'anagram'
+                        key: 'anagram',
+                        'data-link-name': 'Crosswords | Anagram'
```
desbo: might be clearer as 'Show Anagram Helper'

desbo:
+1 other than minor comment

gtrufitt:
@crifmulholland That makes more sense, cheers, have updated that for both crosswords wrapper and anagram helper.
@desbo Ta! Updated.

PR #11065, opened 2015-11-05
Redisplay membership banner

joelochlann:
Here's the banner I'm referring to:

picture 37

We want to redisplay it to people who have already closed it. But we don't want to display it twice to everyone in the future. So I've opted for changing the id of the message. This preserves the history of those who have already closed the original banner (and we use this to distinguish these cases for tracking purposes). Meanwhile those who close this "new" banner without having closed the original will not have it re-displayed to them on the next page view.

@Calanthe @jbreckmckye


joelochlann:
@rich-nguyen

jbreckmckye:
Seems a reasonable approach in the immediate term

:+1:

PR #11066, opened 2015-11-05
Migrate whole app to modular Lodash

OliverJAsh:
As we are about to split the JavaScript application into two parts (“universal.js” (core/critical) and “app.js”?), we need to consider where Lodash will belong in the dependency graph. We know we want to use Lodash in universal.js, but we're not prepared to pay the price for the whole library. Instead, we can use the modular version of Lodash, which will mean we can weigh up the cost of each Lodash module import in universal.js and restrict our usage to suit our needs (e.g. just `forEach`, `map`, and `reduce`). App.js can re-use any Lodash modules already existing in universal.js to avoid duplication. This can be done by excluding universal.js from app.js in the r.js build.

The first commit introduces `lodash-amd` and the last commit removes `lodash`. Because we commit our client-side dependencies for shrinkwrapping, the diff here is unreadable. The commits in-between are the good stuff, so skip the diff in this pull request and go to https://github.com/guardian/frontend/compare/e1e611d349e0a653462c7db46a33c4c5cdcda394...a6f46e221dff1066ba1ddca914cef1467a0421a8. Alternatively you can break down the PR by commit if you find that easier.

I made this change using jscodeshift (AST transformations). You can view the transformers at https://github.com/OliverJAsh/lodash-to-lodash-amd-codemods. However, the transformations didn't catch all cases, so I have applied fixes in subsequent commits.

To do:
- [x] Check tests pass locally
- [x] Check JS validates locally
- [x] Check tests pass and JS validates on TC
- [x] Investigate file size differences and where things get bundled. Is there any duplication? https://docs.google.com/spreadsheets/d/1x0mHTqXewvFoc5edaNogD1yr8Hk2rKcP2hqIvQoznBw/edit?usp=sharing
- [x] Tidy up chain helper

Non-blocker to dos:
- [x] Write lint task to prevent import of whole Lodash
- [x] Restore indentation removed by code transformation
- [x] Restore comments removed by code transformation


piuccio:
really interested in
> Investigate file size differences and where things get bundled
I reckon we should end up with smaller bundles

sndrs:
wondering if it would be better to batch up the changes in this? with 343 files changed, no one is going to be able to realistically check it...

also there are 7 todos attached, only 2 of which are completed at this point. would it be better to complete those first?

OliverJAsh:
If you look at the commits, I migrated different parts of the app in
stages. That should help! Unless you've got any other ideas of how I can do
it?

On Mon, 9 Nov 2015 at 09:02 Alex Sanders <notifications@github.com> wrote:

> wondering if it would be better to batch up the changes in this? with 343
> files changed, no one is going to be able to realistically check it...
>
> also there are 7 todos attached, only 2 of which are completed at this
> point. would it be better to complete those first?
>
> —
> <https://github.com/guardian/frontend/pull/11066#issuecomment-155000753>.
>

-- 

in England Number 908396

OliverJAsh:
Here is a comparison of file sizes: https://docs.google.com/spreadsheets/d/1x0mHTqXewvFoc5edaNogD1yr8Hk2rKcP2hqIvQoznBw/edit?usp=sharing

Analysis:

* Most of Lodash ends up bundled in core (because some Lodash modules are transitive dependencies of core, and we subtract core from other bundles), so the file size of core is very similar. Most bootstraps have seen a very small increase in file size, whereby they request individual Lodash modules which are not provided by core.

piuccio:
Interesting data. I don't have write permission so I've cloned the table to add a diff column. Turns out that 

`admin` increases by +2.23
`video-embed` by +1.69
`core` by 0.19

overall by `0.47`

I suppose `core` increases due to the overhead of defining each individual module, but the other two files are slightly suspicious. I'd check those two first to make sure they're generated correctly and the numbers are correct.

Horrible question: if the file size of `core` (eventually renamed to `universal`?) increases, isn't it better to keep the whole lodash?

OliverJAsh:
I'm going to investigate admin and video embed, but I know they don't exclude core, so it may be a saving due to the removal of all of Lodash from core (somehow).

We don't intend to increase file sizes of the JS as it stands. This change is to give us flexibility so we can split the app in two, including core, as described in the first post of this PR. Only then will we see the benefit of this change.

OliverJAsh:
The reason admin is smaller is because it no longer bundles the whole of Lodash. [Here](https://docs.google.com/spreadsheets/d/1x0mHTqXewvFoc5edaNogD1yr8Hk2rKcP2hqIvQoznBw/edit#gid=201140841) you can see the list of Lodash modules required by admin and core. core requires more Lodash modules, which is why it hasn't decreased much in size. admin doesn't exclude core—it just bundles everything it needs.

sndrs:
tl;dr but seems sensible. 

fingers crossed! :+1:

PR #11067, opened 2015-11-05
Remove the switch for EU cookie messages

JustinPinner:


jbreckmckye:
Déjà vu.

:+1:

JustinPinner:
Thanks @jbreckmckye - before I merge this, I wondered if @gtrufitt might have a look at the change I just made to the mediator.emit lines he added recently. Are those valid emissions to be sending?

gtrufitt:
Hi @JustinPinner - I listen to the emit here: https://github.com/guardian/frontend/blob/0b71e7c128d275b79117bbf7fe9e76b462ab9274/static/src/javascripts/projects/common/modules/experiments/tests/rtrt-email-message.js 

I don't think that will work anymore? Although eventEmitter looks like it takes a regex so you may be able to update those listeners to ``/^modules:ui:cookiesBanner:notShown/``?

PR #11068, opened 2015-11-05
A series of trivial changes to the main README

AaronCritchley:
A few questions:
- I noticed some inconsistencies in spacing after a sentence ends, is the preference to use one or two spaces? Let me know and I'll change them. 
- Code blocks for the most part are not using the laguage definition, one at the bottom is using the Bash 'lint', I think the majority of them should. Again, if I can get confirmation on The Guardians preference I will be happy to implement.

:smile:

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,11 @@
 We're hiring!
 --------
-Ever thought about joining us?<br/>
+Ever thought about joining us?
 http://developers.theguardian.com/join-the-team.html
 
 Frontend
 ========
-The Guardian website frontend.
+[The Guardian](www.theguardian.com) website frontend.
 
 Frontend is a set of Play Framework 2 Scala applications.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,11 +1,11 @@
 We're hiring!
 --------
-Ever thought about joining us?<br/>
```
jbreckmckye: The explicit <br/> is needed to force the line break. If we don't want a break, put it all on one line.

AaronCritchley: Are you sure? The preview still shows it as being broken line by line and I've never come across .md *needing* the explicit tag, if it's being used somewhere else and that's why it needs it, let me know and I'll add it in :+1:

jbreckmckye: > Are you sure?

Yep. Take a look in the browser.

jbreckmckye:
1. Browsers don't typically display redundant whitespace in CDATA. As they are displayed as one space, I would just use one likewise.
2. I'm quite sure if a cabal of top developers at El Grauniad have ever laid out a policy on such things, but I don't think it's necessary. It's mostly boilerplate and I don't really think the syntax highlighting is really going to help anyone just copypasting instructions into their command line. Let's keep the docs easy to edit.

Aside from the comment, this seems fine. :+1:

AaronCritchley:
Comment's been adressed, let me know if you have any other problems, I'll be happy to fix :+1:

piuccio:
nice :+1:

rich-nguyen:
thank you @AaronCritchley, much appreciated!

PR #11071, opened 2015-11-06
Unit tests for adblock messages

Calanthe:
While writing them I found out that the logic in `adblock-messages.js` module wasn't working properly for scenario with first time users and when adblock switch was off. In those cases, users shouldn't see adblock messages too.

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/adblock-messages.js
+++ b/static/src/javascripts/projects/common/modules/commercial/adblock-messages.js
@@ -16,7 +16,7 @@ define([
     }
 
     function noAdblockMsg() {
-        return sharedRules() && userFeatures.isPayingMember();
+        return (sharedRules() && userFeatures.isPayingMember()) || !sharedRules();
     }
 
     function showAdblockMsg() {
--- a/static/src/javascripts/projects/common/modules/commercial/adblock-messages.js
+++ b/static/src/javascripts/projects/common/modules/commercial/adblock-messages.js
@@ -9,18 +9,24 @@ define([
     storage,
     userFeatures
 ) {
-    function sharedRules() {
-        var alreadyVisted = storage.local.get('gu.alreadyVisited') || 0;
+    /*function sharedRules() {
+        var alreadyVisited = storage.local.get('gu.alreadyVisited') || 0;
 
-        return detect.getBreakpoint() !== 'mobile' && detect.adblockInUse() && config.switches.adblock && alreadyVisted > 1;
-    }
+        return detect.getBreakpoint() !== 'mobile' && detect.adblockInUse() && config.switches.adblock && alreadyVisited > 1;
+    }*/
 
     function noAdblockMsg() {
-        return (sharedRules() && userFeatures.isPayingMember()) || !sharedRules();
+        var alreadyVisited = storage.local.get('gu.alreadyVisited') || 0;
+
+        return detect.adblockInUse() && detect.getBreakpoint() !== 'mobile' && (
+                alreadyVisited <= 1 ||
+                !config.switches.adblock ||
+                (config.switches.adblock && alreadyVisited > 1 && userFeatures.isPayingMember()));
     }
 
     function showAdblockMsg() {
-        return sharedRules() && !userFeatures.isPayingMember();
+        var alreadyVisited = storage.local.get('gu.alreadyVisited') || 0;
+        return detect.getBreakpoint() !== 'mobile' && detect.adblockInUse() && config.switches.adblock && alreadyVisited > 1 && !userFeatures.isPayingMember();
     }
 
     return {
--- a/static/src/javascripts/projects/common/modules/commercial/adblock-messages.js
+++ b/static/src/javascripts/projects/common/modules/commercial/adblock-messages.js
@@ -9,18 +9,41 @@ define([
     storage,
     userFeatures
 ) {
-    function noAdblockMsg() {
+    function adblockInUse() {
+        return detect.adblockInUse();
+    }
+
+    function notMobile() {
+        return detect.getBreakpoint() !== 'mobile';
+    }
+
+    function isPayingMember() {
+        return userFeatures.isPayingMember();
+    }
+
+    function visitedMoreThanOnce() {
         var alreadyVisited = storage.local.get('gu.alreadyVisited') || 0;
 
-        return detect.adblockInUse() && detect.getBreakpoint() !== 'mobile' && (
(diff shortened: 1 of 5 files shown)


static/src/javascripts/projects/common/modules/commercial/adblock-messages.js
```diff
@@ -9,18 +9,18 @@ define([
     storage,
     userFeatures
 ) {
-    function sharedRules() {
-        var alreadyVisted = storage.local.get('gu.alreadyVisited') || 0;
-
-        return detect.getBreakpoint() !== 'mobile' && detect.adblockInUse() && config.switches.adblock && alreadyVisted > 1;
-    }
-
     function noAdblockMsg() {
```
jbreckmckye: I realize that you're calling the localStorage value multiple times so that you can make this module testable. But... I think the way we declare and use modules is a bit awkward, really. All of our services are written as stateful singletons, usually with init methods that we trigger in our tests and only-by-convention trigger once in our production code. It would seem much cleaner if we just had constructors for our services that could be instantiated in a service locator...

static/src/javascripts/projects/common/modules/commercial/adblock-messages.js
```diff
@@ -9,18 +9,18 @@ define([
...
+        return detect.adblockInUse() && detect.getBreakpoint() !== 'mobile' && (
+                alreadyVisited <= 1 ||
+                !config.switches.adblock ||
+                (config.switches.adblock && alreadyVisited > 1 && userFeatures.isPayingMember()));
     }
 
     function showAdblockMsg() {
-        return sharedRules() && !userFeatures.isPayingMember();
+        var alreadyVisited = storage.local.get('gu.alreadyVisited') || 0;
+        return detect.getBreakpoint() !== 'mobile' && detect.adblockInUse() && config.switches.adblock && alreadyVisited > 1 && !userFeatures.isPayingMember();
```
jbreckmckye: This line is a bit long. Also, I would put less computationally heavy switches at the start, so they can short-circuit the evaluation.

static/src/javascripts/projects/common/modules/commercial/adblock-messages.js
```diff
@@ -9,18 +9,18 @@ define([
...
-    }
-
     function noAdblockMsg() {
-        return sharedRules() && userFeatures.isPayingMember();
+        var alreadyVisited = storage.local.get('gu.alreadyVisited') || 0;
+
+        return detect.adblockInUse() && detect.getBreakpoint() !== 'mobile' && (
+                alreadyVisited <= 1 ||
+                !config.switches.adblock ||
+                (config.switches.adblock && alreadyVisited > 1 && userFeatures.isPayingMember()));
```
jbreckmckye: This logic seems a bit complex; maybe it would benefit being broken down more.

static/test/javascripts/spec/common/commercial/adblock-messages.spec.js
```diff
@@ -0,0 +1,101 @@
+define([
+    'helpers/injector',
+    'common/utils/storage'
+], function (
+    Injector,
+    storage
+) {
+    var injector = new Injector();
+
+    describe('Adblock messages', function () {
```
jbreckmckye: This looks like it's describing the messages themselves, as opposed to the service

static/test/javascripts/spec/common/commercial/adblock-messages.spec.js
```diff
@@ -0,0 +1,101 @@
...
+            ], function () {
+                adblockMessage = arguments[0];
+                config = arguments[1];
+                detect = arguments[2];
+                userFeatures = arguments[3];
+                done();
+            });
+        });
+
+        describe('If adblock banners and messages should be shown', function () {
```
jbreckmckye: I don't know if this nested describe is necessary

static/test/javascripts/spec/common/commercial/build-page-targeting.spec.js
```diff
@@ -166,6 +166,7 @@ define([
             audienceScienceGateway.getSegments = function () {
                 return {};
             };
+            storage.local.set('gu.alreadyVisited', 0);
```
jbreckmckye: If another test is leaving localStorage dirty, the ideal approach is to add an `afterEach` in that spec which cleans it up after each test.

static/test/javascripts/spec/common/commercial/adblock-messages.spec.js
```diff
@@ -0,0 +1,101 @@
...
+            it('should not show adblock messages for the first time users', function () {
+                storage.local.set('gu.alreadyVisited', 0);
+                config.switches.adblock = true;
+                detect.adblockInUse = function () {
```
jbreckmckye: You could reduce some boilerplate by making your beforeEach do something like

    detect.adblockInUse = function () {return mockUserHasAblock;}
    detect.getBreakpoint = function () {return mockBreakpoint;}

and then in your tests,

    mockUserHasAdblock = true;
    mockBreakpoint = 'desktop';

It might make the tests a little more concise.

static/test/javascripts/spec/common/commercial/adblock-messages.spec.js
```diff
@@ -0,0 +1,101 @@
...
+                    return true;
+                };
+                detect.getBreakpoint = function () {
+                    return 'desktop';
+                };
+
+                expect(adblockMessage.noAdblockMsg()).toBe(true);
+            });
+
+            it('should not show adblock messages for non adblock users', function () {
```
jbreckmckye: For someone reading these tests, I'm not sure if it's clear what the difference is between `noAdblockMsg()` and `!showAdblockMsg()`

static/src/javascripts/projects/common/modules/commercial/adblock-messages.js
```diff
@@ -9,18 +9,41 @@ define([
...
+    }
+
+    function visitedMoreThanOnce() {
+        var alreadyVisited = storage.local.get('gu.alreadyVisited') || 0;
+
+        return alreadyVisited > 1;
+    }
 
-        return detect.getBreakpoint() !== 'mobile' && detect.adblockInUse() && config.switches.adblock && alreadyVisted > 1;
+    function isAdblocSwitchOn() {
```
jbreckmckye: There's a typo here

jbreckmckye:
Other than the typo, it all looks good :+1:

PR #11073, opened 2015-11-06
Selection sharing

zeftilldeath:
Vertically centered the icons

Before
screen shot 2015-11-06 at 15 25 34

After
screen shot 2015-11-06 at 15 25 19

(Don't worry about the green, that's my mac)


rich-nguyen:
:+1:

PR #10937, opened 2015-10-22
Live streaming 404test

Calanthe:
Another 404 AB test variant - this time with live streaming option (including old 404 test cleanup)

Desktop:
screen shot 2015-10-21 at 16 30 11

screen shot 2015-10-21 at 16 36 44

Mobile:
screen shot 2015-10-21 at 16 30 33

screenshot

In order to pop into the test locally:
1. turn on the ab-live-events-survey
2. add `#ab-LiveEventsSurvey=variant` to the url

PR #11080, opened 2015-11-09
Live events AB test removed

Calanthe:
I left other files from the https://github.com/guardian/frontend/pull/10937, because there will be probably another variant of the test soon.


uplne:
:+1:

PR #11084, opened 2015-11-09
Add ESLint rule to prevent whole Lodash import

OliverJAsh:
Instead of importing all of Lodash, we must now be import individual modules, re. #11066.

This lint task will prevent us from accidentally importing all of Lodash, and thereby we can avoid file sizes jumping unexpectedly.

E.g.

``` js
// core.js
define(['lodash'], function () {});
```

Error:

```
$ grunt eslint:static/src
Running "eslint:static/src" (eslint) task

/Users/OliverJAsh/Development/frontend/static/src/javascripts/core.js
  1:9  error  Use Lodash modules instead  no-all-lodash-import
```

If you're using the ESLint plugin for Sublime, you have to add the rules to your global config, unfortunately: https://github.com/roadhump/SublimeLinter-eslint#i-want-to-use-custom-rules-global-eslintignore-file-etc

### DIFF
--- a/dev/eslint-rules/no-all-lodash-import.js
+++ b/dev/eslint-rules/no-all-lodash-import.js
@@ -1,0 +1,36 @@
+var isIdentifier = function (node) {
+    return node.type === 'Identifier';
+};
+var isArrayExpression = function (node) {
+    return node.type === 'ArrayExpression';
+};
+var isDefine = function (node) {
+    return isIdentifier(node) && node.name === 'define';
+};
+
+module.exports = function (context) {
+    return {
+        'CallExpression': function (node) {
+            var callee = node.callee;
+            var firstArg = node.arguments[0];
+            var hasDepsArray = firstArg && isArrayExpression(firstArg);
+
+            if (isDefine(callee) && hasDepsArray) {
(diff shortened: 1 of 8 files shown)


dev/eslint-rules/no-all-lodash-import.js
```diff
@@ -0,0 +1,36 @@
...
+                lodashImports.forEach(function (literalNode) {
+                    context.report({
+                        node: literalNode,
+                        message: 'Use Lodash modules instead'
+                    });
+                });
+            }
+        }
+    };
+};
```
OliverJAsh: Here's the AST: http://astexplorer.net/#/DDNrKEEHYc

sndrs:
:+1: great

johnduffell:
do we need a test or two for this?  I assume you've manually tested but feels like a couple of unit tests could give more confidence

OliverJAsh:
Good point John! See 9cf9256f38521bb080a2ad8730de1644939ad804

johnduffell:
fantastic! now ur testing the thing that tests our code :confounded: :laughing: :+1:

PR #11085, opened 2015-11-09
News and video Site Maps

rich-nguyen:
This is a migration of the old R2 news and video sitemap. Behaviour is roughly the same, except:
- the news site map contains up to 1000 articles from the last two days, rather than the 100 newest articles.
- the video site map was empty, so has similar logic to the new news sitemap.


applications/app/jobs/SitemapLifecycle.scala
```diff
@@ -0,0 +1,42 @@
...
+    }
+  }
+}
+
+object SiteMapJob extends ExecutionContexts with Logging {
+  case class SiteMapContent(
+    news: xml.NodeSeq,
+    video: xml.NodeSeq)
+
+  private val siteMapContent = AkkaAgent[SiteMapContent](SiteMapContent(Nil, Nil))
```
johnduffell: is it better to give an empty sitemap before the job completes successfully or an error?

applications/app/services/NewsSiteMap.scala
```diff
@@ -0,0 +1,104 @@
...
+  private case class Url(
+    location: String,
+    keywords: String,
+    title: String,
+    genres: String,
+    webPublicationDate: DateTime,
+    imageUrl: String) {
+
+    def xml() = {
+      <url>
```
johnduffell: this is really cool!

common/app/common/configuration.scala
```diff
@@ -195,6 +195,7 @@ class GuardianConfiguration(val application: String, val webappConfDirectory: St
 
   object images {
     lazy val path = configuration.getMandatoryStringProperty("images.path")
+    val fallbackLogo = Static("images/fallback-logo.png").path
```
jennysivapalan: Just worth noting the size of fallback-logo.png is currently optimised for facebook.

jennysivapalan:
:+1:

jennysivapalan:
@rich-nguyen can we get this updated and merged?

rich-nguyen:
@jennysivapalan yes! on it. needs router(s) changes of course too.

applications/app/services/NewsSiteMap.scala
```diff
@@ -0,0 +1,104 @@
...
+        val keywords = (keywordTags ++ sectionTag).mkString(", ")
+
+        val genres = item.tones.flatMap(_.webTitle match {
+          case "Blogposts" => Some("Blog")
+          case "Reviews" => Some("Opinion")
+          case "Comment" => Some("OpEd")
+          case _ => None
+        }).mkString(", ")
+
+        val imageUrl: String = item.mainPicture.flatMap(_.largestEditorialCrop.flatMap(_.url))
```
JustinPinner: Is `largestEditorialCrop` always the largest image available? I think R2 used to select the next image size down. Not that it's necessarily an issue, just a difference.

applications/app/services/NewsSiteMap.scala
```diff
@@ -0,0 +1,104 @@
...
+        Url(
+          location = item.webUrl,
+          keywords = keywords,
+          title = item.headline,
+          genres = genres,
+          webPublicationDate = item.webPublicationDate,
+          imageUrl = imageUrl)
+      }
+
+      UrlSet(urls take 1000).xml()
```
JustinPinner: I wonder if we need to return 1000 results? Again, maybe not an issue but an increase in bandwidth by a factor of 10 (and we get a lot of requests for these resources).

JustinPinner:
Looks good :+1:

applications/app/services/NewsSiteMap.scala
rich-nguyen: that's roughly how `largestEditorialCrop` behaves, picking a large non-4:3 crop, but falling back to largest image.

applications/app/services/NewsSiteMap.scala
rich-nguyen: I chose 1000 urls after speaking to chris M, who is unconvinced that these sitemaps have any positive or negative effect. The docs say 1000 urls (published within two days) is the news sitemap limit, so I'm hoping more urls will help news indexing...

https://support.google.com/news/publisher/answer/74288?hl=en-GB

applications/app/jobs/SitemapLifecycle.scala
rich-nguyen: Good question, It's not clear what the effect would be of this. Maybe I should provide stale content (through the cdn) when the agent is cold.

rich-nguyen: done, sending 5xx to CDN to indicate stale should be served.

PR #11087, opened 2015-11-10
Update README.md

patstirling:
Found I had to upgrade to latest version of java (wasn't particularly clear from anywhere else but Kelvin talked me through)


johnduffell:
might be worth mentioning what error you get if it's out of date, and possibly how to update?  Still an improvement though so :+1:

PR #10916, opened 2015-10-20
AB test for related content naming

johnduffell:
I've added an ab test to try different names for related content.
normal:
image
variant:
image

I'm not sure about the proportions in the test, also the exact phrases to be used in the tests may be updated once I spoke to a copy writer.  Any other comments would be appreciated!

PR #11088, opened 2015-11-10
remove onward names ab test

johnduffell:
The AB test had no effect from https://github.com/guardian/frontend/pull/10916 so removing all trace of it.

@Jholder112233 I'll leave the import of ab.js in related.js for your code that needs it too.


Jholder112233:
:+1:

PR #11089, opened 2015-11-10
remove secure-omniture switch

johnduffell:
this doesn't seem to have caused any tracking issues so removing the means to turn it off.

Now all no-js and amp omniture calls will go over https.   This avoids any mixed content errors when you hit an https page (amp, or info section) and it fails to detect that, thus serving an http fallback url.

@mkopka you haven't seen any issues have you?


NataliaLKB:
:+1:

PR #11090, opened 2015-11-10
Create standard.js and enhanced.js

sndrs:
`standard.js` does nothing so far. I'll add omniture in a subsequent PR.

this really just moves `bootstraps/app.js` > `bootstraps/enhanced.js` and set the scene.

cc @OliverJAsh @rich-nguyen @jennysivapalan


OliverJAsh:
:+1:

PR #11092, opened 2015-11-10
Removed omniture tracking from outbrain

uplne:
We don't need this anymore as we can get everything from ophan now.


jbreckmckye:
:+1:

PR #11093, opened 2015-11-10
Changes to the InjectNetworkFrontTest

holder-josh:
Now:
- We don't run the test on iPads or iPhones older than an iPhone 5
- We don't run the test on interactives or liveblogs
- Sample has been altered
- The network front is injected instead of related content, not most popular


static/src/javascripts/projects/common/modules/onward/inject-container.js
```diff
@@ -20,10 +22,12 @@ define([
             if (resp.html) {
                 fastdom.write(function () {
                     var $el = $(containerSelector);
-                    $el.before(resp.html);
-                    $el.css({
-                        display: 'none'
-                    });
+                    $el.after(resp.html);
+                    if (!(config.page && config.page.hasStoryPackage)) {
```
johnduffell: so we used to put the new stuff before the fallback, then hide the fallback. 
now we put the new stuff after, and keep the fallback if there's a story package

static/src/stylesheets/module/content/_article.scss
```diff
@@ -82,4 +82,9 @@
     .fc-container--has-show-more .treats__container {
         bottom: $gs-baseline;
     }
+
+    .fc-container--lazy-load {
+        display: block !important;
```
johnduffell: better to avoid important if possible, but if it's just for a test it won't be worth worrying too much

johnduffell:
looks good to me! :+1: :rocket:

static/src/javascripts/projects/common/modules/onward/inject-container.js
Jholder112233: yep!

static/src/stylesheets/module/content/_article.scss
Jholder112233: Was needed here for the test but will be removed later. It's scoped to only affect divs inside the test

michaelwmcnamara:
QA+1 - changes look good

PR #11094, opened 2015-11-10
Newspaper page for /theguardian

jennysivapalan:
As part of moving newspaper pages over to dotcom I've created `/theguardian` endpoint. Following the [R2 version](http://www.theguardian.com/theguardian) this page
- Queries the Content API to fetch results for today for the newspaper book path `theguardian/mainsection` using the query parameter `use-date=newspaper-edition` (which ensures just content that was in the paper will be returned).
- If today is Sunday then fallback to Saturday edition
- Results which have a page number of 1 will be in a `Front Page` container
- Other results are grouped by `newspaper-book-section` and ordered by page number
- Each grouped content by newspaper-book-section is transformed into a container.
- Containers are limited to 12 results, if the user wants to see more content for a container they can click through to the newspaper-book-section e.g.`theguardian/mainsection/uknews`

screen shot 2015-11-10 at 17 23 34

screen shot 2015-11-10 at 17 23 44

### DIFF
--- a/applications/app/services/TodaysNewspaperQuery.scala
+++ b/applications/app/services/TodaysNewspaperQuery.scala
@@ -98,7 +98,7 @@ object TodaysNewspaperQuery extends ExecutionContexts with Dates with Logging {
       Fixed(containerDefinition),
       CollectionConfigWithId(dataId.getOrElse(""), CollectionConfig.empty.copy(displayName = displayName)),
       CollectionEssentials(trails, Nil, displayName, dataId, None, None)
-    )
+    ).copy(hasShowMoreEnabled = false)
   }
 
   private def getNewspaperPageNumber(content: ApiContent) = content.fields.getOrElse(Map.empty).get("newspaperPageNumber").map(_.toInt)
--- a/applications/app/services/TodaysNewspaperQuery.scala
+++ b/applications/app/services/TodaysNewspaperQuery.scala
@@ -36,7 +36,7 @@ object TodaysNewspaperQuery extends ExecutionContexts with Dates with Logging {
 
       val firstPageContainer = {
         val content = firstPageContent.map(c => FaciaContentConvert.frontendContentToFaciaContent(Content(c)))
-        bookSectionContainer("front-page", Some("Front Page"), content, 0)
+        bookSectionContainer(None, Some("Front Page"), content, 0)
       }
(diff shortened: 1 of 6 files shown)


JustinPinner:
Quite a few hoops to jump through with this one then! Looking good though. Pity we have to fall back on those _._1 and _._2 references but that's expected I guess. And those Facia types... :sweat_drops:

Does it still need the standalone routes BTW?

jennysivapalan:
I realised after raising this PR that this isn't going to work for Sunday as it will return zero results. So another hoop to jump through. I'll look at alternatives for `_._1` and `_._2`.

applications/app/services/TodaysNewspaperQuery.scala
```diff
@@ -0,0 +1,105 @@
...
+
+    //order content for each book section
+    val orderedContentForBookSection: List[BookSectionContent] = unorderedBookSections.map { bookSection =>
+      bookSection.copy(content = orderContentByPageNumber(bookSection.content))
+    }
+
+    //order booksections by first content item in each booksection
+    val pageNumberToFaciaContainer: List[(Int, BookSectionContent)] = orderedContentForBookSection.flatMap { bookSection =>
+      val pageNumberOpt = bookSection.content.headOption.flatMap(content => getNewspaperPageNumber(content))
+      pageNumberOpt.map(_ -> bookSection)
```
johnduffell: if you want to get rid of the _1 and _2 consider instead of using a tuple here, make a case class with two named fields and use it here

jennysivapalan: I had similar approach for grouping the tag and content (which is a stepping stone to creating the BookSectionContent). I felt I was creating case classes for just one line. However if it it makes it more readable then it's I guess that is better. Will give it a go here.

jennysivapalan:
Added fallback for Sunday to use the Saturday edition. Also added the date to the Front Page container.

screen shot 2015-11-11 at 11 35 28

johnduffell:
anyway as discussed looks good to me! :+1: feel free to add the case classes though

jennysivapalan:
Thanks, yep the case classes make it much more readable so have added those in too.

johnduffell:
agreed, looks great :+1:

PR #11096, opened 2015-11-11
Serve core fronts to all ipads not just those with retina screens

michaelwmcnamara:
Currently we serve the core fronts to modern(ish) ipads - iPad3 or later.
This has proven successful in reducing the incidence of complaints about crashes and improving the stability of these iPads. 
This change will serve core fronts to all ipads, not just those with retina screens, with the aim of improving stability for these devices also.


common/app/templates/headerInlineJS/shouldEnhance.scala.js
```diff
@@ -33,11 +33,11 @@
...
     };
 
-    var isRetinaIpad = function() {
-        return (navigator.platform === 'iPad' && window.devicePixelRatio === 2);
+    var isIpad = function() {
+        return (navigator.platform === 'iPad');
     };
 
-    window.shouldEnhance = !personPrefersCore() && !isOlderIOSDevice() && !(@item.isFront && isRetinaIpad());
+    window.shouldEnhance = !personPrefersCore() && !isOlderIOSDevice() && !(@item.isFront && isIpad());
```
JustinPinner: I think we could remove the `olderIPadOnFront` check inside `isOlderIOSDevice` if we do it here instead?

michaelwmcnamara:
Very good idea! - will update.

common/app/templates/inlineJS/blocking/shouldEnhance.scala.js
```diff
@@ -23,21 +23,19 @@
     //
     // NOTE: this moves people into a category where they do not get important things such as commenting
     var isOlderIOSDevice = function () {
-        // This is NOT what we want to be doing long term. It is a stopgap measure only...
-        var olderIPadOnFront = @SplitOlderIPadsSwitch.isSwitchedOn && @item.isFront && window.devicePixelRatio === 1;
 
         if (navigator.platform === 'iPhone' || navigator.platform === 'iPad' || navigator.platform === 'iPod') {
             // I'm intentionally being a bit over zealous in the detection department here
-            return /.*(iPhone|iPad; CPU) OS ([3456])_\d+.*/.test(navigator.userAgent) || olderIPadOnFront;
+            return /.*(iPhone|iPad; CPU) OS ([3456])_\d+.*/.test(navigator.userAgent);
```
JustinPinner: So iPhones and iPads on iOS 3, 4, 5 or 6 will still not have _anything_ enhanced (is there anything _lower_ than iOS 3?). Any iPad running iOS 7 or above will have articles enhanced, while no iPads of any description will have fronts enhanced. iPhones running iOS 7 or above will be given the works for fronts and articles. I think that makes sense, but I just wanted to check these rules are definitely what we want.

JustinPinner:
After a quick sense check of the rules (see above) I think this gets a :+1:

michaelwmcnamara:
Thanks - Can definitely revisit the rules for iphones in a future change. Worth researching what their state is.

PR #11099, opened 2015-11-11
Show Example Page of MPU Container Templates

kelvin-chappell:
This change enables us to show an example page where all containers have an MPU.

It also cleans up some leftover code and does a bit of refactoring.

The key commit is cf7ab32.

/cc @Calanthe @uplne @jbreckmckye

### DIFF
--- a/static/src/stylesheets/module/facia/_items.scss
+++ b/static/src/stylesheets/module/facia/_items.scss
@@ -333,12 +333,6 @@
     @include fc-item--list-compact--media;
 }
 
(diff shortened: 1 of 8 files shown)


static/src/stylesheets/module/facia/_items.scss
```diff
@@ -333,12 +333,6 @@
     @include fc-item--list-compact--media;
 }
 
-.omitted-mpus .l-row--cols-3 .l-row__item--span-1 .fc-item__avatar__media {
```
uplne: I love this

uplne:
Looks good to me :+1:

PR #11095, opened 2015-11-10
Reorganise the assets in the head, in prep for standard/enhance JS

sndrs:
- makes clear what's happening the head (hopefully)
- renames head js to inline blocking js

cc @rich-nguyen @OliverJAsh @jbreckmckye

PR #11102, opened 2015-11-11
Fix offline page

OliverJAsh:
Broken by https://github.com/guardian/frontend/pull/11095

/cc @sndrs

### DIFF
--- a/applications/app/views/offlinePage.scala.html
+++ b/applications/app/views/offlinePage.scala.html
@@ -1,6 +1,9 @@
 @(offlinePage: controllers.OfflinePage)(implicit request: RequestHeader)
 
 @import play.api.libs.json._
+@import conf.switches.Switches._
+@import common.InlineJs
+@import templates.inlineJS.blocking.js._
 
 <!DOCTYPE html>
 <html id="js-context" class="js-off is-not-modern id--signed-out" data-page-path="@request.path">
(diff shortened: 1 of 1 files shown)


applications/app/views/offlinePage.scala.html
```diff
@@ -14,10 +17,65 @@
     @* get the stylesheets downloading ASAP *@
     @fragments.stylesheets(None)
 
-    @* load the app *@
+    @* We don't use the inlineJSBlocking template because we need to specify
+    our own paths *@
```
OliverJAsh: For context: back when I added the offline page, @rich-nguyen and I decided not to re-use the main template, but just do extract the things needed for the offline page. This gives us increased flexibility, allowing us to specify custom paths (see below). We need to set custom paths because we must use bundles in dev, so that the offline page can be tested.

OliverJAsh: @rich-nguyen After talking to @sndrs about this, I think we should try to avoid duplication. I actually created a system to do this beforehand—you could pass in a flag to the template called `alwaysUseBundlesInDev` (or something), but that was quite complicated and too awkward.

We could either go back to a system such as this, or allow `window.curl.paths` to be pre-defined, e.g.:

``` js
// in offline page JS:
window.curlPaths = {
  // bundles
};

// in general JS:
window.curl = {
  api: 'require',
  paths: window.curlPaths || {
    // normal paths
  }
};
```

Can we get a consensus on this before I do the work, so I don't repeat myself? :smile:

sndrs: :100:

rich-nguyen: I still think dryness only makes sense when duplication is excessive, and it allows the reader to better understand the code. I think `alwaysUseBundlesInDev` makes little sense to others, because it's not clear why or when it's needed. Adding a parameter in our main config system for the purpose of a single page doesn't feel right.

Ideally, I don't want to think about the offline page system when I'm doing online page work (~100% of my work). Yes, this implies I might break the offline page, in my ignorance. I'm happy with that, because I never agreed to maintaining it, nor should the whole team have that burden. Which is why I believe a stand-alone, isolated config is preferable.

rich-nguyen:
:+1:

PR #11108, opened 2015-11-12
Store cookies without subdomains on pre-production environments

jbreckmckye:
Sometimes we need to store cookies against the broad `.theguardian.com` superdomain, so that services running on different subdomains can share them. We do this automatically on production, and on development when nextgen is run under `dev.theguardian.com`, but not under `m.thegulocal.com`, or `m.code-dev.theguardian.com`.

This is a requirement to testing an upcoming story to remove the `gu_paying_user` cookie on membership-frontend.


Calanthe:
:+1:

PR #11112, opened 2015-11-12
Hide adblock byline image on print

jbreckmckye:
If a page has an above-nav adblock banner...

image

Then when we print it, the byline image takes over the page:

image
_Above: George Monbiot is disappointed that you're using Adblock Plus. And paper._

There's already a rule to hide `byline-img` components in print, so I've added that behaviour to the `adblock-byline-img` component too. And now the print is a little more sane -

image
_George is still unhappy that you're using an adblocker, but at least you're consuming a little less paper._


Calanthe:
:+1: I like the conclusion ;).

PR #11115, opened 2015-11-13
InjectNetworkFrontTest fixes

holder-josh:
The unexpected adslots that end up on the page when injecting the network front were causing problems with the DFP code. This removes them in the most hacky manner. Hopefully fine for a 5 day test?


Jholder112233:
@uplne this seems to consistently work locally now, does it make sense that this would fix the problem?

uplne:
Yes that is exactly what I've found too. It was caused by merchandising component mostly and it was random because that component is not on the page every time.

So good work :+1: it is a little bit hacky but I suppose for the AB tests it is ok.

Jholder112233:
@uplne nice! thank you very much.

PR #11118, opened 2015-11-13
Always use theguardian.com proxy and over HTTPS

nicl:
Use theguardian.com proxy and always over HTTPS.


nicl:
:+1: (from john!)

PR #11128, opened 2015-11-13
Optimize Load of DFP Creative Templates so that it Doesn't Bring Down the Admin Box

kelvin-chappell:
This separates out caching of creative templates and the creatives built from templates.
An akka agent now holds the active creative templates.
The creatives themselves are being cached in S3 and updated when modified.

There's also a bit of refactoring of DFP API loading generally to make code clearer and add consistent logging around API requests.


kenlim:
:+1:

PR #11129, opened 2015-11-13
Use proxy only for POSTs and secure GETs

nicl:
Re-opened, this time with changed proxy behaviour - only call proxy for secure requests.


static/src/javascripts/projects/common/modules/discussion/api.js
```diff
@@ -1,28 +1,27 @@
...
-            : config.page.discussionApiRoot,
-        Api = {
-            root: root,
-            // TODO get rid of discussion proxy completely when we're changed over to https
-            proxyRoot: (config.switches.discussionProxy ? (config.page.host + '/guardianapis/discussion/discussion-api') : root),
-            clientHeader: config.page.discussionApiClientHeader
-        };
+    var Api = {
+        root: config.page.discussionApiRoot,
+        proxyRoot: (config.switches.discussionProxy ? (config.page.host + '/guardianapis/discussion/discussion-api') : root),
```
johnduffell: this switch has a slightly different purpose now so should change the desc https://github.com/guardian/frontend/blob/master/common/app/conf/switches/FeatureSwitches.scala#L320

johnduffell:
I'd feel more comfortable with a little more comments, but overall :+1:

PR #11132, opened 2015-11-16
Revert "Revert "Use proxy only for POSTs and secure GETs""

nicl:
Reverts guardian/frontend#11130, which in turn is a revert. Inception-style madness.


johnduffell:
:+1:

PR #11134, opened 2015-11-16
Restart InjectNetworkFrontTest & set up dummy test

holder-josh:
We weren't getting a nice 50/50 sample for the InjectNetworkFrontTest AB test and it's the second time this has happened on the 0% to 2% test sample. 

Therefore:
*We're restarting that test with a different offset and doubled in size
*Set up a dummy test on that 0% to 2% sample to see if the same problem persists

Hopefully by doing this we should be able to definitively say whether there's a problem with the AB testing framework's sampling.

@NataliaLKB do you fancy giving me a quick +1? :sweat_smile:


static/src/javascripts/projects/common/modules/experiments/ab.js
```diff
@@ -7,8 +7,8 @@ define([
     'common/modules/analytics/mvt-cookie',
     'common/modules/experiments/tests/high-commercial-component',
     'common/modules/experiments/tests/membership-message-usa',
-    'common/modules/experiments/tests/inject-network-front-test',
-    'common/modules/experiments/tests/most-popular-default',
```
NataliaLKB: Why remove the `most-popular-default` test in this file?

NataliaLKB:
Other than the one question, everything else looks good! :+1:

static/src/javascripts/projects/common/modules/experiments/ab.js
Jholder112233: Whoops, good spot. Hasty merge conflict resolving has let me down. Will fix.

NataliaLKB: Thanks!

NataliaLKB:
:ship:

PR #11136, opened 2015-11-16
remove facia tool switches

piuccio:
they've been migrated to a new switchboard

the switches left are used by facia press or admin


jennysivapalan:
:+1:

PR #11137, opened 2015-11-16
Move top banner below first container server side test

uplne:
This will be part of a larger AB commercial AB test.

Part of that test is to remove top banner and move it below the first container.

This is not going to production yet. It is only for further testing on CODE.

screen shot 2015-11-16 at 15 36 30


static/src/javascripts/projects/common/modules/commercial/front-large.js
```diff
@@ -0,0 +1,35 @@
...
+    createAdSlot
+) {
+
+    function init() {
+        if (typeof config.tests.topBannerPosition === 'undefined' || !config.page.isFront) {
+            return false;
+        }
+
+        var containerIndex = 0,
+            $adSlotWrapper = $.create('<div class="fc-container large-top top-banner-ad-container top-banner-ad-container--desktop top-banner-ad-container--above-nav"></div>'),
```
jbreckmckye: I'm not sure it's obvious what a `large-top` is. It seems to imply a larger upper banner, perhaps?

static/src/stylesheets/module/_adslot.scss
```diff
@@ -128,6 +128,22 @@
...
+    text-align: center;
+    z-index: 0;
+
+    .ad-slot__label {
+        width: 970px; // Size of the large banner 970x250
+        margin: 0 auto;
+    }
+
+    @include mq($until: desktop) {
+        display: none !important;
```
jbreckmckye: I understand why we need this, but the fact we do so says something about our existing style code.

common/app/views/main.scala.html
```diff
@@ -46,21 +46,25 @@
         <a class="u-h skip" href="#maincontent" data-link-name="skip : main content">Skip to main content</a>
 
         @if(!(metaData.isImmersive && metaData.isArticle)) {
-            @if(showAdverts) {
-                @fragments.commercial.topBanner(metaData, edition)
-            }
+            @if(!mvt.TopBannerPosition.isParticipating) {
```
kelvin-chappell: This template will probably need a bit of cleaning up when the test is over.

kelvin-chappell:
Does the test only apply to desktop?  If you're in the test group I don't think you'll see the top ad at all in mobile.
Apart from that :+1:

johnduffell:
is this going out?

PR #11138, opened 2015-11-16
restarting `most popular default` test

NataliaLKB:
Due to the unusual traffic happening this weekend, we are restarting this test to see if the results differ.

cc @Jholder112233


Jholder112233:
:+1: :ship:

PR #11144, opened 2015-11-18
Allow users to insist on receiving enhanced content

JustinPinner:
(Ignore the -ipads part of the branch name!)

As discussed with @michaelwmcnamara - some users don't have problems on older devices, but they can't opt back in to enhanced mode if they're on an iPad or an older iOS device.

@sndrs I'd appreciate your opinion too, is this too broad in scope do you think? It would affect anyone on any device with the appropriate hash or stored preference.

### DIFF
--- a/common/app/templates/inlineJS/blocking/shouldEnhance.scala.js
+++ b/common/app/templates/inlineJS/blocking/shouldEnhance.scala.js
@@ -15,6 +15,17 @@
         }
     };
 
+    var personDemandsFeatures = function () {
+        var locationHash = window.location.hash;
+        if (locationHash === '#featureson' || locationHash === '#nocore' || locationHash === '#gu.prefs.force-core=off') return true;
+        try {
+            var preference = window.localStorage.getItem('gu.prefs.force-core') || 'unknown';
+            return /"value":"off"/.test(preference);
+        } catch (e) {
+            return false;
+        }
+    };
+
     // Guess whether the device is too old, regardless of whether it cuts the mustard
     //
     // 'older' iOS normally indicates a device with lower power (they stop being upgradeable at some point).
@@ -35,7 +46,7 @@
         return (navigator.platform === 'iPad');
     };
 
-    window.shouldEnhance = !personPrefersCore() && !isOlderIOSDevice() && !(@item.isFront && isIpad());
+    window.shouldEnhance = personDemandsFeatures() || (!personPrefersCore() && !isOlderIOSDevice() && !(@item.isFront && isIpad()));
     window.shouldEnhance || console && console.info && console.info("THIS IS CORE");
 })(navigator, window);
 


common/app/templates/inlineJS/blocking/shouldEnhance.scala.js
```diff
@@ -15,6 +15,17 @@
         }
     };
 
+    var personDemandsFeatures = function () {
+        var locationHash = window.location.hash;
+        if (locationHash === '#featureson' || locationHash === '#nocore' || locationHash === '#gu.prefs.force-core=off') return true;
+        try {
+            var preference = window.localStorage.getItem('gu.prefs.force-core') || 'unknown';
+            return /"value":"off"/.test(preference);
```
OliverJAsh: Minor comment, but `JSON.parse` wouldn't be expensive and would make this a tiny bit more readable. Happy to leave as is, though :smile:

common/app/templates/inlineJS/blocking/shouldEnhance.scala.js
```diff
@@ -35,7 +46,7 @@
         return (navigator.platform === 'iPad');
     };
 
-    window.shouldEnhance = !personPrefersCore() && !isOlderIOSDevice() && !(@item.isFront && isIpad());
+    window.shouldEnhance = personDemandsFeatures() || (!personPrefersCore() && !isOlderIOSDevice() && !(@item.isFront && isIpad()));
```
OliverJAsh: Could this be simpler? We could have a function, `isCore`, that returns true or false. Then this becomes:

``` js
window.shouldEnhance = !isCore() || (!isCore() && !isOlderIOSDevice() || !(@item.isFront && isIpad()))
```

common/app/templates/inlineJS/blocking/shouldEnhance.scala.js
```diff
@@ -15,6 +15,17 @@
         }
     };
 
+    var personDemandsFeatures = function () {
```
sndrs: now we've settled on nomenclature, maybe this should be `personDemandsEnhanced`?

common/app/templates/inlineJS/blocking/shouldEnhance.scala.js
```diff
@@ -15,6 +15,17 @@
...
+        var locationHash = window.location.hash;
+        if (locationHash === '#featureson' || locationHash === '#nocore' || locationHash === '#gu.prefs.force-core=off') return true;
```
johnduffell: how come 2 hashes to do the same thing?

common/app/templates/inlineJS/blocking/shouldEnhance.scala.js
johnduffell: there are 3 categories @OliverJAsh : demands features, prefers core, or no preferences

OliverJAsh: If the user demands features, `isCore` would return `false`. If the user prefers core, it would return `true`. If the user has no prefs, we return `false`?

johnduffell: so if they have no preferences, it will enhance even if they're on old ipads

michaelwmcnamara:
QA +1 
Tested with the following results:
ipad
		no pref - shows basic
		gurprefs.force-core=off  - shows enhanced
		gurprefs.force-core=on  - shows basic
		#nocore #featureson  - shows enhanced
		#core  #featuresoff  - shows basic


nonipad 
               - no pref - shows enhanced 
		gurprefs.force-core=off  - shows enhanced
		gurprefs.force-core=on  - shows basic
		#nocore #featureson  - shows enhanced
		#core  #featuresoff  - shows basic

sndrs:
:+1:

PR #11146, opened 2015-11-18
Use render other status for picture content

rich-nguyen:
Rather than return a 410 http status, picture content should show the content removed page.


gtrufitt:
:+1:

PR #11152, opened 2015-11-19
Frequency param in DFP call set to 20plus

uplne:
So we can track user behaviour more deeply and possibly stop showing high relevance commercial components on the lower numbers which will allow us to show more Outbrain.

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/build-page-targeting.js
+++ b/static/src/javascripts/projects/common/modules/commercial/build-page-targeting.js
@@ -98,16 +98,7 @@ define([
             }
         },
         getVisitedValue = function () {
-            var alreadyVisited = storage.local.get('gu.alreadyVisited') || 0,
-                visitedValue;
-
-            if (alreadyVisited > 19) {
-                visitedValue = '20plus';
-            } else {
-                visitedValue = alreadyVisited.toString();
-            }
-
-            return visitedValue;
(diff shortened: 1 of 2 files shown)


jbreckmckye:
How does this all work on the DFP side? Am I right in thinking that DFP can't do any numerical comparisons - it just treats the numbers like string keys? In which case, does this mean we're happy with no longer targeting the range between five and twenty?

uplne:
This is for us. Some commercial components are being clicked on after 5+ visit. This will help merchandising team to decide on after how many visits it is effective to show user component with the high probability of click through.

As a result of that it has a potential of an over one million more views for outbrain.

jbreckmckye:
I see, it's just for tracking purposes rather than DFP targeting. But then, why don't we just pass the value straight up, without any capping? Is it just easier than doing the data processing?

uplne:
I remember that @Calanthe had discussion around this a few months ago and they decided to do it this way.

However I see your point and I am happy to chase it again.

uplne:
Ok, capping is not needed anymore so we pass the value straight up.

kelvin-chappell:
It's very difficult to report on key-value pairs and you could get this data more easily from Ophan or Omniture.  So unless it's actually being used for targeting I'd leave this out altogether.

uplne:
I will open this again when the decision will be clear

PR #11153, opened 2015-11-19
Extending Most Popular Default Switch

NataliaLKB:
Because the traffic on the site has been a bit out of the ordinary as of late, we are extending this test by a few more days. 

cc @Jholder112233


Jholder112233:
:+1:

Issue #11148, opened 2015-11-18
AB testing framework sampling isn't even


Jholder112233:
I've had problems now with numerous tests that ran on the 0-2% sample, where the variant was receiving around half the page views of the control. So I investigated and set up a dummy test which did nothing and always returned true. 

The result:
<img width="958" alt="screen shot 2015-11-18 at 14 33 16" src="https://cloud.githubusercontent.com/assets/2236852/11243682/5858d800-8e01-11e5-995b-eab5f86491c1.png">

The tests which had problems on that sample worked perfectly on a different sample.

Whoever picks this up, come grab me to chat about it!

rich-nguyen:
Very alarming! Let's investigate this together tomorrow, we've had problems like this in the past but thought we'd fixed them.

Jholder112233:
Seems to have been fixed by https://github.com/guardian/frontend/pull/11156

PR #11156, opened 2015-11-19
AB sampling changes

holder-josh:
This may solve https://github.com/guardian/frontend/issues/11148

cc @johnduffell @rich-nguyen


static/src/javascripts/projects/common/modules/experiments/ab.js
```diff
@@ -182,7 +182,7 @@ define([
         // Get this browser's mvt test id.
             mvtCookieId = mvtCookie.getMvtValue();
 
-        if (smallestTestId <= mvtCookieId && largestTestId > mvtCookieId) {
```
johnduffell: interestingly, our range for client side ab tests is 899999 and our range for server side ones is 100000, and we have one id left over that would never get in a test: https://github.com/guardian/frontend/blob/77e10c3bfd40fc4910546796d8aecc1ab7739f5e/static/src/javascripts/projects/common/modules/analytics/mvt-cookie.js#L11

johnduffell: the code that sets the id looks like it sets it correctly in the range 1-1000000 inclusive and does indeed orphan id 1000000 from the server side mvt tests https://github.com/guardian/fastly-edge-cache/blob/d7d52878ae680ef3433351c17f8f5de9c62bd946/theguardiancom/src/main/resources/varnish21/ab-tests.vcl#L27

johnduffell:
:+1: looks good, although interesting facts I just learned that
- we can only run client side ab tests on 99% of the population, and
- the client side ab test groups are interleaved in the population ids, wheras the server side ones are allocated in blocks

PR #11157, opened 2015-11-19
Removed inline-block

zeftilldeath:
Seems a bug with column (-webkit-column-gap etc) was pushing the absolute positioned numbers all over the place.

Managed to solve it by removing inline-block from list items. Let me know if this is stupid in any way.

Before (Note: number 5, 10):
screen shot 2015-11-19 at 17 31 25

After:
screen shot 2015-11-19 at 17 31 42

### DIFF
--- a/static/src/stylesheets/module/_headline-list.scss
+++ b/static/src/stylesheets/module/_headline-list.scss
@@ -31,19 +31,19 @@
             width: 40%;
         }
     }
+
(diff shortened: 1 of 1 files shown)


NataliaLKB:
It _is_ a weird webkit specific problem! 

Looks good :+1:

sndrs:
have you checked this in browserstack? would be a good one for a cross-browser screenshot test

PR #10051, opened 2015-08-06
Add USA Membership/Supporter message test

davidrapson:
This PR adds a new A/B test for a US edition version of the Supporter banner.
- Excludes adblock users to avoid conflicts with similar adblock Supporter message
- Excludes mobile/small-screen devices
- Only shown for US edition
- Only shown on Article pages
- Only shown to visitors who have viewed at least 10 pages.

The banner links to https://membership.theguardian.com/us/supporter

**A**

screen shot 2015-08-06 at 17 03 58

**B**

screen shot 2015-08-06 at 17 04 20

@OliverJAsh @dominickendrick

PR #11158, opened 2015-11-19
Extend useful banner...

rtyley:
It brings in Members, but we should stop it from being an A/B test...

cc @kenlim @Calanthe 

Originally created back in https://github.com/guardian/frontend/pull/10051


uplne:
:+1:

johnduffell:
seems like this has gone now anyway

PR #11159, opened 2015-11-20
Prevent space-finder collisions due to race conditions

jbreckmckye:
## The bug

The asynchronous way we find spaces to insert components, and add them to the page, is a source of several longstanding article bugs. We run space-finder, queue up an action to write an element, and run space-finder again before that element has been written. Naturally, this can mean two elements occupying the same space:

image

We can see the rich-link `aside` element and the `#dfp-ad--im` commercial component next to each other in the DOM:

image

The actual flow of execution makes the cause more obvious:
1. Article-body-ads makes a spacefinder query for an advert. This queues a fastdom read to be run on the next frame.
2. Rich-links makes a spacefinder query for a rich link. This queues a fastdom read to be run on the next frame.
3. On the next frame, after all DOM writes have finished, fastdom runs the queued reads, looking for a free space to put their components. Neither component has been written so far, so both reads return the same paragraph.
4. On the next tick, their `then` methods are resolved. This dispatches asynchronous writes for both components.
5. The components appear next to one another.
## The fix

To stop article body adverts and rich links colliding with one another, they both now use a new `space-filler` service, which acts as a proxy for space-finder that ensures all injections run in the expected order. This pull request refactors `article-body-adverts.js` and `rich-links.js` to use spacefiller. It also updates `article-body-adverts.spec.js` to test how the module interacts directly with space-filler rather than indirectly with space-finder.
## Space-filler in detail
- The service takes a ruleset for finding a space, and a _synchronous_ writer function for injecting a component. It returns you a promise.
- When any pending writes have finished, space-filler will do a search for your component next, and run your writer against any paragraph it finds using Fastdom.
- Once your writer has been either run or discarded, the returned promise resolves as either true (writer was run) or false (no paragraph found).
- All calls to space-filler made after yours will be queued similarly
- If a writer function throws an exception, it is logged to Sentry and the promise is resolved as false, before space-filler moves on to the next query.
## The result

We no longer inject two elements into the same location:

image


static/src/javascripts/projects/common/modules/article/space-filler.js
```diff
@@ -0,0 +1,61 @@
...
+            function promiseInsertion(para) {
+                return new Promise(function (resolve) {
+                    idleFastdom.write(function () {
+                        try {
+                            writer(para);
+                            resolve(true);
+                        } catch (e) {
+                            // log and move on
+                            raven.captureException(e);
+                            resolve(false);
```
desbo: Would it be better to `reject(e)` here so you can use `.catch` rather than checking for success in `.then`?

desbo:
Minor comment aside this looks great, nice one. :+1:

jbreckmckye:
@desbo - using `catch` also means we can intercept bugs in `getParaWithSpace`, too.

PR #11160, opened 2015-11-20
Essential read test changes

holder-josh:
- Made sure the curated and automated versions have different data-link-names 
- Fixed a problem with the curated version only picking 2 stories
- Sets the sample up correctly

All systems go for Monday morning.


NathanielBennett:
:+1: Looks good! ( I thought 0.05% was a little over-cautious)

PR #11161, opened 2015-11-20
add nofollow to report-abuse link

gidsg:
We see many, many hits to these URLs in our logs which we suspect most are bots. This link has no value to bots, so marking `rel=nofollow`

_Bots, nothing to see here, please move along..._
CC @guardian/discussion


NathanielBennett:
:+1: Ship it ( when the tests pass )

PR #11165, opened 2015-11-23
Sort comments by number of recommendations

piuccio:
Because comments API finally allows that. It's possible to sort asc/desc but I don't see the value of allowing users to sort by least recommended comments.


NataliaLKB:
:+1:

OliverJAsh:
Nice :+1:

PR #11166, opened 2015-11-23
Move membership banner for US audience out of A/B tests

regiskuckaertz:
As requested in https://github.com/guardian/frontend/pull/11158, the banner for US readers has been removed from A/B tests and merged with the similar banner for readers of the UK edition.


static/src/stylesheets/module/_release-message.scss
```diff
@@ -432,7 +432,8 @@ $btn-height: gs-height(1) - $gs-baseline/2; //30px
 .site-message--membership-message,
 .site-message--membership-message-uk,
 .site-message--membership-message-uk-redisplayed,
-.site-message--membership-message-usa {
+.site-message--membership-message-us,
+.site-message--membership-message-us-redisplayed {
```
uplne: Do we need to have these class names for each edition? I believe the templates have identical styling so they should be good with just generic class names `.site-message--membership-message` and `.site-message--membership-message-redisplayed`.

uplne: You might also want to remove switch for this AB test. AB tests are always behind the switch.

regiskuckaertz: Will do. Actually even the `-redisplayed` class is redundant.

uplne:
:+1:

PR #11167, opened 2015-11-23
Update height settings to ensure inline form height is set properly on page load

gtrufitt:
There was a big gap below the inlined form as the iframe height was being set after the container height - for some reason I had the callback called the wrong way around, this fixes that...


NataliaLKB:
:+1: :ship:

PR #11171, opened 2015-11-23
Add referrer to dfp call

uplne:
We want to see in DFP from where user came as users from different referrers have different behaviour. So we can target commercial components more precisely and possibly have more space for outbrain.

This adds another parameter to be send to DFP.


kelvin-chappell:
Looks interesting.  :+1:

PR #11172, opened 2015-11-23
Update mobile landscape and phablet form widths

gtrufitt:
Make the email form look a little neater.

### DIFF
--- a/static/src/stylesheets/module/email/_form.scss
+++ b/static/src/stylesheets/module/email/_form.scss
@@ -3,12 +3,10 @@
 }
 
 .email-sub__iframe {
-    @if ($old-ie) {
-        width: 100%;
-    } @else {
-        width: calc(100% + 8px);
-        margin-left: -4px;
-    }
+    // Calc with fallback
+    width: 100%;
+    width: calc(100% + 8px);
+    margin-left: -4px;
 }
 
 .email-sub__form {
--- a/static/src/stylesheets/module/email/_form.scss
+++ b/static/src/stylesheets/module/email/_form.scss
@@ -3,8 +3,13 @@
 }
 
 .email-sub__iframe {
-    // Calc with fallback
+    // Fallback for browsers that don't support calc
     width: 100%;
+}
+
+.email-sub__iframe {
+    // pixel calculations are manual design
+    // adjustments, off grid
     width: calc(100% + 8px);
     margin-left: -4px;
 }
--- a/static/src/stylesheets/module/email/_form.scss
+++ b/static/src/stylesheets/module/email/_form.scss
@@ -7,6 +7,11 @@
     // adjustments, off grid
     width: calc(100% + 8px);
     margin-left: -4px;
+
+    @include mq($from: mobileLandscape, $until: tablet) {
+        margin-right: 50%;
+        width: calc(50% + 8px);
+    }
 }
 
 .email-sub__form {
--- a/static/src/stylesheets/module/email/_form.scss
+++ b/static/src/stylesheets/module/email/_form.scss
@@ -3,14 +3,11 @@
 }
 
 .email-sub__iframe {
-    // pixel calculations are manual design 
-    // adjustments, off grid
-    width: calc(100% + 8px);
-    margin-left: -4px;
-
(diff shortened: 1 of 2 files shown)


static/src/stylesheets/module/email/_form.scss
```diff
@@ -7,6 +7,11 @@
     // adjustments, off grid
     width: calc(100% + 8px);
     margin-left: -4px;
+
+    @include mq($from: mobileLandscape, $until: tablet) {
+        margin-right: 50%;
+        width: calc(50% + 8px);
```
Jholder112233: Do you not need a fallback for browsers that don't support calc?

gtrufitt: IE8 doesn't support media queries, so this calc would be ignored anyway. IE9 supports calc. You could be right about the previous one though.

sndrs: careful with `calc`s, they can be expensive to process on lower powered devices since they generally (afaik) require multiple passes in layout. i'd be nervous to see them creep in a lot...

gtrufitt: Hmm, this is a good point and makes sense in terms of the layout passes. Do you have any reading on this? I couldn't find anything in-particular RE calc performance, though obviously multiple passes on the layout is a common sense performance hit... I'll have a re-think anyway.

sndrs: i can't recall now, that's why i parenthesised that bit – i'll see if i can dig the ref out. i'd be very interested if you found something, will have another look. not a :-1:, just a caution i guess. it could just be myth-making on my part, but it feels instinctively right too, as you say.

gtrufitt: Fair enough, I found a bunch of stuff about the long-gone css expressions, and some anecdotal stuff around calc but no solid numbers. I think I'll stick with `calc` as the alternatives to using it are fairly loop-the-loop layout wise but I am going to raise myself an issue to look into this a bit more and swap it out if need be. I'll let you know if I find anything, shout if you come across anything yourself.

sndrs: we have a few instances of calc in the codebase, mainly commercial but not exclusively. given we trigger layout on scroll relentlessly, i wonder how much difference it makes? would be v interesting to see what diff it makes to scroll without them on an older device...

gtrufitt: I'll see if I can compose a test to find out. My gut feel is that it'll be a micro optimisation, especially compared to bits of JS, but would be good to prove that :D

sndrs: if so that would be great, because it's handy!

sndrs: seems like it might be almost free: https://twitter.com/asanders/status/669116090908336129

gtrufitt: Wow, perfect. Calc all the things! ;) Cheers Alex.

static/src/stylesheets/module/email/_form.scss
```diff
@@ -3,10 +3,12 @@
...
 
 .email-sub__iframe {
-    // pixel calculations are manual design 
-    // adjustments, off grid
-    width: calc(100% + 8px);
-    margin-left: -4px;
+    @if ($old-ie) {
+        width: 100%;
+    } @else {
+        width: calc(100% + 8px);
```
sndrs: `$gs-gutter` :)

sndrs: `$gs-gutter`?

gtrufitt: I shouldn't have removed that comment :)

> // pixel calculations are manual design 
> // adjustments, off grid

@superfrank was making adjustments based on the curve of the earth or something.

sndrs: @superfrank why can they not be part of the grid calcs? if we change the grid, will those still be the correct values?

Jholder112233:
:+1:

PR #11175, opened 2015-11-23
Upgrade reqwest to 2.0.5

desbo:
There's [a bug](https://github.com/ded/reqwest/issues/160) in the version of reqwest we use that throws errors on 204 (empty) responses. We get that type of response from an `OPTIONS` request made in the new email form during a CORS `POST`. 

The fix we want was added in version 2. Although this is a major upgrade, the main difference is to support nodejs via `xhr2` and there's a negligible effect on filesize (<400 bytes uncompressed). Once this builds I'll put it on code and make sure everything still works.
- [x] do the testing

@sndrs @OliverJAsh

### DIFF
--- a/static/src/javascripts/bower.json
+++ b/static/src/javascripts/bower.json
@@ -11,7 +11,7 @@
     "curl": "~0.8.9",
     "qwery": "3.4.2",
     "react": "0.13.2",
-    "reqwest": "1.1.5",
+    "reqwest": "2.0.5",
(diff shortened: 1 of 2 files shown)


piuccio:
nice, totally unrelated, would be nice to convert to `fetch` :trollface:

gtrufitt:
:+1:

sndrs:
:+1:

PR #11182, opened 2015-11-24
Remove OfflinePageSwitch

OliverJAsh:


JustinPinner:
:+1: I think this looks ok - quite a bit has moved around in the side-by-side comparison view :eyes:

PR #11185, opened 2015-11-24
Observer logo

jennysivapalan:
/theobserver gets the Observer logo! Also renamed section navigation so that it does not clash with /observer.

cc @JustinPinner


JustinPinner:
:+1:

PR #11186, opened 2015-11-24
Headline list numbers

zeftilldeath:
More headaches with columns. Still not perfect but this is the best for the time being. The text doesn't wrap under the numbers. 

screen shot 2015-11-24 at 17 34 29


NataliaLKB:
:+1:

PR #11187, opened 2015-11-24
Increase GU_EDITION cookie expiry time

OliverJAsh:
This will improve the UX so users don't have to keep switching their preference every 2 months (as one user complained).

Apparently we did this in the early days of Guardian US to increase traffic on their side, but now it's grown significantly, it's no longer necessary.


NataliaLKB:
:watch:  :+1:

Issue #9957, opened 2015-07-29
Lightbox zoom doesn't work in iOS Safari


OliverJAsh:
And maybe others, but works in Chrome on my Google Nexus 5.

PR #11190, opened 2015-11-25
updating viewport meta tag to be more accessible

NataliaLKB:
Amp have relaxed their requirements their viewport meta tag. 

Discussion here: https://github.com/ampproject/amphtml/issues/481

cc @johnduffell


johnduffell:
awesome! :+1:

paperboyo:
Very interesting discussion there, thanks, @NataliaLKB! May become useful whenever/if we would like to fix our Lightbox behaviour on mobile (https://github.com/guardian/frontend/issues/9957 among many others...).

PR #11177, opened 2015-11-24
Identity - Add a new category with 3 new subscriptions.

mchv:
- Add new category panel - labelled "more from theguardian"
- Add 3 more emails into the preferences for this label

Asked by Aimee Macintosh and Jim Maddock

PR #11191, opened 2015-11-25
Revert "Identity - Add a new category with 3 new subscriptions."

mchv:
Reverts guardian/frontend#11177

Those 3 list ids seems to live in a different business unit of `ExactTarget`, which could not be set through identity API so this is not working.

cc @jennysivapalan


jennysivapalan:
Sad. :+1:

mchv:
@jennysivapalan if only I have been aware before..

Issue #11188, opened 2015-11-25
Firefox galleries: lightbox doesn't work


paperboyo:
Hello,

Windows 10, Firefox 43.0b6 & 45.0a1 and Android 4.1.2, Firefox 43.0b4. All other checked browsers unaffected.

Both gallery and article Lightboxes do not work correctly. Upon opening, console prints `TypeError: this.s is undefined` in `core.js`.

The image does not open fullscreen, just jumps to the top of the page. Page reload opens the Lightbox then, but navigation (arrow buttons, swipe, arrow keys) do not work at all. Close button works.

Regards
Mateusz

paperboyo:
Same thing on Mac 10.9.5 Firefox 42.

OliverJAsh:
Looking into this now. Tracking is broken on those pages ;o

On Wed, 25 Nov 2015 at 09:16 Mateusz <notifications@github.com> wrote:

> Same thing on Mac 10.9.5 Firefox 42.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/guardian/frontend/issues/11188#issuecomment-159544639>
> .
>

-- 

------------------------------
This e-mail and all attachments are confidential and may also be 
privileged. If you are not the named recipient, please notify the sender 
and delete the e-mail and all attachments immediately. Do not disclose the 
contents to another person. You may not use the information for any 
purpose, or store, or copy, it in any way.  Guardian News & Media Limited 
is not liable for any computer viruses or other material transmitted with 
or as part of this e-mail. You should employ virus checking software.
 
Guardian News & Media Limited is a member of Guardian Media Group plc. Registered 
Office: PO Box 68164, Kings Place, 90 York Way, London, N1P 2AP.  Registered 
in England Number 908396

paperboyo:
:beer:

OliverJAsh:
The reason for this is as follows: function declarations inside of try/catch statements do not get hoisted in Firefox, as per http://jsbin.com/wajuhi/3/edit?js,console,output. This will error in Firefox. Error introduced in https://github.com/guardian/frontend/pull/10452 on September 15.

This means that the Omniture script fails to run, however this error shouldn't be causing galleries to break. I've addressed that in #11192.

Oddly, the tracking for page views doesn't seem to have decreased, although we have seen a decrease in clicks:

![image](https://cloud.githubusercontent.com/assets/921609/11398007/a43924a4-9375-11e5-9643-9f266e7ac61b.png)

[...]

PR #11192, opened 2015-11-25
Catch tracking errors to allow normal behaviour

OliverJAsh:
[There is currently an error being thrown for tracking in Firefox](#11193). Although this exception is raised in the tracking code, it also breaks basic functionality, such as [opening a gallery](#11188) or [liveblog view more buttons](#11194). This is because `EventEmitter` will run all the event listeners synchronously, and the event listener that handles galleries is registered after the event listener for tracking.

I propose that we catch errors in the tracking event listener to allow other event listeners to continue.

I am going to fix the tracking error, but this would be nice to have so that if we mess up tracking, it doesn't break unrelated parts of the site.

### DIFF
--- a/static/src/javascripts/projects/common/modules/analytics/omniture.js
+++ b/static/src/javascripts/projects/common/modules/analytics/omniture.js
@@ -16,7 +16,8 @@ define([
     'lodash/objects/assign',
     'lodash/collections/forEach',
     'lodash/arrays/uniq',
-    'lodash/collections/map'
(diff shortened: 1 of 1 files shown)


paperboyo:
Thanks a lot! I was absolutely sure that Lightbox was working in FF just couple of days ago (definitely well after Sept 15th)... I may be wrong, though.

rich-nguyen:
:+1:

Issue #11194, opened 2015-11-25
Firefox liveblog: view more button doesn't work


OliverJAsh:
Caused by https://github.com/guardian/frontend/issues/11193.

OliverJAsh:
Fixed by #11195

PR #11195, opened 2015-11-25
Fix uglify settings

### LINKED ISSUE #11193: Firefox: Omniture script throws error
The reason for this is as follows: function declarations inside of try/catch statements do not get hoisted in Firefox, as per http://jsbin.com/wajuhi/3/edit?js,console,output. This will error in Firefox. The try catch was introduced in #10452 on September 15, but the error only surfaced when I broke the uglify settings in #11181. (Should be `compress: { evaluate: false }`)

This is the cause of #11188 and #11194

OliverJAsh:
Fixes #11193, #11188 and #11194

Broken by #11181. I got the uglify settings wrong! :unamused: 

This is causing #11193, because without `compress: { evaluate: false }`, the code looks like:

```
var s = s_gi();
function s_gi() {}
```

… which exhibits the Firefox bug described in #11193. With `compress: { evaluate: false }`, Uglify will output

```
function s_gi() {}
var s = s_gi();
```

… which, by chance, circumvents the Firefox bug by declaring the function before its usage.

/cc @sndrs @paperboyo

### DIFF
--- a/grunt-configs/uglify.js
+++ b/grunt-configs/uglify.js
@@ -33,8 +33,10 @@ module.exports = function (grunt, options) {
                 dest: 'common/conf/assets/vendor'
             }],
             options: options.isDev ? {} : {
-                // Set to false retain constant expressions, used to avoid writing HTML like </script>.
-                compress: false
+                compress: {
+                    // Set to false retain constant expressions, used to avoid writing HTML like </script>.
+                    evaluate: false
+                }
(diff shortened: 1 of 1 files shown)


rich-nguyen:
:+1:

paperboyo:
:laughing: Wow!

When it comes **just** to timing, that's more like it...

PR #11196, opened 2015-11-25
Move social buttons into links for A/B test

gtrufitt:
After discussing with @superfrank, Amy and Andrea, we decided for the A/B test the quickest way to ensure an uncluttered area around email sign-up was to move social buttons back into links. This is an interim decision until we can re-build the footer the ideal way.

image

Ping/ @crifmulholland @desbo

### DIFF
--- a/common/app/views/fragments/footer.scala.html
+++ b/common/app/views/fragments/footer.scala.html
@@ -47,16 +47,16 @@
                 @currentEdition match {
                     case Uk => {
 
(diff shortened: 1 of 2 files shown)


desbo:
:+1:

crifmulholland:
Fine with me. We can track how this affects things. They saw the bigger increase when we moved them, but still - it's the footer. We need to get them further up the page regardless.

gtrufitt:
Yup, Amy mentioned that there is non-insignificant clicks on these buttons, but as far as a minimal solution design-wise, that we're happy to roll out to 50% this is our best bet. :+1:

PR #11198, opened 2015-11-25
Inject ads in liveblogs every 5th post

regiskuckaertz:
This update addresses the UC where very long liveblogs would only see a handful of ads at the beginning and then no more. Now ads are injected every 5th post, even when live updates have arrived.

The process starts by finding the first post which ends below the viewport–this is the point where the insertion will begin. Ad slots are injected as soon as there's enough room for them, but ads themselves are lazy loaded when they fall inside the `[top - h/2, bottom + h/2]` where `top`, `bottom` and `h` are the position and height of the viewport, respectively.

A switch (`liveblogDynamicAdverts`) controls whether the module is triggered, and if so will take precedence over the current behaviour, itself hidden behind the `liveblogAdverts` switch.

### DIFF
--- a/common/app/conf/switches/CommercialSwitches.scala
+++ b/common/app/conf/switches/CommercialSwitches.scala
@@ -77,6 +77,15 @@ trait CommercialSwitches {
     exposeClientSide = true
   )
 
+  val LiveblogDynamicAdvertsSwitch = Switch(
+    "Commercial",
+    "liveblog-dynamic-adverts",
+    "Dynamically insert inline adverts on liveblogs",
+    safeState = Off,
+    sellByDate = new LocalDate(2015, 12, 14),
+    exposeClientSide = true
+  )
+
   val AudienceScienceSwitch = Switch(
     "Commercial",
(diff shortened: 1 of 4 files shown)


jbreckmckye:
I'm on holiday, so I don't really have time to review this PR in detail, but it sounds like we're implementing behaviour that already exists in the form of space-finder (and the proxy I've been working on, space-filler).

regiskuckaertz:
Oh, alright so let's wait until you come back and see how I could build upon your work to answer this use case. (Thanks for checking while you're away.)

johnduffell:
if this is as specific as every 5th post, we can work that out on the server side and maybe reduce the amount of jumping caused by ads

jbreckmckye:
We might need to sit on this work until the Health Team complete their work making the Live Blog more stable. They're thinking about changing the blog's format to do so, which may invalidate this work.

regiskuckaertz:
Closing because NextGen Health team is working on a revamp of liveblogs.

PR #11203, opened 2015-11-26
Check if we have a byline before replacing on Immersive template

sammorrisdesign-zz:
Before not having a byline was causing the page to crap out

### DIFF
--- a/common/app/views/fragments/standfirst.scala.html
+++ b/common/app/views/fragments/standfirst.scala.html
@@ -16,7 +16,7 @@
     @content.standfirst.map { s =>
         @defining(Edition(request)) { edition =>
             @withJsoup(BulletCleaner(
-                if(content.isImmersive && content.isArticle) {
+                if(content.isImmersive && content.isArticle && content.byline) {
                     if (s.indexOf(content.byline.get) != -1) {
                         ContributorLinks(s, content.contributors).toString()
                     } else {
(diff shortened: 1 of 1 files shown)


common/app/views/fragments/standfirst.scala.html
```diff
@@ -16,7 +16,7 @@
     @content.standfirst.map { s =>
         @defining(Edition(request)) { edition =>
             @withJsoup(BulletCleaner(
-                if(content.isImmersive && content.isArticle) {
```
johnduffell: assuming byline is an `Option`, it would possibly be clearer to 
```
content.byline.map { byline => 
 if (s.contains(byline)) {
   [..]
 } else {
   s + "<p>Words by" + ContributorLinks(byline, content.contributors) + "</p>"
 }
```
I'd also suggest to rename s to standfirst while if you're in there

sammorrisdesign: When I do that I get

```
type mismatch;
 found   : Option[String]
 required: String
```

Happy to change `s` to `standfirst` though

johnduffell: I actually realised why it didn't work - because of the if statement it wasn't sure if it was getting a string or an option[string]  but anyway ship it as it is! :+1:

PR #11205, opened 2015-11-26
Restore empty alt attribute to images on fronts

### LINKED ISSUE #11204: A11y - Add empty alt attributes to fronts images.
Currently the images on fronts do not have any alt attribute at all. 

This causes screen readers to read out the filename of the image:

image

While it is debatable if these images are 'purely decorative', if we're not able to add relevant info (ie: more than just repeating the headline in the alt attribute) then the recommend approach is to treat the images as 'decorative' and add an empty `alt=""` attribute so that screen readers ignore them.

> Purely decorative images must be marked up so they can be ignored by assistive technology with a null alt attribute (alt="") or preferably use CSS techniques. If the image isn't providing the user any informative content or enhancing greater understanding of the content, then the alt attribute must be empty.

http://dev.w3.org/html5/alt-techniques/#sec5

sndrs:
fixes #11204

### DIFF
--- a/common/app/views/fragments/items/elements/facia_cards/image.scala.html
+++ b/common/app/views/fragments/items/elements/facia_cards/image.scala.html
@@ -1,6 +1,7 @@
 @(classes: Seq[String], srcset: String, sizes: String, src: Option[String] = None)
 
 <img class="@RenderClasses(classes: _*)"
+    alt=""
     srcset="@srcset"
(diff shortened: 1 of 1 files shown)


gtrufitt:
:+1:

johnduffell:
:+1: the beauty of dotcom.feedback!

PR #11207, opened 2015-11-26
Get rid of percentage units in srcset sizes

desbo:
This gets rid of percentages in our `sizes` attribute and replaces it with `vw` in line with the [spec](https://html.spec.whatwg.org/multipage/embedded-content.html#source-size-value).

We only use a percentage for `fc-item--standard-mobile` images, which work out to 95% of the viewport width, so it's a simple fix to make.

@johnduffell @sndrs

### DIFF
--- a/common/app/layout/BrowserWidth.scala
+++ b/common/app/layout/BrowserWidth.scala
@@ -4,6 +4,7 @@ object BrowserWidth {
   implicit class RichInt(n: Int) {
     def px = PixelWidth(n)
     def perc = PercentageWidth(n)
+    def vw = ViewportWidth(n)
   }
 }
 
@@ -17,4 +18,8 @@ case class PercentageWidth(get: Int) extends BrowserWidth {
 
 case class PixelWidth(get: Int) extends BrowserWidth {
   override def toString = s"${get}px"
+}
+
+case class ViewportWidth(get: Int) extends BrowserWidth {
+  override def toString = s"${get}vw"
 }
--- a/common/app/layout/WidthsByBreakpoint.scala
+++ b/common/app/layout/WidthsByBreakpoint.scala
@@ -7,7 +7,7 @@ import views.support.Profile
 object FaciaWidths {
   private val MediaMobile = Map[CardType, BrowserWidth](
     (MediaList, 127.px),
-    (Standard, 100.perc)
+    (Standard, 95.vw)
   )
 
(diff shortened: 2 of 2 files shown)


sndrs:
great :+1:

OliverJAsh:
`vw` isn't supported in IE 8, Opera Mini, and some old Android browsers. Unless Picturefill handles that for us? http://caniuse.com/#search=vw

sndrs:
@OliverJAsh

`vw` is better supported than `srcset`, so this should be fine shouldn't it?

`vw`
screen shot 2015-11-26 at 14 45 16

`srcset`
screen shot 2015-11-26 at 14 45 40

desbo:
I think picturefill takes care of it in this function: https://github.com/scottjehl/picturefill/blob/master/src/picturefill.js#L258

I doubt browser support for `vw` will come into play since we only ever use it as part of the `sizes` attribute, and that'll either be handled by picturefill or native picture support (which implies `vw`).

I'm gonna check in IE8 just to be sure

sndrs:
ah do you mean that picturefill could pick `95vw` as the appropriate size?

sndrs:
IE8 doesn't support media queries, it'll never render a `fc-item--standard-mobile`

OliverJAsh:
We have a polyfill for `srcset`, so I'm not concerned about that—just wondering how this will play out for browsers that don't support `vw`.

desbo:
no problems as far as i can tell 
screen shot 2015-11-26 at 15 37 58
screen shot 2015-11-26 at 15 36 17

desbo:
It's fine in Opera Mini as well, so picturefill must be taking care of it. 
@OliverJAsh, let me know if you have any other thoughts otherwise I'll merge

OliverJAsh:
Great, thanks for checking! :+1:

On Thu, 26 Nov 2015 at 16:01 Sam Desborough <notifications@github.com>
wrote:

> It's fine in Opera Mini as well, so picturefill must be taking care of it.
> @OliverJAsh <https://github.com/OliverJAsh>, let me know if you have any
> other thoughts otherwise I'll merge
>
> —
> <https://github.com/guardian/frontend/pull/11207#issuecomment-159948753>.
>

-- 

in England Number 908396

johnduffell:
that's quick work! well done :+1:

PR #11208, opened 2015-11-26
For global development, making section the first tab in most pop

NataliaLKB:
Global development doesn't have many chances to show case their content. Their readers look for more depth rather than breadth when looking for an onward journey.

cc @johnduffell

### DIFF
--- a/onward/app/controllers/MostPopularController.scala
+++ b/onward/app/controllers/MostPopularController.scala
@@ -31,7 +31,10 @@ object MostPopularController extends Controller with Logging with ExecutionConte
     val sectionPopular: Future[List[MostPopular]] = if (path.nonEmpty) lookup(edition, path).map(_.toList) else Future(Nil)
 
     sectionPopular.map { sectionPopular =>
-      val mostPopular = globalPopular.toList ++ sectionPopular
+      val sectionFirst = sectionPopular ++ globalPopular
+      val globalFirst = globalPopular.toList ++ sectionPopular
+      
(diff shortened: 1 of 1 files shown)


onward/app/controllers/MostPopularController.scala
```diff
@@ -31,7 +31,10 @@ object MostPopularController extends Controller with Logging with ExecutionConte
     val sectionPopular: Future[List[MostPopular]] = if (path.nonEmpty) lookup(edition, path).map(_.toList) else Future(Nil)
 
     sectionPopular.map { sectionPopular =>
-      val mostPopular = globalPopular.toList ++ sectionPopular
+      val sectionFirst = sectionPopular ++ globalPopular
```
Jholder112233: Do you need to .toList on this line too on the globalPopular? (just guessing here)

NataliaLKB: You do. It is an option, not a list.

Jholder112233:
:+1:

onward/app/controllers/MostPopularController.scala
```diff
@@ -31,7 +31,10 @@ object MostPopularController extends Controller with Logging with ExecutionConte
...
+      val globalFirst = globalPopular.toList ++ sectionPopular
+      
+      val mostPopular = if ( path.contains("global-development") ) sectionFirst else globalFirst
```
johnduffell: consider making it a startsWith so it's more specific

johnduffell:
:+1: looks good!

PR #11210, opened 2015-11-26
Bugs and tweaks to Immersive template

sammorrisdesign-zz:
- Hides the section rule if it follows an immersive image
- Cleans up some of the pull quote styling
- Forces ads to remaining static to stop broken and unwanted sticky ad behaviour

### DIFF
--- a/static/src/stylesheets/module/content/_article-immersive.scss
+++ b/static/src/stylesheets/module/content/_article-immersive.scss
@@ -553,6 +553,10 @@
                 margin-right: $gs-gutter;
             }
         }
+
+        + .section-rule {
+            display: none;
+        }
     }
 
     figure.element--showcase {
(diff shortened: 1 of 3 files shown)


static/src/stylesheets/module/content/_article-immersive.scss
```diff
@@ -609,4 +620,8 @@
             margin-right: -(gs-span(5) + $gs-gutter);
         }
     }
+
+    .ad-slot--right {
+        position: static !important; // To override sticky ad
```
gtrufitt: I wonder whether it would make more sense to add a specific class to page types that would require this override (and have that managed in one place) rather than starting to add !important's inside specific page type CSS files?

sammorrisdesign:
@gtrufitt Have looked into a js solution and removed the `!important`

gtrufitt:
Good stuff :+1:

PR #11212, opened 2015-11-26
Remodel Content, MetaData and Page

rich-nguyen:
This is a big pull request, it contains preparatory changes for the upcoming Content API upgrade. I couldn't sensibly break this change down, being such a fundamental part of how we render everything.

The aim is to replace the old trait-based model objects with equivalent case classes that contain the same data. Crucially, these new case classes will be easily serialisable, and completely standalone. There are two main benefits from doing all this work:
1. Our frontend model has no external dependencies on other guardian libraries. This simplifies upgrading those libraries, because pressed pages won't break every time an external class changes. In future, where Content API classes are not be serialisable, frontend will be unaffected.
2. The Content system is flat. This means less overriding, so it should be easier to track down data. The existing trait mixin system is fine, but it doesn't scale anymore, it's really hard to understand where data comes from. In this PR, the transformation work happens in one place, [content.scala](common/app/model/content.scala). Traits are still used in places where it makes sense to keep them; that is `Page`, and `ContentType`.
## Testing Checklist

Lots to test in this one, @ScottPainterGNM ! The html should be consistent with how things are in master (when i say "javascript metadata" below, I mean `guardian.config.page`). Would appreciate some help in covering all of these:
- [ ] Commercial - javascript metadata, 
- [ ] /all pages
- [ ] section and tag pages - javascript metadata, pagination should be intact
- [ ] Crossword/sudoku - crossword data should be identical on page
- [ ] Video embed pages - should look the same
- [ ] Gallery, Picture pages and lightbox system should be identical
- [ ] Interactives - pages should boot as before
- [ ] Audio/Video - should have identical javascript metadata, endslate html should be the same too.
- [ ] Offline page - the offline crossword should be unaffected by these changes
- [ ] Contributors and subject listing pages should be identical
- [ ] Preferences page should be identical
- [ ] Article furniture should be present and correct
- [ ] Facia cards should look the same, they have the same css classes as before. Check fronts for visual differences.
- [ ] Commercial components need testing, along with checking the Commercial agents are ok. Will chat to @kenlim 
- [ ] Commercial adsuffix handling should pass the same values as before. Will check with commercial team.
- [ ] Commercial keyword sponsorship system should be intact.
- [ ] Commercial Outbrain container should be placed correctly. Test passes, so check visual placement is ok.
- [ ] Pressed page commercial treatment should exhibit the same behaviour (it uses the DFPAgent, check this works as expected).
- [ ] Navigation - html should be the same
- [ ] RSS page should still be inferred from trails as before, check the content is correct
- [ ] Related Content should appear correct for articles without story package.
- [ ] Live blogs - javascript metadata and blocks should look the same.
- [ ] Analytics - check analytics beacon is firing consistently, and also check the javascript config has the right analytics name for all page types.
- [ ] ShareLinks system should work for lightboxes and page meta, with valid hrefs for each share button.
- [ ] Sport content pages should still have the right tabbing, check football match reports.
- [ ] Tag combiner pages should still render index-like pages.
- [ ] Identity app profile pages should be html identical.
## Changelist

Engineer-oriented description of the changes in this pull request, and a bit of background.
- prefer `Page` to `extends MetaData`. It is no longer possible to override `MetaData` because it is no longer a trait. If you want to change the values of a `MetaData` object, create one, and pass in the values you desire. There are helpers in the `MetaData` companion object, which resemble the signature of the old override-style of doing things. This change makes `MetaData` easily serialisable.
- prefer copy constructor to override. In the few cases where values do need to be overridden, copy can be used to modify values. A good example is `Article` in [content.scala](common/app/model/content.scala), which is a kind of `ContentType`. An `Article` shares a lot of similarities with vanilla `GenericContent`, but overrides a few things by using the copy method.
- prefer composition to `extends`. The old Content trait hierarchy was deeply nested and contained a few multiple inheritances. I've removed traits like `Trail` and `Elements` and replaced them with case classes of the same name. These all get collected together into a class called `Content`, which contains all the data from Content API that frontend needs to render things. Next, the `ContentType` trait contains one of these `Content` objects, to store it's data. Sub-classes of `ContentType` include `Article`, `Video`, `Gallery`, `Interactive`. These are objects that frontend can render pages from.
- A `Page` is anything that contains `MetaData`, and can therefore be rendered as a html page. There are two main kinds of `Page` here, `StandalonePage` for pages that are not based on a single Content API item, and `ContentPage` for CAPI-backed pages like articles.
- The `LiveBlog` class no longer exists, it is simply an instance of `Article` where `isLiveBlog` is true. It's treated in the same way as Immersive; article-like content with little underlying data variation (it's mostly styling and html structural differences).
- We used to refer to the map of javascript key-value pairs as 'metadata', this is a bit confusing so i've called it 'javascriptConfig'.


applications/app/controllers/AllIndexController.scala
```diff
@@ -55,9 +55,9 @@ object AllIndexController extends Controller with ExecutionContexts with ItemRes
 
     loadLatest(path, reqDate).map { _.map { index =>
 
-      val contentOnRequestedDate = index.trails.filter(_.webPublicationDate.sameDay(reqDate))
+      val contentOnRequestedDate = index.contents.filter(_.item.trail.webPublicationDate.sameDay(reqDate))
 
-      val olderDate = index.trails.find(!_.webPublicationDate.sameDay(reqDate)).map(_.webPublicationDate.toDateTime)
+      val olderDate = index.trails.find(!_.trail.webPublicationDate.sameDay(reqDate)).map(_.trail.webPublicationDate.toDateTime)
```
jennysivapalan: What's `trail` and what is the difference between it and `content`?

applications/app/crosswords/CrosswordGridModel.scala
```diff
@@ -1,6 +1,7 @@
 package crosswords
 
 import com.gu.contentapi.client.model.{CrosswordEntry, CrosswordDimensions, Crossword, CrosswordPosition}
+import model.Entry
```
jennysivapalan: Is this needed?

jennysivapalan:
Pheww big PR. I've read the notes and looked at 1/4 of the code change. I assume its consistent through out so giving a :+1: for this

johnduffell:
like Jenny I've not looked at everthing, but overall I really like the idea.
Just my couple of comments:
as discussed it would be great to have descriptions of the more abstractly named classes.
I noticed the `make` method makes a lot of appearances, I guess it's a convention you have for an alternative constructor where apply wouldn't be appropriate?

common/app/views/fragments/head.scala.html
```diff
@@ -1,16 +1,13 @@
...
 <meta charset="utf-8" />
-<title>@views.support.Title(metaData)</title>
+<title>@views.support.Title(page)</title>
 
-@fragments.metaData(metaData)
+@fragments.metaData(page)
 @head
 
 @* stylesheet <link>s - get the stylesheets downloading ASAP *@
 @fragments.stylesheets(projectName)
```
desbo: I'm not sure how `model.Page` differs from `model.metaData` but, just as a heads up, we've just started passing `metaData.isCrosswords` through to stylesheets (https://github.com/guardian/frontend/pull/11219/files#diff-0983be310fb44a148d07ad3af46e62f8R13), will that still be possible?

applications/app/controllers/AllIndexController.scala
rich-nguyen: Good question. A `trail` used to be a trait, defining all the fields needed to render a html trail link. Now, it is simply part of `content`, although still represents the same data fields.

applications/app/crosswords/CrosswordGridModel.scala
rich-nguyen: Yes, it got moved from applications to common

common/app/views/fragments/head.scala.html
rich-nguyen: Thanks, I'll fix this in the next merge

common/app/model/meta.scala
```diff
@@ -295,8 +508,9 @@ trait Elements {
 /**
  * Tags lets you extract meaning from tags on a page.
  */
-trait Tags {
-  def tags: Seq[Tag] = Nil
+final case class Tags(
+  tags: Seq[Tag]) {
```
kelvin-chappell: Perhaps this val should be called `all` or something like that, to distinguish it from a particular subset of tags?

kelvin-chappell:
Nice cleanup. Has been needed for some time even if it's not for capi changes. :+1: 
I haven't tried running it!

common/app/model/meta.scala
rich-nguyen: yeah that would be sensible, would reduce the number of `tags.tags` lines. There's plenty of other tidy up too, (like the filenames in the model folder), but I can't make this PR any bigger!

ScottPainterGNM:
QA :+1:

PR #11213, opened 2015-11-27
Enable top-above-nav ad slot to show full set of ad sizes

kelvin-chappell:
Lifting restriction on sizes in business front because it was having unexpected effect for ad ops.


johnduffell:
was this to try to fix the page jumping when the ads came in?  if so it's a shame it's gone away, but the code looks good :+1:

kelvin-chappell:
The page should still stop jumping in most cases, but this is just in response to a problem where a campaign wasn't being served as expected.  This is a feature that needs a bit more work.

PR #11214, opened 2015-11-27
extend quiz scores switch

johnduffell:
We want to turn off the legacy quiz scores service as it's part of the prototype.  tools team are still working on a quizzes page to replace the hacked together quiz builder/app.  This is still in progress.  There's still value provided by the scores service and it's not causing any effort expended other than extending it occasionally, so we may as well keep it on.


jennysivapalan:
:question: :accept: :+1:

johnduffell:
:confused: :stuck_out_tongue:

PR #11219, opened 2015-11-27
Disable async CSS for crossword pages

desbo:
We have a bug with the `async-css` feature that stops anyone on Safari 6 or below from downloading stylesheets. One of the effects is unusable crosswords: 

unnamed

We've understandably had a lot of complaints about this so this is a temporary hack to disable async CSS for individual crossword pages until we fix the wider issue.

### DIFF
--- a/common/app/views/fragments/stylesheets.scala.html
+++ b/common/app/views/fragments/stylesheets.scala.html
@@ -1,4 +1,4 @@
-@(projectName: Option[String])(implicit request: RequestHeader)
+@(projectName: Option[String], isCrossword: Boolean = false)(implicit request: RequestHeader)
 
 @import conf.switches.Switches._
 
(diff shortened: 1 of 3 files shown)


common/app/views/fragments/stylesheets.scala.html
```diff
@@ -58,8 +58,8 @@
   - Exclude IE Mobile [&(!IEMobile)]
 *@
 <!--[if (IE 9)&(!IEMobile)]>
-    @fragments.stylesheetLink(Static.css.headIE9(projectName))
-    @fragments.stylesheetLink("stylesheets/ie9.content.css")
+    @fragments.stylesheetLink(Static.css.headIE9(projectName), isCrossword)
```
gtrufitt: Not a major deal but I guess you don't need to disable this on `IE9` or `lt IE 9` if we're not seeing the issue in those browsers and keep the performance advantage there?

gtrufitt:
:+1:  other than my minor comment

common/app/views/fragments/stylesheets.scala.html
desbo: I don't know for certain that it doesn't affect those browsers, I think it's better to apply the rule consistently for crossword pages.

gtrufitt: Fair enough :+1:

sndrs:
:+1: this is nasty isn't it. good move though for now :+1:

PR #11220, opened 2015-11-27
Update list id

gtrufitt:
@crifmulholland @desbo Updated ID to list 37, assume this is the only place this is hardcoded currently?


desbo:
Yeah it's the only place, you'd probably be better off adding a new `val` to `listIds` with a more descriptive name though.

gtrufitt:
> Yeah it's the only place, you'd probably be better off adding a new val to listIds with a more descriptive name though.

Ta, done!

desbo:
:+1:

PR #11221, opened 2015-11-27
Enable fluid ads

regiskuckaertz:
This change will enable fluid ads for the top banner only. It is hidden behind the `FluidAdverts` switch.
### SASS

I created a `.ad-slot--fluid` modifier but because currently top banner ads have a fixed width and left padding on desktop, I had to compound classes to cancel it. There may be a more idiomatic way of doing this at the Guardian though.

### DIFF
--- a/common/app/conf/switches/CommercialSwitches.scala
+++ b/common/app/conf/switches/CommercialSwitches.scala
@@ -32,6 +32,15 @@ trait CommercialSwitches {
     exposeClientSide = true
   )
 
+  val FluidAdvertsSwitch = Switch(
+    "Commercial",
+    "fluid-adverts",
+    "Enable fluid ads",
+    safeState = Off,
+    sellByDate = never,
+    exposeClientSide = true
+  )
+
   val CommercialComponentsSwitch = Switch(
     "Commercial",
     "commercial-components",
--- a/common/app/conf/switches/CommercialSwitches.scala
+++ b/common/app/conf/switches/CommercialSwitches.scala
@@ -35,9 +35,9 @@ trait CommercialSwitches {
   val FluidAdvertsSwitch = Switch(
     "Commercial",
     "fluid-adverts",
-    "Enable fluid ads",
+    "Enable fluid ads, which occupy 100% of the width of their parent container but have a fixed height",
     safeState = Off,
-    sellByDate = never,
+    sellByDate = new LocalDate(2015, 12, 15),
     exposeClientSide = true
   )
 
--- a/static/src/javascripts/projects/common/modules/commercial/dfp-api.js
+++ b/static/src/javascripts/projects/common/modules/commercial/dfp-api.js
@@ -415,7 +415,7 @@ define([
             } else if ($adSlot.data('fluid') && cookies.get('adtest') === 'tm2') {
                 $adSlot.addClass('ad-slot--fluid');
                 sizeMapping = defineSlotSizes($adSlot);
-                // regis kuckaertz (Nov 30) – SizeMappingBuilder does not handle 'fluid' very well,
+                // SizeMappingBuilder does not handle 'fluid' very well,
                 // so instead we add it manually ourselves to the end of each array of sizes
                 forEach(sizeMapping, function (sizeMap) { sizeMap[1].push('fluid'); });
                 slot = googletag.defineSlot(adUnit, 'fluid', id).defineSizeMapping(sizeMapping);
--- a/static/src/javascripts/projects/common/modules/commercial/dfp-api.js
+++ b/static/src/javascripts/projects/common/modules/commercial/dfp-api.js
@@ -404,24 +404,24 @@ define([
                 adUnit         = adUnitOverride ?
                     ['/', config.page.dfpAccountId, '/', adUnitOverride].join('') : config.page.adUnit,
                 id             = $adSlot.attr('id'),
-                sizeMapping    = defineSlotSizes($adSlot),
+                slot,
+                size,
+                sizeMapping;
+
+            if ($adSlot.data('out-of-page')) {
+                slot = googletag.defineOutOfPageSlot(adUnit, id);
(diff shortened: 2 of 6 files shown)


johnduffell:
not sure what fluid ads are, but is this likely to cause movement of the page after it's loaded?

regiskuckaertz:
@johnduffell it is meant to be fluid in the sense that it occupies 100% of the viewport width—the same height is kept throughout to prevent layout thrashing. At the moment though, there is a nasty 'appear' effect that animates on the height and I'm still trying to figure out where it comes from.

common/app/conf/switches/CommercialSwitches.scala
```diff
@@ -32,6 +32,15 @@ trait CommercialSwitches {
     exposeClientSide = true
   )
 
+  val FluidAdvertsSwitch = Switch(
+    "Commercial",
+    "fluid-adverts",
+    "Enable fluid ads",
```
johnduffell: might be worth putting a bt more explanation here

johnduffell:
sounds good, jump (and jank) are always painful for users! :+1: although I don't know much about the dfp-api.js code

common/app/conf/switches/CommercialSwitches.scala
```diff
@@ -32,6 +32,15 @@ trait CommercialSwitches {
...
+    safeState = Off,
+    sellByDate = never,
```
jbreckmckye: Never? Never ever... ever?

static/src/javascripts/projects/common/modules/commercial/dfp-api.js
```diff
@@ -404,24 +406,28 @@ define([
...
+                slot,
+                size,
+                sizeMapping;
+
+            if ($adSlot.data('out-of-page')) {
+                slot = googletag.defineOutOfPageSlot(adUnit, id);
+            } else if ($adSlot.data('fluid') && cookies.get('adtest') === 'tm2') {
+                $adSlot.addClass('ad-slot--fluid');
+                sizeMapping = defineSlotSizes($adSlot);
+                // regis kuckaertz (Nov 30) – SizeMappingBuilder does not handle 'fluid' very well,
```
jbreckmckye: You don't need to annotate your name and date of authorship - we already have that information in `git blame`.

static/src/javascripts/projects/common/modules/commercial/dfp-api.js
```diff
@@ -404,24 +406,28 @@ define([
                 adUnit         = adUnitOverride ?
                     ['/', config.page.dfpAccountId, '/', adUnitOverride].join('') : config.page.adUnit,
                 id             = $adSlot.attr('id'),
-                sizeMapping    = defineSlotSizes($adSlot),
+                slot,
+                size,
+                sizeMapping;
+
+            if ($adSlot.data('out-of-page')) {
+                slot = googletag.defineOutOfPageSlot(adUnit, id);
```
jbreckmckye: Does this branch miss out on the defineSizeMapping step?

static/src/javascripts/projects/common/modules/commercial/dfp-api.js
```diff
@@ -404,24 +406,28 @@ define([
...
+            } else if ($adSlot.data('fluid') && cookies.get('adtest') === 'tm2') {
```
jbreckmckye: We're not going to have adtest cookies on for the most part in production - is this just for internal demonstrations? Is this actually going to be released to production?

static/src/javascripts/projects/common/modules/commercial/dfp-api.js
```diff
@@ -404,24 +406,28 @@ define([
...
+                $adSlot.addClass('ad-slot--fluid');
+                sizeMapping = defineSlotSizes($adSlot);
+                // regis kuckaertz (Nov 30) – SizeMappingBuilder does not handle 'fluid' very well,
+                // so instead we add it manually ourselves to the end of each array of sizes
```
jbreckmckye: So are we circumventing the normal API here? Are we actually supposed to manipulate these sizeMap objects?

static/src/javascripts/projects/common/modules/commercial/dfp-api.js
```diff
@@ -404,24 +406,28 @@ define([
...
-                slot = (
-                    $adSlot.data('out-of-page') ?
-                        googletag.defineOutOfPageSlot(adUnit, id) :
-                        googletag.defineSlot(adUnit, size, id)
-                    )
-                    .addService(googletag.pubads())
-                    .defineSizeMapping(sizeMapping)
-                    .setTargeting('slot', slotTarget);
+                size = uniq(
+                    flatten(sizeMapping, true, function (map) { return map[1]; }),
```
jbreckmckye: We don't normally collapse lambdas.

common/app/conf/switches/CommercialSwitches.scala
regiskuckaertz: good call, I'm on it.

regiskuckaertz:
Thanks!

static/src/javascripts/projects/common/modules/commercial/dfp-api.js
regiskuckaertz: this is currently a beta feature of DFP and we are testing it.

static/src/javascripts/projects/common/modules/commercial/dfp-api.js
regiskuckaertz: SizeMappingBuilder is just a helper, but I'm not circumventing it. Instead, I'm building on it by injecting the 'fluid' size at the end.

common/app/conf/switches/CommercialSwitches.scala
regiskuckaertz: good call, I'll set a date in the near future.

PR #11222, opened 2015-11-27
Add progress bar to the immersive articles

sammorrisdesign-zz:
Adds a progress bar to the immersive template for those mammoth long reads. I've tried to build it in a way that means we can use it on other articles if we choose to later.

screen recording 2015-11-27 at 03 51 pm

### DIFF
--- a/static/src/javascripts/projects/common/modules/article/progress.js
+++ b/static/src/javascripts/projects/common/modules/article/progress.js
@@ -5,13 +5,17 @@ define([
     $,
     mediator
 ) {
+    var articleHeight;
+
     return {
         init: function () {
+            this.getArticleHeight();
             this.bindEvents();
         },
 
         bindEvents: function () {
             mediator.on('window:throttledScroll', this.updateProgress.bind(this));
+            mediator.on('window:resize', this.getArticleHeight.bind(this));
         },
 
         updateProgress: function () {
@@ -19,7 +23,11 @@ define([
         },
 
         getProgressAsPercentage: function () {
-            return window.scrollY + "px"
+            return window.scrollY / articleHeight * 100 + "%"
+        },
+
+        getArticleHeight: function () {
+            articleHeight = $('.content--article').offset().height;
         }
     };
 });
--- a/static/src/javascripts/projects/common/modules/article/progress.js
+++ b/static/src/javascripts/projects/common/modules/article/progress.js
@@ -8,15 +8,18 @@ define([
     return {
         init: function () {
             this.bindEvents();
-            console.log("progress");
         },
 
         bindEvents: function () {
-            mediator.on('window:throttledScroll', this.updateProgress);
+            mediator.on('window:throttledScroll', this.updateProgress.bind(this));
         },
 
         updateProgress: function () {
-            console.log("hey");
+            $('.progress__indicator').css('width', this.getProgressAsPercentage);
+        },
+
+        getProgressAsPercentage: function () {
+            return window.scrollY + "px"
         }
     };
 });
--- a/static/src/javascripts/projects/common/modules/article/progress.js
+++ b/static/src/javascripts/projects/common/modules/article/progress.js
@@ -9,8 +9,11 @@ define([
 
     return {
         init: function () {
-            this.getArticleHeight();
-            this.bindEvents();
+            // Check if progress bar is present
+            if ($('.progress').length) {
+                this.getArticleHeight();
+                this.bindEvents();
+            }
         },
 
         bindEvents: function () {
--- a/static/src/javascripts/projects/common/modules/article/progress.js
+++ b/static/src/javascripts/projects/common/modules/article/progress.js
@@ -10,7 +10,7 @@ define([
     return {
         init: function () {
             // Check if progress bar is present
-            if ($('.progress').length) {
+            if ($('.js-progress').length) {
                 this.getArticleHeight();
                 this.bindEvents();
             }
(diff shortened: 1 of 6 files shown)


static/src/javascripts/projects/common/modules/article/progress.js
```diff
@@ -0,0 +1,36 @@
...
+], function (
+    $,
+    mediator
+) {
+    var articleHeight;
+
+    return {
+        init: function () {
+            // Check if progress bar is present
+            if ($('.progress').length) {
```
jbreckmckye: I'm wary of using classes as springboards for scripted functionality. It's not obvious, looking at the template, how the component is instantiated, or that refactoring the styles will break the functionality. It also makes it awkward to re-use the behaviour without re-using the styles.

I would use a `js-progress-bar` class or the like, instead, in the same manner as existing components like `js-site-message`, `js-sport-tabs`, etc. etc.

static/src/stylesheets/module/_progress.scss
```diff
@@ -0,0 +1,13 @@
+.progress {
+    position: fixed;
+    z-index: 1;
```
jbreckmckye: What's the z-index for?

static/src/javascripts/projects/common/modules/article/progress.js
```diff
@@ -0,0 +1,36 @@
...
+        updateProgress: function () {
+            $('.progress__indicator').css('width', this.getProgressAsPercentage);
+        },
+
+        getProgressAsPercentage: function () {
+            return window.scrollY / articleHeight * 100 + '%';
+        },
+
+        getArticleHeight: function () {
+            articleHeight = $('.content--article').offset().height;
```
jbreckmckye: Can article heights mutate after this getter has been called? e.g. if adverts load?

static/src/javascripts/projects/common/modules/article/progress.js
```diff
@@ -0,0 +1,36 @@
...
+            }
+        },
+
+        bindEvents: function () {
+            mediator.on('window:throttledScroll', this.updateProgress.bind(this));
+            mediator.on('window:resize', this.getArticleHeight.bind(this));
+        },
+
+        updateProgress: function () {
+            $('.progress__indicator').css('width', this.getProgressAsPercentage);
```
jbreckmckye: If you had these as just private functions, you wouldn't need to do any `this` binding malarkey.

static/src/javascripts/projects/common/modules/article/progress.js
```diff
@@ -0,0 +1,36 @@
...
+    mediator
+) {
+    var articleHeight;
+
+    return {
+        init: function () {
+            // Check if progress bar is present
+            if ($('.progress').length) {
+                this.getArticleHeight();
+                this.bindEvents();
```
jbreckmckye: We're assuming the element is initialized to empty by its style, which seems a bit... couply, maybe? I guess it's okay, but it would be nice to be able to reason about the lifecycle of the element just by looking at one script.

static/src/stylesheets/module/_progress.scss
```diff
@@ -0,0 +1,13 @@
...
+    height: $gs-baseline / 4;
```
jbreckmckye: So, does this element intercept click / touch events even when it's empty? Maybe it should start out with pointer-events:none or something.

static/src/stylesheets/module/_progress.scss
sammorrisdesign: Without this is appears behind the header and text

sndrs:
this gives me the fear a bit. how clear it is that you're on a special page vs just a page on the guardian? we know, but is it just introducing a mechanism that will appear to be sporadic?

my greater concern is that it's adding processing overhead to the browser when there is a working progress indicator built into the browser UI on another other axis. it seems hard to me to justify the extra bytes and load.

my initial reaction is a :-1: – do we have a problem with people not knowing how far they've read, or not getting the end etc?

sammorrisdesign:
@sndrs The built in scrollbars aren't a good indicator because of all the gubbins we put into the footer. I find it unusable as for the shorter articles, over half of the page could be just footer, comments, related, etc... So I no longer use it on any page anymore.

There are also plans in future to add support for chapters to it, so the user knows how far they're through before they have the opportunity to take a break through that 10,000 word piece.

static/src/stylesheets/module/_progress.scss
sammorrisdesign: Had added a `pointer-events` property here

sammorrisdesign:
@sndrs But, after thinking about the bytes and load thing, I think this would make sense to limit to non-core users. Is there an easy way to switch between core and premium (or whatever it's called) when developing?

uplne:
I don't want to discourage you but I agree with @sndrs Is there a real need from our users for this feature? Does user really need to see how far he/she is in the article?

I would probably see more value in marking/saving the position for the later read which works great on apps for ebooks. And it can be win for membership as well.

static/src/stylesheets/module/_progress.scss
jbreckmckye: Is there a z-index variable we can pull in here? I believe we have some defined for each 'plane' on the page.

jbreckmckye:
One thing to be aware of, too, is that an article's size can mutate whilst the user is reading it, because we have 'inread' adverts that expand as the user scrolls them into view, and because we have interactives that initialize at zero height. Your solution handles that implicitly, but it could mean that every so often the user will perform a small scroll downwards and actually see the bar regress.

PR #11223, opened 2015-11-30
Onward Journeys with Amp List

NataliaLKB:
First iteration on onward journeys in AMP. I thought that most popular would be the easiest thing to test the new `amp-list` component out. Here is what it looks like so far:

image

cc @johnduffell @stephanfowler

### DIFF
--- a/common/app/views/mainAMP.scala.html
+++ b/common/app/views/mainAMP.scala.html
@@ -15,7 +15,7 @@
         <script custom-element="amp-twitter" src="https://cdn.ampproject.org/v0/amp-twitter-0.1.js" async></script>
         <script custom-element="amp-youtube" src="https://cdn.ampproject.org/v0/amp-youtube-0.1.js" async></script>
         <script custom-element="amp-instagram" src="https://cdn.ampproject.org/v0/amp-instagram-0.1.js" async></script>
-        <script custom-template="amp-list" src="https://cdn.ampproject.org/v0/amp-list-0.1.js" async ></script>
+        <script custom-element="amp-list" src="https://cdn.ampproject.org/v0/amp-list-0.1.js" async ></script>
         <script custom-template="amp-mustache" src="https://cdn.ampproject.org/v0/amp-mustache-0.1.js" async ></script>
         <style>body {opacity: 0}</style><noscript><style>body {opacity: 1}</style></noscript>
         <script src="https://cdn.ampproject.org/v0.js" async></script>
@@ -42,15 +42,21 @@
 
             @body
 
-            <amp-list width="600" height="600" layout="responsive" src="https://www.thegulocal.com//most-read-amp.json">
-                <template type="amp-mustache">
-                    <div class="headline-list__item">
-                        <h2 class="fc-item__title">
-                            <a href="{{url}}">{{headline}}</a>
-                        </h2>
-                    </div>
-                </template>
-            </amp-list>
+            <div class="fc-container__inner">
+                <div class="fc-container__header__title">
+                    <span>popular</span>
+                </div>
+                <amp-list width="600" height="1000" layout="responsive" src="@Configuration.ajax.url/most-read-amp.json">
+                    <template type="amp-mustache">
+                        <div class="headline-list__item">
+                            <p class="headline-list__count">{{index}}</p>
+                            <h2 class="fc-item__title">
+                                <a href="{{url}}" class="fc-item__link">{{headline}}</a>
+                            </h2>
+                        </div>
+                    </template>
+                </amp-list>
+            </div>
 
             @fragments.footerAMP(metaData)
         </div>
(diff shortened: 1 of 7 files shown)


common/app/views/mainAMP.scala.html
```diff
@@ -40,6 +42,22 @@
 
             @body
 
+            <div class="fc-container__inner">
+                <div class="fc-container__header__title">
+                    <span>popular</span>
+                </div>
+                <amp-list width="600" height="1000" layout="responsive" src="https://www.thegulocal.com//most-read-amp.json">
```
gidsg: I guess this local URL will need updating? Plus double `//`

NataliaLKB: Thanks @gidsg. updated, but unsure of whether it is the correct page still, as I just make it relative.

dev-build/conf/routes
```diff
@@ -92,6 +92,7 @@ GET            /most-read-facebook.json
 GET            /most-read-twitter.json                                                                                           controllers.MostViewedSocialController.renderMostViewed(socialContext: String = "twitter")
 GET            /most-read.json                                                                                                   controllers.MostPopularController.render(path = "")
 GET            /most-read/*path.json                                                                                             controllers.MostPopularController.render(path)
+GET            /most-read-amp.json                                                                                               controllers.MostPopularController.renderPopularAmp()
```
stephanfowler: Thinking that these endpoints (there'll be more of them) should be named by format, rather than by client. For example: `/most-read-schema.json`

common/app/views/mainAMP.scala.html
```diff
@@ -40,6 +42,22 @@
...
+            <div class="fc-container__inner">
+                <div class="fc-container__header__title">
+                    <span>popular</span>
+                </div>
+                <amp-list width="600" height="1000" layout="responsive" src="@if(Play.isDev) {https://www.thegulocal.com} else {@Configuration.ajax.url}/most-read-amp.json">
+                    <template type="amp-mustache">
+                        <div class="headline-list__item">
+                            <p class="headline-list__count">{{index}}</p>
+                            <h2 class="fc-item__title">
+                                <a href="{{url}}" class="fc-item__link">{{headline}}</a>
```
johnduffell: haha, now I suddenly realised why it's called mustache :stuck_out_tongue_closed_eyes:

stephanfowler:
:+1: nice

common/app/views/mainAMP.scala.html
```diff
@@ -40,6 +42,22 @@
 
             @body
 
+            <div class="fc-container__inner">
+                <div class="fc-container__header__title">
+                    <span>popular</span>
+                </div>
+                <amp-list width="600" height="1000" layout="responsive" src="@if(Play.isDev) {https://www.thegulocal.com} else {@Configuration.ajax.url}/most-read-amp.json">
```
johnduffell: does it not work if you just use configuration.ajax.url every time?  I have mine explicitly configured, but I guess not everyone does

NataliaLKB: It would work normally. In dev mode it just gives you a relative url. That wont work for amp-list though because it requires having a url that starts with `https`. Can you think of a better way of doing this for local development?

common/app/views/mainAMP.scala.html
NataliaLKB: :man:

dev-build/conf/routes
NataliaLKB: Yes that makes sense. I will rename. in this case though it would be `/most-read-mf2.json`. Doesn't exactly roll off the tongue :wink:

common/app/views/mainAMP.scala.html
```diff
@@ -40,6 +42,22 @@
 
             @body
 
+            <div class="fc-container__inner">
+                <div class="fc-container__header__title">
+                    <span>popular</span>
+                </div>
+                <amp-list width="600" height="1000" layout="responsive" src="@if(Play.isDev) {https://www.thegulocal.com} else {@Configuration.ajax.url}/most-read-mf2.json">
```
johnduffell: re the if statement, I don't know the if statement is adding anything.
It would need https enabled in nginx to work, and if someone has set that up, they might as well access the site using the https url anyway.  It would be more amp compliant to do that anyway.  Or is it better to use http where possible in dev mode?

johnduffell: I'm basically just a bit nervous about hard coding the url.

NataliaLKB: Yes I don't like it either, but the problem is that the url needs to start with `https`. Without it starting with that amp-list won't work in development.

johnduffell: ah now I realise - it actually checks the first 6 characters are https: - I raised a similar issue against amp a while ago and they fixed it (for data: urls in iframes) so they might fix it here too

NataliaLKB: That makes sense! As you did previously, I will raise the issue with AMP to allow a relative path here, as long as the currently path is https.

PR #11224, opened 2015-11-30
Prevent multiple submission if user clicks submit multiple times

gtrufitt:
Adds a state flag to prevent multiple submissions if click the submit button repeatedly.


desbo:
:+1:

PR #11230, opened 2015-11-30
Amp-Font

NataliaLKB:
This PR:
- Trials `amp-font`
- Moves `amp-list` to load on only one article

The reason to only load these components on a test article is because we are experimenting with them to provide feedback to the AMP team. Once they are stable, we can add them to all the articles.

`amp-list`/`amp-mustache` is an experiment for amp. To see it, add `#development=1` onto the end of the url and also in your devtools console type in:

```
AMP.toggleExperiment('mustache')
```

cc @stephanfowler


stephanfowler:
:+1: yay

PR #11231, opened 2015-11-30
Restore Spacefinder Fix

jbreckmckye:
@uplne 

A previous fix for a [Spacefinder race condition](https://github.com/guardian/frontend/pull/11159) had to be pulled because a new file, 'space-filler', was being called from both the `commercial.js` and `enhanced.js` bundles.

I've resolved the issue by adding space-filler to `core.js`. See the final commit, b85aefa.


uplne:
:+1:

PR #11232, opened 2015-11-30
Membership adblock banner print fix

Calanthe:
It shouldn't be visible on print pages.

### DIFF
--- a/static/src/stylesheets/print.scss
+++ b/static/src/stylesheets/print.scss
@@ -7,6 +7,7 @@
 .block-share__link,
 .l-header,
 .top-banner-ad-container,
+.adblock-sticky__message,
 .content__secondary-column,
 .content__labels,
 .content-footer,


uplne:
:+1:

static/src/stylesheets/print.scss
```diff
@@ -7,6 +7,7 @@
 .block-share__link,
 .l-header,
 .top-banner-ad-container,
+.adblock-sticky__message,
```
jbreckmckye: Is this the best way to hide elements in print? I think it buries the behaviour somewhat. Could a print-hidden mixin allow us to group together the print behaviour with the rest of the component's behaviour?

I'd be thinking of a mixin like the following:

    @mixin print-hidden {
      @media print {
        & {
          display: none !important;
        }
      }
    }
    
    .hide-me-in-print {
      border: 1px solid blue;
      @include print-hidden
    }

uplne: That is a solid idea

PR #11234, opened 2015-12-01
Hiding the Most Popular tabs for video front

NataliaLKB:
On the video front we only show across the guardian most popular, and geo most popular. However, the tab says "in video". As you can see from the screenshot below, none of these are videos.

image

As a quick fix, I propose we hide the tabs like we do with the network front so at least it wont say "in video".

cc @OliverJAsh @Jholder112233

### DIFF
--- a/static/src/javascripts/projects/facia/modules/onwards/geo-most-popular-front.js
+++ b/static/src/javascripts/projects/facia/modules/onwards/geo-most-popular-front.js
@@ -26,6 +26,7 @@ define([
 
     GeoMostPopularFront.prototype.endpoint = '/most-read-geo.json';
     GeoMostPopularFront.prototype.isNetworkFront = config.page.contentType === 'Network Front';
+    GeoMostPopularFront.prototype.isVideoFront = config.page.pageId === 'video';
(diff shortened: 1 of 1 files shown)


static/src/javascripts/projects/facia/modules/onwards/geo-most-popular-front.js
```diff
@@ -43,7 +44,7 @@ define([
         this.parent = qwery('.js-popular-trails')[0];
 
         if (this.parent) {
-            if (this.isInternational && this.isNetworkFront) {
+            if (this.isInternational && this.isNetworkFront && this.isVideoFront) {
```
Jholder112233: isNetworkFront AND isVideoFront?

NataliaLKB: Thanks for catching @Jholder112233. Will change now!

johnduffell: consider performing or writing tests! ;)

Jholder112233:
:+1: let's see what happens in Code

PR #11235, opened 2015-12-01
adding `lazy` and changing contains to equals

NataliaLKB:
Changes based on @johnduffell's comments in https://github.com/guardian/frontend/pull/11208

John, take a look and let me know if this is better :smile:

### DIFF
--- a/onward/app/controllers/MostPopularController.scala
+++ b/onward/app/controllers/MostPopularController.scala
@@ -31,10 +31,9 @@ object MostPopularController extends Controller with Logging with ExecutionConte
     val sectionPopular: Future[List[MostPopular]] = if (path.nonEmpty) lookup(edition, path).map(_.toList) else Future(Nil)
 
     sectionPopular.map { sectionPopular =>
-      val sectionFirst = sectionPopular ++ globalPopular
-      val globalFirst = globalPopular.toList ++ sectionPopular
-
-      val mostPopular = if ( path.contains("global-development") ) sectionFirst else globalFirst
+      lazy val sectionFirst = sectionPopular ++ globalPopular
+      lazy val globalFirst = globalPopular.toList ++ sectionPopular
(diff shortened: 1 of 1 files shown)


onward/app/controllers/MostPopularController.scala
```diff
@@ -31,10 +31,9 @@ object MostPopularController extends Controller with Logging with ExecutionConte
     val sectionPopular: Future[List[MostPopular]] = if (path.nonEmpty) lookup(edition, path).map(_.toList) else Future(Nil)
 
     sectionPopular.map { sectionPopular =>
-      val sectionFirst = sectionPopular ++ globalPopular
-      val globalFirst = globalPopular.toList ++ sectionPopular
-
-      val mostPopular = if ( path.contains("global-development") ) sectionFirst else globalFirst
+      lazy val sectionFirst = sectionPopular ++ globalPopular
+      lazy val globalFirst = globalPopular.toList ++ sectionPopular
+      lazy val mostPopular = if ( path.equals("global-development") ) sectionFirst else globalFirst
```
johnduffell: does equals really work or does the path include the rest of the article id? if so then you can use == and it actually calls .equals in scala

NataliaLKB: Path is just `global-development` or `commentisfree`.

Is it better to use `==` over equals?

johnduffell: I think == is more scala and .equals is more java.
Google probably has many opinions, but here's one http://stackoverflow.com/questions/7681161/whats-the-difference-between-and-equals-in-scala

NataliaLKB: So many opinions :stuck_out_tongue_winking_eye:! Changed now :smile:

johnduffell:
sorry and looks great! :+1: :speedboat:

PR #11236, opened 2015-12-01
Use the new MMA JSON endpoint with the old frontend

tomverran:
The one bit of behaviour this changes is that the "Annual payment" field will only show up if your subscription is annual as the behaviour of that field in the new API is slightly different


rich-nguyen:
:+1:

tomverran:
@rich-nguyen sorry, made an extra small change - would you mind taking a quick extra look?

markjamesbutler:
LGTM :+1:

PR #11238, opened 2015-12-01
Remove extra http recordings

rich-nguyen:
Every time a dev runs the scala tests, new database files get checked in. This happens because of the crossword agent, which makes a timestamped CAPI query. I've removed the agent in Test mode.


JustinPinner:
It's difficult to see past the hundreds of files that have changed, but the removal of the story package test on galleries and not starting the agent for todays crossword grid both get a :+1: :shipit:

johnduffell:
good spot there rich! :+1:

PR #11240, opened 2015-12-01
Prevent repeat calls to require() when loading omniture in email.js

desbo:
At the moment we make two calls to `require` when loading omniture in the `email` module. This ensures only one call happens.

@gtrufitt


gtrufitt:
:+1:

PR #11243, opened 2015-12-01
Don't attempt to register SW on profile.theguardian.com

OliverJAsh:
Currently this is `404`ing at `https://profile.theguardian.com/service-worker.js`.


rich-nguyen:
:+1:

PR #11245, opened 2015-12-01
The travel feed has been switched off

kelvin-chappell:
So the blended component needs a different feed.

/cc @Calanthe


Calanthe:
:+1:

PR #11246, opened 2015-12-01
Remove crosswords banner

desbo:
Now that we're dropping R2 crosswords we can get rid of the banner about there being a new version.


JustinPinner:
Looks ok to me :+1:

PR #11252, opened 2015-12-02
Fix single content-item merchandising component

kelvin-chappell:
There were 2 problems:
- It was looking up bad short URLs
- It was looking up keyword by context rather than from parameter

/cc @Calanthe


Calanthe:
It looks ok to me but maybe ask also someone who knows scala.

PR #10825, opened 2015-10-09
Sustain: Make old article date clearer

JustinPinner:
re: https://trello.com/c/JK4Y1YkE/7-clearer-date-on-old-articles

Sometimes old articles gain new interest (e.g. Tony Hart) and so to prevent them from gathering too much new popularity we need to indicate that they are older pieces.

In the popular container;
screen shot 2015-10-09 at 09 34 50

In an article;
screen shot 2015-10-08 at 12 32 58

In a gallery;
screen shot 2015-10-08 at 11 16 07

PR #11253, opened 2015-12-02
Sustain: show more prominent content age message on old articles

JustinPinner:
We've changed the rules from the original PR (https://github.com/guardian/frontend/pull/10825).

We now only show the old content message on anything 30 days old or more and that is tagged with `tone/news`. And only if the feature is switched on.

The screenshots below show what it looks like (the crystal maze travel story wouldn't usually show up as it isn't tagged appropriately but the appearance will be the same for correctly tagged content - there just wasn't any matching old stuff in the most popular list at the time) 

screen shot 2015-12-02 at 10 55 39
screen shot 2015-12-02 at 10 57 04
screen shot 2015-12-02 at 12 08 05


paperboyo:
> We now only* show the old content message on anything 30 days old or more and that is tagged with tone/news. And only if the feature is switched on.

* God bless you! Phew!

That said - wasn't there a funny occurrence with an obituary? [I haven't said that. I'm actually against shielding readers from themselves...]

common/app/conf/switches/FeatureSwitches.scala
```diff
@@ -5,6 +5,15 @@ import org.joda.time.LocalDate
 
 trait FeatureSwitches {
 
+  val contentAgeMessageSwitch = Switch(
+    "Feature",
+    "content-age-message",
+    "Show old content message on... old content",
```
jennysivapalan: Should we add here that it's just on tone/news?

common/app/views/support/ContentOldAgeDescriber.scala
```diff
@@ -0,0 +1,39 @@
...
+  def apply(content: model.Content): String = {
+    message(content.trail.webPublicationDate)
+  }
+
+  def apply(apiContent: com.gu.contentapi.client.model.Content): String = {
+    message(apiContent.webPublicationDate)
+  }
+
+  private def message(pubDate: DateTime) = {
+    val warnLimitDays = 30   // config this!
```
jennysivapalan: Not sure it's worth a config but worth removing the comment?

jennysivapalan:
:+1: :older_man: :older_woman:

PR #11255, opened 2015-12-02
Video prerolls skip button tracking fixed in omniture

Calanthe:
The event number `event99` was set some time ago but we didn't fire that event at all. 

After:
screen shot 2015-12-02 at 13 40 18


uplne:
:+1:

PR #11109, opened 2015-11-12
Fix admin pages and offline page in dev environment

OliverJAsh:
This was broken in #11109 

Fixes #11100

PR #11256, opened 2015-12-02
Make sure assets fallback to source file in dev, if missing from the hash list

sndrs:
#11109 broke the model whereby any files that are not compiled in dev (most JS) but which _are_ required by `@Static` in scala (e.g. in [analytics.scala.html](https://github.com/guardian/frontend/blob/master/common/app/views/fragments/analytics.scala.html#L33)) will error because they were missing from the asset hash json.

This PR:
- removes `static/target`, `static/hash` and `static/requirejs` directories before compiling, to ensure they're completely clean
- ensures the path lookup fallsback to `src` dir if it's missing from the asset map in dev, and errors if that `src` file doesn't exist.

@rich-nguyen @jennysivapalan you might want to check this :wink: – that file lookup feels nasty but actual io happens in different places for different ways in (http request, `@Static` etc)

### DIFF
--- a/common/app/assets/assets.scala
+++ b/common/app/assets/assets.scala
@@ -165,5 +165,5 @@ object AssetFinder {
 
 case class AssetNotFoundException(assetPath: String) extends Exception {
   override val getMessage: String =
-    s"Cannot find asset $assetPath. You probably need to run 'make compile'."
+    s"Cannot find asset $assetPath. Have you got the right path? Or do need to run 'make compile', or 'make compile-dev'?."
 }
--- a/common/app/assets/assets.scala
+++ b/common/app/assets/assets.scala
@@ -165,5 +165,5 @@ object AssetFinder {
 
 case class AssetNotFoundException(assetPath: String) extends Exception {
   override val getMessage: String =
-    s"Cannot find asset $assetPath. Have you got the right path? Or do need to run 'make compile', or 'make compile-dev'?."
+    s"Cannot find asset $assetPath. Have you got the right path? Or do you need to run 'make compile', or 'make compile-dev'?."
 }
--- a/common/app/assets/assets.scala
+++ b/common/app/assets/assets.scala
@@ -22,7 +22,7 @@ class AssetMap(base: String, assetMap: String) {
 
     // Avoid memoizing the asset map in Dev.
     if (Play.current.mode == Mode.Dev) {
-      assets().getOrElse(path, throw AssetNotFoundException(path))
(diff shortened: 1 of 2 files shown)


gtrufitt:
:+1: Yay, productivity :chart_with_upwards_trend:

OliverJAsh:
This means that [this line](https://github.com/guardian/frontend/blob/c425ab1829f8c045088be9ca60878221f9d53825/common/app/views/fragments/analytics.scala.html#L33) will now fetch `bootstraps/ophan` from `static/src` (because it's not compiled). Just to check, does Ophan still seem to be working OK in dev?

This should mean that admin pages will work in dev as long as you run `make compile` (as before). Can you confirm that is still the behaviour?

common/app/assets/assets.scala
```diff
@@ -160,5 +165,5 @@ object AssetFinder {
 
 case class AssetNotFoundException(assetPath: String) extends Exception {
   override val getMessage: String =
-    s"Cannot find asset $assetPath. You probably need to run 'make compile'."
+    s"Cannot find asset $assetPath. Have you got the right path? Or do need to run 'make compile', or 'make compile-dev'?."
```
OliverJAsh: Slight typo here, `or do need` => `or do you need`?

makefile
```diff
@@ -43,6 +43,9 @@ shrinkwrap:
 	@npm shrinkwrap --dev && node dev/clean-shrinkwrap.js
 	@node tools/messages.js did-shrinkwrap
 
+clean-assets:
+	@rm -rf static/target static/hash static/requirejs
```
OliverJAsh: And `common/conf/assets`?

sndrs:
@OliverJAsh I can't see ophan loading, but i don't know if it should. should it? if the path to `@Static` doesn't exist, it definitely errors, and doesn't if it does, so that seems right. should ophan load in dev?

sndrs:
>This should mean that admin pages will work in dev as long as you run make compile (as before). Can you confirm that is still the behaviour?

`make watch` runs `make compile-dev` first, so if you're running that you should be fine. as far as i can tell admin runs fine, but i don't know what the symptoms of what broke in #11109 were...

sndrs:
@OliverJAsh do you want to check this out and confirm your fixes from #11109 are still in place? i think they are, but you'd know better...

OliverJAsh:
@sndrs Sure, I'll do that now.

OliverJAsh:
Yup this all works. :+1:

PR #11258, opened 2015-12-02
New JS load infrastructure

OliverJAsh:
This new load infrastructure attempts to strike a happy medium between optimal caching (infrequent cache busting) and optimal load times for first time visits (low number of requests). @sndrs and I have thought through the different options and decided this is the best for us.

Below shows is the execution order of our JS before and after.
# Before

```
└── curl (inline)
    └── boot (inline)
        └── core (manually curated vendor and common modules)
            ├── bootstraps/commercial
            └── bootstraps/enhanced (previously bootstraps/app)
```

`curl` and `boot` (i.e. `require('core')`) were inline on the page
# After

```
└── curl + bootstraps/standard + boot (app.js)
    └── enhanced-vendor
        └── bootstraps/enhanced
```

`curl + bootstraps/standard + boot` is loaded using an injected script tag at the top of the page.

`bootstraps/standard` is everything that runs outside of `bootstraps/enhanced` (https://github.com/guardian/frontend/commit/132d5b74f20694e0edaa675de358c96343749518#diff-27b345caf3191fb7f89f276adf891433R40):
- Raven setup
- A/B tests
- Image upgrades
- Commercial
- Membership access check

If and only if we detect that we should run enhance, `enhanced-vendor` and `bootstraps/enhanced` will be downloaded in parallel and then executed sequentially.
## Notes
- The separate `enhanced-vendor` bundle means we'll have better caching on vendor
  dependencies used in enhanced. Previously, `core` contained vendor and our own
  common modules (many of which had transitive vendor dependencies), and this
  meant `core.js` was being frequently cache busted.
- In the future we might want to refactor what is in standard to avoid large
  dependencies.
- Bonus: previously React appeared duplicated in four different bundles:
  `bootstraps/acessibility`, `bootstraps/crosswords`, `bootstraps/preferences`, and `bootstraps/sudoku`.
  I have removed this duplication and improved caching by fetching React from a
  CDN instead. (Alternatively we could serve React standalone ourselves, if
  people are worried about the SPOF.)
- We might want to rename files to make the load order clearer (i.e. `bootstraps/enhanced/crosswords`)
- We have to manually find the common modules of all `enhanced` bootstraps, which is a bit of a PITA. Webpack and jspm can do this automatically for you.

This spreadsheet shows the file size breakdown for bundles before and after: https://docs.google.com/spreadsheets/d/1M4yqKbKBcSny273MC1DTUFtII26bTY43KAYPhJ8eSNo/edit?usp=sharing
## To do
- [x] Fix `allcreatives.scala.html`
- [x] Test moving commercial out of standard (load enhanced after commercial) to reduce the file size of standard
- [x] Communicate and justify change to team
- [ ] Use our own React
- [x] Make it work in dev
- [x] Browser testing https://docs.google.com/a/guardian.co.uk/spreadsheets/d/1riVzI4pwvG0TlgbTgYKW2_F7dKZXjQHnZ5p33isgrF0/edit?usp=sharing
## Test urls
- http://www.theguardian.com/membership/2015/apr/27/the-spin-london-cycle-festival-save-mop
- http://m.code.dev-theguardian.com/commercial/test-page


applications/app/controllers/WebAppController.scala
```diff
@@ -50,10 +50,11 @@ object WebAppController extends Controller with ExecutionContexts with Logging {
...
           Static("stylesheets/content.css"),
           Static("stylesheets/print.css"),
-          Static("javascripts/core.js"),
-          Static("javascripts/bootstraps/standard.js"),
+          Static("javascripts/curl+boostraps/standard+boot.js"),
+          Static("javascripts/enhanced-vendor.js"),
           Static("javascripts/bootstraps/enhanced.js"),
-          Static("javascripts/bootstraps/crosswords.js")
+          Static("javascripts/bootstraps/crosswords.js"),
+          "https://cdnjs.cloudflare.com/ajax/libs/react/0.13.2/react.min.js"
```
OliverJAsh: I might want to create a variable for this URL, to avoid duplication.

grunt-configs/concat.js
```diff
@@ -0,0 +1,14 @@
+module.exports = function (grunt, options) {
+    return {
+        curlPlusCorePlusBoot: {
+            // https://github.com/gruntjs/grunt-contrib-concat/issues/131
+            // options: { sourceMap: true },
```
OliverJAsh: Will remove this comment but I tried to get grunt concat to output source maps rebased on the input source maps, but it  won't work due to a bug :cry:. Thus, we can't have a source map for this file.

Gruntfile.js
```diff
@@ -123,7 +123,7 @@ module.exports = function (grunt) {
     grunt.registerTask('compile:js', function (fullCompile) {
         grunt.task.run(['clean:js', 'compile:inlineSvgs']);
 
-        grunt.task.run(['concurrent:requireJS', 'copy:javascript', 'uglify:javascript']);
+        grunt.task.run(['concurrent:requireJS', 'copy:javascript', 'concat:curlPlusCorePlusBoot', 'uglify:javascript']);
```
sndrs: `curlPlusStandardPlusBoot`?

sndrs: maybe even `curl+standard+boot`

applications/app/controllers/WebAppController.scala
```diff
@@ -50,10 +50,11 @@ object WebAppController extends Controller with ExecutionContexts with Logging {
           Static("stylesheets/head.content.css"),
           Static("stylesheets/content.css"),
           Static("stylesheets/print.css"),
-          Static("javascripts/core.js"),
-          Static("javascripts/bootstraps/standard.js"),
+          Static("javascripts/curl+boostraps/standard+boot.js"),
```
sndrs: shall we not just call this `bootstraps/standard.js`?

OliverJAsh: I think we could do a better job with filenames to avoid ambiguity. `bootstraps/standard` is the name of the AMD module, maybe we could just call this `standard`?

Gruntfile.js
OliverJAsh: Renamed to `curl+bootstraps-standard+boot`

sndrs: or just `concat:standard`? what we're doing here, at this level, is creating the standard js. that it contains curl and boots enhanced etc is contingent, but that there is a `standard.js` isn't

static/src/javascripts/enhanced-vendor.js
```diff
@@ -0,0 +1,26 @@
...
+    'qwery',
+
+    'reqwest',
+    'videojs',
+    'videojsads',
+    'videojsembed',
+    'videojsima',
+    'videojspersistvolume',
+    'videojsplaylist'
+], function () {});
```
OliverJAsh: Many of these actually end up in standard because they are transitive dependencies. This means they won't appear in enhanced at all (because we subtract standard from enhanced).

That said, I think it's necessary to keep these dependencies listed because, if we removed them from the dependency graph of standard, they would be bundled in each bootstrap where they are used. Keeping them listed here means, if they are removed from standard, we can be sure they will be instead bundled in `enhanced-vendor` instead of with each bootstrap.

Gruntfile.js
OliverJAsh: Yup, makes sense. I'll rename the file to standard.js as well.

applications/app/controllers/WebAppController.scala
sndrs: ah sorry, yeah that's better

static/src/javascripts/enhanced-vendor.js
sndrs: i reckon a comment along these lines at the top of this file would be useful

static/src/javascripts/enhanced-core.js
```diff
@@ -1,7 +1,5 @@
 define([
-    // 3rd party libs
-    'raven',
-    'picturefill',
+    // The following modules are here for bundling purposes only
```
sndrs: maybe expand this comment to mention that it only includes internal modules? just a way of preventing individual bootstraps duplicating modules (is that correct?)

static/src/javascripts/bootstraps/enhanced.js
```diff
@@ -9,7 +9,9 @@ define([
     'common/utils/mediator',
     'common/utils/user-timing',
     'bootstraps/common',
-    'bootstraps/sport'
+    'bootstraps/sport',
+
+    'enhanced-core'
```
sndrs: i wonder if, with all the name changes, `enhanced-core` is confusing? like a contradiction in terms if you've not been following along closely enough? maybe `enhanced-common`? not a big deal though.

OliverJAsh: Good point. Will fix!

static/src/javascripts/boot.js
```diff
@@ -0,0 +1,38 @@
...
+            }
+        };
+    };
+
+    var afterPreFetch = preFetchEnhancedBundles();
+
+    require(['bootstraps/standard', 'domReady!'], function () {
+        if (guardian.isModernBrowser) {
+            afterPreFetch(function () {
+                require(['bootstraps/enhanced']);
```
sndrs: does this definitely not trigger a 2nd download if the `js!` one hasn't completed?

sndrs:
i wonder if a super-lightweight loading script in the head could kick off the download of standard sooner?

sndrs:
we need to be double sure the modules we haven in standard all fail gracefully. reckon it's a def flag for testing job too.

sndrs:
lgtm though, apart from comments. nice one :+1:

static/src/javascripts/boot.js
OliverJAsh: This callback will only be called when `require` has finished downloading both `enhanced-vendor` and `bootstraps/enhanced`, as per `preFetchEnhancedBundles`.

OliverJAsh: The idea is to kick off the download for enhanced ASAP, instead of waiting for standard to execute.

sndrs: >This callback will only be called when require has finished downloading both enhanced-vendor and bootstraps/enhanced

sorry yes of course, it's inside `afterPreFetch`

OliverJAsh:
> i wonder if a super-lightweight loading script in the head could kick off the download of standard sooner?

I'm going to consult Jake’s article for that one: http://www.html5rocks.com/en/tutorials/speed/script-loading/.

`async` would be ideal but browser support :unamused: You can do it with JS but we'll have to add a `link rel="prefetch"` or something I presume?

OliverJAsh:
> we need to be double sure the modules we haven in standard all fail gracefully

How do you mean?

> reckon it's a def flag for testing job too.

Yeah, will need to test just about everything on the site…

sndrs:
> You can do it with JS but we'll have to add a link rel="prefetch" or something I presume?

yeah we could just inject the script tag in the head, get it downloading immediately. will be non-blocking (except the injecting script) and will trigger the download from the head.

sndrs:
>> we need to be double sure the modules we haven in standard all fail gracefully

>How do you mean?

currently they've been run behind `isModern` haven't they?

sndrs:
>currently they've been run behind `isModern` haven't they?  

no i've misremembered that, ignore

OliverJAsh:
@sndrs See 350621f436f6492b5aa234593f58cf25aa5fd21f

OliverJAsh:
Horrible workaround for the `allcreatives` page, but it was better than the other solutions I could think of: 2ef83816ef87cf31afc5e93115fd36722f0d2f90

OliverJAsh:
All comments seen to. Moving on to testing!

desbo:
looks totally sensible to me. just so i understand the react thing: we've defined it as an external resource so whenever something requires `react`, curl will try to load it from the cdn (or cache)?

piuccio:
great idea.
what metrics do you plan to look at to define success?
the spreadsheet of bundle size is not indicative of an improvement for real users with different caching strategies and visiting profiles.

sndrs:
@piuccio we need to do some comms around this when it's ready, but the main driver behind this is to create a ES3-survivable `standard.js` (individual modules may opt themselves out) that contains all 'required' functionality (what counts as required can be controversial :wink:). After that, the modern, enhanced app may or may not run. We should get some perf gains thanks to not inlining a load of stuff, but this is more to be able to pull some important behaviours out of the main app (e.g. for iPads, older UAs or low-powered devices) without adding page weight by inlining.

jbreckmckye:
This is all interesting work, and vital for everyone to know as they start work on the Frontend project. Is there a good place we can document how the scripts etc. get bootstrapped? Partly so that someone checking this project out for the first time can reason how things work, and partly to document to the world the Guardian's approach. I'm not sure what format could work well but it might seem quite helpful.

OliverJAsh:
@desbo That's right. We're using a CDN to maximise the chance that the user will have it cached, from another site that would have also requested the same script.

@jbreckmckye I think that's a good idea. :smile:

sndrs:
@jbreckmckye completely agree, both for new starters and current devs, this needs a bit of comms before merging

OliverJAsh:
Commercial represents 16 kB of the current standard bundle, so we think it would be wise to bundle commercial separately. All modules in standard should aim to be as small as possible. In the long term, we hope commercial will remove large dependencies, and then we can reconsider bundling it with standard again.

By removing commercial from the standard bundle we can improve the load performance. The new load order will look like `standard -> commercial`. *This is no different to the load order we have for commercial today: `core -> commercial`.*

The only downside is enhanced must load after commercial (because the commercial bundle contains common dependencies for commercial and enhanced). This is so we don't have to duplicate bundling of dependencies.

jbreckmckye:
@OliverJAsh - that's reasonable, though I think we should probably do some analysis to see why commercial is quite as large as it is.

OliverJAsh:
@jbreckmckye As a start, here is the dependency graph for commercial: https://gist.github.com/93882f7b76378c6e1008. Maybe we can order those dependencies by file size.

jbreckmckye:
@OliverJAsh Smells like lodash, but would be good to get some facts.

OliverJAsh:
I used [`source-map-explorer`](https://github.com/danvk/source-map-explorer) to answer your question, @jbreckmckye. Here's a breakdown of file contents of `bootstraps/commercial`: http://ill-ring.surge.sh/commercial.html

The command I used:

``` js
source-map-explorer ~/Development/frontend/static/requirejs/bootstraps/commercial.js
```

Note that commercial bundle has subtracted `bootstraps/standard` (as per this pull request), so many dependencies come for free are will not be shown in the tree. The tree shows only those that add to the commercial bundle itself.

OliverJAsh:
Browser testing done. I'm going to push this out tomorrow morning.

common/app/assets/assets.scala
```diff
@@ -152,6 +152,7 @@ class Assets(base: String, assetMapPath: String = "assets/assets.map") extends L
 
      val curl: String = RelativePathEscaper.escapeLeadingDotPaths(inlineJs("assets/curl-domReady.js"))
      val omnitureJs: String = inlineJs("assets/vendor/omniture.js")
+     val reactPath: String = "https://cdnjs.cloudflare.com/ajax/libs/react/0.13.2/react.min.js"
```
rich-nguyen: we should host this content ourselves. Don't want to rely on this CDN for delivering our users content. Another unwritten clientside line in the sand.

sndrs: using a CDN for such a popular lib massively increases the chances we'd hit the http cache. agree we should be able to fall back to our own if it fails, but this seems a particularly strong case for a CDN to me (we don't want to bundle it, it's very popular)?

rich-nguyen:
super stuff. i found a little confusing because of file naming, may i suggest that the output of concatenating `curl.js` + `standard.js` + `boot.js` be called something like `app.js`, rather than `standard.js`?

common/app/assets/assets.scala
OliverJAsh: Agreed. I looked into having a fallback so we could have the best of both worlds, but that's not possible with curl’s API.

I'll raise a subsequent PR for this. Added to the to dos.

rich-nguyen: I don't think we should bundle it, but we should host it ourselves, for when that CDN goes down. This isn't a hypothetical suggestion, it's saving us from fixing an issue that has happened in the past. I don't think, without real numbers, we can claim to be hitting the http cache well using a public CDN. There are n versions of that library, there are x CDNs, there is finite cache space, there is y percentage of other sites using react.

PR #11260, opened 2015-12-02
Change sensitive content policy

jbreckmckye:
Pages marked as 'sensitive' should not exclude badges or third party tags

@kenlim


kenlim:
:+1: once it passes Teamcity

PR #11262, opened 2015-12-03
Load 3rd party code without waiting for DFP

kenlim:
We used to load our 3rd party scripts after the DFP loaded event is fired. This is not guaranteed to happen thanks to Ad blocking, and suppressed ads on sensitive content. 

So I'm removing it.


jbreckmckye:
Once it passes TC let's ship it :+1:

PR #11264, opened 2015-12-03
Replace DFP third-party creatives with template creatives

kelvin-chappell:
This is a new package specifically for replacing creatives in DFP.  It makes use of existing DFP API client code in the admin module.

It enables generating a creative template and replacing existing creatives with new creatives based on the template.


kenlim:
:+1:

PR #11266, opened 2015-12-03
Footer misalignment

superfrank:
The footer is currently misaligned with the grid. The various changes align it with the fronts cards at all breakpoints and should give us a good base for any future changes to the footer.

grid_lineup


static/src/stylesheets/layout/_footer.scss
```diff
@@ -67,14 +67,19 @@ $c-primary-footer-background-side-bar: mix($c-primary-footer-background, #ffffff
...
-    padding-left: $gs-gutter * .5;
+    padding-right: $gs-gutter / 2;
+    padding-left: $gs-gutter / 2;
 
     @include mq(mobileLandscape) {
         padding-right: $gs-gutter;
         padding-left: $gs-gutter;
     }
 
+    @include mq($from: tablet, $until: desktop) {
```
desbo: this mq seems redundant since it contains the same rules as the parent selector. if `mobileLandscape` is the only place it should differ, you can rid of it and say `$from: mobileLandscape, $until: phablet` for the other one

desbo:
:+1: other than comment

gtrufitt:
Removing more code than you've added === :star2: :+1:

PR #11268, opened 2015-12-03
Automatically generate Immersive headers if needed

sammorrisdesign-zz:
At the moment to get these headers we have to inject them into the page and production people have to fiddle in javascript. This automatically generates them if someone places a `h2` under a `figure--immersive` element.

Before:
screen shot 2015-12-03 at 12 20 16

After:
screen shot 2015-12-03 at 12 20 27


Jholder112233:
:+1: but this really needs a comment with a link to this pull request explaining why it's in there. 

Ideally this would be changed in the tools to make this a 'real' feature?

sammorrisdesign:
Happy to add an inline comment if you think it's needed

rich-nguyen:
agreed, it's a bit of a magic rule to fix up the content structure.

PR #11269, opened 2015-12-03
Use members-data-api for card updates

tomverran:
This is due to some backend changes we're making
We also disable the button when you submit a payment to prevent double submits
If there's an error it gets enabled when you change the value of a field

### DIFF
--- a/static/src/javascripts/projects/membership/payment-form.js
+++ b/static/src/javascripts/projects/membership/payment-form.js
@@ -34,7 +34,7 @@ define([
             ERROR_CARD_EXPIRY: 'js-error--card-expiry',
             HIDE: 'is-hidden',
             PAYMENT_ERRORS: 'js-payment-errors',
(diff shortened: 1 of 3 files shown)


static/src/javascripts/projects/membership/payment-form.js
```diff
@@ -34,7 +34,7 @@ define([
             ERROR_CARD_EXPIRY: 'js-error--card-expiry',
             HIDE: 'is-hidden',
             PAYMENT_ERRORS: 'js-payment-errors',
-            FORM_SUBMIT: 'js-submit-input',
+            FORM_SUBMIT: 'submit-input',
```
joelochlann: I would say add the `js-submit-input` class to the HTML rather than amending the JavaScript

joelochlann:
:+1: once tests pass

joelochlann:
:+1:

PR #11271, opened 2015-12-03
Don't print the weather

desbo:
It's not like anyone prints network fronts but it seemed crazy that this is what happens if you do:

screen shot 2015-12-03 at 14 40 48


gtrufitt:
:+1: :zap:

PR #11275, opened 2015-12-04
Masterclasses nav bar

jennysivapalan:
Add guardian masterclasses links to the second nav bar, similar to careers it's not allowed in the all-sections list.
:school_satchel:


JustinPinner:
Looks :+1: to me :mortar_board:

PR #11279, opened 2015-12-04
Remove unused val which causes exception for some tags

OliverJAsh:
This fixes an issue where we're unable to access `/global-development` in preview, and the content does not update in production.

This val was throwing an error because some tags do not have a `/`, e.g. `global-development`.

I believe this issue has only just come up because the val used to be lazy (and thus never evaluated) until #11212

/cc @rich-nguyen @jennysivapalan


jennysivapalan:
Good debugging :bug: :+1:

rich-nguyen:
thanks @OliverJAsh

PR #10152, opened 2015-08-13
Remove 'moz-binding' adblock check which causes false positives for whitelisting Firefox users

jbreckmckye:
The existing code uses two tests to sniff for adblockers - that a sacrificial ad unit gets given `display:none` and that the unit is given a suspicious `moz-binding` CSS property. The latter exists purely to sniff for Firefox Adblock Plus users, but creates false positives when those users whitelist our site.

It is more important to us that we do not give 'adblocker detected' messages to these supportive users than that we "catch" their non-whitelisting peers.

As such, I have removed the `moz-binding` check and the test now relies on display properties alone.

PR #11280, opened 2015-12-04
Sticky navigation small animation fix

Calanthe:
In the situation when there is no top banner (page is sensitive or a paid member uses adblock or we have banner placed after the first container) the animation of slim nav was buggy. Some time ago I added a rule that for sensitive pages there will be no animation of slim nav at all - just a style change from `display: block` to `display: none`. In this PR I added the same functionality for the adblock users and 'the banner below container' scenario.

Before:
slim-navigation-beforr

After:
slim_animation_after

I think that this is the easiest and fastest fix. Maybe in the future there will be a need to refactor the whole animation of the slim nav in js instead of css (`StickyHeader.prototype.showNavigation()` method).

I also fixed the bug when the user wasn't in any test and `config.tests` was an empty object.


paperboyo:
Super! Sticky>smooth is always a right direction ;)

It's probably known, but just wanted to record current scrolling up<>down behaviour of latest release of Firefox with AdBlock:

firefox with adblock

Calanthe:
If I remember correctly, we don't have a proper method to detect adblocks on Firefox.

uplne:
:+1:

paperboyo:
@Calanthe Oh, that (we know, but can't be sure that we are not whitelisted?): https://github.com/guardian/frontend/pull/10152. Boo.

PR #11282, opened 2015-12-04
Change h2 style to match new immersive headers

sammorrisdesign-zz:
Just so we don't have two styles of h2 next to each other following [this PR](https://github.com/guardian/frontend/pull/11268)

Before:
screen shot 2015-12-04 at 14 56 50

After:
screen shot 2015-12-04 at 14 56 35


jbreckmckye:
:+1:

PR #11283, opened 2015-12-04
Add a specific failure for 24*7 support

mchv:
Errors has been [added in content api](https://github.com/guardian/content-api/pull/1229) and [in flexible](https://github.com/guardian/flexible-content/pull/2215)
The article has been chosen with central production.

### DIFF
--- a/common/app/views/main.scala.html
+++ b/common/app/views/main.scala.html
@@ -76,6 +76,23 @@ http://developers.theguardian.com/join-the-team.html
 
         @body
 
+        @********************** 
+         24x7 support training 
+        
+         We intentionally modified the displayed section on this specific article.
+         This is a failure that trainees have to diagnose in frontend.
+        ************************@
+        @if(request.path == "/info/2015/mar/11/-removed-article") {
+            <script type="text/javascript">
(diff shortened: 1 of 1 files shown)


mchv:
@jennysivapalan @rich-nguyen @johnduffell

johnduffell:
I think it's really a great idea to do drills.  However in this case we should think about the best way to simulate a problem on frontend.
basically if something is going wrong the person needs to narrow it down to frontend, and decide if it needs rolling back or scaling.  In this case I'd expect users to roll it back.  For frontend problems we should be focussing on either scaling issues or maybe issues affecting articles giving 5xx errors.  For the latter we should expect people to work out what PR caused the issue and rolling back.  For any long term existing problem that only just appeared, I'd expect 3rd line would need to get involved.

mchv:
@johnduffell In this example I only want people to identify for each of the 3 issues (`headline`, `standfirst` , and `section`) which system is causing it, but not to fix the problem.

If this is PR is merged, I could test this in training and see how this is useful.
So this is much more a first test of what we could do than how we should do this in the future.

johnduffell:
ok thanks for chatting @mchv and I understand now, this works well for training people to work out whether the problems are in CAPI or frontend or composer!  :+1:

Issue #9647, opened 2015-06-29
Thrasher elements overlap sticky nav


paperboyo:
Hello,

I sure hope that they can be prevented from doing that by frontend, instead of having to educate all possible thrashers/ad creators not to use ``z-index``es above certain no.

Possibly related to e.g. https://github.com/guardian/frontend/pull/9426.

Regards
Mateusz

![thrasher elements overlap sticky nav](https://cloud.githubusercontent.com/assets/6032869/8416181/da3a60a8-1e9b-11e5-9cc1-edb9f405b27a.gif)

paperboyo:
Fixed by https://github.com/guardian/frontend/pull/9656. Thanks, @Calanthe!

PR #11284, opened 2015-12-04
Fix search for ad block users

### LINKED ISSUE #10030: Sticky nav: search misbehaving
Hello,

Chrome 44, FF 39 on Win 7; AdBlock Plus default installation; IE11.

In Chrome with AdBlock active, search results are cut off after initial scrolling even one line down, like so (and there is **no way** to get them back):

search cut off chrome with adblock

With AdBlock switched off, everything works as expected.

~~In FF, with or without AdBlock Plus, one can not scroll the search results at all (e.g. to get to their pagination), instead, the content area scrolls.~~

In IE11 search bar hides under the ad after initial scroll (but otherwise everything works as expected):

search ie

Regards
Mateusz

OliverJAsh:
Fixes #10030 

A good proportion of user feedback is users who are using an ad blocker complaining that search doesn't work. It's a trivial fix.

/cc @Calanthe @uplne @paperboyo

### DIFF
--- a/static/src/javascripts/projects/common/modules/navigation/sticky.js
+++ b/static/src/javascripts/projects/common/modules/navigation/sticky.js
@@ -437,7 +437,7 @@ define([
                 // Header is not slim yet
                 this.$els.header.removeClass('l-header--is-slim');
                 this.$els.header.css({
-                    position:  'static',
+                    position:  'relative',
                     width:     '100%',
                     'margin-top': 0,
                     '-webkit-transform': 'translateY(0%)',


static/src/javascripts/projects/common/modules/navigation/sticky.js
```diff
@@ -437,7 +437,7 @@ define([
                 // Header is not slim yet
                 this.$els.header.removeClass('l-header--is-slim');
                 this.$els.header.css({
-                    position:  'static',
+                    position:  'relative',
```
OliverJAsh: The issue was that the `z-index` wasn't being applied because it is only honoured for `relative` and `absolute` positioned elements.

I noticed there were other z-indexes set by the sticky nav along with position static. I don't know what the purpose of these is, as they will not be honoured. Any ideas @Calanthe @uplne?

paperboyo:
Thanks for that, @OliverJAsh! I remember vaguely some issue around thrashers and z-indexes... Oh, yes, that (luckily, git search is working correctly ;-) ): https://github.com/guardian/frontend/issues/9647

paperboyo:
Apparently, my head doesn't, though... it was about sticky nav (which has its own share of search-related problems...) sorry...

gtrufitt:
:+1:

sndrs:
nice! :+1:

sammorrisdesign:
:+1:

static/src/javascripts/projects/common/modules/navigation/sticky.js
Calanthe: Tbh I don't remember why it is `static`. Big thanks for noticing and fixing it!

johnduffell:
ship it!

sammorrisdesign:
:+1:

sammorrisdesign:
:+1:

PR #11288, opened 2015-12-07
Temporarily remove H2 work in immersive template

sammorrisdesign-zz:
Some of this can cause pages to break and should be an option based on the chapters switch.


rich-nguyen:
:+1:

PR #11294, opened 2015-12-07
Rename membership profile CSS to manage-account

tomverran:
Do the same with the JS stripe selectors

mma was indeed too opaque for people outside of membership & subs

This is required as we're adding a tab using the same styles which is not related to membership


joelochlann:
:+1: :smile: much less cryptic, thanks

PR #11295, opened 2015-12-07
Render something when no Main Media is added to an immersive article

sammorrisdesign-zz:
Before we used to just show the loading animation forever. This renders something a little nicer and immediate.

Mobile
screen shot 2015-12-07 at 15 01 36

Desktop
screen shot 2015-12-07 at 15 01 45

### DIFF
--- a/static/src/stylesheets/module/content/_article-immersive.scss
+++ b/static/src/stylesheets/module/content/_article-immersive.scss
@@ -136,7 +136,7 @@
                 background-size: cover;
             }
     
-            .content__head--desktop {
+            &.content__head--desktop {
                 background-image: none !important;
             }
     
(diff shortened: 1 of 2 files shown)


static/src/stylesheets/module/content/_article-immersive.scss
```diff
@@ -23,10 +23,26 @@
     }
 
     .content__logo-container {
-        z-index: 2;
+        position: relative;
+        z-index: 10;
+
+        @include mq($until: desktop) {
+            background-color: rgba(0, 0, 0, .6);
```
sndrs: dose this need to be transparent? will the colour it's over change?

sammorrisdesign: @sndrs Good point, have replaced.

gtrufitt:
:+1:

PR #11296, opened 2015-12-07
Make commercial s3 root configurable

kelvin-chappell:
This makes it easier to test writing to S3 in isolation.


jbreckmckye:
:+1:

PR #11297, opened 2015-12-07
Increase the ceiling on the previous visit count sent for advert targeting

jbreckmckye:
This is a pretty simple change - rather than sending the values 1, 2, 3, 4 and '5plus', we now send the count of the user's previous visits in the format 1, 2, 3, 4, 5, '6-9', '10-15', '11-19', '20-29' and '30plus'. This will allow the Merchandising team to do some more granular ad targeting.

@Calanthe


Calanthe:
I prefer one return but it probably doesn't matter that much :+1:

PR #11298, opened 2015-12-07
Extend ad-free-web switch

jbreckmckye:
We haven't tested the ad-free experience as of yet, and it seems unlikely we'll do so before Christmas, so I've opted to extend this until the end of next January.


Calanthe:
:+1:

PR #11299, opened 2015-12-07
Add templates for use by the digital pack manage-my-account page

tomverran:
These aren't displayed yet but I am pre-merging them to cut down on the complexity of the final PR

### DIFF
--- a/identity/app/views/fragments/profile/digitalPackForm/details.scala.html
+++ b/identity/app/views/fragments/profile/digitalPackForm/details.scala.html
@@ -1,0 +1,105 @@
+<div class="is-hidden js-dig-info">
+    <ul class="details-list u-unstyled js-dig-details">
+        <li class="details-list__item">
+            <div class="fieldset__heading">
+                <h2 class="form__heading">
+                        Subscriber ID
+                </h2>
+            </div>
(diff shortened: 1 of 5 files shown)


identity/app/views/fragments/profile/digitalPackForm/details.scala.html
```diff
@@ -0,0 +1,105 @@
+<div class="is-hidden js-dig-info">
+    <ul class="details-list u-unstyled js-dig-details">
+        <li class="details-list__item">
```
joelochlann: When you have a repeated pattern like this, you can reduce the noise by putting a "micro-template" inside the same template, e.g. https://github.com/guardian/membership-frontend/blob/master/frontend/app/views/patterns/patterns.scala.html#L13-L25

joelochlann:
:+1: Looks good. Just one comment, which can be taken as a suggestion for future iterations if you want

identity/app/views/fragments/profile/digitalPackForm/details.scala.html
tomverran: ah that is nice, I will have a go at that when I next update this template

PR #11301, opened 2015-12-07
Changes to Fronts on Articles

holder-josh:
This is a near-final version which has the correct logic of which section gets which arrangement of containers. I've also added data-link-names to the containers and I've fixed some bugs.


NataliaLKB:
:+1:

PR #11304, opened 2015-12-08
Convert hard coded API URL into a parameter

tomverran:
I considered adding objects which inherited the payment-form prototype object and set the URL themselves but decided it was quite a lot of code for no real benefit


joelochlann:
:+1:

PR #11305, opened 2015-12-08
Prevent the MPU to appear on top of headlines in the popular section

regiskuckaertz:
## The problem

So, there's this small layout bug on tablets:

screen shot 2015-12-08 at 10 04 18

The MPU is absolutely positioned, thus we had a choice: either provide ample space below the headlines or put the MPU back in the multi-column layout. I say 'back' because it originally was; it was presumably taken out because the whole content of the popular section may be overwritten by a more up-to-date version fetched in JS. I say 'may' because the behaviour is different whether on the front page on on a facia page: in the former case, the content get overwritten, in the latter case, it is fed into another tab.

Under this behaviour, there was a risk of generating two impressions—because async.
## The solution

The whole process is quite convoluted and so is the solution I'm afraid. In order to restore the MPU in its initial position, a bona fide `LI` is added after the headlines in the Scala template, where ultimately the ad slot will be added—the only way to restore it in the MC algorithm. But this element does not have the usual `js-fc-slice-mpu-candidate` class; this to prevent the process in `slice-adverts` from adding a slot where it will potentially be overwritten. Instead, it has a new class, `js-fc-slice-mpu-candidate-onload`, and we're going to use it a bit later. This solves our first issue, which is running the risk of a double impression.

Now we need to wait until the most read items have been fetched and fed into the DOM. When it is done, then we can safely ad the slot to the page. One would argue this process is unnecessary on facia pages, but this section is far down the page and branching for this condition wouldn't add much to the UX. So, waiting it is, and there are two events we need to wait for: `modules:geomostpopular:ready` and `modules:onward:geo-most-popular:ready`. As soon as one of them is emitted, `onHeadlines` is called where we can safely look for our `js-fc-slice-mpu-candidate-onload` and insert the ad slot.
## The tests

Much of the tests has been updated: since `sliceAdverts` returns a `Promise`, let's use the monadic way to wait for it to finish, by way of the `then` binding function, instead of waiting for the next layout frame using `fastdom.defer`.

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/slice-adverts.js
+++ b/static/src/javascripts/projects/common/modules/commercial/slice-adverts.js
@@ -4,9 +4,11 @@ define([
     'Promise',
     'common/utils/$',
     'common/utils/config',
+    'common/utils/mediator',
     'common/utils/detect',
     'common/utils/fastdom-idle',
     'common/modules/commercial/create-ad-slot',
+    'common/modules/commercial/dfp-api',
     'common/modules/user-prefs',
     'common/modules/commercial/commercial-features',
     'lodash/objects/defaults',
@@ -19,9 +21,11 @@ define([
     Promise,
     $,
     config,
+    mediator,
     detect,
     idleFastdom,
     createAdSlot,
+    dfp,
     userPrefs,
     commercialFeatures,
     defaults,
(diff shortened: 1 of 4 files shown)


static/src/javascripts/projects/common/modules/commercial/slice-adverts.js
```diff
@@ -70,35 +81,49 @@ define([
...
+                        .insertAfter($.ancestor($adSlice[0], 'fc-container'));
+                    resolve($slot);
+                });
+            }
+        });
+    }
+
+    function onHeadlines() {
+        var $li = $('.js-popular-trails .js-fc-slice-mpu-candidate-onload').first();
+        $li.removeClass('js-fc-slice-mpu-candidate-onload')
```
sndrs: updates to the dom need to be wrapped in `idleFastdom`

regiskuckaertz: good catch, thanks!

regiskuckaertz:
@sndrs Hello, if this is now ok with you, may I proceed?

michaelwmcnamara:
Hi, I'm not getting any Mpu's for the most popular module on this branch, when I view an article page on an ipad.
Master branch seems to be pulling them down ok so I suspect there might be a problem here. Grab me if you want to have a look.

sndrs:
:+1:

PR #11309, opened 2015-12-08
Split format helper functions into their own file

tomverran:


static/src/javascripts/projects/membership/formatters.js
```diff
@@ -0,0 +1,28 @@
+define([], function () {
+    return {
+        formatAmount: function (amount) {
+            return amount ? '£' + (amount / 100).toFixed(2) : 'FREE';
+        },
+        formatDate: function (timestamp) {
+            var date = new Date(timestamp),
```
ostapneko: Could we use [toLocaleDateString](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString) instead?

chrisjowen:
:+1:

PR #11310, opened 2015-12-08
Fronts on Articles changes

holder-josh:
- The test now has two variants with different layouts. Either 2 containers before most popular and 2 after most popular, or 1 container before most popular and 3 after most popular. **Most of fronts-containers.js can be cleared out once we can move this out of the 'test' stage - it's currently a spaghetti of variations**
- Previously the endpoint requests were getting editionalised but then cached. The endpoint's cache isn't fragmented by edition so you often got the wrong edition's containers. This adds the edition to the endpoint.
- Mike (QA) recommended live blogs be excluded from this due to page weight concerns. This can be changed once live blogs are paginated.


NataliaLKB:
As discussed, this code will need a refactor at some point soon. However, recognise we need to get it out! :+1: with a refactor in another PR

Jholder112233:
I accept your :+1:'s terms and conditions.

PR #11317, opened 2015-12-08
Remove A/B Test & Allow 'editionalised' inline email sign-up

gtrufitt:
Push the inline footer email to all users (behind a switch in case it all goes terribly wrong) and allow signing up to different lists (Guardian Today UK, US and AU). This allows us to show the sign-up component in the footer for all editions, as well as use it within articles for different emails.

CC/ @janua


OliverJAsh:
:+1: nice, but you have a merge conflict!

PR #11318, opened 2015-12-08
Login hint

markjamesbutler:
Implementation of LoginHint flow.  See (https://developers.google.com/identity/protocols/OpenIDConnect?hl=en#login-hint)

This should increase conversions by pre-filling the email box and suppressing the account choose for uses that are re-authenticating.

See https://github.com/guardian/identity-federation-api/pull/168 for backend implementation.


nlindblad:
:+1:

PR #11320, opened 2015-12-09
Fix the stripe form DOM elements being saved on the prototype object

tomverran:
…and so multiple stripe forms being bound to one set of elements

This also includes an extra test for the payment form which I did a while back

### DIFF
--- a/static/src/javascripts/projects/membership/payment-form.js
+++ b/static/src/javascripts/projects/membership/payment-form.js
@@ -50,8 +50,8 @@ define([
     };
 
     /**
-     * @param {String} form CSS selector for the form to attach behaviour to
-     * @param {String} successElement CSS selector for an element to show/hide on form success
+     * @param {String|Element} form element or selector which is the form this is for
+     * @param {String|Element} successElement element or selector for the success element to show/hide
      * @param {String} url The API url to hit with a new card token
      * @returns {StripePaymentForm}
      * @constructor
--- a/static/src/javascripts/projects/membership/payment-form.js
+++ b/static/src/javascripts/projects/membership/payment-form.js
@@ -50,7 +50,6 @@ define([
             TIER_FIELD: 'js-tier-field',
             UPDATING: 'js-updating'
         },
-        DOM: {},
         data: {
             CARD_TYPE: 'data-card-type'
         }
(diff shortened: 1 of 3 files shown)


static/src/javascripts/projects/membership/payment-form.js
```diff
@@ -110,8 +109,9 @@ define([
     };
 
     StripePaymentForm.prototype.domElementSetup = function () {
+        this.DOM = {};
```
joelochlann: I'd prefer that all properties were initialised in `.init()`

joelochlann: ...or the constructor

static/src/javascripts/projects/membership/payment-form.js
```diff
@@ -61,14 +60,14 @@ define([
      */
     StripePaymentForm.prototype.updateCard = function (card) {
         var cardTypeClassName = card.type.toLowerCase().replace(' ', '-');
-        var cardTypeElem = this.config.DOM.CARD_TYPE;
-        $(this.config.DOM.CARD_LAST4).text(card.last4);
+        var cardTypeElem = this.DOM.CARD_TYPE;
```
joelochlann: I would argue that to prevent this kind of thing happening again, you should just move the `config` object off the prototype. It can hang directly off of `StripePaymentForm` (`StripePaymentForm.config`) like a "static" property. As discussed, methods on the prototype are fine, but data is generally a bad idea

static/src/javascripts/projects/membership/payment-form.js
```diff
@@ -50,44 +44,53 @@ define([
             TIER_FIELD: 'js-tier-field',
             UPDATING: 'js-updating'
         },
-        DOM: {},
         data: {
             CARD_TYPE: 'data-card-type'
         }
     };
 
+    function StripePaymentForm(form, successElement, url) {
```
joelochlann: Please document that `successElement` is expected to be a bonzo wrapped set! Or maybe better, expect it to be an actual DOM element and wrap it inside the class

tomverran: Ah yes that is rather inconsistent of me, I might change both `form` and `successElement` to be CSS selectors rather than any kind of DOM element, then the constructor can wrap them with bonzo

static/test/javascripts/spec/membership/payment-form.spec.js
```diff
@@ -61,8 +56,8 @@ define([
 
         beforeEach(function () {
             paymentFormFixtureElement = $.create(paymentFormHtml)[0];
-            paymentForm = new PaymentForm(guConfig);
-            paymentForm.init(paymentFormFixtureElement);
+            paymentForm = new PaymentForm(paymentFormFixtureElement, $(), '');
```
joelochlann: should these first two args be selector strings now?

tomverran: wow it is interesting that the karma test still passes
I guess bonzo flattens $($(...))

joelochlann: yeah it will do, jQuery does the same - but it's clearer if we pass in the right thing though!

tomverran: hmm it turns out it is useful to be able to pass in DOM elements in the test to avoid having to modify the actual page we're testing on, how would you feel about specifically supporting passing either dom elements or css selectors to the payment form constructor?

joelochlann:
:+1:

PR #11324, opened 2015-12-09
Video play button alignment issues

zeftilldeath:
Video play buttons were not center aligned on smaller container size. 

Red colour illustrated where the play icon/circle should be. 

off-center

off-center_2

Result:
screen shot 2015-12-09 at 12 18 23

### DIFF
--- a/static/src/stylesheets/module/content/_media-player.scss
+++ b/static/src/stylesheets/module/content/_media-player.scss
@@ -111,7 +111,7 @@ $ima-controls-height: 70px;
             @include mq(mobileLandscape) {
                 @include video-play-button-size($vjs-large-button-size);
             }
-            background-color: rgba(51, 51, 51, .8);
+            background-color: rgba(51, 51, 51, .6);
             .gu-media-wrapper:hover &,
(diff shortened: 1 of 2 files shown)


sndrs:
lgtm - is it ok across all breakpoints?

paperboyo:
Hush-hush. Maybe that's the time when we can drop opacity of the play button dark oval just a tad, so that the below does not look so terrible (there is one like that every day on the site. Trying to educate people to crop off-center will **never** work)?

picture 11

You won't tell anybody, I won't tell anybody, reviewers won't tell anybody, nobody will notice, everything's gonna be fine. Just fine. Only a bit better.

Regards
Mateusz

zeftilldeath:
Cheers @sndrs, yep looks good on all breakpoints. Snuck in a 20% opacity drop @paperboyo. Or did I?

paperboyo:
:gem: Don't know what you're talking about, @zeftilldeath :innocent: In case any designers ask, I always mistake opacity for transparency and vice versa.

giphy

sammorrisdesign:
:+1:

PR #11327, opened 2015-12-09
Extend a/b video preroll switch

kelvin-chappell:
Test ends at end of year.  Switch should probably be removed during the following week.


Calanthe:
:+1:

PR #11328, opened 2015-12-09
Fronts on Articles editionalisation change

holder-josh:
Now we're using the edition that is passed in from the client to determine which edition the containers should be fetched from, instead of looking at the request's edition header.


johnduffell:
:+1: :speedboat:

PR #11331, opened 2015-12-09
Refactor of DFP API client code

kelvin-chappell:
Now the dfp package has two key classes:
- [SessionWrapper](https://github.com/guardian/frontend/blob/master/admin/app/dfp/SessionWrapper.scala) wraps and consistently logs all uses of the client library
- [DfpApi](https://github.com/guardian/frontend/blob/master/admin/app/dfp/DfpApi.scala) is the only point of access from outside package for transactions using the API client


kenlim:
:+1:

PR #11333, opened 2015-12-09
Dedupe dynamic containers on non-network editorial fronts

OliverJAsh:
We don't currently dedupe dynamic containers because deduping would effect their layout when actually we want their layout to be an editorial decision. However, we get a lot of complaints about "deduping not working" as these containers increase in usage.

Katie and I have thought of a solution: dedupe dynamic containers for all editorial fronts except the network front. The network front team are happy to manually dedupe as a compromise for keeping the integrity of their layout inside dynamic containers.

We need to do this so we can deprecate the old fixed containers in favour of the new `dynamic/slow-mpu`.

This is a test so CP can do a quick sanity check before we push it live.

/cc @janua @crifmulholland

### DIFF
--- a/common/app/layout/Front.scala
+++ b/common/app/layout/Front.scala
@@ -344,9 +344,8 @@ object Front extends implicits.Collections {
     ): (Set[TrailUrl], Seq[FaciaContent], (Seq[DedupedItem], Seq[DedupedItem])) = {
     container match {
       case Dynamic(dynamicContainer) =>
-        /** Although Dynamic Containers participate in de-duplication, insofar as trails that appear in Dynamic
-          * Containers will not be duplicated further down on the page, they themselves retain all their trails, no
-          * matter what occurred further up the page.
+        /** We won't remove items from a dynamic container to avoid unexpected layout, but we will add items in the
+          * container to the list of things to remove later.
           */
         dynamicContainer.containerDefinitionFor(
           faciaContentList.map(Story.fromFaciaContent)
--- a/common/app/layout/Front.scala
+++ b/common/app/layout/Front.scala
@@ -12,7 +12,7 @@ import model.meta.{ItemList, ListItem}
 import org.joda.time.DateTime
 import play.api.libs.json.Json
 import play.api.mvc.RequestHeader
-import services.CollectionConfigWithId
+import services.{ConfigAgent, CollectionConfigWithId}
 import slices.{MostPopular, _}
 import views.support.CutOut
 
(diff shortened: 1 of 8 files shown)


common/app/layout/Front.scala
```diff
@@ -476,8 +489,11 @@ object Front extends implicits.Collections {
         (Set.empty[TrailUrl], initialContext, emptyDedupedResultWithPath)
       ) { case ((seenTrails, context, dedupedFrontResult), (pressedCollection, index)) =>
         val omitMPU = pressedPage.metadata.omitMPUsFromContainers(edition)
+        val isNetworkFront = pressedPage.metadata.isNetworkFront
         val container = Container.fromPressedCollection(pressedCollection, omitMPU)
-        val (newSeen, newItems, (usedAndDeduped, usedButNotDeduped)) = deduplicate(seenTrails, container, pressedCollection.curatedPlusBackfillDeduplicated)
+        val isEditorialFront = ConfigAgent.getPriorityForPath(pressedPage.id).getOrElse("editorial") == "editorial"
```
janua: This seems slightly odd... The default behaviour is to say yes it is `editorial`? Maybe that's right.
I would have expected to see `== Some("editorial")`, but that would mean it wouldn't default to editorial.

There would be something seriously wrong if the `ConfigAgent` didn't contain the `pressedPage.id` though.

common/app/slices/DynamicContainer.scala
```diff
@@ -11,7 +11,7 @@ private [slices] case class BigsAndStandards(
   standards: Seq[Story]
 )
 
-private [slices] trait DynamicContainer {
+trait DynamicContainer {
```
janua: You can put this back to `private [slices]`.

common/app/layout/Front.scala
```diff
@@ -369,22 +369,52 @@ object Front extends implicits.Collections {
   def deduplicate(
     seen: Set[TrailUrl],
     container: Container,
-    faciaContentList: Seq[FaciaContent]
-    ): (Set[TrailUrl], Seq[FaciaContent], (Seq[DedupedItem], Seq[DedupedItem])) = {
+    faciaContentList: Seq[FaciaContent],
+    isNetworkFront: Boolean,
+    isEditorialFront: Boolean
+    )(implicit request: RequestHeader): (Set[TrailUrl], Seq[FaciaContent], (Seq[DedupedItem], Seq[DedupedItem])) = {
```
janua: What's using this `implicit request`? Have you thought about trying to not carry it around?

OliverJAsh: `mvt.DedupeDynamicContainers.isParticipating`, below.

common/app/layout/Front.scala
OliverJAsh: Sometimes the value returned for priority is an empty string. I checked in fronts tool and editorial was selected as the priority, so I presume empty string means editorial?

janua: Ah yes, maybe there are a few now laying around that have existed in the past when we didn't have `priority`!

common/app/layout/Front.scala
janua: Could you not do this check upstream in the same way you do it with `isNetworkFront` and `isEditorialFront`?

common/app/layout/Front.scala
OliverJAsh: I will leave a comment to say why :smile:

OliverJAsh:
After more discussions we decided against this. Closing.

PR #11334, opened 2015-12-09
add new test for related content from tags

johnduffell:
So currently we use a heavy algorithm for related content.  We are considering using tags instead.

Since most of the value we add is in the curation rather than scraping, this should be at least as good as the old one, as it plays better to our strengths.

The initial implementation here grabs the first 2 articles from the top 3 tags on the story, and then adds a couple of other to make up the 8.  This can be further refined if CAPI can give the in body links and a better mix, perhaps not based on the date order.

Other improvements would be to move the tag buttons into the same area, but that would need design love.  This would mean they would seem to have a better context.

On first look the new content looks not bad:
http://www.thegulocal.com/us-news/2015/dec/09/petition-calling-for-donald-trump-to-be-banned-from-uk-signed-by-85000
only tagged with donald trump:
image

And another: http://www.thegulocal.com//world/2015/dec/07/keep-calm-carry-on-attacks-whip-up-fear
image

This one: http://www.thegulocal.com//politics/2015/dec/07/uncertainty-over-uk-future-destabilising-eu-donald-tusk
image
image

And: http://www.thegulocal.com/music/2015/aug/28/glastonbury-donates-your-wellies-to-calais-migrants
image
image

@Jholder112233 @stephanfowler @zeftilldeath take a look and feel free to push commits if you like before I get back


onward/app/controllers/RelatedController.scala
```diff
@@ -24,16 +24,29 @@ object RelatedController extends Controller with Related with Containers with Lo
...
+  /*
+  give the first 2 from each tag excluding the current article
+   */
+  def renderTags(path: String) = MemcachedAction { implicit request =>
+    val edition = Edition(request)
+    val keywordIds = request.queryString.getOrElse("keywordIds", Nil).flatMap(_.split(","))
+
+    getRelatedByTags(edition, path, keywordIds) map {
+      case related if related.items.isEmpty => JsonNotFound()
+      case trails => renderRelated(trails.items.sortBy(-_.content.trail.webPublicationDate.getMillis), "related stories")
```
stephanfowler: Contradicting my previous suggestion... I changed this to "related stories". Thinking about it again, "related" is appropriate - because surely the point of tags is to gather related things. But "content" is a bland term. And "stories" is consistent with the "more on this story" title we give to story packages. Thus, "related stories".

stephanfowler:
Nice. Those results look pretty plausible.

static/src/javascripts/projects/common/modules/experiments/tests/alternative-related.js
```diff
@@ -0,0 +1,38 @@
...
+], function (
+    config
+) {
+    return function () {
+        this.id = 'AlternativeRelated';
+        this.start = '2015-12-07';
+        this.expiry = '2016-01-07';
+        this.author = 'John Duffell';
+        this.description = 'gets related content from the top tags instead of the content';
+        this.audience = 0.0;
```
stephanfowler: Go on, give it 10%

Jholder112233:
:+1: for the test

PR #11335, opened 2015-12-10
Test interop of user-features module and membership-frontend

jbreckmckye:
The user-features module relies on a cookie value that membership-frontend can manipulate. This pull request adds a unit test to ensure that interoperation works OK, and to document that relationship between the applications.


uplne:
:+1:

PR #11338, opened 2015-12-10
Fixes pagination for section all pages.

jennysivapalan:
Pagination comes from the metadata and not the section.

cc @rich-nguyen


rich-nguyen:
:+1:

PR #11339, opened 2015-12-10
Fronts on Articles: Fixes IE9 and turns test on

holder-josh:
Fixes the reqwest errors in IE9, enables the test for IE users and changes the audience size.


jbreckmckye:
:+1:

PR #11340, opened 2015-12-10
Rename bootstrap files for clarity

OliverJAsh:
Following on from #11258, this renames the bootstrap files to increase clarity, so we know which bootstraps belong to standard or enhanced.


rich-nguyen:
:+1: not sure if `index` is a misleading module name though.

PR #11341, opened 2015-12-10
Remove BrandedComponent

regiskuckaertz:


common/app/templates/inlineJS/blocking/config.scala.js
```diff
@@ -27,13 +27,6 @@ window.guardian = {
         },
         "tests": { @JavaScript(mvt.ActiveTests.getJavascriptConfig) },
         "modules": { },
-        "images": {
-            "commercial": {
-                "brandedComponentJobs": "@Static("images/commercial/branded-component-jobs.png")",
```
Calanthe: Are you going to remove those images from the codebase as well?

regiskuckaertz: @Calanthe you're right, sorry I thought all the pictures were in src

regiskuckaertz: done

kenlim:
@regiskuckaertz has confirmed that Branded Components are an obsolete component that isn't in use anymore. 

In which case, :+1:

PR #11342, opened 2015-12-10
amp font on all the pages

NataliaLKB:
Amp font is now validating. It still doesn't show up in the cache, but we still have our default styling

cc @stephanfowler


stephanfowler:
:+1: why the hell not

NataliaLKB:
:fireworks: :sparkles:

PR #11344, opened 2015-12-10
Add data link name to discussion post button

rich-nguyen:
There is a wider issue here, we shouldn't have to trap every event. But this has happened before; events that don't get read by the clickstream are being handled by Omniture's library as a page view. Add a `data-link-name` to fix it, like @gtrufitt did in https://github.com/guardian/frontend/pull/11064


JustinPinner:
If `post comment` is a unique name (if that matters) :+1:

OliverJAsh:
:+1:

PR #11346, opened 2015-12-11
Rewrite createAdSlot for improved perfs and less dependencies

regiskuckaertz:
- removed unused dependency: detect
- removed dependency on _.template: the bonzo equivalent is much simpler and much faster

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/create-ad-slot.js
+++ b/static/src/javascripts/projects/common/modules/commercial/create-ad-slot.js
@@ -111,7 +111,7 @@ define([
             .addClass('js-ad-slot ad-slot ad-slot--dfp ' + classes.join(' '));
     }
 
-    return function (name, types, series, keywords, slotTarget) {
+    return function (name, slotTypes, series, keywords, slotTarget) {
         var slotName = slotTarget ? slotTarget : name,
             attributes = {},
             definition,
@@ -147,8 +147,8 @@ define([
             attributes.keywords = keywords;
         }
 
-        if (types) {
-            classes = map((isArray(types) ? types : [types]), function (type) {
+        if (slotTypes) {
+            classes = map((isArray(slotTypes) ? slotTypes : [slotTypes]), function (type) {
                 return 'ad-slot--' + type;
             });
         }
--- a/static/src/javascripts/projects/common/modules/commercial/create-ad-slot.js
+++ b/static/src/javascripts/projects/common/modules/commercial/create-ad-slot.js
@@ -1,5 +1,5 @@
 define([
-    'common/utils/$',
+    'bonzo',
     'common/utils/config',
     'lodash/collections/map',
     'lodash/objects/assign',
@@ -100,7 +100,7 @@ define([
     };
 
     function createAdSlotElement(name, attrs, classes) {
-        return $.create('<div></div>')
+        return $(document.createElement('div'))
             .attr({
                 'id': 'dfp-ad--' + name,
                 'data-link-name': 'ad slot ' + name,
@@ -109,6 +109,7 @@ define([
             })
             .attr(attrs)
             .addClass('js-ad-slot ad-slot ad-slot--dfp ' + classes.join(' '));
+
     }
 
     return function (name, slotTypes, series, keywords, slotTarget) {
--- a/static/src/javascripts/projects/common/modules/commercial/create-ad-slot.js
+++ b/static/src/javascripts/projects/common/modules/commercial/create-ad-slot.js
@@ -1,21 +1,18 @@
 define([
     'common/utils/$',
     'common/utils/config',
-    'common/utils/detect',
-    'common/utils/template',
-    'text!common/views/commercial/ad-slot.html',
     'lodash/collections/map',
+    'lodash/objects/assign',
     'lodash/objects/isArray',
-    'lodash/objects/pairs'
+    'lodash/objects/transform'
 ], function (
     $,
     config,
-    detect,
-    template,
-    adSlotTpl,
     map,
+    assign,
     isArray,
-    pairs) {
+    transform
+) {
     var adSlotDefinitions = {
         right: {
             sizeMappings: {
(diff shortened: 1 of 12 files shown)


static/src/javascripts/projects/common/modules/commercial/create-ad-slot.js
```diff
@@ -102,49 +99,70 @@ define([
...
-            slotName = slotTarget ? slotTarget : name,
+        var slotName = slotTarget ? slotTarget : name,
+            attributes = {},
             definition,
-            dataAttrs,
+            classes,
             $adSlot;
 
-        definition = (slotName.match(/^inline/) && slotName !== 'inline1') ? adSlotDefinitions.inline : adSlotDefinitions[slotName];
+        definition = adSlotDefinitions[slotName] || adSlotDefinitions.inline;
```
jbreckmckye: If an invalid slot name filters through, and the slot is given inline properties rather than undefined ones, will this mask a bug?

jbreckmckye:
How much faster is it to create the element with Bonzo rather than templating? What sort of improvements are we looking at?

static/src/javascripts/projects/common/modules/commercial/create-ad-slot.js
```diff
@@ -102,49 +99,70 @@ define([
...
-        );
-        for (attrName in dataAttrs) {
-            if (dataAttrs[attrName] === false) {
-                $adSlot.attr('data-' + attrName, 'false');
-            }
+
+        assign(attributes, definition.sizeMappings);
+
+        if (definition.label === false) {
+            attributes.label = 'false';
```
jbreckmckye: Maybe these should be titled `dataAttrs` to make it clear what we're manipulating here.

static/src/javascripts/projects/common/modules/commercial/create-ad-slot.js
```diff
@@ -102,49 +99,70 @@ define([
...
-            $adSlot.attr('data-series', series);
+            attributes.series = series;
         }
+
         if (keywords) {
-            $adSlot.attr('data-keywords', keywords);
+            attributes.keywords = keywords;
+        }
+
+        if (types) {
```
jbreckmckye: Maybe this parameter should be renamed something more specific. Like `slotTypes`

static/src/javascripts/projects/common/modules/commercial/create-ad-slot.js
regiskuckaertz: I admit this is a fairly tenuous line of code, but a failure at this point would happen under the confluence of many conditions, and the resulting visual rendering would be at odds with the what is expected. Whitelisting values seems over-engineered at this point.

If you prefer I can revert this line to its previous state.

static/src/javascripts/projects/common/modules/commercial/create-ad-slot.js
regiskuckaertz: @jbreckmckye well, they're all data attributes: size mappings, slot targets, etc.

static/src/javascripts/projects/common/modules/commercial/create-ad-slot.js
regiskuckaertz: @jbreckmckye deal.

static/src/javascripts/projects/common/modules/commercial/create-ad-slot.js
```diff
@@ -102,49 +99,70 @@ define([
...
+            classes = map((isArray(types) ? types : [types]), function (type) {
+                return 'ad-slot--' + type;
+            });
         }
+
+        classes.push('ad-slot--' + name.replace(/((?:ad|fo|sp)badge).*/, '$1'));
+
+        $adSlot = createAdSlotElement(
+            name,
+            transform(attributes, function (result, size, key) {
```
jbreckmckye: This seems a tad esoteric to me. Why not have your attributes named `data-x` in the first instance? It's not really much more boilerplate but it's simpler to reason about, I think. I don't think I'd ever seen _.transform before.

regiskuckaertz:
@jbreckmckye bonzo vs _.template: we're talking orders of magnitude faster here. The lodash variant is a gobbledygook of string manipulations, while bonzo uses the browser's parser via `innerHtml`. I think `template` should be used only for very intricate html fragments, and when it's used pre-compilation (rather than compilation on-the-fly) is a must, because the computational cost is amortized by repetitive calls.

jbreckmckye:
@regiskuckaertz I am sure Bonzo is quicker, but I'm curious to know how much quicker. Tens of milliseconds?

static/src/javascripts/projects/common/modules/commercial/create-ad-slot.js
jbreckmckye: @regiskuckaertz When I first read your PR, I was initially unsure whether you were manipulating data attributes or plain node attributes. It's not a big deal.

static/src/javascripts/projects/common/modules/commercial/create-ad-slot.js
regiskuckaertz: That is what I wanted to do initially, but the fact is I can't change the keys in `definition.sizeMapping` without the risk of breaking the app somewhere else.

static/src/javascripts/projects/common/modules/commercial/create-ad-slot.js
jbreckmckye: I don't really mind, it just makes spotting issues a little easier if we throw errors rather than fall through to undefined behaviour. But I guess the latter could be nicer on the user.

static/src/javascripts/projects/common/modules/commercial/create-ad-slot.js
jbreckmckye: I see. Fair enough, then.

regiskuckaertz:
@jbreckmckye oh, sorry. Yes, probably, multiplied by the number of times `template` is called. You would have the same kind of difference between a program compiled vs the same program generated on the fly and then evaluated. In essence, lodash creates the source code of a javascript function that outputs a string, then feeds it into the browser (via the Function constructor) which can finally evaluate it. This is compounded by the fact that template relies heavily on regular expressions.

Admittedly, I could have written `bonzo(document.createElement('div'))`, that's even faster. Let's do this.

jbreckmckye:
@regiskuckaertz I didn't realize the templating util actually did that (I've always just stuck to plain DOM instead). So, it's actually compiling a script anew under the covers...

regiskuckaertz:
@jbreckmckye yes :) this is convenient, but very inefficient. much more efficient: use `innerHTML` and then `querySelector` to update the nodes and attributes on the generated fragment. this is one of the things I'd like to explore re perf improvements.

jbreckmckye:
:+1:

PR #11348, opened 2015-12-11
Add data link name for lightbox buttons

rich-nguyen:
The gallery lightbox buttons need analytics data-link-names


OliverJAsh:
Makes sense :+1:

paperboyo:
Oh, that's interesting. I thought we've had them just @rich-nguyen's dashboard wasn't working...

PR #11349, opened 2015-12-11
Small tweaks to Fronts On Articles

holder-josh:
- Add tracking to 'view more comments' button
- Initialise the fronts on articles component earlier so the 'view more comments' button appears earlier
- Add a media query to the discussion container offset (for the sticky nav, which only appears >= tablet)


OliverJAsh:
:+1:

PR #11352, opened 2015-12-11
Refactor email setup to catch lazy iframe

gtrufitt:
Moving some functions around so that we can ensure we're initalising the content in the iframe if it loads after the JS (alas, page load is always faster than JS load locally, so a good to have prod only issue...)


OliverJAsh:
:+1:

PR #11356, opened 2015-12-11
Deploys radiator

OliverJAsh:
This adds a radiator to admin (at `/deploys-radiator`) that shows the status of CODE and PROD, and the Git commit difference between them.
- The radiator will group deploys by status and build number. If all the deploys have the same status and build number, we can collapse them into a single box.
- At the top we display the difference between CODE and PROD as a list of Git commits (Oldest PROD build...Latest CODE build).
- The radiator updates every 10 seconds.

It's written in TypeScript (:heart: :heart: :heart: ) and uses jspm for module loading and compilation.
# Preview

# Examples

Here are some examples of what the radiator will look like in various situations.
## Failed deploy

# To do
- [x] Add and merge config: https://github.com/guardian/platform/pull/875
- [x] Add GitHub token to avoid rate limiting


johnduffell:
fantastic improvement, however I don't know if adding typescript is likely to be problematical (build changes or just maintenance wise) so you might want to run that by other people too
other than that it's a great idea :+1:

OliverJAsh:
Thanks!

I agree that is a concern. This radiator is entirely isolated from our other JavaScript in admin so I'm not too worried. I think new projects like this provide a good opportunity to experiment with new technologies, which is what I did here. At first I was going to keep it outside of frontend, but I don't think people will use it if it doesn't live here. Also, if we can make the radiator more fancy technology-wise, people might be more inclined to show it some love.

Personally, I hope I never have to use a dynamic language again. TypeScript :heart:

common/app/common/configuration.scala
```diff
@@ -169,6 +169,10 @@ class GuardianConfiguration(val application: String, val webappConfDirectory: St
     lazy val token = configuration.getStringProperty("github.token")
   }
 
+  object teamcity {
+    lazy val host = configuration.getMandatoryStringProperty("teamcity.host")
```
OliverJAsh: Is mandatory OK? I will make sure we provision beforehand.

rich-nguyen: Fine.

rich-nguyen:
Yes, agree with some of @johnduffell 's concerns regarding the proliferation of technologies that we use to do similar things, especially client-side. Some kind of remove-something/add-something quota would help. I do think Typescript is a candidate for improving our codebase, and unless we make progress with the tough problems (how we replace fundamental parts of client-side tech pipeline), a lot of changes we make could be construed as 'contributing to tech-debt'. Something to talk about in engineering meeting (which we need to bring back).

That aside, I like the new dash, though it's a shame the `*.d.ts` files need to be checked-in, is there a way to avoid this? :+1:

PR #11358, opened 2015-12-14
Remove the essential read test

holder-josh:
This is no longer being worked on


johnduffell:
good idea :+1: go for it!

PR #11359, opened 2015-12-14
Old article message

zeftilldeath:
Couple of tweaks to the new 'Old article message'

Before:
screen shot 2015-12-14 at 11 05 23

After:
screen shot 2015-12-14 at 11 04 38


Jholder112233:
:+1: :gem:

PR #11360, opened 2015-12-14
Remove top margin in facia containers on mobile

regiskuckaertz:
The margin was there (ostensibly) to make space for the header—as the title is floating its height should collapse—but as it has `overflow: hidden` the floating title is counted in nonetheless.

Before:

screenshot_2015-12-12-09-02-08

After:

screen shot 2015-12-14 at 11 18 28


Calanthe:
:+1:

PR #11363, opened 2015-12-14
add the related-alt ab test switch

johnduffell:
Forgot to add the switch, made it hard to turn on the test! @Jholder112233


Jholder112233:
:+1:

PR #11364, opened 2015-12-14
Separate fetching and parsing of merchandising feeds

kelvin-chappell:
This PR is just for jobs and soulmates feeds.

It has several benefits:
- Can use local copy of stale data if it's not possible to fetch a feed
- Have copy of bad data if it's not possible to parse the feed
- Have separate metrics for fetching and parsing
- Can continue to parse last known good data if current feed is bad

Still to do:
- Other feeds (masterclasses, books, travel, best buys etc)
- storing last good version of data separately from last data fetched
- Set alarms based on parsing metrics


commercial/app/commercial/CommercialLifecycle.scala
```diff
@@ -0,0 +1,160 @@
...
+
+      def storeFeed(feed: Feed): Unit =
+        S3.putPrivate(key = s"$merchandisingFeedsRoot/$feedName", value = feed.content, feed.contentType)
+
+      def recordFetch(maybeDuration: Option[Duration]): Unit = {
+        recordEvent(feedName, "fetch", maybeDuration)
+      }
+
+      val msgPrefix = s"Fetching $feedName feed"
+      log.info(s"Fetching $feedName feed from ${fetcher.url} ...")
```
kenlim: Maybe you should just reuse msgPrefix

commercial/app/commercial/CommercialLifecycle.scala
```diff
@@ -0,0 +1,160 @@
...
+    val key = s"${feedName.toLowerCase.replaceAll("[\\s/]+", "-")}-$eventName-time"
+    val duration = maybeDuration map (_.toMillis.toDouble) getOrElse -1d
+    CloudWatch.put("Commercial", Map(s"$key" -> duration))
+  }
+
+  override def onStart(app: PlayApp): Unit = {
+
+    val randomFactor = Random.nextInt(15)
+
+    def randomStartSchedule(minsLater: Int = 0) = s"0 ${randomFactor + minsLater}/15 * * * ?"
```
kenlim: Maybe move the randomFactor val right in here.

commercial/app/commercial/feeds/FeedParser.scala
```diff
@@ -0,0 +1,40 @@
...
+    for (agent <- SoulmatesAgent.agents) yield {
+      new FeedParser[Member] {
+        val feedName = s"soulmates/${agent.groupName}"
+
+        def parse(): Future[ParsedFeed[Member]] = agent.refresh()
+      }
+    }
+  }
+
+  val all = Seq(jobs) ++ soulmates
```
kenlim: Why is the signature between jobs and soulmates so different?

common/app/conf/switches/Switches.scala
```diff
@@ -46,6 +46,9 @@ trait SwitchTrait extends Switchable with Initializable[SwitchTrait] {
 
   def isSwitchedOn: Boolean = delegate.isSwitchedOn && new LocalDate().isBefore(sellByDate)
 
+  // true if switched on and switchboard has been read
```
kenlim: Change to "If the switchboard hasn't been properly initialised, then the "safe state" is returned, not the real switch value. This makes sure the switchboard is initialised before returning the answer"

commercial/app/model/commercial/jobs/JobsFeed.scala
```diff
@@ -1,52 +1,42 @@
...
-      parse(XML.loadString(body.dropWhile(_ != '<')))
-    }
-    } getOrElse {
-      log.warn(s"Missing URL for Jobs feed")
-      Future.failed(FeedMissingConfigurationException("Jobs"))
+  def parsedJobs(feedName: String): Future[ParsedFeed[Job]] = {
+    Switches.JobFeedSwitch.isGuaranteedSwitchedOn flatMap { switchedOn =>
+      if (switchedOn) {
+        val start = currentTimeMillis
+        S3.get(s"$merchandisingFeedsRoot/$feedName") map { body =>
```
kenlim: Should wrap this in a very obvious function that this is the "persistent" object that is giving us the safety.

commercial/app/commercial/CommercialLifecycle.scala
```diff
@@ -0,0 +1,160 @@
...
+
+    def fetchFeed(fetcher: FeedFetcher): Unit = {
+
+      val feedName = fetcher.feedName
+
+      def storeFeed(feed: Feed): Unit =
+        S3.putPrivate(key = s"$merchandisingFeedsRoot/$feedName", value = feed.content, feed.contentType)
```
kenlim: This is persisting our precious object. Should wrap into a function with obvious name that this is special.

commercial/app/commercial/feeds/FeedParser.scala
kelvin-chappell: There are many soulmates feeds and only one jobs feed.

PR #11366, opened 2015-12-14
Wait for DOM ready

OliverJAsh:
We used to wait for `DOMContentLoaded` before running any of our JS. I attempted to copy this across in #11258, but I inevitably broke it. We've had a about three complaints regarding images not loading on fronts, and this is why (Picturefill is running too early).

Here I have switched to using [ded’s `domReady`](https://github.com/ded/domready) (v0.3.0 for IE8 support) because the `domReady!` plugin doesn't fully cover the curl API (i.e. `require.next`). In any case, I don't think it makes sense to tie this to our module loader, and removing it means we can more easily reason about switching module loaders (although I won't be the one to instigate that…).

The `domReady` script is 459 bytes gzipped. I hoped we could switch to the standard version of curl, but sadly not, because [`curl+domReady` is the only distributed version that contains the `js!` plugin](https://github.com/cujojs/curl/tree/master/dist). Not a biggie because it doesn't add much weight (`curl+domReady+js` - `curl` = 957 bytes gzipped).

### DIFF
--- a/common/app/templates/inlineJS/blocking/curlConfig.scala.js
+++ b/common/app/templates/inlineJS/blocking/curlConfig.scala.js
@@ -16,6 +16,7 @@ window.curlConfig = {
             bonzo:                          'components/bonzo/bonzo',
             react:                          'components/react/react',
             classnames:                     'components/classnames/index',
+            domReady:                       'components/domready/ready',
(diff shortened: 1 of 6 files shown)


sndrs:
:+1:

PR #11369, opened 2015-12-15
Fix broken prefetch in old IE

OliverJAsh:
#11258 breaks the JS app in IE 9, because the prefetch fails due to a limitation whereby `XHMLHttpRequest` cannot make cross origin requests. Instead, IE 8 and 9 provide `XDomainRequest` for cross origin requests. reqwest uses this under the hood, so I switched to that.

/cc @gtrufitt


rich-nguyen:
i think we need a module called `prefetch`

gtrufitt:
:+1:

PR #11370, opened 2015-12-15
Show lazy loaded containers if not modern

OliverJAsh:
Several iOS users have complained they are missing containers on fronts. This is because we lazy load containers for breakpoints less than desktop, and the JS for doing this currently lives in standard.

The choices:
1. Move that JS to standard
2. Just show the containers in standard to avoid extra JS

I've sided with the latter. If we have performance problems with extra containers, we can re-introduce the lazy loading JS into standard. What do people think @sndrs @rich-nguyen?

### DIFF
--- a/static/src/stylesheets/module/facia/_container.scss
+++ b/static/src/stylesheets/module/facia/_container.scss
@@ -634,9 +634,10 @@ $header-image-size-desktop: 100px;
 .fc-container--lazy-load {
     @include mq($until: desktop) {
         display: none;
-    }
 
-    .is-not-modern & {
-        display: none;
+        .is-not-modern & {
+            // Unset
+            display: block;
+        }
     }
 }


sndrs:
the reason they are lazy loaded was for performance problems on older iphones, and there's still a switch for it i think? d' be surprised if those reasons had gone, but would it be easier to switch that off if you want to check perf?

OliverJAsh:
I remember hearing that. I'll do some testing to see how big of a difference there is with/without. Now that iOS devices only run standard, it may be less significant.

OliverJAsh:
Tested on an iPad 2 and it seems all OK to me.

jennysivapalan:
:+1:

OliverJAsh:
Merged in https://github.com/guardian/frontend/commit/1082348a9c317d8b3bcbd8fdafbd255b02566887

PR #11371, opened 2015-12-15
De-core header JS

sndrs:
Now that the standard/enhance JS is merged, we have a lot of ways of referring to different feature sets in the codebase. This (hopefully) clears them up, so we have one way of looking at what we do.
- retires `#featuresoff`/`off` and `#core`/`nocore` temporary hash opt-ins in favour of `#standard` or `#enhance`
- replaces `gu.prefs.force-core` with `gu.prefs.enhance`
- allows `#standard` to override enhance pref
- hopefully makes the logic more readable

it also currently allows for the old ones to be supported, but that can be removed over time.

once/if this is merged, I'll update the article.

cc @OliverJAsh @michaelwmcnamara @ScottPainterGNM


OliverJAsh:
I've been calling it standard and enhanced, whereas I see you call it enhance. Shall we decide on one for consistency?

sndrs:
gah, yeah - enhanced is throughout the app, i'll change it to that

common/app/templates/inlineJS/blocking/shouldEnhance.scala.js
```diff
@@ -1,54 +1,98 @@
...
+        var coreKey = 'gu.prefs.force-core', settingCorePref, corePref;
+
+        // update any `force-core` stored pref
         try {
-            var preference = window.localStorage.getItem('gu.prefs.force-core') || 'unknown';
-            return /"value":"off"/.test(preference);
-        } catch (e) {
-            return false;
+            var localStorage = window.localStorage;
+            if (corePref = localStorage.getItem(coreKey)) {
```
OliverJAsh: Nitpicking here, but I think it hurts readability to combine an assignment with an if statement, just because it's more complex.

common/app/templates/inlineJS/blocking/shouldEnhance.scala.js
```diff
@@ -1,54 +1,98 @@
...
+
+    function isIpad() {
+        return (platform === 'iPad');
+    };
+
+    // clean up old names before checking the ones we'll expect from now on
+    normaliseNomenclature();
+
+    // down to business
+    window.shouldEnhance = mustNotEnhance() ? false : mustEnhance() ? true : couldEnhance() && weWantToEnhance();
```
OliverJAsh: `mustNotEnhance` and `mustEnhance` smell like they could be consolidated into a single boolean function? Could be wrong!

OliverJAsh: Ah I see why now!

OliverJAsh: I find this condition hard to get my head around. I'm not sure if this is correct/any better?

``` js
window.shouldEnhance = mustEnhance() || (couldEnhance() && weWantToEnhance() && !mustNotEnhance())
```

OliverJAsh:
:+1: Nice tidy up.

common/app/templates/inlineJS/blocking/shouldEnhance.scala.js
sndrs: the logic is different there - `mustNotEnhance()` needs to take precedence.

common/app/templates/inlineJS/blocking/shouldEnhance.scala.js
```diff
@@ -1,54 +1,100 @@
...
+        var coreKey = 'gu.prefs.force-core', settingCorePref, corePref;
+
+        // update any `force-core` stored pref
         try {
-            var preference = window.localStorage.getItem('gu.prefs.force-core') || 'unknown';
-            return /"value":"off"/.test(preference);
-        } catch (e) {
-            return false;
+            var localStorage = window.localStorage,
+                corePref = localStorage.getItem(coreKey);
```
sndrs: yeah this one makes little difference actually, but on https://github.com/guardian/frontend/pull/11371/files#diff-3cec1f877cbfb319fa115c609baaadeaR46 it save some setting and checking, which for this header stuff i'd rather do.

PR #11372, opened 2015-12-15
A/B test definition for Identity Sign in v2 test

jamespamplin:
The identity team have been working on a new variant for Sign in pages. The source is separated from frontend and is available at https://github.com/guardian/identity-frontend - which is hosted at https://profile-origin.theguardian.com/signin

This PR defines a A/B test definition for recording the outcome of the new pages. This is required by the data team to record test outcomes for reporting.

We'll control the participation of users for the test by routing with fastly on https://profile.theguardian.com - which will route 10% of users to the new service. The new service will automatically record ab event data with omniture and ophan.

I've spoken to the data team about decoupling this process, but we'll have to look at this next year.

/cc @rich-nguyen

### DIFF
--- a/common/app/conf/switches/ABTestSwitches.scala
+++ b/common/app/conf/switches/ABTestSwitches.scala
@@ -39,4 +39,13 @@ trait ABTestSwitches {
     sellByDate = new LocalDate(2016, 1, 6),
     exposeClientSide = true
   )
+
+  val ABIdentitySignInV2 = Switch(
+    "A/B Tests",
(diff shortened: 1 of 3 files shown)


rich-nguyen:
I think you will need to add a switch to your test too, in `ABTestSwitches.scala`, following the existing naming convention.

jamespamplin:
Thanks @rich-nguyen, added Switch in `ABTestSwitches.scala`.

rich-nguyen:
:+1:

PR #11373, opened 2015-12-15
Fix a race condition on the stripe icons

tomverran:
The digital pack stripe form will attempt to set an empty card icon until it becomes visible all the time


tomverran:
Merged into https://github.com/guardian/frontend/pull/11377

PR #11375, opened 2015-12-15
Wait for DOM ready, attempt 2

OliverJAsh:
#11366 didn't work. The body of a module is actually executed at `define` time, not at `require` time. Thus, `bootstraps/standard/main` was executing before the DOM was ready.

https://github.com/guardian/frontend/pull/11375/files?w=1


rich-nguyen:
:+1:

PR #11377, opened 2015-12-15
Digitalpack attempt 2

tomverran:
The original PR worked in DEV but not in PROD due to the digitalpack bootstraps not being compiled properly by grunt. The easiest solution is to use the already existing membership one, in line with the rest of the JS.


afiore:
:+1:

PR #11380, opened 2015-12-15
Cross origin and ajax url handled in ajax.js

gtrufitt:


static/src/javascripts/projects/common/modules/email/email.js
```diff
@@ -103,7 +103,7 @@ define([
         },
         formSubmission = {
             bindSubmit: function ($form, analytics) {
-                var url = config.page.ajaxUrl + '/email';
+                var url = '/email';
```
OliverJAsh: For future reference: we don't add the ajax URL here because [the Ajax helper will only set `crossOrigin: true` (needed for cross origin requests via `XDomainRequest` in IE 8/9) if you pass in a relative URL](https://github.com/guardian/frontend/blob/70dc8934db297e8b3f31e6194761f6232fcf50a0/static/src/javascripts/projects/common/utils/ajax.js#L18-L21). :unamused: 

The short story: remember to set `crossOrigin: true` if you care about IE 8/9, or pass in a relative URL and we'll handle this for you.

static/src/javascripts/projects/common/modules/email/email.js
```diff
@@ -149,7 +149,12 @@ define([
                                 omniture.trackLinkImmediate('rtrt | email form inline | ' + analytics.formType + ' | ' + analytics.listId + ' | subscribe successful');
                             })
                             .then(handleSubmit(true, $form))
-                            .catch(function () {
+                            .catch(function (error) {
+                                /* eslint-disable no-console */
+                                if (window.console && console.error) {
+                                    console.error('Error posting email form', error.stack);
+                                }
+                                /* eslint-enable no-console */
```
OliverJAsh: Use robust

OliverJAsh:
:+1:

PR #11381, opened 2015-12-15
Related Content title change

stephanfowler:
Prior to user-testing of tag-based related content, re-re-name the container without the word "related", in order to not set up a false expectation of what's in it.

@johnduffell


johnduffell:
why not, what's in a name anyway? :+1: :stuck_out_tongue:

crifmulholland:
I like 'further reading'.

(Pull request was slightly cryptic though!)

johnduffell:
@crifmulholland I think in the long term we'd like to move the tags closer and perhaps mention them "read more about <tag_1>, <tag_2> and <tag_3>" although it does depend somewhat on what the final implementation of tag driven related content is in CAPI.

sndrs:
still got it :)

PR #11382, opened 2015-12-16
Fix updating of force-core pref

sndrs:
was setting a boolean, but the format for prefs is `{value: "x"}`. dunno why.

cc @OliverJAsh


OliverJAsh:
:+1:

PR #11384, opened 2015-12-16
Tip to downgrade to node 4

gidsg:
I ran into this issue when running node 5.0. CC @OliverJAsh @rich-nguyen


OliverJAsh:
:+1:

My advice would be not to install Node at all. :tongue:

PR #11387, opened 2015-12-17
use CAPI blocks instead of body to produce live blogs

johnduffell:
This is needed in preparation to make live blogs paginated.

This uses the blocks from capi to render a liveblog, and accepts a page parameter (which is currently ignored)

The next stage is to only render a subset of the blocks based on the page parameter passed in, also add some prev/next controls and metadata.

@stephanfowler


article/app/controllers/ArticleController.scala
```diff
@@ -104,26 +110,35 @@ object ArticleController extends Controller with RendersItemResponse with Loggin
...
   }
 
-  private def loadArticle(path: String, lastUpdate: Option[String], rendered: Option[Boolean])(implicit request: RequestHeader): Future[Result] = {
+  private def loadArticle(path: String, lastUpdate: Option[String], rendered: Option[Boolean], page: Option[String])(implicit request: RequestHeader): Future[Result] = {
     mapModel(path) { model =>
-      (lastUpdate, rendered) match {
-        case (Some(lastUpdate), _) => renderLatestFrom(model, lastUpdate)
-        case (None, Some(false)) => blockText(model, 6)
-        case (_, _) => render(path, model)
+      (lastUpdate, rendered, page) match {
```
philwills: This looks to be getting pretty difficult to scan.

article/app/controllers/ArticleController.scala
```diff
@@ -80,14 +81,19 @@ object ArticleController extends Controller with RendersItemResponse with Loggin
 
   }
 
-  private def render(path: String, page: PageWithStoryPackage)(implicit request: RequestHeader) = page match {
+  private def render(path: String, page: PageWithStoryPackage, pageNo: Option[Int])(implicit request: RequestHeader) = page match {
```
jennysivapalan: you could default `pageNo` to `None` so it's only passed in when needed?

article/app/controllers/ArticleController.scala
```diff
@@ -104,26 +110,35 @@ object ArticleController extends Controller with RendersItemResponse with Loggin
...
+        case (Some(lastUpdate), _, _) => renderLatestFrom(model, lastUpdate)
+        case (None, Some(false), _) => blockText(model, 6)
+        case (_, _, Some(Int(pageNo))) => render(path, model, Some(pageNo))
+        case (_, _, _) => render(path, model, None)
       }
     }
   }
 
+  object Int {
+    def unapply(s : String) : Option[Int] = try {
```
philwills: I'm not a big fan of custom `unapply`s. I'm not convinced their usefulness outweighs the amount of explanation required to those who haven't seen them before.

article/app/controllers/ArticleController.scala
```diff
@@ -80,14 +81,19 @@ object ArticleController extends Controller with RendersItemResponse with Loggin
...
       } else {
-        val htmlResponse = () => views.html.liveBlog(blog)
-        val jsonResponse = () => views.html.fragments.liveBlogBody(blog)
-        renderFormat(htmlResponse, jsonResponse, blog, Switches.all)
+        val blocks = BlocksFor(blog.article.content.blocks, pageNo)
+        blocks match {
+          case Some(blocks) =>
+            val htmlResponse = () => views.html.liveBlog (blog, blocks)
+            val jsonResponse = () => views.html.fragments.liveBlogBody (blog, blocks)
+            renderFormat (htmlResponse, jsonResponse, blog, Switches.all)
```
jennysivapalan: Minor one but we don't normally leave a space between fragment and passing parameters.

article/app/layout/LiveBlogs.scala
```diff
@@ -0,0 +1,32 @@
...
+    case UnclassifiedEvent => ""
+  }
+
+}
+
+object LiveBlogDateFormat {
+
+  def apply(dateTime: DateTime)(implicit request: RequestHeader) = {
+    val fullDate = ISODateTimeFormat.dateTime().withZone(DateTimeZone.UTC).print(dateTime)
+    val hhmm = dateTime.toString(DateTimeFormat.forPattern("HH:mm").withZone(Edition(request).timezone))
```
philwills: Can we not pull the formats out into constants?

article/app/controllers/ArticleController.scala
```diff
@@ -104,26 +110,35 @@ object ArticleController extends Controller with RendersItemResponse with Loggin
       renderFormat(htmlResponse, jsonResponse, article, Switches.all)
   }
 
-  def renderArticle(path: String, lastUpdate: Option[String], rendered: Option[Boolean]) = {
+  def renderArticle(path: String, lastUpdate: Option[String], rendered: Option[Boolean], page: Option[String]) = {
```
jennysivapalan: Does `page` mean page number here? If so why is it a `String`?

common/app/model/content.scala
```diff
@@ -233,7 +237,52 @@ object Content {
     }
   }
 
-  def make(apiContent: contentapi.Content): Content = {
+  object BodyBlocks {
+    def make(blocks: Option[Blocks]): Seq[BodyBlocks] =
+    blocks.toSeq.flatMap { blocks =>
```
philwills: I'm not a fan of shading argument names like this.

article/app/controllers/ArticleController.scala
```diff
@@ -104,26 +110,35 @@ object ArticleController extends Controller with RendersItemResponse with Loggin
...
-      loadArticle(path, lastUpdate, rendered)
+      loadArticle(path, lastUpdate, rendered, page)
     } else MemcachedAction { implicit request =>
-      loadArticle(path, lastUpdate, rendered)
+      loadArticle(path, lastUpdate, rendered, page)
     }
   }
 
-  private def loadArticle(path: String, lastUpdate: Option[String], rendered: Option[Boolean])(implicit request: RequestHeader): Future[Result] = {
+  private def loadArticle(path: String, lastUpdate: Option[String], rendered: Option[Boolean], page: Option[String])(implicit request: RequestHeader): Future[Result] = {
```
jennysivapalan: As above I'm not sure why page is a `String`?

article/app/layout/LiveBlogs.scala
```diff
@@ -0,0 +1,32 @@
...
+import org.joda.time._
+import org.joda.time.format._
+import play.api.mvc.RequestHeader
+
+object BlockFields {
+
+  def apply(block: BodyBlocks) = block.eventType match {
+    case SummaryEvent => " is-summary"
+    case KeyEvent => " is-key-event"
+    case UnclassifiedEvent => ""
```
jennysivapalan: Is the space needed before is-summary and is-key-event?

common/app/model/content.scala
jennysivapalan: @philwills I'm don't follow what you mean, can you explain further?

common/app/model/content.scala
```diff
@@ -233,7 +237,52 @@ object Content {
...
+    bodyHtml: scala.Predef.String,
+    bodyTextSummary: scala.Predef.String,
+    title: scala.Option[scala.Predef.String],
+    attributes: scala.Predef.Map[scala.Predef.String, scala.Predef.String],
+    published: scala.Boolean,
+    createdDate: scala.Option[org.joda.time.DateTime],
+    firstPublishedDate: scala.Option[org.joda.time.DateTime],
+    publishedDate: scala.Option[org.joda.time.DateTime],
+    lastModifiedDate: scala.Option[org.joda.time.DateTime],
+    contributors: scala.Seq[scala.Predef.String]
```
jennysivapalan: can these be shortened? e.g. `contributors: Seq[String]` ?

common/app/model/content.scala
philwills: There is a parameter to `make` called `blocks` of type `Option[Blocks]`, but then inside the flatMap, there's also the parameter of the anonymous function called `blocks` of type `Blocks` which hides the `Option[Blocks]` one from the inner scope.

article/app/controllers/ArticleController.scala
johnduffell: it is then because that's how play passes it in, then later we can parse to an int.  perhaps there is a way to ask play to do the conversion like it does with Boolean?

philwills: I think Play will just do that, but if not, I'd definitely try and convert it here, rather than further down the stack.

article/app/layout/LiveBlogs.scala
johnduffell: yes just because I'm being lazy and using string concatentation rather than mkString

article/app/layout/LiveBlogs.scala
johnduffell: not sure it helps, but you're right in that the code is a mess

common/app/model/content.scala
johnduffell: I actually do it on purpose a lot, but it's an interesting debate.  I definitely don't want to use the outer one in the inner block so from a technical point of view it's a good idea.  The drawback from a readability point of view is github doesn't understand scala well so it's not clear which one I'm using without looking carefully. But if I think too hard that opens a whole debate about why we store code as text files in the first place... :stuck_out_tongue_closed_eyes:

common/app/model/content.scala
johnduffell: oh yeah I copied from a decompiled class :sheep:

common/app/model/content.scala
philwills: From my point of view, it's just too easy to skim over and assume that every reference to `blocks` is a reference to the top-level parameter, whether in Github or IntelliJ.

johnduffell: true, I decided to open my mind a bit to the debate to see what the rest of the internet says, and there's not much.  The best I could find was this: http://stackoverflow.com/questions/7031106/why-does-scala-support-shadow-variables

johnduffell: ok @philwills  I've moved it into its own file and made all the variable names sensible, looks better now :grinning:

johnduffell:
@jennysivapalan @philwills thanks for all your comments!  I think I have addressed everything if you wanted another look - let me know if I missed any =)

philwills:
:+1: to the changes from me.

jennysivapalan:
Looks good :+1:, teamcity is failing though.

common/app/model/content.scala
```diff
@@ -423,7 +424,7 @@ final case class Article (
 
   val lightbox = GenericLightbox(content.elements, content.fields, content.trail, lightboxProperties)
 
-  val isLiveBlog: Boolean = content.tags.isLiveBlog
+  val isLiveBlog: Boolean = content.tags.isLiveBlog && content.fields.blocks.nonEmpty
```
johnduffell: now we only get the blocks if there is live in the url, and if there are no blocks it assumes it's an article not a live blog.  This does cause some formatting issues for old live blogs without live in the url, but Matt W is happy enough with that since they don't get much traffic and they will be fixed anyway once we finish the migration and switch to using blocks for everything.

article/app/controllers/ArticleController.scala
```diff
@@ -160,3 +175,15 @@ object ArticleController extends Controller with RendersItemResponse with Loggin
   }
 
 }
+
+object LongCacheAction {
+  def apply[X](block: RequestHeader => Future[Result]) = {
```
OliverJAsh: Is this covariant needed? I don't see `X` used anywhere, but I'm probably wrong.

article/app/controllers/ArticleController.scala
```diff
@@ -104,42 +111,50 @@ object ArticleController extends Controller with RendersItemResponse with Loggin
...
   }
 
-  def renderArticle(path: String, lastUpdate: Option[String], rendered: Option[Boolean]) = {
-    if (LongCacheSwitch.isSwitchedOn) Action.async { implicit request =>
-      // we cannot sensibly decache memcached (does not support surogate keys)
-      // so if we are doing the 'soft purge' don't memcache
-      loadArticle(path, lastUpdate, rendered)
-    } else MemcachedAction { implicit request =>
-      loadArticle(path, lastUpdate, rendered)
+  def renderLive(path: String, page: Option[Int] = None) =
```
OliverJAsh: Minor note but I misread this as rendering “live” content, when it’s actually rendering ”live blogs”. You might want to rename it

article/app/controllers/ArticleController.scala
philwills: \<pedant\>It's a type parameter, which in this case isn't covariant, that would be `[+X]`\</pedant\>

I agree it seems redundant though.

OliverJAsh: Thanks for the tip :small_orange_diamond:

jennysivapalan:
:+1:

article/app/controllers/ArticleController.scala
johnduffell: oops

article/app/layout/LiveBlogs.scala
```diff
@@ -0,0 +1,35 @@
+package layout
```
rich-nguyen: Is the `layout` directory a bit confusing? It's a fairly loaded name, I thought it was to do with facia-like layout. maybe move this one to `article/app/liveblog/blockformat.scala`?

article/app/controllers/ArticleController.scala
rich-nguyen: I don't think you'll need the page parameter here, in future the page will be inferred from `request` using `inferPage`.

article/app/liveblog/BlocksFor.scala
```diff
@@ -0,0 +1,16 @@
+package liveblog
+
+import model.liveblog.BodyBlock
+
+object BlocksFor {
```
rich-nguyen: I think this object would be more readable if it were called `BlocksForPage`...

article/app/liveblog/BlocksFor.scala
```diff
@@ -0,0 +1,16 @@
...
+  def apply(blocks: Seq[BodyBlock], page: Option[Int]): Option[BlocksFor] = {
+    page match {
+      case Some(pageNo) => None
+      case None => Some(BlocksFor(blocks, 1))//TODO use page to produce different blocks
+    }
+  }
+
+}
+
+case class BlocksFor(blocks: Seq[BodyBlock], pageNo: Int)// TODO prev/next
```
rich-nguyen: ...And this case class could have a name like `BodyBlocks`?

common/app/model/content.scala
```diff
@@ -452,6 +453,7 @@ final case class Article (
   private lazy val soupedBody = Jsoup.parseBodyFragment(fields.body)
   lazy val hasKeyEvents: Boolean = soupedBody.body().select(".is-key-event").nonEmpty
   lazy val isSport: Boolean = tags.tags.exists(_.id == "sport/sport")
+  //@deprecated("use content.blocks", "")
```
rich-nguyen: This looks like a deprecated `@deprecated` annotation!

common/app/model/liveblog/BodyBlock.scala
```diff
@@ -0,0 +1,51 @@
...
+  case object UnclassifiedEvent extends EventType
+}
+
+case class BodyBlock(
+  id: String,
+  bodyHtml: String,
+  bodyTextSummary: String,
+  title: Option[String],
+  attributes: Map[String, String],
+  published: scala.Boolean,
```
rich-nguyen: `Boolean`

common/app/views/support/HtmlCleaner.scala
```diff
@@ -179,87 +177,6 @@ case class PictureCleaner(article: Article, amp: Boolean)(implicit request: Requ
...
-            el.getElementsByClass("block-elements").headOption.foreach(_.before(html.toString()))
-          }
-        }
-      }
-    }
-    body
-  }
-}
-
-case class LiveBlogShareButtons(article: Article)(implicit val request: RequestHeader) extends HtmlCleaner  {
```
rich-nguyen: What happens to block-level sharing now?

rich-nguyen:
lovely stuff :+1:

article/app/layout/LiveBlogs.scala
johnduffell: actually I've got rid of this locally I just haven't pushed yet

common/app/model/content.scala
johnduffell: I want it to be deprecated for future use, but adding a deprecated annotation stops any old usage of it from compiling.  I just wanted it to be a warning but I couldn't do that.

common/app/views/support/HtmlCleaner.scala
johnduffell: still seems to work:
image
it's just done in a play template based on the capi blocks rather than being mixed in to the html in the cleaner

PR #11389, opened 2015-12-18
Extend Fronts on Articles test and rename

holder-josh:
Moving it to a new offset, lowering it to a 6% test and renaming. Leaving it running over xmas.


Jholder112233:
@stephanfowler this is all sensible, right?

stephanfowler:
:+1: LGTM

PR #11391, opened 2015-12-18
Progressively Enhance email form

gtrufitt:
We've got some cross-origin/protocol/content type/header issues that need to be resolved, but in the meantime we can progressively enhance the email form so that it just submits as a normal form in lte ie-9

### DIFF
--- a/static/src/javascripts/projects/common/modules/email/email.js
+++ b/static/src/javascripts/projects/common/modules/email/email.js
@@ -9,6 +9,7 @@ define([
     'Promise',
     'common/utils/mediator',
     'lodash/functions/debounce',
+    'lodash/collections/contains',
     'common/utils/template',
     'common/views/svgs',
     'text!common/views/email/submissionResponse.html',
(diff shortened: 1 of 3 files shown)


static/src/javascripts/projects/common/modules/email/email.js
```diff
@@ -249,19 +251,25 @@ define([
...
-
-                // We're loading through the iframe
-                if (rootEl && rootEl.tagName === 'IFRAME') {
-                    // We can listen for a lazy load or reload to catch an update
-                    setup(rootEl, rootEl.contentDocument.body, true);
-                    bean.on(rootEl, 'load', function () {
+                var browser = detect.getUserAgent.browser,
+                    version = detect.getUserAgent.version;
+                // If we're in lte IE9, don't run the init and adjust the footer
+                if (browser === 'MSIE' && ['7','8','9'].indexOf(version + '') !== -1) {
```
OliverJAsh: FYI, Lodash has a `contains` fn that would be useful for checking the version.

OliverJAsh:
:+1: Nice

PR #11396, opened 2015-12-18
Cleans an R2 html file and presses it to an S3 bucket

jennysivapalan:
Given a url we will have a `PagePresser` that will fetch the page, remove some components and press the file to S3. 

The components removed are
- Ads and promo column
- Share links
- Related component (on examples many of the links redirected to taken down pages)
- Google custom search box.

As a separate PR I will add the removal of omniture tracking and replace is with a small js file for page tracking.

This looks like a big diff as I added examples for testing.

cc @JustinPinner


JustinPinner:
:+1: Won't get up to anything until we get to testing it :smile:

PR #11398, opened 2015-12-18
AB alternative related content - add some audience

stephanfowler:
Puts 1% audience on the variant over the christmas period.

@johnduffell


johnduffell:
:+1: great!

PR #11400, opened 2015-12-18
Use picture element

OliverJAsh:
We currently accidentally serve retina images to high DPR devices on mobile (because the `srcset` contains images for wider breakpoints, but the browser picks this for high DPRs). @paperboyo came up with the idea of using `<picture>` in combination with `srcset/sizes` to avoid this. Instead of:

``` html
<img src="300.jpg" sizes="(min-width: 500px) 600px, 300px" srcset="600.jpg 600w, 300.jpg 300w" />
```

We would do this:

``` html
<picture>
  <source media="(min-width: 500px)" sizes="600px" srcset=" 600.jpg 600w" />
  <img src="300.jpg" />
</picture>
```

This means the `srcset` containing the higher resolution image won't be available until the media query is met.

I ran this as a test locally. When viewing a front as DPR 2, with the picture method all images downloaded in 600 KB, whereas our current method downloaded in 1 MB. That's a big saving!

We run a modified version of picturefill. I added back the necessary parts to polyfill `<picture>`. I've heard that we've suffered with perf issues using picturefill once upon a time, especially Mobile Safari, so I will make sure this gets tested rigorously.

@johnduffell @jennysivapalan @rich-nguyen I'm not too sure about my Scala…
# To do
- [x] Add [comment workaround for IE 9](https://scottjehl.github.io/picturefill/)
- [x] Apply change to content pages

### DIFF
--- a/common/app/views/support/Profile.scala
+++ b/common/app/views/support/Profile.scala
@@ -4,7 +4,7 @@ import java.net.{URI, URISyntaxException}
 import common.Logging
 import conf.switches.Switches.ImageServerSwitch
 import conf.Configuration
-import layout.WidthsByBreakpoint
+import layout.{BreakpointWidth, WidthsByBreakpoint}
 import model._
 import org.apache.commons.math3.fraction.Fraction
 import org.apache.commons.math3.util.Precision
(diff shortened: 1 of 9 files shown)


jennysivapalan:
:+1: for Scala

common/app/views/support/Profile.scala
```diff
@@ -135,19 +135,30 @@ object ImgSrc extends Logging {
...
-    } mkString ", "
+    widths.profiles.map { profile => srcsetForProfile(profile, imageContainer) } mkString ", "
   }
 
-  def srcset(path: String, widths: WidthsByBreakpoint): String = {
-    widths.profiles map { profile =>
-      s"${ImgSrc(path, profile)} ${profile.width.get}w"
-    } mkString ", "
+  def srcsetForBreakpoint(breakpointWidth: BreakpointWidth, widths: WidthsByBreakpoint, maybeUrl: Option[String] = None, maybeImageContainer: Option[ImageContainer] = None) = {
+    widths.breakpointWidthToPixels(breakpointWidth)
```
OliverJAsh: I wonder if the `breakpointWidthToPixels` method makes more sense on the `BreakpointWidth` method, but it needs scope of the `WidthsByBreakpoint`.

oliverlloyd:
This PR is getting a lot of attention, 5 years after it was implemented! For anyone else arriving here wondering why this change was done and if we still need it, check this out: https://blog.twitter.com/engineering/en_us/topics/infrastructure/2019/capping-image-fidelity-on-ultra-high-resolution-devices.html This is Twitter noticing the same problem a year ago and fixing it (most likely) in the same way. It just took them 4 years longer to get there.

OliverJAsh:
I wrote an article on the subject if it helps: https://oliverjash.me/2016/05/12/pitfalls-of-the-img-element-for-serving-responsive-images

PR #11401, opened 2015-12-18
Page pressing service

JustinPinner:
_What is this?_

It's two things: It's a UI component through which someone can enter an **old R2 URL** (or upload a file of **old R2 URLs**), and have them added to an SQS queue. And it's also a scheduled task that monitors that queue, extracting the URL(s), and then uses @jennysivapalan's load-and-clean code to retrieve the current html and remove all the things we don't want pressed to the static file. 

When loaded and cleaned we write the original html to a backup location and the cleaned html to the pressed files repository.

The UI also supports the taking down of individual and batches of URLs.

There will also be some associated Fastly changes to allow the pressed urls to be served from NGW instead of the legacy backend.


admin/app/controllers/R2PressController.scala
```diff
@@ -0,0 +1,86 @@
...
+
+import java.io.File
+
+import common.{ExecutionContexts, Logging}
+import controllers.AuthLogging
+import play.api.mvc.{AnyContent, Controller}
+import services.{R2PressedPageTakedownNotifier, R2PagePressNotifier}
+
+case class R2PagePress(r2url: String) {
+  def trim = this.copy(r2url = r2url.trim)
```
jennysivapalan: can we make this it's only variable rather than copying into `r2url`?

jennysivapalan:
This looks good! I think my only concern is the changes in SQS and ensuring existing behaviour continues to work. Let's grab @janua when he is free to chat about testing.

JustinPinner:
@janua would you take a look at `common/app/common/SQSQueues.scala` and see if you can identify any potential issues these changes may have from the perspective of fronts? Also, _how do we test that_?

admin/app/services/PagePresses.scala
```diff
@@ -0,0 +1,37 @@
...
+import conf.Configuration
+import scala.collection.JavaConversions._
+import com.amazonaws.services.dynamodbv2.model.AttributeValue
+import play.api.Play.current
+import play.api.Play
+
+object PagePresses extends Logging {
+  //TODO: restore prod table when ready!
+  //private lazy val table = if (Play.isProd) "redirects" else "redirects-DEV"
+  private lazy val table = "redirects-DEV"
```
janua: You're going to put it to PROD with for testing I presume?

admin/app/controllers/R2PressController.scala
```diff
@@ -0,0 +1,86 @@
...
+      files.file("r2urlfile").map { theFile =>
+        val rnd = Math.random().toString.replace(".","")
+        val tmpName = s"/tmp/$rnd${theFile.filename}"
+        val tmpFile = new File(tmpName)
+        theFile.ref.moveTo(tmpFile)
+        tmpFile
+      }
+    }
+    uploadedFile.foreach(file => pressFile(file, isBatchTakedown(body)))
+    Ok(views.html.pressR2())
```
janua: Are you sure you want to just ignore what `uploadedFile` was and always return `Ok` regardless of what happened? Wouldn't it be better to match the value?

admin/app/controllers/R2PressController.scala
```diff
@@ -0,0 +1,86 @@
...
+              R2PressedPageTakedownNotifier.enqueue(url)
+            } else {
+              R2PagePressNotifier.enqueue(url)
+            }
+          }
+          case _ =>
+        }
+      }
+    }
+    SeeOther(routes.R2PressController.pressForm().url)
```
janua: Same as above!

admin/app/jobs/R2PagePressJob.scala
```diff
@@ -0,0 +1,113 @@
...
+  val cleaners = Seq(BasicHtmlCleaner, PollsHtmlCleaner)
+
+  val queue: TextMessageQueue[SNSNotification] = (Configuration.r2Press.sqsQueueUrl map { queueUrl =>
+    val credentials = Configuration.aws.mandatoryCredentials
+
+    TextMessageQueue[SNSNotification](
+      new AmazonSQSAsyncClient(credentials).withRegion(Region.getRegion(Regions.EU_WEST_1)),
+      queueUrl
+    )
+  }) getOrElse {
```
janua: You could just `getOrElse` the configuration url directly (and inline if you wanted), makes it much easier to read!

admin/app/jobs/R2PagePressJob.scala
```diff
@@ -0,0 +1,113 @@
...
+      } catch {
+        case e: Exception => log.error(s"Failed to decode r2 url: ${e.getMessage}", e)
+      }
+    } else {
+      log.info("R2PagePressJob is switched OFF")
+    }
+  }
+
+  private def press(message: Message[String]) {
+    val urlIn = (Json.parse(message.get) \ "Message").as[String]
```
janua: I don't understand this, why are you parsing it if you are meant to be using `TextMessageQueue`?

admin/app/services/PagePresses.scala
JustinPinner: That's the plan. Don't want to risk contaminating PROD pressed pages at all.

admin/app/controllers/R2PressController.scala
JustinPinner: The pressing happens asynchronously so we can't wait for the file contents to be processed but maybe It'll be worth adding a test for uploadedFile.nonEmpty though.

admin/app/jobs/R2PagePressJob.scala
JustinPinner: Ah, the AWS terminology. We get a message from the queue (`message.get`), which is a json object. This contains a `Message` field which contains the string value (rather than a json-encoded object) we're interested in - the url to be pressed in this case.

admin/app/controllers/R2PressController.scala
janua: Do you mean asynchronously in scala or async in an architecture level?

admin/app/jobs/R2PagePressJob.scala
janua: Are you sure about that? I thought the bytes from an SQS message could be anything you wanted (As in, it wasn't forced into JSON)

JustinPinner: That's true, but I followed an existing message creation pattern we use which deposits JSON messages on a queue, only for this particular use case they contain a plain string message field which is the URL.

jennysivapalan:
:+1: :+1: :+1: 

:chocolate_bar:

PR #11402, opened 2015-12-18
Test - End of article email sign up from fronts

gtrufitt:
When a user goes to an article from a front, we want to show them the email sign up, we're using an A/B test to segment and roll out to 50%.


janua:
:+1:

PR #11404, opened 2015-12-21
Commercial containers

piuccio:
Implement two new commercial containers for single and multi campaigns.

For the moment it simply renders as a fixed container, small slow 4.

Someone else should pick this up and actually implement the rendering.

@regiskuckaertz @janua

### DIFF
--- a/common/app/slices/CommercialContainer.scala
+++ b/common/app/slices/CommercialContainer.scala
@@ -1,0 +1,13 @@
+package slices
+
+object CommercialContainer {
+  val all: Map[String, Container] = Map(
+    ("commercial/single-campaign", Commercial(SingleCampaign(FixedContainers.fixedSmallSlowIV))),
+    ("commercial/multi-campaign", Commercial(MultiCampaign(FixedContainers.fixedSmallSlowIV)))
+  )
+}
+
+sealed trait CommercialContainer
+
+case class SingleCampaign(get: ContainerDefinition) extends CommercialContainer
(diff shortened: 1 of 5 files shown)


regiskuckaertz:
:metal:

common/app/slices/CommercialContainer.scala
```diff
@@ -0,0 +1,13 @@
+package slices
+
+object CommercialContainer {
+  val all: Map[String, Container] = Map(
+    ("commercial/single-campaign", Commercial(SingleCampaign(FixedContainers.fixedSmallSlowIV))),
+    ("commercial/multi-campaign", Commercial(MultiCampaign(FixedContainers.fixedSmallSlowIV)))
+  )
+}
+
+sealed trait CommercialContainer
```
janua: Not so sure about this name... Maybe `CommercialCampaign`?
The thing on the outside of this is the `Container`, while the thing inside is the `ContainerDefinition`. This middle part has never exited before but I don't think it's strictly containter!

janua:
Great! :+1:

common/app/slices/CommercialContainer.scala
piuccio: The thing about `campaign` is that who knows what types of commercial containers we'll have to create in the future. `campaign` is what these two containers do right now, it does not describe what they are

janua: Time to get out the dictionary :books:

PR #11405, opened 2015-12-21
Create Bashers and add Keep It In The Ground Basher

sammorrisdesign-zz:
### Banner + Thrasher = Bashers (@rich-nguyen came up with that, so blame him)

Merry Christmas from your pal, Samta Claus :santa: 

So a few months ago I created a page banner for [The Keep It In The Ground Campaign Page](www.theguardian.com/environment/series/keep-it-in-the-ground) by using a thrasher. It worked fine but seeing as it's always going to above the metaphorical fold it wasn't ideal. Especially on mobile.

We're creating another one so this seemed like the time to do it properly in frontend on page load. I've tried to mimic the setup for the Treats (see [`Treats.scala`](https://github.com/guardian/frontend/blob/77e10c3bfd40fc4910546796d8aecc1ab7739f5e/common/app/views/support/Treats.scala) and [`Treats.scala.html`](https://github.com/guardian/frontend/blob/77e10c3bfd40fc4910546796d8aecc1ab7739f5e/common/app/views/fragments/nav/treats.scala.html)).

I've also picked an example by using the Keep It In The Ground basher as the first basher.

Before:
loading

After:
screen shot 2015-12-21 at 17 48 24


static/src/stylesheets/inline/basher.KeepItInTheGround.scss
```diff
@@ -0,0 +1,205 @@
...
+    right: 0;
+    bottom: 0;
+    z-index: 10 !important;
+
+    &:hover ~ .keep-it-in-the-ground__content {
+        strong {
+            color: $c-accent-dark;
+        }
+    
+        .keep-it-in-the-ground__button {
```
sndrs: :eyes: 

any way you could shorten this? 
`.keep-it-in-the-ground__anchor:hover ~ .keep-it-in-the-ground__content .keep-it-in-the-ground__button`

common/app/views/support/Basher.scala
```diff
@@ -0,0 +1,15 @@
...
+
+object Basher {
+    def fromId(id: String) = {
+        if (id == "environment/series/keep-it-in-the-ground") {
+            KeepItInTheGround
+        } else {
+            None
+        }
+    }
+}
```
OliverJAsh: This function returns a type of `Any` which is bad because you lose type safety. You should be explicit about what type you are returning:

``` scala
def fromId(id: String): Option[Basher] = {
```

It looks like you want to return an option? So either `None` or `Some(value)`?

common/app/assets/assets.scala
```diff
@@ -86,6 +86,7 @@ class Assets(base: String, assetMapPath: String = "assets/assets.map") extends L
        val knownInlines : PartialFunction[String,String] =
        {
          case "story-package" => "story-package.css"
+         case "KeepItInTheGround" => "basher.KeepItInTheGround.css"
```
OliverJAsh: It would be less magic (and more explicit) if this just mapped `"basher.KeepItInTheGround" => "basher.KeepItInTheGround.css"`.

common/app/views/fragments/basher.scala.html
```diff
@@ -0,0 +1,19 @@
+@(id: String)(implicit request: RequestHeader)
+
+@import views.support.Basher
+
+@if(Basher.fromId(id) != None) {
```
OliverJAsh: If `fromId` returned an `Option` you could do this instead of the `if`:

``` scala
@Basher.fromId(id).map {
  <!-- html -->
}
```

OliverJAsh:
I much prefer this to injecting it with JS. Progressive enhancement! Nice work. We want to do the same for snaps, but that's a bit harder because it has to go through facia press.

common/app/views/fragments/bashers/KeepItInTheGround.scala.html
```diff
@@ -0,0 +1,22 @@
+@()(implicit request: RequestHeader)
+
+<div class="gs-container">
+    <div class="keep-it-in-the-ground">
+        <a class="keep-it-in-the-ground__anchor" href="http://www.theguardian.com/environment/2015/oct/05/a-story-of-hope-the-guardian-launches-phase-two-of-its-climate-change-campaign"></a>
+        <div class="keep-it-in-the-ground__image"></div>
+        <div class="keep-it-in-the-ground__content">
+            <h2 class="keep-it-in-the-ground__headline">Join the Guardian’s campaign and help spread a message of hope that the world can stop climate change<br /> <strong>Join us
+                <div class="keep-it-in-the-ground__button">
```
OliverJAsh: Is this a button? (I presume so from the class.) If so, see `<button>` which has lovely semantics :wink:

OliverJAsh:
Can we make all thrashers into bashers to avoid the page jumping they have?

common/app/views/support/Basher.scala
desbo: agreed, might be better to use a map to make the `Option`

```
val bashersById = Map("environment/series/keep-it-in-the-ground" -> KeepItInTheGround)
def fromId(id: String) = bashersById.get(id)
```

OliverJAsh: :heart_eyes:

common/app/views/fragments/basher.scala.html
```diff
@@ -0,0 +1,19 @@
...
+    <div class="facia-basher" data-link-name="basher">
+        @conf.Static.css.inline("KeepItInTheGround").map { css =>
```
OliverJAsh: You probably don't want to hardcode this CSS for all bashers. Instead you probably want to map the id parameter to CSS?

static/src/stylesheets/inline/basher.KeepItInTheGround.scss
```diff
@@ -0,0 +1,205 @@
...
+    @include mq(desktop) {
+        width: gs-span(4) + $gs-gutter;
+        display: table-cell;
+        vertical-align: top;
+        padding-left: 0;
+    }
+}
+
+.keep-it-in-the-ground__image {
+    background-image: url('http://interactive.guim.co.uk/thrashers/keep-it-in-the-ground-header/hashed/bg.418e6adc.jpg');
```
johnduffell: would https work here?

johnduffell:
@OliverJAsh that would be great, although having said that I don't know if deploying frontend every time we want to change it is ideal if we're using them a lot.  perhaps there's a middle ground worth thinking about where they go into an s3 bucket and we just fetch that bucket on a schedule and stitch it together on the fly?

static/src/stylesheets/inline/basher.KeepItInTheGround.scss
sammorrisdesign: Not really because of the way layout is.

I could move `.keep-it-in-the-ground__anchor` down a level but it wouldn't cover the entire header.

I can't move the button up a level because it has to follow the text.

On the plus side, it is no js.

sammorrisdesign:
@OliverJAsh @johnduffell I'd agree with that. Bashers need to be fairly static and shouldn't too much updating stuff (if they do then js would have be to used, but restricted for obvious reasons). We'd also need a different solution as thrashers have critical content in them, meaning whatever implementation would need to work on apps too.

common/app/views/fragments/basher.scala.html
```diff
@@ -0,0 +1,23 @@
+@(id: String)(implicit request: RequestHeader)
+
+@import views.support.Basher
+
+@Basher.fromId(id).map { id =>
+    <div class="facia-basher" data-link-name="basher">
+        @conf.Static.css.inline(id.toString).map { css =>
+            @if(play.Play.isDev()) {
+                <link rel="stylesheet" type="text/css" href="@Static("stylesheets/inline/basher." + id.toString + ".css")" />
```
OliverJAsh: `href="@Static(s"stylesheets/inline/${css}")"`?

common/app/views/fragments/basher.scala.html
```diff
@@ -0,0 +1,23 @@
+@(id: String)(implicit request: RequestHeader)
+
+@import views.support.Basher
+
+@Basher.fromId(id).map { id =>
```
OliverJAsh: This overrides the `id` identifier which hurts readability. `fromId` returns a `basherType`, so maybe name it that?

common/app/views/fragments/basher.scala.html
sammorrisdesign: because `css` in this instance is the actual css, not a path.

I've moved the map to just the inlining for better readability.

OliverJAsh:
:+1:

OliverJAsh:
… although you have build errors. Try merging in master.

PR #11406, opened 2015-12-22
Fix IE 9 ReferenceError

OliverJAsh:
IE 9 users have been complaining they see an unstyled page:

image

This is happening because we reference `console`, which throws a `ReferenceError` in IE 9 (when the dev tools are not open). As this happens in the `head`, it breaks parsing (AFAIK).

For this reason it's always best to be explicit when referencing properties on the `window` object.

This was broken by #11371.

/cc @sndrs @gklopper @jennysivapalan

### DIFF
--- a/common/app/templates/inlineJS/blocking/shouldEnhance.scala.js
+++ b/common/app/templates/inlineJS/blocking/shouldEnhance.scala.js
@@ -97,6 +97,7 @@
     window.shouldEnhance = mustNotEnhance() ? false : mustEnhance() ? true : couldEnhance() && weWantToEnhance();
 
     // just so we can tell…
+    var console = window.console;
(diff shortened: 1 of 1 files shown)


jennysivapalan:
:+1: :christmas_tree:

gklopper:
:+1:

sndrs:
wow did this fix it? how does that work? 

wouldn't the scope fallback to `window` by default? i'd have expected anything in that closure to be able to refer to anything on `window`, plus the first `console &&` to handle it erroring?

OliverJAsh:
> wouldn't the scope fallback to window by default? i'd have expected anything in that closure to be able to refer to anything on window, plus the first console && to handle it erroring?

It would fallback to `window` for properties which exist, but `window.console` didn't exist hence the `ReferenceError`.

Try typing `abc` into your console. `window.abc` doesn't exist, so you get a `ReferenceError`.

PR #11407, opened 2015-12-22
Fix profile email signup footer

janua:
On `profile.theguardian.com` the email signup footer is completely broken because of relative paths.

Our options are:
- Remove it from the domain
- Hack the paths so that when it is the profile domain we use a full absolute URL
- Move `EmailSignup` out of `applications` and into `common` so that all applications can benefit from it.

I have opted for the last one as this is meant to be site wide.

@rich-nguyen @crifmulholland


rich-nguyen:
:+1:

PR #11409, opened 2015-12-22
Crossword search by

NathanielBennett:
Fixes crossword lookup by ID. Previously, in R2 land, this functionality took an crossword id and a crossword type ( see here: http://www.guprod.gnl/crosswords ), whereas we just had an ID. How did it ever work? This adds the type parameter back in. This means that looking up a crossword takes the same parameters as a crossword page, so re-factored the crossword controller to reflect ths

crossword-search


applications/app/controllers/CrosswordsController.scala
```diff
@@ -107,19 +119,7 @@ object CrosswordSearchController extends Controller with ExecutionContexts {
 
   def lookup() = Action.async { implicit request =>
     lookupForm.bindFromRequest.get match {
```
janua: I know this isn't in your change, but I am surprised to see this...
The `case _` will never happen bellow because `get` throws an exception from `Form[T]`.
I think we should probably be using `Form.fold` here, [see example](https://github.com/guardian/frontend/blob/master/applications/app/controllers/EmailController.scala#L124)

applications/app/controllers/CrosswordsController.scala
```diff
@@ -14,18 +14,20 @@ import services.{IndexPageItem, IndexPage}
...
   protected def withCrossword(crosswordType: String, id: Int)(f: (Crossword, ApiContent) => Result)(implicit request: RequestHeader): Future[Result] = {
     LiveContentApi.getResponse(LiveContentApi.item(s"crosswords/$crosswordType/$id", Edition(request)).showFields("all")).map { response =>
        val maybeCrossword = for {
         content <- response.content
         crossword <- content.crossword }
        yield f(crossword, content)
-       maybeCrossword getOrElse InternalServerError("Crossword response from Content API invalid.")
-    } recover { case _ => InternalServerError("Content API query returned an error.") }
+       maybeCrossword getOrElse noResults
+    } recover { case _ => noResults }
```
janua: Are you sure you don't want to log out the `Throwable`?

janua:
:+1:

applications/app/controllers/CrosswordsController.scala
janua: [It is now here](https://github.com/guardian/frontend/blob/master/common/app/controllers/EmailSignupController.scala#L124)

PR #11410, opened 2015-12-22
Bottom responsive banner on tablets

regiskuckaertz:
So, the responsive takeover at the bottom of the page is not loaded on tablets. On tablets:

img_0064

On a desktop:

screen shot 2015-12-22 at 11 51 01

It turns out the ad loads, but only if we're at the bottom of the page and hit reload then. So, it must be something with how ads are lazy loaded. Lazy loading happens when a scroll event occurs, and in the codebase scrolls are mediated through another event which wraps event handlers in `fastdom.read`:

```
mediator.on('window:throttledScroll', lazyLoad);
```

`lazyLoad` is never called. Replacing the line above with:

```
window.addEventListener('scroll', lazyLoad);
```

... and the whole thing works. For good measure, `lazyLoad` is debounced at 200ms which is more than enough for us to test whether there's a slot that needs to be loaded in the area, and the bulk of the work is offloaded to the next animation frame.

So if the _traditional approach_ is working, there must be something with the way `throttledScroll` works, and the only difference between the two is Fastdom. While the bug in this case is solved, there may be many more lurking in the codebase, and a deeper inspection of how Fastdom works appears to become an urgent issue.

### DIFF
--- a/static/src/javascripts/bootstraps/enhanced/common.js
+++ b/static/src/javascripts/bootstraps/enhanced/common.js
@@ -231,29 +231,6 @@ define([
                 });
             },
 
-            // Adds a global window:throttledScroll event to mediator, which throttles
-            // scroll events until there's a spare animationFrame.
-            // Callbacks of all listeners to window:throttledScroll are run in a
-            // fastdom.read, meaning they can all perform DOM reads for free
-            // (after the first one that needs layout triggers it).
-            // However, this means it's VITAL that all writes in callbacks are delegated to fastdom
-            throttledScrollEvent: function () {
-                var running = false;
-                var emit = function () {
-                    if (!running) {
-                        running = true;
-                        fastdom.read(function () {
-                            mediator.emitEvent('window:throttledScroll');
-                            running = false;
-                        });
-                    }
-                };
-                var callback = userPrefs.get('use-idle-callback') && 'requestIdleCallback' in window ? function () {
-                    window.requestIdleCallback(emit);
-                } : emit;
-                window.addEventListener('scroll', callback);
-            },
-
             checkIframe: function () {
                 if (window.self !== window.top) {
                     $('html').addClass('iframed');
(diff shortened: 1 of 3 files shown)


OliverJAsh:
Interesting. The other difference between the native scroll event and our custom event is that the latter comes through our event emitter. I can't think what the problem might be of the top of my head.

Were you able to look into the problem at all? It would be good to properly diagnose the problem before working around it.

regiskuckaertz:
Hi @OliverJAsh I am looking at it now but this will take time. There is certainly more to it* but I had to push this fix as some ads are not loading on PROD and some people have noticed.

* For one, the behaviour is reproducible on all facia pages, but not on articles where everything works fine, so there must be some side-effect somewhere.

regiskuckaertz:
(to make things clear I changed the title, everything pointed at throttledScoll until I noticed everything was fine on article pages)

regiskuckaertz:
@OliverJAsh I'll move the throttleScroll event initialization to standard and see if it affects perfs too much.  Will keep you posted.

regiskuckaertz:
@OliverJAsh so that's it. I tested on iPad mini and couldn't see any lag/jank at scroll that wasn't there before.

kenlim:
:+1:

static/src/javascripts/bootstraps/enhanced/common.js
```diff
@@ -392,7 +369,6 @@ define([
                 ['c-test-cookie', modules.testCookie],
                 ['c-ad-cookie', modules.adTestCookie],
                 ['c-event-listeners', modules.windowEventListeners],
-                ['c-throttled-scroll-event', modules.throttledScrollEvent],
```
OliverJAsh: We need to add this error handling to standard, but not for your PR :smile:

OliverJAsh:
:+1:

static/src/javascripts/bootstraps/enhanced/common.js
regiskuckaertz: Thanks man!

PR #11411, opened 2015-12-22
Protect against reference errors

OliverJAsh:
This is to prevent errors such as https://github.com/guardian/frontend/pull/11406 from happening again. I've seen this happen multiple times.


rich-nguyen:
:+1:

PR #11413, opened 2015-12-22
Add more email list IDs

janua:
There are more changes coming to email. We now need forms for `listIds` that don't have a triggered email in ExactTarget.

This has the following changes:
- Adds all the extra list ids
- If we can't handle a list id, we now return a `404`. (Whoever it putting out a form should notice fast that this form isn't going to work)
- We no longer need a `triggerId` for the `POST`, if we don't have one we now just `POST` without the `triggerId`.

This can not be released until the work for receiving `listIds` without `triggerIds` happens in the [email-signup](https://github.com/guardian/email-signup).

@crifmulholland @desbo


jennysivapalan:
:+1: :email: :santa:

PR #11416, opened 2015-12-23
Upload omniture script for R2 pressed pages

jennysivapalan:
This PR will allows us to control the R2 omniture tracking outside of a R2 pressed page. We need to upload this file to static S3 directory so that we can refer to it in the HTML. We don't want to combine this with hashed folders as the pressed pages will not dynamically update.

I'm not sure what is the best way to test this other than putting my branch to CODE?


OliverJAsh:
:+1: Looks fine to me, but not sure how to test.

gklopper:
:+1: you should be able to see everything in the right place on CODE.

Just block PROD deploys till then.

PR #11417, opened 2015-12-23
Fixing bug with local admin

kenlim:
The /admin endpoint on dev machines running dev-build is broken because this route is missing.


uplne:
:+1:

PR #11421, opened 2015-12-23
Attempt 2 of omniture script for pressed pages.

jennysivapalan:
Second attempt of https://github.com/guardian/frontend/pull/11416. Turns out there are two `dist-publish-assets-tc` files so I've updated the second one. 

Will chat to @rich-nguyen in the new year to find out why there are two...


OliverJAsh:
:+1:

PR #11424, opened 2015-12-24
Remove Twitter Share Count from ToolTip

scottp719:
Removing the twitter share count from the tooltip on article pages as API no longer available

@OliverJAsh


OliverJAsh:
:+1:

PR #11423, opened 2015-12-23
Omniture for R2 page pressing

jennysivapalan:
Follows https://github.com/guardian/frontend/pull/11396
- Ensures Identity references, `<scripts>` and `<noscripts>` are removed.
- Adds a reference to a call an external script see: https://github.com/guardian/frontend/pull/11421 to add page view for Omniture.
- As suggested by @rich-nguyen the original html should be pushed to a new bucket in case we need to re-do this in the future.
- Adds a switch to push the HTML file to a bucket.

Note the changes in https://github.com/guardian/frontend/pull/11421 is not quite available to use but this requires a Fastly change that I'll do in the new year. :fireworks: 

cc/ @JustinPinner

PR #11429, opened 2015-12-30
R2 polls HTML cleaner

jennysivapalan:
Polls on R2 pages currently load the page without the results and then make a call to the Polls API and update the HTML with the results based on an [R2 javascript file](https://github.com/guardian/r2-frontend/blob/10f041ce16b6be377a88f1b7eef03ee66f02416d/r2frontend-static/src/main/resources/common/scripts/guardian.r2.PollResults.js). All these polls are closed so results cannot be changed. So for pressing these pages I've created a PollsHtmlCleaner which will press the pages with the poll results. 

Note this PR is based off `r2-omniture-pressing` branch (https://github.com/guardian/frontend/pull/11423) as it requires code from here too.

cc @JustinPinner 

:memo:


JustinPinner:
Nice :+1:

jennysivapalan:
Closing this as I merged https://github.com/guardian/frontend/pull/11423

PR #11430, opened 2015-12-30
Commercial content relabeling

regiskuckaertz:
Enables the new designs for Paid for content behind a switch for:

Fronts:
screen shot 2015-12-30 at 13 39 25

Articles:
screen shot 2015-12-30 at 13 38 31

Videos & Audios:
screen shot 2015-12-30 at 13 39 58

Galleries:
screen shot 2015-12-30 at 13 39 49

Interactives:
screen shot 2015-12-30 at 13 39 00

Also a new style for containers of commercial content:
screen shot 2015-12-30 at 13 43 25

... and single cards:
screen shot 2015-12-30 at 13 44 34


common/app/views/fragments/contentMeta.scala.html
```diff
@@ -56,7 +56,7 @@
     @if(item.tags.contributors.length == 1 && item.tags.contributors.headOption.exists(_.properties.twitterHandle.nonEmpty)) { content__meta-container--twitter}
     @if(item.tags.contributors.length == 1 && item.tags.contributors.headOption.exists(_.properties.emailAddress.nonEmpty)) { content__meta-container--email}">
 
-    @if(showBadge) {
+    @if(showBadge && (NewCommercialContent.isSwitchedOff || (NewCommercialContent.isSwitchedOn && (item.tags.isVideo || item.tags.isAudio || (item.tags.isArticle && !item.commercial.isAdvertisementFeature))))) {
```
kelvin-chappell: Logically, condition is same as `showBadge && (NewCommercialContent.isSwitchedOff || (item.tags.isVideo || item.tags.isAudio || (item.tags.isArticle && !item.commercial.isAdvertisementFeature)))`

But it's not clear what this condition means.  If the switch is off, are we only going to show logos on video pages, audio pages and non-paid content?

static/src/javascripts/projects/common/modules/ui/sticky.js
```diff
@@ -1,64 +1,83 @@
...
     bonzo,
-    $,
     config,
     mediator,
     fastdom,
-    defaults,
-    bindAll) {
+    defaults) {
+
+    var header = config.switches.viewability
```
kelvin-chappell: Could this condition be broken down a bit?  It looks very complicated.

common/app/views/fragments/contentMeta.scala.html
regiskuckaertz: Good catch! If the switch is off, then we fallback to the original test (ie `showBadge`). If the switch is on, the badge should only be shown on video/audio pages as well as on articles which are **not** paid for ones (in the latter case, the badge will appear somewhere else)

static/src/javascripts/projects/common/modules/ui/sticky.js
regiskuckaertz: I would've hoped so, but this is copied straight-out from enhanced/common.js. I'll see if this could be done earlier during the bootstrap process.

Calanthe:
:+1:

static/src/stylesheets/module/content/_content.scss
```diff
@@ -407,7 +407,7 @@
     }
 
     @include mq(leftCol) {
-        min-height: gs-height(1) + $gs-baseline;
```
Jholder112233: Hey, just noticed that the dateline min-height had gone awry and tracked it down to this line, do you mind fixing?

Calanthe: We will look at this with @kenlim

PR #9863, opened 2015-07-20
Upgrade to Lodash v3.x.x

OliverJAsh:
All the tests pass and I've done some light QA'ing.

I used https://github.com/lodash/lodash/wiki/Changelog#compatibility-warnings to migrate.

PR #11431, opened 2015-12-30
Pre-compile lodash templates

regiskuckaertz:
Lodash templates can be compiled ahead of time with or without seed data. This is particularly useful if the templates are called several times as it usually yields better performance. See http://jsperf.com/pre-compile-lodash-templates

A next step would be to use a lodash plugin loader, either home grown or [not](https://github.com/tbranyen/lodash-template-loader) (_updated link_), to do that automatically.


static/src/javascripts/projects/common/modules/article/rich-links.js
```diff
@@ -81,8 +81,11 @@ define([
             !isSensitive &&
             !isDuplicate
         ) {
+            if (typeof richLinkTagTmpl === 'string') {
+                richLinkTagTmpl = template(richLinkTagTmpl);
+            }
```
OliverJAsh: Variable re-assignments hurt code readability. Here we're not sure if `richLinkTagTmpl` is a string or something else. I think creating a new variable here (e.g. `richLinkTagTmplString` and `richLinkTagTmpl`) would be better, to make it explicit that there are two types. (On a side note, I wonder if there's an ESLint rule for this?)

Ideally I think we should pre-compile templates as an optimisation at the module loader level, like you said. I would worry that, if we added a half-way solution such as this, we would never get around to doing The Right Thing™. How hard would it be to do that?

regiskuckaertz: @OliverJAsh oh sorry, I tend to see vars as empty vessels and just put anything inside, but you're right it's not the clearest way of writing code. You're right, unless templates were somehow generated at runtime, there's no reason not to pre-compile that during r.js optimisation—and even then, this kind of work is worth offloading to a web worker. I'll give the lodash loader a try, otherwise a slightly modified version of the text plugin should work too.

common/app/templates/inlineJS/blocking/curlConfig.scala.js
```diff
@@ -43,7 +43,8 @@ window.curlConfig = {
 
             // plugins
             text:         'components/requirejs-text/text',
-            inlineSvg:    'projects/common/utils/inlineSvg'
+            inlineSvg:    'projects/common/utils/inlineSvg',
+            ldsh:         'projects/common/utils/lodash-loader'
```
OliverJAsh: Maybe this plugin should be called `template!` instead? Lodash is just the implementation of it.

static/src/javascripts/bootstraps/video-embed.js
```diff
@@ -16,7 +15,7 @@ define([
     'common/modules/video/fullscreener',
     'common/views/svgs',
     'text!common/views/ui/loading.html',
```
OliverJAsh: Could we also update this one to use the loader? I wonder if you've missed any others, might be worth doing a project wide search for `text!foo.html`.

static/src/javascripts/projects/common/modules/commercial/creatives/fluid250.js
```diff
@@ -4,8 +4,7 @@ define([
     'common/utils/$',
     'common/utils/mediator',
     'common/utils/storage',
-    'common/utils/template',
```
OliverJAsh: Can we delete this module, if you haven't already?

OliverJAsh:
Great work on the loader plugin. The diff size is quite large, mainly because it looks like we've gained a copy of lodash in `components` somehow, along we some changes to other dependencies? Can we remove those changes if they're not needed?

common/app/templates/inlineJS/blocking/curlConfig.scala.js
regiskuckaertz: yeah I thought about it indeed. will do.

static/src/javascripts/bootstraps/video-embed.js
regiskuckaertz: I didn't because it's just a string (there are a few other examples), but if it makes things clearer then I'm doing it.

static/src/javascripts/projects/common/modules/commercial/creatives/fluid250.js
regiskuckaertz: I wish we could, but `template` is used in `sport.js` where the string is dynamically assigned

regiskuckaertz:
@OliverJAsh yes, can I just remove `components/lodash`?

static/src/javascripts/projects/common/modules/commercial/creatives/fluid250.js
OliverJAsh: Ahh, I see. If that's the only usage could we remove `common/utils/template` and just require `lodash/utilities/template` directly?

static/src/javascripts/bootstraps/video-embed.js
OliverJAsh: Ahh I see, leave it then. :smile:

OliverJAsh:
@regiskuckaertz Yeah, although I'm not sure how it got added as it's not referenced in `bower.json`. I think it should be fine just to remove the changes to all the other components if they're not needed.

regiskuckaertz:
@OliverJAsh it has to be because I tested another loader before, which probably had lodash as a dependency. The fix to support.js is necessary, otherwise `make compile` will fail. I think the repository is not maintained anymore, that means if we do things the proper way then we need to fork `lodash-amd` and update it from there. What do you think?

static/src/javascripts/projects/common/modules/commercial/creatives/fluid250.js
regiskuckaertz: That was so obvious it didn't cross my mind. Thanks for the tip!

OliverJAsh:
> The fix to support.js is necessary, otherwise make compile will fail. I think the repository is not maintained anymore, that means if we do things the proper way then we need to fork lodash-amd and update it from there. What do you think?

Oh that's interesting. How come it didn't fail the compile beforehand? Is it because we never used to run the module in Node land and now we are because we're compiling templates?

I had a go at updating to Lodash v3 awhile ago but gave up because it was impossible to reason about my changes (no static type system!). https://github.com/guardian/frontend/pull/9863. In the interim I guess we will have to fork the repository, do the fix, and then reference that from `bower.json`. Do you agree?

regiskuckaertz:
1- yes, that is why. There is one reference to window which makes Node complain. Interestingly, this bug does not appear in lodash v2.4.1, only in the AMD version.

2- yes, I think it puts us on the safe side. On it.

OliverJAsh:
Looks good, nice :+1:

PR #11432, opened 2015-12-31
Revert "Remove BrandedComponent"

regiskuckaertz:
There seems to be remaining Branded Component ads on PROD right now.


uplne:
So if it passed tests and you've checked it locally then :+1:

PR #11433, opened 2015-12-31
Prevent gallery lightbox controls from causing FOUC

OliverJAsh:
Similar to https://github.com/guardian/frontend/pull/10924, but this time for gallery lightbox controls, which get injected on any compatible content page.

This is a race condition: if the CSS loads before the JS executes, you don't get a flash, otherwise you do.

This solution will force the JS to wait until the CSS has loaded before injecting the DOM element.
# Before

# After

There is no flash.

### DIFF
--- a/static/src/javascripts/projects/common/modules/gallery/lightbox.js
+++ b/static/src/javascripts/projects/common/modules/gallery/lightbox.js
@@ -23,7 +23,8 @@ define([
     'lodash/collections/map',
     'lodash/functions/throttle',
     'lodash/collections/forEach',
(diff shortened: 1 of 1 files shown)


OliverJAsh:
Nice diff: https://github.com/guardian/frontend/pull/11433/files?w=1

kaelig:
Thanks! Found that one issue particularly annoying.

paperboyo:
Thanks!

sndrs:
:+1: great, been bugging me for yonks

PR #11434, opened 2015-12-31
Crosswords meta restyle

NathanielBennett:
Makes some slight style changes to crossword pages: Mainly making the setter more prominent. But also a left border for the comment count (as per elsewhere) and some other stuff as per Franks' designs. I love doing css. Thanks @superfrank 

Desktop;
dt

Mob
mob


static/src/stylesheets/module/crosswords/_layout.scss
```diff
@@ -158,3 +158,24 @@
...
+
+.crossword__setter {
+    @include mq($from: leftCol) {
+        display: none;
+    }
+}
+
+.crossword__meta-setter {
+    @include mq(leftCol) {
+        margin-top: ($gs-baseline * 6) + 2;
```
desbo: `2` what?

PR #11436, opened 2016-01-04
dedicated crossword print page

desbo:
Instead of the crosswords print CSS we use at the moment, this allows you to add `/print` to the end of any crossword URL and get a separate page like this:

[Cryptic crossword No 26,666 _ Crosswords _ The Guardian.pdf](https://github.com/guardian/frontend/files/75610/Cryptic.crossword.No.26.666._.Crosswords._.The.Guardian.pdf)

This is much closer to the R2 version and should hopefully solve most of the printing problems we're hearing about.


applications/app/views/printableCrossword.scala.html
```diff
@@ -0,0 +1,48 @@
...
+<!DOCTYPE html>
+<html>
+    <head>
+        @fragments.head(crosswordPage, None, Html(""))
+    </head>
+
+    <body class="printable-crossword">
+        <h1 class="printable-crossword__title">
+            @(crosswordPage.crossword.creator match {
+                case Some(creator) => s"${crosswordPage.crossword.name} set by ${creator.name}"
```
NathanielBennett: very minor niggle, the other pages have a colon after "set by"

NathanielBennett:
:+1: 
LGTM

applications/app/views/printableCrossword.scala.html
desbo: ah, I was pretty much rebuilding the R2 print page, which doesn't have that - since this is part of the title of the crossword rather than metadata I think it reads better without it.

gtrufitt:
:+1:

PR #11437, opened 2016-01-04
Increase cache time on form endpoint

rich-nguyen:
There are lots of router/applications alarms relating to request spikes. The fastly logs are pointing to urls for the email form, so increasing the cache time to see if this helps.


gtrufitt:
:+1:

PR #11438, opened 2016-01-04
show clear this and anagram helper buttons on everyman crosswords

NathanielBennett:
This fixes a problem raised by Hugh - the 'Clear this' and 'Anagram helper' buttons sometimes do not appear on Everyman puzzles. This is because Everyman puzzles do not have solutions included until sometime after publication and the code checks for focus AND a solution before showing these controls. So, for these two buttons, just check for focus:

Before:
everyman-before

After: 
everyman-after


NataliaLKB:
:+1:

PR #11440, opened 2016-01-04
Only intercept fetch events in SW if they're to secure endpoints

sndrs:
it doesn't let you fetch them anyway, should stop errors like these:

screen shot 2016-01-04 at 11 43 25
screen shot 2016-01-04 at 11 43 54

### DIFF
--- a/applications/app/templates/serviceWorker.scala.js
+++ b/applications/app/templates/serviceWorker.scala.js
@@ -92,41 +92,43 @@ self.addEventListener('install', function (event) {
 
 this.addEventListener('fetch', function (event) {
     var request = event.request;
-
-    if (doesRequestAcceptHtml(request)) {
-        isCacheUpdated().then(function (isUpdated) {
-            if (!isUpdated) {
-                updateCache().then(deleteOldCaches);
-            }
-        });
-    }
-
     var url = new URL(request.url);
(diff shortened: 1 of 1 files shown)


OliverJAsh:
Sounds sensible and I think it's good for us to start clearing up errors when running in HTTPS mode, although I think the bigger problem we have here is mixed content.

Seeing as this change effects the offline page, you should be able to test if it still works. To do this locally:

* Run `make compile` (you need compiled assets for the offline page as it can't cache the small modules)
* Go to http://localhost:9000/info/developer-blog to install the SW. You might already have a SW installed, in which case either ensure the new one takes over (see Dev Tools) or just do the test in an incognito window if it's easier.
* Go offline (Dev Tools network throttling)
* Refresh and the offline page should appear

johnduffell:
note it's easier to view this one with [?w=0](https://github.com/guardian/frontend/pull/11440/files?w=0)

I can't see a reason not to ship this one? so :+1: .  although in related news now capi stores a secureFile attribute for video files so if we switch to that it should fix the first one anyway

OliverJAsh:
Ship it or close :ship:

sndrs:
whoopsie

OliverJAsh:
Was this a mistake? If we're merging I would like to know the offline page has been tested.

PR #11444, opened 2016-01-04
Adding serverside switch for turning off amp

NataliaLKB:
We want the ability to stop amp pages being found on our site if the need arises. The quickest way to do that is to not show the link to the `amphtml` in the metadata

cc @stephanfowler


stephanfowler:
The switch should probably be in FeatureSwitches.scala, as it's not a serverside AB test

stephanfowler:
:+1: otherwise

NataliaLKB:
@stephanfowler moved it to FeatureSwitches. Will deploy first tomorrow.

Thanks!

PR #11447, opened 2016-01-04
update pasteup-forms

gidsg:
Update `pasteup-forms` to version 6. @guardian/identity @guardian/membership-and-subscriptions  please check you are OK with this (as people who use forms).

This is to support some forthcoming forms within the Discussion frontend app.

### DIFF
--- a/static/src/stylesheets/bower.json
+++ b/static/src/stylesheets/bower.json
@@ -1,7 +1,7 @@
 {
   "name": "frontend-common-css",
   "dependencies": {
-    "pasteup-forms": "~5.0.0",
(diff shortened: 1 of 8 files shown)


rtyley:
This would probably only effect Membership on https://profile.theguardian.com/membership/edit, where we have a dotcom-styled page using data retrived by AJAX from membership - should be fine with us. 

cc @joelochlann @tomverran

emma-p:
Should be fine with Identity too

gidsg:
CC @sndrs please can you review and +1?

sndrs:
:+1:

PR #11449, opened 2016-01-05
Extend fluid ads switch

uplne:
This is blocked for now. Waiting for Google. I extended it for a month so we can check and ask for updates when the right time comes.


regiskuckaertz:
:+1:

PR #11450, opened 2016-01-05
Revert "Only show the digitalpack tab if you go there directly"

tomverran:
As a result the digital pack MMA page will be generally available

This reverts commit 27947fb7f88a75d1666ef405a2e6e3593bb89d41.


ostapneko:
:+1:

PR #11452, opened 2016-01-05
Commercial component manual multiply has an optional link texts

uplne:
After:
screen shot 2016-01-05 at 11 45 47

Component has main link text which can be overridden by specific link text.

Co-author @regiskuckaertz


kenlim:
:+1:

PR #11454, opened 2016-01-05
Fix crossword print URL

desbo:
Move it back to the end of the path to make it consistent with R2.
There's also an incoming fastly change to allow these URLs in nextgen.


gtrufitt:
:+1:

PR #11458, opened 2016-01-05
Support BrowserSync in nginx dev config

desbo:
This means you can use m.thegulocal.com with BrowserSync, and if that isn't running it'll fall back to the usual SBT port.


sndrs:
nice!

sndrs:
:+1: lgtm

PR #11462, opened 2016-01-06
Enable the test that checks switches have a valid date

rich-nguyen:
Putting this back in the test suite because a lot of switches are going past their expiry date.


OliverJAsh:
:+1: 

Was it disabled for a reason?

rich-nguyen:
I think Grant had some problems with the test, so disabled it.

PR #11463, opened 2016-01-06
Restore crossword print styles

gtrufitt:
Putting back the page printing styles (if these don't work user help can point people to the dedicated print page).

@desbo @ScottPainterGNM


desbo:
:+1:

PR #11466, opened 2016-01-06
Content API upgrade (7.12)

OliverJAsh:
This is needed so we can consume content atoms so we can do quizzes.

/cc @rich-nguyen


piuccio:
:heart_eyes:

jennysivapalan:
A big diff but the changes look relatively sane to me :+1: :up: :a:

johnduffell:
the proof of the coding is in the testing I suspect :+1:

piuccio:
just to scare you with random unrelated facts. Last time I tried to upgrade CAPI in fronts tool, live blog started appearing as dead. Once you start testing this it might be a good idea to keep an eye on live blogs as well, just in case...

johnduffell:
@piuccio if this PR doesn't break live blogs, mine will :fearful:

PR #11467, opened 2016-01-06
Crosswords header meta restyle

NathanielBennett:
More restyling to the crossword header styles : Now looks like this: 

Mob:
cw-redesign

Dt:
cw-redesign-deskto


static/src/stylesheets/module/content/_content.scss
```diff
@@ -555,6 +555,25 @@
         min-height: 0;
         padding-bottom: 0;
     }
+
+    .content__labels--crossword & {
+        @include mq(leftCol) {
+            margin-top: ($gs-baseline * 6) + 7;
```
desbo: missing unit on the 7

gtrufitt: Might be worth adding a comment here to explain what the adjustment is

static/src/stylesheets/module/content/_content.global.scss
```diff
@@ -120,11 +120,9 @@
     }
 
     .meta__extras--crossword & {
```
gtrufitt: Does it seem odd that the crossword specific styles are not in the crossword stylesheets? I am not sure what the consistent approach is, modifications within the component or modifications in the feature stylesheets?

NathanielBennett: @gtrufitt  Point. I'll try and shift this over ...

gtrufitt:
:+1: other than couple of comments :)

PR #11468, opened 2016-01-06
First attempt at commercial containers

kelvin-chappell:
This fills in the design implementation built by @regiskuckaertz and @Calanthe.

Checking it in now so that work on branding can be done separately from rendering work.

Still to do in further PRs:
- More clientside work to make it presentable at all breakpoints
- Support other content types
- Add tracking data
- Add logos, sponsors and links

image


common/app/views/fragments/containers/facia_cards/commercialContainer.scala.html
```diff
@@ -1,14 +1,151 @@
...
+                                                        </li>
+                                                    }
+                                                    case _ => {}
+                                                }
+                                            }
+                                        }
+                                    }
+                                }
+                            </ul>
+                            <aside class="commercial__meta">Paid for by <img src="ben-jerrys.png"></aside>
```
Calanthe: Should it be hardcoded?

kelvin-chappell: For now.  Going to bring logos in in a separate PR.

Calanthe:
:+1: although I need to fix the styling (eg padding of the text and the header)

PR #11472, opened 2016-01-06
Loading public assets in dev mode

rich-nguyen:
When running play in dev mode, and building assets using `make compile`, public files such as react.js could not be located. 

I've moved as much dev-based asset loading code into `DevAssetsController.scala` as possible.


rich-nguyen:
we are unable to read files from `static/hash`, this directory doesn't exist in Prod. We rely on the java classpath to load the asset map as a resource instead. In dev, we avoid continuously copying the asset map to `conf` (the classpath folder), and load the asset map directly from `hash`.

OliverJAsh:
:+1: Nice tidy up

PR #11474, opened 2016-01-06
extend quiz score switch

johnduffell:
People are still getting value out of the temporary quiz score service.  Eventually we will write quizzes properly in frontend (apparently you know more @rich-nguyen ) at which point we'll make a proper solution for quiz scores


rich-nguyen:
:+1:

PR #11475, opened 2016-01-06
Don't call s.tl for untagged page clicks

rich-nguyen:
The omniture analytics module was firing track links which didn't have a tag. These were interpreted as page views. Now, same-page clicks that do not have a `data-link-name` won't be tracked by omniture at all.

### DIFF
--- a/static/src/javascripts/projects/common/modules/analytics/omniture.js
+++ b/static/src/javascripts/projects/common/modules/analytics/omniture.js
@@ -88,7 +88,7 @@ define([
             storage.session.set(NG_STORAGE_KEY, storeObj);
         } else {
             // Do not perform a same-page track link when there isn't a tag.
-            if (!spec.tag) {
+            if (spec.tag) {
                 // this is confusing: if s.tl() first param is "true" then it *doesn't* delay.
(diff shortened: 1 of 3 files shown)


static/src/javascripts/projects/common/modules/analytics/omniture.js
```diff
@@ -87,9 +87,12 @@ define([
             try { sessionStorage.setItem(R2_STORAGE_KEY, storeObj.tag); } catch (e) {/**/}
             storage.session.set(NG_STORAGE_KEY, storeObj);
         } else {
-            // this is confusing: if s.tl() first param is "true" then it *doesn't* delay.
-            delay = spec.samePage ? true : spec.target;
-            this.trackLink(delay, spec.tag, { customEventProperties: spec.customEventProperties });
+            // Do not perform a same-page track link when there isn't a tag.
+            if (!spec.tag) {
```
OliverJAsh: Shouldn't this condition be the inverse?

rich-nguyen: I just noticed!

OliverJAsh:
:+1:

PR #11477, opened 2016-01-07
add doc to create sns queue for pressing locally

johnduffell:
@rich-nguyen good point, I forgot to add that one


rich-nguyen:
:+1: nice

PR #11478, opened 2016-01-07
Move omniture to standard js

sndrs:
Previously, omniture was inlined in the `<head>`, adding around 30kb (gzipped) to every page. 

This prepends both the library and our pageview code to the `standard` app. 

It runs before curl/the main app to minimise tracking changes, but it does tie tracking to `app.js`.

`videoEmbed` (and pages in dev) still both inline it, because it exists outside the main app module system (`videoEmbed` might want to inline everything, but that's another PR...).

@rich-nguyen and @gklopper, you might want to cast an eye over this?


grunt-configs/concat.js
```diff
@@ -1,10 +1,23 @@
 module.exports = function (grunt, options) {
     return {
+        options: {
+            process: function (src, path) {
+                // make sure omniture handles an error
+                // this image fallback is duplicated in /common/app/views/fragments/omnitureScript.scala.html
+                if (path.match(/vendor\/omniture\.js/)) {
+                    src = 'try {' + src + '} catch (e) {(new Image()).src = window.guardian.config.page.beaconUrl + "/count/omniture-library-error.gif";}';
```
rich-nguyen: This is a little shim-like, i think we assess what we have left as a 'standalone analytics application', and build that as an asset that can be pre-pended to `app.js`

gklopper:
How much later does this load than before and how much more likely is it that it will fail to load?

common/app/views/fragments/inlineJSBlocking.scala.html
```diff
@@ -57,4 +36,23 @@
...
     @InlineJs(cloudwatchBeacons(page.metadata).body, "cloudwatchBeacons.js")
+
+    // finally, load the standard.js app async
+    // do this last so that any config stuff it needs is definitely on `window.`
+    @Html(curlConfig().body)
+
+    @if(play.Play.isDev()) {
+        // in prod, these are concatenated in /grunt-configs/concat.js
+        @Html(omniture().body)
+        @Html(Static.js.curl);
```
rich-nguyen: Needs something like `<script src=@Static('bootstraps/standard/omniture-pageview.js')></script>`

rich-nguyen:
@gklopper I reckon the success rate of the initial page view from omniture will suffer from the time delay and network quality involved with waiting for `app.js`. However it won't suffer from the success rate of initialising our app (ie anything from `boot.js`), and there still won't be a dependency on curl from omniture.

gklopper:
OK, assuming a beacon (or something) to measure the difference then :+1: for the concept.

sndrs:
> How much later does this load than before and how much more likely is it that it will fail to load?

in pseudo code, before it was:
```html
<head>
<script>omniture lib</script>
<script>
gu1, gu2...
omniture pageview
...gu3, gu4
</script>
<script src="standard.js" /> // curl, boot.js etc.
</head>
```

now:
```html
<head>
<script>
gu1, gu2...
...gu3, gu4
</script>
<script src="standard.js" /> // omniture lib, omniture pageview, curl, boot.js etc.
</head>
```

PR #11479, opened 2016-01-07
Fix asset map lookups for any JS not compiled in dev

OliverJAsh:
Broken by https://github.com/guardian/frontend/pull/11472

On master, after `make compile-dev`:

image

This was happening because we used to allow invalid asset map lookups, but now this throws an error (correctly).

This proposes a whitelist for assets which are not compiled in dev but we still want them to be accessible via the asset map. Coincidentally, this also solves a problem we had with the dev workflow for the offline page.

### DIFF
--- a/common/app/dev/DevAssetsController.scala
+++ b/common/app/dev/DevAssetsController.scala
@@ -10,12 +10,12 @@ import play.Play
 
 object DevAssetsController extends Controller with ExecutionContexts {
   // This asset map is used for finding files on local disk.
-  private val localAssets = new common.Assets.Assets("static/hash/")
+  private val localAssets = new common.Assets.Assets(base = "")
 
-  private val findDevAsset: PartialFunction[String, Asset] = {
-    case path if new File(s"static/hash/$path").exists() => Asset(s"static/hash/$path")
-    case path if new File(s"static/src/$path").exists() => Asset(s"static/src/$path")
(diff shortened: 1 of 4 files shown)


rich-nguyen:
ace :+1:

common/app/dev/DevAssetsController.scala
```diff
@@ -25,11 +25,17 @@ object DevAssetsController extends Controller with ExecutionContexts {
...
 
     // In dev mode we must consistently reload the asset map, to account for changes to the map.
-    val asset = localAssets.lookup.assets().getOrElse(path, findDevAsset.lift(path).getOrElse {
-        throw AssetNotFoundException(path)
-      })
+    val resolvedPath = findDevAsset.lift(
+      localAssets.lookup
+        .assets()
+        .get(path)
+        .map(_.path)
```
rich-nguyen: needs a comment like `first try to pass the resolved asset map path to findDevAsset`

common/app/dev/DevAssetsController.scala
```diff
@@ -25,11 +25,17 @@ object DevAssetsController extends Controller with ExecutionContexts {
...
+        .getOrElse(path)
```
rich-nguyen: needs a comment like `if that fails, try to pass the original path to findDevAsset`

PR #11481, opened 2016-01-07
0% AB test removing the sticky nav

holder-josh:
This is an initial 0% AB test that replaces the current sticky.js code with a much simplified version that only sticks the top ad.

The current sticky nav implementation is a blocker on the 'toast' element of the upcoming liveblog work (which has a sticky top button) and is also the cause of all of these Github issues: https://github.com/guardian/frontend/labels/sticky-nav

Putting this out there as not prod-ready code but something we can show to people and get feedback on.

Final implementation could be simpler as some of the injected CSS here could be done in the CSS files. 

May need to add some extra workarounds for the Apple banner/adblock banner. 

@OliverJAsh, you may be interested in this?

nav

### DIFF
--- a/static/src/javascripts/projects/common/modules/navigation/newSticky.js
+++ b/static/src/javascripts/projects/common/modules/navigation/newSticky.js
@@ -1,0 +1,76 @@
+define([
+    'fastdom',
+    'common/utils/$',
+    'common/utils/mediator',
+    'lodash/functions/bindAll'
+], function (
+    fastdom,
+    $,
+    mediator,
+    bindAll) {
+
+    function StickyHeader() {
+
+    }
+
+    StickyHeader.prototype.init = function () {
+        fastdom.read(function () {
+            this.$adBanner = $('.js-top-banner-above-nav');
+            this.$header = $('.js-header');
+            this.headerHeight = this.$header.height();
+            this.adHeight = this.$adBanner.height();
+        }.bind(this));
+
+        fastdom.write(function () {
+            this.$adBanner.css({
+                'width': '100%',
+                'z-index': '1020'
+            });
+
+            this.$header.css({
+                'z-index': '1019'
+            });
+        }.bind(this));
+
+        bindAll(this, 'stickOrUnstick', 'updateHeights');
+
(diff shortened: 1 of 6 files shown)


rich-nguyen:
I haven't even checked the code. :+1:

ScottPainterGNM:
screen shot 2016-01-07 at 16 28 00

static/src/javascripts/projects/common/modules/navigation/newSticky.js
```diff
@@ -0,0 +1,80 @@
...
+                    this.$header.css({
+                        'margin-top': this.adHeight + 'px'
+                    });
+                }.bind(this));
+            }.bind(this));
+        }
+    };
+
+    return new StickyHeader();
+});
```
OliverJAsh: As noted in person, one of the things I didn't like about old sticky nav was the OOP. I think we could make this less buggy and easier to read without the class and object mutation. I think we generally prefer that style for the rest of our code (more so in Scala).

Not important for the test, though, so don't let this stop you! :smile:

paperboyo:
sticky_no_more_lossy100

crifmulholland:
:+1:  for the comments

static/src/javascripts/projects/common/modules/navigation/newSticky.js
```diff
@@ -0,0 +1,80 @@
...
+    'lodash/functions/bindAll',
+    'common/utils/detect'
+], function (
+    fastdom,
+    $,
+    mediator,
+    bindAll,
+    detect) {
+
+    function StickyHeader() {
```
jbreckmckye: Why not just have the init method's content in the constructor?

Then you could just define your module as

    function StickyHeader() {
         setup();
    }

    return new StickyHeader();

This way you wouldn't have to know if newSticky had already been initialized when you grab the service - you wouldn't be splitting definition and generation. In your test code you could just do

     let stickyHeader;
     beforeEach(()=> {
         stickyHeader = new stickyHeaderService.constructor();
     });

Jholder112233:
Closing this as it's been replaced with https://github.com/guardian/frontend/pull/11503

Would be good to get comments on that code.

PR #11482, opened 2016-01-07
Moving adtest cookie setter to Standard mode.

kenlim:
The adtest cookie setter is essential for testing ads on the platform. Unfortunately, it was only run in Enhanced mode. 

Moving the function to Standard mode only increased the package size by 200B. The function only manipulates a domain cookie, so it shouldn't be a significant performance hit.


OliverJAsh:
Might be worth sharing information on here about size differences?

jbreckmckye:
:+1:

PR #11483, opened 2016-01-07
New Basher – This Is The NHS

sammorrisdesign-zz:
It's a new Basher! This one is for the NHS project that will be going live on the Monday. Here's what it looks like...

Mobile:
screen shot 2016-01-07 at 18 03 58

Tablet:
screen shot 2016-01-07 at 18 20 49

Desktop:
screen shot 2016-01-07 at 17 54 50


common/app/views/fragments/bashers/ThisIsTheNHS.scala.html
```diff
@@ -0,0 +1,41 @@
+@()(implicit request: RequestHeader)
+
+<div class="gs-container">
+    <div class="this-is-the-nhs">
+        <h1 class="title title--first">this is</h1>
```
jbreckmckye: A classical BEM approach would have us replace `this-is-the-nhs title` with `this-is-the-nhs__title` (or perhaps `nhs-campaign__title` for ease of typing). Off the top of my head, I can think of a few advantages to doing things this way:

1. It is clear that we are using a campaign-specific style rather than a generic `title` widget. It is also clear that an `nhs-campaign__grid` class is particular to the component, rather than a site-wide layout helper.
2. There is no chance of namespace collision with other generic styles, like a non-campaign `title` class.
3. We are keeping specificity low by avoiding a descendant combinator that we do not actually need.
4. As is, there is an invisible coupling between the div and the H1. It is not _explicit_ that the class on the container also styles the children. BEM makes the dependencies between containers and children very obvious.

What do you think?

common/app/views/fragments/bashers/ThisIsTheNHS.scala.html
```diff
@@ -0,0 +1,41 @@
...
+        <div class="stripe stripe--white"></div>
+
+        <div class="line">
+            <svg class="line__image" width="2200" height="419.084" viewBox="0 0 2200 419.084"><path clip-rule="evenodd" stroke="#FDBA12" stroke-width="12" stroke-miterlimit="10" d="M0 287.004h778l14.2-47.5 25.2 108 27.2-161 22.2 197 18.2-96.5h433l20.4-84.5 20.4 142.447 20.4-190.446 20.4 216 20.4-83.5h312l27.4-140.5 26.6 195.5 28-288 34 269 21-36h327" fill="none"/></svg>
+        </div>
+
+        <div class="circle"></div>
+        <div class="circle circle--lower"></div>
+
+        <img class="hand hand--left" src="http://static.guim.co.uk/sys-images/Guardian/Pix/pictures/2016/1/7/1452167267487/handleftcompressor.png" />
```
jbreckmckye: Isn't there a helper we would normally use to inject these URIs?

static/src/stylesheets/inline/basher.ThisIsTheNHS.scss
```diff
@@ -0,0 +1,424 @@
...
+        height: gs-height(6);
+    }
+
+    @include mq(desktop) {
+        height: gs-height(8);
+    }
+
+    .title {
+        position: absolute;
+        z-index: 8;
```
jbreckmckye: Why eight?

static/src/stylesheets/inline/basher.ThisIsTheNHS.scss
```diff
@@ -0,0 +1,424 @@
...
+    }
+
+    .description {
+        position: absolute;
+        top: 0;
+        max-width: gs-span(4);
+        @include fs-header(1);
+        color: #ffffff;
+        margin: 0 $gs-gutter / 2;
+        letter-spacing: -.2px;
```
jbreckmckye: I am always glad to see PRs which pay attention to these things.

static/src/stylesheets/inline/basher.ThisIsTheNHS.scss
```diff
@@ -0,0 +1,424 @@
...
+        }
+
+        @include mq(wide) {
+            left: 96px;
+        }
+    }
+
+    .line__image {
+        position: absolute;
+        top: -99999px;
```
jbreckmckye: Why is this so huge?

IIRC creating large planes will reserve a lot of memory in first-generation iPads.

static/src/stylesheets/inline/basher.ThisIsTheNHS.scss
```diff
@@ -0,0 +1,424 @@
...
+        height: 44px;
+        border-radius: 50%;
+        background-color: $c-accent;
+        display: none;
+
+        @include mq(mobileLandscape) {
+            display: block;
+        }
+
+        @include mq(tablet) {
```
jbreckmckye: Idea - rather than defining multiple components (`circle`, `hand`, etc.) and declaring their device-specific flavours within each block, would it perhaps make more sense to declare breakpoints and then, within them, the various changes that each entails? If I were thinking about the design, or describing it to others, that's how I'd go about organizing my description ("In mobile, we don't show the second hand, and we remove the lower circle, etc. etc.").

common/app/views/fragments/bashers/ThisIsTheNHS.scala.html
uplne: @jbreckmckye very nice summary

Making atoms like title, stripe etc makes sense if they will be reused across multiple components (molecules). And then it should be split into separate files, eg. _stripe.scss so it will be easy to find.

However as this is probably very campaign specific and might be removed in the future, it will be much better to follow BEM 'single class' approach as @jbreckmckye suggested. It is very logical, easy to read, easy to find (one component) and there are no nesting/specificity issues involved.

gtrufitt: :star2:

common/app/views/fragments/bashers/ThisIsTheNHS.scala.html
sammorrisdesign: Not to specific images, is there?

static/src/stylesheets/inline/basher.ThisIsTheNHS.scss
sammorrisdesign: To center it no matter what the width of the element, I can reduce the number though.

common/app/views/fragments/bashers/ThisIsTheNHS.scala.html
jbreckmckye: @sammorrisdesign I mean for injecting the domain. (The answer may be 'no').

static/src/stylesheets/inline/basher.ThisIsTheNHS.scss
sammorrisdesign: Both approaches have their downfalls though. What you're suggesting would make it difficult to find out about inheritance from smaller breakpoints. Unless I scope each breakpoint between two sizes, which would result in duplication.

Maybe more of a preference thing, unless there's something I'm missing.

common/app/views/fragments/bashers/ThisIsTheNHS.scala.html
sammorrisdesign: Soz came from hastily prepared thrasher, am switching it all to live under the `this-is-the-nhs` block.

static/src/stylesheets/inline/basher.ThisIsTheNHS.scss
gtrufitt: FWIW, it seems to be consistent with the codebase to have the mq within the component as done here.

jbreckmckye: @sammorrisdesign I see, so some styles are applied to a superset of breakpoints.

If that were the case I would be tempted to duplicate the styles for the sake of clarity. Inheritance is difficult to reason about.

common/app/views/fragments/bashers/ThisIsTheNHS.scala.html
```diff
@@ -0,0 +1,41 @@
+@()(implicit request: RequestHeader)
+
+<div class="gs-container">
+    <div class="this-is-the-nhs">
+        <h1 class="title title--first">this is</h1>
+        <h1 class="title title--second">the NHS</h1>
```
uplne: This is small thing and won't make any difference but why did you split this into two H1s?

As I understand, from SEO perspective it splits the weight of H1. It is ok to use multiple H1s in HTML5 but the recommendation is to use them as per block tag e.g. per section, not multiple H1s inside one section.

So wrapping the title--second in <span> inside <h1> should work better.

static/src/stylesheets/inline/basher.ThisIsTheNHS.scss
jbreckmckye: @gtrufitt It is - though I am not sure everyone agrees our current style code is very good. In any case, as a self-contained element with various design permutations, I think it would be nice if we could describe each permutation in its own block, so that it would be very easy to reason how each breakpoint would look like from reading each. But - I am not zealous.

common/app/views/fragments/bashers/ThisIsTheNHS.scala.html
jbreckmckye: Good :eyes:

sammorrisdesign: They're both absolutely positioned relative to the parent though. Wouldn't nesting the second line inside the first, mess this up?

static/src/stylesheets/inline/basher.ThisIsTheNHS.scss
gtrufitt: This is true and probably something for the client-side code style discussion, I think though that I prefer this style personally, instead of duplicating styles, purely that I think it matches the base style, ``--modification`` mindset like bem.

common/app/views/fragments/bashers/ThisIsTheNHS.scala.html
gidsg: Yeah, there is `Configuration.static.path` please use this as it will use the secure version if the user is on HTTPS.

common/app/views/fragments/bashers/ThisIsTheNHS.scala.html
gtrufitt: @sammorrisdesign Wrap both in an h1, then wrap these in spans and absolutely position them (without absolute or relative positioning on the h1)?

jbreckmckye: Couldn't we just do it as two spans inside an empty block (the H1)

sammorrisdesign: I know guys :disappointed:. that's why I deleted the comment because @sammorrisdesign dumb.

static/src/stylesheets/inline/basher.ThisIsTheNHS.scss
jbreckmckye: I'm happy to defer to others' opinions on this.

common/app/views/fragments/bashers/ThisIsTheNHS.scala.html
gtrufitt: I'll delete my comment too and no one will be any the wiser :dash:

common/app/views/fragments/bashers/ThisIsTheNHS.scala.html
sammorrisdesign: Cheers @gidsg, didn't know that existed

static/src/stylesheets/inline/basher.ThisIsTheNHS.scss
uplne: I've been in many discussions about this and tried both. They both have their pros and cons and in this case it is, from my experience, mostly only on personal preference.

I personally (as @gtrufitt) prefer our current style as in most cases I develop and test breakpoints per atom (circle, line etc) instead of per whole component. In that case having everything inside atom is easier than scrolling between breakpoints for the whole component.

sammorrisdesign:
Alright I think I've caught everything on this. Good to go?

gtrufitt:
:+1:

jbreckmckye:
:ship:

gtrufitt:
Hahaha, all that work then Jimmy closes it ;p

sammorrisdesign:
Nice passive aggressive no @jbreckmckye

jbreckmckye:
I've done that by accident more times than I'm happy to admit...

gtrufitt:
I feel like the size and position of the button is at odds with the amount
it actually gets used.

On 8 January 2016 at 15:07, Jimmy Breck-McKye <notifications@github.com>
wrote:

> I've done that by accident more times than I'm happy to admit...
>
> —
> <https://github.com/guardian/frontend/pull/11483#issuecomment-170026631>.
>

-- 

in England Number 908396

PR #11484, opened 2016-01-07
Hotfix broken embeds (CAPI client upgrade)

OliverJAsh:
[Updated CAPI client](https://github.com/guardian/frontend/pull/11466) represents Embed assets types as Image asset types. Doh. I think this is because CAPI client is missing the Embed asset type: https://github.com/guardian/content-api-scala-client/blob/0c9c4475c6ccd0e4cdf4910e40f47d3691b94cdf/src/main/thrift/content/v1.thrift#L158
# Before

http://www.theguardian.com/us-news/2015/dec/10/kern-county-california-police-killings-misconduct-district-attorney

image
# After

image

/cc @rich-nguyen

### DIFF
--- a/common/app/conf/switches/FeatureSwitches.scala
+++ b/common/app/conf/switches/FeatureSwitches.scala
@@ -5,6 +5,15 @@ import org.joda.time.LocalDate
 
 trait FeatureSwitches {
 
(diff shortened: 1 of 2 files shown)


rich-nguyen:
:+1:

PR #11486, opened 2016-01-08
Add logo to single-campaign commercial container

kelvin-chappell:
Eg.
image

It's in the wrong place at the moment - left justified instead of right justified.

@Calanthe


kenlim:
:gift_heart:

PR #11487, opened 2016-01-08
Remove pan domain

janua:
This removes the pandomain dependency and adds extra logging into `Notification.scala`.

@rich-nguyen @johnduffell


johnduffell:
:+1:

PR #11490, opened 2016-01-08
Report abuse page non ajax

gidsg:
Adds a form at `/discussion/report-abuse/:commentId` to Report Abuse on comments for non AJAX users. This is to replace the form currently provided by D2 microapp e.g. http://discussion.theguardian.com/components/report-abuse/27576482 which we would like to deprecate.

_N.B_ The styling will look a bit uglier that that shown until the changes in https://github.com/guardian/frontend/pull/11447 are merged. I believe @jamespamplin is looking into this.
## Tasks after merging this PR:
- [ ] Ship sister PR https://github.com/guardian/fastly-edge-cache/pull/542 to make routes HTTPS only
- [ ] Add link to new route 

screen shot 2016-01-08 at 12 17 22
screen shot 2016-01-08 at 12 17 07
screen shot 2016-01-08 at 12 27 03

CC @guardian/discussion @adamnfish

### DIFF
--- a/discussion/app/views/fragments/reportCommentForm.scala.html
+++ b/discussion/app/views/fragments/reportCommentForm.scala.html
@@ -4,63 +4,65 @@
         <label class="label">@errorMessage.get</label>
     </div>
 }
-@defining(userForm.error("categoryId")) { categoryIdError =>
-    @defining(userForm.error("email")) { emailError =>
-        @defining(userForm.error("reason")) { reasonError =>
+@defining(userForm.apply("email").value) { emailUserInput =>
+    @defining(userForm.error("categoryId")) { categoryIdError =>
+        @defining(userForm.error("email")) { emailError =>
+            @defining(userForm.error("reason")) { reasonError =>
 
-            @helper.form(controllers.routes.CommentsController.reportAbuseSubmission(commentId), 'class -> "form") {
-            @helper.CSRF.formField
-                <fieldset class="fieldset">
-                    <div class="fieldset__heading" />
-                    <input type="hidden" name="commentId" value="@commentId">
-                    <div class="fieldset__fields">
-                        <div class="form-field @{if(categoryIdError) {"form-field--error"}}">
-        <label class="label" for="categoryId">Category</label>
+                @helper.form(controllers.routes.CommentsController.reportAbuseSubmission(commentId), 'class -> "form") {
+                    @helper.CSRF.formField
+                    <fieldset class="fieldset">
+                        <div class="fieldset__heading" />
+                        <input type="hidden" name="commentId" value="@commentId">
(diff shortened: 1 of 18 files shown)


paperboyo:
"Email" field says it's optional. Why does it give "Please enter a valid email address" warning when empty?

gidsg:
Yeah @paperboyo, it is optional but if the user enters and invalid email it comes back without their original input. I will see if I can fix this.

paperboyo:
Thanks, @gidsg! I took into account that it may warn only on invalid while allowing empty. Haven't thought that empty may mean "I just forgot what you've put here"...

gidsg:
OK @paperboyo I've updated it now to retain the user input in the case of an invalid email address.
screen shot 2016-01-08 at 15 25 03

paperboyo:
Thank you, @gidsg!

gidsg:
Please may someone from the .com health team review this? CC @rich-nguyen

discussion/app/views/fragments/reportCommentForm.scala.html
```diff
@@ -0,0 +1,68 @@
+@(commentId: Int, userForm: Form[discussion.model.DiscussionAbuseReport], errorMessage: Option[String] = None)(implicit request: RequestHeader)
+@if(errorMessage) {
+    <div class="form-field--error">
+        <label class="label">@errorMessage.get</label>
+    </div>
+}
+@defining(userForm.apply("email").value) { emailUserInput =>
+    @defining(userForm.error("categoryId")) { categoryIdError =>
+        @defining(userForm.error("email")) { emailError =>
+            @defining(userForm.error("reason")) { reasonError =>
```
rich-nguyen: can you define a tuple instead?

discussion/conf/application.conf
```diff
@@ -35,8 +35,14 @@ play {
...
-
-  ws {
+  filters {
+      csrf {
+          cookie.name = GU_CSRF
+          cookie.secure = true
+          sign.tokens = true
+      }
+  }
+    ws {
```
rich-nguyen: indentation

rich-nguyen:
:+1:

discussion/app/views/fragments/reportCommentForm.scala.html
```diff
@@ -0,0 +1,67 @@
...
+    <div class="form-field--error">
+        <label class="label">@errorMessage.get</label>
+    </div>
+}
+@defining(
+    userForm.apply("email").value,
+    userForm.error("categoryId"),
+    userForm.error("email"),
+    userForm.error("reason")
+) { formValidationErrors =>
```
rich-nguyen: I mean something like `(email, categoryError, emailError, reason)`

PR #11494, opened 2016-01-08
Adding active adunit reports for Facebook IA and Mobile apps

kenlim:
As requested by Jamie and Kyle.


kelvin-chappell:
It's unfortunate about the mandatoryProperty dependency but apart from that I can't see *any* thing wrong with it. :ship:

PR #11495, opened 2016-01-08
Remove unused module parameter

jbreckmckye:
This is just a little change to remove an 'options' parameter in dfp-api.js that isn't actually set by anything - either in production or in test code.


kelvin-chappell:
Could it be being set by a DFP template?

jbreckmckye:
@kelvin-chappell Is that possible? How would I check?

uplne:
@kelvin-chappell what do you mean by that? DFP templates shouldn't have access to DFP init

kelvin-chappell:
That's true.  :+1:

PR #11496, opened 2016-01-08
Add logging inside sns match

janua:
No logging after `log.info(s"Issuing SNS notification: ${request.getSubject}:${request.getMessage}")` comes out.

Is it reaching `client.publishFuture`?

@rich-nguyen


rich-nguyen:
:+1:

PR #11498, opened 2016-01-08
Change logging in Notification

janua:
No longer need the extra logging.

Also rejig the logging around.

@rich-nguyen


rich-nguyen:
:+1:

PR #11499, opened 2016-01-08
Update email article Ab test

gtrufitt:
Revisit the in-article email sign-up after a fairly successful initial test, to extend out and test mid-article (3 paras from end) vs end of article.


desbo:
:+1:

PR #11501, opened 2016-01-08
Add api-key=dotcom query string parameter when querying the discussionApi

TBonnin:


TBonnin:
Q: to add test data I removed the `data/discussion` content and rerun the tests locally. Only 3 new test files have been recreated where there were 12 before. Were they stale test data?

discussion/app/discussion/DiscussionApi.scala
```diff
@@ -34,21 +40,24 @@ trait DiscussionApi extends Http with ExecutionContexts with Logging {
...
-                    |${if(params.displayThreaded) "" else "&displayThreaded=false"}
-                    |&showSwitches=true""".stripMargin.replace("\n", "")+
-                    params.maxResponses.map{i => "&maxResponses="+ i}.getOrElse("")
+    val parameters = List(
+      "pageSize" -> params.pageSize,
+      "page" -> params.page,
+      "orderBy" -> params.orderBy,
+      "displayThreaded" -> (params.displayThreaded match {
+        case false => "false"
+        case _ => None}),
```
jennysivapalan: Will this evaluate to `displayThreaded = "" ` in the `_` case?

discussion/app/discussion/DiscussionApi.scala
```diff
@@ -149,8 +176,10 @@ trait DiscussionApi extends Http with ExecutionContexts with Logging {
     }
   }
 
-  override protected def getJsonOrError(url: String, onError: (WSResponse) => String, headers: (String, String)*) =
+  override protected def getJsonOrError(url: String, onError: (WSResponse) => String, headers: (String, String)*) = {
+    println(url)
```
jennysivapalan: Can remove the `println`

jennysivapalan:
Looks good :+1: 
Yes they were probably stale - they are used in teamcity mainly. I think we only recreate new ones when tests are run locally but perhaps Discussion hasn't been run for a while.

discussion/app/discussion/DiscussionApi.scala
TBonnin: No. DisplayThreaded won't be added to the query string if None. That's the default behavior of the NetAPorter.uri when adding query string parameters lib https://github.com/NET-A-PORTER/scala-uri/blob/3c438f082392a722e1c0913b69fe21f458de5af2/src/main/scala/com/netaporter/uri/Uri.scala#L65

discussion/app/discussion/DiscussionApi.scala
TBonnin: oups good catch. thank you

rich-nguyen:
:+1:

PR #11503, opened 2016-01-08
Updated: 0% remove sticky nav test

holder-josh:
Paired with @OliverJAsh for this.

An updated version of https://github.com/guardian/frontend/pull/11481

This manages to skip all of the jank of the expanding adslot by adjusting the scrollY to compensate.

When this goes to prod we would move the sticky nav code up the chain of common.js, but at the moment that would require the current sticky nav executing earlier for everyone. 

_This is an initial 0% AB test that replaces the current sticky.js code with a much simplified version that only sticks the top ad._

_The current sticky nav implementation is a blocker on the 'toast' element of the upcoming liveblog work (which has a sticky top button) and is also the cause of all of these Github issues: https://github.com/guardian/frontend/labels/sticky-nav_


common/app/views/support/Commercial.scala
```diff
@@ -33,7 +33,7 @@ object Commercial {
     def adSizes(metaData: MetaData, edition: Edition): Map[String, Seq[String]] = {
       Map(
         "mobile" -> Seq("1,1", "88,70", "728,90"),
-        "desktop" -> Seq("1,1", "88,70", "728,90", "940,230", "900,250", "970,250")
+        "desktop" -> Seq("1,1", "940,230", "900,250", "970,250")
```
OliverJAsh: rm

static/src/javascripts/projects/common/modules/navigation/newSticky.js
```diff
@@ -0,0 +1,90 @@
...
+                                    $adBanner.css({
+                                        'position': 'fixed',
+                                        'top': 0
+                                    });
+                                }
+                                $header.css('margin-top', adHeight + 'px');
+                            }
+                        });
+                    });
+                };
```
OliverJAsh: Despite the fact `render` relies on values wrapped in promises, it is in fact a synchronous function because either one of the promises in `Promise.race` will always be fulfilled.

OliverJAsh: I wonder if the heart of this function would be better like so:

``` js
// Reset
$adBanner.css({
    'position': '',
    'top': ''
});
$header.css('margin-top', '');

// Set
if (window.scrollY > 0) {
    if (window.scrollY > (headerHeight)) {
        $adBanner.css({
            'position': 'absolute',
            'top': headerHeight + 'px'
        });
    } else {
        $adBanner.css({
            'position': 'fixed'
        });
    }
    $header.css('margin-top', adHeight + 'px');
}
```

First we reset all used CSS values, then we set the ones needed depending on the context.

Just an idea?

static/src/javascripts/bootstraps/enhanced/common.js
```diff
@@ -31,6 +31,7 @@ define([
     'common/modules/identity/cookierefresh',
     'common/modules/navigation/navigation',
     'common/modules/navigation/sticky',
+    'common/modules/navigation/newSticky',
```
johnduffell: best to avoid new in names otherwise you get the next-next-gen problem ;)

static/src/javascripts/projects/common/modules/navigation/newSticky.js
```diff
@@ -0,0 +1,90 @@
...
+                    return fastdom.read(function () {
+                        var adSlotPadding = parseInt($('#' + adId, $adBanner).css('padding-bottom')) * 2;
+                        // We must read the iframe attribute height to avoid reading the clientHeight mid-transition
+                        return Number($('iframe', $adBanner).attr('height')) + adSlotPadding;
+                    });
+                });
+
+                var getLatestAdHeight = function () {
+                    return Promise.race([newAdHeightPromise, oldAdHeightPromise]);
+                };
```
OliverJAsh: Thinking about it, we can simplify this file massively by just making this function read the height attribute.

static/src/javascripts/projects/common/modules/navigation/newSticky.js
```diff
@@ -0,0 +1,90 @@
...
+                var newAdHeightPromise = topAdRenderedPromise.then(function () {
+                    return fastdom.read(function () {
+                        var adSlotPadding = parseInt($('#' + adId, $adBanner).css('padding-bottom')) * 2;
+                        // We must read the iframe attribute height to avoid reading the clientHeight mid-transition
+                        return Number($('iframe', $adBanner).attr('height')) + adSlotPadding;
+                    });
+                });
+
+                var getLatestAdHeight = function () {
+                    return Promise.race([newAdHeightPromise, oldAdHeightPromise]);
```
johnduffell: this is kind of a slightly misleading way of saying getOrElse, does something like that exist?

static/src/javascripts/projects/common/modules/navigation/newSticky.js
```diff
@@ -0,0 +1,90 @@
...
+
+                mediator.on('window:throttledScroll', render);
+                render();
+                topAdRenderedPromise.then(render);
+
+                newAdHeightPromise.then(function (adHeight) {
+                    var diff = adHeight - oldAdHeight;
+
+                    if (window.scrollY !== 0) {
+                        window.scrollTo(window.scrollX, window.scrollY + diff);
```
johnduffell: great work!

static/src/javascripts/projects/common/modules/navigation/newSticky.js
```diff
@@ -0,0 +1,90 @@
...
+                var getLatestAdHeight = function () {
+                    return Promise.race([newAdHeightPromise, oldAdHeightPromise]);
+                };
+
+                var render = function () {
+                    getLatestAdHeight().then(function (adHeight) {
+                        fastdom.write(function () {
+                            if (window.scrollY === 0) {
+                                // Reset
+                                $header.css('margin-top', '');
```
johnduffell: for readability I'd consider splitting the css changes into functions with a useful name

johnduffell:
looks great to me! just seems so simple :+1:

static/src/javascripts/projects/common/modules/navigation/newSticky.js
```diff
@@ -0,0 +1,90 @@
...
+                render();
+                topAdRenderedPromise.then(render);
+
+                newAdHeightPromise.then(function (adHeight) {
+                    var diff = adHeight - oldAdHeight;
+
+                    if (window.scrollY !== 0) {
+                        window.scrollTo(window.scrollX, window.scrollY + diff);
+                    }
+                });
```
OliverJAsh: The above isn't actually tied to the sticky ad, it's just part of the top ad itself. Should we move it?

static/src/javascripts/projects/common/modules/navigation/newSticky.js
OliverJAsh: Scroll lookups trigger layout, so we should wrap this in fast Dom

static/src/javascripts/projects/common/modules/navigation/newSticky.js
```diff
@@ -0,0 +1,90 @@
...
+
+                var getLatestAdHeight = function () {
+                    return Promise.race([newAdHeightPromise, oldAdHeightPromise]);
+                };
+
+                var render = function () {
+                    getLatestAdHeight().then(function (adHeight) {
+                        fastdom.write(function () {
+                            if (window.scrollY === 0) {
```
OliverJAsh: We need to `fastdom.read` this

static/src/javascripts/projects/common/modules/navigation/newSticky.js
```diff
@@ -0,0 +1,90 @@
...
+                                // Reset
+                                $header.css('margin-top', '');
+                                $adBanner.css({
+                                    'position': '',
+                                    'top': ''
+                                });
+                            } else {
+                                if (window.scrollY > (headerHeight)) {
```
OliverJAsh: We need to `fastdom.read` this

OliverJAsh:
Just did some testing and it seems like we're not picking up the ad resize correctly 100% of the time. Let's chat on Monday.

static/src/javascripts/projects/common/modules/navigation/newSticky.js
```diff
@@ -0,0 +1,90 @@
...
+], function (
+    fastdom,
+    Promise,
+    $,
+    mediator,
+    detect) {
+
+    return function () {
+        if (detect.getBreakpoint() !== 'mobile') {
+            fastdom.read(function () {
```
regiskuckaertz: Because the window of an animation frame is very small and there are many other scripts vying for it, I'd limit `fastdom.read|write`s to the bare minimum (in this case, `$adBanner|$header.height()`) and start by creating a Promise that pipes the necessary DOM elements into it.

regiskuckaertz:
Great work! That top slot is something that's been bugging me since I arrived. I was wondering if, instead of animating on min-height, it wouldn't be better to position the ad absolutely from the beginning and instead `transform: translate()` the `body` element with an initial `margin-top` for the case where the ad is a 728×90. We could even alert the browser of the change as soon as the script starts by means of the `will-change` property.

paperboyo:
:gem:!

static/src/javascripts/projects/common/modules/navigation/newSticky.js
johnduffell: do promises run on a background thread?  I was under the impression they still used the main thread

regiskuckaertz: promises don't create threads no, but `fastdom` batches reads and writes in a single `requestAnimationFrame` call, and there are many of them throughout the code, so it's best to keep them as short as possible.

OliverJAsh: > I'd limit fastdom.read|writes to the bare minimum (in this case, $adBanner|$header.height()) and start by creating a Promise that pipes the necessary DOM elements into it.

Can you elaborate? I don't entirely understand.

OliverJAsh:
> Great work! That top slot is something that's been bugging me since I arrived. I was wondering if, instead of animating on min-height, it wouldn't be better to position the ad absolutely from the beginning and instead transform: translate() the body element with an initial margin-top for the case where the ad is a 728×90. We could even alert the browser of the change as soon as the script starts by means of the will-change property.

We currently transition the `height` on the iframe as this is what Google Tag sets on our behalf. Even so, there may be a way to improve the performance of the transition, but this change would not be related to the sticky ad. (The ad resizes/transitions whether or not we have a sticky ad.) I would suggest investigating this separately to avoid complicating the focus of this work.

static/src/javascripts/projects/common/modules/navigation/newSticky.js
```diff
@@ -0,0 +1,92 @@
...
+                    .then(getLatestAdHeight)
+                    .then(function (newAdHeight) {
+                        fastdom.read(function () {
+                            var diff = newAdHeight - oldAdHeight;
+
+                            if (window.scrollY !== 0) {
+                                window.scrollTo(window.scrollX, window.scrollY + diff);
+                            }
+                        });
+                    });
```
OliverJAsh: It might be worth leaving a comment to say why this lives here. (Why is it related to the sticky ad again? I can't remember, even though we just spoke about it.)

Jholder112233: Because if the ad wasn't sticky then there would be no need to adjust the margin on the header and therefore no need to therefore adjust the scroll position.

static/src/javascripts/projects/common/modules/navigation/newSticky.js
regiskuckaertz: Yes, so in your case

```
Promise.resolve([
  $('.js-top-banner-above-nav'),
  $('.js-header')
]).then(function (elems) {
  var $adBanner = elems[0];
  var $header = elems[1];
  return fastdom.read(function () {
    return [$adBanner.height(), $header.height()];
  });
}).then(...)
```

This is to reduce as much as possible the time needed to run the fastdom callback. It's a bit clearer I think since `fastdom.read` has been recently renamed `fastdom.measure`.

OliverJAsh: I see, yes we can move the query selectors outside the frame.

PR #11506, opened 2016-01-11
Deploys radiator improvements

OliverJAsh:
- Use build revision instead of first commit (useful for re-run builds where there are no changes)
- If the update fails, schedule another update
- Model API errors using Scala’s `Either` type. Thanks @adamnfish!


rich-nguyen:
:+1:

PR #11508, opened 2016-01-11
Fail the build when tests aren't passing

rich-nguyen:
This makes sure that switches which expire over the weekend cause a test failure. And that test failures prevent a riff raff upload/deploy.


jennysivapalan:
:+1:

PR #11518, opened 2016-01-11
Adjust names of buckets and objects for page press tests

JustinPinner:
As I had to modify platform to adjust permissions to buckets, I took the opportunity to improve the names we're using for the page pressing service.


jennysivapalan:
:+1:

PR #11519, opened 2016-01-11
Add logos to multiple-campaign commercial container

kelvin-chappell:
It looks like this at the moment:

image

So more styling will be required.

/cc @Calanthe


kelvin-chappell:
@kenlim has looked at this code and I've refactored some of it as a result.

paperboyo:
Looking at the Woodland one particularly... Where are they? Who manages them? Are they optimised (could call _via_ imgIX to force optimisation)?

kelvin-chappell:
@paperboyo The logos are stored in the ad server at the moment.  They're managed by Ad Ops. I don't know if they're optimised.

paperboyo:
Thanks, @kelvin-chappell, will try to have a closer look...

PR #11520, opened 2016-01-11
Remove topLargeSlot AB test

uplne:
We've got all data. Now is time to analyse. However as this is going to expire lets remove the ab test.

It was passing 'ad' and 'noad' custom parameters into DFP.


Calanthe:
:+1:

PR #11522, opened 2016-01-11
Age notice only shows when it is supposed to on AMP

NataliaLKB:
Shows the old content message on anything 30 days old or more and that is tagged with tone/news. 

This is specifically only for the onward journeys on AMP


johnduffell:
great! :+1:

PR #11523, opened 2016-01-11
Press R2 interactives

jennysivapalan:
I suspect this will take a few more iterations but this code is to ensure R2 interactives will be pressed with their scripts. It looks like R2 interactives are in a `div` with an id called `interactive-content`, scripts tags in this `div` will not be removed. 

Some interactives rely on `swfObject`, in this case a script tag will be added -  I have copied the code of this script tag from R2 `gu-core.js`

cc @JustinPinner


JustinPinner:
Nice! :+1: :flashlight: :page_with_curl: :facepunch:

PR #11525, opened 2016-01-12
An expired switch should retain it's value

rich-nguyen:
This is needed to ensure that we don't accidentally schedule unwanted behaviour when a switch expires. The expiry is only used to fail the tests.

### DIFF
--- a/common/app/conf/switches/Switches.scala
+++ b/common/app/conf/switches/Switches.scala
@@ -44,7 +44,7 @@ sealed trait SwitchTrait extends Switchable with Initializable[SwitchTrait] {
 
   val delegate = DefaultSwitch(name, description, initiallyOn = safeState == On)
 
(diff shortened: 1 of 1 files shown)


OliverJAsh:
:+1:

johnduffell:
:+1: excellent

kelvin-chappell:
Doesn't this mean expired switches will always be off?  So if a switch that was previously on expires just after midnight on a bank holiday Monday there will be unexpected behaviour until someone deploys the next build (with the switch extended or removed).

rich-nguyen:
@kelvin-chappell it means an expired switch will take whatever value has been set on the switchboard. The previous behaviour was "expired switches will always take their default state". That was causing a midnight expiry to change a switch value.

kelvin-chappell:
Oh yes.  I was looking at it the wrong way around.  As in I thought red was green.  I'm not colour-blind. :confounded:

PR #11527, opened 2016-01-12
Fix liveblog article 404s

johnduffell:
when I changed it to treat liveblogs without /live/ in the url as articles, I forgot I needed to update the route for fetching the JSON.  As a result, it was treating all json requests to be for articles, which meant it couldn't fetch the blocks (parsed in the old way)

This meant that liveblogs on fronts are causing 404s which was causing a big spike in the error rate.

This PR updates it to treat /live/*.json as a live blog so it can follow the correct old code path.


rich-nguyen:
:+1:

PR #11528, opened 2016-01-12
Get the tone from the page to style the email form

gtrufitt:
Update the email form styling so that it uses the tone of the article it is embedded in.

image

image

### DIFF
--- a/common/app/views/fragments/email/emailSignUp.scala.html
+++ b/common/app/views/fragments/email/emailSignUp.scala.html
@@ -12,7 +12,7 @@
 @formId = @{ componentClass + "-email-sub-form" }
 @inputId = @{ componentClass + "-email-sub-input" }
 
-@wrapperClass = @{"email-sub js-email-sub" + " email-sub--" + componentClass }
+@wrapperClass = @{"email-sub js-email-sub" + " email-sub--" + componentClass + " js-email-sub--" + componentClass}
 @formClass = @{ "email-sub__form js-email-sub__form" + " email-sub__form--" + componentClass }
 
 <div class="@wrapperClass">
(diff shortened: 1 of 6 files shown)


common/app/views/fragments/email/emailSignUp.scala.html
```diff
@@ -12,7 +12,7 @@
 @formId = @{ componentClass + "-email-sub-form" }
 @inputId = @{ componentClass + "-email-sub-input" }
 
-@wrapperClass = @{"email-sub js-email-sub" + " email-sub--" + componentClass }
+@wrapperClass = @{"email-sub js-email-sub" + " email-sub--" + componentClass + " js-email-sub--" + componentClass}
```
OliverJAsh: Tip! You could use string interpolation, e.g. `s"js-email-sub--${componentClass}"`. Don't worry about updating it though!

static/src/javascripts/projects/common/modules/email/email.js
```diff
@@ -210,6 +211,19 @@ define([
...
 
             },
+            setTone: function ($el) {
+                var toneClass = 'tonal__main--tone-',
+                    toneClassRegexp = new RegExp('(' + toneClass + ')([a-z-]+)'),
+                    tone;
+
+                if ($el.hasClass('js-email-sub--article')) {
+                    // Get tone from class
+                    tone = $('.tonal__main')[0].className.match(toneClassRegexp)[2];
```
OliverJAsh: Minor comment but I think it's fine to declare a var where it's used rather than explicitly hoisting. Linters can warn you if you use a variable before it is defined (`/*eslint no-use-before-define: 2*/`)

OliverJAsh: Idea: I wonder if we should use `Option` in JavaScript. There are two values here that could be undefined: the element and the result of the regexp match. `Option` allows you to declare a value may be undefined and easily handle those cases.

I'm just thinking out loud, ignore me!

OliverJAsh:
:+1:

PR #11530, opened 2016-01-12
Liveblog ads every 5th

regiskuckaertz:
Reopening [this PR](https://github.com/guardian/frontend/pull/11198). The code has been refactored to rely on spacefiller/spacefinder. Both had to go through some refactoring as well in order for it to work, mainly by adding new configuration options and making them more generic.


static/src/javascripts/projects/common/modules/article/space-filler.js
```diff
@@ -1,66 +1,47 @@
...
-                });
-            }
-
-            function onInsertionError(e) {
-                raven.captureException(e);
-                return false;
-            }
+        function insertionPromise(slots) {
+            return fastdom.write(function () {
+                return writer(slots);
```
jbreckmckye: It looks like this changes the signature of the function. `space-filler` used to return a true or false to indicate if a space had been found; now it returns the result of the writer, which is different (which means the jsDoc is no longer up to date).

One issue with that might be that if the writer itself returns a promise, then the writer has the ability to hold the rest of the space-filling writes hostage. You'd hope that people wouldn't send asynchronous writers to space-filler - they won't really work - but if they do, you want that mistake to break their code, not someone else's.

common/app/conf/switches/CommercialSwitches.scala
```diff
@@ -86,6 +86,15 @@ trait CommercialSwitches {
     exposeClientSide = true
   )
 
+  val LiveblogDynamicAdvertsSwitch = Switch(
+    "Commercial",
+    "liveblog-dynamic-adverts",
+    "Dynamically insert inline adverts on liveblogs",
+    safeState = Off,
+    sellByDate = new LocalDate(2016, 2, 14),
```
kelvin-chappell: Sunday.

static/src/javascripts/projects/common/modules/article/space-filler.js
```diff
@@ -1,66 +1,47 @@
...
+     * Given a set of spacefinder rules, applies a writer to the first matching paragraph.
+     * Uses fastdom to avoid layout-thrashing, but queues up asynchronous writes to avoid race conditions. We don't
+     * seek a slot for a new component until all the other component writes have finished.
+     *
+     * @param rules - a spacefinder ruleset
+     * @param writer - function, takes a para element and injects a container for the new content synchronously. It should NOT use Fastdom.
+     *
+     * @returns {Promise} - when insertion attempt completed, resolves 'true' if inserted, or 'false' if no space found
+     */
+    function fillSpace(rules, writer) {
```
jbreckmckye: I think the function name now implies we're passing a space to be filled

static/src/javascripts/projects/common/modules/article/space-filler.js
```diff
@@ -1,66 +1,47 @@
...
 ], function (
-    $,
-    idleFastdom,
+    fastdom,
     Promise,
     raven,
     spacefinder
 ) {
-    return new SpaceFiller();
+    var lastIteration = Promise.resolve(false);
```
jbreckmckye: Moving this out of the object makes the module stateful. This means we can no longer reset the state during unit testing.

static/src/javascripts/projects/common/modules/article/space-filler.js
```diff
@@ -1,66 +1,47 @@
...
-    'common/utils/$',
-    'common/utils/fastdom-idle',
+    'common/utils/fastdom-promise',
     'Promise',
     'raven',
     'common/modules/article/spacefinder'
 ], function (
-    $,
-    idleFastdom,
+    fastdom,
```
jbreckmckye: We're calling the module fastdom, but is that appropriate any longer? Increasingly we're actually building a Guardian-flavoured proxy for non-immediate DOM writes, and using the proxy as a hook for various optimizations. I think the name 'fastdom' for this module gives us the worst of both worlds: it constrains us not to diverge too radically from the underlying implementation, yet it also misleads readers. 

Instead I would be bold and embrace the fact we're building a specialized DOM proxy for the Guardian's distinct requirements. To that end, I might call it something like `lazyDom` instead. What do you think?

static/src/javascripts/projects/common/modules/article/open-module.js
```diff
@@ -31,23 +33,21 @@ define([
...
-                                        .addClass('element--supporting')
-                                        .insertBefore(space);
-                                    $('.submeta-container--break').removeClass('submeta-container--break');
-                                });
-                            }
-                        });
-                    }
+                spaceFiller.fillSpace(getSpacefinderRules(), function (spaces) {
+                    ajax({
+                        url: config.page.openModule,
```
jbreckmckye: It would be nice if we could make the writer synchronous, otherwise we run the risk of race conditions - we have no guarantee the found space will still exist when the AJAX response returns. I'm not sure if we want to resolve that (existing) issue here and now, though...

jbreckmckye:
There's quite a few commits here - should we maybe get into the habit of `git squashing` them if possible? It's something some of the other projects do.

static/src/javascripts/projects/common/modules/article/spacefinder.js
```diff
@@ -25,37 +22,91 @@ define([
...
-        };
+        // filter:(slot:Element, index:Integer, slots:Collection<Element>) -> Boolean
+        // will run each slot through this fn to check if it must be counted in
+        filter: null,
+        // startAt:Element
+        // will remove slots before this one
+        startAt: null,
+        // stopAt:Element
+        // will remove slots from this one on
+        stopAt: null,
```
jbreckmckye: Is the meaning of `stopAt` ambiguous? By the looks of things, an element matching `stopAt` is not included - is that what everyone will expect?

static/src/javascripts/projects/common/modules/article/spacefinder.js
```diff
@@ -25,37 +22,91 @@ define([
...
-        // test one element vs an array of other elements for the given rules
-        _testElems = function (para, others, rules) {
-            return chain(others).and(every, function (other) {
-                return _testElem(para, other, rules);
-            }).valueOf();
-        };
+        // filter:(slot:Element, index:Integer, slots:Collection<Element>) -> Boolean
+        // will run each slot through this fn to check if it must be counted in
+        filter: null,
+        // startAt:Element
```
jbreckmckye: I would add a line before each comment/property block.

static/src/javascripts/projects/common/modules/article/spacefinder.js
```diff
@@ -25,37 +22,91 @@ define([
...
+        // will remove slots before this one
+        startAt: null,
+        // stopAt:Element
+        // will remove slots from this one on
+        stopAt: null,
+        // reverse:Boolean
+        // will reverse the order of slots (this is useful for lazy loaded content)
+        reverse: false
```
jbreckmckye: Would `fromBottom` make more sense?

static/src/javascripts/projects/common/modules/commercial/liveblog-dynamic-adverts.js
```diff
@@ -0,0 +1,101 @@
...
+], function (
+    bonzo,
+    fastdom,
+    config,
+    mediator,
+    createAdSlot,
+    spaceFiller,
+    dfp
+) {
+    var INTERVAL = 5;      // number of posts between ads
```
jbreckmckye: Technically, the house style is to only use on var declaration. I'm not a huge fan of that rule, though...

static/src/javascripts/projects/common/modules/commercial/liveblog-dynamic-adverts.js
```diff
@@ -0,0 +1,101 @@
...
+            return windowHeight = document.documentElement.clientHeight;
+        })
+        .then(getSpaceFillerRules)
+        .then(fill);
+    }
+
+    return {
+        init: init,
+        settings: {
+            INTERVAL: INTERVAL,
```
jbreckmckye: Why expose this? AFAIK the only place you're reading this is the test, where the values could be hard-coded.

static/src/javascripts/projects/common/modules/commercial/liveblog-dynamic-adverts.js
OliverJAsh: Is it a house style? I don't think we have much of a up-to-date style guide, and it's not enforced by either linter. (I'm not a fan either)

static/src/javascripts/projects/common/modules/article/space-filler.js
regiskuckaertz: indeed that may be problematic. I had in mind that the result of the writer might be used somehow, but I think it's creating more harm than anything. Though I think rejecting the promise is a better behaviour if no space is found.

static/src/javascripts/projects/common/modules/article/space-filler.js
regiskuckaertz: well, yes it does. It's not a "para" we're looking for, but a space to be filled by the writer.

static/src/javascripts/projects/common/modules/article/space-filler.js
regiskuckaertz: I have a strong opinion too about which and how the guardian should choose open source libraries. As it seems Promises are the way forward (temporarily I hope), so let's embrace Promises and use them the way they're supposed to: all new async libraries should be Promise-based, and all existing ones should be "promisified". This way, the coding style will be harmonious and we can avoid anti-patterns like the ones we can see at the moment. That might mean indeed more of these proxies will be written eventually.

In any case `lazydom` seems like a good choice, as a matter of fact I raised [another PR](https://github.com/guardian/frontend/pull/11533) to do just that and will rename it that way.

static/src/javascripts/projects/common/modules/article/open-module.js
regiskuckaertz: maybe raise an issue on the board? or on github in the meantime?

static/src/javascripts/projects/common/modules/article/spacefinder.js
regiskuckaertz: I think it makes sense if you think backwards: (1) the default case is where `stopAt` is null and all the remaining elements are included and (2) the case where `stopAt` is the last slot and all the remaining elements but this one are included. It almost sounds like a recursive definition :-)

static/src/javascripts/projects/common/modules/commercial/liveblog-dynamic-adverts.js
regiskuckaertz: I'm not a fan of magic numbers, that's why. It's worth mentioning that when we start coding in ES6 lingo, this code will go straight to:

```
const INTERVAL = 5;
const OFFSET = 1.5;
```

static/src/javascripts/projects/common/modules/commercial/liveblog-dynamic-adverts.js
regiskuckaertz: just for unit tests, but indeed it's not that important.

regiskuckaertz:
@jbreckmckye thanks a lot for your review, I'll update the code as soon as I have time today. Can you show me what `git squash` is when we have 5 mins?

static/src/javascripts/projects/common/modules/article/space-filler.js
jbreckmckye: Sounds good!

static/src/javascripts/projects/common/modules/article/space-filler.js
jbreckmckye: We're not actually passing a space, though, which might be inferred from the name. Also, I'd argue that the name no longer makes clear that we're injecting the content at the first match.

rich-nguyen:
The issues with the original pr are still unresolved. We are still are in the process of implementing paginated live blogs, and I personally believe we should be establishing a line in the sand about number of mpu's per single piece of content.

static/src/javascripts/projects/common/modules/article/space-filler.js
jbreckmckye: I'm not sure if I would reject the promise in that case; I think the idiom is more for reject to be used for exceptions, and this isn't something I'd normally use `throw` for. But I guess it is fine to choose our own promise interface so long as we document it.

static/src/javascripts/projects/common/modules/article/spacefinder.js
jbreckmckye: Plus, someone can always read the documentation ;-)

static/src/javascripts/projects/common/modules/commercial/liveblog-dynamic-adverts.js
jbreckmckye: If no-one else objects, I am happy to see multiple var declarations. I think they are just easier to work with.

static/src/javascripts/projects/common/modules/article/open-module.js
jbreckmckye: I am not even 100% sure if this is causing bugs right now - you are right, we shouldn't handle it in this PR.

regiskuckaertz:
@rich-nguyen They are, in fact. The business logic is that ads are inserted every screenful, with a minimum of 5 posts in between. This might change based on feedback, that is why this behaviour is behind a switch at the moment. This PR will work alongside the new paginated liveblog without any change.

static/src/javascripts/projects/common/modules/article/open-module.js
```diff
@@ -31,23 +33,21 @@ define([
...
+                        crossOrigin: true,
+                        method: 'get'
+                    }).then(function (resp) {
+                        if (resp.html) {
```
OliverJAsh: When wouldn't there be a `resp.html`? Do we need to throw an error if there isn't one? (Silent errors are the worst.)

rich-nguyen:
@regiskuckaertz we haven't shipped paginated live blogs yet? I don't have any issues with the refactoring here, if those are the issues you are referring to.

The argument that "business logic dictates we must do X" is something we often have to apply technical sense to.

static/src/javascripts/projects/common/modules/article/space-filler.js
OliverJAsh: Are we suggesting `lazydom` as a name for a promise wrapped version of `fastDom`? I think it's better to be explicit. I've seen promisfy used as terminology before, perhaps `promisifiedFastDom`, but it is a bit long. My point is explicitness is better as it won't cause confusion.

static/src/javascripts/projects/common/modules/commercial/liveblog-dynamic-adverts.js
OliverJAsh: :+1:

static/src/javascripts/projects/common/modules/article/space-filler.js
regiskuckaertz: Agreed. Might be sth worth discussing during the next dotcom engineering meeting because I feel there is no consensus on many of these things.

common/app/conf/switches/CommercialSwitches.scala
regiskuckaertz: @kelvin-chappell argh, I looked at my calendar, but I'm still used to Mo-Su weeks, not Su-Sa like it is here in the UK :'(

static/src/javascripts/projects/common/modules/article/space-filler.js
regiskuckaertz: Don't you find it a bit annoying to pollute the namespace and API coverage just for the sake of testing? This is sth that is direly missing in AMD, some kind of module introspection. I'll add a reset method to `fillSpace`.

static/src/javascripts/projects/common/modules/commercial/liveblog-dynamic-adverts.js
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    function onUpdate() {
+        stopListening();
+        Promise.resolve(getSpaceFillerRules(windowHeight, true)).then(fill);
+    }
+
+    function init() {
+        return fastdom.read(function () {
+            return windowHeight = document.documentElement.clientHeight;
```
OliverJAsh: We reassign variables in here a lot, which hurts readability. I wonder if it's possible to avoid that?

static/src/javascripts/projects/common/modules/article/space-filler.js
jbreckmckye: The question is, when is it no longer appropriate to use the name fastdom? The proxy we're looking at here is fastdom wrapped in an idle callback wrapped in a promise interface. Should the name reflect all those details?

I wonder if we should just say, 'this is the Guardian proxy for writing to the DOM, it'll do things as fast as possible but at the cost of synchronicity, and that's our interface'. I tend to think that dropping the name 'fastdom' will make developers feel more free to make the proxy thicker - they are less bound to strictly follow the behaviour of fastdom, so they can add more optimizations and filters. This may or may not be a good thing, and may not be necessary yet in any case.

The more I think of it, the less I like the idea of a big, thick proxy. So perhaps keeping the word 'fastdom' in our name is a subtle way to keep the wrapper thin, at least for the moment.

regiskuckaertz: if you don't mind I'll do it [there](https://github.com/guardian/frontend/pull/11533).

static/src/javascripts/projects/common/modules/article/space-filler.js
jbreckmckye: In an ideal worldI suppose I would rather declare my modules as constructors, and then instantiate them in some kind of service locator. I definitely agree that it is not a good idea to change the API for the sake of testing.

That's why I originally implemented space-filler as an instance of a constructor - so I could avoid a test interface whilst allowing tests to create their own module instances with the `object.constructor` approach. I don't know if it's the ideal approach, or if it solves every problem. The most obvious alternative, I guess, is to write the test utils as pseudo-privates (`_reset`, etc.), but then you have to wrap everything in eslint switches, which feels a bit awkward.

How were modules written in your last project?

static/src/javascripts/projects/common/modules/article/open-module.js
regiskuckaertz: Let's raise an issue and deal with it in another PR, if that's fine?

static/src/javascripts/projects/common/modules/commercial/liveblog-dynamic-adverts.js
regiskuckaertz: Are you referring to `prevSlot` and `prevIndex`? I might use functional composition to isolate the backtracking logic, but there will be a need to keep track of the previous "thing" anyway. Think of it like the pivot in quicksort :-)

static/src/javascripts/projects/common/modules/article/space-filler.js
regiskuckaertz: We tried to keep our modules as simple as possible, which means doing one thing only. Consequently the surface of our APIs was quite small (and easy to debug). For UI widgets we resorted to a basic plugin architecture that built behaviour on top of the module, again with as few hooks as necessary (mainly creation and destruction). I have to say OO in javascript feels like monkey patching, I tend to use it only for objects with a very short lifespan.

common/app/conf/switches/CommercialSwitches.scala
paperboyo: No way it's Sunday! There is a [norm](https://en.wikipedia.org/wiki/ISO_8601) for that! :wink:

static/src/javascripts/projects/common/modules/article/open-module.js
regiskuckaertz: it's worth pointing out that when.js has some excellent sugarcoating on top of promises, some of which could make our code easier to read:

```
PromiseInst.then(... do some ajax stuff ...).**timeout(1000)**
PromiseInst.then(... return an array ...).**spread**(function(elem1, elem2, ...) { })
PromiseInst.**yield**(valueOrPromise)
```

as well as a "promisification" library that works on objects, functions, and callbacks.

static/test/javascripts/spec/common/content/space-filler.spec.js
```diff
@@ -37,31 +35,30 @@ define([
...
         it('If it finds a space, it calls the writer', function (done) {
             mockSpacefinderResult = document.createElement('p');
-            spaceFiller.insertAtFirstSpace(mockRules, function mockWriter() {
+            spaceFiller.fillSpace(mockRules, function mockWriter() {
                 done();
             });
         });
 
-        it('If it finds a space, it resolves the promise with `true`', function (done) {
+        it('If it finds a space, it resolves the promise with the return value of the writer', function (done) {
```
jbreckmckye: Is this no longer going to be true?

static/src/javascripts/projects/common/modules/article/space-filler.js
```diff
@@ -1,66 +1,52 @@
...
+                return true;
+            });
+        }
+    }
 
-            function onInsertionError(e) {
-                raven.captureException(e);
-                return false;
-            }
+    fillSpace.reset = function () {
```
jbreckmckye: I'm still not sure this approach is preferable to what the module originally did...

regiskuckaertz:
@jbreckmckye I reverted to a singleton. This is an interesting topic for dotcom-hygiene.

jbreckmckye:
:+1:

static/src/javascripts/projects/common/modules/article/space-filler.js
```diff
@@ -1,21 +1,18 @@
...
 ], function (
-    $,
-    idleFastdom,
+    fastdom,
     Promise,
     raven,
     spacefinder
 ) {
-    return new SpaceFiller();
+    var lastIteration;
```
jbreckmckye: Won't this still be stateful? The reason I originally made this a private variable of the SpaceFiller module was that this way, a test could re-initialize the whole thing and be sure that it was working with a tabula rasa. I guess it won't impact the test as it stands, though.

PR #11532, opened 2016-01-12
Split S3 and Dynamo into code and prod targets

JustinPinner:
**For Page Pressing**

Make sure we have a clean separation of CODE and PROD resources. 

We used to be ok with a single bucket for archived files because we didn't have any means of pressing pages without doing them as a one-off dev task or manually as individual updates.

Now we must prevent work done in CODE from overwriting PROD archived data.

cc @jennysivapalan


jennysivapalan:
:+1: :zap:

PR #11533, opened 2016-01-12
Fastdom Promise complying with original Fastdom api

regiskuckaertz:
Makes the promisified Fastdom comply with [the original API](https://github.com/wilsonpage/fastdom#fastdommeasurecallback-context) which accepts a second parameter.

Minor improvement by making `promisify` a higher order function, saving a function call every time `fastdom.read|write` is called.

/cc @OliverJAsh @sndrs

### DIFF
--- a/static/src/javascripts/projects/common/utils/fastdom-promise.js
+++ b/static/src/javascripts/projects/common/utils/fastdom-promise.js
@@ -5,20 +5,22 @@ define([
     fastdom,
     Promise
 ) {
-    function fastDomAction(action, fn) {
-        return new Promise(function (resolve) {
-            fastdom[action](function () {
-                resolve(fn());
+    function promisify(fdaction) {
+        return function(fn, ctx) {
+            return new Promise(function (resolve, reject) {
+                fdaction(function () {
+                    try {
+                        resolve(fn());
+                    } catch (e) {
+                        reject(e);
+                    }
+                }, ctx);
             });
-        });
+        }
     }
 
     return {
-        read: function (fn) {
-            return fastDomAction('read', fn);
-        },
-        write: function (fn) {
-            return fastDomAction('write', fn);
-        }
+        read: promisify(fastdom.read),
+        write: promisify(fastdom.write)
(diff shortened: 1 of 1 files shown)


OliverJAsh:
:+1: Looks fine but I would question the usage of contexts and `this`.

sndrs:
looks as thought it's updated too, might be worth updating us? plus there seems to be a more official take on it https://github.com/wilsonpage/fastdom/blob/master/extensions/fastdom-promised.js?

regiskuckaertz:
Well, yeah :) There are many instances of `fastdom(function() {}.bind(this))` in the code and I find this unsightly, also perhaps less efficient.

regiskuckaertz:
@sndrs it's using ES6 Maps, might be a blocker no? Also +1 for upgrading to the v1 API.

static/src/javascripts/projects/common/utils/fastdom-promise.js
```diff
@@ -5,20 +5,22 @@ define([
...
     }
 
     return {
-        read: function (fn) {
-            return fastDomAction('read', fn);
-        },
-        write: function (fn) {
-            return fastDomAction('write', fn);
-        }
+        read: promisify(fastdom.read),
```
piuccio: shouldn't it be `read: promisify(fastdom.read.bind(fastdom)` ? otherwise `fdaction` get called with global scope

sndrs:
> it's using ES6 Maps, might be a blocker no?

yeah that's a good spot!

regiskuckaertz:
@sndrs I have to admit I don't see the point of using a Map, if there isn't any logic to check if the promise has already been resolved—which could be an improvement if and only if the callback had no side effect... which admittedly never happens, by definition the DOM is mutable.

static/src/javascripts/projects/common/utils/fastdom-promise.js
regiskuckaertz: oh, good catch, thanks!

PR #11534, opened 2016-01-12
Repositioned selection sharing buttons

zeftilldeath:
Something I've been meaning to do for a while now. 

For people that habitually highlight text whilst reading it, the sharing buttons cover the text they're about to read. I've repositioned them to remedy this. 

Lots of people in dotcom feedback have had the same gripe:
https://groups.google.com/a/guardian.co.uk/forum/#!searchin/dotcom.feedback/sharing/dotcom.feedback/KaBFVsJp5Zg/MLAo5M-PEgAJ

Before:
screen shot 2016-01-12 at 17 02 50

After:
screen shot 2016-01-12 at 17 02 24

### DIFF
--- a/static/src/javascripts/projects/common/modules/ui/selection-sharing.js
+++ b/static/src/javascripts/projects/common/modules/ui/selection-sharing.js
@@ -57,7 +57,7 @@ define([
         if (selection && selection.rangeCount > 0 && selection.toString()) {
             range = selection.getRangeAt(0);
             rect = clientRects.getBoundingClientRect(range);
-            top = $body.scrollTop() + rect.bottom;
+            top = $body.scrollTop() + rect.top;
             twitterMessage = range.toString();
 
             if (!isValidSelection(range)) {
--- a/static/src/stylesheets/module/_social.scss
+++ b/static/src/stylesheets/module/_social.scss
@@ -191,8 +191,8 @@ category: Common
 
 .selection-sharing {
     position: absolute;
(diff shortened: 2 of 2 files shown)


johnduffell:
ship it!! :speedboat: :racehorse: :+1:

paperboyo:
Yeah, but what if the reader starts selecting text which sits at top of the browser window?

selection-at-top

(messing around a z-index may not be worth it as sticky situation is ending soon, right?)

Also, slightly related, is the [feedback](https://mail.google.com/mail/u/1/#label/dotcom.feedback/1523aa3502a3afa5) about this functionality breaking a screen reader. Maybe the whole of it should have a switch in accessibility options? [just thinking out loud]

zeftilldeath:
@paperboyo True, but I see that as the edge of the page. It's the equivalent of highlighted text at the bottom of the viewport previously going off page. (and yeah you're right sticky is on the way out.)

Would look a bit weird too: 
screen shot 2016-01-13 at 13 57 55

Screen reader thing is interesting, probably should be looked at!

paperboyo:
Noticed one more thing. They are hidden under the image when it’s showcased main media (other main medias are fine). The bubbles are not functioning, but actually appear nicer being a bit transparent (this could help readability when they do cover text...):

sharing popups under showcased image

URL: http://www.theguardian.com/world/2016/jan/19/alexander-litvinenko-the-man-who-solved-his-own-murder

PR #11535, opened 2016-01-12
New sticky ad test changes

holder-josh:
This simplifies the logic in an effort to remove jank on the initial scroll.

The CSS addition would be moved to the stylesheets when/if productionised.

### DIFF
--- a/static/src/javascripts/projects/common/modules/navigation/newSticky.js
+++ b/static/src/javascripts/projects/common/modules/navigation/newSticky.js
@@ -43,25 +43,18 @@ define([
                         fastdom.read(function () {
                             var scrollY = window.scrollY;
                             fastdom.write(function () {
-                                // Reset
-                                $header.css('margin-top', '');
-                                $adBanner.css({
-                                    'position': '',
-                                    'top': ''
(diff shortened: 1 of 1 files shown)


Jholder112233:
cc @OliverJAsh 

Still got problems with white space between the ad expanding and the margin-top animating, but that can be solved, just not easily in a AB test.

static/src/javascripts/projects/common/modules/navigation/newSticky.js
```diff
@@ -43,25 +43,18 @@ define([
...
-                                    $header.css('margin-top', adHeight + 'px');
+                                if (scrollY > headerHeight) {
+                                    $adBanner.css({
+                                        'position': 'absolute',
+                                        'top': headerHeight + 'px'
+                                    });
+                                } else {
+                                    $adBanner.css({
+                                        'position': 'fixed',
+                                        'top': 0
```
OliverJAsh: Minor comment but `top: ''` would unset the inline style

Jholder112233:
Going to do a bit more testing on this one

PR #11538, opened 2016-01-13
handle any liveblog with /live/ in the first 2 dir

johnduffell:
fix formatting on http://www.theguardian.com/film/filmblog/live/2015/feb/22/oscars-2015-live-red-carpet-and-arrivals-ceremony-and-winners

before it was being treated as an article but it actually did have live in the url so it should have coped with it using the new codepaths


NataliaLKB:
:+1:

PR #11539, opened 2016-01-13
Extend period for Identity Sign in V2 AB test

jamespamplin:
We're adding a new variant (B) for the new Identity Sign in guardian/identity#54

To support this, we need to extend IdentitySignInV2 test period, audience (now 20%), and add variant B.


rich-nguyen:
:+1:

jamespamplin:
Thanks @rich-nguyen

PR #11540, opened 2016-01-13
Sticky ad test tweaks

holder-josh:
Fixes the top ad on render to remove jank on first scroll. As a result of this we also have to animate the margin-top addition to the header.

### DIFF
--- a/static/src/stylesheets/module/_adslot.scss
+++ b/static/src/stylesheets/module/_adslot.scss
@@ -69,6 +69,8 @@
     .js-on & {
         z-index: 1022;
         width: 100%;
+        transform: translateZ(0);
+        will-change: transform;
     }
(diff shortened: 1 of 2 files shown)


Jholder112233:
@OliverJAsh let's :ship:?

static/src/stylesheets/module/_adslot.scss
```diff
@@ -69,6 +69,8 @@
     .js-on & {
         z-index: 1022;
         width: 100%;
+        transform: translateZ(0);
+        will-change: transform;
```
OliverJAsh: We should document this to say it's necessary for `position: fixed`.

OliverJAsh:
We did it! :+1: Nice

PR #11541, opened 2016-01-13
Adding clientside variable to AMP ad calls.

kenlim:
AMP does not permit the running of JS on their pages. 

But this amp-3p-iframe-src meta tag lets us run something specific.


common/app/views/mainAMP.scala.html
```diff
@@ -8,6 +8,8 @@
     <head>
         @* "utf-8" meta tag needs to be first according to AMP spec *@
         <meta charset="utf-8">
+            <meta name="amp-3p-iframe-src" content="https://static.theguardian.com/amp/remote.html">
```
NataliaLKB: indentation

NataliaLKB:
Otherwise :+1:

PR #11544, opened 2016-01-13
Ensure & are not encoded for values in omniture script.

jennysivapalan:
Page name for Life & Style was not correctly being sent through. I think & is the only case where we need to ensure it is encoded.

cc @JustinPinner


JustinPinner:
:+1: :footprints: :shipit:

PR #11545, opened 2016-01-13
Update amp params

kenlim:
Adding p=amp and the content type + platform to the ad unit. (As required by Commercial for auditing purposes.)


stephanfowler:
:+1:

common/app/views/support/cleaner/AmpAdCleaner.scala
```diff
@@ -108,7 +109,7 @@ case class AmpAdCleaner(edition: Edition, uri: String, article: Article) extends
         "edition" -> edition.id.toLowerCase(),
         "se" -> parseIds(article.trail.tags.series).mkString(","),
         "ct" -> article.metadata.contentType,
```
NataliaLKB: Should this be the variable `contentType`?

kenlim: I believe most content objects copy the value from here.

PR #11548, opened 2016-01-13
Amp pages now support ?adtest= param setting

kenlim:
If you attach ?adtest=tester1 to an AMP url, AMP will use the adtest param to set a new targeting param of at=tester1 to DFP... just like in NextGen


NataliaLKB:
:+1:

PR #11549, opened 2016-01-13
paginate business and politics blogs

johnduffell:
So in preparation for the pagination of live blogs, we will try hard pagination in a basic way.

This PR adds pagination for those once it gets to 29, the 30th post will cause the oldest 10 to be moved to another page.  This will happen every time the main page hits 30.

There will be a rudimentary navigation link at the bottom of each page.

Key events will only be scraped from the current page, but hopefully this can change very soon.

Metadata should be there to keep google happy.

The threshold for hiding old posts used to be set to 10, this has been extended to 29 in the intervening period so that the two mechanisms don't conflict.

### DIFF
--- a/article/app/controllers/ArticleController.scala
+++ b/article/app/controllers/ArticleController.scala
@@ -87,7 +87,9 @@ object ArticleController extends Controller with RendersItemResponse with Loggin
       if (request.isAmp) {
         NotFound
       } else {
-        val blocks = BodyBlocks(blog.article.content.fields.blocks, pageNo)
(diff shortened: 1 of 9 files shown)


NataliaLKB:
:+1: woohoo!

piuccio:
nice but pagination shouldn't be done with search queries `?page=`, for SEO reasons it's preferable if pages are part of the URL, `/page2`

stephanfowler:
@piuccio we've registered in Google webmaster which query params affect content; basically just `page`. We use it already, for example: http://www.theguardian.com/world/middleeast?page=2

PR #11550, opened 2016-01-14
Use configured environment name on page press form

JustinPinner:
I had set the environment value to "TEST" - but now we should use the configured environment name to avoid confusion.

cc @jennysivapalan


jennysivapalan:
+1

PR #11551, opened 2016-01-14
Deploy rad: Add super quick and basic styling for big screen

gtrufitt:
@OliverJAsh Very basic and appalling css but quick fix so it fits on the tv - now you just need to get the tv to stay on ;p

cc/ @rich-nguyen

### DIFF
--- a/static/src/deploys-radiator/app/main.css
+++ b/static/src/deploys-radiator/app/main.css
@@ -4,6 +4,10 @@ body {
     padding: 1rem;
 }
 
+h1 {
+    font-size: 3em;
+}
+
 .deploys {
     list-style: none;
     display: flex;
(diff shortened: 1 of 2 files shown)


static/src/deploys-radiator/app/main.css
```diff
@@ -71,3 +75,9 @@ a {
 a:hover {
     text-decoration: underline;
 }
+
+div > div {
+    float: left;
+    width: 31%;
+    margin: 0 1%;
+}
```
OliverJAsh: I've been using flexbox but the browser on the Pi doesn't support it. We need to try and get Firefox/Chrome on there :tongue:

OliverJAsh:
:+1:

PR #11552, opened 2016-01-14
Update omniture.js

mkopka:
Updated base library code to upgrade to omniture latest library version 1.6
Also included new clickmap plug in as partaking in Adobe clickmap Beta.


jbreckmckye:
:+1:

PR #11553, opened 2016-01-14
Small improvement to switches tests

kelvin-chappell:
I was perplexed.  The switch tests I was expecting to fail were passing.
This was because if you inspect the state of an akka agent you can't guarantee it's up-to-date.
So I've modified the tests a bit to get predictable results.  
This is only likely to affect you if you run the switch tests in isolation.

I've also removed the `SwitchTrait` which had become redundant.

### DIFF
--- a/common/app/conf/switches/Switches.scala
+++ b/common/app/conf/switches/Switches.scala
@@ -3,8 +3,7 @@ package conf.switches
 import java.util.concurrent.TimeoutException
 
 import common._
-import conf.Configuration.environment
-import org.joda.time.{DateTime, Days, Interval, LocalDate}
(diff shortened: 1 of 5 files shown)


common/app/conf/switches/Switches.scala
```diff
@@ -69,23 +68,17 @@ sealed trait SwitchTrait extends Switchable with Initializable[SwitchTrait] {
 
   def expiresSoon = daysToExpiry < 7
 
-  def hasExpired = daysToExpiry == 0
+  def hasExpired = daysToExpiry <= 0
```
johnduffell: :dog:

johnduffell:
good stuff! :+1:

PR #11555, opened 2016-01-14
fix liveblog layout on preview

johnduffell:
Preview is treating liveblogs as articles rather than liveblogs which is causing the layout to be dodgy.  On article it actually looks at the path to see it it contains /live/ before deciding whether it's a liveblog.

On preview it seems to go through standalone's routes which just gets the items from capi and then finds a controller to deal with it.  This could be a live blog so making it retrieve the blocks, otherwise the liveblogs will appear in a bad format.

This shouldn't cause a big impact on bandwidth as I think standalone is only used for preview etc.

@rich-nguyen I think my assumptions are good there regarding where standalone is used?


stephanfowler:
:+1:

rich-nguyen:
:+1:

PR #11557, opened 2016-01-14
Apply a "frontendarchive" omniture param for cleaned, pressed pages.

JustinPinner:
So that we can differentiate NGW archived pages vs. legacy R2-sourced pages in omniture data.

cc @jennysivapalan


jennysivapalan:
:+1: :chart_with_upwards_trend:

PR #11558, opened 2016-01-14
use per-action filter and don't cache form (Discussion Report Abuse non-ajax)

gidsg:
Forms with CSRF tokens are uncacheable, so previously trying to cache the form was causing unexpected behaviour. Also, using a per-action filter rather than a global filter.

CC @guardian/discussion


gidsg:
CC @rich-nguyen please may someone from Dotcom Health team review this?

jennysivapalan:
:+1:

PR #11563, opened 2016-01-14
Fix preview liveblogs

johnduffell:
Autoupdate on liveblogs asks for the json version with lastBlock=.....

However preview runs standalone app which didn't understand these parameters so was ignoring them.  Updating it to send json article requests to the code that can handle these parameters.

Before it was auto updating constantly with the whole page, which made preview useless.


rich-nguyen:
:+1:

PR #11567, opened 2016-01-14
A/B: Restrict injection of email form

gtrufitt:
Restrict to only injecting when the page’s last 5 elements are paragraphs

### DIFF
--- a/static/src/javascripts/projects/common/modules/experiments/tests/rtrt-email-form-article-promo.js
+++ b/static/src/javascripts/projects/common/modules/experiments/tests/rtrt-email-form-article-promo.js
@@ -7,7 +7,8 @@ define([
     'common/modules/email/email',
     'common/utils/detect',
     'lodash/collections/contains',
-    'common/utils/config'
+    'common/utils/config',
+    'lodash/collections/every'
 ], function (
     $,
     bean,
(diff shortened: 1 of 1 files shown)


static/src/javascripts/projects/common/modules/experiments/tests/rtrt-email-form-article-promo.js
```diff
@@ -32,16 +32,26 @@ define([
...
             var host = window.location.host,
                 escapedHost = host.replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, '\\$&'), // Escape anything that will mess up the regex
                 urlRegex = new RegExp('^https?:\/\/' + escapedHost + '\/(uk\/|us\/|au\/|international\/)?([a-z-])+$', 'gi'),
                 browser = detect.getUserAgent.browser,
-                version = detect.getUserAgent.version;
+                version = detect.getUserAgent.version,
+                allArticleEls = $('> *', $articleBody),
+                lastFiveElsParas = true;
+
+            for (var i = allArticleEls.length - 1; i > allArticleEls.length - 6; i--) {
```
desbo: this is maybe a little tricky to read and could be simplified to a single expression, with the paragraph check as a standalone function e.g. `lastFiveElsParas = allArticleEls.slice(allArticleEls.length - 5).every(isParagraph)`

PR #11568, opened 2016-01-14
Create new id privacy tab and move marketing prefs there

danmassie:
A number of new privacy / data settings are going to be made available to our users. We need to create a new home for these within edit profile.

Marketing preferences are currently at the bottom of the email preferences area. We should move these under a new tab within edit profile.

By moving these settings to the edit profile, we're forcing the user to re-authenticate.

Before:
screenshot from 2016-01-14 15 36 44

After:
upload_1_13_2016_at_11_23_30_am

upload_11_19_2015_at_2_53_21_pm

### DIFF
--- a/identity/app/views/fragments/profile/privacyForm.scala.html
+++ b/identity/app/views/fragments/profile/privacyForm.scala.html
@@ -1,0 +1,42 @@
+@(idUrlBuilder: services.IdentityUrlBuilder,
+  idRequest: services.IdentityRequest,
+  user: com.gu.identity.model.User,
+  privacyForm: Form[form.PrivacyFormData])(implicit request: RequestHeader, messages: play.api.i18n.Messages)
+
+@import views.html.fragments.form.{checkboxField}
+@import form.IdFormHelpers._
+@import helper._
+
+
+<form class="form js-public-profile-form" novalidate action="@idUrlBuilder.buildUrl("/privacy/edit", idRequest)" role="main" method="post">
+    @views.html.helper.CSRF.formField
+
+    @if(privacyForm.globalError.isDefined) {
+        <div class="form__error">@privacyForm.globalErrors.map(_.message).mkString(", ")</div>
+    }
+
+    <fieldset class="fieldset">
+
+        <div class="fieldset__heading">
+            <h2 class="form__heading">Marketing</h2>
+        </div>
+
+        <div class="fieldset__fields">
+            <ul class="u-unstyled">
+                <li class="form-field">
+                    <label class="label">Would you like to receive information from the Guardian and their partners?</label>
+                    <div class="form-field__note">The Guardian and their partners would like to occasionally send you information about their products, services and events.</div>
+
+                    <div class="form-fields-group">
+                        @checkboxField(Checkbox(privacyForm("receiveGnmMarketing"), '_label -> "Receive email from Guardian News and Media Ltd."))(nonInputFields, messages)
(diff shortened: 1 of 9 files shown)


paperboyo:
:gem:! Ela, Fergal and @johnduffell had a great hackday idea around that - well worth talking to them!

danmassie:
@paperboyo Chris Wilk is the man to speak to. The identity team are about to implement a few features in this area.

paperboyo:
Thanks, @danmassie. I'm just a matchmaker. And a blind one, too. Also, true privacy = best marketing, I think.

johnduffell:
privacy and security are all our favourite things :+1: although our hackday project was a little more ambitious than could be done in a PR

paperboyo:
What about marketing?! :wink:

No-no, of course, I just got excited and wanted to mention your work as always when this topic comes up, because I think it’s The Thing™. As always, unacceptably, I hijacked someone’s PR to do so. Sorry, @danmassie! :sweet_potato:

identity/app/views/fragments/profile/privacyForm.scala.html
```diff
@@ -0,0 +1,43 @@
...
+  idRequest: services.IdentityRequest,
+  user: com.gu.identity.model.User,
+  privacyForm: Form[form.PrivacyFormData])(implicit request: RequestHeader, messages: play.api.i18n.Messages)
+
+@import views.html.fragments.form.{checkboxField}
+@import form.IdFormHelpers._
+@import helper._
+
+
+<form class="form js-public-profile-form" novalidate action="@idUrlBuilder.buildUrl("/privacy/edit", idRequest)" role="main" method="post">
```
jamespamplin: Do we have to build the URL in the view directly, or can we build it in the Controller (or elsewhere) instead?

identity/test/controllers/EditProfileControllerTest.scala
```diff
@@ -88,6 +88,46 @@ class EditProfileControllerTest extends path.FreeSpec with ShouldMatchers with M
...
+          receiveGnmMarketing = Some(receiveGnmMarketing),
+          receive3rdPartyMarketing = Some(receive3rdPartyMarketing)
+        )
+      )
+      when(api.saveUser(Matchers.any[String], Matchers.any[UserUpdate], Matchers.any[Auth]))
+        .thenReturn(Future.successful(Right(updatedUser)))
+
+      val result = controller.submitPrivacyForm().apply(fakeRequest)
+
+      Await.result(result, 10.seconds)
```
jamespamplin: Even though this probably wont fail, 10 seconds is quite long to wait, could this be just 1 second?

danmassie: It is a long time. Its a copy and paste of the other other submit tests.

identity/app/views/fragments/profile/privacyForm.scala.html
danmassie: We should build it elsewhere , but I'm not refactoring all of the views as part of this. At the moment this copies the pattern used elsewhere.

jamespamplin: Ok I understand. Keep it consistent for now, we can fix it properly later. Thanks.

danmassie:
@jamespamplin timeout reduced

jamespamplin:
:+1: looks good to me

PR #11569, opened 2016-01-14
Adjust page press timings

JustinPinner:
We can't pull more than 10 messages from an SQS queue at once (according to Amazon docs). Reduce this limit and increase the duty cycle to compensate.

cc @jennysivapalan


jennysivapalan:
10?! Rubbish. :+1:

PR #11570, opened 2016-01-14
Manual multiple soulmates

regiskuckaertz:
This change will allow manual multiple to use the same styling as the soulmates merchandising component:

manual-multiple-soulmates

Sadly the left-hand side has a different **content** than all the other styles so there had to be some forking in the template.


uplne:
:+1:

PR #11571, opened 2016-01-14
Top banner test extension

uplne:
This is an extension of the already (not) running server side AB test https://github.com/guardian/frontend/pull/11137 (but will after this) which should test to move top banner below the first container for desktop. It has some UX advantages. Banner animation won't be that visible anymore and possibly the banner can be loaded later (lazy loading).

However it is also moving banner back on mobile similarly below the first container.


static/src/javascripts/projects/common/modules/commercial/top-banner-below-container.js
```diff
@@ -13,24 +13,35 @@ define([
...
-        }
+        // For mobile we want to add inline1 MPU after the first container
+        if (detect.getBreakpoint() === 'mobile' && config.tests.topBannerPosition) {
+            $adSlot = bonzo(createAdSlot('inline1', 'container-inline'));
+
+            if ($containers.length > 0) {
+                return $adSlot.insertAfter($containers[containerIndex]);
+            }
+        // For desktop we want to place topAboveNav after first container
+        } else {
```
regiskuckaertz: This else statement will be fulfilled if `breakpoint !== 'desktop'` **or** `breakpoint === 'mobile' && !config.tests.topBannerPosition`. Are you sure the latter is correct? I wonder if you can't remove the test on `topBannerPosition` since it's already been tested at the beginning of the function.

regiskuckaertz:
:+1:

PR #11572, opened 2016-01-15
Discussion omniture tracking

gidsg:
- Add `data-link-name` to previously untracked discussion events
- Disable recommend button if discussion is closed for recommendations

CC @chriswilk @guardian/discussion

### DIFF
--- a/discussion/app/views/fragments/comment.scala.html
+++ b/discussion/app/views/fragments/comment.scala.html
@@ -70,7 +70,7 @@
                     data-comment-id="@comment.id"
                     data-user-id="@comment.profile.userId"
                     data-recommend-count="@comment.numRecommends"
-                    title="@comment.numRecommends recommendations">
+                    title="@comment.numRecommends recommendations" data-link-name="Recommend comment">
                     <button class="u-button-reset d-comment__recommend-button rounded-icon">
                         <span class="d-comment__recommend-pulse"></span>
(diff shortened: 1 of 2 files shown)


discussion/app/views/fragments/comment.scala.html
```diff
@@ -43,7 +43,7 @@
                     </a>
                 }
                 <div class="d-comment__timestamp">
-                    <a href="@Configuration.discussion.url/comment-permalink/@comment.id" class="d-comment__timestamp block-time">
+                    <a href="@Configuration.discussion.url/comment-permalink/@comment.id" class="d-comment__timestamp block-time" data-link-name="Comment permalink">
```
gidsg: @rich-nguyen Is it possible to track this, as it is an outbound request, the tracking call fails to complete as they are leaving the site

rich-nguyen: I'd check with the data team, but we have enabled outbound tracking in omniture.

rich-nguyen:
:+1:

discussion/app/views/fragments/comment.scala.html
gidsg: Haha, data team said ask @rich-nguyen (!) but I am minded to remove it.

rich-nguyen: I think removing it makes sense.

PR #11575, opened 2016-01-15
Add AB tests for related content variants

marialivia16:
Three new variants have been added to CAPI which can be accessed using the `related-type` parameter:
- `tags-only`: using tags to find related articles;
- `tags-headline`: using tags and headlines to find related articles;
- `in-body-links`: using the inline body links as related and, if that is not returning 10 articles, the results will be backfilled by tags related content.

This PR has been implemented in 2 steps:
- Revert the following PR https://github.com/guardian/frontend/pull/11334
- Add AB tests for the 3 new variants.

@johnduffell @stephanfowler


common/app/conf/switches/ABTestSwitches.scala
```diff
@@ -4,10 +4,10 @@ import org.joda.time.LocalDate
...
 
-  val AlternativeRelated = Switch(
+  val RelatedVariants = Switch(
     "A/B Tests",
-    "ab-alternative-related",
-    "show alternative related content based on the tags to users in the test",
+    "ab-related-variants",
+    "show related content based on the new variants",
     safeState = Off,
     sellByDate = new LocalDate(2016, 1, 25),
```
johnduffell: probably worth making the switch longer

onward/app/controllers/RelatedController.scala
```diff
@@ -25,32 +25,20 @@ object RelatedController extends Controller with Related with Containers with Lo
   def render(path: String, isMf2: Boolean = false) = MemcachedAction { implicit request =>
     val edition = Edition(request)
     val excludeTags = request.queryString.getOrElse("exclude-tag", Nil)
-    related(edition, path, excludeTags) map {
+    val relatedType = request.queryString.getOrElse("related-type", Nil)
```
johnduffell: you could put .headOption on the end of this and treat it as an Option[String] afterwards (which is basically what it is in reality)

onward/app/services/repositories.scala
```diff
@@ -22,10 +21,14 @@ trait Related extends ConciergeRepository {
...
 
-      val response = getResponse(LiveContentApi.item(path, edition)
+      val requestBuilder = LiveContentApi.item(path, edition)
         .tag(tags)
         .showRelated(true)
-      )
+
+      val response = relatedType match {
+        case Nil => getResponse(requestBuilder)
+        case someType => getResponse(requestBuilder.stringParam("related-type", someType.head))
```
johnduffell: here if you are using an Option[String] instead you could case None => and case Some(type) => to avoid having to do a .head later.

johnduffell:
looks great :+1: but I think if this is to be a rigorous test rather than just getting it out there for people to play with, it's worth sitting down with the data team to go through it and make sure the results will be usable

marialivia16:
@johnduffell Thanks! We should set up a meeting with the data team to discuss the results of the variants and set up the sample sizes.

stephanfowler:
:+1: great! Just get it out as it is, using 0% audience, so we can give it an internal user test first.

PR #11576, opened 2016-01-15
Amp ads right aligned instead of left

NataliaLKB:
Aligning Ads in AMP right, just the website.

Before:
image

After
image

cc @zeftilldeath @stephanfowler


stephanfowler:
:+1:

PR #11578, opened 2016-01-15
Manual multiple soulmates edit

regiskuckaertz:
Move test logic into the JS to avoid polluting the template

/cc @uplne


uplne:
:+1:

PR #11582, opened 2016-01-15
Ad calls from AMP now include Krux audience segments

kenlim:


NataliaLKB:
:+1:

amp/remote-3.html
```diff
@@ -0,0 +1,67 @@
...
+
+            var fragments = url.split('adtest=');
+            if (fragments.length === 2) {
+                config.targeting.at = fragments[1]
+            }
+            return config;
+        }
+
+        function attachKruxSegments(config) {
+            if (localStorage.getItem('kxsegs') !== null) {
```
stephanfowler: You should first test for the availability of localStorage. Eg in frontend we do it like this:
https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/common/utils/storage.js#L28

PR #11584, opened 2016-01-15
Sticky ad test now works with framebuster ads

holder-josh:
Plus a whole load of fixes.

Still a 0% test

cc @OliverJAsh

### DIFF
--- a/static/src/javascripts/projects/common/modules/navigation/newSticky.js
+++ b/static/src/javascripts/projects/common/modules/navigation/newSticky.js
@@ -79,11 +79,11 @@ define([
                             $header.css({ 'margin-top': state.adHeight });
 
                             $adBanner.css({ 'max-height': state.adHeight });
-                            var userHasScrolledPastHeader = scrollY > headerHeight;
+                            var userHasScrolledPastHeader = scrollY > state.headerHeight;
                             if (userHasScrolledPastHeader) {
                                 $adBanner.css({
                                     'position': 'absolute',
-                                    'top': headerHeight
+                                    'top': state.headerHeight
                                 });
                             } else {
                                 $adBanner.css({ 'position': 'fixed' });
@@ -112,7 +112,8 @@ define([
                             return render({
                                 adHeight: adHeight,
                                 previousAdHeight: previousAdHeight || adHeight,
-                                firstRender: options.firstRender || false
+                                firstRender: options.firstRender || false,
+                                headerHeight: headerHeight
                             }).then(function () {
                                 previousAdHeight = adHeight;
                             });
--- a/static/src/javascripts/projects/common/modules/navigation/newSticky.js
+++ b/static/src/javascripts/projects/common/modules/navigation/newSticky.js
@@ -59,11 +59,11 @@ define([
                 };
 
                 var render = function (state) {
-                    fastdom.read(function () {
+                    return fastdom.read(function () {
                         var scrollY = window.scrollY;
                         var scrollX = window.scrollX;
 
-                        fastdom.write(function () {
+                        return fastdom.write(function () {
                             // Reset so we have a clean slate
                             $header.css({
                                 'transition': '',
@@ -108,13 +108,14 @@ define([
                 var update = (function () {
                     var previousAdHeight;
                     return function (options) {
-                        getCachedAdHeight().then(function (adHeight) {
-                            render({
+                        return getCachedAdHeight().then(function (adHeight) {
+                            return render({
                                 adHeight: adHeight,
                                 previousAdHeight: previousAdHeight || adHeight,
                                 firstRender: options.firstRender || false
+                            }).then(function () {
+                                previousAdHeight = adHeight;
                             });
-                            previousAdHeight = adHeight;
                         });
                     };
                 })();
@@ -127,10 +128,11 @@ define([
                     // will move into stylesheets when productionised
                     $adBanner.css({'overflow': 'hidden', 'transition': 'max-height 1s cubic-bezier(0, 0, 0, 0.985)'});
 
-                    update({ firstRender: true });
-                    window.addEventListener('scroll', function () { update({}); });
-                    newAdHeightPromise
-                        .then(function () { update({}); });
+                    update({ firstRender: true }).then(function () {
+                        window.addEventListener('scroll', function () { update({}); });
+                        newAdHeightPromise
+                            .then(function () { update({}); });
+                    });
                 });
             });
         }
--- a/static/src/javascripts/projects/common/modules/navigation/newSticky.js
+++ b/static/src/javascripts/projects/common/modules/navigation/newSticky.js
@@ -3,16 +3,18 @@ define([
     'Promise',
     'common/utils/$',
     'common/utils/mediator',
+    'common/utils/config',
     'common/utils/detect'
 ], function (
     fastdom,
     Promise,
     $,
     mediator,
+    config,
     detect) {
 
     return function () {
-        if (detect.getBreakpoint() !== 'mobile') {
+        if (detect.getBreakpoint() !== 'mobile' && config.page.contentType !== 'Interactive') {
             fastdom.read(function () {
                 var adId = 'dfp-ad--top-above-nav',
                     $adBanner = $('.js-top-banner-above-nav'),
--- a/static/src/javascripts/projects/common/modules/navigation/newSticky.js
+++ b/static/src/javascripts/projects/common/modules/navigation/newSticky.js
@@ -126,6 +126,7 @@ define([
 
                 fastdom.write(function () {
                     // will move into stylesheets when productionised
+                    $(document.body).addClass('new-sticky-ad');
                     $adBanner.css({'overflow': 'hidden', 'transition': 'max-height 1s cubic-bezier(0, 0, 0, 0.985)'});
 
                     update({ firstRender: true }).then(function () {
--- a/static/src/javascripts/projects/common/modules/navigation/newSticky.js
+++ b/static/src/javascripts/projects/common/modules/navigation/newSticky.js
@@ -122,7 +122,7 @@ define([
                 // Side effects
                 //
 
-                fastdom.write(function() {
+                fastdom.write(function () {
                     // will move into stylesheets when productionised
                     $adBanner.css({'overflow': 'hidden', 'transition': 'max-height 1s cubic-bezier(0, 0, 0, 0.985)'});
 
--- a/static/src/javascripts/projects/common/modules/navigation/newSticky.js
+++ b/static/src/javascripts/projects/common/modules/navigation/newSticky.js
@@ -16,8 +16,8 @@ define([
             fastdom.read(function () {
                 var adId = 'dfp-ad--top-above-nav',
                     $adBanner = $('.js-top-banner-above-nav'),
+                    $adBannerInner = $('#' + adId, $adBanner),
                     $header = $('.js-header'),
-                    oldAdHeight = $adBanner.height(),
                     headerHeight = $header.height();
 
                 var topAdRenderedPromise = new Promise(function (resolve) {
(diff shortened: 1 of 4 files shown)


static/src/javascripts/projects/common/modules/navigation/newSticky.js
```diff
@@ -28,68 +28,109 @@ define([
...
                 //
                 // Side effects
                 //
 
-                mediator.on('window:throttledScroll', render);
-                render({ firstRender: true });
-                topAdRenderedPromise.then(render);
+                fastdom.write(function () {
+                    // will move into stylesheets when productionised
+                    $adBanner.css({'overflow': 'hidden', 'transition': 'max-height 1s cubic-bezier(0, 0, 0, 0.985)'});
```
OliverJAsh: Why do we need `overflow: hidden`? I've forgotten.

static/src/javascripts/projects/common/modules/navigation/newSticky.js
```diff
@@ -28,68 +28,109 @@ define([
...
-
-                                previousAdHeight = adHeight;
+                var getLatestAdHeight = function () {
+                    var $iframe = $('iframe', $adBanner);
+                    var slotWidth = $iframe.attr('width');
+                    var slotHeight = $iframe.attr('height');
+                    // iframe may not have been injected at this point
+                    var isFluidAd = $iframe.length > 0 && [slotWidth, slotHeight].join(',') === '88,70';
+                    // fluid ads are currently always 250px high
+                    var fluidAdInnerHeight = 250;
```
OliverJAsh: I think this needs a comment to say why we can't just calculate the client height. We can't calculate the client height because fluid ads are rendered asynchronously.

static/src/javascripts/projects/common/modules/navigation/newSticky.js
Jholder112233: Otherwise the top ad overlaps the header before the top banner slot expands

OliverJAsh: I think it's worth adding a comment for that

Jholder112233: Will do once it's production-ready and can be moved into the stylesheet

OliverJAsh:
Looks good to me but let's get someone else to review our work! :smile:

static/src/javascripts/projects/common/modules/navigation/newSticky.js
```diff
@@ -28,68 +28,112 @@ define([
...
+                            var diff = state.adHeight - state.previousAdHeight;
+                            var adHeightHasIncreased = diff > 0;
+                            if (!scrollIsAtTop && adHeightHasIncreased) {
+                                // If the user is not at the top and the ad height has increased,
+                                // we want to offset their scroll position
+                                window.scrollTo(scrollX, scrollY + diff);
+                            } else if (!state.firstRender) {
+                                // Otherwise we want to transition the change when it happens.
+                                // Avoid an initial transition when we apply the margin top for the first time
+                                $header.css('transition', 'margin-top 1s cubic-bezier(0, 0, 0, 0.985)');
```
sndrs: does this need vendor prefixing?

static/src/javascripts/projects/common/modules/navigation/newSticky.js
```diff
@@ -28,68 +28,112 @@ define([
...
-                                previousAdHeight = adHeight;
+                    return isFluidAd
+                        ? Promise.resolve(fluidAdHeight)
+                        : getClientAdHeight();
+                };
+
+                var oldAdHeightPromise = getLatestAdHeight();
+                var newAdHeightPromise = topAdRenderedPromise.then(getLatestAdHeight);
+
+                var getCachedAdHeight = function () {
```
sndrs: i feel funny about calling promises `xPromise`, like I would about calling a string `xString`. either way, `getCachedAdHeight` returns a promise and isn't names like that. my feeling is that the former is overkill, but we should be consistent?

static/src/javascripts/projects/common/modules/navigation/newSticky.js
```diff
@@ -28,68 +28,112 @@ define([
...
+                    return Promise.race([newAdHeightPromise, oldAdHeightPromise]);
```
OliverJAsh: We should check `.race` is in the polyfill for `Promise`

static/src/javascripts/projects/common/modules/navigation/newSticky.js
```diff
@@ -28,68 +28,112 @@ define([
...
+                            }
                         });
                     });
                 };
 
+                var update = (function () {
+                    var previousAdHeight;
+                    return function (options) {
+                        return getCachedAdHeight().then(function (adHeight) {
+                            return render({
```
sndrs: does render not return an instance of `fastdom`? is it thenable? i'm guessing it must be...?

static/src/javascripts/projects/common/modules/navigation/newSticky.js
sndrs: should stack these for readability

static/src/stylesheets/module/_adslot.scss
```diff
@@ -178,8 +178,13 @@
     .top-banner-ad-container--desktop & {
         margin: 0 auto;
         display: none;
-        min-height: 90 + ($gs-row-height / 2);
+        min-height: 90px;
         padding-bottom: $gs-row-height / 2;
+        padding-top: $gs-row-height / 2;
+
+        .ad-slot__label {
+            margin-top: -$gs-row-height / 2;
```
sndrs: i'd tend to opt for `0 - $gs-row-height` (or similar) for legibility, mainly for times like this, where there are a lot of hyphens in the name and calcs with it 

i.e. spaces around operators, since `$my-value+$other-value - $another-value*$final-value` is valid but not nec clear

sndrs:
:+1: apart from comments - nice work!

static/src/stylesheets/module/_adslot.scss
```diff
@@ -119,7 +120,6 @@
 }
 .top-banner-ad-container--reveal iframe {
     transform: translateZ(0);
-    transition: height 1s cubic-bezier(0, 0, 0, .985);
```
OliverJAsh: Isn't this still needed for the old one?

static/src/javascripts/projects/common/modules/navigation/newSticky.js
```diff
@@ -28,68 +28,112 @@ define([
...
+
+                            // Set
+                            $header.css({ 'margin-top': state.adHeight });
+
+                            $adBanner.css({ 'max-height': state.adHeight });
+                            var userHasScrolledPastHeader = scrollY > headerHeight;
+                            if (userHasScrolledPastHeader) {
+                                $adBanner.css({
+                                    'position': 'absolute',
+                                    'top': headerHeight
```
OliverJAsh: Here and L82 should be `state.headerHeight`

static/src/stylesheets/module/_adslot.scss
```diff
@@ -119,7 +120,6 @@
 }
 .top-banner-ad-container--reveal iframe {
     transform: translateZ(0);
```
sndrs: vendor prefix?

static/src/stylesheets/module/_adslot.scss
```diff
@@ -71,6 +71,7 @@
         width: 100%;
         transform: translateZ(0);
```
sndrs: vendor prefix?

static/src/javascripts/projects/common/modules/navigation/newSticky.js
```diff
@@ -28,68 +28,112 @@ define([
...
+
+                var render = function (state) {
+                    return fastdom.read(function () {
+                        var scrollY = window.scrollY;
+                        var scrollX = window.scrollX;
+
+                        return fastdom.write(function () {
+                            // Reset so we have a clean slate
+                            $header.css({
+                                'transition': '',
```
sndrs: vendor prefix?

static/src/javascripts/projects/common/modules/navigation/newSticky.js
sndrs: vendor prefix?

static/src/javascripts/projects/common/modules/navigation/newSticky.js
```diff
@@ -28,68 +28,112 @@ define([
...
 
-                mediator.on('window:throttledScroll', render);
-                render({ firstRender: true });
-                topAdRenderedPromise.then(render);
+                fastdom.write(function () {
+                    // will move into stylesheets when productionised
+                    $adBanner.css({'overflow': 'hidden', 'transition': 'max-height 1s cubic-bezier(0, 0, 0, 0.985)'});
+
+                    update({ firstRender: true }).then(function () {
+                        window.addEventListener('scroll', function () { update({}); });
```
OliverJAsh: mediator `throttledScroll`? I'm impartial because we're already using fastdom and inside `update`

static/src/javascripts/projects/common/modules/navigation/newSticky.js
OliverJAsh: This is an interesting thing to discuss. Imagine you have a promise for a user. If we called this `user`, you would either have to shadow the identifier in the `then` callback, or call it something else:

``` js
user.then(user => {
  // do something with user
});
```

I want to avoid doing this because shadowing hurts readability, so the `<identifier>Promise` is just one way of doing that. Strings aren't wrapped values, they are just values themselves, so they don't have this problem.

We have the same problem with futures and options in Scala, and it is common convention to prefix identifiers with their wrapped type, i.e. `maybeUser` for options or `futureUser` for futures.

--

In the example above, the value of `getLatestAdHeight` is a function which returns a promise. I think we only need the above convention when dealing with promise values themselves.

static/src/javascripts/projects/common/modules/navigation/newSticky.js
OliverJAsh: We are using a promise wrapped version of fastdom. `render` returns a promise.

static/src/stylesheets/module/_adslot.scss
OliverJAsh: Don't we have autoprefixer?

static/src/stylesheets/module/_adslot.scss
OliverJAsh: :+1: 

You can also multiple a value by -1 to get the negative equivalent

static/src/javascripts/projects/common/modules/navigation/newSticky.js
sndrs: i wonder if it's more immediately comprehensible to name things by their intent or role than what they literally are (which is what i think is happening in the scala convention)? 

in your example, you'll probably be doing either `getUser` or `createUser` or some such. you only end up shadowing `user` because the promise was badly named (i agree). 

the first instance of `user` there is not interesting because it's a `Promise` (specifically), but because it hasn't yet happened.

if i invented an amazing thing called a `Probably`, and we replaced `Promise`, `getUser` and `createUser` would still be meaningful because they only capture the behaviour, not the implementation

OliverJAsh: What would you suggest instead of the following?

``` js
var userPromise = getUser();
userPromise.then(user => console.log(user));
```

sndrs:
```javascript
getUser().then(user => console.log(user));
```
:smile: 

don't mean to be facetious – are there times that it's not as straightforward as that?

sndrs: taking to slack...

static/src/stylesheets/module/_adslot.scss
sndrs: doh! yeah not needed in css, but will be in JS

PR #11592, opened 2016-01-18
Share collections

piuccio:
hides the implementation details of collection backfill inside facia scala client

@janua

### DIFF
--- a/facia-press/app/frontpress/FapiFrontPress.scala
+++ b/facia-press/app/frontpress/FapiFrontPress.scala
@@ -124,15 +124,8 @@ trait FapiFrontPress extends QueryDefaults with Logging with ExecutionContexts {
   }
 
   private def getBackfill(collection: Collection): Response[List[PressedContent]] = {
-    collection
-      .collectionConfig
-      .apiQuery
-      .map { query =>
-        FAPI.backfill(query, collection, searchApiQuery, itemApiQuery)
-      }
-      .getOrElse {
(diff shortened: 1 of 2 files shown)


facia-press/app/frontpress/FapiFrontPress.scala
```diff
@@ -124,15 +124,8 @@ trait FapiFrontPress extends QueryDefaults with Logging with ExecutionContexts {
...
-    collection
-      .collectionConfig
-      .apiQuery
-      .map { query =>
-        FAPI.backfill(query, collection, searchApiQuery, itemApiQuery)
-      }
-      .getOrElse {
-        Response.Right(Nil)
-      }.map(_.map(PressedContent.make))
+    FAPI.backfillFromConfig(collection, searchApiQuery, itemApiQuery)
```
rich-nguyen: why does `backfillFromConfig` take a `Collection`, and not the config in the collection? Should it be called `backfillFromCollection`?

rich-nguyen:
:+1:

facia-press/app/frontpress/FapiFrontPress.scala
janua: You should leave this name but use `collection.collectionConfig` instead; backfill definitely comes from config. You will have to change `facia-scala-client` and release it again.

piuccio:
changes made, it now takes a collection config.

PR #11593, opened 2016-01-18
Roll out pagination v1 to all liveblogs

johnduffell:
This seems to have worked well on business live and politics live so it's time to roll to all.

Sport needs a longer page size - it will go up to 69 before it splits off a page of 10 whereas others will only get to 29.

I've also fixed a little bug with the "newer updates" link.


stephanfowler:
:+1: - think you mean: sport gets 1st page up to 119, then pages of 50?

PR #11597, opened 2016-01-18
Manual blended component

regiskuckaertz:
People from the merch team now have the ability to select particular items to show up in the blended component, with the exception of travels and soulmate profiles.

/with @kelvin-chappell


kelvin-chappell:
:+1: 
Will need to monitor performance of commercial servers in case book lookup has some effect.

PR #11598, opened 2016-01-18
Rtrt email interactives event

gtrufitt:
@sammorrisdesign This is the counterpart to the mods I've made on the interactive (which I can't push right now...).

### DIFF
--- a/static/src/stylesheets/module/email/_header.scss
+++ b/static/src/stylesheets/module/email/_header.scss
@@ -62,3 +62,15 @@
     }
 }
 
+// PLAIN MODS
+.email-sub--plain {
+    .email-sub__message__headline {
+        color: guss-colour(guardian-brand, $pasteup-palette);
+        line-height: 2rem;
+    }
+
+    .email-sub__message__description {
+        display: none;
+    }
+}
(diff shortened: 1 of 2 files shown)


static/src/stylesheets/module/email/_header.scss
```diff
@@ -62,3 +62,15 @@
     }
 }
 
+// PLAIN MODS
```
johnduffell: what's a plain mod?

gtrufitt: The plain version of the component style modification (Ie: BEM modifier). Comment probably makes more sense in context when you can see there are a different modifications: https://github.com/guardian/frontend/blob/RTRT_Email-interactives-event/static/src/stylesheets/module/email/_header.scss

johnduffell:
:+1:

static/src/stylesheets/module/email/_header.scss
johnduffell: I'm still confused but I'm not 100% up on BEM terminology, maybe run it by someone else to confirm it makes sense!

PR #11599, opened 2016-01-18
Top banner position test with cm prefix

uplne:
This PR is solving multiple things:
1. prefixing topBannerPosition test with cm- prefix. We are sending only ab tests variables to DPF with this prefix
2. Extending test for another month


regiskuckaertz:
:+1:

PR #11596, opened 2016-01-18
Add A/B test data to the feedback email

jbreckmckye:
This pull request adds details of the user's running A/B tests to the technical feedback form. We'd like to do this because we're about to trial a new way of loading adverts, and we'd be keen to spot any user issues arising of the back of it.

A/B test data is added to the email in the format of `testName=value, testName=value`. It includes tests the user is not participating in:
image

In the instance the user has no test data at all - because no tests are running, or because localStorage is unavailable - the email summarises the data as 'No tests running':
image

@ScottPainterGNM

PR #11601, opened 2016-01-18
A/B test header-bidding for advertising

jbreckmckye:
**What is this?**

The commercial team is interested in a new approach to loading adverts termed 'header-bidding'. Normally, we load adverts by making a request to DFP that returns the creative. Under a header-bidding system, we first run an auction - making requests to multiple third parties which return how much they're willing to pay for the slot - before sending the winner's details up to DFP. This approach improves revenue, but it's also a lot slower, and that's why we'd like to test this carefully.

**A/B test**

In this A/B test, we're performing a header-bidding auction with [prebid.js](https://github.com/prebid/Prebid.js) before loading each ad slot. We're not actually sending the winner's details up to DFP - we're just interested in the performance consequences in doing it, both for the adverts (because slow adverts have low 'viewability' figures), and for the page as a whole.

The main ways we'll be looking for issues will be -
- sending A/B cohort data up to Moat, which is the tool that tracks ad viewability
- watching performance and engagement figures in Ophan (e.g. under https://dashboard.ophan.co.uk/perf/timings?ab=PrebidPerformance:variant)
- catching errors users themselves report now that [we've added A/B data to the feedback form](https://github.com/guardian/frontend/pull/11596)

The test has the key `prebid-performance`. The cohort size is 2%. Like all A/B tests, it has a switch with a safe-state set to false.

**What is the technical change?**

I have added a temporary service named `prebid-service` that tries to encapsulate most of the test's implementation. When its `testEnabled` flag is true, `dfp-api` will asynchronously load the header-bidding implementation script, `prebid.js`, and route all slot loads through the prebid service. This will run a header bidding auction and await its resolution before making the normal DFP call.

A couple of ad-slot types are excluded from the test, because they must always be loaded ASAP:
- **pageskin-inreads** are a kind of video advert that open up as the user scrolls them into view. They have to be ready to play as soon as that happens.
- **merchandising-high** slots contain Outbrain widgets. We don't want to do anything to jeopardise Outbrain impressions.


static/src/javascripts/projects/common/modules/commercial/prebid-test/prebid-service.js
```diff
@@ -0,0 +1,133 @@
...
+        };
+
+        this.slotIsInTest = function (slot) {
+            return ['dfp-ad--pageskin-inread', 'dfp-ad--merchandising-high'].indexOf(slot) === -1;
+        };
+
+        var pendingAdverts = new Promise.resolve(true);
+
+        this.loadSlots = function (slotIds) {
+            slotIds = Array.isArray(slotIds) ? slotIds : [slotIds];
```
regiskuckaertz: just in case, `Array.prototype.isArray` does not exist in IE8

jbreckmckye: That shouldn't even be there. Good spot!

static/src/javascripts/projects/common/modules/commercial/prebid-test/prebid-service.js
```diff
@@ -0,0 +1,133 @@
...
+                    referrer : 'http://www.theguardian.com/uk'
+                }
+            };
+        });
+    }
+
+    function getSlotAdSizes(slotElementId) {
+        var $slot = bonzo(document.getElementById(slotElementId));
+
+        var validBreakpoints = detect.breakpoints.filter(function (breakpoint) {
```
regiskuckaertz: again, for IE8 users, it might be preferable to use lodash functions instead.

jbreckmckye: Yes, I'd meant to do that. Will convert them.

static/src/javascripts/projects/common/modules/commercial/dfp-api.js
```diff
@@ -290,8 +292,12 @@ define([
             googletag.enableServices();
             // as this is an single request call, only need to make a single display call (to the first ad
             // slot)
-            googletag.display(keys(slots).shift());
-            displayed = true;
+            if (prebidService.testEnabled) {
```
regiskuckaertz: Shouldn't this rather be `if (prebidService.testEnabled && prebidService.slotIsInTest(slotKey))` to avoid `loadSlot` to fail when trying to access a non-existing `slots` array?

jbreckmckye: We should apply the guard everywhere, yes.

static/src/javascripts/projects/common/modules/commercial/dfp-api.js
```diff
@@ -290,8 +292,13 @@ define([
             googletag.enableServices();
             // as this is an single request call, only need to make a single display call (to the first ad
             // slot)
-            googletag.display(keys(slots).shift());
-            displayed = true;
+            var firstSlot = keys(slots).shift();
```
regiskuckaertz: In fact, this whole block could simply become `loadSlot(keys(slots).shift())`

regiskuckaertz:
:+1:

static/src/javascripts/projects/common/modules/commercial/prebid/prebid-service.js
```diff
@@ -0,0 +1,130 @@
...
+
+        slotPlacements = filter(slotPlacements, function matchViewport(placement) {
+            return detect.isBreakpoint(placement.breakpoint);
+        });
+
+        slotPlacements = filter(slotPlacements, function matchSlotSize(placement) {
+            return find(slotSizes, function (slotSize) {
+                return slotSize[0] === placement.width && slotSize[1] === placement.height;
+            });
+        });
```
uplne: Nice. Very readable.

uplne:
Looks good. Nicely crafted. :+1:

static/src/javascripts/projects/common/modules/commercial/prebid/prebid-service.js
```diff
@@ -0,0 +1,130 @@
...
+            var dimensions = sizeString.split(',');
+            return map(dimensions, function (widthOrHeightString) {
+                return parseInt(widthOrHeightString, 10);
+            });
+        });
+    }
+
+    return new PrebidTestService();
+
+});
```
OliverJAsh: This is clean :-)

Do you need classes (constructors) at all? You're deliberately avoiding keeping state on the class object, which is great, but it's one small step for someone to change that. (See `save-for-later.js`.)

Could this instead be done just using objects and functions? E.g.

``` js
var prebidTestService = {
  loadDependencies: function () {
    require(['js!prebid.js']);
  }
}
```

Or even just:

``` js
var loadPreBidTestDependencies = function () {
  require(['js!prebid.js']);
}
```

I think this approach might be safer in terms of future maintainability, but it also uses less features of the language, which I think makes this simpler.

static/src/javascripts/projects/common/modules/commercial/prebid/prebid-service.js
OliverJAsh: Minor comment but you could use `utils/chain.js` here instead of reassigning:

``` js
chain(placements)
    .and(filter, () => {})
    .and(filter, () => {})
```

OliverJAsh:
We should also measure how ethical this is. It may perform smoothly (because it's not in the critical path), but if it adds a lot to our page weight or keeps the radio alive for a long time (bad for the battery), I would question whether this is tolerable.

Interested to hear the results of this test!

jbreckmckye:
@OliverJAsh I absolutely agree that there are big questions about power consumption. There is a Battery Status API supported by Chrome for Android - it would be really interesting to log its data, but the data would be very noisy, even if it could be recorded with high precision - we wouldn't be able to differentiate battery drain due to theguardian.com from drain due to another tab, drain due to a background Spotify process, and so on.

Have we ever thought about measuring power usage before? I would be interested to know if we'd had any ideas about this.

static/src/javascripts/projects/common/modules/commercial/prebid/prebid-service.js
jbreckmckye: Good point.

static/src/javascripts/projects/common/modules/commercial/prebid/prebid-service.js
jbreckmckye: I did consider that. My thinking was that with a constructor -

- it was easy to write a stateless module, which in turn made things a bit nicer to test. It's true you can break this quite easily (as per `save-for-later`), but it's also relatively easy to fix this by moving initialization into the constructor
- you don't have to add an `init` method. You don't need a third party bootstrap script to initialize your dependency, and when you're looking at the module itself, you don't have to worry that someone might be calling `init` a second time innocently. If you want a clean module, you have to call `new service.Constructor` - in which case, it's obvious that you're doing something unusual, and you aren't going to wipe out the state some other module is depending on.
- it's a minor stylistic point, but because your 'class' definition is in a function, you can describe the public interface at the top of a file, even though your return statement is at the very bottom. But then again, it's hardly arduous to pageDown a file. You also get to declare your 'type' in a function rather than an object literal, so you don't have to mess about with joining commas. But that is fairly trivial.

I do have my own biases, though - I worked at a Java shop for many years, and so several of the conventions I picked up were chosen to make scripts look a lot more Java-esque. Conventions like always having an `init` method might just be a house style I need to embrace.

PR #11602, opened 2016-01-18
Update email article AB test with keyword blacklist

gtrufitt:
Avoid showing the Guardian Today email form on NHS pages to avoid clash with their own email campaign


OliverJAsh:
:+1:

PR #11603, opened 2016-01-18
Story packages amp

NataliaLKB:
Adding Series and Story Packages to AMP.

Logic:
- If there is a story package then use it
- if there is a series then use it
- if neither a story package or a series exist, then use related content

For all of these only render 4 trails.

This PR also includes putting onward journeys to all AMP pages because it is no longer an experiment

cc @johnduffell @stephanfowler


onward/conf/routes
```diff
@@ -62,6 +62,7 @@ GET        /gallery/most-viewed                       controllers.MostViewedGall
 # Experimental
 GET        /cards/opengraph/*path.json                controllers.CardController.opengraph(path)
 GET        /tagged.json                               controllers.TaggedContentController.renderJson(tag: String)
+GET        /series/mf2/*path.json                     controllers.SeriesController.renderMf2SeriesStories(path)
```
johnduffell: not sure if it's worth putting the mf2 at the end because it's only the format that's changing not the content?  e.g. /series/*path.mf2.json or something similar

johnduffell:
looks good! :+1:

stephanfowler:
:+1: nice!

PR #11604, opened 2016-01-19
Add guard to rtrt ab test

rich-nguyen:
This test is causing issues on profile pages (these pages don't have a `keywords` key)

### DIFF
--- a/static/src/javascripts/projects/common/modules/experiments/tests/rtrt-email-form-article-promo.js
+++ b/static/src/javascripts/projects/common/modules/experiments/tests/rtrt-email-form-article-promo.js
@@ -48,7 +48,8 @@ define([
                 version = detect.getUserAgent.version,
                 allArticleEls = $('> *', $articleBody),
                 lastFiveElsParas = every([].slice.call(allArticleEls, allArticleEls.length - 5), isParagraph),
-                pageIsBlacklisted = contains(config.page.keywords.split(','), 'NHS');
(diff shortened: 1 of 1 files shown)


OliverJAsh:
:+1: 

Meanwhile, how can we prevent issues like this from re-occurring? I make these mistakes all the time. We shouldn't have to rely entirely on manual testing to give us confidence.

This is exactly the sort of thing static typing would have solved.

If I understand the problem correctly, we have two sorts of pages—let's call them `ContentPage` and `StandalonePage` (e.g. front). Only `ContentPage` has `keywords`. So you could match on the type and pull out `keywords`. You would be forced to handle the other cases whereby there is no `keywords`, or else it won't compile.

Also, how did this break discussion? Is our swimlaning not working?

rich-nguyen:
@OliverJAsh we encode a lot of application logic through our use of javascript objects. This means it isn't clear when a required value is missing or not. I think you're suggesting that the use of an `Option`-like type would have equipped the engineer better. I agree, although we may need to consider how we implement modules like `config`. There is no safe implementation, eg. `getKeywords`, provided, probably due to the amount of boilerplate we would add.

This is all moot, however. The best approach is to accept that errors can happen, and provide a robust way to isolate exceptions. An AB test exception should not affect the identity system.

OliverJAsh:
> An AB test exception should not affect the identity system.

:+1: 

> I think you're suggesting that the use of an Option-like type would have equipped the engineer better

That's another way to do it. We could even do that without static typing.

PR #11605, opened 2016-01-19
Make the page press service switch permanent.

JustinPinner:
The page press service switch was added to protect against adverse effects on the admin box.

It has been running for a while and hasn't caused an issue yet, but we've still got a few things to press and while this shouldn't take too long, we may be dealing with straggler urls for some time yet.

We might want to turn the page pressing service off again at some point but we don't know when yet, so rather than extending the switch by a few days at a time I've made it permanent until we kill the service.


NataliaLKB:
:+1:

PR #11606, opened 2016-01-19
Catch errors from A/B tests (swimlane!)

OliverJAsh:
Errors in our A/B tests currently propagate which means any subsequent code will not run. This proposes we swimlane the A/B tests so, if they do error, subsequent code will still run.

This will avoid situations such as #11604
# Before

image
# After

image


rich-nguyen:
:+1:

mchv:
:+1: for the solution

gtrufitt:
Solid. :+1:

PR #11609, opened 2016-01-19
Remove expired bounce test

rich-nguyen:
This switch has now expired. We should still be measuring this, we need more detail about the time when the page view has registered.


JustinPinner:
Code changes look :+1: How will we make sure to revisit this in more / better detail?

rich-nguyen:
@sndrs is currently looking at ways to implement good omniture reporting without embedding the script. So we'll need to know about bounce rate as a related metric of this work

NataliaLKB:
:+1: :ship:

gklopper:
thanks Rich

PR #11610, opened 2016-01-19
Amp list overflow element

NataliaLKB:
#### How it looks:

image
#### Behaviour:

If the amp-list component is within 600px, or 75%, from the top of the screen only one list item will show. This is to ensure that there is no page jank for the user. Alongside this an overflow element will appear. When clicked the overflow element disappears and the rest of the stories show. 

If the amp-list component isn't in view, it will resize as normal and no overflow element will show.

@zeftilldeath is currently working on improving the styling :raised_hands:


zeftilldeath:
Latest:

screen shot 2016-01-19 at 16 05 27

NataliaLKB:
Looks great @zeftilldeath

stephanfowler:
:+1: like

johnduffell:
code looks great! however one tiny point on the sytling @zeftilldeath - the button seems a bit unbalanced in terms of size, it seems to dominate the size of the picture
also in future maybe we can consider changing it to say "more headlines" or "more highlights" etc

zeftilldeath:
@johnduffell Good point John have switched to medium sized button (screenshot now reflects this). Stefan had suggested that labeling too.

johnduffell:
looks better now

NataliaLKB:
@johnduffell though I agree that labelling like that would be clearer to the user, unfortunately there isn't a nice way of doing that with amp-list. The list itself doesn't know what stories are going to load before it happens, and so I would have to hard code the sections. 

It is best to leave it as is for now, as "More stories" is fine.

PR #11611, opened 2016-01-19
Upgrade scalatest to 2.2.6

rich-nguyen:
Earlier today, a test failed, but the build still deployed because the exit code was still 0, despite the error being caught and reported. I can't repro locally, but I'm hoping it's related to this recent issue in Scalatest: https://github.com/scalatest/scalatest/issues/432.


johnduffell:
maybe not, but why not give it a go :+1:

PR #11612, opened 2016-01-19
Discussion report abuse - return 200 in error cases

gidsg:
Previously if there was an error in form validation or in the server-side DAPI response we returned either a `400` or `500` response. However, those were being intercepted by the router and displaying the generic error page, rather than returning the form with the errors highlighted, so now we return a `200` regardless of the outcome of the form submission.

CC @guardian/discussion @rich-nguyen

### DIFF
--- a/discussion/app/controllers/CommentsController.scala
+++ b/discussion/app/controllers/CommentsController.scala
@@ -114,13 +114,13 @@ object CommentsController extends DiscussionController with ExecutionContexts {
     Action.async { implicit request =>
     val scGuU = request.cookies.get("SC_GU_U")
       userForm.bindFromRequest.fold(
-        formWithErrors => Future.successful(BadRequest(views.html.discussionComments.reportComment(commentId, reportAbusePage, formWithErrors))),
+        formWithErrors => Future.successful(Ok(views.html.discussionComments.reportComment(commentId, reportAbusePage, formWithErrors))),
         userData => {
           postAbuseReportToDiscussionApi(userData, scGuU).map {
             case success if success.status == 200 => NoCache(Redirect(routes.CommentsController.reportAbuseThankYou(commentId)))
-            case error => InternalServerError(views.html.discussionComments.reportComment(commentId, reportAbusePage, userForm.fill(userData), errorMessage = Some(ReportAbuseFormValidation.genericErrorMessage)))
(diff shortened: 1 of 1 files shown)


johnduffell:
it feels like the behaviour was on the whole correct before - had you ruled out tweaking the router config already?  seems like we're hitting the old story about theguardian.com being generally read only, but discussion app form POST is a special case

johnduffell:
could be as simple as removing the 400 from this line: https://github.com/guardian/platform/blob/master/provisioning/puppet/modules/frontend/templates/etc/nginx/router.conf.erb#L248

gidsg:
Yes, good suggestion @johnduffell  that would work for validation errors. But what about where DAPI returns an error, there the user has not made a `BadRequest` so a `500` would seem a more truthful response than `400`?

rich-nguyen:
I agree with @johnduffell, it really is a 400, and the client should receive a 400. It may be possible to override the global 400 handler by providing your own 400 in the context of your specific nginx location, rather than modifying the default error_page directive.

mchv:
@gidsg `500` should continue to report `500` **and** should be fixed on server side, apart from an unexpected issue you should never return a `500`. Hope that make sense

gidsg:
OK, so in the case of an error from the proxy request to DAPI, I should return a `200` and display the friendly message?

johnduffell:
I think you should return 400 for validation errors, 200 for success, and 500 for something that went wrong unexpectedly.
The remaining question is, since at the moment we throw a friendly error page from nginx for 400 backend errors, should we remove that friendly page altogether (under the assumption that we don't really throw 400 errors under normal situations) or should we just  put an exception in for these particular requests to pass them through.
I'm going to have a quick scan through the router logs to see if we serve many 400 errors.

nicl:
I think there's a general problem here (throughout this story) that, as has been mentioned, theguardian.com is not really set up for forms. We can hack away special solutions for Discussion (and that might be the right solution for this card), but it's just adding more special logic for Discussion, when really it would be nice to have a more general solution here.

Unless we think we're unlikely to have more forms on dotcom (I don't know how likely this is.)

gidsg:
OK, I have made the suggested change to allow 400s through the router. So closing this in favour of https://github.com/guardian/platform/pull/891

PR #11613, opened 2016-01-19
Generate liveblog timeline server-side

johnduffell:
The timeline is scraped from the page at the moment.  With pagination, we still want the latest stuff to appear in the key events regardless of what page you're on.

This PR has two non visible changes:
- changes the events to be generated on the server side, this means we can always generate then from the latest posts rather than the current page.  
- makes the auto update blocks generated from the templates rather than the cleaners so the share buttons are back (although see the note below about permalinks)

At present the links are all hard coded to go to the "latest" page, and only generate events on that page (so there may be some missing from older pages)  Once permalinks are fixed, we can use those instead, and possibly shrink the page size depending what they want downstairs.

Future tweaks needed are to update the timeline regardless of which page you're on, also permalinks at present only work while the posts are on the live page, not on archive pages.


OliverJAsh:
Less JS. Awesome. :+1:

PR #11614, opened 2016-01-19
Send longer A/B test identifiers to DFP

jbreckmckye:
Right now, when we pass user A/B test participation data to DFP, we only pass the first letter of the name of each test variant. This causes us problems when they begin with the same initial -

```
'stickyNavOn' => 's'
'stickyNavOff' => 's' // Not unique
```

This test tries to pass as much of the variant name as it can, respecting the 40 character limit for data in a DFP key value pair.


Calanthe:
:+1:

PR #11615, opened 2016-01-19
Fix paid-for logo bug

kelvin-chappell:
This fixes a bug where pre-logo-load sponsor text was being shown in wrong places occasionally.
Now there should be no confusion between paid-for keywords and paid-for series.


JonNorman:
:+1:

PR #11616, opened 2016-01-19
Stop r.js filtering directories with the 'test' substring in their name

jbreckmckye:
I had a recent issue with my project refusing to `make compile`, because a directory of mine had the 'test' substring in its name ('prebid-test'). This was because r.js filters files and directories that match `requirejs.js`'s fileExclusionRegExp key.

I've made a change so that directories with names like 'prebid-test' are still copied, but that doesn't cause the `/test/` directory to be ported as well. We can't tell require.js to target a specific folder path, unfortunately - the test is applied to the _parts_ of the path, rather than the path as a whole - but we can at least only ignore things with the _exact_ name `path`.


jbreckmckye:
@OliverJAsh - this is a fix for the issue with my branch yesterday

OliverJAsh:
Do we still need to exclude `test`? I think that was used to exclude `static/src/javascripts/test` which no longer exists?

jbreckmckye:
@OliverJAsh - You're correct. The directory was still present locally, so it looked like something that still needed to be filtered.

OliverJAsh:
Ahh, it's odd that it didn't appear as an untracked folder in your Git?

OliverJAsh:
:+1: although the build has failed due to some switch

PR #11617, opened 2016-01-19
Sticky ad: honour Rubicon ad height changes (and tidy up)

OliverJAsh:
/cc @Jholder112233 

https://github.com/guardian/frontend/pull/11617/files?w=1

### DIFF
--- a/static/src/javascripts/projects/common/modules/navigation/newSticky.js
+++ b/static/src/javascripts/projects/common/modules/navigation/newSticky.js
@@ -95,7 +95,7 @@ define([
                         $header.css({ 'margin-top': state.adHeight });
 
                         $adBanner.css({ 'max-height': state.adHeight });
-                        var userHasScrolledPastHeader = state.scrollY > state.headerHeight;
+                        var userHasScrolledPastHeader = state.pageYOffset > state.headerHeight;
                         if (userHasScrolledPastHeader) {
                             $adBanner.css({
                                 'position': 'absolute',
@@ -105,13 +105,13 @@ define([
                             $adBanner.css({ 'position': 'fixed' });
                         }
 
-                        var scrollIsAtTop = state.scrollY === 0;
+                        var scrollIsAtTop = state.pageYOffset === 0;
                         var diff = state.adHeight - state.previousAdHeight;
                         var adHeightHasIncreased = diff > 0;
                         if (!scrollIsAtTop && adHeightHasIncreased) {
                             // If the user is not at the top and the ad height has increased,
                             // we want to offset their scroll position
-                            window.scrollTo(state.scrollX, state.scrollY + diff);
+                            window.scrollTo(state.pageXOffset, state.pageYOffset + diff);
                         } else if (!state.firstRender) {
                             // Otherwise we want to transition the change when it happens.
                             // Avoid an initial transition when we apply the margin top for the first time
@@ -122,7 +122,7 @@ define([
 
                 var getScrollCoords = function () {
                     return fastdom.read(function () {
-                        return [window.scrollX, window.scrollY];
+                        return [window.pageXOffset, window.pageYOffset];
                     });
                 };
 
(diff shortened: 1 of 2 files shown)


static/src/javascripts/projects/common/modules/navigation/newSticky.js
```diff
@@ -61,72 +61,80 @@ define([
                 };
 
                 var render = function (state) {
-                    return fastdom.read(function () {
```
regiskuckaertz: That's clean! What do you think of moving all the logic out of the promise and refactoring the write to essentially:

```
function render(state) {
  /* build headerCss, adBannerCss and scroll objects here */
    $header.css(headerCss);
    $adBanner.css(adBannerCss);
    window.scrollTo(scroll.x, scroll.y);
  };
}
```

static/src/javascripts/projects/common/modules/navigation/newSticky.js
```diff
@@ -61,72 +61,80 @@ define([
...
+                        var diff = state.adHeight - state.previousAdHeight;
+                        var adHeightHasIncreased = diff > 0;
+                        if (!scrollIsAtTop && adHeightHasIncreased) {
+                            // If the user is not at the top and the ad height has increased,
+                            // we want to offset their scroll position
+                            window.scrollTo(state.scrollX, state.scrollY + diff);
+                        } else if (!state.firstRender) {
+                            // Otherwise we want to transition the change when it happens.
+                            // Avoid an initial transition when we apply the margin top for the first time
+                            $header.css('transition', 'margin-top 1s cubic-bezier(0, 0, 0, 0.985)');
```
regiskuckaertz: I wonder if it is there is an opportunity to use classes to switch on/off the positioning and transitioning.

static/src/javascripts/projects/common/modules/navigation/newSticky.js
```diff
@@ -61,72 +61,80 @@ define([
...
+                        }
                     });
                 };
 
+                var getScrollCoords = function () {
+                    return fastdom.read(function () {
+                        return [window.scrollX, window.scrollY];
```
Jholder112233: Best to switch these to pageY/XOffset

Jholder112233:
Also whilst shipping this you may as well add the listener for the postMessage

Jholder112233:
:+1: :star:

static/src/javascripts/projects/common/modules/navigation/newSticky.js
```diff
@@ -53,88 +46,122 @@ define([
...
+                            resolve();
+                        }
+                    });
+                });
+
+                var newRubiconAdHeightPromise = new Promise(function (resolve) {
+                    window.addEventListener('message', function (event) {
+                        var data = JSON.parse(event.data);
+                        var $iframe = getAdIframe();
+                        var isRubiconAdEvent = data.type === 'set-ad-height';
```
OliverJAsh: /cc @jbreckmckye @Calanthe @uplne FYI, we've added a `postMessage` to the Rubicon expand script so we can know when an ad changes its size. This might be useful for other things, too :smile:

static/src/javascripts/projects/common/modules/navigation/newSticky.js
OliverJAsh: Yeah that's not a bad idea :smile:

static/src/javascripts/projects/common/modules/navigation/newSticky.js
jbreckmckye: Do we ever have multiple rubicon ads running on a page? We might need to do something to differentiate the messages' source frames if we do. (I hope not, because this is really awkward when you don't have control over the iframe)

OliverJAsh: So the line below :-P

jbreckmckye: Fair enough!

PR #11618, opened 2016-01-19
Amp video style fix

NataliaLKB:
Making sure we are utilising poster for videos in AMP. Also added a fallback.

Before:
image

After:
image


common/app/views/support/cleaner/AmpEmbedCleaner.scala
```diff
@@ -28,8 +31,7 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
         // All videos need to start with https for AMP.
         // Temperary code until all videos returned from CAPI are https
```
johnduffell: worth noting that this isn't the case any more, there's a secureFile field now

johnduffell:
:+1:

common/app/views/support/cleaner/AmpEmbedCleaner.scala
NataliaLKB: Yes you are right, I will update this in a new PR

PR #11620, opened 2016-01-20
Make switches a bit better

johnduffell:
I noticed that switches seem to fail the test a day before the beginning of the last day.  In the process of fixing that, I found a couple of other bits that needed tidying.  I also fixed up that annoying y2k1 bug where we assume the year 2100 will never happen but the switchboard thinks it will happen in 30661 days.

### DIFF
--- a/admin/app/controllers/RadiatorController.scala
+++ b/admin/app/controllers/RadiatorController.scala
@@ -24,7 +24,7 @@ object RadiatorController extends Controller with Logging with AuthLogging with 
   // put it in your properties file as github.token=XXXXXXX
   lazy val githubAccessToken = Configuration.github.token.map{ token => s"?access_token=$token" }.getOrElse("")
 
-  def switchesExpiringThisWeek = {
+  def switchesExpiringSoon = {
     Switches.all.filter { switch =>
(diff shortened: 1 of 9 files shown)


admin/app/controllers/RadiatorController.scala
```diff
@@ -26,7 +26,7 @@ object RadiatorController extends Controller with Logging with AuthLogging with
 
   def switchesExpiringThisWeek = {
     Switches.all.filter { switch =>
-      switch.sellByDate.isBefore(new LocalDate().plusDays(7))
+      Switch.expiry(switch).expiresSoon
```
kelvin-chappell: Perhaps `expiresSoon` should be `expiresThisWeek` if the two concepts are coupled together.

johnduffell: or I could change switchesExpiringThisWeek to switchesExpiringSoon?

kelvin-chappell:
:+1:

admin/app/controllers/RadiatorController.scala
kelvin-chappell: Yes.  That would be better.

PR #11622, opened 2016-01-20
test top banner transition hunch

sndrs:
this is just a pref to test a hunch that transitioning the height of the top banner contributes to crashes we're currently seeing. 

i cannot reproduce the crashes, but if people who see them constantly report a change with this pref, i'll beacon it to see if anything wider can be detected. 

if not, i'll just remove it.

optin with http://www.theguardian.com/uk#gu.prefs.no-banner-transition=true

cc @gtrufitt

### DIFF
--- a/common/app/templates/inlineJS/blocking/applyRenderConditions.scala.js
+++ b/common/app/templates/inlineJS/blocking/applyRenderConditions.scala.js
@@ -87,6 +87,15 @@
         }
     }
 
+    (function (window) {
+        try {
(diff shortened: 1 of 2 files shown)


OliverJAsh:
This could be interesting! You might want to wrap it in a switch to keep us honest about removing it.

sndrs:
i will if it goes as far as beacons. i'll remove or do that in a few hours

gtrufitt:
Sweet, I'll give this a try when it's out!

sndrs:
cc @colster too

PR #11623, opened 2016-01-20
Remove "more galleries" component in paid-content-redesign

jbreckmckye:
This is part of the paid content redesign. It is a small change to hide the 'more galleries' component when a gallery is paid for by a client (because those links take the user away from partner content).

@Calanthe


Calanthe:
:+1:

PR #11624, opened 2016-01-20
add azed to types in crossword search page

NathanielBennett:
Adds azed crosswords to the search form here: http://www.theguardian.com/crosswords/search. 

Since there's still a lode of missing azeds in the api ( see http://www.theguardian.com/crosswords/series/azed ) and they're rendered at normal articles(with a different form of URL to other crosswords), rather than crosswords, we'll continue to exclude them from the lookup form until we can implement a workable solution (I've run this by @ScottPainterGNM and @crifmulholland ): 

What you think @desbo ?


desbo:
looks good to me :+1:

PR #11625, opened 2016-01-20
don't transition the height of the top banner, ever

sndrs:
makes #11622 permanent.

it didn't affect crashes, but it definitely makes page load feel quieter

cc @Calanthe @uplne

### DIFF
--- a/common/app/templates/inlineJS/blocking/applyRenderConditions.scala.js
+++ b/common/app/templates/inlineJS/blocking/applyRenderConditions.scala.js
@@ -87,15 +87,6 @@
         }
     }
 
-    (function (window) {
-        try {
-            var bannerPref = JSON.parse(window.localStorage.getItem('gu.prefs.no-banner-transition')).value;
-            if (bannerPref) {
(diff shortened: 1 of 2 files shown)


OliverJAsh:
Could you post a recording of before/after? So instead of a transition, the page just jumps?

sndrs:
you can view exactly what it does by opting in to the temporary pref i added in #11622 http://www.theguardian.com/uk#gu.prefs.no-banner-transition=true

uplne:
As I remember the whole purpose of the transition (UX/product decision) was to make it a little bit less painful for the user. That is the only reason why we haven't removed it months ago.

sndrs:
@uplne i've updated it to set a min-height on the iframe. this should mean there's nothing to transition when a 90px ad loads, but if it's a larger one, the page will still slide to accommodate it

screen recording 2016-01-20 at 04 48 pm
screen recording 2016-01-20 at 04 50 pm

OliverJAsh:
I think this is fine :+1:

PR #11626, opened 2016-01-20
AMP tones 

zeftilldeath:
Added special report tone to onward journeys in AMP. 

screen shot 2016-01-20 at 14 45 19

Also prevented them from wrapping if close to an ad:

screen shot 2016-01-20 at 14 45 28

FAO @NataliaLKB


NataliaLKB:
:+1: much better!

PR #11628, opened 2016-01-20
Mobile merch component kicking desktop inline MPUs

regiskuckaertz:
When:
- a page is set up to have inline merchandising components
- there's only merch components targeted at mobile visitors
- a visitor is on desktop

Then there is a chance no inline MPU will be registered for this visitor. This is the result of two things:
- that inline slot insertion and inline slot rendering are decoupled
- that space finding rules on desktop only allow inline ad slots 1300px apart, whether or not these slots will ultimately be filled

There are many possible solutions to this problem. Let's first discard the idea of working upstream and prevent this case to appear ever—we don't want to change the current business logic. Then, logically, we can wait for DFP to return something for inline merch components and then only can we insert inline MPUs. This is essentially what this PR does, as it makes the "insert body adverts" a two-pass algorithm, i.e. if no MPU could be added on a first try, then we wait until DFP has finished and see if merch components are empty, in which case a second pass of inserting MPUs is triggered. To do that, we must first remove empty slots before.

Another solution is part of [another PR](https://github.com/guardian/frontend/pull/11629). It doesn't use the two-pass trick and thus is less effective, but it works.


uplne:
What about testing it?

regiskuckaertz:
You're right, I'll add a test case for the second-pass at inserting slots. Yeah!

static/src/javascripts/projects/common/modules/commercial/article-body-adverts.js
```diff
@@ -31,97 +43,114 @@ define([
...
+        if (config.switches.viewability && detect.getBreakpoint() !== 'mobile') {
+            return addArticleAds(2, rules).then(function (countAdded) {
+                if (countAdded === 0) {
+                    mediator.on('modules:commercial:dfp:rendered', onAdRendered);
+                }
 
-        reset: function () {
-            ads = [];
+                return countAdded === 2 ?
+                    addArticleAds(8, getLongArticleRules()) :
```
uplne: It might be useful for our future selves to document more what these magic count parameters mean.

uplne:
:+1:

PR #11632, opened 2016-01-20
Onward Journeys on AMP only take 4 stories, 5 for popular

NataliaLKB:
On mobile, there are just too make stories to scroll through at the bottom of an article. this makes it so there are only 4 that show per section except for popular which shows 5. This is because it seems weird to show "top 4", "top 5" is better.

cc @stephanfowler


stephanfowler:
:+1:

PR #11634, opened 2016-01-20
Small tweaks to feedback wording

scottp719:
@johnduffell 

Reasoning behind this is that many people leave feedback that can be a little blue to say the least, then when we reply they are embarrassed as assumed they were shouting into the void. Just tweaking the wording to make it a little more clear that we do read everything (just don't necessarily commit to responding).


johnduffell:
haha great! :+1: :speedboat: happy happy users!

PR #11636, opened 2016-01-21
make liveblog key event links scroll correctly

johnduffell:
Key events are scrolling to the top on click, but they should scroll to the right block.


Jholder112233:
:+1: :gem:

PR #11637, opened 2016-01-21
Disable sticky MPU in liveblogs

regiskuckaertz:
In liveblogs, there's no `DIV.js-content__article-body` and thus the sticky-mpu script fails during initialization. It has been decided to disable the stickiness in all cases where such an element does not exist.

Small improvement though: access to layout-related properties is now sealed in Fastdom reads.


static/src/javascripts/projects/common/modules/commercial/ads/sticky-mpu.js
```diff
@@ -20,14 +18,23 @@ define([
...
+
+        element = document.querySelector(config.page.hasShowcaseMainElement ? '.media-primary' : '.content__article-body');
+        if (!element) {
+            return;
+        }
+
+        return fastdom.read(function () {
+            return element[config.page.hasShowcaseMainElement ? 'offsetHeight' : 'offsetTop'];
+        }).then(function (articleBodyOffset) {
+            this.$adSlot.parent().css('height', (articleBodyOffset + mpuHeight) + 'px');
```
johnduffell: should the `then` happen within a fastdom.write?

johnduffell:
great! :+1: just the small q about whether the dom writing should be explicitly in a fastdom.write or whether it won't affect things

static/src/javascripts/projects/common/modules/commercial/ads/sticky-mpu.js
regiskuckaertz: yes, definitely! thanks for spotting this @johnduffell :-)

static/src/javascripts/projects/common/modules/commercial/ads/sticky-mpu.js
```diff
@@ -20,14 +18,25 @@ define([
...
+        var element;
 
         if (this.$adSlot.data('name') !== 'right') {
             return;
         }
-        articleBodyOffset = config.page.hasShowcaseMainElement ? $('.media-primary').dim().height : qwery('.content__article-body')[0].offsetTop;
-        this.$adSlot.parent().css('height', (articleBodyOffset + mpuHeight) + 'px');
-        new Sticky(this.$adSlot[0], { top: this.opts.top }).init();
+
+        element = document.querySelector(config.page.hasShowcaseMainElement ? '.media-primary' : '.content__article-body');
```
uplne: I would move declaration of the variable here. We had many discussions about this and we are already moving to this pattern which makes more sense with ES6 anyway.

uplne:
:+1:

PR #11642, opened 2016-01-21
Update survey message to allow headline

gtrufitt:
New message styling to allow a headline - for use in survey message


jbreckmckye:
Looks good :+1:

PR #11643, opened 2016-01-21
Improve grid alignment of new sponsored content containers

jbreckmckye:
They used to look like this...

image

Now they look like this...

_Desktop_
image

_Tablet_
image

_Mobile_
image

I still need to fix the styling of the 'paid for' logos. This PR just handles the layout of the boxes.

I'm not sure if the way I've done this is the best possible. I'm not terribly au fait with Scala or Play templates, so I'm happy to hear any suggestions about a better approach.

@Calanthe

### DIFF
--- a/static/src/stylesheets/module/commercial/_paidfor.scss
+++ b/static/src/stylesheets/module/commercial/_paidfor.scss
@@ -1,6 +1,12 @@
 /* ==========================================================================
    Paid for component
    ========================================================================== */
+
+.fc-container .fc-container__inner--paidfor {
+    border: none;
+    padding: 0;
+}
+
 .commercial--paidfor {
 
     .commercial__header,
@@ -55,6 +61,7 @@
     .lineitems {
         display: flex;
         flex-flow: row nowrap;
+        width: auto;
     }
 
     .lineitem {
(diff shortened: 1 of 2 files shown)


Calanthe:
In my opinion, it's fine. Maybe ask @kelvin-chappell about the scala template.

I remember having some problems with IE and those flexboxes. Did you check how does it look like in IE?

static/src/stylesheets/module/commercial/_paidfor.scss
```diff
@@ -1,6 +1,12 @@
 /* ==========================================================================
    Paid for component
    ========================================================================== */
+
+.fc-container .fc-container__inner--paidfor {
```
regiskuckaertz: Isn't `.fc-container__inner--paidfor` enough on its own? unless `.fc-container__inner` has already been nested elsewhere of course.

kelvin-chappell:
Template code looks good so apart from comments from @Calanthe and @regiskuckaertz  :+1:

static/src/stylesheets/module/commercial/_paidfor.scss
jbreckmckye: Specificity issues. I do not want to open a can of worms by trying to refactor the original module.

regiskuckaertz: I know what you mean. :+1:

jbreckmckye:
Speaking to Pat, he'd like us to solve IE-related issues later. I'm going to merge this for the moment as there's some other critical work we need to do for Monday.

PR #11644, opened 2016-01-21
Add frontend profile name to project AWSCredentialsProviderChain

janua:
The default profile in `janus` is `frontend` for our project.

This is an annoying mismatch; you CAN change the profile in `janus` and I have added `localStorage` caching for it and also added `--alias <profileName>` to `feria`, but I think the real fix is to align with `janus`.

@rich-nguyen Not sure whether to leave `nextgen` profile and have `frontend` take precedence over `nextgen` or just remove it. Thoughts?

See https://github.com/guardian/feria/pull/3 and https://github.com/guardian/janus/pull/191

Goo tool is also on `nextgen`, I will PR this also.

### DIFF
--- a/common/app/common/configuration.scala
+++ b/common/app/common/configuration.scala
@@ -414,6 +414,7 @@ class GuardianConfiguration(val application: String, val webappConfDirectory: St
       val provider = new AWSCredentialsProviderChain(
         new EnvironmentVariableCredentialsProvider(),
         new SystemPropertiesCredentialsProvider(),
+        new ProfileCredentialsProvider("frontend"),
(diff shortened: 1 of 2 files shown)


rich-nguyen:
Having both `nextgen` and `frontend` for a short period of time gives devs a chance to transition. Then we should remove `nextgen` once all those permanent access keys have been revoked.

johnduffell:
:+1: and agreed with rich, consider adding a switch to remind us to remove it.  we can eradicate nextgen from the readme by then

janua:
@johnduffell Added a switch now.

johnduffell:
:star2: :+1:

PR #11645, opened 2016-01-21
Are we ready for videos on HTTPS?

NataliaLKB:
Since @marialivia16 did the great work of adding a secureFile field to videos, we should use it! However we will still uses the insecure url if the secure one isn't there.

cc @rich-nguyen @johnduffell 

Todo in another PR:
- Audio

### DIFF
--- a/common/app/model/Asset.scala
+++ b/common/app/model/Asset.scala
@@ -82,20 +82,23 @@ object VideoAsset {
     VideoAsset(
       fields = Helpers.assetFieldsToMap(asset),
       mimeType = asset.mimeType,
-      url = asset.file )
+      url = asset.file,
+      secureUrl = asset.typeData.get.secureFile )
   }
 }
 
 case class VideoAsset(
   fields: Map[String,String],
   url: Option[String],
-  mimeType: Option[String]) {
+  mimeType: Option[String],
+  secureUrl: Option[String]) {
 
   val width: Int = fields.get("width").map(_.toInt).getOrElse(0)
   val height: Int = fields.get("height").map(_.toInt).getOrElse(0)
   val encoding: Option[Encoding] = {
-    (url, mimeType) match {
-      case (Some(url), Some(mimeType)) => Some(Encoding(url, mimeType))
+    val urlToUse = if (secureUrl.isEmpty) url else secureUrl
+    (urlToUse, mimeType) match {
+      case (Some(urlToUse), Some(mimeType)) => Some(Encoding(urlToUse, mimeType))
       case _ => None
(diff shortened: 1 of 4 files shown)


johnduffell:
:+1: :speedboat:

common/app/model/Asset.scala
```diff
@@ -82,20 +82,23 @@ object VideoAsset {
     VideoAsset(
       fields = Helpers.assetFieldsToMap(asset),
       mimeType = asset.mimeType,
-      url = asset.file )
+      url = asset.file,
+      secureUrl = asset.typeData.get.secureFile )
```
philwills: Any reason to `get` rather than `flatMap` here?

I presume that in the vast majority of cases this would make no difference as videos would have typeData, but it'd be a bit more consistent with other accesses of this.

rich-nguyen: `secureUrl = asset.typeData.flatMap(_.secureFile)`

rich-nguyen: Or even better `secureUrl = asset.typeData.flatMap(_.secureFile).orElse(asset.file)`

rich-nguyen: I think if we did that, we wouldn't need the url field at all in `VideoAsset`

NataliaLKB: Yes that is much better! Will make the change now.

rich-nguyen: @philwills Spooky coincidence! That must have caused a race condition down at Github. "two users, same message, same time = DDOS. Shutdown github immediately"

NataliaLKB: https://github.com/guardian/frontend/pull/11650

common/app/views/support/HtmlCleaner.scala
```diff
@@ -185,10 +185,10 @@ object VideoEncodingUrlCleaner {
   def apply(url: String): String = url.filter(_ != '\n')
 }
 
-object AmpVideoSrcCleaner {
+object AmpSrcCleaner {
   def apply(videoSrc: String) = {
-    // All videos need to start with https for AMP.
-    // Temperary code until all videos returned from CAPI are https
+    // All media sources need to start with https for AMP.
+    // Temperary code until all media urls returned from CAPI are https
```
gidsg: Typo 
> Temperary

common/app/model/Asset.scala
philwills: @rich-nguyen :ghost:

mchv:
:+1: for https

PR #11646, opened 2016-01-21
Updated old article message

zeftilldeath:
Tidied up the old article message.

screen shot 2016-01-21 at 17 44 03

screen shot 2016-01-21 at 17 43 54

screen shot 2016-01-21 at 17 43 36

### DIFF
--- a/common/app/views/fragments/contentAgeNotice.scala.html
+++ b/common/app/views/fragments/contentAgeNotice.scala.html
@@ -2,7 +2,7 @@
 
 @if(ageMessage.nonEmpty) {
     <div class="old-article-message">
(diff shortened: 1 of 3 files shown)


JustinPinner:
Will this work on non-white backgrounds? Also, the original designs specified the clock icon I think.

zeftilldeath:
@JustinPinner Great point with the dark backgrounds, have added a variant. Reverted to clock too. Have updated screenshots.

JustinPinner:
Looking good :+1:

paperboyo:
Better typography + clock without the hole = win!
Q: “isn’t it too small now?”; A: “no, it was too big before”

PR #11648, opened 2016-01-21
Fix sponsors' logos on section fronts

kelvin-chappell:
They stopped working on automated fronts.  This change should bring them back again.

/cc @JonNorman


JonNorman:
:+1:

PR #11651, opened 2016-01-21
Add News Alert endpoints to admin-jobs app

TBonnin:
TODO:
- [x] Create NewsAlertNotification models
- [x] Create endpoint to handle POST /news-alerts/alert
- [x] Save breaking news output to S3
- [x] Handle state (ie: keep in the output the previous notifications for other "editions")
- [x] Manage concurrency to avoid race conditions
- [x]  Add Authentication

The data coming from n10n is expected to be formatted like this:

```
{
"uid":"eaf51128-aecc-4fcb-bf5f-2064ad8g455a" //UUID
"title":"This is a breaking news title", //String
"message":"This is a breaking news message", //String
"thumbnailUrl":"http://i.guimcode.co.uk.global.prod.fastly.net/img/media/54c2dc737fc82bf793dd919694e3ea7111cf2d82/0_169_3936_2363/140.jpg", //URI optional
"link":"http://gu.com/p/4fgcd", //URI
"imageUrl":"http://i.guimcode.co.uk.global.prod.fastly.net/img/media/54c2dc737fc82bf793dd919694e3ea7111cf2d82/0_169_3936_2363/140.jpg", //URI optional
"publicationDate":"2016-01-18T12:21:01.000Z", //Datetime
"urlId":"category/2016/feb/01/slug-of-article", //URI
"topics":["breaking/uk", "breaking/sport"] // Set of Strings
}
```

### DIFF
--- a/admin-jobs/app/controllers/BreakingNewsApi.scala
+++ b/admin-jobs/app/controllers/BreakingNewsApi.scala
@@ -7,7 +7,6 @@ import play.api.libs.json._
 import services.S3
 
 import scala.concurrent.Future
-import scala.util.{Failure, Success, Try}
 
 trait S3BreakingNews extends S3 {
   override lazy val bucket = Configuration.aws.bucket
@@ -23,37 +22,28 @@ trait BreakingNewsApi extends Logging with ExecutionContexts {
   val s3: S3BreakingNews
   val breakingNewskey = s3.getKeyForPath("breaking-news")
 
-  def getBreakingNews : Future[Option[JsValue]] = {
+  def getBreakingNews: Future[Option[JsValue]] = {
     Future {
-      Try(s3.get(breakingNewskey)) match {
-        case Success(content) => content match {
-          case Some(s) => Try(Json.parse(s)) match {
-            case Success(json) =>
-              Some(json)
-            case Failure(e) =>
-              log.error(s"Cannot deserialize Breaking News json (${e.getMessage})")
-              None
-          }
-          case None =>
-            None
-        }
-        case Failure(e) =>
-          log.error(s"Cannot fetch Breaking News json from S3 (${e.getMessage})")
-          None
-      }
+      s3.get(breakingNewskey)
+    }.map {
+      case Some(s) => Some(Json.parse(s))
+      case _ =>
+        val e = new Exception("No Breaking News content")
+        log.error(e.getMessage)
+        throw e
+    }.recover {
+      case e: Exception =>
+        log.error(s"Cannot fetch Breaking News json (${e.getMessage})")
+        throw e
     }
   }
-
-  def putBreakingNews(json: JsValue) : Future[Boolean] = {
+  def putBreakingNews(json: JsValue): Future[Unit] = {
     Future {
-      Try(s3.putPublic(breakingNewskey, json.toString, "application/json")) match {
-        case Success(_) =>
-          true
-        case Failure(e) =>
-          log.error(s"Cannot write Breaking News json to S3 (${e.getMessage})")
-          false
-      }
-
+      s3.putPublic(breakingNewskey, json.toString, "application/json")
+    }.recover {
+      case e: Exception =>
+        log.error(s"Cannot write Breaking News json (${e.getMessage})")
+        throw e
     }
   }
 
--- a/admin-jobs/app/controllers/BreakingNewsApi.scala
+++ b/admin-jobs/app/controllers/BreakingNewsApi.scala
@@ -1,0 +1,64 @@
+package controllers
+
+import common.{ExecutionContexts, Logging}
+import conf.Configuration
+import play.Play
+import play.api.libs.json._
+import services.S3
+
+import scala.concurrent.Future
+import scala.util.{Failure, Success, Try}
+
+trait S3BreakingNews extends S3 {
+  override lazy val bucket = Configuration.aws.bucket
+  lazy val stage = if (Play.isTest) "TEST" else Configuration.environment.stage.toUpperCase
+  val namespace = "notifications"
+  lazy val location = s"$stage/$namespace"
+  def getKeyForPath(path: String): String = s"$location/$path.json"
+}
+object S3BreakingNews extends S3BreakingNews
+
+trait BreakingNewsApi extends Logging with ExecutionContexts {
+
+  val s3: S3BreakingNews
+  val breakingNewskey = s3.getKeyForPath("breaking-news")
+
+  def getBreakingNews : Future[Option[JsValue]] = {
+    Future {
+      Try(s3.get(breakingNewskey)) match {
+        case Success(content) => content match {
+          case Some(s) => Try(Json.parse(s)) match {
+            case Success(json) =>
+              Some(json)
+            case Failure(e) =>
+              log.error(s"Cannot deserialize Breaking News json (${e.getMessage})")
+              None
+          }
+          case None =>
+            None
+        }
+        case Failure(e) =>
+          log.error(s"Cannot fetch Breaking News json from S3 (${e.getMessage})")
+          None
+      }
+    }
+  }
+
+  def putBreakingNews(json: JsValue) : Future[Boolean] = {
+    Future {
+      Try(s3.putPublic(breakingNewskey, json.toString, "application/json")) match {
+        case Success(_) =>
+          true
+        case Failure(e) =>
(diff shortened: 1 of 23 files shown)


admin-jobs/app/models/BreakingNews.scala
```diff
@@ -0,0 +1,61 @@
...
+import java.util.UUID
+import play.api.libs.json._
+import play.api.libs.functional.syntax._
+
+case class BreakingNewsCollection(name: String, href: String, id: UUID, alerts: Set[NewsAlertNotification])
+
+case class BreakingNews(alerts: Set[NewsAlertNotification]) {
+
+  val collections = List(
+    BreakingNewsCollection("UK alerts", "uk", UUID.fromString("5dff149e-3cc8-4b23-ba7d-c7596d8e39e5"), alerts.filter(_.isOfType(NewsAlertTypes.Uk))),
```
TBonnin: I don't think the UUID for each collection is used anywhere. I kept it anyway. Let me know if this is unecessary

TBonnin:
@rich-nguyen I would like to get your feedback on this.

TBonnin:
I am assuming from now that the data coming from n10n is formatted like this:
```
{
"id":"c3def817-ab16-4d1d-9eae-c090a5c753f9",
"title":"This is a breaking news title",
"message":"This is a breaking news message",
"thumbnailUrl":"http://i.guimcode.co.uk.global.prod.fastly.net/img/media/54c2dc737fc82bf793dd919694e3ea7111cf2d82/0_169_3936_2363/140.jpg",
"link":"http://gu.com/p/4fgcd",                          "imageUrl":"http://i.guimcode.co.uk.global.prod.fastly.net/img/media/54c2dc737fc82bf793dd919694e3ea7111cf2d82/0_169_3936_2363/140.jpg",
"publicationDate":"2016-01-18T12:21:01.000Z",
"topics":["breaking/us"]
}
```

We can tweak that later.

admin-jobs/app/controllers/BreakingNewsApi.scala
```diff
@@ -0,0 +1,64 @@
...
+          }
+          case None =>
+            None
+        }
+        case Failure(e) =>
+          log.error(s"Cannot fetch Breaking News json from S3 (${e.getMessage})")
+          None
+      }
+    }
+  }
```
TBonnin: I feel like there might be a better way to achieve the same result here without those nested matches :wink: but I don't know how

rich-nguyen: I think a `map` and `recover` could help here, will chat to you

admin-jobs/app/controllers/BreakingNewsApi.scala
```diff
@@ -0,0 +1,64 @@
...
+
+  def putBreakingNews(json: JsValue) : Future[Boolean] = {
+    Future {
```
rich-nguyen: it looks like `s3.putPublic` is a blocking request, so a Future type is misleading here.

admin-jobs/app/models/BreakingNews.scala
rich-nguyen: I think you're right, the UUID isn't needed here.

admin-jobs/app/models/NewsAlertNotification.scala
```diff
@@ -0,0 +1,76 @@
...
+            case Success(uri) => JsSuccess(uri)
+            case Failure(_) => error
+          }
+        case _ => error
+      }
+    }
+  }
+
+  // publicationDate (DateTime)
+  implicit val dtf = new Format[DateTime] {
```
rich-nguyen: I'm surprised you needed this one?

admin-jobs/app/models/NewsAlertNotification.scala
```diff
@@ -0,0 +1,76 @@
...
+  implicit val jWrites = Json.writes[NewsAlertNotification]
+  implicit val jReads: Reads[NewsAlertNotification] = (
+    (__ \ "id").read[UUID] and
+      (__ \ "title").read[String] and
+      (__ \ "message").read[String] and
+      (__ \ "thumbnailUrl").readNullable[URI] and
+      (__ \ "link").read[URI] and
+      (__ \ "imageUrl").readNullable[URI] and
+      (__ \ "publicationDate").read[DateTime] and
+      (__ \ "topics").read[Set[NewsAlertType]].filter(ValidationError("Notification should have at least one topic"))(_.nonEmpty)
```
rich-nguyen: Maybe we could find a way to avoid all this boilerplate. If topics was a `List[String]`, then you could a val in `NewsAlertNotification` which could provide a `List[NewsAlertType]`. Lifts the deserialisation up one level into the case class.

rich-nguyen:
Brill :+1: , especially the test coverage.

admin-jobs/app/models/BreakingNews.scala
TBonnin: ok I will remove

admin-jobs/app/models/NewsAlertNotification.scala
TBonnin: That's what I had done at first (ie: `implicit val jf = Json.format[NewsAlertNotification]`) but then I decided to validate the number of topics and make sure there is at least one.

admin-jobs/app/models/NewsAlertNotification.scala
TBonnin: I was surprised too but I couldn't make it work without it

admin-jobs/app/controllers/BreakingNewsApi.scala
TBonnin: I thought that was the point of a Future, ie: perform the block of code concurrently. :astonished:

admin-jobs/app/controllers/NewsAlertController.scala
```diff
@@ -1,10 +1,48 @@
...
+      case Some(json) => Cached(30)(Ok(json))
+      case None => NoContent
+    }.recover{
+      case _ => InternalServerError(Json.toJson(NewsAlertError("Error while accessing alerts")))
+    }
   }
 
+  def create() : Action[NewsAlertNotification] = Action.async(BodyJson[NewsAlertNotification]) { request =>
+    val receivedNotification : NewsAlertNotification = request.body
+    val result = breakingNewsUpdater ? NewNotificationRequest(receivedNotification)
```
TBonnin: I decided here to wait for the answer from the updater actor so the success/failure can be passed by to the endpoint caller. Given the load of this endpoint that shouldn't be an issue.
However if we want to increase the throughput we could fire and forget to the actor and returns an 201(Accepted)

admin-jobs/app/models/NewsAlertNotification.scala
piuccio: sorry for jumping in the discussion without really understanding things. Topics are not fixed, new ones can be easily created in the tool, what happens when a topic cannot be parsed?

TBonnin: @piuccio 
According to my discussion with Alex it seems like topics for breaking news are actually fixed, or to phrase it differently we (on the frontend) would, at least for now, only care about `breaking/au`, `breaking/us`, `breaking/uk` or `breaking/international` topics.

The objective of this PR is to decouple breaking-news from Fronts by generating the json file used by the website every time a new notification with one of the topic listed above is sent from n10n. At the moment we are not changing how the website deals with Breaking News content.

Let me know if this is not clear.

piuccio: makes sense, what I meant is that in the future topics will evolve in many directions, you might have a `science` or `minute` topic... frontend and apps should simply ignore topics that are not relevant for them.

TBonnin: That's exactly what the endpoint does at the moment. If topics are not valid the request is dropped with a 400

piuccio: :+1:

TBonnin: After discussing with @rich-nguyen I change my mind. All strings will be accepted as topics and won't trigger an error. However, non breaking news topics will be ignored.

TBonnin:
Last check @rich-nguyen 
Let me know if you have any concern about this PR. If not I will deploy it tomorrow morning

PR #11652, opened 2016-01-21
Make GLab image links accessible

jbreckmckye:
This is another change for the sponsored content redesign. We add unseen anchor text for the sake of screenreaders.


uplne:
:+1: solid templating

PR #11653, opened 2016-01-21
Introducing Toast: new liveblog autoupdates

holder-josh:
_Currently set up as a 0% AB test, hence so much code duplication. We need to do this initially to test this thoroughly before switching over. Once switched, the net addition of code should be negligible/even positive_
# What is Toast?

One problem with our current liveblog implementation is that the page jumps when a new liveblog block is fetched and prepended to the liveblog body, subsequently pushing the sentence you were reading down or even out of view.

jumping

Toast is a new method of informing readers that there has been an update published in the liveblog they're currently reading. If the reader has scrolled down the page, we'll alert them to a new update with a small notification button, which they can click to read the latest update.

nojumping23

If the user is at the top of the liveblog awaiting an incoming update, we insert it straight into the DOM as we do now. Our definition of being at the top of the liveblog is that the top of the first post is inside their browser window.

It also fetches in the latest posts if you scroll back to the top of the liveblog.

refesh3

This ties in with John's liveblog pagination work, so if you're on an archived liveblog page (i.e. not the first page), we don't inject updates into the page, we alert users to them and clicking the button takes them to the 'live' (read: first) page.

I've reimplemented everything from the old AutoUpdate module, including the polling backoff when not visible. It gradually drops from polling once every 10 seconds (on desktop) all the way to once every 20 minutes (same as before). For those who likes graphs (like me), here's the [dropoff rate](https://cloud.githubusercontent.com/assets/2236852/12481227/91250e9a-c03f-11e5-8709-1c4f28718b92.png).


static/src/inline-svgs/icon/refresh.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24"><path fill="#FFF" d="M12 .5C18.4.5 23.5 5.6 23.5 12c0 6.3-5.2 11.5-11.5 11.5-3.5 0-6.6-1.5-8.7-4l-.7 4H1.3v-6.9l.5-.5h6.7v1.3l-3.8.6c1.8 2.1 4.3 3.4 7.3 3.4 5.2 0 9.4-4.2 9.4-9.4S17.2 2.5 12 2.5c-3.2 0-6 1.6-7.7 4l-.7.3-.9-.6-.2-.7c2.1-3 5.6-5 9.5-5z"/></svg>
```
sndrs: need to remove the `xmlns` so IE8 won't freak

paperboyo:
ohmygodatlast :gem:!

article/app/controllers/ArticleController.scala
```diff
@@ -105,13 +109,13 @@ object ArticleController extends Controller with RendersItemResponse with Loggin
...
 
-  def renderLiveBlogJson(path: String, lastUpdate: Option[String], rendered: Option[Boolean]) = {
+  def renderLiveBlogJson(path: String, lastUpdate: Option[String], rendered: Option[Boolean], isLivePage: Option[Boolean]) = {
     LongCacheAction { implicit request =>
       mapModel(path, blocks = true) { model =>
-        (lastUpdate, rendered) match {
-          case (Some(lastUpdate), _) => renderLatestFrom(model, lastUpdate)
-          case (None, Some(false)) => blockText(model, 6)
-          case (_, _) => render(path, model, None)
+        (lastUpdate, rendered, isLivePage) match {
```
rich-nguyen: You're matching a tuple3, but without pattern matching the third item?

static/src/javascripts/bootstraps/enhanced/liveblog.js
```diff
@@ -126,8 +127,10 @@ define([
         },
 
         createFilter: function () {
-            new LiveFilter($('.js-blog-blocks')[0]).ready();
-            new NotificationCounter().init();
+            //new LiveFilter($('.js-blog-blocks')[0]).ready();
```
rich-nguyen: are there any references to `LiveFilter`? Delete this line, and the module?

rich-nguyen: and `createFilter` should probably be renamed now.

static/src/javascripts/projects/common/modules/experiments/affix.js
```diff
@@ -63,7 +64,7 @@ define([
             markerBottomTop = this.$markerBottom.offset().top,
             elHeight      = this.$element.dim().height,
 
-            topCheck      = scrollTop >= markerTopTop,
+            topCheck      = scrollTop >= markerTopTop - 48,
```
rich-nguyen: comment needed for magic number

rich-nguyen:
:+1: ace

sndrs:
well nice

stephanfowler:
Love it.

PR #11654, opened 2016-01-21
Editionalise life and style

rich-nguyen:
Adds lifeandstyle section to the editionalised list, so UK, US and AU users who go to `/lifeandstyle` will be redirected to `/uk/lifeandstyle`


JustinPinner:
:+1:

PR #11655, opened 2016-01-21
Test moving omniture to the footer

dominickendrick:
This PR tests moving Omniture calls to the footer of the page, rather than the head.
This is in preparation for moving from two Omniture calls back to one.
We will test this for around 30 mins to check the impact against omniture confidence.

@mkopka @sndrs

### DIFF
--- a/common/app/conf/switches/PerformanceSwitches.scala
+++ b/common/app/conf/switches/PerformanceSwitches.scala
@@ -14,6 +14,15 @@ trait PerformanceSwitches {
     exposeClientSide = false
   )
 
+  val NonBlockingOmniture = Switch(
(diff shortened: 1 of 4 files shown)


sndrs:
for some context regarding #11478, removing the top up call will now precede work to try moving omniture to standard.js again. 

but the single call will need to send logged in status, which we currently detect in the non-blocking JS, and don't want to be a blocking script. 

therefore this is a kind of a trial run to see the affect of moving omniture to non-blocking inline JS (so it can use the same login status). 

the ideal is still to not inline it at all.

gtrufitt:
:+1:

PR #11656, opened 2016-01-21
Missing label in ie

uplne:
We have missing 'Advertisement' label in IE<11.

DFP sometimes returns two iframes and one of them will never go to readyState. That with combination of some, probably accidental, changes to return value meant that condition to add label where not met. And this is specific to IE.

In this PR I've also added test to make sure that we don't add label only in the case when we get 'gu-style' response.


static/src/javascripts/projects/common/modules/commercial/dfp-api.js
```diff
@@ -645,32 +643,38 @@ define([
...
+                        ) {
+                            bean.off(updatedIFrame, 'readystatechange');
+                            resolve(breakoutIFrame(updatedIFrame, $slot));
+                        }
+                    });
+                } else {
+                    resolve(breakoutIFrame(iFrame, $slot));
+                }
+            }).then(function (items) {
+                return new Promise.resolve(find(items, function (item) {
```
regiskuckaertz: You don't need to use `Promise.resolve` in a `then` callback: `return` is how you implicitly resolve a Promise, while `throw` is how you reject it. This is in stark contrast with how it works inside a `new Promise(fn)`, but the constructor was never meant to be used directly.

static/src/javascripts/projects/common/modules/commercial/dfp-api.js
```diff
@@ -645,32 +643,38 @@ define([
...
+                var iFrame = find($('iframe', $slot), function (iframe) { return iframe.id.match('__hidden__') === null; });
+
+                // No iFrame, no work to do
+                if (typeof iFrame === 'undefined') {
+                    reject();
+                }
+                // IE needs the iFrame to have loaded before we can interact with it
+                else if (iFrame.readyState && iFrame.readyState !== 'complete') {
+                    bean.on(iFrame, 'readystatechange', function (e) {
+                        var updatedIFrame = e.srcElement;
```
regiskuckaertz: Just to make sure, would this make sense: `updateIFrame = e.srcElement || e.target`

uplne:
@regiskuckaertz this bad boy is ready to go out

regiskuckaertz:
let's make this beast roar :+1:

PR #11660, opened 2016-01-22
Fix the min-height on author block

holder-josh:
Went awry in https://github.com/guardian/frontend/pull/11430

Before: 
screen shot 2016-01-22 at 10 09 31

After:
screen shot 2016-01-22 at 10 09 25


sndrs:
:+1:

PR #11662, opened 2016-01-22
Identity - Opt out of third party data profiling

danmassie:
By default all new users and existing users are opted in to third party data profiling. This PR gives them the option to opt out. On the privacy tab in the edit profile page there is a new checkbox to allow users to do this.

upload_1_13_2016_at_2_18_30_pm


identity/app/views/fragments/profile/privacyForm.scala.html
```diff
@@ -32,7 +32,27 @@ <h2 class="form__heading">Marketing</h2>
                         @checkboxField(Checkbox(privacyForm("receive3rdPartyMarketing"), '_label -> "Receive email from other organisations"))(nonInputFields, messages)
                     </div>
                 </li>
+            </ul>
+        </div>
+    </fieldset>
+    <fieldset class="fieldset">
+        <div class="fieldset__heading"></div>
```
jamespamplin: Are these empty `fieldset__heading`s needed?

danmassie: Unfortunately its all required for the layout :(

jamespamplin:
:+1: looks good to me

PR #11664, opened 2016-01-22
Expose paid-content flag on clientside

kelvin-chappell:
`guardian.config.page.isAdvertisementFeature`

We will need to do some renaming in the codebase eventually, as ad features are now called paid content.

/cc @uplne


uplne:
Do we need some test for this? Other than that :+1:

kelvin-chappell:
I'll merge this and then add a test to avoid regression.

PR #11667, opened 2016-01-22
Added 'networkFrontID' to each Edition to enable reliable identification of network fronts

JonNorman:
As described on the [Trello card](https://trello.com/c/G9TdopQq), the International front was being classed as a section rather than a network-front when making ad calls (**ct** parameter) as well as omniture calls (**c9** parameter).

This has now been fixed.

### DIFF
--- a/common/app/common/Edition.scala
+++ b/common/app/common/Edition.scala
@@ -13,6 +13,7 @@ abstract class Edition(
     val timezone: DateTimeZone,
     val locale: Locale,
     val homePagePath: String,
+    val networkFrontId: String,
     val editionalisedSections: Seq[String] = Seq(
       "", // network front
       "business",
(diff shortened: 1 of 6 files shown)


common/app/common/Edition.scala
```diff
@@ -13,6 +13,7 @@ abstract class Edition(
     val timezone: DateTimeZone,
     val locale: Locale,
     val homePagePath: String,
```
kelvin-chappell: Perhaps you can get rid of `homePagePath` now that it looks redundant?

JonNorman: It appears to be used in a few other places so haven't removed it however the `homePagePath` is now derived from `networkFrontId` by first prepending a '/'. The `toLowerCase` method has also been removed as suggested.

kelvin-chappell:
Looks good :+1:

PR #11668, opened 2016-01-22
fixing bug with iframes on amp

NataliaLKB:
Problem and suggested solution documented here: https://github.com/ampproject/amphtml/issues/1526#issuecomment-173944932

cc @stephanfowler @kenlim


stephanfowler:
:+1:

PR #11669, opened 2016-01-22
Changing number of minimum related content items for amp

NataliaLKB:
Like theguardian.com, we will show related content, even if there are less than 4 numbers.


stephanfowler:
:+1: & :+1:

PR #11670, opened 2016-01-22
Increase user survey to 2%

gtrufitt:
The survey has run for an hour with low response rate, so increasing to 2% over the weekend.


NataliaLKB:
:+1:

PR #11671, opened 2016-01-22
Liveblog button padding

stephanfowler:
Before:

lb-button-before

After:

lb-button-after

cc/ @johnduffell


Jholder112233:
So much for coding dry January :+1:

johnduffell:
is css classed as code? :stuck_out_tongue: :+1:

stephanfowler:
CSS is like vaping.

PR #11672, opened 2016-01-22
Using remote instead of remote-3 for custom domain serving on amp

NataliaLKB:
The code added in `remote-3.html` is currently breaking other iframes in AMP pages. For example, `amp-twitter` is not working right now.

Lets go back to the basic version recommeded by the AMP team over the weekend, and then on Monday come up with a revised solution for krux.

cc @kenlim @stephanfowler

### DIFF
--- a/amp/remote.html
+++ b/amp/remote.html
@@ -5,7 +5,7 @@
     <script>
         (function() {
             var v = location.search.substr(1);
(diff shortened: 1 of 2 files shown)


stephanfowler:
:+1:

PR #11673, opened 2016-01-22
Add /email route to generate article email templates

desbo:
At the moment, we're using R2 for paths like http://www.theguardian.com/football/series/thefiver/latest/htmlemail to generate emails for the latest article in a series.

These changes bring that functionality to the current codebase by:
1. adding an `/latest/email` path to series, which redirects you to...
2. a new `/email` path on articles, which gives you markup suitable for injection into an email template (i.e. it uses tables and is painful to write). 

For the new email template there's a `InlineStyles` object which takes any of the CSS from the `<style>` tag in that page and applies the rules to the appropriate element's `style` attribute. This probably needs more attention but it's okay for simple stylesheets. 

The added benefit of all this is that we can generate emails for any article, similar to how `/amp` works. We'll probably test this on The Fiver initially, but there's a bit more to do on the templates first (I'll do that in a separate PR, this one is big enough already).

3dmail-1_e0


crifmulholland:
This is amazing

common/app/views/emailEmbed.scala.html
```diff
@@ -0,0 +1,30 @@
+@(metaData: model.Page)(body: Html)(implicit request: RequestHeader)
+
+@import conf.switches.Switches._
+@import common.InlineJs
+@import templates.inlineJS.blocking.js._
+
+<!doctype html>
+<head>
+    @if(play.Play.isDev() || !InlineCriticalCss.isSwitchedOn) {
```
gtrufitt: I guess you wouldn't want this switch to prevent putting the css int the head, because your inlining would then break?

article/app/views/fragments/emailArticleBody.scala.html
```diff
@@ -0,0 +1,20 @@
...
+@import conf.switches.Switches._
+
+@defining(page.article) { article =>
+    @fragments.headTonal(article, article.commercial.isAdvertisementFeature && NewCommercialContent.isSwitchedOn, showMeta = false)
+
+    @fragments.emailMainMedia(article)
+
+    @fragments.contentMeta(article, showExtras = false)
+
+    <div class="content__article-body from-content-api" itemprop="@if(article.tags.isReview){reviewBody} else {articleBody}"
```
gtrufitt: Very minor & for future updates: I am not sure whether itemProp would be useful within the email template but maybe some like the [Schema](https://developers.google.com/schemas/reference/) would work nicely in the email (orani**z**ation for example).

gtrufitt:
This is really cool! 🆒A couple of minor things but :+1:

common/app/views/emailEmbed.scala.html
desbo: hmm, I didn't write this, this file was just renamed from `mainEmail.scala.html` from the earlier email work - don't think it's anything that'd be processed by the inliner.

article/app/views/fragments/emailArticleBody.scala.html
desbo: thanks, I'll add some schema properties when I do the template.

applications/app/controllers/LatestIndexController.scala
```diff
@@ -15,7 +15,11 @@ object LatestIndexController extends Controller with ExecutionContexts with impl
   def latest(path: String) = Action.async { implicit request =>
     loadLatest(path).map { _.map { index =>
       index.page match {
-        case tag: Tag if tag.isSeries || tag.isBlog => index.trails.headOption.map(latest => Found(latest.metadata.url)).getOrElse(NotFound)
+        case tag: Tag if tag.isSeries || tag.isBlog => index.trails.headOption.map(latest => {
+          if (request.isEmail) Found(latest.metadata.url + "/email")
```
crifmulholland: We just need to check whether Exact Target follows this before scraping, or whether we need to keep the url masked. The best way to find out will be release anyways!

johnduffell:
@stephanfowler noticed apps issues with using /amp on the end of amp urls so was thinking about having a common url format for rendering articles in another way (e.g ?format=amp)  I notice you're following an amp stle of adding extra parts to the path, but it might be worth a chat with stephan

desbo:
@johnduffell to note the latest on that, we're sticking with /email for this and amp is probably moving to its own subdomain.

PR #11674, opened 2016-01-22
Fix campaign logos for new sponsored content design

jbreckmckye:
This pull request implements improvements in the layout and appearance of sponsor logos when the new design for paid-for content is enabled. This will go out on Monday.

Screenshots (see the logos in the bottom right of each block):

Container with a single campaign, desktop:
image

Container with a single campaign, tablet:
image

Container with a single campaign, mobile:
image

Container with multiple campaigns, desktop:
image

Container with multiple campaigns, tablet:
image

Container with multiple campaigns, mobile:
image

### DIFF
--- a/static/src/stylesheets/module/commercial/_paidfor.scss
+++ b/static/src/stylesheets/module/commercial/_paidfor.scss
@@ -135,7 +135,7 @@
 
 .commercial--paidfor {
 
-    /* Resets for the A/B test. When (and if) the new design is implemented, the old code
+    /* Resets for new design. When (and if) the new design is committed to, the old code
        will be torn down and these resets should no longer be necessary
     */
 
--- a/static/src/stylesheets/module/commercial/_paidfor.scss
+++ b/static/src/stylesheets/module/commercial/_paidfor.scss
@@ -130,6 +130,90 @@
     }
 }
 
+/* Badges
+   ========================================================================== */
+
+.commercial--paidfor {
+
+    /* Resets for the A/B test. When (and if) the new design is implemented, the old code
+       will be torn down and these resets should no longer be necessary
+    */
+
+    .ad-slot--paid-for-badge__help {
+        display: none;
+    }
+
+    .ad-slot--paid-for-badge--front {
+        border: 0;
+        padding: 0;
+        margin: 0;
+    }
+
+    .ad-slot--paid-for-badge .ad-slot--paid-for-badge__link {
+        margin: 0;
+        float: none;
+    }
+
+    .ad-slot--paid-for-badge__link:after {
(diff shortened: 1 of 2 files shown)


static/src/stylesheets/module/commercial/_paidfor.scss
```diff
@@ -130,6 +130,90 @@
...
+    }
+
+    .ad-slot--paid-for-badge__link:after {
+        content: none;
+    }
+
+    .ad-slot.ad-slot--paid-for-badge--front {
+        margin: auto;
+        width: auto;
+        min-height: unset;
```
regiskuckaertz: I'm not sure `unset` is supported by any browser. Might be preferable to use `0` instead.

regiskuckaertz:
:+1: not sure about the `unset` though, the behaviour in this case might be to fall back to `inherit` or `initial` which is not what you want.

static/src/stylesheets/module/commercial/_paidfor.scss
jbreckmckye: Good point

uplne: Yeah this is very interesting. Unset is definitely mentioned in the documentation but the initial value of min-height is 0. Would be interesting to know if it will work as expected across all browsers. Have you tested it on other browsers?

static/src/stylesheets/module/commercial/_paidfor.scss
```diff
@@ -130,6 +130,90 @@
...
+        max-width: 100%;
+        max-height: 100%;
+        vertical-align: middle;
+        display: inline-block;
+    }
+
+    .ad-slot--paid-for-badge__link:before {
+        content: '';
+        height: 100%;
+        vertical-align: middle;
```
uplne: This almost calls for some %placeholder

static/src/stylesheets/module/commercial/_paidfor.scss
jbreckmckye: 0 seems like a safe option; everything that supports `min-height` should apply zero properly. Roger Johannson recommends it too: https://twitter.com/rogerjohansson/status/324947861148164096

uplne:
Go for it :+1:

static/src/stylesheets/module/commercial/_paidfor.scss
jbreckmckye: Yeah, is there a mixin for this already?

uplne: Unfortunately I don't know.

PR #11675, opened 2016-01-22
Point GLabs 'About...' CTA to "content funding" page

jbreckmckye:
This is another change for the upcoming sponsored content redesign. It points the URI of the 'Learn more about...' link to a new page (http://www.theguardian.com/content-funding, which doesn't exist yet).

image

This change only applies to templates used when the NewCommercialContent switch is on.


regiskuckaertz:
:+1:

PR #11676, opened 2016-01-25
Use secureFile for images and audio and embedded media

NataliaLKB:
Videos are now using SecureFile field if available. Audio, Images and other embedded media should do the same

cc @rich-nguyen


rich-nguyen:
:+1:

PR #11678, opened 2016-01-25
Sponsored CAPI component restyling

uplne:
After:
screen shot 2016-01-25 at 10 56 58

It is behind the switch so till the switch is off it looks like old CAPI component.
It uses @jbreckmckye amazing work on badges position so it works like a magic. I just needed to add '.commercial--dfp-single' into the equation.

TODO: after launch we will have to do massive clean up of redundant code which is behind the switch

CC @jbreckmckye @regiskuckaertz

### DIFF
--- a/commercial/app/views/contentapi/item.scala.html
+++ b/commercial/app/views/contentapi/item.scala.html
@@ -36,18 +36,6 @@
                     <span class='u-h'>Guardian Labs</span>
                 </a>
             </div>
-
-            <div class="ad-slot--adbadge ad-slot--paid-for-badge ad-slot--paid-for-badge--front">
-                <div class="ad-slot--paid-for-badge__inner">
-                    <h3 class="ad-slot--paid-for-badge__header">@optSponsorLabel:</h3>
-                    <a href="@optClickMacro@for(linkUrl <- optCapiLink) {@linkUrl}" class="ad-slot--paid-for-badge__link" data-link-name="logo link">
-                        @for(logoUrl <- optLogo) {<img src="@logoUrl" alt="" class="ad-slot--paid-for-badge__logo" />}
-                    </a>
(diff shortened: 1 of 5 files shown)


regiskuckaertz:
Shouldn't the colon after `paid for by` disappear?

uplne:
@regiskuckaertz nice catch. Fixed.

commercial/app/views/contentapi/item.scala.html
```diff
@@ -1,8 +1,83 @@
...
+    @if(NewCommercialContent.isSwitchedOn) {
+    <div class="commercial commercial--dfp-single paid-content--advertisement-feature commercial--tone-paidfor" data-link-name="merchandising | capi | single @optOmnitureId">
+        <div class="commercial__inner">
+            <div class="commercial__header commercial--paidfor">
+                <div class="paidfor-meta">
+                    <span class="paidfor-meta__label">Paid Content</span>
+                    <div class="paidfor-label paidfor-meta__more has-popup">
+                        <button class="u-button-reset paidfor-label__btn popup__toggle" data-toggle="paidfor-popup--capi-component">About @inlineSvg("arrow-down", "icon")</button>
+                        <div class="popup is-off paidfor-popup paidfor-popup--capi-component">
+                            <h3 class="paidfor-popup__title">Paid stories are paid for and approved by an advertiser and created by the Guardian Labs commercial editorial team</h3>
```
regiskuckaertz: The popup content has been updated, you should check it out (text+link)

uplne: @regiskuckaertz I've updated the text and created card so we should look at it one more time to possibly create fragment for this as it is across multiple components as you said.

regiskuckaertz:
:+1: once the link is fixed yeah!

PR #11680, opened 2016-01-25
GLabs links to Australia

jbreckmckye:
The GLabs logo should point to the GLabs Australia pages when the user is reading the Australian edition (or as I prefer to call it, "upside-down mode"). This link was missed from a previous PR.

### DIFF
--- a/common/app/views/fragments/containers/facia_cards/commercialContainer.scala.html
+++ b/common/app/views/fragments/containers/facia_cards/commercialContainer.scala.html
@@ -104,7 +104,16 @@
                                 </div>
                             </div>
                             <h3 class="commercial__title">@container.displayName</h3>
-                            <a class="commercial__cta" href="@LinkTo("/guardian-labs")">@inlineSvg("glabs-logo", "logo")</a>
+                            <a class="commercial__cta"
+                                @if(Edition(request).id == "AU") {
+                                    href="@LinkTo("/guardian-labs-australia")"
+                                } else {
+                                    href="@LinkTo("/guardian-labs")"
+                                    }
+                            >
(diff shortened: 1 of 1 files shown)


common/app/views/fragments/containers/facia_cards/commercialContainer.scala.html
```diff
@@ -104,7 +104,16 @@ <h3 class="paidfor-popup__title">Paid stories are paid for and controlled by an
...
-                            <a class="commercial__cta" href="@LinkTo("/guardian-labs")">@inlineSvg("glabs-logo", "logo")</a>
+                            <a class="commercial__cta"
+                                @if(Edition(request).id == "AU") {
+                                    href="@LinkTo("/guardian-labs-australia")"
+                                } else {
+                                    href="@LinkTo("/guardian-labs")"
+                                    }
+                            >
+                                @inlineSvg("glabs-logo", "logo")
+                                <span class='u-h'>Guardian Labs</span>
```
regiskuckaertz: Food for thought: that kind of hidden label should really be in the SVG as a `title` element. Much more accessible and reusable.

regiskuckaertz:
:+1:

common/app/views/fragments/containers/facia_cards/commercialContainer.scala.html
jbreckmckye: We could embed a title in the SVG itself, but I can think of two drawbacks:
1. We probably always want a title on our links, so that ATs read them out promptly when a user tabs over navigation points on the page. Adding an SVG title as well would mean inundating an AT user with repetitive information.
2. Embedding a title in our SVGs might make dynamic internationalization harder.

PR #11681, opened 2016-01-25
Sponsorship: video height fix

uplne:
As much as I don't like it I had to do some massive CSS specificity override.

Before:
screen shot 2016-01-25 at 11 41 09

After:
screen shot 2016-01-25 at 11 40 49

@regiskuckaertz @jbreckmckye


jbreckmckye:
:+1: It's a shame we have to escalate specificity, but at least the selector indicates its intent.

PR #11682, opened 2016-01-25
Hide popup label to prevent broken layout

regiskuckaertz:
This will only happen when the viewport is between `leftCol` and `wide`.

Before (top left):
picture 3

After:
picture 1

### DIFF
--- a/static/src/stylesheets/module/commercial/_paidfor.scss
+++ b/static/src/stylesheets/module/commercial/_paidfor.scss
@@ -31,13 +31,9 @@
         overflow: visible;
     }
 
-    .paidfor-meta__more .popup__toggle__label {
-        @include mq(leftCol) {
-            display: none;
-        }
-
-        @include mq(wide) {
-            display: inline;
+    .paidfor-meta__more > .popup__toggle {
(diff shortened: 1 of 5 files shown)


jbreckmckye:
Would we actually want to hide the about link in this instance? Perhaps it should just appear on a second line without the left margin.

regiskuckaertz:
Alright, doing that then.

regiskuckaertz:
That way.

picture 1

static/src/stylesheets/module/commercial/_paidfor.scss
```diff
@@ -31,6 +31,12 @@
         overflow: visible;
     }
 
+    .paidfor-meta__more > .popup__toggle {
```
uplne: You know that this is one of the most expensive selectors

uplne:
:+1:

static/src/stylesheets/module/commercial/_paidfor.scss
jbreckmckye: Surely, that's less expensive than `.foo .bar`?

PR #11686, opened 2016-01-25
Revert "Introducing Toast: new liveblog autoupdates"

holder-josh:
Toast works, but having some issues with liveblogs when not opted in to the 0% test.


uplne:
:+1:

rich-nguyen:
:cry:

Jholder112233:
@rich-nguyen think I just deleted a CSS module the non-test users are still using, will be back shortly :wink:

PR #11688, opened 2016-01-25
Move the popup to the right between leftCol and wide breakpoints

regiskuckaertz:
Before:
picture 1

After:
picture 2


jbreckmckye:
:+1:

PR #11689, opened 2016-01-25
Reduce size of GLabs container text

jbreckmckye:
Go from this...
image

to this...
image


regiskuckaertz:
Much better! Tight columns means tighter leading. :+1:

PR #11690, opened 2016-01-25
Enable HTTPS on /science

desbo:
See also: https://github.com/guardian/fastly-edge-cache/pull/549

### DIFF
--- a/common/app/common/LinkTo.scala
+++ b/common/app/common/LinkTo.scala
@@ -28,7 +28,7 @@ trait LinkTo extends Logging {
   /**
     * email is here to allow secure POSTs from the footer signup form
     */
-  val httpsEnabledSections: Seq[String] = Seq("info", "email")
+  val httpsEnabledSections: Seq[String] = Seq("info", "email", "science")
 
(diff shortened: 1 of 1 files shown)


Jholder112233:
This is sadly going to break all of these interactives http://www.theguardian.com/science/science+content/interactive even the most recent one has a lot of http links.

What's the reason for the move?

desbo:
@Jholder112233 commercial are using it to measure the impact of HTTPS on ads. I've checked with them and it's fine to exclude interactives from the test - will keeping everything under `science/ng-interactive` on regular HTTP be enough?

Jholder112233:
@desbo that will cover all of the interactives served by NGW, there are still some on R2. Not sure if your change will affect those? They live on science/interactive

johnduffell:
awesome that https is moving forward!

desbo:
@Jholder112233 I've updated the fastly PR to not apply any of the HTTPS rules to interactives URLs.
Could you take a look and +1 here and there if you're happy?

Jholder112233:
:+1:

PR #11691, opened 2016-01-25
validate email input (report abuse non-ajax)

gidsg:
I forgot to validate the email input before, so this comparison wasn't working previously.
CC @guardian/discussion


mchv:
:+1:

PR #11696, opened 2016-01-25
Changing height for badge

uplne:
This is just to make it a little bit nicer before we find some long term solution.

Before:
screen shot 2016-01-25 at 17 44 59

After:
screen shot 2016-01-25 at 17 44 00


regiskuckaertz:
:+1:

PR #11698, opened 2016-01-26
Changed text sponsored by to supported by

uplne:
From here:
screen shot 2016-01-26 at 11 37 25

To here:
screen shot 2016-01-26 at 11 37 28


regiskuckaertz:
:+1:

PR #11699, opened 2016-01-26
Reduce height of soulmates component on large screens

regiskuckaertz:
Before:
picture 3

After:
picture 4


static/src/stylesheets/module/commercial/_soulmates.scss
```diff
@@ -19,7 +19,7 @@ $c-soulmates-high: colour(neutral-1);
         }
 
         @include mq(wide) {
-            min-height: 260px;
+            min-height: 230px;
```
uplne: Magic numbers. Can we use something like $gs-row-height * 6 or 6.5?

uplne:
Other than comment :+1:

PR #11700, opened 2016-01-26
live blog block-level permalinks

johnduffell:
This PR removes integer page numbers ?page=1 and replaces them with block page numbers ?page=with:block-1a2b3c4d5e6f7890.

The next/previous page buttons will always reference the first block on the page.  The permalinks on the date blocks refer to the current block specifically.

Still to do: share buttons, key events

Short urls (gu.com) actually link to the whole liveblog.  I don't propose to introduce block level short urls, as they are primarily for use in the printed paper anyway.  Twitter can shorten its own urls, so it will actually take out one redirect from the equation if we use the real urls.


NataliaLKB:
:+1:

PR #11701, opened 2016-01-26
Added a test to make sure sponsored merch components don't use the new design

regiskuckaertz:
Before:
picture 1

After:
picture 2

### DIFF
--- a/commercial/app/views/contentapi/items.scala.html
+++ b/commercial/app/views/contentapi/items.scala.html
@@ -4,7 +4,7 @@
 @import views.html.fragments.inlineSvg
 
 @defining(("1_0", "2014-08-14")) { case (version, date) =>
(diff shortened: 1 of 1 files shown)


commercial/app/views/contentapi/items.scala.html
```diff
@@ -4,7 +4,7 @@
 @import views.html.fragments.inlineSvg
 
 @defining(("1_0", "2014-08-14")) { case (version, date) =>
-    @if(NewCommercialContent.isSwitchedOn) {
+    @if(NewCommercialContent.isSwitchedOn && optSponsorType.getOrElse("") != "fc-container--sponsored") {
```
kelvin-chappell: I think this should work: `optSponsorType.contains("fc-container-sponsored")`

uplne:
:+1:

PR #11705, opened 2016-01-26
Add a try catch to Rubicon listener in new sticky ad test

holder-josh:
Adds a try catch to Rubicon listener in new sticky ad test to prevent errors when non-json DFP events are caught by this listener.

@OliverJAsh, is this what you meant?


OliverJAsh:
Yup! :+1:

PR #11709, opened 2016-01-26
Fix broken About button in commercial components

regiskuckaertz:
The initial call to `Toggles` does not account for lazy loaded components, thus we must call it again in case there's a popup in the content returned from `/commercial`

@jbreckmckye @uplne

### DIFF
--- a/commercial/app/views/debugger/allcreatives.scala.html
+++ b/commercial/app/views/debugger/allcreatives.scala.html
@@ -645,7 +645,9 @@
                                 omnitureId: '[%OmnitureID%]'
                             });
 
-                            commercialComponent.postLoadEvents['capi@if(component("type") == "single"){Single}'] = function () {
+                            var oldFn = commercialComponent.postLoadEvents['capi@if(component("type") == "single"){Single}'];
+                            commercialComponent.postLoadEvents['capi@if(component("type") == "single"){Single}'] = function (el) {
+                                oldFn.call(commercialComponent, el);
                                 $capiType.removeClass('lazyloaded');
                             };
                             commercialComponent.create();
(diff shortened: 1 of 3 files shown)


static/src/javascripts/projects/common/modules/commercial/creatives/commercial-component.js
```diff
@@ -93,10 +95,18 @@ define([
             };
         };
 
+    function createToggle(el) {
+        if (el.querySelector('.popup__toggle')) {
+            new Toggles(el).init();
```
jbreckmckye: Does this component short circuit if you try initializing a toggle twice?

regiskuckaertz: Yes, this is taken care of in the `Toggles` module.

commercial/app/views/debugger/allcreatives.scala.html
```diff
@@ -645,7 +645,9 @@ <h2 class="test-page-head">
                                 omnitureId: '[%OmnitureID%]'
                             });
 
-                            commercialComponent.postLoadEvents['capi@if(component("type") == "single"){Single}'] = function () {
+                            var oldFn = commercialComponent.postLoadEvents['capi@if(component("type") == "single"){Single}'];
```
jbreckmckye: In this context I'm not sure if `oldFn`'s meaning is immediately obvious, at least until you read a little further. What about `originalPostLoadEvents`?

regiskuckaertz:
@jbreckmckye thanks, I had the same feeling too :)

jbreckmckye:
Ready to go :+1:

PR #11711, opened 2016-01-26
Make sure logo badge fits inside its div

uplne:
Another massive change.

Before:
screen shot 2016-01-26 at 15 41 04

After:
screen shot 2016-01-26 at 15 44 38


jbreckmckye:
:+1:

PR #11712, opened 2016-01-26
Sponsored audio articles have layout bugs

jbreckmckye:
This is a fix to sponsored audio articles. 

image

It comprises two changes:
1. An empty standfirst created a large empty space between the media player and the paragraph below,
2. When the large space was removed, the 'Save for Later' button was cut off. This is because that cell is absolutely positioned, and the container uses a min-height to reserve space for it.

_After:_
image

I've changed three things:
1. We now only insert the standfirst on the mediaBody is it's actually needed.
2. I've removed an old comment which suggested that we needed to inject an empty standfirst for the sake of scripting; I couldn't find any evidence of this by grepping scripts (*.js, *scala.js) for 'standfirst'.
3. I've bumped up the min-height of the audio content container when the sidebar has a sponsorship logo to show.

### DIFF
--- a/static/src/stylesheets/module/content/_media.global.scss
+++ b/static/src/stylesheets/module/content/_media.global.scss
@@ -79,7 +79,7 @@
 
     // increase height if we have sponsorship logos
     &.paid-content--advertisement-feature .content__main-column--audio {
-        min-height: 325px;
+        min-height: $gs-row-height * 9;
     }
 }
 .content__headline:before {
--- a/static/src/stylesheets/module/content/_media.global.scss
+++ b/static/src/stylesheets/module/content/_media.global.scss
@@ -70,9 +70,17 @@
         }
     }
 }
-// if it's audio and there's body text, we display an advert
-.content--has-body .content__main-column--audio {
-    min-height: 274px;
+
+.content--has-body {
+    // if it's audio and there's body text, we display an advert
+    .content__main-column--audio {
+        min-height: 274px;
+    }
+
+    // increase height if we have sponsorship logos
+    &.paid-content--advertisement-feature .content__main-column--audio {
+        min-height: 325px;
+    }
 }
 .content__headline:before {
     .content--media--audio & {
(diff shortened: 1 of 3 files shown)


static/src/stylesheets/module/content/_media.global.scss
```diff
@@ -70,9 +70,17 @@
...
     }
 }
-// if it's audio and there's body text, we display an advert
-.content--has-body .content__main-column--audio {
-    min-height: 274px;
+
+.content--has-body {
+    // if it's audio and there's body text, we display an advert
+    .content__main-column--audio {
+        min-height: 274px;
```
regiskuckaertz: @uplne would suggest using `$gs-row-height` here :-)

static/src/stylesheets/module/content/_media.global.scss
```diff
@@ -70,9 +70,17 @@
...
+    }
+
+    // increase height if we have sponsorship logos
+    &.paid-content--advertisement-feature .content__main-column--audio {
```
regiskuckaertz: This is a matter of style I guess, but you could save a few strokes this way:

```
.content--has-body {
  .content__main-column--audio {
    .paid-content--advertisement-feature& { ... }
  }
}
```

regiskuckaertz:
Anyway, great job! :+1:

static/src/stylesheets/module/content/_media.global.scss
jbreckmckye: I tried that; it didn't seem to work. I assumed you must just not be able to modify parent selectors in Sass.

static/src/stylesheets/module/content/_media.global.scss
jbreckmckye: You are right.

static/src/stylesheets/module/content/_media.global.scss
jbreckmckye: I had tried that; the compiler didn't seem to like using the parent selector with a modifier.

PR #11713, opened 2016-01-26
liveblog permalinks second go

johnduffell:
I managed to lose the page test for bringing new live blocks, this meant that blocks were coming in whichever page I was on. @Jholder112233 make sure I've got the condition right in liveblog.js


static/src/javascripts/bootstraps/enhanced/liveblog.js
```diff
@@ -153,7 +153,7 @@ define([
             if (config.page.isLive) {
                 if (ab.isInVariant('LiveblogToast', 'toast')) {
                     AutoUpdateNew();
-                } else {
```
Jholder112233: :+1: for this

PR #11714, opened 2016-01-26
Invalidate all identity csrf cookies

danmassie:
Due to the application secret being rotated users are experiencing issues accessing pages that check csrf cookies. Existing cookies are not currently removed on sign out and existing cookies signed with the old secret are resulting in 500s. This updates the sign out action to discard the csrf cookie and renames the cookie to effectively remove it for all users.

### DIFF
--- a/identity/app/controllers/SignoutController.scala
+++ b/identity/app/controllers/SignoutController.scala
@@ -43,9 +43,10 @@ class SignoutController @Inject()(returnUrlVerifier: ReturnUrlVerifier,
       DiscardingCookie("gu_paying_member", "/", Some(conf.id.domain), secure = false)
       )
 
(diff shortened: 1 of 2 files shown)


jamespamplin:
:+1: looks good to me

rich-nguyen:
:+1:

rich-nguyen:
Sounds like we should update the behaviour so that this doesn't happen again. Maybe it's possible to sever the dependency on the app secret? Identity could have it's own app secret, for example.

danmassie:
I'll create a ticket for that Rich.

rich-nguyen:
thanks @danmassie, add me as a contact for that one, I can give a few pointers.

PR #11715, opened 2016-01-26
Fixed the case where we get a Bonzo object

regiskuckaertz:
In dfp-api, the constructor if given a Bonzo object, while everywhere else it's getting a DOM element.


uplne:
:+1:

jbreckmckye:
:ship:

PR #11716, opened 2016-01-26
The Minute. Initial minimal template.

gtrufitt:
The Minute is a new template type, initially the format will be used for US election coverage. This PR contains the initial architecture for the new template type and a very basic version of the design. We will be iterating on this heavily over the next few weeks but for the sake of keeping PRs small(-ish) and having the beginnings of the US Campaign Minute visual aesthetic we have this inital PR.

The minute re-uses the live blog editor, providing us with the blocks format. In the initial architecture @desbo has transformed the live blog mark-up for use with the new template. You will notice liveblog classes and elements at this stage but the next step will be to make the minute template distinct and refactor components for agnostic use.

inital_mobile

inital_wide

cc/ @superfrank @rich-nguyen


static/src/javascripts/projects/common/modules/article/spacefinder.js
```diff
@@ -175,7 +175,7 @@ define([
         var body;
 
         rules = rules || defaultRules;
-        body = rules.bodySelector ? document.querySelector(rules.bodySelector) : document;
+        body = (rules.bodySelector && document.querySelector(rules.bodySelector)) || document;
```
gtrufitt: Minor change but don't want to break ads! Quick thoughts @regiskuckaertz ?

regiskuckaertz: oh interesting, I rarely come across such syntactic fine-tuning: is it more efficient? you could do without the brackets but it would hinder readability.

gtrufitt: Now I think about this, I am probably hiding another problem - `document.querySelector(rules.bodySelector)` was returning null, I guess because this JS was being run maybe when it shouldn't (live blog template). Efficiency wise, [the tenerary looks faster](https://jsperf.com/ternary-vs-double-ampersand-and-or), so that's even more reason to find the root cause.

regiskuckaertz: hmm I see. Is the `.js-liveblog-body` class still applied to the body of the liveblog?

gtrufitt: Yeah, it is, I'll strip that out, thanks for pointing me to that :)

-- 

in England Number 908396

regiskuckaertz: Cheers! But I think there might be something else leading to it. If *the spacefinder is running* (saying this I feel like Mr Spock) that means the JS is either trying to insert adverts or rich links in the page. Both may be looking for a `.js-article__body` instead, so you might want to disable them in The Minute templates.

Great work by the way, it's looking hot!

gtrufitt: Thanks! Yeah, so I had a quick chat with @jbreckmckye and he pointed me towards the commercial-feature-policies so I added a new policy for the minute and turned off all the money makers :dancer: (as the plan is for this to be sponsored so we need dfp, but not any of the in-article modules). Thanks for your help!

static/src/javascripts/projects/common/modules/commercial/commercial-feature-policies.js
```diff
@@ -34,6 +34,21 @@ define([
         }
     };
 
+    policies.minuteArticle = function () {
```
jbreckmckye: :sunglasses:

common/app/model/content.scala
```diff
@@ -334,6 +334,7 @@ object Article {
     val javascriptConfig: Map[String, JsValue] = Map(
       ("contentType", JsString(contentType)),
       ("isLiveBlog", JsBoolean(content.tags.isLiveBlog)),
+      ("isMinuteArticle", JsBoolean(content.tags.isUSMinuteSeries)),
```
rich-nguyen: Does every content item need this value set? This seems like a job for `conditionalConfig`

common/app/model/meta.scala
```diff
@@ -238,7 +238,8 @@ final case class MetaData (
   customSignPosting: Option[NavItem] = None,
   javascriptConfigOverrides: Map[String, JsValue] = Map(),
   opengraphPropertiesOverrides: Map[String, String] = Map(),
-  twitterPropertiesOverrides: Map[String, String] = Map()
+  twitterPropertiesOverrides: Map[String, String] = Map(),
+  isMinute: Boolean = false
```
rich-nguyen: Again, I don't think `isMinute` is worth adding as metadata to the top-level of content taxonomy

common/app/model/content.scala
```diff
@@ -376,7 +377,8 @@ object Article {
       iosType = Some("Article"),
       javascriptConfigOverrides = javascriptConfig,
       opengraphPropertiesOverrides = opengraphProperties,
-      twitterPropertiesOverrides = twitterProperties
+      twitterPropertiesOverrides = twitterProperties,
+      isMinute = content.tags.isUSMinuteSeries
```
rich-nguyen: A bit inconsistent here, is it `USMinute` or is it `Minute`? What will it be in the future?

common/app/model/content.scala
gtrufitt: A minute article is along the same lines (template/content wise) as a live blog - it is designed as a new template but the format, in terms of frontend templates, css, js, composer (eventually), will be at the same level of difference from an article as a liveblog. With that in mind, what are the guidelines / advantages of using `conditionalConfig` for this and what makes that usage make sense in this case but not with  `isLiveBlog`?

common/app/model/content.scala
gtrufitt: Thoughts @desbo - shall we make the tag isMinute and add a new identifier (not sure where) for the specific campaign minute type (so we can add usMinute classes?). We mentioned refactoring this later on, but maybe we should update it now?

desbo: Ultimately we want to mark a piece of content with both the general Minute identifier and one for the specific topic it covers. Right now all we have is the US campaign minute series tag so we should probably just treat everything as `USMinute` until we have something more general.

common/app/model/content.scala
rich-nguyen: yes, `isLiveBlog` is at the wrong level

gtrufitt: Cool.

crifmulholland:
This looks great!

rich-nguyen:
:+1:

superfrank:
ql8bnjm

PR #11717, opened 2016-01-26
Add election 2016 to the US nav bar

jennysivapalan:
As requested by the US office:

screen shot 2016-01-26 at 18 07 13


crifmulholland:
Looks like we'll have to drop something from it too - can't really have it scrolling to the side on desktoo

jennysivapalan:
Okay as discussed I have removed science (:cry:) from the brief nav for US edition

screen shot 2016-01-27 at 13 54 51

JustinPinner:
Personally, I'd have dumped Fashion instead. Oh well :+1:

PR #11719, opened 2016-01-27
Change healthcheck temporarily because problems with /uk

kelvin-chappell:


gklopper:
:-1: :smile:

kelvin-chappell:
No longer needed.  Grant has found the problem.

PR #11721, opened 2016-01-27
Remove Egyptian from paid content fronts

jbreckmckye:
Containers on paid content fronts use the Egyptian font for sublinks, when everything on those pages should be styled with Guardian sans:

image

This dangerous and complicated pull request uses something named "CSS" to change the "font" of the "text".

image

@regiskuckaertz


regiskuckaertz:
:+1: let's be fearless and launch this beast in the wild

PR #11722, opened 2016-01-27
Fixing TakeoverEmptyMPUs bug

kenlim:
It was previously possible to create an TakeoverEmptyMPU entry that pointed to www.theguardian.com. This broke the app. 

We've implemented a check in the form with better messaging to prevent new entries. 

And we've implemented a filter to stop existing entries from breaking the app.


JonNorman:
:+1:

JonNorman:
:+1:

PR #11723, opened 2016-01-27
Harmonise padding in paidfor band

regiskuckaertz:
Before:
picture 3

After:
picture 4

I know this is not the place nor the time, but it's sad that paddings are not the same on fronts/content pages.


jbreckmckye:
:+1:

PR #11724, opened 2016-01-27
format defines nicely

sndrs:
just scratching a silly itch (hangover from #11066)

before

``` javascript
define([
    'lodash/objects/isEqual'
], function (
    isEqual) {
```

now

``` javascript
define([
    'lodash/objects/isEqual'
], function (
    isEqual
) {
```

nothing major, it just always bugs me :)


jbreckmckye:
:star: It always annoyed me, too.

PR #11725, opened 2016-01-27
Increase log line exceptions to improve diagnosing of issues

jennysivapalan:
Following @gklopper example https://github.com/guardian/frontend/pull/11718/files  I've increased the number of log lines reported in exceptions from 3 to 15 across the board.

Admin seemed to never have a limit but I've made it consistent with other applications.


rich-nguyen:
:+1:

johnduffell:
personally I was going to change to unlimited, why stop at 15?

PR #11727, opened 2016-01-27
Remove unused traits which should fix the flaky test build.

jennysivapalan:
`WidthsByBreakpointTest` fails in Teamcity on 
`
sbt.ForkMain$ForkError: org.jboss.netty.channel.ChannelException: Failed to bind to: /0.0.0.0:19001
`

This test looks to be a unit test so doesn't need an app server, I've removed it and it ran locally fine. Hopefully the PR will build in Teamcity too.


johnduffell:
doh didn't spot that it didn't even need it! :+1: I think it's caused by them running in parallel and it using a default port

PR #11728, opened 2016-01-27
AB test: Update userzoom survey audience & message

gtrufitt:
Update AB test name to ensure `notintest` is reset, localstorage will still be checked for users that have seen the message.


desbo:
Plus One

PR #11730, opened 2016-01-27
clear up console by not running code that we know will fail in dev

sndrs:
it's become pretty tricky to find errors in the console in dev.

this prevents code from running that we know cannot succeed in dev, while trying to make it plain how to re-enable it if you're working on it (specifically the service worker)

before:
screen shot 2016-01-27 at 16 19 41

after:
screen shot 2016-01-27 at 16 18 10

### DIFF
--- a/common/app/views/fragments/head.scala.html
+++ b/common/app/views/fragments/head.scala.html
@@ -10,7 +10,9 @@
 @* stylesheet <link>s - get the stylesheets downloading ASAP *@
 @fragments.stylesheets(projectName, getContent(page).exists(_.tags.isCrossword))
 
-<link rel="prefetch" href="@Static("javascripts/app.js")">
+@if(!play.Play.isDev()) {
+    <link rel="prefetch" href="@Static("javascripts/app.js")">
+}
 
 @*
 Protect against ReferenceErrors in IE:
--- a/static/src/javascripts/bootstraps/enhanced/main.js
+++ b/static/src/javascripts/bootstraps/enhanced/main.js
@@ -124,9 +124,9 @@ define([
             });
         }
 
-        var isDev = window.location.hostname === 'localhost';
-        if ((window.location.protocol === 'https:' || isDev)
-            && config.page.section !== 'identity') {
+        // invert the comments on the following two lines to force service worker registration for local dev
+        // if (true) {
+        if (window.location.protocol === 'https:' && config.page.section !== 'identity') {
             var navigator = window.navigator;
             if (navigator && navigator.serviceWorker) {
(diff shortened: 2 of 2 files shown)


OliverJAsh:
Hey! Could you post a full screenshot of the console beforehand so I can see the full error messages? I would like to understand what exactly is causing these errors. I don't have the same errors but that's likely because my environment is different (I'm not signed in, etc.)

I'm happy to disable the SW in dev, but if I can understand the exact issues, I will see if I can clean them up in dev without disabling the SW altogether. The SW should work in dev but it shouldn't be as noisy as it is.

I believe push notifications are right around the corner so the SW may need to be enabled in dev for that, I'm not sure.

:+1:

common/app/views/fragments/head.scala.html
```diff
@@ -10,7 +10,9 @@
 @* stylesheet <link>s - get the stylesheets downloading ASAP *@
 @fragments.stylesheets(projectName, getContent(page).exists(_.tags.isCrossword))
 
-<link rel="prefetch" href="@Static("javascripts/app.js")">
+@if(!play.Play.isDev()) {
+    <link rel="prefetch" href="@Static("javascripts/app.js")">
+}
```
OliverJAsh: Which console errors was this causing for you, out of curiosity?

desbo:
what if we used a query param or cookie to enable SW in dev rather than having to flip those comments - can see that slipping into a few git commits unintentionally

sndrs:
this is the full console 

screen shot 2016-01-27 at 17 02 17

common/app/views/fragments/head.scala.html
sndrs: it triggers one of these

screen shot 2016-01-27 at 17 04 15

sndrs:
> what if we used a query param or cookie to enable SW in dev rather than having to flip those comments - can see that slipping into a few git commits unintentionally

yeah i wondered about that too, maybe you're right

OliverJAsh:
Thanks. Looks like you need to be logged in to reproduce this, so I'll give that a go.

johnduffell:
it's a great idea to clean things up in dev, I've had trouble with this too and it's difficult to get confidence that my changes are good without this. :+1: I do also have discussion errors due to the fact I don't have it set up locally which I guess are separate.

sndrs:
@desbo updated to use `#force-sw` instead of editing the source to force it in local.

happy to see this reverted if we can sort out the errors properly, but will merge this for now if were ok with this option till then?

desbo:
sounds good to me :+1:

PR #11731, opened 2016-01-27
Added switch for Plista content and logic to serve Outbrain/Plista widgets

JonNorman:
I've added a switch: _plista-for-outbrain-au_ which, when enabled, will cause the Outbrain recommended content widget to be replaced by that of Plista for AU edition content.

A separate switchboard has been created to allow the toggling of this switch without exposing the other switches on our switchboard: `/dev/switchboard-plista`. The Plista controller ensures that the state of the switches absent from the Plista switchboard maintain their values when being written back to S3.

image

The Plista widget is integrated in a similar fashion to Outbrain with the following caveats:
- there is no Ophan tracking for the Plista widget as there currently is for Outbrain and
- the Plista widget code is currently not rendering any HTML inside its container. This is currently sitting with the Plista team to investigate as the widget ceased to work despite no changes to the code i.e. it appears to be something relating to the script that is requested from the Plista server.


static/src/javascripts/projects/common/modules/commercial/third-party-tags.js
```diff
@@ -11,19 +11,19 @@ define([
...
-    audienceScienceGateway,
-    audienceSciencePql,
-    imrWorldwide,
-    remarketing,
-    krux,
-    outbrain
-) {
+    'common/modules/commercial/third-party-tags/outbrain',
+    'common/modules/commercial/third-party-tags/plista'
+], function (Promise,
```
jbreckmckye: Most modules in Frontend have their dependencies written in the original file's style.

static/src/javascripts/projects/common/modules/commercial/third-party-tags/plista.js
```diff
@@ -0,0 +1,87 @@
...
+             detect,
+             template,
+             identity,
+             plistaTpl,
+             contains) {
+    var containerOuterClassName = 'js-plista',
+        containerInnerClassName = 'js-plista-container';
+
+    return {
+        load: function () {
```
jbreckmckye: How many of these methods need to be public?

static/src/javascripts/projects/common/modules/commercial/third-party-tags/plista.js
```diff
@@ -0,0 +1,87 @@
...
+], function (fastdom,
+             $,
+             config,
+             mediator,
+             detect,
+             template,
+             identity,
+             plistaTpl,
+             contains) {
+    var containerOuterClassName = 'js-plista',
```
jbreckmckye: I believe we are moving away from the single-`var` style.

static/src/javascripts/projects/common/modules/commercial/third-party-tags/plista.js
```diff
@@ -0,0 +1,87 @@
...
+                lib.widgets.push({name: w, pre: u});
+            }
+        },
+
+        getSection: function () {
+            return config.page.section.toLowerCase().match('news')
+            || contains(['politics', 'world', 'business', 'commentisfree'], config.page.section.toLowerCase()) ? 'sections' : 'all';
+        },
+
+        identityPolicy: function () {
```
jbreckmckye: Repeated logic for turning features on or off should probably live in the commercial-feature-policies file. Each entry there is a policy for a particular domain concept, e.g. 'sensitive content', 'logged in'. Perhaps this should be extracted into there?

static/src/javascripts/projects/common/modules/commercial/third-party-tags/plista.js
```diff
@@ -0,0 +1,87 @@
...
+            fastdom.write(function () {
+                $containerOuter.css('display', 'block');
+                $containerInner.append($.create(template(plistaTpl)));
+
+                this.embed('462925f4f131001fd974bebe', 'innerArticle', 'au', undefined, undefined, containerInnerClassName);
+            }.bind(this));
+        },
+
+        // a modification of the code provided by Plista; altered to be a lazy load rather than during DOM construction
+        embed: function (p, w, g, u, c, containerTag) {
```
jbreckmckye: If we're writing this ourselves, that means we'll be maintaining it ourselves, so we should probably use the naming conventions we'd use elsewhere in our own code, rather than trying to resemble the widget code plista gave us (which will only diverge further from our implementation over time).

static/src/javascripts/projects/common/modules/commercial/third-party-tags/plista.js
```diff
@@ -0,0 +1,87 @@
...
+
+        identityPolicy: function () {
+            return (!identity.isUserLoggedIn() || !(identity.isUserLoggedIn() && config.page.commentable));
+        },
+
+        hasHighRelevanceComponent: function () {
+            return detect.adblockInUse() || config.page.edition.toLowerCase() === 'int';
+        },
+
+        init: function () {
```
jbreckmckye: I'm having a little difficulty seeing where the if-test ends in this function - it looked to me as though the function body is part of the expression, almost.

static/src/javascripts/projects/common/modules/commercial/third-party-tags/plista.js
```diff
@@ -0,0 +1,87 @@
...
+            if (
+                !config.page.isFront
+                && !config.page.isPreview
+                && this.identityPolicy()
+                && config.page.section !== 'childrens-books-site') {
+                if (this.hasHighRelevanceComponent()) {
+                    this.load();
```
jbreckmckye: If we want to guard the `load` action, should we perhaps make it private?

static/src/javascripts/projects/common/modules/commercial/third-party-tags/plista.js
```diff
@@ -0,0 +1,87 @@
...
+                } else {
+                    mediator.on('modules:commercial:dfp:rendered', function (event) {
+                        if (event.slot.getSlotId().getDomId() === 'dfp-ad--merchandising-high' && event.isEmpty) {
```
jbreckmckye: Is this repeated from somewhere else? Perhaps dfp-api should emit a `commercial:dfp:emptyMerchandisingHigh` event.

regiskuckaertz:
We'll need to sync our changes before going live, as [the rules to load Outbrain have changed](https://github.com/guardian/frontend/pull/11758). One easy fix would be to keep the `init` logic inside each module and let `third-party-tags` decide which to initialise.

static/src/javascripts/projects/common/modules/commercial/third-party-tags/plista.js
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    function identityPolicy() {
+        return !(identity.isUserLoggedIn() && config.page.commentable);
+    }
+
+    function shouldServe() {
+        return commercialFeatures.outbrain &&
+                config.switches.plistaForOutbrainAu &&
+                config.page.edition == 'AU' &&
```
regiskuckaertz: You could in fact remove that line and the previous one, since you already know they are both true (see `third-party-tags`)

static/src/javascripts/projects/common/modules/commercial/third-party-tags/plista.js
```diff
@@ -0,0 +1,98 @@
...
+    // a modification of the code provided by Plista; altered to be a lazy load rather than during DOM construction
+    function embed(publickey, widgetName, geo, u, categories) {
+        var name = 'PLISTA_' + publickey;
+        var lib = window[name];
+        var $container = $(selectors.container);
+        var $plista = $(selectors.widget);
+
+        $container.append('<div data-display="plista_widget_' + widgetName + '"></div>');
+        $container.css('display', 'block');
+        $plista.append(plistaTpl);
```
regiskuckaertz: Shouldn't it rather be `plistaTpl()`?

static/src/javascripts/projects/common/modules/commercial/third-party-tags/plista.js
```diff
@@ -0,0 +1,98 @@
...
+        });
+    }
+
+    function init() {
+        if (shouldServe()) {
+            if (loadInstantly()) {
+                module.load();
+            } else {
+                mediator.on('modules:commercial:dfp:rendered', function (event) {
+                    if (event.slot.getSlotId().getDomId() === 'dfp-ad--merchandising-high' && event.isEmpty) {
```
regiskuckaertz: The documented API is `event.slot.getSlotElementId()`, I corrected it in my PR so you'd better too (in the test as well), since one never knows when Google deprecates this one.

static/src/javascripts/projects/common/modules/commercial/third-party-tags/plista.js
```diff
@@ -0,0 +1,98 @@
...
+
+        if (!lib || !lib.publickey) {
+            window[name] = {
+                publickey: publickey,
+                widgets: [{name: widgetName, pre: u}],
+                geo: geo,
+                categories: categories,
+                dataMode: 'data-display'
+            };
+            require(['js!' + (document.location.protocol === 'https:' ? 'https:' : 'http:') + '//static-au.plista.com/async/' + name + '.js']);
```
regiskuckaertz: You can actually leave out the `(document.location.protocol === 'https:' ? 'https:' : 'http:')`

regiskuckaertz:
:+1: the front-end part looks ok to me, but I can't vouch for the back-end part

static/src/javascripts/projects/common/modules/commercial/third-party-tags.js
```diff
@@ -22,8 +26,17 @@ define([
...
     remarketing,
     krux,
-    outbrain
-) {
+    identity,
+    outbrain,
+    plista) {
+
+    function loadExternalContentWidget() {
+        if (config.switches.plistaForOutbrainAu && config.page.edition.toLowerCase() == 'au') {
```
jbreckmckye: We should probably prefer `===`

static/src/javascripts/projects/common/modules/commercial/third-party-tags/plista.js
```diff
@@ -0,0 +1,93 @@
...
+    function shouldServe() {
+        return commercialFeatures.outbrain &&
+                !config.page.isFront &&
+                !config.page.isPreview &&
+                identityPolicy();
+    }
+
+    // a modification of the code provided by Plista; altered to be a lazy load rather than during DOM construction
+    function embed(publickey, widgetName, geo, u, categories) {
+        var name = 'PLISTA_' + publickey;
```
jbreckmckye: Should this key live in PROD.properties, DEV.properties, etc?

admin/app/controllers/SwitchboardPlistaController.scala
```diff
@@ -0,0 +1,94 @@
...
+import tools.Store
+import model.NoCache
+
+object SwitchboardPlistaController extends Controller with AuthLogging with Logging with ExecutionContexts {
+
+  val SwitchPattern = """([a-z\d-]+)=(on|off)""".r
+
+  def renderSwitchboard() = AuthActions.AuthActionTest.async { implicit request =>
+    log("loaded plista Switchboard", request)
+    Future { Store.getSwitchesWithLastModified } map { switchesWithLastModified =>
```
kelvin-chappell: Doesn't seem much value in wrapping this in a future.

kelvin-chappell: Add new method to Store to getPlistaSwitchWithLastModified.

admin/app/controllers/SwitchboardPlistaController.scala
```diff
@@ -0,0 +1,94 @@
...
+      val configuration = switchesWithLastModified.map(_._1)
+      val nextStateLookup = Properties(configuration getOrElse "")
+
+      Switches.all foreach { switch =>
```
kelvin-chappell: Then this can be simplified.

admin/app/controllers/SwitchboardPlistaController.scala
```diff
@@ -0,0 +1,94 @@
...
+        nextStateLookup.get(switch.name) foreach {
+          case "on" => switch.switchOn()
+          case _ => switch.switchOff()
+        }
+      }
+
+      val lastModified = switchesWithLastModified.map(_._2).map(_.getMillis).getOrElse(System.currentTimeMillis)
+      NoCache(Ok(views.html.switchboardPlista(Configuration.environment.stage, lastModified)))
```
kelvin-chappell: And view only needs to take the one switchWithLastModified.

kelvin-chappell: Pass switch and lastModified into view.

admin/app/controllers/SwitchboardPlistaController.scala
```diff
@@ -0,0 +1,94 @@
...
+
+    // if any of the switches have a later modified date than locally, redirect
+    if (remoteSwitches.map(_._2).exists(_.getMillis > localLastModified)) {
+      Future {
+        NoCache(Redirect("/dev/switchboard-plista").flashing("error" -> "A more recent change to the switch has been found, please refresh and try again."))
+      }
+    } else {
+      log("saving plista switchboard", request)
+      val requester = UserIdentity.fromRequest(request).get.fullName
+      // for switches not present on this page, we need to persist their current values
```
kelvin-chappell: You only need the change in the plista switch and the remote switches that you've fetched.  You could make a map out of the remote switches and then add the plista switch to it, which would update the state of the plista switch.  Then save the new map in whatever form the method expects.

admin/app/views/switchboardPlista.scala.html
```diff
@@ -0,0 +1,32 @@
+@import conf.switches.Switch
+@(env: String, lastModified: Long)(implicit flash: Flash)
```
kelvin-chappell: Pass plista switch in as well.

admin/app/views/switchboardPlista.scala.html
```diff
@@ -0,0 +1,32 @@
...
+@import controllers.admin.routes.UncachedAssets
+
+@admin_main("Switchboard - Plista", env, isAuthed = true) {
+<link href="@UncachedAssets.at("css/switchboard.css")" rel="stylesheet">
+<link href="@UncachedAssets.at("css/radiator.css")" rel="stylesheet">
+@if(flash.get("error").isDefined) { <h1 style="color:#bd362f">@flash.get("error").get</h1> }
+<div class="row-fluid">
+    <form id="switchboard" action="/dev/switchboard-plista" method="POST">
+        <input type="hidden" name="lastModified" value="@lastModified" />
+        @conf.switches.Switches.all.filter(_.name=="plista-for-outbrain-au").map{case switch: Switch =>
```
kelvin-chappell: No need to include all switches in view.

admin/app/views/switchboardPlista.scala.html
```diff
@@ -0,0 +1,32 @@
...
+            <h4>Plista Activation Switch</h4>
+            <p>When enabled, the following switch will swap in the Plista content recommendation widget
+                in the place of Outbrain (for AU edition only)</p>
+            <p>When the switch has been set - click the 'Save' button to confirm.</p>
+            <p><b>Note: this switch should only be enabled in the CODE environment, not PROD</b></p>
+            <div class="well">
+                <div class="checkbox">
```
kelvin-chappell: Maybe disable this in the prod environment.
`@if(env == 'prod') // then disable checkbox, or something like that`

common/app/conf/switches/FeatureSwitches.scala
```diff
@@ -41,6 +41,15 @@ trait FeatureSwitches {
     exposeClientSide = true
   )
 
+  val PlistaForOutbrainAU = Switch(
```
kelvin-chappell: If this switch is for the replacement of outbrain for plista, it should have an expiry date.
Then we will need another permanent switch to turn plista on or off when it goes wrong.

kelvin-chappell:
Looks good apart from comments!

admin/app/controllers/SwitchboardPlistaController.scala
```diff
@@ -0,0 +1,81 @@
...
+      val currentState = Properties(remoteSwitches.map(_._1) getOrElse "")
+      val currentConfig = Switches.all.filterNot(_ == Switches.PlistaForOutbrainAU).map{switch => switch.name + "=" + currentState.get(switch.name).getOrElse("off")}
+      val plistaConfig = Switches.PlistaForOutbrainAU.name + "=" + plistaSetting
+
+      saveSwitchesOrError(requester, currentConfig :+ plistaConfig)
+    }
+  }
+
+  private def saveSwitchesOrError(requester: String, updates: Seq[String]) = try {
+    val current =  Switches.all map { switch =>
```
kelvin-chappell: Don't think you need this block.

admin/app/controllers/SwitchboardPlistaController.scala
```diff
@@ -0,0 +1,81 @@
...
+      switch.name + "=" + (if (switch.isSwitchedOn) "on" else "off")
+    }
+
+    Store.putSwitches(updates mkString "\n")
+    SwitchesUpdateCounter.increment()
+
+    log.info("plista switches successfully updated")
+
+    val changes = updates filterNot { current contains _ }
```
kelvin-chappell: The only change is going to be to the Plista switch.

admin/app/controllers/SwitchboardPlistaController.scala
```diff
@@ -0,0 +1,81 @@
...
+    Notification.onSwitchChanges(requester, Configuration.environment.stage, changes)
+    changes foreach { change =>
```
kelvin-chappell: Similarly here.

admin/app/controllers/SwitchboardPlistaController.scala
```diff
@@ -0,0 +1,83 @@
...
+
+    def saveSwitchesOrError(updates: Seq[String], newState: String) = try {
+
+      val requester = UserIdentity.fromRequest(request).get.fullName
+      Store.putSwitches(updates mkString "\n")
+      SwitchesUpdateCounter.increment()
+
+      log.info("plista switches successfully updated")
+
+      val alterationMade = Switches.PlistaForOutbrainAU.isSwitchedOn match{
```
kelvin-chappell: This would probably be better as an `if` because the cases are just booleans.

admin/conf/routes
```diff
@@ -56,7 +56,10 @@ GET         /ophan/ads/render-time
 
 #Development endpoints
 GET         /dev/switchboard                                                                        controllers.admin.SwitchboardController.renderSwitchboard()
```
kelvin-chappell: Please add the new routes to dev-build routes file as well.

kelvin-chappell:
Good work :+1: when comments acted on

PR #11733, opened 2016-01-27
Fix three Toast bugs

holder-josh:
cc @OliverJAsh,  just a few minor bug fixes for toast:
- Loading button no longer shows when it shouldn't (was related to adding the loading class when there are no new blocks)
- IE doesn't support window.scrollY
- Remove unnecessary margin on the toast button


static/src/javascripts/projects/common/modules/ui/newAutoupdate.js
```diff
@@ -54,7 +54,7 @@ define([
         };
 
         var scrolledPastTopBlock = function () {
-            return $liveblogBody.offset().top < window.scrollY;
+            return $liveblogBody.offset().top < window.pageYOffset;
         };
         var isLivePage = !(window.location.href.search('[?&]page=') !== -1);
```
johnduffell: no need the double negative
`var isLivePage = (window.location.href.search('[?&]page=') === -1);`

johnduffell:
:+1:

static/src/javascripts/projects/common/modules/ui/newAutoupdate.js
piuccio: isn't it `window.location.search` ? https://developer.mozilla.org/en-US/docs/Web/API/Location

johnduffell: good point,t hat's what I used and it seems(!) to work
https://github.com/guardian/frontend/blob/master/static/src/javascripts/bootstraps/enhanced/liveblog.js#L156

Jholder112233: Yep, error there that needs fixing. @OliverJAsh, do you mind fixing that before shipping this? 

Thanks very much.

michaelwmcnamara:
[ ] - Loading button showing when it shouldn't - Fixed
[ ] - Formatting on toast popup - Fixed
[ ] - Toast notifications not working on IE - not fixed - no toast is displayed in IE
[ ] - check all on later pages of liveblogs - still to come

static/src/javascripts/projects/common/modules/ui/newAutoupdate.js
OliverJAsh: I think you have done it correctly, actually. `window.location.href.search` is a function on the `String` prototype, which is what you want. `window.location.search` is a string of query params.

Question is, how well supported is `window.location.href.search`…?

OliverJAsh: Will change to `window.location.search.indexOf` as I think that makes sense

piuccio: [String.search](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/search), I like that, just a bit confusing when you read it

OliverJAsh:
Seems to be working on all the IEs when opted into this test as well as `RemoveStickyNav=new`. Test URL: http://localhost:9000/media/live/2016/feb/05/julian-assange-is-being-arbitrarily-detained-un-panel-finds-live-updates#ab-LiveblogToast=toast,RemoveStickyNav=new

OliverJAsh:
This is ready to go. I'll ship it when I'm done with sticky :smile:

PR #11734, opened 2016-01-27
Fix invisible text on sponsored fronts

jbreckmckye:
Sponsored content fronts inherit styles from the existing facia containers. Some containers with the `--story-package` type have a dark background colour set with `!important` [[story-package.scss:69](https://github.com/guardian/frontend/blob/master/static/src/stylesheets/inline/story-package.scss#L69)]. This makes the dark copy on the sponsored items effectively invisible:

image

This is a production issue, and fairly high-priority. For the moment I've resolved the issue by shadowing the !important selector in the sponsored styles rather than trying to unpick the original code and see if its specificity hacks can be removed. This `important` is scoped within a check for the new paid content theme, so it should at least stay relatively contained.

### DIFF
--- a/static/src/stylesheets/module/commercial/_capi.scss
+++ b/static/src/stylesheets/module/commercial/_capi.scss
@@ -117,6 +117,11 @@
         border-top: 0;
     }
 
+    .fc-container--story-package .tone-news--item.fc-item {
+        // This is to override an !important background colour in story-package items
+        background-color: colour(paid-article-subheader) !important;
+    }
+
     .tone-media--item {
         .u-faux-block-link--hover {
             background-color: colour(neutral-6);


sndrs:
cor that's a real sign that something's not right!

:+1: to get it fixed

johnduffell:
:+1: maybe we should try harder to avoid !important (as a coding standard), I have had trouble with them competing before too

gtrufitt:
When everything is `!important`, nothing is `!important` :D

johnduffell:
haha it's just not very guardian :stuck_out_tongue:

jbreckmckye:
Merging, but would really like to fix this a bit more thoroughly.

PR #11737, opened 2016-01-28
make liveblog timeline use permalinks

johnduffell:
at the moment the timeline (key events panel) uses anchor links and only populates from the first page.

This PR changes it to pull the events from all pages, and uses the block permalink to find it.  This does have the side effect of making the page refresh, but it should allow the page size to be made a lot smaller if desired.

A bit more work is needed to make the smooth scroller work if the block is actually on the current page - this would involve altering the data-block-id to only be present if the block is on the current page (server side logic) and then the client side smooth scroller could only capture the event if the data-block-id is defined. I have added a switch to remind us to do that when I'm available.

@stephanfowler @OliverJAsh I don't think I should ship it today, feel free to close for the time being if you can't get it out

### DIFF
--- a/article/app/liveblog/LiveBlogPageModel.scala
+++ b/article/app/liveblog/LiveBlogPageModel.scala
@@ -13,7 +13,7 @@ object LiveBlogPageModel {
 
     endedPages.sliding(3).toList.map {
       case List(later, curr, earlier) =>
-        LiveBlogPageModel(curr.page, main, later.self, earlier.self, curr.self)
+        LiveBlogPageModel(curr.page, blocks, later.self, earlier.self, curr.self)
     }.find(hasRequestedBlock)
   }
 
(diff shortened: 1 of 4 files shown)


stephanfowler:
@johnduffell - I just chatted to Matt W who has noticed the key events behaviour and is keen to have this fix. So we'll probably ship it.

stephanfowler:
> "this would involve altering the data-block-id to only be present if the block is on the current page"
I think this risks leaking out URLs that aren't proper permalinks. I'd rather prevent that risk, and not worry about smooth scrolling etc. So, the way you have it in this PR is the right outcome.

johnduffell:
@stephanfowler actually it wouldn't cause a risk because the code can scroll based on client side logic and the data-block-id couldn't really leak out - the actual href would still be the proper permalink.  When it actually scrolls it can update the url to the proper permalink too

stephanfowler:
Ak, OK. I understood it as: you were proposing to omit `?page=with:block-123` on the server side.

johnduffell:
the changes would purely be to data-block-id.  I'll take a look when the switch nears expiry

michaelwmcnamara:
+1

NataliaLKB:
Merging and deploying for @johnduffell :+1:

johnduffell:
Thanks for that!

PR #11738, opened 2016-01-28
Extract cardstyle

janua:
This just optimises the imports and extracts `cardStyle` out into a `val`.

@rich-nguyen


rich-nguyen:
:+1:

PR #11741, opened 2016-01-28
Override story-package styles on paid-for fronts

jbreckmckye:
This continues the [fix](https://github.com/guardian/frontend/pull/11734) from this morning. We still have hard-to-read text on live, because of hover styles:

image

There's also still poor contrast in the non-hover state with the standfirst:

image

This PR applies overrides for the story-package styles on fc-containers. I'd like to take a look at stopping the Scala from adding the story-package class full stop, so we can remove all the overrides entirely, but that might take a little while as I don't know Scala well.

Fix screenshot, hover state:
image

Fix screenshot, non-hover state:
image


regiskuckaertz:
:+1:

PR #11743, opened 2016-01-28
Use DAPI webUrl for permalink URL 

gidsg:
Previously comment permalinks were built using a host value from `Configuration` e.g. `@Configuration.discussion.url/comment-permalink/@comment.id` however, we provide this URL in the DAPI response as `webUrl` so this PR changes frontend to read `webUrl` in all instances.

This means we can remove the config value `Configuration.discussion.url` which was the old Discussion D2 microapp and allow us to update the `webUrl` in future without having to modify Frontend.

CC @guardian/discussion @johnduffell


jamesgorrie:
Nice and tidy.
:+1:

PR #11745, opened 2016-01-28
Add hasHeader to metadata part 2

gtrufitt:
Adding to match up with https://github.com/guardian/frontend/blob/9ef511edf54676066302d85961c224337642beee/common/app/model/Formats.scala#L147

cc/ @rich-nguyen @janua


rich-nguyen:
:+1:

PR #11756, opened 2016-01-29
Sponsored popup abstract

uplne:
Created fragment for paid for content meta block so we don't need to edit it in multiple places.
- I did the requested change to replace 'created' with 'produced'

screen shot 2016-01-29 at 13 56 03

CC @kelvin-chappell @jbreckmckye @Calanthe


common/app/views/fragments/commercial/paidForMeta.scala.html
```diff
@@ -0,0 +1,14 @@
+@()(implicit request: RequestHeader)
```
kelvin-chappell: Is the `request` needed here?

kelvin-chappell:
Apart from comment :+1:

common/app/views/fragments/commercial/paidForMeta.scala.html
uplne: LInkTo required it

kelvin-chappell: Then :ship:

PR #11758, opened 2016-01-29
Load Outbrain in place of the low-pri merch component

regiskuckaertz:
These are the new rules for when and how Outbrain should be displayed, across all editions, under the same condition (i.e. not a front, not on children's book site, anonymous user or article w/o comments)
1. if the visitor has an ad blocker: same as before, otherwise
2. if the high priority merchandising component is empty: same as before, otherwise
3. if the high priority merchandising component is not empty _and_ the low priority is empty: move the outbrain widget down to where the low-pri component should have appeared and load it from there

The latter case uses new codes given by Outbrain, and this is not looking too good right now:
picture 1

Thus, point **3** is hidden behind a switch until they can fix this.

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/third-party-tags/outbrain-codes.js
+++ b/static/src/javascripts/projects/common/modules/commercial/third-party-tags/outbrain-codes.js
@@ -38,7 +38,7 @@ define(function () {
     };
 
     function getCode(data) {
-        if (data.slot === 'outbrain') {
+        if (!data.slot in outbrainCodes || data.slot === 'outbrain') {
             return outbrainCodes.outbrain[data.section][data.breakpoint === 'wide' ? 'desktop' : data.breakpoint];
         } else {
             return outbrainCodes.merchandising[data.edition][data.breakpoint === 'wide' ? 'desktop' : data.breakpoint];
--- a/static/src/javascripts/projects/common/modules/commercial/third-party-tags/outbrain-codes.js
+++ b/static/src/javascripts/projects/common/modules/commercial/third-party-tags/outbrain-codes.js
@@ -1,0 +1,49 @@
+define(function () {
+    var outbrainCodes = {
+        outbrain: {
+            1: {
+                mobile:  { code: 'MB_4' },
+                desktop: { image: 'AR_12', text: 'AR_14' },
+                tablet:  { image: 'MB_6', text: 'MB_8' }
+            },
+            2: {
+                mobile:  { code: 'MB_5' },
+                desktop: { image: 'AR_13', text: 'AR_15' },
+                tablet:  { image: 'MB_7', text: 'MB_9' }
+            }
+        },
+
+        merchandising: {
+            UK: {
+                mobile:  { code: 'MB_10' },
+                desktop: { code: 'AR_28' },
+                tablet:  { code: 'MB_11' }
+            },
+            US: {
+                mobile:  { code: 'CRMB_55' },
+                desktop: { code: 'CR_13' },
+                tablet:  { code: 'CRMB_56' }
+            },
+            AU: {
+                mobile:  { code: 'CRMB_57' },
+                desktop: { code: 'CR_14' },
+                tablet:  { code: 'CRMB_58' }
+            },
+            INT: {
+                mobile:  { code: 'CRMB_59' },
+                desktop: { code: 'CR_15' },
+                tablet:  { code: 'CRMB_60' }
+            }
+        }
+    };
+
+    function getCode(data) {
+        if (data.slot === 'outbrain') {
+            return outbrainCodes.outbrain[data.section][data.breakpoint === 'wide' ? 'desktop' : data.breakpoint];
+        } else {
+            return outbrainCodes.merchandising[data.edition][data.breakpoint === 'wide' ? 'desktop' : data.breakpoint];
+        }
+    }
+
+    return getCode;
+})
--- a/static/src/javascripts/projects/common/modules/commercial/third-party-tags/outbrain-codes.js
+++ b/static/src/javascripts/projects/common/modules/commercial/third-party-tags/outbrain-codes.js
@@ -1,4 +1,6 @@
-define(function () {
+define([
+    'common/modules/commercial/third-party-tags/outbrain-sections'
+], function (getSection) {
     var outbrainCodes = {
         outbrain: {
             1: {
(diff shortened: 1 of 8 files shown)


static/src/javascripts/projects/common/modules/commercial/third-party-tags/outbrain-sections.js
```diff
@@ -0,0 +1,10 @@
+define(function () {
+    var sections = ['politics', 'world', 'business', 'commentisfree'];
+
+    function getSection(section) {
+        section = section.toLowerCase();
+        return /news/.test(section) || sections.indexOf(section) !== -1 ? 1 : 2;
```
uplne: Is 1 and 2 good identifier for this? I would suggest we use something more descriptive: inisde-sections : outside-sections. Maybe that's not the best wording but you see the point. It will be more obvious in the outbrain-codes.js, especially as the logic is split into two modules.

static/src/javascripts/projects/common/modules/commercial/third-party-tags/outbrain-codes.js
```diff
@@ -0,0 +1,51 @@
+define([
+    'common/modules/commercial/third-party-tags/outbrain-sections'
+], function (getSection) {
+    var outbrainCodes = {
```
jbreckmckye: What are these, exactly?

static/src/javascripts/projects/common/modules/commercial/third-party-tags/outbrain.js
```diff
@@ -16,109 +18,152 @@ define([
...
-                    }
-                }
-            };
+    var selectors = {
+        outbrain: {
+            widget: '.js-outbrain',
+            container: '.js-outbrain-container'
+        },
+        merchandising: {
+            widget: '.fc-container--commercial',
```
jbreckmckye: Should we make this target a `js-` class?

static/test/javascripts/spec/common/commercial/third-party-tags/outbrain.spec.js
```diff
@@ -66,151 +74,141 @@ define([
...
+                });
+                mediator.emit('modules:commercial:dfp:rendered', eventStub).then(function () {
+                    mediator.emit('modules:commercial:dfp:rendered', eventStubLo);
+                });
 
-                sut.init();
-                expect(sut.load).toHaveBeenCalled();
+                mediator.emit = oldEmit;
             });
         });
```
uplne: We should include test for the case when high and low priority are both on the page. To make sure that in that case we definitely not load outbrain.

static/src/javascripts/projects/common/modules/commercial/third-party-tags/outbrain.js
```diff
@@ -16,109 +18,152 @@ define([
...
+            html += outbrainTpl({ widgetCode: codes.code || codes.text });
+        }
+        return html;
+    }
 
-            fastdom.write(function () {
-                $outbrain.css('display', 'block');
-                $container.append($.create(template(outbrainTpl, { widgetCode: widgetCode })));
+    function load(target) {
+        var slot          = target in selectors ? target : 'outbrain',
```
jbreckmckye: Should we start using multiple `var` declarations?

static/src/javascripts/projects/common/modules/commercial/third-party-tags/outbrain-codes.js
uplne: Yes, we need some description/documentation for the future.

static/src/javascripts/projects/common/modules/commercial/third-party-tags/outbrain.js
```diff
@@ -16,109 +18,152 @@ define([
...
-            require(['ophan/ng'], function (ophan) {
-                ophan.record({
-                    outbrain: {
-                        widgetId: widgetCode
-                    }
-                });
-            });
-        },
+            var onAllAdsLoaded = function () {
+                off();
```
jbreckmckye: `off` is a bit of a terse name, I think.

static/src/javascripts/projects/common/modules/commercial/third-party-tags/outbrain.js
```diff
@@ -16,109 +18,152 @@ define([
...
+    function loadInstantly() {
+        return !document.getElementById('dfp-ad--merchandising-high') ||
+            detect.adblockInUse();
+    }
+
+    function init() {
+        if (config.switches.outbrain
+            && !config.page.isFront
+            && !config.page.isPreview
+            && config.page.section !== 'childrens-books-site'
```
jbreckmckye: Is our intention to turn off Outbrain on sensitive content pages here, or just the children's books site? The latter is usually done alongside the former. There is a sensitive content policy in `commercialFeaturePolicies.js` that could do this.

static/src/javascripts/projects/common/modules/commercial/third-party-tags/outbrain-sections.js
regiskuckaertz: yes, I really struggled with this. I used the same vocabulary than Outbrain, but yours is clearer.

static/src/javascripts/projects/common/modules/commercial/third-party-tags/outbrain.js
regiskuckaertz: yes, probably. Will do.

static/src/javascripts/projects/common/modules/commercial/third-party-tags/outbrain.js
regiskuckaertz: I do prefer multiple vars, it's more readable.

static/src/javascripts/projects/common/modules/commercial/third-party-tags/outbrain.js
regiskuckaertz: good catch @jbreckmckye, thanks! I thought about adding the whole test as a policy, I think it belongs there, but would prefer to postpone it to when we do refactor the codebase.

uplne: From our discussion with @jbreckmckye a few months ago:

> Policies would normally describe particular states of the page, rather than individual features. For example, we would have a userLoggedIn policy or a sensitivePage policy (is Outbrain supposed to be disabled on all sensitive content, incidentally?). This allows us to describe what the policies in these cases are. Otherwise I don't think we really take advantage of centralizing this logic.

Just something to think about before you will do that.

jbreckmckye: As it was already there, I think you're probably right.

static/src/javascripts/projects/common/modules/commercial/third-party-tags/outbrain.js
```diff
@@ -16,109 +19,151 @@ define([
...
+        make the call instantly when we detect adBlock in use.
+     */
+    function loadInstantly() {
+        return !document.getElementById('dfp-ad--merchandising-high') ||
+            detect.adblockInUse();
+    }
+
+    function init() {
+        if (commercialFeatures.outbrain &&
+            !config.page.isFront &&
```
uplne: I think there is a policy for nonFrontPages so you might be able to add outbrain into that policy as you moved it there now.

uplne:
:+1:

uplne:
:ship:

PR #11761, opened 2016-01-29
Adding apple touch icon. In the hope that we can serve it from static.

JustinPinner:
Adding the apple-touch-icon.png manually to the static images bucket didn't work (access denied in spite of granting everyone read rights). @rich-nguyen suggested making it part of the frontend repo so here we are.

:apple: :boom:


rich-nguyen:
yes, that should work. bear in mind the final asset path will be hashed. :+1:

PR #11762, opened 2016-01-29
Extend prebid.js A/B test

jbreckmckye:
This extends the prebid.js performance A/B test for a week. We'll determine how long this test needs to run next week in collaboration with Data Science.

@Calanthe


Calanthe:
:+1:

PR #11763, opened 2016-01-29
Remove redundant ipad split features switch

JustinPinner:
As discussed with @michaelwmcnamara - this switch isn't used any more.

cc @sndrs @OliverJAsh


sndrs:
:+1:

PR #11765, opened 2016-01-29
Autoplay video

jamesgorrie:
Whenever you land on a video page - autoplay.
As discussed with Piers / Yannick / Editorial.

### DIFF
--- a/static/src/javascripts/projects/common/modules/onward/history.js
+++ b/static/src/javascripts/projects/common/modules/onward/history.js
@@ -141,12 +141,6 @@ define([
         saveSummary(summary);
     }
 
-    function isRevisit(pageId) {
-        return (find(getHistory(), function (page) {
-            return (page[0] === pageId);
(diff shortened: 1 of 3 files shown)


rich-nguyen:
This was specifically added on editorial request. It used to be autoplay everywhere. People were complaining that they were seeing the same video (and adverts) over and over, but were revisiting the same video page for the purposes of seeing the comments again. Shouldn't this be a user option?

Also, does mobile remain non-autoplay?

> plus ça change, plus c’est la même chose

jamesgorrie:
> Shouldn't this be a user option?
Potentially.

> It used to be autoplay everywhere...  does mobile remain non-autoplay
It now only autoplays on video pages and desktop.

jamesgorrie:
I'm speaking to Editorial on Monday - so I will let you know then.

jamesgorrie:
Closed as there was a misunderstanding with Editorial.

PR #11764, opened 2016-01-29
Editionalising onward journeys on amp

NataliaLKB:
This will make sure that the edition is no longer being hardcoded, and is actually coming from the user's preferences/location.

cc @stephanfowler

PR #11767, opened 2016-01-29
adding properties file for AMP

NataliaLKB:
Forgot to change the properties file for AMP


NataliaLKB:
Testing this in conjunction with #11764 on CODE. Will revert and deploy properly on Monday

PR #11771, opened 2016-01-30
Promote FA Cup in hierarchy of competitions

phamann:
### tl;dr

As a user, I don't want to be diagnosed with RSI due to excessive scrolling whilst trying to find the FA cup results.

google chrome 14

The FA cup was in a very low position within the definition sequence, said `Seq` is used to control the display order on competition listing pages. The majority of users accessing `/football/live` on FA cup days – especially on the UK edition – would be expecting to find the results at the top of the page and not have to scroll through 4+ irrelevant competitions first and ultimately contracting early RSI.


phamann:
FYI obvs **haven't** tested this locally... #yolo

paperboyo:
:wave::sunglasses:

rich-nguyen:
:+1: toptier

JustinPinner:
The thing is, I think the order of this list is editorially decided. Or it used to be anyway.

daithiocrualaoich:
Should have just waited a few weeks until Arsenal are knocked out. Now you're just going to have to revert this when they are. #midpack

mattandrews:
While we're here, can we just replace the Football pages with cycling instead?

paperboyo:
tango

PR #11772, opened 2016-01-30
Remove unnecessary star specificity

kaelig:
Note that this should probably follow standard BEM notation:

Block: `.stars`
Elements should probably be: `.stars__item` with an `s` instead of `.star__item`.


rich-nguyen:
classic pr! :+1:  :fr:

PR #11774, opened 2016-02-01
Clear out old clean-up code from the header JS

sndrs:
been a month, time to go


jbreckmckye:
I thought there was a module somewhere that let us clear out old legacy localstorage keys, but I can't find it any longer...

sndrs:
this needs to run outside the module system, unfortunately

jbreckmckye:
I can't remember what the module is, but it essentially clears localstorage entries for keys that were once used, but aren't used any more. You're not relying on the property any more so you shouldn't need the value to be cleared in the critical path scripts.

jbreckmckye:
:+1:

PR #11776, opened 2016-02-01
The Campaign Minute. Version 1, desktop styling

gtrufitt:
The Campaign starts today - this converts the current [mobile-on-desktop](http://www.theguardian.com/us-news/live/2016/jan/28/trump-dumps-debate) styling into nearer to the desktop view. 

v1_wide_scroll

v1_breakpoints

cc/ @superfrank @desbo @crifmulholland

### DIFF
--- a/article/app/views/fragments/headerImmersive.scala.html
+++ b/article/app/views/fragments/headerImmersive.scala.html
@@ -46,22 +46,37 @@
         <div class="content__head--mobile">
             <div class="content__wrapper content__wrapper--headline">
                 <div class="gs-container">
+                @if(!article.isUSMinute) {
                     @article.content.blogOrSeriesTag.map { series =>
-                    <h3 class="content__series-label">
-                        <a class="content__series-label__link" href="@LinkTo {/@series.id}">@series.name</a>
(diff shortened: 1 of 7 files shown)


paperboyo:
Are these images lightboxable into an ad hoc gallery? Just asking (earlier design included a button)...

gtrufitt:
@paperboyo Not currently a part of the design, no, but will talk to the designer-types and see whether that's something we can add later on.

paperboyo:
Thanks, @gtrufitt. It doesn't work in Liveblogs either... Not sure if it would be easy to add. Apparently, in-article Lightbox is quite popular.

gtrufitt:
@paperboyo It's easy enough to add functionally but don't know whether we would want a different design treatment for it in minute articles/us-election :)

desbo:
:+1:

paperboyo:
Sure thing. My take on this - not that it matters - is that it should be available on all article types. If, what I think may be the case sometimes, it isn't because "no, i don't want all these gray buttons on images", we could decide to show them only on hover in these situations.

PR #9553, opened 2015-06-19
Avoid early return

OliverJAsh:
Many of us have been confused about whether bucketing happens before or after `canRun`. (It happens after.)

I try to avoid early returns, so when I see a function, there's only one place I need to look to see what the function returns. More importantly, early returns can lead to bugs, when developers mistakenly assume the whole function will run and have code that must always run be below an early return (or add an early return without realising some code below must always run). For the same reason, the code is harder to refactor as you have to take into account that escape point.

I made this change because the logic is clearer when reversed.

Non-white space diff: https://github.com/guardian/frontend/pull/9553/files?w=1

PR #11777, opened 2016-02-01
Fix AJAX error messages

desbo:
The current version treats the third parameter to the error callback as the XHR object when it should be the first one ([source](https://github.com/ded/reqwest/blob/master/src/reqwest.js#L366)). 

Before: 
screen shot 2016-02-01 at 12 38 54

After:
screen shot 2016-02-01 at 12 38 37

### DIFF
--- a/static/src/javascripts/projects/common/utils/ajax-promise.js
+++ b/static/src/javascripts/projects/common/utils/ajax-promise.js
@@ -8,22 +8,21 @@ define([
     Promise
 ) {
     return function wrappedAjax(params) {
-        var promise = new Promise(function (resolve, reject) {
+        return new Promise(function (resolve, reject) {
             ajax(params)
-            .then(function (value) {
-                resolve(value);
-            })
-            .fail(function (request, text, err) {
-                var statusText = (err && err.statusText) || '';
-                var statusCode = (err && err.status) || '';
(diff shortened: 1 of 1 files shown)


gtrufitt:
:+1:

OliverJAsh:
Nice thing to target! I've always been confused by these error messages.

static/src/javascripts/projects/common/utils/ajax-promise.js
```diff
@@ -8,22 +8,21 @@ define([
...
-                var statusCode = (err && err.status) || '';
-                var errorText = 'Error retrieving data (' + text + ') (Status: ' + statusCode + ') (StatusText: ' + statusText + ')';
+                .then(resolve)
+                .fail(function (res, msg, err) {
+                    if (err) {
+                        return reject(err);
+                    }
 
-                var error = err ? err : new Error(errorText);
+                    if (res && res.status) {
```
OliverJAsh: `else if` and then `else` below might be clearer

desbo: hmm, i much prefer early returns to `if`/`else if` chains :/

gtrufitt: :D https://github.com/guardian/frontend/pull/9553

sndrs:
:+1:

PR #11779, opened 2016-02-01
Urgent fix to us-election banner

sammorrisdesign-zz:
Whoops...

It ended up looking like this on all over non-us-election dead blogs

screen shot 2016-02-01 at 13 49 44


gtrufitt:
:+1:

PR #11780, opened 2016-02-01
Use new avatar subdomain

nicl:
The Avatar API now lives on avatar.theguardian.com. This means we can use it and stop using the deprecated GU_U cookie for authentication.

### DIFF
--- a/static/src/javascripts/projects/common/modules/discussion/user-avatars.js
+++ b/static/src/javascripts/projects/common/modules/discussion/user-avatars.js
@@ -11,7 +11,8 @@ define([
     function avatarify(el) {
         var container = bonzo(el),
             updating = bonzo(bonzo.create('<div class="is-updating"></div>')),
-            avatar = bonzo(bonzo.create('<img class="user-avatar__image" alt="" />'));
+            avatar = bonzo(bonzo.create('<img class="user-avatar__image" alt="" />')),
+            userId = container.data('userid');
 
         container
             .removeClass('is-hidden');
@@ -23,6 +24,10 @@ define([
         avatarApi.getActive()
             .then(function (response) {
                 avatar.attr('src', response.data.avatarUrl);
+                updating.remove();
+                avatar.appendTo(container);
+            }, function () {
+                avatar.attr('src', avatarApi.deterministicUrl(userId));
                 updating.remove();
                 avatar.appendTo(container);
             });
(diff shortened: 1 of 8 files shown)


nicl:
cc @guardian/identity @guardian/discussion

static/src/javascripts/projects/common/modules/discussion/user-avatars.js
```diff
@@ -1,8 +1,8 @@
 define([
     'common/utils/$',
     'bonzo',
-    'common/modules/discussion/api'
-], function ($, bonzo, discussionApi) {
+    'common/modules/avatar/api'
```
sndrs: why `/api` and not just `avatar`?

sndrs: should it also go in `modules/identity/avatar`?

static/src/javascripts/projects/common/modules/discussion/user-avatars.js
```diff
@@ -13,14 +13,21 @@ define([
             updating = bonzo(bonzo.create('<div class="is-updating"></div>')),
             avatar = bonzo(bonzo.create('<img class="user-avatar__image" alt="" />')),
             userId = container.data('userid');
+
         container
```
sndrs: all dom touching should be done inside `fastdom`

static/src/javascripts/projects/common/modules/discussion/user-avatars.js
nicl: I think it deserves a separate folder as it is potentially useful for Discussion as well as Identity code/features.

Re: avatar versus api, no strong preference, was just following what other modules do here - for example discussion/api.js.

static/src/javascripts/projects/common/modules/discussion/user-avatars.js
nicl: @sndrs thanks. How important is this here? The code is written to rely on the return value of the bonzo create operations, whereas fastdom seems to return simply an op identifier. The profile page this impacts has hardly any dom elements on it (https://profile.theguardian.com/public/edit) and I'm a bit reluctant to start rewriting existing code here unless absolutely necessary.

sndrs: sorry - this isn't part of this PR, should have looked closer

sndrs:
it'd be good to fastdom all these dom read/writes at some point, but :+1: for this

PR #11782, opened 2016-02-01
Change image urls for US Election liveblog header

sammorrisdesign-zz:


gtrufitt:
This looks brittle? :+1: to merge but there must be a better way?!

PR #11784, opened 2016-02-01
Restrict number of cards in commercial containers

kelvin-chappell:
Until we have a 'more' CTA for the container.

Before:
image

Now:
image

/cc @Calanthe @jbreckmckye


Calanthe:
:+1:

PR #11787, opened 2016-02-01
No longer require R2 non hash file as we inlined the omniture script.

jennysivapalan:
cc @JustinPinner


JustinPinner:
:no_entry_sign: :hash: :checkered_flag: :+1:

PR #11788, opened 2016-02-01
Hide 'About this content' helptext

jbreckmckye:
In the new commercial content design, there's an 'about this content' message that accompanies sponsor badges.

image

This is unnecessary, because we now have an 'About' popup that links to a special 'content funding' page

image
image

So we now hide the message:

image
image

**Changes**
- `template.js` now tells templates when they're rendering under the new design
- the logo templates now suppress the 'about this content' messages
- I removed an unnecessary new commercial content class in `item.scala.html` (it's in the branch of the code that only runs when `NewCommercialContent.isSwitchedOff`.)

When we come to clean up the commercial content redesign, these switches should hopefully act as markers for content to delete permanently.


static/src/javascripts/projects/common/modules/commercial/creatives/template.js
```diff
@@ -56,6 +56,7 @@ define([
         this.params.logoguardian = svgs('logoguardian');
         this.params.marque36iconCreativeMarque = svgs('marque36icon', ['creative__marque']);
         this.params.logoFeatureLabel = config.switches.newCommercialContent ? 'Paid for by' : 'Brought to you by:';
+        this.params.isNewCommercialContent = !config.switches.newCommercialContent;
```
regiskuckaertz: it seems counter-intuitive to me, that two vars with the same name have opposite values

jbreckmckye: Wait, I got the logic the wrong way around!

regiskuckaertz:
:+1: once you get that right :)

PR #11794, opened 2016-02-02
Minor spacing tweaks to the Minute header on mobile

sammorrisdesign-zz:
A few sizing changes to the minute header as it was breaking on mobile. Code by @superfrank 

cc @desbo @gtrufitt


gtrufitt:
:+1:

PR #11796, opened 2016-02-02
coin based adblock banner

tomverran:
image

I'd hoped to do this entirely in CSS but the SVG images necessitated template changes


tomverran:
@Calanthe Here's the PR I mentioned just now - this is the first commercial banner I've done so let me know if I've made any rookie mistakes!

Thanks

static/src/inline-svgs/commercial/logo-membership.svg
```diff
@@ -0,0 +1,129 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
jbreckmckye: This SVG seems to have a non-standard 'inkspace' namespace on several elements. The inline SVG processor, using SVGO, can't handle them. You can check your file against the [W3 validator](https://validator.w3.org/check). You might have more luck using Inkscape's "Export to Plain SVG" functionality.

paperboyo:
Hi, @tomverran. The adblock-coins.svg (of 100kB!) is nothing more than transparent PNG embedded within an SVG image. I’m not sure all browsers are happy wit that (looking at you Safari!). If they are, we would better make pluses and equal sign vector in that SVG anyway. For now here is a pure PNG of that whole image (18kB):

ai_image_opt_128

Unfortunately, logo-membership.svg needs optimising but it’s invalid - I can’t open it anywhere...

Let me know if I can help any more.

Regards
Mateusz

tomverran:
Thanks for the pointers @paperboyo and @jbreckmckye!

I've re-exported the membership logo without the Inkscape metadata. I've updated the coin SVG to embed the optimised PNG but otherwise left it as-is - it is only an SVG at all due to there not being a simple way to reference hashed non SVG assets clientside. The image is now 24K which doesn't seem like too bad an overhead over the raw PNG and hopefully gzipping will help to mitigate much of it, let me know if that seems okay.

jbreckmckye:
> it is only an SVG at all due to there not being a simple way to reference hashed non SVG assets clientside

You mean in the (non-scala) commercial templates? Yes, you'd need to pipe the asset path into the JavaScript. This might already be done elsewhere, but I couldn't say for sure.

One disadvantage of using the base64-encoded PNG is that it's a bit larger - not _criminally_ so, but still larger than referencing it through a stylesheet, which would be fairly easy:

`background-image: url(@Static("images/commercial/money-money-money.png"));`

If you use a CSS background image, you could treat your image as a sprite - three coins rendered in three blocks, with two SVG / text plus symbols between them. I wouldn't go that far if you don't feel comfortable with frontend work, though.

static/src/stylesheets/module/_release-message.scss
```diff
@@ -466,6 +466,81 @@ $btn-height: gs-height(1) - $gs-baseline/2; //30px
     background: colour(membership-default);
 }
 
+
+
+/*
+ * Coin based adblock message
+ */
+.adblock-sticky__message.adblock-sticky__coin {
```
jbreckmckye: Why are we using this combinator? BEM class names are supposed to enact namespacing by themselves. Was this needed for specificity?

tomverran:
@jbreckmckye Hi, I've updated the CSS so the coin banner is a separate element, sharing only `adblock-sticky__link` with the other banners, does this generally look okay?

Also would I be right in thinking that to reference the png in a stylesheet I'd have to make the stylesheet some kind of scala / twirl template?

jbreckmckye:
Looks great - let's :ship: it!

PR #11798, opened 2016-02-02
Removed disabling Outbrain on non-front pages (duh)

regiskuckaertz:
Fixes an issue with [this PR](https://github.com/guardian/frontend/pull/11758) where Outbrain would not appear on article pages because of a wrong guard condition


jbreckmckye:
:+1: once it passes TC

PR #11631, opened 2016-01-20
Ab survey message

gtrufitt:
Add an (initially 0%) A/B test for a userzoom survey for the data team.

PR #11800, opened 2016-02-02
set up userzoom email survery for mobile

NathanielBennett:
Run a userzoom survery for the data team for mobile users, following on from this test: https://github.com/guardian/frontend/pull/11631 which is running for desktop users. This test has an audience of 30% and runs until Sunday coming. 

@gtrufitt . This look OK to you?

### DIFF
--- a/static/src/javascripts/projects/common/modules/experiments/tests/userzoom-survey-message-mobile-v3.js
+++ b/static/src/javascripts/projects/common/modules/experiments/tests/userzoom-survey-message-mobile-v3.js
@@ -68,7 +68,6 @@ define([
                 arrowWhiteRight: svgs('arrowWhiteRight')
             },
             createMessage = function () {
-                console.log("Create message");
                 new Message(messageId, messageOptions).show(template(messageTemplate, messageTemplateOptions));
                 userPrefs.set('survey-message-seen', true);
             };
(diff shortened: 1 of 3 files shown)


static/src/javascripts/projects/common/modules/experiments/tests/userzoom-survey-message-mobile-v3.js
```diff
@@ -0,0 +1,86 @@
...
+    return function () {
+        this.id = 'UserzoomSurveyMessageMobileV3';
+        this.start = '2016-02-02';
+        this.expiry = '2016-02-07';
+        this.author = 'Nathaniel Bennett';
+        this.description = 'Segment the userzoom data-team survey for mobile';
+        this.audience = 0.3;
+        this.audienceOffset = 0.0;
+        this.successMeasure = 'Gain qualitative feedback via a survey';
+        this.audienceCriteria = '30% of UK visitors to article page, on mobile, that haven\'t seen the message previously';
```
gtrufitt: This will show 15% of users at it'll segment into two buckets, one of which is noop, so if you actually want 30% then you'll need 0.6

gtrufitt:
Cool with me :+1:

static/src/javascripts/projects/common/modules/experiments/tests/userzoom-survey-message-mobile-v3.js
NathanielBennett: Yeah, he just wanted it to be 30% compared to the 20% on desktop, so I'll check with him but I'm guessing that's good

gtrufitt: They never asked me to update it to 20%, it's been at 10% (after segmentation) so if that's the case, you'll need to update the desktop audience too :)

PR #11802, opened 2016-02-02
Added inline merchandise test back into the default Commercial model

regiskuckaertz:
Inline merchandising components are not showing up anymore, because they're hidden by default _unless_ the page is sponsored.

/cc @kelvin-chappell


kelvin-chappell:
:+1:

PR #11803, opened 2016-02-02
Changing font urls to one with the correct mime type

NataliaLKB:
Currently the fonts we are requesting in AMP have the wrong Content-Type. Changing the url to this one fixes this issue, but I am unsure if it is the correct font. 

@zeftilldeath, can you check to see if this is correct?


paperboyo:
Kerning should be on, AFAIK...

NataliaLKB:
Thanks @paperboyo. I think @zeftilldeath agrees that we need kerning on. I will close this PR until we find a better solution.

PR #11789, opened 2016-02-01
[WIP] Move second call omniture data inline

dominickendrick:
This is WIP @sndrs

PR #11804, opened 2016-02-02
inject GU_U data into `guardian` object

sndrs:
Adds userdata that's pulled from the `GU_U` cookie to `guardian.config`, so that it can be shared across features that need it (soon, both showing the username and the main omniture call) without having to duplicate the read.

This is in prep for #11789

cc @dominickendrick


dominickendrick:
:+1:

nicl:
@sndrs I believe GU_U is going away soon. Might be worth checking with @guardian/identity on this.

sndrs:
@nicl ooh cheers

PR #11807, opened 2016-02-02
Update US Election logo

sammorrisdesign-zz:
To make this all look a little more intentional

screen shot 2016-02-02 at 15 03 10

screen shot 2016-02-02 at 15 03 07


zeftilldeath:
:+1: Much nicer.

PR #11808, opened 2016-02-02
Align paid-for containers with fronts grids

jbreckmckye:
This pull request changes the layout of paid-for containers on fronts from this -

image
image

to this -

image
image

Turns out we were using `fc-slice` without an `fc-slice-wrapper`. This also means we can remove an earlier fix for the widths of `commercial__body' elements.

### DIFF
--- a/commercial/app/views/contentapi/items.scala.html
+++ b/commercial/app/views/contentapi/items.scala.html
@@ -28,7 +28,8 @@
                     </a>
                 </div>
                 <div class="commercial__body">
-                    <ul class="u-unstyled l-row l-row--cols-4 l-row--layout-m fc-slice">
+                    <div class="fc-slice-wrapper">
+                        <ul class="u-unstyled l-row l-row--cols-4 l-row--layout-m fc-slice">
                         @for(item <- items) {
                             <li class="fc-slice__item l-row__item l-row__item--span-1">
                                 <div class="fc-item fc-item--has-image fc-item--has-metadata @item match {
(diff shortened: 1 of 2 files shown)


sammorrisdesign:
Looks like you've lost some of the internal styling too though. `line-height` on the headlines is massive now. Also the padding/margin around the logo bottom right has massively increased.

jbreckmckye:
@sammorrisdesign Somehow that completely passed me by.

jbreckmckye:
@sammorrisdesign Those styling differences are on master, too, it would seem. The two components are not actually the same; I am finding very difficult to reproduce the containers we see on prod on my local zone. It is... vexatious. Let me see if I can get a like for like comparison.

jbreckmckye:
@sammorrisdesign I've updated the description to compare like-for-like components.

commercial/app/views/contentapi/items.scala.html
```diff
@@ -28,7 +28,8 @@ <h3 class="commercial__title">
                     </a>
                 </div>
                 <div class="commercial__body">
-                    <ul class="u-unstyled l-row l-row--cols-4 l-row--layout-m fc-slice">
+                    <div class="fc-slice-wrapper">
```
regiskuckaertz: It's a shame we have to add a `DIV` for that. There are so many elements in the DOM, consuming memory for the sake of presentation; the opportunities to get rid of that are legions. Could `DIV.commercial__body.fc-slice-wrapper` work?

jbreckmckye: No. Although, it seems there are some templates we want to change that don't use fc-slice in the first place. So I may need another solution anyway.

PR #11813, opened 2016-02-02
Experimental new component architecture

OliverJAsh:
@desbo and I are working on a new component architecture to bring consistency to our front-end codebase. In this pull request I have converted the new sticky ad to use the experimental component architecture. @desbo is also going to have a go at rewriting a feature to use it.

The main principle of the new component architecture is that all DOM read/write side effects (events and DOM mutations) should be isolated to make it easier for developers to reason about code in any given component. This basically means:
- All DOM write side effects (DOM mutations) must go in a single `render` function
- DOM read side effects (events) should not be intertwined with DOM write side effects.

We have provided a `mountComponent` function which helps you to manage component state using a pattern very similar to [Redux](http://redux.js.org/) (the implementation is tiny). The basic idea is you define a pure reducer function which takes the previous state and an action object and returns the new state. In response to certain events (e.g. `click`) you then dispatch actions which will cause the state to update based on the reducer function, and then the render function will be invoked with the new state.

Here's what the component helper function looks like:

``` js
var counterTextEl = {};
var incrementBtnEl = {};
var decrementBtnEl = {};

mountComponent({
    initialState: 0,
    // Takes old state and returns new state. This should be pure
    reducer: (previousState, action) => {
        switch (action.type) {
            case 'INCREMENT':
                return previousState + 1;
            case 'DECREMENT':
                return previousState - 1;
            default:
                return previousState;
        }
    },
    render: state => {
        /* DOM write side effects */
        counterTextEl.textContent = state;
    },
    attachListeners: dispatch => {
        /* DOM read side effects (events) which dispatch actions */
        incrementBtnEl.addEventListener('click', () => dispatch({ type: 'INCREMENT' }));
        decrementBtnEl.addEventListener('click', () => dispatch({ type: 'DECREMENT' }));
    }
});
```

Full example at http://jsbin.com/kepetu/2/edit?html,js,output.

Once this experiment is finished and we have settled on the right component architecture, we will be sure to communicate our findings to the team.

Feedback is very welcome. Remember this is still early stages! :smile:

### DIFF
--- a/static/src/javascripts/bootstraps/enhanced/common.js
+++ b/static/src/javascripts/bootstraps/enhanced/common.js
@@ -139,7 +139,8 @@ define([
 
             initialiseStickyHeader: function () {
                 if (config.switches.viewability
-                    && !(config.page.isProd && config.page.contentType === 'Interactive')
+                    && !(config.switches.disableStickyNavOnMobile && detect.getBreakpoint() === 'mobile')
+                    && config.page.contentType !== 'Interactive'
                     && config.page.contentType !== 'Crossword'
                     && (!config.switches.newCommercialContent || !config.page.isAdvertisementFeature)
                     && config.page.pageId !== 'offline-page') {
--- a/static/src/javascripts/projects/common/modules/navigation/newSticky.js
+++ b/static/src/javascripts/projects/common/modules/navigation/newSticky.js
@@ -177,41 +177,38 @@ define([
         });
     };
 
+    var reducer = function (previousState, action) {
+        switch (action.type) {
+            case 'SCROLL':
+                return assign({}, previousState, {
+                    firstRender: false,
+                    previousAdHeight: previousState.adHeight,
+                    pageXOffset: action.scrollCoords[0],
+                    pageYOffset: action.scrollCoords[1]
+                });
+            case 'NEW_AD_HEIGHT':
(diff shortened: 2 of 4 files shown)


static/src/javascripts/bootstraps/enhanced/common.js
```diff
@@ -139,7 +139,8 @@ define([
 
             initialiseStickyHeader: function () {
                 if (config.switches.viewability
-                    && !(config.page.isProd && config.page.contentType === 'Interactive')
```
OliverJAsh: Not sure why we only disabled the sticky nav for interactive pages in prod. @uplne, any ideas? (Re. https://github.com/guardian/frontend/commit/8b5f2ddd15cfa2e85cf460d20a5286b4b01558a4)

OliverJAsh:
We've decided to iterate on this more before using [a fork of frontend](https://github.com/OliverJAsh/frontend) before merging anything into master, so I'm closing this. Please redirect eyes/feedback to https://github.com/OliverJAsh/frontend/pull/1

static/src/javascripts/bootstraps/enhanced/common.js
jamesgorrie: This would also return true for `isDev && config.page.contentType !== 'Interactive'`.

PR #11819, opened 2016-02-03
Use discussion subdomain

nicl:
The result of this change is that we no longer need to proxy through the
www.theguardian.com subdomain; all the special logic here can be
removed once this goes live.

As part of this PR two switches are removed as they are no longer
required:
- discussion https flag (we are https-only now as our auth mechanism
  requires this)
- discussion proxy (the new subdomain and also move to always-https mean
  that the proxy is no longer required)

cc @guardian/discussion @johnduffell


discussion/test/package.scala
```diff
@@ -57,7 +58,13 @@ object DiscussionApiHttpRecorder extends HttpRecorder[WSResponse] {
...
   import play.api.Play.current
   protected val clientHeaderValue: String =""
-  protected val apiRoot = conf.Configuration.discussion.apiRoot
+
+  protected val apiRoot =
+    if (Configuration.environment.isProd)
+      Configuration.discussion.apiRoot
+    else
+      Configuration.discussion.apiRoot.replaceFirst("https://", "http://") // CODE SSL cert is defective and expensive to fix
+
```
jennysivapalan: If this is a test don't we just want the non prod one?

jennysivapalan:
:+1: :speech_balloon:

satterly:
:+1:

discussion/test/package.scala
nicl: Hmm, maybe. I wasn't sure what people might do for testing!

PR #11820, opened 2016-02-03
Initialise inlined embeds to for inline styling

rich-nguyen:
There was a missing class that deals with initial styling for some embeds, `facia-snap-embed`. Furthermore, the javascript inlines some height styling after initial render, so this needs to be executed for both inline and normal embeds on fronts.
### before

screen shot 2016-02-03 at 13 33 24
### after

screen shot 2016-02-03 at 13 34 02


OliverJAsh:
:+1:

PR #11822, opened 2016-02-03
corrent mimetype for hls

jamesgorrie:
We're not currently taking advantage, on iOS, that we have HLS.
This allows iOS to detect and use it.

CC @fredex42

### DIFF
--- a/common/app/model/Encoding.scala
+++ b/common/app/model/Encoding.scala
@@ -5,7 +5,8 @@ case class Encoding(format: String, url: String, rawFormat: String)
 object Encoding {
 
   val typeMapping = Map(
-    "mp4" -> "video/mp4"
+    "mp4" -> "video/mp4",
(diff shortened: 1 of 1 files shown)


rich-nguyen:
:+1: keep an eye on ios video play. probably need a graph that reports on the success rate of video plays, outside of omniture.

jamesgorrie:
Yea - I need to have a chat to a few people (including yourself) about stats.

jamesgorrie:
@rich-nguyen Would you say Ophan (we have it in there) is good enough for this and non-blocking?

rich-nguyen:
yep, but what you'll need to assess the trend longer than two weeks.

jamesgorrie:
I'll be getting a rolling report for the video starts and ends for the next two weeks from omniture.

PR #11825, opened 2016-02-03
Fix logos on sponsored galleries

kelvin-chappell:
They got lost on sponsored galleries when the paid content changes went out.

/cc @Calanthe @jbreckmckye


Calanthe:
:+1:

PR #11775, opened 2016-02-01
Update style of existing paid content containers

kelvin-chappell:
Before:
image

Now:
image

/cc @Calanthe @jbreckmckye

PR #11826, opened 2016-02-03
Fix paid containers on paid fronts

kelvin-chappell:
They should no longer look like paid containers.

This is just a revert of https://github.com/guardian/frontend/pull/11775 and a removal of the _fc-container--advertisement-feature_ class from containers so that they blend in on paid fronts.

Before:
image

Now:
image

/cc @Calanthe @jbreckmckye


Calanthe:
:+1:

PR #11827, opened 2016-02-03
Unique id for sponsored merch component toggle

uplne:
CC @kelvin-chappell

### DIFF
--- a/common/app/views/fragments/commercial/paidForMeta.scala.html
+++ b/common/app/views/fragments/commercial/paidForMeta.scala.html
@@ -1,4 +1,4 @@
-@(dataId: String = s"")(implicit request: RequestHeader)
+@(dataId: Option[String] = None)(implicit request: RequestHeader)
 
 @import views.html.fragments.inlineSvg
 
--- a/common/app/views/fragments/commercial/paidForMeta.scala.html
+++ b/common/app/views/fragments/commercial/paidForMeta.scala.html
@@ -1,12 +1,12 @@
-@()(implicit request: RequestHeader)
+@(dataId: String = s"")(implicit request: RequestHeader)
 
 @import views.html.fragments.inlineSvg
 
 <div class="paidfor-meta">
     <span class="paidfor-meta__label">Paid content</span>
     <div class="paidfor-label paidfor-meta__more has-popup">
-        <button class="u-button-reset paidfor-label__btn popup__toggle" data-toggle="js-paidfor-popup">About @inlineSvg("arrow-down", "icon")</button>
-        <div class="popup is-off paidfor-popup js-paidfor-popup">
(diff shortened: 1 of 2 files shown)


common/app/views/fragments/commercial/paidForMeta.scala.html
```diff
@@ -1,12 +1,12 @@
-@()(implicit request: RequestHeader)
+@(dataId: String = s"")(implicit request: RequestHeader)
```
kelvin-chappell: Don't need the `s` prefix if it's a string literal.  
But if this doesn't need to be specified by the caller it would be better to have the type `Option[String]` with a default value of `None`.

common/app/views/fragments/containers/facia_cards/commercialContainer.scala.html
```diff
@@ -13,13 +13,12 @@
...
     case SingleCampaign(_) => {
         @if(Switches.NewCommercialContent.isSwitchedOn) {
-
         <div class="fc-container fc-container--commercial">
             <div class="ad-slot ad-slot--dfp ad-slot--merchandising ad-slot--commercial-component">
                 <div class="commercial commercial--paidfor commercial--tone-paidfor commercial--paidfor-single">
                     <div class="commercial__inner">
                         <div class="commercial__header">
-                            @fragments.commercial.paidForMeta()
+                            @fragments.commercial.paidForMeta(container.dataId)
```
kelvin-chappell: Then this would be `Some(container.dataId)`

kelvin-chappell:
:+1: with changes

Calanthe:
What about other templates which use `paidForMeta` too? Like `/commercial/app/views/contentapi/item.scala.html`, `/commercial/app/views/contentapi/items.scala.html`, `/common/app/views/fragments/guBand.scala.html` and also `/common/app/views/fragments/containers/facia_cards/commercialContainer.scala.html` line 87?

kelvin-chappell:
With a default argument the other templates should just work, but I think @Calanthe has a conflicting change which also fixes this problem.

uplne:
@Calanthe others should be ok as they are always only one module at the time in the page. These commercialContainers are basically normal containers on fronts so there can be many of them.

I believe it's not worth it to create some complex solution for it now. If such a case will come in the future we can think about it.

I am going to fix the line 87. Nice catch.

Calanthe:
@uplne it won't work on the test page because we have many of components there

I fixed that issue by adding a random value to the data attr: https://github.com/guardian/frontend/blob/paid-content-card-with-CAPI/common/app/views/fragments/commercial/paidForMeta.scala.html

uplne:
Sure but that is against our caching strategy, isn't it? Page with component with this random attribute will be different for each page view.

Calanthe:
@uplne yeah you are right. I will remove my changes.

PR #11828, opened 2016-02-03
Deploys radiator: group commits by commit author username, only show …

OliverJAsh:
…name

We used to group commits by author name consecutively. Now we just group commits by author name (including non-consecutive commits), and only show the name. This will simplify the interface and stop names from appearing twice.

If your name appears incorrectly on the dashboard, please check `git config --global user.name` and also your GitHub profile username (as this is the user often used for the merge commit).
# After

image


gtrufitt:
:+1: Rad.

NataliaLKB:
:+1: :ship:

PR #11830, opened 2016-02-03
Add election top stories to minute onwards container

gtrufitt:
Utilising some of the work from @Jholder112233 to pull the us election top stories into the minute onwards container. Josh - does this look ok? Am I doing anything silly? Should I be importing some other module and using that or anything?


Jholder112233:
Gone but never forgotten :gem: 

:+1: assuming it works as expected

PR #11831, opened 2016-02-03
Paid content card with cAPI

Calanthe:
This PR adds the new way of loading content for the paid content Card. Now instead of manual typing in DFP, whole data will be taken from cAPi (based on the short link url). The article image, header and text can be manually overwritten via DFP. Link to the new DFP template: https://www.google.com/dfp/main?networkCode=59666047&networkCode=59666047#delivery/CreateCreativeTemplate/creativeTemplateId=10041927

screen shot 2016-02-03 at 15 27 43

When adops will move content to this new template I will remove the old `paidfor-content` component.

Also I added a js unit test but the whole test in `commercial-component.spec.js` is turned off and it seems  like a bigger issue, which needs to be fixed in the near future. 

cc @kelvin-chappell


kelvin-chappell:
:+1:

PR #11833, opened 2016-02-03
Frontend fetches breaking-news content from /news-alert/alerts

TBonnin:
/news-alert/alerts is returing the content of a json file
stored on S3 which is generated by the admin-jobs app


gtrufitt:
:+1:

PR #11834, opened 2016-02-03
Move adblock detection to non-blocking inline JS, and store it on global `guardian` object

sndrs:
This is more preparation for #11789 (removing omniture topup call)

Move adblock detection code to the bottom of the page, so that it can also be used by the inline omniture call.

has the bonus of only adding one test element.

cc @Calanthe @uplne @jbreckmckye @dominickendrick @regiskuckaertz

### DIFF
--- a/common/app/templates/inlineJS/nonBlocking/detectAdblock.scala.js
+++ b/common/app/templates/inlineJS/nonBlocking/detectAdblock.scala.js
@@ -1,0 +1,24 @@
+try {
+    ((document, window) => {
+        var ad = document.createElement('div');
+        ad.style.position = 'absolute';
+        ad.style.left = '-9999px';
+        ad.style.height = '10px';
+        ad.innerHTML = '&nbsp;';
+        ad.setAttribute('class', 'ad_unit');
+
+        window.requestAnimationFrame(() => {
+            document.body.appendChild(ad);
+
+            window.requestAnimationFrame(() => {
+                var adStyles = window.getComputedStyle(ad);
+                window.guardian.adBlock = {
+                    active: adStyles.getPropertyValue('display') === 'none',
+                    ffAdblockPlusInstalled: adStyles.getPropertyValue('-moz-binding').match('elemhidehit') !== null
+                }
+            })
+        });
+    })(document, window);
+} catch (e) {
+    @if(play.Play.isDev) {throw (e)}
+}
--- a/common/app/templates/inlineJS/nonBlocking/detectAdblock.scala.js
+++ b/common/app/templates/inlineJS/nonBlocking/detectAdblock.scala.js
@@ -12,9 +12,9 @@ try {
 
             window.requestAnimationFrame(() => {
                 var adStyles = window.getComputedStyle(ad);
-                window.guardian.adBlock = {
-                    active: adStyles.getPropertyValue('display') === 'none',
-                    ffAdblockPlusInstalled: adStyles.getPropertyValue('-moz-binding').match('elemhidehit') !== null
+                window.guardian.adBlockers = {
+                    generic: adStyles.getPropertyValue('display') === 'none',
+                    ffAdblockPlus: adStyles.getPropertyValue('-moz-binding').match('elemhidehit') !== null
                 }
             })
         });
--- a/static/src/javascripts/projects/common/utils/detect.js
+++ b/static/src/javascripts/projects/common/utils/detect.js
@@ -397,30 +397,18 @@ define([
     }
 
     function adblockInUse() {
-        if (!detect.cachedAdblockInUse) {
-            var sacrificialAd = createSacrificialAd(),
-                contentBlocked = isHidden(sacrificialAd);
-            sacrificialAd.remove();
-            detect.cachedAdblockInUse = contentBlocked;
-            return contentBlocked;
+        if (window.guardian.adBlocker) {
+            return window.guardian.adBlocker.generic;
         }
-
-        return detect.cachedAdblockInUse;
-
-        function isHidden(bonzoElement) {
-            return bonzoElement.css('display') === 'none';
-        }
+        return false;
     }
 
     /** Includes Firefox Adblock Plus users who whitelist the Guardian domain */
     function getFirefoxAdblockPlusInstalled() {
-        var sacrificialAd = createSacrificialAd();
-        var adUnitMozBinding = sacrificialAd.css('-moz-binding');
-        if (adUnitMozBinding) {
-            return adUnitMozBinding.match('elemhidehit') !== null;
-        } else {
-            return false;
+        if (window.guardian.adBlocker) {
+            return window.guardian.adBlocker.ffAdblockPlus;
         }
+        return false;
     }
 
     function createSacrificialAd() {
--- a/static/src/javascripts/projects/common/utils/detect.js
+++ b/static/src/javascripts/projects/common/utils/detect.js
@@ -397,16 +397,16 @@ define([
     }
 
     function adblockInUse() {
-        if (window.guardian.adBlocker) {
-            return window.guardian.adBlocker.generic;
+        if (window.guardian.adBlockers) {
+            return window.guardian.adBlockers.generic;
         }
         return false;
     }
 
     /** Includes Firefox Adblock Plus users who whitelist the Guardian domain */
     function getFirefoxAdblockPlusInstalled() {
-        if (window.guardian.adBlocker) {
-            return window.guardian.adBlocker.ffAdblockPlus;
+        if (window.guardian.adBlockers) {
+            return window.guardian.adBlockers.ffAdblockPlus;
         }
         return false;
     }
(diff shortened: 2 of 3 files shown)


static/src/javascripts/projects/common/utils/detect.js
```diff
@@ -397,36 +397,12 @@ define([
...
-            detect.cachedAdblockInUse = contentBlocked;
-            return contentBlocked;
-        }
-
-        return detect.cachedAdblockInUse;
-
-        function isHidden(bonzoElement) {
-            return bonzoElement.css('display') === 'none';
-        }
+        return !!window.guardian.adBlockers && window.guardian.adBlockers.generic;
```
jamesgorrie: Isn't `window.guardian.adBlockers` always going to be set?

sndrs: it should be yeah – this is just a massive guard against this code running before its gets set. it shouldn't ever, because `detect` (currently) does not run till domready

it's really a just-in-case

common/app/templates/inlineJS/nonBlocking/detectAdblock.scala.js
```diff
@@ -0,0 +1,24 @@
+try {
+    ((document, window) => {
+        var ad = document.createElement('div');
+        ad.style.position = 'absolute';
+        ad.style.left = '-9999px';
```
sndrs: this bugs me - will try a switch if this merges to use a different method to hide it. i'm curious as to whether it'll [affect iOS crashes](http://www.zeldman.com/2012/03/01/replacing-the-9999px-hack-new-image-replacement/)

static/src/javascripts/projects/common/utils/detect.js
jamesgorrie: :ghost:

common/app/templates/inlineJS/nonBlocking/detectAdblock.scala.js
gtrufitt: :thought_balloon:

sndrs: dreaming of a :smoking::gun:

jbreckmckye: I thought I'd committed a fix to the -9999px hack - evidently not. I believed that some older user agents would reserve a large canvas in memory, and this could cause scroll jank.

gtrufitt: Aye, this is a known performance issue, but whether it causes the crashes... We'll see. Would be amazing if it was!

common/app/templates/inlineJS/nonBlocking/detectAdblock.scala.js
```diff
@@ -0,0 +1,24 @@
...
+
+            window.requestAnimationFrame(() => {
+                var adStyles = window.getComputedStyle(ad);
+                window.guardian.adBlockers = {
+                    generic: adStyles.getPropertyValue('display') === 'none',
+                    ffAdblockPlus: adStyles.getPropertyValue('-moz-binding').match('elemhidehit') !== null
+                }
+            })
+        });
+    })(document, window);
```
jamesgorrie: Is it worth detroying the dom element once you have the info you need about it?

sndrs: yeah maybe – it's not doing a lot though, might cause less upset to the page to just leave it?

common/app/templates/inlineJS/nonBlocking/detectAdblock.scala.js
```diff
@@ -0,0 +1,24 @@
...
+    ((document, window) => {
+        var ad = document.createElement('div');
+        ad.style.position = 'absolute';
+        ad.style.left = '-9999px';
+        ad.style.height = '10px';
+        ad.innerHTML = '&nbsp;';
+        ad.setAttribute('class', 'ad_unit');
+
+        window.requestAnimationFrame(() => {
+            document.body.appendChild(ad);
```
jbreckmckye: I know this should be fine in practice, but it does feel a bit weird to have this running in a loosely asynchronous fashion.

sndrs: agreed – i actually began converting `detect.adblockInUse`/`getFirefoxAdblockPlusInstalled` into promises, but that means making a quite a bit of a change elsewhere.

any ideas?

regiskuckaertz: I wonder if you really need any `rAF` call. since the script is inline at the top of the body element, the browser is anyway currently building the DOM, and so the render tree is itself either inexistant or incomplete.

common/app/templates/inlineJS/nonBlocking/detectAdblock.scala.js
```diff
@@ -0,0 +1,24 @@
...
+
+            window.requestAnimationFrame(() => {
```
regiskuckaertz: In fact you might want to wait for the `DOMContentLoaded` event here. I know the browser is doing speculative work ahead, but are you sure at this point the `ad_unit` class is already parsed?

jbreckmckye: Actually, you are not waiting for the browser to apply styles to `ad_unit`. You are waiting for the browser to send a DOM mutation event of some kind to a listening extension, and for that extension to mutate the DOM on behalf of the user. There are actually quite a few opportunities for race conditions, and this is one reason why we're not currently able to track AdblockPlus in Internet Explorer.

sndrs: in my tests, ghostery doesn't even seem to hide it?

regiskuckaertz: waw o_O

regiskuckaertz: so instead you could attach a mutation observer to the element—or mutation events for IE9–10—couldn't you? According to PageFair, visitors using IE with an adblocker are a very small minority.

common/app/templates/inlineJS/nonBlocking/detectAdblock.scala.js
regiskuckaertz: try [this](http://www.zeldman.com/2012/03/01/replacing-the-9999px-hack-new-image-replacement/) perhaps

common/app/templates/inlineJS/nonBlocking/detectAdblock.scala.js
jamesgorrie: Non-plussed, just asking.

common/app/templates/inlineJS/nonBlocking/detectAdblock.scala.js
jbreckmckye: This was the subject of a long and fairly involved investigation. For the moment, I don't believe we're happy with holding the omniture call up with an asynchronous function.

sndrs: the second one at least is needed to be sure the element has gone into the DOM? will come and talk offline about the investigation

regiskuckaertz: Do ad blockers change the value of the `style` attribute? If yes, then you certainly don't need/want to wait until the browser has calculated layout: you can be notified long before, using mutation observers (even mutation events for older browsers—in this case it's rather safe). Then you read the value of the `style` attribute, **not** `getComputedStyle`. Here's a quick [secret gist](https://gist.github.com/regiskuckaertz/2c69fd7df2394c26ea57).

jbreckmckye: @regiskuckaertz We did test whether we could use mutation observers to intercept ABP changes, at least when trying to improve FF ABP false-positives, and the answer was: no. This might be a design decision not to allow web pages to sniff information about running plugins.

We spent several weeks investigating Adblock detection; I can show you the data we collected, if you like. Long story short - it's complicated, different plugins have different implementations, and the only real commonality you can rely upon is that they all reject connections to URIs matching particular text patterns.

sndrs: after a few good chats offline, I'm going to update this account fully for the async nature of the test (reading and writing to the DOM), but I won't be changing the technique. 

that's a whole other can of worms, and moving this is really only to allow us to consolidate the omniture calls, rather than improve detection. 

a good discussion and v interesting area though for sure...

common/app/templates/inlineJS/nonBlocking/detectAdblock.scala.js
jbreckmckye: Let's not fix this as part of this PR

sndrs:
closing this to reopen it in smaller chunks, incorporating async

PR #11835, opened 2016-02-03
Video contrib ads update

jamesgorrie:
After chatting to people -> [this](https://github.com/guardian/videojs-ima/pull/1) always seems to be the least sustainable way to manage dependencies.

At least this way the hack is abstracted out into our code base as opposed to trying to wedge it into someone else's.

The reason for this is to finally move to bump the version on the [google-ima](https://github.com/googleads/videojs-ima) to try and get prerolls working better (and to be able to do it easily later).


jamesgorrie:
CC @Calanthe - what do you think?

dominickendrick:
:+1:

PR #11836, opened 2016-02-03
Make sure dynamically inserted add slots are lazyloaded

regiskuckaertz:
Whether ads should be lazyloaded or not, currently `DfpApi.addSlot` will synchronously display an ad for each dynamically inserted slot. This PR changes that and allows the scroll handler to deal with them.

cc @uplne @jbreckmckye @Calanthe


static/src/javascripts/projects/common/modules/commercial/dfp-api.js
```diff
@@ -412,7 +423,11 @@ define([
                         isLoading: false,
                         slot: defineSlot($adSlot)
```
uplne: What about putting this inside the else branch so we won't run defineSlot if we don't need to.

uplne:
:+1:

PR #11837, opened 2016-02-03
AMP fonts and twitter

zeftilldeath:
Coloured in the twitter bird, and changed the medium weight of the font location.


NataliaLKB:
:+1:

PR #11838, opened 2016-02-03
The Minute: Update twitter loading to prevent fastdom error

gtrufitt:
I had to remove `js-article_body` (as per liveblog approach, where needed we will add the new class `js-article__body--minute-article`) but this broke twitter. Re-adding the minute-article specific body class fixes this.


sndrs:
:+1:

PR #11840, opened 2016-02-03
Moving AB test tracking to first omniture call

dominickendrick:
This PR:
- Adds you previous page ab test participations into the current page omniture tracking
- Fires a second Omniture call if you are added or removed from a test
- Does not fire a second call if your test allocation does not change

Depends on  #11789
## QA
- [ ] With no current test, if I'm not allocated to any new test, a 2nd tracking call shouldn't be fired
- [ ] With some current tests, if I'm not allocated to any new test, a 2nd tracking call shouldn't be fired
- [ ] With no current tests, if I am allocated to any new test, a 2nd tracking call should fire
- [ ] With some current tests, if I am allocated to any new test, a 2nd tracking call should fire
- [ ] With some current tests, if I am removed from a test, a second tracking call should fire

### DIFF
--- a/common/app/templates/inlineJS/blocking/analytics.scala.js
+++ b/common/app/templates/inlineJS/blocking/analytics.scala.js
@@ -130,6 +130,48 @@ try {
         s.prop20    = tpA[2] + ':' + tpA[1];
         s.eVar20    = 'D=c20';
 
+        var participationsKey = 'gu.ab.participations';
+        var abTestsParticipations;
+
+        try {
+            var participations = window.localStorage.getItem(participationsKey);
+
+            abTestsParticipations = makeOmnitureABTag(participations);
+
+            s.eVar51    = abTestsParticipations;
+            s.list1     = abTestsParticipations;
+
+            // This is set globally so we can check if the use ab test participations change once the ab test runs.
+            // If it does, we fire a second tracking call in modules/analytics/omniture.js
+            guardian.config.abTestsParticipations = abTestsParticipations;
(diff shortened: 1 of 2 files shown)


dominickendrick:
@sndrs @desbo 
@ScottPainterGNM it would be good to verify the test plan with you!

common/app/templates/inlineJS/blocking/analytics.scala.js
```diff
@@ -130,6 +130,43 @@ try {
...
+            // If it does, we fire a second tracking call in modules/analytics/omniture.js
+            guardian.config.abTestsParticipations = abTestsParticipations;
+        } catch (e) { }
+
+        function makeOmnitureABTag(currentParticipations) {
+            var participations = JSON.parse(currentParticipations);
+            var tag = [];
+
+            for (var key in participations.value) {
+                tag.push(['AB', key, participations.value[key].variant].join(' | '));
```
sndrs: might want `.hasOwnProperty(key)` check in these?

desbo: the result of json.parse isn't part of any prototype chain so we don't need any `hasOwnProperty` check.
screen shot 2016-02-04 at 10 59 11

sndrs: very true, it's probably overkill. ignore me :boom:

johnduffell:
this sounds like a great idea, but is this to :ship: ? or :closed_book: ? @dominickendrick

PR #11841, opened 2016-02-04
Removed obsolete tracking code in commercial components

regiskuckaertz:
Neither `OASToken` nor `OmnitureToken` are used anywhere—the `clickMacro` is set beforehand by the scala template and Omniture is already tracking everything (this has been confirmed)


jbreckmckye:
:+1:

PR #11842, opened 2016-02-04
Updated copy for the membership supporter banner for the US edition.

paulbrown1982:
The engaged reader banner appears on the US edition homepage for logged out customers. Its copy is slightly wrong, and this change corrects that, as per https://trello.com/c/F4RsgETQ/354-copy-change-in-front-end-engaged-user-banner-served-in-the-us


regiskuckaertz:
:+1:

PR #11843, opened 2016-02-04
update videojs ima plugin

jamesgorrie:
Finally updating the IMA plugin.

This should hopefully help us stay ontop of the videojs progressions in the future.
We will be writing tests too to make sure updates in the future don't muck anything up.


NataliaLKB:
:+1:

PR #11844, opened 2016-02-04
Improve commercial container card code

kelvin-chappell:
This is a refactor to try to make the code more readable.

/cc @kenlim


JonNorman:
A lot easier to follow than previously:+1:

PR #11845, opened 2016-02-04
Merch components soulmates cleanup

uplne:
Removing some unused styles and sprites as we moved to the new left side design.
screen shot 2016-02-03 at 17 36 11


Calanthe:
nice :+1:

PR #11847, opened 2016-02-04
Removes unused div container for plista/outbrain

JonNorman:
**Issue**
Only one of Plista and Outbrain widgets will be served on the page. The original Scala HTML template contained two container divs - one for each widget. As a result, there would be an empty container being rendered regardless of which widget is being served.

**Fix**
The widget container div is now rendered immediately prior to the initialisation of the respective Plista/Outbrain widget, thus only one div is being rendered.

**Screenies**
_before:_
image

_after:_
image

### DIFF
--- a/article/test/ArticleFeatureTest.scala
+++ b/article/test/ArticleFeatureTest.scala
@@ -603,28 +603,5 @@ import collection.JavaConversions._
       }
     }
 
-    scenario("Outbrain") {
-
-      Given("I am on an article")
-      OutbrainSwitch.switchOn()
-      goTo("/society/2014/oct/15/lord-freud-unreserved-apology-comment-disabled-people-mimimu-wage") {
-        browser =>
-          import browser._
-          Then("Then the Outbrain placeholder should be rendered")
-          var outbrainPlaceholder = $(".js-outbrain")
-          outbrainPlaceholder.length should be(1)
-      }
-
-      Given("I am on a live blog")
-      goTo("/politics/blog/live/2014/oct/15/cameron-and-miliband-at-pmqs-politics-live-blog") {
-        browser =>
-          import browser._
-          Then("Then the Outbrain placeholder should not be rendered")
-          $(".js-outbrain").isEmpty should be(true)
-
-      }
-
-    }
-
   }
 }
--- a/static/src/javascripts/projects/common/modules/commercial/third-party-tags.js
+++ b/static/src/javascripts/projects/common/modules/commercial/third-party-tags.js
@@ -41,20 +41,20 @@ define([
     function loadExternalContentWidget() {
 
         var externalTpl = template(externalContentContainerStr);
-        var documentAnchorId = '.js-external-content-widget-anchor';
+        var documentAnchorClass = '.js-external-content-widget-anchor';
 
         function renderWidgetContainer(widgetType) {
-            $(documentAnchorId).append(externalTpl({widgetType: widgetType}));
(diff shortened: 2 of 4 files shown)


static/src/javascripts/projects/common/modules/commercial/third-party-tags.js
```diff
@@ -28,13 +33,28 @@ define([
...
+
+        function renderWidgetContainer(widgetType) {
+            $(documentAnchorId).append(externalTpl({widgetType: widgetType}));
+        }
+
         if (config.switches.plistaForOutbrainAu && config.page.edition.toLowerCase() === 'au') {
-            plista.init();
+            fastdom.write(function () {
+                renderWidgetContainer('plista');
+            }).then(plista.init());
```
desbo: You might want  `then(plista.init)` here, otherwise it'll be invoked on parse rather than after `then`.

JonNorman: Ah thanks - meant to have removed that line.

article/test/ArticleFeatureTest.scala
```diff
@@ -603,28 +603,5 @@ import collection.JavaConversions._
       }
     }
 
-    scenario("Outbrain") {
```
jbreckmckye: Anything to kill Selenium tests is always a win in my book :heavy_plus_sign:

static/src/javascripts/projects/common/modules/commercial/third-party-tags.js
JonNorman: Thanks - didn't realise that it required a function pointer but it seems obvious now.

static/src/javascripts/projects/common/modules/commercial/third-party-tags.js
```diff
@@ -28,13 +33,28 @@ define([
...
-    plista) {
+    plista,
+    externalContentContainerStr,
+    fastdom,
+    template) {
 
     function loadExternalContentWidget() {
+
+        var externalTpl = template(externalContentContainerStr);
+        var documentAnchorId = '.js-external-content-widget-anchor';
```
jbreckmckye: It's not an ID anymore...

jbreckmckye:
Aside from comments, looks great :ship:

static/src/javascripts/projects/common/modules/commercial/third-party-tags.js
```diff
@@ -14,8 +15,12 @@ define([
     'common/modules/commercial/third-party-tags/krux',
     'common/modules/identity/api',
     'common/modules/commercial/third-party-tags/outbrain',
-    'common/modules/commercial/third-party-tags/plista'
+    'common/modules/commercial/third-party-tags/plista',
+    'text!common/views/commercial/external-content.html',
+    'common/utils/fastdom-promise',
```
uplne: Utils should be next to other utils just to keep it clean. If you look across another components you will find the pattern is:

Promise,
utils,
modules,
text

JonNorman: Good to know - made the changes.

regiskuckaertz:
:+1: but let's flag the issue where you can just add a `fc-container--${widgetType}` class instead of creating new elements for when you've finished testing (lodash templates and DOM mutations are expensive).

JonNorman:
@regiskuckaertz agreed - I think that there might be a fair amount of tinkering with the Plista widget now that it is available in Code - no doubt this will be revisited.

PR #11849, opened 2016-02-04
Fix profile avatars

nicl:
At the moment, when viewing another user's profile page, a user sees _their own_ avatar. E.g. see:

https://profile.theguardian.com/user/id/1678506

This bug was introduced as part of: https://github.com/guardian/frontend/pull/11780

This commit fixes this bug.


NataliaLKB:
Once the tests are :white_check_mark:  :+1:

nicl:
@NataliaLKB thanks!

PR #11852, opened 2016-02-04
Insert ads in liveblogs at the correct location

regiskuckaertz:
The computation of the position at which the first dynamic ad should be inserted was off because all the coordinates are relative to the top of the `.js-liveblog-body`. The only way to get around this was to add a new spacefinder rule that will reject every slot that is located in a certain range from the top of the page. This is done via the `absoluteMinAbove` rule.

Additionally, I refactored the code to remove useless variables and function calls as well as reduce the footprint of code running inside animation frames.

/cc @michaelwmcnamara @uplne @jbreckmckye @Calanthe

### DIFF
--- a/static/src/javascripts/projects/common/modules/article/spacefinder.js
+++ b/static/src/javascripts/projects/common/modules/article/spacefinder.js
@@ -78,8 +78,6 @@ define([
             new Promise(function (resolve) {
                 var loadedCount = 0;
                 bean.on(body, 'load', notLoaded, function onImgLoaded(e) {
-                    var idx = notLoaded.indexOf(e.currentTarget);
-                    if (idx === -1) return;
                     loadedCount += 1;
                     if (loadedCount === notLoaded.length) {
                         bean.off(body, 'load', onImgLoaded);
--- a/static/src/javascripts/projects/common/modules/article/spacefinder.js
+++ b/static/src/javascripts/projects/common/modules/article/spacefinder.js
@@ -77,7 +77,7 @@ define([
             Promise.resolve(true) :
             new Promise(function (resolve) {
                 var loadedCount = 0;
-                bean.on(body, 'load', notLoaded, function onImgLoaded(e) {
+                bean.on(body, 'load', notLoaded, function onImgLoaded() {
                     loadedCount += 1;
                     if (loadedCount === notLoaded.length) {
                         bean.off(body, 'load', onImgLoaded);
@@ -89,9 +89,7 @@ define([
     }
 
     function onRichLinksUpgraded(body) {
-        var unloaded = qwery('.element-rich-link--not-upgraded', body);
-
-        return unloaded.length === 0 ?
+        return qwery('.element-rich-link--not-upgraded', body).length === 0 ?
             Promise.resolve(true) :
             new Promise(function (resolve) {
                 mediator.once('rich-link:loaded', resolve);
--- a/static/src/javascripts/projects/common/modules/article/spacefinder.js
+++ b/static/src/javascripts/projects/common/modules/article/spacefinder.js
@@ -118,11 +118,11 @@ define([
     }
 
     function _enforceRules(slots, rules, bodyTop, bodyHeight) {
-        var filtered;
+        var filtered = Promise.resolve(true);
 
         // enforce absoluteMinAbove rule
         if (rules.absoluteMinAbove > 0) {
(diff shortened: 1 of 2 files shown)


static/src/javascripts/projects/common/modules/article/spacefinder.js
```diff
@@ -57,46 +58,41 @@ define([
...
-                });
-            })),
-            new Promise(function (resolve) {
-                window.setTimeout(resolve, 5000);
-            })
-        ]));
+        return notLoaded.length === 0 ?
+            Promise.resolve(true) :
+            Promise.race(
+                [new Promise(expire)].concat(
```
uplne: So in here you are interested only in one image or all images?

regiskuckaertz:
@uplne you were right :smile: thanks! also I thought about some more improvements.

static/src/javascripts/projects/common/modules/article/spacefinder.js
```diff
@@ -57,46 +62,38 @@ define([
...
-                    resolve();
-                } else {
-                    reject();
-                }
-            }).catch(function () {
-                return new Promise(function (resolve) {
-                    mediator.once('rich-link:loaded', resolve);
-                });
-            }),
+        return qwery('.element-rich-link--not-upgraded', body).length === 0 ?
```
uplne: Is this class used for styling too? We like to use js-hook but I can imagine that in this situation it maybe weird to do?!

static/src/javascripts/projects/common/modules/article/spacefinder.js
```diff
@@ -120,25 +117,36 @@ define([
...
+            }));
+        }
+
         // enforce minAbove and minBelow rules
-        filtered = Promise.resolve(filter(slots, function (slot) {
-            var farEnoughFromTopOfBody = slot.top >= rules.minAbove,
-                farEnoughFromBottomOfBody = slot.top + rules.minBelow <= bodyHeight;
-            return farEnoughFromTopOfBody && farEnoughFromBottomOfBody;
-        }));
+        filtered = filtered.then(function (slots) {
```
uplne: if (rules.absoluteMinAbove > 0)  then filtered is undefined not a promise at this moment

static/src/javascripts/projects/common/modules/article/spacefinder.js
```diff
@@ -120,25 +117,36 @@ define([
...
+            return filter(slots, function (slot) {
+                var farEnoughFromTopOfBody = slot.top >= rules.minAbove,
+                    farEnoughFromBottomOfBody = slot.top + rules.minBelow <= bodyHeight;
```
uplne: Let's use multiple vars as we decided lately. In this case I would actually find it really helpful as they are quite long lines to read.

static/src/javascripts/projects/common/modules/article/spacefinder.js
regiskuckaertz: something for dotcom I guess

static/src/javascripts/projects/common/modules/article/spacefinder.js
```diff
@@ -147,12 +155,14 @@ define([
         if (rules.selectors) {
             forOwn(rules.selectors, function (params, selector) {
                 filtered = filtered.then(function (slots) {
+                    return [slots, qwery(rules.bodySelector + selector)];
+                }).then(function (args) {
```
uplne: I don't know how good you already understand the code but I feel like this file should probably be the most well commented/documented piece of functionality. I can imagine that the next person in 6 months looking at this will have hard times again to understand what and especially why exactly this piece of code is doing. It might not be the task for you for this PR but we should finally find time to document it properly. Unless you feel it is documented well enough already.

uplne:
Are you confident enough that this thing is tested properly?

regiskuckaertz:
Yes, it's already on prod, these are just enhancements. There's probably going to be some fine-tuning of the offsets, but it's definitely good to go.

static/src/javascripts/projects/common/modules/article/spacefinder.js
regiskuckaertz: TBH, I didn't change that much of the logic in the spacefinder, but indeed I find it quite difficult to follow. Something to think about indeed!

uplne:
:+1:

PR #11853, opened 2016-02-04
Push back credentials switch again

janua:
Indeed again.
I have sent out an email regarding user accounts, and the deadline for this is Wednesday the 10th.

There is one last set of users left; loose random users. I will email them all individually and give them a deadline of the 12th.

@rich-nguyen


rich-nguyen:
:+1:

PR #11740, opened 2016-01-28
liveblogs share block permalinks

johnduffell:
at present sharing blocks uses the short urls with a hash on them.  Apparently short urls are primrily for the paper, and were used for sharing before twitter could handle long urls.  We use short urls for all sharing at the moment.

Sharing the full urls saves a redirect so makes things quicker.  Also it means we can use the full permalink in liveblog block sharing.

This PR changes it to use the full url with expanded campaign code.

I haven't tested this with all sharing, especially could do with testing on galleries sharing individual images as that uses the same block level sharing.
It would also need checking with data people to make sure the campaign codes change won't break anything.

I think this change is quite risky for causing issues with on site sharing and needs careful monitoring after go live.  As such it needs someone to take full responsibility for getting it into prod.  If it wants to go live someone should have a look at the issue starting with this PR and talking to all the necessary people outside the team to make sure nothing is missed.  Otherwise it can wait until I'm available and I'll handle it!

@stephanfowler @OliverJAsh take care!

PR #11855, opened 2016-02-04
Only use long share links for liveblog blocks

OliverJAsh:
We are reverting to short URLs for all share links but the ones for liveblog blocks, because we didn't mean to change the others in https://github.com/guardian/frontend/pull/11740.


NataliaLKB:
What does using the short urls for the other share blocks mean? Does it break them?

NataliaLKB:
Other than my question, :+1:

NataliaLKB:
Discussed in person! Makes sense. :ship:

PR #11856, opened 2016-02-04
Fix for snap cards on fronts

rich-nguyen:
Non-inlined snap cards were not executing the enhancement step because of this.


jennysivapalan:
:+1:

PR #11858, opened 2016-02-04
Send the R2 header to get the R2 page

jennysivapalan:
Headers set in Platform repo. cc @JustinPinner


JustinPinner:
:+1: :top: :older_man: :older_woman:

PR #11859, opened 2016-02-04
Using amp.theguardian.com as domain for amphtml link

NataliaLKB:
:zap:

### DIFF
--- a/common/app/common/LinkTo.scala
+++ b/common/app/common/LinkTo.scala
@@ -18,6 +18,7 @@ import scala.collection.JavaConversions._
 trait LinkTo extends Logging {
 
   lazy val host = Configuration.site.host
+  lazy val ampUrl = Configuration.amp.url
 
   private val ProdURL = "^http://www.theguardian.com/.*$".r
   private val GuardianUrl = "^(http://www.theguardian.com)?(/.*)?$".r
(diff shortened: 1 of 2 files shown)


common/app/common/LinkTo.scala
```diff
@@ -49,18 +50,20 @@ trait LinkTo extends Logging {
...
-      ProcessedUrl(urlFor(path, edition, secure = true) + format)
+      ProcessedUrl(urlFor(path, edition, secure = true, isAmp = true) + format)
     case GuardianUrl(_, path) =>
       ProcessedUrl(urlFor(path, edition))
     case otherUrl =>
       ProcessedUrl(otherUrl, true)
   }
 
-  private def urlFor(path: String, edition: Edition, secure: Boolean = false): String = {
+  private def urlFor(path: String, edition: Edition, secure: Boolean = false, isAmp: Boolean = false): String = {
```
jamesgorrie: It feels like have an argument of `urlHost: Option[String] = None` might be more generic and then you could have something like:
```scala
val url = s"${urlHost.getOrElse(host)}/$editionalisedPath"
```

Just a thought, perhaps I'm wrong.

jamesgorrie:
I am assuming in the [config the `amp.url` so it's being set to `""`](https://github.com/guardian/frontend/blob/master/common/app/common/configuration.scala#L185) thus not getting [`https`](https://github.com/guardian/frontend/pull/11859/files#diff-a44dd7de0fcab3c0565272c878abea39R68)

common/app/common/LinkTo.scala
NataliaLKB: It is a good thought. While working on this I have been trying to do so in a way that makes it explicit that a change is for amp. First and foremost, so it is easy to search for and remove. 

As amp continues I am weary of this approach. However, I still think it is best to wait until after launch and the business decides how they want to proceed.

NataliaLKB:
@jamesgorrie thanks for taking a look! I have updated the tests to reflect my change so they should work just fine now

common/app/common/LinkTo.scala
jamesgorrie: >  makes it explicit

Always the best way.

jamesgorrie:
:+1:

PR #11860, opened 2016-02-04
Reformat dfp-api.js

jbreckmckye:
**Summary**

This pull request reformats dfp-api.js so that -
- functions are declared as functions, rather than function variables
- the single-var-declaration style is replaced by the multiple-declaration style
- private methods are not listed under the 'public functions' block

**Why do this?**
- named functions are easier to debug, because we can see their names in the callstack. We should also see these names in any stacktraces logged to Sentry. You can do this with function vars, but it's repetitive - first-class functions have this attribute by default.
- functions-as-functions can be hoisted, so are easier to work with and move around than function variables
- the single var style seems to have fallen out of favour in this project
- the single var style is awkward because adding a new item means swapping a semicolon for a comma (which adds noise to the diff), because items have to be indented arbitrarily, and because anyone who jumps into the code has to scroll to verify whether they are looking at a var, an object field or - once we use ES6 - a let/const value.
- it helps pave the way for further, incremental improvements to this gigantic file

**Reviewing**

It's a bit of a chunky PR, so you might want to view it commit-by-commit, and with whitespace changes disabled.

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/dfp-api.js
+++ b/static/src/javascripts/projects/common/modules/commercial/dfp-api.js
@@ -795,7 +795,7 @@ define([
     /**
      * Module
      */
-    var dfp = {
+    return {
         init:           init,
         addSlot:        addSlot,
         refreshSlot:    refreshSlot,
@@ -816,5 +816,4 @@ define([
         }
     };
 
-    return dfp;
 });
--- a/static/src/javascripts/projects/common/modules/commercial/dfp-api.js
+++ b/static/src/javascripts/projects/common/modules/commercial/dfp-api.js
@@ -176,7 +176,6 @@ define([
         }
     };
     var renderStartTime = null;
-
     var recordFirstAdRendered = once(function () {
         beacon.beaconCounts('ad-render');
     });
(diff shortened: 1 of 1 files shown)


static/src/javascripts/projects/common/modules/commercial/dfp-api.js
```diff
@@ -132,667 +132,696 @@ define([
...
-            },
-            '1,1': function (event, $adSlot) {
-                if (!event.slot.getOutOfPage()) {
-                    $adSlot.addClass('u-h');
-                    var $parent = $adSlot.parent();
-                    // if in a slice, add the 'no mpu' class
-                    if ($parent.hasClass('js-fc-slice-mpu-candidate')) {
-                        $parent.addClass('fc-slice__item--no-mpu');
-                    }
+    var resizeTimeout        = 2000;
```
jamesgorrie: So happy to know this is the style.

jamesgorrie:
Looks good. :+1:

regiskuckaertz:
Marvelous! :+1: 

You can name function expressions, i.e. `var x = function y(...) { ... }` and they will appear in stack traces (you could name them in the `callbacks` object). However, you can use `y` only inside the function itself.

The only significant difference is function expressions inherit the current scope, thus it is more [a matter of design](https://github.com/guardian/frontend/blob/841229002f97594ad3aced0fcd5effcbebf85bdf/static/src/javascripts/projects/common/modules/commercial/third-party-tags/outbrain.js#L87) than one of stylistic preference.

PR #11861, opened 2016-02-05
track clicks on video caption

jamesgorrie:
There is talk about adding a "to video page" link on the video player.

There is also talk about having a letter-box video - it'd be good to see if people are using this as it could suggest they want big fancy videos.

### DIFF
--- a/article/app/views/fragments/mainMedia.scala.html
+++ b/article/app/views/fragments/mainMedia.scala.html
@@ -30,7 +30,7 @@
                     }
 
                     @mainVideo.videos.caption.map { caption =>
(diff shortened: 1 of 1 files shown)


jamesgorrie:
I couldn't find a convention for link naming, so just went with the most understandable.

paperboyo:
Ad. 1 - that would be lovely as having to make the whole caption a link is a lot of work + properly ancient + requires manual work so it’s inconsistent. The best would be for our logo not to be burned-in but clickable SVG that would take a reader to the video page **at the time** that the video is playing currently.

Ad. 2 - I’m personally against it. It’s just an absurd half-measure. We do have fullscreen!

rich-nguyen:
:+1:

PR #11863, opened 2016-02-05
AMP header ch-ch-ch-chaaaaanges

zeftilldeath:
Resolved a safari issue, and improved the breakpoints AESTHETICALLY!

### DIFF
--- a/common/app/views/mainAMP.scala.html
+++ b/common/app/views/mainAMP.scala.html
@@ -50,32 +50,33 @@
 
         <div class="main-body">
 
-            <header class="main-header @if(page.metadata.id == "film/2016/feb/02/alan-rickman-david-bowie-tributes-2016-berlin-film-festival") {variant-kicker}">
-                <a href="@LinkTo{/}" class="logo-wrapper" name="top">
-                    @fragments.inlineSvg("guardian-logo-160", "logo")
-                </a>
-
-                @* this is for testing only *@
-                @if(page.metadata.id == "film/2016/feb/02/alan-rickman-david-bowie-tributes-2016-berlin-film-festival") {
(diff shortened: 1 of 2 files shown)


zeftilldeath:
@stephanfowler changed conditonal to display header on any bowie article. Thanks @JustinPinner please take a looksie.

common/app/views/mainAMP.scala.html
```diff
@@ -50,32 +50,33 @@
...
-                            <a href="http://www.theguardian.com/culture">arts</a>
-                        </div>
-                        <div class="main-navigation_link">
-                            <a href="http://www.theguardian.com/lifeandstyle">life</a>
-                        </div>
-                    </div>
-                }
-            </header>
+            @* this is for testing only *@
+            @defining("""bowie""".r) { hasHeader =>
```
JustinPinner: Just a thought, but if it's possible that we'll ever see a capital B, you might need a regex like this: [B|b]owie. If not - seems reasonable. But WHY are we doing this?!

NataliaLKB: Echoing @JustinPinner 's point about the capital B!

__For the Why:__ 
this is for guerrilla testing in the amp demo which only crawls a small section of our articles. We have found that Bowie related articles are one of the things that exists. Therefore this can be tested with all bowie articles to help mimic an "organic" search.

Do you have a preferred search term you would rather us use? :wink:

stephanfowler: Capitals aren't relevant, as this is matching against our articles urls, which are alway lowercase.

I think "bowie" is a good search term. Also quite a good songwriter.

PR #11866, opened 2016-02-05
Remove duplicate log configurations

rich-nguyen:
I accidentally named the configuration of common `logback-test`, which means it conflicts with the logback config for the app under test. Renaming it to `logback` means the config will only be selected for the common tests.


JustinPinner:
:+1: :scroll: :back:

PR #11867, opened 2016-02-05
posh console message

sndrs:
played with the console hiring message on the train last night, came up with this:

screen shot 2016-02-05 at 12 47 45

it's a bit more _brandish_ and will only be shown in the most modern browsers. 

might take a poll of :+1:s against :-1: and merge on that basis...


crifmulholland:
:+1:

jamesgorrie:
:+1: I was going to suggest it.

rich-nguyen:
:+1: can we also remove `THIS IS ENHANCED`?

sndrs:
> can we also remove THIS IS ENHANCED?

it becomes quite tricky to work out if your enhanced or not without it. i guess it could only show in dev?

paperboyo:
:+1: Now, how you load custom fonts into the Console? ;-)

Also, I would add `font-family: Helvetica Neue, Calibri, sans-serif;`. Calibri exists on all modern Windows machines and it’s way nicer than Arial.

rich-nguyen:
@sndrs that sounds dev only, yep. it would be nice if our production code didn't have to do this kind of logic, but our dev system made it easy to see this kind of thing.

sndrs:
posh it is

PR #11868, opened 2016-02-05
Merch components logo color fix

uplne:
Before:
screen shot 2016-02-05 at 12 58 58

After:
screen shot 2016-02-05 at 12 56 49


regiskuckaertz:
:+1:

PR #11869, opened 2016-02-05
Fix extra padding in top ad slot when the page has a skin

OliverJAsh:
# Before

When the page has a skin, such as it does now on http://www.theguardian.com/uk/culture, there is a bug whereby the top ad has too much padding:

image
# After

image
# Why

This is because the `margin-top` no longer needs to be unset since we changed the way padding works in https://github.com/guardian/frontend/pull/11584/files#diff-080ab32aa8365abdced6a7d722b4d6a5R194.

/cc @uplne @Calanthe @jbreckmckye

### DIFF
--- a/static/src/stylesheets/module/_adslot.scss
+++ b/static/src/stylesheets/module/_adslot.scss
@@ -221,7 +221,8 @@
             text-align: center;
 
             .ad-slot__label {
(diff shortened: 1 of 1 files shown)


rich-nguyen:
:+1:

PR #11870, opened 2016-02-05
Any facia item with an image may be snappable

rich-nguyen:
This allows containers which have items that were considered listed (part of a list) to perform their snap upgrade. Space may be limited, but the agreement is to let the snap embed choose it's own height; the container grows as needed.
### before

screen shot 2016-02-05 at 13 43 33
### after

screen shot 2016-02-05 at 13 43 49

### DIFF
--- a/common/app/views/fragments/items/facia_cards/contentCard.scala.html
+++ b/common/app/views/fragments/items/facia_cards/contentCard.scala.html
@@ -10,7 +10,7 @@
 
 @import Function.const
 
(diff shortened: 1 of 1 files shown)


OliverJAsh:
:+1: Makes sense!

sndrs:
won't `!isList` still be needed?

rich-nguyen:
we think no, but I did contemplate keeping it. Straightforwards to add something like `!isList || hasImage`

PR #11871, opened 2016-02-05
Sticky ad: use throttled scroll

OliverJAsh:
Previously we were queueing the animation frame every time the scroll event happened. `throttledScroll` optimises the scroll event so that we only queue an animation frame once the previous animation frame has finished, thus we remove redundant JS execution.

There is no perceived difference in performance.


desbo:
:+1:

sndrs:
:+1:

PR #11872, opened 2016-02-05
Images from static-secure.guim.co.uk are now mapped to correct image host

TBonnin:
cc @sndrs


sndrs:
brilliant :+1:

rich-nguyen:
nice! :+1:

PR #11874, opened 2016-02-05
AMP analytics everywhere

philwills:
Also moved to pull the amp.json from a server where it can insert cookie values into the response.

cc @NataliaLKB


NataliaLKB:
:+1: :sparkles:

PR #11876, opened 2016-02-05
Sticky ad: support expandables

OliverJAsh:
Expandable ads leak outside of their containers upon user interaction, e.g. mouse over or click.
# Before

Previously these ads did not work with the sticky ad because we had set `overflow: hidden` on the sticky ad container.

The reason we were using `overflow: hidden` is because, when the inner ad element increases in height, we transition the height of the outer element, but we don't want the inner ad element to overflow during this transition.
# After

image
## 

This fixes it by only applying `overflow: hidden` whilst we are transitioning. In order to achieve this I've had to introduce a pattern similar to Redux for managing the state (whereas before we were just read the same state for every render), although I think this is a good pattern to use regardless of this change.

JSBin: http://jsbin.com/yogana/5/edit?js,output

/cc @stephanfowler


static/src/javascripts/projects/common/modules/navigation/newSticky.js
```diff
@@ -2,179 +2,234 @@ define([
...
+
+    var topAdRenderedPromise = new Promise(function (resolve) {
+        mediator.on('modules:commercial:dfp:rendered', function (event) {
+            var dfpAdSlotId = 'dfp-ad--top-above-nav';
+            var isEventForTopAdBanner = event.slot.getSlotElementId() === dfpAdSlotId;
+            if (isEventForTopAdBanner) { resolve(); }
+        });
+    });
+
+    var newRubiconAdHeightPromise = new Promise(function (resolve) {
```
jamesgorrie: Does anything rely on these promises failing e.g. tracking?

jamesgorrie: Ah, just looking further down - you just have them in a race :racehorse:

OliverJAsh: As in if the promises end up rejected? If they are rejected it's most definitely an error on our side

jamesgorrie: Yea, but we're not catching it anywhere and letting us know there was an error. e.g. if `'modules:commercial:dfp:rendered'` is never fired, this promise will just hang. But that's maybe being superstitious about failing.

OliverJAsh: These promises represents values that may never happen, which is fine. If we're not loading a Rubicon ad, `newRubiconAdHeightPromise` will never fire.

`topAdRenderedPromise` should always fire, as long as there is a top ad, but I guess the only way to reject this promise would be by setting a timeout, i.e. give up after 10 seconds.

OliverJAsh:
Amongst my refactoring to use a Redux like pattern, the important change described above is best seen at https://github.com/guardian/frontend/commit/a96ba3f2f88cc2bad4c3ce115c57cb5438b4497d.

static/src/javascripts/projects/common/modules/navigation/newSticky.js
```diff
@@ -2,179 +2,235 @@ define([
...
+        var headerHeightPromise = fastdom.read(function () {
+            return $header.height();
+        });
+        return Promise.all([getLatestAdHeight(), headerHeightPromise, getScrollCoords()])
+            .then(function (args) {
+                var adHeight = args[0];
+                var headerHeight = args[1];
+                var scrollCoords = args[2];
+                return {
+                    adIsResizing: false,
```
NataliaLKB: Instead of `adIsResizing`, would `inTransition`, or something like that, be a more accurate name? As the ad itself isn't the thing moving, but the nav itself is moving to compensate for the ad height.

OliverJAsh: :+1:. I've also just realised we never unset this flag to `false` when the ad has resized but we didn't transition ([because the user has scrolled](https://github.com/guardian/frontend/pull/11876/files#diff-fc8d0618fdb58d64c439406ddfebd992R143)). I might have to rethink this.

static/src/javascripts/projects/common/utils/create-store.js
```diff
@@ -0,0 +1,27 @@
+define([], function () {
+    // Mini Redux
+    var createStore = function (reducer) {
```
NataliaLKB: Really like this pattern. It is fairly easy to reason about. :star2:

NataliaLKB:
Really great change!!! Some nitpicks and such, but once those are fixed :+1:

NataliaLKB:
:ship:

static/src/javascripts/projects/common/modules/navigation/newSticky.js
```diff
@@ -2,179 +2,233 @@ define([
...
-                    $adBanner = $('.js-top-banner-above-nav'),
-                    $adBannerInner = $('#' + adId, $adBanner),
-                    $header = $('.js-header');
-
-                var getClientAdHeight = function () {
-                    return fastdom.read(function () {
-                        return $adBannerInner[0].clientHeight;
+    var getAdIframe = function () { return $('iframe', $adBanner); };
+
+    var topAdRenderedPromise = new Promise(function (resolve) {
```
regiskuckaertz: I've created [a promisified function](https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/common/modules/commercial/third-party-tags/outbrain.js#L85) that tracks when a slot has loaded, would it make sense I make a module out of it?

OliverJAsh: That looks perfect. <3

So I just need to do:

``` js
var topAdRenderedPromise = trackAd(dfpAdSlotId);
```

It looks like this will return `Promise[Boolean]`. What does the boolean represent?

regiskuckaertz: It means there's a creative that has been loaded in the slot. When false, the slot is empty.

OliverJAsh: Thanks, can you move it outside of `outbrain` too, so I can use it?

static/src/javascripts/projects/common/modules/navigation/newSticky.js
```diff
@@ -2,179 +2,233 @@ define([
...
+                        var padding = parseInt($adBannerInner.css('padding-top'))
+                            + parseInt($adBannerInner.css('padding-bottom'));
+                        var clientHeight = parseInt(data.value.height) + padding;
+                        resolve(clientHeight);
                     });
-                };
+                }
+            }
+        });
+    });
```
OliverJAsh: @regiskuckaertz I wonder if this would be helpful to you. Your `trackAd` will tell you when DFP renders, but Rubicon ads (which are loaded through DFP) can render themselves again asynchronously, which is what this promise accounts for.

static/src/javascripts/projects/common/modules/navigation/newSticky.js
regiskuckaertz: Sure :santa:

static/src/javascripts/projects/common/modules/navigation/newSticky.js
regiskuckaertz: Very good! Questions: for the purpose of reusability, is there anything specific to your use case I should take out of the function? I'm guessing the top banner test and perhaps the height?

OliverJAsh: Perhaps you could have a `trackRubiconAdResize` function which returns `Promise[Dimensions]`?

FYI, here's the part of the Rubicon script which we added to do the messaging: https://gist.github.com/OliverJAsh/28732cf29cd4f9df5b20#file-a-rb-L40-L47

regiskuckaertz: Yep, that makes sense. Wonderful, so I'll do it asap!

For my own curiosity, is there a security reason why you stringified the object?

OliverJAsh: Good point, it appears that is redundant as it's alright serialised by `postMessage` for you. It's a big effort to update the Rubicon script (manual updates in ~40 different places), so maybe save this for when we next need to make a change?

regiskuckaertz: Alright :fist:

PR #11878, opened 2016-02-05
Extract dfp-api.js into its own package

jbreckmckye:
The dfp-api.js file is fairly big - 830 lines, in fact - and isn't all that easy to navigate. It contains some large, sprawling and complex private functions that can only be tested indirectly, and this makes things harder to refactor.

I have done three things in this PR:
- I have created a new `dfp` package for `dfp-api` and all its submodules. 
- Into that package I've pulled in `dfp-ophan-tracking.js`, renamed just 'ophan-tracking.js'. I did think of renaming `dfp-api` the same way, but reasoned the file would be harder to search for. 
- I have extracted the largest private method in dfp-api, `breakoutIFrame`, into its own module. I have, naturally, tested that breakout ads still work.

Over time, we'll be able to use a divide-and-conquer approach with `dfp-api.js` - extracting complicated dependencies and refactoring each in turn.


regiskuckaertz:
Nice one @jbreckmckye :+1:

PR #11879, opened 2016-02-05
Fail the riff raff upload in sbt

rich-nguyen:
Attempt 2. The bash script fails to report the right exit code on ubuntu. 

Added a new dynamic task that runs the tests on all projects, and executes the riff raff upload only if the tests passed. This is a more complicated solution than the bash script, but it is faster, because sbt is only started once instead of twice.


jennysivapalan:
:+1:

rich-nguyen:
Update: the sbt logger plugin was suppressing the test failure, and confusing the build.

PR #11880, opened 2016-02-08
Dedupe JavaScript config

OliverJAsh:
The JS config for our templates is currently duplicated. This PR dedupes it so that all templates share the same config. It now lives as `templates/javaScriptConfig.scala.js`.

Test:
- [x] Video embed https://embed.theguardian.com/embed/video/commentisfree/video/2015/oct/21/in-defence-of-lad-culture-camaraderie-fun-and-friendships-video
- [x] JsonComponent? http://www.theguardian.com/politics/blog/live/2016/feb/09/labour-trident-corbyn-thornberry-burnham-andy-burnham-says-politics-.json
- [x] Fronts as JSON? http://localhost:9000/uk.json


applications/app/views/videoEmbed.scala.html
```diff
@@ -78,7 +78,7 @@
                                 || 'XDomainRequest' in window
                             )
                         ),
-                        config: @fragments.javaScriptConfig(model.SimpleContentPage(video))
+                        config: @Html(templates.js.javaScriptConfig(model.SimpleContentPage(video)).body)
```
jamesgorrie: Why do we use `@Html` here and not `@Javascript`?

OliverJAsh: My understanding: because this file is a `.scala.html` file, and we don't want to escape the characters here, we wrap it in `Html` so that it does not get HTML encoded.

I tried `JavaScript` and here's what happened:

``` html
config: 


    {
        &quot;page&quot;:
```

… etc.

jamesgorrie: Ah okay, because it's coming through the HTML template parsing, you need to turn the encoding off... :+1:

jamesgorrie:
:+1: much tidier.

PR #11881, opened 2016-02-08
Don't show last payment information if you're cancelling friend membership

tomverran:
Currently it says "your last subscription payment was taken on `<sign up date>` which is a bit surprising if you're on a free membership


tomverran:
@AWare

PR #11882, opened 2016-02-08
Link to comments on Amp

NataliaLKB:
Since we can't load comments on amp (at the moment) we thought it bests to add a link to the comments on our site.

cc @stephanfowler

### DIFF
--- a/common/app/views/fragments/amp/stylesheets/buttons.scala.html
+++ b/common/app/views/fragments/amp/stylesheets/buttons.scala.html
@@ -48,6 +48,7 @@
     content: "";
     background-color: #eaeaea;
     border-radius: 1.125rem;
+    -webkit-border-radius: 1.125rem;
     position: absolute;
     top: 0;
     left: 0;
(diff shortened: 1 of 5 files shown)


common/app/views/fragments/amp/stylesheets/buttons.scala.html
```diff
@@ -48,6 +48,7 @@
     content: "";
     background-color: #eaeaea;
     border-radius: 1.125rem;
+    -webkit-border-radius: 1.125rem;
```
stephanfowler: Doesn't autoprefixer do all that -vendor- stuff for us now?

stephanfowler:
:+1:

common/app/views/fragments/amp/stylesheets/buttons.scala.html
NataliaLKB: As far as I can tell we do that while compiling our sass (mentioned in our gulp and grunt files).

If that is the case that doesn't work for AMP.

PR #11883, opened 2016-02-08
Decomission header-bidding A/B test

jbreckmckye:
This PR removes the test introduced in #11601 - the test is about to expire, and we have collected all the data we need.
- Removed A/B test switch
- Removed prebid vendor scripts
- Removed vendor script curl entry
- Removed prebid-service and dependencies
- Removed ab.js entry and experiment module
- Removed prebid code path in dfp-api.js


uplne:
:+1:

Issue #6872, opened 2014-10-30
Missing Bird Photo


Smylers:
This article features a picture of a bird on the live site, but not on Beta:
http://www.theguardian.com/science/grrlscientist/2012/aug/07/3

I've checked Firefox and Chromium on Linux, and IE on Windows.

The missing image is externally hosted, on Flickr, and its `<img>` element seems to be entirely missing from the HTML source on Beta.

Other articles with Flickr images display just fine, such as: http://www.theguardian.com/science/grrlscientist/2012/aug/07/1

paperboyo:
The poor bird is an embed. Very much an issue still, even more so, because of a lack of an R2 fallback. Duplicate of https://github.com/guardian/frontend/issues/5000.

ScottPainterGNM:
Still an issue with this, when previewing in R2 the bird is displayed and missing from the Nextgen site.

gidsg:
This is a known issue that may not be fixed. We have chosen currently not to import arbitrary embeds into CAPI. We have migrated certain popular embed types e.g. Twitter, YouTube only.
CC @guardian/content-platforms

paperboyo:
Thanks, @gidsg! But in https://github.com/guardian/frontend/issues/5000 there are examples of _our own_ graphics that are not showing... Are they to be forgotten about as well?

Regards
Mateusz

gidsg:
@paperboyo it does not matter who originated the embed, it only matters currently if it is one of the whitelisted types which we have migrated, this was based on the usage stats from the R2 database with guidance from editorial. Doing the migration against a whitelist ensures clean content that we can render consistently in the future, which is difficult for arbitrary embed code.

mchv:
@paperboyo yes the point that our own graphics or code does not matter. This is not going to be fixable until existing r2 content is migrated as flexible content. When this content will editable in Composer you will be able to fix it manually.

ScottPainterGNM:
In which case, considering the comments above I think this issue can be closed as frontend can't serve what isn't there essentially.

paperboyo:
Thanks for explanation. In my very humble opinion, it's a bit strange that ``http://static.guim.co.uk/`` is not on a whitelist (at least ``/sys-images/`` should be safe, if not ``/ni/``). But very good to hear what @mchv is saying! Can't wait to be able fix all the thingz :wink: 

Regards
Mateusz

PR #11885, opened 2016-02-08
Better page presser resilience; send full exception to log and cope with iframes

JustinPinner:
The page presser can fail silently while cleaning a page. This was happening frequently for interactive pages where the script object we look for is encoded within in an iframe, and we ended up with a null reference exception. 

We now send any exception to the log for better auditing, and we've enabled the extraction of the script object we're interested in from iframes if any exist.

cc @jennysivapalan


paperboyo:
This may be **totally** unrelated, I the apologise... But will this salvage https://github.com/guardian/frontend/issues/6872 ?

jennysivapalan:
:+1:

JustinPinner:
@paperboyo sorry, no. This isn't fixing those kinds of issues, we're just archiving old content.

PR #11886, opened 2016-02-08
Remove ad-free experience

jbreckmckye:
_Alas, Ad-Free Web - I knew him well, Horatio..._

image

The switch for ad-free-web is going to expire - imminently - and in the four months since we released it, we've neither tested it with users nor made any plans to in future. For that reason, I'm decomissioning AFW until we actually have a test in place.

Hopefully it'll be resurrected at some point - perhaps as part of the new focus on membership.

**What has been removed?**
- the preflight scripts no longer check the AFW cookie, set guardian.config.commercial properties, or put adfree classes on the document
- CSS no longer listens to those adfree classes
- the user-features module no longer reads the cookie or writes it, and it does not count a missing AFW cookie as grounds to refresh its data
- no other scripts query `userFeatures.isAdfree`, and there is no longer an AFW commercial features policy.

**Changes in other projects**
We'll also need to remove the response key in Membership-Attribute-Service


paperboyo:
:saddest_panda_of_them_all:

Calanthe:
:+1: :(

ScottPainterGNM:
dawson

OliverJAsh:
:+1: :-(

PR #11887, opened 2016-02-08
Show age content warning on most popular on rhs of article

TBonnin:
@JustinPinner and @zeftilldeath for the styling

This is how it looks (ignore the **0** days ;) )
picture 1

### DIFF
--- a/onward/app/views/fragments/rightMostPopular.scala.html
+++ b/onward/app/views/fragments/rightMostPopular.scala.html
@@ -5,6 +5,7 @@
 @import implicits.FaciaContentFrontendHelpers._
 @import common.LinkTo
 @import views.support.ImgSrc
+@import views.html.fragments
 
 @popular.map { pop =>
     <div class="js-right-most-popular right-most-popular right-most-popular--image component--rhc hide-on-childrens-books-site" data-component="right-most-popular" data-importance="-1">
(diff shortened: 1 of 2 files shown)


onward/app/views/fragments/rightMostPopular.scala.html
```diff
@@ -18,7 +19,14 @@ <h3 class="content__meta-heading">Most popular</h3>
...
                         }
-                        <h4 class="right-most-popular-item__headline media__body">@trail.properties.linkText</h4>
+                        <div class="media__body">
+                            <h4 class="right-most-popular-item__headline">@trail.properties.linkText</h4>
+                            @trail.properties.maybeContent.map { content =>
+                                @if(content.tags.tags.exists(_.id == "tone/news")) {
+                                    @fragments.contentAgeNotice(ContentOldAgeDescriber(content))
+                                }
+                            }
+                        </div>
```
TBonnin: `rightMostPopular.scala.html` template doesn't seem to be in use anymore. However I still included this change in it just in case.

JustinPinner:
Assuming `<h4 class="right-most-popular-item__headline"` becoming a child of `<div class="media__body">` doesn't cause any weird problems (@zeftilldeath?) :+1:

zeftilldeath:
@TBonnin @JustinPinner Looks like it makes sense to me! :tiger:

PR #11889, opened 2016-02-08
Capi paid card styling fixes

Calanthe:
Small styling fixes for paid card cAPI component:
- The 'About' dropdown text is smaller now (everywhere also on the paid merchandising content and sticky band) 
- Article text link is underlined on hover state

screen shot 2016-02-08 at 12 55 57

### DIFF
--- a/static/src/stylesheets/module/commercial/paidfor/_paidfor-content.scss
+++ b/static/src/stylesheets/module/commercial/paidfor/_paidfor-content.scss
@@ -57,15 +57,32 @@
     }
 }
 
-.paidfor-box {
+.paidfor-box--header,
+.paidfor-box--text,
+.paidfor-box--logo {
     color: inherit;
     display: block;
-    padding: $gs-baseline / 3 $gs-gutter / 4 $gs-baseline;
+    padding: 0 $gs-gutter / 4;
 
     &:hover,
     &:focus {
         text-decoration: none;
     }
+}
+
+.paidfor-box--header {
+    padding-top: $gs-baseline / 3;
+}
+
+.paidfor-box--text {
+    &:hover,
+    &:focus {
+        text-decoration: underline;
+    }
+}
+
+.paidfor-box--logo {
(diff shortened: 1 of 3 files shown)


static/src/stylesheets/module/commercial/paidfor/_paidfor-content.scss
```diff
@@ -57,10 +57,12 @@
     }
 }
 
-.paidfor-box {
+.paidfor-box--header,
```
jamesgorrie: Is it not worth having this as `paidfor-box` and then adding the modifiers?
As far as I remember the BEM style would be to have `class="paidfor-box paidfor-box--header"` in the HTML?

uplne: In theory the whole thing is not using BEM. The correct way would be: .paidfor (main object) and then .paidfor__box (element of the object) and .paidfor__box--text (modification of the element).

uplne: And in totally ideal world we shouldn't even create new popup styles as we already have popup 
screen shot 2016-02-08 at 14 08 45

We should reuse that, add modifications (bg color) and some elements.

jamesgorrie: Having `.paidfor__box.paidfor__box--text` and `.paidfor-box.paidfor-box--text` are almost the same think I think. `.paidfor-box` could definitely be it's own block element.

I think my concern here is that having a modifier only (which is what these DOM elements have) moves away from BEM completely (a modifier should be used to modify and existing block).

In terms of the dropdown - that makes sense too... Is it abstracted enough to be reused?

Calanthe: Yeah, I wasn't sure about my approach. Thanks for suggestions, I will fix it.

uplne:
:+1:

PR #11891, opened 2016-02-08
Identity signout needn't scrub a deprecated cookie

jbreckmckye:
Following #11886, this pull request stops Identity scrubbing the deprecated 'gu_adfree_user' cookie on logout. If a user still has the cookie, it will be deleted on sight by the common.js cookie cleanup task.


uplne:
:+1:

PR #11893, opened 2016-02-09
Launch Outbrain widget in place of the low-pri merchandising component

regiskuckaertz:
This is to launch the Outbrain widget in place of the low-priority merchandising component, when the hi-pri loads but not the lo-pri one. I've removed the switch and also the second DIV, which was unnecessary.

picture 2

[See it working](http://localhost:9000/technology/2016/jan/06/oculus-rift-kickstarter-free-headsets-virtual-reality)


uplne:
:+1:

PR #11894, opened 2016-02-09
Editing route to contain edition 

NataliaLKB:
Will go with a corresponding fastly change.

Like how we do it with `/`, we need to separate the cache into editions for this route.


rich-nguyen:
:+1:

PR #11896, opened 2016-02-09
Fix double breaking-news banner

TBonnin:
**The problem:**
picture 3

**The explanation:**
In order to make sure the breaking-news banner doesn't sit on top of
some footer content at the bottom of the page, we are adding height to the
page, by adding a copy of the news alert inline at the end of the content
which means the scroll height of the page increases by exactly the height of the alert.
In case the breaking news is loaded before the content of the page/footer (likely to happen on interactives since loading might be slow and interactives have no footer), this banner copy would be visible.

**The fix:**
This commit sets `visibility:hidden` to this banner copy (ie: content of the
banner is hidden but the element still takes up space)

cc @sndrs


sndrs:
:+1:

PR #11900, opened 2016-02-09
Sticky ad: only transition when the ad has actually increased in height

OliverJAsh:
The logic in https://github.com/guardian/frontend/pull/11876 was wrong.

/cc @NataliaLKB


michaelwmcnamara:
QA +1

jamesgorrie:
:+1:

PR #11901, opened 2016-02-09
Fix minute logo positioning

gtrufitt:
Prevent the svg logo container from heading over its width and the svg centering within its container.

_Before_
image

_After_
image


superfrank:
:+1:

PR #11903, opened 2016-02-09
Identity telephone number

markjamesbutler:
Add phone number to edit profile tab to allow updates/deletes.  

[Trello](https://trello.com/c/K45QDrNe/320-touchpoint-add-optional-telephone-field)

screen_shot_2016-02-08_at_17 09 47


identity/app/views/fragments/profile/accountDetailsForm.scala.html
```diff
@@ -3,7 +3,7 @@
 @import views.html.fragments.form.{inputField, textareaField}
 @import form.IdFormHelpers._
 @import form.Input
-@import model.Countries
+@import model.{Countries, CountryCodes}
 @import org.joda.time.DateTime.now
 @import helper._
```
susiecoleman: Currently the 2 submit buttons in one form mean that it is not possible to use the enter button to submit updated details as the browser will just use the first submit button in the form and delete the phone number instead.

It is possible to get round this by adding a hidden button at the top of the form:
<button type="submit" class="is-hidden"></button>

susiecoleman:
Looks good just a minor UX fix. Reviewed with @nlindblad :+1:

identity/app/views/fragments/profile/accountDetailsForm.scala.html
jamespamplin: @susiecoleman is right, but there is a way to do it without adding a 3rd submit button, and without JS. Requires fiddling more with CSS though.

- Move the "Save Changes" button to the top of the form, then make the `<form>` element have CSS `position: relative`, and then absolutely position the button at the bottom of the form with:
```css
button {
  position: absolute;
  bottom: 0;
}
```
- Move the "Delete Phone Number" button to the bottom of the form, then set its position manually within the form (not very scalable as would need to be a constant).

Simplest solution using JS is to make the "Delete Phone Number" button a normal button (not a submit), then add the `type="submit" on click: (should work, but untested)
```html
<button name="thing" onclick="this.type='submit'">Delete Phone Number</button>
```

PR #11904, opened 2016-02-09
Bump dependency update plugin

kelvin-chappell:


rich-nguyen:
:+1: what's the benefit?

kelvin-chappell:
No particular general benefit.  It just gets rid of a warning in my dev environment because my global plugin is at v0.1.10.

rich-nguyen:
ok, nice tidy up.

PR #11905, opened 2016-02-09
Up the chance of the coin banner to 50% in the UK

tomverran:
So instead of the coin banner being in with the contributor banners it is now split out and a variation of it (only one at the moment) will be put up against a contributor banner to choose which banner to finally display.


dominickendrick:
:+1:

PR #11906, opened 2016-02-09
Email sign up: Make AB test functionality perm

gtrufitt:
We've run the test long enough and extended the switch a few times so have moved the functionality for the Guardian Today out of the AB test. This only replicates the bottom-of-article version but we can review the AB test results to decide whether mid-article is going to make sense going forward.


desbo:
:+1:

PR #11907, opened 2016-02-09
amp pages are requested via amp.guardian.com or ?amp=1 query parameters

TBonnin:
cc @NataliaLKB

### DIFF
--- a/common/app/implicits/Requests.scala
+++ b/common/app/implicits/Requests.scala
@@ -2,10 +2,10 @@ package implicits
 
 import common.Edition
 import play.api.mvc.RequestHeader
+import conf.Configuration
 
 trait Requests {
 
-  val AMP_SUFFIX = "/amp"
   val EMAIL_SUFFIX = "/email"
 
   implicit class RichRequestHeader(r: RequestHeader) {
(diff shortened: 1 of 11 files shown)


NataliaLKB:
:star2:

NataliaLKB:
Asked @OliverJAsh to review

dev-build/conf/routes
```diff
@@ -395,7 +395,6 @@ GET            /$path<[^/]+(/[^/]+)?>
 
 GET            /$path<[\w\d-]*(/[\w\d-]*)?/gallery/.*>/lightbox.json                                                             controllers.GalleryController.lightboxJson(path)
 GET            /$path<[\w\d-]*(/[\w\d-]*)?/gallery/.*>.json                                                                      controllers.GalleryController.renderJson(path)
-GET            /$path<[\w\d-]*(/[\w\d-]*)?/gallery/.*>/amp                                                                       controllers.GalleryController.render(path)
```
OliverJAsh: Does this route also need to be removed in the non-dev-build `routes`?

common/app/implicits/Requests.scala
```diff
@@ -22,14 +22,13 @@ trait Requests {
 
     lazy val isRss: Boolean = r.path.endsWith("/rss")
 
-    lazy val isAmp: Boolean = r.path.endsWith(AMP_SUFFIX)
+    lazy val isAmp: Boolean = r.getQueryString("amp").isDefined || r.host == Configuration.amp.host
```
OliverJAsh: `isAmp` is true if the amp query param is defined and equal to `1`? Not sure it matters though…

OliverJAsh:
This looks OK to me, but it's changing some parts of the system which I'm not completely familiar with. Had a good look over it though :-) :+1: 

Build is failing…

NataliaLKB:
@OliverJAsh Thanks for taking a look!

We should have fixed the build now.

dev-build/conf/routes
NataliaLKB: This is seems to be the only place this is referenced. It was probably left over from a few months ago when we thought about galleries on amp. That won't happen now for a while.

PR #11908, opened 2016-02-09
Expose server side test definitions to the client

OliverJAsh:
`window.guardian.config.tests` is currently a map of participations: `Map[TestId, Boolean]`.

I want to add support for server-side tests to the [A/B Tests Chrome extension](https://chrome.google.com/webstore/detail/ab-tests/nehbenedinjacnhlkjbdneedibcagmno). By exposing information about these tests to the client, I will be able to expose it in the extension, like so:

image

This proposes changing `window.guardian.config.tests` to an interface with two keys: `participations` (same as before) and `definitions`. For example:

image

### DIFF
--- a/common/app/mvt/MultiVariateTesting.scala
+++ b/common/app/mvt/MultiVariateTesting.scala
@@ -7,6 +7,7 @@ import org.joda.time.LocalDate
 import play.api.mvc.RequestHeader
 import views.support.CamelCase
 import conf.switches.Switches.ServerSideTests
+import play.api.libs.json._
 
 // To add a test, do the following:
 // 1. Create an object that extends TestDefinition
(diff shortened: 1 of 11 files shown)


sndrs:
why separate them out? at the point of an interface, why do i mind how they're implemented?

if you do want them separate, maybe `serverSide` and `clientSide`, or `A/B` and `MVT` are more familiar  than `participations` and `definitions`?

OliverJAsh:
`config.tests` only contains server-side tests to my knowledge, and this is no different. If that's what you mean?

sndrs:
oh sorry i completely misunderstood what this is. i assumed `tests` was ab tests, and this was adding mvt/serverside to that list. feels like we should lock down naming a bit!

common/app/mvt/MultiVariateTesting.scala
```diff
@@ -30,13 +31,27 @@ object ActiveTests extends Tests {
...
+      "definitions" -> Json.toJson(tests.map(test => Map(
+        "id" -> CamelCase.fromHyphenated(test.name),
+        "description" -> test.description
+      ))),
+      "participations" -> JsObject(participations)
+    )).toString()
   }
 }
 
+sealed case class Variant(name: String)
```
desbo: purely out of interest - why did you make this `sealed`?

desbo: never mind, I've just seen the extensions further down.

rich-nguyen:
Does this grow the user's payload, for no additional benefit to the user?

Why not add server side test information to the s3 json file which currently stores only client-side tests?

https://github.com/guardian/frontend/blob/master/tools/ab-test-info/ab-test-info.js

OliverJAsh:
It does grow the payload :unamused: although only by a tiny fraction. I think your solution would be cleaner, let's chat.

OliverJAsh:
Closing as I'm going to try uploading this information to S3 instead, as it would also be useful for the audience team to have

PR #11909, opened 2016-02-09
Abstract add tracking pixel in a module

regiskuckaertz:
Let's get rid of these unsightly repetitions.

cc @jbreckmckye @uplne


static/src/stylesheets/module/commercial/_creatives.scss
```diff
@@ -1,5 +1,7 @@
 .creative__tracking-pixel {
     position: absolute;
+    width: 1px;
```
jbreckmckye: I wonder if the styles should just be inlined onto the element itself.

static/src/javascripts/projects/common/modules/commercial/creatives/add-tracking-pixel.js
```diff
@@ -0,0 +1,17 @@
+define([
+    'common/utils/template',
+    'text!common/views/commercial/tracking-pixel.html'
+], function (
+    template,
+    trackingPixelStr
+) {
+    var trackingPixelTpl;
```
jbreckmckye: I'm not a big fan of abbreviated variable names.

If you want to make this more concise, I would just drop the 'trackingPixel' prefix (which can be inferred from the context) and merely call this a `template`.

static/src/javascripts/projects/common/modules/commercial/creatives/add-tracking-pixel.js
```diff
@@ -0,0 +1,17 @@
...
+    function addTrackingPixel($adSlot, url) {
+        if (!trackingPixelTpl) {
```
jbreckmckye: It might be shorter - and more idiomatic - to write

    trackingPixelTpl = trackingPixelTpl || template(trackingPixelStr);

static/src/javascripts/projects/common/modules/commercial/creatives/add-tracking-pixel.js
uplne: He already have 'template' there so how is he going to solve this one. Tricky.

regiskuckaertz: this is temporary anyway, I have my precompiled templates nearly ready for take out :smile:

jamesgorrie: Might be a silly question due to lack of context, but why not just `trackingPixelTpl = template(trackingPixelStr);`?

static/src/javascripts/projects/common/modules/commercial/creatives/add-tracking-pixel.js
regiskuckaertz: I keep coming back to it, although I also used `trackingPixelTpl || (trackingPixelTpl = template(trackingPixelStr));` which I find more efficient. But thanks, doing it!

static/src/stylesheets/module/commercial/_creatives.scss
regiskuckaertz: There may be many tracking pixels in the page. What would be the advantage?

jbreckmckye: My element and the styles that concern it would be together.

static/src/javascripts/projects/common/modules/commercial/creatives/add-tracking-pixel.js
regiskuckaertz: You're right, we know for sure this is getting called many times.

static/src/stylesheets/module/commercial/_creatives.scss
regiskuckaertz: I think the existing solution is more efficient. As a DOM element is added, there is zero computation required to update the CSSOM and make it to the render tree. Whereas with a style attribute the HTML tokenizer and CSSOM builder need to work more, especially as items are added after the `DOMContentLoaded` event.

jbreckmckye: I think we are sacrificing readability for a micro-optimization.

Are there figures comparing the two approaches?

uplne: I agree with @jbreckmckye

regiskuckaertz: I don't. The class was there before my PR. I just removed two presentation attributes.

jbreckmckye: It's not important. Let's forget about it.

uplne: Yes, there is not really anything wrong or good. It is more subjective so if you don't agree then keep it as it is. However because it was there before doesn't mean it was right. It's ok to question things.

jamesgorrie: I would say that, as an outsider, I would find either as easy to fathom.
Performance - probably negligible, unless someone wants to test it?
Ambivalence = doesn't matter?

regiskuckaertz: I'm happy to do it—yes, I agree we should question all the choices ever made on code that is not "ours", and critique is second to none when it comes to improving things—but I find no compelling case for it. I may be missing the point. Why don't we inline styles for all similar small components then?

jbreckmckye: The only reason was that if there's a one-to-one mapping between a piece of markup and its styling, and that styling the markup's raison d'etre, then separating the two across two different files is just diffusing your implementation.

Follow your discretion.

regiskuckaertz: It makes sense, it's just that I have a difficult time reconciling this with the SoC between HTML and CSS. Obviously I have a huge mental block here :blush:

I imagine myself coming across a snippet with inline styles, my first action would be to move it to a CSS file; we're so used to switch back and forth between the two when developing/debugging anyway, it's literally why CSS was invented.

uplne: Push your boundaries dude.

uplne:
I am going to do a bold move here and give him :+1:

static/src/stylesheets/module/commercial/_creatives.scss
jbreckmckye: > It makes sense, it's just that I have a difficult time reconciling this with the SoC between HTML and CSS.

Separation of concerns does not mean separation of technologies. If an element 'needs' styling to do its job, all extracting CSS does is split a single concern between two pieces of code that have tight coupling but a loose contract. If the style hook breaks - because a class gets renamed, or something with high specificity tramples your element - there is no way of knowing this before runtime.

> we're so used to switch back and forth between the two when developing/debugging anyway, it's literally why CSS was invented.

But not with the benefit of a crystal ball, and we can see many trends against this approach at work already. Consider how React templates pair markup and the scripts that manipulate them. Or how Vue.js 'vuefiles' embed style snippets alongside the templates they're paired to. In both cases, we marry markup with the things that give that markup its behaviour.

jbreckmckye:
Looks good to me, too :+1:

static/src/stylesheets/module/commercial/_creatives.scss
regiskuckaertz: I think with this argument you just shot yourself in the foot. Let's pursue this discussion over a beer. What say we and @uplne and whoever else wants to join go for a drink this week?

uplne: Do you know about Google gwt?

PR #11910, opened 2016-02-09
R2 pressing - ensure https version of file is available

jennysivapalan:
Some content cannot be migrated from R2 to Flex content so we need to press them. Some of these pages are in sections that have moved to https and so will currently not be served out of the archive.

The plan:
- In R2 pressing (Admin), along with a press of the original R2 page we will press a "secure" version of page. This secure version updates all links on the page source to https. This generally looks okay as it's mainly assets which has a secure version too.
- We will have two rows in the DynamoDb for a path one for http and one for https. The http will point to `path` whereas https will point to `path.secure`.
- In Archive if the request is on https fetch the secure version of the or non https the non-secure version. There is no fallback if secure version does not exist so we need to repress every R2 page to ensure this will work in the future.

cc @JustinPinner

### DIFF
--- a/admin/app/jobs/R2PagePressJob.scala
+++ b/admin/app/jobs/R2PagePressJob.scala
@@ -7,7 +7,8 @@ import common._
 import conf.Configuration
 import conf.switches.Switches.R2PagePressServiceSwitch
 import org.jsoup.Jsoup
-import pagepresser.{PollsHtmlCleaner, BasicHtmlCleaner}
+import org.jsoup.nodes.Document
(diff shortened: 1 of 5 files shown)


JustinPinner:
:+1: :lock: :tada:

rich-nguyen:
I've had an idea. Why don't we try having [protocol relative urls](http://www.paulirish.com/2010/the-protocol-relative-url/)?

jennysivapalan:
Thanks @rich-nguyen  -  I will give the PRURL :cat: a go!

jennysivapalan:
Closing as protocol relative urls will work nicely here. Will update the presser with a new PR.

PR #11911, opened 2016-02-10
Changing amp logo to be `/`, and not editionalise

NataliaLKB:
LinkedTo editionalises, which is great, but unfortunately not useful for AMP. The link to the home page our article pages in the amp carousal are being cached with an edition.

To prevent this, lets not use LinkedTo :)


jamesgorrie:
:+1:

PR #11912, opened 2016-02-10
Change links for R2 pressed page to protocol relative url. 

jennysivapalan:
This will help as we migrate to https (http://www.paulirish.com/2010/the-protocol-relative-url/).

cc @rich-nguyen @JustinPinner


rich-nguyen:
:+1: , but maybe retain https rather than //

JustinPinner:
:+1: I think I'm going to have a few merge conflicts :cry:

PR #11913, opened 2016-02-10
Some additional processing is needed when archiving old interactives

JustinPinner:
**tl;dr We can now extract the correct params for the simplified pressed page tracker for interactives.**

I had to create an html cleaner specifically for interactive content due to inconsistencies in the Jsoup parser.

In doing so I changed the object model a bit to make some activities specific to the object from which they are running.

I've tested with a few different types of content too, and everything looks ok so far.

cc @jennysivapalan


admin/app/pagepresser/HtmlCleaner.scala
```diff
@@ -120,56 +52,76 @@ object BasicHtmlCleaner extends HtmlCleaner {
...
+}
 
-  private def addSwfObjectScript(document: Document): Document = {
+object BasicHtmlCleaner extends HtmlCleaner {
 
-    val swfScriptOpt = try {
-      val source = Source.fromInputStream(getClass.getClassLoader.getResourceAsStream("resources/r2/interactiveSwfScript.js"), "UTF-8").getLines().mkString
-      Some(source)
+  override def canClean(document: Document) = {
+    document.getElementsByAttribute("data-poll-url").isEmpty
```
jennysivapalan: You'll need to check this is not an interactive too?

admin/app/pagepresser/HtmlCleaner.scala
```diff
@@ -120,56 +52,76 @@ object BasicHtmlCleaner extends HtmlCleaner {
...
+      parse(omnitureNoScript.getElementsByTag("img").attr("src")).query.paramMap
+    } else {
+      Map.empty
     }
-    swfScriptOpt.foreach { script =>
-      val html = "<script type=\"text/javascript\">" + script + "</script>"
-      document.head().append(html)
+  }
+
+  def basicClean(document: Document): Document = {
```
jennysivapalan: Should this go in the abstract class? All the implementations call it and don't override it?

jennysivapalan: Plus then `InteractiveHtmlCleaner` doesn't rely on `BasicHtmlCleaner`

jennysivapalan:
A couple comments but in general :+1: 
My PR will conflict but the method I added should just go into `basicClean`.

JustinPinner:
@jennysivapalan You were right - this worked better as three distinct cleaners, one for polls, one for interactives and one for anything else. The separation should be quite easy to follow, and although only 6 files are changed here, that's most of the pressing feature! I'm still not happy with the omniture extraction from interactives, but what's here works for the use cases I've found so far. It could probably be prettier though.

admin/app/pagepresser/InteractiveHtmlCleaner.scala
```diff
@@ -0,0 +1,100 @@
...
+
+  override def extractOmnitureParams(document: Document) = {
+    val omnitureNoScript = if(document.getElementById("omnitureNoScript") != null) {
+      document.getElementById("omnitureNoScript")
+    } else {
+      // Dear reviewer, this horrible code is here because I've found examples of interactives that have nested iframes.
+      // Yes, honestly. Nested iframes. Which Jsoup struggles to parse properly.
+      // Anyway, the omnitureNoScript element we're seeking may exist at some nested level and so this is an attempt at
+      // working arounf the shortcomings of Jsoup's parser. It's horrible and I'm sorry. But I need to get on with
+      // pressing these damned things. If you know a better way, please say.
```
jennysivapalan: :pensive:

jennysivapalan: I think three attempts it more than enough. We should just pass the interactives in small groups. Hopefully it is just a few that we can manually fix up.

jennysivapalan:
Looks good :+1: 
I need a merge and a :+1: on https://github.com/guardian/frontend/pull/11937 and then we should be good to re-press a few urls and check all is as expected?

jennysivapalan:
@JustinPinner I tried running this through locally today in the hope of getting it merged and going. Two issues I found.
1. Some urls that we have pressed already have started to be taken down in R2 e.g. competitions. I've got a small change locally so that we can use either Dotcom or R2 (using the R2 headers) via a switch. It's a bit hacky but it would only be needed for a short time. 
2. The `replaceLinks` method isn't working on this branch. It's running through the code fine and making the replacements but when I check the file on S3 it hasn't made the replacements. I haven't got to why yet so let's chat tomorrow?

JustinPinner:
@jennysivapalan What do you think of this link replacement change?

JustinPinner:
Actually, let me fix that merge conflict first...

JustinPinner:
Ok @jennysivapalan - how's this look to you? :eyes:

jennysivapalan:
look good, thanks fir changing it :+1:

PR #11914, opened 2016-02-10
Revert "corrent mimetype for hls"

jamesgorrie:
Reverts guardian/frontend#11822

I was watching iOS - everything was good, but FF android borked.


michaelwmcnamara:
+1 Revert of earlier change. No new changes that build on said change so this should be fine.

PR #11915, opened 2016-02-10
adds support for embeds/links from s3-uploader

akash1810:
S3-Uploader is a replacement for the R2 dependant batch uploader. You upload a file and are given a fastly backed URL to it. This file can then be added to any article. If the file is a gif, jpeg or png it will be embedded else it will be linked.

This PR adds `uploads.guim.co.uk` to `fastlyOriginCdns` so that images can be served through imigix.

Requires https://github.com/guardian/fastly-image-service/pull/5.

### DIFF
--- a/common/conf/env/DEVINFRA.properties
+++ b/common/conf/env/DEVINFRA.properties
@@ -55,10 +55,6 @@ commercial.masterclasses_url=http://theguardian.com/guardian-masterclasses
 commercial.soulmates_url=https://soulmates.theguardian.com
 commercial.travel_url=http://www.guardianholidayoffers.co.uk
 
-images.media.token=not-a-real-token
-images.static.token=not-a-real-token
-images.uploads.token=not-a-real-token
-
 # Headlines AB Test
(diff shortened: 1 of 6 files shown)


jamesgorrie:
Your tests are breaking...

akash1810:
fixed tests

rich-nguyen:
:+1:

common/conf/env/DEVINFRA.properties
```diff
@@ -57,6 +57,7 @@ commercial.travel_url=http://www.guardianholidayoffers.co.uk
 
 images.media.token=not-a-real-token
```
rich-nguyen: Think these are redundant, remove?

jamesgorrie: Aren't they there for the tests, otherwise they break?

akash1810: > Aren't they there for the tests, otherwise they break?

The build agents have a `frontend.properties` file which overrides values in `DEVINFRA.properties`, so these values never actually got used.

PR #9986, opened 2015-07-30
Anchor links on liveblog fix

Calanthe:
This PR fixes anchor links on liveblogs when the viewability changes are turned on.

Before:
screen shot 2015-07-30 at 16 51 44

After:
screen shot 2015-07-30 at 16 54 42

It is related to the https://github.com/guardian/frontend/issues/9963

PR #9867, opened 2015-07-20
Anchor links with sticky situation fixed

Calanthe:
Before:
screen shot 2015-07-20 at 17 26 16

After:
screen shot 2015-07-20 at 17 26 26

This is the fix for https://github.com/guardian/frontend/issues/9776

PR #11916, opened 2016-02-10
Sticky ad: tidy

OliverJAsh:
Comments and tidy up, see commits for breakdown.


static/src/javascripts/projects/common/modules/navigation/newSticky.js
```diff
@@ -107,48 +113,36 @@ define([
...
-            'transition': ''
-        });
-
-        // Set
         // TODO: move into stylesheets when productionised
         els.$document.addClass('new-sticky-ad');
-        els.$header.css({ 'margin-top': state.adHeight });
-        els.$adBanner.css({ 'max-height': state.adHeight });
+
+        var transitionTimingFunction = 'cubic-bezier(0, 0, 0, .985)';
```
jamesgorrie: For my info, `.985`?

static/src/javascripts/projects/common/modules/navigation/newSticky.js
```diff
@@ -107,48 +113,36 @@ define([
...
+        var transitionDuration = '1s';
+
+        els.$header.css({
+            'transition': state.shouldTransition
```
jamesgorrie: Would this not set the transition to an invalid CSS value?
Perhaps just not setting it if `!shouldTransion`.

static/src/javascripts/projects/common/modules/navigation/newSticky.js
```diff
@@ -107,48 +113,36 @@ define([
...
+
+        els.$adBanner.css({
+            'position': userHasScrolledPastHeader ? 'absolute' : 'fixed',
+            'top': userHasScrolledPastHeader ? state.headerHeight : '',
+            'height': state.adHeight,
+            // Stop the ad from overflowing while we transition
+            'overflow': state.shouldTransition ? 'hidden' : '',
+            'transition': state.shouldTransition
+                ? ['height', transitionDuration, transitionTimingFunction].join(' ')
+                : ''
```
jamesgorrie: [See here](https://github.com/guardian/frontend/pull/11916/files#r52463403), and line 137?

static/src/javascripts/projects/common/modules/navigation/newSticky.js
OliverJAsh: We've just copied this from what we had with the old sticky nav: https://github.com/guardian/frontend/blob/master/static/src/stylesheets/module/_adslot.scss#L120

It looks like this: 
image

@stephanfowler and I did think about changing it to ease-in

static/src/javascripts/projects/common/modules/navigation/newSticky.js
OliverJAsh: Setting it to `''` actually unsets the inline style

static/src/javascripts/projects/common/modules/navigation/newSticky.js
OliverJAsh: Setting it to `''` actually unsets the inline style. I used to do what you're saying (`only set it if `shouldTransition`) but this way I can use a single object to set all styles for a given element. Not sure if the `''` to unset is clear though

static/src/javascripts/projects/common/modules/navigation/newSticky.js
```diff
@@ -179,41 +173,39 @@ define([
         });
     };
 
+    var reducer = function (previousState, action) {
+        switch (action.type) {
+            case 'SCROLL':
+                return assign({}, previousState, {
+                    previousAdHeight: previousState.adHeight,
```
jamesgorrie: As you're assigning, this isn't necessary is it?

jamesgorrie: Oh wait, just reread it.

static/src/javascripts/projects/common/modules/navigation/newSticky.js
jamesgorrie: If it clears it, all good :+1:

jamesgorrie:
Looks good form here :+1:

static/src/javascripts/projects/common/modules/navigation/newSticky.js
sndrs: `''` or `null` should clear it

Calanthe:
@OliverJAsh  @Jholder112233 I just realised that after introducing the sticky nav, we had some bugs with anchor links which were partially fixed: here https://github.com/guardian/frontend/pull/9986 and here https://github.com/guardian/frontend/pull/9867. Now when the sticky nav is gone, we should double check and revert those fixes.

OliverJAsh:
Thanks @Calanthe!

PR #11917, opened 2016-02-10
Host Amp 3rd Party iframe on www.theguardian.com

kenlim:
This will allow us to serve the Google AMP 3rd party iframe from within the www.theguardian.com domain.


stephanfowler:
:+1:

PR #11918, opened 2016-02-10
Add image missing metadata (height and width)

TBonnin:
Image metadata height and width are required for amp pages
screen shot 2016-02-10 at 15 34 10

@NataliaLKB

### DIFF
--- a/common/app/views/fragments/img.scala.html
+++ b/common/app/views/fragments/img.scala.html
@@ -41,6 +41,8 @@
 
         @if(amp) {
             <meta itemprop="url" content="@ImgSrc.getFallbackUrl(picture)">
+            <meta itemprop="width" content="@ImgSrc.getFallbackAsset(picture).fold(0)(_.width)">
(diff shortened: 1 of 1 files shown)


jamesgorrie:
:+1:

common/app/views/fragments/img.scala.html
```diff
@@ -41,6 +41,8 @@
 
         @if(amp) {
             <meta itemprop="url" content="@ImgSrc.getFallbackUrl(picture)">
+            <meta itemprop="width" content="@ImgSrc.getFallbackAsset(picture).fold(0)(_.width)">
+            <meta itemprop="height" content="@ImgSrc.getFallbackAsset(picture).fold(0)(_.height)">
```
TBonnin: I don't think image metadata needs to be amp specific but since `url` was already in the `if` I went with the flow ;)

NataliaLKB:
Thanks @TBonnin !!!

PR #11920, opened 2016-02-10
Use replaceLinks at the end of the html cleaner.

jennysivapalan:
It seems ordering is important in the cleaner. Using replaceLinks at the last step.

cc @JustinPinner @rich-nguyen


rich-nguyen:
:+1:

PR #11921, opened 2016-02-10
Improve the mobile styling of The Minute

gtrufitt:
- Full height content blocks
- Enhance tweets on mobile
- Different figure styling on mobile
- Split logos

mobile_style_v1


common/app/model/content.scala
```diff
@@ -197,15 +197,17 @@ final case class Content(
     } getOrElse (None,None)
 
     val articleMeta = if (tags.isUSMinuteSeries) {
-      Some("isMinuteArticle", JsBoolean(tags.isUSMinuteSeries))
-    } else None
+      List(
+        Some("isMinuteArticle", JsBoolean(tags.isUSMinuteSeries)),
+        Some("enhanceMobileTweets", JsBoolean(true))
```
rich-nguyen: ??? why do we need an always true meta?

gtrufitt: It's always true if it is the USMinuteSeries as inside the if statement. Seemed redundant to add tags.isUSMinuteSeries here too but probably doing it wrong :/

rich-nguyen: exactly, so why can't the js just check `isMinuteArticle`? The human logic is "a US minute article should enhance tweets."

gtrufitt: Hmm, yeah, I guess I was pre-empting future use of that meta for other things, but will revert this and use isMinuteArticle instead instead of trying to predict the future...

rich-nguyen:
:+1:

PR #11924, opened 2016-02-10
Paid Card cAPI mostpop fix

Calanthe:
Before:
screen shot 2016-02-10 at 17 42 13

After:
screen shot 2016-02-10 at 17 42 26

cc @uplne 

Also added small css fixes so the Paid Card (and the whole most popular component) is a little smaller.


static/src/javascripts/projects/common/modules/commercial/creatives/commercial-component.js
```diff
@@ -103,13 +105,30 @@ define([
         }
     }
 
+    function paidforCardLoaded(el) {
+        var height,
+            $adSlot = $(el);
```
uplne: We are slowly adopting multiple var approach.

static/src/javascripts/projects/common/modules/commercial/creatives/commercial-component.js
```diff
@@ -103,13 +105,30 @@ define([
...
+
+        if ($adSlot.hasClass('ad-slot--mostpop')) {
```
uplne: Should we go for js-hook here?

static/src/javascripts/projects/common/modules/commercial/creatives/commercial-component.js
```diff
@@ -103,13 +105,30 @@ define([
...
+    }
+
     CommercialComponent.prototype.postLoadEvents = {
         bestbuy: function (el) {
             new Tabs().init(el);
         },
         capi: createToggle,
         capiSingle: createToggle,
-        paidforCard: createToggle
+        paidforCard: paidforCardLoaded
```
uplne: Is the name of this method describing its purpose properly? Should we rather use something like `adjustComponentHeight` or something like that.

static/src/javascripts/projects/common/modules/commercial/creatives/commercial-component.js
Calanthe: Does it mean that I have to use that approach? I am used to one var and I think it's cleaner so I prefer to stay with it.

static/src/javascripts/projects/common/modules/commercial/creatives/commercial-component.js
Calanthe: We don't have `js-ad-slot--mostpop`, we only got `js-ad-slot` but it will target all ad slots. In this example I am interested only in the `mostpop` slot.

static/src/javascripts/projects/common/modules/commercial/creatives/commercial-component.js
Calanthe: You are right. I had problems with naming it. But that function does not only change height. It is more general because it also initialise tabs. I couldn't think of any short name.

static/src/javascripts/projects/common/modules/commercial/creatives/commercial-component.js
jbreckmckye: At the moment we have two styles. Dotcom use the multi-var approach quite extensively, and we've adopted multi-var in files like `dfp-api` too. We should probably choose just one.

I know a few people aren't all that keen on the single-var format and I'd probably count myself amongst them. If nothing else, I'd say that single-var will become an antipattern in ES6, because it becomes less obvious in large declarations whether an entity is a `var`, a `let` or a `const`. I also think it's more awkward to write, and that having to swap semi-colons for commas when adding variables adds white noise to git diffs.

> Does it mean that I have to use that approach?

I don't think it matters that much for short declarations - it's when you have a screen full of vars that I think it gets awkward to work with. We did discuss how hard we should enforce consistencies like this at the last dotcom-hygiene meeting, but we didn't have a clear answer.

static/src/javascripts/projects/common/modules/commercial/creatives/commercial-component.js
```diff
@@ -8,6 +8,7 @@ define([
     'raven',
     'common/utils/$',
     'common/utils/config',
+    'common/utils/fastdom-promise',
```
jbreckmckye: If we are not using the promise, do we still need the `fastdom-promise` dependency?

static/src/javascripts/projects/common/modules/commercial/creatives/commercial-component.js
```diff
@@ -103,13 +105,30 @@ define([
...
+        var height,
+            $adSlot = $(el);
+
+        if ($adSlot.hasClass('ad-slot--mostpop')) {
+            fastdom.read(function () {
+                height = $adSlot.dim().height;
+            });
+
+            fastdom.write(function () {
+                $('.js-most-popular-footer .tabs__pane').css('height', height);
```
jbreckmckye: Are there ever instances where fastdom will run the `write` before the `read` to preserve re-layout calls?

static/src/javascripts/projects/common/modules/commercial/creatives/commercial-component.js
jbreckmckye: What about something like `initMostPopularCard`?

static/src/javascripts/projects/common/modules/commercial/creatives/commercial-component.js
Calanthe: I don't think so. Based on the examples in documentation is seems that `read` will be always before `write`. Now I noticed that I could move `$('.js-most-popular-footer .tabs__pane')` to the `read` operation.

static/src/javascripts/projects/common/modules/commercial/creatives/commercial-component.js
```diff
@@ -103,13 +105,32 @@ define([
...
 
+    function initMostPopularCard(el) {
+        var height;
+        var $adSlot = $(el);
+        var $mostPopTabs;
+
+        if ($adSlot.hasClass('ad-slot--mostpop')) {
+            fastdom.read(function () {
+                height = $adSlot.dim().height;
+                $mostPopTabs = $('.js-most-popular-footer .tabs__pane');
```
regiskuckaertz: You can move this out of the Fastdom batch, querying the DOM does not trigger layout.

static/src/javascripts/projects/common/modules/commercial/creatives/commercial-component.js
```diff
@@ -103,13 +105,32 @@ define([
...
+    }
+
     CommercialComponent.prototype.postLoadEvents = {
         bestbuy: function (el) {
             new Tabs().init(el);
         },
         capi: createToggle,
         capiSingle: createToggle,
-        paidforCard: createToggle
+        paidforCard: initMostPopularCard
```
regiskuckaertz: This tight coupling between `createToggle` and `initMostPopularCard` should stay out of each function's logic. We could compose them, with the only contract that each function must return the element:

`paidforCard: compose(createToggle, initMostPopularCard)`

jbreckmckye: @regiskuckaertz You wouldn't even need to return the elements, necessarily. Perhaps we could do something like this -

    function join() {
        var actions = Array.apply(null, arguments);
        return function joined() {
            forEach(actions, function (action) {
                action.apply(this, arguments);
            })
        };
    }

and then

    paidForCard : join(createToggle, initMostPopularCard)

regiskuckaertz: Even better! This way the logic inside each callback is completely sealed from the rest. We should definitely build a library of higher-order functions like this one, it would help write much more readable code.

regiskuckaertz:
:+1:

PR #11927, opened 2016-02-11
Update top employers URL

sndrs:
From:

`http://www.theguardian.com/careers/britains-top-employers`

To:

`http://www.theguardian.com/careers/top-employers-uk`


rich-nguyen:
:+1:

PR #11929, opened 2016-02-11
Update styling of paid containers

kelvin-chappell:
Thanks to @jbreckmckye _commercial/single-campaign_ containers now:
- Hold more cards
- Have _show more_ functionality

Before:
image

Now:
image


static/src/javascripts/projects/common/modules/commercial/badges.js
```diff
@@ -69,8 +69,13 @@ define([
 
             return new Promise(function (resolve) {
                 idleFastdom.write(function () {
-                    $('.js-container__header', container)
-                        .after($adSlot);
+                    var placeholder = $('.js-badge-placeholder', container);
+
+                    if (placeholder[0]) {
+                        placeholder.replaceWith($adSlot[0]);
```
regiskuckaertz: Bear in mind that this will replace *all* elements matched by the query, should there ever be more than one.

Calanthe:
:+1: It looks so better now!

regiskuckaertz:
:+1: nice work!

static/src/javascripts/projects/common/modules/commercial/badges.js
jbreckmckye: @regiskuckaertz I think that is less weird than only replacing the first candidate.

regiskuckaertz: yeah definitely. I guess I was bugged by the test, it's the first time I see it this way rather than testing the value of `placeholder.length`. On a sidenote: I've never actually tested if bonzo would detect that you pass a `Node` to `replaceWith` and thus needs to clone it, otherwise the same node will just move around and end up in the last selected node.

jbreckmckye: Good point - I will check.

jbreckmckye: Apparently,
    bonzo.replaceWith() will replace each item in the initial collection with the supplied content.

PR #11932, opened 2016-02-11
Use front type to determine rendering of paid containers

kelvin-chappell:
We only want paid containers to have the new styling if they appear on editorial fronts.
This change uses the `ConfigAgent` to determine the type of front the container is on.

/cc @janua

### DIFF
--- a/common/app/views/support/Commercial.scala
+++ b/common/app/views/support/Commercial.scala
@@ -1,8 +1,10 @@
 package views.support
 
+import com.gu.facia.api.models.{EditorialPriority, FrontPriority}
 import common.Edition
 import common.dfp.AdSize.{leaderboardSize, responsiveSize}
 import common.dfp._
+import conf.switches.Switches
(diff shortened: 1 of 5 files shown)


common/app/views/support/Commercial.scala
```diff
@@ -81,6 +83,13 @@ object Commercial {
 
   object container {
 
+    def shouldRenderAsCommercialContainer(frontPriority: Option[FrontPriority], container: FaciaContainer): Boolean = {
+      Switches.NewCommercialContent.isSwitchedOn &&
+      Switches.PaidContainerUpdate.isSwitchedOn &&
+      frontPriority.contains(EditorialPriority) &&
```
kelvin-chappell: Everything is using scala 2.11 so safe to use an `Option.contains`.

JonNorman:
Looks good - can expand to Dynamic containers once happy with Fixed containers.
:+1:

PR #11933, opened 2016-02-11
Make inline component as wide as the thinnest element on mobile

regiskuckaertz:
This PR gives inline components the same width/margins than rich links, so that they take the same space and not more. The only exception is for the inline book component, which keeps a smaller width so that book covers display nicely.

**Before**
picture 2

**After**
picture 1

### DIFF
--- a/static/src/stylesheets/module/_adslot.scss
+++ b/static/src/stylesheets/module/_adslot.scss
@@ -470,13 +470,8 @@
     margin: 5px $gs-gutter $gs-baseline 0;
 
     @include mq($until: mobileLandscape) {
-        width: $gs-gutter * 6.5;
         margin-bottom: $gs-baseline / 2;
         margin-right: $gs-gutter / 2;
-    }
-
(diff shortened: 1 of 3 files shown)


static/src/stylesheets/module/_adslot.scss
```diff
@@ -466,12 +466,12 @@
     }
 }
 .ad-slot--im {
-    box-sizing: border-box;
-    max-width: 50%;
     float: left;
+    margin: 5px $gs-gutter $gs-baseline 0;
```
uplne: My only concern is if it is supposed to use the same styles why we just don't use the same class as rich links are? Do we need to abstract some class to something like `'.inlinebox` and move there these margin styles and reuse it?

regiskuckaertz: It's not a complete 1:1 similarity, as rich links get wider and have a negative left margin on desktop.

uplne:
:+1:

PR #9489, opened 2015-06-16
Viewability AB test

uplne:
This PR is:
- removing recommendation 1 and 2 (MtRec1 and MtRec2) AB tests.
- moving everything from Recommendation 1 to new Viewability test
- creating improved version of sticky header
- AB test with 0 audience for internal testing. We will be slowly increasing the audience starting from next week.
- I am also cleaning some unused CSS 

Most of this PR is just moving already running MtRec1 features to new Viewability test (lazy loaded ads, sticky MPU, dynamically loaded ads on article pages, comments ads, spacefinder 2.0 and the sticky header). The biggest change is the sticky header (/navigation/sticky.js). We need to move features from MtRec1 to Viewability as it all needs to launch together.

There are still some issues we need to address with sticky header in following PR.

screen shot 2015-06-16 at 10 01 23

PR #9084, opened 2015-05-01
Sticky nav UX tests

uplne:
Because sticky nav tests are running only on the US edition and for UX testing we need them to run on UK edition, I've moved tests to its own file and added two extra 0 audition tests just for UX (internal) testing purposes.

This will be reverted after UX test will happen early next week.

CC @kelvin-chappell

PR #11936, opened 2016-02-11
Remove old sticky code

OliverJAsh:
Searched the codebase for all things sticky, and this is what I found. I also consulted https://github.com/guardian/frontend/pull/9489/files, https://github.com/guardian/frontend/pull/9084/files, amongst other PRs.

@uplne @Calanthe Is there anything here that you think I may have missed?

To do:
- [ ] Ensure the banner has a transition when not enhanced


uplne:
I remember that we did some changes to the search popup so it might be worth checking. Otherwise :+1:

PR #11938, opened 2016-02-11
Merch components with optional titles

uplne:
After: 
screen shot 2016-02-11 at 15 15 32

@regiskuckaertz I've thought about it and I can't see any smart abstraction for this which won't be just over-engineering things. There are actually a lot of differences. So unless you have some clever idea I will leave it as I think it makes sense how it is for now.


regiskuckaertz:
Let's focus on getting things done now and consult each other when we've got time and space to look at this from a distance.

:+1: bro

PR #11940, opened 2016-02-11
Make amp header permanent

NataliaLKB:
image

cc @zeftilldeath @stephanfowler


zeftilldeath:
This is even bigger news than Gravitational Waves. :earth_africa: :+1:

PR #11945, opened 2016-02-11
Change Outbrain widget to be consistent with other containers in AMP

JonNorman:
#### Before

image
#### After

image
#### Overflow element

The `overflow` element contains the _More stories_ button that appears when the page is loaded prior to the amp container contents. In such cases the height of the `amp-iframe` is determined by the `height` property specified as an attribute of the `amp-iframe`.

The `height` property has been set so as to show a single card and then the button, however on narrower screens, the Outbrain header text wraps, causes the contents of the frame to become taller and results in the `overflow` covering part of the ad:

**Nexus 7 - 600x960**:
image

**Nexus 6 - 412x659**:
image

An initial height for the `amp-iframe` must be given and I don't know a way of determining the user agent when setting the height property of the `amp-iframe`. The `overflow` element must also be part of the `amp-iframe` element so it cannot be set to float below the Outbrain `iframe`.

@NataliaLKB - if you have some time tomorrow, it'd be great to discuss **a)** whether this matters and (if so) **b)** a good approach to circumventing it.


NataliaLKB:
@JonNorman It already looks much better! Lets discuss tomorrow :)

common/app/views/fragments/amp/outbrain.scala.html
```diff
@@ -4,13 +4,17 @@
...
-        sandbox="allow-top-navigation allow-scripts allow-same-origin allow-popups allow-popups-to-escape-sandbox"
-        layout="responsive"
+        <amp-iframe height=210
+        sandbox="allow-scripts allow-same-origin allow-popups"
+        layout="fixed-height"
         frameborder="0"
         resizable
-        src="https://widgets.outbrain.com/hub/amp.html#widgetIds=AMP_1&htmlURL=@URLEncode(htmlUrl)&ampURL=@URLEncode(s"$htmlUrl/amp")&styleFile=">
-            <div overflow tabindex=0 role=button aria-label="Outbrain Recommendations">Outbrain</div>
+        src="https://widgets.outbrain.com/hub/amp.html#widgetIds=AMP_1&htmlURL=@URLEncode(htmlUrl)&ampURL=@URLEncode(s"$htmlUrl?amp")"
```
regiskuckaertz: Minor detail: ampersands are special characters in HTML, you should use the entity reference: `&amp;`

uplne:
Is this styling right? 
screen shot 2016-02-15 at 10 49 09

JonNorman:
> Is this styling right?

@uplne No - it is something I want to fix, but would like to discuss the options available.

**Background**
The `amp-iframe` element in which the Outbrain `iframe` is loaded must be given an initial `height` (see [AMP Layout spec](https://www.ampproject.org/docs/guides/responsive/control_layout.html)). If the content of the `iframe` does not fit that initial height, then the `overflow` element/button is displayed which  (when clicked) resizes the `amp-iframe` to fit the (now loaded) `iframe`contents.

**Question**
*What value should be given to the `height` attribute of the `amp-iframe`?*

The other elements on the page typically show a single card before the `overflow` element. To emulate this behaviour, the `height` of the `amp-iframe` is set at 210px. When the media size is sufficiently wide (e.g. Nexus 7), the bottom of the first ad appears around 205 px and so this works well. When the media size is narrower (e.g. Nexus 6), the Outbrain header wraps and thus the first ad is pushed lower down than before (the bottom now being around 240px) which then causes the `overflow` element to overlay it.

**Solutions**
1) setting the initial height to always be the full height of the Outbrain `iframe` to begin with (i.e. no `overflow`), which would be inconsistent with the other onward elements on the page;
2. somehow passing details of the screen width to the template and to adjust the height parameter accordingly when creating the `amp-iframe`; or
3. tolerating the overlay in the cases where the user has scrolled to the `iframe` before it has had time to load but being unhappy about it.

Happy to discuss any other options that I've missed.

regiskuckaertz:
@JonNorman correct me if I'm wrong, but AMP allows iframes to be resized. Only it has to occur under certain circumstances, namely the iframe is BTF and a 60 fps refresh rate is kept. If one of these conditions is not met, then clicking on the button should trigger the resize.

regiskuckaertz:
Well, the resize request needs to come from within the iframe by means of `window.postMessage`. Can we ask Outbrain to do that?

JonNorman:
@regiskuckaertz you are correct wrt the resize request that can be sent. The Outbrain `iframe` already does this: 
image

However, the problem lies wherein the user scrolls to the `amp-iframe` prior to the Outbrain `iframe` having loaded. At this point, the resize request is being made when the user is already viewing the `amp-iframe` and as a result, the height is not changed, and the `overflow` is used. It is therefore the *initial* height that is used, which is static.

NataliaLKB:
> What value should be given to the height attribute of the amp-iframe?

What question do you have about this? We have already given the default height [here](https://github.com/guardian/frontend/blob/jn-outbrain-amp-styling/common/app/views/fragments/amp/outbrain.scala.html#L7) and as @regiskuckaertz says, the iframe will resize anyway :smile: . Keep the same behaviour as all the others - Outbrain should style them with set heights anyway. This isn't a problem we can fix on our end.

JonNorman:
>What question do you have about this?

It is the screenshot that @uplne referenced. When the screen is narrow to the point that the Outbrain header wraps, the `overflow` element covers part of the first ad in the Outbrain `iframe` and looks a bit rubbish; as I said above, I think this is an edge case and something we should leave as-is.

That said, I've just scrolled to the comment and I think what he was pointing out was actually the lack of margin **not** the `overflow` positioning(?!). On that point, the Outbrain component is now consistent with the other components on the page - if we want to add a margin, then I suggest we add it to all of the amp components.

Apologies for the massive discussion on this - I think I've drawn out far more than it merits!

NataliaLKB:
> lack of margin not the overflow positioning

Yes I think you are correct and that is what he was mentioning. We discussed a fix for that last week - outbrain does need the margin, and it is no problem to add some :smile: , shall I come chat at some point today?

common/app/views/fragments/amp/stylesheets/buttons.scala.html
```diff
@@ -30,10 +30,18 @@
     margin: 0 20px 0;
 }
 
+.cta--show-more.cta--show-more__unindent {
```
NataliaLKB: I am not sure we want to unindent this. When we add padding/margin around the outbrain container to make it inline with the rest of the components, it shouldn't include the overflow element.

NataliaLKB:
@JonNorman Thanks for doing a bit more work on this.

Just the last comment I had about the unindenting of the overflow element. If I remember correctly I don't think thats what we want to do. I could be wrong though so please just do what you think is best.

It might be worth just waiting until outbrain finishes with their changes before continuing with this PR anyway, but its your choice :smile: 

:+1:

PR #11955, opened 2016-02-12
Fix liveblog updates 

stephanfowler:
When there are no key events, there's no timeline. The case wasn't being caught, and a js error was killing the liveblog module.

@Jholder112233 @OliverJAsh

### DIFF
--- a/static/src/javascripts/bootstraps/enhanced/liveblog.js
+++ b/static/src/javascripts/bootstraps/enhanced/liveblog.js
@@ -60,7 +60,7 @@ define([
     function createScrollTransitions() {
 
         var curBinding,
-            timeline      = qwery('.timeline')[0],
+            timeline      = qwery('.timeline'),
             selectedClass = 'live-blog__key-event--selected';
 
         function unselect() {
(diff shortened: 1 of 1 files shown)


Jholder112233:
:+1: :clap:

Got an expired switch though.

OliverJAsh:
Great detective work!

I've seen a few bugs in JS caused by false assumptions such as this: we assumed `[0]` would always return something, but that's not true.

It's worth considering whether we could use `Option`s in JS (like in Scala) to declare a value as optional and thus force us to deal with the scenario where the value is undefined. Instead of `array[0]` you could do `headOption(array).map(...).getOrElse(...)`. This would mean we would get no exception and we can eradicate this class of bugs.

mchv:
@stephanfowler great to see you figure out what was the issue!

> It's worth considering whether we could use Options in JS (like in Scala) to declare a value as optional and thus force us to deal with the scenario where the value is undefined. Instead of array[0] you could do headOption(array).map(...).getOrElse(...). This would mean we would get no exception and we can eradicate this class of bugs.

@OliverJAsh did you have a look to [monapt](https://github.com/jiaweihli/monapt) and this [article from heap analytics](http://blog.heapanalytics.com/goodbye-coffeescript-hello-typescript/)

OliverJAsh:
@mchv Yup, I'm using Montapt very lightly in our dashboard: https://github.com/guardian/frontend/blob/b65afe214200b1fec78449cd4ac6c3c32f2e587a/static/src/deploys-radiator/app/main.ts#L20. Also enjoyed that article! Have you got any experience of using Options in JS?

OliverJAsh:
Closing in favour of https://github.com/guardian/frontend/pull/11959

PR #11958, opened 2016-02-12
Add a min-width to the engaged reader banner button

tomverran:
before

image

after

image

necessary as the text is the only clickable part of the button

### DIFF
--- a/static/src/stylesheets/module/_release-message.scss
+++ b/static/src/stylesheets/module/_release-message.scss
@@ -150,6 +150,7 @@
     @include mq(tablet) {
         display: table-cell;
         vertical-align: middle;
+        min-width: 85px;
         margin-top: 0;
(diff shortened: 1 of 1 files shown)


static/src/stylesheets/module/_release-message.scss
```diff
@@ -150,6 +150,7 @@
     @include mq(tablet) {
         display: table-cell;
         vertical-align: middle;
+        min-width: 85px;
```
Calanthe: Try to use variables like `$gs-gutter` from the `_grid-system.scss` file.

tomverran:
hi @Calanthe I used $gs-column-width in the end which is a bit wider than it needs to be but it still looks ok

gtrufitt:
:+1:

PR #11959, opened 2016-02-12
Fix liveblog updates bug

OliverJAsh:
This PR is to replace https://github.com/guardian/frontend/pull/11955

> When there are no key events, there's no timeline. The case wasn't being caught, and a js error was killing the liveblog module.

This bug was introduced by https://github.com/guardian/frontend/commit/99ff155c4cd559abbca42823bba9be887826c248

Here's the stacktrace of the bug:

image

/cc @stephanfowler @Jholder112233


Jholder112233:
How does this related to #11955?

static/src/javascripts/projects/common/modules/ui/autoupdate.js
```diff
@@ -176,7 +176,8 @@ define([
 
             this.notificationBar = new NotificationBar({attachTo: $('.js-update-notification')[0] });
 
-            $(options.attachTo).addClass('autoupdate--has-animation');
+            var identity = function (val) { return val; };
+            $(options.attachTo.filter(identity)).addClass('autoupdate--has-animation');
```
rich-nguyen: is `identity` the best way to check the validity of `attachTo`? Seems a bit...javascript-y :)

rich-nguyen:
:+1:

OliverJAsh:
@Jholder112233 #11955 correctly identified the problem, but the fix was incorrect.

Jholder112233:
Cool, thanks for fixing this @OliverJAsh - :ship:

static/src/javascripts/projects/common/modules/ui/autoupdate.js
OliverJAsh: As discussed in person, `options.attachTo` is an array of `node | undefined`. The error occured when you tried calling `addClass` on a bonzo element that wrapped `undefined`. Here we filter out the `undefined` values.

I don't like the idea of `attachTo`—these elements would be much better explicitly passed in as options rather than as a list—but now's not the time for a refactor…

PR #11960, opened 2016-02-12
Recommendation button for 'more replies' comments should not be disabled

TBonnin:
Comments loaded via the 'Load X more replies' always had their
recommendation button disabled.
Chrome seems to dispatch the click event anyway, even when the button is
disabled. That would explain why the recommendation feature for those
comments was still working on Chrome
On the other hand, Firefox doesn't seem to dispatch the click event when
button is disabled
Read more at https://code.google.com/p/chromium/issues/detail?id=243199

### DIFF
--- a/discussion/app/controllers/CommentsController.scala
+++ b/discussion/app/controllers/CommentsController.scala
@@ -46,10 +46,10 @@ object CommentsController extends DiscussionController with ExecutionContexts {
         Cached(60) {
           if (request.isJson)
             JsonComponent(
-              "html" -> views.html.fragments.comment(comment).toString
+              "html" -> views.html.fragments.comment(comment, comment.discussion.isClosedForRecommendation).toString
             )
           else
-            Ok(views.html.fragments.comment(comment))
+            Ok(views.html.fragments.comment(comment, comment.discussion.isClosedForRecommendation))
         }
     }
   }


rich-nguyen:
nice :+1:

TBonnin:
@sndrs

discussion/app/controllers/CommentsController.scala
```diff
@@ -46,10 +46,10 @@ object CommentsController extends DiscussionController with ExecutionContexts {
         Cached(60) {
           if (request.isJson)
             JsonComponent(
-              "html" -> views.html.fragments.comment(comment).toString
+              "html" -> views.html.fragments.comment(comment, comment.discussion.isClosedForRecommendation).toString
             )
           else
-            Ok(views.html.fragments.comment(comment))
+            Ok(views.html.fragments.comment(comment, comment.discussion.isClosedForRecommendation))
```
TBonnin: `isClosedForRecommendation` default value is `false` 
https://github.com/guardian/frontend/blob/master/discussion/app/views/fragments/comment.scala.html#L1

sndrs:
great!

TBonnin:
Note: It means that relying on button `disabled=true` to prevent recommendation of comments when recommendations are supposed to be closed won't work on Chrome... :(

PR #11961, opened 2016-02-12
Phone number styling improvements

markjamesbutler:
Small change to inline and width.

screen shot 2016-02-12 at 15 35 55


susiecoleman:
:+1:

johnduffell:
actually pasteup-forms comes from https://github.com/guardian/pasteup-forms/blob/master/src/_forms.scss and this is just a copy of that that so presumably we should be updating the original unless we intend to fork it

PR #11962, opened 2016-02-12
margin tweaks on logo/blocks, link style updated

superfrank:
minute


static/src/stylesheets/module/content/_article-minute.scss
```diff
@@ -469,6 +472,17 @@ $minute-layout-margin-nudge: gs-span(1);
         color: $us-minute-contrast-colour;
     }
 
+    a:link {
```
gtrufitt: can do this as 

`
a {
 border-bottom: 1px solid rgba(255, 255, 255, .4);
&:hover { styles}
`

gtrufitt:
:+1: After the anchor stuff

PR #11963, opened 2016-02-12
Refactor the fluid 250 adverts

regiskuckaertz:
This is initially to allow the Merchandising team to use a simplified version of that component, but I caught the bug and decided to get rid of as much redundancy as I could.

I did not try to get too fancy with the code, so the result is still as ugly as it initially was, just a few KB smaller.


uplne:
In this case, I think, the most important thing is to make sure that all parties know what is happening and everything is 100% tested.

regiskuckaertz:
What do you mean? I made it so that nothing changes in DFP—which explains the ugly detection in `breakoutIframe`

regiskuckaertz:
I'll add some unit tests :smiley:

jbreckmckye:
All looks OK :+1:

uplne:
Is this worth unit testing? Definitely don't focus too much on testing DOM manipulations.

uplne:
Definitely nice work removing a lot of code :+1: :hammer:

regiskuckaertz:
There's definitely a lot going on in this supposedly simple module. A lot of it comes from the sheer number of parameters and their respective values in DFP. We should add this to the list of TODOs for next quarter. Anyway, let's merge this mozafucka.

uplne:
Whatever you add to TODO has ~5% chance to be really done.

regiskuckaertz:
@uplne actually you got that already covered in your doc, under the heading **Creatives** :fist:

OliverJAsh:
Nice work removing 361 lines from the codebase :-) :+1:

regiskuckaertz:
@OliverJAsh yeah :wine_glass:

PR #11965, opened 2016-02-12
Remove the nextgen profile from the AWSCredentialsProviderChain

janua:
Will switch this over on monday as there could be unforeseen consequences.

@rich-nguyen


rich-nguyen:
:+1:

PR #11968, opened 2016-02-15
Fix logical fallacy to detect wider columns for comments in articles

regiskuckaertz:
Currently a `.discussion__ad-wrapper-wider` class is added to the container of the comments MPU in all cases except when the page is both a liveblog and a minute article—i.e. always.

Here is what happens:
picture 1


jbreckmckye:
One step at a time...

:+1:

regiskuckaertz:
I wish one day there won't be an expired switch coming in the way :sob:

gtrufitt:
Whoops, my bad :+1:

sndrs:
:dizzy: :dizzy_face:

regiskuckaertz:
http://giphy.com/gifs/fcK30LKXjG6Tm

PR #11969, opened 2016-02-15
Force schema.org/NewsArticle for all amplified articles

TBonnin:
Google search carousel doesn't support other schema type at the moment
This commit forces reviews to be NewsArticles (only amp'ed page though)
A feature switch has also be added to force us to reevaluate this
workaround on a regular basis


TBonnin:
@stephanfowler @NataliaLKB

stephanfowler:
:+1:

PR #11970, opened 2016-02-15
Add global async adblock detection

sndrs:
This is now the first part of a series of PRs to simplify #11834, for #11789.

Deprecates the synchronous adblock method and updates current instances to use that. 

I'll remove them individually, subsequently, because the various calls to `detect.adblockInUse` are a bit of a thicket and the ramifications of `then`ing them a bit thorny

@jbreckmckye @regiskuckaertz


common/app/templates/inlineJS/nonBlocking/detectAdblock.scala.js
```diff
@@ -0,0 +1,26 @@
+try {
+    ((document, window) => {
```
jbreckmckye: Will this get transpiled to an ES5 function?

sndrs: yeah, closure compiler is used for inline JS

static/src/javascripts/projects/common/utils/detect.js
```diff
@@ -428,6 +437,26 @@ define([
...
     }
+    // end sync adblock detection
+
+    var getAdBlockers = new Promise(function (resolve) {
+        if (has(window.guardian.adBlockers, 'generic')) {
+            // adblock detection has completed
+            resolve(window.guardian.adBlockers);
+        } else {
+            // Push a listener for when the JS loads
+            window.guardian.adBlockers.onDetect = resolve.bind(this, window.guardian.adBlockers);
```
OliverJAsh: You shouldn't need to bind `resolve`?

common/app/templates/inlineJS/nonBlocking/detectAdblock.scala.js
```diff
@@ -0,0 +1,26 @@
...
+        ad.style.position = 'absolute';
+        ad.style.left = '-9999px';
+        ad.style.height = '10px';
+        ad.innerHTML = '&nbsp;';
+        ad.setAttribute('class', 'ad_unit');
+
+        window.requestAnimationFrame(() => {
+            document.body.appendChild(ad);
+
+            window.requestAnimationFrame(() => {
```
OliverJAsh: May be worth leaving a comment to say this is to avoid forced sync layout, as some people might not realise this.

static/src/javascripts/projects/common/utils/detect.js
sndrs: callbacks in `then` expect a reference to the detect object

OliverJAsh:
Nice :+1:

common/app/templates/inlineJS/nonBlocking/detectAdblock.scala.js
```diff
@@ -0,0 +1,28 @@
...
+    ((document, window) => {
+        window.guardian.adBlockers = {};
+        var ad = document.createElement('div');
+        ad.style.position = 'absolute';
+        ad.style.left = '-9999px';
+        ad.style.height = '10px';
+        ad.innerHTML = '&nbsp;';
+        ad.setAttribute('class', 'ad_unit');
+
+        // avoid a forced sync layout
```
jbreckmckye: Long term, making adblock detection async is very helpful, because it opens up the way to more accurate, non-synchronous heuristics - like seeing if an ad request gets blocked at the network level.

static/src/javascripts/projects/common/utils/detect.js
regiskuckaertz: I likez it :smile: You could use `null` instead of `this` to make sure the context is right. Not sure if that's the case, but is there a polyfill for `bind` already loaded at this stage? (mostly for IE8 users)

common/app/templates/inlineJS/nonBlocking/detectAdblock.scala.js
```diff
@@ -0,0 +1,28 @@
...
+        window.requestAnimationFrame(() => {
+            document.body.appendChild(ad);
+
+            // avoid a forced layout, and be sure the element has been added to the DOM
+            window.requestAnimationFrame(() => {
+                var adStyles = window.getComputedStyle(ad);
+                window.guardian.adBlockers.generic = adStyles.getPropertyValue('display') === 'none';
+                window.guardian.adBlockers.ffAdblockPlus = adStyles.getPropertyValue('-moz-binding').match('elemhidehit') !== null;
+                try {
+                    window.guardian.adBlockers.onDetect();
```
jbreckmckye: Rather than binding the adBlockers object to the promise resolution in detect, perhaps we should just pass the object to onDetect right here?

regiskuckaertz:
This is a minor detail, but I've read from multiple sources that `try/catch` blocks are not optimised by the browser. Improving on this could be done two ways AFAIK.

1. Calling an optimised function inside the `try/catch`:

```
function foo() {
 // move all ad block detection code here
}
try {
 foo();
} catch { ... }
```

2. [Getting rid of the `try/catch` altogether by using `window.onerror` instead](http://ryanmorr.com/reinventing-the-try-catch-block/).

static/src/javascripts/projects/common/utils/detect.js
```diff
@@ -396,7 +400,12 @@ define([
...
     }
 
-    function adblockInUse() {
+    // sync adblock detection is deprecated.
+    // it will be removed once sticky nav and omniture top up call are both removed
+    // this is soon - it's not worth refactoring them when they're off soon
+    //
+    // ** don't forget to remove them from the return object too **
+    function adblockInUseSync() {
         if (!detect.cachedAdblockInUse) {
```
jbreckmckye: Might want to rename this cache property.

common/app/templates/inlineJS/nonBlocking/detectAdblock.scala.js
sndrs: yeah miles nicer. i'll include this in a subsequent PR

jbreckmckye: Super pedantic, but should this technically be an array of listeners? I guess I'm splitting hairs here...

jbreckmckye:
Looks good :+1:

jbreckmckye:
@regiskuckaertz Yes. Try blocks introduce dynamic scope. Each one adds a new variable object at the top of the scope chain of the current execution context. Functions with try/catches therefore perform like functions augmented using `with` and are always run in interpreted mode.

The difference may be moot, though, because this code is only run once and then discarded anyway - meaning it may not even pass through the optimizer in the first place. And you probably wouldn't want it to anyway - the optimizer doesn't come for free and - last I understood - can't be used on running code without blocking its execution.

common/app/templates/inlineJS/nonBlocking/detectAdblock.scala.js
```diff
@@ -0,0 +1,29 @@
...
+        // avoid a forced sync layout, and open door to more accurate detection
+        // e.g. detecting network behaviour
+        window.requestAnimationFrame(() => {
+            document.body.appendChild(ad);
+
+            // avoid a forced layout, and be sure the element has been added to the DOM
+            window.requestAnimationFrame(() => {
+                var adStyles = window.getComputedStyle(ad);
+                window.guardian.adBlockers.generic = adStyles.getPropertyValue('display') === 'none';
+                window.guardian.adBlockers.ffAdblockPlus = adStyles.getPropertyValue('-moz-binding').match('elemhidehit') !== null;
```
gtrufitt: I know this is merged but I think this is causing `Cannot read property 'match' of null` in the console ( I guess because `getPropertyValue('-moz-binding')` only has a return value in firefox?). Would put in a fix, but didn't want to mess about too much with something that seemed fairly fragile from the conversations on this PR!

jbreckmckye: I just noticed this too. I will put in a fix.

gtrufitt: :+1:

sndrs: what's this happening in? `window.getComputedStyle(document.body).getPropertyValue('-moz-binding')` = `''` in chrome?

gtrufitt:
```
window.getComputedStyle(document.body).getPropertyValue('-moz-binding')
null
```

in Chrome so calling `.match ` on `null` value

sndrs: that's really weird, I get:

```
"" === window.getComputedStyle(document.body).getPropertyValue('-moz-binding')
true
```

gtrufitt: :frowning: 

image

PR #11973, opened 2016-02-15
Fix broken template in fluid 250

regiskuckaertz:
A tentative compilation of an already compiled template resulted in this:
picture 2

This is fixed:
picture 3


uplne:
:+1:

PR #11974, opened 2016-02-15
International coin & contributor banners

tomverran:
So this PR enables USD contributor banners for the US and International editions:

image

It also enables a US currency version of the coin banner for the international edition only

image


tomverran:
Hi @jbreckmckye or @Calanthe would you mind taking a look at this?

Thanks!

static/src/javascripts/projects/common/modules/adblock-banner.js
```diff
@@ -49,7 +50,7 @@ define([
                 customCssClass: this.customCssClass,
                 cursor: svgs('cursor'),
                 rectangleLogo: svgs('logomembership'),
-                adblockCoins: svgs('adblockCoins'),
+                adblockCoins: svgs('adblockCoins' + this.edition),
```
jbreckmckye: Won't this try and fetch the wrong SVGs if we're under the UK or AUS editions?

static/src/javascripts/projects/common/views/svgs.js
```diff
@@ -107,7 +109,9 @@ define([
         externalLink: externalLink,
         tick: tick,
         glabsLogoSmall: glabsLogoSmall,
-        adblockCoins: adblockCoins,
+        adblockCoinsUK: adblockCoins,
+        adblockCoinsUS: adblockCoinsUS,
+        adblockCoinsINT: adblockCoinsUS, //deliberate
```
jbreckmckye: I think any proprietary SVG mappings for our banner should happen in `adblock-banner` rather than the generic SVG fetch module.

static/src/javascripts/projects/common/modules/commercial/donot-use-adblock.js
```diff
@@ -72,35 +76,54 @@ define([
...
+                    supporterLink: 'https://membership.theguardian.com/supporter?INTCMP=ADBLOCK_BANNER_MACASKILL',
+                    quoteText: 'The Guardian enjoys rare freedom and independence. Support our journalism',
+                    quoteAuthor: 'Ewen MacAskill',
+                    customCssClass: 'macaskill',
+                    imageAuthor: '//i.guim.co.uk/img/static/sys-images/Guardian/Pix/contributor/2015/8/18/1439913873894/Ewen-MacAskill-R.png?w=300&q=85&auto=format&sharp=10&s=0ecfbc78dc606a01c0a9b04bd5ac7a82'
+                }
+            ],
+
+            internationaliseBanner = function (country) {
+                return function (contributor) {
```
jbreckmckye: This seems an odd signature to me - I would make this perform the merge directly. If you need to curry functions with parameters when passing it to a `map` (as you seem to be later) you can use Lodash's `_.partial` helper.

static/src/javascripts/projects/common/modules/commercial/donot-use-adblock.js
```diff
@@ -72,35 +76,54 @@ define([
...
-                customCssClass: 'monbiot',
-                imageAuthor: '//i.guim.co.uk/img/static/sys-images/Guardian/Pix/contributor/2015/7/9/1436429159376/George-Monbiot-L.png?w=300&amp;q=85&amp;auto=format&amp;sharp=10&amp;s=903233b032379d7529d7337b8c26bcc9'
-            },
-            {
-                supporterLink: 'https://membership.theguardian.com/supporter?INTCMP=ADBLOCK_BANNER_MUIR',
-                quoteText: 'Support and become part of the Guardian',
-                quoteAuthor: 'Hugh Muir',
-                customCssClass: 'muir',
-                imageAuthor: '//i.guim.co.uk/img/static/sys-images/Guardian/Pix/pictures/2014/3/13/1394733739000/HughMuir.png?w=300&amp;q=85&amp;auto=format&amp;sharp=10&amp;s=c1eeb35230ad2a215ec9de76b3eb69fb'
+            contributorBannersUK = [
```
jbreckmckye: I would keep the `contributorBanners` edition-independent and turn them into UK banners with the internationalize function you're using for the others.

static/src/javascripts/projects/common/modules/commercial/donot-use-adblock.js
```diff
@@ -110,21 +133,31 @@ define([
...
             ],
 
-            coinBannersRelevantToEdition = filter(coinBanners, function (message) {
-                return !message.edition || message.edition == config.page.edition;
-            }),
+            coinBannersRelevantToEdition = filter(coinBanners, relevantToEdition),
+            contributorBannersRelevantToEdition = filter(contributorBannersUK.concat(
+                map(contributorBannersUK, internationaliseBanner('INT')),
+                map(contributorBannersUK, internationaliseBanner('US'))
+            ), relevantToEdition),
```
jbreckmckye: It seems that what you're really trying to do is create a set of contributorBanners that have their edition set to INT, UK or US if that's the edition we're currently reading - is that correct? It seems there might be a simpler way, e.g.

    function getContributorBanners() {
        var haveBanners = edition.match(/^(UK|US|INT)$/);
        return haveBanners ? map(banners, internationalise(edition)) : [];
    }

tomverran: That is indeed much better! Attending to this and the other comments now, thanks!

tomverran:
Hi @jbreckmckye I've made some alterations to try and accommodate more banner variations without so many code changes.

The idea is that banners get configured in `adblock-banner-config.js`, which merges together `defaults`, `variations` and `editions` to produce a final list of banners which `donot-use-adblock.js` can then choose from.

I also wrote tests to eliminate the super slow make JS change -> recompile everything -> reload page cycle which I'd been using. :snail:

static/src/javascripts/projects/common/modules/commercial/adblock-banner-config.js
```diff
@@ -0,0 +1,127 @@
...
+    ];
+
+
+    return {
+        banners: banners,
+        getBanners: function (edition) {
+            if (!edition) {
+                return [];
+            }
+            var out = [];
```
jbreckmckye: You can use `return array.map(x => fx)` rather than `accumulator = []; array.forEach(x => {accumulator.push(x)}); return accumulator`.

static/test/javascripts/spec/common/adblock-banner.spec.js
```diff
@@ -0,0 +1,22 @@
...
+
+    describe('Adblock banner', function () {
+
+        var ukBanners;
+        beforeEach(function () {
+            document.innerHtml = '<div class="top-banner-ad-container--desktop"></div>';
+            ukBanners = adblockBannerConfig.getBanners('UK');
+        });
+
+        it('should be able to render every type of banner', function () {
```
jbreckmckye: I'm not sure if this test is necessary - we already know Lodash templates and Fastdom work.

static/test/javascripts/conf/adblock.js
```diff
@@ -0,0 +1,9 @@
+module.exports = function (config) {
```
jbreckmckye: Why do we need this configuration file? We are already pulling in specs in /test/javascripts/spec/common/*

static/src/javascripts/projects/common/modules/commercial/adblock-banner-config.js
tomverran: Ah yes my fear with map, at least lodash's map, is that the thing you're mapping gets passed by reference and can be accidentally modified.

I shall have a go at using it though, I do agree it'd be nicer

static/test/javascripts/spec/common/adblock-banner.spec.js
tomverran: This test is probably a bit misleading in that what I'm actually testing is that the configs for the templates provide all the variables etc that the templates need. 

I could probably include it alongside all the other config tests and name it slightly better.

static/src/javascripts/projects/common/modules/commercial/adblock-banner-config.js
jbreckmckye: I see - well, Lodash's map does not mutate the original collection:

image

static/test/javascripts/spec/common/adblock-banner.spec.js
jbreckmckye: If you wish to test that the model variables are present, it may be preferable to just read them off the objects themselves. Testing the actual render means writing an async test and changing the contract of the code-under-examination to return a promise that informs the test when the fastdom write has finished.

static/test/javascripts/conf/adblock.js
tomverran: We are, but the common tests take ~30s to run while the adblock ones are instant
I am happy to remove these files though and keep them only locally

jbreckmckye: You can run an individual Jasmine unit test by changing the name of the describe block to `fdescribe` - the 'f' standing for 'focus'. The advantage of this approach is that you can never accidentally commit it - there is a linting rule - run on pushing upstream - that will bounce tests with `fdescribe` and `fit` blocks.

tomverran: :+1: Much better!

static/test/javascripts/spec/common/adblock-banner.spec.js
tomverran: So I've split out the synchronous template rendering and am now explicitly testing that - I'm still keen on testing the actual rendering of the actual templates as it can be difficult to see exactly which templates depend on which variables being present

jbreckmckye: Exactly what kinds of failures are you trying to test, and how would your test capture them?

tomverran: I'm trying to get early feedback if I remove a variable from the configuration that is still referenced in a template - This test saved me some time as I ported the configuration across from donot-use-adblock to adblock-banner-config

jbreckmckye: Ah, so it'd be a kind of integration test? You'd be looking to make sure that none of the template variables were replaced with `undefined`s.

If you do that, you might find it easier to test not the result of the render, but the result of the function that will be passed to fastdom. This would make testing a lot easier, because otherwise you have to unit test an asynchronous function and give it a way of telling the test framework that it's finished.

jbreckmckye:
:+1: Let's ship it!

tomverran:
Thanks for all your help @jbreckmckye :+1:

PR #11976, opened 2016-02-15
Full fix: Force schema.org/NewsArticle for all amplified articles

TBonnin:
Complement of #11969 (which was an incomplete fix). The body type need to change depending on the schema type.


TBonnin:
@stephanfowler @NataliaLKB I need you again :) My first patch was incomplete.

stephanfowler:
:+1:

NataliaLKB:
@TBonnin thanks for finding this!

PR #11978, opened 2016-02-15
track previous content type

jamesgorrie:
Seems it's available in omniture but we aren't populating it.

[There's other examples in the vendor file, but I assumed that was supplied](https://github.com/guardian/frontend/blob/master/static/public/javascripts/vendor/omniture.js#L89-L90).


rich-nguyen:
:+1:

PR #11979, opened 2016-02-16
Align inline MPUs insertion logic across all breakpoints

regiskuckaertz:
We don't discriminate between mobile/tablet/desktop anymore.


uplne:
It simplify things but it means more data :+1:

regiskuckaertz:
Yes, indeed. This is somehow related to the tests @michaelwmcnamara did with the new dynamic ads in liveblogs, there is a cost for the user and we should keep track of the corresponding metrics.

PR #11980, opened 2016-02-16
Add fallback image metadata to amp article if no images

TBonnin:
A value for the image field is required for AMP articles
https://developers.google.com/structured-data/rich-snippets/articles
which is a kind of ridiculous requirement for a webpage. :(
In order to please (trick?) the validator we are adding an invisible
image metadata (fallback logo) if the article content has no images


TBonnin:
cc @NataliaLKB @stephanfowler

NataliaLKB:
:+1:

NataliaLKB:
This may look so weird in the carousel! However better than it not loading at all. :ship: :smile:

TBonnin:
That's already the fallback image we are using for open graph metadata. Therefore a twitter card for an article with no image already looks like this
screen shot 2016-02-16 at 13 20 07

PR #11981, opened 2016-02-16
Adds a switch for the r2 headers to be passed with the request url.

jennysivapalan:
Uses a switch to determine if to hit R2 or Dotcom. This is required as we need to repress a few urls to be ready for https but they have been taken down in R2 and we are using the pressed versions. 

This is a bit manual but it's just required for a few examples.
cc @JustinPinner


JustinPinner:
:+1: It's nasty but it's temporary :clock11:

PR #11983, opened 2016-02-16
Fix edition-deriving code for top mobile ad slot

kelvin-chappell:
So that it uses all geotargets to derive edition rather than just the first one.

/cc @regiskuckaertz


regiskuckaertz:
:+1:

PR #11985, opened 2016-02-16
The minute fullscreen mobile

gtrufitt:
Fullscreen images for the minute on mobile.

mobile_full_screen_images

We use `object-fit: cover` along with `@supports` to ensure browsers that only support object-fit get the css related to the full screen images.


common/app/layout/WidthsByBreakpoint.scala
```diff
@@ -200,7 +200,8 @@ object ContentWidths {
 
   object MinuteMedia extends ContentRelation {
     override val inline = WidthsByBreakpoint(
-      mobile = Some(95.vw),
+      // In the minute inline images match the immersive header size
+      mobile = Some(1000.px),
```
OliverJAsh: `100vw`?

gtrufitt: Nope, it's like the immersive header where the ratio stays the same, so wider than viewport but editors select images that crop into the middle.

static/src/javascripts/bootstraps/enhanced/common.js
```diff
@@ -140,6 +140,8 @@ define([
                     && !(config.switches.disableStickyAdBannerOnMobile && detect.getBreakpoint() === 'mobile')
                     && config.page.contentType !== 'Interactive'
                     && config.page.contentType !== 'Crossword'
+                    && !config.page.isImmersive
+                    && !config.page.isUsMinute
```
OliverJAsh: Are all minutes immersive?

static/src/stylesheets/module/content/_article-minute.scss
```diff
@@ -455,84 +544,49 @@ $minute-layout-margin-nudge: gs-span(1);
...
-        color: $us-minute-contrast-colour;
-    }
-
-    .svg  & .caption--img:before {
-        @extend %svg-i-camera-white-large;
-        margin-right: 5px;
-    }
+// Background Image Layout
+@supports (object-fit: cover) {
+    .block--background-image {
```
OliverJAsh: Is this only used in minutes?

static/src/javascripts/bootstraps/enhanced/common.js
gtrufitt: At the moment they are but that may or may not be the case in the future and there is JS hanging off isImmersive config that shouldn't be applied to the minute template hence the separation.

static/src/stylesheets/module/content/_article-minute.scss
gtrufitt: Currently, yeah. Wasn't quite sure whether I should scope this directly to `minute-block ` or make it generic like this (this is a new class so doesn't apply anywhere else at the moment). And if I make it generic, should I put it elsewhere (all block stuff is in liveblog template at the moment...). Thoughts on the architecture here?

OliverJAsh: I was wondering if this should be called `block--minute-article--background-image`? I don't think this is worth abstracting until we can safely identify a pattern?

gtrufitt: Yup, I think you're right. I'll pull these back into the `block--minute-article` scope, was see-sawing on that anyway so no argument from me :)

common/app/layout/WidthsByBreakpoint.scala
OliverJAsh: Makes sense now we spoke in person. Maybe the comment should elaborate more? I didn't know what the immersive header was.

If I understand correctly, we set this to a large pixel size because we use `object-fit` on the image.

OliverJAsh:
Nice work! :+1:

PR #11986, opened 2016-02-16
Stop FF ABP detection throwing exceptions outside Firefox

jbreckmckye:
The adblock detection script throws exceptions outside of Firefox:

image

This line is the culprit:
image
_If the element does not have a mozBinding, we try and call .match on a null_

This pull request squashes the bug:
image

@sndrs @gtrufitt


gtrufitt:
:cool: :+1:

sndrs:
:+1: cheers

PR #11988, opened 2016-02-16
Ad banner: transition transform instead of margin-top

OliverJAsh:
This aims to improve the rendering performance of the ad banner when it expands by using [hardware acceleration](http://www.paulirish.com/2012/why-moving-elements-with-translate-is-better-than-posabs-topleft/). To do this we've ditched `margin-top` and replaced it with `transform`.

With this change, there is a large perceptible improvement in rendering performance and the frame times (seen in the animations below) support this.
# Before

~20-30 fps when transitioning (on my Mac)

no transform
# After

~60 fps when transitioning

transform


static/src/stylesheets/base/_base.scss
```diff
@@ -19,8 +19,7 @@ body {
 
 
 @if $old-ie == false {
-    html,
-    body {
```
OliverJAsh: We had to move this style because of this weird browser behaviour when combining `overflow-x` with `transform`: http://jsbin.com/yocanu/3/edit?html,css,output

jbreckmckye:
:+1: : This is absolutely great. Not sure why it wasn't implemented via a translateY to begin with, though...

Jholder112233:
You've probably covered this, but if we just translate the whole page using a transform, won't the footer be cut off at the bottom?

OliverJAsh:
Nope, it won't. The browser just extends the document size.

On Thu, 18 Feb 2016 at 11:15 Josh Holder <notifications@github.com> wrote:

> You've probably covered this, but if we just translate the whole page
> using a transform, won't the footer be cut off at the bottom?
>
> —
> <https://github.com/guardian/frontend/pull/11988#issuecomment-185665130>.
>

-- 

in England Number 908396

Jholder112233:
Should have tested it myself before commenting, just tried it and you are of course right :ship:

regiskuckaertz:
Damn you bet me to it! It's the first thing I wanted to fix as I started working here. Great job! :+1: :tophat:

static/src/stylesheets/module/_adslot.scss
```diff
@@ -69,8 +69,6 @@
     .sticky-ad-banner & {
         z-index: 1022;
         width: 100%;
-        transform: translateZ(0);
-        will-change: transform;
```
regiskuckaertz: To have any positive effect, the `will-change` property should be set before the animation, and then removed after it has ended.

OliverJAsh: In this case the element is already promoted to its own layer because of `position: fixed`. `will-change: transform` was a lie because we never changed transform.

OliverJAsh: (`position: fixed` is set in JS)

static/src/javascripts/projects/common/modules/commercial/sticky-ad-banner.js
```diff
@@ -124,11 +124,12 @@ define([
         var transitionTimingFunction = 'cubic-bezier(0, 0, 0, .985)';
         var transitionDuration = '1s';
 
-        els.$header.css({
+        els.$page.css({
+            'will-change': 'transform',
```
gtrufitt: IIRC `will-change` [shouldn't be added at the time of animation](https://dev.opera.com/articles/css-will-change-property/#give-the-browser-enough-time-to-work) but should be added in anticipation of the event (and removed if that event never happens). Maybe in the render function you can add a threshold for adding the will-change and track scroll direction in the state? It should also be removed after the animation is complete. It's worth thinking about whether `will-change` is useful on `els.$adBanner` too as it's not just a replacement for the promoting to a layer hack but more a hint to the browser on *all* the things it can optimise for and get ready to change.

OliverJAsh: Yeah, this is a good point. Here I'm always adding it, so it will definitely be there before the event happens (good), but it won't ever be removed (not good).

We only want to add it when the resize event occurs. When that condition holds true we want to add it one animation frame prior to changing the `transform` value, and we want to remove it one animation frame after the transition is complete. Still thinking about how I can do that in a functional way, which is why this PR is still pending.

As for `will-change` on the ad banner element, good point. I can't say `will-change: transform` though, that would be lying to the browser (and for no need as it's already promoted to a layer thanks to `position: fixed`). I could say `will-change: height`, in which case I'm curious what performance optimisations the browser would make?

gtrufitt: Interesting problem - this is an action type of `NEW_AD_HEIGHT`? Maybe you could use `fastdom.defer` to push the css change to the next frame? Why do you need to remove it one animation frame after the animation end rather than in the same frame?

With the banner element, I think `will-change: height` is what I was implying, though perf optimisation wise, I don't know what the browser would do, maybe the `will-change` helps the browser to understand that sibling or parent re-paints are on their way or defer actions that would be overridden by the height change. I need to do a @regiskuckaertz and hunt down some source code to tell me...

regiskuckaertz: :joy: 

The strangest thing about `will-change` is that it's not supposed to appear in a stylesheet. I'm not sure one animation frame is enough to let the browser do whatever groundwork it can do, I don't think you need to be that extreme and in fact could set the property when the DOM is ready for instance.

As for removing it, why not wait for the `transitionend` event?

gtrufitt: > As for removing it, why not wait for the transitionend event?

Yeah, the `AD_BANNER_TRANSITION_END` action :)

>  I'm not sure one animation frame is enough to let the browser do whatever groundwork it can do

I don't know, I reckon it might be. Isn't the point with `will-change` that it is a just-in-time tool that should be attached to tell the browser exactly when it should be performing it's performance optimisations. [The spec](https://www.w3.org/TR/css-will-change-1/#dont-waste) says you should always let the browser handle things as normal if you're not expecting this change to happen soon.

jbreckmckye: My understanding - and this is possibly rather dated - is that `will-change` essentially tells a browser that you want to promote an a node to its own compositor layer, reserving GPU memory for that specific purpose. There might be other consequences - turning off certain optimizations, or caching useful properties for faster reuse - but at a very high level, `will-change` is an alias for the old `translate3d` hack.

Thinking in terms of compositor memory is informative. Obviously, memory is a bounded resource, particularly scarce on mobile, so a complex page with lots of animations (i.e. several adverts) could quite feasibly exhaust it.

What happens when the compositor runs out of memory? I can think of four consequences:

1. If the browser cannot de-promote a `will-change` element from a compositor layer in a graceful, timely manner, it might throw an exception and crash;
2. If the browser is being asked to promote a new element with `will-change`, it might arbitrarily de-promote a `will-change` element that's older;
3. If the browser is being asked to promote a new element with `will-change`, it might just ignore the request and queue it for later consideration.
4. The browser bails out to software composition and possibly starts using a pagefile (yuk!)

The upshot is that we shouldn't add `will-change` arbitrarily, and that setting it too early consumes resources we'd really rather the browser allocated. At the same time, we don't want to set it too _late_. Promoting an element costs time as well as resource, and if we start the animation before promotion has completed, we incur that cost _and_ our animation remains unoptimized. This may well be worse than not using `will-change` at all.

Vendors are coy about the exact 'sweet spot' between setting the property too early and too late. Opera provide a little advice [here](https://dev.opera.com/articles/css-will-change-property/), suggesting 200ms might be the kind of gap to aim for, so it makes sense to assume a `will-change` optimization can take at least entire frames.

regiskuckaertz: :bowtie: sweet!

> but at a very high level, will-change is an alias for the old translate3d hack.

When `will-change: transform|opacity`, yes definitely. Not sure it would be smart of the browser to do that when the element's bounding box is part of the document flow though, or would it? For instance, Blink/Webkit will un-promote a node when `will-change: contents`. Basically [all the properties which trigger layout](https://csstriggers.com/).

It really feels like this property should never have been a property in the first place, but rather an API to manipulate the render tree. I'm very curious to see if this will change (no pun intended) with Houdini.

johnduffell:
this is awesome... is it ready to go out?

jbreckmckye:
If we can achieve a performance benefit without the `will-change` property, let's just ship that and investigate how well it works with / without the additional optimization later.

OliverJAsh:
Yup, @jbreckmckye, I'll do that!

gtrufitt:
:+1:

common/app/views/main.scala.html
```diff
@@ -34,65 +34,73 @@
...
+            }
+        }
 
-                @fragments.header(page)
+        <div id="page">
 
-                @if(adBelowNav) {
-                    @fragments.commercial.topBannerBelowNav()
-                }
+            @fragments.header(page)
```
johnduffell: this should still be inside the hasHeader check

PR #11989, opened 2016-02-16
Remove international message

rich-nguyen:
This is an introductory message, we are getting user responses that this message is no longer needed.

@ScottPainterGNM


ScottPainterGNM:
:+1: When checks have completed - Looks good though!

jennysivapalan:
:+1:

PR #11991, opened 2016-02-16
Pass full list of series tags to omniture

rich-nguyen:
Article content which is part of multiple series needs to pass the name of each series tag in a comma separated list


jennysivapalan:
:+1:

PR #11992, opened 2016-02-16
Use async adblock for membership messages

sndrs:
First part of using async adblock detection, for #11789 and following #11970. Also cleans up following https://github.com/guardian/frontend/pull/11970#discussion_r52889127.

@jbreckmckye @regiskuckaertz @OliverJAsh you might want to check the tests in particular...

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/commercial-feature-policies.js
+++ b/static/src/javascripts/projects/common/modules/commercial/commercial-feature-policies.js
@@ -68,16 +68,15 @@ define([
         }
     };
 
-    policies.membershipMessages = function () {
-        if (!detect.adblockInUseSync() &&
-            detect.getBreakpoint() !== 'mobile' &&
-            config.page.contentType === 'Article' &&
-            !userFeatures.isPayingMember()
-        ) {
-            return {
-                membershipMessages : true
-            };
-        }
+    policies.checkWeCanShowMembershipMessages = function () {
+        return {
+            checkWeCanShowMembershipMessages: detect.adblockInUse.then(function (adblockInUse) {
+                return !adblockInUse &&
+                    detect.getBreakpoint() !== 'mobile' &&
+                    config.page.contentType === 'Article' &&
+                    !userFeatures.isPayingMember();
+            })
+        };
     };
 
     policies.identityPages = function () {
--- a/static/test/javascripts/spec/common/commercial/membership-messages.spec.js
+++ b/static/test/javascripts/spec/common/commercial/membership-messages.spec.js
@@ -4,14 +4,16 @@ define([
     'qwery',
     'common/utils/$',
     'helpers/fixtures',
-    'helpers/injector'
+    'helpers/injector',
+    'Promise'
 ], function (
     bonzo,
     fastdom,
     qwery,
     $,
     fixtures,
-    Injector
+    Injector,
+    Promise
 ) {
     var commercialFeatures, membershipMessages,
         showMembershipMessages, alreadyVisited, storage, config;
@@ -51,15 +53,20 @@ define([
 
         describe('If user already member', function () {
             beforeEach(function (done) {
-                showMembershipMessages = commercialFeatures.membershipMessages;
+                showMembershipMessages = commercialFeatures.checkWeCanShowMembershipMessages;
                 alreadyVisited = storage.local.get('gu.alreadyVisited');
-                commercialFeatures.membershipMessages = false;
+                window.guardian.adBlockers.onDetect = new Promise(function (resolve) {
+                    resolve({
+                        generic: false,
+                        ffAdblockPlus: false
+                    });
+                });
                 storage.local.set('gu.alreadyVisited', 10);
                 done();
             });
 
             afterEach(function () {
-                commercialFeatures.membershipMessages = showMembershipMessages;
+                commercialFeatures.checkWeCanShowMembershipMessages = showMembershipMessages;
                 storage.local.set('gu.alreadyVisited', alreadyVisited);
             });
 
@@ -72,15 +79,20 @@ define([
 
         describe('If user not member', function () {
             beforeEach(function (done) {
-                showMembershipMessages = commercialFeatures.membershipMessages;
+                showMembershipMessages = commercialFeatures.checkWeCanShowMembershipMessages;
                 alreadyVisited = storage.local.get('gu.alreadyVisited');
-                commercialFeatures.membershipMessages = true;
+                window.guardian.adBlockers.onDetect = new Promise(function (resolve) {
+                    resolve({
+                        generic: true,
+                        ffAdblockPlus: true
+                    });
+                });
                 fixtures.render(conf);
                 done();
             });
 
             afterEach(function () {
-                commercialFeatures.membershipMessages = showMembershipMessages;
+                commercialFeatures.checkWeCanShowMembershipMessages = showMembershipMessages;
                 storage.local.set('gu.alreadyVisited', alreadyVisited);
                 fixtures.clean(conf.id);
             });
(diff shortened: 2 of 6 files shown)


regiskuckaertz:
:+1: looks great to me, nice job!

static/src/javascripts/projects/common/modules/commercial/commercial-feature-policies.js
```diff
@@ -68,16 +68,15 @@ define([
...
-        if (!detect.adblockInUseSync() &&
-            detect.getBreakpoint() !== 'mobile' &&
-            config.page.contentType === 'Article' &&
-            !userFeatures.isPayingMember()
-        ) {
-            return {
-                membershipMessages : true
-            };
-        }
+    policies.checkWeCanShowMembershipMessages = function () {
```
jbreckmckye: I'm not sure this is aligned with the way the feature policies are supposed to work. Granted, the way they operate is opaque, and that is my fault - I wrote it and, in retrospect, overengineered it.

The idea is that a policy is maps a condition or a 'mode' - 'ad free web', 'sensitive content', etc. - to a feature change - 'ad free web disables foo, bar, baz, bam...'. So a policy is named after an assertion and returns a map of booleans to show its changes. The maps of each policy then get flattened together, assuming that they contain bools.

This code will work, but it's clear that the commercial feature policies module isn't fit for purpose. The reduction function is now misleading.

For the moment, I don't think this PR needs to refactor the features module, but I would rename the returned flag after a feature 'membershipMessages' rather than a function.

sndrs: do you mean use `policies.membershipMessages` as before? i only changed it to imply it wasn't now simply a set of values – very happy to change the name back if it's not helped?

jbreckmckye: It probably is helpful to show that something isn't just a boolean... would `membershipMessagesPromise` work?

sndrs: i'm pretty uneasy about naming things after their implementation, rather than their role. if promises are replaced, this function will presumably still be async (maybe even `then`able) but not a promise. `checkX.then` seems so much easier to understand, to my mind, than `xPromise.then`?

OliverJAsh: Perhaps `membershipMessages` should return `Promise[Boolean]` rather than return `Map[String, Promise[Boolean]]`? Unless you need to return a map/object, then the function could return `Promise[Map[String, Boolean]]`, so there is only one outer promise.

``` js
policies. checkWeCanShowMembershipMessages = function () {
    return detect.adblockInUse.then(adBlockInUse => {
        return !adBlockInUse
            && …
        }
    })
};
```

static/test/javascripts/spec/common/commercial/membership-messages.spec.js
```diff
@@ -51,15 +53,20 @@ define([
...
-                showMembershipMessages = commercialFeatures.membershipMessages;
+                showMembershipMessages = commercialFeatures.checkWeCanShowMembershipMessages;
                 alreadyVisited = storage.local.get('gu.alreadyVisited');
-                commercialFeatures.membershipMessages = false;
+                window.guardian.adBlockers.onDetect = new Promise(function (resolve) {
+                    resolve({
+                        generic: false,
+                        ffAdblockPlus: false
+                    });
+                });
```
OliverJAsh: I'm a bit confused as to why we set `window.guardian.adBlockers.onDetect` to a promise here, whereas elsewhere this property is set to a function?

static/src/javascripts/projects/common/modules/commercial/commercial-feature-policies.js
jbreckmckye: @OliverJAsh Policies return maps that then get squashed together. The idea - back at the time - was that this meant a policy could map an abstract idea to a set of feature changes, e.g.

    policies.nonEnhancedUsers = ()=> {
        if (config.isEnhanced === false) {
            return {
                resizableTopNav : false,
                highPerformanceTopNav : true,
                viewabilityTracking : false
            }
        }
    }

et cetera, et cetera.

@sndrs Is naming things `promise` naming them after their implementation, or their contract / interface? I would say `checkWeCanShowMessages` sounds like a function to me.

sndrs: the whole file is set up like that, and a `mapValues` creates a simple object out of them. 

i'd rather see this I think:
```javascript
policies.checkWeCanShowMembershipMessages = detect.adblockInUse.then(function (adBlockInUse) {
    return !adBlockInUse
        && …
});
```
but i don't really want to get in to refactoring commercial features as part of this :)

@jbreckmckye if you'd rather the name stayed the same, i have no problem with that, you touch this code far more than me anyway. shall i change it back?

static/test/javascripts/spec/common/commercial/membership-messages.spec.js
sndrs: good point, i don't know :wink:

static/src/javascripts/projects/common/modules/commercial/commercial-feature-policies.js
jbreckmckye: > i'd rather see this I think: {code}

I was thinking of refactoring the module on similar lines. The notion of 'policies' which mapped 'concepts' to 'feature diffs' sounded interesting on paper, but has caused confusion in practice, and most changes to this file have involved creating policies for the case of turning on / off individual flags, which is quite boilerplatey. The idea just hasn't worked well.

>  shall i change it back?

No, I agree the invocation with `then` is more readable, and like that the name suggests a different interface. I think it's better than reverting to the previous name.

sndrs:
@OliverJAsh my approach in the tests (using `guardian.adBlockers` was wrong I think - I've updated it, would you mind taking another look?

static/src/javascripts/projects/common/modules/commercial/commercial-feature-policies.js
OliverJAsh: Still confused how this is a function called `checkWeCanShowMembershipMessages` that returns an object with a property of the same name, `checkWeCanShowMembershipMessages`.

I generally think that bad abstractions are worse than duplication of code. Would it be better just to have a `checkWeCanShowMembershipMessages` function that returns a `Promise[Boolean]` and worry about abstracting it when/if we need to?

static/test/javascripts/spec/common/commercial/membership-messages.spec.js
```diff
@@ -51,57 +53,60 @@ define([
...
-                it('should show a message to engaged readers', function () {
+                it('should show a message to engaged readers', function (done) {
                     config.page = { edition: 'UK' };
                     storage.local.set('gu.alreadyVisited', 10);
-                    membershipMessages.init();
-                    var message = document.querySelector('.js-site-message.site-message--membership-message-uk');
-                    expect(message).not.toBeNull();
+                    membershipMessages.init().then(function () {
+                        var message = document.querySelector('.js-site-message.site-message--membership-message-uk');
+                        expect(message).not.toBeNull();
```
OliverJAsh: This is more of a note on the tests than your PR…

If all DOM side effects were in a `render` function, I suspect it would clean up these tests quite a bit. Instead of interfacing with local storage and mocking global variables, you would just pass in mock state into `render` (`{ edition, alreadyVisited, canShowMembershipMessages }`) and verify the outcome.

OliverJAsh:
Only minor notes, otherwsie :+1: Nice work!

static/test/javascripts/spec/common/commercial/membership-messages.spec.js
sndrs: agreed

static/src/javascripts/projects/common/modules/commercial/commercial-feature-policies.js
sndrs: yeah i think @jbreckmckye's in agreement; beyond the scope of this, but worth looking at

OliverJAsh: Could we just refactor and move this into `membership-messages.js` until the abstraction becomes clear then?

static/test/javascripts/spec/common/commercial/membership-messages.spec.js
jbreckmckye: By decoupling rendering the template from injecting it into the page, we could make a lot of our modules synchronous in behaviour. This would be a big improvement; lots of current code seems to labour under promise boilerplate.

Issue #11944, opened 2016-02-11
Main images on amp pages are cut off on smaller devices


michaelwmcnamara:
Main images are cut off on amp pages. Not padding this time. Looks like either the image size is too big or the aspect ratio is wrong.
Example - see page https://amp.theguardian.com/stage/2015/dec/08/lazarus-review-david-bowie-jukebox-musical-off-broadway-new-york on an iphone4 (note google dev tools, set to emulate an iphone4 will also show the issue)
![image](https://cloud.githubusercontent.com/assets/7239344/12983204/22dd2a74-d0e0-11e5-9a9c-b3d356bce641.png)

Seen on multiple pages.

michaelwmcnamara:
Same page on iphone4 - non amp.
https://www.theguardian.com/stage/2015/dec/08/lazarus-review-david-bowie-jukebox-musical-off-broadway-new-york
![image](https://cloud.githubusercontent.com/assets/7239344/12983289/8da6f6b4-d0e0-11e5-9420-0c98db8ff597.png)

michaelwmcnamara:
Had to add this one - its too funny!
from this:
![image](https://cloud.githubusercontent.com/assets/7239344/12985100/c4d85440-d0e8-11e5-8ab4-fafa3c0123aa.png)

to this:
![image](https://cloud.githubusercontent.com/assets/7239344/12985067/9ec7ee32-d0e8-11e5-8c86-425ef0bd42c3.png)

Issue #11943, opened 2016-02-11
In body images are cut off on right-hand-side


michaelwmcnamara:
Seen on iPhone4 running iOS9 using Safari browser and also on Android Chrome, and the Android emulator in Chrome dev-tools.
Example page:
https://amp.theguardian.com/music/2016/jan/15/david-bowies-last-days-an-18-month-burst-of-creativity
Suspect the culprit is excessive padding on the left hand side, which pushes the image off the screen.
Example images:

![image](https://cloud.githubusercontent.com/assets/7239344/12982890/cacb76b6-d0de-11e5-8d36-f3aa6b88cdf1.png)

![image](https://cloud.githubusercontent.com/assets/7239344/12982875/bd32db02-d0de-11e5-8fd0-08e66b53df07.png)

PR #11993, opened 2016-02-16
Taking out srcset for responsive images on amp, use fallback src instead

NataliaLKB:
Fixes both #11944 #11943 

It appears as though srcset for amp-images are not behaving as expected. I propose that for now we just use the fallback source and make a test page after launch for the amp team to look at. I can't find another publisher using srcset to compare their approach.

cc @stephanfowler @michaelwmcnamara


rich-nguyen:
:+1:

PR #11996, opened 2016-02-17
Pass commissioning desk trackingIds to Omniture

rich-nguyen:
Capi provides a new tag type, `Tracking`. This contains Ids that we should pass to omniture through prop5.


jennysivapalan:
:+1:

Issue #11966, opened 2016-02-12
AMP - captions on images have extra top padding


Jholder112233:
<img width="559" alt="screen shot 2016-02-12 at 18 58 29" src="https://cloud.githubusercontent.com/assets/2236852/13017132/a9adfd1e-d1ba-11e5-9add-d51802b133cc.png">

https://amp.theguardian.com/music/2015/nov/12/prince-interview-paisley-park-studios-minneapolis

@NataliaLKB @zeftilldeath

NataliaLKB:
@Jholder112233 Thanks for catching this!

The problem is we have a block share element there between the picture and the caption. The svgs are being loaded, but just don't have any styling. @stephanfowler @zeftilldeath do we want to show more share buttons (style these), or should I remove them?

Jholder112233:
Hey @NataliaLKB, looks like the .rounded-icon class has display: inline-block on it, so the height of the "inline-expand-image" span is still there. Going to be one for an !important sadly!

NataliaLKB:
Yes, you are right! I will try to think of a way to not use `!important`

Jholder112233:
Maybe the cleanest solution is to clean them out of the dom in the first place?

I wouldn't be worried about putting an !important in for AMP though, feels like a valid use. (I've put plenty of them in the codebase when necessary for overrides)

NataliaLKB:
Yes I think cleaning them out in the first place is the answer - as we don't show them at all on mobile anyway.

NataliaLKB:
Closing! Works on code now: http://amp.code.dev-theguardian.com/music/2015/nov/12/prince-interview-paisley-park-studios-minneapolis

Thanks @Jholder112233  :smile:

PR #11997, opened 2016-02-17
Display:none on share block (between caption and pictures) in amp

NataliaLKB:
fixing: #11966

### DIFF
--- a/common/app/views/fragments/amp/stylesheets/main.scala.html
+++ b/common/app/views/fragments/amp/stylesheets/main.scala.html
@@ -246,6 +246,12 @@
         list-style: none;
     }
 
+    @* temporarily hiding these for after launch *@
+    .inline-expand-image, .content__dateline-lm , .block-share, .witness-cta-wrapper, .block-share {
+        display: none;
+        visibility: hidden;
+    }
+
     @fragments.amp.stylesheets.tones()
     @fragments.amp.stylesheets.embeds()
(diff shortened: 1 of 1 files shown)


common/app/views/fragments/amp/stylesheets/main.scala.html
```diff
@@ -246,6 +246,12 @@
         list-style: none;
     }
 
+    @* temporarily hiding these for after launch *@
```
jamesgorrie: until after launch?

jamesgorrie:
:+1:

common/app/views/fragments/amp/stylesheets/main.scala.html
NataliaLKB: Haha yes... Thanks you!

common/app/views/fragments/amp/stylesheets/main.scala.html
```diff
@@ -246,6 +246,12 @@
...
+    .inline-expand-image, .content__dateline-lm , .block-share, .witness-cta-wrapper, .block-share {
+        display: none;
+        visibility: hidden;
```
OliverJAsh: You shouldn't need both? `display: none` removes the node from the render tree entirely.

NataliaLKB: Force of habit. Thanks @OliverJAsh :smile:

Issue #11971, opened 2016-02-15
AMP - Cartoons do not render in recommended content


michaelwmcnamara:
In the recommended content containers at the bottom of the amp pages, cartoon pages do not render.
example:

bottom container of recommendations on page: https://amp.theguardian.com/politics/2016/feb/15/lords-overruled-recording-laws-vellum-goat-calf-skin

![image](https://cloud.githubusercontent.com/assets/7239344/13048043/117b731e-d3db-11e5-8192-7c276cca7a5f.png)

NataliaLKB:
This will be fixed with @TBonnin 's change. Related to #11972.

Mind you, we should add a fix to say if each content object is empty, don't show. this will mean a blank space will show instead. What do you think Thomas?

TBonnin:
The problem seems to be the container endpoint returning empty items. We should probably fix this

TBonnin:
This can't be reproduced at the moment and unless Cartoons become the first container of a Section front or one of the first 4 containers of a network front it won't be visible on an amp page.
(See https://github.com/guardian/frontend/pull/12000 to see the containers we show on amp pages)
Closing for now.

PR #12000, opened 2016-02-17
Show correct containers below amp articles

TBonnin:
Layout:

```
| ...                            |
| Article content                |
| ...                            |
|--------------------------------|
| Related                        |
|--------------------------------|
| Outbrain Widget                |
|--------------------------------|
| First container either from:   |
|    - Section front             |
|    or                          |
|    - Network front             |
|--------------------------------|
| Popular                        |
|--------------------------------|
| Container from Network front   |
|--------------------------------|
| Container from Network front   |
|--------------------------------|
| Container from Network front   |
|--------------------------------|
| ...                            |
```


common/app/views/fragments/amp/onwardTemplateAmp.scala.html
```diff
@@ -10,49 +10,51 @@
...
-                                        {{/isComment}}
-                                    </span>
-                                    {{headline}}
-                                </h2>
-                                {{#isComment}}
-                                <div class="fc-item__byline">
-                                    {{byline}}
-                                </div>
-                                {{/isComment}}
+                    {{#headline}} @* Don't show if headline is empty*@
```
TBonnin: diff is confusing because of intendation but the change is simple:
- Check if headline is non empty before to display a row

dev-build/conf/routes
```diff
@@ -327,10 +327,13 @@ GET            /commercial/magento/books
...
 GET            /series-mf2/*path.json                                                                                            controllers.SeriesController.renderMf2SeriesStories(path)
 
+GET            /container/count/:count/offset/:offset/mf2.json                                                                   controllers.FaciaController.renderSomeFrontContainersMf2(count: Int, offset: Int, section = "", edition = "")
+GET            /container/count/:count/offset/:offset/section/:section/mf2.json                                                  controllers.FaciaController.renderSomeFrontContainersMf2(count: Int, offset: Int, section, edition = "")
+GET            /container/count/:count/offset/:offset/edition/:edition/mf2.json                                                  controllers.FaciaController.renderSomeFrontContainersMf2(count: Int, offset: Int, section = "", edition)
+GET            /container/count/:count/offset/:offset/section/:section/edition/:edition/mf2.json                                 controllers.FaciaController.renderSomeFrontContainersMf2(count: Int, offset: Int, section, edition)
+
```
TBonnin: I went for 4 different routes. That is not the most scalable but it make the API contract clearer. (ie: No need to check the implementation to know what value to pass for the optional parameter section and/or edition)

facia/app/controllers/FaciaController.scala
```diff
@@ -25,6 +25,14 @@ trait FaciaController extends Controller with Logging with ExecutionContexts wit
...
   val frontJsonFapi: FrontJsonFapi
 
+  private def getEditionFromString(edition: String) = {
+    val editionToFilterBy = edition match {
+      case "international" => "int"
+      case _ => edition
+    }
+    Edition.all.find(_.id.toLowerCase() == editionToFilterBy).getOrElse(Edition.all.head)
+  }
+
```
TBonnin: Just exposing this function so it can be reused both in `renderSomeFrontContainers` and `renderSomeFrontContainersMf2`

TBonnin:
@NataliaLKB do you mind reviewing this PR? Thx

cc @stephanfowler

dev-build/conf/routes
```diff
@@ -327,10 +327,13 @@ GET            /commercial/magento/books
...
 GET            /related-mf2/*path.json                                                                                           controllers.RelatedController.renderMf2(path)
-GET            /container-mf2/:num/:offset/:section.json                                                                         controllers.FaciaController.frontContainersMf2EditionRedirect(num, offset, section)
-GET            /container-mf2/:num/:offset/:section/:edition.json                                                                controllers.FaciaController.renderSomeFrontContainersMf2(num, offset, section, edition)
 GET            /series-mf2/*path.json                                                                                            controllers.SeriesController.renderMf2SeriesStories(path)
 
+GET            /container/count/:count/offset/:offset/mf2.json                                                                   controllers.FaciaController.renderSomeFrontContainersMf2(count: Int, offset: Int, section = "", edition = "")
```
NataliaLKB: Two comments here:

1. Have been thinking and I don't think we need the `.json` here. It adds complexity while editionalising in fastly. 

2. So we need this endpoint to serve on amp.theguardian.com. Right now this wont based on [this in platform](https://github.com/guardian/platform/blob/master/provisioning/puppet/modules/frontend/templates/etc/nginx/router.conf.erb#L809).

dev-build/conf/routes
NataliaLKB: I do agree - no great approach to solve this problem. 

But this does seem to be as good as can be given the requirements. Thank you!

Lets just monitor this for now.

facia/app/controllers/FaciaController.scala
NataliaLKB: This may not be needed, as we get the edition from the request anyway. We should test it without and see whether it works. If it does, we can remove this completely.

facia/test/FaciaControllerTest.scala
```diff
@@ -134,4 +135,39 @@ import org.scalatest._
     val future = test.faciaController.renderFront("inline-embeds")(request)
     contentAsString(future) should include ("""<div class="keep-it-in-the-ground__wrapper""")
   }
+
+  it should "render correct amount of fronts in mf2 format (no section or edition provided)" in {
```
NataliaLKB: Yay Tests! :+1: :white_check_mark:

NataliaLKB:
This PR also fixes https://github.com/guardian/frontend/issues/11971

NataliaLKB:
This is great. Other than the couple comments I have, which we should discuss, looks really good.

facia/app/controllers/FaciaController.scala
TBonnin: I am happy to reevaluate passing the edition as a parameter once we can reliably get it from the request.

dev-build/conf/routes
TBonnin: 1. I thought having the json extension was an "unwritten" rules we were trying to follow as mush as possible
2. I created a PR to change the router regex https://github.com/guardian/platform/pull/920

NataliaLKB:
:+1: :ship:

PR #12001, opened 2016-02-17
Updating krux tag to Guardian Amp

NataliaLKB:
We have created a new Guardian Amp krux tag. This PR uses that for amp, instead of Nextgen one. 

@stephanfowler @blackyjnz


stephanfowler:
:+1:

PR #12002, opened 2016-02-17
Fix styling of dropdowns on liveblogs

OliverJAsh:
The dropdown buttons currently leak off the side of the page.
# Before

image
# After

image

@jamesgorrie I realise we also use these dropdowns in football somewhere, so that will also need testing, but I can't find a page that demonstrates this. Do you know where I can look?

To do:
- [x] Test football http://www.theguardian.com/football/2016/feb/07/chelsea-manchester-united-premier-league-match-report

### DIFF
--- a/static/src/stylesheets/module/content/_live-blog.scss
+++ b/static/src/stylesheets/module/content/_live-blog.scss
@@ -478,12 +478,11 @@ $block-padding-right: $gs-gutter;
 .dropdown--live-feed {
     border-top: 0;
 
+    $content-gutter: $gs-gutter/2;
+
     @include mq($until: desktop) {
-        margin-right: $gs-gutter*-1;
-        margin-left: $gs-gutter*-1;
-        padding-right: $gs-gutter;
-        padding-left: $gs-gutter;
-        padding-top: $gs-baseline*4;
+        margin-right: ($content-gutter)*-1;
+        margin-left: ($content-gutter)*-1;
     }
 
     .dropdown__button {
@@ -491,10 +490,6 @@ $block-padding-right: $gs-gutter;
             background-color: #ffffff;
             padding-right: $gs-gutter;
             padding-left: $gs-gutter;
-            position: absolute;
-            top: 0;
-            right: 0;
-            left: 0;
         }
 
         @include mq(desktop) {
@@ -522,7 +517,7 @@ $block-padding-right: $gs-gutter;
 
         @include mq($until: desktop) {
             background-color: colour(neutral-8);
-            padding: $gs-baseline 0;
+            padding: $gs-baseline $content-gutter;
         }
 
         @include mq(desktop) {
--- a/static/src/stylesheets/module/content/_live-blog.scss
+++ b/static/src/stylesheets/module/content/_live-blog.scss
@@ -478,11 +478,11 @@ $block-padding-right: $gs-gutter;
 .dropdown--live-feed {
     border-top: 0;
 
-    $content-gutter: $gs-gutter/2;
+    $content-gutter: $gs-gutter / 2;
 
     @include mq($until: desktop) {
-        margin-right: ($content-gutter)*-1;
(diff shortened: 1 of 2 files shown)


static/src/stylesheets/module/content/_live-blog.scss
```diff
@@ -478,12 +478,11 @@ $block-padding-right: $gs-gutter;
...
 
+    $content-gutter: $gs-gutter/2;
+
     @include mq($until: desktop) {
-        margin-right: $gs-gutter*-1;
-        margin-left: $gs-gutter*-1;
-        padding-right: $gs-gutter;
-        padding-left: $gs-gutter;
-        padding-top: $gs-baseline*4;
+        margin-right: ($content-gutter)*-1;
```
sndrs: slightly unsure about abstracting `$gs-gutter` away? wonder if it makes it harder to skim?

i think it's also useful to add white space around operators in sass, so you can see immediately that it's not a name, it's an operation

OliverJAsh: I've used this variable further down as well. If someone changes it, they won't accidentally forget to change it there too…

Will add some white space

sndrs: it's not certain these two things will always be laid out with the same relationship, though, is it? 

by tying them both to a new variable, changing one necessarily changes the other - isn't that the kind of side-effect that we want to remove from our current scss?

OliverJAsh: In this case I am negating the page margin on the parent element and setting it again as padding on a descendant. In the future this may be refactored so that is no longer the case, in which case the variable can be removed. For the time being though, I think it's valid to use a variable here.

I don't see how the usage of variables apply differently here to how they would in any other code: variables are useful when you need to share a value.

> isn't that the kind of side-effect that we want to remove from our current scss?

I wasn't in the CSS guild meeting. Can you explain?

rich-nguyen:
@OliverJAsh here's one: http://www.theguardian.com/football/2016/feb/07/chelsea-manchester-united-premier-league-match-report

static/src/stylesheets/module/content/_live-blog.scss
```diff
@@ -478,12 +478,11 @@ $block-padding-right: $gs-gutter;
 .dropdown--live-feed {
     border-top: 0;
 
+    $content-gutter: $gs-gutter / 2;
```
sndrs: I really want to convince you that abstracting this just increases entropy and adds congitive overhead! one more go:

someone wants to update to an attribute that uses this variable. they don't recognise it, so before changing the spacing, they have to first find its definition (40 lines up the page). they then have to find all subsequent uses of it, and make sure the change they want to make doesn't break anything else using it. if it does, they probably have to undo this. 

without abstracting it, it's a simple one line edit you can be reasonably confident has no unexpected effects…
 
if that doesn't convince you, can we rename it so its value is clearer? `$content-gutter` is a bit generic?

jamesgorrie: Strange too as it's not a complete abstraction i.e. `($gs-gutter / 2)*-1` which makes the abstraction confusing in itself?

jamesgorrie: Ah wait, there's another use further down...

jamesgorrie:
:+1:

static/src/stylesheets/module/content/_live-blog.scss
jamesgorrie: Thinking about it - variable setting is such a small, if even, and abstraction, more of convenience. Agreed with the generic naming, which could lead to someone using it out of scope.

OliverJAsh: Interesting discussion. Variables have pros/cons, I will rename this.

PR #10395, opened 2015-09-04
Do not serve woff2 fonts for Edge browser

rich-nguyen:
This tweaks the detect library so that we exclude Edge browsers from woff2 candidacy. Fixes #9992 hopefully.

Issue #9992, opened 2015-07-31
Microsoft Edge Browser does not serve Guardian Fonts


ScottPainterGNM:
The new Microsoft Browser being shipped with Windows 10 is currently not displaying the Guardian fonts on the site, the version of IE being shipped with it does however do so.

OliverJAsh:
;o

ScottPainterGNM:
Microsofttttt!!!! *shakes fist*

paperboyo:
:-( Told them half a year ago... I'm surprised only IE11 got fixed... 

http://answers.microsoft.com/en-us/insider/forum/insider_internet-insider_ie/fonts-embedded-as-data-uri-do-not-display-in-ie11/a6fa26a8-2e83-4347-a192-94c80973d63d?auth=1

OliverJAsh:
Is it just our fonts or all web fonts? If it is just ours, what's special about our fonts?

paperboyo:
When I first encountered it, both IE11 and Spartan (later Edge) wouldn't display any webfonts embedded as data URIs (eg. https://robert.accettura.com/wp-content/uploads/2009/07/embedded-font-face.html). I have no way of checking it at the moment...

ScottPainterGNM:
We have a Windows 10 box built next to me if you want to have a play, weird to only fix IE11 though

paperboyo:
Unfortunately, I'm on holidays currently playing in http://www.polin.pl/en (great!):

![mk2](https://cloud.githubusercontent.com/assets/6032869/9035224/4af87d0c-39d6-11e5-9814-e6c8fde234c1.jpg)

Back in two weeks... :sunglasses:

ScottPainterGNM:
@sndrs This was the GitHub issue raised for Edge fonts

paperboyo:
Unfortunately, the fonts are not working in Edge. See https://github.com/guardian/frontend/pull/10395#issuecomment-153706458. Reopen?

OliverJAsh:
This is still an issue? Raised by @theefer

![image](https://cloud.githubusercontent.com/assets/921609/13078725/a08bcb84-d4b8-11e5-976b-2367ed80a607.png)

paperboyo:
> This is still an issue?

Yep. Many machines, many Edges. If our Win10 testing is on RTM - easy to see. Thanks for reopening!

rich-nguyen:
Yeah, let's sniff the sniffing, and see what's sniffing.

PR #12004, opened 2016-02-17
Exclude edge browser from woff2 selection

rich-nguyen:
It looks like there has been some UA capitalisation since my last attempt to exclude Edge browser from woff2, #10395. This should address a regression of issue #9992, which was fixed for a short period of time.

I've checked on browserstack, and this seems to work. Prod, however is another matter.


sndrs:
:+1:

PR #12007, opened 2016-02-17
Naughty MPU overlaps "Most popular"

regiskuckaertz:
When DFP returns a 300x600 when a 300x250 is expected in the right-hand column, there is a possibility that the MPU will overlap the "Most popular" widget when it gets sticky.

A "possibility" is quite an understatement, since I have only been able to reproduce the problem once and never since. But I noticed the height of the component was cached—my bad—and figured it could be the problem, so I changed that.

I also took it upon myself to improve the logic inside `sticky.js` which was not very readable and ended up  doing some cleanup all the way to the top: `Sticky` → `stickyMpu` → `articleBodyAdverts` → `Dfp`.


static/src/javascripts/projects/common/modules/commercial/ads/sticky-mpu.js
```diff
@@ -1,43 +1,36 @@
...
 ) {
 
-    var mpuHeight = 275,
-        StickyMpu = function ($adSlot, options) {
-            this.$adSlot = $adSlot;
-            this.opts    = defaults(options || {}, {
-                top: 0
-            });
-        };
+    var MPU_HEIGHT = 275;
```
jamesgorrie: For my info - is this a convention?

jamesgorrie: For constants? Or should we just be avoiding mutation?

static/src/javascripts/projects/common/modules/commercial/article-aside-adverts.js
```diff
@@ -3,66 +3,66 @@ define([
...
         // are article aside ads disabled, or secondary column hidden?
         if (!commercialFeatures.articleAsideAdverts || colIsHidden) {
             return false;
         }
 
-        $mainCol = config.page.contentType === 'Article' ? $('.js-content-main-column') : false;
-        $componentsContainer = $('.js-components-container', '.js-secondary-column');
-        $adSlotContainer = $(opts.adSlotContainerSelector);
+        $mainCol = $(mainColumnSelector);
+        $componentsContainer = $(componentsContainerSelector, $col[0]);
```
jamesgorrie: Will `'.js-secondary-column'` always be there?

static/src/javascripts/projects/common/modules/commercial/ads/sticky-mpu.js
regiskuckaertz: Just a convention from back when I used to write C :blush: 

I recently switched back to it because we're not far from using `const` and `let` instead of `var` and this way it's going to be a little easier to discriminate between the two.

static/src/javascripts/projects/common/modules/commercial/article-aside-adverts.js
regiskuckaertz: good catch, thanks!

regiskuckaertz: Actually the guard just above (`colIsHidden`) will prevent `$col[0]` from failing. This is arguably not very elegant.

static/src/javascripts/projects/common/modules/commercial/ads/sticky-mpu.js
jamesgorrie: Not really I would argue as we use [`const` like variables](https://github.com/guardian/frontend/pull/12007/files#diff-482df1b51d47e064a1c5e155b31ef0ecR25) all over the shop - and this denotes something is different (when I first saw it I thought it was some for of `ENUM`). Just my :dollar:'s worth.

jamesgorrie:
Loks good :+1:

static/src/javascripts/projects/common/modules/commercial/ads/sticky-mpu.js
regiskuckaertz: oh, you're right, I should have been more consistent. Let's use lowercase then.

PR #12009, opened 2016-02-17
Move amp related regexes up in facia routes

TBonnin:
...in order to make sure amp mf2 requests are not wrongfully
catched by the '/container/:id.json' regex


NataliaLKB:
:+1: Once tests pass

PR #12011, opened 2016-02-18
Adblock detect tidy

sndrs:
- fix `guardian.adBlockers` race condition (sporadic `onDetect` is undefined error)
- coerce FF adblock detection into a boolean

### DIFF
--- a/common/app/templates/inlineJS/blocking/config.scala.js
+++ b/common/app/templates/inlineJS/blocking/config.scala.js
@@ -20,6 +20,7 @@ window.guardian = {
     css: {
         loaded: false
     },
+    adBlockers: {},
     config: @JavaScript(templates.js.javaScriptConfig(page).body)
 };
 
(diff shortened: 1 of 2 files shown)


jbreckmckye:
I thought the onDetect call was wrapped in a try-catch to handle cases when it didn't exist. Or was that just for the purpose of robusting the inline scripts against failures in a foreign function?

sndrs:
i think it's raised when the JS app tries to attach the `resolve` function to `guardian.adBlockers.onDetect`, but `guardian.adBlockers` has not yet been created (e.g. if they app loads v fast from the cache). 

@OliverJAsh had this

pasted image at 2016_02_17 06_40 pm

and I've seen it sporadically too

jbreckmckye:
I see, it's an exception from the 'other side' of the race condition.

:+1:

sndrs:
🌑🌘🌗🌖🌕💥

PR #12015, opened 2016-02-18
Adding first two charts, updated via AWS

scottp719:
This updates the first version of the Quality Dashboard page in admin to render the first two charts using data stored in the omniture-dashboard S3 bucket.

Thanks to @dominickendrick for the help with the Scala!


gtrufitt:
:+1:

PR #12017, opened 2016-02-18
Refactor CSS inliner

desbo:
This should be more robust now and work on a much bigger range of stylesheets (including the [Zurb one](http://foundation.zurb.com/emails) I'm using for the new design). I've added some tests too.

It also includes an upgrade to Jsoup, the older version had a bug that escaped quotes in CSS attribute selectors.

@fcarr @rich-nguyen


rich-nguyen:
:+1: looks.good.

PR #12019, opened 2016-02-18
Simplify commercial features module

jbreckmckye:
I wasn't very happy with the existing commercial features / commercial feature policies modules - the 'policy' idea clearly wasn't providing much value and the abstraction was clunky and opaque. This pull request (hopes?) to make things much easier to read - courtesy of some generous advice from @regiskuckaertz.

**What has changed?**
- the behaviour is now all together in a single module
- it's not a lot easier to see what flags influence an individual switch when debugging why a feature isn't active
- we now decorate the config object with a debug map that makes debugging commercial features much easier
- if the module throws an error, this will now be contained and logged properly in Sentry. This is valuable because any errors in this module could kill the majority of advertising features.

The module itself still returns the same properties, with the exception of `features.checkWeCanShowMembershipMessages`, which is now `features.async.membershipMessages`. @sndrs - I know you preferred the former style because it allows us to write a much more naturalistic `.then` chain, but it seemed out of kilter with an interface whose keys were named simply after the features they turned on and off. I am happy to change this back, though.

This pull request is a proposal rather than a fait accompli - I am very happy to discuss its design and consider other options.

@uplne @Calanthe @OliverJAsh


static/src/javascripts/projects/common/modules/commercial/commercial-features.js
```diff
@@ -1,38 +1,123 @@
...
+        var externalAdvertising =
+            !isSSL &&
+            !noadsUrl &&
+            !userPrefs.isOff('adverts');
+
+        var sensitiveContent =
+            config.page.shouldHideAdverts ||
+            config.page.section === 'childrens-books-site';
+
+        var minuteArticle = config.page.isMinuteArticle;
```
regiskuckaertz: This is nitpicking but I would name it `isXXX`, just like the next ones.

static/src/javascripts/projects/common/modules/commercial/commercial-features.js
```diff
@@ -1,38 +1,123 @@
...
-                    // This gives any policy the ability to veto a commercial feature
-                    activeSwitches[switchKey] = switchValue;
-                }
-            });
-        });
-
-        return activeSwitches;
+    try {
+        config.commercial = config.commercial || {};
+        return config.commercial.featuresDebug = new CommercialFeatures();
```
regiskuckaertz: Why not simply `config.commercial.features`?

regiskuckaertz:
Great work! It's going to be so much easier now. :+1:

OliverJAsh:
I like it! :-) :+1:

static/src/javascripts/projects/common/modules/commercial/commercial-features.js
jbreckmckye: Good spot.

static/src/javascripts/projects/common/modules/commercial/commercial-features.js
jbreckmckye: I'd like to hint that production code shouldn't hang off this object - it should `require` in the service instead, which makes sure all the keys are set.

sndrs:
> The module itself still returns the same properties, with the exception of `features.checkWeCanShowMembershipMessages`, which is now `features.async.membershipMessages`.

this has proven a far trickier issue than I ever would have imagined. I really have no good ideas beyond a mild preference for something that reads naturally and instructively. not sure I'm totally sold, personally, on prefixing with `async` as a general solution, but as I say, I have no better ideas, and it conveys the character, so why not :rocket:

jbreckmckye:
@sndrs I think this worth discussing; I've always found promises tricky to name.

When you read code that uses them, it is pleasant to see them written as a verb -

    doSomething.then(x = > y);

But verbs only make sense as functions. This makes sense -

    doSomething().then(x => y);

But this seems weird -

    const doSomething = somethingDoer();
    doSomething.then(x => y);

You can name things after 'actions' -

    const somethingAction = doSomething();

But then we're back to the awkwardness of following a noun with a `then` -

   somethingAction.then(x => y)

You can also use a 'promise' suffix:

    const doSomethingPromise = doSomething();
    doSomethingPromise.then(x = > y);

But you haven't really solved the awkwardness of chaining 'then' with a noun, and you now have a kind of Hungarian notation to boot, too. This is `not a solution`.

We could evade the issue in this instance by turning the features field into a method -

    commercialFeatures.getMembershipMessages().then(x => y)

How does this look? It might be nice to somehow signal this function is async, though.

OliverJAsh:
@jbreckmckye Nice breakdown of the various techniques :-). I would add an option in there about naming promises by their value. After all, that's what they are — async values.

E.g. if `doSomething` returned a user, you could call the returned value `userPromise`. Looking at the promise, I don't need to care where it came from, all I need to know is its value. I don't feel any awkwardness about falling a noun with `.then`, because the promise is explicit by its name. This avoids problems with shadowing as well:

``` js
user.then(user => {}) // shadowing!
userPromise.then(user => {})
```

This appears to be the convention with futures and options in Scala (`futureUser`, `maybeUser`).

PR #12021, opened 2016-02-18
Removing files for amp that are not needed anymore

NataliaLKB:
Replaced with: https://github.com/guardian/frontend/blob/master/facia/public/amp_remote.html

cc @kenlim


rich-nguyen:
:+1:

PR #12027, opened 2016-02-18
Enable re-pressing from archive-originals sources

JustinPinner:
@jennysivapalan - hopefully this isn't too cumbersome to track the changes, and the UI isn't very nice but it'll get us going.

Thanks,
Justin.

:recycle: :page_with_curl: :boom:


jennysivapalan:
Looks fine to me, I didn't quite follow all the changes in `R2PagePressJob` but we can def test this out so it's :+1: from me.

PR #12030, opened 2016-02-19
Add caching duration to renderMf2Series result

TBonnin:


stephanfowler:
:+1: seem reasonable. No need for speed here.

PR #12022, opened 2016-02-18
Set amp-list credentials mode to 'include' so cookies are sent with the xhr request

TBonnin:
"Editionalization" redirect is based on cookie (X-GU-Edition). See
fastly vcl config

PR #12031, opened 2016-02-19
Revert "Set amp-list credentials mode to 'include' so cookies are sent with the xhr request"

NataliaLKB:
Reverts guardian/frontend#12022

cc @TBonnin


TBonnin:
:sheep:

PR #12032, opened 2016-02-19
Try out a lighter adblocking technique

sndrs:
this basically uses the same method as the current one, but only ads the element once, and on the main viewport canvas.

before, it was actually adding the element 5 times:

screen shot 2016-02-19 at 13 09 22

and `9999px` off screen.

this may have no effect, but i'm hopeful it may help iPad crashes

cc @jbreckmckye @regiskuckaertz

### DIFF
--- a/static/src/javascripts/projects/common/utils/detect.js
+++ b/static/src/javascripts/projects/common/utils/detect.js
@@ -8,6 +8,7 @@
 define([
     'common/utils/mediator',
     'common/utils/$',
+    'common/modules/user-prefs',
     'lodash/arrays/last',
     'lodash/arrays/findIndex',
     'lodash/objects/defaults',
@@ -19,10 +20,12 @@ define([
     'lodash/collections/pluck',
     'lodash/arrays/initial',
     'lodash/arrays/rest',
+    'lodash/functions/memoize',
     'Promise'
 ], function (
     mediator,
     $,
+    userPrefs,
     last,
     findIndex,
     defaults,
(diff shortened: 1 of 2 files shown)


static/src/javascripts/projects/common/utils/detect.js
```diff
@@ -405,38 +409,58 @@ define([
...
-    function adblockInUseSync() {
-        if (!detect.cachedAdblockInUse) {
-            var sacrificialAd = createSacrificialAd(),
-                contentBlocked = isHidden(sacrificialAd);
-            sacrificialAd.remove();
-            detect.cachedAdblockInUse = contentBlocked;
-            return contentBlocked;
-        }
+    var adblockInUseSync, getFirefoxAdblockPlusInstalledSync, createSacrificialAd;
+    if (userPrefs.get('lighter-adblock')) {
```
jbreckmckye: I actually probably wouldn't bother putting this in a separate branch - I'd just replace the original function entirely. This also make it easy to check the new approach remains accurate in the wild - we just have to look at Ophan and check adblockers are still being detected.

sndrs: ah ok yeah will do then

sndrs:
updated to remove the pref

regiskuckaertz:
Nice :mountain_cableway:

sndrs:
is that a :+1:? :)

jbreckmckye:
It's a :palm_tree: from me.

PR #12033, opened 2016-02-19
Move most pop MPU back in the normal flow

regiskuckaertz:
Chromium has a bug with the balancing algorithm where it can't compute the height of the column container. Which gives:
picture 3

To circumvent the problem, we move the MPU back into normal document flow and centre it horizontally below the headlines:
picture 2

cc @stephanfowler @jbreckmckye


jbreckmckye:
Do you know if there are any tickets for this issue on the Chromium project? Perhaps they give an indication as to when they can fix this? How did you track down the cause?

regiskuckaertz:
There is none, I'm creating one now. I should create two in fact, one for `column-fill` and one for `calc`. There is a comment on caniuse.com but no source is provided, nevertheless from a few tests it is fairly obvious the new multicol implementation they released a while ago is incomplete.

regiskuckaertz:
Here is what the spec says about `column-fill`:

> In continuous media, this property will only be consulted if the length of columns has been constrained.

`min-height` and `max-height` set constraints on the length of columns, so Chromium should behave like Safari.

static/src/javascripts/projects/common/modules/onward/popular.js
```diff
@@ -41,16 +42,10 @@ define([
         this.fetch(qwery('.js-popular-trails'), 'html');
     };
 
-    MostPopular.prototype.mobileMaximumSlotsReached = function () {
-        return (detect.getBreakpoint() === 'mobile' && $('.ad-slot--inline').length > 1);
-    };
-
     MostPopular.prototype.prerender = function () {
-        if (commercialFeatures.popularContentMPU && !this.mobileMaximumSlotsReached()) {
```
jbreckmckye: We no longer care about the ads-per-page count on mobile?

regiskuckaertz: A while ago I removed the limitation, but now that you mention it, that was for inline MPUs in article, not fronts. I'm putting this back. Thanks!

static/src/stylesheets/module/facia/_slices.scss
```diff
@@ -239,6 +210,29 @@ Hence why a greater depth of selector specificity is needed.
     }
 }
 
+.fc-slice__popular-mpu {
+    .ad-slot {
+        min-height: 274px;
```
jbreckmckye: Where does 274px come from? I notice _items.scss has the following entry -

    .fc-slice__item--mpu-candidate {
        @include mq(tablet) {
            min-height: 25p + $mpu-ad-label-height;
        }
    }

I don't know if perhaps something could be rationalized between them...

regiskuckaertz: That's a good point. Same for the `286px` a few lines below. I'll try and see what we can do here.

jbreckmckye: Well, it's all existing code, so it's okay if we don't have the chance to resolve it in this PR - but any cleanup is always an improvement!

regiskuckaertz:
@jbreckmckye turns out there are vars for a standard MPU size. We should perhaps think about creating a map with all our sizes—both standard and Guardian-specific.

jbreckmckye:
:+1:

PR #12034, opened 2016-02-19
Refactor full-height into state and renderer funcs

gtrufitt:
This is a refactor and bugfix for the 100vh fixed height images to use the separation of state and render pattern. It also fixed a bug with reseting the height on orientationchange as this doesn't fire resize event.

cc/ @OliverJAsh


OliverJAsh:
:+1: !

PR #12035, opened 2016-02-19
Use cAPI blocks instead of flexible-content-body-parser

desbo:
We're currently using https://github.com/guardian/flexible-content-body-parser to parse CAPI responses into blocks, but CAPI gives us the blocks directly so there's no need for the parser. This removes the only use of it.

@rich-nguyen


rich-nguyen:
:+1:

Nice, negative line count :tada:

PR #12036, opened 2016-02-19
Deploys radiator: add polyfills, use float instead of flexbox

OliverJAsh:
This is because the browser on the Pi is very old…


rich-nguyen:
:+1:

Issue #11967, opened 2016-02-13
Annihilate faux bold in footer


paperboyo:
Hello,

Let the battle begin! Currently, for unknown reason, we use Guardian Sans (even though the rest of footer uses Guardian TextSans) of which we just serve a Regular weight. But the links in the footer use it emboldened. This is ugly. We have two choices:

- switch off bold:

![unbolden](https://cloud.githubusercontent.com/assets/6032869/13030029/a57f1428-d294-11e5-8652-621718cfef06.gif)

- switch to TextSans:

![sans to textsans](https://cloud.githubusercontent.com/assets/6032869/13030035/c3923080-d294-11e5-8ca8-0ac23ac7c2fe.gif)

Is there an easy way to find all occurrences of faux font styling? I know of three more, but solving these won’t be as easy, although possible and will be done :napoleon_hat:

Regards
Mateusz

crifmulholland:
@superfrank - what do you think?

sndrs:
TextSans! Annihilate the bold!

superfrank:
All of the footer should be in: "Guardian Text Sans Web". With "Sign up to our daily email" in bold and all links in regular.

This just looks like an oversight (probably by myself!)

superfrank:
Guardian Sans should only be used for advertorial content I believe. Do you know if we're accidentally doing this elsewhere?

gtrufitt:
This was my bad I believe.. I've updated them to GuardianTextSansWeb regular and now they look all lovely like.

PR #12037, opened 2016-02-19
Fix for #11967 - Footer font

gtrufitt:
# Footer Font
## What does this change?

Bugfix for the footer font issue raised by @paperboyo in #11967 
### Before

image
### After

image
## Notify

@superfrank


regiskuckaertz:
:+1:

PR #12039, opened 2016-02-19
Unhide paid content "About" popup when sticky

regiskuckaertz:
In paid content, when the user scrolls so that the band becomes sticky, the popup will be hidden because of a `position: overflow`:
picture 1

Also, when deciding whether an element becomes sticky or not, the script takes into account the height of the band if it is present—this behaviour should be disabled if the band is itself the element.

### DIFF
--- a/static/src/javascripts/projects/common/modules/ui/sticky.js
+++ b/static/src/javascripts/projects/common/modules/ui/sticky.js
@@ -15,11 +15,11 @@ define([
     var isNewCommercialContent = config.switches.newCommercialContent && config.page.isAdvertisementFeature;
     var paidforBandHeight;
 
-    function initPFBand() {
+    function initPFBand(element) {
         paidforBandHeight = 0;
         if (isNewCommercialContent) {
             var paidforBand = document.querySelector('.paidfor-band');
-            if (paidforBand) {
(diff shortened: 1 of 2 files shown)


static/src/javascripts/projects/common/modules/ui/sticky.js
```diff
@@ -15,11 +15,11 @@ define([
     var isNewCommercialContent = config.switches.newCommercialContent && config.page.isAdvertisementFeature;
     var paidforBandHeight;
 
-    function initPFBand() {
+    function initPFBand(element) {
```
uplne: Why not to call it initPaidForBand. Someone who will touch this in half a year will save 2.5 seconds by not wondering what PF means :smile:

uplne:
:+1:

static/src/javascripts/projects/common/modules/ui/sticky.js
jbreckmckye: I have to agree - I'm not a huge fan of abbreviations.

regiskuckaertz: You're right guys, thanks for the tip!

PR #12042, opened 2016-02-19
Short circuit commercial code in IE8

jbreckmckye:
# Stop running ads in legacy browsers

Commercial code already doesn't run under Internet Explorer 8 due to a bug (apparently with `pre-fetch-modules.js` - this could bear a closer look itself). We don't really care about this - IE8 numbers are very low, and we don't want to clog their (already rather ponderous) experience with processing-intensive ads. This pull request makes the matter formal.
## What does this change?

Internet Explorer 8 no longer tries to run any commercial code. This should change no actual behaviour, because - as stated above - none of it appears to have actually run for quite some time.

To do this, I've split the existing `window.guardian.isModernBrowser` - which checks if we're enhanced and whether we pass a 'Cut the mustard' check - into `isModernBrowser` and `isEnhanced` flags. Our commercial IE8 check uses `isModernBrowser`. All the existing checks to the property now use `isEnhanced`.

There is one exception, though, in the form of Ophan's tracker-js. See below in the RFC.

We can start writing commercial code in naive ES5 by default, and we can potentially pull out unnecessary lodash helpers like `forEach` and `map` which are implemented natively. This will let us write more concise code going forward and whittle away IE shims from our dependency tree.

Nope.
## Request for comment

I've made a change to a property that Ophan [tracks](https://github.com/guardian/ophan/blame/master/tracker-js/assets/coffee/ophan/core.coffee) - `isModernBrowser`. It was originally a cut the mustard check, but was then changed to check if the scripts were in 'enhanced' mode, too. I've now changed the property's meaning back. Which value would we prefer Ophan to track?


static/src/javascripts/boot.js
```diff
@@ -70,9 +70,15 @@ define([
 
 
 
-    var shouldRunEnhance = guardian.isModernBrowser;
+    var shouldRunEnhance = guardian.isEnhanced;
+    var shouldRunCommercial = guardian.isEnhanced;
```
regiskuckaertz: You wrote "Our commercial IE8 check uses `isModernBrowser`." Shouldn't this rather be `guardian.isModernBrowser`?

regiskuckaertz:
the name of that branch :laughing:

static/src/javascripts/boot.js
jbreckmckye: :eyes: Er... that was a test! You passed!

(You are absolutely right!)

regiskuckaertz:
:+1: nice nice nice! pending decision on which property is right for Ophan.

jbreckmckye:
For the sake of conservatism, I'm going to create a PR such that Ophan will continue tracking the old property, but will ask that team if there's any interest in tracking the separate cut-the-mustard check too

common/app/templates/inlineJS/blocking/config.scala.js
```diff
@@ -4,19 +4,21 @@
...
-            && "sessionStorage" in window
-            && "bind" in Function
-            && (
-                ("XMLHttpRequest" in window && "withCredentials" in new XMLHttpRequest())
-                    || "XDomainRequest" in window
-            )
-    ),
+    isModernBrowser : isModernBrowser,
+    isEnhanced:
+        window.shouldEnhance && isModernBrowser,
```
OliverJAsh: Is `window.shouldEnhance` used anywhere else? Feels like this and `window.isEnhanced` are the same thing. Can `shouldEnhance` be inlined here?

jbreckmckye: `shouldEnhance` is defined over in `shouldEnhance.scala.js` and used once, here. It's moderately substantial, so inlining could make the file quite large. I'm happy to change things around, but that does increase the scope of this pull request.

OliverJAsh: I understand. I think it might be a good idea in order to avoid people consuming both when they should really only be consuming this? Your call though :-)

jbreckmckye: I see what you mean. I guess it's pretty easy to just make `shouldEnhance.scala.js` decorate the isEnhanced field. This would mean we wouldn't pollute the namespace with temporary values, and be a fairly minor (i.e. low risk) change overall.

jbreckmckye: I've decided to nix the above change because it doesn't look like I'll have time to manually test the changes in the offline page.

Issue #12025, opened 2016-02-18
Amp - Ads should be suppressed if capi field shouldHideAdverts: "true"


michaelwmcnamara:
Ads should be suppressed if the following capi field is set:
shouldHideAdverts: "true"

example:
www.theguardian.com/uk-news/2015/may/18/hillsborough-inquest-teenage-liverpool-fan-philip-hammond

http://amp.theguardian.com/uk-news/2015/may/18/hillsborough-inquest-teenage-liverpool-fan-philip-hammond

PR #12043, opened 2016-02-19
Hide ads on amp pages if shouldHideAdverts is true

TBonnin:
## What does this change?

Hides ads on amp pages if shouldHideAdverts is set to true for this article
Fix for https://github.com/guardian/frontend/issues/12025
## Request for comment

@NataliaLKB @johnduffell

### DIFF
--- a/article/app/views/package.scala
+++ b/article/app/views/package.scala
@@ -36,6 +36,10 @@ object MainCleaner {
 object BodyCleaner {
   def apply(article: Article, html: String, amp: Boolean)(implicit request: RequestHeader) = {
     implicit val edition = Edition(request)
+
+    val shouldShowAds = !article.content.shouldHideAdverts && article.metadata.section != "childrens-books-site"
+    def ListIf[T](condition: Boolean)(value: => T): List[T] = if(condition) List(value) else Nil
+
     val cleaners = List(
       InBodyElementCleaner,
       InBodyLinkCleaner("in body link", amp),
@@ -58,14 +62,11 @@ object BodyCleaner {
       ImmersiveLinks(article.isImmersive),
       TimestampCleaner(article),
       MinuteCleaner(article)
-    )
-    val nonAmpCleaners = List(
-      VideoEmbedCleaner(article)
-    )
-    val ampOnlyCleaners = List(
-      AmpAdCleaner(edition, request.uri, article),
-      AmpEmbedCleaner(article)
-    )
-    withJsoup(BulletCleaner(html))((if (amp) ampOnlyCleaners else nonAmpCleaners) ++ cleaners :_*)
+    ) ++
+      ListIf(!amp)(VideoEmbedCleaner(article)) ++
+      ListIf(amp)(AmpEmbedCleaner(article)) ++
+      ListIf(amp && shouldShowAds)(AmpAdCleaner(edition, request.uri, article))
+
+    withJsoup(BulletCleaner(html))(cleaners :_*)
   }
 }


article/app/views/package.scala
```diff
@@ -36,7 +38,7 @@ object MainCleaner {
 object BodyCleaner {
   def apply(article: Article, html: String, amp: Boolean)(implicit request: RequestHeader) = {
     implicit val edition = Edition(request)
-    val cleaners = List(
+    val cleaners = ListBuffer(
```
TBonnin: Making this list mutable. Since it is only a local variable it shouldn't be a issue.

article/app/views/package.scala
```diff
@@ -59,13 +61,16 @@ object BodyCleaner {
...
-    withJsoup(BulletCleaner(html))((if (amp) ampOnlyCleaners else nonAmpCleaners) ++ cleaners :_*)
+
+    if (amp) {
+      cleaners += AmpEmbedCleaner(article)
+      if(!article.content.shouldHideAdverts) {
+        cleaners += AmpAdCleaner(edition, request.uri, article)
+      }
+    } else {
+      cleaners += VideoEmbedCleaner(article)
+    }
```
TBonnin: If there is a more idiomatic way of creating this list I am happy to do it.

jbreckmckye:
Can content from the children's book site appear on AMP? Those pages are often not flagged as sensitive, but cannot show advertising - ASA regulations prohibit it on anything that might be deemed childrens' content.

TBonnin:
@jbreckmckye Thanks for pointing this out. Indeed this logic happens there for the website. https://github.com/guardian/frontend/blob/f3941b51b79b63abf92fc9c53dd2f7af8201b694/static/src/javascripts/projects/common/modules/commercial/commercial-feature-policies.js#L24  

Duplicating this kind of logic for amp and non-amp will for sure end up in oversights and bugs in the future and I don't think we can always count on attentive reviewers like you to catch this kind of mistmatches.
This might be a very stupid question but why is CAPI not returning `shouldHideAdverts=false` for all `/childrens-book-site/...` content?

jbreckmckye:
@TBonnin - I don't really know how CAPI works or the taxonomy of backend content. There must presumably be some reason we haven't done this, but I'm not sure who to ask about it.

NataliaLKB:
@TBonnin @jbreckmckye maybe ask @mchv or @cb372

mchv:
@jbreckmckye let's have a chat about it when you will be available

article/app/views/package.scala
johnduffell: if you're going to have a mutable list, it's better if it's local, but the same arguments about being able to reason about your code apply regardless.  I'd recommend keeping it immutable.
Having said all that, there may be a better (but still functional) way of building up the list as a logic for whats in it becomes more complicated.

TBonnin: The problem with using an immutable list in this case is you'll have to make a bunch of copies and temporary vals, or use weird filtering. Unless I am missing something...

TBonnin:
@mchv @jbreckmckye have you chatted about this?

jbreckmckye:
@TBonnin We have. It does make sense; the property is an aspect of presentation, and that's not CAPI's concern. So any rationalization needs to happen in the rendering tier.

article/app/views/package.scala
```diff
@@ -58,14 +64,11 @@ object BodyCleaner {
...
-    )
-    val ampOnlyCleaners = List(
-      AmpAdCleaner(edition, request.uri, article),
-      AmpEmbedCleaner(article)
-    )
-    withJsoup(BulletCleaner(html))((if (amp) ampOnlyCleaners else nonAmpCleaners) ++ cleaners :_*)
+    ) ++
+      ListIf(!amp)(VideoEmbedCleaner(article)) ++
+      ListIf(amp)(AmpEmbedCleaner(article)) ++
+      ListIf(amp && shouldShowAds)(AmpAdCleaner(edition, request.uri, article))
```
TBonnin: @johnduffell @dominickendrick you might like that better :)

article/app/views/package.scala
```diff
@@ -36,6 +38,10 @@ object MainCleaner {
 object BodyCleaner {
   def apply(article: Article, html: String, amp: Boolean)(implicit request: RequestHeader) = {
     implicit val edition = Edition(request)
+
+    val shouldShowAds = !article.content.shouldHideAdverts && article.metadata.section != "childrens-books-site"
```
TBonnin: Got it @jbreckmckye 
I guess the section check should be centralised somewhere instead of having this check scattered in multiple places in scala/javascript code. However I feel this goes behind the scope of this PR and could be the objective of another pull request

JonNorman: We have [`commercial-features.js`](https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/common/modules/commercial/commercial-features.js) which does codify some of these checks (see `var sensitiveContent`), which you could plug in.

article/app/views/package.scala
johnduffell: clever! looks good :yum: thanks

johnduffell:
:+1: I'm happy anyway, improved on the previous, but feel free to deal with the duplication too

article/app/views/package.scala
```diff
@@ -8,6 +8,8 @@ import play.api.mvc.RequestHeader
 import views.support._
 import views.support.cleaner.{AmpEmbedCleaner, VideoEmbedCleaner, CmpParamCleaner, AmpAdCleaner}
 
+import scala.collection.mutable.ListBuffer
```
johnduffell: dn't forget to remove the unneeded import

TBonnin: Good catch. I will remove. Thx

article/app/views/package.scala
TBonnin: @JonNorman That's happening client side but on amp page ads are included server side during page generation.

NataliaLKB:
:+1: Thanks for adding the bit about Children's books!

PR #12046, opened 2016-02-21
Week-end refactoring: lazyload

regiskuckaertz:
I suggest that all coding patterns resembling:

```
new Something().init(); // notice there's no assignment here
```

are in fact procedures:

```
doSomething();
```

`lazyload` is one fairly unimportant example, and there are many such examples that could be rewritten in a more idiomatic way.

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/creatives/commercial-component.js
+++ b/static/src/javascripts/projects/common/modules/commercial/creatives/commercial-component.js
@@ -28,7 +28,7 @@ define([
     $,
     config,
     mediator,
-    LazyLoad,
+    lazyload,
     Tabs,
     Toggles,
     isArray,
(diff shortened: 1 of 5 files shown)


static/src/javascripts/projects/common/modules/commercial/creatives/commercial-component.js
```diff
@@ -149,7 +149,7 @@ define([
             error: function () {
                 bonzo(this.adSlot).hide();
             }.bind(this)
-        }).load();
+        });
```
OliverJAsh: :+1: because this is simpler :-) No contexts, no constructors, no prototypes — fewer language features

OliverJAsh:
Nice! :+1:

sndrs:
:+1: great

jbreckmckye:
I hate those `init` methods :+1:

regiskuckaertz:
@OliverJAsh @sndrs @jbreckmckye Thanks guys! Next one in line: https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/common/modules/commercial/dfp/breakout-iframe.js#L46

It's going to be a bit more work, but clarity and simplicity are worth the effort.

PR #12052, opened 2016-02-22
no more vpaid

jamesgorrie:
# Remove VPAID video tech
## What does this change?

Removes references to videojs vpaid as we don't actually have that tech available. 
[The Google Ads IMA plugin should cover this for us](https://github.com/googleads/videojs-ima).

We stop getting the vpaid not supported error

Na
## Request for comment

@Calanthe @rich-nguyen 
##

### DIFF
--- a/common/app/conf/switches/CommercialSwitches.scala
+++ b/common/app/conf/switches/CommercialSwitches.scala
@@ -50,15 +50,6 @@ trait CommercialSwitches {
     exposeClientSide = true
   )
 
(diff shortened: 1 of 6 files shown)


rich-nguyen:
And https://github.com/guardian/frontend/blob/master/static/public/flash/bower.json#L7

cool :+1:

jamesgorrie:
Thanks @rich-nguyen, missed my `grep`.

PR #12054, opened 2016-02-22
Extending to banner position test extra month

uplne:
We have sufficient data for this test and it seems that odds are against this amazing feature. However the user experience gain is great so I would like to wait for product owner (on holiday) to go through data one more time and possibly setup some process so it won't be lost in the abyss of the past.

### DIFF
--- a/common/app/mvt/MultiVariateTesting.scala
+++ b/common/app/mvt/MultiVariateTesting.scala
@@ -22,7 +22,7 @@ object CMTopBannerPosition extends TestDefinition(
   List(Variant1),
   "cm-top-banner-position",
   "Test viewability and revenue changes when top banner is moved below first container on fronts and removed from articles",
-  new LocalDate(2016, 2, 25)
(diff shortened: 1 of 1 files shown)


rich-nguyen:
:+1:  maybe up the traffic to, by adding another variant to the variant list? gets you some more data.

paperboyo:
> user experience gain is great

Skew the data, then. Fake it. Where do I have to click? How many times? I’ve got loads of friends, you know... This data thing will bury us all. :pouting_cat: :bar_chart: 

https://www.youtube.com/watch?v=GAXLHM-1Psk

*all views mine, whose else, right?!

uplne:
We have enough data :disappointed: It is really only :money_with_wings: talking in this case. But I am idealist so I want to push for it more.

rich-nguyen:
@uplne good stuff. For me, that's not idealism. That's just doing the right thing.

paperboyo:
> That's just doing the right thing

+1!

PR #12055, opened 2016-02-22
make multimedia URIs non-secure

jamesgorrie:
## What does this change?

A simple wedge to not use https URIs on the multimedia domain as something went wrong with the dang certificates.

We will be removing this once the certs are setup again.

Videos on that domain will start working again.

No

@rich-nguyen


rich-nguyen:
:+1:

PR #12058, opened 2016-02-23
Move deploys radiator installation to make install 

sndrs:
## What does this change?

`make reinstall` wouldn't fix a corrupt package in the deploys app because it didn't touch it, and the `shell:makeDeploysRadiator` grunt task also only installed, it didn't clean or let you reinstall. 

this connects them both

### DIFF
--- a/grunt-configs/shell.js
+++ b/grunt-configs/shell.js
@@ -38,9 +38,6 @@ module.exports = function () {
 
         makeDeploysRadiator: {
             command: [
-                'npm install',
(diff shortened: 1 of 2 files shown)


OliverJAsh:
The reason `makeDeploysRadiator` installed and compiled was because I wanted to keep it self-contained. If you wanted to continue along that path you could make `makeDeploysRadiator` also do a clean at the start. Your call, though! :+1:

PR #12062, opened 2016-02-23
Display ads - changes to style on mobile

uplne:
# Changes to styles on mobile
## What does this change?

Ad have full width on mobile and the image is taking whole height of the ad slot.

It looks good.

Nope
## Screenshots

Before:
screen shot 2016-02-23 at 15 34 58
screen shot 2016-02-23 at 15 35 04

After:
screen shot 2016-02-23 at 15 07 26
screen shot 2016-02-23 at 15 07 54


regiskuckaertz:
:+1:

PR #12063, opened 2016-02-23
Refactored breakout iframe using promises

regiskuckaertz:
## What does this change?

This fixes the "jump" that is happening in Firefox (and sometimes Safari) in the top slot, when the component is a breakout script. It is caused by the iframe taking up space until it is hidden, which happens only after its content has been added to the DOM. When the iframe is hidden first, the jump does not happen anymore.

I also cleaned up the code a little and moved all the creatives in `breakout-iframe` where they are called (unless @jbreckmckye there was a reason not to... if yes let me know, happy to revert it)
## Screenshots

picture 1

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/dfp/breakout-iframe.js
+++ b/static/src/javascripts/projects/common/modules/commercial/dfp/breakout-iframe.js
@@ -40,8 +40,8 @@ define([
             return type;
         }
 
-        $breakoutEl = $('.breakout__html', iFrameBody);
-        if ($breakoutEl.length) {
+        $breakoutEl = $('.breakout__html, .breakout__script', iFrameBody);
+        if ($breakoutEl.hasClass('breakout__html')) {
             fastdom.write(function () {
                 $iFrame.hide();
                 $breakoutEl.detach();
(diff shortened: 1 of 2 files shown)


static/src/javascripts/projects/common/modules/commercial/dfp/breakout-iframe.js
```diff
@@ -22,57 +30,45 @@ define([
...
-                                });
-                            }, 50);
-                        });
-                    }
-                    shouldRemoveIFrame = true;
+        if (($breakoutEl = $('.breakout__html', iFrameBody)).length) {
+            fastdom.write(function () {
+                $iFrame.hide();
+                $breakoutEl.detach();
+                $iFrameParent.append($breakoutEl.html());
```
jbreckmckye: Do we actually need to serialize the HTML and retokenize it? I'm pretty sure you can `.append` Bonzo collections directly.

static/src/javascripts/projects/common/modules/commercial/dfp/breakout-iframe.js
```diff
@@ -22,57 +30,45 @@ define([
...
-                        $('.ad--responsive', $iFrameParent[0]).each(function (responsiveAd) {
-                            window.setTimeout(function () {
-                                idleFastdom.write(function () {
-                                    bonzo(responsiveAd).addClass('ad--responsive--open');
-                                });
-                            }, 50);
-                        });
-                    }
-                    shouldRemoveIFrame = true;
+        if (($breakoutEl = $('.breakout__html', iFrameBody)).length) {
```
jbreckmckye: Using the value of an assignment feels a little opaque to me.

static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
```diff
@@ -45,21 +45,7 @@ define([
...
-    'common/modules/commercial/creatives/expandable-v2',
-    'common/modules/commercial/creatives/expandable-v3',
-    'common/modules/commercial/creatives/expandable-video',
-    'common/modules/commercial/creatives/expandable-video-v2',
-    'common/modules/commercial/creatives/fluid250',
-    'common/modules/commercial/creatives/fluid250GoogleAndroid',
-    'common/modules/commercial/creatives/foundation-funded-logo',
-    'common/modules/commercial/creatives/scrollable-mpu',
-    'common/modules/commercial/creatives/scrollable-mpu-v2',
-    'common/modules/commercial/creatives/template'
```
jbreckmckye: Yeah, sorry, I never noticed that these dependencies needed to be moved over.

static/src/javascripts/projects/common/modules/commercial/dfp/breakout-iframe.js
regiskuckaertz: Oh that's a good one!

static/src/javascripts/projects/common/modules/commercial/dfp/breakout-iframe.js
```diff
@@ -22,58 +30,50 @@ define([
...
-                            });
-                        } else {
-                            // evil, but we own the returning js snippet
-                            eval(breakoutContent);
+        $breakoutEl = $('.breakout__html', iFrameBody);
+        if ($breakoutEl.length) {
+            fastdom.write(function () {
+                $iFrame.hide();
+                $breakoutEl.detach();
+                $iFrameParent.append($breakoutEl.html());
```
jbreckmckye: It's a corner case, but this may not work if a creative ever comes back with more than one element classed `.breakout-html`. Bonzo's .html() method only returns the innerHTML of the first element in a Bonzo collection. Would it be safer just to append the `breakoutEl` rather than the serialized markup? I guess it's a corner case.

jbreckmckye:
:+1:

PR #12064, opened 2016-02-23
Enchilada margins at the right article position

uplne:
# Fix: Enchilada margins at the right article position
## What does this change?

Styling issue.

Looks better.
## Screenshots

Before:
screen shot 2016-02-23 at 17 16 50

After:
screen shot 2016-02-23 at 17 20 22


jamesgorrie:
:+1:

PR #12065, opened 2016-02-23
Override media tones in paid content styles

jbreckmckye:
# Turn white links black on paid-content media pages
## What does this change?

On a paid-content page page like http://www.theguardian.com/netflix-love/video/2016/feb/19/isy-sutties-real-dating-disaster-the-cats-tale-video, several elements use the `tone-media` styles that assume a dark background, and appear white. This gives them poor contrast in the paid-content theme.
## Screenshots

_Before:_
image

_After:_
image
## Request for comment

I've fixed this by adding an override for the tonal styles in paid-content mode. Long term, though, we might need a better solution - perhaps we should create a separate `tone-paidfor` style, and use that for styling paidfor articles rather than `paid-content--advertisement-feature`. This might be a bit of refactor, though.

### DIFF
--- a/static/src/stylesheets/module/commercial/_capi.scss
+++ b/static/src/stylesheets/module/commercial/_capi.scss
@@ -191,7 +191,7 @@
         background-color: unset;
     }
 
-    // Tone overrides
+    // Media tone overrides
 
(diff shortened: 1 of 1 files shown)


regiskuckaertz:
:+1:

PR #12066, opened 2016-02-23
Fix toast bug: send correct block ID

OliverJAsh:
# Steps to reproduce
1. Go to a live liveblog with Toast enabled (currently disabled in PROD)
2. Wait ~30 seconds
# Expected

Toast will appear if there are updates, with the correct number of updates displayed.
# Actual

Toast appears with a large number of updates (all blocks!).
# Comments

This is because we are requesting new updates in the format of `<liveblogUrl>.json?lastUpdate=<blockId>&isLivePage=true`, when we should be querying updates in the format of `<liveblogUrl>.json?lastUpdate=block-<blockId>&isLivePage=true`. Note the missing prefix on the block ID.

This was happening because the `data-most-recent-block` attribute only contained the block ID without the `block-` prefix.

I have changed this so that we always query the DOM to find the latest block, whose ID always contains the `block-` prefix.
# Screenshots

image

/cc @Jholder112233

### DIFF
--- a/article/app/views/liveblog/liveBlogBody.scala.html
+++ b/article/app/views/liveblog/liveBlogBody.scala.html
@@ -75,7 +75,7 @@
                         <div class="js-article__container" data-component="body">
                             @* Remove LiveFilter and the fragment once Toast comes out of AB test status *@
                             @fragments.liveFilter(article.fields.isLive)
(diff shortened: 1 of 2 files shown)


Jholder112233:
Yep :+1:

PR #12067, opened 2016-02-23
Replace article structure with space-filler

gtrufitt:
## What does this change?

This replaces my 'are the last 5 elements paragraphs' rudimentary check with the super-powered spacefinder check via space-filler for the in-article email sign-up insertion.

It's less likely to cause clashes and no longer breaks certain contracts.
## Request for comment

@crifmulholland on the rules, @jbreckmckye because you're spacefinderman

### DIFF
--- a/static/src/javascripts/projects/common/modules/email/email-article.js
+++ b/static/src/javascripts/projects/common/modules/email/email-article.js
@@ -13,7 +13,8 @@ define([
     'lodash/collections/find',
     'text!common/views/email/iframe.html',
     'common/utils/template',
-    'common/modules/article/space-filler'
+    'common/modules/article/space-filler',
+    'common/modules/analytics/omniture'
 ], function (
     $,
     bean,
@@ -29,7 +30,8 @@ define([
     find,
     iframeTemplate,
     template,
-    spaceFiller
+    spaceFiller,
+    omniture
 ) {
 
     var listConfigs = {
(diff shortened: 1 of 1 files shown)


static/src/javascripts/projects/common/modules/email/email-article.js
```diff
@@ -101,34 +116,26 @@ define([
...
-                    $iframeEl[listConfig.insertMethod || 'appendTo']($insertEl && $insertEl.length > 0 ? $insertEl : $articleBody);
-                });
+                if (listConfig.insertMethod) {
+                    fastdom.write(function () {
+                        listConfig.insertMethod($iframeEl);
+                    });
+                } else {
+                    spaceFiller.fillSpace(getSpacefinderRules(), function (paras) {
+                        $iframeEl.insertBefore(paras[0]);
+                        emailInserted = true;
```
jbreckmckye: Given that this is async, is there a risk that `init()` will be called in the interval before spaceFiller has had a chance to run your write function?

gtrufitt: Are you referring to this `init`?

```
bean.on(iframe, 'load', function () {
    email.init(iframe);
});
```

If so, as it's inserting an iframe, that `load` event won't fire until it's inserted in the page and made the request so there shouldn't ever be a race condition there I don't think.

jbreckmckye: Fair enough. Just wanted to mention that `spaceFiller` returns a promise, if you need it.

crifmulholland:
The rules look good to me. The only other thing we need is tracking when it's shown - can we please fire an event when rendered onto the page?

static/src/javascripts/projects/common/modules/email/email-article.js
gtrufitt: Aye, thanks, saw that but didn't need it in this case, cheers!

jbreckmckye:
:+1:

PR #12068, opened 2016-02-23
The New Fiver

desbo:
## What does this change?

It's a new email template for articles that we're going to test out on The Fiver. It replaces the R2 system that obviously won't be around for much longer.

It uses the [Zurb Email](http://foundation.zurb.com/emails.html) framework to take some of the hassle out of making it responsive. The big ink.css file here is a part of the Zurb library, so it doesn't need reviewing.

no
## Screenshots

screen shot 2016-02-23 at 17 45 49
## Request for comment

please can i have a comment


crifmulholland:
This is game changing!!

article/app/controllers/ArticleController.scala
```diff
@@ -161,7 +161,7 @@ object ArticleController extends Controller with RendersItemResponse with Loggin
 
   def renderArticle(path: String) = {
     LongCacheAction { implicit request =>
-      mapModel(path) {
+      mapModel(path, request.isEmail) {
```
jennysivapalan: I'd put the variable name of the boolean value here:
`mapModel(path, blocks = request.isEmail)`

jennysivapalan: Also why does email need to use `blocks`? Or more why can't we use `'blocks` everywhere?

article/app/views/fragments/emailArticleBody.scala.html
```diff
@@ -1,20 +1,68 @@
...
+    }
 
     @fragments.emailMainMedia(article)
 
-    @fragments.contentMeta(article, showExtras = false)
+    @paddedRow(Seq("author")) {
+        @article.trail.byline.map { byline =>
+            <h3 class="byline">@byline</h3>
+
+        @if(article.content.tags.contributors.length == 1) {
```
jennysivapalan: Could you use a `map` here or does it just have to be if there is just one contributor?

article/app/controllers/ArticleController.scala
desbo: We can (and should) use them everywhere but it would involve rewriting the article templates which is outside the scope of this PR.

desbo: I mean rewriting the main (non-email) article templates – since this email one was more of a blank slate it seemed like a good place to start using cAPI blocks!

jennysivapalan: Yep that makes sense. Just wondering what does Amp do as it has to come through `renderArticle` too? Does Amp use blocks @NataliaLKB @johnduffell? We should keep the methods for using `mapModel` with blocks consistent?

article/app/views/fragments/emailArticleBody.scala.html
desbo: the fragment it calls just uses `headOption` from the contributors, so i think it's designed not to show multiple contact links.

common/app/views/support/Zurb.scala
```diff
@@ -0,0 +1,31 @@
+package views.support
+
```
gtrufitt: I really like this, makes the email views really clear and clean!

static/src/inline-svgs/icon/instagram.svg
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 16.0.1, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
```
gtrufitt: You probably want to run this through SVGO, and then remove the xmlns attribute which SVGO doesn't remove

gtrufitt:
This is aces! Nice work, I can't comment on all the Scala stuff but it's a :+1: from me outside of that.

article/app/views/fragments/emailArticleBody.scala.html
jennysivapalan: Ah yes, so in this case you could remove the `if` statement as the fragment will take care of it for you.

article/app/controllers/ArticleController.scala
johnduffell: we only started using blocks for liveblogs, everything else uses the body.  it shouldn't be much work to move things over, but at the moment any article still in R2 will have no block.  Amp doesn't do anything special re blocks.

jennysivapalan: ok thanks @johnduffell.

jennysivapalan:
Scala looks good :+1: 
:email: :soccer:

common/app/views/support/Zurb.scala
desbo: thanks! it's actually not a million miles away from the michael jackson thing, with simpler tags expanding to <table> based markup

static/src/inline-svgs/icon/instagram.svg
gtrufitt: Whoops, sorry @You

common/app/views/support/Zurb.scala
gtrufitt: Yeah, definitely, really nice architecture

static/src/inline-svgs/icon/instagram.svg
desbo: lol

PR #12069, opened 2016-02-23
Dont show any onward journey's on amp if in section children's books

NataliaLKB:
## What does this change?

We should not be showing any onward journey's on the Children's Books site. This is because it is possible that not child appropriate content may show in most popular, headlines, or other onward containers.

I have also refactored slightly to remove a redundant import
## Does this affect other platforms?

This only effects AMP :zap: 
## Request for comment

please can i have a comment :smile:


rich-nguyen:
:+1:

PR #12073, opened 2016-02-24
Remove old sticky code

### LINKED ISSUE #12072: Sticky ad banner: transition broken on /uk/culture in Firefox
# Steps to reproduce

1. Go to http://www.theguardian.com/uk/culture in Firefox

# Expected

The ad banner should transition when the ad renders and thus expands in height.

# Actual

The ad overlaps the banner, and then the banner transitions to an incorrect height. This only happens sometimes, suggesting a race condition.

# Demo

OliverJAsh:
I have a feeling this will fix #12072.

It appears that the iframe is transitioning before the `render` functions run due to this CSS.

/cc @regiskuckaertz

### DIFF
--- a/static/src/stylesheets/module/_adslot.scss
+++ b/static/src/stylesheets/module/_adslot.scss
@@ -121,18 +121,6 @@
 .top-banner-ad-container--responsive {
     min-height: ($gs-row-height * 7) + 16;
 }
-.top-banner-ad-container--reveal iframe {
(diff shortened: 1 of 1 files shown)


johnduffell:
great! ship it! :+1:

regiskuckaertz:
Oh, yes! :+1:

NataliaLKB:
:ship:

PR #12074, opened 2016-02-24
remove liveblog smooth scroller completely

johnduffell:
This PR removes the old scroll transitions from liveblogs.  Because all timeline links are permalinks now, they will send you to a url including a block ID, thus causing a page refresh rather than just jumping to the right place.  We decided it would be possible to implement the transitions if the block is already on the same page, but probably not worth it 100%.  However we will monitor reaction and see if people don't like it.
@OliverJAsh hopefully this won't conflict with your stuff?


OliverJAsh:
No additions, beautiful :+1:

PR #12075, opened 2016-02-24
Hide low-relevance Outbrain widget behind a switch

regiskuckaertz:
## What does this change?

Outbrain hasn't fully tested the new widget, as it's not displaying correctly in the Australian edition:
unnamed

I just hid it behind a switch so that they can test on CODE and verify that everything is working as expected.

/cc @JonNorman


JonNorman:
:+1:

PR #12061, opened 2016-02-23
Removed NewCommercialContent switch

JonNorman:
## What does this change?

The `NewCommercialContent` switch has been removed as it is now to be a permanent change. The switch has been deleted and all the logic that used the switch has been altered accordingly.
## What is the value of this and can you measure success?

value: one less switch to worry about
success: paid content still working on the site
## Request for comment

-- @kelvin-chappell 

_Does this PR meet the [contributing guidelines](https://github.com/guardian/frontend/blob/issue_pr_templates/.github/CONTRIBUTING.md#submission)?_

PR #12076, opened 2016-02-24
Make colons in branded content logos consistent

kelvin-chappell:
This is a small fix following on from #12061


JonNorman:
Looks good :+1: - it'd be nice to have the three headers codified once (sponsored, paid-for and foundation) rather than hard-coded in all over the place, but think that can be addressed separately as part of the larger refactoring exercise.

Issue #12026, opened 2016-02-18
Should Outbrains be suppressed if any of legally sensitive, suppress related content, or should hide adverts are set?


michaelwmcnamara:
Currently outbrains ignores the 
shouldHideAdverts, showInRelatedContent, and legallySensitive flags in CAPI and just displays anyway.
Is this correct?
example: http://www.theguardian.com/australia-news/2016/feb/05/church-ethos-of-easy-forgiveness-failed-to-stop-child-sex-abuse-says-peter-jensen

![image](https://cloud.githubusercontent.com/assets/7239344/13154162/86647ac2-d66f-11e5-8b98-18dddae12f45.png)

ScottPainterGNM:
Might be worth pinging this question to Jamie Morrison to investigate?

regiskuckaertz:
Outbrain is not shown when either `shouldHideAdverts` is true (it is false in this example) or the page belongs to the children's books section (here it is australia news). You can see other ads showing up too.

I couldn't find any reference to a `legallySensitive` flag in the global config, not anywhere. Is this something that should be added? In this case it currently does not apply, neither to Outbrain nor any other ad.

/cc @ScottPainterGNM @JonNorman

JonNorman:
I believe that `legallySensitive` is a new flag being introduced - Huma pointed me to this [Useful Links](https://docs.google.com/document/d/1iqCgyKaLepXKDOWQ12FYipwqyeLHYPUpFmsZTHKScHk) doc which, on the second page, details these flags.

NataliaLKB:
@michaelwmcnamara I think this is okay to close, what do you think?

ScottPainterGNM:
@NataliaLKB pretty sure this can be closed off now!

PR #12078, opened 2016-02-24
AMP: Only show Outbrain when shouldHideAdverts is false

JonNorman:
## What does this change?

Relates to #12026 - Outbrain is currently served on AMP regardless of `shouldHideAdverts`.
This is inconsistent with desktop/regular articles; this astoundingly complex change remedies this.

value: consistency with the desktop/regular articles
measure: self-evident

Only AMP
## Request for comment

-- @michaelwmcnamara - I believe that there are still questions around what the behaviour should be for some of the CAPI flags e.g. `legallySensitive`/`sensitive` and how they should be handled, but this at least harmonises the behaviour of Outbrain across both platforms.

### DIFF
--- a/common/app/views/mainAMP.scala.html
+++ b/common/app/views/mainAMP.scala.html
@@ -69,7 +69,10 @@
                     @fragments.amp.onwardTemplateAmp("related-mf2/" + page.metadata.id + ".json")
                 }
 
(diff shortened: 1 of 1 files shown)


jamesgorrie:
:+1:

PR #12081, opened 2016-02-24
Outbrain: Use same codes across all editions for the low-rel slot

regiskuckaertz:
Despite having sent us four sets of codes, one for each edition, OB now wants us to use only one. Let's save a few bytes then!

cc @JonNorman


jbreckmckye:
Fair enough :+1:

JonNorman:
I think there may be a (now-redundant) test that wants to be removed in outbrain.spec.js, line 333.

PR #12084, opened 2016-02-24
Rename dfp-api 'slots' as 'adverts'

jbreckmckye:
This pull request isn't going to get anyone excited. It merely renames the 'slots' array in dfp-api to 'adverts'. Why? Because the name 'slot' is already used by DFP. That means the current data structure looks a little like -

``` javascript
var slots = [
    {
        isRendered: Boolean,
        isLoading: Boolean,
        slot : googletag.defineSlot(...)
    }
];
```

Meanwhile, our code is full of references to `slot` variables, and it's never clear whether they're talking about slots-the-domain-object or slots-the-dfp-entity.

This PR makes a couple of simple changes:
- The `slots` array is now `adverts`
- Anything that handles the slot element is called an `$adSlot` or `adSlot` (still perhaps a little too similar to `slot`, but this is the convention everywhere else)
- Any reference to a `slot` should now exclusively refer to a [Googletag Slot](https://developers.google.com/doubleclick-gpt/reference#googletagslot).

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
+++ b/static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
@@ -683,8 +683,9 @@ define([
         init:           init,
         addSlot:        addSlot,
         refreshSlot:    refreshSlot,
-        getSlots:       getSlots,
+
         // Used privately but exposed only for unit testing
+        getSlots:       getSlots,
         shouldLazyLoad: shouldLazyLoad,
         getCreativeIDs: getCreativeIDs,
         checkForBreakout: checkForBreakout,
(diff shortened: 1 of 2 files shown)


regiskuckaertz:
Man, I'm kind of a [SNOOT](http://grammar.about.com/od/grammarfaq/f/whatisasnoot.htm) so this is definitely welcome. Besides, isn't naming things the most difficult thing in CS?

On a sidenote, I really think we should better take care of the `adverts` object. Right now we delete the entry every time it loads, but this is bad for two reasons:

* `delete` is bad for performances
* listeners who want to know if an ad has been rendered after the fact will hopelessly wait on the `rendered` event... something my `trackAd` could address

Definitely :+1:

jamesgorrie:
Also way easier for a newcomer to the codebase to understand.
:+1:

jbreckmckye:
@regiskuckaertz I believe that `delete` forces V8 to bail-out of its 'hidden classes' mode, and start treating the object as a plain old dictionary. I guess this is probably still 'quick enough' for a non-hot-code-path but is definitely something to avoid on objects that get read several times per message.

PR #12085, opened 2016-02-24
Fix image picking behaviour in Facia

rich-nguyen:
##### What does this change?
- Removes duplicated image picking code in facia and facia-press, in favour of trusting the `trailPicture` selected by the frontend common library.
- Reimplements rss feed media, and title.
- Updates and simplifies stale image picking logic, was based on pre-facia breakpoints.
- Simplifies element slimming, done in facia-press.
##### What is the value of this and can you measure success?

Success is a coconut. 
##### Before

screen shot 2016-02-24 at 17 14 44
##### After

screen shot 2016-02-24 at 17 14 54


common/app/common/TrailsToRss.scala
```diff
@@ -132,16 +131,21 @@ object TrailsToRss extends implicits.Collections {
         .distinctBy(faciaContent => faciaContent.properties.maybeContentId.getOrElse(faciaContent.card.id))
         .flatMap(_.properties.maybeContent)
 
-    fromFaciaContent(pressedPage.metadata.webTitle, faciaContentList, pressedPage.metadata.url, pressedPage.metadata.description)
+    val webTitle = if (pressedPage.metadata.contentType != GuardianContentTypes.NetworkFront) {
+      s"${pressedPage.metadata.webTitle} | The Guardian"
+    } else {
+      "The Guardian"
```
janua: Does this fix the RSS title now?

janua:
:+1:

common/app/common/TrailsToRss.scala
rich-nguyen: Yeah

PR #12086, opened 2016-02-24
Liveblog pagination navigation

OliverJAsh:
## What does this change?

This adds richer navigation controls for paginated liveblogs.
### Mobile

image
### Desktop/tablet

image
## Request for comments
- I am using flexbox for layout because it is now well supported (inclusive of the old syntax) and it means I can properly separate my HTML from the layout styles. Here's what the gracefully degraded version looks like in IE 9 (using floats without adding much complexity to the code as agreed with @zeftilldeath):
- I've tried to write the CSS in a way that avoids unsetting, so all classes are orthogonal.
- I've used `<a>`’s without `href`s as placeholders when there is no link (i.e. newer page link on newest page). Apparently this is valid HTML5
## To do:
- [ ] Rename arrow icons
- [ ] Fix conflict with Toast
- [ ] Update design


article/app/liveblog/LiveBlogPageModel.scala
```diff
@@ -29,7 +40,17 @@ object LiveBlogPageModel {
...
+  blocks: Seq[B],
+  main: Seq[B] /*for key events - TODO remove*/ ,
+  newer: PageReference,
+  older: PageReference,
+  canonical: PageReference,
+  pageNumber: Int,
+  pagesLength: Int,
+  newest: PageReference,
+  oldest: PageReference
+)
```
OliverJAsh: I wasn't sure whether to add these properties to this case class. Let's chat in person when you look at this? :-)

paperboyo:
Could I ask why a different/inconsistent and less usable (can't get quickly to specific page) design, please?

choose-one

zeftilldeath:
Because navigating to page '4' has no context. The user has no idea of what
time span the live blog covers/regularity of posts. It's a live blog,
something that is most likely to be read sequentially (As that's how the
story will have developed). Not at all comparable to wading through
hundreds of unrelated comments to an article.

zeftilldeath:
@paperboyo Because navigating to page '4' has no context. The user has no idea of what time span the live blog covers/regularity of posts. It's a live blog, something that is most likely to be read sequentially (As that's how the story will have developed). Not at all comparable to wading through hundreds of unrelated comments to an article.

paperboyo:
> The user has no idea of what time span the live blog covers/regularity of posts.

True that. Nor does they have with comments, though. I would still like it more consistent (don't care which way). Just my 2gr, obviously!

jamesgorrie:
> True that. Nor does they have with comments

Comments very often don\t have flow, a lot of comments relate directly to the article, not the comment before or after - whereas a liveblog post generally relates to something before or after.

Saying that, similar to video, some people (seeking instant gratification) still fast-forward to the middle and usually the end.

paperboyo:
Similarity to comments, true or false, have nothing to do with difference in styling, IMHO. Only with providing intermediate page-steps. And even then, only partially (comments also are filed chronologically but can be nested in threads, intermediate pages pagination is of questionable usefulness there, too, because as comments number grows, they find themselves on a different page). That last point, actually, could be more useful in Liveblogs, as there are no new blocks inserted in the middle of a flow.

My biggest (still tiny) objection is to it looking different.

article/app/liveblog/LiveBlogPageModel.scala
```diff
@@ -27,13 +46,27 @@ object LiveBlogPageModel {
...
-case class BlockPage(blockId: String) extends PageReference { val suffix = Some(s"?page=with:block-$blockId#block-$blockId") }
+case class LiveBlogPageModel[B](
+  allBlocks: Seq[B] /*for key events - TODO remove*/ ,
+  currentPage: PageReference[B],
+  pagination: Option[Pagination[B]]
+)
+
+case class FirstPage[B](blocks: Seq[B]) extends PageReference[B] {
+  val suffix = ""
+  val pageNumber = 1 }
```
johnduffell: slightly odd bracket being on the same line

article/app/liveblog/LiveBlogPageModel.scala
```diff
@@ -27,13 +46,27 @@ object LiveBlogPageModel {
...
-sealed trait PageReference {
-  def suffix: Option[String]
-}
-case object NoPage extends PageReference { val suffix = None }
-case object FirstPage extends PageReference { val suffix = Some("") }
-case class BlockPage(blockId: String) extends PageReference { val suffix = Some(s"?page=with:block-$blockId#block-$blockId") }
+case class LiveBlogPageModel[B](
+  allBlocks: Seq[B] /*for key events - TODO remove*/ ,
+  currentPage: PageReference[B],
+  pagination: Option[Pagination[B]]
```
johnduffell: maybe we should call them controls or something

article/app/views/fragments/liveFilter.scala.html
```diff
@@ -0,0 +1,63 @@
...
+@import model.liveblog.BodyBlock
+@import model.Article
+@(isLive: Boolean, article: Article, pageModel: LiveBlogPageModel[BodyBlock])
+
+<div class="live-toolbar js-live-toolbar modern-visible">
+    @pageModel.pagination.map { pagination =>
+        <div class="liveblog-navigation">
+            <div class="liveblog-navigation__detail">
+                <!-- div necessary for flexbox -->
+                <div><span class="liveblog-navigation__page-number">@pageModel.currentPage.pageNumber</span> of <span class="liveblog-navigation__pages-length">@pagination.pagesLength</span></div>
```
johnduffell: maybe more readable to call it numberOfPages not pagesLength

article/app/views/liveBlog.scala.html
```diff
@@ -1,16 +1,16 @@
 @import _root_.liveblog.LiveBlogPageModel
 @import model.liveblog.BodyBlock
-@(model: LiveBlogPage, blocks: LiveBlogPageModel[BodyBlock])(implicit request: RequestHeader)
+@(model: LiveBlogPage, pageModel: LiveBlogPageModel[BodyBlock])(implicit request: RequestHeader)
 
 @import common.LinkTo
 
 @main(model){
-    @blocks.earlier.suffix.map { suffix =>
+    @pageModel.pagination.flatMap(_.older).map(_.suffix).map { suffix =>
```
johnduffell: are we getting rid of these controls/putting the new controls here instead?

article/app/views/fragments/liveFilter.scala.html
```diff
@@ -0,0 +1,63 @@
+@import _root_.liveblog.LiveBlogPageModel
```
johnduffell: this file is called liveFilter but it's more of a navigation control rather than a filter

Pearson82:
@paperboyo @zeftilldeath @jamesgorrie @OliverJAsh 

I agree with all the points people have made above, but we should make an effort to unify these two styles as they feel like essentially the same thing despite slightly different contexts. 

If the intermediate numbers are redundant on comments then we should get rid. We don't want to create two patterns for the same thing – although at the wrath of the commenters :)

johnduffell:
looks great from a scala POV and looking forward to having nice controls, thanks! :+1: feel free to get someone expert to check the css though
and agreed about the comments thing, how many people want to skip to a specific page of the comments after all?  maybe only useful when you're on newest first and lots of comments are being posted :stuck_out_tongue_closed_eyes:

article/app/views/liveBlog.scala.html
OliverJAsh: Discussed in person :+1:

NataliaLKB:
css gets :+1: from me

Jholder112233:
Are the controls also duplicated at the bottom of the page's posts? As that feels like the place when people will want to navigate pages, rather than at the top of the liveblog.

zeftilldeath:
@Jholder112233 Yeah, in fact there's a slight change from the screenshots now. The controls are at the bottom of the first liveblog page, then at the top AND bottom of liveblog archive pages. Have tidied up the styling a bit too to be clearer on touch target etc.

screen shot 2016-02-25 at 17 24 51

Jholder112233:
@zeftilldeath Much clearer :gem:

OliverJAsh:
@Pearson82 @zeftilldeath @stephanfowler Sorry I merged this without responding to the discussion about bringing consistency to our pagination controls across the site. I agree we should, maybe we could have a consistent design with optional numbered links?

paperboyo:
> maybe we could have a consistent design with optional numbered links?

:+1:

PR #12087, opened 2016-02-25
Some juicy css specificity overrides

uplne:
I am not enjoying this PR. All this madness is to fix this:

screen shot 2016-02-24 at 17 40 31

And it happens because of this:
screen shot 2016-02-24 at 17 40 44

### DIFF
--- a/static/src/stylesheets/module/commercial/gustyle/_gu-comcontent.scss
+++ b/static/src/stylesheets/module/commercial/gustyle/_gu-comcontent.scss
@@ -35,6 +35,7 @@
     color: #ffffff;
 }
 
+.content--immersive .from-content-api .gu-display__content-title,
 .gu-display__content-title {
     @include f-headlineSans;
     @include font-size(20, 20);
@@ -85,6 +86,7 @@
     color: #ffffff;
 }
 
+.content--immersive .from-content-api .gu-display__content-color--dark,
 .gu-display__content-color--dark {
     color: #000000;
 }


static/src/stylesheets/module/commercial/gustyle/_gu-comcontent.scss
```diff
@@ -35,6 +35,7 @@
     color: #ffffff;
 }
 
+.content--immersive .from-content-api .gu-display__content-title,
 .gu-display__content-title {
     @include f-headlineSans;
```
jamesgorrie: Would using `!important` with a note be more explicit and easier to fathom for someone seeing this later?

static/src/stylesheets/module/commercial/gustyle/_gu-comcontent.scss
```diff
@@ -35,6 +35,7 @@
     color: #ffffff;
 }
 
+.content--immersive .from-content-api .gu-display__content-title,
```
jamesgorrie: Would using !important with a note be more explicit and easier to fathom for someone seeing this later?

jbreckmckye: It may well be, although it's not the real fix for this problem in the long term.

The long term fix is to try and disentangle our CSS and give this commercial content a 'blank slate' to work upon.

regiskuckaertz: One line in the sand I have in mind is to seldom use the descendant selector. We control the markup, markup & style go hand in hand (markup *is* presentation, especially here), so we should never avoid the opportunity to use more efficient selectors, like the child one.

`.something h2` is a recipe for disaster.

regiskuckaertz:
:+1:

static/src/stylesheets/module/commercial/gustyle/_gu-comcontent.scss
jbreckmckye: > One line in the sand I have in mind is to seldom use the descendant selector.

Yes.

> .something h2 is a recipe for disaster.

**Yes**.

PR #12089, opened 2016-02-25
Switch h2 rules for sign-up positioning

gtrufitt:
# I got some numbers wrong
## What does this change?

It makes sure the email sign-up has 200px space _below_ an h2, instead of above so that we don't get the sign-up form immediately after the heading.

It is a bugfix to avoid bad flow of articles.


zeftilldeath:
:lollipop: :+1:

PR #12091, opened 2016-02-25
Don't use space-finder for G-today AU

gtrufitt:
## What does this change?

The method of insertion for The Guardian Today in Australia will now use a basic insert to the bottom of the article, instead of using space-finder.

This was a request from Aus editorial, so keeps them happy but also gives us a comparison to be able to measure sign-ups when the form is bottom of article vs mid-article.

Nope.
## Request for comment

@crifmulholland


rich-nguyen:
:+1:

PR #12093, opened 2016-02-25
Amp tweaks

zeftilldeath:
Firstly a small fix to prevent this from happening:
screen shot 2016-02-24 at 11 08 05

Then typography, The bold weight of Georgia is **SO FAT**, that it warrants using the medium cut of Egyptian Web we have available as a replacement. Examples of why:

dustbag

CC @NataliaLKB


NataliaLKB:
:+1:

PR #12094, opened 2016-02-25
Layout variations for commercial containers

regiskuckaertz:
## What does this change?

Improve the layout of commercial containers:
picture 1
picture 2

Images get huge when there's only 1 or 2 items in a commercial container:
unnamed1
unnamed2

With some mind-blowing functional expression @kelvin-chappell

### DIFF
--- a/common/app/views/fragments/containers/facia_cards/commercialContainer.scala.html
+++ b/common/app/views/fragments/containers/facia_cards/commercialContainer.scala.html
@@ -37,7 +37,7 @@
                         <div class="commercial__body">
                             @for(layout <- container.containerLayout) {
                                 @for(slice <- layout.slices) {
-                                    <ul class="lineitems l-row l-row--cols-4">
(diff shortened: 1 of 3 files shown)


common/app/views/fragments/containers/facia_cards/commercialContainer.scala.html
```diff
@@ -37,7 +37,7 @@ <h3 class="commercial__title">@container.displayName</h3>
                         <div class="commercial__body">
                             @for(layout <- container.containerLayout) {
                                 @for(slice <- layout.slices) {
-                                    <ul class="lineitems l-row l-row--cols-4">
+                                    <ul class="lineitems l-row l-row--cols-@math.min(slice.columns.flatMap(_.cards).map(_.item).filter(_.isInstanceOf[ContentCard]).length, 4)">
```
kelvin-chappell: Perhaps move this expression into https://github.com/guardian/frontend/blob/master/common/app/views/support/Commercial.scala#L83?

PR #12095, opened 2016-02-25
Send immersives into email template

desbo:
Slightly change the `ArticleController` logic so that immersives go through to the email template as well as regular articles. Also never render embeds in emails since no email client will show them.


gtrufitt:
:+1:

PR #12096, opened 2016-02-25
Add support for new `packages` field that is going to replace `storyPackages`

TBonnin:
# Add support for new `packages` field
## What does this change?

New story packages are now returned by CAPI into the `packages` field which is going to replace `storyPackages`
This change checks for this new `packages` field content and fall back over old `storyPackages` if empty.

App team is making the same change at the moment. @davidfurey
## Screenshots

Before
screen shot 2016-02-25 at 16 10 13

After
screen shot 2016-02-25 at 16 10 00
## Request for comment

@rich-nguyen


common/app/model/RelatedContent.scala
```diff
@@ -25,8 +25,17 @@ case class RelatedContent (
 
 object RelatedContent {
   def apply(parent: ContentType, response: ItemResponse): RelatedContent = {
-    // It's misleading to use storyPackage here rather than relatedContent. A tidy up should rename this file.
-    val items = response.storyPackage.map { item =>
+    // It's misleading to use storyPackage here rather than relatedContent. A tidy up should rename this object
```
TBonnin: I agree this is very misleading. It took me more than it should have to understand how `RelatedContent` works.
I am happy to spend some time and refractor this once this PR is out. @rich-nguyen

project/Dependencies.scala
```diff
@@ -25,7 +25,7 @@ object Dependencies {
   val commonsHttpClient = "commons-httpclient" % "commons-httpclient" % "3.1"
   val commonsIo = "commons-io" % "commons-io" % "2.4"
   val cssParser = "net.sourceforge.cssparser" % "cssparser" % "0.9.18"
-  val contentApiClient = "com.gu" %% "content-api-client" % "7.27"
+  val contentApiClient = "com.gu" %% "content-api-client" % "7.28"
```
TBonnin: Necessary to include packages related bugfix https://github.com/guardian/content-api-scala-client/blob/master/CHANGELOG.md#728

rich-nguyen:
:+1: You may need to run tests again, because you've upgraded content-api.

common/app/model/RelatedContent.scala
rich-nguyen: Yes, `RelatedContent`could do with another look, now that the content model is a bit more stable.

PR #12098, opened 2016-02-25
Improve FaciaControllerTest (mf2.json rendering)

TBonnin:
## What does this change?

Instead of checking if response is not of JsUndefined type,
tests are now checking that response contains an `items` array
with the correct size
## Request for comment

@rich-nguyen


rich-nguyen:
cool :+1:

rich-nguyen:
do you need to change the test so that it tests a non-default editionalised section, eg `au/sport`?

TBonnin:
Good point @rich-nguyen. I'll do that

TBonnin:
@rich-nguyen Done :smile: https://github.com/guardian/frontend/commit/501c36e618ed09a7068cb227423d647cb741e44b

NataliaLKB:
:+1:

PR #12099, opened 2016-02-25
Refactor commercial-component

regiskuckaertz:
There was lots of useless work going on in this module and it was (to me at least) very hard to reason about.

It runs 10–20x faster than the previous version, but we're talking milliseconds here so it's not really a good selling point.


static/src/javascripts/projects/common/modules/commercial/creatives/commercial-component.js
```diff
@@ -32,72 +24,54 @@ define([
...
+        };
+    }
+
+    function soulmatesGroupUrlBuilder(url) {
+        return function (params) {
+            return buildComponentUrl(url + params.soulmatesFeedName, params);
         };
+    }
+
+    function complexUrlBuilder(url, withT, withKeywords) {
```
uplne: What does it mean withT?

regiskuckaertz: `t` is the name of the parameter. I'll change it to `withSpecificID`, that is its real use.

static/src/javascripts/projects/common/modules/commercial/creatives/commercial-component.js
```diff
@@ -32,72 +24,54 @@ define([
...
+    map,
+    reduce
 ) {
+    var urlBuilders = {
+        bestbuy:        defaultUrlBuilder('money/bestbuys'),
+        soulmates:      defaultUrlBuilder('soulmates/mixed'),
+        soulmatesTest:  defaultUrlBuilder('soulmates-test/mixed'),
+        capiSingle:     defaultUrlBuilder('capi-single'),
+        capi:           defaultUrlBuilder('capi'),
+        paidforCard:    defaultUrlBuilder('paid'),
```
uplne: Let's order this:

defaultUrlBuilder,
complexUrlBuilder,
bookUrlBuilder,
soulmatesGroupUrlBuilder

johnduffell:
looks awesome to me, if it's good code we should get this out the door! :+1: if not close

PR #12101, opened 2016-02-26
Added missing parameter in the test blended component

regiskuckaertz:
@jbreckmckye any comment?


jbreckmckye:
I don't know how these components work, but if they're fixed locally, by all means, ship them. :+1:

PR #12104, opened 2016-02-26
Layout variations for commercial containers (2/2)

regiskuckaertz:
Turns out there's an extra case I did not see [the first time](https://github.com/guardian/frontend/pull/12094).

cc @kelvin-chappell @jbreckmckye


kelvin-chappell:
:+1:

PR #12105, opened 2016-02-26
Determine branded container logo from first card

kelvin-chappell:
## What does this change?

The logos on sponsored/paid/foundation-funded containers were determined by parsing the capi query that builds the container.  This change uses the metadata of the first card in the container to determine the logo instead.
- We're ready to replace ad-server served logos with capi-served logos.
- Targeting is more accurate.
- A branded container can be built with a mixture of manual and automated content as long as the first card targets the correct logo.

/cc @JonNorman


JonNorman:
:+1: looks good - would it be worth replacing the (now removed) tests with an analogous one for the new functionality?

kelvin-chappell:
Not sure where the tests should go at the moment or what should be tested so will set up new tests when required.

PR #12106, opened 2016-02-26
Email template updates

desbo:
Minor fixes and updates:
- lock banner size
- show block titles (fixes missing headers in The Minute)
- remove social links in footer
- add proper links for prefs/unsubscribe
- use png rather than svg icons for better client support
- add blockquote styling

thanks to @zeftilldeath for the last two


common/app/views/fragments/email/footer.scala.html
```diff
@@ -1,23 +1,9 @@
...
-
-    <a class="social__link" href="https://www.instagram.com/guardian">
-        @fragments.inlineSvg("instagram", "icon") 272k
-    </a>
-}
+@import views.support.EmailHelpers._
 
 @fullRow(Seq("email__links")) {
-    <a class="email__link" href="">Manage your emails</a> | <a class="email__link" href="">Unsubscribe</a>
+    <a class="email__link" href="https://profile.theguardian.com/email-prefs">Manage your emails</a> | <a class="email__link" href="http://click.mail.theguardian.com/?qs=699722b7e50d4eb4c88091e8d2b9b341b0bba21aed1e528c64a42b5605713f23">Unsubscribe</a>
```
gtrufitt: What's the query string there? Should it be hardcoded or is it some kind of identifier that should get inserted at send time?

gtrufitt:
:+1: :cool:

common/app/views/fragments/email/footer.scala.html
desbo: not sure, i copied it from the current fiver email, will try to find out

PR #12107, opened 2016-02-26
Email sign-up: Single click for signed-in users

gtrufitt:
Looking to get this in Monday morning.
## What does this change?

This removes the email input for signed in users so that they are able to click the sign up button with no friction. It also checks whether the user has already signed up so they are not shown sign-up forms for emails they already get. Right thing, right time, mate.

This change:
- Improves the user experience of signing up for signed-in users, single click, minimal friction
- Allows us to ensure that signed in users  that are signing-up are tied to one email address, allowing management in email preference correctly
- I have added the state of the user (signed in / out) to the omniture hit to ensure that we can segment the differences between these sign ups

No.
### Signed Out Article

not signed in article
### Signed In Article

signed in article
### Signed Out Footer

not signed in footer
### Signed In Footer

signed in footer
## Request for comment
- @NathanielBennett - Can you look at the ID stuff I've added as you've worked with it before
- @crifmulholland - Checkity check, the chinese chicken.
- @desbo @OliverJAsh - Scan of the JS to see if I've made any sillies or not handled errors anywhere you can see
- @superfrank For reference, I'm going to be doing more on the compact version after this is in, so lets chat then

### DIFF
--- a/static/src/javascripts/projects/common/modules/email/email-article.js
+++ b/static/src/javascripts/projects/common/modules/email/email-article.js
@@ -15,7 +15,7 @@ define([
     'text!common/views/email/iframe.html',
     'common/utils/template',
     'common/modules/article/space-filler',
-    'common/modules/analytics/omniture'
+    'common/modules/analytics/omniture',
     'common/utils/robust'
 ], function (
     $,
@@ -34,7 +34,7 @@ define([
     iframeTemplate,
     template,
     spaceFiller,
-    omniture
+    omniture,
     robust
 ) {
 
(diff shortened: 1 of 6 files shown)


paperboyo:
Font police: is all live text rendered in Helvetica by any chance?

Have a nice weekend!

gtrufitt:
> Font police: is all live text rendered in Helvetica by any chance?

@paperboyo No, it's all "Guardian Text Sans Web" I believe. 

See http://www.theguardian.com/football/blog/2016/feb/29/premier-league-talking-points for an example.

Is that what you meant?

static/src/javascripts/projects/common/modules/email/email-article.js
```diff
@@ -123,8 +128,17 @@ define([
                 }
             };
         },
+        buildUserSubscriptions = function (response) {
+            if (response && response.status !== 'error' && response.result && response.result.subscriptions) {
+                userListSubscriptions = pluck(response.result.subscriptions, 'listId');
```
desbo: might be worth using `map` here, syntax would be the same (using a string instead of a fn gets that property), it's more well-known, and `pluck` has been dropped from lodash 4.

paperboyo:
Sorry, @gtrufitt, must buy :eyeglasses:. Policed too small main media, though, so all good ;-)

gtrufitt:
@paperboyo :sweat_smile:

NathanielBennett:
:+1: L(R)GTM

desbo:
:+1:

PR #12108, opened 2016-02-29
don't show breaking news if the user cannot hide it

sndrs:
## What does this change?

stops breaking news being displayed if storage is unavailable, and cannot therefore be hidden
## Request for comment

@stephanfowler


paperboyo:
:gem:

OliverJAsh:
Nice! :+1:

gtrufitt:
:cool:

PR #11106, opened 2015-11-12
Remove self-help section from tech feedback page to better gauge crashing

OliverJAsh:
We haven't had any feedback about crashing since the end of September. Is this because users are now guided to the self-help page? Feedback is a good indicator of whether we still have a problem, so this temporarily takes away the self-help section to help us better gauge whether users are still having problems.

/cc @johnduffell

PR #12112, opened 2016-02-29
Restore self help section

OliverJAsh:
Reverts #11106

We removed this to track crashing, but whilst crashing doesn't appear to be so bad, we can better help users by providing the link to the self-help article.

/cc @ScottPainterGNM


ScottPainterGNM:
:+1: LGTM

PR #12113, opened 2016-02-29
Added fallback when blended component would not find any item by id

regiskuckaertz:
## What does this change?

The blended component template in DFP allows the merch team to specify an item for each component, via its ID. Except that if this ID leads to nothing, the whole component fails to display.

This change makes sure the component falls back to fetching a targeted item.

cc @kelvin-chappell


kelvin-chappell:
:+1:

PR #12114, opened 2016-02-29
Fix URLs in InlineStyles

desbo:
just a quick fix

### DIFF
--- a/common/test/common/InlineStylesTest.scala
+++ b/common/test/common/InlineStylesTest.scala
@@ -71,8 +71,12 @@ class InlineStylesTest extends FlatSpec with Matchers {
   }
 
   it should "work with urls in values" in {
-    val styles = "background: url(http://www.example.com/image.png)"
+    val http = "background: url(http://www.example.com/image.png)"
+    val https = "background: url(https://www.example.com/image.png)"
+    val relative = "background: url(//www.example.com/image.png)"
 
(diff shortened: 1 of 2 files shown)


common/test/common/InlineStylesTest.scala
```diff
@@ -69,4 +69,10 @@ class InlineStylesTest extends FlatSpec with Matchers {
 
     CSSRule.makeStyles(styles) should be(ListMap("font-family" -> "Arial", "background" -> "red url(data:image/png;base64,ABCDEF)"))
   }
+
+  it should "work with urls in values" in {
+    val styles = "background: url(http://www.example.com/image.png)"
+
+    CSSRule.makeStyles(styles) should be(ListMap("background" -> "url(http://www.example.com/image.png)"))
```
sndrs: do you think we should force it to be a relative protocol?

desbo: i don't know if the css inliner itself should be opinionated about that, obviously the url here is just for this test and in real usage the urls will probably come from something like the `Static` helper in scala - it may be an idea to upgrade that to always use relative, but i think right now all images are served over https which might be better?

desbo: although come to think of it, i should add test cases for every form of url

sndrs:
:+1:

PR #12115, opened 2016-02-29
Email add killswitch

gtrufitt:
## What does this change?

Adds killswitch for in-article email signup insertion (the switch already existed, I just hadn't added the check). It also corrects a minor CSS bug.

To do what switches do best, allow us to shut this down if things go pear shaped.


ScottPainterGNM:
dsc_0007

gtrufitt:
@ScottPainterGNM I'm taking that as a +1

ScottPainterGNM:
Ha I think Pear Bear should replace the thumbs up in fact

PR #12116, opened 2016-02-29
Extending Plista-for-AU switch

JonNorman:
## What does this change?

Switch needs extending for another month whilst Plista make changes on their end.


regiskuckaertz:
:+1:

PR #12120, opened 2016-02-29
Ship Toast

OliverJAsh:
Toast has been in production for long enough now that we're happy/confident with it! This PR removes all the old auto update logic.

/cc @Jholder112233


static/src/javascripts/projects/common/modules/ui/autoupdate.js
```diff
@@ -25,190 +34,189 @@ define([
...
-                    this.notificationBar.notify(this.unreadBlocks);
-                    mediator.emit('modules:autoupdate:unread', this.unreadBlocks);
+        var scrolledPastTopBlock = function () {
+            return $liveblogBody.offset().top < window.pageYOffset;
+        };
+        var isLivePage = window.location.search.indexOf('?page=') === -1;
+
+
+        var setUpListeners = function () {
+            bean.on(document.body, 'click', '.toast__button', function () {
```
regiskuckaertz: Are `.toast__button`s injected in the page? If not then you can save a query either by attaching the listener on the button itself (here it will fire every time there's a click) or (I learned this while reading bean.js source) by passing it instead:

`bean(document.body, 'click', [$toastButton[0]], ...)`

static/src/javascripts/projects/common/modules/ui/autoupdate.js
```diff
@@ -25,190 +34,189 @@ define([
...
+        var newBlocks;
+        var currentUpdateDelay = options.minUpdateDelay;
+        var latestBlockId;
+        var updating = false;
+
+        // Cache selectors
+        var $liveblogBody = $('.js-liveblog-body');
+        var $toastButton = $('.toast__button');
+        var $toastSpaceReserver = $('.toast__space-reserver');
+        var $toastText = $('.toast__text', this.$toastButton);
```
regiskuckaertz: Is the `this` necessary here?

OliverJAsh: This PR is actually just moving the contents of `newAutoupdate.js` instead `autoupdate.js`, to replace the previous system. So I haven't made any changes to the code per se, just renamed and deleted bits.

Not sure why `this` is here :confused:

regiskuckaertz: oh, right! good riddance legacy code :smile: by all means :+1:

johnduffell:
fanstastic! :+1: :speedboat: liveblogs really needed some love and now they feel more modern

PR #12123, opened 2016-03-01
Making story packages in AMP have an absolute url instead of a relative one

NataliaLKB:
# Absolute vs Relative
## What does this change?

Makes sure that links on story packages on amp takes us to www.theguardian.com!

Only affects Amp
## Request for comment

@stephanfowler


stephanfowler:
:+1:

PR #12124, opened 2016-03-01
Updating name and wording on email formerly known as 'Best of Cif' to 'Best of Guardian Opinion'

michaelwmcnamara:
This changes the name of the "Best of Cif" email signup to "Best of Guardian Opinion"
Descriptions have also been updated.

In addition  email signup name of "Best of Comment is Free - Australia" - has been changed to "Best of Guardian Opinion" and description updated.

### DIFF
--- a/common/app/controllers/EmailSignupController.scala
+++ b/common/app/controllers/EmailSignupController.scala
@@ -38,7 +38,7 @@ object ListIds {
   val guardianTodayUs = 1493
   val guardianTodayAu = 1506
 
-  val theBestOfCiF = 2313
+  val theBestOfOpinion = 2313
   val theFiver = 218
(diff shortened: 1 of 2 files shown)


gtrufitt:
Should we be updating the `data-link-names` here too @crifmulholland ? Eg: `Subscribe to Best of Comment is free - Australia`

michaelwmcnamara:
Good point! Will wait for confirmation from Chris.
Possible also the body text here?
http://www.theguardian.com/commentisfree/2014/dec/04/best-of-comment-is-free-australia-subscribe-by-email

crifmulholland:
Hi, no that's fine. We can just keep it as-is for now.

gtrufitt:
:+1:

PR #12128, opened 2016-03-01
Add close button to in-article email sign-up

gtrufitt:
## What does this change?

This allows users to remove the email sign-up from in an article. It stores this preference in localstorage so that the user does not see the same email sign-up on the same machine.

This improves the user experience of in-article sign-up forms regardless of the position of this insertion putting the decision to remove in the user's hands (and remembering that decision).

Nope
## Screenshots

image

### DIFF
--- a/static/src/javascripts/projects/common/modules/email/email-article.js
+++ b/static/src/javascripts/projects/common/modules/email/email-article.js
@@ -16,7 +16,8 @@ define([
     'common/utils/template',
     'common/modules/article/space-filler',
     'common/modules/analytics/omniture',
-    'common/utils/robust'
+    'common/utils/robust',
+    'common/modules/user-prefs'
 ], function (
     $,
     bean,
@@ -35,7 +36,8 @@ define([
     template,
     spaceFiller,
     omniture,
-    robust
+    robust,
+    userPrefs
 ) {
 
     var listConfigs = {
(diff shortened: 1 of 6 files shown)


static/src/stylesheets/module/email/_form.scss
```diff
@@ -163,6 +163,28 @@
...
+    top: 0;
+    right: 0;
+}
+
+.email-sub--hidden {
+    @include u-h();
+}
+
+.email-sub--close-icon svg {
+    fill: darken(guss-colour(neutral-2, $pasteup-palette), 10%);
```
sndrs: this can just be `colour(neutral-2)`.

guss is a bit crazy, it's something we'll tidy up

gtrufitt: Ah yeah, cool.

static/src/javascripts/projects/common/modules/email/email-article.js
```diff
@@ -138,10 +140,17 @@ define([
         },
         listCanRun = function (listConfig) {
             // Check our lists canRun method and make sure that the user isn't already subscribed to this email
-            if (listConfig.canRun && canRunList[listConfig.canRun]() && !contains(userListSubscriptions, listConfig.listId)) {
+            if (listConfig.canRun &&
```
johnduffell: canRun is a misleading name (I know it's existing code) but I think this is more like listName?

johnduffell:
the cross seems quite small, is it big enough on mobile to target with fat fingers?

johnduffell:
awesome idea, power to the users - they love it! :+1:

gtrufitt:
> the cross seems quite small, is it big enough on mobile to target with fat fingers?

Yeah, good shout, it has padding around it though so the click area is nice and big :+1: 

> power to the users

:muscle:

static/src/javascripts/projects/common/modules/email/email-article.js
gtrufitt: Good point, I'll update this.

paperboyo:
I love it. The smallest problem possible: in beta and release Firefoxes on both Win and Mac, the close button has no hover state. Works well in Chrome:

no hand

PR #12130, opened 2016-03-01
Live Blog pagination styling and toast z-index

zeftilldeath:
Firstly z-index added to toast so it shows over those pesky ads. Also centered a slightly mis-aligned toast.
screen shot 2016-03-01 at 16 18 06

Secondly, re-visited the visuals of this beast, to make them more in line with the comments pagination styling.  

screen shot 2016-03-01 at 12 22 03

Tiny mobile breakpoint:
screen shot 2016-03-01 at 16 33 49

Please check the css @OliverJAsh

### DIFF
--- a/article/app/views/fragments/liveBlogNavigation.scala.html
+++ b/article/app/views/fragments/liveBlogNavigation.scala.html
@@ -15,9 +15,9 @@
                 href="/@{article.content.metadata.id}@newest.suffix"
                 title="Go to page @newest.pageNumber"
             }
-            class="liveblog-navigation__link link-navigation__link--secondary link-navigation__link--secondary--newer @if(pagination.newest.isEmpty) { liveblog-navigation__link--disabled }"
+            class="liveblog-navigation__link liveblog-navigation__link--secondary liveblog-navigation__link--secondary--newer @if(pagination.newest.isEmpty) { liveblog-navigation__link--disabled }"
             data-link-name="newest page">
-                @fragments.inlineSvg("arrow-left-double", "icon", List("link-navigation__link--secondary--newer__icon"))
+                @fragments.inlineSvg("arrow-left-double", "icon", List("liveblog-navigation__link--secondary--newer__icon"))
                 Newest
             </a>
 
@@ -45,10 +45,10 @@
                 href="/@{article.content.metadata.id}@oldest.suffix"
                 title="Go to page @oldest.pageNumber"
             }
-            class="liveblog-navigation__link link-navigation__link--secondary link-navigation__link--secondary--older @if(pagination.oldest.isEmpty) { liveblog-navigation__link--disabled }"
+            class="liveblog-navigation__link liveblog-navigation__link--secondary liveblog-navigation__link--secondary--older @if(pagination.oldest.isEmpty) { liveblog-navigation__link--disabled }"
             data-link-name="oldest page">
                 Oldest
-                @fragments.inlineSvg("arrow-right-double", "icon", List("link-navigation__link--secondary--older__icon"))
+                @fragments.inlineSvg("arrow-right-double", "icon", List("liveblog-navigation__link--secondary--older__icon"))
             </a>
         </div>
     </div>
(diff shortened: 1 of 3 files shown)


static/src/stylesheets/module/content/live-blog/navigation.scss
```diff
@@ -42,82 +39,103 @@
...
+        height: 14px;
+        top: 11px;
     }
-}
 
-.link-navigation__link--secondary--newer {
-    margin-right: $gs-gutter / 4;
-    @include mq(mobileLandscape) {
-        margin-right: $gs-gutter / 2;
+    &--newer {
```
OliverJAsh: In the past we've avoided this syntax because it breaks searchability, i.e. you can no longer do a project wide search for `link-navigation__link--secondary--newer` and yield all the results.

zeftilldeath: Ok good to know, did wonder about that.

static/src/stylesheets/module/content/live-blog/navigation.scss
```diff
@@ -42,82 +39,103 @@
...
 .liveblog-navigation__link {
     display: inline-block;
+    position: relative;
+    color: colour(neutral-1);
+    background: transparent;
+    border: 1px solid colour(neutral-4);
+    height: $gs-baseline * 3;
+    line-height: 38px;
+    font-weight: bold;
+    vertical-align: middle;
```
OliverJAsh: Few questions here (sorry!):

Do you need `background: transparent` here?

Secondly, I wonder if it's possible to style this without hardcoding heights and line heights as magic numbers. This would be preferred because it's less brittle and thus less likely to break in the future, e.g. if the text size changes, you would have to remember to change this as well, I think?

static/src/stylesheets/module/content/live-blog/navigation.scss
```diff
@@ -42,82 +39,103 @@
...
+        border-color: colour(neutral-2);
         text-decoration: none;
     }
 
-    .inline-icon {
-        vertical-align: middle;
-        display: inline-block;
+    svg {
+        fill: colour(neutral-2);
+        position: absolute;
```
OliverJAsh: Out of interest, is this positioned absolutely in order to vertically align it correctly?

static/src/stylesheets/module/content/live-blog/navigation.scss
zeftilldeath: Removed `background: transparent`. Heights hardcoded to fit alongside other large buttons such as toast, The buttons will grow with a larger text size though :smiley:

static/src/stylesheets/module/content/live-blog/navigation.scss
zeftilldeath: Absolute gave me greater control for optical alignment with less minus margin hacky stuff. As the buttons are only ever going to be this size, it allows for less lines of css.

article/app/views/fragments/liveBlogNavigation.scala.html
```diff
@@ -45,7 +45,7 @@
                 href="/@{article.content.metadata.id}@oldest.suffix"
                 title="Go to page @oldest.pageNumber"
             }
-            class="liveblog-navigation__link link-navigation__link--secondary liveblog-navigation__link--secondary--older @if(pagination.oldest.isEmpty) { liveblog-navigation__link--disabled }"
+            class="liveblog-navigation__link link-navigation__link--secondary link-navigation__link--secondary--older @if(pagination.oldest.isEmpty) { liveblog-navigation__link--disabled }"
```
OliverJAsh: This should be `liveblog-navigation__link--secondary--older`, I just forgot to rename `link-navigation__link--secondary` to `liveblog-navigation__link--secondary`…

OliverJAsh:
Looks good! :+1:

article/app/views/fragments/liveBlogNavigation.scala.html
zeftilldeath: Thanks for clarifying that, updating.

zeftilldeath:
Thanks for all your help @OliverJAsh

PR #12132, opened 2016-03-01
Align items with the grid

regiskuckaertz:
## What does this change?

Items in paid containers were not aligned according to the grid:
picture 1

Into this:
picture 6

@jbreckmckye @uplne this is fairly urgent so it cannot wait our refactoring.


uplne:
beaut

:+1:

PR #12133, opened 2016-03-01
Sv enchilada article fix

uplne:
This PR removes some margins around enchilada format on content pages. It also reduces size on tablet  breakpoint. Simple but beautiful.

screen shot 2016-03-01 at 16 48 12
screen shot 2016-03-01 at 16 48 23


static/src/javascripts/projects/common/modules/commercial/gustyle/gustyle.js
```diff
@@ -33,12 +35,19 @@ define([
...
+            this.$slot.addClass(classList);
             this.$slot.prepend($.create(template(labelTpl, { data: merge(templateOptions) })));
 
             toggles = new Toggles(this.$slot[0]);
             toggles.init();
         }.bind(this));
     };
+
+    Gustyle.prototype.isContentPage = function () {
+        return !!config.page.isContent;
```
regiskuckaertz: Do we really want a level of indirection for that?

jbreckmckye: Generally speaking I do think a viewmodel should act as a proxy for everything the view depends upon. So I can see advantages as well as disadvantages.

regiskuckaertz: Yeah but I guess a property then would be more appropriate... somehow like you did for the commercial-features refactoring.

regiskuckaertz:
:+1: once you wage in on the property thing

PR #12137, opened 2016-03-02
make pagination link to the pagination rather than the first block

johnduffell:
When you use pagination on liveblogs, it will drop you at the pagination controls rather than the first block.  I've also fixed up the metadata in the head to link without the fragments.
When coming back to the first page, the original behaviour of dropping you at the top still applies.
comments welcome, and @OliverJAsh take a look!

### DIFF
--- a/article/app/views/fragments/liveBlogNavigation.scala.html
+++ b/article/app/views/fragments/liveBlogNavigation.scala.html
@@ -4,7 +4,7 @@
 @(article: Article, pageModel: LiveBlogPageModel[BodyBlock])
 
 @pageModel.pagination.map { pagination =>
-    <div class="liveblog-navigation">
+    <div id="liveblog-navigation" class="liveblog-navigation">
         <div class="liveblog-navigation__detail">
             <!-- div necessary for flexbox -->
             <div><span class="liveblog-navigation__page-number">@pageModel.currentPage.pageNumber</span> of <span class="liveblog-navigation__pages-length">@pagination.numberOfPages</span></div>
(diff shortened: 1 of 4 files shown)


article/app/views/fragments/liveBlogNavigation.scala.html
```diff
@@ -22,7 +22,7 @@
             </a>
```
OliverJAsh: Do you need to add it to the above as well?

article/app/views/liveblog/liveBlogBody.scala.html
```diff
@@ -81,8 +81,8 @@ <h1 itemprop="headline" class="content__headline js-score">@Html(article.trail.h
                                     </button>
                                 </div>
                             </div>
-                            @pageModel.pagination.map { pagination =>
-                                @if(!pagination.isNewestPage) {
+                            @if(pageModel.currentPage.isArchivePage) {
+                                @pageModel.pagination.map { pagination =>
```
OliverJAsh: You could get rid of this map because we do the same map in the fragment.

OliverJAsh:
Nice :+1:

article/app/views/fragments/liveBlogNavigation.scala.html
johnduffell: as discussed this is always page 1 which should navigate to the top, as there's no nagivation available to snap down to

article/app/views/liveblog/liveBlogBody.scala.html
johnduffell: good point, done!

PR #12138, opened 2016-03-02
Add trackAd module

regiskuckaertz:
## What does this change?

Adds a module allowing third-parties to detect when an advert has been rendered. Adds another module to detect when a rubicon ad is resized.

Promisifies the `modules:commercial:dfp:rendered` event, handles adding/removing event listeners and caches promises.
## Request for comment

@OliverJAsh can you please take a look at the sticky-ad changes, especially the rubicon ones (which I don't know how to test)

@JonNorman would like you to have a look at the changes in plista.js
## 


regiskuckaertz:
This design has a fairly obvious pitfall: it relies on the fact that the events haven't *already* been dispatched. But this is a problem that needs to be solved in our DFP API.

static/src/javascripts/projects/common/modules/commercial/track-ad.js
```diff
@@ -0,0 +1,41 @@
...
+                unlisten();
+                reject(new Error('Slot ' + id + ' was never loaded'));
+            };
+
+            function unlisten() {
+                mediator.off('modules:commercial:dfp:rendered', onAdLoaded);
+                mediator.off('modules:commercial:dfp:alladsrendered', onAllAdsLoaded);
+            }
+
+            mediator.on('modules:commercial:dfp:rendered', onAdLoaded);
```
jbreckmckye: You might be able to just use `mediator.once`

static/src/javascripts/projects/common/modules/commercial/track-ad.js
```diff
@@ -0,0 +1,41 @@
...
+            mediator.on('modules:commercial:dfp:alladsrendered', onAllAdsLoaded);
+        });
+
+        return promises[id];
+    };
+
+    return new TrackAd();
```
OliverJAsh: Would this be simplier if the module just returned `_.memoize(function (id) { })`?

jbreckmckye: We've written this with a constructor, but the way all the instances share the same `promises` map implies this could just as well be written as a singleton.

static/src/javascripts/projects/common/modules/commercial/track-rubicon-ad-resize.js
```diff
@@ -0,0 +1,60 @@
...
+) {
+    function trackRubiconAdResize(id) {
+        return trackAd.track(id)
+            .then(getIframeId)
+            .then(listenForRubicon)
+            .then(getDims);
+
+        function getIframeId(isLoaded) {
+            if (!isLoaded) {
+                return false;
```
OliverJAsh: Currently this fn returns either `Boolean | String`. Would it be simpler if the fn returned `String | undefined`? In Scala this would be modelled as `Option[String]`.

``` js
if (isLoaded) { return iframe.id }
```

OliverJAsh: That's a very minor point btw, just a thought exercise more than anything!

static/src/javascripts/projects/common/modules/commercial/track-rubicon-ad-resize.js
```diff
@@ -0,0 +1,60 @@
...
+                            data.value.id === iFrameId
+                        ) {
+                            resolve([data.value.width, data.value.height]);
+                            window.removeEventListener('message', onMessage);
+                        }
+                    };
+
+                    window.addEventListener('message', onMessage);
+                }) :
+                false;
```
OliverJAsh: This function returns either `Boolean | Promise[Object]`. It's nice when a function returns the same type for consistency and so there are no surprises, but here I guess it doesn't really matter!

OliverJAsh:
Really nice work :+1: It's great to extract common functionality :smile:

static/src/javascripts/projects/common/modules/commercial/third-party-tags/plista.js
```diff
@@ -74,9 +76,10 @@ define([
         if (shouldServe()) {
             if (loadInstantly()) {
                 module.load();
+                return Promise.resolve(true);
             } else {
-                mediator.on('modules:commercial:dfp:rendered', function (event) {
-                    if (event.slot.getSlotElementId() === 'dfp-ad--merchandising-high' && event.isEmpty) {
+                return trackAd.track('dfp-ad--merchandising-high').then(function (isLoaded) {
```
jbreckmckye: Would a method name like, I don't know, `waitFor`, read a little better?

```javascript
trackAd.waitFor('dfp-ad--merchandising-high').then(...)
```

static/test/javascripts/spec/common/commercial/third-party-tags/outbrain.spec.js
```diff
@@ -164,49 +179,44 @@ define([
...
+                    expect(args[0]).toBe(true);
+                    expect(args[1]).toBe(false);
                     done();
                 });
-                mediator.emit('modules:commercial:dfp:rendered', eventStub).then(function () {
-                    mediator.emit('modules:commercial:dfp:rendered', eventStubLo);
-                });
-
-                mediator.emit = oldEmit;
+                // .then(sut.init)
```
jbreckmckye: Should this be deleted?

static/src/javascripts/projects/common/modules/commercial/third-party-tags/plista.js
regiskuckaertz: that is a very nice suggestion. Taking it!

static/test/javascripts/spec/common/commercial/third-party-tags/plista.spec.js
```diff
@@ -111,12 +111,14 @@ define([
                 fixtures.render(fixturesMerch);
 
                 spyOn(sut, 'load');
-                sut.init();
+                var p = sut.init();
```
jbreckmckye: What's a `p`? Plista?

static/src/javascripts/projects/common/modules/commercial/track-ad.js
regiskuckaertz: The initial version was a function. I'm using the constructor form so that the cache can be reset during testing. This is not great since I am against any use of classes in JS, but it's that or adding a reset function. Considering a second parameter `force` that would be false by default though.

Thoughts?

static/src/javascripts/projects/common/modules/commercial/track-rubicon-ad-resize.js
regiskuckaertz: Nope, you're right. I wouldn't be surprised if there was some optimisation done when parameters and return values remain consistent.

static/src/javascripts/projects/common/modules/commercial/track-rubicon-ad-resize.js
regiskuckaertz: Good point. Would you be good with `Option[Promise[Object]]`? In a promise chain, I prefer it if a callback can return a value rather than a promise that resolves to a value, to avoid one page in the scheduler.

static/test/javascripts/spec/common/commercial/third-party-tags/outbrain.spec.js
regiskuckaertz: oops, nope :sweat_smile:

static/src/javascripts/projects/common/modules/commercial/track-ad.js
OliverJAsh: Ahh, I see. Your module could return a closure so you can easily get a new copy of the memoized function:

``` js
return () => { return memoize((id) => {} }
```

static/src/javascripts/projects/common/modules/commercial/track-rubicon-ad-resize.js
OliverJAsh: I guess the choices are `Promise[Object | undefined]` or `undefined | Promise[Object]`. The latter means we would return two different types, but that's due to the lack of Options in JS, and if you don't do it there then you have to do it inside of the promise instead.

static/src/javascripts/projects/common/modules/commercial/track-ad.js
regiskuckaertz: Yes, I suppose there must be some indirection. Unit tests be damned :disappointed:

regiskuckaertz: Not sure the closure would work in fact. If you call it from two different modules, you lose all previous calls.

regiskuckaertz: I think the optional parameter is a more useful option. I can imagine a future where we would reload a slot (e.g. in ajax-based pages, etc.)

OliverJAsh: Ah yes. In our tests we do use Squire which is supposed to help here, allowing you to import a module into multiple contexts, each with their own state.

regiskuckaertz: wOOt. Can you show me an example?

On 2 March 2016 at 15:30, Oliver Joseph Ash <notifications@github.com>
wrote:

> In static/src/javascripts/projects/common/modules/commercial/track-ad.js
> <https://github.com/guardian/frontend/pull/12138#discussion_r54738480>:
>
> > +                reject(new Error('Slot ' + id + ' was never loaded'));
> > +            };
> > +
> > +            function unlisten() {
> > +                mediator.off('modules:commercial:dfp:rendered', onAdLoaded);
> > +                mediator.off('modules:commercial:dfp:alladsrendered', onAllAdsLoaded);
> > +            }
> > +
> > +            mediator.on('modules:commercial:dfp:rendered', onAdLoaded);
> > +            mediator.on('modules:commercial:dfp:alladsrendered', onAllAdsLoaded);
> > +        });
> > +
> > +        return promises[id];
> > +    };
> > +
> > +    return new TrackAd();
>
> Ah yes. In our tests we do use Squire which is supposed to help here,
> allowing you to import a module into multiple contexts, each with their own
> state.
>
> —
> <https://github.com/guardian/frontend/pull/12138/files#r54738480>.
>



-- 
–Regis

-- 

in England Number 908396

jbreckmckye: How are you going to reset the cache during testing? The module will always have the same promise map, am I correct? Are you trying to reset the cache between tests? It is tricky - this'd be a lot simpler with a service layer ;-)

I guess you could put a `reset` method on your class. It's a bit awkward, but I'm not sure if there are any better alternatives at present.

The only other way to reset the map between specs might be to make the injector mock the module with itself. I am not sure if this will work, and is quite a bulky solution in any case:

```javascript
injector.require(['tracker'], function () {
    injector.mock('tracker', new arguments[0].constructor());
});
```

OliverJAsh: I think if you do:

``` js
injector.require(['tracker'], function (trackerA) {
});

injector.require(['tracker'], function (trackerB) {
});
```

`trackerA` and `trackerB` will be separate instances?

static/src/javascripts/projects/common/modules/commercial/track-rubicon-ad-resize.js
jamesgorrie: `Promise[Object | undefined]` which equates to Promise[Option[Object]]` would make sense.
You could then:
```JS
  if (frameId) { return undefined } // None
  else {
    return { magic: true }
  }
});
```

PR #12139, opened 2016-03-02
Add logging to help understand PA errors

TBonnin:
## What does this change?

Add logging and metric to better understand why Football scores sometimes don't update, especially if/when PA API returns errors in xml reponses

Next time the PA API stops returning up-to-date data, we will be able to see the type of errors they return and communicate them to their customer support
The new metric will also allow us to assess how often those errors happen. Until now we have been relying on readers or football desk to report this issue

Nope
## Request for comment

@rich-nguyen @JustinPinner

### DIFF
--- a/common/app/common/metrics.scala
+++ b/common/app/common/metrics.scala
@@ -188,7 +188,12 @@ object PaMetrics {
     "AP api returned error"
   )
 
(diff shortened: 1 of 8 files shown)


rich-nguyen:
:+1:

JustinPinner:
Looks :+1: to me

PR #12140, opened 2016-03-02
Bugfix: Move header back into hasHeader check

gtrufitt:
## What does this change?

Fixes immersive headers so we check the hasHeader metadata.
## Request for comment

@OliverJAsh @johnduffell


johnduffell:
looks good to me :yum:  :+1: thanks Gareth!

PR #12141, opened 2016-03-02
Add email name to footer text

desbo:
With this, if you're looking at an email version of an article, and that article meets the criteria for a particular email, we'll show a bit of text in the footer to explain that you've received it because you follow (for example) The Fiver.


gtrufitt:
:+1:

PR #12142, opened 2016-03-02
Amp multiple tweaks

zeftilldeath:
More AMP tweaks.

Review tone in OJs now showing correctly. 

Changed `top` to `padding` on the logo wrapper so when a user hits back to top they go to the top, not 3px off.

Prevented forced bold version of Egyptian showing when `<strong>` tags are wrapped within h2s etc. 

Added a clear to the comments/meta section.

### DIFF
--- a/common/app/views/fragments/amp/stylesheets/fonts.scala.html
+++ b/common/app/views/fragments/amp/stylesheets/fonts.scala.html
@@ -43,3 +43,7 @@
     font-family: "Guardian Egyptian Web", Georgia, serif;
     font-weight: 500;
 }
+
+b, strong {
+    font-weight: 500;
+}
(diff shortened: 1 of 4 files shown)


common/app/views/fragments/amp/stylesheets/fonts.scala.html
```diff
@@ -43,3 +43,7 @@
     font-family: "Guardian Egyptian Web", Georgia, serif;
     font-weight: 500;
 }
+
+b, strong {
```
jamesgorrie: Are we sure this doesn't affect anything anywhere else on the site?

jamesgorrie: ah, I just realised these are in `amp/fragments`...

jamesgorrie:
:+1:

PR #12145, opened 2016-03-02
Add initial AMPScript tags to emails

desbo:
I have no idea if this will work and the only way to find out is to deploy it!


gtrufitt:
:+1:

PR #12147, opened 2016-03-02
Minor change when commercial is switched off

regiskuckaertz:
There's no need to add slots everywhere if they are not to be filled.


uplne:
:+1:

PR #12148, opened 2016-03-02
Email template: Add icon layout and 100% width to image

gtrufitt:
Minor bugfix to improve layout of icons and fix image width


desbo:
:+1:

PR #12152, opened 2016-03-02
Revert "Add trackAd module"

regiskuckaertz:
The Promise in article-body-adverts will not resolve until the inline slot is loaded. Problem is, dfp waits for it before making any ad call... A deadlock.


uplne:
:+1:

uplne:
giphy 2

OliverJAsh:
Hope this makes it back in? :+1:

regiskuckaertz:
@OliverJAsh yeah I'll do that as soon as I have time. deadlocks are always a fun riddle :)

PR #12153, opened 2016-03-02
Email template: Add link and title attribute to header

gtrufitt:
## What does this change?

Wraps the header of emails in an anchor, adding title attr (and alt on image) so that when hovered it says "View _email name_ online and links to the website


jamesgorrie:
:+1:

PR #12154, opened 2016-03-03
Add the morning briefing series to the blacklist

gtrufitt:
## What does this change?

Makes sure the non-series version of the morning briefing email sign-up doesn't appear in the series articles.

It has different wording and positioning that fits with the series better.


TBonnin:
:+1:

PR #12159, opened 2016-03-03
defer binding of video events to analytics

jamesgorrie:
# Video analytics
## What does this change?

It seems `bindGlobalEvents` doesn't just bind, but triggers events. Who'd have thunk it.
Tested against what we were fixing last time and that works too.

The reason for not sorting this out properly is the larger discussions about the player.

:dancers: 

:hatched_chick: 
## Screenshots

:gun: 
## @akash1810 


akash1810:
:+1:

PR #12160, opened 2016-03-03
rss: compile regex pattern only once

TBonnin:
## What does this change?

Small performance improvement: Instead of compiling the regex pattern every time `stripInvalidXMLCharacters` is called, we now compile it only once and reuse it
=> divide the number of time regex compilation happens by 3

Nope
## Request for comment

@rich-nguyen


rich-nguyen:
:+1:

PR #12163, opened 2016-03-03
Add series name to series emails without a custom header

gtrufitt:
## What does this change?

Adds the series name to emails without a custom header.
### Series, no custom header.

image
### Series, custom header

image
### Not a series

image


desbo:
:+1:

PR #12173, opened 2016-03-04
This adds ipod device type and devices on iOS7 to is Older Device

michaelwmcnamara:
Since mobile aggregator got switched off we have had a bunch of users on older devices trying to access the web and crashing. Many cant even access the content page with the link to switch to core.
This change incorporates iPods into our old device checks.
It also adds iOS7 into the older device bin


johnduffell:
go for it! :+1:

johnduffell:
good catch scott :speedboat:

Issue #12053, opened 2016-02-22
Amp - Recommended content does not run on firefox


michaelwmcnamara:
On amp content recommended content containers are not displaying - scroll below the content and you are left with a screen of loading "..."'s (see screenshot).
In the last month firefox has accounted for 0.8% of mobile phone visits (1 million out of 132 million).
In terms of usage this is very small, but the visual impact is pretty ugly - one to fix when we get to it.

Example url: https://amp.theguardian.com/us-news/2016/feb/21/bill-clinton-monica-lewinsky-hillary-president

screenshot:
![image](https://cloud.githubusercontent.com/assets/7239344/13225186/acb134c4-d983-11e5-9b0f-9253fb09d6ef.png)

TBonnin:
FYI errors in console :(
![screen shot 2016-02-23 at 10 18 40](https://cloud.githubusercontent.com/assets/233326/13248516/acd65072-da16-11e5-8f8c-7e38c43782f6.png)

NataliaLKB:
This has been fixed:

![image](https://cloud.githubusercontent.com/assets/8774970/13603297/170dfd3c-e534-11e5-8819-45511e3e881c.png)
(on firefox)

PR #12176, opened 2016-03-07
Credentials on Amp List

NataliaLKB:
# Credentials on Amp List

Take Two. First attempt here: https://github.com/guardian/frontend/pull/12022 (which I had to revert)
## What does this change?
- Adds amp specific response headers 
## This PR should fix:
- https://github.com/guardian/frontend/issues/12053
- Cookies should be sent with a request (cookie based editionalisation perhaps? :open_mouth: )

This should only affect AMP
## Request for comment

@TBonnin can you take a look? Happy to discuss this with you :). Would really like to get your feedback

### DIFF
--- a/common/app/conf/Filters.scala
+++ b/common/app/conf/Filters.scala
@@ -67,6 +67,19 @@ object SurrogateKeyFilter extends Filter with ExecutionContexts {
   }
 }
 
+object AmpFilter extends Filter with ExecutionContexts with implicits.Requests {
+  override def apply(nextFilter: (RequestHeader) => Future[Result])(request: RequestHeader): Future[Result] = {
+    if (request.isAmp) {
+      val domain = "https://" + request.domain
+      val ampHeader = "AMP-Access-Control-Allow-Source-Origin" -> Configuration.amp.corsOrigins.find(_ == domain).getOrElse("*")
+
+      nextFilter(request).map(_.withHeaders(ampHeader))
+    } else {
+      nextFilter(request)
+    }
+  }
+}
+
 object Filters {
   // NOTE - order is important here, Gzipper AFTER CorsVaryHeaders
   // which effectively means "JsonVaryHeaders goes around Gzipper"
@@ -76,6 +89,7 @@ object Filters {
     BackendHeaderFilter,
     RequestLoggingFilter,
     GNUFilter,
-    SurrogateKeyFilter
+    SurrogateKeyFilter,
+    AmpFilter
   )
 }
--- a/common/app/conf/Filters.scala
+++ b/common/app/conf/Filters.scala
@@ -70,14 +70,7 @@ object SurrogateKeyFilter extends Filter with ExecutionContexts {
 object AmpFilter extends Filter with ExecutionContexts with implicits.Requests {
   override def apply(nextFilter: (RequestHeader) => Future[Result])(request: RequestHeader): Future[Result] = {
     if (request.isAmp) {
-      val domain = request.headers.get("Origin") match {
-        case None => {
-          "https://" + request.domain
-        }
-        case Some(requestOrigin) => {
-          requestOrigin
-        }
-      }
+      val domain = request.headers.get("Origin").getOrElse("https://" + request.domain)
       val exposeAmpHeader = "Access-Control-Expose-Headers" -> "AMP-Access-Control-Allow-Source-Origin"
       val ampHeader = "AMP-Access-Control-Allow-Source-Origin" -> Configuration.amp.corsOrigins.find(_ == domain).getOrElse("*")
 
--- a/common/app/conf/Filters.scala
+++ b/common/app/conf/Filters.scala
@@ -70,10 +70,18 @@ object SurrogateKeyFilter extends Filter with ExecutionContexts {
 object AmpFilter extends Filter with ExecutionContexts with implicits.Requests {
   override def apply(nextFilter: (RequestHeader) => Future[Result])(request: RequestHeader): Future[Result] = {
     if (request.isAmp) {
-      val domain = "https://" + request.domain
+      val domain = request.headers.get("Origin") match {
+        case None => {
+          "https://" + request.domain
+        }
+        case Some(requestOrigin) => {
+          requestOrigin
+        }
+      }
+      val exposeAmpHeader = "Access-Control-Expose-Headers" -> "AMP-Access-Control-Allow-Source-Origin"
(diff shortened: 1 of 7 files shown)


common/app/conf/Filters.scala
```diff
@@ -67,6 +67,27 @@ object SurrogateKeyFilter extends Filter with ExecutionContexts {
...
+  override def apply(nextFilter: (RequestHeader) => Future[Result])(request: RequestHeader): Future[Result] = {
+    if (request.isAmp) {
+      val domain = request.headers.get("Origin") match {
+        case None => {
+          "https://" + request.domain
+        }
+        case Some(requestOrigin) => {
+          requestOrigin
+        }
+      }
```
TBonnin: could you use a more idiomatic `request.headers.get("Origin").fold("https://" + request.domain)(_)` here?
Or `request.headers.get("Origin").map(_).getOrElse("https://"+request.domain)'

A good read about Option matching: http://blog.originate.com/blog/2014/06/15/idiomatic-scala-your-options-do-not-match/

common/app/conf/Filters.scala
```diff
@@ -67,6 +67,27 @@ object SurrogateKeyFilter extends Filter with ExecutionContexts {
...
+      val exposeAmpHeader = "Access-Control-Expose-Headers" -> "AMP-Access-Control-Allow-Source-Origin"
+      val ampHeader = "AMP-Access-Control-Allow-Source-Origin" -> Configuration.amp.corsOrigins.find(_ == domain).getOrElse("*")
```
TBonnin: I know `*` is set as a fallback in case the domain cannot be found in our config. But given this is not recommended should we log/track if this happens. That could allow us to monitor it and add a domain in case Google makes any changes for instance

common/app/model/Cors.scala
```diff
@@ -1,24 +1,27 @@
...
 
     val responseHeaders = (defaultAllowHeaders ++ request.headers.get("Access-Control-Request-Headers").toList) mkString ","
+    val corsOrigins = if(request.isAmp) amp.corsOrigins else ajax.corsOrigins
 
     request.headers.get("Origin") match {
       case None => result
       case Some(requestOrigin) => {
+        val allowedOrigin = corsOrigins.find(_ == requestOrigin).getOrElse("*")
         val headers = allowedMethods.map("Access-Control-Allow-Methods" -> _).toList ++ List(
-          "Access-Control-Allow-Origin" -> corsOrigins.find(_ == requestOrigin).getOrElse("*"),
```
TBonnin: Same comment as above

common/app/model/Cors.scala
```diff
@@ -1,24 +1,27 @@
...
 
   def apply(result: Result, allowedMethods: Option[String] = None)(implicit request: RequestHeader): Result = {
 
     val responseHeaders = (defaultAllowHeaders ++ request.headers.get("Access-Control-Request-Headers").toList) mkString ","
+    val corsOrigins = if(request.isAmp) amp.corsOrigins else ajax.corsOrigins
 
     request.headers.get("Origin") match {
       case None => result
       case Some(requestOrigin) => {
+        val allowedOrigin = corsOrigins.find(_ == requestOrigin).getOrElse("*")
```
TBonnin: Same comment as above

common/conf/env/PROD.properties
```diff
@@ -23,6 +23,7 @@ ajax.cors.origin=http://www.theguardian.com,https://www.theguardian.com,https://
 
 amp.host=amp.theguardian.com
 amp.scheme=https://
+amp.cors.origin=https://www.theguardian.com,https://amp.theguardian.com, https://cdn.ampproject.org
```
TBonnin: Did Google/AmpProject confirm amp content would always be served from `https://cdn.ampproject.org`

TBonnin:
I made a few tiny comments but overall it looks good to me :+1:

common/app/conf/Filters.scala
NataliaLKB: Great blog post and good suggestion. Will update.

common/app/conf/Filters.scala
NataliaLKB: Yes that makes sense. Not sure how to track this, any ideas? Otherwise I will ask around :smile:

common/conf/env/PROD.properties
NataliaLKB: Yes they did, we are reliant on it for our fonts as well. Other publishers are reliant on it as well so they will inform us if anything changes.

TBonnin: Got it

common/app/conf/Filters.scala
TBonnin: Logging it (and then create a filter in awslog) could be a start.

common/app/conf/Filters.scala
```diff
@@ -67,6 +67,20 @@ object SurrogateKeyFilter extends Filter with ExecutionContexts {
   }
 }
 
+object AmpFilter extends Filter with ExecutionContexts with implicits.Requests {
+  override def apply(nextFilter: (RequestHeader) => Future[Result])(request: RequestHeader): Future[Result] = {
+    if (request.isAmp) {
+      val domain = request.headers.get("Origin").fold("https://" + request.domain)(_)
```
johnduffell: is this folding over an option? I had to google to work out what that even means :confounded: but apparently you could just just change to something like ```
val domain = request.headers.get("Origin").getOrElse("https://" + request.domain)
```

File: common/conf/env/CODE.properties [hunk 7]
```diff
@@ -25,6 +25,7 @@ oas.siteId.host=m.code.dev-theguardian.com
 
 amp.host=amp.code.dev-theguardian.com
 amp.scheme=https://
+amp.cors.origin=https://m.code.dev-theguardian.com,https://amp.code.dev-theguardian.com, https://cdn.ampproject.org
```
johnduffell: if it's a list, consider plural names?

johnduffell:
look sgood apart from that :+1:

common/conf/env/CODE.properties
NataliaLKB: I would rather keep consistent with what we have had before: https://github.com/guardian/frontend/blob/credentials-amp-list/common/conf/env/CODE.properties#L22

Also, only one every gets used at once, so in a way they are just all options for one origin.

common/app/conf/Filters.scala
NataliaLKB: I like fold, but you are right that it isn't the most intuitive. Happy to change!

PR #12177, opened 2016-03-07
Fixed alignments and colours in paid containers

regiskuckaertz:
Paid containers are like black holes for styled items. There was some colour and grid alignment issues that are fixed in this PR.

Before:
picture 1

After:
picture 2


gtrufitt:
:+1:

PR #12179, opened 2016-03-07
Add shouldHideAdverts check to email sign-up

gtrufitt:
## What does this change?

Adds a check for `shouldHideAdverts` config for email sign-up

This prevents us showing the email sign-up promo on sensitive or inappropriate content


rich-nguyen:
:+1:

PR #12181, opened 2016-03-07
Move all email sign-ups to bottom of article except film

gtrufitt:
## What does this change?

Moves all email sign-ups to the bottom of articles instead of mid article except for film today.

Lessens the annoyance of editors and users for having a big sign-up mid-article until we design a smaller one. We are leaving film today as we need to compare the CTR to outbrain for bottom-of-article sign-ups.
## Request for comment

ping @crifmulholland


rich-nguyen:
:+1:

PR #12182, opened 2016-03-07
Separate jobs feed reading and parsing switches

kelvin-chappell:
## What does this change?

Separate switches for reading jobs from external source, and parsing job instances for the merchandising component.

We can stop trying to read a known broken feed without bringing down the component.

Don't think other platforms will be affected.

/cc @kenlim


kenlim:
:+1:

PR #12183, opened 2016-03-07
Remove --production flag

rich-nguyen:
The npm install --production flag keeps installing devDependencies. Removing for quick fix, should take 30-50 seconds off the build


desbo:
:+1:

PR #12185, opened 2016-03-07
Encode raw image uri strings when using ImgSrc

rich-nguyen:
Fixes issues with image asset urls that are not encoded. Here's one:

http://www.theguardian.com/football/2009/jun/03/newcastle-united-search-buyer-alan-shearer

The raw image url was being used by the facebook open graph field `rawOpenGraphImage` in the `Content` class, during construction.

@OliverJAsh


sndrs:
:+1:

PR #12186, opened 2016-03-07
only autoplay on liveblogs & articles

jamesgorrie:
# Video autoplay

Turns out I didn't really understand how the framework runs, `canRun` I thought was whether to run the test, but is actually can the person be put into the test. This stops people, who are in the test, have the autoplay code run on Video pages and fronts. 
## What does this change?

:arrow_up:
## Request for comment

@johnduffell & @akash1810 
## 

### DIFF
--- a/static/src/javascripts/bootstraps/enhanced/media/main.js
+++ b/static/src/javascripts/bootstraps/enhanced/media/main.js
@@ -326,7 +326,9 @@ define([
                         resolve();
                     }
 
-                    if (abVideoAutoplay.variant === 'autoplay') {
+                    var contentType = config.page.contentType;
+                    var isArticleOrLiveBlog = contentType === 'Article' || contentType === 'LiveBlog';
+                    if (abVideoAutoplay.variant === 'autoplay' && isArticleOrLiveBlog) {
                         // Annoyingly we pass the `parentNode` as the video is absolutely positioned.
                         var parentNode = player.el().parentNode;
                         var firstAutoplay = true;
--- a/common/app/conf/switches/ABTestSwitches.scala
+++ b/common/app/conf/switches/ABTestSwitches.scala
@@ -36,7 +36,7 @@ trait ABTestSwitches {
     "ab-article-video-autoplay",
     "Autoplay embedded videos in article",
     safeState = Off,
-    sellByDate = new LocalDate(2016, 3, 21),
+    sellByDate = new LocalDate(2016, 3, 22),
     exposeClientSide = true
   )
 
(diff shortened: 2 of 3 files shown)


static/src/javascripts/bootstraps/enhanced/media/main.js
```diff
@@ -326,7 +326,9 @@ define([
                         resolve();
                     }
 
-                    if (abVideoAutoplay.variant === 'autoplay') {
+                    var contentType = config.page.contentType;
+                    var isArticleOrLiveBlog = contentType === 'Article' || contentType === 'LiveBlog';
+                    if (abVideoAutoplay.variant === 'autoplay' && isArticleOrLiveBlog) {
```
johnduffell: actually that's not enough, you have to use isVariant('abVideoAutoplay', 'autoplay') to make sure you do all right the checks

jamesgorrie: Strangely I would prefer not to do this as this also requires the switch to be on - which I don't want to do until I have run it past a few people. Having this would stop the `#ab-TESTNAME` from working.

jamesgorrie: Happy to relook at this in a different PR where we can figure out where we think this is meant ot work.

johnduffell: if it's just for user testing/messing around then don't worry too much, however when you actually want to use the test data you have to be more careful as it's easy to much up the samples.  A good tip is to rename the test before you do the official testing so it doesn't pick up any data from people who were bucketed by old versions of the code.

akash1810:
:+1:

static/src/javascripts/bootstraps/enhanced/media/main.js
jamesgorrie: Brilliant - it would be good to find a way to force someone into a test though, even while using the `isInVariant`.

akash1810: > it would be good to find a way to force someone into a test

The [chrome extension](https://chrome.google.com/webstore/detail/ab-tests/nehbenedinjacnhlkjbdneedibcagmno) can be used to force yourself into a test.

johnduffell: isInVariant it just using the same logic it uses to run the code in your module.  Often the code in the module is empty, but if the test is off/expired but you are in the bucket, you can end up with some weird situation where you're half in the test.  Also the code for updating your local storage relies on the switch etc.  So I think in reality the code needs updating/refactoring to support test mode where you are in the test on the client side but it doesn't send back analytics

PR #12191, opened 2016-03-07
Header bidding test for US edition

jbreckmckye:
If you read the US edition of www.theguardian.com, you will have a 10% chance of seeing adverts selected by 'header bidding'. This means that the browser will ask third parties what they can pay for an advert, send the details of the best bid up to DFP and then await its decision whether to award the adslot.

image

We believe that if we can tender adverts to more customers, we have a higher chance of finding more lucrative deals. We'll need to examine the data carefully, though, and make sure that our higher revenues aren't offset by poorer advert loading times (which harms impressions).
## Application flow
1. If the user is part of the A/B test, initialize a PrebidService that calls the Prebid.js header-bidding vendor library.
2. For all lazy-loaded adverts, load the adslot using the prebid service
3. The service will load only one advert at a time. Prebid.js can run only one auction at a time. We can tender multiple adverts per auction, but because we must make multiple HTTP requests per advert, this would run the risk of congesting the browser's network resources.
4. For any advert, Prebid will ask one third party (at this time) - Rubicon - the best price it can provide for the current advert, given its keywords and sizes.
5. We then add the details of the bidder, the price and the creative ID onto the slot's targeting, so it can be sent to DFP. The price is floored to a bucket, so that it can be matched against a particular DFP line item that represents a header bidding result for a particular partner, at a particular price point (e.g. Rubicon, CPM = $12.00)
6. DFP compares the value of the matching line item to the value of other running campaigns. It might decide not to show the winning bid if there is a roadblock campaign, or another high-value campaign is running behind.
7. If the auction winner is chosen, DFP returns a 'dummy creative', a creative containing nothing more than a script which tells the Prebid API on the client it can go ahead and display the advert.
8. Prebid fetches and renders the creative.
## Notes
- The Prebid API is proxied by a PrebidService that loads its dependencies, and handles the vagaries of setting up prebid auctions
- Auction functions are wrapped in Robust checks, so that errors can be reported in Sentry
- If an auction takes more than 2000 ms (for instance, because a third party service is down), we will short-circuit the request and fallback to showing either the best response it has so far, or an ordinary DFP advert
- Pageskin roadblock adverts and high-relevance merchandising components are excluded from header bidding
## RFC


static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
```diff
@@ -471,34 +485,43 @@ define([
...
+            var depth = 0.5;
+
+            var advertsToLoad = values(adverts).filter(function (advert) {
+                return !advert.isRendered
+                    && !advert.isLoading
+                    // if the position of the ad is above the viewport - offset (half screen size)
+                    && (scrollBottom > document.getElementById(advert.adSlotId).getBoundingClientRect().top + scrollTop - viewportHeight * depth);
+            });
+
+            advertsToLoad.forEach(function (advert) {
```
regiskuckaertz: `advertsToLoad.forEach(loadSlot)` :smiley:

static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
```diff
@@ -471,34 +485,43 @@ define([
...
+                loadSlot(advert);
             });
         }
     }
 
     function instantLoad() {
-        chain(adverts).and(keys).and(forEach, function (adSlotId) {
-            if (contains(['dfp-ad--pageskin-inread', 'dfp-ad--merchandising-high', 'dfp-ad--im'], adSlotId)) {
-                loadSlot(adSlotId);
+        values(adverts).forEach(function (advert) {
```
regiskuckaertz: It's great to finally see array functions used instead of the lodash counterparts. Why not go all the way and use: `Object.keys(adverts).forEach(...)`? I feel like `values()` is one we could get rid of very easily.

static/src/javascripts/projects/common/modules/commercial/dfp/PrebidService.js
```diff
@@ -0,0 +1,148 @@
...
+            }
+        }];
+    }
+
+    function getAllValidSlotSizes(advert) {
+        var sizesForAllBreakpoints = concatAll(values(advert.sizes));
+        return uniq(sizesForAllBreakpoints, arrayValue);
+
+        function concatAll(arrays) {
+            return Array.prototype.concat.apply([], arrays);
```
uplne: Why not concat from lodash?

jbreckmckye: It's not in the lodash version we bundle (2.4.1?)

static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
jbreckmckye: Good point - let's try and reduce the size of our dependency tree

static/src/javascripts/projects/common/modules/commercial/dfp/PrebidService.js
```diff
@@ -0,0 +1,148 @@
...
+                    if (isNewAdvert) {
+                        pbjs.addAdUnits([adUnits[adSlotId]]);
+                    }
+                    pbjs.requestBids({
+                        timeout : PREBID_TIMEOUT,
+                        adUnits : [adUnits[adSlotId]],
+                        bidsBackHandler : function () {
+                            try {
+                                pbjs.setTargetingForGPTAsync([adSlotId]);
+                                window.googletag.display(adSlotId);
```
regiskuckaertz: Don't you need to update the data structure in dfp-api somehow? i.e. set the `isLoading` and `isRendered` accordingly?

static/src/javascripts/projects/common/modules/commercial/dfp/PrebidService.js
regiskuckaertz: I like this line :smile: I would have written `arrays.reduce((res, input) => res.concat(input), [])` but it is harder to read.

static/src/javascripts/projects/common/modules/commercial/dfp/PrebidService.js
jbreckmckye: DFP is still rendering the advert itself, so `slotRenderEnded` still fires.

static/src/javascripts/projects/common/modules/commercial/dfp/PrebidService.js
jbreckmckye: I always find reductions a little opaque when they're used to transform a data structure.

regiskuckaertz: indeed that's what `mapcat` is for

static/src/javascripts/projects/common/modules/commercial/dfp/PrebidService.js
regiskuckaertz: :+1:

static/src/javascripts/projects/common/modules/commercial/dfp/QueueAsync.js
```diff
@@ -0,0 +1,19 @@
...
+    'Promise'
+], function (
+    Promise
+) {
+    /**
+     * Queue up asynchronous functions to run one-at-a-time.
+     * When the first function's promise resolves, the second runs, and so on.
+     * If any function throws an error, that is captured by the errorHandler.
+     */
+    return function QueueAsync(errorHandler) {
```
Calanthe: I like the use of this function!

regiskuckaertz:
This is looking really good, great work @jbreckmckye !! I'll give it a +1 as soon as the tests pass.

jbreckmckye:
This all seems to be passing, now.

regiskuckaertz:
:+1:

static/src/javascripts/projects/common/modules/commercial/dfp/QueueAsync.js
regiskuckaertz: Yes, there is potential for reuse in spacefiller I guess

PR #12192, opened 2016-03-07
Taking advantage of empty ad space on AMP

NataliaLKB:
# Taking advantage of empty ad space on AMP
## What does this change?

All amp articles have spaces in the article body that are reserved for an ad. If an ad takes a while to load, or doesn't load at all (no js or ad blocker) than all the user sees is a blank grey space. 

We should take advantage of this space, and use it as a branding opportunity! If people like it, lets do it on the site as well :smile: 

Just affects AMP
## Screenshots

image
## Request for comment

@stephanfowler


stephanfowler:
:+1: yes!

johnduffell:
really clever! :+1:

common/app/views/fragments/amp/stylesheets/ads.scala.html
```diff
@@ -1,5 +1,8 @@
...
...kxheWVyXzEiIHhtbG5zPSJodHRwOi8vd3d3LnczLm9yZy8yMDAwL3N2ZyIgeG1sbnM6eGxpbms9Imh0dHA6Ly93d3cudzMub3JnLzE5OTkveGxpbmsiIHg9IjBweCIgeT0iMHB4IiB2aWV3Qm94PSIwIDAgODYgNzYiIGVuYWJsZS1iYWNrZ3JvdW5kPSJuZXcgMCAwIDg2IDc2IiB4bWw6c3BhY2U9InByZXNlcnZlIj48cGF0aCBmaWxsPSIjZGNkY2RjIiBkPSJNNTAuOCwxNS45YzAtMTEuOC0zLjUtMTMuNi03LjgtMTMuNmMtNC4yLDAtNy43LDEuNi03LjcsMTMuNmMwLDEyLDMuNSwxMyw3LjcsMTNDNDcuMiwyOC45LDUwLjgsMjcuNyw1MC44LDE1LjkgTTM1LjMsNjAuOGMtNS40LDAtNyw0LTcsNi45YzAsNC4zLDMuOCw4LjIsMTUsOC4yYzEyLjcsMCwxNi4yLTMuNiwxNi4yLTguMmMwLTQuMS0zLjEtNi45LTguMi02LjlIMzUuM3ogTTI1LjEsMC41QzEwLjMsNy4zLDAsMjIuMywwLDM5LjZDMCw1MS40LDQuNyw2MiwxMi40LDY5LjhjMC0wLjIsMC0wLjQsMC0wLjZjMC03LjcsNy41LTEwLjUsMTQuMS0xMS45djBjLTYuMy0xLjQtOS40LTUuOS05LjQtMTAuNWMwLTYuMiw2LjktMTEuNSwxMC4zLTEzLjlMMjcsMzIuN2MtNS45LTMuMy05LjgtOS05LjgtMTYuOEMxNy4yLDkuMiwyMC4xLDQsMjUuMSwwLjUgTTg2LDM5LjZDODYsMjEuOSw3NS4zLDYuNyw2MCwwLjFjNS4xLDMuMyw4LjMsOC41LDguNSwxNWMwLDAuNiwwLjEsMSwwLjEsMS40YzAsMTIuOC0xMC40LDE5LjYtMjUuNiwxOS42Yy0zLjksMC02LjUtMC4yLTkuNy0xLjJjLTEuNSwxLTIuNiwyLjctMi42LDQuM2MwLDIuMSwxLjksMy44LDQuMiwzLjhoMjEuMWMxMy4xLDAsMTkuNSw1LjQsMTkuNSwxNi45YzAsMy45LTAuOCw3LjMtMi40LDEwLjNDODEsNjIuNSw4Niw1MS42LDg2LDM5LjYiLz48L3N2Zz4=);background-size: 105px;
```
TBonnin: The concern about base64 url is that browsers won't cache the image and would have to decode it every time. Does the network saving worth the loading cost (especially on mobile devices)? I don't know though

NataliaLKB: Not sure which is the best solution :confused: . I will look into it before adding any more things like this to the AMP pages

PR #12196, opened 2016-03-08
stop mobile chrome jumping on immersives

johnduffell:
Current situation - as the address bar switches in and out the vh changes and causes everything to resize.  This is really annoying.
jumping
With a long transition (hack), you can't really notice it moving
still
This only affects chrome apparently, and once [this bug](https://bugs.chromium.org/p/chromium/issues/detail?id=428132) is resolved in stable chrome, we can take this out again - although their fix may involve it being hidden behind the url bar.
@sammorrisdesign does that seem sensible? @NataliaLKB

### DIFF
--- a/static/src/stylesheets/module/content/_article-immersive.scss
+++ b/static/src/stylesheets/module/content/_article-immersive.scss
@@ -664,6 +664,7 @@
             height: 100vh;
             min-height: 520px;
             max-height: 800px;
+            transition: height 999s; // remove when fixed: https://bugs.chromium.org/p/chromium/issues/detail?id=428132
         }
 
         .content__header {
--- a/static/src/stylesheets/module/content/_article-immersive.scss
+++ b/static/src/stylesheets/module/content/_article-immersive.scss
@@ -664,7 +664,7 @@
             height: 100vh;
             min-height: 520px;
             max-height: 800px;
-            transition: height 999s; // remove when fixed: https://bugs.chromium.org/p/chromium/issues/detail?id=428132
+            transition: height 999999s; // remove when fixed: https://bugs.chromium.org/p/chromium/issues/detail?id=428132
         }
 
         .content__header {


sammorrisdesign:
Hahahah this solution is great. Although does this not mean the rest of the article slowly moves over the next 16 and a half minutes?

NataliaLKB:
@johnduffell the problem with this fix in isolation is that you are not compensating for orientation change on mobile. However, if that is not an issue then merge away!

If it is, read on:

@gtrufitt has a fix for the Minute that does deal with this edge case, however next week we can plan to find the most performant solution and use it across all immersives, the minute, as well as the new design for galleries by @zeftilldeath

NataliaLKB:
@sammorrisdesign you are right! In my own implementation of this I have a much larger time so it virtually doesn't move :smile:

johnduffell:
@NataliaLKB good point about the orientation change - what do you think @sammorrisdesign ?
Natalia has a fix in mind for next week which should fix the orientation issue here and for galleries so the question is should we ship this one in the mean time.

I'll up the transition to 99999 seconds too ;)

gtrufitt:
You could just apply the `js-is-fixed-height` from https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/common/modules/ui/full-height.js in the meantime which will just add a fixed height (and perform the same way as this fix) but handles orientation change too.

sammorrisdesign:
Yeah, I think @NataliaLKB is right. If we have a fix coming it might be worth waiting until that.

Not that I don't really enjoy this solution.

NataliaLKB:
I think the main question is:
For immersives is orientation change something we care enough to compensate for? 
If so, then use @gtrufitt 's suggestion. 
If not, then lets merge this PR

johnduffell:
sounds great, I'll close this one and be patient :hand:

johnduffell:
oh, it's a shame we still have this problem 😢

gtrufitt:
Note, for Chromium, there is a bug here: https://bugs.chromium.org/p/chromium/issues/detail?id=428132#

Hopefully this will be looked at, in Chrome at least, soon: https://groups.google.com/a/chromium.org/forum/#!msg/blink-dev/BK0oHURgmJ4/YLXY6uA6BQAJ

In the meantime, and for Safari, we should still look at using the `full-height.js` I guess.

PR #12197, opened 2016-03-08
Article - Recommendation discovery - People who read this also read - baseline

mchv:
### Recommendation discovery

Recommendations are widely used as a mechanism to encourage a customer to buy more products / a reader to consume more content. We currently we do not provide recommendations to our readers so we have created a small team ( @MahanaC @cb372 @LATaylor-guardian @thomaska @cb372 ) that will spend 4 weeks to test a series of hypotheses to understand which recommendations improve onward journeys for our readers.
#### Objectives
- Our primary aim is to prove that content recommendation using a collaborative filtering algorithm can increase the number of articles viewed and increase the time spent per visit.
- Our secondary aim is to understand whether providing effective recommendations can encourage loyalty - increasing the number of visits per week.
#### Recommendation UI

For the discovery we are going to replace/override the related content component. The user interface will not differ, only the content inside this section and the title presented to the user will be different
#### First test - People who read this also read

Our first test will be around recommending content based on what other users have read.
#### Associated changes

The changes add a new A/B test that only change the title from `related content` to `people who read this also read`.  The content is not modified. The results will serve as our baseline to measure success when the content will come from recommendation engine and not related content.

### DIFF
--- a/static/src/javascripts/projects/common/modules/onward/related.js
+++ b/static/src/javascripts/projects/common/modules/onward/related.js
@@ -1,6 +1,7 @@
 define([
     'bonzo',
     'qwery',
+    'fastdom',
     'common/utils/$',
     'common/utils/config',
     'common/utils/mediator',
@@ -13,6 +14,7 @@ define([
 ], function (
     bonzo,
     qwery,
+    fastdom,
     $,
     config,
     mediator,
@@ -66,6 +68,17 @@ define([
                 showCount: false
             }).init();
         } else if (fetchRelated) {
+
+
+            if (ab.isInVariant('RelatedContentDisplayAsRecommendation', 'people-who-read-this-also-read-title')) {
+                var relatedContentSection = document.getElementById('related-content');
+                if (relatedContentSection) {
+                    fastdom.write(function () {
+                        $('.fc-container__header__title', relatedContentSection).textContent = 'people who have read this also read';
+                    });
+                }
+            }
+
             container = document.body.querySelector('.js-related');
 
             if (container) {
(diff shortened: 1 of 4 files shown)


cb372:
Looks good to me. The build failure appears to be unrelated to the changes. Could we get a +1 from a frontend person please?

static/src/javascripts/projects/common/modules/onward/related.js
```diff
@@ -66,6 +66,15 @@ define([
                 showCount: false
             }).init();
         } else if (fetchRelated) {
+
+
+            if (ab.isInVariant('RelatedContentDisplayAsRecommendation', 'people-who-read-this-also-read-title')) {
+                var relatedContentSection = document.getElementById('related-content');
+                if (relatedContentSection) {
+                    relatedContentSection.querySelector('.fc-container__header__title').textContent = 'people who have read this also read';
```
sndrs: it would be more standard to use `$` here, but not a blocker. `.textContent` needs to be inside a `fastdom.write` callback though

sndrs:
:+1: apart from comment

mchv:
@sndrs Thanks! Updated the code to use `$` and `fastdom`

static/src/javascripts/projects/common/modules/onward/related.js
```diff
@@ -66,6 +68,17 @@ define([
                 showCount: false
             }).init();
         } else if (fetchRelated) {
+
+
+            if (ab.isInVariant('RelatedContentDisplayAsRecommendation', 'people-who-read-this-also-read-title')) {
+                var relatedContentSection = document.getElementById('related-content');
+                if (relatedContentSection) {
+                    fastdom.write(function () {
+                        $('.fc-container__header__title', relatedContentSection).textContent = 'people who have read this also read';
```
johnduffell: classes that are used from JS should always start with js-<something> so people know to check the js and not refactor them without thinking.  Since this is an AB test don't worry though, but if it's productionised watch that one

johnduffell:
:+1: great start

PR #12200, opened 2016-03-08
Article - recommendation discovery - Fix name to match convention set-up

mchv:
This is a fix necessary for A/B switching working client side, see https://github.com/guardian/frontend/pull/12197


rich-nguyen:
:+1:

PR #12202, opened 2016-03-08
Fix paid containers logo size

regiskuckaertz:
Logos in commercial containers are currently capped at 140x90px. Since most of them are horizontal, the `max-width` triggers first and results in logos that are unreadable.

picture 1

It turns out that the width limit makes no sense, at least in this instance—there's plenty of room. The problem is vertical space, because the logo will ultimately create a huge negative space below container items that is likely to draw attention away from the content.

As a first step to a better solution, we'll cap the height at 60px. Next is getting all people involved in setting up these components to ask brands for a horizontal version of their logo.


uplne:
:+1: 

giphy

PR #12203, opened 2016-03-08
Show dismiss button only for merchandising components

uplne:
I had to do some additional checks to make sure that dismiss button is only next to merchandising components.

And there are two attributes telling us that:
1. id
2. data-link-name for article inline components

@regiskuckaertz


regiskuckaertz:
:+1:

uplne:
@regiskuckaertz better?

regiskuckaertz:
Great :+1:

PR #12204, opened 2016-03-08
Article - recommendation discovery - only modify related section when it has been fully loaded

mchv:
Associated changes wait  for the the related content component to load successfully before trying to modify the title.

See #12197


OliverJAsh:
:+1:

PR #12207, opened 2016-03-08
use ab util methods to determine if we should run autoplay tests

jamesgorrie:
# Use A/B util to autoplay videos

Using the actual util functions to do this to make sure we aren't running tests when we shouldn't etc as per @johnduffell's [suggestion](https://github.com/guardian/frontend/pull/12186#discussion_r55348671)

Makes things work more accurately

Nope
## Request for comment

@akash1810 @ScottPainterGNM 
## 


akash1810:
:+1:

PR #12208, opened 2016-03-08
Fix redirect loop on tag pages ending with /email

desbo:
Currently there's a redirect loop on `technology/email` caused by `ModelOrResult` not treating a path ending in `/email` as modified, and returning a 302 to the same thing.

I've added `/email` to the check so that if we're on `technology/email` the item gets returned (and the index page rendered) rather than the 302.


gtrufitt:
:+1:

PR #12209, opened 2016-03-08
Update Rubicon Fastlane request parameters

jbreckmckye:
Rather than sending page keywords and Krux segments, we're going to send Rubicon just the page section, and allow them to fish audience data out of the cookies sent up to rubiconproject.

This means we don't have to give the prebidService pageTargeting data any longer, and allows us to call buildPageTargeting in the way we used to, directly in setPageTargeting.


jbreckmckye:

regiskuckaertz:
Let's roll this out :+1:

uplne:
340x

PR #12210, opened 2016-03-08
Fix A/B test readme wording

desbo:
Just a minor update to say that the audience is split between however many variants are defined, not necessarily 50/50.


jamesgorrie:
It does turn this into MVTing not ABTing too. But one step at a time.
:+1:

PR #12157, opened 2016-03-03
404 test: next in series

OliverJAsh:
I'm working with Andrea to 404 test a “next in series” container at the bottom of 4 different series articles (Experience, Alanis, What I'm Really Thinking, and Blind Date) so we can track interest over the next week. These articles go live tomorrow and Saturday.

Very hacky CSS/JS, but we'll tear it out in a week anyway.

The data is hardcoded and comes directly from editorial.
# Screenshot

image
# How will this be tracked?
- clicks on the button
- we will email the next article to people who complete the survey, and track this return journey

PR #12218, opened 2016-03-09
Next in series: add test page

OliverJAsh:
This is an extension of the 404 test for “next in series” functionality, added in #12157.

We need to link users to a Guardian page in order for us to complete tracking on the link. (As users are currently navigated away from the Guardian, we don't complete tracking the click.)

image
# To do:
- [x] Will these URLs route to onward in production? If not, add to Nginx, or move the routes
- [x] Move routes behind switch


onward/conf/routes
```diff
@@ -73,3 +73,9 @@ GET        /info/tech-feedback/*path                  controllers.TechFeedbackCo
 GET        /most-read-mf2.json                        controllers.MostPopularController.renderPopularMicroformat2()
 GET        /related-mf2/*path.json                    controllers.RelatedController.renderMf2(path)
 GET        /series-mf2/*path.json                     controllers.SeriesController.renderMf2SeriesStories(path)
+
+GET        /404-test/series/experience                controllers.SeriesController.renderSeries404Test(name = "experience")
```
gtrufitt: Can you just do `series/$seriesName<experience|alanis|blind-date|etc>` and `enderSeries404Test(seriesName: String)` rather than repeat all these? 

Or `series/:seriesName` 

IIRC you will need to open a platform PR as you mention in your TODO :)

onward/app/controllers/SeriesController.scala
```diff
@@ -32,6 +32,26 @@ object SeriesController extends Controller with Logging with Paging with Executi
...
+      case "experience" => Some("guardian-reminder-experience")
+      case "alanis" => Some("guardian-reminder-alanis")
+      case "blind-date" => Some("guardian-reminder-blinddate")
+      case "what-im-really-thinking" => Some("guardian-reminder-whatimreallythinking")
+      case "yotam" => Some("guardian-reminder-yotam")
+      case _ => None
+    }
+    maybeSurveyId
+      .map(id => s"https://www.surveymonkey.co.uk/r/${id}")
+      .map(url => Ok(views.html.series404Test(url, metaData))).getOrElse(NotFound)
```
gtrufitt: Worth caching this? `Cached(60)OK(...`

onward/app/controllers/SeriesController.scala
```diff
@@ -32,6 +32,26 @@ object SeriesController extends Controller with Logging with Paging with Executi
...
   }
 
+  val metaData = MetaData.make(
+    id = "series-404-test",
+    section = "",
+    analyticsName = "series-404-test",
+    webTitle = "You’ve found our ‘Remind me’ prototype"
+  )
+  def renderSeries404Test(name: String) = Action { implicit request =>
+    val maybeSurveyId: Option[String] = name match {
```
desbo: a `Map` might be more straightforward here

onward/conf/routes
OliverJAsh: Will do, thanks! :-)

onward/app/controllers/SeriesController.scala
OliverJAsh: Will do, thanks! :-)

onward/app/views/series404Test.scala.html
```diff
@@ -0,0 +1,18 @@
...
+
+@main(model.SimplePage(metaData)) { } {
+    <div class="l-side-margins">
+        <div class="content">
+            <div class="gs-container">
+                <div class="content__main-column">
+                    <h1 itemprop="headline" class="content__headline">You’ve found our ‘Remind me’ prototype</h1>
+                    <div class="js-article__container u-cf from-content-api" itemprop="articleBody">
+                        <p>Thanks for your interest in this feature.</p>
+                        <p>‘Remind me’ is an early stage prototype at the moment — we want to learn whether Guardian readers will find this useful.</p>
```
jamesgorrie: I asked this a while back - but do we store that you've shown interest to get back to you later?

jamesgorrie:
:+1:

crifmulholland:
As we chatted about, the tracking should work when going to external pages. The solution shouldn't be to make a page onsite (although I'm sure the result looks better too).

onward/app/views/series404Test.scala.html
OliverJAsh: That's what the survey below does — it captures your email.

OliverJAsh:
Let's see if adding a test page increases the number of received clicks, which will confirm if external link tracking is working or not.

crifmulholland:
:+1:

PR #12219, opened 2016-03-09
A match report is not necessary to have a minute by minute liveblog

mchv:
This is currently preventing link from the football live match pages to go to the match minutes by minutes.


johnduffell:
if that's the rules everyone agrees today then there's no reason not to! we can change back if opinions move on :+1:

TBonnin:
:+1:

PR #12222, opened 2016-03-09
Commercial code doesn't need _.forEach

jbreckmckye:
Our commercial code no longer runs under IE8, so why pull in the _.forEach dependency?


jamesgorrie:
I am glad I saw this, as it's great.
:+1:

PR #12224, opened 2016-03-10
Fix localStorage check in breaking news

desbo:
The old value was always `undefined` so the rest of the function was never being called.


gtrufitt:
:+1:

OliverJAsh:
Static typing would have saved us ;-)

sndrs:
good one!

PR #12225, opened 2016-03-10
Survey page refactor and tracking

uplne:
A little bit better way of doing this + updated tracking.

@kelvin-chappell @JonNorman @regiskuckaertz


kelvin-chappell:
:+1:

PR #12226, opened 2016-03-10
Fixed inline image for videos in commercial containers

regiskuckaertz:
## What does this change?

Items in commercial containers are not pulling in images for videos:
picture 1

Now it does:
picture 2

@kelvin-chappell


kelvin-chappell:
:+1:

PR #12227, opened 2016-03-10
Do not load Foresee survey JS on profile pages

niklasvincent:
# Do not load Foresee survey JS on profile pages
## What does this change?

This change excludes the profile pages (https://profile.theguardian.com) from triggering the Foresee survey JS.

Currently, when Foresee is triggered, it will try to load `/assets/javascripts/vendor/foresee/20150703/foresee-surveydef.js?build=10` from the same domain.

In the case of the profile pages, that means `https://profile.theguardian.com/assets/javascripts/vendor/foresee/20150703/foresee-surveydef.js?build=10`, which is not accessible (all assets for the profile pages are on `https://assets.guim.co.uk`).

When browsers try to load the asset directly from `https://profile.theguardian.com`, it causes a 500 error, which has lead to a constant stream of alerts in our monitoring.

This can be reproduced by hitting `https://profile.theguardian.com/account/edit#forceForesee` whilst logged in (see screenshot further down).

Similarly to how the existing code checked to see if

```
config.page.contentType === 'Network Front'
```

an additional check for

```
config.page.contentType === 'userid'
```

is done and evaluated.

The `contentType` for Identity Frontend is [set here](https://github.com/guardian/frontend/blob/master/identity/app/model/IdentityPage.scala#L14).

Reduce the number of non-critical 500 errors for Identity Frontend to below our pre-determined threshold.

At the moment this makes it hard to rely on our monitoring for Identity Frontend to distinguish actual issues affecting users.

Only web.
## Screenshots

screen shot 2016-03-10 at 11 41 12
## 

### DIFF
--- a/static/src/javascripts/projects/common/modules/analytics/foresee-survey.js
+++ b/static/src/javascripts/projects/common/modules/analytics/foresee-survey.js
@@ -15,12 +15,13 @@ define([
     function load() {
 
         var isNetworkFront = config.page.contentType === 'Network Front',
+            isProfilePage = config.page.contentType === 'userid',
             sampleRate = detect.isBreakpoint({max: 'mobile'}) ? 0.008 : 0.006, // 0.8% mobile and 0.6% rest
             sample = Math.random() <= sampleRate,
             hasForcedOptIn = /forceForesee/.test(location.hash);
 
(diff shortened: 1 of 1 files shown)


gtrufitt:
:+1:

mchv:
@nlindblad great fix!

PR #12228, opened 2016-03-10
Supporter banner for international edition

joelochlann:
## What does this change?

This displays a membership "engaged reader" banner (encouraging people with more than 10 page views to become a Supporter) to readers of the International edition, since on we now have local landing pages for [Europe](https://membership.theguardian.com/eu/supporter) and the [Rest Of The World](https://membership.theguardian.com/int/supporter) and a geolocated redirect [here](https://membership.theguardian.com/supporter).

It also fixes a bug that was introduced in #11166 (specifically, [here](https://github.com/guardian/frontend/blob/975b4a9ee3c1ae8c5023d22307ff8a3b06179f56/static/src/javascripts/projects/common/modules/commercial/membership-messages.js#L76-L80)). This caused a banner closed by the user to immediately redisplay on the next page load. (When closed a second time, it was gone for good.) This was based on a misinterpretation of #11065, where I changed the id of the message to redisplay to everyone who had closed it before that point in time (but _not_ to display to everyone twice). I've added a unit test to make sure this doesn't happen again.

To avoid irritating our users I thought a bit about how my change here would affect the possible states a user could be in at this point. There are four cases (for a given edition, say UK):
##### 1.) has never closed any membership banner

**currently:** they see banner. closed once, it will reappear. closed again, it will not.
**after this change:** they see banner. closed once, it will not reappear.
##### 2.) has closed only `membership-message-uk`

**currently:** they see banner. closed once, it will reappear.
**after this change:** they will not see banner (since they've already closed it)
##### 3.) has closed only `membership-message-uk-redisplayed`

**currently:** they do not see banner.
**after this change:** they will see the banner. closed once, it will not reappear.
##### 4.) has closed both `membership-messaged-uk` and `membership-message-uk-redisplayed`

**currently:** they do not see banner.
**after this change:** they do not see banner.

Case (3) is the only problematic one, but will only affect people who closed the banner once between  #11065 and #11166. And once the banner appears again, closing it once more will keep it closed. So I think it's acceptable.

Value is to drive international readers towards their appropriate localised supporter page in the hopes that they will sign up. Success will be measured based on signup rates for the MEMBERSHIP_SUPPORTER_BANNER_INT campaign code.

Just web
## Screenshots

picture 165
## Request for comment

@regiskuckaertz

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/membership-messages.js
+++ b/static/src/javascripts/projects/common/modules/commercial/membership-messages.js
@@ -22,7 +22,8 @@ define([
 
     var membershipEndpoints = {
         UK:   'https://membership.theguardian.com/supporter',
-        US:   'https://membership.theguardian.com/us/supporter'
+        US:   'https://membership.theguardian.com/us/supporter',
+        INT:   'https://membership.theguardian.com/supporter'
     };
 
     var defaultData = {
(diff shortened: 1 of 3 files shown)


regiskuckaertz:
:+1:

PR #12232, opened 2016-03-10
Add HSTS CDN test endpoint

johnduffell:
I want to have a test that the cdn won't let through any HSTS header.  However we need to have an endpoint that will return it so we know it's gone.  This is that endpoint.


OliverJAsh:
:+1:

PR #12233, opened 2016-03-10
Header bidding: section should be sent to Rubicon Fastlane as an inventory key

jbreckmckye:
It looks like we're not sending the page context up to Rubicon properly in the header-bidding test. This PR should fix that.


uplne:
like_a_boss_cat_meme

:+1:

PR #12235, opened 2016-03-10
Enable batch redirects

JustinPinner:
# Upload a file of from/to URLs into the redirects form
## What does this change?

The admin redirects page can now upload a batch of redirects from a file, instead of just one-by-one manually.

We have approximately 3 million redirects coming our way as a result of shutting down R2. I don't fancy entering those individually through the existing form. Can we measure it? Sure, by watching the Dynamo DB table grow.

No.
## Screenshots

screen shot 2016-03-10 at 16 06 27

### DIFF
--- a/admin/app/controllers/RedirectController.scala
+++ b/admin/app/controllers/RedirectController.scala
@@ -1,5 +1,7 @@
 package controllers.admin
 
+import java.io.File
+
 import play.api.mvc.Controller
 import common.Logging
 import play.api.data._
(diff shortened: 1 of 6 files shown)


admin/app/controllers/RedirectController.scala
```diff
@@ -30,4 +32,57 @@ object RedirectController  extends Controller with Logging {
     SeeOther(routes.RedirectController.redirect().url)
   }
 
+  def redirectBatchPost() = AuthActions.AuthActionTest { implicit request =>
+
+    val body = request.body
+    val uploadedFile = body.asMultipartFormData.flatMap { files =>
+      files.file("urlfile").map { theFile =>
+        val rnd = Math.random().toString.replace(".","")
```
johnduffell: I assume that's replacing an actual dot rather than regex dot "any char"

admin/app/controllers/RedirectController.scala
```diff
@@ -30,4 +32,57 @@ object RedirectController  extends Controller with Logging {
...
+        val tmpName = s"/tmp/$rnd${theFile.filename}"
+        val tmpFile = new File(tmpName)
```
johnduffell: why not just use File.createTempFile(prefix, suffix);

admin/app/views/redirects.scala.html
```diff
@@ -44,6 +44,37 @@
...
+            <legend>Redirect a batch of URLs from a tab-separated file (fromUrl\ttoUrl\r\n): </legend>
+            <div class="form-group">
+                <label for="urlfile" class="control-label col-sm-1">File:</label>
+                <div class="col-sm-6">
+                    <input type="file" class="form-control" id="urlfile" name="urlfile" />
+                </div>
+            </div>
+            <div class="form-group">
+                <div class="controls col-sm-offset-1 col-sm-11">
+                    <input class="btn btn-default" type="submit" value="Upload & Redirect" />
```
johnduffell: &amp;

johnduffell:
great! do it :smiley: :+1:

admin/app/controllers/RedirectController.scala
JustinPinner: Yep

PR #12236, opened 2016-03-10
Instant reveal knowledge quizzes, and styling fixes

sndrs:
## What does this change?
- where possible, we reveal the truth about your answer as you select it, by styling form state. A tiny bit of JS was needed to select which message to show at the end.
- moves the quiz CSS to the head of articles
- updates the image styling 
## Screenshots

screen shot 2016-03-10 at 16 43 59

screen shot 2016-03-10 at 16 44 40

screen shot 2016-03-10 at 15 54 58

screen shot 2016-03-10 at 15 55 13

screen shot 2016-03-10 at 16 37 52
## Request for comment

@Pearson82 @rich-nguyen


common/app/views/fragments/atoms/quiz.scala.html
```diff
@@ -2,32 +2,35 @@
...
-                playForm(s"answers[$index]"),
-                maybeResults.flatMap(_.getAnswerFor(question))
-            )
-        }
-    @maybeResults.map { results =>
-        @renderEndMessage(results)
-    }.getOrElse {
+<form action="@postUrl(quiz)" method="POST" class="@RenderClasses(Map(
+    ("atom-quiz", true),
+    ("atom-quiz--" + quiz.quizType, true),
```
jamesgorrie: Were we at some stage trying to avoid dynamic class names?

static/src/javascripts/projects/common/modules/atoms/quiz.js
```diff
@@ -0,0 +1,38 @@
...
+                                }
+                                bucket--;
+                            } while (bucket >= 0);
+                        }
+                    });
+                }
+            }
+        }
+    };
+});
```
jamesgorrie: What's an `atom`?

static/src/javascripts/projects/common/modules/atoms/quiz.js
```diff
@@ -0,0 +1,38 @@
...
+            if (HTMLFormElement.prototype.checkValidity) {
+                var $quiz = $('.js-atom-quiz--instant-reveal');
+                if ($quiz.length > 0) {
+                    bean.on(document, 'click', Array.prototype.slice.call($quiz), function (e) {
+                        var quiz = e.currentTarget;
+                        if (quiz.checkValidity()) {
+                            var $bucket__message = null,
+                                bucket = $(':checked + .atom-quiz__answer__item--is-correct', quiz).length;
+                            do {
+                                $bucket__message = $('.js-atom-quiz__bucket-message--' + bucket, quiz);
```
jamesgorrie: Same question around dynamic style names (to be able to find / replace / analyze properly).

static/src/javascripts/projects/common/modules/atoms/quiz.js
```diff
@@ -0,0 +1,38 @@
...
+                                if ($bucket__message.length > 0) {
+                                    fastdom.write(function () {
+                                        $bucket__message.css({
+                                            'display': 'block'
+                                        });
+                                    });
+                                    break;
+                                }
+                                bucket--;
+                            } while (bucket >= 0);
```
jamesgorrie: As far as I remember this will run once even if `bucket === 0`.

OliverJAsh: You could coerce the node list into an array and loop over that instead of doing the while, which will give you the index of each item. Saves you having to re-assign a variable each time?

static/src/javascripts/projects/common/modules/atoms/quiz.js
```diff
@@ -0,0 +1,38 @@
...
+    bean,
+    $,
+    fastdom
+) {
+    return {
+        handleCompletion: function () {
+            if (HTMLFormElement.prototype.checkValidity) {
+                var $quiz = $('.js-atom-quiz--instant-reveal');
+                if ($quiz.length > 0) {
+                    bean.on(document, 'click', Array.prototype.slice.call($quiz), function (e) {
```
OliverJAsh: You could import `toArray` from Lodash for readability.

static/src/javascripts/projects/common/modules/atoms/quiz.js
jamesgorrie: Yea, I'd agree completely, I'm not a massive fan of `while`s, unless you need to rely on external state (time etc)...

common/app/views/fragments/atoms/quiz.scala.html
sndrs: yes v good point

static/src/javascripts/projects/common/modules/atoms/quiz.js
sndrs: a new type of content from CAPI, of which this is the first. they're content types that can be edited independently of the main content, and included across content too. widgets, if you like

OliverJAsh:
Added a :+1: as a reaction but here it is as a comment: :+1:

static/src/javascripts/projects/common/modules/atoms/quiz.js
sndrs: > As far as I remember this will run once even if bucket === 0.

that's ok, you could get 0/𝑥

sndrs:
i love the reactions

common/app/model/content/Atoms.scala
```diff
@@ -21,7 +21,8 @@ final case class Quiz(
   title: String,
   path: String,
   quizType: String,
-  content: QuizContent
+  content: QuizContent,
+  revealAtEnd: Boolean = false
```
rich-nguyen: I think we should omit default values where possible, this one is never used.

static/src/javascripts/projects/common/modules/atoms/quiz.js
sndrs: > You could coerce the node list into an array and loop over that instead of doing the while, which will give you the index of each item. Saves you having to re-assign a variable each time?

i need to iterate down from the total count, checking of the presence of an element using the index in the selector  – how would you do that?

common/app/views/fragments/atoms/quiz.scala.html
```diff
@@ -2,32 +2,35 @@
...
+    ("js-atom-quiz--instant-reveal", !quiz.revealAtEnd && quiz.quizType == "knowledge")
+))">
+    @quiz.content.questions.zipWithIndex.map { case (question, index) =>
+        @renderQuestion(question,
+            playForm(s"answers[$index]"),
+            maybeResults.flatMap(_.getAnswerFor(question))
+        )
+    }
+    @renderFooter(maybeResults)
+    @if(!maybeResults) {
```
rich-nguyen: since `maybeResults` is an option, `if(maybeResults.isDefined)`

static/src/javascripts/projects/common/modules/atoms/quiz.js
sndrs: here it's actually iterating through a list of potential elements based on your score:

```
<div class="js-atom-quiz__bucket-message--5">nice</div>
<div class="js-atom-quiz__bucket-message--4">ok</div>
```
etc.

we can't know what the buckets and total etc will be

static/src/javascripts/projects/common/modules/atoms/quiz.js
OliverJAsh: Does this cover it?

``` js
var buckets = $(':checked + .atom-quiz__answer__item--is-correct', quiz);
toArray(buckets).forEach((bucket, index) => {});
```

You could reverse the array if you need to work from the end of the array.

sndrs: ah ok no i don't think so – the `$(':checked + .atom-quiz__answer__item--is-correct', quiz)` lookup gets you the total correct answers. 

the loop, currently, will then look for the first bucket element (`'.js-atom-quiz__bucket-message--' + bucket`) with a smaller or equal total to yours (so 5/10 goes in bucket 4 of `['bucket--4', 'bucket--6']` for example

OliverJAsh: Ah, makes sense! Sorry I didn't notice the `break`.

How about this (ES6 speak can be converted to Lodash/ES5)? May still be misunderstanding the code, please shout if so!

``` js
var buckets = $(':checked + .atom-quiz__answer__item--is-correct', quiz);
var $bucketMessage = toArray(buckets)
    .map((bucket, index) => $('.js-atom-quiz__bucket-message--' + index, quiz)[0])
    .find(bucket => bucket);
if ($bucketMessage) {
    fastdom.write(function () {
        $bucketMessage.css({
            'display': 'block'
        });
    });
}
```

common/app/model/content/Atoms.scala
sndrs: is it always present in the capi response?

rich-nguyen: Yep

PR #12237, opened 2016-03-11
Limit adslot sizes for header bidding trial

jbreckmckye:
After speaking to the adops team, it would appear we only intend to serve header-bidded adverts in 728x90 and 300x250 formats.

Prebid now only requests those sizes from Rubicon (size IDs 2 and 15, respectively):

image


uplne:
fabulous-reaction-gif

:+1:

PR #12238, opened 2016-03-11
Replace liveblog-adverts with liveblog-dynamic-adverts

regiskuckaertz:
It has been confirmed that the new insertion algorithm is the right one for the future.

@uplne @jbreckmckye @Calanthe


uplne:
this show was pure gold too bad it only had _21ef28978aeb271dec3ba9409fa4a747

What can go wrong on Friday...

PR #12241, opened 2016-03-11
Cleaning up story packages fetch

TBonnin:
## What does this change?

Old story packages has been migrated to the new system.
Falling back to the old system is not necessary anymore
## What is the value of this?

Feature stays the same from a reader perspective. Just cleaning the code here

No
## Request for comment

@JustinPinner


rich-nguyen:
:+1:

PR #12242, opened 2016-03-11
Facia press sends status update to a stream

piuccio:
# My Awesome Pull Request
## What does this change?

Facia press sends status updates on a kinesis stream every time it presses a front, either successfully or not

We can create all sort of services that consume this information, e.g. be notified of errors or burst cache of front pages

No
## Screenshots

Json received by a lambda consuming the stream
screen shot 2016-03-11 at 14 01 07
## Request for comment

@janua how do you like it?
@johnduffell this requires a cloudformation change to create a kinesis stream, but in the meantime we can merge it, it'll just do nothing until we configure a stream

### DIFF
--- a/facia-press/app/frontpress/FapiFrontPress.scala
+++ b/facia-press/app/frontpress/FapiFrontPress.scala
@@ -2,27 +2,24 @@ package frontpress
 
 import com.amazonaws.services.s3.AmazonS3Client
 import com.gu.contentapi.client.GuardianContentClient
-import com.gu.contentapi.client.model.{SearchQuery, ItemQuery, ItemResponse}
+import com.gu.contentapi.client.model.{ItemQuery, ItemResponse, SearchQuery}
 import com.gu.facia.api.contentapi.ContentApi.{AdjustItemQuery, AdjustSearchQuery}
 import com.gu.facia.api.models.Collection
 import com.gu.facia.api.{FAPI, Response}
 import com.gu.facia.client.{AmazonSdkS3Client, ApiClient}
 import common.FaciaPressMetrics.{ContentApiSeoRequestFailure, ContentApiSeoRequestSuccess}
 import common._
-import conf.{Configuration, LiveContentApi}
 import conf.switches.Switches.FaciaInlineEmbeds
+import conf.{Configuration, LiveContentApi}
 import contentapi.{CircuitBreakingContentApiClient, QueryDefaults}
+import model._
 import model.facia.PressedCollection
 import model.pressed._
-import model._
-import org.apache.commons.lang.exception.ExceptionUtils
+import play.api.Play.current
 import play.api.libs.json._
 import play.api.libs.ws.WS
-import play.api.Play.current
 import services.{ConfigAgent, S3FrontsApi}
-import views.support.{Item460, Item140, Naked}
 
-import scala.collection.JavaConversions._
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.Try
 
@@ -52,7 +49,11 @@ object LiveFapiFrontPress extends FapiFrontPress {
   def pressByPathId(path: String): Future[Unit] =
     getPressedFrontForPath(path)
       .map { pressedFront => S3FrontsApi.putLiveFapiPressedJson(path, Json.stringify(Json.toJson(pressedFront)))}
-      .asFuture.map(_.fold(e => throw new RuntimeException(s"${e.cause} ${e.message}"), _ => ()))
+      .asFuture.map(_.fold(
+        e => {
+          StatusNotification.notifyFailedJob(path, isLive = true, e)
+          throw new RuntimeException(s"${e.cause} ${e.message}")},
+        _ => StatusNotification.notifyCompleteJob(path, isLive = true)))
 
   def collectionContentWithSnaps(
     collection: Collection,
@@ -74,7 +75,12 @@ object DraftFapiFrontPress extends FapiFrontPress {
   def pressByPathId(path: String): Future[Unit] =
     getPressedFrontForPath(path)
       .map { pressedFront => S3FrontsApi.putDraftFapiPressedJson(path, Json.stringify(Json.toJson(pressedFront)))}
-      .asFuture.map(_.fold(e => throw new RuntimeException(s"${e.cause} ${e.message}"), _ => ()))
+      .asFuture.map(_.fold(
+        e => {
+          StatusNotification.notifyFailedJob(path, isLive = false, e)
+          throw new RuntimeException(s"${e.cause} ${e.message}")
+        },
+        _ => StatusNotification.notifyCompleteJob(path, isLive = false)))
 
   def collectionContentWithSnaps(
     collection: Collection,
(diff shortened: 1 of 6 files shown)


johnduffell:
Fast work!! I was planning to work out why kinesis is better than sqs queues at some point :stuck_out_tongue: but I haven't yet

facia-press/app/frontpress/FapiFrontPress.scala
```diff
@@ -2,27 +2,24 @@ package frontpress
 
 import com.amazonaws.services.s3.AmazonS3Client
 import com.gu.contentapi.client.GuardianContentClient
-import com.gu.contentapi.client.model.{SearchQuery, ItemQuery, ItemResponse}
+import com.gu.contentapi.client.model.{ItemQuery, ItemResponse, SearchQuery}
```
johnduffell: :unamused:

facia-press/app/frontpress/FapiFrontPress.scala
```diff
@@ -52,7 +49,11 @@ object LiveFapiFrontPress extends FapiFrontPress {
   def pressByPathId(path: String): Future[Unit] =
     getPressedFrontForPath(path)
       .map { pressedFront => S3FrontsApi.putLiveFapiPressedJson(path, Json.stringify(Json.toJson(pressedFront)))}
-      .asFuture.map(_.fold(e => throw new RuntimeException(s"${e.cause} ${e.message}"), _ => ()))
+      .asFuture.map(_.fold(
+        e => {
```
johnduffell: consider changing this to an onComplete with a case Failure/case Success just to make it more readable

johnduffell:
where are the tests? :confounded: 
also, just wondering how much ongoing risk this has of killing the presser

facia-press/app/frontpress/StatusNotification.scala
```diff
@@ -0,0 +1,67 @@
...
+    ))}
+
+  def putMessage(message: StatusNotificationMessage): Unit = {
+    Configuration.faciatool.frontPressStatusNotificationStream match {
+      case Some(streamName) =>
+        client.putRecordAsync(
+          new PutRecordRequest()
+            .withStreamName(streamName)
+            .withPartitionKey(partitionKey)
+            .withData(ByteBuffer.wrap(Json.toJson(message).toString.getBytes)),
```
johnduffell: is getBytes guaranteed to use UTF-8 encoding?  assuming that's what you're after

facia-press/app/frontpress/StatusNotification.scala
```diff
@@ -0,0 +1,67 @@
...
+
+  def notifyCompleteJob(front: String, isLive: Boolean) = {
+    putMessage(StatusNotificationMessage(
+      status = "ok",
+      front = front,
+      isLive = isLive
+    ))}
+
+  def putMessage(message: StatusNotificationMessage): Unit = {
+    Configuration.faciatool.frontPressStatusNotificationStream match {
```
johnduffell: do you think it's worth a switch as well?

johnduffell:
looks awesome, thanks fabio! just a few points to consider in the comments above, otherwise :+1:

facia-press/app/frontpress/FapiFrontPress.scala
piuccio: I don't want to change anything because I don't know what I'm doing, just adding an extra line. Maybe @janua wants to change to `onComplete` :tongue:

facia-press/app/frontpress/StatusNotification.scala
piuccio: switch as a circuit breaker, I'll do that

piuccio:
I think there's little risk on press, if kinesis fails the exception does not kill press because of the asynchronous behaviour. But I'm a total noob so I can easily be wrong. @janua can enlighten us

tests what?

facia-press/app/frontpress/FapiFrontPress.scala
janua: I think it's fine as it is. This isn't a `Future`, rather an `Either`. With `onSuccess` you would just be using a `Success` with another nested match inside for the `Either.` The fold is to help avoid all the boilerplate.

If you want a maintain readability then just use the `map` with a match on `Left` and `Right`.

piuccio:
> I was planning to work out why kinesis is better than sqs queues at some point

kinesis allows multiple consumers, sqs only one

facia-press/app/frontpress/StatusNotification.scala
janua: Yes! Add a switch for the transition with a view to removing it.

janua:
I don't think we need tests for this, what would we test? It's just putting something onto kinesis under certain circumstances.

About an exception in the asynchronous behaviour; it all depends on how dependant the asynchronous behaviour is. Just follow the flow of the code, and if you ever see any two `Future`s joining, then there are dependencies between the asynchronous behaviour!

facia-press/app/frontpress/FapiFrontPress.scala
johnduffell: didn't realise it was an either (I'm just trying to illustrate my point about readability by misreading it :stuck_out_tongue: ) but `Either` way I'm happy, just in general we should be careful about overusing scala's power to do everything on one line :)

janua: You didn't..... :scream: #scalapun
Yes, I am all for more readable code!

johnduffell:
maybe not worth testing, I think they would notice reasonbly fast if there was a problem (after we extended cache times) but it wouldn't be a disaster as long as we kept the times fairly short

johnduffell:
excited now :laughing:

piuccio:
added a switch and this is a proof that errors are sent to kinesis

screen shot 2016-03-11 at 14 57 30

janua:
@piuccio Not sure I like that `status` message. Couldn't it just be `error` and `ok`? Then attach another message in `message`?

piuccio:
@janua sure, done

screen shot 2016-03-11 at 15 06 00

janua:
How are you doing this so fast??

piuccio:
:sunglasses:

janua:
Scala :sunglasses:

piuccio:
build is green do I have thumbs up? I'm not afraid of merging on Friday afternoon! :smiling_imp:

johnduffell:
hey I gave you thumbs up ages ago :stuck_out_tongue: https://github.com/guardian/frontend/pull/12242#issuecomment-195392826

janua:
:+1:
Only thing is the `aws` dependency bump. Does it match what's in `facia-scala-client` and anything elsewhere? And why did you bump it?

piuccio:
doesn't match scala client and I've bumped it because nobody bumps dependencies. totally unnecessary but I like keeping deps up to date

janua:
This particular dependency has caused us major consequences in the past that were extremely hard to diagnose and realise there was a problem. I'd keep it matched to what is in `facia-scala-client` if possible.

piuccio:
I remember that, but it happened becuase frontend was using 1.8 while fapi was on 1.10, two different major releases.
I wanted to see what happens when minor versions are different, but I've reverted that change just because it's Friday

piuccio:
tested on code, updates are pressed correctly, dunno how to test admin, I'll deploy on prod and switch it off

rtyley:
The data from this stream is used in https://github.com/guardian/front-pressed-lambda !

PR #12243, opened 2016-03-11
Toast bug fix: prevent overlapping elements above

OliverJAsh:
Toast currently overlaps previous elements, e.g. the Witness button:

Turns out this bit of CSS is no longer necessary…

image

/cc @johnduffell


johnduffell:
:+1: great! thanks :grinning:

PR #12244, opened 2016-03-11
Email sign up: Update when shown

gtrufitt:
## What does this change?
- Changes the Guardian today to be shown only when coming from the network front (not any front)
- Only shows email sign-up for any email once per session
- Remove morning briefing as end of testing period
- Adds a Guardian today specific kill switch

Reduces the overload of shown email sign-ups, moving nearer to the right thing, right time concept. Allows us to quickly shut down the far reaching sign-up quick sharp.


NathanielBennett:
:+1: LGTM

PR #12245, opened 2016-03-11
Breakout html for DFP surveys

Calanthe:
## What does this change?

There is a need to have simple polls served by DFP (because it's the best for the geolocation targeting). I use the same method as we used to have for the old merchandising templates, where html snippets where injected directly into ad slots (instead of json type template which are in use now), outside of the iframe.

We will be able to show surveys to the users and save the results in [formstack](https://www.formstack.com/) (service which generates also form templates).

no
## Screenshots

screen shot 2016-03-11 at 17 45 42

This will require some additional CSS and JS (validation) work directly in the new DFP template.
## 

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/dfp/breakout-iframe.js
+++ b/static/src/javascripts/projects/common/modules/commercial/dfp/breakout-iframe.js
@@ -34,7 +34,6 @@ define([
         return new Promise(function (resolve, reject) {
             /*eslint-disable no-eval*/
             var $iFrame = bonzo(iFrame);
-            var $iFrameParent = bonzo(iFrame.parentNode);
             var iFrameBody = iFrame.contentDocument.body;
             var $breakoutEl;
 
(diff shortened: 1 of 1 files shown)


static/src/javascripts/projects/common/modules/commercial/dfp/breakout-iframe.js
```diff
@@ -45,16 +44,11 @@ define([
...
                     $iFrame.hide();
                     $breakoutEl.detach();
-                    $iFrameParent.append($breakoutEl[0].children);
-                }).then(function () {
-                    var $responsiveAds = $('.ad--responsive', $iFrameParent[0]);
-                    resolve(fastdom.write(function () {
-                        $responsiveAds.addClass('ad--responsive--open');
-                    }));
-                });
+                    $slot.append($breakoutEl[0].innerHTML);
```
regiskuckaertz: Using `.children` instead of `.innerHTML` will save the browser a whole lot of work.

Calanthe: I have only `.innerHTML`.

PR #12246, opened 2016-03-14
Enable plista switchin production environment

JonNorman:
Currently the switch on the [Plista-specific switchboard](https://frontend.gutools.co.uk/dev/switchboard-plista) is disabled in DEV. We are now ready to go live with it - this change enables the switch to be turned off/on in DEV.

**/cc** @kelvin-chappell


kelvin-chappell:
:+1:

PR #12247, opened 2016-03-14
Do not run picture cleaner on atoms

rich-nguyen:
Image html elements may appear in a capi response that contains atoms. Do not clean these ones; atoms will be expanded in the `AtomsCleaner`


JustinPinner:
:+1:

PR #12248, opened 2016-03-14
Email sign-up: OB refactor

gtrufitt:
## What does this change?

We need to show the non-compliant (merchandising) OB 'widget' when the email sign-up is shown. To ensure that OB is loaded early as possible in all cases but still check whether email sign-up will show, I have refactored so that we can check if a sign-up form will be shown before showing it, therefore allowing us to insert the correct OB component.

The value of this is to ensure that we are meeting our contract and so they can test the impressions of the widget.
## Request for comment

@regiskuckaertz @jbreckmckye I'm messing with Outbrain so can you guys have a glance please.

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/third-party-tags/outbrain.js
+++ b/static/src/javascripts/projects/common/modules/commercial/third-party-tags/outbrain.js
@@ -152,11 +152,13 @@ define([
                         module.load('merchandising');
                     }
                 } else {
-                    if (emailRunChecks.getEmailInserted() && emailRunChecks.getEmailShown() === 'theGuardianToday') {
+                    if (config.switches.emailInArticleOutbrain &&
+                        emailRunChecks.getEmailInserted() &&
+                        emailRunChecks.getEmailShown() === 'theGuardianToday') {
                         // The Guardian today email is already there
                         // so load the merchandising component
                         module.load('merchandising');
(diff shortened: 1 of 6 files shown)


regiskuckaertz:
The Outbrain part looks great to me :+1:

gtrufitt:
I've updated this to now handle email slightly differently - use merchandising codes but still place OB into the high relevance container: https://github.com/guardian/frontend/pull/12248/files#diff-9bc737895e7827869f14c0647c206986R40

I've also handled the ad-blocker scenario https://github.com/guardian/frontend/pull/12248/files#diff-9bc737895e7827869f14c0647c206986R161

static/src/javascripts/projects/common/modules/commercial/third-party-tags/outbrain.js
```diff
@@ -120,6 +126,30 @@ define([
             detect.adblockInUseSync();
     }
 
+    function checkEmailSignup() {
```
regiskuckaertz: I feel like a small improvement would have this function return a new promise on the first call and a cached promised on the second call.

regiskuckaertz:
Let's move the widget only in the `merchandising` case and the roll this little guy out :dancers:

gtrufitt:
Thanks @regiskuckaertz - made those changes. Really appreciate your support on this one! :+1:

PR #12250, opened 2016-03-14
Vary membership adblock message by edition

joelochlann:
## What does this change?

We were running a test involving four variations of the ad block message. Based on [these results](https://docs.google.com/spreadsheets/d/1dZER-_FK_LaJ2Q4rSX-Dt0COzDoP7lSDDF2m5hr5YkE/edit?ts=56e6deb8#gid=0) the weekly price variant has the best Supporter conversion rate so we're going with that. At the same time, we need to put in international variations by edition since we now have custom landing pages and also custom banners at the top of the page in local currency.

Also increases min width on the button area of the banner to fix a styling issue where the button appears empty between the tablet and desktop breakpoints. 

Value is not to confuse users by potentially giving them a price in their local currency at the top of the page, but in pounds at the bottom

Just web
#### before

picture 177
#### after

picture 178
#### before

picture 173
#### after

picture 174
## Request for comment

@jbreckmckye

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/donot-use-adblock.js
+++ b/static/src/javascripts/projects/common/modules/commercial/donot-use-adblock.js
@@ -33,39 +33,46 @@ define([
 ) {
     function showAdblockMessage() {
         var adblockLink = 'https://membership.theguardian.com/supporter',
-            message = sample([
-                {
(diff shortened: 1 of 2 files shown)


jbreckmckye:
:+1:

PR #12251, opened 2016-03-14
Redisplay uk supporter banner with colours

joelochlann:
## What does this change?
- Redisplays the UK supporter banner to everyone who has previously closed it (by changing the message id).
- Tweaks the copy on that banner.
- Provides some new colours for the banner and rotates through them on successive page views.

https://trello.com/c/AuLCGUN5/399-update-copy-and-colours-on-engaged-reader-banner-and-display-it-to-readers-in-the-uk

The aim is to drive Supporter signup. Success would be an increase in signups associated with the MEMBERSHIP_SUPPORTER_BANNER_UK code.

No
## Screenshots

picture 185
picture 184
picture 183
picture 182
picture 180
picture 179

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/membership-messages.js
+++ b/static/src/javascripts/projects/common/modules/commercial/membership-messages.js
@@ -33,7 +33,9 @@ define([
     var messages = {
         UK: {
             campaign:      'MEMBERSHIP_SUPPORTER_BANNER_UK',
-            code:          'membership-message-uk',
+            // increment the number at the end of the code to redisplay banners
+            // to everyone who has previously closed them
+            code:          'membership-message-uk-1',
             minVisited:    10,
             data: {
                 messageText: [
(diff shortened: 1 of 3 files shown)


static/src/javascripts/projects/common/modules/commercial/membership-messages.js
```diff
@@ -78,12 +88,20 @@ define([
 
     function show(edition, message) {
         var data = defaults({ linkHref: formatEndpointUrl(edition, message) }, message.data, defaultData);
+        // Rotate through six different colours on successive page views
+        var colour = storage.local.get('gu.alreadyVisited') % 6;
```
joelochlann: For anyone reviewing this: `null % 6` evaluates to `0`... I've checked :stuck_out_tongue_winking_eye:

rich-nguyen:
> Success would be an increase in signups

How do you intend on measuring this, in the absence of a baseline for comparison? Otherwise, a sharp change in signups could be attributed to something else. 

Code looks fine :+1:

static/src/javascripts/projects/common/modules/commercial/membership-messages.js
johnduffell: if you're worried about hidden behaviour you can always make the is-defined check explicit ;)

johnduffell:
sounds great! :+1: I agree with Rich when he says we should try to find a way to be confident the work was worth it

Issue #12214, opened 2016-03-09
Frontend needs to be able to accept /master/ assets for images set in frontstool (slideshows, image replacements, cutouts), for video poster frames and for video trails on cards


paperboyo:
[EDIT: Please see last comment]

Hello,

All calls for images from mediaservice should prioritise a `/master/` asset first. Masters have minimal compression and are of original pixel dimensions (per crop). Only they can guarantee highest image quality. But not all requests for imagery ask for masters.

As a side effect of that, currently, the image picking logic sometimes fails to choose mediaservice crops big enough for imgIX to perform `drp=2` resize. This results in too small images of very low quality (`q=20`) being upsized by browsers.

E.g.:
Current call (the big picture container on lifestyle front, noticed also in slideshows): https://i.guim.co.uk/img/media/74ae0e3c934b9f5ccb78316f333c7fcfeb0aa22b/0_250_5498_3299/1000.jpg?w=700&q=20&auto=format&usm=12&fit=max&dpr=2&s=9d7c0807512de89c75741f9d2f3a35aa

Should be: https://i.guim.co.uk/img/media/74ae0e3c934b9f5ccb78316f333c7fcfeb0aa22b/0_250_5498_3299/master/5498.jpg?w=700&q=20&auto=format&usm=12&fit=max&dpr=2&s=693bf2274017bb1f7372fb58bdb4715a

Please note the `/master/` part of the filename.

Related to https://github.com/guardian/frontend/pull/12079

Just because we've discussed it: @rich-nguyen.

Regards
Mateusz

paperboyo:
Fixed by https://github.com/guardian/frontend/issues/12254. Thanks! The only two types left to fix are fronts tool's Image Replacements and Slideshows. Will close this after I know if we can fix those too.

paperboyo:
Hi,

So apparently frontend needs to introduce some changes before facia can move in the right direction.

A call for volunteers who hate low-res slideshows and image replacements on fronts: began! 🍰 awaits.

Regards
Mateusz

paperboyo:
Video poster frames and trail pictures on playable cards on fronts also need to use masters (they don't). Poster frames should also be responsive (they aren't).

paperboyo:
[...]

PR #12254, opened 2016-03-15
Prioritise master images during trail image picking

rich-nguyen:
This ensures that we pick master images for trail images, I think that's the main issue causing  https://github.com/guardian/frontend/issues/12214

### DIFF
--- a/IMAGES.md
+++ b/IMAGES.md
@@ -3,7 +3,7 @@
 ## How are trail pictures picked in Frontend?
 
 Use the rules below like a filter to narrow down the content api response into a list of candidate images.
-From the candidate images, choose the largest 5:3 image, or simply the largest one if a 5:3 wasn't found.
+From the candidate images, choose the master 5:3 image, or simply the largest 5:3 if a master wasn't found.
 
 ### Default rules
 1. Use the trail pic (known in content api as 'image element with thumbnail relation'), if it contains an image with width >= 460.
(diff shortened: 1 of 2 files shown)


TBonnin:
Nitpicking :) : why not looking for the largest image only if a master one doesn't not exist

Otherwise :+1:

rich-nguyen:
@TBonnin I added a lazy for that, thanks

paperboyo:
Great, thanks @rich-nguyen! Will this affect/fix slideshow behaviour or will I have to bother @piuccio for that?

rich-nguyen:
Let's double check, it's not clear if slideshows will be corrected. My belief is yes, because my fix should affect how facia-press picks images for slideshow.

paperboyo:
Just checked (on International front). Slideshow is **not** using `/master/`. Video trails aren't using `/master/`. That would not surprise me (although, I would like it fixed still). But some cards are still **not** using `/master/` (I've checked, there is one). Example: https://i.guim.co.uk/img/media/cdf9df24553f14bddfaedc4ced55ff47aec06d85/0_220_4928_2958/1000.jpg?w=220&q=20&auto=format&usm=12&fit=max&dpr=2&s=0cc7a56c805743cfeaa4561b3932a2fc

no masters
(I don't know how to properly identify a card for you)

Found them by filtering a Network Img filter with `-master`.

paperboyo:
Ok, my bad: so the trail above is a fronts tool's Image Replacement. Others like it, also do not call for `/master/`. So these and slideshow would probably need an intervention from @piuccio? Not sure about trails behind a video play button, though (which, funnily enough, **won't** accept an Image Replacements, but should :wink:)...

PR #12257, opened 2016-03-15
Upgrade Prebid version to 0.7.0

jbreckmckye:
This upgrades the vendor library used in the US header bidding trial and fixes a bug we've spotted in the live trial, where the Rubicon adapter struggles to assign bid values to any advert after the first auctioned.


uplne:
:+1:

PR #12261, opened 2016-03-16
Use the same S3 client for facia collections/config

johnduffell:
in facia press we have a job to get the facia config from the s3 bucket and put it in an agent.  We also have the actual presser that goes to get the collections from the same bucket and return the pressed json.

These were using separate clients but the latter one was just using the default credentials.  This meant it didn't work if you were using the frontend credentials, only if you had default credentials with access.

This updates them to use the same client, which seems to work.

@rich-nguyen thanks for helping understand this problem! @janua @piuccio fyi/any comments?
@NataliaLKB

### DIFF
--- a/facia-press/app/frontpress/FapiFrontPress.scala
+++ b/facia-press/app/frontpress/FapiFrontPress.scala
@@ -12,6 +12,7 @@ import common._
 import conf.switches.Switches.FaciaInlineEmbeds
 import conf.{Configuration, LiveContentApi}
 import contentapi.{CircuitBreakingContentApiClient, QueryDefaults}
+import fronts.FrontsApi
 import model._
 import model.facia.PressedCollection
 import model.pressed._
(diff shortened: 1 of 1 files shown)


piuccio:
Thanks for looking into press.

I'm not sure if the job agent is needed for facia press, @janua or IntelliJ should give you an answer. The agent is used by facia tool to cache the config, is press pressing on a cached version or is it refreshing it on every press?
In the latter, you can simply remove the job.

Worth mentioning that we are in the middle of migrating the bucket. The config is going to be moved to another bucket in `cmsFronts` account while the pressed files will stay where they are.
This means you'll need two different clients, one for reading and the other for writing.

johnduffell:
to be honest I'm not sure what it's used for, I tried to use intellij to understand what's going on and failed :cry: so now I'm waiting for @janua to comment :stuck_out_tongue:

NataliaLKB:
@johnduffell @janua is away for two weeks on holiday :disappointed:

NataliaLKB:
Since @janua is away, and this does fix a problem, lets ship this!

rich-nguyen:
:+1:

PR #12262, opened 2016-03-16
Do not ask for old story-packages when calling CAPI

TBonnin:
## What does this change?

Now story packages have been migrated to the new system it is not necessary to have the `show-storyPackage` query string param when calling CAPI
## What is the value of this

Shorter CAPI responses :)

No


rich-nguyen:
:+1:

PR #12265, opened 2016-03-16
Remove connection speed detection

jbreckmckye:
This used to be passed up to Omniture analytics (as 'prop48'), but seems to have been dropped at some point over the last few years. As it's no longer used, we should probably delete the implementation.


OliverJAsh:
:+1: :100: and 1 LOC removed!

PR #12268, opened 2016-03-16
Remote file for loading KRUX for Facebook IA

marialivia16:
Adding a remote file for Facebook Instant Articles to send data to KRUX.
- Creating a global JS object to expose data for KRUX. Example of how the parameters are sent to the iframe:
  `https://www.theguardian.com/facebook-ia/remote.html#param_1=value_1&...&param_n=value_n`
- Loading KRUX.
- Inserting ad using the KRUX segments.

@NataliaLKB

### DIFF
--- a/facia/public/facebook_ia_remote.html
+++ b/facia/public/facebook_ia_remote.html
@@ -10,160 +10,149 @@
 <div id="c" style="position:absolute;top:0;left:0;bottom:0;right:0;">
 
     <script>
-        (function () {
+        var scripts = parseParameters();
 
-            function parseParameters() {
-                var script = document.createElement("script");
-                script.id = "ad-slot-id";
+        var slot = scripts[scripts.length - 1];
 
-                var args = window.location.hash.substr(1).split('&');
+        var slotName = "ad-slot-" + scripts.length;
 
-                for (var i = 0; i < args.length; ++i) {
-                    var arg = args[i].split("=");
-                    var argName = arg[0];
-                    var argValue = decodeURIComponent(arg[1]);
+        var config = {} ;
 
-                    var currValue = script.getAttribute(argName);
-                    if (currValue) {
-                        script.setAttribute(argName, currValue + "," + argValue);
-                    } else {
-                        script.setAttribute(argName, argValue);
-                    }
+        function parseParameters() {
+            var script = document.createElement("script");
+            script.id = "ad-slot-id";
+
+            var args = window.location.hash.substr(1).split('&');
+            var i, argsLength = args.length;
+
+            for (i = 0; i < argsLength; ++i) {
+                var arg = args[i].split("=");
+                var argName = arg[0];
+                var argValue = decodeURIComponent(arg[1]);
+
+                var currValue = script.getAttribute(argName);
+                if (currValue) {
+                    script.setAttribute(argName, currValue + "," + argValue);
+                } else {
+                    script.setAttribute(argName, argValue);
                 }
-                document.getElementsByTagName("div")[0].appendChild(script);
-
-                return document.querySelectorAll("script[id$='" + script.id + "']");
             }
+            document.getElementsByTagName("div")[0].appendChild(script);
 
-            var networkId = "59666047";
+            return document.querySelectorAll("script[id$='" + script.id + "']");
+        }
 
-            var scripts = parseParameters();
+        function toTitleCase(str) {
+            return str.replace(/\w\S*/g, function(txt){return txt.charAt(0).toUpperCase() + txt.substr(1).toLowerCase();});
+        }
 
-            var slot = scripts[scripts.length - 1];
+        function exposeMetaDataForKrux() {
+            config = {
+                'pageId' : slot.getAttribute("url"),
+                'edition' : toTitleCase(slot.getAttribute("edition")),
+                'seriesId' : slot.getAttribute("se"),
+                'contentType' : toTitleCase(slot.getAttribute("ct")),
+                'keywordIds' : slot.getAttribute("keywordIds"),
+                'keywords' : slot.getAttribute("k").split(',').map(function (x) { return toTitleCase(x)}),
+                'authorIds' : slot.getAttribute("authorIds"),
+                'blogIds' : slot.getAttribute("bl"),
+                'referrer' : 'facebook-ia',
+                'section' : slot.getAttribute("section")
+            };
+        }
 
-            var slotName = "ad-slot-" + scripts.length;
+        function loadKrux() {
+            var krux = document.createElement('script');
+            krux.setAttribute("class", "kxct");
+            krux.setAttribute("data-id", "J_RxuxWJ");
+            krux.setAttribute("data-timing", "async");
+            krux.setAttribute("data-version", "1.9");
+            krux.setAttribute("type", "text/javascript");
 
-            var config = {} ;
+            krux.innerHTML = "window.Krux||((Krux=function(){Krux.q.push(arguments)}).q=[]); " +
+                "(function(){ " +
+                "var k=document.createElement('script');k.type='text/javascript';k.async=true; " +
+                "var m,src=(m=location.href.match(/\bkxsrc=([^&]+)/))&&decodeURIComponent(m[1]); " +
+                "k.src = /^https?:\\/\\/([a-z0-9_\\-\\.]+\\.)?krxd\\.net(:\\d{1,5})?\\//i.test(src) ? src : src === 'disable' ? '' : " +
+                "(location.protocol==='https:'?'https:':'http:')+'//cdn.krxd.net/controltag?confid=J_RxuxWJ' ; " +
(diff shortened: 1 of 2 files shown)


NataliaLKB:
As discussed in person, we need to determine whether these query params are just ignored by fastly (:+1:) or cleaned (:disappointed:)

NataliaLKB:
With @JustinPinner and @johnduffell 's help [this line](https://github.com/guardian/fastly-edge-cache/blob/78ef4e0e33cafbca1e6ba0b040faa81cb0e56ddd/theguardiancom/src/main/resources/varnish21/backend-nextgen.vcl#L70) was discovered in fastly. This doesn't clean the character `#` in the url so we are good!

NataliaLKB:
@kelvin-chappell , would you be able to take a look at the code for this as well?

facia/public/facebook_ia_remote.html
```diff
@@ -0,0 +1,175 @@
...
+
+            var slotName = "ad-slot-" + scripts.length;
+
+            var guardian = {} ;
+
+            function toTitleCase(str) {
+                return str.replace(/\w\S*/g, function(txt){return txt.charAt(0).toUpperCase() + txt.substr(1).toLowerCase();});
+            }
+
+            function exposeMetaDataForKrux() {
```
NataliaLKB: We don't have to expose the data in this way, we could make it a much simpler object

facia/public/facebook_ia_remote.html
```diff
@@ -0,0 +1,175 @@
...
+                        'section' : slot.getAttribute("section")
+                    }
+                }
+                };
+            }
+
+            function loadKrux() {
+                var krux = document.createElement('script');
+                krux.setAttribute("class", "kxct");
+                krux.setAttribute("data-id", "KS_cfybw");
```
NataliaLKB: Once we meet with Steven, he can show you how to create a new control tag for this. This line of code and one other (where it says "KS_cfyb") can be replaced with the new id. Lets do this before we merge this in.

facia/public/facebook_ia_remote.html
```diff
@@ -0,0 +1,175 @@
...
+            }
+
+            function insertAd() {
+                var slotDiv = document.createElement("div");
+                slotDiv.setAttribute("id", slotName);
+                slot.parentElement.appendChild(slotDiv);
+                return insertScript(buildSlotCode());
+            }
+
+            function init(){
```
NataliaLKB: Minor comment, to stay consistent, can we add a space between the `()` and the `{` in this function?

johnduffell:
# in urls is never sent to the backend, so fastly won't have the chance to strip it out

NataliaLKB:
@johnduffell I think you mean: 
`#` in urls is never sent to the backend, so fastly won't have the chance to strip it out

:wink:

facia/public/facebook_ia_remote.html
```diff
@@ -50,8 +50,8 @@
             }
 
             function exposeMetaDataForKrux() {
-                guardian = { 'config' : {
-                    'page' : {
+                guardian = {
```
NataliaLKB: Try just `config =`, and remove guardian all together. It serves no purpose to have an object within an object. Or you could call it `guardianConfig` or something else if you prefer :smile:

kelvin-chappell:
There's probably a cleaner way to build the ad call, for people who are better at javascript than me, but it all looks like it should work.
Does it work with multiple ads on the page? (if there ever are multiple ads on the page).
If so, :+1:

NataliaLKB:
@kelvin-chappell it would run for every ad on the page. Great for the ads, not much for Krux :disappointed: . I am not sure if there is a way in FIA to specify only load krux in once per page. That is something that maybe Chris Wilk can advise on.

This is much better than nothing, and we should definitely put this solution out. However, I think it is an issue for Krux to count each page as multiple views.

facia/public/facebook_ia_remote.html
```diff
@@ -0,0 +1,172 @@
...
+
+            var networkId = "59666047";
+
+            var scripts = parseParameters();
+
+            var slot = scripts[scripts.length - 1];
+
+            var slotName = "ad-slot-" + scripts.length;
+
+            var config = {} ;
```
cb372: Doesn't this need to be exported to the global namespace, as all this stuff is happening inside a `function() { ... }`? e.g.

```
window.config = config;
```

@NataliaLKB

facia/public/facebook_ia_remote.html
```diff
@@ -0,0 +1,172 @@
...
+                    } else {
+                        script.setAttribute(argName, argValue);
+                    }
+                }
+                document.getElementsByTagName("div")[0].appendChild(script);
+
+                return document.querySelectorAll("script[id$='" + script.id + "']");
+            }
+
+            var networkId = "59666047";
```
NataliaLKB: To help with readability, could we have all the variables declared at the top?

facia/public/facebook_ia_remote.html
```diff
@@ -0,0 +1,172 @@
...
+    <meta charset="utf-8">
+    <meta name="robots" content="noindex">
+    <script>
+    </script>
+</head>
+<body style="margin:0">
+<div id="c" style="position:absolute;top:0;left:0;bottom:0;right:0;">
+
+    <script>
+        (function () {
```
NataliaLKB: Apologies, I missed this wrapping function the first time I went through. Either you can do what @cb372 said and use `window.config`, or make it a global variable by not wrapping it in this function here.

facia/public/facebook_ia_remote.html
```diff
@@ -0,0 +1,172 @@
...
+
+            function parseParameters() {
+                var script = document.createElement("script");
+                script.id = "ad-slot-id";
+
+                var args = window.location.hash.substr(1).split('&');
+
+                for (var i = 0; i < args.length; ++i) {
```
NataliaLKB: I think it is best to declare the variables outside the for loop. Like this:

```
var i = 0;
var argLength = args.length;

for (i = 0; i < argLength; i++) ...
```

Currently, args.length is being re-evaluated each time this loop is run.

facia/public/facebook_ia_remote.html
NataliaLKB: I could be wrong, but as far as I can tell, `networkId` is only used once, so I would probably put it inside that function, not on its own :smile:

facia/public/facebook_ia_remote.html
```diff
@@ -0,0 +1,172 @@
...
+                kruxSegments = (ref = localStorage.getItem('kxsegs')) != null ? ref : "";
+                if (kruxSegments !== "") {
+                    filling = kruxSegments.split(',').join("','");
+                    output = "['" + filling + "']";
+                    targeting += ".setTargeting('x', " + output + ")";
+                }
+                return targeting;
+            }
+
+            var pageLevelCodePart1 = "var googletag = googletag || {}; googletag.cmd = googletag.cmd || []; (function() { var gads = document.createElement(\"script\"); gads.async = true; gads.type = \"text/javascript\"; var useSSL = \"https:\" == document.location.protocol; gads.src = (useSSL ? \"https:\" : \"http:\") + \"//www.googletagservices.com/tag/js/gpt.js\"; var node = document.getElementsByTagName(\"script\")[0]; node.parentNode.insertBefore(gads, node); })();";
```
NataliaLKB: the variables `pageLevelCodePart1` and `pageLevelCodePart2` are only used once as well,  I would move them to be inside the function that they are used. This means they only stay within that function scope, and also helps with readability.

facia/public/facebook_ia_remote.html
```diff
@@ -0,0 +1,172 @@
...
+            }
+
+            function insertAdConfig() {
+                var isFirstAd = scripts.length === 1;
+                if (isFirstAd) {
+                    insertScript(pageLevelCodePart1);
+                    insertScript(pageLevelCodePart2);
+                }
+                var size = slot.getAttribute("data-ad-size");
+                return insertScript(buildSlotDeclaration(size));
```
NataliaLKB: @kelvin-chappell out of curiosity, do all these scripts need to be separate? It would be ideal to decrease the amount of dom manipulations that we have and make only one script tag with all of these in it. What do you think?

kelvin-chappell: There's one script that runs for the entire page, and another script per ad slot.  It could in principle be put together into a single script.  The ad call has to refer to the id of the div in which the ad is going to be inserted.  So the `slotName` var holds the id of the div for the ad, and that gets pushed into the `defineSlot` and `display` methods of the ads library.  The div has to be in the dom before the `display` method is called on it.

marialivia16:
@NataliaLKB @johnduffell @kelvin-chappell Can you give this a final review, please? It would be great to merge it and make it live today as soon as possible as the testing will take a long time. Thanks!

NataliaLKB:
@marialivia16 thanks for making these changes. :+1:

PR #12269, opened 2016-03-16
Explicitly cache redirects from the short urls controller

rich-nguyen:
This attempts to fix router alarms we are regularly seeing. I think this is happening because of short urls.

I see latency spikes in router correlated to applications latency spikes. They are very brief. The fastly logs show an increase in 503s, and all these errors come from applications, specifically `/p/4hhjp` urls.

None of these responses were being cached in `ShortUrlsController` because the `Cached` helper does not add cache headers to 303 responses (and rightly so, we use internal redirecting).

@jennysivapalan @JustinPinner

### DIFF
--- a/applications/app/controllers/ShortUrlsController.scala
+++ b/applications/app/controllers/ShortUrlsController.scala
@@ -19,7 +19,7 @@ object ShortUrlsController extends Controller with Logging with ExecutionContext
       response.content.map(_.id).map { id =>
         Redirect(LinkTo(s"/$id"), queryString)
       }.getOrElse(NotFound)
-    }.recover(convertApiExceptionsWithoutEither).map(Cached(60))
+    }.recover(convertApiExceptionsWithoutEither).map(Cached.explicitlyCache(1800))
   }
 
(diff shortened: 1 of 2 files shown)


JustinPinner:
Nice. :+1:

applications/app/controllers/ShortUrlsController.scala
```diff
@@ -19,7 +19,7 @@ object ShortUrlsController extends Controller with Logging with ExecutionContext
       response.content.map(_.id).map { id =>
         Redirect(LinkTo(s"/$id"), queryString)
       }.getOrElse(NotFound)
-    }.recover(convertApiExceptionsWithoutEither).map(Cached(60))
+    }.recover(convertApiExceptionsWithoutEither).map(Cached.explicitlyCache(1800))
```
jennysivapalan: Looks good but 30 mins seems like a long time for new content?

rich-nguyen: we could soft purge those, if needed, shouldn't happen too often. Alternatively, we can give fresher content a smaller cache time.

PR #12271, opened 2016-03-16
Improve accessibility for flashing elements facility

johnduffell:
Currently there is a handy page at /help/accessibility-help to let you turn off flashing elements if they cause you problems.  This works by looping through all .js-flashing-image and removing from the dom, on facia and liveblog pages.

This wasn't working for a rich link on an article, turns out that there were 3 reasons
- rich links for live blogs don't get the js-flashing-element class
- the code to strip them doesn't run on articles
- the code to strip them runs before the rich links are enhances.

I decided to rewrite the code to add a class to the body element, and use css to target the flashing-image classes with a display: none.

I did lose a couple of tests because they weren't easy to adapt, and people will let us know if we break it.  Comments?

We should probably make it so interactives don't enhance when this setting is enabled.  I don't think getting the interactive teams to add this class to everything that might move is really worth it.

@OliverJAsh I know you have looked at this in the past so any comments appreciated!


OliverJAsh:
Nice!

> We should probably make it so interactives don't enhance when this setting is enabled.

That would mean excluding those users from some of our best content, because we rarely do progressive enhancement well when it comes to our interactives (I think?).

OliverJAsh:
:+1:

static/test/javascripts/spec/facia/accessibility.spec.js
```diff
@@ -93,39 +93,39 @@ define([
...
-            accessibility.saveState({
-                'flashing-elements': true
-            });
-            helpers.shouldHideFlashingElements(function () {
-                expect(qwery('.js-flashing-image').length).toBe(2);
-                done();
-            });
-        });
-    });
+    //describe('Accessibility helpers', function () {
```
jamesgorrie: Why the comment?

johnduffell: that was just to make it obvious I was removing tests but I planned to either rewrite or delete completely!  fixed now

paperboyo:
> We should probably make it so interactives don't enhance when this setting is enabled. I don't think getting the interactive teams to add this class to everything that might move is really worth it.

+1 for what Olly said. Best or not best, it's an important part of what we do, and not all of it flashy. Flashes!

Asking Visuals to add this class to **all** that moves would probably be asking to much, but asking them to empathize with its purpose and the need to consider it for certain content parts where it's possible - that would be a noble goal, I should think.

PR #12272, opened 2016-03-16
Short urls should be cacheable

rich-nguyen:
Short url responses are nearly cacheable, they pass the right cache headers, but fastly can't cache 303s


JustinPinner:
:+1:

johnduffell:
I'll update the fastly tests to reflect this too

PR #12273, opened 2016-03-16
Fix branding of containers

kelvin-chappell:
The facia model has changed so that container config now has a backfill.query instead of a capiQuery, which means that the commercial branding on containers is failing.

This change fixes it.


JonNorman:
:+1:

PR #12274, opened 2016-03-16
ESLint: "no-use-before-define": [2, "nofunc"]

OliverJAsh:
Depends on #12270 

[As discussed in Slack](https://theguardian.slack.com/archives/dotcom-future/p1458143951000087):

> I would like to introduce this ESLint rule: no-use-before-define—values must always be defined before they can be used. http://eslint.org/docs/rules/no-use-before-define
> 
> We currently have large modules which sometimes refer to values several hundred lines of code before they are defined. Hoisting means the variable could be declared anywhere above or below (hoisted). This rule would mean values would always be defined above. For me that's a big readability improvement.

We've decided to exclude function declarations from this rule because there is no consensus on that.


static/src/javascripts/bootstraps/enhanced/accessibility.js
```diff
@@ -82,13 +84,13 @@ define([
 
         React.render(
             React.createElement(Accessibility),
-            document.getElementById(module.DOM_ID),
+            document.getElementById(DOM_ID),
```
piuccio: this change (and in main.js) has the tiny side effect of making `module.DOM_ID` read only. It's good that tests don't change that value but maybe you can juggle `var module` around in order to keep the same behaviour

OliverJAsh: We could define `module` at the top, along with `module.DOM_ID`, and then add `init` to the `module` object afterwards before exporting. I deliberately wanted to avoid this mutation though.

I've seen other modules use this technique as a hacky means of testing. Perhaps those module should be using Squire to do proper dependency injection instead.

piuccio: I prefer not having to rely on other libraries and their magic to write tests.

```js
var module = {
   DOM_ID: 'js-accessibility-preferences',
   init: init
};

function init () {}

return module;
```

^ not a mutation and should pass lint

OliverJAsh: Good point, thanks! :-)

gtrufitt:
Checked my email stuff, still works :D :+1:

PR #12275, opened 2016-03-16
Revert "Short urls should be cacheable"

johnduffell:
Reverts guardian/frontend#12272

this is somehow causing fastly to strip the campaign codes, roll back now and we can look tomorrow as it's not obvious why this is happening

@JustinPinner @rich-nguyen


rich-nguyen:
we fixed it in fastly https://github.com/guardian/fastly-edge-cache/pull/591

PR #12276, opened 2016-03-17
Remove JSCS

desbo:
eslint obsoletes jscs, so let's get rid of it to reduce cognitive overhead and speed our builds up*!

_\* by 5 seconds_


OliverJAsh:
<3

Just to check, there are definitely no JSCS rules we need to carry over? I can't recall any myself.

I can't see the diff of the shrinkwrap. By any chance did you run `make shrinkwrap` instead of `npm shrinkwrap`, as this does some clean up for better diffs?

OliverJAsh:
:+1:

desbo:
@OliverJAsh if we exclude the preset ("crockford") these are the eslint equivalents of what was in `.jscsrc`

```
    "quote-props": 0,
    "one-var": 0,d 
    "vars-on-top": 0 
```

all of those are in `eslint:recommended` so the only thing we're throwing away is the `crockford` preset in jscs. since we're already using the `eslint:recommended` preset I think it's ok not to have two (possibly conflicting) presets running at the same time, what do you think?

desbo:
I did run `make shrinkwrap` by the way.

OliverJAsh:
It might be worth checking what the difference is between the JSCS and ESLint presets? I think it will be fine though!

Obviously the shrinkwrap cleaner isn't helping much :-P

jbreckmckye:
:fireworks:

desbo:
here's the `crockford` preset, eslint style:
```
{
  "rules": {
    "curly": [
      2,
      "all"
    ],
    "keyword-spacing": [
      2,
      {
        "overrides": {
          "if": {
            "after": true
          },
          "else": {
            "after": true
          },
          "for": {
            "after": true
          },
          "while": {
            "after": true
          },
          "do": {
            "after": true
          },
          "switch": {
            "after": true
          },
          "case": {
            "after": true
          },
          "return": {
            "after": true
          },
          "try": {
            "after": true
          },
          "catch": {
            "after": true
          },
          "function": {
            "after": true
          },
          "typeof": {
            "after": true
          }
        }
      }
    ],
    "space-before-blocks": [
      2,
      "always"
    ],
    "wrap-iife": 2,
    "space-before-function-paren": [
      2,
      "always"
    ],
    "one-var": [
      2,
      "never"
    ],
    "var-on-top": 2,
    "no-empty": 2,
    "array-bracket-spacing": [
      2,
      "never",
      {
        "singleValue": true
      }
    ],
    "space-in-parens": [
      2,
      "never"
    ],
    "no-underscore-dangle": 2,
    "comma-style": [
      2,
      "last"
    ],
    "space-unary-ops": [
      2,
      {
        "words": false,
        "nonwords": false
      }
    ],
    "space-infix-ops": 2,
    "no-with": 2,
    "indent": [
      2,
      4,
      {
        "SwitchCase": 1
      }
    ],
    "no-mixed-spaces-and-tabs": 2,
    "no-trailing-spaces": 2,
    "comma-dangle": [
      2,
      "never"
    ],
    "brace-style": 2,
    "eol-last": 2,
    "new-cap": 2,
    "dot-notation": 2,
    "no-multi-str": 2,
    "key-spacing": [
      2,
      {
        "afterColon": true
      }
    ]
  }
}
```
generally these are much stricter than `eslint:recommended`, and I don't agree with a lot of them. Instead of going through each of these and deciding which to keep I suggest we drop this and accept `eslint:recommended` as our baseline, which we're free to extend in the future.

OliverJAsh:
:+1: Great!

PR #12279, opened 2016-03-17
Fix odd Omniture error in dev

OliverJAsh:
# The issue

You may have noticed the following obscure runtime error appearing in your console, in dev only, since Chrome 49:

image

This error occurs because `window.s` is `undefined`. Why is it `undefined`?

Omniture is wrapped in a try/catch. If there's an error, we beacon it. But we don't log it.

If I add a console log to the catch (as we should have been doing), we get another obscure, but more helpful, runtime error:

image

This error occurs in the Omniture script because `k.AppMeasurement` returns `undefined`, where `k` is the `window`. The fact this expression returns `undefined` is an oddity of JavaScript’s function declarations when used inside try/catch, as is occurring in the Omniture script.

Here's a digest of what's going on in Omniture:

``` js
foo();
function foo() {
    // Correctly logs function definition
    console.log(window.foo)
}

try {
    bar();
    function bar() {
        // You would expect this function declaration to
        // be on the window object, but it's not because,
        // unlike the above example, it's in a try/catch.
        // Thus, it logs undefined.
        console.log(window.bar);
    }
} catch (e) {
    console.log(e.stack)
}
```

http://jsbin.com/siceji/4/edit?html,console,output

Why has this error suddenly appeared? [Chrome were previously doing the wrong thing according the ECMAScript standard, but since Chrome 49 they fixed this](https://bugs.chromium.org/p/chromium/issues/detail?id=587122).

Why don't we see this error in production? Uglify re-orders the Omniture code so the function declarations appear before their usages, [which means this works](http://jsbin.com/siceji/5/edit?html,console,output).
# The fix

This PR fixes the problem by referring to the function definition using local scope instead of the window:

``` js
try {
    bar();
    function bar() {
        // Correctly logs function definition
        console.log(bar);
    }
} catch (e) {}
```

Unfortunately this meant modifying the Omniture source code below the lines that very clearly state:

```
============== DO NOT ALTER ANYTHING BELOW THIS LINE ! ===============
```

:confused: 

/cc @uplne


rich-nguyen:
I think I've seen this kind of thing happen in [Firefox](http://stackoverflow.com/questions/27581205/javascript-code-fails-in-try-catch-block-only-in-firefox), which seemed to lead first in following the ecma standard for try-catch declaration semantics.

I'm glad you've caught this one, but the fix is a :-1: for me. We will almost certainly lose this fix if we upgrade, and it's really opaque. There isn't a nice spot for comments, either.

Maybe we should use the uglified omniture script in dev? At least that would allow us to work the same way in both dev and prod.

I would actually prefer removing the try-catch, as mad as that sounds. The module has clearly not been written with catching in mind. We don't iterate/change that module, either. And there is something extremely uncomfortable about knowing that we _rely_ on uglify to run Prod, Dev, or both.

rich-nguyen:
I've just checked `omniture-library-error` in cloudwatch, and it's averaging 800 errors a day. That's 0.0032% of our page views falling into the try catch.

OliverJAsh:
> And there is something extremely uncomfortable about knowing that we rely on uglify to run Prod, Dev, or both.

:+1: 

Removing the try-catch would be good. I'd rather that than add several seconds to `make compile-dev` due to uglification. I'll update this PR!

sndrs:
after chatting with rich, it struck us that omniture is probably the one piece of code that actually reports its' own failings, and in that sense, removing the try/catch seems less dangerous.

neither removing it or requiring uglification are great though. wonder if we can ask adobe to fix it?

OliverJAsh:
> wonder if we can ask adobe to fix it?

If this was an open-source project I would open a PR ;-)

I'll chat to Matt Kopka to see if we can file an issue and get it fixed upstream.

OliverJAsh:
@rich-nguyen @sndrs Please see https://github.com/guardian/frontend/pull/12279/commits/e119989c291fae0a8883d8ada8a6b4405727c603

common/app/views/fragments/omnitureScript.scala.html
```diff
@@ -1,21 +1,12 @@
...
 
 @item.map{ item =>
     <script>
         @* must always be set before the Omniture file is parsed *@
         window.s_account = '@OmnitureAnalyticsAccount(item)';
+        @Html(Static.js.omnitureJs)
     </script>
 }
-
-<script>
```
rich-nguyen: have you tested this works? is it possible `s_account` could be referenced before the assignment?

rich-nguyen:
:+1: aside from comment in https://github.com/guardian/frontend/pull/12279/commits/e119989c291fae0a8883d8ada8a6b4405727c603

common/app/views/fragments/omnitureScript.scala.html
```diff
@@ -1,20 +1,12 @@
...
 
 @import conf.Static
-@import conf.Configuration
 @import views.support.OmnitureAnalyticsAccount
 
 @item.map{ item =>
     <script>
         @* must always be set before the Omniture file is parsed *@
         window.s_account = '@OmnitureAnalyticsAccount(item)';
+        @Html(Static.js.omnitureJs)
```
sndrs: feels weird having the lib inside a map, given it's a loop?

common/app/views/fragments/omnitureScript.scala.html
sndrs: only ref should be in the app, which won't fire till dom ready – should be ok?

common/app/views/fragments/omnitureScript.scala.html
OliverJAsh: As far as I can see, `item` is an Option, so we can have to map over it here?

sndrs: sure, so maybe
```html
<script>
    @* must always be set before the Omniture file is parsed *@
    @item.map{ item =>
        window.s_account = '@OmnitureAnalyticsAccount(item)';
    }
    @Html(Static.js.omnitureJs)
</script>
```
?

OliverJAsh: If the Option is a `None`, we don't want to include the Omniture library, or do we?

sndrs: oh yeah that's a good point, probably not

common/app/views/fragments/omnitureScript.scala.html
OliverJAsh: Yup, tested. The Omniture library still appears after the definition of `s_account`.

rich-nguyen: `s_account` is referenced in the omniture vendor js file that we are inlining through `omnitureJs`. Is there any possible way it could be referenced before our assignment?

johnduffell:
good detective work there! it's a shame omniture library is so opaque

common/app/views/fragments/omnitureScript.scala.html
OliverJAsh: As discussed, there should be no possibility of that happening.

PR #12280, opened 2016-03-17
Move DFP init at the top of the chain

regiskuckaertz:
## What does this change?

Before setting up ad slots and loading creatives, the googletag script needs to be loaded. Currently, it happens only when all the ad slots have been inserted into the page, but this feels like a waste of time. We can already load scripts asynchronously further ahead (googletag, prebid, etc.) and set up basic parameters like page targeting and listeners in advance.

The result should be that ads load slightly faster, although this is hard to measure given how many scripts are executed in "parallel" (via huge stacks of promises and fastdom queues).

@uplne @jbreckmckye @Calanthe


static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
```diff
@@ -722,6 +732,7 @@ define([
      */
     var dfp = {
         init:           init,
+        load:           load,
```
jbreckmckye: I might have called this `loadAds`, so it's clear what we're displaying ads that already exist, rather than kicking off our DFP dependencies.

static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
```diff
@@ -16,7 +15,7 @@ define([
     'common/utils/url',
     'common/utils/user-timing',
     'common/utils/sha1',
-    'common/utils/fastdom-idle',
```
jbreckmckye: Don't we want to use the requestIdleCallback stuff any more?

static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
```diff
@@ -303,33 +303,40 @@ define([
...
+            if (prebidEnabled) {
+                prebidService = new PrebidService();
+            }
 
-        window.googletag.cmd.push = raven.wrap({ deep: true }, window.googletag.cmd.push);
+            googletag.cmd.push = raven.wrap({ deep: true }, googletag.cmd.push);
 
-        window.googletag.cmd.push(function () {
-            renderStartTime = new Date().getTime();
+            googletag.cmd.push(
```
jbreckmckye: Does GPT's push implementation accept multiple arguments?

static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
regiskuckaertz: fair enough.

static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
regiskuckaertz: I didn't have any problem during my tests, but I'll run some more just to make sure. Would be foolish of them to replace a native function and changing the signature though.

jbreckmckye:
:+1:

PR #12282, opened 2016-03-17
Enable Chrome’s web app install banners

OliverJAsh:
[Chrome will prompt loyal users to add the website to their home screen](https://developers.google.com/web/updates/2015/03/increasing-engagement-with-app-install-banners-in-chrome-for-android) if the following conditions are true. We are fulfilling all but one (in bold) which this PR fixes:

> - You have a web app manifest file, which defines how your app appears on the user’s system and how its hould be launched
>   - The manifest must have a `short_name`, a name for display in the banner,
>   - A start URL (e.g. / or index.html) which must be loadable,
>   - **At least an `144x144` PNG icon**
>   - Your icon declarations should include a mime type of image/png
> - You have a service worker registered on your site. We recommend a simple custom offline page service worker
> - Your site is served over HTTPS (service worker requires HTTPS for security)
> - The user has visited your site at least twice, with at least five minutes between visits.

— https://developers.google.com/web/updates/2015/03/increasing-engagement-with-app-install-banners-in-chrome-for-android

Seeing as we're not yet fully HTTPS, I doubt many people will see this.

Value: users will have another way to come back to the site
Measure: page views with the campaign code of `INTCMP=webapp`
## Screenshot

image
## To do
- [ ] Bump version in name to ensure cache bust

### DIFF
--- a/applications/app/templates/webAppManifest.scala.js
+++ b/applications/app/templates/webAppManifest.scala.js
@@ -10,6 +10,11 @@
     },
     {
         "type": "image/png",
+        "sizes": "144x144",
+        "src": "@{JavaScript(Static("images/favicons/144x144.png").path)}"
+    },
+    {
+        "type": "image/png",
         "sizes": "114x114",
         "src": "@{JavaScript(Static("images/favicons/114x114.png").path)}"
     }


paperboyo:
:gem:

rich-nguyen:
Are you sure we need a 144, when we have a 152? I copied this from a @PaulKinlan blog, I think

gtrufitt:
Rich is right, it's a *minimum of 144*, however it may be a (very very) small file size saving serving up the smaller icon? 

@rich-nguyen I wondered why the start_url has hardcoded `/uk` in it though? Something do to with not being able to use the campaign code without a valid url? If so, maybe we should feed back on that. I guess we would need to internationalise the manifest (ie `start_url` would be based on current edition) in the future as we make more stuff https.

crifmulholland:
This is great... 

Could this happen for crosswords section specifically? i.e. Add crosswords homepage to homescreen? That might give us some better stats?

gtrufitt:
> Could this happen for crosswords section specifically? i.e. Add crosswords homepage to homescreen?

I hadn't thought of this, this is a *really* exciting idea - add to homescreen not just for editions, but for sections, different icon, different name etc - we can have different 'apps' for each game on the site (crosswords can open as a [fullscreen](https://w3c.github.io/manifest/#dfn-display-modes-values) app). Not tested but seems like it should be technically possible according to the [W3C spec](https://w3c.github.io/manifest/#start_url-member).

paperboyo:
Great idea!

Not sure about different sections as that may not be what reader would expect, though. Would an explicit  decision ("would you like a shortcut to the whole of the Guardian or to http://www.theguardian.com/sport+type/video front") be possible or too much friction?

gtrufitt:
> Would an explicit decision ("would you like a shortcut to the whole of the Guardian or to http://www.theguardian.com/sport+type/video front") be possible or too much friction?

I don't think this is possible right now (or preferable). There is a `beforeinstallprompt` that would let us prevent the banner being shown, however there is no check to enable us to see whether we're *allowed* to show the banner. I think the nicest UX for this option would be user initiated - We can check something like `window.isInstallPromptAllowed` (not real) and then show a button that when the user clicks we  can then tell the browser to show the banner. But I can see this as being too susceptible to mis-use so unlikely this would be added (maybe @PaulKinlan could give us some advice here and maybe tell us whether 'multiple apps on a domain' use-case is just weird).

johnduffell:
This looks great!  It would be nice if we could customise the text a bit more on the banner to convince people why they want to add it.  I guess it uses the name from the [manifest](https://www.theguardian.com/2015-06-24-manifest.json) at the moment, but it's not 100% clear what will  be added to the home screen if I agree.
In terms of the icon yes it does sound like 152 _should_ be enough though.

crifmulholland:
Looks like it just doesn't apply the standard rounded corners if it's
off-line
On 22 Mar 2016 11:45 am, "John Duffell" <notifications@github.com> wrote:

> This looks great! It would be nice if we could customise the text a bit
> more on the banner to convince people why they want to add it. I guess it
> uses the name from the manifest
> <https://www.theguardian.com/2015-06-24-manifest.json> at the moment, but
> it's not 100% clear what will be added to the home screen if I agree.
> In terms of the icon yes it does sound like 152 *should* be enough though.
>
> —
> You are receiving this because you commented.
> <https://github.com/guardian/frontend/pull/12282#issuecomment-199775949>
>

-- 

in England Number 908396

OliverJAsh:
Correct, we already have an icon large enough that this should work. Upon investigation, the real reason why the install banner currently doesn't appear is because our manifest file is served over HTTP when it must be served over HTTPS. So that's what we need to fix! @rich-nguyen

PR #12283, opened 2016-03-17
Remove old foresee content

rich-nguyen:
Nowadays foresee servers are delivered through dfp, there is no dependency on the old system.


rich-nguyen:
I will merge this on monday, when the new ad ops creative goes live for the latest survey

johnduffell:
cool, good marks for deletion! :+1: :speedboat:

PR #12285, opened 2016-03-17
Sticky ad banner file renamed

Calanthe:
## What does this change?

This PR renames the `sticky-ad-banner.js` file into `sticky-top-banner.js` because on CODE when files are not combined into one file, Adblock plugin blocks every file with the `ad` in its name.

There is no js error which blocks all other js scripts from executing.

no
## Screenshot

before:
screen shot 2016-03-17 at 18 14 24
## @OliverJAsh 


jbreckmckye:
:+1:

OliverJAsh:
:+1: Thanks!

PR #12286, opened 2016-03-18
Forward campaign codes on /latest/email routes

desbo:
This allows us to enter the campaign code in ExactTarget and retain it after we redirect to the actual article.

### DIFF
--- a/applications/app/controllers/LatestIndexController.scala
+++ b/applications/app/controllers/LatestIndexController.scala
@@ -16,8 +16,11 @@ object LatestIndexController extends Controller with ExecutionContexts with impl
     loadLatest(path).map { _.map { index =>
       index.page match {
         case tag: Tag if tag.isSeries || tag.isBlog => index.trails.headOption.map(latest => {
-          if (request.isEmail) Found(latest.metadata.url + "/email")
-          else                 Found(latest.metadata.url)
+          if (request.isEmail) {
+            request.campaignCode.map(code => Redirect(latest.metadata.url + "/email", Map("CMP" -> Seq(code))))
+              .getOrElse(Found(latest.metadata.url + "/email"))
+          }
+          else Found(latest.metadata.url)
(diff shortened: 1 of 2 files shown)


applications/app/controllers/LatestIndexController.scala
```diff
@@ -16,8 +16,11 @@ object LatestIndexController extends Controller with ExecutionContexts with impl
     loadLatest(path).map { _.map { index =>
       index.page match {
         case tag: Tag if tag.isSeries || tag.isBlog => index.trails.headOption.map(latest => {
-          if (request.isEmail) Found(latest.metadata.url + "/email")
-          else                 Found(latest.metadata.url)
+          if (request.isEmail) {
+            request.campaignCode.map(code => Redirect(latest.metadata.url + "/email", Map("CMP" -> Seq(code))))
```
rich-nguyen: you could map `campaignCode` to a `Map` with `CMP` defined, and then `getOrElse Map()` for empty map

rich-nguyen:
:+1:

applications/app/controllers/LatestIndexController.scala
desbo: what about a fold :smiling_imp: 
`request.campaignCode.fold(Map[String, Seq[String]]())(c => Map("CMP" -> Seq(c)))`

rich-nguyen: `request.campaignCode.fold(Map())(c => Map("CMP" -> Seq(c)))` :)

desbo: without the type it gives me this... :(
`type mismatch;
 found   : (String, Seq[String])
 required: (Nothing, Nothing)`
is there a way around that?

rich-nguyen: no, i thought it was inferred, but not here. github needs scalac compiler

PR #12287, opened 2016-03-18
Restore standfirsts on 'paid-for' audio content

jbreckmckye:
Commercialized audio content doesn't currently show standfirsts -

image

This is the side effect of an errant CSS rule that caters to an oddity of video content - most videos don't have body content, only standfirsts, injected into the page a second time to substitute the body. We have CSS to hide the first standfirst (in the header) and style the second standfirst (acting the body), but real standfirsts in audio pages are getting caught in the crossfire.

Paid-for audio content now shows standfirsts as expected:

image


Calanthe:
:+1:

PR #12258, opened 2016-03-15
Adblocking response AB test - part 1

Calanthe:
## What does this change?

This PR introduces the first part of an AB test which will test how users will response to adblock messages. 
## What is the value of this and can you measure success?

The message (which is not decided yet, there will be 2 different variants) will be shown to adblock users on content pages to encourage them either to whitelist our site or become a paid member. The message box will be on a dark background layer which will cover the whole content and will not be dismissible.
## Does this affect other platforms - Amp, Apps, etc?

no
## Screenshots

screen shot 2016-03-15 at 16 44 36
## TODO in the next PRs:
- A proper links, text, cta's, audience and data links.
- `Enter subscriber number` page which based on the provided number will check if the user is a subscriber
- Logic with redirect to the last seen article to show it after whitelisting, entering subscriber's number or logging as a member. 
## Request for comment

@uplne @jbreckmckye 
## 

_Does this PR meet the [contributing guidelines](https://github.com/guardian/frontend/blob/issue_pr_templates/.github/CONTRIBUTING.md#submission)?_

PR #12288, opened 2016-03-18
Adblocking response AB test  - text, links, buttons

Calanthe:
## What does this change?

Another part of the Adblocking response AB test https://github.com/guardian/frontend/pull/12258
Texts, links, data attributes and second variant added.

screen shot 2016-03-18 at 14 23 04
## 


static/src/javascripts/projects/common/views/commercial/survey/survey-simple.html
```diff
@@ -22,6 +22,16 @@ <h3 class="survey-text__header">
                 <p class="survey-text__info">
```
jbreckmckye: Maybe instead of using separate variables - `surveyTextFirst`, `surveyTextSecond`, etc - you could just put the paragraphs in an array, e.g.

```html
<% surveyText.forEach(function(para) {%>
    <p class='survey-text__info'>
        <%=para%>
    </p>
<% }) %>
```

static/src/stylesheets/module/commercial/surveys/_survey.scss
```diff
@@ -42,9 +42,9 @@
         box-sizing: border-box;
 
         @include mq(tablet) {
-            left: 50%;
+            left: 42%;
```
jbreckmckye: Is this supposed to be horizontally centered? Maybe you could keep the left:50% and just use `margin-left: -$gs-column-width * 5`, instead? That way it'd look a bit more like 'idiomatic' centering?

jbreckmckye:
Other than suggestions, this looks good :+1:

PR #12289, opened 2016-03-18
Also stop flashing dot on facia

johnduffell:
I rejigged how we hide flashing dots when someone turned their accessibility settings that way in https://github.com/guardian/frontend/pull/12271

However I didn't include the css file on facia meaning it didn't help on fronts e.g. /tone/minutebyminute

This PR is just to add in the missing line.
@OliverJAsh if you have a minute :smile_cat:

### DIFF
--- a/static/src/stylesheets/facia.scss
+++ b/static/src/stylesheets/facia.scss
@@ -31,3 +31,4 @@
 @import 'module/facia/snaps/_facia';
 @import 'module/crosswords/_main';
 @import 'module/email/_main';
+@import 'module/_accessibility';


OliverJAsh:
:+1:

Maybe we need a common stylesheet?

johnduffell:
thanks for the +1, and we could oli, but for this it's such a small thing I left it separate. I'll get this one out for now as it's more broken than before and we can think about separating it later.

johnduffell:
+also, the word common is always a disaster in package naming, so I'm going to try to think of a better name :stuck_out_tongue_closed_eyes:

PR #12291, opened 2016-03-18
R1 urls should be normalised always. Remove redundant switch.

JustinPinner:
## What does this change?

Removes a switch that I added temporarily.

We can all be a little happier when the switch is removed. If you really want to you can sing :notes: "Ding dong the switch is dead" :notes: over and over. But I wouldn't recommend it. 

No.
## Free-for-all :scream: 


rich-nguyen:
:+1:

PR #12293, opened 2016-03-18
Upgrades capi and fapi

rich-nguyen:
Upgrade to the latest facia client, featuring collection-based backfills. Currently need to sync our aws and capi dependencies to facia clients, due to risky library eviction during jar packaging.


rich-nguyen:
I didn't realise this was missing a :+1:, sorry!

PR #12295, opened 2016-03-21
Restore min-heights and paddings to all standfirsts

jbreckmckye:
A side effect of [this](https://github.com/guardian/frontend/pull/12287) PR is that standfirsts no longer have a minimum height. This causes problems when a sponsored content badge is over a certain height - the element is absolutely positioned and doesn't contribute to its container's height:

image

This PR restores the padding and min-height rules to all standfirsts, as existed before #12287. 

image

This does leave us with an unusually large standfirst on audio articles (I have yet to confirm whether it is actually intentional), but this seems a smaller problem and more bearable in the short term than broken badges:

image


Calanthe:
:+1:

PR #12296, opened 2016-03-21
Removing dismiss 404

uplne:
We have enough data to analyse from this test so it is safe to remove it.

CC @Calanthe I tried to remove only obsolete things. Can you please double check that I left everything you need there?


regiskuckaertz:
:+1:

PR #12259, opened 2016-03-15
Live blog browser alerts

NathanielBennett:
# My Awesome Pull Request
## What does this change?

(My Awesome pull request, yeah, right ..)
Building on previous work done by @OliverJAsh and @sndrs, we're going to try live notificatons on live-blogs on chrome. (We'll use cookies to put them on https). Initially this will be a zero percent ab test that the loyalty team can play with internally, before we roll out to a subset of of users. Using a service worker to obtain a browserId, we store the browserId against a a pageID in Dynamo. This work by @janua here: https://github.com/guardian/frontend-notifications handles sending the notfications via gcm. At present, we just display key events The extension to service-worker.js here displays them. 

Relies on https://github.com/guardian/platform/pull/937
## What is the value of this and can you measure success?

We want to see if we can use notifications as a loyalty driver. There's a campaign code in the notification link to enable us to track returning page views
## Does this affect other platforms - Amp, Apps, etc?

Nope
## Screenshots

(First time upsell)
live-blog-follow-new

(Notification)
new-notification

(User has already followed other blogs, but not this one)
follow

(User is already following this live-blog)
live-blog-following
## 

_Does this PR meet the [contributing guidelines](https://github.com/guardian/frontend/blob/issue_pr_templates/.github/CONTRIBUTING.md#submission)?_

PR #12297, opened 2016-03-21
load notification bootstray via require

NathanielBennett:
# My Awesome Pull Request

(not)
## What does this change?

Load the notifications module as a standard require module

Notifications would work( As opposed to not at all )

None
## Screenshots

See here: https://github.com/guardian/frontend/pull/12259
## 


sndrs:
does this mean we shipping notifications code to every enhanced user?

OliverJAsh:
It does. When we're in a HTTPS world, all enhanced users will need this so it makes sense to include it in the enhanced file. In the interim it's such a small module, it shouldn't add much to enhanced. Is that ok?

OliverJAsh:
:+1: from me if @sndrs is ok with this

sndrs:
just for paper trail, chatted in slack about moving this to liveblog bootstrap rather than main enhanced. :+1: for that from me

PR #12298, opened 2016-03-21
Refactor SCSS to remove duplication between facia and content bootstraps

johnduffell:
As per @OliverJAsh comment on https://github.com/guardian/frontend/pull/12289

We should have a common stylesheet between facia and content, as it's full of duplication anyway.

### DIFF
--- a/static/src/stylesheets/content.scss
+++ b/static/src/stylesheets/content.scss
@@ -1,14 +1,13 @@
 @charset 'UTF-8';
 
+@import '_facia-content';
+
 /* ==========================================================================
    Variables and mixins
    ========================================================================== */
 
-@import '_vars';
-@import '_mixins';
 @import 'module/facia/_mixins';
 @import 'module/_discussion-mixins';
-@import '_extends-only';
 
 
 /* ==========================================================================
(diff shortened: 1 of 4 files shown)


static/src/stylesheets/content.scss
```diff
@@ -1,14 +1,13 @@
 @charset 'UTF-8';
 
+@import '_facia-content';
+
 /* ==========================================================================
    Variables and mixins
    ========================================================================== */
 
-@import '_vars';
```
sndrs: i'm not sure about abstracting these, they're kind of 'settings' partials, it's not clear there's present now?

sndrs:
we have a `_head.common.scss`, maybe call it `_common.scss`?

static/src/stylesheets/content.scss
johnduffell: will do, I didn't really know why things are there I just had a bash at it!

johnduffell:
I just think the word common always causes problems as it isn't specific, it just says what it isn't rather than what it is.  So it turns into a dumping ground for anything no-one knows where they go (look at the common project in the scala for a prime example)  I prefer to say what things are in explicitly rather than just inviting anything.  Happy to discuss and change if necessary though.

OliverJAsh:
Perhaps call it `facia-and-content` or `facia+content` to make it clear? I was confused at first!

OliverJAsh:
Nice work! :+1:

sndrs:
`_head.common.scss` currently is included in every `head.*.scss`, so that makes sense to me? this file will also be included in both main css files (i.e. there are none which will not have this). happy for any way of naming them to prevail, but i think they should be consistent?

johnduffell:
on reflection since this is a file rather than a container that includes specifically named files (rather than being a container) we're not likely to get the dumping ground problem I mentioned

PR #12301, opened 2016-03-21
float meta content at leftcol breakpoint, so that it stays in the flow

sndrs:
## What does this change?

float the content header meta at left-col breakpoint, rather than use absolute positioning, so that it stays in the flow and pushes the content down if it extends beyond the header content
## Screenshots

before
screen shot 2016-03-21 at 12 59 44

after
screen shot 2016-03-21 at 12 58 13
## Testing

@michaelwmcnamara @ScottPainterGNM I've given this a good test across content types, browsers and breakpoints and think it's ok, but if it breaks something, it's likely to be in quite a big way, so a QA pass would be a good idea to confirm:
- [x] fixes commentisfree/audio/2016/mar/17/podcast-feminism-jessica-valenti in US edition
- [x] has no other effect on content, across types, browsers and breakpoints


static/src/stylesheets/module/content/_content.scss
```diff
@@ -473,8 +473,7 @@
     border-bottom: 1px dotted colour(neutral-5);
 
     @include mq(leftCol) {
-        position: absolute;
```
OliverJAsh: Does this mean we can remove a `position: relative` from some parent?

OliverJAsh:
:+1: from a code perspective!

static/src/stylesheets/module/content/_content.scss
sndrs: done (line 470), just not associated with the same change by gh

paperboyo:
That wouldn't fix the byline hiding under showcased first inline element, would it? Like in: http://www.theguardian.com/lifeandstyle/2015/may/01/last-bites-david-everitt-mathias-final-meal

rich-nguyen:
:+1: for IE screenshots. I'm already more confident this will work.

michaelwmcnamara:
+1 - Looks ok on IE couldn't see any new issues introduced on IE or Chrome, firefox, safari  or mobile devices

sndrs:
great, thanks @michaelwmcnamara!

PR #12303, opened 2016-03-21
Remove zipObject dependency from dfp-api.js

jbreckmckye:
We don't need lodash's zipObject dependency if we use native array methods to build the adverts map.

I think it's much more transparent this way, too.


regiskuckaertz:
Awesome, that's been itching me too for a while. :+1: when the tests succeed.

PR #12305, opened 2016-03-21
Add series view counts to onward/history

desbo:
This gives you a count of hits against any series tags recorded (see the test for an example). I've also increased how long we store history data for. It'll help us work out how to promote series content to different users.

@stephanfowler


stephanfowler:
:+1:

PR #12307, opened 2016-03-21
Retire facia containers

rich-nguyen:
We are retiring the following facia containers. @piuccio has already removed them from facia tool.

fixed/large/fast-XV
fixed/small/fast-X
fixed/medium/slow-VIII
fixed/small/slow-II
fixed/small/slow-VI


johnduffell:
cool :+1: thanks!

PR #12309, opened 2016-03-21
Remove toast div that covers content

johnduffell:
At the moment, there's a toast div sitting on top of the content all the time, blocking up some links near the top of the screen.  This just moves the div to the back when toast isn't popped.

Before:
coveredlink

After:
linkworks

@OliverJAsh I just used a new class in the end!


gtrufitt:
:+1:

OliverJAsh:
:+1:

PR #12315, opened 2016-03-22
Add function to get most viewed series from history

desbo:
A little extension that'll tell us which series tag has the highest number of views.
@stephanfowler


stephanfowler:
:+1:

PR #12316, opened 2016-03-22
Don't run liveblog-adverts on sensitive pages

jbreckmckye:
If we try and run liveblog-adverts on sensitive pages, we end up seeing big grey boxes:

image

This happens because DFP isn't actually running on sensitive content.

This PR adds a commercial-features flag that lets us short-circuit liveblog-adverts on sensitive content.

We now no longer see adverts on liveblogs with the 'should hide adverts' switch, but continue to see ads on others (like the politics liveblog)


JonNorman:
:shipit:

PR #12317, opened 2016-03-22
Fix radiator by making it use new teamcity

johnduffell:
The radiator is currently hanging.  This updates to use the external url for teamcity.

I also updated it to not think riffraff is internal at the same time.

Unfrotunately even though [this url](http://teamcity.gu-web.net:8111/externalStatus.html?buildTypeId=dotcom_master) works, the [js one](http://teamcity.gu-web.net:8111/externalStatus.html?buildTypeId=dotcom_master&js=1) doesn't.  No idea why, but I couldn't work it out after some searching and I didn't want it to block my changes.  Suggestions welcome.


TBonnin:
That would do ;)

PR #12319, opened 2016-03-22
Add test to move series onward container up

desbo:
This is a server-side test that'll go out to 5% of our audience. Anyone in that segment who views a series article will see the onward container as the first item at the end of an article. We're testing whether the position of the container drives the number of clicks to other articles in the series. 

Before:
screen shot 2016-03-22 at 12 18 35

After (in test):
screen shot 2016-03-22 at 12 21 46

@rich-nguyen


gtrufitt:
:+1:

PR #12320, opened 2016-03-22
Put paid branding decisions behind switch

kelvin-chappell:
There are two ways to decide how a paid container should be branded:
1. Use the backfill query that populated the container
2. Use the metadata of the cards in the container

The latter is the better way but we're currently using the former.
This switch enables us to iron out all the edge cases until we're ready to stop using the backfill query altogether.

/cc @JonNorman


JonNorman:
:+1:

PR #12321, opened 2016-03-22
No 500s from missing tags

JustinPinner:
# Tag combiner requests can fail with 500 errors
## What does this change?

Prevents a 500 if one of the tags in a combiner request doesn't exist in the CAPI results.

We don't like 500s, they're not good for us and we should return a 404 if we can't service the request due to not matching on a query term.

No.


common/app/services/repositories.scala
```diff
@@ -71,13 +71,14 @@ trait Index extends ConciergeRepository with QueryDefaults {
       trails match {
         case Nil => Right(NotFound)
         case head :: _ =>
-          //we can use .head here as the query is guaranteed to return the 2 tags
```
rich-nguyen: !!! I bet @gklopper did this.

(he did)

rich-nguyen:
:+1:

common/app/services/repositories.scala
gklopper: Never trust a manager

PR #12322, opened 2016-03-22
Check if marketingcheckbox exists

gtrufitt:
Another one for @OliverJAsh's static typing campaign :/

### DIFF
--- a/static/src/javascripts/projects/common/modules/email/email.js
+++ b/static/src/javascripts/projects/common/modules/email/email.js
@@ -262,7 +262,8 @@ define([
                 return function (event) {
                     var emailAddress = $('.' + classes.textInput, $form).val(),
                         listId = $('.' + classes.listIdHiddenInput, $form).val(),
-                        markCheckbox = $('.' + classes.markCheckbox, $form)[0].checked ? 'marketing allowed' : 'marketing disallowed',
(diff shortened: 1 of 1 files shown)


NataliaLKB:
:+1:

static/src/javascripts/projects/common/modules/email/email.js
```diff
@@ -262,7 +262,8 @@ define([
                 return function (event) {
                     var emailAddress = $('.' + classes.textInput, $form).val(),
                         listId = $('.' + classes.listIdHiddenInput, $form).val(),
-                        markCheckbox = $('.' + classes.markCheckbox, $form)[0].checked ? 'marketing allowed' : 'marketing disallowed',
+                        $markCheckbox = $('.' + classes.markCheckbox, $form),
+                        markCheckbox = $markCheckbox.length && $markCheckbox[0].checked ? 'marketing allowed' : 'marketing disallowed',
```
OliverJAsh: Was it a `Cannot read property 'foo' of undefined`? :tongue: 

We could use Options to avoid this class of errors.

gtrufitt:
> We could use Options to avoid this class of errors.

Could we? Tell me more! :laughing:

PR #12326, opened 2016-03-22
Use tracking names instead of tracking ids

rich-nguyen:
Omniture is reporting tracking tags with the id of the tag, instead of the user-friendly tag name.


jamespamplin:
:+1: looks fine to me

PR #12328, opened 2016-03-22
Deploys radiator: upgrade deps and break up app

OliverJAsh:
As the title says. This will make it easier to work on :-)


static/src/deploys-radiator/app/model-helpers.ts
```diff
@@ -0,0 +1,23 @@
...
+);
+
+export const isDeployMostRecent = (deploy: Deploy, deploys: List<DeployRecord>): boolean => {
+    const mostRecent = getStartedDeploysFor(deploy.projectName, deploys).last();
+    return mostRecent ? (deploy.uuid === mostRecent.uuid) : false;
+};
+
+export const getMostRecentDeploys = (deploys: List<DeployRecord>): List<DeployRecord> => (
+    deploys.filter(deploy => isDeployMostRecent(deploy, deploys)).toList()
+);
```
TBonnin: Would it be better to give this file have a less generic name like `deploy-helpers` since all functions are related to `Deploys`

static/src/deploys-radiator/app/render.ts
```diff
@@ -0,0 +1,173 @@
...
+                                })
+                                .toList()
+                        ))
+                    ])
+                ))
+                .toList()
+            )
+        ]
+    ]);
+};
```
TBonnin: :eyes: This is not really easy to read but I guess you didn't want to introduce dependencies or write boilerplate to handle viewModels. Right?

TBonnin:
I am not familiar with Typescript at all so it is hard to give a thorough review but other than the html rendering which is quite hard to comprehend I don't have any issue with merging this PR :+1:

static/src/deploys-radiator/app/render.ts
OliverJAsh: When you say you think it's not easy to read, what do you mean exactly? I'm curious! :-)

TBonnin: The multiple levels of indentations and the mix between presentation and logic (what's related to dom element creation and what's inlined businees logic) mostly.
That's definitely subjective though.

johnduffell:
is this going?

static/src/deploys-radiator/app/render.ts
OliverJAsh: Thanks for the feedback, I agree! I will create a model and use that in another PR.

PR #12329, opened 2016-03-23
Separate card lists in commercial containers

kelvin-chappell:
So that the initial and extra lists are separate, which should make the model easier to work with in the template.

/cc @regiskuckaertz


regiskuckaertz:
Yeah! :+1:

PR #12330, opened 2016-03-23
Fix paid popup styling

johnduffell:
This PR is to fix the styling in the paidfor popup on https://www.theguardian.com/guardian-labs

image
changes to
image

This is because there was some conflicting styling on popup and popup__paidfor which were applied to the same divs and therefore depending on the order.  This showed up after th order changed in https://github.com/guardian/frontend/pull/12298

I have basically separated out the conflicting styles into separate classes so it doesn't override any more.  I've also removed the duplicate import of modules/_popup as it's imported in the head anyway.

Thanks for your advice @OliverJAsh 
@Calanthe thanks for diagnosing the problem

### DIFF
--- a/static/src/stylesheets/module/commercial/paidfor/_paidfor-popup.scss
+++ b/static/src/stylesheets/module/commercial/paidfor/_paidfor-popup.scss
@@ -4,9 +4,6 @@
     left: 50%;
     top: 100%;
     padding: 16px;
-}
-
-.paidfor-popup {
     color: #ffffff;
     transform: translate(-50%, 0);
     width: 240px;
@@ -34,13 +31,13 @@
     }
 }
 
-.paidfor-popup__title {
+.popup--paidfor__title {
     font-size: get-font-size(textSans, 3);
     line-height: get-line-height(textSans, 1);
     margin: 0 0 1em;
 }
 
-.paidfor-popup__link {
+.popup--paidfor__link {
     color: #676767;
     font-size: get-font-size(textSans, 2);
     line-height: get-line-height(textSans, 1);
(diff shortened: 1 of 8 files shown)


static/src/stylesheets/module/commercial/paidfor/_paidfor-popup.scss
```diff
@@ -1,12 +1,15 @@
...
     left: 50%;
     top: 100%;
-    width: 240px;
     padding: 16px;
+}
+
+.paidfor-popup {
+    color: #ffffff;
+    transform: translate(-50%, 0);
+    width: 240px;
```
OliverJAsh: Why can't these styles also appear in the `popup--paidfor` modifier?

OliverJAsh:
:+1: 

Less overrides and less dependency on the cascade. Great!

static/src/stylesheets/module/commercial/paidfor/_paidfor-popup.scss
johnduffell: they could, I didn't know how far to go, especially as there are paidfor-popup__<etc> classes as well and I wasn't sure if I should change them and what I should change to if so

OliverJAsh: I guess child objects would become `popup--paidfor__`? That ways it's consistent.

johnduffell: ok I'll do that, sounds sensible.  I can feel the ball of wool unravelling as I pull the end...

PR #12333, opened 2016-03-23
Restore jQuery script to pressed interactive pages that need it.

JustinPinner:
## What does this change?

When we pressed R2 pages we removed all scripts that were not contained within the interactive element. As a result, jQuery got removed but is needed in many cases. This will re-write the jQuery script reference when we re-press from preserved sources.

Restores functionality to degraded interactive pressed pages. Success = pages work again.

No.
## Request for comment

@johnduffell @jennysivapalan
**Note: The Firestorm article may still exhibit problems after re-pressing so don't include it in the batch.**


jennysivapalan:
Nice :+1:

PR #12334, opened 2016-03-23
Use unsafe-url for referrer policy

desbo:
This preserves referrer URLs when navigating from HTTPS->HTTP so Ophan tracking won't break for navigations between protocols.

The danger of this policy is that it'd be possible to find out which secure URLs someone had visited, but that won't really affect us since we redirect https-enabled URLs from http via fastly anyway. When we upgrade to https properly, I'll get rid of this.

https://www.w3.org/TR/referrer-policy/#referrer-policy-state-unsafe-url

@philwills @rich-nguyen


philwills:
:+1: thanks @desbo

PR #12337, opened 2016-03-23
Slightly nicer error messaging when you haven't setup your AWS creds

jamespamplin:
- Produces a better error when you haven't configured your AWS credentials locally
- Use configuration instead of hard coded string for fronts bucket


johnduffell:
great idea! :+1:

PR #12338, opened 2016-03-23
Fix "post comment" button styling

johnduffell:
The Post your comment text was languishing at the bottom of the button itself.  This is of course related to the change of order I introduced in https://github.com/guardian/frontend/pull/12298.

I have rejigged the CSS to not rely on one padding overriding the other - only one should apply at a time.

I could have used extend or mixin, but after googling, it seems mixin wins out slightly over extend!

@OliverJAsh thanks for the help!

PS the file in pasteup https://github.com/guardian/pasteup-forms/blob/master/src/_forms.scss doesn't seem to match the _forms.scss that I edited so it presumably isn't being used by the build as we just have another copy?

### DIFF
--- a/static/src/stylesheets/components/pasteup-forms/src/_forms.scss
+++ b/static/src/stylesheets/components/pasteup-forms/src/_forms.scss
@@ -250,7 +250,7 @@ category: Common
     min-height: ($gs-baseline/3)*40;
 }
 
-@mixin submit-input--behaviour {
+.submit-input, .submit-input--behaviour {
     background: $pasteup-forms-submit-background;
     border: 0 none;
     color: #ffffff;
@@ -273,11 +273,7 @@ category: Common
 }
 
 .submit-input {
-    @include submit-input--behaviour;
     padding: 11px $gs-gutter/2;
-}
-.submit-input--behaviour {
-    @include submit-input--behaviour;
 }
 
 .submit-input[disabled] {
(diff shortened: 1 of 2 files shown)


static/src/stylesheets/components/pasteup-forms/src/_forms.scss
```diff
@@ -273,6 +272,14 @@ category: Common
...
 }
 
+.submit-input {
+    @include submit-input--behaviour;
+    padding: 11px $gs-gutter/2;
+}
+.submit-input--behaviour {
+    @include submit-input--behaviour;
+}
+
```
OliverJAsh: You could do this without mixins:

``` css
.submit-input,
.submit-input--behaviour {
   /* common styles */
}

.submit-input {
  padding: foo;
}
```

Also, this is not how we usually use BEM. Modifiers are supposed to be used in combination with the block class, e.g. `submit-input submit-input--behaviour`, in order to compose styles.

Perhaps the block class should be all the common styles, and the modifier should just contain the padding? This would mean we have to opt-in to it everywhere else in the codebase though.

johnduffell:
@OliverJAsh I went for the simplest version of what you said.  I'm feeling like our CSS relies on this cascade in a lot of places.

OliverJAsh:
> PS the file in pasteup https://github.com/guardian/pasteup-forms/blob/master/src/_forms.scss doesn't seem to match the _forms.scss that I edited so it presumably isn't being used by the build as we just have another copy?

Bower requires us to check in downloaded code to guarantee version correctness. The upstream version should match what we have in our codebase, unless someone has accidentally edited our downstream version (easy mistake to make).

rich-nguyen:
:+1:

PR #12339, opened 2016-03-23
Refactor breakingnews and add tests

sndrs:
adds tests for breaking news, following the bug that went unnoticed and was fixed in #12224 

this is, shamefully, my first real set of tests. at this point, they seem like overkill, but they exist now. feedback v welcome.

cc @stephanfowler @OliverJAsh @desbo @TBonnin


static/src/javascripts/projects/common/modules/onward/breaking-news.js
```diff
@@ -32,137 +32,197 @@ define([
...
+            alerts
+                .filter(function (alert) {return alert.href === 'global';})
+                .map(function (alert) {return alert.content;}),
+            alerts
+                .filter(function (alert) {return alert.href === edition;})
+                .map(function (alert) {return alert.content;}),
+            alerts
+                .filter(function (alert) {return section && alert.href === section;})
+                .map(function (alert) {return alert.content;})
+        ]);
```
OliverJAsh: Could we remove these three `map`s and just have one at the end of the `flatten`?

static/test/javascripts/spec/common/onward/breaking-news.spec.js
```diff
@@ -0,0 +1,286 @@
...
+
+        describe('user cannot dismiss alerts', function () {
+            beforeEach(function () {
+                sinon.stub(storage.local, 'isAvailable').returns(false);
+            });
+
+            it('should not try and fetch the json', function (done) {
+                mockBreakingNewsWith([]).then(function () {
+                    fail('user cannot use local storage, but we seem to think things are okish');
+                    done();
```
gtrufitt: Not worth changing it all now, but I wonder whether we could use `.finally(done)` (if we're using when.js) as you have it in all `resolve` and `reject`. I am not sure why the es6 implementation doesn't have that method though. Apparently you could do `.then(done)` (after the catch if there is one): https://gist.github.com/jish/e9bcd75e391a2b21206b (And seems like Matt Andrews had this though at some point too... https://github.com/matthew-andrews/Promise.prototype.finally)

Anyway, you have lots of these so I am more talking out loud than recommending changing it :)

gtrufitt: Oh and in the situations you had:
```
function (e) {
   fail(e);
   done();
}
```
You could have had: 
```
.then(function () {..}, fail).finally(done)
```

I think!

static/src/javascripts/projects/common/modules/onward/breaking-news.js
```diff
@@ -32,137 +32,197 @@ define([
...
+
+    // don't show alerts if they're over a certain age
+    function filterAlertsByAge(alerts) {
+        return alerts.filter(function (alert) {
+            var alertTime = alert.frontPublicationDate;
+            return alertTime && relativeDates.isWithinSeconds(new Date(alertTime), 1200); // 20 mins
+        });
+    }
+
+    // don't show alerts if they're over a certain age
```
gtrufitt: Don't think this comment applies to this function :)

static/src/javascripts/projects/common/modules/onward/breaking-news.js
```diff
@@ -32,137 +32,197 @@ define([
...
+            return fetchBreakingNews()
+                .then(parseResponse)
+                .then(getRelevantAlerts)
+                .then(pruneKnownAlertIDs)
+                .then(filterAlertsByDismissed)
+                .then(filterAlertsByAge)
+                .then(pickNewest)
+                .then(alert);
+        } else {
+            return Promise.reject('cannot dismiss');
```
OliverJAsh: You can reject the promise with an `Error` object so the browser will log a stack trace.

static/src/javascripts/projects/common/modules/onward/breaking-news.js
desbo: seems like you could get rid of `flatten` altogether with one `filter` and a condition that combines those three checks

static/src/javascripts/projects/common/modules/onward/breaking-news.js
gtrufitt: Not really an error though?

static/src/javascripts/projects/common/modules/onward/breaking-news.js
OliverJAsh: I had wondered the same but I think the separate filers are done to deliberately effect ordering, as per the comment above: `global > current edition > current section`.

static/test/javascripts/spec/common/onward/breaking-news.spec.js
piuccio: The best way (IMO) to do this is Jasmine (which unfortunately doesn't support promises like mocha or others) is

```js
mockBreakingNewsWith([]).then(() => {
   done.fail('message')
}, () => {
   // some expectations
})
.then(done)
.catch(done.fail)
```

It's easier when you're testing for resolved promises

```js
returnsAPromise()
.then(() => {  /* expectations */ })
.then(done)
.catch(done.fail)
```

gtrufitt:
BN code is much easier to reason about and a nice set of tests :+1: from me.

static/src/javascripts/projects/common/modules/onward/breaking-news.js
OliverJAsh: A rejected promise is the same thing as an async error. If this isn't an error maybe this function should just return nothing instead of a rejected promise?

static/test/javascripts/spec/common/onward/breaking-news.spec.js
```diff
@@ -0,0 +1,286 @@
...
+                    })]);
+                    breakingNews().then(function(result) {
+                        // make sure the DOM has finished updating
+                        requestAnimationFrame(function () {
+                            resolve(result);
+                        });
+                    }, reject);
+                    server.restore();
+                }, function (e) {
+                    console.log(e);
```
piuccio: is this what you really want?

piuccio:
As we're often complaining about the organization of our files, I'd split `common/modules/onward/breaking-news.js` in two, one file (call it **the model**) that does fetch and filter, and another one (call it **the view**) that does dom manipulation.

The model can do all you're doing until line 153 and then

```js
function alert (alert) {
   if (alert) {
      mediator.emit('breaking-news:show', alert);
   }
}
```

It'll come handy when we migrate to react, because all you have to change is **the view**. In the meantime (or if we never migrate (which we shouldn't, imo)) your files are smaller, loosely coupled and have a single responsibility; which is better according to some.

As I'm not in any frontend teams, don't let my comments stop your deploy, it's just food for thoughts.

johnduffell:
yay for tests!! :tada: :balloon:

static/test/javascripts/spec/common/onward/breaking-news.spec.js
sndrs: `done.fail` – nice!

static/src/javascripts/projects/common/modules/onward/breaking-news.js
sndrs: no...

static/src/javascripts/projects/common/modules/onward/breaking-news.js
sndrs: it's only rejecting for testing, kind of belts and braces. the test could be less robust (just check nothing's happened)?

static/test/javascripts/spec/common/onward/breaking-news.spec.js
sndrs: is it not? :)

it's just because it turns out we override `inject.require` with a promised version, and for ages i couldn't work out why nothing was working. turns out the promise i didn't know existed was rejecting but i wasn't catching it i.e. it was test code issue, not code that was being tested. just there incase it happens again.

piuccio: ok, I don't know how injector works, but it felt weird that if it fails you log an error and then your test is stuck because you're not resolving or rejecting the 'parent' promise (line 43).

static/test/javascripts/spec/common/onward/breaking-news.spec.js
```diff
@@ -0,0 +1,286 @@
...
+                        webTitle: 'Breaking News',
+                        collections: collections
+                    })]);
+                    breakingNews().then(function(result) {
+                        // make sure the DOM has finished updating
+                        requestAnimationFrame(function () {
+                            resolve(result);
+                        });
+                    }, reject);
+                    server.restore();
```
piuccio: just noticed this, I'm sure it works fine but it's a bit brittle, it relies on the fact that breaking news is synchronously issuing the request and the mock server implementation synchronously registers it.

Something more robust is

```js
Promise.resolve().then(() => {
   return breakingNews()
})
.then(() => {
   server.restore();
   raf(() => resolve(result));
})
.catch(err => {
   server.restore();
   reject(err)
})
```

Or 
```js
Promise.resolve().then(() => {
   return breakingNews()
})
.then(() => {
   raf(() => resolve(result));
})
.catch(reject)
.then(() => {
   server.restore();
})
```

The first `Promise.resolve()` is extra safety in case `breakingNews()` throws an exception. Exceptions in the `.then` callback make the promise fail, while exceptions in a `Promise` constructor blow up your code badly.

static/src/javascripts/projects/common/modules/onward/breaking-news.js
sndrs: yeah i agree it feels weird. i did fiddle with it for a bit actually, but i made a mistake somewhere in it, and i didn't really want to change the way it worked, just reorder it to make it easier to test, so i changed it back. this is how it works currently. maybe i'll have a quick 2nd look

static/test/javascripts/spec/common/onward/breaking-news.spec.js
```diff
@@ -52,13 +52,16 @@ define([
                         webTitle: 'Breaking News',
                         collections: collections
                     })]);
-                    breakingNews().then(function(result) {
+                    Promise.resolve().then(function () {
```
desbo: why do you need this first promise?

sndrs: based on @piuccio's recommendation, that

> The first Promise.resolve() is extra safety in case breakingNews() throws an exception. 

seems sensible to me?

OliverJAsh: Out of interest, is there a reason why you have to coerce the exception into a rejected promise? Is it so that you can only have one `catch`, instead of two (like below)?

``` js
try {
  breakingNews().catch(reject);
} catch (e) {
  reject(e);
}
```

Related, I found this article interesting (which also documents your `Promise.resolve` solution): [Promise-based functions should not throw exceptions](http://www.2ality.com/2016/03/promise-rejections-vs-exceptions.html)

sndrs:
awaiting #12430 for merge

PR #12340, opened 2016-03-23
Remove ruby

rich-nguyen:
This PR removes ruby and its packages: sass, scss-lint and hologram. We already use node-sass, the only thing left was to port our scss-lint task to sass-lint.

Surprisingly, lots of linting errors came through, most of it genuine stuff not picked up in old linter. This is because they have different rulesets and parsing behaviour. Lint time is 10 seconds quicker too. Still, I don't really mind which one we use, it's just odd that we still need Ruby, just for sass linting.


jamespamplin:
:+1: looks good to me, great stuff!

piuccio:
image

PR #12341, opened 2016-03-23
place cta so that it remains in place on updates

NathanielBennett:
# My Awesome Pull Request
## What does this change?

Fixes the Notifications CTA in a different DOM position so that it isn't shifted down on subsequent updates

No
## Screenshots

Before: 
before

After:
after
## 

### DIFF
--- a/common/app/views/fragments/liveNotificationsCallToActionPlaceHolder.scala.html
+++ b/common/app/views/fragments/liveNotificationsCallToActionPlaceHolder.scala.html
@@ -1,0 +1,2 @@
+@()(implicit request: RequestHeader)
+<aside id="notifications" class="notifications js-notifications" role="complementary"></aside>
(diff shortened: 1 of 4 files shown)


common/app/views/fragments/liveNotificationsCallToActionPlaceHolder.scala.html
```diff
@@ -0,0 +1,2 @@
+@()(implicit request: RequestHeader)
+<aside id="notifications" class="notifications js-notifications" role="complementary"></aside>
```
gtrufitt: Good use of aside and role

static/src/stylesheets/module/preferences/_notifications.scss
```diff
@@ -2,11 +2,13 @@
     display: none;
 }
 
-.from-content-api .notifications-subscribe-link__headline {
+.notifications-subscribe-link__headline {
```
gtrufitt: Yay, I hate having to prefix with `.from-content-api` :D

gtrufitt:
:+1:

PR #12342, opened 2016-03-23
Revert "Merge pull request #12283 from guardian/remove-foresee"

rich-nguyen:
We need to revert this because the adops survey is not ready yet.


johnduffell:
fair enough :+1:

PR #12344, opened 2016-03-24
create video status endpoint

jamesgorrie:
# Get video info endpoint

We currently need to know whether videos are expired as we don't have this information available from flex when the video is embedded. 

I could have used the [video json](http://gu.com/film/video/2016/mar/01/sacha-baron-cohan-appears-as-ali-g-at-oscars-video.json) as it has the `"status": "GONE"` but the file is a few kilobytes big, and this call might happen across multiple videos on a page.

There is also a question around caching, but it seems the media endpoints don't have much in place for that? Is this correct, or shall we be changing that? @sndrs?
## What does this change?

Nothing, it adds value.

We will have an endpoint, it serves JSON, valuable JSON.

Nope, but we eventually be doing something similar there.
## @akash1810 

### DIFF
--- a/dev-build/conf/routes
+++ b/dev-build/conf/routes
@@ -411,6 +411,7 @@ GET            /$path<[\w\d-]*(/[\w\d-]*)?/gallery/.*>                          
 GET            /$path<[\w\d-]*(/[\w\d-]*)?/(cartoon|graphic|picture)/.*>.json                                                    controllers.ImageContentController.renderJson(path)
 GET            /$path<[\w\d-]*(/[\w\d-]*)?/(cartoon|graphic|picture)/.*>                                                         controllers.ImageContentController.render(path)
 
+GET            /$path<[\w\d-]*(/[\w\d-]*)?/(video|audio)/.*>/info.json                                                           controllers.MediaController.renderInfoJson(path)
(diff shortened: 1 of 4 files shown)


dev-build/conf/routes
```diff
@@ -411,6 +411,7 @@ GET            /$path<[\w\d-]*(/[\w\d-]*)?/gallery/.*>
 GET            /$path<[\w\d-]*(/[\w\d-]*)?/(cartoon|graphic|picture)/.*>.json                                                    controllers.ImageContentController.renderJson(path)
 GET            /$path<[\w\d-]*(/[\w\d-]*)?/(cartoon|graphic|picture)/.*>                                                         controllers.ImageContentController.render(path)
```
gtrufitt: Think you also need to update standalone routes

akash1810:
:+1:

dev-build/conf/routes
jamesgorrie: What do you mean?

gtrufitt: This file: https://github.com/guardian/frontend/blob/master/standalone/conf/standalone.routes#L139 Do you need to add your new route there? IIRC standalone is used for preview or something - @rich-nguyen ?

rich-nguyen: Correct, @gtrufitt

jamesgorrie: What is that file for?

jamesgorrie: Just reread it, sorry, for preview.

PR #12347, opened 2016-03-24
Header bidding test audience changes

Calanthe:
## What does this change?

The header bidding AB test's audience is increased to 100% (still US only).


uplne:
:+1:

PR #12348, opened 2016-03-24
Revert "float meta content at leftcol breakpoint, so that it stays in the flow"

sndrs:
Reverts guardian/frontend#12301


jamespamplin:
:+1: :shipit:

PR #12349, opened 2016-03-24
Some refactoring and make logos work

kelvin-chappell:
/cc @regiskuckaertz

### DIFF
--- a/commercial/app/views/jobs/jobsV2.scala.html
+++ b/commercial/app/views/jobs/jobsV2.scala.html
@@ -5,6 +5,8 @@
   omnitureId: Option[String],
   clickMacro: Option[String])(implicit request: RequestHeader)
 
+@import views.html.commercial.containerWrapper
+
 @containerWrapper(Seq("legacy", "jobs", "tone-jobs"),
                   omnitureId.map(id => s"merchandising | jobs | ${id}"),
                   rowLayout = Some("2x1x1"),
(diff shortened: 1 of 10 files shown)


commercial/app/views/jobs/jobsV2.scala.html
```diff
@@ -5,6 +5,8 @@
   omnitureId: Option[String],
   clickMacro: Option[String])(implicit request: RequestHeader)
 
+@import views.html.commercial.containerWrapper
+
```
regiskuckaertz: Isn't `@import jobCard` missing here?

common/app/views/commercial/containerWrapper.scala.html
```diff
@@ -28,6 +28,6 @@ <h1 class="adverts__title">@title</h1>
     </header>
     <div class="adverts__body">
         @innards
-        @for(badge        <- optBadge)   {<div class="adverts__badge js-badge">@badge</div>}
```
regiskuckaertz: That line remains necessary for the two cAPI components

regiskuckaertz:
:+1: aside from minor comments, thanks a lot!

commercial/app/views/jobs/jobsV2.scala.html
kelvin-chappell: It was imported at the top of the file but I've moved it down now to make it more consistent.

common/app/views/commercial/containerWrapper.scala.html
kelvin-chappell: Oops! Reinstated it now.

PR #12352, opened 2016-03-24
Make switch expiry time clearer

kelvin-chappell:
This is a very sophisticated change to give a better expectation of when switches will expire.

/cc @johnduffell


johnduffell:
haha oh yeah good spot! :+1: :speedboat:

piuccio:
how about 5pm? that's when we start leaving the office

piuccio:
oh that's not when it actually expires, never mind

johnduffell:
the expiry is just when the builds will start failing

PR #12353, opened 2016-03-24
fix admin creative templates page to show creatives

JonNorman:
Fixes up the admin creatives page:
- the creatives now display underneath their respective templates
- removed the example creative in favour of using the `previewUrl` from the first creative of each template.

_cc_ @kelvin-chappell


kelvin-chappell:
Looks good :+1:

PR #12359, opened 2016-03-29
Fix ansi-styles version

desbo:
One of the dependencies in our `npm-shrinkwrap.json` was unpublished (https://github.com/chalk/ansi-styles/issues/17) so we need to update it.

This definitely took longer than doing a manual find and replace and I nearly went insane writing it, but I imagine it'll happen again so here's the command to do it:

```
grep -nHC1 "2\.2\.0" npm-shrinkwrap.json | grep -A1 ansi-styles | grep npm-shrinkwrap.json: | cut -d: -f2 | xargs -I"{}" -n1 sed -i {}s/\.0/\.1/ npm-shrinkwrap.json
```

@sammorrisdesign @jbreckmckye

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -770,7 +770,7 @@
           "version": "1.1.1",
           "dependencies": {
             "ansi-styles": {
-              "version": "2.2.0",
+              "version": "2.2.1",
               "dependencies": {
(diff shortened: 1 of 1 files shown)


jbreckmckye:
:+1:

PR #12357, opened 2016-03-24
Test new membership propositions

joelochlann:
This is a test of 3 different variants of a new membership proposition. Three thrashers to display on :

picture 181
picture 186
picture 187

And three landing pages:

screencapture-m-thegulocal-com-commercial-guardian-explore-1458839675438
screencapture-m-thegulocal-com-commercial-guardian-experience-1458839694702
screencapture-m-thegulocal-com-commercial-guardian-explain-1458839836924

@Calanthe 
- [x] Resolve conflicts
- [x] Rename a few things to make them less general, as per @Calanthe's comments
- [x] Display on home (UK) front, uk news front, and politics front
- [ ] Refine styling of thrashers
- [ ] Optimise/remove images as per @Calanthe's comments

PR #12360, opened 2016-03-29
Set test expiry to weekday

joelochlann:
Sets the membership test expiry to a weekday to fix a test failure caused by #12357. (I accidentally merged that PR with a failing test)

@jbreckmckye


jbreckmckye:
:+1:

PR #12362, opened 2016-03-29
Catch ad callback exceptions

jbreckmckye:
On pages like [this](http://www.theguardian.com/commentisfree/audio/2016/mar/17/podcast-feminism-jessica-valenti), a DFP advert callback for creating a sticky right-column ad throws an ugly exception:

image

This pull request:
- fixes the error (guards against trying to find an element in an array which is null)
- makes sure the 'parseAds' stage of the ad lifecycle is monitored by Sentry

Now our unhandled error is logged with an accompanying console warning:

image


desbo:
:+1:

PR #12363, opened 2016-03-29
Fix HTML escaping in ExactTarget variables

desbo:
to stop this happening:
screen shot 2016-03-29 at 15 17 25


jbreckmckye:
:+1:

Issue #10689, opened 2015-09-29
Special Characters not displaying correctly when user signed in


ScottPainterGNM:
See - https://github.com/guardian/frontend/issues/4012#issuecomment-72757249

Looks like in tags and on profile pages it displays correctly but not so when displaying a signed in users name on the nav bar

![image](https://cloud.githubusercontent.com/assets/11922253/10160331/b0935c3c-6694-11e5-874c-f1d8673f4fec.png)

![image](https://cloud.githubusercontent.com/assets/11922253/10160346/c3da2ef6-6694-11e5-999e-dd7b71aa3f8f.png)

ScottPainterGNM:
@paperboyo FYI

gtrufitt:
I've raised this with the identity team, and it's scheduled to be fixed as part of their Q3 work, as well as being able to register with diacritics as part of a username (which is currently not possible, so made it difficult to test the original issue!)

paperboyo:
That's great, thąnks, @gtrufitt!

nlindblad:
Fixed in https://github.com/guardian/frontend/pull/12364 and on the Identity side.

<img width="574" alt="screen shot 2016-04-05 at 12 39 29" src="https://cloud.githubusercontent.com/assets/164991/14280594/9e636cba-fb2b-11e5-8049-066eb90e5cce.png">

PR #12364, opened 2016-03-29
Fix display name UTF-8 decoding issue

niklasvincent:
# Fix display name UTF-8 decoding issue
## What does this change?

Currently display names containing special characters don't display correctly in the header, which is bad from an internationalisation point of view.

The Identity team is looking at URI encoding the display name before it gets passed to the cookie, which means client side implementations will have to run the display name field through `decodeURIComponent(str)`.

This PR contains the necessary front-end changes to comply with the new encoding.

**Before:**

screen shot 2016-03-29 at 12 58 38

```
window.guardian.config.user
Object {id: "21840015", displayName: "Niklas JÃ´se Lindblad", accountCreatedDate: 1459244427000, emailVerified: true, rawResponse: "WyIyMTg0MDAxNSIsIiIsIk5pa2xhcyBKw7RzZSBMaW5kYmxhZC…zZJOkfDTCcWQOOR7CAtAhQTeije3RGtlpIn3EnVhJDBT0jkuQ"}
```

**After:**

screen shot 2016-03-29 at 12 58 33

```
window.guardian.config.user
Object {id: "21840015", displayName: "Niklas Jôse Lindblad", accountCreatedDate: 1459244427000, emailVerified: true, rawResponse: "WyIyMTg0MDAxNSIsIiIsIk5pa2xhcyBKw7RzZSBMaW5kYmxhZC…zZJOkfDTCcWQOOR7CAtAhQTeije3RGtlpIn3EnVhJDBT0jkuQ"}
```

It provides a nicer user experience by displaying people's names correctly.

User Help has seen a few e-mails on this topic, which should hopefully drop to zero after this fix.

Unless Amps uses display name, otherwise no.
## Screenshots

**Before:**

screen shot 2016-03-29 at 12 58 38

**After:**

screen shot 2016-03-29 at 12 58 33
## Request for comment

Would really appreciate feedback on this change.
## 


common/app/templates/inlineJS/nonBlocking/getUserData.scala.js
```diff
@@ -5,7 +5,7 @@ try {
     ((isVeryModern, document, window) => {
         if (isVeryModern) {
             function decodeBase64(str) {
-                return decodeURIComponent(encodeURIComponent(atob(str.replace(/-/g, '+').replace(/_/g, '/').replace(/,/g, '='))));
+                return decodeURIComponent(escape(atob(str.replace(/-/g, '+').replace(/_/g, '/').replace(/,/g, '='))));
```
jbreckmckye: Isn't `escape` deprecated?

paperboyo:
:gem:

Finally fixes https://github.com/guardian/frontend/issues/10689. Thanks a lot, @nlindblad!

static/test/javascripts/spec/common/identity/Api.spec.js
```diff
@@ -34,8 +34,8 @@ define([
                 };
                                getCookieStub = sinon.stub();
                                getCookieStub.withArgs('GU_U').returns(
-                                   'WyIyMzEwOTU5IiwiamdvcnJpZUBnbWFpbC5jb20iLCJqYW1lc2dvcnJpZSIsIjUzNCIsMTM4Mjk1MzAzMT' +
-                                   'U5MSwxXQ.MC0CFBsFwIEITO91EGONK4puyO2ZgGQcAhUAqRa7PVDCoAjrbnJNYYvMFec4fAY'
+                                   'WyIyMzEwOTU5IiwiamdvcnJpZUBnbWFpbC5jb20iLCJBbcOpbGllIErDtHNlIiwiNTM0IiwxMzgyOTUzMDMxNTkxLDFd' +
```
jbreckmckye: This won't test the JavaScript you've changed.

In fact, we currently have no easy way to test the inlined scripts, excepting Selenium tests (which we probably don't want to use).

johnduffell:
this is my PR of the day! :grinning: our users will be happy again, thanks

static/test/javascripts/spec/common/identity/Api.spec.js
johnduffell: agreed about the selenium tests I wonder what such a test would achieve - if we did break it, we would notice pretty fast!

nlindblad: @jbreckmckye: Yeah, I should add a note on that. I just wanted to make sure that the fallback is doing the right thing as well (which it does already, but without any tests verifying that).

ScottPainterGNM:
:gem:

common/app/templates/inlineJS/nonBlocking/getUserData.scala.js
desbo: I think we can get rid of it now that we're encoding names in identity.

jamespamplin:
:+1: great stuff! Glad you found a way around using `escape` :)

sndrs:
:100: :star:

PR #12365, opened 2016-03-29
Remove AB test - Marketing Checkbox

gtrufitt:
Remove the marketing checkbox A/B test


common/app/views/fragments/email/signup/emailSignUp.scala.html
```diff
@@ -43,7 +43,6 @@
                 <input class="js-email-sub__listid-input" type="hidden" name="listId" value="@listId" />
             </div>
             <button type="submit" class="email-sub__submit-input js-email-sub__submit-input button button--tertiary button--large">@fragments.inlineSvg("envelope", "icon", Seq("submit-input__icon"))@submitText</button>
-            <small class="email-sub__small js-email-sub__small email-sub__small--is-hidden">@smallText</small>
```
desbo: looks like we can get rid of the `smallText` parameter as well

gtrufitt: Good shout, removed that. Is just using the default param, so not passed in anywhere.

PR #12366, opened 2016-03-29
Add kicker field to commercial card content

kelvin-chappell:
And there's some refactoring too!
And tests.  :egg:


JonNorman:
Looks good! The card model is a lot clearer now and the tests look a lot more effective!
:+1:

PR #12370, opened 2016-03-29
Fix js error in admin AB tests dashboard

rich-nguyen:
The detect ad blocker system expects a valid key in the guardian config object. I reckon it should always be present, rather than guarded, so added it to the admin_main template


desbo:
:+1:

sndrs:
ooh nice one

PR #12371, opened 2016-03-29
Expose each membership test variant to 3% of users

joelochlann:
Turn on #12357


johnduffell:
this will be a good test! look forward to the results :+1:

PR #12373, opened 2016-03-30
show video expired error if video is expired

jamesgorrie:
# Add Expired video message if this is the case
## What does this change?

We now show a expired message if the video has expired.
The best solution here would be to have the CAPI response know if a video is expired to avoid the XHR call, and this will be coming when we introduce the video atom, this is a stopgap for now.

Hopefully we don't get charged for the use of expired video.

No - but they might have to do a similar thing.
## Screenshots

vide-removed
## Request for comment

@akash1810 
@rrees - is this similar to what you were thinking?
## 


static/src/javascripts/bootstraps/enhanced/media/main.js
```diff
@@ -256,126 +274,141 @@ define([
...
-                    events.bindGlobalEvents(player);
-                    events.bindContentEvents(player);
-                    if (withPreroll) {
-                        events.bindPrerollEvents(player);
-                    }
+        isPlayerExpired.then(function(expired) {
+            if (expired) {
+                player.ready(function() {
+                    player.error({
+                        code: 0,
```
akash1810: What does this code relate to? I'm tempted to say its not an HTTP code?

jamesgorrie: No. As we aren't 100% sure why something might be invalid i.e. it's our own logic.

akash1810:
Some of the tests around expiry are failing. Overall it looks ok (the diff looks a lot bigger than it is...)

jamesgorrie:
> Some of the tests around expiry are failing.

I think I have old A/B tests that make it fail.

> Overall it looks ok

Only OK?

>  (the diff looks a lot bigger than it is...)

That's what happens when you add an `if`.

rrees:
Yes, this is what I was thinking of. I don't the code that well so it looks fine to me :shipit: 

However the big if-else might be more elegant if the content of the branches was extracted into functions. Then it would be easier to change the behaviour of the valid/invalid branch in future.

akash1810:
> Only OK?

Oops! :+1: :tada: :fireworks: 

Though that merging of master is making me want to retract some of that excitement!

jamesgorrie:
> was extracted into functions

I would, but this code is a little scarey and has a lot of dependant variables (which are, on occasion, mutated), so it'll just be passing those into other functions. OIf other player details aren't sorted soon, I'll have a bash at cleaning this up.

PR #12374, opened 2016-03-30
Extending top banner position test for another month

uplne:
For 1% of happy users. There is still no decision what to do with this so nothing changed from the last time but we NEED this to stay in PROD.


jamesgorrie:
Does the data show conclusive evidence? 

:+1: to the code.

uplne:
@jamesgorrie it's tricky because data are proving that viewability metric is worse (as expected) but user experience is amazing. That's why it is difficult to fully commit to 100% of the audience.

Issue #8118, opened 2015-01-30
Video trails twitch on fronts before play


paperboyo:
Hello,

All desktop platforms/browsers tested. Go to any front with a video player(s), start any video.

Before hitting play, there is a trail pic shown, after hitting play but before video starts, the same trail displays but cropped differently, after that I am not sure what happens myself (the trail is shown the third time again cropped differently?).

The same videos play correctly on their respective video pages. Ideally, the videos of fronts should play as cleanly as on video pages.

Below a presentation of above in slow motion (buffering courtesy of Virgin’s ‘Superfast and super value Up to 120Mb Broadband’):

![twitch1](https://cloud.githubusercontent.com/assets/6032869/5984849/577e2440-a8d2-11e4-97fd-0619e4cabeb3.gif)

Regards
Mateusz

johnduffell:
we haven't got around to fixing this yet so it's probably not going to be fixed any time soon

paperboyo:
Ok, I _think_ it may be because there is a progression of trail pic>poster frame[>possibly another poster frame]. We should just get rid of poster frames on fronts. This would shave e.g. ~1,7MB off current video front. We will save on bandwidth and will have no twitching. I may be wrong, though!

PR #12375, opened 2016-03-30
don't show poster on fronts

jamesgorrie:
# Don't use posters on fronts

As trail slots and videos are of a different ratio - so are the trail images and video posters.
This creates a jerky experience when pressing play on a video on a front when you go from trail image => poster => video.

We also save some bandwidth as per #8118.

Better UX for all and bandwidth.

Nope
## Screenshots

It's hard to Gif as it's milliseconds, but just press play [here](http://theguardian.com/video) for the bad experience and with a little imagination, you can image that it doesn't.
## Request for comment

@akash1810 @paperboyo
## 

### DIFF
--- a/common/app/views/fragments/items/facia_cards/contentCard.scala.html
+++ b/common/app/views/fragments/items/facia_cards/contentCard.scala.html
@@ -58,7 +58,7 @@
                 )) { player =>
                     <div class="fc-item__media-wrapper u-faux-block-link__promote media__container--hidden js-video-player">
                         <div class="fc-item__video-container">
-                            @video(player, false, item.cardTypes.showVideoEndSlate)
+                            @video(player, false, item.cardTypes.showVideoEndSlate, showPoster = false)
                         </div>
                     </div>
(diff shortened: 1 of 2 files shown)


paperboyo:
Fixes https://github.com/guardian/frontend/issues/8118.

Savings depend on number of videos on a given front and requested image format (FF: current UK front: -100KB, Culture: -250KB, Video: -1,6MB).

Thanks a lot, @jamesgorrie! :beer:

akash1810:
Interesting :+1:.

You can see the current state with network throttling (native in Chrome, or use Charles). The poster image helps the UX *a little* whilst the video is buffering in the background.

paperboyo:
> The poster image helps the UX a little whilst the video is buffering in the background.

True that. But it's often not synchronised with the trail (sometime it's a completely different pic), so that there is this twitching progression shown in https://github.com/guardian/frontend/issues/8118 (worse, the faster the connection/machine is). I, personally, think the savings and less twitching are worth it, but not my call obviously...

PR #12376, opened 2016-03-30
Adjust start & expiry of membership test to run for a week

joelochlann:
This test should run for a week. The start & expiry dates didn't reflect the actual time I started running the test (yesterday).


jamesgorrie:
:+1:

PR #12377, opened 2016-03-30
Admin login: Allow multiple oauth callback urls

TBonnin:
## What does this change?

Allow to configure multiple oauth callbacks for the admin pages depending on the domain it is accessed from

Short term goal: keep a way to access admin page while changing frontend.gutools.co.uk dns 

No
## Request for comment

@dominickendrick

### DIFF
--- a/admin/app/conf/AdminConfiguration.scala
+++ b/admin/app/conf/AdminConfiguration.scala
@@ -1,7 +1,7 @@
 package conf
 
 import com.gu.conf.ConfigurationFactory
-import conf.Configuration.OAuthCredentials
+import conf.Configuration.OAuthCredentialsWithMultipleCallbacks
 
 case class OmnitureCredentials(userName: String, secret: String)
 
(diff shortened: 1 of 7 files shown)


admin/app/conf/AdminConfiguration.scala
```diff
@@ -42,12 +42,13 @@ object AdminConfiguration {
     lazy val appName = configuration.getStringProperty("api.dfp.applicationName")
   }
 
-  lazy val oauthCredentials: Option[OAuthCredentials] =
+  lazy val oauthCredentials: Option[OAuthCredentialsWithMultipleCallbacks] =
```
dominickendrick: Could this be rename to `OAuthCredentialsWithMultipleHosts`

TBonnin: I went the other way and instead of having 2 properties and then reassemble the callback url, I now store in the config a list of full callback urls. Tell me what you think

common/conf/env/DEV.properties
```diff
@@ -77,7 +77,7 @@ faciatool.show_test_containers=true
 
 #Admin Oauth
 admin.oauth.clientid=265127054270-97f63kb58ns533alnhlb1vj4jds79q2f.apps.googleusercontent.com
-admin.oauth.callback=http://localhost:9000/oauth2callback
+admin.oauth.callbacks=http://localhost:9000/oauth2callback,http://127.0.0.1:9000/oauth2callback
```
TBonnin: No need for `127.0.0.1` to be honest but this is how I tested it locally. :)

dominickendrick:
👍

PR #12378, opened 2016-03-30
Manually floor header-bidding CPMs to appropriate buckets

jbreckmckye:
This fixes an issue where prices between $5 and $15 were being artificially capped at $5.00, because Prebid.js applies that ceiling to the low-granularity CPM value (but none of the others, bizarrely).

This change puts the CPM bucketing behaviour squarely in our control.


jamesgorrie:
:pencil2: :+1:

PR #12380, opened 2016-03-30
Git ignore /data and /docs/styleguide

joelochlann:
These both seem to contain autogenerated files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 artifacts.zip
 /dist
+/data
 
 frontend-article.properties
 frontend-core-navigation.properties


rich-nguyen:
`/data` is used by the CI build to stub external dependencies, and `/docs/styleguide` shouldn't be used anymore. Is there anything in `styleguide`? Those files were recently removed.

joelochlann:
Ah sorry, I needed to run a `git clean -fd` to remove `/docs/styleguide`. Does it make sense to ignore `/data` though?

rich-nguyen:
not really, because the CI build will perform a git checkout of our codebase, and will read those files during test. It makes sense for `data` to be more understandable, though. We could add a readme, or explainer?

PR #12312, opened 2016-03-22
Adblocking response AB test - Firefox excluded from the test

Calanthe:
## What does this change?

Firefox is now excluded from the Adblocking response AB test. 
## 

_Does this PR meet the [contributing guidelines](https://github.com/guardian/frontend/blob/issue_pr_templates/.github/CONTRIBUTING.md#submission)?_

PR #12302, opened 2016-03-21
Final changes for the adblocking response test

Calanthe:
## What does this change?

Minor changes in link styling, membership and signup links. 
## Screenshots

screen shot 2016-03-21 at 15 13 25
## 

_Does this PR meet the [contributing guidelines](https://github.com/guardian/frontend/blob/issue_pr_templates/.github/CONTRIBUTING.md#submission)?_

PR #12299, opened 2016-03-21
Adblocking response AB test - overlay text and data links changes

Calanthe:
## What does this change?

The continuation of the https://github.com/guardian/frontend/pull/12258 test. Changes in overlay text and data links.
## Screenshots

Variant A:
screen shot 2016-03-21 at 14 22 09

Variant B:
screen shot 2016-03-21 at 14 21 50
## 

_Does this PR meet the [contributing guidelines](https://github.com/guardian/frontend/blob/issue_pr_templates/.github/CONTRIBUTING.md#submission)?_

PR #12290, opened 2016-03-18
Subscriber number page changes

Calanthe:
## What does this change?

This PR introduces changes on the subscriber number page:
- new header, 
- no footer and description text,
- changes in correct and incorrect statuses
- new valid subscriber number format
## Screenshots

screen shot 2016-03-18 at 15 22 16
## 

_Does this PR meet the [contributing guidelines](https://github.com/guardian/frontend/blob/issue_pr_templates/.github/CONTRIBUTING.md#submission)?_

PR #12383, opened 2016-03-30
Decommission adblock-blocking test

jbreckmckye:
The 'Adblock Blocker' A/B test is about to expire tomorrow, and we're satisfied we have all the data we need. This pull request removes the adblocker bounce overlay added by @Calanthe in PRs #12258, #12281, #12288, #12290, #12299, #12302, #12304 and #12312.

It specifically removes:
- the A/B test switch, experiment, and AB script entry;
- the static page that allowed subscribers to opt-out of the test;
- the survey-simple module that's no longer used by anything else;
- the detect.js feature alias used only by this test;
- style code no longer called in any other templates;
- SVGs specific to the survey overlay

It also adds a module to common.js which cleans up extant localStorage values that may persist on users' machines. Like the cookie cleanup task, the idea is to add entries as storage values get purged, to keep the resource free of junk data.


uplne:
Nice work. I like the cleanup method :+1:

PR #12384, opened 2016-03-30
completely remove posters from fronts

jamesgorrie:
# Completely remove posters from fronts

Extension of #12375. For the same reasons
We now don't include them in the flash / fallback either, which strangely download even when they're invisible.
## Request for comment

@akash1810 @sndrs @paperboyo
## 


akash1810:
:+1:

PR #12386, opened 2016-03-30
Fix video loading

jamesgorrie:
We were accidentally setting canonicalUrl to '/null'.
This is a hotfix as tracking is down, I will look at a longer term fix later.


akash1810:
:+1:

PR #12389, opened 2016-03-31
Give paid containers access to surrounding section block

kelvin-chappell:
This is part of the commercial containers refactor.

It switches to the new container template at a higher level so that the section block that currently surrounds the container is included as part of the container.

/cc @regiskuckaertz


JonNorman:
Looks good! :+1:

PR #12391, opened 2016-03-31
Cache liveblog archive pages for longer

johnduffell:
Liveblogs have a cache time of 5 seconds as they update frequently.  This also applied to archive pages, which won't update as much.  Since frontend fetches the entire JSON from capi on every render, this was causing a high load for very long live blogs (e.g. the 17 page long one last week)

This PR implements a simple solution to reduce the load by increasing the cache time for archive pages to be the same as for an equivalent article (probably 300 seconds for a liveblog that's being updated regularly)

After that most of the load would come from the first page.  A separate piece of work is under way to retrieve only the last n blocks from capi - this will help with live updates and loading the first page.

I had to refactor things a bit because we were assuming one resource in capi mapped to one http response from frontend.  This meant the cache time was baked into the model we populate from capi data.  I have integrated the live blog page model more closely into the capi model and separated out the cache time a little more clearly.

@rich-nguyen @cb372


common/test/recorder/HttpRecorder.scala
```diff
@@ -19,8 +19,6 @@ trait HttpRecorder[A] extends ExecutionContexts {
   final def load(url: String, headers: Map[String, String] = Map.empty)(fetch: => Future[A]):Future[A] = {
 
     val fileName = name(url, headers)
-    println(url)
-    println(fileName)
```
johnduffell: @TBonnin :stuck_out_tongue_closed_eyes:

TBonnin: Damn!!!! Thanks for fixing

article/app/controllers/ArticleController.scala
```diff
@@ -192,18 +172,66 @@ object ArticleController extends Controller with RendersItemResponse with Loggin
...
+      maybeRequiredBlockId.map(blockId => block => blockId == block.id),
+      _.id
+    )
+    liveBlogPageModel match {
+      case Some(pageModel) =>
+
+        val cacheTime =
+          if (!pageModel.currentPage.isArchivePage && liveBlog.fields.isLive) liveBlog.metadata.cacheTime
+          else {
+            if (LongCacheSwitch.isSwitchedOn) {
```
johnduffell: not completely happy to replicate all this logic here as well

common/app/model/Formats.scala
```diff
@@ -75,7 +75,6 @@ object MetaDataFormat {
     contentType: String,
     hasHeader: Boolean,
     schemaType: Option[String],
-    cacheSeconds: Int,
```
johnduffell: not sure if it's worth storing the cache time in the pressed json, unless I missed something?

johnduffell:
I've had another go @rich-nguyen - see what you think?

johnduffell:
conflicts already! :cry: come on @rich-nguyen :stuck_out_tongue:

article/app/controllers/ArticleController.scala
```diff
@@ -192,18 +172,66 @@ object ArticleController extends Controller with RendersItemResponse with Loggin
...
+        case (minute: Article, None) if minute.isUSMinute =>
+          Left(MinutePage(minute, RelatedContent(minute, response)))
+        case (liveBlog: Article, None) if liveBlog.isLiveBlog =>
+          createLiveBlogModel(liveBlog, response, None)
+        case (liveBlog: Article, Some(Some(requiredBlockId))) if liveBlog.isLiveBlog =>
+          createLiveBlogModel(liveBlog, response, Some(requiredBlockId))
+        case (article: Article, None) =>
+          Left(ArticlePage(article, RelatedContent(article, response)))
+        case _ =>
+          Right(NotFound)
```
cb372: Left for success, right for failure? Controversial!

article/app/controllers/ArticleController.scala
```diff
@@ -192,18 +172,66 @@ object ArticleController extends Controller with RendersItemResponse with Loggin
...
-      case article: Article => ArticlePage(article, RelatedContent(article, response))
+    val page = pageParam.map(ParseBlockId.apply)
+    val supportedContentResult = ModelOrResult(supportedContent, response)
+    val content: Either[PageWithStoryPackage, Result] = supportedContentResult.left.flatMap { content =>
+      (content, page) match {
+        case (minute: Article, None) if minute.isUSMinute =>
+          Left(MinutePage(minute, RelatedContent(minute, response)))
+        case (liveBlog: Article, None) if liveBlog.isLiveBlog =>
+          createLiveBlogModel(liveBlog, response, None)
+        case (liveBlog: Article, Some(Some(requiredBlockId))) if liveBlog.isLiveBlog =>
```
cb372: I'm sure there's a good reason for it, but the nested `Option` here smells strange to me

article/test/liveblog/PageParserTest.scala
```diff
@@ -1,18 +1,18 @@
 package liveblog
 
-import controllers.PageParser
+import controllers.ParseBlockId
 import org.scalatest.{FlatSpec, Matchers}
 
 class PageParserTest extends FlatSpec with Matchers {
```
cb372: Could rename this test to `ParseBlockIdTest`?

cb372:
Looks good to my unqualified eye

common/app/model/Formats.scala
```diff
@@ -110,7 +109,7 @@ object MetaDataFormat {
         part2.contentType,
         part2.hasHeader,
         part2.schemaType,
-        part2.cacheSeconds,
```
rich-nguyen: The only reason to keep the cache seconds in the pressed json is to keep serialisation simple. So we wouldn't need too many rules describing which fields get 'slimmed', which ones become default, etc. Here, the deserialisation is a bit more complicated.

article/app/controllers/ArticleController.scala
johnduffell: maybe you could class it as a success to come up with a Result without having to do any more processing :stuck_out_tongue_closed_eyes:

article/app/controllers/ArticleController.scala
johnduffell: the outer one is whether the user specified a block id, and the inner one was whether the one they specified was actually valid.  If they specified one which is invalid, it's a 404.  Maybe I could do some matching onto case classes or something clever, or at least add a comment though.

article/test/liveblog/PageParserTest.scala
johnduffell: good point

rich-nguyen:
to recap:
- the `LiveBlogPageModel` could be simpler without the typeclass, given it always takes `BodyBlock`
- the format should match the original model as much as possible, including `cacheTime`
- feel free to name things more sensibly now, eg `LiveBlogPage` -> `LiveBlog`

Great :+1:

PR #12392, opened 2016-03-31
Change header bidding price buckets

jbreckmckye:
Prices between $10 and $20 should be aliased to $1 increments.


uplne:
Beautiful :+1:

jbreckmckye:
:+1:

PR #12393, opened 2016-03-31
Pass AB test data to ophan

dominickendrick:
# Pass AB test data to ophan
## What does this change?

This moves the AB test data tracking in ophan from the register component loading module to the ab test module.

This will fire the AB test tracking data earlier in the page load, so should ensure it fires the tracking event early on **every** page view.
We currently depended on the `register` module loading, which only passes this data to ophan after 5 seconds. So moving it earlier should reduce the number of tracking events lost due to bouncing.

Ophan
## @philwills @rich-nguyen 


philwills:
Is this running anywhere so that we can see the data flowing through Ophan? Happy to help show how this can be done.

OliverJAsh:
> which only passes this data to ophan after 5 seconds

Out of interest, how come it waits 5 seconds?

Does it matter if we record `abTestRegister` before `register`?

desbo:
@OliverJAsh `register` appears to be used to record the timing/registration of other js modules, so the delay is probably so it fires at a point when any modules should've comfortably finished loading. 

There's no dependency between `abTestRegister` and `register`, it was just that the A/B tests happened to be submitted along with the `register` event. Decoupling the A/B test tracking from the register module means it can be sent much earlier and without a reliance on the `register` event firing.

static/src/javascripts/bootstraps/standard/main.js
```diff
@@ -118,6 +118,7 @@ define([
             if (guardian.isEnhanced) {
                 ab.segmentUser();
                 ab.run();
+                ab.trackEvent();
```
OliverJAsh: Minor point, but should/could this be called in `run` instead?

OliverJAsh:
:+1:

static/src/javascripts/bootstraps/standard/main.js
desbo: I don't think so - segmenting users, running tests, and firing the tracking event are distinct operations.
Would there be any benefit in conflating it with `run`?

OliverJAsh: Guess so, I wasn't sure if run meant “run the tests” or “initialise this
module”.

On Fri, 1 Apr 2016 at 16:45 Sam Desborough <notifications@github.com> wrote:

> In static/src/javascripts/bootstraps/standard/main.js
> <https://github.com/guardian/frontend/pull/12393#discussion_r58224743>:
>
> > @@ -118,6 +118,7 @@ define([
> >              if (guardian.isEnhanced) {
> >                  ab.segmentUser();
> >                  ab.run();
> > +                ab.trackEvent();
>
> I don't think so - segmenting users, running tests, and firing the
> tracking event are distinct operations.
> Would there be any benefit in conflating it with run?
>
> —
> You are receiving this because you were mentioned.
>
>
> <https://github.com/guardian/frontend/pull/12393/files/1775a7dbaf7c49a18f4caece9adf8e9861501a23#r58224743>
>

-- 

in England Number 908396

jamesgorrie: Why not `init` then?

desbo: `run` means "run the tests" ! There's no `init` function in the A/B module at the moment, but maybe it'd be better to move these three calls into one, rather than having them inside the bootstrap module.

dominickendrick:
@philwills I've now run through this locally with @tomseddon

PR #12395, opened 2016-03-31
make special report highlight yellow

sndrs:
## What does this change?

screen shot 2016-03-31 at 16 08 07

becomes

screen shot 2016-03-31 at 16 07 52

contrast ratio of 4:1, AA 

cc @rich-nguyen

### DIFF
--- a/static/src/stylesheets/_vars.scss
+++ b/static/src/stylesheets/_vars.scss
@@ -120,6 +120,8 @@ $frontend-palette: (
 
     paidfor-background:     guss-colour(news-support-3, $pasteup-palette),
 
+    special-report-accent:  #fce800,
+
     soulmates-colour:       #332e2e,
     soulmates-accent-colour: #484848,
     soulmates-border-colour: #e6e1c3,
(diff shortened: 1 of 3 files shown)


Pearson82:
cool, looks more legible than the blue @sndrs

desbo:
:+1:

paperboyo:
Not really related but we have another double WCAG AA & AAA fail for news-over-review:

wcag-fail

static/src/stylesheets/_vars.scss
```diff
@@ -120,6 +120,8 @@ $frontend-palette: (
 
     paidfor-background:     guss-colour(news-support-3, $pasteup-palette),
 
+    special-report-accent:  #fce800,
```
jbreckmckye: Is this an 'official' Guardian colour? It isn't on http://guardian.github.io/pasteup-palette/demo/ ...

sndrs: it isn't no, it's a special one for special reports

jbreckmckye: Do these follow some other palette, then? Or do we have free reign to just choose whatever our accessibility tool recommends?

PR #12396, opened 2016-03-31
Add dynamic container content to paid container model

kelvin-chappell:
This is part of the commercial containers refactor.

It's an initial attempt at modelling the dynamic container content.

/cc @regiskuckaertz


JonNorman:
Looks good to me :+1:

PR #12397, opened 2016-03-31
Upgrade fapi to 1.3.0

rich-nguyen:
Use the latest version of facia scala client


JustinPinner:
_Looks_ simple enough :+1:

PR #12399, opened 2016-04-01
Replace outbrain-on-amp switch with generic outbrain switch

JonNorman:
This PR removes the AMP-specific switch for outbrain (which is about to expire).

Now that AMP has been stable for a while, we can use the old outbrain switch to control outbrain on both web and AMP.

/@stephanfowler


stephanfowler:
:+1:

PR #12400, opened 2016-04-01
Breaking-News: Notification message should be used as headline

TBonnin:
## What does this change?

Notification message should be used as the breaking-news headline, instead of the notification title
Title is always 'The Guardian', which is only used by the Android app

No

@johnduffell


johnduffell:
fair enough :+1:

PR #12403, opened 2016-04-01
Admin javascript app bootstrap

rich-nguyen:
Currently, the admin javascript does not initialise. This was because of various problems with the curl config and guardian config objects expected on window. Rather than spoofing these objects, this pr uses the standard js templates to inject prod-like configuration.


OliverJAsh:
:+1:

PR #12408, opened 2016-04-01
Fix teamcity link bugs (number instead of build id and external host)

TBonnin:
## What does this change?

Fix the link to the teamcity build:
- Display build number instead of build ID
- Use external host to build link (instead of internal host)

No
## Request for comment

@OliverJAsh


OliverJAsh:
:+1:

PR #12409, opened 2016-04-01
Add test implementation of Google Analytics

gklopper:
# My Awesome Pull Request
## What does this change?

This allows us to do a limited evaluation rollout of Google Analytics. We want a chance to play with some real data and have a look at things like any effect on page performance.

We are considering moving to GA. We are going to let people play with this data and see if it improves their use of analytics.

The audience team is in on this too.

Not directly. Apps is aware we are doing this, and the Android app already has GA
## Screenshots

image
## 

*Does this PR meet the [contributing guidelines]

### DIFF
--- a/common/app/templates/inlineJS/blocking/googleAnalytics.scala.js
+++ b/common/app/templates/inlineJS/blocking/googleAnalytics.scala.js
@@ -10,7 +10,7 @@ try {
     We are more interested in "users" than in raw page views.
     *@
     if (window.localStorage) {
-        var gaStorageKey = 'gu_ga_participation';
+        var gaStorageKey = 'gu.ga.participation';
         gaParticipation = window.localStorage.getItem(gaStorageKey) || gaParticipation;
         window.localStorage.setItem(gaStorageKey, gaParticipation);
     }
--- a/common/app/templates/inlineJS/blocking/googleAnalytics.scala.js
+++ b/common/app/templates/inlineJS/blocking/googleAnalytics.scala.js
@@ -1,0 +1,35 @@
+@(page: model.Page)
+@import implicits.Strings.string2encodings
+
+var gaRandomParticipation = Math.floor(Math.random() * 100) < @GoogleAnalyticsAccount.samplePercent;
+var gaParticipation = gaRandomParticipation ? 'in' : 'out';
+
+try {
+    @*
+    Once a person is in the test group we want them to stay in.
+    We are more interested in "users" than in raw page views.
+    *@
+    if (window.localStorage) {
+        var gaStorageKey = 'gu_ga_participation';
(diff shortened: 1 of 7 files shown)


common/app/views/support/GoogleAnalyticsAccount.scala
```diff
@@ -0,0 +1,28 @@
+package views.support
+
+import conf.Configuration.environment
+import model.Page
+import conf.switches.Switches.GoogleAnalyticsSwitch
+
+object GoogleAnalyticsAccount {
+
+  private val prod = "UA-33592456-1"
```
rich-nguyen: Are these ok public?

gklopper: Yes, they appear in the source of the page

OliverJAsh:
:+1: !

common/app/templates/inlineJS/blocking/googleAnalytics.scala.js
```diff
@@ -0,0 +1,35 @@
...
+    We are more interested in "users" than in raw page views.
+    *@
+    if (window.localStorage) {
+        var gaStorageKey = 'gu.ga.participation';
+        gaParticipation = window.localStorage.getItem(gaStorageKey) || gaParticipation;
+        window.localStorage.setItem(gaStorageKey, gaParticipation);
+    }
+} catch (ex) {
+    @* if for some reason we cannot read/write to localStorage don't be in the test *@
+    gaParticipation = 'out';
```
OliverJAsh: Should we log the error? `console.error(error.stack);`

sndrs:
yay! :+1: 

@dominickendrick reminds me we should remove the footer analytics switch when this merges - this shouldn't be blocking

sndrs:
plus we can finally enable this: http://caniuse.com/#stats_import :gem:

OliverJAsh:
> plus we can finally enable this: http://caniuse.com/#stats_import

That would be super useful!

common/app/templates/inlineJS/blocking/googleAnalytics.scala.js
gklopper: Mozilla warns me against using `console.error` https://developer.mozilla.org/en-US/docs/Web/API/Console/error

sndrs: it's not strictly an error, it's more of a safe way to feature detect – does it need an error message?

OliverJAsh: I was thinking more from the point of debuggability. Imagine this code isn't working but you have no error to say why.

> Mozilla warns me against using console.error

We use it along with other `console` methods frequently, but we tend to guard the statement with `if (console && console.log) …`. In practice the only JS runtime that doesn't have it is old IE.

PR #12411, opened 2016-04-04
Extend ABNextInSeries

OliverJAsh:
Not sure what we're doing with this yet, but it's turned off in PROD for now.


jamespamplin:
:+1: looks good to me, although not sure why the tests are failing

Issue #10434, opened 2015-09-08
crosswords: older crosswords contain false special instruction with print link


ScottPainterGNM:
http://www.theguardian.com/crosswords/quick/12001

![image](https://cloud.githubusercontent.com/assets/11922253/9740502/f9b08b0e-564c-11e5-86e7-689517e028ff.png)

Not sure how many this covers but from Quick 13000 onwards that I have looked at it doesn't appear

paperboyo:
Very loosely related (similar cleanup job): https://github.com/guardian/frontend/issues/8129

ScottPainterGNM:
Fixed by #10442

ScottPainterGNM:
https://groups.google.com/a/guardian.co.uk/forum/#!topic/crosswords.beta/WSl1F0XMbns - this is still getting reported as an error on older crosswords

JonNorman:
Do these "special instructions" only appear on older crosswords? I think the fix is quite simple and am happy to put it in, but it would be good to know what else might be affected first...

PR #12412, opened 2016-04-04
Fix for older crosswords with printable instructions not being linked properly (#10434)

JonNorman:
### What does this change?

Fixes issue #10434 wherein the "special instructions" are being HTML escaped on older crosswords.
### Does this affect other platforms - Amp, Apps, etc?

No
### Screenshots

**before**:
image

**after**:
image

-- /@ScottPainterGNM


JonNorman:
The steps to print older crosswords are different to the current crosswords; instead of clicking a link to a new print-friendly page, it serves an image of the crossword that is print friendly. This PR just fixes the link, it doesn't attempt to replicate the same behaviour as for current crosswords:

jbreckmckye:
:+1:

PR #12414, opened 2016-04-04
Dont play expired video

jamesgorrie:
# Add Expired video message if this is the case

We now show a expired message if the video has expired.

The best solution here would be to have the CAPI response know if a video is expired to avoid the XHR call, and this will be coming when we introduce the video atom, this is a stopgap for now.

I have made sure no Ophan calls are broken.

Hopefully we don't get charged for the use of expired video.

No - but they might have to do a similar thing.
## Screenshots

vide-removed
## Request for comment

@akash1810 @fredex42
## 


akash1810:
slightly confused on the difference between this and https://github.com/guardian/frontend/pull/12373

jamesgorrie:
@akash1810 There is none - we pulled it out as we were paranoid, superstitious (and it was Friday).

akash1810:
ah ok. :+1:

PR #12417, opened 2016-04-04
Fix element--inline images in The Minute at non-mobile breakpoints

desbo:
There was a bug in the cleaner for The Minute that removed the container
for inline images. This container was later added as an empty element
after the image, and its padding was rendered as empty space.

This restores the image container and updates the styles as required.

Before: 
screen shot 2016-04-04 at 14 44 37

After:
screen shot 2016-04-04 at 14 46 39

(full-height images still work on mobile, which is what the original update to the cleaner was for):
screen shot 2016-04-04 at 14 47 28

### DIFF
--- a/static/src/stylesheets/module/content/_article-minute.scss
+++ b/static/src/stylesheets/module/content/_article-minute.scss
@@ -571,7 +571,7 @@ $minute-layout-margin-nudge: gs-span(1);
                     width: 100%;
                 }
 
-                .element--inline__outer {
+                .element--inline__image-wrapper {
                     height: 100vh;
                 }
 
--- a/static/src/stylesheets/module/content/_article-minute.scss
+++ b/static/src/stylesheets/module/content/_article-minute.scss
@@ -571,8 +571,12 @@ $minute-layout-margin-nudge: gs-span(1);
                     width: 100%;
                 }
 
-                .gu-image {
+                .element--inline__outer {
                     height: 100vh;
+                }
+
+                .gu-image {
+                    height: 100%;
                     width: 100vw;
                     object-fit: cover;
                 }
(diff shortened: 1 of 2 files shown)


jamespamplin:
:+1: looks good to me

static/src/stylesheets/module/content/_article-minute.scss
```diff
@@ -571,8 +571,12 @@ $minute-layout-margin-nudge: gs-span(1);
                     width: 100%;
                 }
 
-                .gu-image {
+                .element--inline__outer {
```
jamesgorrie: Do we add `element`to state?
Seems like `element--inline-outer` would make more BEM sense?

desbo: hmm, the parent element has the `element--inline` class and `outer` is a descendent. 
that made sense to me, but i'm not a BEM master :(

jamesgorrie: It's normally `block__element--modifier`.

To stop creating layout decisions (i.e. `element--inline__outer` should always be in `element--inline`) adding complexity, I would say just create the style and name it, aka, `inline-outer`, which doesn't specify usage.

That's my personal understanding of BEM though, some people like to use it more complexly.

@sndrs opinion?

sndrs: `.inline-element`, `.inline-element--outer`?

desbo: This is the (simplified) markup: 

```
<figure class="element element--inline">
  <div class="element--inline__outer"></div>
</figure>
```

`outer` depends on `element--inline` and shouldn't be used anywhere else. Since it's a child element rather than a modifier, shouldn't it be separated by underscores?

sndrs: yeah ok, that works for me i think. it's unorthodox, but it seems to follow the principle?

sndrs: insofar as `B__E--M` is probably a simplistic description of the principle

desbo: yeah, i think it's a bit odd because `outer` is a bad name for that thing, and i'm treating `element--inline` as the block. it's the same as the `.person--female__hand` example here, http://csswizardry.com/2013/01/mindbemding-getting-your-head-round-bem-syntax/ but you don't see it much.
i'll rename it from `outer` so it's less confusing. thanks!

PR #12419, opened 2016-04-04
Scale back survey overlays on mobile breakpoints

jbreckmckye:
The survey overlays on mobile are full-page in scale:

image

We don't like this - we think it's too aggressive and disruptive.

We want something more like this instead:

image

We want the overlay to look more dialog-like, with negative space around it with the page's underlying content peeking through.

The dialog is vertically centred and takes whatever height it needs:

_iPhone 5 portrait_:
image

_iPhone 4 landscape_:
image

_iPhone 6 landscape_:
image

The desktop view remains slightly different - it's positioned a little higher on the page:

image


jamesgorrie:
Much nicer :+1:

PR #12420, opened 2016-04-04
Make DEV JS compile errors consistent with PROD.

gklopper:
## What does this change?

We cannot have templates 503 at runtime in PROD with no chance of discovering the problem in pre-production.

You now get the same errors in DEV as in PROD (and like PROD you can use the switch to move between modes)

We no longer get "random" JS compile errors in PROD.

Development environments
## 


sndrs:
merging #12426 should deal with these conflicts

sndrs:
:+1:

PR #12422, opened 2016-04-05
Expose test results in Teamcity

jamespamplin:
## What does this change?
- Log `sbt test` output in teamcity with `sbt-teamcity-logger` plugin from Jetbrains
- Log JS test output in teamcity with `karma-teamcity-reporter`
- See [Build 7835](http://teamcity.gu-web.net:8111/viewLog.html?buildTypeId=dotcom_pullrequests&buildId=7835) for output.
- Should make it faster to determine what has broken a build
- Exposes number of tests run per build - useful if number changes significantly between builds, which may mean that something is wrong.
- Build only
## Screenshots

dotcom-teamcity-test-results-1
dotcom-teamcity-test-results-2

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -12,7 +12,8 @@ resolvers ++= Seq(
   Resolver.sonatypeRepo("releases"),
   Resolver.typesafeRepo("releases"),
   Resolver.url("guardian sbt-plugins", new URL("https://dl.bintray.com/guardian/sbt-plugins/"))(Resolver.ivyStylePatterns),
-  Resolver.url("sbt sbt-plugins", new URL("https://dl.bintray.com/sbt/sbt-plugin-releases/"))(Resolver.ivyStylePatterns)
+  Resolver.url("sbt sbt-plugins", new URL("https://dl.bintray.com/sbt/sbt-plugin-releases/"))(Resolver.ivyStylePatterns),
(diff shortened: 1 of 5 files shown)


jamespamplin:
Screenshots and link to build added

sndrs:
:+1: nice

project/plugins.sbt
```diff
@@ -24,3 +25,5 @@ addSbtPlugin("com.timushev.sbt" % "sbt-updates" % "0.1.10")
 addSbtPlugin("com.gu" % "sbt-riffraff-artifact" % "0.8.1")
 
 addSbtPlugin("com.typesafe.sbt" % "sbt-git" % "0.8.5")
+
+addSbtPlugin("org.jetbrains.teamcity.plugins" % "sbt-teamcity-logger" % "0.3.0")
```
rich-nguyen: I think that this was the plugin caused the following problem: https://github.com/JetBrains/sbt-tc-logger/issues/9

desbo:
cool!

project/plugins.sbt
jamespamplin: Added b200be6c269c401a8994b47c5315a0d3c91d496e to workaround this issue. Commented on the JetBrains ticket to see if we can fix it properly in the future. This will do for now I think though.

TBonnin:
😍  ✖️ 💯

PR #12423, opened 2016-04-05
Enable page re-pressing in a post-R2 era

JustinPinner:
## What does this change?

Now that R2 is properly gone, we can't use it as a page pressing source so we don't need to keep the remnants of it in our code. However, there is a requirement to be able to re-press from the preserved sources to fix any issues that may be spotted in the near future.

This PR removes the ComboCleaner object and relocates its (now future-free) code into HtmlCleaner. It also removes the special headers and corresponding switch that the presser used to reach around the CDN and into R2.

Lastly, the "From preserved src" checkbox now defaults to a checked state on the admin page.

We can re-press preserved R2 pages if we need to. Hopefully we won't need to.

No.
## Request for comment

@johnduffell (and/or anyone else)
## 


admin/app/jobs/R2PagePressJob.scala
```diff
@@ -85,11 +83,11 @@ object R2PagePressJob extends ExecutionContexts with Logging {
   private def parseAndClean(originalDocSource: String): Future[String] = {
     val cleaners = Seq(PollsHtmlCleaner, InteractiveHtmlCleaner, SimpleHtmlCleaner)
     val archiveDocument = Jsoup.parse(originalDocSource)
-    cleaners.filter(_.canClean(archiveDocument))
+    val doc = cleaners.filter(_.canClean(archiveDocument))
```
johnduffell: may be worth adding a type on doc, as otherwise it will always compile even if we make it stop returning a Document by mistake, as toString always exists on anything (which is the main problem with toString, but we can't easily get around that)

JustinPinner: Good point. Will do.

johnduffell:
sure, although might be worth completely removing the non preserved source options as it's not actually possible any more! :+1: good work though thanks!

JustinPinner:
@johnduffell it is possible if we run a local instance :scream:  Not that we _want_ to but it's a useful thing to keep for a while.

PR #12424, opened 2016-04-05
remove test code for cholita autoplay

jamesgorrie:
Some ab code wasn't removed so is erroring.
There should probably be better checking in the ab code, but I will look at that once this is out.


rich-nguyen:
:+1: I just saw this in dev too. Ideally, the ab test shouldn't be so embedded in our main entry point code. What do you think?

jamesgorrie:
:+1: via @akash1810

jamesgorrie:
@rich-nguyen you shouldn't  publically expose a method that can break so brittally.

Potentially looking at whether we need to do this, if not, remove it. Probably also cater for the `null` case.

Where art though sweet `Option`s.

PR #12427, opened 2016-04-05
The Canadian Issue

markjamesbutler:
# The Canadian issue
## What does this change?

Fixes: Canada appears to be missing from the phone number drop-down on (https://profile.theguardian.com/account/edit)

UserHelp don't receive reports of errors that Canada is missing from the drop down.  It additionally, fixes a bug involving selecting the wrong country when initially loading page.  For example, if the user has set a +1 country code, on page load the US Virgin Islands would be selected as default.  This is due to this country being the last selected item in the options, as there are multiple options with the value 1.

As the USA and Canada share country codes we cannot easily determine which country the user is located base on their phone.  Unfortunately the API implementation of this issue does not save the country with the telephone number thus allowing us to select the correct country from the drop down.

Please note a longer term fix for this issue is to store the country against the phone number so this can be shown in the dialog.

No
## Screenshots

screen shot 2016-04-05 at 16 53 19
## 


identity/app/model/CountryCodes.scala
```diff
@@ -2,242 +2,14 @@ package model
 
 object CountryCodes {
```
kelvin-chappell: Perhaps this should be called something like `PhoneNumbers` to make it clear what the codes are?

markjamesbutler: Good suggestion. :)

emma-p:
Should there be a label explaining what the dropdown is for? I don't think it is obvious if you remove the country name

markjamesbutler:
@emma-p How about this as a pragmatic fix.  We should iterate on this in later releases as the UX around phone number entry is quite complicated.

screen shot 2016-04-12 at 18 23 26

emma-p:
:+1: That looks good

identity/app/model/PhoneNumbers.scala
```diff
@@ -0,0 +1,15 @@
+package model
+
+object PhoneNumbers {
+
+  lazy val countryCodes = List(1, 7, 20, 27, 30, 31, 32, 33, 34, 36, 39, 40, 41, 43, 44, 45, 46, 47, 48, 49, 51, 52, 53,
```
jamespamplin: This doesn't look very maintainable, what happens if you've left out another country or its code? Would a Map of Country to Country Code be more maintainable?

identity/app/model/PhoneNumbers.scala
```diff
@@ -1,15 +1,23 @@
...
+
+  /**
+    * Unfortunately the country code map is not accessible in PhoneNumberUtil. This code uses reflection to obtain access
+    * then transform it to the correct structure.
+    */
+  lazy val countryCodes = {
+    val m = ru.runtimeMirror(phoneNumberUtil.getClass.getClassLoader)
+    val countryCallingCodeToRegionCodeMapSymb = ru.typeOf[PhoneNumberUtil].decl(ru.TermName("countryCallingCodeToRegionCodeMap")).asTerm
+    val im = m.reflect(phoneNumberUtil)
+    val countryCallingCodeToRegionCodeMap = im.reflectField(countryCallingCodeToRegionCodeMapSymb).get.asInstanceOf[java.util.HashMap[java.lang.Integer,java.util.List[java.lang.String]]]
```
jamespamplin: Crazy code, but still better than hardcoding ourselves.

jamespamplin:
:+1: good stuff

PR #12428, opened 2016-04-05
Remove LiveStats module

desbo:
It's no longer in use and the same data is available in Ophan.
@rich-nguyen is about to do the equivalent of this in the backend.


rich-nguyen:
:+1:

PR #12429, opened 2016-04-05
Aws cloudwatch audit

rich-nguyen:
This change removes a large number of cloudwatch metrics that are topping up our monthly AWS bill.

Another thing it does is restrict the Cloudwatch helper. This means that anonymous `cloudwatch.put` usage is no longer possible, which is annoying, but very soon I will add something that puts an expiry date on the cloudwatch metric, forcing us to delete metrics after they are used.

Big savings:
- removed cloudwatch alarms in non-prod environments.
- removed some system metrics, which count per-instance.
- removed fastly metrics, which were counting every service, in every region.
- there are almost no references to Memcached, so I can now downgrade the cluster.


admin/app/services/Fastly.scala
```diff
@@ -22,14 +22,15 @@ case class FastlyStatistic(service: String, region: String, timestamp: Long, nam
...
-  private val regions = List("usa", "europe", "ausnz", "apac")
+  private val regions = List("usa", "europe", "ausnz")
 
   def apply(): Future[List[FastlyStatistic]] = {
 
     val futureResponses: Future[List[String]] = Future.sequence{
       regions map { region =>
-        val request = WS.url(s"https://api.fastly.com/stats?by=minute&from=45+minutes+ago&to=15+minutes+ago&region=${region}"
-        ).withHeaders("Fastly-Key" -> AdminConfiguration.fastly.key).withRequestTimeout(20000)
+        val request = WS.url(s"https://api.fastly.com/stats/service/${fastly.serviceId}?by=minute&from=45+minutes+ago&to=15+minutes+ago&region=$region")
```
rich-nguyen: We were cloudwatch uploading every fastly service, and every json key...

common/app/conf/switches/FeatureSwitches.scala
```diff
@@ -205,15 +205,6 @@ trait FeatureSwitches {
     exposeClientSide = false
   )
 
-  val QuizScoresService = Switch(
```
rich-nguyen: I removed as many memcached references as possible, and i'll downsize the cluster

admin/app/conf/AdminConfiguration.scala
```diff
@@ -29,6 +29,7 @@ object AdminConfiguration {
 
   object fastly {
     lazy val key = configuration.getStringProperty("fastly.key").getOrElse(throw new RuntimeException("Fastly key not configured"))
+    lazy val serviceId = configuration.getStringProperty("fastly.serviceId").getOrElse(throw new RuntimeException("Fastly key not configured"))
```
johnduffell: should be fastly.serviceId not nconfigured.  and I'm surprised we don't have a helper method to do this automatically correctly?

admin/app/jobs/AnalyticsSanityCheckJob.scala
```diff
@@ -1,58 +1,64 @@
...
-  def run() {
-
-    val rawPageViews = CloudWatch.rawPageViews.map(
-      _.getDatapoints.headOption.map(_.getSum.doubleValue()).getOrElse(0.0)
-    )
-
-    val analyticsPageViews = CloudWatch.analyticsPageViews.map(
-      _.getDatapoints.headOption.map(_.getSum.doubleValue()).getOrElse(0.0)
-    )
+  private val rawPageViews = new AtomicLong(0L)
```
johnduffell: why not a normal long?  I'm sure that's atomic for sets and gets anyway

admin/app/jobs/FastlyCloudwatchLoadJob.scala
```diff
@@ -10,21 +13,50 @@ object FastlyCloudwatchLoadJob extends ExecutionContexts with Logging {
   // Samples in CloudWatch are additive so we want to limit duplicate reporting.
   // We do not want to corrupt the past either, so set a default value (the most
   // recent 15 minutes of results are unstable).
+
+  // This key is (service, name, region)
   val latestTimestampsSent = mutable.Map[(String, String, String), Long]().
     withDefaultValue(DateTime.now().minusMinutes(15).getMillis())
 
+  // Be very explicit about which metrics we want. It is not necessary to cloudwatch everything.
+  val allFastlyMetrics: List[SamplerMetric] = List(
```
jamespamplin: You could simplify this to make it a bit more DRY, but not a massive issue, nor a blocker.

admin/app/jobs/FastlyCloudwatchLoadJob.scala
```diff
@@ -10,21 +13,50 @@ object FastlyCloudwatchLoadJob extends ExecutionContexts with Logging {
   // Samples in CloudWatch are additive so we want to limit duplicate reporting.
   // We do not want to corrupt the past either, so set a default value (the most
   // recent 15 minutes of results are unstable).
+
+  // This key is (service, name, region)
```
johnduffell: +1 for documentation, but if possible a case class would document it in the code instead of a comment

admin/app/services/Fastly.scala
johnduffell: :sob:

admin/app/jobs/AnalyticsSanityCheckJob.scala
rich-nguyen: Apparently not! And primitive Double is not atomic either.

admin/app/jobs/FastlyCloudwatchLoadJob.scala
rich-nguyen: Agreed. I went for readability, so that the next person who has to audit our cloudwatch can grep all the metrics.

admin/app/conf/AdminConfiguration.scala
rich-nguyen: well spotted

common/app/common/metrics.scala
```diff
@@ -39,384 +32,151 @@ object SystemMetrics extends implicits.Numbers {
     }
   }
 
-
   lazy val garbageCollectors: Seq[GcRateMetric] = ManagementFactory.getGarbageCollectorMXBeans.map(new GcRateMetric(_))
 
-
   // divide by 1048576 to convert bytes to MB
```
kelvin-chappell: Perhaps make this explicit with a def called something like `asMb`?

archive/app/controllers/ArchiveController.scala
```diff
@@ -144,7 +144,7 @@ object ArchiveController extends Controller with Logging with ExecutionContexts
     request.headers.get("User-Agent").getOrElse("no user agent") match {
       case GoogleBot(_) =>
         log.warn(s"404,${RequestLog(request)}")
-        Googlebot404Count.increment()
+        app.Global.Googlebot404Count.increment()
```
jamespamplin: Rather than accessing `Googlebot404Count` via `Global`, you could move it into a companion object `ArchiveMetrics`.

common/app/common/metrics.scala
```diff
@@ -39,384 +32,151 @@ object SystemMetrics extends implicits.Numbers {
...
-object S3Metrics {
-  object S3ClientExceptionsMetric extends CountMetric(
-    "s3-client-exceptions",
-    "Number of times the AWS S3 client has thrown an Exception"
-  )
-
-  object S3AuthorizationError extends CountMetric(
-    "s3-authorization-errors",
-    "Number of requests to S3 by facia that have resulted in a 403"
+  val BuildNumberMetric = GaugeMetric(
```
kelvin-chappell: How is this metric used out of interest?

common/app/common/metrics.scala
```diff
@@ -39,384 +32,151 @@ object SystemMetrics extends implicits.Numbers {
...
+    description = "Build number",
+    get = () => buildNumber,
+    metricUnit = StandardUnit.None
   )
 }
 
 object ContentApiMetrics {
-  object ElasticHttpTimingMetric extends FrontendTimingMetric(
-    "elastic-content-api-calls",
+  val ElasticHttpTimingMetric = TimingMetric(
```
kelvin-chappell: Is there a difference between `ElasticHttp` and `ContentApi`?

kelvin-chappell:
I like it.  Very bold! :+1:

common/app/common/metrics.scala
rich-nguyen: Sounds good

archive/app/controllers/ArchiveController.scala
rich-nguyen: good idea

common/app/common/metrics.scala
rich-nguyen: I've used it to work out when a build was live, using average. Bit like riffraff, except you can overlay the graph on another metric, and correlate step changes with deployments. Build number usually looks like an ascending staircase (the exception being reverts)

common/app/common/metrics.scala
rich-nguyen: None! I think we should get rid of the term `Elastic` from our codebase now.

johnduffell:
:moneybag: :dollar:  :+1:

common/app/common/metrics.scala
kelvin-chappell: Makes sense.

PR #12430, opened 2016-04-05
Tidy JS tests

sndrs:
## What does this change?

cleans up un-caught promises in the tests, and tightens a couple up

I couldn't get the tests in #12339 to pass until updating the config test to use `Injector`. doing that caused a small string of other, unrelated tests to fail until these non-fail errors were all handled. 
## Request for comment

@regiskuckaertz making the liveblog test async revealed a bug it, so this actually [disables it](https://github.com/guardian/frontend/compare/js-tests-tidy?expand=1#diff-a612948a62cc990b622968b9ededa738R56) (I think the selector is causing issues).

cc @uplne 
## 

:100:


desbo:
looks good to me, would be good to get that broken test fixed and reinstated before merging though

sndrs:
agree, but that test has always been broken, just it never got picked up

regiskuckaertz:
:+1: I can have a look at that test later

PR #12432, opened 2016-04-06
Fix block level links in notifications

janua:
This fixes the links in notifications that link to live blog block level.

The anchor link should be at the end.

@NathanielBennett


rich-nguyen:
:+1:

PR #12433, opened 2016-04-06
Adjust spacing for lists in standfirsts

sammorrisdesign-zz:
## What does this change?

Makes the standfirsts with lists in them look designed

The value is that the site looks like theguardian.com

I hope amp hasn't done this otherwise @zeftilldeath will be put in front of Design Court
## Screenshots

Before:
screen shot 2016-04-06 at 11 10 16

After:
screen shot 2016-04-06 at 11 10 24


rich-nguyen:
:+1:

PR #12436, opened 2016-04-06
Tidy application.conf

rich-nguyen:
Remove memcached conf from servers which don't use memcached. Removed Image entirely.


jamespamplin:
:+1: good stuff

PR #12440, opened 2016-04-06
Refactor paid containers

regiskuckaertz:
This is a major rewrite of paid content containers, complete with new markup and new styles, that is intended to help us retake control of our own code (_yeah, let's retake control_), reduce the time it takes to solve bugs and reduce the overall number of bugs in the future.

Overall reduction in maintenance costs and easier newcomer induction.

No
## Screenshots

Load http://localhost:9000/commercial-containers using the facia app with the CODE stage
## Request for comment

@kelvin-chappell @jbreckmckye @uplne @Calanthe @JonNorman


static/src/stylesheets/module/commercial/_adverts-capi.scss
```diff
@@ -20,28 +17,50 @@
...
 
-    * + .advert {
-        border-left-color: guss-colour(paid-article-subheader-bg, $frontend-palette);
+    .adverts__row > * + *::before,
+    .adverts__row--wrap::before,
+    .adverts__row--wrap::after {
+        background: guss-colour(paid-article-subheader-bg, $frontend-palette);
+    }
+
+    .adverts__more > summary {
```
jbreckmckye: Should these styles know this much about our HTML structure?

static/src/stylesheets/module/commercial/_adverts.scss
```diff
@@ -165,42 +250,39 @@
...
+            flex-basis: calc(33.33% - #{$gs-gutter});
+        }
     }
 }
 
-.adverts__row--1x1x1 {
-    > * {
-        flex-basis: calc(33.33% - #{$gs-gutter});
+@include mq(tablet) {
+    .adverts__row--2x1x1 > :nth-child(1),
```
jbreckmckye: I like this!

static/src/stylesheets/module/commercial/_adverts.scss
```diff
@@ -241,6 +335,11 @@
             display: none;
         }
     }
+
+    &:not([open]) .inline-minus,
```
jbreckmckye: What if you had `when-open` and `when-closed` classes, instead?

static/src/stylesheets/module/commercial/_adverts-capi.scss
regiskuckaertz: I don't know. I've always resisted the urge to add class names to everything, especially if the underlying markup is fixed. But this is not a great argument.

static/src/stylesheets/module/commercial/_adverts.scss
regiskuckaertz: Then I need javascript to toggle the classes, but this moots the intent of the `details`.

kelvin-chappell:
Great work! :+1:
It may be possible to simplify the capi cards model but I can do that separately after merge.

regiskuckaertz:
@kelvin-chappell Oh yeah, that redundancy is exactly what we're trying to get rid of, so you're most welcome to do that :-)

I would rather the `data-sponsorhip` et al attributes be on the `ASIDE.advert`, because the surrounding `div` is just there for layout, but this can also be improved in the future.

Let's goooooo!

static/src/stylesheets/module/commercial/_advert.scss
```diff
@@ -39,3 +42,64 @@
...
+            margin-bottom: 0;
+            margin-left: $gs-gutter / 2;
+        }
+
+        .advert__image {
+            width: 100%;
+        }
+    }
+
+    .advert--large--inverse {
```
uplne: This is smelly. Can't it be `advert--large` and `advert--inverse` and you can mix and match in the template?

When I see modification of the modification, usually there is always a different way how to approach it. And it is same across the whole file.

regiskuckaertz: good idea! I'll put that in the next update. Thanks dude

regiskuckaertz: @uplne giphy 1

PR #12441, opened 2016-04-06
Paid content card's height in the most popular fix

Calanthe:
## What does this change?

This PR fixes the bug when the paid content card is being cut by the most popular container. In order to show the whole paid content card inside the most popular container, that container should have a height at least the same as the paid content card. But there is no point of setting this height, when the height of the paid content card is smaller than the most popular container.
## Screenshots

Before:
screen shot 2016-04-06 at 15 21 39

After:
screen shot 2016-04-06 at 15 24 20
## 


regiskuckaertz:
It is sad we have to go through that kind of rigmaroles, but refactoring the most pop container is a tough one.

:+1:

PR #12442, opened 2016-04-06
Use cards to determine container branding

kelvin-chappell:
Use cards to determine the branding of sponsored and foundation-funded containers.
This uses the same logic as the paid containers.

/cc @JonNorman


JonNorman:
:+1:

PR #12447, opened 2016-04-06
Add completion event to A/B tests

desbo:
@dominickendrick and I have been looking adding a "complete" state to the A/B data we send to Ophan so it's more simple to find out conversion/success rates for our tests.

Related PR: https://github.com/guardian/frontend/pull/12393

This updates the structure of the event we send to Ophan and adds a completion step to the _next in series_ test. You can add  a `success` function to your variant to specify what needs to happen for the test to be completed. That function is passed a callback that will fire the completion data to Ophan.

Later on we'll add dashboards to display the success rate of tests.

### DIFF
--- a/static/src/javascripts/projects/common/modules/experiments/ab.js
+++ b/static/src/javascripts/projects/common/modules/experiments/ab.js
@@ -21,6 +21,7 @@ define([
     'lodash/collections/map',
     'lodash/collections/find',
     'lodash/objects/pick',
+    'lodash/utilities/noop',
     'common/utils/chain'
 ], function (
     reportError,
@@ -45,6 +46,7 @@ define([
     map,
     find,
     pick,
+    noop,
     chain
 ) {
 
@@ -164,7 +166,10 @@ define([
                 variantId = participations[test.id].variant;
             var variant = getVariant(test, variantId);
             if (variant) {
+                var success = variant.success || noop;
+
                 variant.test();
+                success(ab.trackEvent);
             } else if (variantId === 'notintest' && test.notInTest) {
                 test.notInTest();
             }
--- a/static/src/javascripts/projects/common/modules/experiments/ab.js
+++ b/static/src/javascripts/projects/common/modules/experiments/ab.js
@@ -178,7 +178,7 @@ define([
                 var active = tests[0];
                 var server = tests[1];
 
-                if (isParticipating(active) && testCanBeRun(active)) {
+                if (active && isParticipating(active) && testCanBeRun(active)) {
                     var variant = getTestVariantId(active.id);
 
                     if (variant && variant !== 'notintest') {
--- a/static/src/javascripts/projects/common/modules/experiments/ab.js
+++ b/static/src/javascripts/projects/common/modules/experiments/ab.js
@@ -14,11 +14,13 @@ define([
     'common/modules/experiments/tests/membership',
     'common/modules/experiments/tests/loyal-adblocking-survey',
     'lodash/arrays/flatten',
+    'lodash/arrays/zip',
     'lodash/collections/forEach',
     'lodash/objects/keys',
     'lodash/collections/some',
     'lodash/collections/filter',
     'lodash/collections/map',
+    'lodash/collections/reduce',
     'lodash/collections/find',
     'lodash/objects/pick',
     'lodash/utilities/noop',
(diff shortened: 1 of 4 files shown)


static/src/javascripts/projects/common/modules/experiments/ab.js
```diff
@@ -157,6 +161,60 @@ define([
...
+            // Encountering an error should invalidate the logging process.
+            abLogObject = {};
+            reportError(error, false);
+        }
+
+        return abLogObject;
+    }
+
+    function createAbObject(id, variantName, complete) {
+        var abData = {};
```
regiskuckaertz: Couldn't you get rid of this object and just return its content? This would spare you the trouble of merging all the time.

static/src/javascripts/projects/common/modules/experiments/ab.js
```diff
@@ -157,6 +161,60 @@ define([
...
+    function recordOphanAbEvent(data) {
+        require(['ophan/ng'], function (ophan) {
+            ophan.record({
+                abTestRegister: data
+            });
+        });
+    }
+
+    function recordTestComplete(test, variantId) {
+        var testName = test.id;
```
regiskuckaertz: Is this `var` really necessary?

static/src/javascripts/projects/common/modules/experiments/ab.js
desbo: good call. I'm gonna do a general refactor here.

static/src/javascripts/projects/common/modules/experiments/ab.js
desbo: nope 8)

static/src/javascripts/projects/common/modules/experiments/ab.js
```diff
@@ -157,6 +165,61 @@ define([
...
+    function abData(variantName, complete) {
+        return {
+            'variantName': variantName,
+            'complete': complete
+        };
+    }
+
+    function getAbLoggableObject() {
+        try {
+            return reduce(zip(getActiveTests(), getServerSideTests()), function(log, tests) {
```
regiskuckaertz: Silly question: is there a 1:1 mapping between those sets?

regiskuckaertz:
:+1: Looks good! You might want to remove unused dependencies thought.

static/src/javascripts/projects/common/modules/experiments/ab.js
desbo: nah, generally server side tests will be a lot smaller but `zip` will continue for the length of the longest array, adding `undefined` if there's nothing in the other one. that's why i had to add the check for `server` on line 189. 

thinking about it i should add an equivalent check for `active` to cover the unlikely event that there are more server than client-side tests. not a silly question at all!

desbo:
they're all still in use :deciduous_tree:

static/src/javascripts/projects/common/modules/experiments/tests/next-in-series.js
```diff
@@ -64,6 +66,10 @@ define([
                         var el = $.create(render(series));
                         el.insertAfter($articleBody);
                     }
+                },
+
+                success: function(complete) {
+                    bean.on($('.next-in-series-test__remind-me-link')[0], 'click', complete);
```
OliverJAsh: Will the tracking for this complete before the navigation completes? Not sure it matters, but curious!

desbo: I think it's very unlikely but not guaranteed not to happen. I'll change it to work by checking if you're on the survey URL from the appropriate referrer instead

static/src/javascripts/projects/common/modules/experiments/tests/next-in-series.js
```diff
@@ -64,6 +68,18 @@ define([
...
+                success: function(complete) {
+                    var onSurvey = window.location.href.match(path.replace(/\/$/, '')).length > 0;
+
+                    var referredFromSeries = some(pluck(allSeries, 'pageId'), function (id) {
+                        return window.document.referrer.match(id).length > 0;
+                    });
+
+                    if (onSurvey && referredFromSeries) {
+                        complete();
+                    }
```
OliverJAsh: :+1: Nice

PR #12448, opened 2016-04-06
Removing force schema switch for amp

NataliaLKB:
Removing a switch for amp which forces the schema. The schema is still being forced, but I am just removing the switch.

All AMP pages to pass Google's validation, and show in the AMP carousel, need to have a schema of `NewsArticle`.

We put the switch in as a reminder to check with Google about this and to see whether they would start accepted other schemas like review. They are not planning to support other schemas anytime soon, and will let us know when they do. Therefore, I am removing the switch.

@TBonnin


jamesgorrie:
:+1:

PR #12451, opened 2016-04-07
Split switches for fixed and dynamic containers

kelvin-chappell:
So we can switch them on and off independently of each other.

This also extends all the refactor switches for another week.


JonNorman:
:+1:

regiskuckaertz:
Yeah! :+1:

PR #12452, opened 2016-04-07
Refactored jobs component

regiskuckaertz:
Was already using the new container but had the wrong margins.

/cc @kelvin-chappell @uplne


kelvin-chappell:
:+1:

PR #12455, opened 2016-04-07
Gimbap simple

uplne:
# New (simple) version of Gimbap commercial component
## What does this change?

There are some changes to original gimbap component to make it more modular. 
- splitted `gimbap-wrap`, `gimbap-logo` and `gimbap-branding`
- simple `effect` version to test

This is for testing. New original gimbap component (in multiple versions) is being tested for a little over a week and it is already showing uplift in CTR. CTR is the main metric.

Nope
## Screenshots

screen shot 2016-04-07 at 10 42 49
screen shot 2016-04-07 at 10 40 31
screen shot 2016-04-07 at 10 40 20

With effect:
gimbap-effects

Tested on all browsers (including IE9+) and devices.

@regiskuckaertz


regiskuckaertz:
Instead of having two wrapper templates `gimbap` and `gimbap-simple` (what does 'simple' mean in this case?), I would rather you create a template for the wrapper and one for the gimbap component, so that you can get rid of all the redundancy. WDYT?

uplne:
There is quite a lot of difference (in the templates) for these two modules even if they are similar.

I can create two templates which have some redundancy or I can create one template for each `wrap` and `gimbap` and have some logic (arguably a lot of logic) inside of them. I've chosen the former.

+ keep in mind that they are all just tests. We don't know which one will stay so I would be careful with too much abstraction.

Simple means that it is much simpler (inside the gimbap, not necessarily inside the wrapper) than the original gimbap component.

uplne:
@regiskuckaertz I've created a gimbap-simple-blob which is the sub-template for simple gimbap. I've left original untouched as there is more logic in the current template. So in theory it doesn't add any advantage but because I started working on `gimbap richmedia` version it should.

static/src/stylesheets/module/commercial/gimbap/_gimbap-logo.scss
```diff
@@ -0,0 +1,101 @@
...
+        .logo--part1,
+        .logo--part2 {
+            fill: #ffffff;
+        }    
+    }
+}
+
+.gimbap-logo--tone-soulmates {
+    svg {
+        width: 85px; /* "Make your own magic" */
```
regiskuckaertz: giphy 4

static/src/stylesheets/module/commercial/gimbap/_gimbap-wrap.scss
```diff
@@ -0,0 +1,271 @@
...
+    @include mq(desktop) {
+        > .gimbap:nth-child(n+4) {
+            display: none;
+        }
+    }
+}
+
+.gimbap-wrap__title {
+    font: 900 1.375rem/2.25rem 'Guardian Egyptian Web', 'Guardian Text Egyptian Web', Georgia, serif;
+    color: guss-colour(guardian-brand-dark, $pasteup-palette);;
```
regiskuckaertz: beware the two `;;`

static/src/stylesheets/module/commercial/gimbap/_gimbap.scss
```diff
@@ -61,6 +45,87 @@ $gimbap-width-quarter: calc(25% - 20px);
...
+}
+
+.gimbap--reversed {
+    color: #ffffff;
+
+    .gimbap__cta {
+        color: rgba(0, 0, 0, .65);
+    }
+
+    &.gimbap--tone-members { background: $gimbap-members-reversed; }
```
regiskuckaertz: Shouldn't these go where you define the tones?

regiskuckaertz:
:+1: aside minor comments

static/src/stylesheets/module/commercial/gimbap/_gimbap.scss
uplne: In theory yes.

Lets think about the story in which you have a task to change the color tone of `jobs` `reversed` modification. Would you first go to `.gimbap--reversed` and try to change it there or would you go to `.gimbap--tone-jobs` and try to change it there?

PR #12461, opened 2016-04-07
New header: add opt-in controller

OliverJAsh:
This adds or removes the `new_header_opt_in` cookie as used by https://github.com/guardian/fastly-edge-cache/pull/604

### DIFF
--- a/applications/app/controllers/NewHeaderCookieController.scala
+++ b/applications/app/controllers/NewHeaderCookieController.scala
@@ -1,0 +1,23 @@
+package controllers
+
+import model.Cached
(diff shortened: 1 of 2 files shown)


johnduffell:
description - you mean "new_header_opt_in" (underscores)

johnduffell:
:+1:

desbo:
:cool: 
maybe we should have a generic opt-in/out controller with a cookie name parameter

johnduffell:
@desbo yes yes! I was thinking of saying the same - we are just splattering our top level name space with these optin urls.  Even /opt/in/https would be better!  And could call the cookie the same as the last part of the url for consistency!

PR #12462, opened 2016-04-07
Make commercial cards consistent

kelvin-chappell:
So that cards in Capi merchandising components are the same as cards in paid containers.

/cc @regiskuckaertz


commercial/app/views/contentapi/itemV2.scala.html
```diff
@@ -30,7 +31,7 @@
 }{
 
     <div class="adverts__row">
-        @itemLargeCard(item, clickMacro)
+        @itemLargeCard(card, omnitureId)
```
regiskuckaertz: so no more `clickMacro`?

regiskuckaertz:
Thanks, that's awesome! :+1:

commercial/app/views/contentapi/itemV2.scala.html
kelvin-chappell: It gets passed in as part of the `targetUrl` of the card content.
That's built here: https://github.com/guardian/frontend/blob/kc-capi-card/common/app/common/commercial/CardContent.scala#L79

regiskuckaertz: :heart_eyes:

PR #12465, opened 2016-04-07
Add generic opt-in/out controller

desbo:
This way we don't have to keep copy/pasting these controllers. I think changing the routes might be an idea as well, but will need a bit of comms.

@johnduffell @OliverJAsh


OliverJAsh:
Perfect :+1: 

Perhaps we could change the URLs to the new format, but keep the old ones around for the transition…

applications/conf/routes
```diff
@@ -48,11 +48,11 @@ GET        /preferences
...
 
-GET        /https/optin                                                         controllers.HTTPSCookieController.optIn()
-GET        /https/optout                                                        controllers.HTTPSCookieController.optOut()
+GET        /https/optin                                                         controllers.OptInController.HTTPS.optIn()
+GET        /https/optout                                                        controllers.OptInController.HTTPS.optOut()
 
-GET        /new-header/optin                                                    controllers.NewHeaderCookieController.optIn()
-GET        /new-header/optout                                                   controllers.NewHeaderCookieController.optOut()
+GET        /new-header/optin                                                    controllers.OptInController.Header.optIn()
+GET        /new-header/optout                                                   controllers.OptInController.Header.optOut()
```
johnduffell: if you restructure the urls I'll be even more happy! ;)

johnduffell: however the existing https one would have to stay as it's too well known

johnduffell:
:+1:

applications/conf/routes
desbo: i think olly's idea of using them both for a while will work

PR #12466, opened 2016-04-07
Refactor cAPI component

regiskuckaertz:
Behold, the updated capi components join the list of those using the refactored container.

@uplne **with** the new `advert--inverse`, Monsieur.

/cc @kelvin-chappell @uplne


kelvin-chappell:
Looking good :+1:

static/src/stylesheets/module/commercial/_advert.scss
```diff
@@ -69,7 +69,7 @@
         }
     }
 
-    .advert--large--inverse {
+    .advert--inverse {
```
uplne: Well done but you have `.advert--large--thumbnail` below on the line 85 and it is the same thing. I should mentioned it properly in the previous comment.

regiskuckaertz: No worries. I'll fix that too.

static/src/stylesheets/module/commercial/_adverts-capi.scss
```diff
@@ -18,6 +18,7 @@
     }
 
     .adverts__row > * + *::before,
+    &.adverts--legacy-single .adverts__row > .advert::after,
```
uplne: My head is spinning. Can this be moved to the template so you will only add class, something like `adverts__row--legacy` and it will have only after pseudo element in it?

uplne:
Other than comments cool stuff 👍

PR #12467, opened 2016-04-07
Remove pasteup-buttons, establish new pasteup styles folder

sammorrisdesign-zz:
## What does this change?

This PR removes the dependancy with the bower component `pasteup-buttons`. It replaces it with a completely uncomponented version of `buttons`. The reason is that we're going to make it easier, or at least make people more aware, to make changes to our shared pasteup resources.

In the short term, the pasteup-buttons bower module that is being used by other teams could diverge from what we're using in frontend.


sndrs:
:+1:

PR #12469, opened 2016-04-07
remove cacheSeconds from pressed json

johnduffell:
Because https://github.com/guardian/frontend/pull/12391 was a breaking change to facia json, I had to add the new parameter and keep the old one.  Once I have deployed, I will merge this which will get rid of the redundant parameter!  Fun :cold_sweat: 
@janua @rich-nguyen


janua:
:+1:

PR #12470, opened 2016-04-07
Upgrade frontend to capi 8.2

rich-nguyen:
Upgrades frontend to the latest capi and fapi libraries. `UseThrift` is set to false, initially.


johnduffell:
:+1: :confounded:

PR #12472, opened 2016-04-07
Introduce onboarding message for first-time commenters

nicl:
# My Awesome Pull Request
## What does this change?

Introduces an onboarding stage for first-time commenters. When pressing 'post' the user is shown a welcome message linking to the community standards and showing a preview of the comment.

The hope is that the onboarding will educate users about our community standards and reduce the number of abusive comments.

No.
## Screenshots

screen shot 2016-04-07 at 18 27 22
## Request for comment

I've gone over this with @OliverJAsh in the initial stages (all faults entirely my own yada yada) but I'm fumbling around here so all feedback welcome!

cc @guardian/discussion

### DIFF
--- a/common/app/views/fragments/commentBox.scala.html
+++ b/common/app/views/fragments/commentBox.scala.html
@@ -21,13 +21,24 @@
         </div>
     </div>
 
+    <div class="d-comment-box__onboarding">
+        <h3>Welcome, you’re about to make your first comment!</h3>
+        <p>Before you post, we’d like to thank you for joining the debate - we’re glad you’ve chosen to participate and we value your opinions and experiences.</p>
+        <p>Please keep your posts respectful and abide by the <a href="/community-standards">community guidelines</a> - and if you spot a comment you think doesn’t adhere to the guidelines, please use the ‘Report’ link next to it to let us know.</p>
+        <p>Please preview your comment below and click ‘post’ when you’re happy with it.</p>
+        <div class="d-comment-box__onboarding-preview-body"></div>
+        <button type="submit" class="u-button-reset button button--large button--primary submit-input--behaviour d-comment-box__onboarding-submit">Post your comment</button>
+        <button type="button" class="u-button-reset button button--large button--primary submit-input--behaviour d-comment-box__onboarding-cancel">Cancel</button>
+    </div>
+
     <div class="d-comment-box__content">
         <div class="d-comment-box__messages"></div>
-        <div class="d-comment-box__community-standards">Please keep comments respectful and abide by the <a href="/community-standards">community guidelines</a>.</div>
         <div class="d-discussion__error d-comment-box__premod">
             @fragments.inlineSvg("status-alert", "icon")
             <span class="d-discussion__error-text">Your comments are currently being pre-moderated (<a href="/community-faqs#311" target="_blank">why?</a>)</span>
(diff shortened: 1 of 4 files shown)


common/app/views/fragments/commentBox.scala.html
```diff
@@ -21,13 +21,24 @@
         </div>
     </div>
 
+    <div class="d-comment-box__onboarding">
+        <h3>Welcome, you’re about to make your first comment!</h3>
```
mchv: Very small suggestion.  We may add the user firstname to have a more personal message, like `Welcome John`.

nicl: That was on the original spec in fact! So definitely a good idea. I'll look to re-add it.

static/src/javascripts/projects/common/modules/discussion/comment-box.js
```diff
@@ -145,6 +146,25 @@ CommentBox.prototype.prerender = function() {
     }
 };
 
+CommentBox.prototype.showOnboarding = function(e) {
+    e.preventDefault();
+
+    // Check if new commenter as they may have already commented on this article
+    if (this.hasState('onboarding-visible') || !this.options.newCommenter) {
```
OliverJAsh: It looks like we only call this method iff they are a new commenter?

nicl: @OliverJAsh the imagined scenario is a user that comments once (for the first time) and then comments again without reloading the page.

nicl: To expand on this, once they have commented we set newCommenter to false (see ln155 just below) but we don't rebind the events. Maybe we should? I wasn't sure about rebinding outside of the ready function.

common/app/views/fragments/commentBox.scala.html
nicl: @mchv this is now done!

screen shot 2016-04-11 at 11 08 24

nicl:
@OliverJAsh thanks for looking over it. Do you have any more thoughts on this?

static/src/javascripts/projects/common/modules/discussion/comment-box.js
OliverJAsh: I see! You could unbind the events, but I think this looks fine.

static/src/javascripts/projects/common/modules/discussion/comment-box.js
```diff
@@ -379,13 +416,12 @@ CommentBox.prototype.verificationEmailFail = function() {
...
  * @param {Event=} e (optional)
  */
-CommentBox.prototype.previewComment = function(e) {
+CommentBox.prototype.previewComment = function(callback) {
     var self = this,
         comment = {
             body: this.getElem('body').value
         };
 
-    e.preventDefault();
```
OliverJAsh: How come this was removed?

OliverJAsh:
:+1: Well done for fighting the complexity here!

static/src/javascripts/projects/common/modules/discussion/comment-box.js
nicl: It didn't seem to do anything!

PR #12473, opened 2016-04-08
Commercial component - no ad call without isbn

Calanthe:
## What does this change?

This PR fixes the bug which in a situation when there is no isbn provided (for a commercial component) there was a redundant ad call made with a 404 status. Right now, when there is no isbn, there will be no ad call.
## Screenshots

Before:
screen shot 2016-04-08 at 10 29 47


jbreckmckye:
This is a very welcome change; I hate how many redundant errors clog up the console sometimes.

:+1:

PR #12474, opened 2016-04-08
Adblocking response survey - FF excluded and proper links set

Calanthe:
## What does this change?

This PR excludes FF from the test (because we can't detect adblock properly) and sets up proper survey links. The audience is still 0% because we need to test if a dummy ophanId value is passed to the survey servers correctly.


kelvin-chappell:
:+1:

PR #12478, opened 2016-04-08
Adblocking response AB test - correct ophan id value

Calanthe:
## What does this change?

This PR adds a correct ophan id value to the AB test. The audience is still 0% so we can test it.


kelvin-chappell:
:+1:

PR #12480, opened 2016-04-08
Refactor blended component

regiskuckaertz:
This one took a bit longer because it needed additional markup and styles.

I also kept the same layout for each component as inside their own wrapper, which is a departure from the existing blended component... those images are too small and thus distracting. Let's see how people inside the guardian react.

Before:
picture 1

After:
picture 1

/cc @uplne @kelvin-chappell


commercial/app/controllers/commercial/Multi.scala
```diff
@@ -29,73 +29,144 @@ object Multi
...
       val omnitureId = request.getParameter("omnitureId")
 
-    val eventualContent = componentsAndSpecificIds map {
-      case ("jobs", jobId) if jobId.nonEmpty =>
-        Future.successful {
-          JobsAgent.specificJobs(Seq(jobId)).headOption orElse {
-            JobsAgent.jobsTargetedAt(segment).headOption
-          } map {
-            views.html.jobs.jobFragment(_, clickMacro)
+      var eventualContent = if(conf.switches.Switches.v2BlendedTemplate.isSwitchedOn) {
```
kelvin-chappell: Small point but maybe import `conf.switches.Switches`

kelvin-chappell:
Looks good. :+1: 

When we remove the switch it may be worth seeing if the controller and presentation code can be separated more cleanly, so that the controller doesn't have to pass blocks of html to the template.

regiskuckaertz:
Oh yeah! I actually tried passing a `Seq[Any]` but this is not a good pattern and thought about creating a proxy class with four attributes, one for each type, only one of which would be `!= Nil` and use a `match` in the template; but I still don't know if this is the right way to do it.

PR #12481, opened 2016-04-08
Start fetching travel offer feeds again

kelvin-chappell:
This change is the first step in reinstating travel offer merch components.
It:
- separates fetching and parsing switches logically for all merchandising feeds
- separates fetching and parsing physically for the new travel feed
- removes a redundant property

Parsing is still to come in another PR
(in case I don't get it done this afternoon so someone else can take it on next week).

### DIFF
--- a/commercial/app/views/travel/travelProminent.scala.html
+++ b/commercial/app/views/travel/travelProminent.scala.html
@@ -1,20 +1,18 @@
 @(offers: Seq[model.commercial.travel.TravelOffer], omnitureId: Option[String], clickMacro: Option[String])(implicit request: RequestHeader)
 
-@import conf.Configuration
-
 @defining(("1_1", "2014-07-04")) { case (version, date) =>
     <div class="commercial commercial--travel commercial--tone-travel commercial--high commercial--prominent" data-link-name="commercial-high | travel | @omnitureId">
         <div class="commercial__inner">
             <div class="commercial__header">
                 <h3 class="commercial__title">
-                    <a href="@clickMacro@Configuration.commercial.travel_url/" data-link-name="merchandising-travel-v@{version}_@{date}-high-title">
+                    <a href="@{clickMacro}http://www.guardianholidayoffers.co.uk/" data-link-name="merchandising-travel-v@{version}_@{date}-high-title">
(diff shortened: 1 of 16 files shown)


commercial/app/views/travel/travelProminent.scala.html
```diff
@@ -1,20 +1,18 @@
...
 
-@import conf.Configuration
-
 @defining(("1_1", "2014-07-04")) { case (version, date) =>
     <div class="commercial commercial--travel commercial--tone-travel commercial--high commercial--prominent" data-link-name="commercial-high | travel | @omnitureId">
         <div class="commercial__inner">
             <div class="commercial__header">
                 <h3 class="commercial__title">
-                    <a href="@clickMacro@Configuration.commercial.travel_url/" data-link-name="merchandising-travel-v@{version}_@{date}-high-title">
+                    <a href="@{clickMacro}http://www.guardianholidayoffers.co.uk/" data-link-name="merchandising-travel-v@{version}_@{date}-high-title">
```
JonNorman: would it be worth defining this URL as a variable at the top?

JonNorman:
Looks good :+1: - might it be worth grouping these switches under a new group - e.g. `External Feeds` given that we currently have ~50 switches in the Commercial group on the switchboard and we will now have two switches per feed?

commercial/app/views/travel/travelProminent.scala.html
kelvin-chappell: This URL doesn't actually work.  I just stuck it there to get rid of the property.  The new feed has complete target URLs so this will go altogether in the next step.

kelvin-chappell:
I don't think we necessarily need to have separate switches for each feed, and the parsing switch should probably become optional.  But I agree if we did end up with two switches for each feed.

commercial/app/views/travel/travelProminent.scala.html
regiskuckaertz: @kelvin-chappell the correct URL should be https://holidays.theguardian.com/

kelvin-chappell: @regiskuckaertz Thanks.  That's in the feed so I'm going to replace the whole URL.

PR #12483, opened 2016-04-08
Adblocking response AB test - final changes

Calanthe:
## What does this change?

Final changes to the Adblocking response AB test: proper audience and a click event attached also to the survey link.

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/survey/survey-simple.js
+++ b/static/src/javascripts/projects/common/modules/commercial/survey/survey-simple.js
@@ -43,14 +43,17 @@ define([
                 $(document.body).append(this.bannerTmpl);
 
                 if (this.config.showCloseBtn) {
-                    bean.on(document, 'click', $('.js-survey-close'), function () {
-                        $('.js-survey-overlay').addClass('u-h');
-                        if (this.shouldClosePermanently) {
-                            this.closePermanently();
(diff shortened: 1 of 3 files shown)


static/src/javascripts/projects/common/modules/commercial/survey/survey-simple.js
```diff
@@ -43,17 +43,20 @@ define([
...
 
                 if (this.config.showCloseBtn) {
-                    bean.on(document, 'click', $('.js-survey-close'), function () {
-                        $('.js-survey-overlay').addClass('u-h');
-                        if (this.shouldClosePermanently) {
-                            this.closePermanently();
-                        }
-                    }.bind(this));
+                    bean.on(document, 'click', $('.js-survey-close'), this.handleClick.bind(this));
+                    bean.on(document, 'click', $('.js-survey-link__takepart'), this.handleClick.bind(this));
```
uplne: I don't think we need to use BEM syntax in js hooks

uplne:
Other than the small comment 👍

static/src/javascripts/projects/common/modules/commercial/survey/survey-simple.js
Calanthe: I remember about it in the future.

PR #12484, opened 2016-04-08
Add subscribeDate to the notifications subscription database

janua:
This adds the subscribe date as a unix timestamp to each new record in the `frontend-notifications` database to track who is subscribed to what.

@NathanielBennett


NathanielBennett:
:+1:

PR #12487, opened 2016-04-08
Parse new travel offers feed

kelvin-chappell:
This parses the new feed.  It works, and generates a travel component.

Work to do:
- doesn't match against Capi tags, only against countries
- I think sorting needs looking at
- prices need formatting
- images are of variable size and quality


commercial/app/model/commercial/travel/TravelOffer.scala
```diff
@@ -2,30 +2,70 @@ package model.commercial.travel
...
+      offerUrl = text("url"),
+      imageUrl = text("image"),
+      fromPrice = parseDouble(text("@fromprice")),
+      earliestDeparture = DateTime.parse(text("@earliestdeparture")),
+      keywordIdSuffixes = Nil,
+      countries = tagText("Country"),
+      category = tagText("Holiday Type").headOption,
+      tags = Nil,
+      duration = parseInt(text("@duration")),
+      position = -1
```
JonNorman: Is this ever set anywhere else? If not, do we need to keep this attribute?

JonNorman:
:+1: - the travel feed is currently turned off so the subsequent work can be done in a separate PR.

PR #12489, opened 2016-04-08
Enable Content API with thrift

rich-nguyen:
Adding a switch which, when enabled, makes requests to CAPI in thrift format. Dev tests show a small bandwidth saving, but the main test will be how much CPU saving we make from improved parsing performance.

### DIFF
--- a/common/app/common/package.scala
+++ b/common/app/common/package.scala
@@ -4,8 +4,8 @@ import java.util.concurrent.TimeoutException
 
 import akka.pattern.CircuitBreakerOpenException
 import com.gu.contentapi.client.GuardianContentApiThriftError
+import com.gu.contentapi.client.model.v1.ErrorResponse
 import conf.switches.Switch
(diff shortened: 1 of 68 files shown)


johnduffell:
:+1: the graphs will be interesting to see! :grinning:

common/app/common/package.scala
```diff
@@ -14,6 +14,10 @@ import play.twirl.api.Html
 
 object `package` extends implicits.Strings with implicits.Requests with play.api.mvc.Results {
 
+  def isCommercialExpiry(error: ErrorResponse): Boolean = {
```
cb372: As I mentioned in my email, this is quite likely to blow up when using Thrift :(

CAPI will return the error response in JSON and the Scala client will try to parse it as Thrift.

Let's chat IRL next week about how to work around this.

desbo:
cool!

PR #12491, opened 2016-04-08
Extend header bidding test over weekend

jbreckmckye:
So the US team can continue serving FastLane ads until we descaffold the test next week


JonNorman:
Well I suppose so :+1:

PR #12494, opened 2016-04-08
badges on amp should be amp-img

NataliaLKB:
## What does this change?

The US election badge wasn't validated on amp:
image

This was preventing some important AMP stories to not appear in the AMP carousel. This change fixes that :)

On the webmaster search console for amp it is hard to tell exactly how many stories are affected as it seems to be under the label `Unknown syntax error`. This label has 9,993 stories under it, but not all will be fixed by this change. It will help all of our articles on US elections pass amp validation though.

Just AMP :)


OliverJAsh:
:+1:

PR #12495, opened 2016-04-08
Series description amp

NataliaLKB:
## What does this change?

Adding Series container descriptions onto amp onward journeys

Just Amp
## Screenshots

BEFORE:
image

AFTER:
image

### DIFF
--- a/onward/app/controllers/SeriesController.scala
+++ b/onward/app/controllers/SeriesController.scala
@@ -67,11 +67,13 @@ object SeriesController extends Controller with Logging with Paging with Executi
   private def rendermf2Series(series: Series)(implicit request: RequestHeader) = {
     val displayName = Some(series.displayName)
     val seriesStories = series.trails.items take 4
-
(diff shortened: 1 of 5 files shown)


OliverJAsh:
:+1: Nice!

onward/app/controllers/SeriesController.scala
```diff
@@ -67,11 +67,13 @@ object SeriesController extends Controller with Logging with Paging with Executi
   private def rendermf2Series(series: Series)(implicit request: RequestHeader) = {
     val displayName = Some(series.displayName)
     val seriesStories = series.trails.items take 4
-
+    val description = series.tag.metadata.description.getOrElse("").replaceAll("<.*?>", "")
+
```
OliverJAsh: Is there no way to insert raw HTML into Amp?

NataliaLKB: No, it is rendered by mustache, so it only accepts strings unfortunately.

PR #12496, opened 2016-04-10
RTFM - Fix Google Analytics sample rate

gklopper:
# My Awesome Pull Request
## What does this change?

While reading the docs, I discovered Google analytics has a built in sampling percentage. We now use that instead of our (my) own convoluted random percentage code.

Also did some code cleanup while I was in there and moved analytics templates to their own package.

Also now force tracking over HTTPS.

We still stick inside the free tier of Google analytics

No
## Screenshots

N/A
## 


common/app/views/main.scala.html
```diff
@@ -80,7 +80,7 @@
 
         @fragments.inlineJSNonBlocking()
 
-        @fragments.analytics(page)
+        @fragments.analytics.analytics(page)
```
sndrs: how about this instead? seems a bit easier to work out what's where
```
@fragments.analytics.omniture(page)
@fragments.analytics.google(page)
```

gklopper: @sndrs I have done something like that. The other file did not only contain Omniture, so I broke it up a bit.

sndrs:
:+1: really hope this goes well!

PR #12498, opened 2016-04-11
Refactor blended component

regiskuckaertz:
Second try, this time with a fresh copy of master.

/cc @JonNorman @uplne

### DIFF
--- a/commercial/app/views/jobs/jobsBlended.scala.html
+++ b/commercial/app/views/jobs/jobsBlended.scala.html
@@ -1,0 +1,14 @@
+@(job: model.commercial.jobs.Job, clickMacro: Option[String])
+
+<div class="advert__blended-container advert__blended-container--jobs">
+    <div class="advert__blended-container__title">
+        <a href="@{clickMacro}http://jobs.theguardian.com"
+            data-link-name="merchandising-hybrid-jobs-v1_0_2014-08-14-visit-site">
+            @fragments.inlineSvg("logo-jobs-26", "commercial", Seq("logo--tone-jobs"))
+            <span class="u-h">Jobs</span>
+        </a>
+    </div>
+    <div class="advert__blended-container__body">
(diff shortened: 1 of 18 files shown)


commercial/app/views/jobs/jobsBlended.scala.html
```diff
@@ -0,0 +1,14 @@
+@(job: model.commercial.jobs.Job, clickMacro: Option[String])
+
+<div class="advert__blended-container advert__blended-container--jobs">
+    <div class="advert__blended-container__title">
```
uplne: This is difficult. In theory you should think about creating a separate object `blended-component` so you can avoid having element of the element.

In the logic of BEM there are two ways:
`advert__body--blended`

or 

`blended__body`

That's why we have `gimbap-wrap` component for wrapping and then `gimbap` for inside and `gimbap-logo` for handling logos etc. Each is a separate object with it's own elements which you can create modifications of.

When you start creating elements of the elements it get's complicated. At least that's my experience.

uplne: I would probably rather see multiple objects:
`advert`
`advert-blended`
`advert-something-else`

It is easier to reason about. So if I want to work on advert blended I will work in `advert-blended.scss` file. It is easy to follow.

static/src/stylesheets/module/commercial/_adverts-soulmates.scss
```diff
@@ -85,8 +85,42 @@
     }
 }
 
+.advert-blended--soulmates {
+    background: #f5f3e7;
```
uplne: Should this go from our palette? If not should we put it into variable in top of the file or to some sort of banding section?

uplne:
does it work on IE9+?

uplne:
Other than that I think it looks good 👍

PR #12501, opened 2016-04-11
Intersperse storyPackages stories

TBonnin:
## What does this change?

Reorder the stories within a story package container to mix stories from multiple
packages rather than showing stories from package 2 before stories from
package2

We suspect reordering the stories will bring more click on story packages' stories

No. @piuccio and I decided not to make any change for now about how we show story packages on amp
## Request for comment

@piuccio @OliverJAsh @dominickendrick

### DIFF
--- a/common/app/mvt/MultiVariateTesting.scala
+++ b/common/app/mvt/MultiVariateTesting.scala
@@ -68,8 +68,29 @@ object SeriesOnwardPosition extends TestDefinition(
     }
 }
 
+object ABIntersperseMultipleStoryPackagesStories extends TestDefinition(
+  List(Variant8), // 1% of our audience
+  "intersperse-multiple-story-packages-stories",
+  "To test if mixing storyPackages stories (when article has more than one storyPackage) results in more clicks",
+  new LocalDate(2016, 5, 3)
+)
+object ABIntersperseMultipleStoryPackagesStoriesControl extends TestDefinition(
+  List(Variant9), // 1% of our audience
+  "intersperse-multiple-story-packages-stories-control",
+  "Control for the intersperse-multiple-story-packages-stories A/B test",
+  new LocalDate(2016, 5, 3)
+)
+
 object ActiveTests extends Tests {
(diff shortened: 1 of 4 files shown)


common/app/mvt/MultiVariateTesting.scala
```diff
@@ -68,8 +68,29 @@ object SeriesOnwardPosition extends TestDefinition(
...
+  "To test if mixing storyPackages stories (when article has more than one storyPackage) results in more clicks",
+  new LocalDate(2016, 5, 3)
+)
+object IntersperseMultipleStoryPackagesStoriesControl extends TestDefinition(
+  List(Variant5, Variant6, Variant7, Variant8, Variant9), // 5% of our audience
+  "intersperse-multiple-story-packages-stories-control",
+  "Control group for intersperse-multiple-story-packages-stories A/B test",
+  new LocalDate(2016, 5, 3)
+)
+
```
OliverJAsh: Beware that other tests will be using these buckets/variants, and that may or may not corrupt your data.

OliverJAsh:
Looks good to me :+1: !

common/app/mvt/MultiVariateTesting.scala
TBonnin: Is there any way to prevent this?

TBonnin:
Scratch this PR. I cannot access the ABTest this way in RelatedContent. Working on a fix

common/app/mvt/MultiVariateTesting.scala
OliverJAsh: I guess you could pick variants which are not being used, but that might reduce your audience size. How big does the test need to be?

TBonnin: Ok. I am gonna use buckets which are not being used

TBonnin:
I fixed the issue. @piuccio and I decided to run the A/B test on article page only

piuccio:
looks great to me 👍

PR #12502, opened 2016-04-11
Update notification subscriptions

janua:
## What does this change?

Changes all subscription records to carry a `subscribedDate`.
Changes the `save` and `delete` endpoints so that they return and expose any errors that may have occurred during subscribing.
Also added more logging.

No

@NathanielBennett


NathanielBennett:
:+1: Cool. That's what I should have done. Check that service-worker.js don't need to changed?

janua:
@NathanielBennett Don't think that needs to change, as the `service-worker.js` doesn't call these endpoints.

PR #12490, opened 2016-04-08
Add Pasteup2016 release process to frontend

sndrs:
## What does this change?

@sammorrisdesign and I are in the process of retiring the current guss/pasteup repos and packages and centralising everything we use from them in the frontend codebase itself.

We are the main users of them, and the general source of truth for them. This will mean we won't have to jump through the release hoops to update the site (which some people have actually stopped doing anyway).

As part of this, we are adding [a new package on NPM](https://www.npmjs.com/package/pasteup) which will distribute Pasteup2016/next/2.0/🌟, and which will be published directly from frontend.

We will retire those repos/packages gracefully once this is all in place.
## Do I need to do anything about this?

No, not until you hear more. We'll send an email when it's in use, this is just preparing the ground.
## What is the value of this and can you measure success?

Hopefully, making edits and managing the components will become massively simpler, while 3rd parties are just as able to make use of them.
## Does this affect other platforms - Amp, Apps, etc?

Possibly, yes.
## Screenshots

screen shot 2016-04-08 at 17 18 21

screen recording 2016-04-08 at 04 47 pm

screen recording 2016-04-08 at 04 47 pm-2
## Request for comment

@sammorrisdesign 
## 

_Does this PR meet the [contributing guidelines](https://github.com/guardian/frontend/blob/issue_pr_templates/.github/CONTRIBUTING.md#submission)?_

PR #12503, opened 2016-04-11
Add buttons to new pasteup2016 library

sammorrisdesign-zz:
## What does this change?

This moves `pasteup-buttons` over to the new pasteup components library (of which you can read more about [here](https://github.com/guardian/frontend/pull/12490).
## Request for comment

@sndrs 
## 


sndrs:
do you reckon this should get some screen shots, to make sure buttons are actually being rendered properly?

sndrs:
think we should get some other eyes on this: @desbo @uplne @Calanthe @regiskuckaertz @OliverJAsh anyone fancy having a look?

uplne:
I think it's good. My only question is if we still need to keep the pasteup name (legacy reasons?). Can we just put it into `objects` folder because basically `button` is just an object with modifications. It is not component nor module.

uplne:
👍

sndrs:
'pasteup' will still be released to npm as a module, for 3rd parties to consume, this just moves it's home to front-end, not various other repos

PR #12505, opened 2016-04-11
Harmonise the disclosure widget in commercial containers with the ones in standard containers

regiskuckaertz:
The existing 'Show more' buttons carry a lot of baggage with them, which we don't want in our new adverts component: the 'Loading' state plus all the styling and markup that comes with `fc-container-*`. Hence I had to partially recreate its behaviour. I'm using the standard `DETAILS` element which is the defacto disclosure widget in HTML, with some polyfilling for browsers not supporting it.

Next I'm just overcoating the standard behaviour with some text replacement that tells the reader, once the widget is expanded there will be **less** content if she ever clicks the button again... Shocking! The result? If the user ever clicks the middle-to-right part of the button, because of the dramatic change in size, she will have to re-target it if she ever wants to click it again. 🎯 

/cc @uplne

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/adverts.js
+++ b/static/src/javascripts/projects/common/modules/commercial/adverts.js
@@ -44,7 +44,7 @@ define([
     function onOpenClick(event) {
         var summary = event.currentTarget;
         var details = summary.parentNode;
-        var label = summary.querySelector('.button__label');
+        var label = summary.querySelector('.js-button__label');
         if (details.hasAttribute('open')) {
             fastdom.write(function () {
                 label.textContent = 'More ' + summary.getAttribute('data-text');
--- a/static/src/javascripts/projects/common/modules/commercial/adverts.js
+++ b/static/src/javascripts/projects/common/modules/commercial/adverts.js
@@ -1,8 +1,8 @@
 define([
-    'bean',
     'qwery',
+    'bean',
     'fastdom'
-], function (bean, qwery, fastdom) {
+], function (qwery, bean, fastdom) {
     var shouldWePolyfill = !('open' in document.createElement('details'));
 
     return {
--- a/static/src/javascripts/projects/common/modules/commercial/adverts.js
+++ b/static/src/javascripts/projects/common/modules/commercial/adverts.js
@@ -1,0 +1,59 @@
+define([
+    'bean',
+    'qwery',
+    'fastdom'
+], function (bean, qwery, fastdom) {
+    var shouldWePolyfill = !('open' in document.createElement('details'));
+
+    return {
+        init: init
+    };
+
+    function init() {
+        if (shouldWePolyfill) {
+            bean.on(document, 'click', 'summary', onClick);
+            bean.on(document, 'keypress', 'summary', onKeyPress(onClick));
+        }
+
+        var showMores = qwery('.adverts__more > summary');
+        bean.on(document, 'click', showMores, onOpenClick);
+        bean.on(document, 'click', showMores, onKeyPress(onOpenClick));
+    }
+
+    function onClick(event) {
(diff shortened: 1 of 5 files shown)


static/src/javascripts/projects/common/modules/commercial/adverts.js
```diff
@@ -0,0 +1,59 @@
+define([
+    'bean',
+    'qwery',
+    'fastdom'
+], function (bean, qwery, fastdom) {
```
uplne: Let's follow the structure we have in all other files. Just for consistency.

static/src/javascripts/projects/common/modules/commercial/adverts.js
```diff
@@ -0,0 +1,59 @@
...
+    var shouldWePolyfill = !('open' in document.createElement('details'));
+
+    return {
+        init: init
+    };
+
+    function init() {
+        if (shouldWePolyfill) {
```
uplne: What about js hooks?

static/src/javascripts/projects/common/modules/commercial/adverts.js
```diff
@@ -0,0 +1,59 @@
...
+            if (event.keyCode === 0x20 || event.keyCode === 0x0D) {
+                handler(event);
+            }
+        };
+    }
+
+    function onOpenClick(event) {
+        var summary = event.currentTarget;
+        var details = summary.parentNode;
+        var label = summary.querySelector('.button__label');
```
uplne: Same here (js-hook). If for some reason I will change button__label to button__text it will stop working. Arguably it may never happen but...

static/src/javascripts/projects/common/modules/commercial/adverts.js
regiskuckaertz: yeah, what about js hooks?? I dont know what that is 🎅

static/src/javascripts/projects/common/modules/commercial/adverts.js
uplne: Should we change all show more buttons across the website?

static/src/javascripts/projects/common/modules/commercial/adverts.js
regiskuckaertz: oh I see. Not in this case, it is native behaviour that we want to simulate, so it is bound to all `SUMMARY` elements.

static/src/stylesheets/module/commercial/_adverts.scss
```diff
@@ -348,6 +348,10 @@
     &[open] .inline-plus {
         display: none;
     }
+
+    &:not([open]) > *:not(summary) {
```
uplne: I can't get used to this. For me the complexity of this selector is too much. Look at it as if statement in the JS. if adverts__more && !open && (child !== summary) you see the point.

static/src/javascripts/projects/common/modules/commercial/adverts.js
regiskuckaertz: (1) you're right, I'm changing it.
(2) if it was me, yes, not only because it is the right element, but also because this 'More/Less' behaviour should be removed... it is bad UX.

static/src/stylesheets/module/commercial/_adverts.scss
regiskuckaertz: This rule is only for browsers that don't support the element yet, we'll get rid of it when it's safe to. I agree but the `open` attribute is how the HTML element works (try in Chrome).

uplne: Maybe worth comment?

uplne:
👍

PR #12506, opened 2016-04-11
Use a case class to deserialise fastly api data

rich-nguyen:
This fixes the admin fastly CDN pages, which were not reading the Fastly API correctly.


JustinPinner:
Much tidier 👍 ✨

PR #12507, opened 2016-04-11
Liveblog jankiness ad-related bug fix

Calanthe:
## What does this change?

This PR fixes the bug which caused jankiness on liveblogs pages. Now only adslots located below the viewport should be loaded.

Smaller jankiness on liveblogs pages

cc @regiskuckaertz

### DIFF
--- a/static/src/javascripts/projects/common/modules/article/spacefinder.js
+++ b/static/src/javascripts/projects/common/modules/article/spacefinder.js
@@ -117,7 +117,7 @@ define([
         };
     }
 
-    function _enforceRules(slots, rules, bodyTop, bodyHeight) {
+    function _enforceRules(slots, rules, bodyHeight) {
         var filtered = Promise.resolve(slots);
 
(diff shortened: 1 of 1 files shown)


ScottPainterGNM:
Sweeeeeet!

static/src/javascripts/projects/common/modules/article/spacefinder.js
```diff
@@ -226,15 +227,15 @@ define([
...
+                    rect.top,
                     rect.height,
                     map(slots, _mapElementToDimensions)
                 ];
             });
         }
 
         function enforceRules(data) {
-            return _enforceRules(data[2], rules, data[0], data[1]);
+            return _enforceRules(data[2], rules, data[1]);
```
regiskuckaertz: Could you get rid of `data[0]` in `getMeasurements` if it isn't used anywhere?

regiskuckaertz:
👍  Let's see how this fares!

PR #12510, opened 2016-04-11
pasteup palette 2016

sndrs:
## What does this change?
- detaches all pasteup/guss components in frontend from bower (because of the scope of the affects of this)
- specifically removes `guss-colour` and `pasteup-palette`, replacing the output of all with `pasteup/palette`
- colours are now just `$vars`
## 


static/src/stylesheets/_vars.scss
```diff
@@ -23,8 +23,7 @@ $guss-rem-baseline: 16px;
 $inArticleInlineImgWidth: 114px;
 $headerHeight: 36px;
 
-@import 'components/guss-colours/_colours';
-@import 'components/pasteup-palette/src/_palette';
+@import 'pasteup/palette/src';
```
uplne: I would suggest to move the whole colors definitions into `branding.scss`?

sndrs: thats an interesting point - we may still jiggle things around once this is all localised to front-end. at the moment, this is mainly to duplicate the current behaviour, and tidy up any of the particularly major hoops we've been jumping through (colour, guss-colour etc)

uplne:
This thing killed my macbook. But from what I could see, in general, it looks good.

Just the naming really. Maybe we can call it `core` or something like that as these things are core settings and objects which are the same for the whole project. And it makes sense right away, at least for me, instead of pasteup which I can imagine has no meaning for new people. Especially if it is not "library" anymore but part of the project.

👍

uplne:
And I've just read that it will be still supported as library so my comment above is irrelevant.

sndrs:
i'm not sure, i think it's quite astute. will look at what we have after this part is done and see if it makes sense - `stylesheets/components` -> `npm i guardian-components` makes quite a lot of sense to me

PR #12511, opened 2016-04-11
Fix fastly dashboard in admin

rich-nguyen:
The admin dashboard was looking up the wrong dashboard name.


jamespamplin:
:+1: thanks for fixing this

PR #12512, opened 2016-04-11
Gimbap should follow grid on wide breakpoint

uplne:
I don't really have anything to add. It's now taking 100% and it should take only 940px;

@regiskuckaertz


regiskuckaertz:
rock n roll dude

giphy

PR #12513, opened 2016-04-11
Add link to Guardian holidays to more section (UK only)

lps88:
## What does this change?

Add new holidays link in the 'more' dropdown for UK header


JonNorman:
Looks good to me! :+1:

Issue #8792, opened 2015-04-07
Inconsistently with other types, pagination unavailable for type/gallery


paperboyo:
Hello,

If I want a list of all articles, ``/type/article`` (redirected to ``/articles``), gives me a paginated, chronological view.

Ditto for ``/type/picture`` (redirected to ``/pictures``).

But if I want a paginated, chronological view of all galleries, ``/type/gallery`` redirects to ``/inpictures`` which has all sorts of containers, but no paginated, chronological view ("latest" container doesn't count as there is no pagination to see older items). One way to get what I want, consistently with above type searches, is to manually add a pagination ``/inpictures?page=1``.

Also, inconsistently with, lat's say, ``/world``, ``/inpictures`` has no ``/all`` user-accessible link at the bottom of the page in Topics.

Regards
Mateusz

michaelwmcnamara:
Confirmed - type/gallery redirects to /inpictures which is a curated front.
Response code for /type/gallery is 302 which suggests this is a policy choice rather than just failing soft due to a missing page.
I will investigate and confirm if this configuration is deliberate.

paperboyo:
Thanks a lot, @michaelwmcnamara! Don't you think that, independently of your investigation, it would be a good idea to add "All today's stories" link (``http://www.theguardian.com/inpictures/all``) to the bottom of ``http://www.theguardian.com/inpictures`` just as on every other front (bar network front)?

Regards
Mateusz

johnduffell:
this hasn't been fixed so far so it's not likely to be fixed any time soon

PR #12514, opened 2016-04-11
Show /all links in Topics area on curated fronts

JustinPinner:
## What does this change?

On wider breakpoints, some fronts have been given treats with screen shot 2016-04-11 at 17 18 12 links to the /all tag page;
screen shot 2016-04-11 at 17 05 04

On narrower breakpoints these treats are hidden as their columns are removed, and so is the link to the /all page;
screen shot 2016-04-11 at 17 06 20

And in this state, the live site doesn't include any link to today's stories in their Topics area;
screen shot 2016-04-11 at 17 07 53

On some fronts, we do already see an "All today's stories" link in the Topics section near the end of the page, but this is due to the curated content they contain, and not because of the presence of the treats.

Now, if we don't have a pre-existing /all link in Topics and we do have a treat with an /all link, we'll show that in the Topics area;
screen shot 2016-04-11 at 17 14 07

Even at very narrow breakpoints;
screen shot 2016-04-11 at 17 24 21

It gives readers on mobile devices the opportunity to explore more of the content related to the front they are viewing. If it's particularly successful we may see an uplift in traffic to /all pages (and therefore content) from mobile devices, but at the very least we're making a navigation route available consistently across formats. 

No.
## Request for comment

cc @rich-nguyen @janua 
## 

### DIFF
--- a/common/app/model/PressedPage.scala
+++ b/common/app/model/PressedPage.scala
@@ -103,6 +103,9 @@ case class PressedPage (
       s"/${Paths.withoutEdition(id).getOrElse(id)}/all"
     }
   }
+
+  def alwaysAvailableAllPath = allPath.orElse(Some(s"/${Paths.withoutEdition(id).getOrElse(id)}/all"))
+
   val navSection: String = metadata.section
 
   val keywordIds: Seq[String] = frontKeywordIds(id)
--- a/common/app/model/PressedPage.scala
+++ b/common/app/model/PressedPage.scala
@@ -99,12 +99,17 @@ case class PressedPage (
       Paths.withoutEdition(id)
     ).flatten
 
-    tagAndSectionIds.find(validIds contains) map { id =>
+    tagAndSectionIds.find(validIds contains).map { id =>
       s"/${Paths.withoutEdition(id).getOrElse(id)}/all"
-    }
+    }.orElse(allPathFromTreats)
   }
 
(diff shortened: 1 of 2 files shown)


janua:
LGTM :+1:

paperboyo:
That’s :gem:, but it won’t fix the fact that the information infrastructure/browsing is neither predictable nor consistent enough. Keeping with the long-established Grauniad’s tradition, we are much better at it at the arcane URL level than at the UI one :sadpanda:. Finally fixes: https://github.com/guardian/frontend/issues/8792, though! BTW, how are they “today’s”?!

JustinPinner:
@paperboyo They're "today's" stories because the /all page generally starts displaying content from the current date. I'm not sure I can address your other comments right now. Is this a conversation that's been had already or should be had soon? What kind of forum do you need?

paperboyo:
@JustinPinner “today’s” may give a false hint that it will take me to a place where only a limited number of items live. While, in truth, this is the only place where _all_ items live.

This PR is great and brings more consistency! My comments were general (too general, as always!) and the problem is long-standing so there is no rush. There are tons of stuff I would like to see when it comes to better browsing/surfacing/discovery UX and I’m sure that you guys are discussing it all the time (although, I wouldn’t know.) Any forum is good, including, or even preferably, a :beers: one :wink:

PR #12515, opened 2016-04-11
Add CSP reporting

desbo:
This adds a `/csp-report` endpoint that logs violations (sent via https://github.com/guardian/fastly-edge-cache/pull/606) to DynamoDB.

Also make the dynamoDB reporter a bit more general, and update the CSS reporter accordingly.


diagnostics/conf/routes
```diff
@@ -11,7 +11,10 @@ GET        /counts.gif                      controllers.DiagnosticsController.an
 GET        /robots.txt                      controllers.Assets.at(path="/public", file="robots.txt")
 
 POST       /css                             controllers.DiagnosticsController.css
-OPTIONS    /css                             controllers.DiagnosticsController.cssOptions
+OPTIONS    /css                             controllers.DiagnosticsController.postOptions
+
+POST       /csp-report                      controllers.DiagnosticsController.csp
```
johnduffell: I'm pretty sure diagnostics runs on beacon.guim.co.uk in prod, but in your fastly PR it'll go to www.theguardian.com, so not sure if this will work

project/Frontend.scala
```diff
@@ -49,7 +49,8 @@ object Frontend extends Build with Prototypes {
       jerseyCore,
       jerseyClient,
       cssParser,
-      w3cSac
+      w3cSac,
+      scanamo
```
johnduffell: is this library needed everywhere or just in diagnostics?

diagnostics/app/model/diagnostics/csp/CSP.scala
```diff
@@ -0,0 +1,48 @@
...
+object CSPReport {
+  implicit val jsonReads: Reads[CSPReport] = (
+    (__ \ "blocked-uri").read[String] and
+      (__ \ "document-uri").read[String] and
+      (__ \ "effective-directive").read[String] and
+      (__ \ "original-policy").read[String] and
+      (__ \ "referrer").read[String] and
+      (__ \ "status-code").readNullable[Int] and
+      (__ \ "violated-directive").read[String]
+    )(withTimestamp _)
```
TBonnin: I think you can use `Reads.pure` so you don't need `withTimestamp`

```
implicit val jsonReads: Reads[CSPReport] = (
     (__ \ "blocked-uri").read[String] and
       (__ \ "document-uri").read[String] and
       (__ \ "effective-directive").read[String] and
       (__ \ "original-policy").read[String] and
       (__ \ "referrer").read[String] and
       (__ \ "status-code").readNullable[Int] and
       (__ \ "violated-directive").read[String] and
       Reads.pure(new Instant().getMillis)
     )(CSPReport.apply _)
```

diagnostics/app/model/diagnostics/DynamoDbReport.scala
```diff
@@ -0,0 +1,61 @@
...
+          log.error(s"Failed to log CSS report to dynamo DB", error)
+      }
+    }
+  }
+}
+
+object CSPDynamoDbReport extends DynamoDbReport[CSPReport] {
+  def report(report: CSPReport): Unit = {
+    Scanamo.put(client)("cspReport")(report)
+    ()
```
johnduffell: is this line returning unit needed?

TBonnin:
How many entries are we expected to record?

diagnostics/app/model/diagnostics/DynamoDbReport.scala
```diff
@@ -0,0 +1,61 @@
...
+        case error: Throwable =>
+          log.error(s"Failed to log CSS report to dynamo DB", error)
+      }
+    }
+  }
+}
+
+object CSPDynamoDbReport extends DynamoDbReport[CSPReport] {
+  def report(report: CSPReport): Unit = {
+    Scanamo.put(client)("cspReport")(report)
```
johnduffell: might be tricky to debug without doing something with the putItemResult that comes back, at least log something if it fails even if you don't want to return a 500 to the client for whatever reason

desbo: just added error logging :)

johnduffell: also this doesn't seem to return a Future, so it might be better to wrap in a blocking { ... } so that the scala runtime creates enough threads on the box

johnduffell: actually, you can get a future if you get @philwills  to ship a new version :grinning:  https://github.com/guardian/scanamo/pull/11

desbo:
@TBonnin It'll be `number of hits to https pages * number of mixed-mode errors on page`. A lot of pages probably won't throw any mixed mode errors and the worst cases will probably be around 5-10. The amount of https hits will vary depending on how widely we roll it out. 

Maybe I should restrict it to a certain number of entries within a timeframe since a lot of them will be duplicates. What do you think?

diagnostics/conf/routes
desbo: i think it's fine to do it as a cors request, can just use beacon as the report url

TBonnin:
@desbo 
I am curious to see what kind of write throughput would be needed and how much it would cost?
To limit the db throughput we can sample the reporting (either randomly or by limiting "reporting" users to a smaller audience).

Do we have an idea of the traffic we get on https pages? That would help making an estimate here

desbo:
It's gonna be huge actually because we've already rolled out HTTPS to a few sections :sweat_smile: 

I had a quick look and I think a sample rate of about 0.01% would be plenty, probably a few hundred submissions per 24 hours.

TBonnin:
😁 sounds good to me
I am not super familiar with dynamodb but I think the put requests fail when the provisioned throughput is reached. I don't know if scanamo has any retry logic though?

diagnostics/conf/routes
johnduffell: yes I think so

diagnostics/app/controllers/DiagnosticsController.scala
```diff
@@ -42,9 +42,12 @@ object DiagnosticsController extends Controller with Logging {
   }
 
   def csp = Action(jsonParser) { implicit request =>
-    if (conf.switches.Switches.CspReporting.isSwitchedOn) {
+    val sampleRate = 0.01
+
+    if (conf.switches.Switches.CspReporting.isSwitchedOn && random.nextDouble <= sampleRate) {
```
johnduffell: probably no need to use doubles here, I'd consinder use integers.  might be worth just using the time in milliseconds as it's quick to execute and doing `if (time % 1000 == 1)`

desbo:
@TBonnin don't think it does, and i think you're right about dynamo dropping requests. one of the main aims of this work is to see how widespread the problems are, so it's hard to say in advance if that'll happen or not!

TBonnin:
@desbo Indeed.
Are you planning to create a simple dashboard to see the results of these reporting?

common/app/conf/switches/MonitoringSwitches.scala
```diff
@@ -42,6 +42,15 @@ trait MonitoringSwitches {
     exposeClientSide = true
   )
 
+  val CspReporting = Switch(
+    "Monitoring",
+    "csp-reporting",
+    "Enables logging of CSP violations",
+    safeState = Off,
+    never,
```
TBonnin: Should we set a expiration date? so we are forced to reevaluate it in a set period of time in case we dropped the ball for whatever reason

diagnostics/app/controllers/DiagnosticsController.scala
```diff
@@ -40,9 +40,17 @@ object DiagnosticsController extends Controller with Logging {
     TinyResponse.noContent()
   }
 
-  def cssOptions = postOptions
+  def csp = Action(jsonParser) { implicit request =>
+    val shouldReport = (new Instant().getMillis) % 10000 == 1
```
TBonnin: Correct me if I am wrong but this is only random if we assume this instruction timestamps are evenly distributed throughout time, which in reality is probably not the case. 
Is there a reason why this is not using `scala.util.Random`?

johnduffell: haha, that was my fault, I suggested to use Int instead of double and then I wasn't sure how fast scala.util.Random would be instead of time.  It doesn't matter how evenly distributed it is, because we're not relying on that for the sample distribution, just for the sample size.  Unless some pages are more likely to be hit in a certain millisecond than others?
If the performance of scala.util.Random is good (I think it's pseudo random rather than using entropy) I wouldn't argue too strongly either way

TBonnin: `Unless some pages are more likely to be hit in a certain millisecond than others?` 
That's what I am questioning 😁 

Also a quick test in scala repl shows `scala.util.Random` is much faster than `Instant().getMillis`
```
scala> def time[R](block: => R): R = {
     | val t0 = System.nanoTime()
     | val result = block
     | println("Elapsed time: " + (System.nanoTime - t0) / 1e6 + "ms")
     | result
     | }
time: [R](block: => R)R

scala> time {0.to(1000000000).foreach(ignored => (new Instant().getMillis) % 10000 )}
Elapsed time: 61791.900968ms

scala> time {0.to(1000000000).foreach(ignored => scala.util.Random.nextInt(10000) )}
Elapsed time: 24994.601662ms
```

johnduffell: OK you win! :stuck_out_tongue_winking_eye:

desbo: did this count as a premature optimisation? :imp:

desbo:
@TBonnin yup, I'll add something to the dashboard to show (at least) the top pages generating insecure requests, and the top urls/domains that were serving content over http. this'll hopefully give us a clear list of things to fix

diagnostics/app/controllers/DiagnosticsController.scala
johnduffell: haha must be! lucky @TBonnin is here to put me right =)

TBonnin:
lgtm

PR #12516, opened 2016-04-11
Remove the breaking-news-from-admin-jobs switch

TBonnin:
## What does this change?

Remove the breaking-news-from-admin-jobs switch.
It has been enabled in prod for more than a week now and no problem have been noticed or reported

Less code, less bugs :)

No
## Request for comment

@sndrs


NataliaLKB:
:+1:

PR #12520, opened 2016-04-12
Switching to static jobs feed

kenlim:
The 3rd party is giving us a new static URL to poll. I'm making another poller that we can remove later.


commercial/app/commercial/feeds/FeedFetcher.scala
```diff
@@ -111,10 +111,16 @@ object FeedFetcher {
     }
   }
 
-  private val jobs: Option[FeedFetcher] = {
-    Configuration.commercial.jobsUrlTemplate map { template =>
-      new SingleFeedFetcher(JobsFeedMetaData(template))
-    }
+  private val jobs: Seq[FeedFetcher] = {
+      val dynamicDateJob = Configuration.commercial.jobsUrlTemplate map { template =>
+        new SingleFeedFetcher(JobsFeedMetaData(template))
```
rich-nguyen: Maybe rename to `DynamicJobsFeedMetaData`?

rich-nguyen:
:+1:

JonNorman:
Just updated the name of the jobs `FeedMetaData` to be "job" for both feeds - this means that both `FeedFetcher` instances (static and dynamic) will aim to write to the same file. Given that one of them is expected to fail, there shouldn't actually be any overwriting.

PR #12524, opened 2016-04-12
Use notifications endpoint

janua:
This changes the specific `gcmBrowserId` over to the more general `browserEndpoint` when saving who subscribes to what.

This purposely does not change any of the behaviour of fetching messages for a given Chrome browser; that still uses the GCM IDs.

When using the standardised spec, we can potentially use encrypted payloads, so will never need to store the individual ID anywhere.

This is a change to move to the more standardised spec; see https://github.com/guardian/frontend-notifications/pull/12.

@NathanielBennett


NathanielBennett:
:+1:

PR #12525, opened 2016-04-12
Updating message to new style

kenlim:
It used to say "This content is sponsored" but we have updated the messaging to now read "Supported by..."

Just updating the message to match


rich-nguyen:
:+1:

PR #12527, opened 2016-04-12
sup

NathanielBennett:
# My Awesome Pull Request

Or, so much pain for so little
## What does this change?

There was a problem with subscribing to live notfications on football livebloigs on android. Turns out this is because of the extra components in the left hand column for football. At the mobile breakpoiint, the tables are requested, but not inserted into the dom and for some reason, this meant the notifications click handler didn't get registered. If you look at a liveblog at the mobile breakpoint like this one ( https://www.theguardian.com/football/live/2016/apr/10/sunderland-v-leicester-premier-league-live ) and then expand it the tables don't appear so there's no change in functionality. A mobile browser will just make three less ajax calls

Live notifications work on mobile for football blogs. Un mucho grande audience.

Nope
## Screenshots

match-components
## 

### DIFF
--- a/static/src/javascripts/bootstraps/enhanced/football.js
+++ b/static/src/javascripts/bootstraps/enhanced/football.js
@@ -192,6 +192,7 @@ define([
                             renderTable(competition + '/' + resp.group, extras, dropdownTemplate);
                         }
 
+                        // Other games today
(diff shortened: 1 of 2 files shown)


NathanielBennett:
I hit return prematurely whilst typig git commit -m ''. I type the single quotes first

janua:
:+1:

PR #12529, opened 2016-04-12
Accept Thrift arraybuffer in model

rich-nguyen:
The `Tags` object was using a `Seq` to hold tags, whose type was implicitly defined by the CAPI client. In thrift format, this `Seq` was an `ArrayBuffer`, so our pattern matches began to fail.


jamespamplin:
:+1: looks good to me

PR #12532, opened 2016-04-12
don't reuse CompilerOptions across threads

johnduffell:
Deploys have been super slow recently due to boxes not coming up.

Upon looking into it we're getting some errors from google closure compiler caused by CompilerOptions [not being thread safe](https://groups.google.com/forum/#!topic/closure-compiler-discuss/7w-qplzkYf8)

This just uses one per request (although after that it's memoised into a TrieMap anyway)

This probably isn't causing the box issues, but it shouldn't be happening anyway and it'll make the logs clearer.

@jamespamplin @janua

### DIFF
--- a/common/app/common/JsMinifier.scala
+++ b/common/app/common/JsMinifier.scala
@@ -13,7 +13,7 @@ import scala.util.Try
 
 object JsMinifier {
 
-  val compilerOptions = {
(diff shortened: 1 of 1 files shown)


janua:
:+1:

rich-nguyen:
Great work

PR #12311, opened 2016-03-21
Remove notification call

gtrufitt:
Bug - Notification CTA showing on all liveblogs

PR #12533, opened 2016-04-12
Fix exception in test

OliverJAsh:
Fixes exception re. https://github.com/guardian/frontend/pull/12447

The `success` function was throwing an exception because `String.prototype.match` returned `null` instead of the expected `array`.

The effected test variant is out to a total of 50% of users. Unfortunately this means other A/B tests may not have run for those users, and/or we may not have tracked their presence to Ophan.

/cc @desbo @dominickendrick 

https://github.com/guardian/frontend/pull/12311#issuecomment-199518409


dominickendrick:
👍

PR #12535, opened 2016-04-12
Deploys radiator: remove timestamp

OliverJAsh:
This is useless because we can't configure the time on the Pi :unamused:


sndrs:
:+1:

PR #12536, opened 2016-04-12
Release pasteup master

sndrs:
## What does this change?

prevents you from releasing a new version of pasteup if you're:
- not on master
- have a dirty branch

we need to make sure that released versions are actually code we have in master, not just something you've done and then change/delete etc.
## 


static/src/stylesheets/pasteup/publish.js
```diff
@@ -12,10 +12,34 @@ var inquirer = require('inquirer');
...
+    return new Promise(function (resolve, reject) {
+        git.branch(function (branch) {
+            if (branch === 'master') {
+                resolve();
+            } else {
+                reject(new Error('notmaster'));
+            }
+        })
+    });
+}
```
OliverJAsh: Minor comment but, given the name, I would expect this function to return a boolean

static/src/stylesheets/pasteup/publish.js
```diff
@@ -12,10 +12,34 @@ var inquirer = require('inquirer');
...
+function isClean() {
+    return isGitClean().then(function (isClean) {
+        if (isClean) {
+            return true;
+        } else {
+            throw new Error('notclean');
+        }
+    })
+}
+
```
OliverJAsh: Same here

static/src/stylesheets/pasteup/publish.js
sndrs: yeah good point

johnduffell:
looks like a great idea! 👍 for the principle and the code looks good from my limited knowledge

PR #12539, opened 2016-04-13
sponsored containers now decide their branding based on cards contents

JonNorman:
Further work on the container refactoring.

The branding of a container is currently decided by inspecting the DFP query of the first card to see whether it is sponsored or not. This change uses the new container model and allows the container to infer the correct branding, derived from its contents, with the cards also inferring whether they should be branded or not (e.g. in the case where one card is sponsored and the rest are not).

This also fixes a bug that causes all ad-badge calls to fail if one of the badge calls cannot be made - now all badge calls that can be made will be made irrespective of the other badges being requested.

This behaviour has been tested against the [test page](http://localhost:9000/commercial-containers), which contains an example of each type of container.

/@regiskuckaertz


regiskuckaertz:
Niiiiice 👍  once the failing test passes

PR #12544, opened 2016-04-13
Liveblogs ads jankiness fix v2

Calanthe:
## What does this change?

The continuation of the https://github.com/guardian/frontend/pull/12507. It was causing misplacement of ads on standard articles. Now I take scroll position in consideration only on liveblog pages.
## Request for comment

@regiskuckaertz


static/src/javascripts/projects/common/modules/article/spacefinder.js
```diff
@@ -60,7 +60,11 @@ define([
 
         // fromBotton:Boolean
         // will reverse the order of slots (this is useful for lazy loaded content)
-        fromBottom: false
+        fromBottom: false,
+
+        // considerScroll:Boolean
+        // will take scroll position into consideration (this is useful for liveblog ads)
+        considerScroll: false
```
regiskuckaertz: I added the `absoluteMinAbove` setting for this purpose, thinking we should measure from the top of the page (hence the `absolute`), and this is what we want to change. It is not used anywhere than in liveblogs, so you could actually use that setting instead of creating a new one.

Calanthe: I removed the `absoluteMinAbove` setting because now we calculate the `top` value based on slot's position relative to the viewport, not absolute value from the top of the page. So we don't need to filter slots again and compare to that `absoluteMinAbove` setting.

regiskuckaertz:
Looks gooddie 👍

PR #12546, opened 2016-04-13
Revert "sponsored containers now decide their branding based on cards contents"

JonNorman:
Reverts guardian/frontend#12539


regiskuckaertz:
👍

PR #12548, opened 2016-04-13
Make sure Seq-based objects pattern match as List

rich-nguyen:
I've gone through the all code since https://github.com/guardian/frontend/pull/12529 and fixed any places where a `Seq` reference is used to pattern match with a `Nil` case. `Nil` is a subclass of `List`, not `Seq`. That means it's possible for a non-`List` object to passed in, and the pattern match will fail with an exception.

### DIFF
--- a/sport/app/rugby/feed/rugbyOptaDeserialisation.scala
+++ b/sport/app/rugby/feed/rugbyOptaDeserialisation.scala
@@ -255,7 +255,7 @@ object Parser {
   }
 
   private def chainAttributes(attributes: Iterable[MetaData]): MetaData = {
-    attributes match {
+    attributes.toList match {
       case Nil => Null
       case head :: tail => new UnprefixedAttribute(head.key, head.value, chainAttributes(tail))
     }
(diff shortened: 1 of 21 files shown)


johnduffell:
:+1: that's such a tough one, how often do we forget that List has :: and Nil, but Seq is just a trait.  This is where pattern matching just gets the danger of runtime type checking. :fearful:

sport/app/rugby/feed/rugbyOptaDeserialisation.scala
```diff
@@ -255,7 +255,7 @@ object Parser {
   }
 
   private def chainAttributes(attributes: Iterable[MetaData]): MetaData = {
-    attributes match {
```
TBonnin: For my own knowledge:
Would the following work?
```
attributes match {	
        case Seq() => Null
```

rich-nguyen: Yes. But a case like `head :: tail` is not `Seq` - based either. So I decided not to use `Seq()`, because it would only work in situations when the pattern match doesn't involve head-tail matching.

TBonnin: Got it. thx

PR #12550, opened 2016-04-13
Crosswords use CAPI's dateSolutionAvailble if present, or solutionAvailable if not.

JustinPinner:
## What does this change?

Enables crossword solution UI elements if CAPI dateSolutionAvailable or solutionAvailable fields say so.

If dateSolutionAvailable exists, and is less than or equal to today: show elements. If it doesn't exist but solutionAvailable is true: show elements. Otherwise hide elements. If this all works: SUCCESS!

No.
## Request for comment

@desbo @philmcmahon
## 


sndrs:
:+1:

desbo:
:+1:

static/src/javascripts/projects/common/modules/crosswords/main.js
```diff
@@ -597,7 +597,13 @@ define([
         },
 
         hasSolutions: function () {
```
philmcmahon: I'm not sure but I think you might need to wrap the new logic in an if('solution' in this.props.data.entries[0]). It's unlikely but I think it's possible that a crossword could have a dateSolutionAvailable in the past or solutionAvailable set to true but there to be no actual solution available in the response from capi.

PR #12551, opened 2016-04-13
allow me to switch on the test

jamesgorrie:
I very obviously don't understand the test framework.
I thought the hash override would override switches too.

@akash1810


akash1810:
:+1: though I too don't have the best knowledge on the test framework...

Issue #11515, opened 2016-01-11
"This article is X months old" glitches interactive template 


Jholder112233:
![screen shot 2016-01-11 at 15 01 25](https://cloud.githubusercontent.com/assets/2236852/12236966/4588a33e-b874-11e5-82dc-82b563927847.png)

http://www.theguardian.com/politics/ng-interactive/2015/sep/18/jeremy-corbyns-universe-labour-leaders-allies-and-enemies

PR #12554, opened 2016-04-13
Fix old message layout #11515

katebee:
## What does this change?

This fixes the layout of the old article message for tablet layout.

Fixes a bug where elements were overlaying.

This fix is targeted only at interactives with the news tone tag.
## Screenshots

picture 1

### DIFF
--- a/static/src/stylesheets/module/content/_content.scss
+++ b/static/src/stylesheets/module/content/_content.scss
@@ -1130,3 +1130,15 @@
         }
     }
 }
+
+.content--interactive {
+    .tonal__main--tone-news {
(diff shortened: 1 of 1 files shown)


ScottPainterGNM:
Would this help with the similar issue with the interactive and immersive templates possibly? See #11515 These have been raised with the desktop breakpoint but the styling is similar with the furniture appearing horizontally across the top of the page.

ScottPainterGNM:
Should have read the title, ignore me..... :see_no_evil:

johnduffell:
great, looks much better! thanks :speedboat:  :+1:

PR #12555, opened 2016-04-13
Service worker: only handle asset requests

OliverJAsh:
Occasionally HTTPS pages will request HTTP assets. Browsers block some of these requests (e.g. JS) and not others (e.g. images). In both cases you get a mixed content warning.

However, images will still fail because the service worker will try to request them via `fetch`, which will not allow any HTTP requests whatsoever.

We should only be handling requests necessary for the offline page (just requests to assets.guim.co.uk). This will mean other requests, such as images, will not be intercepted, so they can continue to load.

/cc @michaelwmcnamara


NataliaLKB:
:+1:

PR #12556, opened 2016-04-13
Adding in fix for paid-containers refactor

JonNorman:
This reverts the revert we put in after having merged #12539 and finding a bug.

The reason we had to revert initially is because we were creating the new container model regardless of whether the switch was on or not, at which point it was passed to (and then used inside) the `container` view.

In addition, the `shouldRenderAsPaidContainer` has been refactored and altered such that it only returns true when the the `containerModel.branding.sponsorshipType == "advertisement-features"`.

/@regiskuckaertz


static/src/javascripts/projects/common/modules/commercial/badges.js
```diff
@@ -69,12 +69,12 @@ define([
...
             return new Promise(function (resolve) {
                 idleFastdom.write(function () {
-                    var placeholder = $('.js-badge-placeholder', container);
+                    var placeholder = $('.js-badge-placeholder', item);
 
                     if (placeholder.length) {
                         placeholder.replaceWith($adSlot);
                     } else {
-                        $('.js-container__header', container).after($adSlot);
+                        $(opts.fallback, item).after($adSlot);
```
regiskuckaertz: Small detail but as `fallback` is mandatory, let's make it explicit by making it a bona fide function parameter.

regiskuckaertz:
👍  Yihaa!

PR #12557, opened 2016-04-14
etag should be in quotes!

johnduffell:
Turns out etag should be
ETag: "etag goes here"

note the quotes... I was mentally discounting them!

@jamespamplin


jamespamplin:
:+1:

PR #12560, opened 2016-04-14
Change password max length to 72 when changing password

jranks123:
- change max length on input boxes on change password fields to 72
- Change max length copy 
- set new max length of 72 in the validation of user input

This affects the change password, reset password and the legacy sign up pages


emma-p:
:+1:

PR #12561, opened 2016-04-14
Yes it is happening. New Merchandising component.

uplne:
# New awesome component
## What does this change?

Just another cool way how to make people buy our great services.

More £££

Nope. I wish tho.
## Screenshots

screen shot 2016-04-14 at 11 30 27
screen shot 2016-04-14 at 11 30 37
screen shot 2016-04-14 at 11 35 08

CC @regiskuckaertz


Calanthe:
Sometimes it is hard to read the text (especially the one at the bottom). I hope that the real ads will have more appropriate images.

paperboyo:
+1 what @Calanthe said. Or a permanent image wash.

uplne:
It will definitely depends on choosing a proper images @Calanthe @paperboyo 

screen shot 2016-04-14 at 14 34 05

paperboyo:
We could play safe by adding something permanent over an image or playing tricks with the text (0.5px shadow or something). We should actually do similar in couple of places - most importantly for our logo at RHS of immersive templates (yes, it **is** here: http://www.theguardian.com/society/2016/feb/09/in-defence-of-pretentiousness).

static/src/javascripts/projects/common/modules/commercial/creatives/template-preprocessor.js
```diff
@@ -7,6 +7,7 @@ define([
     'text!common/views/commercial/creatives/manual-multiple-button.html',
     'text!common/views/commercial/creatives/manual-title.html',
     'text!common/views/commercial/creatives/gimbap/gimbap-simple-blob.html',
+    'text!common/views/commercial/creatives/gimbap/gimbap-richmedia-blob.html',
```
regiskuckaertz: This module is getting quite large, I'll add this to the list of things to improve during my tenure in the health team 🎅

uplne:
@paperboyo that's a good point. We actually added some text shadow to other component but I will add it to this one too.

static/src/stylesheets/module/commercial/gimbap/_gimbap-richmedia.scss
```diff
@@ -0,0 +1,72 @@
+%gimbap-zindex {
+    position: relative;
+    z-index: 1;
+}
+
+.gimbap-richmedia__title {
+    @extend %gimbap-zindex;
```
regiskuckaertz: Nice!

static/src/stylesheets/module/commercial/gimbap/_gimbap-richmedia.scss
```diff
@@ -0,0 +1,72 @@
...
+    @include fs-headline(3);
```
regiskuckaertz: I've been using `get-font-size` recently, especially since most of the time I overwrite `font-weight`, `line-height`, etc.

static/src/stylesheets/module/commercial/gimbap/_gimbap-richmedia.scss
```diff
@@ -0,0 +1,72 @@
...
+
+    font-weight: bold;
+    line-height: .8;
+    color: #ffffff;
+
+    .has-flex & {
```
regiskuckaertz: Why? You don't need this guard since browsers which don't support flexbox won't understand `align-self`. Plus this way it works without JS 💥

regiskuckaertz:
👍

PR #12565, opened 2016-04-14
Don't show empty adslot when there is no ad loaded

Calanthe:
## What does this change?

This PR fixes the intend bug which occurs when there is no isbn number provided and empty ad slot makes unwanted space in the article body. When there is no add call, the empty ad slot should be hidden.

No intend in the article body.
## Screenshots

Before:
screen shot 2016-04-14 at 14 59 57

After:
screen shot 2016-04-14 at 15 01 40


static/src/javascripts/projects/common/modules/commercial/creatives/commercial-component.js
```diff
@@ -56,7 +56,7 @@ define([
         return function (params) {
             var isbn = config.page.isbn || params.isbn;
             if (isbn) {
-                return buildComponentUrl(url, merge(params, { t: config.page.isbn || params.isbn }));
+                return buildComponentUrl(url, merge(params, isbn));
```
jbreckmckye: `isbn` is an object?

jbreckmckye:
:+1:

PR #12566, opened 2016-04-14
Updated healthcheck for training-preview

jamespamplin:
Healthcheck is currently failing for training-preview. This is due to a CAPI reindex of content on CODE whereby the previous healthcheck no longer exists.
## What does this change?

Healthcheck on training-preview

Fixes training-preview - used by editorial for training on composer code

No
## Screenshot

Of the page being healthchecked:

training-preview-healthcheck

### DIFF
--- a/training-preview/app/controllers/TrainingHealthCheck.scala
+++ b/training-preview/app/controllers/TrainingHealthCheck.scala
@@ -40,7 +40,7 @@ class TrainingHttp extends contentapi.Http with ExecutionContexts {
 
 object TrainingHealthCheck extends AllGoodHealthcheckController(
  9016,
- "/world/2012/sep/11/barcelona-march-catalan-independence"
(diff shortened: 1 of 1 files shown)


LATaylor-guardian:
@jamespamplin Why are you checking for a particular id? Are you trying to test simply whether CAPI is up? In this case would it more resilient to check in some manner that does not rely on a particular piece of content being indexed? For example hitting the base url and checking the CAPI returned status in the response.

jamespamplin:
@LATaylor-guardian All the other healthchecks in dotcom work similar to this, by checking the response of a rendered page.

But you're right, checking the route response from CAPI may be more resilient to changes, but for now I think we should keep the healthcheck consistent with the rest of the apps.

TBonnin:
+1 for this PR

+1 for @LATaylor-guardian idea of not relying on a specific piece of content :)

PR #12567, opened 2016-04-14
Initialise phone number reflection on application startup

markjamesbutler:
# Initialise phone number reflection on application startup
## What does this change?

This moves the potentially slow reflection code into application startup rather than executing on rendering of the account edit page.

Faster rendering of the account edit page.  Blocking reflection code is removed from page rendering.

No
## Screenshots

None
## 

### DIFF
--- a/identity/app/conf/IdentityLifecycle.scala
+++ b/identity/app/conf/IdentityLifecycle.scala
@@ -2,6 +2,7 @@ package conf
 
 import common.{AkkaAsync, Jobs, ExecutionContexts}
 import jobs.{TorExitNodeList, BlockedEmailDomainList}
+import model.PhoneNumbers
 import play.api.GlobalSettings
 
 trait IdentityLifecycle extends GlobalSettings with ExecutionContexts {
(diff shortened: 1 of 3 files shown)


jamespamplin:
:+1: looks good to me

johnduffell:
great! 👍

johnduffell:
is reflection really that slow?  I thought it was a bad idea because it takes away compile time safety?  perhaps swapping the hard coded list for reflection is swapping one thing for another.  Personally I'd prefer the list.  Good on you for fixing the Canada problem though, I got completely stumped by that one.

janua:
@johnduffell Good question...
http://stackoverflow.com/questions/435553/java-reflection-performance
Which leads to http://docs.oracle.com/javase/tutorial/reflect/index.html

johnduffell:
@janua thanks for googling that for me 😝 I guess they mean the JIT can't compile it very well.

Interestingly, in that document oracle think the drawbacks of reflection are Performance Overhead, Security Restrictions and Exposure of Internals.  They didn't think about compile time safety?

johnduffell:
come on, ship it! 😝

PR #12569, opened 2016-04-14
Fix CSP error messages

desbo:
These are happening, but for some reason the second parameter isn't showing up in CloudWatch.

### DIFF
--- a/diagnostics/app/model/diagnostics/DynamoDbReport.scala
+++ b/diagnostics/app/model/diagnostics/DynamoDbReport.scala
@@ -72,7 +72,7 @@ object CSPDynamoDbReport extends DynamoDbReport[CSPReport] {
       .withAttributeUpdates(count.asJava)
 
     client.updateItemFuture(update) onFailure {
-      case error: Throwable => log.error(s"Failed to log CSP report to DynamoDB", error)
+      case error: Throwable => log.error(s"Failed to log CSP report to DynamoDB: $error", error)
     }
   }
 }


diagnostics/app/model/diagnostics/DynamoDbReport.scala
```diff
@@ -72,7 +72,7 @@ object CSPDynamoDbReport extends DynamoDbReport[CSPReport] {
       .withAttributeUpdates(count.asJava)
 
     client.updateItemFuture(update) onFailure {
-      case error: Throwable => log.error(s"Failed to log CSP report to DynamoDB", error)
+      case error: Throwable => log.error(s"Failed to log CSP report to DynamoDB: $error", error)
```
NathanielBennett: Does that work? $error isnt a string ... ${error.getMessage] mebbe. You don't need another arg for string interpolation
Or am I being **really** stupid?

desbo: because it's being interpolated it'll call `toString` on the error.
the second arg is meant to show the full trace in cloudwatch, but that's not working

NathanielBennett: Yeah so can't you just do 

case error: Throwable => log.error(s"Failed to log CSP report to DynamoDB: $error")

?

desbo: you can do that if you want to exclude the error object

`log.error` can also take two parameters, a string and a `Throwable`, which allows stack trace reporting as well as the message: https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/scala/play/api/Logger.scala#L139

NathanielBennett: OK, my bad
:+1:

PR #12570, opened 2016-04-14
Set correct credentials for Logback kinesis appender

TBonnin:
## What does this change?

Set correct credentials for Logback kinesis appender

Also refactor LogbackConfig so it doesn't rely on the Configuration
object

Applications can then publish logs to the kinesis stream

No
## Request for comment

@jamespamplin


jamespamplin:
:+1: much nicer, great stuff

PR #12572, opened 2016-04-15
Extend Header Bidding

regiskuckaertz:
The header bidding AB test expired two days ago, but the US wants it to run like forever.

/cc @JonNorman @kenlim


JonNorman:
:+1:

JonNorman:
awkward... :+1:

PR #12573, opened 2016-04-15
Add Logstash trait to all apps

TBonnin:
## What does this change?

Add Logging trait to all application Global object

Make sure logs are sent to Logstash when using Play default Logger

No
## Request for comment

@jamespamplin


jamespamplin:
:+1: good stuff!

PR #12574, opened 2016-04-15
Move the AB Header Bidding test to a bona fide commercial switch

regiskuckaertz:
Header bidding is a go in the US and they want it forever for everybody — something which the AB test does not make possible.


JonNorman:
Lovely :+1:

PR #12575, opened 2016-04-15
Fix GET /news-alert/alerts on preview

TBonnin:
## What does this change?

GET /news-alert/alerts is currently failing on preview. This patch fixes the problem by adding admin-jobs to the standalone app and creating the correct route in standalone routes file

No more error when fetching /news-alert/alerts from preview
## Request for comment

@jamespamplin


jamespamplin:
:+1: thanks for fixing this

Issue #12216, opened 2016-03-09
Lightobx is requesting too big images. Lightbox is requesting too small images.


paperboyo:
Hello,

- [ ] Portrait images in Lightbox are often requested at crazily high resolutions, because they are all a fixed 1920px wide at widest BP. Example: http://www.theguardian.com/environment/gallery/2015/sep/14/the-british-wildlife-photography-awards-2015-winners-in-pictures#img-7 (2878px deep). Lightbox should take viewport into consideration and, maybe to be safe when changing orientation at the very least, should never request images' longer side to be longer than the vieport's longer side. If that makes sense which it probably doesn't, but the current behaviour is still broken.

- [ ] For some reason, gallery pages and the Lightbox, never triggers `dpr=2` logic from https://github.com/guardian/frontend/pull/12079. 

Related: https://github.com/guardian/frontend/pull/12079

There is so much more we could do to the Lightbox starting with fixing the mobile experience. Or small things like still occurring https://github.com/guardian/frontend/issues/6144. Or the big things. Really big :-)

Regards
Mateusz

rich-nguyen:
This issue is two-fold:
- We only infer image resizing parameters based on the breakpoint width, which we map to fixed image content widths. This is a landscape-only solution, and is used by all frontend, including lightbox.
- The lightbox client-side system generates its own image html using `<img>`, but the modern conventional server-side system uses `<picture>`. The img and srcset parameters for lightbox images come from the server through server-prepared lightbox JS config.

The advantage to using srcsets is we don't account for viewport, the browser does. But this assumes the srcset is correct, and its not possible for the browser to do anything sensible for this image because the srcset is specifying 'height-over-sized' image content, as mentioned.

Broadly, here are the goals for this task:
1. Update the lightbox and gallery systems to use `<picture>` and `media=`.
2. Modify server logic to use a new concept, `HeightsByBreakpoint`, to complement the existing `WidthsByBreakpoint`. This would fix all portrait images on the website.

Sounds simple!

paperboyo:
Big thanks for the thorough explanation and a plan. One small thing we _may_ consider **only** for Lightbox images is using `fm=pjpg` imgIX parameter to (only perceptually?) speed up the carousel. Would require some tests, as I'm pretty sure, `q` is not quality-normalised between `jpg` and `pjpg` (not that it is among most other formats...).

rich-nguyen:
[...]

Issue #12215, opened 2016-03-09
Make Lightbox images fill the screen again


paperboyo:
Hello,

To cut long story short, images in the Lightbox should always fill the screen.

Long story:
The lack of this obvious functionality was [discussed here](https://github.com/guardian/frontend/pull/5934), where the [only argument for not having it](https://github.com/guardian/frontend/pull/5934#issuecomment-55766272) was to punish readers, so that the users would learn to use higher quality images. This problem is mostly gone/solved because of mediaservice. There are plenty of arguments **for** having this feature, too many for me to list them unprompted :)

Also, since we've used imgIX, images in Lightbox actually **were** filling it. The lack of `fit=max` meant we were requesting them upsized from imgIX. That was crazy for all sorts of reasons, but meant that this particular feature was there. It changed in https://github.com/guardian/frontend/pull/12079. So we are back to this (e.g. for all old galleries):

![get the loupe](https://cloud.githubusercontent.com/assets/6032869/13638033/4e3015c4-e602-11e5-82eb-c223f8a3a176.jpg)

Let's get rid of that!

Regards
Mateusz

paperboyo:
This has been “fixed”. Because of some bug in PNG, we upsize images up. Bit silly, but they look better. Anyhow, closing as no longer relevant.

PR #12576, opened 2016-04-15
Gallery redesign WIP: Initial PR

NataliaLKB:
## What does this change?

@zeftilldeath designed what galleries should look like, this is the initial PR implementing the design.

This will start as a 0% A/B test while we iterate. Next week the aim is to start rolling it out an audience 
## What is next?

These will all be done in future prs
- [x] Use picture element instead of srcset. We will refactor the scala img helper to reuse that, but outside this PR
- [x] Try to eliminate jank, at least where pictures are loading. Had a talk with @OliverJAsh yesterday about how to do this.
- [ ] Add animation for when you scroll down (ask myself or @zeftilldeath for the video if you want to see)
- [ ] On mobile, option to only see images and not captions.
- [ ] Set a limit for caption length on both mobile and desktop
- [x] Fastly Change persistent cookie for optin so that people in the business can take a look without adding a header (https://github.com/guardian/fastly-edge-cache/pull/614)
- [ ] Fix 100vh jumping on mobile across immersives. talked to @gtrufitt about this
- [x] MPU slot styling

No
## Screenshots

Mobile:
image
image

Desktop:
image
image
## Request for comment

@OliverJAsh


applications/app/views/fragments/newGalleryBody.scala.html
```diff
@@ -0,0 +1,124 @@
+@(gallery: model.Gallery, related: model.RelatedContent, index: Int)(implicit request: RequestHeader)
+@import common.{Edition, LinkTo}
+@import layout.ContentWidths.GalleryMedia
+@import views.support.TrailCssClasses.toneClass
+@import views.support.{ImgSrc, RenderClasses}
+
+
+@*
+
+TODO:
```
NataliaLKB: I left a lot of todos in the code for future iteration. Apologies in advance if it is confusing!

paperboyo:
OMG!!111ONE! :heart_eyes: 

"Try to eliminate junk" & "Fix 100vh jumping on mobile" - maybe FulscreenAPI? Anyway: :beers: for whomever succeeds!

Maybe also worth looking at the very small related https://github.com/guardian/frontend/issues/6144?

At the end, when all is done and everybody's drinking :champagne:, we may think of joining galleries from the same series? Maybe? Like in this screenshot (or even without the disruptive endslate altogether): https://github.com/guardian/frontend/pull/5934#issuecomment-55599747?

swipe...swipe...swipe...swipe...ad_impression...swipe...swipe...swipe...swipe...ad_impression...swipe...swipe...swipe...swipe...

jamesgorrie:
@akash1810 Look at this beauty (not talking about @paperboyo )...

applications/app/views/fragments/newGalleryBody.scala.html
```diff
@@ -0,0 +1,145 @@
...
+                    src="@ImgSrc.getFallbackUrl(imageElement.images)" />
+                </picture>
+            </div>
+        </figure>
+    </li>
+
+
+
+
+
```
sammorrisdesign: whitespace

common/app/views/fragments/immersiveGalleryHeadTonal.scala.html
```diff
@@ -0,0 +1,90 @@
+@(item: model.ContentType, showBadge: Boolean = false, showMeta: Boolean = true)(implicit request: RequestHeader)
+
+@import views.support.TrailCssClasses.toneClass
+@import views.support.ContributorLinks
+
+<header class="content__head content__head--immersive content__head--desktop is-fixed-height tonal__head tonal__head--@toneClass(item)">
```
sammorrisdesign: Is this that different from the normal `headerImmersive.scala.html`?

I wonder if it's worth sharing parts of this if not the whole thing.

static/src/stylesheets/module/content/_gallery.scss
```diff
@@ -245,3 +245,313 @@
...
+    .content__main-column {
+        max-width: none;
+
+        @include mq(tablet) {
+            margin-left: gs-span(3);
+        }
+    }
+
+    @include mq(tablet) {
+        .content__dateline, .content__labels, .byline, .content__meta-container, .meta__break, .content__headline, .content__standfirst {
```
sammorrisdesign: Shouldn't these be written as

```sass
.content__dateline,
.content__labels,
.byline
```

static/src/stylesheets/module/content/_gallery.scss
```diff
@@ -245,3 +245,313 @@
...
+    .content__head--immersive.is-fixed-height {
+        .content__header {
+            position: relative;
+        }
+
+        .content__head--mobile {
+            position: relative;
+            background-position: center center;
+            -webkit-background-size: cover;
+            background-size: cover;
```
sammorrisdesign: Doesn't autoprefixer mean we don't have to do add the `-webkit` rule?

NataliaLKB: Not sure, haven't written much scss in the project so far. Will remove now!

NataliaLKB: Removed. Thank you for reviewing! More comments, the better 👍

sammorrisdesign:
There are a bunch of sass whitespace and consistency things but I'm not going to be a pedant.

Only other thing I could think of is that the header is called `immersive` and the body is called `new`. Probably will be clearer to give it a single name.

Otherwise it's great that we're sorting these galleries out and it looks good. Probably will be better to get someone who knows what they're doing with scala for that side of things though.

NataliaLKB:
@sammorrisdesign, do let me know about the sass whitespace and consistency things! Its useful to me, and easy to fix. 

Regarding naming, once galleries are replaced, the `new` prefix can be removed.

common/app/views/fragments/immersiveGalleryHeadTonal.scala.html
NataliaLKB: I will take a look if it is worth moving headerImmersive to `common/`. It is a good point that it may not be necessary to use both.

sammorrisdesign:
But aren't the `immersiveGalleryHead` and `newGalleryBody` linked together? I'm just thinking in terms of someone coming to this fresh, they might not realise they are linked.

static/src/stylesheets/module/content/_gallery.scss
```diff
@@ -245,3 +245,321 @@
...
+        .content__standfirst {
+            padding-bottom: 0;
+            padding-top: 0;
+            &:before {
+                content: none;
+            }
+        }
+    }
+
+
```
sammorrisdesign: Double return inconsistent with single return below. I think we tend to have single lines after blocks. But should at least be consistent in same file.

static/src/stylesheets/module/content/_gallery.scss
```diff
@@ -245,3 +245,321 @@
...
+        .tonal__header {
+            background-color: rgba(0, 0, 0, 0);
+        }
+        .meta__numbers {
+            color: $neutral-3;
+        }
+        .content__dateline, .sharecount__heading {
+            color: $neutral-2;
+        }
+        .byline a, .sharecount__value {
```
sammorrisdesign: Same as above too. Selectors should get their own line each

static/src/stylesheets/module/content/_gallery.scss
```diff
@@ -245,3 +245,321 @@
...
+    .content__labels {
+        border: 0;
+        margin-bottom: 0;
+        width: auto;
+        float: none;
+    }
+
+    .block-share--gallery {
+        position: relative;
+        .block-share__item--facebook,
```
sammorrisdesign: Again inconsistent with other parts of file. Having an empty line before a selector. I think the preferred option is to have an empty line before.

static/src/stylesheets/module/content/_gallery.scss
```diff
@@ -245,3 +245,321 @@
...
+        @include mq($until: phablet) {
+            margin-left: -$gs-gutter;
+            margin-right: -$gs-gutter;
+        }
+        @include mq($until: mobileLandscape) {
+            margin-left: -$gs-gutter / 2;
+            margin-right: -$gs-gutter / 2;
+        }
+
+        @include mq(tablet) {
```
sammorrisdesign: This media query has empty line before, one above doesn't. Again I'd opt for spacing them out but it's personal preference.

NataliaLKB: Agreed, it is nicer to read. Done :)

NataliaLKB:
> But aren't the `immersiveGalleryHeadTonal` and `newGalleryBody` linked together? I'm just thinking in terms of someone coming to this fresh, they might not realise they are linked.

Thats fair. I will look first to seeing I can get rid of `immersiveGalleryHeadTonal`, and using the immersiveHeader.

applications/app/controllers/OptInController.scala
```diff
@@ -20,10 +20,12 @@ object OptInController extends Controller {
...
       case "https" => HTTPS.opt(choice)
       case "header" => Header.opt(choice)
+      case "gallery" => gallery.opt(choice)
       case _ => NotFound
     }
   }
 
   val HTTPS = OptInFeature("https_opt_in")
   val Header = OptInFeature("new_header_opt_in")
+  val gallery = OptInFeature("gallery_redesign_opt_in")
```
johnduffell: just be careful about splitting the cache too much, otherwise the backend has to serve the same content twice as much

applications/conf/routes
```diff
@@ -51,6 +51,8 @@ GET        /https/optin
 GET        /https/optout                                                        controllers.OptInController.handle(feature = "https", choice = "out")
 GET        /new-header/optin                                                    controllers.OptInController.handle(feature = "header", choice = "in")
 GET        /new-header/optout                                                   controllers.OptInController.handle(feature = "header", choice = "out")
+GET        /gallery/optin                                                       controllers.OptInController.handle(feature = "gallery", choice = "in")
+GET        /gallery/optout                                                      controllers.OptInController.handle(feature = "gallery", choice = "out")
```
johnduffell: the above routes are just legacy ones, from now on you just need to use the /opt/ route

johnduffell: although, worth adding a comment there to explain that to other people

common/app/views/fragments/immersiveGalleryHeadTonal.scala.html
johnduffell: on the topic of common, we should really only keep things that are genuinely common in here not just because stuff is needed in more than one project.  We have a lot of stuff that's only common between two or more of facia/applications/article which needn't be compiled into admin/diagnostics/etc.  I wonder if it would be better to start putting things in a new module.. I've been wondering for a while though 💭

applications/conf/routes
NataliaLKB: Ah! Thanks!

NataliaLKB: Should these ones be deleted?

applications/app/controllers/OptInController.scala
NataliaLKB: Its true, but we need to be able to test the value of this before we roll it out to everyone.

common/app/views/fragments/immersiveGalleryHeadTonal.scala.html
NataliaLKB: A new module is an interesting idea. Would this be just to share things between projects like article, applications, and facia and not compile to other applications?

Does it makes more sense in this instance to try to combine these to files with the necessary if statements in common, or have too similar files one in article, and one in applications.

johnduffell: yeah I mean your first paragraph, I feel like the group of semi-common stuff is as big as the actual common stuff, so it makes sense to keep it separate.  However your PR is probably not the right place to do too much!  As for this file, I haven't investigated how much is in common with the other file, but if it's more common than not it makes sense to combine them.

NataliaLKB: Thanks for your comments @johnduffell and @sammorrisdesign.

The only really common part between the two headers is the style tag. I have put that new template in the common folder named very verbosely `immersiveHeaderBackgroundStyling`. It is in common as it is shared between articles and applications - this can change when we have the separate module that @johnduffell may create.

`headerImmersive` remains in article, and `galleryHeader` is in applications as that is where they are used. 

Hopefully that makes things more clear. Let me know if this is better!

article/app/views/fragments/headerImmersive.scala.html
```diff
@@ -15,7 +12,7 @@
                 @fragments.inlineSvg("guardian-logo-160", "logo", Seq("content__logo"))
             </a>
             @if(article.isUSMinute) {
-                <a class="logo--us-election" href="@LinkTo {/us-news/us-elections-2016}">
```
NataliaLKB: removing space for consistency

common/app/views/fragments/header.scala.html
```diff
@@ -9,7 +9,7 @@
     @fragments.newHeader()
 } else {
     <header id="header"
-            class="l-header u-cf @if(page.metadata.hasSlimHeader) {l-header--is-slim l-header--no-navigation} js-header"
+            class="l-header u-cf @if(page.metadata.hasSlimHeader) {l-header--is-slim l-header--no-navigation} @if(mvt.ABGalleryRedesignVariant.isParticipating){l-header--is-slim l-header--no-navigation l-header--new-gallery} js-header"
```
OliverJAsh: Does this mean you'll get the slim header if you're in the test but not on a gallery?

NataliaLKB: Updated 😄

static/src/stylesheets/module/content/_gallery.scss
```diff
@@ -245,3 +245,333 @@
...
+// TODO: put in pasteup
+$multimedia-support-5: #161616;
+
+// The main image needs to take up 100vh and the slim header needs to sit just on top of it.
+.l-header--new-gallery {
+    position: absolute;
+    z-index: 1;
+    right: 0;
+    left: 0;
+}
```
OliverJAsh: Out of curiosity, should the image appear as 100vh inclusive or exclusive of the header? Currently it's inclusive of the header if I understand correctly?

If it's the former, this approach might be problematic on devices with a landscape orientation:

image

applications/app/views/fragments/newGalleryBody.scala.html
```diff
@@ -0,0 +1,127 @@
...
+            ),
+            "content", "content--media", "content--gallery", "tonal", "new-gallery", "content--immersive", s"tonal--${toneClass(gallery)}"
+        )"
+        itemscope itemtype="@gallery.metadata.schemaType" role="main">
+
+        @if(gallery.commercial.isAdvertisementFeature){
+            @fragments.guBand()
+        }
+
+        @fragments.galleryHeader(gallery, gallery.commercial.isSponsored(Some(Edition(request))) || gallery.commercial.isFoundationSupported || gallery.commercial.isAdvertisementFeature)
```
OliverJAsh: The gallery body (this file) contains the header. This slightly threw me off! Perhaps we could have `gallery` and `newGallery` files instead?

applications/app/views/fragments/galleryHeader.scala.html
```diff
@@ -0,0 +1,93 @@
...
+            @@media (min-width: @{layout.Wide.minWidth}px) {
+                .content__header--background-image {
+                    background-image: url('http:@Html(ImgSrc.findNearestSrc(picture.images, Profile(width = Some(2200))).getOrElse(""))');
+                }
+            }
+        }
+    </style>
+
+    <div class="content__header tonal__header">
+        <div class="content__header--background-image">
```
OliverJAsh: In BEM, “modifiers” tend to live on the same element as the “block”, e.g. `content__header content__header--background-image`. If you need it to my a descendent, should it be a child element instead, i.e. `conent__header__background-image`?

applications/app/views/fragments/galleryHeader.scala.html
```diff
@@ -0,0 +1,93 @@
...
+            @@media (min-width: @{layout.Phablet.minWidth}px) {
+                .content__header--background-image {
+                    background-image: url('http:@Html(ImgSrc.findNearestSrc(picture.images, Profile(width = Some(1300))).getOrElse(""))');
+                }
+            }
+            @@media (min-width: @{layout.Wide.minWidth}px) {
+                .content__header--background-image {
+                    background-image: url('http:@Html(ImgSrc.findNearestSrc(picture.images, Profile(width = Some(2200))).getOrElse(""))');
+                }
+            }
```
OliverJAsh: Could we use `object-fit` on an image for the same purpose?

Without the `img` in the DOM, it won't be machine readable (SEO, Pocket, screen readers, etc.).

static/src/stylesheets/module/content/_gallery.scss
```diff
@@ -245,3 +245,333 @@
...
+    // landscape gallery images now are wider than they were before.
+    .gallery__img-container--landscape,
+    .gallery__img--landscape {
+        width: 100%;
+        height: auto;
+    }
+
+    .content__header--background-image {
+        height: 100vh;
+        position: relative;
```
OliverJAsh: Why do you need this? I'm sure there's a good reason for it, perhaps like a comment to say why?

static/src/stylesheets/module/content/_gallery.scss
```diff
@@ -245,3 +245,333 @@
...
+        background-position: center center;
+        background-size: cover;
+        display: block;
```
OliverJAsh: I don't think this is needed as the style applies on a div.

static/src/stylesheets/module/content/_gallery.scss
```diff
@@ -245,3 +245,333 @@
...
+    }
+
+    // immersive Overwrites
+    .content__head--immersive.is-fixed-height {
+        .content__header {
+            position: relative;
```
OliverJAsh: Minor comment but if you're unsetting, `position: static` is the default.

static/src/stylesheets/module/content/_gallery.scss
```diff
@@ -245,3 +245,333 @@
...
+    .content__series-label,
+    .content__section-label {
+        display: inline-block;
+    }
+
+    .content__labels {
+        border: 0;
+        margin-bottom: 0;
+        width: auto;
+        float: none;
```
OliverJAsh: Instead of overriding these styles, perhaps we should refactor `content__labels` so it only has common styles and then create another class such as `content__labels--article` with article specific styles. This way we won't need to have any overrides. What do you think?

Overrides lead to lots of mess in CSS :-(

NataliaLKB: Yes that makes sense. So we could have `content__labels` `content__labels--gallery` and `content__labels--article`

applications/app/views/fragments/galleryHeader.scala.html
sndrs: `object-fit` won't work in IE, but is this a background image because it's supporting content? not clear why it's not an `img`

applications/app/views/fragments/newGalleryBody.scala.html
```diff
@@ -0,0 +1,127 @@
...
+                        }
+                    }
+
+                    <!--[if IE 9]></video><![endif]-->
+
+                    <img class="gallery__img gallery__img@if(image.width >= image.height) {--landscape} else {--portrait}"
+                    itemprop="contentUrl"
+                    alt="@image.altText.getOrElse("")"
+                    src="@ImgSrc.getFallbackUrl(imageElement.images)" />
+                </picture>
```
OliverJAsh: Will it be possible to re-use `img.scala.html` for the `<picture>` element, like all other pages do?

applications/app/views/fragments/galleryHeader.scala.html
```diff
@@ -0,0 +1,93 @@
...
+        }
+    </style>
+
+    <div class="content__header tonal__header">
+        <div class="content__header--background-image">
+            <div class="content__wrapper content__wrapper--headline">
+                <div class="gs-container">
```
sndrs: does this need its own div?

applications/app/views/fragments/galleryHeader.scala.html
```diff
@@ -0,0 +1,93 @@
...
+            </div>
+        </div>
+
+        <div class="gs-container">
+            @fragments.contentMeta(item)
+        </div>
+
+        @if(item.fields.standfirst.isDefined) {
+            <div class="content__wrapper--standfirst">
+                <div class="gs-container">
```
sndrs: again, does it need to be a separate div?

applications/app/views/fragments/galleryHeader.scala.html
NataliaLKB: not sure, but happy to give it a try. If we go that route, we should also change immersive articles  and the minute to be the same

applications/app/views/fragments/newGalleryBody.scala.html
```diff
@@ -0,0 +1,127 @@
...
+        itemscope itemtype="@gallery.metadata.schemaType" role="main">
+
+        @if(gallery.commercial.isAdvertisementFeature){
+            @fragments.guBand()
+        }
+
+        @fragments.galleryHeader(gallery, gallery.commercial.isSponsored(Some(Edition(request))) || gallery.commercial.isFoundationSupported || gallery.commercial.isAdvertisementFeature)
+
+        <div class="content__main tonal__main tonal__main--@toneClass(gallery)">
+            <div class="gs-container">
```
sndrs: there's a few of these ;)

applications/app/views/fragments/newGalleryBody.scala.html
```diff
@@ -0,0 +1,127 @@
...
+                    @fragments.witnessCallToAction(gallery.content)
+                    @fragments.submeta(gallery)
+                </div>
+            </div>
+        </div>
+    </article>
+
+    @if(gallery.content.showInRelated &&
+        !gallery.commercial.isAdvertisementFeature ) {
+        <div class="gallery__most-popular facia-container fc-container--media hide-on-childrens-books-site">
```
sndrs: can these all be collapsed into one div?

applications/app/views/fragments/newGalleryBody.scala.html
```diff
@@ -0,0 +1,127 @@
...
+            </div>
+        </div>
+    }
+    @fragments.contentFooter(gallery, related, "media")
+</div>
+
+@galleryItem(adSlots: Seq[String], adInterval: Int, image: model.ImageAsset, rowNum: Int, imageElement: model.ImageElement) = {
+
+    <li id="img-@rowNum" class="gallery__item js-gallery-item" data-link-name="Gallery item | @rowNum">
+        <figure itemscope itemtype="http://schema.org/ImageObject">
```
sndrs: i think this will be the third instance of very similar code in the codebase, is it similar enough to reuse an existing image fragment? or if not, should they be made similar enough and reused?

static/src/stylesheets/module/content/_gallery.scss
```diff
@@ -245,3 +245,341 @@
         }
     }
 }
+
+// New gallery stuff
+
+// TODO: put in pasteup
+$multimedia-support-5: #161616;
```
sndrs: this should be very simple now - and should go in `_vars.scss` at least?

static/src/stylesheets/module/content/_gallery.scss
```diff
@@ -245,3 +245,341 @@
...
+            }
+
+            &.inline-icon, .inline-icon {
+                fill: $neutral-3;
+            }
+        }
+    }
+
+    .social-icon, .block-share__item {
+        background-color: rgba(0, 0, 0, 0);
```
sndrs: why a background colour of `0` opacity?

applications/app/views/fragments/galleryHeader.scala.html
gtrufitt: FWIW I did use object fit for 100vh images in the minute: https://github.com/guardian/frontend/pull/11985/files#diff-c557d31de28ab1c0964b852829df0b8fR548

static/src/stylesheets/module/content/_gallery.scss
```diff
@@ -245,3 +245,341 @@
...
+            padding: $gs-gutter / 2;
+        }
+
+        @include mq(tablet) {
+            @include fs-textSans(2, true);
+            position: absolute;
+            width: gs-span(3);
+            top: 0;
+            right: auto;
+            margin-left: -(gs-span(3));
```
sndrs: `0 - gs-span(3)`? feels simpler to read (not a blocker)

static/src/stylesheets/module/content/_gallery.scss
```diff
@@ -245,3 +245,341 @@
...
+        color: $neutral-4;
+        margin-bottom: 0;
+
+        @include mq(tablet) {
+            padding-right: $gs-gutter / 2;
+        }
+
+        a {
+            color: $neutral-4;
+            border-bottom: 1px solid #767676;
```
sndrs: what colour is this then?

static/src/stylesheets/module/content/_gallery.scss
```diff
@@ -245,3 +245,341 @@
...
+            }
+        }
+
+        > strong:first-child {
+            @include fs-headline(2);
+            color: $neutral-5;
+            display: block;
+            margin-bottom: $gs-baseline/2;
+
+            & + br {
```
sndrs: `.new-gallery .gallery__caption > strong:first-child + br`  👀 

is that really a selector you want to be writing! :stuck_out_tongue:

static/src/stylesheets/module/content/_gallery.scss
sammorrisdesign: That looks like some `$neutral-2`

applications/app/controllers/OptInController.scala
desbo: will you incorporate opting in into the a/b test somehow? otherwise it seems like opting in won't give you the new gallery :)

common/app/mvt/MultiVariateTesting.scala
```diff
@@ -47,6 +47,17 @@ object ABNewHeaderVariant extends TestDefinition(
...
 }
 
+object ABGalleryRedesignVariant extends TestDefinition(
+  variants = Nil,
+  name = "ab-gallery-redesign-variant",
+  description = "Feature switch (0% test) for gallery redesign",
+  sellByDate = new LocalDate(2016, 5, 10)
+) {
+  override def isParticipating(implicit request: RequestHeader): Boolean = {
+    request.headers.get("X-GU-ab-gallery-redesign").contains("variant") && switch.isSwitchedOn && ServerSideTests.isSwitchedOn
```
desbo: do you need this header? I'm not sure there's any need to override `isParticipating` here

static/src/stylesheets/module/content/_gallery.scss
NataliaLKB: Happy to move to `_vars.scss` but would prefer just to put it in paste-up palette, is there documentation on how to do this?

sndrs: we're in the process of moving and documenting, but you can now just the value to `static/src/stylesheets/pasteup/palette/_src.scss`.

we need to define the relationship between `_vars` and `palette/_src.scss`, but we can take care of that

static/src/stylesheets/module/content/_gallery.scss
NataliaLKB: Once the `.new-gallery` class is removed it will be exactly like: https://github.com/guardian/frontend/blob/320b993b0bd6c51e97bd8ed9adc92cf52f571f41/static/src/stylesheets/module/content/_gallery.scss#L96

common/app/mvt/MultiVariateTesting.scala
NataliaLKB: Thanks for asking @desbo. As far as I am aware, if we want to run a serverside test on 30% of people (made up number) we need a new header, and to override `isParticipating`

applications/app/controllers/OptInController.scala
NataliaLKB: Do you mean in fastly? I have branch I pushed up waiting to go! Just want to get this out before I make another PR. If that isn't what you mean, come talk to me? Not sure I follow.

applications/app/views/fragments/newGalleryBody.scala.html
```diff
@@ -0,0 +1,127 @@
...
+@import views.support.TrailCssClasses.toneClass
+@import views.support.{ImgSrc, RenderClasses}
+
+<div class="l-side-margins l-side-margins--media l-side-margins--new-gallery">
+
+    <article id="article" class="@RenderClasses(
+            Map("content--advertisement-feature" -> gallery.commercial.isAdvertisementFeature,
+                "paid-content--advertisement-feature" -> gallery.commercial.isAdvertisementFeature
+            ),
+            "content", "content--media", "content--gallery", "tonal", "new-gallery", "content--immersive", s"tonal--${toneClass(gallery)}"
```
OliverJAsh: Do we need `content--immersive`? Seems to me like it just creates more work for us as we need to override its styles later on.

paperboyo:
Oh, apart from already cited https://github.com/guardian/frontend/issues/6144 it _might be_ prudent to add https://github.com/guardian/frontend/issues/12215 & (at least half of) https://github.com/guardian/frontend/issues/12216 to the grand plan. There is also https://github.com/guardian/frontend/issues/9957, but my personal opinion is that currently zooming in is very broken on mobile, because it zooms in **both** the image and the UI (including captions, up to a point of no return). If we can’t make it so that only the image zooms in and the Lightbox’s UI stays put, there is not much point in having it at all, so no point extending it (“fixing”) for iOS...

common/app/views/fragments/header.scala.html
```diff
@@ -9,7 +9,7 @@
     @fragments.newHeader()
 } else {
     <header id="header"
-            class="l-header u-cf @if(page.metadata.hasSlimHeader) {l-header--is-slim l-header--no-navigation} js-header"
+            class="l-header u-cf @if(page.metadata.hasSlimHeader) {l-header--is-slim l-header--no-navigation} @if(mvt.ABGalleryRedesignVariant.isParticipating && page.metadata.contentType.toLowerCase == "gallery"){l-header--is-slim l-header--no-navigation l-header--new-gallery} js-header"
```
desbo: to avoid repetition of `mvt.ABGalleryRedesignVariant.isParticipating && page.metadata.contentType.toLowerCase == "gallery"`, you might want to move it to a method inside the variant

applications/app/views/fragments/newGalleryBody.scala.html
NataliaLKB: As a note, Olly has separated this out into `content--immersive` for shared styles and `content--immersive-article` only used for immersive articles

applications/app/views/fragments/newGalleryBody.scala.html
NataliaLKB: I removed a couple from `galleryHeader` :)

applications/app/views/fragments/newGalleryBody.scala.html
NataliaLKB: What are the instances? I only see the one in galleryBody, which this file will replace soon

static/src/stylesheets/module/content/_gallery.scss
NataliaLKB: Note: Olly has come up with a fix for this and is implementing it on a bigger PR as it will affect other things as well

common/app/views/fragments/header.scala.html
NataliaLKB: Changed this to `shouldShow` method which includes the check of whether you are on a gallery or not.

NataliaLKB:
@desbo @sndrs would one of you be able to give another look? I believe all the comments have been addressed.

applications/app/controllers/OptInController.scala
NataliaLKB: Note: This was discussed in person and the fastly change that makes this possible has already been made in a separate PR ready to go :)

applications/app/views/fragments/newGalleryBody.scala.html
sndrs: does this div still need to be a wrapper? could `gs-container` go on the div on line 26?

applications/app/views/fragments/newGalleryBody.scala.html
sndrs: not sure if this has changed or if i was in the wrong place, but i was referring to https://github.com/guardian/frontend/blob/2e00099b6509024fd5a9f04aa7daea03e08281ac/common/app/views/fragments/img.scala.html#L112, https://github.com/guardian/frontend/blob/2e00099b6509024fd5a9f04aa7daea03e08281ac/common/app/views/fragments/items/elements/facia_cards/image.scala.html#L6 and the stuff on around 78 here. it may be they need to be separate, in which case great. was a question really - do we want the picture choosing technique duplicated like this or abstracted into one 'correct' method

static/src/stylesheets/module/content/_gallery.scss
sndrs: is there any way round that? that's not a nice one either!

static/src/stylesheets/module/content/_gallery.scss
```diff
@@ -245,3 +245,322 @@
...
+        position: relative;
+        left: 0;
+    }
+
+    // TODO: Replace all media tones instead of overwriting the defaults
+    &.tonal--tone-media {
+
+        // Overwriting media tone header so that you can see the image instead
+        .tonal__header {
+            background-color: rgb(0, 0, 0);
```
sndrs: we don't have any other `rgb` colours in the codebase. for the sake of consistency, should they be `#000000`? they'll work fine so it's not a blocker.

applications/app/views/fragments/newGalleryBody.scala.html
NataliaLKB: That makes more sense! 

Yes Olly and I are planning on refactoring the scala img helper so we can use it. It didn't feel right to do it in this PR though because it is already big enough.

static/src/stylesheets/module/content/_gallery.scss
NataliaLKB: Yep, actually changing them to `$multimedia-support-5`

static/src/stylesheets/module/content/_gallery.scss
NataliaLKB: I have been playing around with it, but I don't think there is. The problem is more of a tooling problem than anything else. We don't get properly structured data back so the caption looks like this:

image

The first `br` is display none and the "title" (what is in the `strong` tags) is display block with the correct margin-bottom. 

If we make all `br`s display none, then in long captions the editors wont have the ability to make "paragraphs" or breaks between sentences.

If we leave it and take out the display block in the title, we don't get the correct spacing between the title and the rest of the caption.

The best thing to do is change the tooling to allow us to template this properly. I am happy to talk to the tools team about this, but it won't be for this PR.

applications/app/views/fragments/newGalleryBody.scala.html
NataliaLKB: It does need to wrap the `content__main-column`, but I have moved it to the one above `content__main`

sndrs:
👍 do it!

23c2931e00000578-0-image-m-33_1417770875109

johnduffell:
haha @sndrs

dominickendrick:
🚀 🎸 📷

PR #12577, opened 2016-04-15
Remove old job parser switch

lps88:
## What does this change?

remove switch and also small fix to fetch method


JonNorman:
Looks good! :shipit:

PR #12578, opened 2016-04-15
UK excluded from the loyal survey AB test

Calanthe:
## What does this change?

This PR excludes UK from the loyal survey AB test


uplne:
👍

PR #12580, opened 2016-04-15
CtaController: take open-cta switch into account

TBonnin:
## What does this change?

Do not make request to callToAction url if open-cta switch is off

The open-cta switch value used in javascript can be cached and therefore turning off the switch doesn't result in stopping immediately all calls to the callToAction url.
Checking the switch server side would ensure this request is not made even if the client has the old switch value cached

No
## Request for comment

@jamespamplin


jamespamplin:
:+1: looks good to me

PR #12582, opened 2016-04-15
Log better error when MostPopularAgent update requests fail

TBonnin:
## What does this change?

Better logging when the MostPopularAgents update requests fail

This is one of the most recurrent error on preview and training-preview. (see below)
Trying to find out what is the issue is

No
## Screenshots

screen shot 2016-04-15 at 17 18 49
## Request for comment

@jamespamplin

### DIFF
--- a/onward/app/feed/MostPopularAgent.scala
+++ b/onward/app/feed/MostPopularAgent.scala
@@ -5,8 +5,8 @@ import common._
 import services.OphanApi
 import play.api.libs.json.{JsArray, JsValue}
 import model.RelatedContentItem
-import scala.concurrent.duration._
 import scala.concurrent.Future
+import scala.util.control.NonFatal
 import ContentApiClient.getResponse
 
 object MostPopularAgent extends Logging with ExecutionContexts {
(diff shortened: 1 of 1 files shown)


onward/app/feed/MostPopularAgent.scala
```diff
@@ -58,9 +58,11 @@ object GeoMostPopularAgent extends Logging with ExecutionContexts {
       } yield {
         getResponse(ContentApiClient.item(urlToContentPath(url), Edition.defaultEdition))
           .map(_.content.map(RelatedContentItem(_)))
-          .fallbackTo{
-            log.error(s"Error requesting $url")
-            Future.successful(None)}
+          .recover {
+            case e: Throwable =>
```
jamespamplin: Generally would prefer matching on `scala.util.control.NonFatal` rather than all throwables. `NonFatal` still allows things like Out of Memory, and other fatal errors to fail.

TBonnin: Thanks. Done this in 37fc1d6

PR #12583, opened 2016-04-15
added SwitchGroup case class with optional description

JonNorman:
## What does this change?

Currently a `Switch` has a `group` which is a string - this PR turns `group` into a `SwitchGroup`, which has both a `name` and a `description`.

It enables us to add useful caveats/descriptions that might provide some useful context to the switches in a given group.

This morning, there was a fairly large bug with one of the AB tests, partly due to conflation between the switch expiry date and the AB test expiry date. Whilst it might seem obvious, the ability to add a description to a switch group could be generally useful and seemed like the best way of persisting this knowledge.

**I am happy to be shot down about this and if so will scrap this away to the dustbin of closed PRs**

Not really.
## Screenshots

image
## Request for comment

/@kenlim


rich-nguyen:
👍

PR #12585, opened 2016-04-15
Only send crossword solutions out to clients if CAPI says it's ok.

JustinPinner:
## Following on from https://github.com/guardian/frontend/pull/12550...

At some point, CAPI are going to start returning all crossword solutions regardless of whether their other data (`dateSolutionAvailable` and `solutionAvailable`) say yes or no to the availability of those solutions. And when they do, we'll be ready for 'em.
## What does this change?

We're going to prevent the solutions being delivered to the client based on the aforementioned CAPI fields, so this is just a relatively small shift of decision functionality away from the client-side and back to the server-side. In fact the change made to the client-side code by the linked PR can now revert to its former state.

Not that crossword users are likely to be a cheaty bunch, but if we ship the solutions out in the page data of prize puzzles before the closing date, well that just kind of makes a mockery of the concept, right?

No
## @sndrs @desbo @philmcmahon

### DIFF
--- a/common/app/model/CrosswordData.scala
+++ b/common/app/model/CrosswordData.scala
@@ -81,7 +81,12 @@ object CrosswordData {
     // group are incorrectly stored on the first group entry. We normalize
     // the data to store the separator locations on their corresponding entries.
 
-    val entries = crossword.entries.map(Entry.fromCrosswordEntry)
+    val shipSolutions = crossword.dateSolutionAvailable.map(_.toJodaDateTime.isBeforeNow).getOrElse(crossword.solutionAvailable)
+
+    val entries = crossword.entries.collect {
+      case entry if shipSolutions || entry.solution.isEmpty => Entry.fromCrosswordEntry(entry)
+      case entry => Entry.fromCrosswordEntry(entry.copy(solution = None))
+    }
 
     val entryGroups = entries
       .groupBy(_.group)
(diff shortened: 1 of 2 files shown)


common/app/model/CrosswordData.scala
```diff
@@ -81,7 +81,12 @@ object CrosswordData {
     // group are incorrectly stored on the first group entry. We normalize
     // the data to store the separator locations on their corresponding entries.
 
-    val entries = crossword.entries.map(Entry.fromCrosswordEntry)
+    val shipSolutions = crossword.dateSolutionAvailable.map(_.toJodaDateTime.isBeforeNow).getOrElse(crossword.solutionAvailable)
+
+    val entries = crossword.entries.collect {
+      case entry if shipSolutions || entry.solution.isEmpty => Entry.fromCrosswordEntry(entry)
+      case entry => Entry.fromCrosswordEntry(entry.copy(solution = None))
```
jamespamplin: This may be slightly easier to read if you invert the check:

```scala
val entries = crossword.entries.collect {
  case entry if !shipSolutions => Entry.fromCrosswordEntry(entry.copy(solution = None))
  case entry => Entry.fromCrosswordEntry(entry)
}
```

jamespamplin:
:+1: small suggestion, but otherwise looks good to me

PR #12586, opened 2016-04-18
Fabric advert template - first commit

jbreckmckye:
This is the first pass of a template for the all new **fabric adverts** that The Guardian will soon be showcasing at AdWeek. 

Fabric is a 250 high, fluid width ad that can be served without need for a responsive takeover in DFP. Ultimately, it's part of a collection ('fabric', 'focus' and 'frame') that underpin the Guardian's resolve for **'fewer, better ads'**.

Fabric will do the following things:
- It will always be 250px high. This will let us move ads over to a consistent slot height, and reduce the page yank that comes when we can't reserve the right ad height on initial page render.
- It will be flush width in single-column layouts
- It'll be expandable to 500px high when embedded in mobile fronts
- It will support 'fixed' style backgrounds
- It will support 'parallax' style background scrolling
- It will appear after the first container on mobile fronts

This pull request implements these primary requirements:
1. It gives us a switch for requesting a new type of adverts (88x71 size)
2. It implements a model and view for the new ad type
3. It serves this new advert to inlines and top slot ads
4. It implements the flush appearance in single-column (i.e. mobile) layouts, and a full-width layout in the top slot
5. It implements a desktop-only fixed background (I'm investigating approaches to making this work on mobile)
6. It makes top-slot fabric ads play nicely with sticky banners
7. It lets me expand the advert when it's in a mobile front

Here are some screenshots:

_Mobile front advert with a fixed background:_
image

_The selfsame advert expanded:_
image

_As a top slot:_
image

_Inside a desktop container:_
image


static/src/stylesheets/module/commercial/_creatives.scss
```diff
@@ -730,6 +730,55 @@
...
+    background: #303030;
+}
+
+.fabric--expanded {
+    height: 500px;
+}
+
+// Label
+
+.fabric__label-wrapper {
```
uplne: Do we need it's own label? Advert label is the same across the page for display ads. Maybe instead of having `ad-slot__label` and `fabric__label` we should create new object `ad-slot-label` and then have `ad-slot-label--full or whatever` for fabric and `ad-slot-label__content or __mark` as an element for the filling.

uplne:
Do you think it is worth it to have any unit tests?

static/src/stylesheets/module/commercial/_creatives.scss
jbreckmckye: I guess we could, but is this worth refactoring before we possibly deprecate the old templates anyway?

jbreckmckye:
We could do, but right now I'm not sure if the abstractions here are even
going to survive.

On 18 April 2016 at 14:44, Steve Vadocz <notifications@github.com> wrote:

> Do you think it is worth it to have any unit tests?
>
> —
> You are receiving this because you authored the thread.
> <https://github.com/guardian/frontend/pull/12586#issuecomment-211385239>
>



-- 

Jimmy Breck-McKye

Client-side developer

Guardian News & Media

-----

jimmy.breck-mckye@theguardian.com <alex.breuer@theguardian.com>

-----

@jbreckmckye <https://twitter.com/jbreckmckye>

-----

Kings Place, 90 York Way,

London N1 9GU

theguardian.com

​-----

​​Download the Guardian app for Android
<https://play.google.com/store/apps/details?id=com.guardian&hl=en_GB> and
iOS​ <https://itunes.apple.com/gb/app/the-guardian/id409128287?mt=8>

-- 

in England Number 908396

static/src/stylesheets/module/commercial/_creatives.scss
uplne: I don't know but that would be the proper way how to do it ;-)

uplne:
I will leave the comments on your judgement. Other than that it is beautiful 👍

common/app/views/support/Commercial.scala
```diff
@@ -34,9 +34,10 @@ object Commercial {
     }
 
     def adSizes(metaData: MetaData, edition: Edition): Map[String, Seq[String]] = {
+      val fabricAdverts = if (FabricAdverts.isSwitchedOn) Some("88,71") else None
```
rich-nguyen: I actually prefer this, because it's symbolic, rather than the strings below, like "88,70". Not sure why we don't use a pre-defined key value to do the targeting, @kenlim ?

rich-nguyen:
👍 it looks this isn't fully production-ready, but makes sense given that you will continue to iterate this on code.

As discussed, a configurable solution to parallax backgrounds should also give you an elegant way to do fixed backgrounds aswell.

jbreckmckye:
@uplne I do, but I'm not sure if the time is right - it's still not clear how background effects will work, and if this will have any impact on expansion and templating. So I'm a little hesitant to write unit tests right now.

static/src/stylesheets/module/commercial/_creatives.scss
```diff
@@ -730,6 +730,55 @@
...
+        // Full width on mobile fronts
+        width: auto;
+    }
+}
+
+.fabric {
+    height: 250px;
+    transition: height .8s;
+
+    // This is just for demonstration purposes, whilst the ad is populated with ipsum text
```
Calanthe: To be honest I don't like how the white & blue font looks like on that colourful background. Is this a version that we are going to demonstrate during the ad week?

Calanthe:
Can you also include here an animated pic with the animated expansion of the ad?

static/src/stylesheets/module/commercial/_creatives.scss
jbreckmckye: No, we're not.

jbreckmckye:
@Calanthe - you can see it in the flesh at http://m.code.dev-theguardian.com/world?adtest=jbmk - view it under a mobile screen size and click on the inline ad.

Calanthe:
@jbreckmckye I cannot see nay inline ads, only Glabs card at the bottom.

static/src/javascripts/projects/common/views/commercial/creatives/fabric.html
```diff
@@ -0,0 +1,17 @@
+<div
+    class="fabric"
+    style="
+        background-image: url(<%=backgroundImage%>);
+        background-color: <%=backgroundColour%>;
+        background-repeat: <%=backgroundRepeat%>;
+        background-position: <%=backgroundPosition%>;
+        background-attachment: <%=(backgroundScroll === 'reveal' ? 'fixed' : 'scroll')%>;
```
regiskuckaertz: Beware, I've had bad experiences with expressions in lodash template once they've been included in the optimized build. The `text` plugin is escaping some characters where it shouldn't... you might want to offset that computation in your module.

jbreckmckye: What characters are problematic, and how does it choke on them?

regiskuckaertz: TBH I'm not sure but I'd say the quotes are dangerous. They are probably escaped in a way that does not bode well with `template`. It will convert this into a string that is then fed into a `Function` constructor, meta-programming ftw.

regiskuckaertz:
Awesome work! I think we should abstract ad sizes in a Map in a future PR (both in JS and Scala), @rich-nguyen is right numbers are very cryptic.

static/src/javascripts/projects/common/views/commercial/creatives/fabric.html
jbreckmckye: So, we've seen this issue before?

Possibly I will end up moving this logic into some kind of proper viewmodel anyway.

regiskuckaertz: I saw it when I tried to precompile lodash templates, and yet again when we worked on gimbap with @uplne. I think it makes better sense to treat lodash templates as dumb vessels for values.

PR #12587, opened 2016-04-18
A switch that has been badly configured should still be readable

rich-nguyen:
Discovered this one in dev, but it could happen in prod too.

If you rely on a parsed switch value, using something like `isGuaranteedSwitchedOn`, then a badly configured switch value will never be considered initialized, and you get a future failed, initialization timeout.

This tweaks the behaviour so that if the switch file has been read at least once, the switch is initialised, even when there is a problem parsing/finding the switch. The value will be safe state, as it was before the switch file was read, but future will be completed.

### DIFF
--- a/common/app/conf/GlobalSettings.scala
+++ b/common/app/conf/GlobalSettings.scala
@@ -55,17 +55,20 @@ trait SwitchboardLifecycle extends GlobalSettings with ExecutionContexts with Lo
     super.onStop(app)
   }
 
-  def refresh() {
+  def refresh(): Unit = {
     log.info("Refreshing switches")
-    services.S3.get(Configuration.switches.key) map { response =>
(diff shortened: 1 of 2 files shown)


JonNorman:
:+1:

Is it worth actively checking features that depend on `isGuaranteedSwitchedOn` whilst this is deployed? Just thinking that there may be some dormant code that has been timing out that may now be activated, though this should only happen for switches with a `safeState` of `on`.

rich-nguyen:
I think we're ok in prod. The most likely scenario to cause the issue is when a switch name does not appear in the switch file, and that switch value is being resolved through a future. For dev, that could happen a lot, because there may not be an admin server targeting your switches file. In prod, the issue could arise when switches change name, or a new switch is introduced. So we should be fine.

PR #12588, opened 2016-04-18
Commercial Jobs switch names

rich-nguyen:
Renamed the jobs feed switches to match travel's naming. Removed `jobsUrlTemplate` so that we fully rely on the static url system. @lps88

### DIFF
--- a/common/app/common/configuration.scala
+++ b/common/app/common/configuration.scala
@@ -308,7 +308,6 @@ class GuardianConfiguration(val application: String, val webappConfDirectory: St
     lazy val masterclassesToken = configuration.getStringProperty("masterclasses.token")
     lazy val liveEventsToken = configuration.getStringProperty("live-events.token")
     lazy val liveEventsImagesUrl = "https://membership.theguardian.com/events.json"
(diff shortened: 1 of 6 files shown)


common/app/common/configuration.scala
```diff
@@ -308,7 +308,6 @@ class GuardianConfiguration(val application: String, val webappConfDirectory: St
     lazy val masterclassesToken = configuration.getStringProperty("masterclasses.token")
     lazy val liveEventsToken = configuration.getStringProperty("live-events.token")
     lazy val liveEventsImagesUrl = "https://membership.theguardian.com/events.json"
-    lazy val jobsUrlTemplate = configuration.getStringProperty("jobs.api.url.template")
     lazy val jobsStaticUrl= configuration.getStringProperty("jobs.api.url")
```
kelvin-chappell: A small point but could this just be `jobsUrl`?
The fact that it's static doesn't seem to make any difference to how it's used.

kelvin-chappell:
👍

common/app/common/configuration.scala
rich-nguyen: yep, it doesn't need to be static anymore.

PR #12589, opened 2016-04-18
Logstash Logging: only add kinesis appender to root logger

TBonnin:
## What does this change?

Previous implementation would append a new kinesis appender every time a new logger would be initialized.
This fix ensure the kinesis appender is appended only once to the root logger when the app starts

No more wasted resources.

No
## Request for comment

@jamespamplin @johnduffell


jamespamplin:
:+1: looks good to me

johnduffell:
good work! 😰 👍 🚤 😝

PR #12584, opened 2016-04-15
Video Container!

sammorrisdesign-zz:
## What does this change?

What doesn't this change?!? I was going to split this up into separate PRs but 99% of it related. Here's a full list of changes...
- `frontend` now recognises the new container type of `fixed/video`
- It will inject a completely custom container of `videoContainer.scala.html` (in a similar way to nav list or commercial containers do at the moment). This container is a horizontal scrollable list on mobile and a carousel on desktop. It only plays inline.
- Adds a new image format of `Video700`
- Adds a new `formattedDuration` fragment/helper that will format the duration of a video in a readable format
- Adds a chevron that points to the left.
## Does this affect other platforms - Amp, Apps, etc?

It will render as a regular container on other platforms as it doesn't know what a `fixed/video` container is
## Screenshots

screen shot 2016-04-15 at 17 46 43

videocontainer
## Request for comment

@rich-nguyen @janua @jamesgorrie 
## 

_Does this PR meet the [contributing guidelines](https://github.com/guardian/frontend/blob/issue_pr_templates/.github/CONTRIBUTING.md#submission)?_

PR #12591, opened 2016-04-18
Add video overlay option to video player

sammorrisdesign-zz:
## What does this change?

Adds an optional field to the `Video` template to add a video overlay to the player. This adds a headline, tag if needed and video duration. At the moment it's not visible to users. But it needs to go in as it will be used in my [Video Container PR](https://github.com/guardian/frontend/pull/12584/) and some stuff by @blongden73.
## Request for comment

@rich-nguyen @jamesgorrie 
## 

### DIFF
--- a/common/app/model/Asset.scala
+++ b/common/app/model/Asset.scala
@@ -104,11 +104,10 @@ case class VideoAsset(
     }
   }
 
-  // The video duration in seconds
-  val duration: Int = fields.get("durationSeconds").getOrElse("0").toInt +
-                           (fields.get("durationMinutes").getOrElse("0").toInt * 60)
   val durationSeconds: Int = fields.get("durationSeconds").getOrElse("0").toInt
   val durationMinutes: Int = fields.get("durationMinutes").getOrElse("0").toInt
+  // The video duration in seconds
+  val duration: Int = durationSeconds + (durationMinutes * 60)
   val blockVideoAds: Boolean = fields.get("blockAds").exists(_.toBoolean)
   val source: Option[String] = fields.get("source")
   val embeddable: Boolean = fields.get("embeddable").exists(_.toBoolean)
--- a/common/app/model/Asset.scala
+++ b/common/app/model/Asset.scala
@@ -107,8 +107,9 @@ case class VideoAsset(
   // The video duration in seconds
   val duration: Int = fields.get("durationSeconds").getOrElse("0").toInt +
                            (fields.get("durationMinutes").getOrElse("0").toInt * 60)
+  val durationSeconds: Int = fields.get("durationSeconds").getOrElse("0").toInt
+  val durationMinutes: Int = fields.get("durationMinutes").getOrElse("0").toInt
   val blockVideoAds: Boolean = fields.get("blockAds").exists(_.toBoolean)
-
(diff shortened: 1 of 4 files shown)


common/app/model/Element.scala
```diff
@@ -94,6 +94,11 @@ final case class VideoMedia(videoAssets: List[VideoAsset]) {
     }.flatten.sorted
   }
   val duration: Int = videoAssets.headOption.map(_.duration).getOrElse(0)
+  val formattedDuration: String = {
+    videoAssets.headOption.map(_.durationMinutes).getOrElse(0).toString() + ":" + "%02d".format(videoAssets.headOption.map(_.durationSeconds).getOrElse(0))
+  }
+  val durationMinutes: Int = videoAssets.headOption.map(_.durationMinutes).getOrElse(0)
```
jamesgorrie: Should these not be options, as, if the time isn't there, it should be `None` not 0?

common/app/model/Asset.scala
```diff
@@ -107,8 +107,9 @@ case class VideoAsset(
   // The video duration in seconds
   val duration: Int = fields.get("durationSeconds").getOrElse("0").toInt +
                            (fields.get("durationMinutes").getOrElse("0").toInt * 60)
+  val durationSeconds: Int = fields.get("durationSeconds").getOrElse("0").toInt
```
jamesgorrie: On [this point](https://github.com/guardian/frontend/pull/12591/files#r60062191) - Could we not have:
`val durationSeconds: Option[Int] = fields.get("durationSeconds").map(_.toInt)`

common/app/views/fragments/media/video.scala.html
```diff
@@ -67,6 +67,21 @@
                     </object>
                 }
             </video>
+            @if(showOverlay){
+            <div class="video-overlay">
```
jamesgorrie: Is this already styled, or is that done in the other PR?

sammorrisdesign: That's done in a different PR

common/app/model/Asset.scala
sammorrisdesign: I think we would still want to display 0:32, if there is no `durationMinutes`

jamesgorrie: We receive `durationMinutes: "0"` from the API - the reason for the change above is to not have the duration set to 0:00 when it isn't a media element.

jamesgorrie: Sorry, I am being a fool, I didn't realise these were video specific.

You could make this slightly more readable by:
```
val durationSeconds: Int = fields.get("durationSeconds").map(_.toInt).getOrElse(0)
val durationMinutes: Int = fields.get("durationMinutes").map(_.toInt).getOrElse(0)
val duration: Int = durationSeconds + (durationMinutes * 60)
```

Saying that - why do we need `duration` in seconds?

jamesgorrie:
Just a bit around code clarity, but that's just opinion. 👍

common/app/model/Asset.scala
sammorrisdesign: Ha no worries. That's a good point on using the new vals for legibility.

I believe videojs needs the duration in seconds to do its magic

PR #12593, opened 2016-04-18
Use isGuaranteedSwitchOn to get value of Logstash-logging switch

TBonnin:
## What does this change?

Since the appender configuration is happening very early on in the lifecycle of the app, the switchboard was not read yet and always returned the safe state of the logstash-logging switch (which is off)

This PR makes sure the logstash-logging switch value has been read before returning its state.

Make logging to logstash possible :)

No
## Request for comment

@johnduffell @jamespamplin

### DIFF
--- a/common/app/common/Logback/LogbackConfig.scala
+++ b/common/app/common/Logback/LogbackConfig.scala
@@ -53,22 +53,22 @@ object LogbackConfig {
       try {
         val rootLogger = LoggerFactory.getLogger(SLFLogger.ROOT_LOGGER_NAME)
         asLogBack(rootLogger).map { lb =>
-            lb.info("Configuring Logback")
-            val context = lb.getLoggerContext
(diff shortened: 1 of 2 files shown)


common/app/common/Logback/LogbackConfig.scala
```diff
@@ -53,22 +53,22 @@ object LogbackConfig {
...
+              config.region,
+              config.awsCredentialsProvider,
+              bufferSize
             )
-            lb.addAppender(appender)
-            lb.info("Configured Logback")
+          )
+          lb.addAppender(appender)
+          lb.info("Configured Logback")
         } getOrElse(PlayLogger.info("not running using logback"))
```
TBonnin: There is not change here, only an indentation fix :)

jamespamplin:
:+1: looks good to me

johnduffell:
👍 🎉

PR #12595, opened 2016-04-18
when picking a creative example, make sure that the preview URL is po…

JonNorman:
We have a [Commercial test page](https://frontend.gutools.co.uk/analytics/commercial/templates) that provides a list of all the DFP creative templates along with their currently running creatives.

Each `Template` model that is represented on this page currently picks the `previewUrl` from the first creative it has - regardless of whether said creative has a populated `previewUrl`. This PR alters this behaviour to always pick a populated `previewUrl` from amongst the available creatives.

/@kelvin-chappell

### DIFF
--- a/common/app/common/dfp/DfpData.scala
+++ b/common/app/common/dfp/DfpData.scala
@@ -1,7 +1,5 @@
 package common.dfp
 
-import java.net.URLEncoder
-
 import common.Edition
 import common.dfp.AdSize.{leaderboardSize, responsiveSize}
 import org.joda.time.DateTime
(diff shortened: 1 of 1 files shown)


common/app/common/dfp/DfpData.scala
```diff
@@ -457,7 +456,7 @@ case class GuCreativeTemplate(id: Long,
                               snippet: String,
                               creatives: Seq[GuCreative]) {
 
-  val examplePreviewUrl: Option[String] = creatives.headOption flatMap { creative => creative.previewUrl }
+  lazy val examplePreviewUrl: Option[String] = creatives flatMap {_.previewUrl} headOption
```
kelvin-chappell: Would be more efficient to find first creative with a preview URL, but this is more readable and given the small number of creatives this is better.

kelvin-chappell:
👍

common/app/common/dfp/DfpData.scala
JonNorman: Yeah - using `find` would be quicker but it looked a bit clunky and I went for aesthetics!

PR #12596, opened 2016-04-18
Move back image to html element instead of css background

uplne:
I was testing this approach while working on the gimbap richmedia and totally deployed it by an accident.

This approach doesn't work as scaling effect on background image isn't very smooth. So back to img.

@jbreckmckye @regiskuckaertz @Calanthe


regiskuckaertz:
👍

PR #12597, opened 2016-04-19
Fix commercial test page not working in IE9

regiskuckaertz:
cc @uplne


uplne:
Beautiful. I almost wish we work in the same team 👍

PR #12598, opened 2016-04-19
Add proxy pass and referer for idapi

katebee:
## What does this change?

Adds a simple proxy server to nginx
The proxy server will proxy the referer so that access is no longer denied when attempting to perform identity based actions.

Permits users to use the identity and membership features on localhost (login, saved for later, etc.)
## Request for comment

@jamespamplin


NataliaLKB:
👍 great!

jamespamplin:
:+1: great stuff

PR #12599, opened 2016-04-19
Adding the liveEvents FeedFetcher job back into the feeds to fetch

JonNorman:
It looks  like the `liveEvents` feedFetch job was inadvertently merged out of the `all` list in `FeedFetcher` between the commit 7391e91 and ea0b951.

Just adding it back in.


rich-nguyen:
👍

PR #12601, opened 2016-04-19
Fix ab test switch indicator

dominickendrick:
## What does this change?

This fixes the ab test switch indicator
Previously it was showing that all tests were turned off. This is not true!
## Screenshots

**before** :sad_face:
screen shot 2016-04-19 at 12 31 12

**after** :happy_face:
screen shot 2016-04-19 at 12 30 37

### DIFF
--- a/admin/app/views/abtests.scala.html
+++ b/admin/app/views/abtests.scala.html
@@ -39,7 +39,7 @@
 
     <script type="text/javascript">
         window.abSwitches = {
-            @conf.switches.Switches.all.filter(_.group.name == "A/B Tests").map{ s =>
+            @conf.switches.Switches.all.filter(_.group == conf.switches.SwitchGroup.ABTests).map{ s =>
                 "@CamelCase.fromHyphenated(s.name)": @s.isSwitchedOn,
             }
         };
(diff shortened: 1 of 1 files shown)


admin/app/views/abtests.scala.html
```diff
@@ -39,7 +39,7 @@ <h3 class="abtests-expired-title">Expired tests <a href="#">show</a></h3>
 
     <script type="text/javascript">
         window.abSwitches = {
-            @conf.switches.Switches.all.filter(_.group == "A/B Tests").map{ s =>
+            @conf.switches.Switches.all.filter(_.group.name == "A/B Tests").map{ s =>
                 "@CamelCase.fromHyphenated(s.name)": @s.isSwitchedOn,
```
JonNorman: You could replace `group.name == "A/B Tests"` with `group == SwitchGroup.ABTests`

That way, if the name changes (or the group is deleted) it should flag as a compilation error.

rich-nguyen:
👍

admin/app/views/abtests.scala.html
dominickendrick: nice !

PR #12602, opened 2016-04-19
Lo and behold: the super sticky top slot

regiskuckaertz:
Some big campaign will be running for 1 year under the personal-investments tag and it has been decided that the top slot needs to be sticky all the way through in the corresponding pages.

cc @kenlim @JonNorman @uplne

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/sticky-top-banner.js
+++ b/static/src/javascripts/projects/common/modules/commercial/sticky-top-banner.js
@@ -6,7 +6,8 @@ define([
     'common/utils/mediator',
     'common/utils/config',
     'common/utils/detect',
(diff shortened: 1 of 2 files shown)


JonNorman:
AFAIK the `keywordIds` is a sequence of comma separated `<section>/<keyword>` strings. Do we only want to make the ad super-sticky when matching against the `section` tag or should it be super sticky on any personal-investment content that may be outside the personal-investment section - i.e. against the `keyword`? I think the regex currently only matches on `section`; if that is how we want it then I'll shut right up and give it a big :+1:

regiskuckaertz:
Good catch! That's indeed a keyword tag, I updated the regex to match that.

PR #9016, opened 2015-04-27
Sync local date to server, for relative dates

stephanfowler:
Use the page's render time as the basis of "now", for the purpose of relative time calculations. Because if a user's device has its local time/zone incorrectly set, our relative dates can be wildly out.

PR #12603, opened 2016-04-19
Serve appropriate ETag for facia pages

johnduffell:
This makes it so we can serve an etag based on the content body of facia pages.

Based on looking at the incoming traffic after a couple of hours, I can estimate how many requests we don't need to serve with a full response because the content has not changed.

If this is a decent number, I can change it to serve 304s instead of 200s.  I will also keep a note whether doing the hashing has any negative impact on CPU on the instances.

The [adjusted time](https://github.com/guardian/frontend/pull/9016) on the client side stuff isnt' really compatible with caching, so I've removed it

@JustinPinner @TBonnin @gklopper any comments?

### DIFF
--- a/facia/app/controllers/FaciaController.scala
+++ b/facia/app/controllers/FaciaController.scala
@@ -3,6 +3,7 @@ package controllers
 import common._
 import controllers.front._
 import layout.{CollectionEssentials, FaciaContainer, Front}
+import model.Cached.StringResult
 import model._
 import model.facia.PressedCollection
 import model.pressed.CollectionConfig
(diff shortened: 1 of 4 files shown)


facia/app/controllers/FaciaController.scala
```diff
@@ -148,13 +149,17 @@ trait FaciaController extends Controller with Logging with ExecutionContexts wit
...
+          Cached.withEtag(faciaPage) {
+            if (request.isRss) {
+              val body = TrailsToRss.fromPressedPage(faciaPage)
+              StringResult(Ok(body).as("text/xml; charset=utf-8"), body)
+            }
             else if (request.isJson)
-              JsonFront(faciaPage)
-            else
-              Ok(views.html.front(faciaPage))
+              StringResult.withoutString(JsonFront(faciaPage))
```
JustinPinner: `StringResult.withoutString` seems confusing. Could we call it something more intuitive?

piuccio:
facia is serving pressed files right? Maybe instead of doing the hashing yourself you can use the etag of pressed.json

JustinPinner:
I think this is ok, especially given its current experimental nature. 👍

Issue #12059, opened 2016-02-23
Crossword Fronts - SVG misalignment issue


ScottPainterGNM:
# ISSUE
Looking at the crossword SVGs used on the browse page https://www.theguardian.com/crosswords . The numbers are showing, and as a result of the site's font styles, they're overlapping their boxes.

## Steps to Reproduce
Go to any crossword browse page such as https://www.theguardian.com/crosswords

## Actual Results (include screenshots)
The numbers are overlapping the boxes
![picture 30](https://cloud.githubusercontent.com/assets/11922253/13250053/b1c14110-da1f-11e5-8ae8-b9843013f732.png)

## Expected Results (include screenshots)
Either these should be aligned or just removed if easier (i.e. just show the grid not the clue numbers at all)

## URL
https://www.theguardian.com/crosswords
## OS and Browser details

- Device: Mac OSX
- Browser (inc version): Firefox 43, Chrome 48.0.2564
- OS: El Capitan

*Please add relevant labels*
Done

paperboyo:
Good spot! Also, even though the text is styled as `Guardian Text Sans Web`, for some reason, it comes out as whatever's default `serif`!!111!ONE!! Most probably our fonts do not work when svg is embedded as an `<img>`, but I wouldn't know.

katebee:
Relevant stack-overflow: http://stackoverflow.com/questions/20577316/svg-font-face-works-in-svg-but-not-when-included-in-a-page

`.crossword__cell-number` is the class responsible for styling the text on the individual crossword page (e.g. https://www.theguardian.com/crosswords/quick/14335) fonts cannot be loaded by a svg inside an img tag.

Swapping the img tag for a object tag appears to fix this :smile:

**Any immediate caveats with swapping the crossword img tag for an object one? Otherwise I will go ahead and make a pull request** 

<img width="644" alt="picture 4" src="https://cloud.githubusercontent.com/assets/8607683/14643214/68eca43c-0645-11e6-8384-c2dc97222861.png">

katebee:
PR #12604 should fix this

PR #12604, opened 2016-04-19
Fix the font displayed in the crossword preview, issue #12059

katebee:
## What does this change?

Fonts cannot be loaded by a svg inside an img tag, however, placing the svg inside an object tag will allow scriptable content; the cell numbers can then have the correct, sans-serif font displayed.

Aesthetic improvement to the crossword Front.
#### BEFORE

crossword-before
#### AFTER

crossword-after

additional screenshots: https://www.browserstack.com/screenshots/915efaa38c33d9e5f5b9639dea9c1c466068cfd6
## Request for comment

@sndrs


paperboyo:
Cool! I vaguely remember there being some browser differences in font rendering within different flavours of SVG... but can't remember for the love of...

sndrs:
👍 great

PR #12605, opened 2016-04-19
Logs error and request headers when GET /discussion/p/XXXX fails

TBonnin:
## What does this change?

Logs error and request headers when GET /discussion/p/XXXX fails

Understand where those requests come from

No
## Request for comment

@OliverJAsh


OliverJAsh:
👍

PR #12607, opened 2016-04-20
Add superSticky flag to JavascriptConfig

regiskuckaertz:
The stickiness of the top banner is a little problematic, in that it is only enabled under [a combination of circumstances](https://github.com/guardian/frontend/compare/rkz-supersticky-topslot?expand=1#diff-4edda315e7471b54637cf5b743a6b4eeR137).

For the personal investments campaign, some of these flags need to be overridden (i.e. the type of content and if the page is an adfeature), which is why I had no other choice but to add another flag in the global page config.

Admittedly it's not the stickiness that is problematic, rather the campaign, but: 💰💰💰💰

cc @rich-nguyen @kelvin-chappell @JonNorman


common/app/model/Section.scala
```diff
@@ -15,9 +16,16 @@ object Section {
...
+    // The PI campaign will run for one year, during which all the related pages must provide a sticky
+    // banner at the top that sticks all the way through, therefore overriding other config flags
+    // such as isAdvertisementFeature
+    lazy val isPersonalInvestmentsCampaign: Boolean = keywordIds.exists(t => t.endsWith("/personal-investments"))
+    lazy val isPersonalInvestmentsCampaignRunning: Boolean = DateTime.now().isBefore(new DateTime(2017, 4, 26, 0, 0))
+
     val javascriptConfigOverrides: Map[String, JsValue] = Map(
         "keywords" -> JsString(webTitle),
         "keywordIds" -> JsString(keywordIds.mkString(",")),
+        "hasSuperStickyBanner" -> JsBoolean(isPersonalInvestmentsCampaign && isPersonalInvestmentsCampaignRunning),
```
kelvin-chappell: Perhaps this condition could be encapsulated in a `PersonalInvestmentsCampaign` object that `Section` and `PressedPage` both use, to avoid repetition.

regiskuckaertz: Yep, doing that now.

rich-nguyen:
👍  looks good

PR #12610, opened 2016-04-20
Don't send HTTP body if clients already have the content

johnduffell:
This changes facia so that if the client has the content already (as illustrated by the if-none-match header) the server will respond with 304 not modified and an empty body.

This will save bandwidth - current stats show that 25% (878 out of 3528) of requests for fronts (rss and html are supported) during a sample period (1/6th of the traffic between 9:35 and 9:54 this morning) were when the client had the correct content.  A further 123 of the requests, the client had no content to revalidate.  So this is a significant saving on bandwidth.

@JustinPinner I've also tidied up that weird code I wrote last time!  I hope it is more logical now.
@piuccio time to save some bandwidth ;)

### DIFF
--- a/common/app/model/Cached.scala
+++ b/common/app/model/Cached.scala
@@ -1,9 +1,10 @@
 package model
 
+import conf.switches.Switches
 import conf.switches.Switches._
 import org.joda.time.DateTime
 import org.scala_tools.time.Imports._
-import play.api.mvc.{Request, Action, Result}
+import play.api.mvc._
 import scala.concurrent.Future
 import scala.concurrent.duration.Duration
 import scala.concurrent.ExecutionContext.Implicits.global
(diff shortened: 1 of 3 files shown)


cb372:
Awesome! Although I'm a bit surprised that we haven't been using ETag across the board until now.

By the way, couldn't we get nginx to do both the ETag header calculation and the If-None-Match check in the router? That would mean less code for us to maintain.

johnduffell:
@cb372 thanks for the comment - we could if this is as far as we'd like to go, but if it turns out to save a lot of requests, I'd like to try to save the cpu as well by doing the check before we render the html.  In fact, the ideal solution would be to pass the etag on to CAPI/S3 and get them to only serve their body if it's changed.

common/app/model/Cached.scala
```diff
@@ -41,33 +42,32 @@ object Cached extends implicits.Dates {
...
+      }
+      new RevalidatableResult(result, Hash(hashLong.toString))
+    }
   }
 
-  def withEtag(page: Page)(stringResult: StringResult): Result = {
+  def withRevalidation(page: Page)(stringResult: RevalidatableResult)(implicit request: RequestHeader): Result = {
     val cacheSeconds = page.metadata.cacheTime.cacheSeconds
-    if (cacheableStatusCodes.contains(stringResult.result.header.status)) cacheHeaders(cacheSeconds, stringResult.result, stringResult.body) else stringResult.result
+    if (cacheableStatusCodes.contains(stringResult.result.header.status)) cacheHeaders(cacheSeconds, stringResult.result, Some((stringResult.hash, request.headers.get("If-None-Match")))) else stringResult.result
```
JustinPinner: This is a very long line. Maybe consider breaking it into a more traditional if-else format?

JustinPinner:
It's still pretty tricksy for my little brain to absorb, but I think I get it. 👍

johnduffell:
Sum of network out per minute for the facia instances
image
can you guess when I turned it on? 😝 

In case it's not clear, we went from 275MB a minute to 200MB a minute, i.e. saving over a third of network traffic and therefore saving a grand total of $10 a day on AWS bandwidth 😞 that's $3650 a year just for facia pages😆

piuccio:
do we get a share of that 3k?

jamespamplin:
Wow, great stuff @johnduffell! Lets get etags on our other content too :)

johnduffell:
@piuccio of course if we were soft purging using the kinesis stream you added, that would have saved some of it, so the saving would be smaller 😛

PR #12608, opened 2016-04-20
Refactor travel component

regiskuckaertz:
Last one in line! With a few display bugs fixed too.

picture 1

/cc @kelvin-chappell @JonNorman @uplne

PR #12612, opened 2016-04-20
Format TravelOffer price

JonNorman:
This is a slight extension to #12608, which formats the price a little more clearly.

**Before:**
image

**After:**
image

(also removed a `toInt` call given that the attribute is now a `String`)

/@regiskuckaertz


regiskuckaertz:
Looks good to me 👍

Issue #11241, opened 2015-12-01
Selected text -> Tweet intent is 141 characters


adamamyl:
## Steps to reproduce

 - Open a Guardian article
 - [Copy some text](http://amyl.org.uk/gutweets-bug/gufrontend-bugreport-tweet-too-long-guardian.png) (screenshot)
 - Hit the Tweet icon
 - [Try and tweet](http://amyl.org.uk/gutweets-bug/gufrontend-bugreport-tweet-too-long-twitter.png) (screenshot)
 - Notice -1 characters
 - [Tweet](https://twitter.com/adamamyl/status/671351527638966272) about it, and raise a bug report.

## Environment

Device: Macbook
Browser: Chrome Version 46.0.2490.86 (64-bit) and Firefox 35.0.1
OS: MacOS, El Cap
URL: Various I've tried, but in this instance http://www.theguardian.com/world/2015/dec/01/thai-printer-international-new-york-times-blank-space

PR #12613, opened 2016-04-20
Fix twitter selection sharing #11241 #12615

katebee:
## What does this change?
- Sharing highlighted text via twitter no longer slices a section that is 141 characters, it now only takes the 140 required for a valid tweet.
- Sharing text containing semicolons no longer truncates the text; encodeURIComponent will convert the semicolon to UTF-8 (encodeURI does not cover all special characters).

e6da9c9 fixes issue #11241 (too many characters being passed to a tweet)
1deb267 fixes issue #12615 (semicolons in selected text cause truncation of tweet)
## Screenshots

**141 char issue BEFORE:**
tweetbefore

**141 char issue AFTER:**
tweetafter

**Semicolon issue BEFORE:**
semicolons missing

**Semicolon issue AFTER:**

semicolons present
## Request for comment

@gtrufitt

### DIFF
--- a/static/src/javascripts/projects/common/modules/ui/selection-sharing.js
+++ b/static/src/javascripts/projects/common/modules/ui/selection-sharing.js
@@ -40,7 +40,7 @@ define([
         $emailAction,
         twitterShortUrl = config.page.shortUrl + '/stw',
         twitterHrefTemplate = 'https://twitter.com/intent/tweet?text=%E2%80%9C<%=text%>%E2%80%9D&url=<%=url%>',
-        twitterMessageLimit = 115, // 140 - t.co length - 3 chars for quotes and url spacing
+        twitterMessageLimit = 114, // 140 - t.co length - 3 chars for quotes and url spacing
         emailShortUrl = config.page.shortUrl + '/sbl',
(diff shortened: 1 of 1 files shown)


katebee:
While I was checking #e6da9c9 I found another bug (#12615), this PR fixes both issues

gtrufitt:
🆒 👍 Just for reference, why did it reduce a character?

katebee:
Why it reduced a character:

> twitterMessageLimit = 114, // 140 - t.co length - 3 chars for quotes and url spacing

The limit of 115 didn't account for the ellipsis character added to the tweet when the selected text exceeds 140 characters.

gtrufitt:
Hmm, interesting, it was definitely right at some point! 😄 Maybe the url length changed or something.

gtrufitt:
Actually, the comment says t.co url and your screenshot has a gu.com url so probably that

katebee:
Ah, yep! I assumed it was the ellipsis throwing it off, missed the change of the url length...

Any potential issues with the change to using `encodeURIComponent` instead of `encodeURI`? Otherwise I will merge tomorrow morning.

gtrufitt:
I don't think so off the top of my head, I made a similar change (but from urlEncoded) myself a while back: https://github.com/guardian/frontend/commit/7a2cba628f251afa3885cd494cc59d839aa63e39

PR #12614, opened 2016-04-20
Video fade out

blongden73:
Adds an element to the video page which fades the screen when the video is playing. 

Some javascript to handle the fading in and out. When the mouse moves the page fades back there is a timeout on the transition of 3 seconds. So if the mouse is inactive again it comes back in. 

The screen fades back in as the page scrolls and gets to the related content at the bottom. 

picture 33

screenfade

Request comment @sammorrisdesign @rich-nguyen @jamesgorrie

### DIFF
--- a/applications/app/views/fragments/mediaBody.scala.html
+++ b/applications/app/views/fragments/mediaBody.scala.html
@@ -22,6 +22,12 @@
 
         @fragments.headTonal(media)
         <div class="content__main tonal__main tonal__main--@toneClass(media)">
+            @media match {
+                case video: Video => {
+                    <div class="video__fader"></div>
+                }
+                case _ => {}
+            }
             <div class="gs-container">
                 <div class="content__main-column content__main-column--media content__main-column--@mediaType">
                     <div class="media-body">
--- a/applications/app/views/fragments/mediaBody.scala.html
+++ b/applications/app/views/fragments/mediaBody.scala.html
@@ -24,7 +24,7 @@
         <div class="content__main tonal__main tonal__main--@toneClass(media)">
             @media match {
                 case video: Video => {
-                    <div class="video__fader"></div>
+                    <div class="video-fader"></div>
                 }
                 case _ => {}
             }
--- a/static/src/javascripts/bootstraps/enhanced/media/main.js
+++ b/static/src/javascripts/bootstraps/enhanced/media/main.js
@@ -367,9 +367,7 @@ define([
                 var related = $('.content-footer').offset().top;
                 var scroll = $(window).scrollTop();
                 var position = Math.floor(related - scroll);
-                console.log(position);
                 if (position <= 200){
-                    console.log('tihsihsih');
                     fader.removeClass('is-faded');
                     clearTimeout(faderTimeout);
                     bean.off(document.body, 'mousemove', setFaderTimeout);
--- a/static/src/javascripts/bootstraps/enhanced/media/main.js
+++ b/static/src/javascripts/bootstraps/enhanced/media/main.js
@@ -364,14 +364,16 @@ define([
             bonzo(player.el()).removeClass(endState);
             bean.on(document.body, 'mousemove', setFaderTimeout);
             bean.on(window, 'scroll', function (){
-                var related = $('#more-media-in-section').offset().top;
+                var related = $('.content-footer').offset().top;
                 var scroll = $(window).scrollTop();
                 var position = Math.floor(related - scroll);
-                if (position <= 0){
+                console.log(position);
+                if (position <= 200){
+                    console.log('tihsihsih');
                     fader.removeClass('is-faded');
                     clearTimeout(faderTimeout);
                     bean.off(document.body, 'mousemove', setFaderTimeout);
-                }else if (position > 0 && !player.paused()){
+                }else if (!player.paused()){
                     fader.addClass('is-faded');
                     bean.on(document.body, 'mousemove', setFaderTimeout);
                 }
--- a/static/src/javascripts/bootstraps/enhanced/media/main.js
+++ b/static/src/javascripts/bootstraps/enhanced/media/main.js
@@ -358,9 +358,8 @@ define([
                 bean.on(document.body, 'mousemove', function() {
                     fader.removeClass('is-faded');
                     if(videoCheck === false){
-                        console.log('video is paused');
+                        return;
                     }else{
-                        console.log('video is playing');
                         clearTimeout(faderTimeout);
                         faderTimeout = setTimeout(function() {
                         fader.addClass('is-faded');
--- a/static/src/javascripts/bootstraps/enhanced/media/main.js
+++ b/static/src/javascripts/bootstraps/enhanced/media/main.js
@@ -349,7 +349,12 @@ define([
             });
         });
         player.on('playing', function () {
+            $('.video__fader').addClass('isPlaying').css('opacity', '1');
             bonzo(player.el()).removeClass(endState);
+        });
+        
+        player.on('pause', function(){
+           $('.video__fader').removeClass('isPlaying').css('opacity', '0'); 
         });
     }
 
--- a/static/src/javascripts/bootstraps/enhanced/media/main.js
+++ b/static/src/javascripts/bootstraps/enhanced/media/main.js
@@ -338,6 +338,7 @@ define([
 
     function initEndSlate(player, endSlatePath) {
         var fader = $('.video__fader');
+        var videoCheck = false;
         var faderTimeout;
         var endSlate = new Component(),
             endState = 'vjs-has-ended';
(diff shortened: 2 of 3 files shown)


static/src/stylesheets/module/content/_media.global.scss
```diff
@@ -15,8 +15,28 @@
...
 
+    .video__fader {
+        position: fixed;
+        z-index: 9999;
+        top: 0;
+        bottom: 0;
+        left: 0;
+        right: 0;
+        background-color: rgba(0, 0, 0, .8);
+        opacity: 0;
```
sammorrisdesign: Do you need this is `visibility` is hidden?

static/src/stylesheets/module/content/_media.global.scss
```diff
@@ -15,8 +15,28 @@
...
+        transition: all 1s ease-in-out;
```
sammorrisdesign: Shouldn't be `all`, should just be `visibility` or `opacity`

static/src/javascripts/bootstraps/enhanced/media/main.js
```diff
@@ -348,8 +358,31 @@ define([
                 bonzo(player.el()).addClass(endState);
             });
         });
+
         player.on('playing', function () {
+            fader.addClass('is-faded');
             bonzo(player.el()).removeClass(endState);
+            bean.on(document.body, 'mousemove', setFaderTimeout);
+            bean.on(window, 'scroll', function (){
```
Calanthe: I think it's better to use `mediator.on('window:throttledScroll', debounce(fn, 200));` instead if `scroll`.

static/src/javascripts/bootstraps/enhanced/media/main.js
```diff
@@ -337,6 +337,16 @@ define([
     }
 
     function initEndSlate(player, endSlatePath) {
+        var fader = $('.video__fader');
+        var faderTimeout;
+        function setFaderTimeout(){
+            fader.removeClass('is-faded');
+            clearTimeout(faderTimeout);
+            faderTimeout = setTimeout(function() {
```
Calanthe: The `window.requestAnimationFrame` method is more efficient than `setTimeout()`

static/src/javascripts/bootstraps/enhanced/media/main.js
```diff
@@ -348,8 +358,31 @@ define([
...
+            bean.on(document.body, 'mousemove', setFaderTimeout);
+            bean.on(window, 'scroll', function (){
+                var related = $('.content-footer').offset().top;
+                var scroll = $(window).scrollTop();
+                var position = Math.floor(related - scroll);
+                if (position <= 200){
+                    fader.removeClass('is-faded');
+                    clearTimeout(faderTimeout);
+                    bean.off(document.body, 'mousemove', setFaderTimeout);
+                }else if (!player.paused()){
```
Calanthe: [minor] add a space character before `else`

jamesgorrie:
As we spoke about it might be worth wacking this behind a switch so it won't conflict with any of the test work we want to do with the video series stuff... Happy to get it merged though.

johnduffell:
very smart! I like it!

rich-nguyen:
Shouldn't this be desktop-only?

applications/app/views/fragments/mediaBody.scala.html
```diff
@@ -22,6 +22,12 @@
 
         @fragments.headTonal(media)
         <div class="content__main tonal__main tonal__main--@toneClass(media)">
+            @media match {
+                case video: Video => {
+                    <div class="video__fader"></div>
```
jamesgorrie: I'd probably just have it as `video-fader` as it's an element in it's own right.

static/src/javascripts/bootstraps/enhanced/media/main.js
jamesgorrie: Definitely better, nothing should be on the `window` `scroll` event/.

static/src/stylesheets/module/content/_media.global.scss
```diff
@@ -15,8 +15,28 @@
...
+        visibility: hidden;
+        
+        &.is-faded {
```
jamesgorrie: Could we BEM here please.
i.e. 
```CSS
.video-fader { /*...*/ }
.video-fader--active { /*...*/ }
```

static/src/javascripts/bootstraps/enhanced/media/main.js
```diff
@@ -348,8 +358,31 @@ define([
                 bonzo(player.el()).addClass(endState);
             });
         });
+
         player.on('playing', function () {
+            fader.addClass('is-faded');
             bonzo(player.el()).removeClass(endState);
+            bean.on(document.body, 'mousemove', setFaderTimeout);
```
jamesgorrie: You're going to want to `throttle` and `debounce` this too.

static/src/stylesheets/module/content/_media.global.scss
jamesgorrie: Yes, otherwise it will overlay the content, thus making links etc inactive.

static/src/javascripts/bootstraps/enhanced/media/main.js
```diff
@@ -348,8 +358,31 @@ define([
...
+                    fader.addClass('is-faded');
+                    bean.on(document.body, 'mousemove', setFaderTimeout);
+                }
+            });
+        });
+
+        //add in on video stop remove class
+        player.on('pause', function(){
+            clearTimeout(faderTimeout);
+            bean.off(document.body, 'mousemove', setFaderTimeout);
```
jamesgorrie: `throttle` and `debounce` please.

jamesgorrie:
When we test this - similar to autoplay - it would be good to have some for of, "Please don't do this" button, if not just to allow people to disable it, but to allow us to get feedback on whether we're pissing people off.

static/src/stylesheets/module/content/_media.global.scss
```diff
@@ -15,8 +15,28 @@
...
+            visibility: visible;
+            opacity: 1;
+            transition: all 1s ease-in-out;
+        }
+    }
+
     .player {
         margin-bottom: $gs-baseline/2;
+        z-index: 99999;
```
jamesgorrie: This will inevitably give us grief in the future.
We've tried to add some [logic around applying `z-index`s](https://github.com/guardian/frontend/blob/master/static/src/stylesheets/base/_zindex.scss) - it'd be good to see this used.

static/src/javascripts/bootstraps/enhanced/media/main.js
```diff
@@ -337,6 +337,16 @@ define([
     }
 
     function initEndSlate(player, endSlatePath) {
+        var fader = $('.video__fader');
```
jamesgorrie: Why is this in the `initEndSlate`?
I think I can see why this is done, but given the complexity of this file already, it'd be good to shift it else where.

fredex42:
I like it! I think it would work fine on iPad but probably not a smaller screen

jamesgorrie:
@rich-nguyen

> Shouldn't this be desktop-only?  

I've realised that, because this is in the `endSlate` function - this inherits it from there. It'd be good to see this logic more self contained and not adding to the spaghetti that is the `media/main` file.

sammorrisdesign:
@fredex42 Wouldn't the video play in the native play on iPad and not inline. Meaning you wouldn't be able to see the fade anyway. (Unless they've changed the way videos work on iPad)

paperboyo:
> Wouldn't the video play in the native play on iPad.

Not on Chrome on Android phone, though. Wouldn't know about this iPad thing.

static/src/stylesheets/module/content/_media.global.scss
gtrufitt: Does this not cause problems when you open an image overlay with a video on the page? Feels like the player would be on top when it shouldn't be? Should you just apply a higher z-index when you know the fade is there? Also should you be checking whether a modal window is open or does it only appear when playing?

blongden73: @gtrufitt @jamesgorrie  Yeah this only happens when the video is playing. If I set the z-index to 1000 this should be inline with the logic around z-index and will be behind ads, overlays etc...

blongden73:
@fredex42 @sammorrisdesign Checked on mobile, and tablet, the video plays straight away in the native player so the fade wont happen on them.

static/src/javascripts/bootstraps/enhanced/media/main.js
```diff
@@ -348,10 +351,47 @@ define([
...
+            var position = Math.floor(related - scroll);
+                if (position <= 200){
+                    fader.removeClass('video-fader--active');
+                    clearTimeout(faderTimeout);
+                    bean.off(document.body, 'mousemove', setFaderTimeout);
+                } else if (!player.paused()){
+                    fader.addClass('video-fader--active');
+                    bean.on(document.body, 'mousemove', setFaderTimeout);
+                }
+        }
```
OliverJAsh: I don't entirely understand the need for this function. Could you try to explain why it's needed?

johnduffell:
is this one still going out?

jamesgorrie:
There's a lot, performance wise that needs to be done.
We would also need to run it as a test, so as not to add complexity without value.

johnduffell:
fair point - if it's up in the air it might be worth closing until it's ready

jamesgorrie:
@blongden73 are you happy to put this on hold?

johnduffell:
closing, please reopen when you're ready

PR #12617, opened 2016-04-20
Minor commercial refactor fixes

regiskuckaertz:
Two things:
1. Rob didn't want masterclasses without pictures in the blended component
2. The width of the component header has been adjusted at `leftCol` and `wide` breakpoints

cc @kelvin-chappell @uplne


commercial/app/controllers/commercial/Multi.scala
```diff
@@ -75,15 +75,15 @@ object Multi
               }
             case ("masterclasses", eventBriteId) if eventBriteId.nonEmpty =>
               Future.successful {
-                MasterclassAgent.specificMasterclasses(Seq(eventBriteId)).headOption orElse {
-                  MasterclassAgent.masterclassesTargetedAt(segment).headOption
+                MasterclassAgent.specificMasterclasses(Seq(eventBriteId)).filter(!_.mainPicture.isEmpty).headOption orElse {
+                  MasterclassAgent.masterclassesTargetedAt(segment).filter(!_.mainPicture.isEmpty).headOption
```
JonNorman: More of an FYI than a suggestion to make a change but there is a `filterNot` method that can be used rather than negating an expression within a `filter` method.

JonNorman:
:+1: Looks good to me!

PR #12618, opened 2016-04-20
Create frontend model from capi sponsorship model

kelvin-chappell:
This integrates the new sponsorship data from Capi responses into the frontend content model.  
There's no change in what's rendered yet so this bit doesn't need to be behind a feature switch.

/cc @steppenwells

### DIFF
--- a/common/app/model/Tag.scala
+++ b/common/app/model/Tag.scala
@@ -218,7 +218,7 @@ object TagProperties {
       bylineImageUrl = tag.bylineImageUrl,
       podcast = tag.podcast.map(Podcast.make),
       references = tag.references.map(Reference.make),
-      activeBrandings = tag.activeSponsorships.map(_.map(Branding.make)).getOrElse(Nil)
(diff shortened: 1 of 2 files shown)


common/app/model/Tag.scala
```diff
@@ -114,6 +117,88 @@ case class Reference(
...
+  override val name: String = "foundation"
+}
+
+case object PaidContent extends SponsorshipType {
+  override val name: String = "paid-content"
+}
+
+object SponsorshipType {
+
+  def make(name: String): SponsorshipType = name match {
```
janua: Is this function intentionally partial?

kelvin-chappell: No.  I'll complete it.

janua: :clap:

kelvin-chappell:
@janua Is that better?

janua:
LGTM :+1:

PR #12620, opened 2016-04-20
Pageskin adcall filtering

rich-nguyen:
This reduces the appearance of the out-of-page ad slot html. The server should only render this ad slot when the ad unit for the request has a targeted page skin. Currently this unfilled dfp call costs us hundreds of pounds everyday.


JonNorman:
:+1: looks good

PR #12624, opened 2016-04-21
Failed requests are logged as warning

TBonnin:
## What does this change?

Requests can failed for external reasons (non existing paths for
example). It's better not to add noise to our error logs.
Also if an internal error happens while processing a requests, it is
likely to be logged as well

Better logs structure, easier to monitor internal errors by removing noise

No
## Request for comment

@jamespamplin

### DIFF
--- a/discussion/app/controllers/CommentsController.scala
+++ b/discussion/app/controllers/CommentsController.scala
@@ -155,7 +155,6 @@ object CommentsController extends DiscussionController with ExecutionContexts {
     }
       .recover {
         case NonFatal(e) =>
-          log.error(s"Error fetching comments from request with following headers: ${request.headers}", e)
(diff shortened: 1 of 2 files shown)


discussion/app/controllers/CommentsController.scala
```diff
@@ -155,7 +155,6 @@ object CommentsController extends DiscussionController with ExecutionContexts {
     }
       .recover {
         case NonFatal(e) =>
-          log.error(s"Error fetching comments from request with following headers: ${request.headers}", e)
```
TBonnin: Also removing this logging which was added to find out where traffic which is crawling our discussions was coming from.

jamespamplin:
:+1: looks good to me

PR #12626, opened 2016-04-21
Fix fluid250 parallax

regiskuckaertz:
Ahem, some objects needed to be wrapped in Bonzo before being able to call `addClass` on them. My bad 💨 

cc @uplne


uplne:
👍

PR #12627, opened 2016-04-21
Add switch to enable constant DFP request of 1x1 ad slot

JonNorman:
## What does this change?

@rich-nguyen [put in a change](https://github.com/guardian/frontend/pull/12620) that stops us requesting a 1x1 (out-pf-page) ad slot on every page.

Surveys depend on this slot being there and so this switch allows us to revert back to this behaviour when surveys are live.

**NB**
Surveys are soon to be delivered via MPU slots and so will no longer depend on the 1x1 ad slot, meaning that this switch can then be discarded.

Surveys can be run in their current form on the desktop site.

Nope.


jbreckmckye:
👍

PR #12628, opened 2016-04-21
Make soulmates colors a little bit more readable

uplne:
Not that it is that much better but...

screen shot 2016-04-21 at 14 23 31

CC @regiskuckaertz


regiskuckaertz:
No var? Boom 👍

uplne:
Nice, I've put it into variable

PR #12630, opened 2016-04-21
Add treats to video container

sammorrisdesign-zz:
## What does this change?

This adds treats to video container.

This will allow editors to showcase new series and that.
## Screenshots

screen shot 2016-04-21 at 15 05 34
## Request for comment

@jamesgorrie


jamesgorrie:
Nice 👍

PR #12632, opened 2016-04-21
Soulmates tone for gimbap richmedia

uplne:
Merch people asked me for the soulmates version of the richmedia gimbap. We will see if they will be able to create some nice copy to match the template...but why not.

screen shot 2016-04-21 at 15 16 11

CC @regiskuckaertz


regiskuckaertz:
Tones will rule the world! 👍

PR #12633, opened 2016-04-21
Spruce up the switchboard

JonNorman:
## What does this change?

We have a lot of switches, whilst cleaning up and sorting the Commercial switches I thought it would be useful to be able to add a small list of the switch groups to the top of the page.

Quite a small change, happy to throw away if people don't think it would be useful.

A cleaner switchboard

No
## Screenshots

image


JonNorman:
The `git diff` on `switchboard.scala.html` looks a lot worse than it is, all I've changed is defining the `switchGroup` at the top and the subsequent contents `<ul>`

common/app/conf/switches/CommercialSwitches.scala
```diff
@@ -213,14 +195,32 @@ trait CommercialSwitches {
...
-    SwitchGroup.Commercial,
+    SwitchGroup.CommercialFeeds,
     "book-lookup",
     "If this switch is on, book data will be looked up using a third-party service.",
     safeState = Off,
     sellByDate = never,
     exposeClientSide = false
   )
 
+  val MembersAreaSwitch = Switch(
```
regiskuckaertz: This should probably be done in a separate PR

common/app/conf/switches/CommercialSwitches.scala
```diff
@@ -213,14 +195,32 @@ trait CommercialSwitches {
...
+    SwitchGroup.Commercial,
+    "gu-members-area",
+    "If this switch is on, content flagged with membershipAccess will be protected",
+    safeState = On,
+    sellByDate = never,
+    exposeClientSide = false
+  )
+
+  val LCMortgageFeedSwitch = Switch(
```
regiskuckaertz: *ibid*

admin/app/views/switchboard.scala.html
```diff
@@ -1,44 +1,54 @@
...
+        <blockquote>
+            <em>"Death by switches."</em>
+            <footer>Nick Haley</footer>
+        </blockquote>
+    </p>
+    <div>
+        <h3>Switch Groups</h3>
+        <ul class="multi-column--3">
+        @for((group, _) <- switchGroups){
+            <li><a href="#@group.name">@group.name</a></li>
```
regiskuckaertz: Is there a Scala thing that automatically encode string so that they're URL-safe?

common/app/conf/switches/CommercialSwitches.scala
JonNorman: I didn't (intentionally) create any new switches - this looks to be an artefact of these switches being removed subsequent to the raising of this PR. I'll merge `master`, which should fix this.

admin/app/views/switchboard.scala.html
jamespamplin: You could try using `views.support.URLEncode("...")`

common/app/conf/switches/CommercialSwitches.scala
JonNorman: Actually ignore that - it is just because I moved some of the Commercial switches around, the diff is misleading; these switches are not new in my commit.

regiskuckaertz:
Looks good to me, I'd be happy to give you a +1 though someone that has ... history with the switchboard would make more sense.

johnduffell:
if you have whitespace changes you can ignore by adding [?w=0](https://github.com/guardian/frontend/pull/12633/files?w=0)

johnduffell:
looks good! good improvement, but I fear you're fixing the wrong problem though and we should just delete some of them....
👍

PR #12635, opened 2016-04-21
Paid containers - visual fixes to align with colour theme

lps88:
## What does this change?

Some colours changes on 'in pictures' text and borders of the connect buttons on following page, to fit in with the tone
http://m.code.dev-theguardian.com/reservation-las-vegas


jbreckmckye:
👍

PR #12637, opened 2016-04-21
Increase the distance between elements and inline ads

regiskuckaertz:
## What does this change?

Currently, inline ads can be inserted into the body of articles just 35px below `.element`s, which can result in two visual elements competing for attention:

picture 7

I increased the minimum space to 200px. The value is less visual noise inside articles; but it is also likely to result in a significant drop of impressions.

picture 6

/cc @jbreckmckye @uplne


jbreckmckye:
How will we know if the drop of impressions is too much to bear?

regiskuckaertz:
Good question! I'm sending an email to Pat, it's his call I guess. We can run a sample for a few days if he wants to.

jbreckmckye:
👍 Code is fine, but might be worth asking how we can track / validate the consequences of this. We could do a 50% A/B test, for example, and count impressions in Moat, or maybe even revenue in DFP (I think?)

PR #12639, opened 2016-04-21
don't change any more og:urls to https

johnduffell:
Facebook doesn't like us changing the og:url of articles, if we do they treat it as a new article.   This has already happened for some sections on https migration.

This PR prevents it happening in future - all future section migrations won't affect the og:url.  This is as per their [helpful instructions](https://developers.facebook.com/docs/sharing/webmasters/crawler#updating) on pretending to them we haven't changed anything.

This is as a subsequent PR to https://github.com/guardian/fastly-edge-cache/pull/613 - that one has to ship first as they want to be able to access stuff on the old url, not just via a redirect.


common/app/model/meta.scala
```diff
@@ -290,15 +290,23 @@ final case class MetaData (
...
+    val httpsEnabledSectionsApr2016: Seq[String] = Seq("info", "email", "science", "crosswords", "technology")
+    // for all other sections, keep the old og:url even once the migration happens
+    def ogUrl = if (!httpsEnabledSectionsApr2016.contains(section)) webUrl.replaceFirst("^https:", "http:") else webUrl
+
+    Map(
+      "og:site_name" -> "the Guardian",
+      "fb:app_id"    -> Configuration.facebook.appId,
+      "og:type"      -> "website",
+      "og:url"       -> ogUrl
+    ) ++ (iosId("applinks") map (iosId => List(
```
jamespamplin: I know it wasn't introduced in this PR, but can we separate the ios properties to its own `val`, for greater readability

common/app/model/meta.scala
```diff
@@ -290,15 +290,23 @@ final case class MetaData (
...
-    "og:url"       -> webUrl) ++ (iosId("applinks") map (iosId => List(
-    "al:ios:url" -> s"gnmguardian://$iosId",
-    "al:ios:app_store_id" -> "409128287",
-    "al:ios:app_name" -> "The Guardian"
-  )) getOrElse Nil)
+  def opengraphProperties: Map[String, String] = {
+    // the cat is already out of the bag for these sections
+    val httpsEnabledSectionsApr2016: Seq[String] = Seq("info", "email", "science", "crosswords", "technology")
+    // for all other sections, keep the old og:url even once the migration happens
+    def ogUrl = if (!httpsEnabledSectionsApr2016.contains(section)) webUrl.replaceFirst("^https:", "http:") else webUrl
```
jamespamplin: `httpsEnabledSectionsApr2016` is only used by the `og:url` property, so you can scope the value and block to just that property, or separate it out to its own function - for greater readability.

johnduffell:
noone in audience had strong opinions on whether to switch the old sections back to http or not, so to keep the code base consistent I'm switching everything back to http

TBonnin:
Simpler this way 👍

PR #12640, opened 2016-04-21
Aim to limit football competition agents to polling the current season only

JustinPinner:
I opened this PR mainly to run tests (lots of unrelated tests failed locally), and to put the idea out there.
## What does this change?

We currently poll PA for football season data going way back (to the 1880s in some cases). There's probably not much point in polling for updates to data beyond the current season, so this PR attempts to limit the date range it considers.

We'll reduce the number of calls to PA for old data.

No.
## Request for comment

@TBonnin

### DIFF
--- a/sport/app/football/feed/Competitions.scala
+++ b/sport/app/football/feed/Competitions.scala
@@ -97,6 +97,14 @@ trait Competitions extends LiveMatches with Logging with implicits.Collections w
     DateTimeComparator.getInstance.asInstanceOf[Comparator[DateTime]]
   )
 
+  private def mostRecentCompetitionSeasons(competitions: List[Season], numToTake: Int): List[Season] = {
+    competitionDefinitions.flatMap{ compDef =>
+      competitions.filter(_.competitionId == compDef.id)
+        .sortBy(_.startDate.toDateTimeAtStartOfDay.getMillis).reverse
+        .take(numToTake)
+    }
+  }
+
   val competitionDefinitions = Seq(
     Competition("500", "/football/championsleague", "Champions League", "Champions League", "European", tableDividers = List(2, 6, 21)),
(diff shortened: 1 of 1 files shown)


sport/app/football/feed/Competitions.scala
```diff
@@ -136,15 +144,23 @@ trait Competitions extends LiveMatches with Logging with implicits.Collections w
...
   def refreshCompetitionData() = {
     log.info("Refreshing competition data")
-    FootballClient.competitions.map(_.flatMap { season =>
-      competitionAgents.find(_.competition.id == season.id).map { agent =>
-        val newCompetition = agent.competition.startDate match {
-          case Some(existingStartDate) if season.startDate.isAfter(existingStartDate.toDateTimeAtStartOfDay) => agent.update(agent.competition.copy(startDate = Some(season.startDate)))
-          case None => agent.update(agent.competition.copy(startDate = Some(season.startDate)))
-          case _ =>
+    FootballClient.competitions.map { allComps =>
+      mostRecentCompetitionSeasons(allComps, numToTake = 1).map { season =>
```
TBonnin: why not assign a default value to the numToTake param in the function def, so you don't need to specify it here (and anyway it would be 1 most of the time)

johnduffell:
looks great!

sport/app/football/feed/Competitions.scala
johnduffell: or even just make it `mostRecentCompetitionSeason` and hard code it to call `headOption` instead of `take(num)` on the basis that if someone needs more later they can edit the code

TBonnin:
other than the comment. lgtm.
Thanks Justin

johnduffell:
🚤

PR #12644, opened 2016-04-22
Added test cookie variable to test page

JonNorman:
## What does this change?

The [commercial pageskins page](https://frontend.gutools.co.uk/analytics/commercial/pageskins) doesn't include the ad-test in the link to the page skin (for pageskins that are behind ad-tests).

Now it does.

A better experience when troubleshooting page skins.

No
## Screenshots

**before:**
image

**after:**
image


common/app/common/dfp/PageSkinSponsorship.scala
```diff
@@ -11,33 +11,21 @@ case class PageSkinSponsorship(lineItemName: String,
...
-  implicit val pageskinSponsorShipReads: Reads[PageSkinSponsorship] = (
-    (JsPath \ "lineItem").read[String] and
-      (JsPath \ "lineItemId").read[Long] and
-      (JsPath \ "adUnits").read[Seq[String]] and
-      (JsPath \ "editions").read[Seq[Edition]] and
-      (JsPath \ "countries").read[Seq[String]] and
-      (JsPath \ "isR2Only").read[Boolean] and
-      (JsPath \ "isAdTest").read[Boolean]
-    )(PageSkinSponsorship.apply _)
+  implicit val pageskinSponsorShipFormat: Format[PageSkinSponsorship] = (
```
kelvin-chappell: Nice simplification.

kelvin-chappell:
Looks good but in prod, links that are behind a test cookie should go to the code site. 
(Because they don't work in prod to avoid the page shrinking if the edition has a test running on it).

admin/app/tools/DfpLink.scala
```diff
@@ -24,21 +25,25 @@ object DfpLink {
...
-      None
+    lazy val domain = if (adTest.isDefined) commercial.testDomain else site.host
+
+    val relativePath =
+      if (path endsWith "/front/ng")
+        getRelativePath(2)
+      else if (path endsWith "front")
+        getRelativePath(1)
+      else
+        getRelativePath(0)
```
kelvin-chappell: This should never happen because pageskins are only allowed on fronts.  It might be worth adding a test to make sure this is the case.

kelvin-chappell:
If there's a test to make sure pageskins can only run on fronts: 👍

admin/app/views/commercial/pageskins.scala.html
```diff
@@ -45,22 +51,17 @@
     <h2>Pageskinned Ad Units</h2>
     <p>Last updated: @pageSkinnedAdUnits.updatedTimeStamp</p>
 
-    <h3>Deliverable to NGW</h3>
+    <h3>Deliverable without test cookie</h3>
```
rich-nguyen: `Live Deliverable`? Then below could be `Test Deliverable (behind cookie)`?

JonNorman: Much better wording!

JonNorman:
I clarified a few points around valid page-skins with ad ops. I've reinstated the validation that was there before and I've refactored the `PageSkin` Object a bit.

kelvin-chappell:
Looks good but I wouldn't deploy it this afternoon! 👍

JonNorman:
Ah I guess it can wait until Tuesday.

PR #12645, opened 2016-04-22
Don't show email sign-up on sensitive content

gtrufitt:
## What does this change?

Add isSensitive js config and checks for it in email sign-up insertion. Also moves film today sign-up to bottom-of-page.

Prevents showing the chirpy 'want stories like this in your inbox?' on things like obits.
## CC

@crifmulholland


jamespamplin:
:+1: looks good to me, thanks for this!

gtrufitt:
Ta @jamespamplin

PR #12647, opened 2016-04-22
Fix button colors in refactored manual commercial components

regiskuckaertz:
This addresses an issue where the refactored manual components would not use the correct button colour.

Before:
picture 8

After:
picture 9

cc @Calanthe @uplne


Calanthe:
👍

PR #12649, opened 2016-04-22
Fix discussion exceptions

regiskuckaertz:
## What does this change?

Exceptions thrown by the discussions loader cannibalise the console when run locally.

No more unsightly exceptions (but still reported by raven).

### DIFF
--- a/static/src/javascripts/projects/common/modules/discussion/loader.js
+++ b/static/src/javascripts/projects/common/modules/discussion/loader.js
@@ -73,7 +73,9 @@ Loader.prototype.initTopComments = function() {
                 this.setState('has-top-comments');
             }
         }.bind(this)
-    );
+    ).catch(function () {
+        /* ignore */
+    });
 };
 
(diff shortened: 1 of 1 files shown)


static/src/javascripts/projects/common/modules/discussion/loader.js
```diff
@@ -73,7 +73,9 @@ Loader.prototype.initTopComments = function() {
                 this.setState('has-top-comments');
             }
         }.bind(this)
-    );
+    ).catch(function () {
+        /* ignore */
```
jamesgorrie: Isn't that a little dangerous?

gtrufitt: Can we send them to sentry so if there are prod errors they get logged nicely?

regiskuckaertz: Yep I'll improve that now.

jamesgorrie:
Nice 👍

PR #12650, opened 2016-04-22
Fabric v1 - again

jbreckmckye:
After the [first implementation](https://github.com/guardian/frontend/pull/12586) of Fabric, the requirements came under question - this pull request implements the new needs.

Commercial is currently under pressure to start building fabric adverts by the end of next week. This PR is a (somewhat hasty) way to meet that problem - it clones the Fluid250 and ExpandableVideoV2 templates with some slight changes to meet the specific requirements of Fabric. I'm hoping that next week I'll have opportunities to refine this, but in the meanwhile, it's very important that we can get something into the CODE environment for Carlo's team to start experimenting with.

**What does this PR do?**
- creates a Fabric-v1 and Fabric-expandable template
- replaces the top slot on mobile with a substitute slot (either after the first container on fronts, or on the first inline in articles) that requests a new ad size - 88x71 - that targets specifically the new creative type.

You'll notice that I've essentially cloned the original CSS and JavaScript by process of simple repetition. This is a tactical decision - it's code we can develop separately during the week without risk of harming live creatives that use the existing code. The commercial team is keen to 'branch' creatives in this fashion - it gives them jerry-rigged way to use versioned dependencies.

_Inline full width ad after first container on mobile - a substitute for the top nav ad_
image

_Equivalent placement in an article_
image

@regiskuckaertz


static/src/javascripts/projects/common/modules/commercial/creatives/fabric-expanding-v1.js
```diff
@@ -0,0 +1,94 @@
...
+                    videoSrcAutoplay = videoSrcAutoplay.replace(this.isClosed ? 'autoplay=0' : 'autoplay=1', this.isClosed ? 'autoplay=1' : 'autoplay=0');
+                }
+                $('.ad-exp__close-button').toggleClass('button-spin');
+                $('.ad-exp__open-chevron').removeClass('chevron-up').toggleClass('chevron-down');
+                this.$ad.css('height', this.isClosed ? this.openedHeight : this.closedHeight);
+                $('.slide-video, .slide-video .ad-exp__layer', $(this.$adSlot[0])).css('height', this.isClosed ? this.openedHeight : this.closedHeight).toggleClass('slide-video__expand');
+                this.isClosed = !this.isClosed;
+                setTimeout(function () {
+                    $('#YTPlayer').attr('src', videoSrcAutoplay);
+                }, 1000);
```
regiskuckaertz: 1000 b/c of the animation?

jbreckmckye: I guess so? It's a clone of expanding-v2, really.

static/src/javascripts/projects/common/modules/commercial/sticky-top-banner.js
```diff
@@ -76,8 +76,11 @@ define([
         var $iframe = getAdIframe();
         var slotWidth = $iframe.attr('width');
         var slotHeight = $iframe.attr('height');
+        var slotSize = [slotWidth, slotHeight].join(',');
```
regiskuckaertz: DFP uses the sizes of the winning creative to set the `width` and `height` on the `iframe`? I didn't know that 😮

regiskuckaertz:
Let's roll this out and figure out how we could improve the code later 👍

static/src/javascripts/projects/common/modules/commercial/sticky-top-banner.js
jbreckmckye: I think the idea is that most of the adverts in question apply height properties to the `$adSlot`. It might be nicer if the creative object itself could be responsible for this and exposing events.

regiskuckaertz:
@sndrs they're not 😆 we're really trying to take a bird's eye view at our work and rethink everything from the ground up—this shouldn't block work on new creatives though.

PR #12651, opened 2016-04-22
Press mainMedia embeds into immersive pages

sammorrisdesign-zz:
## What does this change?

At the moment if we paste some `html` into composer's main media embed slot, it gets `fenced`. This means we slow down the load of the page and limit what the interactive developer can do. This PR is an experiment that will `un-iframe` main media embed code. With one caveat, the article must have been modified in the last sixty days. After that the embed will be re-iframed and the article will be essentially "archived".

To be double safe I've also put this behind a switch.
## Request for comment

@gklopper


common/app/views/support/HtmlCleaner.scala
```diff
@@ -426,6 +426,18 @@ case class ImmersiveLinks(isImmersive: Boolean) extends HtmlCleaner {
   }
 }
 
+case class ImmersiveMainEmbed(isImmersive: Boolean, isSixtyDaysModified: Boolean) extends HtmlCleaner {
+  override def clean(document: Document): Document = {
+      if(immersiveMainEmbedSwitch.isSwitchedOn && isSixtyDaysModified) {
+        val srcdoc = document.getElementsByTag("iframe").attr("srcdoc")
+        if(isImmersive && srcdoc != null) {
```
gtrufitt: Can you add `isImmersive` into the if statement above?

common/app/views/support/HtmlCleaner.scala
```diff
@@ -426,6 +426,18 @@ case class ImmersiveLinks(isImmersive: Boolean) extends HtmlCleaner {
   }
 }
 
+case class ImmersiveMainEmbed(isImmersive: Boolean, isSixtyDaysModified: Boolean) extends HtmlCleaner {
+  override def clean(document: Document): Document = {
+      if(immersiveMainEmbedSwitch.isSwitchedOn && isSixtyDaysModified && isImmersive) {
+        val srcdoc = document.getElementsByTag("iframe").attr("srcdoc")
+        if(srcdoc != null) {
+            document.getElementsByTag("body").html(srcdoc)
```
gklopper: Does this replace the entire body? I'm not sure that is what we want.

Will this have unexpected consequences on any bit of immersive content created in the last 60 days?

gklopper:
I'm not sure all the side effects on older immersive content have been thought through. Likewise I'm worried about this having unintended consequences on new immersive content not specifically written for it.

common/app/views/support/HtmlCleaner.scala
sammorrisdesign: The cleaner seems to add wrap what it's given in `body` tags. the `document` in this case being just the main media.

sammorrisdesign:
@gklopper I'm happy to put a published after today statement in there to doubly safeguard it.

static/src/stylesheets/module/content/_article-immersive.scss
```diff
@@ -95,6 +95,34 @@
...
     }
 
+    .content__main-embed {
+        position: absolute;
+        top: 0;
+        right: 0;
+        left: 0;
+        bottom: 0;
+        z-index: 1;
+    }
```
OliverJAsh: Do we not already have these styles for images in the main media?

static/src/stylesheets/module/content/_article-immersive.scss
```diff
@@ -671,6 +680,7 @@
             bottom: 0;
             left: 0;
             right: 0;
+            z-index: 2;
```
OliverJAsh: Why is this needed? We need to document things like this

OliverJAsh:
> limit what the interactive developer can do.

What do you need to do that you can't do when the iframe is fenced?

Any extra behaviour you can achieve when the embed is not ifamed will surely be lost after 60 days once the embed is iframed? Is that what we want?

Will embeds be designed to gracefully degrade when they are iframed? Or will they simply fail after 60 days?

Is there a risk that an embed could break the page when it is not iframed? All it takes is one bad embed.

OliverJAsh:
I have some questions, but as this is behind a switch, we can continue to review this decision. 👍

jamesgorrie:
I echo the question of - if we are going to turn this off after 60 days - what happens to the unfenced functionality? Implicitly this seems to be saying we want to break the law for a bit while traffic is high? 

Specifying what functionality is lacking and then adding it permanently seems like a better solution.

sammorrisdesign:
@OliverJAsh So yes, anything added in the `mainMedia` will have to be built to be non crucial.

One example of something that we'll be able to do smoother is [The County](http://www.theguardian.com/us-news/2015/dec/01/the-county-kern-county-deadliest-police-killings). If we had this system we could achieve that header without js. Equally the page colouring which has degraded considerably since launch, would be removed.

I agree that things can break and the responsibility relies with the developer. However, things can break now we just thrash layout to provide a more janky experience before they do break.

@jamesgorrie Given how prone bespoke elements are to break, I think this solution works well for that. The example above works well here too. The page styling isn't required but does add a lot. However, changes we've made in `frontend` since have naturally caused some regressions.

It's also worth noting that `thrashers` are also pressed into the page and we haven't had any problems. In the event of issues, we could easily remove any breaking code.

OliverJAsh:
I think this needs further discussion.

In the meantime, as a workaround for @blongden73’s requirements, we've created an interactive embed that removes the headline:

``` js
define([], function () {
    return {
        boot: function (el, context, config, mediator) {
            var headerEl = document.querySelector('.content__header');
            if (headerEl) {
                headerEl.remove();
            }
        }
    };
});
```

johnduffell:
is anything happening with this or has it gone back to the drawing board for a rethink?

sammorrisdesign:
Not sure I got a 👍 and then a "this needs a rethink"

From my point of view this is good to go. I'd like to merge it in, it's behind a switch if we notice old things going sideways.

gklopper:
I'm worried that this simply unpacks all recent iframe type interactives and they have not been written to take that into account (i.e. broken inetractives will happen).

So I better not get a call this weekend about broken interactives 🔥 🔥 🔥 .

But if you are going to be reallllly careful then 👍

sammorrisdesign:
@gklopper If you don't want to get a call this weekend, I can always merge it in on Monday ;)

johnduffell:
personally I don't like 60 days cutoffs, we don't want to effectly change things 60 days after publication because stuff might just be breaking.  Can we just choose how we want to do it and implement it that way?  If this is just a test and you are gathering data to support a proper change ( i guess so as it has a switch) then can we just have a list of urls that it tries it on, and if that goes well open it up more widely?

gklopper:
Sam is just doing a test to see how this works. 

If you look at it in that perspective then the 60 day thing makes sense (i.e. it may well revert and we cannot come to rely on it in the meantime).

We are probably going to schedule a reasonable amount of interactive work in Q2 (and I doubt this lives beyond that), so learning from this would be great (i.e. does this actually improve Google search results)

johnduffell:
fair enough after all it has a switch.
however merge conflicts! 💀

PR #12654, opened 2016-04-22
use etags/if-none-match for all responses

johnduffell:
This PR follows on from https://github.com/guardian/frontend/pull/12610 and extends the etag and revalidation logic across the whole site.

This will remove any unnecessary resending of responses and therefore save us plenty of bandwidth out from the apps and the router.

@cb372 @jamespamplin @piuccio take a look!


piuccio:
forgot to ask in the other PR, how's the CPU doing?

johnduffell:
@piuccio can't really notice anything obvious
image

OliverJAsh:
Awesome change!

TBonnin:
Responding witch a `revalidatable` result is `Cached(XX)(RevalidatableResult(PlayResult))`
Not revalidating is `Cached.withoutRevalidation(XX)(PlayResult)`

Is there any way to be consistant with those 2 approach?

johnduffell:
@TBonnin I've made a trait with 2 types, and used that for both revaidatable and non revalidatable results.  In the process, I could simplify the logic in a couple of places, but I did need to provide an implicit request header (for the if-none-match header) in various places that weren't actually revalidatable.  See what you think?

common/app/model/Cached.scala
```diff
@@ -29,41 +31,55 @@ object Cached extends implicits.Dates {
...
-  def explicitlyCache(seconds: Int)(result: Result): Result = cacheHeaders(seconds, result)
+  def explicitlyCache(seconds: Int)(result: Result): Result = cacheHeaders(seconds, result, None)
+
+  def apply(seconds: Int, cacheableResult: CacheableResult, ifNoneMatch: Option[String]) = cacheableResult match {
+    case RevalidatableResult(result, hash) =>
+      if (cacheableStatusCodes.contains(result.header.status)) {
+        val hashMaybeRequestHash = (hash, ifNoneMatch)//FIXME could be comma separated
+        cacheHeaders(seconds, result, Some(hashMaybeRequestHash))
+      } else
+        result
```
TBonnin: I am not a huge fan of the if/else without brackets

jamesgorrie: MIght be nicer if `result` was on the same line?

common/app/model/Cached.scala
```diff
@@ -29,41 +31,55 @@ object Cached extends implicits.Dates {
...
-  }
+  case class Hash(string: String)
 
-  def apply(page: Page)(result: Result): Result = {
-    val cacheSeconds = page.metadata.cacheTime.cacheSeconds
-    if (cacheableStatusCodes.contains(result.header.status)) cacheHeaders(cacheSeconds, result) else result
-  }
+  sealed trait CacheableResult
+  case class RevalidatableResult(result: Result, hash: Hash) extends CacheableResult
+  case class WithoutRevalidationResult(result: Result) extends CacheableResult
```
TBonnin: I would propose `NoRevalidationResult` here to save a few character but this is up to you

common/app/model/Cached.scala
```diff
@@ -29,41 +31,55 @@ object Cached extends implicits.Dates {
...
+      val hashLong: Long = writeable.transform(content).foldLeft(z = 1L){
         case (accu, nextByte) => 31 * accu + nextByte
       }
       new RevalidatableResult(result, Hash(hashLong.toString))
     }
+
+    def Ok[C](content: C)(implicit writeable: Writeable[C]) = {
+      apply(Results.Ok(content), content)
+    }
+  }
```
TBonnin: Do we need to support more result status code? I mean not only 200 OK but also other code from the 2XX family

johnduffell: that's just a convenience method to save me hassle, we only cache 200s and 404s anyway, and we don't return any other 2xx codes as far as I could see

TBonnin:
lgtm

common/app/model/Cached.scala
johnduffell: might be nicer if I wrote the right logic 🐑

johnduffell: there's no embarassed smiley so I went for a sheep 😝

johnduffell:
Initial indications are that we'll save around 16GB per hour from this PR, which is 4 times as good as what we got from #12610 making an additional saving of around $1200 dollars a month on AWS bandwidth.  I haven't worked out how much we're saving on fastly.

The blue line is bandwidth, the red line is requests, and the orange line is the number of requests we didn't have to serve the body.
image

TBonnin:
👏 👏 👏  This is awesome

rich-nguyen:
I can't wait to see our aws bill in May. the cloudwatch bill is down by $2000 for April too.

johnduffell:
you can actually see in fastly when deploys that change the body happen
image

johnduffell:
overnight we had less editing of articles and deploying of frontend so things improved a little.  Interestingly there was a drop at midnight when we change the html of fronts to include the new date.  Not sure if articles change their source too, but we would do well to avoid that.
I did a rough totalling of 6 hour periods and got savings of approx 720GB per day or around 20%.  So we saved around $75 in a day. If this were replicated over a month we would actually save around $2200 per month!

I'll see what happens over the following weekend, as theoretically we should get better savings with no deployments and less updates, however on the flip side there's less traffic in the first place.

johnduffell:
including a whole week in the results, looks like it's gone approx 120G/h to 80 G/h, saving around 1TB per day - however this also includes the impact of https://github.com/guardian/frontend/pull/12728:
image

johnduffell:
Actually, although we did save all the bandwidth, around half was down to turning off the server side testing framework, but I hadn't done the analysis and I did both at a similar time.  So it's good that we saved it but we can't give all the credit to etags! For graphs etc, see https://github.com/guardian/fastly-edge-cache/pull/615

PR #12655, opened 2016-04-22
Podcast layout oddity fix

Calanthe:
## What does this change?

This PR fix the bug with layout oddity on podcast pages.
## Screenshots

Before:
screen shot 2016-04-22 at 13 51 18

After:
screen shot 2016-04-22 at 17 23 15


uplne:
I have a strong feeling that this is not the last time we are changing the height of this :-) 👍

PR #12656, opened 2016-04-22
Email preferences: copy changes and example urls

rrees:
## What does this change?

The copy on the email preferences page now reflects the actual email's frequencies and sending patterns. Where that information was duplicated in the description it has been removed.

The Fiver and Opinion have been duplicated so they appear in multiple themes. This is a bit crappy but I didn't want to change the model to allow an email to be in multiple themes without a better reason.

A link to be able to see the current version of the email has been added to give more context as to what the user might be subscribing to.

The previous information was incorrect or missing and therefore not useful for users trying to manage their email settings.

The changes make the page more valuable as a landing page for promotional activity.

No clear metrics are associated with this change.

This is purely on-platform.
## Screenshots

email-prefs-links

### DIFF
--- a/identity/app/model/EmailSubscriptions.scala
+++ b/identity/app/model/EmailSubscriptions.scala
@@ -12,10 +12,31 @@ case class EmailSubscription(
   subscribedTo: Boolean = false,
   exampleUrl: Option[String] = None
 )
+
 object EmailSubscription {
   def apply(emailSubscription: EmailSubscription) = emailSubscription
 }
 object EmailSubscriptions {
+  val theFiver = EmailSubscription(
+    "The Fiver",
+    "sport",
+    "Football",
+    "The Fiver is theguardian.com/sport's free football email. Every weekday we round up the day's news and gossip in our own belligerent, sometimes intelligent and — very occasionally — funny way. The Fiver is delivered every Monday to Friday at around 5pm — hence the name.",
+    "Weekday afternoons",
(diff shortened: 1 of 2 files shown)


gtrufitt:
👍 Hooray for example URLs!

identity/app/model/EmailSubscriptions.scala
```diff
@@ -220,19 +228,21 @@ object EmailSubscriptions {
...
       "news",
       "Environment",
-      "In each weekly edition our editors highlight the most important stories of the week including data, opinion pieces and background guides. We'll also flag up our best video, picture galleries, podcasts, blogs and green living guides.",
-      "",
+      "In each weekly edition our editors highlight the most important environment stories of the week including data, opinion pieces and background guides. We'll also flag up our best video, picture galleries, podcasts, blogs and green living guides.",
+      "Every Friday",
       "38",
-      subscribedTo = subscribedListIds.exists{ x => x == "38" }
+      subscribedTo = subscribedListIds.exists{ x => x == "38" },
+      exampleUrl = Some("http://www.theguardian.com/environment/series/green-light/latest/email")
```
gtrufitt: Should we use [guardian.page.host](https://github.com/guardian/frontend/blob/5e03ef5dbbd7d1a49f598c3e2b3761690b505b37/common/conf/env/PROD.properties#L52) for these?

PR #12658, opened 2016-04-25
Supported by articles in Paid for by containers

regiskuckaertz:
This small update to accommodate the rare but possible case where a 'supported by' article is linked inside a paid for container along with other 'paid for by' articles.

Before:
picture 2

After:
picture 1

cc @uplne I have a list of styles in adverts.scss that will need to be removed when you refactor badges 🏇

### DIFF
--- a/static/src/stylesheets/module/commercial/_adverts-capi.scss
+++ b/static/src/stylesheets/module/commercial/_adverts-capi.scss
@@ -122,6 +122,10 @@
             min-height: $gs-baseline * 11;
             width: 100%;
         }
+
+        .advert__standfirst {
+            margin-bottom: $gs-baseline / 2;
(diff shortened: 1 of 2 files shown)


kelvin-chappell:
Wouldn't this case break the business rules?  I thought paid containers should only have paid content.  In which case it would be better to filter out non-paid content before rendering the container.

regiskuckaertz:
We thought so too, but have received confirmation that it may happen. Tim Hughes:

> I could see that happening, actually. For example, with Nespresso we ran parallel paid-for and supported-by campaigns, one promoting the coffee and the other articles around its sustainability, and those could have been mixed together.

Calanthe:
👍

PR #12659, opened 2016-04-25
Fix discussion email status code

nicl:
Ensures users see a sensible error message when they comment and their email is not verified by using the correct Discussion code.


gidsg:
:+1: :mega:

PR #12660, opened 2016-04-25
Remove Firefox-specific Adblocker detection

jbreckmckye:
Could detecting an adblocker on a user's computer breach EU data privacy laws? [This campaigner](https://twitter.com/alexanderhanff/status/722506381451010048) says so, and he thinks he has the case to prove it. The Guardian is seeking advice on the matter, but for now, we're taking the precaution of 'genericizing' our adblocking - making sure that it can't spot specific extensions on specific browsers - so that all that remains is a heuristic that an adblocker _might_ be installed because an ad _appears_ to have been hidden.

This pull request removes the Firefox AdblockPlus detection path from the inlined detection script.

@sndrs @Calanthe


Calanthe:
👍  This is interesting! I wonder what will be the advice from the lawyers.

common/app/templates/inlineJS/nonBlocking/detectAdblock.scala.js
```diff
@@ -16,17 +16,14 @@ try {
...
-                window.guardian.adBlockers.generic = adStyles.getPropertyValue('display') === 'none';
-
-                // Only tells us if FF ABP is installed - not whether it is active
-                var adMozBinding = adStyles.getPropertyValue('-moz-binding');
-                window.guardian.adBlockers.ffAdblockPlus = !!adMozBinding && adMozBinding.match('elemhidehit') !== null;
 
+                adBlockers.hiddenAds = adStyles.getPropertyValue('display') === 'none';
                 try {
-                    window.guardian.adBlockers.onDetect(window.guardian.adBlockers);
+                    adBlockers.onDetect(adBlockers.hiddenAds);
```
sndrs: maybe `active`, or just `guardian.adsBlocked = true`? at first, i thought this was a collection of the ads which were hidden

jbreckmckye: The object also carries the onDetect method.

sndrs: good point. `guardian.adBlockers.active` then? not a blocker, i just didn't understand what it was at first, and therefore seems possible someone else won't.

jbreckmckye: No, fair enough

uplne:
So in theory it is also illegal to detect if you are using IE9 and thus give you different CSS styles?

rich-nguyen:
@uplne exactly, this line of thinking just doesn't make sense. If we include the browser's capabilities in an interpretation of 'user information', then all responsive websites are breaching the law. Even implicit feature detection would be outlawed.

PR #12662, opened 2016-04-25
fix styling of span

gidsg:
## What does this change?

Fix styling of discussion error message
## Screenshots

Before:
screen shot 2016-04-25 at 12 32 58

After:
screen shot 2016-04-25 at 12 33 08
## Request for comment

CC @nicl


nicl:
:+1: :bread: worth removing the spans though as not sure they're needed anymore.

PR #12663, opened 2016-04-25
Extend commercial container refactor switches

kelvin-chappell:
This extends switches to next Wednesday to give chance for any problems to appear before stripping out old code.

And removing dynamic-container-specific code because it's redundant at the moment.
(We can reinstate this code when it's required but it just seems to add complexity to a much cleaner model.)


common/app/views/commercial/containers/paidContainerV2.scala.html
```diff
@@ -39,14 +39,11 @@
...
             case "fixed/large/slow-XIV"          => { @fixedLargeSlowXIV(frontId, containerIndex, containerModel) }
-            case "dynamic/fast"                  => { @dynamicFast(frontId, containerIndex, containerModel) }
-            case "dynamic/slow"                  => { @dynamicSlow(frontId, containerIndex, containerModel) }
-            case "dynamic/package"               => { @dynamicPackage(frontId, containerIndex, containerModel) }
             case _ => {
             }
         }
 
-        @if(containerModel.content.showMoreCards != Nil && containerModel.content.showMoreCards.length > 0){
+        @if(containerModel.content.showMoreCards != Nil && containerModel.content.showMoreCards.nonEmpty){
```
kelvin-chappell: Actually first half of this condition is redundant.

regiskuckaertz:
Great! 👍

PR #12665, opened 2016-04-25
Revert "Fabric v1 - again"

jbreckmckye:
This seems to have caused some problems now that article inlines have the name 'article-inline-X'

@regiskuckaertz


regiskuckaertz:
👍

PR #12666, opened 2016-04-25
Fix resend link when email not validated

nicl:
Fixes email validation resend link when a user has attempted to comment and DAPI has refused with a 403 and error code 'EMAIL_NOT_VALIDATED'.

@gidsg


gidsg:
:+1: :warning: :tokyo_tower:

PR #12668, opened 2016-04-25
Asset metrics dashboard: show size and size change for all js/css files

TBonnin:
## What does this change?

This is the first step of fixing the Asset Metrics admin dashboard
https://frontend.gutools.co.uk/metrics/assets

The dashboard now shows two lists of files (js and css) with each file' size
and size evolution during the time period (14 days)

The 2nd step objective would be to show a graph when a file is clicked
that would give a more detailled overview of this file size evolution

Tracking the size of all js/css files overtime

No
## Screenshots

screen shot 2016-04-25 at 15 53 32
## Request for comment

@sndrs @johnduffell

### DIFF
--- a/admin/app/model/AdminLifecycle.scala
+++ b/admin/app/model/AdminLifecycle.scala
@@ -10,7 +10,7 @@ import football.feed.MatchDayRecorder
 import jobs._
 import play.api.GlobalSettings
 import services.EmailService
-import tools.{CloudWatch, LoadBalancer}
+import tools.{AssetMetricsCache, CloudWatch, LoadBalancer}
 
 import scala.concurrent.Future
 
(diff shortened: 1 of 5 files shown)


admin/app/model/AdminLifecycle.scala
```diff
@@ -93,6 +93,10 @@ trait AdminLifecycle extends GlobalSettings with Logging {
       VideoEncodingsJob.run()
     }
 
+    Jobs.scheduleEveryNMinutes("AssetMetricsCache", 60 * 6) {
```
TBonnin: Fetching the metrics from Cloudwatch is done via a schedule job in order to avoid unnecessary requests that cost money.
I set the cron job to run every 6 hours because it is unlikely we would like to track those stats at a smaller granularity but I am happy to discuss about this value

admin/app/tools/charts/charts.scala
```diff
@@ -128,11 +128,12 @@ class AwsLineChart(
       .getOrElse(throw new IllegalStateException(s"Don't know how to get a value for $dataPoint"))
 
   protected def toLabel(date: DateTime): String = date.withZone(format.timezone).toString("HH:mm")
+
+  lazy val latest = dataset.lastOption.flatMap(_.values.headOption).getOrElse(0.0)
```
TBonnin: Moving `latest` so it is also available for `AwsLineChart` and not just for `AwsDailyLineChart`

sndrs:
👍

PR #12669, opened 2016-04-25
Remove the `retry-failed-accuweather-requests` switch

TBonnin:
## What does this change?

Lowering the timeout value of the requests to accuweather API and
retrying in case of timeout exceptions successfully resulted in lower
errors.
This patch is removing the flag and make the retrying permanent

No
## Request for comment

@johnduffell @jamespamplin


NataliaLKB:
👍

PR #12671, opened 2016-04-25
Report empty dfp ad responses

rich-nguyen:
When dfp responds with an empty advert response, report the issue.


JonNorman:
:+1: Looks great! A far more robust way of spotting targeting issues than just browsing the site(!)

PR #12673, opened 2016-04-25
Remove A/B dependency from enhanced/common bootstrap

desbo:
A minor thing. We do the A/B bootstrapping in standard so it's not used here.


TBonnin:
👍

PR #12674, opened 2016-04-25
Revert "Fix resend link when email not validated"

nicl:
Reverts guardian/frontend#12666

It broke commenting :(


jbreckmckye:
👍

PR #12677, opened 2016-04-26
Split brain

piuccio:
## What does this change?

Fronts tool lives in cmsFronts account but uses frontend S3 as data storage.

At the time of writing, fronts tool reads the master config from frontend and writes in both buckets. Ideally it would only write in cmsFronts.

This PR introduces a switch to use cmsFronts bucket as master config.

Changes:
- all reads (facia, press, admin ...) go through `ConfigAgent` which chooses the account / bucket depending on a feature switch.
- press writes with `S3FrontsApi` which always writes into frontend account

I sleep better at night

It affects most developers, they'll need cmsFronts credentials through Janus

It depends on cloudformation and platform updates
## Request for comment

@janua @johnduffell @anyone_in_the_team?

### DIFF
--- a/admin/app/jobs/RefreshFrontsJob.scala
+++ b/admin/app/jobs/RefreshFrontsJob.scala
@@ -1,10 +1,9 @@
 package jobs
 
-import common.Logging
-import services.{S3FrontsApi, FrontPressNotification}
-import play.api.libs.json.{JsString, JsObject, JsValue, Json}
-import conf.switches.Switches._
+import com.gu.facia.api.models.{CommercialPriority, EditorialPriority, TrainingPriority}
+import common.{ExecutionContexts, Logging}
 import conf.Configuration
+import services.{ConfigAgent, FrontPressNotification}
 
 sealed trait FrontType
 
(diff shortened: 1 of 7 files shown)


common/app/common/configuration.scala
```diff
@@ -398,6 +402,32 @@ class GuardianConfiguration(val application: String, val webappConfDirectory: St
...
+      )
+
+      // this is a bit of a convoluted way to check whether we actually have credentials.
+      // I guess in an ideal world there would be some sort of isConfigued() method...
+      try {
+        val creds = provider.getCredentials
+        Some(provider)
+      } catch {
+        case ex: AmazonClientException =>
+          log.error("amazon client cross account exception")
```
johnduffell: `s"amazon client cross account exception $ex"` might be nice?

admin/app/jobs/RefreshFrontsJob.scala
```diff
@@ -63,8 +54,7 @@ object RefreshFrontsJob extends Logging {
   def runAll(): Option[Seq[Unit]] = {
     Configuration.aws.frontPressSns.map(Function.const {
       log.info("Putting press jobs on Facia Cron (MANUAL REQUEST)")
-
-      for {update <- getAllCronUpdates.getOrElse(Nil)}
+      for {update <- getAllCronUpdates}
```
johnduffell: if it's only one line seems neater to use () instead of {} - I think the {} only changes the behaviour of multi line to not need `;` - but that's only a minor comment 😛

common/app/common/configuration.scala
piuccio: valid point, I've changed to `log.error("...", ex)`

johnduffell:
are you force pushing an amended commit? 😒

admin/app/jobs/RefreshFrontsJob.scala
piuccio: In scala I tend to go for the policy of least amount of changes possible 👅  But I've replaced curly with round.

johnduffell:
look good to me! 👍 😴 sleep well 💤

piuccio:
> are you force pushing an amended commit?

yes, I hate multiple commits, they make history hard to read. Is it confusing from a review point of view?

johnduffell:
yeah I suppose you're right there!  although if that, I think you can squash when you merge using the drop down on the merge button in github!  I was just trying to find the new commits since I already reviewed the rest and getting completely confused 😜

common/app/common/configuration.scala
```diff
@@ -398,6 +402,32 @@ class GuardianConfiguration(val application: String, val webappConfDirectory: St
       Try(configuration.getStringProperty("admin.pressjob.low.push.rate.inminutes").get.toInt)
         .getOrElse(60)
 
+    lazy val stsRoleToAssume = configuration.getStringProperty("frontpress.role")
```
janua: Can we name this something better? It's not very clear what this is.

common/app/common/configuration.scala
```diff
@@ -377,6 +377,10 @@ class GuardianConfiguration(val application: String, val webappConfDirectory: St
   }
 
   object faciatool {
+    lazy val crossAccountSourceBucket = configuration.getMandatoryStringProperty("frontpress.buckets.source")
+    lazy val sameAccountSourceBucket = configuration.getMandatoryStringProperty("aws.bucket")
```
janua: I find the naming here a bit weird.
`sameAccountSourceBucket` ??

And a property called `frontpress.buckets.source`? Maybe we should refer to account names in either properties or values or both.

gtrufitt:
> I sleep better at night

😴 Pretty much the most important reason to make a change.

PR #12680, opened 2016-04-26
Insulated Soulmates feed parsing against malformed JSON items

JonNorman:
## What does this change?

We had a case of a soulmates member having an age of "n/a", which broke the parsing of the feed. This PR makes the parsing of the feed more "best effort"; if the parsing of one of the `Member` objects fails, then it is thrown away and we continue in trying to parse the rest of the feed.

No more babies will be thrown out with bath waters when we get a malformed `Member` in our soulmates feed, we will just ignore it.

I highlighted the problem and the feed has since been fixed; I've added to the test suite, which verifies this behaviour.

No


JonNorman:
A better way of parsing soulmates appears...#12685

PR #12683, opened 2016-04-26
Re-enable sentry logging

rich-nguyen:
The switch had been removed, but the raven configuration was still referencing it.


kelvin-chappell:
👍

OliverJAsh:
Nice catch, wondered why we had so few errors!

PR #12684, opened 2016-04-26
Ensure that adKeywords don't join an undefined array

rich-nguyen:
I introduced a bug with dfp response reporting. It was joining an undefined array.


JonNorman:
Let the errors flow!
:+1:

PR #12685, opened 2016-04-26
Different way of parsing soulmates

kelvin-chappell:
This is a different approach to dealing with exceptions in the soulmates feed.

/cc @JonNorman

### DIFF
--- a/commercial/app/model/commercial/soulmates/Member.scala
+++ b/commercial/app/model/commercial/soulmates/Member.scala
@@ -1,5 +1,9 @@
 package model.commercial.soulmates
 
+import play.api.data.validation.ValidationError
+import play.api.libs.functional.syntax._
+import play.api.libs.json._
+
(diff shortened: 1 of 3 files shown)


JonNorman:
Awesome! This is much nicer than my approach to doing it. :+1:

When I went to use `asOpt`, it didn't seem to work as instead of getting a `None` I was still getting the exception (and thus `flatMap` didn't help)...but this is great, I'll close the PR I raised :)

JonNorman:
Just one small additional comment: is there a good reason that the SoulmatesFeedTest is `@DoNotDiscover` and `with ConfigredTestSuite`?

kelvin-chappell:
@JonNorman If you ask @rich-nguyen he'll be able to explain the test suite business.

rich-nguyen:
looks great, just what me and @JonNorman were going to do 👍

PR #12686, opened 2016-04-26
Update password copy for reset password

jranks123:
. The copy was not updated when we changed the new max password length
for reset password so this change updates the copy accordingly


rich-nguyen:
👍

PR #12687, opened 2016-04-26
Refactor supported by containers

regiskuckaertz:
The cAPI commercial components support both paid for and supported by sponsorships, which I was not aware of.

/cc @kelvin-chappell @JonNorman @uplne @Calanthe @jbreckmckye


commercial/app/controllers/commercial/ContentApiOffersController.scala
```diff
@@ -17,6 +17,12 @@ object ContentApiOffersController extends Controller with ExecutionContexts with
     "foundation-supported" -> "fc-container--foundation-supported"
   )
 
+  private val sponsorTypeToClassRefactor = Map(
```
JonNorman: We could use a `case object` here as with our Switches i.e. `On` and `Off`. It would abstract away from string matching, which I think would be a good thing...

e.g.
```
sealed trait SponsorType
case object PaidFor extends SponsorType { override def toString = "paidfor"}
case object Sponsored extends SponsorType { override def toString = "supported"}
case object Foundation extends SponsorType { override def toString = "supported"}
```

we can then refer to `SponsorType.PaidFor` etc

static/src/stylesheets/module/commercial/_adverts-capi.scss
```diff
@@ -70,10 +72,49 @@
...
+    &::before {
+        content: '';
+        position: absolute;
+        top: 0;
+        left: 0;
+        right: 0;
+        border-top: 1px solid $news-main-2;
+
+        @include mq(tablet) {
+            left: calc(50% - 370px);
```
jbreckmckye: I'm assuming the idea is to horizontally center something with a 50% width. Could something like the following handle all viewport sizes?

````
left: 0;
width: 50%;
transform: translateX(50%);
````

File: static/src/stylesheets/module/commercial/_adverts-capi.scss [hunk 3]
```diff
@@ -70,10 +72,49 @@
...
+            right: calc(50% - 650px);
+        }
+    }
+
+    .adverts__logo {
+        color: $guardian-brand-dark;
+    }
+
+    .adverts__row > * + *::before {
+        background: none;
```
jbreckmckye: What does this do?

static/src/stylesheets/module/commercial/_adverts-capi.scss
regiskuckaertz: Alas, no 😢  The top border should span the container's content box but there is no CSS to do that unfortunately, and since I'm using padding to center the content horizontally I have to make this small sacrifice.

static/src/stylesheets/module/commercial/_adverts-capi.scss
regiskuckaertz: Removing the separator. It's not clear, I should `content: none` instead and document this line.

commercial/app/controllers/commercial/ContentApiOffersController.scala
regiskuckaertz: Nice! I did it 🚬

kelvin-chappell: I like this typing, but rather than overriding `toString` maybe create a new property to make it more explicit what it's for.

regiskuckaertz: Will do

static/src/stylesheets/module/commercial/_adverts-capi.scss
regiskuckaertz: Now that I come think of it, I can use this technique and remove half the properties in the media queries. Sweet 🏄

commercial/app/views/contentapi/itemV2.scala.html
```diff
@@ -67,7 +77,7 @@
     <a class="adverts__badge__link" href="@clickMacro@optCapiLink" data-link-name="logo link">
         @for(logoUrl <- optLogo) {<img class="adverts__badge__logo" src="@logoUrl" alt="">}
     </a>
-    @for(sponsorType <- optSponsorType if sponsorType != "paidfor";
+    @for(sponsorType <- optSponsorType if sponsorType != PaidFor;
```
kelvin-chappell: Don't need semicolon.  

```
@for {
       ...
       ...
    } {
      ...
   }
```

kelvin-chappell:
Apart from minor comments, 👍

static/src/stylesheets/module/commercial/_adverts-capi.scss
uplne: Just keep in mind that calc() is not supported on opera mini and some older android versions. Opera mini seems to have almost 5% global usage. If anything I would check with QA to get the proper number as https://frontend.gutools.co.uk/analytics/quality shows only up to 600k visitors.

commercial/app/controllers/commercial/ContentApiOffersController.scala
kelvin-chappell: Looks good 👍

PR #12688, opened 2016-04-26
Onward video series

jamesgorrie:
## What does this change?

We're looking to see if people want to see more series.
We know this wasn't the case for certain content, especially articles, but we think it might be for video.

We will be measuring whether we have more people use the related content component in it's series-ified version. @harrysalmon

No
## Screenshots

more-from-desktop
more-from-mini
## Request for comment

@akash1810


akash1810:
Can't comment inline as on mobile, but..

Do we need to `showFields(all)`? Could we just ask for what we're rendering?

jamesgorrie:
> Do we need to showFields(all)
I'll refine the query once we've done the test, it doesn't add much overhead.

onward/app/controllers/MostViewedVideoController.scala
```diff
@@ -1,12 +1,39 @@
...
+    val edition = Edition(request)
+    getResponse(ContentApiClient.search(edition)
+      .tag(series)
+      .showTags("all")
+      .showFields("all")
+    ).map { response =>
+      val videos = response.results.toList.map(apiContent => {
+        val content = Content.make(apiContent)
+        Video.make(content)
+      }).take(6)
```
akash1810: could we use capi's page-size param?

dev-build/conf/routes
```diff
@@ -114,6 +114,7 @@ GET            /tagged.json
 GET            /:mediaType/section/:sectionId/*seriesId.json                                                                     controllers.MediaInSectionController.renderSectionMediaWithSeries(mediaType, sectionId, seriesId)
 GET            /:mediaType/section/:sectionId.json                                                                               controllers.MediaInSectionController.renderSectionMedia(mediaType, sectionId)
 GET            /video/most-viewed.json                                                                                           controllers.MostViewedVideoController.renderMostViewed()
+GET            /video/in-series/*series.json                                                                                     controllers.MostViewedVideoController.renderInSeries(series: String)
```
akash1810: do the other `routes` need this path added to too?

jamesgorrie: Good point - done

akash1810:
👍

onward/app/controllers/MostViewedVideoController.scala
jamesgorrie: Yip - good ol Copy / Paste + nothinking

PR #12690, opened 2016-04-26
Fix media query for high DPR images

OliverJAsh:
We are currently serving desktop images to all high DPR devices, regardless of screen width. This is because the `min-width` in our media query is not being taken into effect.

As an example, on a high DPR device of any screen width, the following media query returns `true`:

``` js
window.matchMedia('(min-width: 980px) and (-webkit-min-device-pixel-ratio: 1.25), (min-resolution: 120dpi)').matches
// => true
```

Our intention is that this media query should only return true if the screen is 980px or wider _and resolution is deemed high DPR_, but this is not happening.

That is because the precedence is wrong: the media query is interpreted as `(x and y) or z` where `x` is our `min-width`, when we want it to be interpreted as `x and (y or z)`.

It seems you can't use parentheses to control precedence in media queries, so I've worked around this by duplicating the `min-width` media query, resulting in a media query of the form `(x and y) or (x and z)`.
# Before

image

`w=620&dpr=2` resulting in an image 1240px wide (too big)
# After

image

`w=465&dpr=2` resulting in an image 930px wide (correct)

/cc @paperboyo @sndrs

### DIFF
--- a/applications/app/views/fragments/newGalleryBody.scala.html
+++ b/applications/app/views/fragments/newGalleryBody.scala.html
@@ -81,7 +81,7 @@
                         GalleryMedia.inline
                     }) { widths =>
                         @widths.breakpoints.map { breakpointWidth =>
-                            <source media="(min-width: @breakpointWidth.breakpoint.minWidth.getOrElse("0")px) and (-webkit-min-device-pixel-ratio: 1.25), (min-resolution: 120dpi)"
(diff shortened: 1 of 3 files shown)


sndrs:
👍 great spot

rich-nguyen:
👍  nice one

paperboyo:
Great detective work! :clap:

janua:
What size difference is in the file between the two?

OliverJAsh:
@janua In the example I gave above, the difference is 42291-28939=13352 (bytes). Of course it will be different throughout the site, as this picture element is used everywhere.

janua:
:clap:

OliverJAsh:
Did a quick test on the UK front. The total size of all downloaded images on an iPhone 5 (high DPR):

Before
787KB

After
649KB

PR #12693, opened 2016-04-26
Gustyle commercial template _blank fixes

Calanthe:
## What does this change?

All of the links in the gustyle commercial template should be opened in the new tab


uplne:
👍

PR #12695, opened 2016-04-26
Fabric expandables should be based on 'expandable-v-n' templates, not expandable videos

jbreckmckye:
Switches the implementation of _fabric-expanding-v1_ to that of _expandable-video-v2_.


jbreckmckye:
The strategy for our ads templates seems to be to fork existing code whenever we want to make changes. This allows the ads team to jerry-rig a kind of versioned dependency system. I wonder if the templating system @regiskuckaertz is proposing might be a better way of implementing this.

regiskuckaertz:
Let's do that! 👍 

Yes, the problem with templates in JS is there is an incentive to avoid redundancy and abstract the code as much as possible (a mistake I made with fluid250). That is not how it should be. Each creative should have its own markup, style and logic; we can think about bundling common styles/scripts working backwards.

jbreckmckye:
Indeed. I remember feeling quite squeamish about copy-pasting the old templates, until I reasoned that I was mis-applying the principle of DRY. "Don't Repeat Yourself" is about de-duplicating _sources of knowledge_, not implementation. Each template is its _own_ source of knowledge about its _own_ creative type.

PR #12696, opened 2016-04-26
Refactor slice adverts for the new Fabric creatives

regiskuckaertz:
This complete refactoring of slice-adverts abstracts the selection of containers/slices from the insertion of ad slots from the nature of ad slots. It makes it more easy to factor in the new **fabric** creative and provide a fallback in case it does not load.

cc @jbreckmckye @rich-nguyen


static/src/javascripts/projects/common/modules/commercial/slice-adverts.js
```diff
@@ -1,120 +1,118 @@
...
-    'common/utils/fastdom-idle',
+    'common/utils/fastdom-promise',
+    'common/modules/commercial/track-ad',
     'common/modules/commercial/create-ad-slot',
+    'common/modules/commercial/dfp/dfp-api',
     'common/modules/user-prefs',
-    'common/modules/commercial/commercial-features',
-    'lodash/objects/defaults',
-    'lodash/collections/contains',
-    'lodash/collections/map',
```
uplne: nice!

static/src/javascripts/projects/common/modules/commercial/slice-adverts.js
```diff
@@ -1,120 +1,118 @@
...
-                    }
+                    return true;
                 }
-            }
 
-            return Promise.all(chain(adSlices).slice(0, maxAdsToShow).and(map, function ($adSlice, index) {
-                    // When we are inside the AB test we are adding inline1 manually so index needs to start from 2.
-                    var inlineIndexOffset = (config.tests.cmTopBannerPosition) ? 2 : 1;
+                // Slots are inserted every containerGap + 1 slice
+                if (!item.adSlice || index % (containerGap + 1) !== 0) {
```
jbreckmckye: This isn't quite the same as the original behaviour. Here, we're taking all odd-numbered containers that have advert candidates. The original would find slots with candidates, then skip scanning until `containerGap` items afterwards.

What this means (unless I've missed a subtlety of the logic) is that if only even-numbered containers had slice-candidates, the original code would display those adverts, but our new code would not display any of them.

e.g.
````
Index     Container     Old     New     What stops
          has slice?                    ad?

0         X             X       X       !item.adSlice
1         √             √       X       (1 % 2) !== 0
2         X             X       X       !item.adSlice
3         √             √       X       (3 % 2) !== 0
````

File: static/src/javascripts/projects/common/modules/commercial/slice-adverts.js [hunk 3]
```diff
@@ -1,120 +1,118 @@
...
+                    return false;
+                }
+
+                var containerId  = item.container.getAttribute('data-id');
+                var isContainerClosed = prefs[containerId] === 'closed';
+                var isFrontFirst = isNetworkFront && index === 0;
+                return !(isFrontFirst || isContainerClosed);
```
jbreckmckye: I know there's a little more overhead to running more filter functions, but would these benefit from being separated into their own distinct filters?

static/src/javascripts/projects/common/modules/commercial/slice-adverts.js
```diff
@@ -1,120 +1,118 @@
...
+            })
+            // limit to maxAdsToShow
+            .slice(0, maxAdsToShow)
+            // create ad slots for the selected slices
```
jbreckmckye: These comments are far more readable than trying to name the lambdas.

static/src/javascripts/projects/common/modules/commercial/slice-adverts.js
```diff
@@ -1,120 +1,120 @@
...
+            .map(function (item, index) {
+                // When we are inside the AB test we are adding inline1 manually so index needs to start from 2.
+                var inlineIndexOffset = (config.tests.cmTopBannerPosition) ? 2 : 1;
```
jbreckmckye: If Steve can merge his A/B test change, we can get rid of this

static/src/javascripts/projects/common/modules/commercial/slice-adverts.js
```diff
@@ -1,120 +1,120 @@
...
+                var adName = 'inline' + (index + inlineIndexOffset);
+                var adSlot = hasFabricAd && index === 0 ?
+                    (fabricAdSlot = createAdSlot('fabric', 'container-inline')) :
```
jbreckmckye: How do we feel about using assignments as expression values? It does make things concise, but once the code grows, does it bury the logic?

jbreckmckye:
> `8e490a : Use function declaration syntax for God's sake`

Yes!

static/src/javascripts/projects/common/modules/commercial/slice-adverts.js
```diff
@@ -1,120 +1,120 @@
...
+                return {
+                    anchor: isMobile ? item.container : item.adSlice,
+                    adSlot: adSlot
+                };
+            });
 
-                            resolve(null);
-                        });
+        if (hasFabricAd) {
+            // if there is no fabric creative, we fall back to loading
```
jbreckmckye: Do we actually need this?

If we want the first container to show any advert if a fabric is not available, the best way to do that would be to give the fabric ads higher-priority targeting, or to deliver them as part of a roadblock.

On the other hand, if we don't want the first container to show any other advert (which was Carlo's view), then we either have to insert the slot in the second container (if it's a network front) and accept some adjacent ads, or leave the first two containers empty (which means no yield).

Personally, I think the only way we can do this is to allow the top slot to support «any» ad type. I can ask Patrick to see if he's okay with this.

static/src/javascripts/projects/common/modules/commercial/slice-adverts.js
```diff
@@ -1,120 +1,120 @@
...
+            adSlots.forEach(isMobile ? insertOnMobile : insertOnTabletPlus);
+        });
+
+        function insertOnMobile(item) {
+            // add a mobile advert after the container
+            item.anchor.parentNode.insertBefore(item.adSlot, item.anchor.nextSibling);
+        }
+
+        function insertOnTabletPlus(item) {
+            // add a tablet+ ad to the slice
```
jbreckmckye: Nice

static/src/javascripts/projects/common/modules/commercial/track-ad.js
```diff
@@ -0,0 +1,30 @@
+define([
+    'Promise',
+    'common/utils/mediator'
+], function(Promise, mediator) {
+    return trackAd;
+
+    function trackAd(id) {
```
jbreckmckye: I'd forgotten about this util. Nice to have it as a module.

static/src/javascripts/projects/common/modules/commercial/slice-adverts.js
regiskuckaertz: Oh yes, I was just waiting for him to do it 😄

static/src/javascripts/projects/common/modules/commercial/track-ad.js
regiskuckaertz: Yeah I failed to do that by trying to be too smart in a previous PR, I think this version is fine. The weak link is those events, `dfp-api` should ultimately keep track of the state of each slot—currently they're removed and forgotten as soon as the ad is rendered.

static/src/javascripts/projects/common/modules/commercial/slice-adverts.js
regiskuckaertz: Good catch!! Ouch, I just need to amend the filter, this is an easy fix 😅

static/src/javascripts/projects/common/modules/commercial/slice-adverts.js
regiskuckaertz: I could split in two filters indeed, i.e. from `A ∨ (B ∧ C)` to `(A ∨ B) ∧ (A ∨ C)`

Where `A` is `showMpuInAllContainers`. Let's see how this look.

regiskuckaertz: The problem is `index` would lose some its semantics and `isFrontFirst` not correct anymore.

jbreckmckye: Fair enough!

static/src/javascripts/projects/common/modules/commercial/slice-adverts.js
jbreckmckye: (Speaking to Patrick, it seems we're happy with the fabric slot supporting any inline types in addition to Fabric)

regiskuckaertz:
@jbreckmckye can you please have another look?

static/src/javascripts/projects/common/modules/commercial/slice-adverts.js
```diff
@@ -1,117 +1,103 @@
...
-                    adSlices.push($adSlice.first());
-                    index++;
-                } else {
-                    if ($adSlice.length && !isFrontFirst && (!prefs || prefs[containerId] !== 'closed')) {
-                        adSlices.push($adSlice.first());
-                        index += (containerGap + 1);
-                    } else {
-                        index++;
-                    }
+                    return true;
```
jbreckmckye: The original code, in this case, only pushes the slices it finds in each container. This pushes all of the containers themselves. If the switch is on and some of the containers lack ad candidates, could the call to `anchor.nextSibling` below occasionally throw an exception?

static/src/javascripts/projects/common/modules/commercial/slice-adverts.js
```diff
@@ -1,117 +1,103 @@
...
-                    var adName        = 'inline' + (index + 1),
-                        $mobileAdSlot = bonzo(createAdSlot(adName, 'container-inline'))
-                            .addClass('ad-slot--mobile'),
-                        $tabletAdSlot = bonzo(createAdSlot(adName, 'container-inline'))
-                            .addClass('ad-slot--not-mobile');
+                var containerId  = item.container.getAttribute('data-id');
+                var isContainerClosed = prefs[containerId] === 'closed';
+                var isFrontFirst = isNetworkFront && index === 0;
+                var isFarEnough = lastIndex === -1 || index - lastIndex > containerGap;
+                if (!item.adSlice || isFrontFirst || isContainerClosed || !isFarEnough) {
```
jbreckmckye: Nice

static/src/javascripts/projects/common/modules/commercial/slice-adverts.js
regiskuckaertz: Good question! `nextSibling` can be `null` indeed, but the great thing about `insertBefore` is that it falls back to `appendChild` is the second parameter is `null`, so no exception.

jbreckmckye:
Great stuff - 👍

PR #12698, opened 2016-04-27
use isInVarient properly

jamesgorrie:
I'm going to have a think about whether it is just me who is stupid, or if there is something to make the ab framework a little more clear.


akash1810:
👍 

oops, apologies for missing this!

PR #12699, opened 2016-04-27
Clean targeted names before creating PaidForTag

rich-nguyen:
some target names have unicode characters that break the capi lookup

### DIFF
--- a/common/app/common/dfp/DfpData.scala
+++ b/common/app/common/dfp/DfpData.scala
@@ -2,6 +2,7 @@ package common.dfp
 
 import common.Edition
 import common.dfp.AdSize.{leaderboardSize, responsiveSize}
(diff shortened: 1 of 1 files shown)


JonNorman:
:+1:

kelvin-chappell:
Well spotted! This is something I've never seen before. 

How come it isn't affecting ad targeting generally?

rich-nguyen:
Most ad targeting has key values that are fairly sanitised. In this case, the new CAPI tag refererred to a dfp keyword which reused an old key value pair. There isn't a way in DFP to delete the key-value history, and keywords are free-form, so there's always the possibility of introducing a subtle problem. We could do better with our diagnosis tools, because the paid-for admin page just says `Targeted Tag does not exist`. Which is usually enough.

PR #12701, opened 2016-04-27
Add correct class names to containers and items

regiskuckaertz:
Fixes a bug where the paid for containers did not have the correct class names following the changes in #12687

cc @JonNorman


JonNorman:
If the new `ContainerModel` is only used for `PaidFor` containers then this should be fine, but would have thought we should interrogate the `containerModel.branding` to see whether a container is paid-for, sponsored or foundation-sponsored etc...

:+1:

PR #12702, opened 2016-04-27
Add fake series tests

desbo:
A couple of tests that don't actually do anything, but cover 100% of our audience and only run on series pages. The only difference is whether they should run on sensitive content or not. 

We'll use them see how well the current A/B framework supports queries about visitor habits.


gtrufitt:
👍

PR #12703, opened 2016-04-27
Small fix: color on the badge help link

regiskuckaertz:
cc @uplne

### DIFF
--- a/static/src/stylesheets/module/commercial/_adverts-capi.scss
+++ b/static/src/stylesheets/module/commercial/_adverts-capi.scss
@@ -109,6 +109,10 @@
     .adverts__row > * + *::before {
         content: none;
     }
+
+    .adverts__badge__help {
+        color: $neutral-2-contrasted;
+    }
 }
 
 .advert--capi {


static/src/stylesheets/module/commercial/_adverts-capi.scss
```diff
@@ -109,6 +109,10 @@
     .adverts__row > * + *::before {
         content: none;
     }
+
+    .adverts__badge__help {
```
uplne: It's ugly. Isn't there any other way than creating an element of element? There is a new _brandbadge.scss object. Maybe it should be something like that. `.brandbadge__help` then makes sense.

regiskuckaertz: Now that you mention it, we could totally use the `.badge` namespace, which is free at the moment.

regiskuckaertz: But this is part of a badge refactoring, not within this PR actual intent.

regiskuckaertz: (thanks for the advice though 😄 )

uplne: Yes we can. I thought about it but I've chicken out because it could be useful for something else. But yeah, lets go for it.

uplne: You know what I mean...badge is pretty broad term, so do we want to take it for commercial? But I suppose why not.

uplne:
👍

PR #12705, opened 2016-04-27
Fabric expandable video v1

jbreckmckye:
Creates a new fabric template, forked from _expandable-video-v2_, but with its heights coerced to 250px (closed) and 500px (open).

@tijanamiletic @regiskuckaertz


regiskuckaertz:
Yep 👍

PR #12707, opened 2016-04-27
add pagination to more-in-series

jamesgorrie:
## What does this change?

Adding pagination to the test as, without fail, if we went with it, it would have pagination, and adds a lot to the value of what this container might be, so seems a better indication of whether it's useful to our audience.

I know there are a few things that we don't cater for i.e
- What happens when we run out of pages
- Clicking through to a video won't keep your place in the pagination
- etc - it's a week test, so should be okay.

Same as the test before, we will be looking at who goes from a video in a series, to another.

No.
## Screenshots

more-in-pagination
## Request for comment

@akash1810 @blongden73

### DIFF
--- a/onward/app/views/fragments/videosInSeries.scala.html
+++ b/onward/app/views/fragments/videosInSeries.scala.html
@@ -1,4 +1,4 @@
-@(videos: List[model.Video], seriesTitle: Option[String])(implicit request: RequestHeader)
+@(videos: List[model.Video], seriesTitle: Option[String], series: String, nextPage: Option[Int], prevPage: Option[Int])(implicit request: RequestHeader)
 
 @import views.support.Seq2zipWithRowInfo
 
@@ -8,7 +8,7 @@ data-component="most-viewed-video">
     <div class="most-viewed-container__header">
         <h3 class="most-viewed-container__heading u-cf">
             <div class="most-viewed__more-from">More from</div>
-            @seriesTitle.getOrElse("More in this series")
+               @seriesTitle.getOrElse("More in this series")
         </h3>
     </div>
     <ul class="u-cf u-unstyled most-viewed--media most-viewed--video items--media items--video items--playlist" data-link-name="Most viewed video">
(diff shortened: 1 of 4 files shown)


onward/app/views/fragments/videosInSeries.scala.html
```diff
@@ -8,13 +8,26 @@
...
+            <!-- TODO: reverse routing - do we support it? -->
+            <!-- TODO: Buttons here, not really links, but it's easy for styling -->
+        @prevPage.map{ p =>
+            <a class="most-viewed-navigation__link most-viewed-navigation__link--prev" data-page="@p" data-link-name="prev" href="">
+            @fragments.inlineSvg("arrow-left", "icon")
+            </a>
+        }
+        @nextPage.map{ p =>
+            <a class="most-viewed-navigation__link most-viewed-navigation__link--next" data-page="@p" data-link-name="next" href="">
+            @fragments.inlineSvg("chevron-right", "icon")
```
akash1810: I'm assuming `arrow-` and `chevron-` look different? Is this desired here?

onward/app/views/fragments/videosInSeries.scala.html
```diff
@@ -1,4 +1,4 @@
-@(videos: List[model.Video], seriesTitle: Option[String])(implicit request: RequestHeader)
+@(videos: List[model.Video], seriesTitle: Option[String], series: String, nextPage: Option[Int], prevPage: Option[Int])(implicit request: RequestHeader)
```
akash1810: does it make sense for `seriesTitle` to be an `Option[String]` and `series` to be `String`

onward/app/views/fragments/videosInSeries.scala.html
jamesgorrie: You assume the codebase is consistent. They look they same e.g. screenshot.

onward/app/views/fragments/videosInSeries.scala.html
jamesgorrie: Keeping it consistent with the naming elsewhere - this will be refactored once in.

onward/app/views/fragments/videosInSeries.scala.html
akash1810: > e.g. screenshot.

Oh yeah!

akash1810:
👍

PR #12708, opened 2016-04-27
Move metrics into separate static object

Calum-Campbell:
## What does this change?

Metric values where extrapolated outside of trait into separate static object due to a compiler error. Compiler was crashing trying to access private val within the trait. 
## Request for comment

@JonNorman

### DIFF
--- a/commercial/app/commercial/CommercialLifecycle.scala
+++ b/commercial/app/commercial/CommercialLifecycle.scala
@@ -14,6 +14,16 @@ import scala.util.control.NonFatal
 
 import common.AkkaAgent
 
+private [commercial] object CommercialLifecycleMetrics {
+
+  val metricMap = Map(
+    "fetch-failure" -> AkkaAgent(0.0),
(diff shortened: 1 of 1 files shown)


JonNorman:
Nice work :+1:

commercial/app/commercial/CommercialLifecycle.scala
```diff
@@ -23,24 +33,15 @@ trait CommercialLifecycle extends GlobalSettings with Logging with ExecutionCont
...
-    metricMap
+    CommercialLifecycleMetrics.metricMap
       .get(keyName)
       .foreach(agent => agent.send(_ +1.0))
   }
 
-
   private def updateMetrics(): Unit = {
-    metricMap.foreach{
+    CommercialLifecycleMetrics.metricMap.foreach{
```
TBonnin: Just curious: why not move the whole `recordEvent` and `updateMetrics` methods to the `CommercialLifecycleMetrics`, so you keep the map of agents private and not expose it?

PR #12709, opened 2016-04-27
Manual display card styling fixes

Calanthe:
## What does this change?

The manual display card on mobile and tablet breakpoints should look like and behave similar to the current inline slots (in the same breakpoints). This means that the width should be 250px (275px with the grey label) and it should be positioned in the centre. 

//edit
After another discussion with Patrick, we came to conclusion that the card will look better on mobile if it will behave like it was before (full width) and the width will be limited only on the widest (horizontal) mobile breakpoint.
## Screenshots

Before:
screen shot 2016-04-27 at 17 11 08

After:
Mobile (full width):
screen shot 2016-05-05 at 12 13 30

Mobile widest breakpoint (centred and max width = 275px):
screen shot 2016-05-05 at 12 13 44

Tablet (fit the container):
screen shot 2016-05-05 at 12 14 07

Desktop (fit the container):
screen shot 2016-05-05 at 12 14 17


paperboyo:
General PITA says: text inside buttons not vertically aligned.

regiskuckaertz:
I thought this was supposed to be full-bleed 😞

Calanthe:
@regiskuckaertz I updated this PR. Now on mobile is full-bleed as it was before.

@paperboyo Thanks for noticing! This is weird because I just used a standard button styling so the alignment should be correct by default. I made a small change in styles (changing line-height) but I am not sure if overwriting those general styles is the way we should do it.

paperboyo:
@Calanthe Yeah, overriding seems like a bad idea...

regiskuckaertz:
Looks good to me @Calanthe 👍

PR #12712, opened 2016-04-28
Remove travel offers refactor switch

kelvin-chappell:
The first to go!

/cc @regiskuckaertz


regiskuckaertz:
wuhuuuuu 👍 

(please wait for my PR to go through, we have a expired switch that will break tests)

regiskuckaertz:
(I think books, masterclasses, jobs and soulmates are good to go too, I will re-enable capi and manual today)

regiskuckaertz:
I think `travelFragment` and `travelProminent` should go too, no?

kelvin-chappell:
`travelFragment` is part of the blended component, isn't it?

`travelProminent` appears to respond to the `layout` parameter.  If you think that's not used I'll remove it.  I'm pretty sure it's not used.

regiskuckaertz:
oh yeah right. Haha, I forgot about the `layout` parameter 😅  But this is a trivial change in the new code, you can safely remove the `travelProminent` template.

kelvin-chappell:
I went too far in that last commit.  We still need the `layout` parameter.

regiskuckaertz:
So, I will post an update of

```
<div class="adverts__row">
    @offers.map { offer => @travelCard(offer, clickMacro) }
</div>
```

to handle the `layout` parameter for travel and books

PR #12716, opened 2016-04-28
A fabric layer two animation must be not be positioned

rich-nguyen:
This fixes a visual issue where the fabric layer two animation is visible in it's final position before the animation begins. We do need to keep the inline `background-position` in some cases, so it has to be removed conditionally.


jbreckmckye:
Great stuff. With this we should have the full gamut of Fabric template features 👍

PR #12717, opened 2016-04-28
Add prominent layouts to books and travels components

regiskuckaertz:
picture 1
picture 2

@kelvin-chappell


kelvin-chappell:
Got some conflicts now I'm afraid but looks good 👍

PR #12721, opened 2016-04-28
Update the blended component to look more like the legacy version

regiskuckaertz:
The new blended component now looks like the old blended component

picture 3

Only 1000% better!

@kelvin-chappell


kelvin-chappell:
🎆 👍

PR #12706, opened 2016-04-27
New galleries: refactor hero image to improve display

OliverJAsh:
# Before

The header and top ad overlapped the hero image.

image
# After

The header and top ad no longer overlap the hero image.

image
# Notes

All the good stuff happens in https://github.com/guardian/frontend/commit/c34f1cd4cb39eb1898ad4e44c5f4bfd94b13be78. The other commits are small refactors in preparation for this refactor.

This took quite a lot of refactoring to achieve. Our HTML is now structured like so:

```
.immersive-header-content
  .top-ad
  .header
  .main-media
```

The container, `immersive-header-content`, has a `min-height` of `100vh` and uses flexbox. This allows us to flex `main-media` to whatever space remains after `top-ad` and `header`.

To pull this off we've had to pull the `main-media` and content title out of the article and up next to the header. We still have a hidden content title in the article for SEO purposes.

According to [Google’s Structured Data Testing Tool](https://developers.google.com/structured-data/testing-tool/), there appears to be no difference in our structured data (/cc @johnduffell):

image
# To do

In future PRs:
- [ ] Use picture el
- [ ] Apply these immersive styles to other immersive content (articles) and remove styles for the old technique

PR #12722, opened 2016-04-28
Refactor img helper

NataliaLKB:
## What does this change?

There are some places where you don't want to use the whole scala img helper, you just want to use `<picture>` element. This change makes that much easier. 

I will mention that @OliverJAsh 's PR https://github.com/guardian/frontend/pull/12706 can also use this instead of the current `` being used.

Makes the code a bit more dry, and hopefully easier to read (renaming to `imageFigure` and `image` instead of just `img`)


NataliaLKB:
As a side benefit, this PR also reserves the space in Galleries for all images before they load, greatly reducing JANK.

applications/app/views/fragments/galleryHeader.scala.html
```diff
@@ -6,7 +6,7 @@
 <header class="content__head is-fixed-height tonal__head tonal__head--@toneClass(gallery.item)">
     @* Hidden because we visually show this data in the head, but needed
     here for SEO. *@
-    <h1 class="is-hidden" itemprop="headline" class="is-hidden">@Html(gallery.item.trail.headline)</h1>
```
OliverJAsh: Oops, thanks!

common/app/views/fragments/image.scala.html
```diff
@@ -0,0 +1,59 @@
...
+    @picture.largestImage.map{ largestImage =>
+        <div
+        class="u-responsive-ratio@if(isFeature && isShowcase) { u-responsive-ratio--letterbox}"
+            @if(isFeature && isShowcase) {} else {style="padding-bottom: @{"%.2f".format(100 * largestImage.height.toDouble / largestImage.width)}%"}
+        >
+    }
+
+    <picture>
+        @* IE 9 needs this workaround as per https://scottjehl.github.io/picturefill/ *@
+            <!--[if IE 9]><video style="display: none;"><![endif]-->
```
OliverJAsh: Indentation is a bit off here and the IE comment below.

common/app/views/main.scala.html
```diff
@@ -31,8 +31,7 @@
         @defining(
             page.item.content.trail.trailPicture.head
         ) { picture =>
-            <img src="http:@Html(ImgSrc.findNearestSrc(picture, Profile(width = Some(2200))).getOrElse(""))"
-            class="immersive-main-media__img">
+            @fragments.image(picture, imageAltText = "")
```
OliverJAsh: We could use the caption as alt text?

common/app/views/fragments/image.scala.html
```diff
@@ -0,0 +1,59 @@
+@import layout.ContentWidths.MainMedia
+@import views.support.ImgSrc
+
+@(
+    picture: model.ImageMedia,
+    isShowcase: Boolean = false,
+    isFeature: Boolean = false,
+    widths: layout.WidthsByBreakpoint =  MainMedia.inline,
```
OliverJAsh: Galleries main media will currently be using the default `widths` specified here. Do we need to specify our own `widths`?

applications/app/views/fragments/galleryHeader.scala.html
NataliaLKB: 😉

common/app/views/main.scala.html
NataliaLKB: Would this mean the caption would be read twice?

OliverJAsh:
I think there are two(?) other instances of the picture el in our source code. Could we de-dupe those too?

NataliaLKB:
There is one other instance of `<picture>` used by facia cards. The parameters are quite a bit different, which is probably why it wasn't using the img helper to begin with. 

It would add much more complexity to the now simple image helper. I am not sure if its worth it.

common/app/views/fragments/image.scala.html
NataliaLKB: Thanks, this has been addressed 😄

common/app/views/main.scala.html
OliverJAsh: I think we also need a `widths` value here, we'll probably have to create a new one?

NataliaLKB: Yes agreed. I will add that to this PR.

applications/app/views/fragments/newGalleryBody.scala.html
```diff
@@ -74,30 +74,7 @@ <h2 class="fc-container__header__title">
...
-                            sizes="@breakpointWidth.width"
-                            srcset="@ImgSrc.srcsetForBreakpoint(breakpointWidth, widths.breakpoints, maybePath = None, maybeImageMedia = Some(imageElement.images))" />
-                        }
-                    }
-
-                    <!--[if IE 9]></video><![endif]-->
-
-                    <img class="gallery__img gallery__img@if(image.width >= image.height) {--landscape} else {--portrait}"
-                    itemprop="contentUrl"
-                        style="max-width: calc(@{image.width.toFloat/image.height} * 96vh)"
```
OliverJAsh: As discussed, let's make sure the max-width still applies with the new img

common/app/views/main.scala.html
OliverJAsh: Related: I was thinking perhaps we should rename `ImmersiveMedia.immersive` to `ImmersiveMedia.inline`?

OliverJAsh: Actually ignore that!

applications/app/views/fragments/newGalleryBody.scala.html
NataliaLKB: This has been fixed, please take a look when you have time @OliverJAsh

OliverJAsh:
👍

OliverJAsh:
Awesome work 😄

NataliaLKB:
😊

PR #12723, opened 2016-04-28
Add Fabric dependency

jbreckmckye:
We have to specify creative templates as dependencies in breakout-iframe.js, else the _require_ for them in the iframe parser will break in prod - they're not available on any bundle.

This adds the new fabric expandable video.

@regiskuckaertz


regiskuckaertz:
👍

PR #12725, opened 2016-04-28
Fix parallax scrolling for fabric ads

rich-nguyen:
This fixes an issue with fabric content caused by me when I changed the initialisation order of fabric-v1


jbreckmckye:
👍

PR #12727, opened 2016-04-28
Changing optimal image widths for galleries

NataliaLKB:
## What does this change?

For the new gallery design, images appear bigger on most breakpoints. This change means that the correct image width will be available. 

Downside: Images for the old galleries will be requested too big on desktop breakpoints. However, this will only be for a week until Galleries are fully replaced.

cc @OliverJAsh

### DIFF
--- a/common/app/layout/WidthsByBreakpoint.scala
+++ b/common/app/layout/WidthsByBreakpoint.scala
@@ -236,10 +236,13 @@ object ContentWidths {
 
   object GalleryMedia {
     val inline = WidthsByBreakpoint(
-      mobile          = Some(445.px),
-      mobileLandscape = Some(610.px),
-      phablet =         Some(620.px),
-      tablet =          Some(700.px)) // desktop, leftCol, and wide are also 700px
+      mobile          = Some(480.px),
+      mobileLandscape = Some(660.px),
+      phablet =         Some(700.px),
+      tablet =          Some(700.px), // TODO: Change to 480 when new galleries is merged
+      desktop =         Some(720.px),
+      leftCol =         Some(880.px),
+      wide =            Some(1010.px))
 
     val lightbox = WidthsByBreakpoint(
       mobile =          Some(465.px),


OliverJAsh:
👍

common/app/layout/WidthsByBreakpoint.scala
```diff
@@ -236,10 +236,13 @@ object ContentWidths {
...
-      mobileLandscape = Some(610.px),
-      phablet =         Some(620.px),
-      tablet =          Some(700.px)) // desktop, leftCol, and wide are also 700px
+      mobile          = Some(480.px),
+      mobileLandscape = Some(660.px),
+      phablet =         Some(700.px),
+      tablet =          Some(700.px), // TODO: Change to 480 when new galleries is merged
+      desktop =         Some(720.px),
+      leftCol =         Some(880.px),
+      wide =            Some(1010.px))
```
paperboyo: This is displayed at 1040px wide on widest BP at 100% forcing upsizing by the browser which, in turn, reduces quality and introduces interpolation artifacts. Other sizes are also not exactly as they appear on page. BTW, I wonder how this works for portrait images that are now shown fluidly at the viewport height...

OliverJAsh: > BTW, I wonder how this works for portrait images that are now shown fluidly at the viewport height...

It looks pretty bad on a portrait device because the image is downloaded by width rather than height. We can update the value in our `sizes` attribute to fix this, so it downloads by the max of height/width, but it's a bit fiddly: codepen.io/tigt/post/when-responsive-images-get-ugly#height-and-width-constrained-srcset. I got it to work on http://samefourchords.com/ though :-)

paperboyo: So it's almost always scaled. Sometimes up, sometimes down... For fixed size portrait images, @rich-nguyen's plan would work: https://github.com/guardian/frontend/issues/12216#issuecomment-196869603, but for the fluid ones, we would have to have a clever mechanism, similar to what @gtrufitt mentioned re:immersives (https://github.com/guardian/frontend/pull/12775#issuecomment-216855041), so that only _meaningful_ viewport height changes initiates a new request... That's quite a bit of work, no? (I will be 👀 , though).

What about the simple case of landscapes shown at different width to requests - maybe at least this can be fixed before the rollout?

P.s. I like how it works at https://samefourchords.com/! But there are too many new requests for my liking.

too-many-requests

I would adjust the logic, so that the smaller images (for when the viewport's height shrinks) are not new requests, but downsized, already downloaded ones. I would also initiate a higher res uploads at looser steps, maybe. As a last point I would love for the page (image in view) to always stay put and not disappear from view when you change viewport height (orientation). See?! Now I'm commenting on your private projects , oh gawd! 🙀  Zakopane's cool, if a tad too touristy...

paperboyo: Unnecessarily downloading lower resolution sources from the set when the viewport shrinks (instead of just shrinking already downloaded ones) looks like a bug (or bad policy decision) on Firefox’s part. All other browsers I tested do not trigger extra downloads (test page https://samefourchords.com/2014/11/01/zakopane-poland):

## Chrome 51:
chrome 51
---
## Firefox 47 (same on 49 :sob:):
firefox 47
---
I couldn’t find a related bugzilla entry or anything at all TBH, just someone having the same problem: https://www.sitepoint.com/community/t/img-srcset-firefox-loads-too-many-images/219162...

One could argue that lower resolution images may look better at native res not having to undergo resizing and not exhibiting related interpolation artifacts, but one would be wrong, of course.

paperboyo:
Unnecessarily downloading lower resolution sources from the set when the viewport shrinks (instead of just shrinking already downloaded ones) looks like a bug (or bad policy decision) on Firefox’s part. All other browsers I tested do not trigger extra downloads (test page https://samefourchords.com/2014/11/01/zakopane-poland):

## Chrome 51:
chrome 51
---
## Firefox 47 (same on 49 :sob:):
firefox 47
---
I couldn’t find a related bugzilla entry or anything at all TBH, just someone having the same problem: https://www.sitepoint.com/community/t/img-srcset-firefox-loads-too-many-images/219162...

One could argue that lower resolution images may look better at native res not having to undergo resizing and not exhibiting related interpolation artifacts, but one would be wrong, of course.

PR #12728, opened 2016-04-28
remove server side ab buckets

johnduffell:
As suggested by @TBonnin in https://github.com/guardian/fastly-edge-cache/pull/615 I have temporarily removed the unused server side ab test buckets.  See that PR for more details on why.


TBonnin:
👍

PR #12731, opened 2016-04-28
AB test https

JustinPinner:
## What does this change?

Sets up an (initially 0%) AB test for site-wide https.

Lets us discover if there are any areas that need further work before https can be enabled universally.

No. Or at least it _shouldn't_.
## Screenshots

One thing at a time! I just want to get it running without breaking anything first.
## Request for comment

@sndrs @desbo and anyone else who's an AB veteran.

### DIFF
--- a/static/src/javascripts/projects/common/modules/experiments/tests/https-test.js
+++ b/static/src/javascripts/projects/common/modules/experiments/tests/https-test.js
@@ -18,7 +18,7 @@ define([
         function getCookieValue(cookieName) {
             var cookieArray = document.cookie.split(';'),
                 cookieVal;
-            for(i = 0; i < cookieArray.length; i++) {
+            for(var i = 0; i < cookieArray.length; i++) {
                 var c = cookieArray[i].split('='),
                     n = c[0].trim(),
                     v = c[1].trim();
@@ -38,8 +38,8 @@ define([
             {
                 id: 'mvt-https',
                 test: function () {
-                    if(document.location.href.startsWith("http:")) {
-                        document.cookie = "https_opt_in=true; expires=Thu, 26 May 2016 22:59:59 GMT";
+                    if(document.location.href.startsWith('http:')) {
+                        document.cookie = 'https_opt_in=true; expires=Thu, 26 May 2016 22:59:59 GMT';
                         document.location = document.location.href.replace('http:','https:');
                     }
                 }
--- a/static/src/javascripts/projects/common/modules/experiments/tests/https-test.js
+++ b/static/src/javascripts/projects/common/modules/experiments/tests/https-test.js
@@ -4,7 +4,7 @@ define([
     cookies
 ) {
     return function () {
-        this.id = 'https-test';
+        this.id = 'HttpsTest';
         this.start = '2016-04-28';
         this.expiry = '2016-05-26';
         this.author = 'Justin Pinner';
--- a/static/src/javascripts/projects/common/modules/experiments/tests/https-test.js
+++ b/static/src/javascripts/projects/common/modules/experiments/tests/https-test.js
@@ -1,0 +1,49 @@
+define([
+], function (
+) {
+    return function () {
+        this.id = 'https-test';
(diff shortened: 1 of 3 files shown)


desbo:
It might be better to do this in fastly or play so we can send a redirect without having to wait for the JS to load?

static/src/javascripts/projects/common/modules/experiments/tests/https-test.js
```diff
@@ -0,0 +1,36 @@
...
+        this.audienceCriteria = 'Small percentage of users';
+        this.dataLinkNames = '';
+        this.idealOutcome = 'Nobody reports that things are broken';
+        this.hypothesis = 'We believe that most of our content should serve successfully over https, with a few known exceptions.';
+
+        this.canRun = function () {
+            return !cookies.get('https_opt_in');
+        };
+
+        this.variants = [
```
harrysalmon: At the moment there's no control group, even if it's a zero offset it'd be good to see it included from the start to check it.

static/src/javascripts/projects/common/modules/experiments/tests/https-test.js
```diff
@@ -0,0 +1,36 @@
...
+        this.audience = 0;
+        this.audienceOffset = 0;
+        this.successMeasure = 'Pages are served over https';
+        this.audienceCriteria = 'Small percentage of users';
+        this.dataLinkNames = '';
+        this.idealOutcome = 'Nobody reports that things are broken';
+        this.hypothesis = 'We believe that most of our content should serve successfully over https, with a few known exceptions.';
+
+        this.canRun = function () {
+            return !cookies.get('https_opt_in');
```
harrysalmon: It https_opt_in going to be the same value for the test and control groups?

static/src/javascripts/projects/common/modules/experiments/tests/https-test.js
```diff
@@ -0,0 +1,36 @@
...
+        };
+
+        this.variants = [
+            {
+                id: 'mvt-https',
+                test: function () {
+                    if(document.location.href.startsWith('http:')) {
```
harrysalmon: Could this logic exist in canRun rather than within the variant?

JustinPinner: It can. I'll move it.

static/src/javascripts/projects/common/modules/experiments/tests/https-test.js
JustinPinner: Ok

static/src/javascripts/projects/common/modules/experiments/tests/https-test.js
JustinPinner: It's a pre-existing cookie which may or may not exist when the test runs. The original plan was to only execute this test when somebody was going to be moved into https. Now I think we do need a proper control group with the cookie set to false as long as it isn't already present. I just tested that and it doesn't appear to have any effect on existing https sections. Which is good.

JustinPinner:
@desbo We don't have a mechanism for defining Fastly-level AB tests do we?

desbo:
I think you'd have to do something similar to what @dominickendrick did for his headline test, which is to assign a header based on the MVT id and create a server-side test to read from it. 

https://github.com/guardian/fastly-edge-cache/blob/master/theguardiancom/src/main/resources/varnish21/ab-tests.vcl#L42
https://github.com/guardian/frontend/blob/master/common/app/mvt/MultiVariateTesting.scala#L28

It is a little hacky and might not be as straightforward to analyse, but it'd probably be better for the user since it won't trigger a reload after the JS loads.

johnduffell:
looks good to me! doing it in fastly would be better from a user point of view IMO just because you can base it on a changing range of mvt ids (people can already opt in to https anyway) but actually it's only a one off penalty/reload to get people into the group, and after that you can get data easily. 👍

johnduffell:
this one is gathering dust/merge conflicts 💀

PR #12735, opened 2016-04-29
trackAds module

regiskuckaertz:
Second trial, this time with a **vengeance**.

cc @OliverJAsh @jbreckmckye @JonNorman

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/article-body-adverts.js
+++ b/static/src/javascripts/projects/common/modules/commercial/article-body-adverts.js
@@ -7,7 +7,8 @@ define([
     'common/modules/commercial/dfp/dfp-api',
     'common/modules/commercial/track-ad',
     'common/modules/commercial/create-ad-slot',
-    'common/modules/commercial/commercial-features'
+    'common/modules/commercial/commercial-features',
+    'lodash/functions/memoize'
 ], function (
     Promise,
     $,
@@ -17,7 +18,8 @@ define([
     dfp,
     trackAd,
     createAdSlot,
-    commercialFeatures
+    commercialFeatures,
+    memoize
 ) {
 
     /* bodyAds is a counter that keeps track of the number of inline MPUs
@@ -111,8 +113,8 @@ define([
     // the decoupling between the spacefinder algorithm and the targeting
     // in DFP: we can only know if a slot can be removed after we have
     // received a response from DFP
-    function waitForMerch() {
-        trackAd('dfp-ad--im').then(function (isLoaded) {
+    var waitForMerch = memoize(function waitForMerch() {
+        return trackAd('dfp-ad--im').then(function (isLoaded) {
             return isLoaded ? 0 : addArticleAds(2, getRules());
         }).then(function (countAdded) {
             return countAdded === 2 ?
@@ -123,7 +125,7 @@ define([
                 $('.ad-slot--inline').each(dfp.addSlot);
             }
         });
-    }
+    });
 
     function init() {
         if (!commercialFeatures.articleBodyAdverts) {
@@ -152,6 +154,10 @@ define([
     }
 
     return {
-        init: init
+        init: init,
+
+        '@@tests': {
+            waitForMerch: waitForMerch
+        }
     };
 });
--- a/static/src/javascripts/projects/common/modules/commercial/article-body-adverts.js
+++ b/static/src/javascripts/projects/common/modules/commercial/article-body-adverts.js
@@ -3,9 +3,9 @@ define([
     'common/utils/$',
     'common/utils/config',
     'common/utils/detect',
-    'common/utils/mediator',
     'common/modules/article/space-filler',
     'common/modules/commercial/dfp/dfp-api',
+    'common/modules/commercial/track-ad',
     'common/modules/commercial/create-ad-slot',
     'common/modules/commercial/commercial-features'
 ], function (
@@ -13,9 +13,9 @@ define([
     $,
     config,
     detect,
-    mediator,
     spaceFiller,
     dfp,
+    trackAd,
     createAdSlot,
     commercialFeatures
 ) {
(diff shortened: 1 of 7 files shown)


static/src/javascripts/projects/common/modules/commercial/track-ad.js
```diff
@@ -0,0 +1,30 @@
+define([
+    'Promise',
+    'common/utils/mediator'
+], function(Promise, mediator) {
+    return trackAd;
+
+    function trackAd(id) {
```
jbreckmckye: Should this be a method? And should it be working over mediator?

static/src/javascripts/projects/common/modules/commercial/sticky-top-banner.js
```diff
@@ -35,13 +37,7 @@ define([
...
-    var topAdRenderedPromise = new Promise(function (resolve) {
-        mediator.on('modules:commercial:dfp:rendered', function (event) {
-            var dfpAdSlotId = 'dfp-ad--top-above-nav';
-            var isEventForTopAdBanner = event.slot.getSlotElementId() === dfpAdSlotId;
-            if (isEventForTopAdBanner) { resolve(); }
-        });
-    });
+    var topAdRenderedPromise = trackAd('dfp-ad--top-above-nav');
 
     var getAdIframe = function () { return $('iframe', $adBanner); };
```
OliverJAsh: Is the below worth abstracting i.e. `trackRubiconAd`?

static/src/javascripts/projects/common/modules/commercial/track-ad.js
regiskuckaertz: dfp-api: good idea!
mediator: I don't understand your question

static/src/javascripts/projects/common/modules/commercial/track-ad.js
```diff
@@ -0,0 +1,30 @@
...
+        return new Promise(function (resolve, reject) {
+            var onAdLoaded = function (event) {
+                if (event.slot.getSlotElementId() === id) {
```
rich-nguyen: maybe we could have mediator emit an event specific to the slot. So instead of listening to `modules:commercial:dfp:rendered`, we listen to `modules:commercial:dfp:rendered:slotId`

static/src/javascripts/projects/common/modules/commercial/sticky-top-banner.js
regiskuckaertz: I don't know to be honest, this is a very isolated case, although we are thinking of using the messaging API more often. What do @jbreckmckye @uplne @Calanthe think?

static/src/javascripts/projects/common/modules/commercial/track-ad.js
jbreckmckye: We're basically broadcasting an event over mediator, which is itself just listening to a googletag event. If this method existed inside DFP, would we even need to pass the render messages to mediator at all?

static/src/javascripts/projects/common/modules/commercial/track-ad.js
```diff
@@ -0,0 +1,30 @@
...
+                    unlisten();
+                    resolve(!event.isEmpty);
```
jbreckmckye: Should _trackAd_'s interface be to return a success code, or should the promise return the event for the advert in question, given its generic-sounding name?

static/src/javascripts/projects/common/modules/commercial/track-ad.js
regiskuckaertz: I guess it depends if we want to let modules listen to all rendering events (unless mediator handles wildcard listeners?)

The next step is to get rid of mediator altogether. Knowing if an ad has rendered is a stateful query, events are stateless and transient.

static/src/javascripts/projects/common/modules/commercial/track-ad.js
regiskuckaertz: Do we have a use case for that? I'm wary of storing in memory objects we don't really control, such as the [`Slot`](https://developers.google.com/doubleclick-gpt/reference#googletagslot) object in this case.

static/src/javascripts/projects/common/modules/commercial/track-ad.js
regiskuckaertz: Yes indeed. That what I'd like to do next, but it's a bit more work—trackAd is stateful (this is why `Promise` is much better than `mediator`).

static/src/javascripts/projects/common/modules/commercial/track-ad.js
jbreckmckye: I don't have a use case for it, I just wasn't sure if I'd expect a boolean as part of the interface. I suppose it's fine.

regiskuckaertz: You're right, with such a name one would expect to know when an ad _changes_ state, not only when it is rendered.

static/src/javascripts/projects/common/modules/commercial/sticky-top-banner.js
jbreckmckye: What are we proposing, exactly?

static/src/javascripts/projects/common/modules/commercial/track-ad.js
jbreckmckye: I would agree that we should track lifecycle with promises, not events. The biggest issue with events is that your listeners must always be registered before the adverts are loaded, which is a kind of coupling across concerns. The other is that each thing that cares about the state of your adverts maintains its own copy of the ad's state.

You could implement a promise-based interface in dfp-api without too much trouble, I reckon:

```javascript
dfpApi.trackAd = memoize((adSlotId) => {
	const advert = adverts[adSlotId];
	const isRendered = advert && advert.isRendered;

	return new Promise((resolve, reject) => {
		if (isRendered) {
			resolve(true);
		} else {
			mediator.once('modules:commercial:dfp:rendered:' + adSlotId, event => {
				const adLoaded = !event.isEmpty;
				resolve(adLoaded);
			});
		}
	});
});
```

static/src/javascripts/projects/common/modules/commercial/article-body-adverts.js
```diff
@@ -151,6 +154,10 @@ define([
     }
 
     return {
-        init: init
+        init: init,
+
+        '@@tests': {
```
jbreckmckye: Not necessarily saying this is the right approach, but you _could_ avoid testing the internals of the module if your init function returned the promise that came from `waitForMerch()`.

jbreckmckye:
I don't have any objections to this as is. Let's :ship:

johnduffell:
this one's been around a while, is it dead or alive? 💀

regiskuckaertz:
I want to add memoization as suggested by @jbreckmckye , otherwise there is no point in providing this API, and also moving closer to DFP. When I have time though 😅

johnduffell:
fair enough! might be better to close and reopen

PR #12737, opened 2016-04-29
Fixed backgrounds in Fabric

jbreckmckye:
Fabric-v1 needs to be able to support fixed-position backgrounds.

These only run on desktop, so we can just use the `background-attachment: fixed` CSS property.

@regiskuckaertz @rich-nguyen


regiskuckaertz:
Looks good to me, although these variable names will need some care in the future 😲

👍

PR #12738, opened 2016-04-29
Verbose logging in config agent

piuccio:
The trait now logs which client is in use, otherwise there's no way to check that cross account communication is actually working


JonNorman:
:+1:

PR #12739, opened 2016-04-29
Explicit dependency on STS

piuccio:
## What does this change?

Make STS explicit in the dependencies. For some reason all apps work fine without except preview.

Preview works
## Request for comment

@TBonnin @janua

### DIFF
--- a/project/Frontend.scala
+++ b/project/Frontend.scala
@@ -22,6 +22,7 @@ object Frontend extends Build with Prototypes {
       awsDynamodb,
       awsS3,
       awsSns,
+      awsSts,
       awsSqs,
       contentApiClient,
       faciaScalaClient,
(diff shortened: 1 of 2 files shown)


project/Frontend.scala
```diff
@@ -189,11 +190,11 @@ object Frontend extends Build with Prototypes {
     adminJobs
   )
 
-  val preview = application("preview").dependsOn(withTests(common), standalone).settings(
+  val preview = application("preview").dependsOn(commonWithTests, standalone).settings(
     RoutesKeys.routesImport += "scala.language.reflectiveCalls"
   )
 
-  val trainingPreview = application("training-preview").dependsOn(withTests(common), standalone).settings(
```
TBonnin: Are `commonWithTests ` and `withTests(common)` not supposed to be the same?

piuccio: they are, all the other projects are using `commonWithTests`, defined like this `val commonWithTests = withTests(common)`. Just wanted to make consistent

TBonnin: ok. lgtm
Does this fix the issue we had earlier on preview though?

piuccio: yup, at least on my machine

TBonnin:
👍

PR #12740, opened 2016-04-29
Remove standfirst on multi cAPI component

regiskuckaertz:
Before:
picture 3

After:
picture 4

cc @JonNorman @kelvin-chappell


JonNorman:
Nice work :+1: - if we begin to see more differences between multi and single cAPI components, we might want to design and pass through a `CardOptions` object rather than paramaterising each difference; but this looks great!

regiskuckaertz:
Yeah, from my own perspective users should see no difference at all, whether the component is coming from DFP or editorial is really nobody's business.

PR #12741, opened 2016-04-29
Show capi-driven logos on section fronts

kelvin-chappell:
This will render a logo on sponsored section fronts when the switch is on.
There's also some improvement to the model for tag pages and content pages.

Eg:

image

/cc @steppenwells @Calanthe @uplne @lps88


common/app/model/Tag.scala
```diff
@@ -286,4 +287,8 @@ case class Tag (
   val isFootballTeam = properties.references.exists(_.`type` == "pa-football-team")
   val isFootballCompetition = properties.references.exists(_.`type` == "pa-football-competition")
   val contributorImagePath = properties.bylineImageUrl.map(ImgSrc(_, Contributor))
+
+  override def branding(edition: Edition): Option[Branding] = {
```
JonNorman: Does the notion of an "active" branding not apply to `Tag` as with `Section`?

kelvin-chappell: A capi `Tag` actually corresponds to a frontend `TagProperties`, which has an `activeBrandings` property: https://github.com/guardian/frontend/blob/master/common/app/model/Tag.scala#L240

kelvin-chappell: `Tag` and `Section` are both types of `Page`, and they both have a `branding` method which works out the branding based on the edition in scope.

applications/app/controllers/GalleryController.scala
```diff
@@ -56,7 +54,7 @@ object GalleryController extends Controller with RendersItemResponse with Loggin
     val htmlResponse: (() => Html) = () =>
       views.html.gallery(model, model.related, model.index)
     val jsonResponse = () =>
-      views.html.fragments.galleryBody(model.gallery, model.related, model.index)
+      views.html.fragments.galleryBody(model.gallery, model, model.related, model.index)
```
JonNorman: Can we just pass the `model` through here? Is `galleryBody` ever called where these attributes don't come from the same model?

JonNorman:
Looks good! :+1:

Only other small comment was the parameters into the fragments - it looks like we could just pass the model through, but up to you whether it is worth it / clearer.

applications/app/controllers/GalleryController.scala
kelvin-chappell: Good idea.  I've reduced the number of parameters to 1 to the gallery templates now.

JonNorman:
🚂

applications/app/controllers/GalleryController.scala
```diff
@@ -2,14 +2,12 @@ package controllers
 
 import com.gu.contentapi.client.model.v1.{Content => ApiContent, ItemResponse}
 import common._
-import contentapi.ContentApiClient
-import conf._
 import conf.switches.Switches
-import model._
+import contentapi.ContentApiClient
+import model.{GalleryPage, _}
```
jamesgorrie: Why not just `_`?

kelvin-chappell: That's the power of Intellij smart import optimisation!

kelvin-chappell: I'll fix it.

PR #12744, opened 2016-04-29
Fabric ads in first slice

jbreckmckye:
Fabric adverts should always appear in the first slot of mobile fronts - even if there's no slice candidate.

This PR also adds a unit test to verify the ad slot is being inserted.

@rich-nguyen - the way fabric works is a little implicit in the code at the moment. I'd like to rejig things, but I wouldn't mind getting a fix into code for Tijana.


rich-nguyen:
👍

PR #12745, opened 2016-04-29
High merchandising admin report

rich-nguyen:
Added a page so we can see line items which target high merchandising slots. This will be used to work out which keywords we target, and hence which content pages need to call dfp for high-merchandising.

Includes some commercial admin tidy up too.
screen shot 2016-04-29 at 16 57 05


admin/app/controllers/admin/CommercialController.scala
```diff
@@ -31,8 +31,8 @@ object CommercialController extends Controller with Logging with AuthLogging wit
     NoCache(Ok(views.html.commercial.commercialMenu(environment.stage)))
   }
 
-  def renderCommercial = AuthActions.AuthActionTest { implicit request =>
-    NoCache(Ok(views.html.commercial.commercial(environment.stage)))
+  def renderCommercialRenderTimes = AuthActions.AuthActionTest { implicit request =>
+    NoCache(Ok(views.html.commercial.commercialRenderTimes(environment.stage)))
   }
```
JonNorman: Just looking at this route (https://frontend.gutools.co.uk/analytics/commercial), it doesn't look like it actually gives us anything (useful) - shall we scrap it altogether?

rich-nguyen: It looks broken, yes. I just renamed it to render times, to keep the urls tidy. We should look at fixing it, definitely.

JonNorman:
Nice one, looks shiny :+1:

PR #12746, opened 2016-04-29
remove iOS crash metrics

sndrs:
## What does this change?

removes the iPhone confidence switch and related cloud watch metrics

we're not looking at it currently, and it costs us money. they're easy to add back if we need them again.

no
## Screenshots

🙈🙊🙉
## Request for comment

@rich-nguyen @michaelwmcnamara


rich-nguyen:
👍  great, thanks

PR #12749, opened 2016-05-03
Add US briefing email banner

desbo:
## What does this change?

Adds a new banner for US briefing emails, the first of which is due to go out today.

🙅 
## Screenshots

✉️


gtrufitt:
👍

PR #12720, opened 2016-04-28
Smaller fabric-expandable videos on mobile

jbreckmckye:
When we're on mobile, we want videos in fabric-expandables to take a smaller form factor, like they do in the old expandable-v3s:

image

@regiskuckaertz

PR #12751, opened 2016-05-03
Fabric-expandable video position should be mobile-only

jbreckmckye:
The side effect of [this](https://github.com/guardian/frontend/pull/12720) fix is that because videos are no longer flush to their container's lower edges, they peek out in the desktop view:

image

This PR limits the video's bottom margin - added for the sake of the mobile view - to mobile breakpoints exclusively.

image


rich-nguyen:
👍

PR #12753, opened 2016-05-03
Add badges to immersives

JonNorman:
## What does this change?

Immersive templates don't inject adbadges into the page for sponsored content, now they do.
## Screenshots

image
## Request for comment

/@regiskuckaertz


regiskuckaertz:
👍

sammorrisdesign:
How good quality are these logos? Do they have transparency?

JonNorman:
It varies - they are stored in DFP, so it depends how the creatives have been set up. I'm not aware of specific quality constraints. I've seen a mix of transparent and white-background badges on the site.

PR #12756, opened 2016-05-03
Rename offline-page to offline-crossword

sndrs:
## What does this change?

prepares space for alternative offline offering (for now, a simple cloudwatch ping for the rest of the site)

_update_
some context for this: getting metrics from the current offline page (by stroring a view in localstorage and pinging cloudwatch if we detect it again) limits us to users of the dev blog, who go offline, then come back to another secure page on the guardian subsequently. 

this clears out the old metrics stuff (which didn't work) and makes space for a metrics-only offline page that can be served regardless of the section, but which can leave the current crossword in place for dev blog

makes it easier to add more than one offline treatment
## Request for comment

@OliverJAsh @rich-nguyen

### DIFF
--- a/applications/app/templates/serviceWorker.scala.js
+++ b/applications/app/templates/serviceWorker.scala.js
@@ -27,7 +27,7 @@ var cachePageAndAssetResponses = function (jsonResponse, assetResponses) {
     var cacheName = [getISODate(), staticCacheName].join('-');
     return caches.open(cacheName).then(function (cache) {
         return jsonResponse.clone().json().then(function (jsonResponseJson) {
-            var pageRequest = new Request('/offline-page');
+            var pageRequest = new Request('/offline-crossword');
             var pageResponse = new Response(jsonResponseJson.html, { headers: { 'Content-Type': 'text/html' } });
             return Promise.all([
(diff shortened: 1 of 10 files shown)


applications/app/templates/serviceWorker.scala.js
```diff
@@ -23,11 +20,11 @@ var fetchAll = function (inputs) {
     }));
 };
 
-var cachePageAndAssetResponses = function (jsonResponse, assetResponses) {
```
OliverJAsh: Maybe you could name space all of the offline crossword code by wrapping them in an IIFE? It would be good to group the notifications stuff in the same way!

``` js
// Offline crossword
(() => {
  var updateCache = () => {};
})();
```

OliverJAsh:
Looks fine to me! 👍  I'm happy to help testing this locally if needed

applications/app/templates/serviceWorker.scala.js
sndrs: yeah good thinking, I will do

sndrs: i'm going to do this in a separate PR, with a bit of further rejigging (caching regex etc)

PR #12758, opened 2016-05-03
Hover state and image width in refactored components

regiskuckaertz:
## What does this change?
1. The hover/focus states for paidfor/supportedby containers is now in line with normal articles (i.e. no underline but a darker shade for the background)
2. The image for the soulmates component with 3 cards now occupies 100% of the card width (by removing the height constraint)

Before:
picture 2

After:
picture 1

@uplne

### DIFF
--- a/static/src/stylesheets/module/commercial/_adverts-capi.scss
+++ b/static/src/stylesheets/module/commercial/_adverts-capi.scss
@@ -117,6 +117,14 @@
 
 .advert--capi {
     border-top: 1px solid;
+    transition: background .1s;
+
+    &:hover,
+    &:focus {
+        .advert__title {
+            text-decoration: none;
+        }
+    }
 
     .advert__title,
     .advert__meta,
@@ -173,6 +181,11 @@
     background: $paid-article-subheader;
     border-top-color: $paid-article-brand;
 
+    &:hover,
+    &:focus {
+        background: darken($paid-article-subheader, 7%);
+    }
+
     .advert__title {
         @include f-headlineSans;
         font-weight: 400;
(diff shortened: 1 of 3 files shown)


uplne:
Screenshots would be nice

static/src/stylesheets/module/commercial/_adverts-capi.scss
```diff
@@ -173,6 +181,11 @@
     background: $paid-article-subheader;
     border-top-color: $paid-article-brand;
 
+    &:hover,
+    &:focus {
+        background: darken($paid-article-subheader, 7%);
```
uplne: My concern with this darken + whatever magic number pattern is that it will get messy pretty soon. I've used it on one project and it reaaaaaaallly fast became totally random percentages across all components. Based on the current mood basically.

uplne: So we should come up with some logic and stick to it.

uplne:
Other than comments 👍

static/src/stylesheets/module/commercial/_adverts-capi.scss
regiskuckaertz: What do you suggest?

regiskuckaertz: I'm afraid here I have to stick to the existing values. In the grand scheme of things, there should be a function that outputs the hover shade, yes; something out of our hands unfortunately.

uplne: what I mean is to have some rules, that darken can be 5% or 10% only. So we won't end up with gazillion shades of grey. It is really a minor thing but it will add up and guardian website shouldn't be a circus but some solid website with patterns.

regiskuckaertz:
Pictures! 🎆

static/src/stylesheets/module/commercial/_adverts-capi.scss
regiskuckaertz: yeah I agree, although with colours it's very hard to do, unless you switch to the lab colour space—adding 5% to an RGB colour can produce very different optical results based on the amount of white in the colour.

PR #12760, opened 2016-05-03
Remove fixed container refactor switch

kelvin-chappell:
I think this completes the great commercial container refactor!
[Apart from the manual content templates.]

There'll be more code to go when the great logo switchover takes place.


regiskuckaertz:
Awesome! Also, I will remove all the styles and scripts that we don't need anymore — I had to hold on to that until the manual components were not used anymore.

👍

PR #12761, opened 2016-05-03
never ue get]

jamesgorrie:
## What does this change?

Removing unsafe `.get` on Option

THings work

@sammorrisdesign @akash1810


sammorrisdesign:
👍 

Great PR title too

akash1810:
👍

PR #12764, opened 2016-05-03
Fix FrontsApi initialization bug

kelvin-chappell:
`java.lang.NoClassDefFoundError: Could not initialize class fronts.FrontsApi`


jamesgorrie:
👍 Nice catch

PR #12766, opened 2016-05-03
New gallery styling changes at smaller breakpoints

zeftilldeath:
Main change being; images are centered at small breakpoint now. Have added additional rule work to divide them comfortably.

screen shot 2016-05-03 at 16 38 00

Commits should make sense (For once).


OliverJAsh:
👍  after you've pushed!

PR #12768, opened 2016-05-03
Update related content to include legallySensitive check as per CAPI

gtrufitt:
## What does this change?
- Updates Frontend to match CAPI vocabulary for showRelatedContent - which is ['show related content' AND 'not legally sensitive'](https://github.com/guardian/content-api/blob/master/concierge/src/main/scala/com.gu.contentapi.concierge/view/handler/RelatedContentHandler.scala#L18-L20)
- Updates Outbrain to not show when related content is not shown to prevent outbrain being the first container after the article (as per Jamie Morrison's guidance of Outbrain behaviour)
- We match the behaviour of CAPI (the team describe this behaviour as "expected as indicating 2 subjects/persons related make us legally liable/responsible.") and can therefore rely on the showRelatedContent flag client-side more confidently
- This should reduce the number of 404s we see a bit as this is what we were returning on all legallySensitive articles that requested related content.
- Won't show Outbrain on legally sensitive articles when related content is not shown

Maybe - @guardian/android-developers @guardian/ios-developers - it looks as if you deal with an empty related content response from CAPI with a 'no related content at this time' message (at least in iOS), however you still show Outbrain, it is worth checking if this behaviour should be the same as I am implementing here.

@NataliaLKB - Does AMP show Outbrain?
## Request for comment

@guardian/commercial-dev


static/src/javascripts/projects/common/modules/commercial/third-party-tags/outbrain.js
```diff
@@ -165,6 +165,7 @@ define([
         if (commercialFeatures.outbrain &&
             !config.page.isFront &&
             !config.page.isPreview &&
+            config.page.showRelatedContent &&
```
jbreckmckye: Maybe this should go into `commercialFeatures.js`?

NataliaLKB:
@gtrufitt amp does show outbrain

static/src/javascripts/projects/common/modules/commercial/third-party-tags/outbrain.js
gtrufitt: Good shout, moved!

PR #12770, opened 2016-05-03
Fix logo position in sponsored galleries

regiskuckaertz:
## What does this change?

Correctly positions the logo in sponsored galleries.

Before:
picture 4

After:
picture 3

cc @JonNorman @jbreckmckye


JonNorman:
:+1:

PR #12773, opened 2016-05-03
S3 access logging

johnduffell:
we want to log the size of everything going out of S3 into facia.  this might help us know why we're getting so much out of s3 all the time

@TBonnin @sihil @adamnfish

### DIFF
--- a/common/app/services/S3.scala
+++ b/common/app/services/S3.scala
@@ -34,6 +34,7 @@ trait S3 extends Logging {
 
       val request = new GetObjectRequest(bucket, key)
       val result = client.getObject(request)
+      log.info(s"S3 got ${result.getObjectMetadata.getContentLength} bytes from ${result.getKey}")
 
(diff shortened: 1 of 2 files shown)


common/app/services/S3.scala
```diff
@@ -34,6 +34,7 @@ trait S3 extends Logging {
 
       val request = new GetObjectRequest(bucket, key)
       val result = client.getObject(request)
+      log.info(s"S3 got ${result.getObjectMetadata.getContentLength} bytes from ${result.getKey}")
```
jamespamplin: Can we possibly log the time for the request too? Also how is the error case logged?

johnduffell: we could but it would be more tricky, and the problem we're investigating is too much bandwidth not too long time.  the error case is probably logged elsewhere, Idon't think we're seeing errors

jamespamplin: ok, fair enough

jamespamplin:
:+1: looks good to me

PR #12774, opened 2016-05-04
Change wording of CTA on travel component

JonNorman:
A request from Andy Turner to change the wording on the CTAs within travel components.

@regiskuckaertz


regiskuckaertz:
👍

PR #12776, opened 2016-05-04
remove video series test

jamesgorrie:
## What does this change?

Removing the test. Results to come.

Everything still works and the world will continue as normal.

Nope.
## Request for comment

@akash1810 @harrysalmon


harrysalmon:
The presto-SQL for analysis of this is stored here, can't think of a better way to link them.
https://github.com/guardian/data-science/blob/master/mvt/video_series.sql

sndrs:
👍 gorgeous

PR #12777, opened 2016-05-04
Make the service worker more discreet

sndrs:
## What does this change?

Further to #12756 and a forthcoming tracking of offline views, this partitions the `fetch` parts of the service worker in more discrete silos.

It's also caches some regexes to avoid recreating them at each request.

should make it easier to add new behaviour to the service worker, and make it run quicker
## Request for comment

@OliverJAsh – you might want to give this a thorough exam 😉


applications/app/templates/serviceWorker.scala.js
```diff
@@ -8,201 +8,267 @@
...
+ * FETCH HANDLERS
+ */
 
-var needCredentialsWorkaround = function (url) {
-    var whitelist = ['https://discussion.theguardian.com/discussion-api'];
-    return whitelist.some(function (entry) {
-        return new RegExp('^' + entry).test(url);
-    });
-};
+var requestForDevBlog = (function (event) {
```
OliverJAsh: I don't think you need `event` here?

applications/app/templates/serviceWorker.scala.js
```diff
@@ -8,201 +8,267 @@
...
+    var isForDevBlog = (function (url, request) {
```
OliverJAsh: I don't think you need these params

applications/app/templates/serviceWorker.scala.js
```diff
@@ -8,201 +8,267 @@
...
-                    return response || fetch(request, needCredentialsWorkaround(request.url) ? { credentials: 'include' } : {});
-                })
-        );
+            );
+        };
     }
-});
+})();
+
+var requestForAsset = (function (event) {
```
OliverJAsh: Unused params

applications/app/templates/serviceWorker.scala.js
```diff
@@ -8,201 +8,267 @@
...
+    var isAssetRequest = (function (url) {
```
OliverJAsh: Unused params

OliverJAsh:
Nice organisation work!

I can't speak for the notifications part of the SW, but your changes to the offline crossword page look good to me! 👍 

As mentioned on Slack, #12756 requires from changes to Fastly and Nginx to get it working.

applications/app/templates/serviceWorker.scala.js
```diff
@@ -8,201 +8,267 @@
...
-    var whitelist = ['https://discussion.theguardian.com/discussion-api'];
-    return whitelist.some(function (entry) {
-        return new RegExp('^' + entry).test(url);
-    });
-};
+var requestForDevBlog = (function (event) {
+    var isForDevBlog = (function (url, request) {
+        var devBlogPathRegex = /^\/info\/developer-blog($|\/.*$)/;
+        return function (url, request) {
+            return isSameHost(url.host) && url.pathname.match(devBlogPathRegex) && requestAcceptsHTML(request);
```
OliverJAsh: Minor comment but I always use `RegExp.prototype.test` (which returns `boolean`) instead of `String.prototype.match` (which returns `null` or `array`), just because you don't forget about checking the type. In this case there's not much in it though

applications/app/templates/serviceWorker.scala.js
```diff
@@ -8,201 +8,267 @@
...
+        var url = new URL(request.url);
 
-    if (doesRequestAcceptHtml(request)) {
-        isCacheUpdated().then(function (isUpdated) {
-            if (!isUpdated) {
-                updateOfflineCrosswordCache().then(deleteOldCaches);
-            }
-        });
-    }
+        if (isForDevBlog(url, request)) {
```
OliverJAsh: Perhaps `isForDevBlog` can just take a `request` and construct the URL itself?

applications/app/templates/serviceWorker.scala.js
```diff
@@ -8,201 +8,267 @@
...
+ * EVENT LISTENERS
+ */
 
-self.addEventListener('activate', function(event) {
+this.addEventListener('install', function (event) {
+    event.waitUntil(updateOfflineCrosswordCache());
 });
 
-self.addEventListener('push', function(event) {
+self.addEventListener('activate', function (event) {});
```
OliverJAsh: I think this can just be deleted? Not sure why we had it

applications/app/templates/serviceWorker.scala.js
sndrs: oh yeah? https://github.com/guardian/frontend/commit/bb32671d146056fe4510866fd88b4d042a5e2f2a#diff-a11e736f6f4b342fee7da3437b919fd0R114

😄

applications/app/templates/serviceWorker.scala.js
sndrs: fair enough, didn't want to make too many edits to functionality, but yeah, not doing much by the look of it

applications/app/templates/serviceWorker.scala.js
sndrs: oh yeah good shout - i think i've always done these like this, and they're doing nothing - nice one

applications/app/templates/serviceWorker.scala.js
OliverJAsh: That's a comment to myself then ;-)

PR #12779, opened 2016-05-04
Remove some redundant commercial container code

kelvin-chappell:
We no longer need a special single-sponsor commercial container.

The multiple-sponsor commercial container can go when we've switched on the `cardsDecideContainerBranding` switch.

### DIFF
--- a/common/app/views/fragments/containers/facia_cards/commercialContainer.scala.html
+++ b/common/app/views/fragments/containers/facia_cards/commercialContainer.scala.html
@@ -1,149 +1,58 @@
 @(containerType: slices.CommercialContainerType, container: layout.FaciaContainer, frontProperties: model.FrontProperties)(implicit request: RequestHeader)
 @import common.{Edition, LinkTo}
-@import conf.switches.Switches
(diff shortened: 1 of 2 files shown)


JonNorman:
These PRs are so satisfying :+1:

common/app/views/fragments/containers/facia_cards/commercialContainer.scala.html
```diff
@@ -1,149 +1,58 @@
...
-                                                    </li>
-
-                                                }
-                                                case _ => {}
-                                            }
-                                        }
-                                    }
-                                    </ul>
-                                }
+<div class="fc-container fc-container--commercial">
```
regiskuckaertz: Bonus point if you can use `@containerWrapper` 😍

kelvin-chappell: The rest of this template is all going to go as well in the next few weeks so doesn't seem worth changing it.  It's just that we can't show multiple logos on containers until the other work is done.

PR #12780, opened 2016-05-04
Add technology and business to LinkTo httpsEnabledSections

JustinPinner:
## What does this change?

Technology and business are now on https so links to those resources should be updated.

It saves unnecessary redirects from http to https on inbound links.

No.
## Screenshots

N/A
## Request for comment

@desbo and/or anyone else in frontend
**Note**: tests may fail - I need to open the PR to see what's affected


desbo:
👍

JustinPinner:
Thanks for the +1 @desbo. Unfortunately I found that LinkTo wasn't working as expected for editionalised secure sections, so I've tweaked it a bit to accommodate that behaviour. Could you have a second look when you get a minute? Thanks v much.

desbo:
ah yes, I missed that! 
looks like it'll cover it to me, good tests! 👍

PR #12781, opened 2016-05-04
Spacefinder: wait for interactives to load

regiskuckaertz:
When an article contains interactives, spacefinder may get geometry wrong if the interactives haven't been loaded yet and therefore think there is space enough for inline slots:

picture 2

Alas, there is no way of knowing if an interactive has been "enhanced" unless we wait for the load event of the containing iframe... which may not even exist when spacefinder is running. Event delegation is a no go, so I had to use a `MutationObserver`, which was kind of fun.

picture 1

@uplne

### DIFF
--- a/static/src/javascripts/projects/common/modules/article/spacefinder.js
+++ b/static/src/javascripts/projects/common/modules/article/spacefinder.js
@@ -91,7 +91,7 @@ define([
             return !(iframe.length && isIframeLoaded(iframe[0]));
         });
 
-        return notLoaded.length === 0 ?
+        return notLoaded.length === 0 || !('MutationObserver' in window) ?
             true :
             Promise.all(notLoaded.map(function (interactive) {
                 return new Promise(function (resolve) {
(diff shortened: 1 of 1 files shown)


static/src/javascripts/projects/common/modules/article/spacefinder.js
```diff
@@ -79,12 +79,54 @@ define([
...
+        var notLoaded = qwery('.element-interactive', body).filter(function (interactive) {
+            var iframe = qwery(interactive.children).filter(isIframe);
+            return !(iframe.length && isIframeLoaded(iframe[0]));
+        });
+
+        return notLoaded.length === 0 ?
+            true :
+            Promise.all(notLoaded.map(function (interactive) {
+                return new Promise(function (resolve) {
+                    new MutationObserver(function (records, instance) {
```
gtrufitt: Do we polyfill this? Do we need to for IE10,9?

regiskuckaertz: I'll count this as a progressive enhancement, in which case I should probably add a test for `MutationObserver`. Thanks for the shout out!

uplne: progressive enhancement ftw. after that you can 🚢 👍

gtrufitt: 💛 👍

PR #12786, opened 2016-05-04
Show capi-driven logos on facia fronts

kelvin-chappell:
This change will show logos from capi on facia fronts, instead of logos from DFP, when switch is on.

The logo data is stored in `FrontProperties` so that it gets pressed with the front.

/cc @steppenwells

### DIFF
--- a/facia/app/views/fragments/brandedFrontBody.scala.html
+++ b/facia/app/views/fragments/brandedFrontBody.scala.html
@@ -1,0 +1,61 @@
+@(faciaPage: model.PressedPage)(implicit request: RequestHeader)
+@import common.Edition
+@import common.commercial.ContainerModel
+@import model.{PaidContent, Sponsored}
+@import views.support.RenderClasses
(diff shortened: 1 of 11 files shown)


facia/app/views/fragments/brandedFrontBody.scala.html
```diff
@@ -0,0 +1,61 @@
+@(faciaPage: model.PressedPage)(implicit request: RequestHeader)
+@import common.Edition
+@import common.commercial.ContainerModel
+@import model.{PaidContent, Sponsored}
+@import views.support.RenderClasses
+
+@faciaPage.branding(Edition(request)) match {
+
+    case Some(branding) => {
```
JonNorman: if we have no branding here then we call the `frontBody` fragment. Would it be worth putting the branded frontBody code into a separate fragment too?

JonNorman:
Looks good to me :+1:

PR #12787, opened 2016-05-04
Manual single: restore the call to action

regiskuckaertz:
I was not dealing with ctas correctly in the manual single component, and the cta was not showing up:
picture 1

Now it does (plus minor typographic improvements):
picture 2

cc @uplne @JonNorman


JonNorman:
:+1:

PR #12788, opened 2016-05-04
Add primitive offline page view metrics

sndrs:
## What does this change?

Adds a simple offline view in the service worker that increments a count in a cookie (so we can get at it on http too). If we find that value in the cookie when coming back to the site, we ping cloudwatch.

All non-developer-blog pages on https will get this if you're offline (and can register a service worker).

Should give us some kind of idea how many people try to view get to us offline.

no.
## Screenshots

screen shot 2016-05-04 at 17 12 26
## Request for comment

@OliverJAsh


common/app/templates/inlineJS/blocking/cloudwatchBeacons.scala.js
```diff
@@ -1,2 +1,22 @@
...
+                var data = cookies[i].split('=');
+                if (data[0].indexOf('gu.offlineViews') !== -1) {
+                    offlineViews = parseInt(data[1]);
+                    break;
+                }
+            }
+            for (var i = 0; i < offlineViews; i++) {
+                (new Image()).src = window.guardian.config.page.beaconUrl + '/count/offline-page-view.gif';
+            }
+            document.cookie = 'gu.offlineViews=0;Max-Age=0';
```
OliverJAsh: Out of interest how come we went with cookies instead? Disadvantage being extra overhead in requests to the server

OliverJAsh:
Looks great! Looking forward to hearing the stats on this one 👍

applications/app/templates/serviceWorker.scala.js
```diff
@@ -150,6 +150,24 @@ var requestForAsset = (function () {
...
...data[1]); break;}}document.cookie=\'gu.offlineViews=\' + (offlineViews + 1) + \';Max-Age=31536000\';})(); </script></head><body style="background-color: #005689;color:white; font-family: Helvetica,Arial,sans-serif;font-size: 12px"><div style="display: flex;align-items: center;justify-content: center;flex-direction:column;position: absolute;top:0;right:0;bottom:0;left:0"><svg width="36" height="36" viewBox="0 0 36 36" fill="white" style="height: 64px;width: 64px;"><path d="M21.3 8.8c0-4.9-1.5-5.7-3.3-5.7-1.8 0-3.2.7-3.2 5.7s1.5 5.5 3.2 5.5c1.8-.1 3.3-.6 3.3-5.5m-6.5 18.8c-2.3 0-2.9 1.7-2.9 2.9 0 1.8 1.6 3.4 6.3 3.4 5.3 0 6.8-1.5 6.8-3.4 0-1.7-1.3-2.9-3.4-2.9h-6.8zM10.5 2.4C4.3 5.2 0 11.4 0 18.7c0 4.9 2 9.4 5.2 12.6V31c0-3.2 3.1-4.4 5.9-5-2.6-.6-3.9-2.5-3.9-4.4 0-2.6 2.9-4.8 4.3-5.8l-.2-.1c-2.5-1.4-4.1-3.8-4.1-7 0-2.7 1.2-4.9 3.3-6.3M36 18.8C36 11.4 31.5 5 25.1 2.3c2.1 1.4 3.4 3.5 3.5 6.3l.1.6c0 5.4-4.4 8.2-10.7 8.2-1.6 0-2.7-.1-4.1-.5-.6.4-1.1 1.1-1.1 1.8 0 .9.8 1.6 1.8 1.6h8.8c5.5 0 8.2 2.2 8.2 7.1 0 1.6-.3 3.1-1 4.3 3.3-3.4 5.4-7.9 5.4-12.9"></path></svg><p style="padding: 2rem 2rem 0;text-align:center;">Sorry, you do not have any network at the moment.</p></div></body></html>', {
```
OliverJAsh: Not sure it matters as this is a test, but since this file is `scala.js`, this could be a HTML template that you reference.

common/app/templates/inlineJS/blocking/cloudwatchBeacons.scala.js
OliverJAsh: Ahh is it because you can't access local storage from the SW?

sndrs: means we can read cookies set on http from https. can't do that with localstorage, so we could only record the view if you return to a secure page

applications/app/templates/serviceWorker.scala.js
sndrs: ooooh yes. i'll not merge it tonight, just incase, so I'll do that now.

common/app/templates/inlineJS/blocking/cloudwatchBeacons.scala.js
sndrs: > Ahh is it because you can't access local storage from the SW?

we're doing it from the page to get over the SW/sync api gulf, so it could work. the protocol was the issue

applications/app/templates/serviceWorker.scala.js
```diff
@@ -150,6 +150,24 @@ var requestForAsset = (function () {
     }
 })();
 
+// handle requests for generic pages
+var requestForPage = function (event) {
```
gtrufitt: What's preventing this from being called on the dev blog as well? I might be missing something to do with `fetch` being a global promise that's resolved or something fun like that. Can there be a race condition?

sndrs: hmmm, that's a good point. currently, it's work as you'd expect, but I think you're right – that must be a race condition?

applications/app/templates/serviceWorker.scala.js
```diff
@@ -86,70 +87,76 @@ var isSameHost = function (host) {
...
+
+var isRequestForGenericPage = function (request) {
+    var url = new URL(request.url);
+    return requestAcceptsHTML(request) && isSameHost(url.host);
+}
+
+// handle requests for generic pages
+var handleGenericPageRequest = function (event) {
+    event.respondWith(
+        fetch(event.request).catch(function () {
```
gtrufitt: Is there ever a situation where we would want to check the cache for an html page? I think that would be ridiculous in most cases but don't know if you need to handle it?

gtrufitt: Will we respond with the offline page if we hit a random network error in an ajax request for html? The email sign-up for example? Probably an unlikely situation, but possible if you're on a train and lose network? Maybe you need to clarify if it's a request to a front or article page?

applications/app/templates/serviceWorker.scala.js
```diff
@@ -159,8 +166,17 @@ this.addEventListener('install', function (event) {
     event.waitUntil(updateOfflineCrosswordCache());
 });
 
-this.addEventListener('fetch', requestForDevBlog);
-this.addEventListener('fetch', requestForAsset);
+this.addEventListener('fetch', function (event) {
+    var request = event.request;
+
+    if(isRequestForDevBlog(request)) {
```
gtrufitt: Much safer 😎

gtrufitt:
A couple of minor comments but looks good to me 👍

applications/app/templates/serviceWorker.scala.js
sndrs: it will, but that's ok. you'd get the browser's default offline page anyway (as discussed offline)

PR #12789, opened 2016-05-04
Commercial components: remove active state colour

regiskuckaertz:
When active, links in commercial components should keep the same colour.

cc @uplne @JonNorman


rich-nguyen:
👍

PR #12791, opened 2016-05-05
Extending out-pf-page-slot-always switch

JonNorman:
Will discuss with Ad Ops as to when we can remove this switch this morning, but will extend for now.


kelvin-chappell:
👍

jamespamplin:
:+1: thanks

PR #12793, opened 2016-05-05
Manual components without capped height

regiskuckaertz:
Previously, only for manual soulmates components, now for all manual components.

cc @uplne


kenlim:
👍

PR #12794, opened 2016-05-05
Remove identity-register-membership-standfirst ab test

jamespamplin:
Removes `identity-register-membership-standfirst` ab test. Test definition was required for tableau dashboards, which are no longer used.


gtrufitt:
:+

PR #12795, opened 2016-05-05
High merch slot server side match

Calum-Campbell:
## What does this change?

Checks for a match between high-merchandising tags and page tags on the server side not on the client side. If no tags are matched the page should not render a 'fc-container--commercial-high'  div on the page. 

Should reduce the dfp server returning an empty ad response errors. Currently at 16k events in the last 2 days.
Can be seen and assessed on Sentry. 

Shouldn't do as template only used on this platform.


common/app/common/dfp/HighMerchandiseComponentAgent.scala
```diff
@@ -0,0 +1,12 @@
+package common.dfp
+
+import model.Tag
+
+trait HighMerchandiseComponentAgent {
+
+  protected def highMerchandisingTargetedTags: HighMerchandisingTargetedTagSet
+
+  def hasHighMerchandiseTarget(tags: Seq[Tag]): Boolean = {
+    tags exists highMerchandisingTargetedTags.hasTag
```
kelvin-chappell: You might want to put a feature switch in here as well so that if the switch is off, this is always true.

kelvin-chappell:
This looks good but have you discussed it with Jamie and/or merchandising team?
If the high-relevance merchandising slot is only targeting tags, 👍 
But I'm pretty sure there are other criteria as well.

common/app/common/dfp/DfpAgent.scala
```diff
@@ -88,6 +90,22 @@ object DfpAgent
       maybeTagSet getOrElse InlineMerchandisingTagSet()
     }
 
+    def grabHighMerchandisingTargetTagsFromStore(): HighMerchandisingTargetedTagSet = {
+
+      val maybeTagList = for {
+        jsonString <- stringFromS3(dfpHighMerchandisingTagsDataKey)
+        report <- HighMerchandisingTargetedTagsReportParser(jsonString)
+      } yield report.lineItems.items
```
rich-nguyen: I think these two for comprehensions can be combined

common/app/common/dfp/TagSponsorship.scala
```diff
@@ -65,6 +65,30 @@ object InlineMerchandisingTargetedTagsReportParser extends Logging {
...
+        "items" -> tagSet.items
+      )
+    }
+  }
+}
+
+case class HighMerchandisingTargetedTagSet(items: Set[String] = Set.empty){
+
+  private def hasTagId(tagId: String): Boolean = tagId.split('/').lastOption exists { keyword =>
+    items contains keyword
```
rich-nguyen: something like `items.exists(tagId.endsWith(_))`

common/app/common/dfp/TagSponsorship.scala
```diff
@@ -65,6 +65,30 @@ object InlineMerchandisingTargetedTagsReportParser extends Logging {
...
+  }
+
+  def hasTag(tag: Tag):Boolean = hasTagId(tag.id)
```
rich-nguyen: inline `hasTagId`?

rich-nguyen:
Have we checked that outbrain still functions as normal?

rich-nguyen:
As @kelvin-chappell , a switch would be handy here, and we need to double check the other criteria which may be used for high merch targeting. 👍

regiskuckaertz:
@rich-nguyen in outbrain.js we test the existence of `merchandising-high` so this change shouldn't be a problem

regiskuckaertz:
Nice work! Can we then remove specs from [`create-ad-slot`](https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/common/modules/commercial/create-ad-slot.js#L64)?

regiskuckaertz:
Also important to remove [this](https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/common/modules/commercial/front-commercial-components.js#L22)

common/app/common/dfp/HighMerchandiseComponentAgent.scala
```diff
@@ -0,0 +1,16 @@
...
+
+  protected def highMerchandisingTargetedTags: HighMerchandisingTargetedTagSet
+
+  def hasHighMerchandisingTarget(tags: Seq[Tag]): Boolean = {
+    if (highMerchandisingComponentSwitch.isSwitchedOn)
+      true
+    else
+      tags exists highMerchandisingTargetedTags.hasTag
+  }
+}
```
Calum-Campbell: @kelvin-chappell Switch added, any comments appreciated.

common/app/common/dfp/TagSponsorship.scala
```diff
@@ -65,6 +65,26 @@ object InlineMerchandisingTargetedTagsReportParser extends Logging {
...
+}
+
+case class HighMerchandisingTargetedTagSet(items: Set[String] = Set.empty){
+  
+  def hasTag (tag: Tag): Boolean = items.contains(tag.id.split('/').lastOption.get)
+
+  def nonEmpty = items.nonEmpty
+
+}
+
```
Calum-Campbell: @rich-nguyen refactored to one def, not using exists function however, Is there a better way using exists?

common/app/common/dfp/DfpAgent.scala
```diff
@@ -88,6 +90,17 @@ object DfpAgent
...
+
+      val tags = for {
+        jsonString <- stringFromS3(dfpHighMerchandisingTagsDataKey).toSeq
+        report <- HighMerchandisingTargetedTagsReportParser(jsonString).toSeq
+        lineItems <- report.lineItems.items
+        tag <- lineItems.tags
+      } yield tag
+      HighMerchandisingTargetedTagSet(tags.toSet)
+    }
+
```
Calum-Campbell: @rich-nguyen comprehensions combined ^^

common/app/common/dfp/HighMerchandiseComponentAgent.scala
kelvin-chappell: Effect is right but could be simplified to:
```
highMerchandisingComponentSwitch.isSwitchedOn || 
(tags exists highMerchandisingTargetedTags.hasTag)
```

Calum-Campbell: Sounds good, thought there would be a more concise way to do it. Thanks

Calum-Campbell:
@rich-nguyen , as @regiskuckaertz says, outbrain still functioning fine, due to testing of high-merchandising in outbrain.js. 
@regiskuckaertz - yep redundant js files/lines now removed, cheers.

regiskuckaertz:
You probably need to also remove where the module is loaded, i.e. in commercial.js

regiskuckaertz:
Other than that small change, the JS looks good to me 👍

common/app/common/dfp/DfpAgent.scala
rich-nguyen: nice

PR #12796, opened 2016-05-05
Fix incorrect button colour in manual components

regiskuckaertz:
Before:
picture 6

After:
picture 7

cc @JonNorman


JonNorman:
:+1:

PR #12797, opened 2016-05-05
Facia HealthCheck request is not inline anymore

TBonnin:
## What does this change?

Instead of inlining the backend request every time the /_healthcheck endpoint is
hit, we are now polling on a regular schedule, cache the
result and use this result to serve the response to the /_healthcheck
endpoint

If this works out, we would enable it for all apps

Note: I didn't touch the current way we are handling healthcheck https://github.com/guardian/frontend/blob/master/common/app/conf/AllGoodHealthcheckController.scala but created new traits/object and duplicate some code. It will be easy then to delete the old way 😉 

The number of healthcheck requests will stay stable and won't increase anymore on the number of when ELBs are scaled up

No
## Request for comment

@johnduffell @jamespamplin


common/app/conf/HealthCheck.scala
```diff
@@ -0,0 +1,153 @@
...
+
+  def runChecks = cache.fetchPaths(paths)
+
+  private def healthCheckResponse(condition: => Boolean) = Action.async {
+    Future.successful {
+      val response = cache.get().map {
+        case (url, r) => s"GET ${url} '${r.formattedResult}' '${r.formattedDate}'"
+      }
+        .mkString("\n")
+      if(condition) Ok(response) else ServiceUnavailable(response)
```
TBonnin: We are currently returning an `Internal Server error` if /_healthcheck fails. https://github.com/guardian/frontend/blob/master/common/app/conf/AllGoodHealthcheckController.scala#L59

I think it make more sense to return `Service unavailable`
What do you think?

common/app/conf/HealthCheck.scala
```diff
@@ -0,0 +1,153 @@
...
+/*
+* Healthcheck endpoints are requested periodically
+* Results are stored in a cache
+* Goal: Requests made to the healthcheck endpoint by the ELBs are not 'inlined'
+* */
+
+private[conf] case class HealthCheckResult(url: String,
+                                           result: Either[Throwable, Int], // Status code if success, throwable otherwise
+                                           date: DateTime = DateTime.now,
+                                           expiration: Duration = 10.seconds) {
```
TBonnin: I am happy to discuss this 10 seconds expiration value :)

common/app/conf/HealthCheck.scala
```diff
@@ -0,0 +1,153 @@
...
+  def apply(url: String, error: Throwable, date: DateTime) = {
+    new HealthCheckResult(url, Left(error), date)
+  }
+  def apply(url: String, status: Int, date: DateTime, expiration: Duration) = {
+    new HealthCheckResult(url, Right(status), date, expiration)
+  }
+  def apply(url: String, error: Throwable, date: DateTime, expiration: Duration) = {
+    new HealthCheckResult(url, Left(error), date, expiration)
+  }
+}
```
TBonnin: Those are only convenience constructors, but is there a way to avoid this kind of duplication

facia/conf/routes
```diff
@@ -6,8 +6,8 @@
 # For dev machines
 GET        /assets/*path                                                            dev.DevAssetsController.at(path)
 
-GET        /_healthcheck                                                            conf.HealthCheck.healthcheck()
-GET        /_fronts_cdn_healthcheck                                                 conf.HealthCheck.cdnHealthcheck()
+GET        /_healthcheck                                                            conf.HealthCheckController.healthCheck()
+GET        /_fronts_cdn_healthcheck                                                 conf.HealthCheckController.healthCheck()
```
TBonnin: Since we are caching result, we can use the same endpoint for normal and cdn healthcheck.
Eventually we can remove `/_fronts_cdn_healthcheck ` all together

facia/app/conf/HealthCheck.scala
```diff
@@ -8,3 +8,13 @@ object HealthCheck extends AllGoodHealthcheckController(9008, "/uk") {
     NoCache(if (isOk) Ok("OK") else ServiceUnavailable("Service Unavailable"))
   }
 }
+
+object HealthCheckController extends CachedHealthCheckController {
+  override val paths = Seq("/uk/business")
```
TBonnin: Instead of checking `/uk` which is quite heavy, I changed it to `/uk/business` let me know what you think

common/app/conf/HealthCheck.scala
```diff
@@ -0,0 +1,153 @@
...
+
+/*
+* Healthcheck endpoints are requested periodically
+* Results are stored in a cache
+* Goal: Requests made to the healthcheck endpoint by the ELBs are not 'inlined'
+* */
+
+private[conf] case class HealthCheckResult(url: String,
+                                           result: Either[Throwable, Int], // Status code if success, throwable otherwise
+                                           date: DateTime = DateTime.now,
```
johnduffell: no need to store date and expiration, just do the calculation of expires when the class is constructed

common/app/conf/HealthCheck.scala
johnduffell: if they're only for tests, move them into the test file

common/app/conf/HealthCheck.scala
johnduffell: as long as the ELB will treat it as unhealthy I'm happy

common/app/conf/HealthCheck.scala
TBonnin: The goal was so you can specify your own creation date and expiration delay. I agree though the expire date can be calculated at construction

common/app/conf/HealthCheck.scala
TBonnin: Those are mostly used in tests but not only

facia/app/conf/HealthCheck.scala
```diff
@@ -8,3 +8,13 @@ object HealthCheck extends AllGoodHealthcheckController(9008, "/uk") {
     NoCache(if (isOk) Ok("OK") else ServiceUnavailable("Service Unavailable"))
   }
 }
+
+object HealthCheckController extends CachedHealthCheckController {
```
johnduffell: all this extending works, but it would be nice if we could minimise the use of inheritance and make it more functional, just to make it easier to follow the code

common/app/conf/HealthCheck.scala
johnduffell: I can't find anywhere else than tests! if you remove them what doesn't compile?

johnduffell:
great work, this will really reduce the load on the system 👍 🐎

common/app/conf/HealthCheck.scala
TBonnin: You are right. I wanted to but I ended up only using this in the tests. Will move

common/app/conf/HealthCheck.scala
```diff
@@ -0,0 +1,127 @@
...
+  }
+  def formattedDate: String = if(expired) s"${date} (Expired)" else date.toString
+}
+
+private[conf] trait HealthCheckFetcher extends ExecutionContexts with Logging {
+  import play.api.Play.current
+
+  protected def fetchResult(baseUrl: String, port: Int, path: String): Future[HealthCheckResult] = {
+    WS.url(s"$baseUrl$path")
+      .withHeaders("X-Gu-Management-Healthcheck" -> "true")
```
jamespamplin: Can we set a `User-Agent` header too? Can be useful for the logs

common/app/conf/HealthCheck.scala
```diff
@@ -0,0 +1,127 @@
...
+      .withRequestTimeout(10000).get()
+      .map(result => HealthCheckResult(path, Right(result.status)))
```
jamespamplin: It might be good to get the `statusText` and the body as a string to be logged for a "bad" response (non 2xx). So perhaps you could catch bad responses here, returning something like: `Left(HealthCheckBadResponseException(statusCode, statusText, body: String))`.

common/app/conf/HealthCheck.scala
```diff
@@ -0,0 +1,127 @@
...
+    }
+    val baseUrl = s"http://localhost:$port"
+    Future.sequence(paths.map(fetchResult(baseUrl, port, _)))
+  }
+
+}
+
+private[conf] trait HealthCheckCache extends HealthCheckFetcher {
+
+  protected val cache = AkkaAgent[Map[String, HealthCheckResult]](Map.empty)
```
jamespamplin: Looks like you are not using the keys on this collection, so you could just use a Sequence or a List.

common/app/conf/HealthCheck.scala
```diff
@@ -0,0 +1,127 @@
...
+    }
+  }
+  def anySuccessful: Boolean = get().values.exists(_.recentlySucceed)
+
+  def fetchPaths(testPort: Int, paths: Seq[String]): Future[Unit] = {
+    log.info("Fetching HealthChecks...")
+    fetchResults(testPort, paths:_*)
+      .map { m =>
+      m.foreach { case healthCheckResult =>
+        cache.alter(_.updated(healthCheckResult.url, healthCheckResult))
```
jamespamplin: Not too sure on how alter works, but it looks like you'll alter the cache for each path checked, since you have all the results at this stage, you could just have a single `send` or `alter` statement with the whole new list of results.

common/app/conf/HealthCheck.scala
jamespamplin: 502 Service Unavailable makes more semantic sense to me

common/app/conf/HealthCheck.scala
```diff
@@ -0,0 +1,127 @@
...
+}
+object HealthCheckCache extends HealthCheckCache
+
+trait CachedHealthCheckController extends Controller with Results with ExecutionContexts with Logging {
+
+  val paths: Seq[String]
+  val testPort: Int
+
+  def healthCheck(): Action[AnyContent]
+  private[conf] val cache: HealthCheckCache = HealthCheckCache
```
jamespamplin: We may run into problems here if multiple `CachedHealthCheckController`s are defined on the same app, as `HealthCheckCache` is a shared resource. Might be safer to have an instance of the cache per controller.

facia/app/conf/HealthCheck.scala
```diff
@@ -3,8 +3,12 @@ package conf
 import model.NoCache
 import play.api.mvc.Action
 
-object HealthCheck extends AllGoodHealthcheckController(9008, "/uk") {
-  def cdnHealthcheck = Action{ request =>
-    NoCache(if (isOk) Ok("OK") else ServiceUnavailable("Service Unavailable"))
-  }
+object HealthCheckController extends CachedHealthCheckController {
+  override val paths = Seq("/uk/business")
+  override val testPort = 9008
```
jamespamplin: Is it possible not to hardcode this? Is this helpful: http://stackoverflow.com/a/21353658 ?

common/app/conf/HealthCheck.scala
TBonnin: why not but what the difference to use the `user-agent` header rather than a custom `X-Gu-Management-Healthcheck`?

common/app/conf/HealthCheck.scala
TBonnin: Good idea.

common/app/conf/HealthCheck.scala
TBonnin: I was originally using it to update the specific map value but as you pointed it out below I now should update the all cache all together since I have all the necessary value. Good catch thx

common/app/conf/HealthCheck.scala
TBonnin: Yep. Thanks

common/app/conf/HealthCheck.scala
TBonnin: Totally. Thanks for pointing this mistake out.

facia/app/conf/HealthCheck.scala
TBonnin: I do agree it would be better not to hardcode this. Let me see what I can do. Thx

common/app/conf/HealthCheck.scala
jamespamplin: `User-Agent` normally is logged in access logs, although we have a custom access logger for our app level logs, so I'm not sure if its logged at all. Can you check for me please? Its useful for debugging so we can distinguish between Healthcheck requests and "real" requests.

TBonnin: we don't log any request headers at the moment but I am planning to add that soon. I totally agree about distinguishing between healthcheck and real traffic though.

common/app/conf/HealthCheck.scala
TBonnin: Do we need the body though?

jamespamplin: Can be useful for debugging 500 style errors. Although if we're using the default play error handlers, play omits the stack trace from the error page in PROD mode. So probably not that useful, we can leave that out.

TBonnin:
@jamespamplin I have implemented all your great feedback. Let me know what you think. Thanks again

jamespamplin:
:+1: changes look good, thanks for that!

PR #12800, opened 2016-05-05
expired switches now turn red, rather than orange

JonNorman:
Currently when a switch is actually expired, the colour goes back to orange. Whilst this doesn't answer the broader question around how we manage our switches, it at least keeps the visual cue of "this switch needs immediate attention" that is otherwise missing (expired switches are currently coloured orange).

**before**
image
**after**
image


desbo:
👍

PR #12803, opened 2016-05-05
Extend plista switch

JonNorman:
This switch is used by the AU team to balance the meeting of Outbrain and Plista targets. It needs to be present for the foreseeable future.


regiskuckaertz:
👍

PR #12805, opened 2016-05-06
Adjust layout of commercial components on mobile

regiskuckaertz:
Essentially made sure components fit in a 400px high window by hiding and moving elements around:
picture 2

Now
picture 1

@uplne @piuccio

### DIFF
--- a/static/src/stylesheets/module/commercial/_adverts-soulmates.scss
+++ b/static/src/stylesheets/module/commercial/_adverts-soulmates.scss
@@ -1,4 +1,16 @@
 .adverts--soulmates {
+    @include mq($until: tablet) {
+        > .adverts__header {
+            flex-direction: row;
+            justify-content: space-between;
+        }
+
+        .adverts__row > :nth-child(3),
+        .adverts__row > :nth-child(4) {
+            display: flex;
+        }
+    }
+
     > .adverts__header .inline-logo-soulmates > svg {
         width: 100px;
         height: 21px;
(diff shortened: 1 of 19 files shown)


piuccio:
isn't it too narrow to have two cards on one line?
Anyway it looks much better than the previous solution.

Would it be too hard to bottom align `book now` buttons?

uplne:
Debugging is twice as hard as writing the code in the first place. Therefore, if you write the code as cleverly as possible, you are, by definition, not smart enough to debug it.

regiskuckaertz:
I always keep an ace in my sleeve :-)

@piuccio I could using `margin-top: auto` but then I would lose the minimum top margin between it and what's before.

piuccio:
maybe flexbox?

regiskuckaertz:
im already using it 😃

static/src/stylesheets/module/commercial/_adverts-soulmates.scss
```diff
@@ -96,6 +108,7 @@
 
 .advert-blended--soulmates {
     background: $soulmates-background;
+    box-sizing: border-box;
```
uplne: Do you actually understand what `border-box` does?

regiskuckaertz: In this case the combination of `padding` and `flex-basis` drives Chrome crazy and prevents it from breaking words; `border-box` solves this.

uplne: Great

uplne:
👍

PR #12807, opened 2016-05-06
Fix wrong omniture id for manual multiple

regiskuckaertz:
The Omniture ID for manual components need to report the type of component (multiple/single)

cc @JonNorman @uplne


jbreckmckye:
👍

PR #12808, opened 2016-05-06
Deploys radiator: tidy up TypeScript

OliverJAsh:
Just simplifying bits a little :-)

### DIFF
--- a/static/src/deploys-radiator/app/main.ts
+++ b/static/src/deploys-radiator/app/main.ts
@@ -57,7 +57,7 @@ const run = (): Promise<void> => {
             .sortBy(deploy => deploy.build)
             .first();
 
-        return [ latestCodeDeploy, oldestProdDeploy ];
+        return [ latestCodeDeploy, oldestProdDeploy ] as [ DeployRecord, DeployRecord ];
     });
 
     const buildsPromise = latestDeploysPromise.then(([ latestCodeDeploy, oldestProdDeploy ]) => (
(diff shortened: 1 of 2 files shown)


static/src/deploys-radiator/app/render.ts
```diff
@@ -120,17 +120,12 @@ const renderGroupDeployListNode = (deploys: List<DeployRecord>) => {
...
-    // TODO: Use tuple instead
-    deployPair: Array<DeployRecord>,
-    commits: List<GitHubCommit>
-) => VirtualDOM.VNode =
+const renderPage =
     (
-        [ codeDeploys, prodDeploys ],
-        [ latestCodeDeploy, oldestProdDeploy ],
-        commits
-    ) => {
```
TBonnin: Make sense to me

static/src/deploys-radiator/app/main.ts
```diff
@@ -57,7 +57,7 @@ const run = (): Promise<void> => {
             .sortBy(deploy => deploy.build)
             .first();
 
-        return [ latestCodeDeploy, oldestProdDeploy ];
```
TBonnin: I know it is less code but explicitly showing the return type might be handy. No?

TBonnin:
I am not an expert but it lgtm 👍

static/src/deploys-radiator/app/main.ts
OliverJAsh: Here I have added the return type as a type assertion. TypeScript uses arrays as tuples: without the type assertion it would think this is a type of `Array<DeployRecord>` instead of `[ DeployRecord, DeployRecord ]` (tuple).

TBonnin: Got it. Thx

PR #12811, opened 2016-05-06
New galleries: improve main media quality for portrait devices

OliverJAsh:
When the device is in portrait orientation, we want to constrain the image by height rather than width. As `sizes` only allows widths and not heights, I'm using the technique described at http://codepen.io/tigt/post/when-responsive-images-get-ugly#height-and-width-constrained-srcset
# Before

image
# After

image
# To do
- [x] Replace media query with `orientation: portrait`?
- [x] Update media query for hidpi

/cc @NataliaLKB @paperboyo


common/app/views/fragments/image.scala.html
```diff
@@ -10,6 +10,15 @@
     isImmersiveMainMedia: Boolean = false
 )
 
+@immersivePortraitSource(largestImage: model.ImageAsset, hidpi: Boolean = false) = {
+    <source media="(max-aspect-ratio: 1/1)"
+    sizes="@{100 * (largestImage.width.toDouble / largestImage.height.toDouble)}vh"
+    srcset="@{(500 to 4000 by 250)
```
OliverJAsh: @paperboyo What do you think to these breakpoints?

NataliaLKB:
👍 This looks good to me, but I am not that familiar with the picture element yet. I am happy to deploy it... but could @sndrs or @rich-nguyen take a quick look?

The plan is to test this on gallery immersive media and if it works well then use it on all media.

sndrs:
👍 lgtm

jamesgorrie:
Nice! So often a missed test.

rich-nguyen:
I was still reading the useful docs! looks good though.

paperboyo:
Don’t know if it’s possible to “fix”, but currently _if the image is bound by its height_ (the clever mechanism of expressing height using width kicks in), image is requested at `x px&q=55&dpr=1`, instead of correct `x/2 px&q=20&dpr=2`. This results in images of the same dimensions, but we request a much bigger file (because of q=55)!

To test: go http://www.theguardian.com/artanddesign/gallery/2016/may/07/the-20-photographs-of-the-week

Open devtools in Chrome:

Nexus 5x, portrait (incorrect): https://i.guim.co.uk/img/media/80a99a26bd383997d3c4352504686c60ee67868c/0_160_2016_1210/master/2016.jpg?w=3250&q=55&auto=format&usm=12&fit=max&s=f792690ca16c3fc09777899e10907d81 (image is bound by its original 2016px width, weights 415KB)

Nexus 5x, landscape (correct): https://i.guim.co.uk/img/media/80a99a26bd383997d3c4352504686c60ee67868c/0_160_2016_1210/master/2016.jpg?w=740&q=20&auto=format&usm=12&fit=max&dpr=2&s=08ed3f6677060b4dd14dda10bb285f60 (76KB)

Now - the images shouldn’t be the same weight, because they are requested at different dimensions, but would the first request follow our logic of requesting `q=20&dpr=2` - its size would be (I’ve checked): **149KB instead of 415KB**! (URL: [our imgIX test bucket]80a99a26bd383997d3c4352504686c60ee67868c/0_160_2016_1210/master/2016.jpg?w=1625&q=20&dpr=2&auto=format&usm=12&fit=max)

Now, how to modify the clever mechanism @OliverJAsh found to play ball with our dpr/q ratio logic, I wouldn’t know :dumbpanda: This Picture Element stuff is already so complicated - and it’s before we consider Art Direction :laughing:

common/app/views/fragments/image.scala.html
```diff
@@ -10,6 +10,19 @@
     isImmersiveMainMedia: Boolean = false
 )
 
+@immersivePortraitSource(largestImage: model.ImageAsset, hidpi: Boolean = false) = {
+    <source media="@if(hidpi) {
+        (orientation: portrait) and (-webkit-min-device-pixel-ratio: 1.25), (orientation: portrait) and (min-resolution: 120dpi)
+    } else {
+        (orientation: portrait)
+    }"
+    sizes="@{100 * (largestImage.width.toDouble / largestImage.height.toDouble)}vh"
```
rich-nguyen: Should this be a whole number?

rich-nguyen: (also known as an integer). We could also a helper to `ImageAsset`

common/app/views/fragments/image.scala.html
```diff
@@ -10,6 +10,19 @@
...
+    srcset="@{(500 to 4000 by 250)
```
paperboyo: Of course, I don’t really get what that does, so I can be wrong, but wouldn’t that mean that there are situations when the requested picture may be 250px smaller that it’s currently shown at?

rich-nguyen: That's 15 srcset entries, do we need that many? Also, a comment to explain these numbers might help future readers.

OliverJAsh:
@paperboyo Good catch! I think we just need to fix the order of the `<source>` elements so that the high DPR source element comes first.

@NataliaLKB Want to try that?

common/app/views/fragments/image.scala.html
OliverJAsh: The number of provided images here is totally up for question. In an ideal world we wouldn't have to list specific breakpoints like this, the browser would download the exact size using [client hints](https://developers.google.com/web/updates/2015/09/automating-resource-selection-with-client-hints), but we're not there yet. (Maybe we'll never be due to cache fragmenting issues?) In the interim, the lower the interval we set here (currently 250) the better for the user as it will provide a closer match without rounding up/down to a much smaller/bigger image.

@paperboyo I think the browser rounds to the nearest image, e.g. a screen size of 600px would download the 500px image, whereas a screen size of 700px would download the 750px image. We should test it!

common/app/views/fragments/image.scala.html
```diff
@@ -23,6 +36,13 @@
     @* IE 9 needs this workaround as per https://scottjehl.github.io/picturefill/ *@
     <!--[if IE 9]><video style="display: none;"><![endif]-->
 
+    @if(isImmersiveMainMedia) {
+        @picture.largestImage.map { largestImage =>
+          @immersivePortraitSource(largestImage)
+            @immersivePortraitSource(largestImage, hidpi = true)
```
OliverJAsh: @NataliaLKB I think if you switch the order of these two it will fix the issue @paperboyo mentioned.

common/app/views/fragments/image.scala.html
NataliaLKB: Yes that makes sense! It should be a helper, as we use it enough. I will make a new PR now with these improvements

common/app/views/fragments/image.scala.html
NataliaLKB: Thanks!

common/app/views/fragments/image.scala.html
paperboyo: Apart from https://github.com/guardian/frontend/pull/12727#issuecomment-217738761 (which we shouldn’t really care about) and the size of HTML, what are other downsides of using more finegrained steps? Cache?

I misunderstood. If `srcset="@{(500 to 4000 by 250)` means that there will always be 15 steps, the situation I was afraid of cannot occur. Does it mean we will be providing sources even over the specific image’s maximal resolution (`/master/[width].jpg` part of an URL)? That seems excessive, but may not be a problem if it’s only for immersive content when we target unknown viewport dimensions.


---NOT REALLY RELATED---
This reminds me of another tiny problem. Currently, if specific image’s maximal resolution is less than what could fully satisfy our `w=x&dpr=2&q=20` request, the image gets compressed at `q=20`. This is not a problem if its maximum resolution still satisfies what, say, would be `dpr=1,5`. But some of our imagery, wink, wink, nudge, nudge, do not even satisfy `dpr=1`! :laughing:

To pick on the usual offender, this comic native res is only 650px: http://www.theguardian.com/commentisfree/picture/2016/may/05/steve-bell-on-david-cameron-and-the-london-mayoral-campaign-cartoon

The above means that it will actually look worse on devices with `dpr`>1, because the image dimensions will be the same, but the `q` will be 20...

Not sure how to solve it, if only because the cutoff point would have to be discretionary. Shut upping now...

NataliaLKB: > Does it mean we will be providing sources even over the specific image’s maximal resolution

That is a good point! I am making a change now that will increment by 250 until the largest image width, rather than by 4000

paperboyo: :smiley: But let’s this last step actually be the image’s original width, so that the situation I was afraid of won’t occur (last step being 248px smaller that it could just because image’s widths is `last step-1px`). If that makes sense...

common/app/views/fragments/image.scala.html
NataliaLKB: > Should this be a whole number?

If it was a whole number this is what would happen: 

Say you have an image with a width of 500px and height of 300px. The calculation would then be:

`100 * (500 / 300)vh`

With a double you would get:
`166.66... vh`

With integers it would be:
`100vh` (100 * 1vh)

In some cases it makes sense to use integers, but in this case we need a double (or float?)

PR #12812, opened 2016-05-06
Adding main media for hero image on galleries

NataliaLKB:
Since composer has just launched a change to add main media to galleries, we should use it!

cc @OliverJAsh


NataliaLKB:
@OliverJAsh gave me a 👍  in person yesterday! Merging this today

PR #12814, opened 2016-05-06
Read the series Id from the keywordIds value

regiskuckaertz:
Sponsored series fronts don't have a `seriesId` value, so instead we are looking into `keywordIds` for a  matching keyword starting with `series/`.

cc @JonNorman @rich-nguyen @kenlim


kenlim:
:+1:

PR #12815, opened 2016-05-06
Better js bundle support in dev

rich-nguyen:
We had a problem recently where we couldn't run frontend using requirejs-built javascript. The following changes in this PR are intended to make our asset build a bit more sane:
- `make compile-dev` builds files to `target`, and doesn't affect `hash`
- `make compile` builds files to `target`, and then hashes them into `hash`

Developers can add `assets.useHashedBundles=true` to `frontend.properties`. This means we can simulate prod's asset loading easily. By default, this setting is false, and you will build assets as normal, using `make compile-dev`.

Other changes:
- Do not use the asset map or `hash` build folder when `useHashedBundles` is false. This greatly simplifies dev asset loading, and dev asset building, because all assets will come from the `target`  folder (for css) and `src` folder (for js). 
- Also tidied some of the older parts of the gruntfile and it's configs.
- Removed extra asset-hash steps in gruntfile.
- `Prebid.js` is loaded as a hashed asset url, not a canonical url.
- Removed the requirejs intermediate build folder.
- We no longer copy files into `hash` without `hash`.
- We no longer copy sass .scss files into `hash` and asset map. Weird.
- We now use the XHR `pre-fetch-modules.js` optimiser to preload stuff in both Prod and non-Prod. This reduces the number of dev-mode special cases by one.
- The gulp css tasks now modify the css files in `target`, rather than `hash`.


desbo:
The changes to the build pipeline look good, but isn't it better to run in prod mode with the `stage` command so you get the actual prod app?

rich-nguyen:
Yes, I think that's a valid option too. Part of the motivation of the new property is to clearly mark which parts of the server-side, client-side, and build-time logic are intended for which asset-loading mode. So now there is a clear link between `make compile` with "hashed, js bundles", and `make compile-dev` with "src-loaded js".

The ability to run hashed bundles in dev isn't essential, but means a developer can specifically choose to opt-in to bundle-based js loading. There are lots of other parts of Frontend behaviour which also vary based on Prod mode, which the developer may not care about, or want to configure.

grunt-configs/copy.js
```diff
@@ -4,57 +4,30 @@ module.exports = function (grunt, options) {
...
-                    src: ['**/*.js'],
-                    dest: options.staticTargetDir + 'javascripts/vendor'
+                    dest: options.staticTargetDir + 'javascripts'
                 },
                 {
                     expand: true,
                     cwd: 'static/src/javascripts/vendor',
                     src: [
-                        'formstack-interactive/0.1/boot.js',
-                        'vast-client.js',
```
desbo: not sure what `vast-client` is but did you mean to get rid of it?

desbo:
Apart from that minor comment, looks good! 👍

grunt-configs/copy.js
rich-nguyen: It's a js video library that we don't use anymore.

PR #12822, opened 2016-05-09
Remove List ID requirement in Email Signup

janua:
This changes the `EmailSignupController` so that it no longer requires a valid `listId`.

Now, it will now render forms with any `listId` and accept `POST`s with any `listId`.

I have left the list of `listId`s at the top of the file; even though this will eventually not match what is being used anymore.

@gtrufitt @crifmulholland @desbo


desbo:
👍

PR #12824, opened 2016-05-09
temporary reprieve for server side ab testing framework

johnduffell:
The stats for server side mvt test showed that 8 of the buckets alone were costing $1k+ per month just in bandwidth, so I'm going to remove the rest and we can just add tests specifically for the time they're needed and only against the URLs that are in the tests.

There are a couple of tests ending on 17th so I'll wait for them to finish before updating everything to explain how to do them in future.

### DIFF
--- a/common/app/conf/switches/PerformanceSwitches.scala
+++ b/common/app/conf/switches/PerformanceSwitches.scala
@@ -72,9 +72,9 @@ trait PerformanceSwitches {
   val ServerSideBucketsSwitch = Switch(
     SwitchGroup.Performance,
     "server-side-buckets",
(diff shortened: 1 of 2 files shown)


TBonnin:
IMHO it would be a good thing to unify how we define a/b tests. I think @desbo and @dominickendrick are working on improving our A/B test framework.
I am in charge of the remaining test using bucket 8 and 9. I will remove them when I end the test itself

johnduffell:
Yeah I'll update the docs at some point, unless one of the guys wants to update as part of other tweaking.

johnduffell:
didn't spot your comment about fastly - no it doesn't, until we actually remove the buckets - this is just updating comments effectively

TBonnin:
👍

PR #12825, opened 2016-05-09
If advertisement feature, don't show new gallery design

NataliaLKB:
## What does this change?

The new gallery design affects paid advertisement features, which we were not aware of (sorry 😞 ). They are supposed to look like:

image

And right now look like:
image

The main problems are the tonal changes, the lack of a badge, and the paid for content bar. 

At the moment we are not sure whether we should modify them to use the new design, or if we should keep the old template and styles around just for them.

Until that is decided this PR makes them use the old design and template again.
## Screenshots

They should look the same as they used to, except for the slim header. Like this:

image
## Request for comment

cc @regiskuckaertz do you have time to take a look at this?

### DIFF
--- a/applications/app/views/gallery.scala.html
+++ b/applications/app/views/gallery.scala.html
@@ -1,9 +1,7 @@
 @(page: model.GalleryPage)(implicit request: RequestHeader)
 
-@import conf.switches.Switches.galleryRedesign
-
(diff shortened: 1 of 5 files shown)


regiskuckaertz:
👍  YEAH

PR #12826, opened 2016-05-09
Cache healthcheck requests

TBonnin:
## What does this change?

Second attempt of removing healthcheck from inline requests. 
The first attempt was here https://github.com/guardian/frontend/pull/12797.
The only addition is passing the test port so there is no conflict of address when running tests in parallel 0c9a5e2

The number of healthcheck requests will stay stable and won't increase anymore on the number of when ELBs are scaled up

No


johnduffell:
cool looks fine, I just checked the one commit with the updates in it 😛 👍 thanks! 🚤

PR #12827, opened 2016-05-09
pass in the 'ids' field for the travel component call

JonNorman:
The travel component was ignoring the Vibe IDs passed through from the DFP template.

/@kelvin-chappell


kelvin-chappell:
I assume it works, but is there an easy way of testing it?
👍

JonNorman:
Good point; the proof is in the pudding - this pudding in fact: http://localhost:9000/travel/2016/may/05/the-shack-cafe-surfing-dewata-beach-sri-lanka?adtest=6

image

PR #12830, opened 2016-05-09
Reduce frequency of CSP reports

desbo:
## What does this change?

Limits CSP reporting to every 1 in 100 rather than every request.

We'll be able to turn on CSP reporting for longer to gain more insight into where errors are coming from.
## Request for comment

@TBonnin


TBonnin:
👍 
Do we know how many reports we ingest at the moment?

desbo:
@TBonnin let's go through it later :)

PR #12831, opened 2016-05-09
Fix the download link for mp3s

johnduffell:
We changed the first part of the link in this old commit: https://github.com/guardian/frontend/commit/c94e43a4333ef5703c718f3026787ec325162773#diff-cddbf1a20ebe144ec9aa35d41a46dbb4R462
Not sure why, but that server doesn't have a valid http cert anyway, so updating it to use the new url.

@mchv @JustinPinner

### DIFF
--- a/common/app/model/content.scala
+++ b/common/app/model/content.scala
@@ -501,7 +501,7 @@ object Audio {
 final case class Audio (override val content: Content) extends ContentType {
 
   lazy val downloadUrl: Option[String] = elements.mainAudio
-    .flatMap(_.audio.encodings.find(_.format == "audio/mpeg").map(_.url.replace("static.guim", "download.guardian")))
+    .flatMap(_.audio.encodings.find(_.format == "audio/mpeg").map(_.url))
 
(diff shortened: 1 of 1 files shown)


desbo:
👍

mchv:
@gklopper @jamesgorrie do you remember why we were replacing `static.guim` by `download.guardian`?

jamesgorrie:
@fredex42 - do you know why we did this?

fredex42:
Hi all,

I think it dates back to way back when, when there were seperate CDN contracts for audio/video compared to the static image serving.  download.theguardian.tv was a part of the Level 3 deal for audio/video; I remember a big argument years back after an interactive developer uploaded a bunch of very large video files to static.guim and served them; we nearly blew the monthly bandwidth limit on that contract in a few days (and it wasn't even a popular interactive)

fredex42:
By the way, download.theguardian.tv shows practically zero traffic, or at least it was last time I looked at it.

mchv:
so 👍 for me

PR #12833, opened 2016-05-09
Add commercial audit

kenlim:
Adds audit to site

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/commercial-audit.js
+++ b/static/src/javascripts/projects/common/modules/commercial/commercial-audit.js
@@ -13,22 +13,23 @@ define([
 
         if (config.switches.commercialAudit) {
             window.addEventListener('message', receiveMessage, false);
-
-            function receiveMessage(event) {
-                var origin = event.origin || event.originalEvent.origin;
-                var message = event.data;
-
-
-                if (message.startsWith('Tracker beacon: ')) {
-                    reportError(new Error('###Ad Beacon'), {
-                        feature: 'commercial',
-                        message: message,
-                        source: origin
-                    }, false);
-                }
-                return;
-            }
         }
 
+    }
+
+    function receiveMessage(event) {
+        var origin = event.origin || event.originalEvent.origin;
+        var message = event.data;
+
+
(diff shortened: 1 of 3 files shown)


static/src/javascripts/projects/common/modules/commercial/commercial-audit.js
```diff
@@ -0,0 +1,34 @@
...
+    'common/utils/config',
+    'common/utils/report-error'
+], function (config, reportError) {
+
+    return {
+        init: init
+    };
+
+    function init() {
+        var guardian = window.guardian;
```
regiskuckaertz: You don't need these two lines since you load the module on line 2

static/src/javascripts/projects/common/modules/commercial/commercial-audit.js
```diff
@@ -0,0 +1,34 @@
...
+        }
+
+    }
+
+    function receiveMessage(event) {
+        var origin = event.origin || event.originalEvent.origin;
+        var message = event.data;
+
+
+        if (message.startsWith('Tracker beacon: ')) {
```
regiskuckaertz: `startsWith` is not supported in IE so you might want to use `indexOf` instead:

```
message.indexOf(...) === 0
```

regiskuckaertz: ... also, open question: is this single test secure enough that nobody will be able to exploit it?

Depending on the test duration, this may be futile.

regiskuckaertz:
Aside from minor comments, looks good to me 👍

static/src/javascripts/projects/common/modules/commercial/commercial-audit.js
```diff
@@ -0,0 +1,31 @@
...
+        }
+
+    }
+
+    function receiveMessage(event) {
+        var origin = event.origin || event.originalEvent.origin;
+        var message = event.data;
+
+
+        if (message.indexOf('Tracker beacon: ') === 0) {
```
johnduffell: the PR description isn't too helpful here, but it's causing a JS error, hope this isn't an essential feature
image

PR #12834, opened 2016-05-09
Instances will stop trying as hard when they're overloaded

johnduffell:
This is to allow our servers to survive if they get overwhelmed by traffic.

It works by limiting the amount of work the server will take on to the amount it can complete within a reasonable time.

It makes our servers monitor the latency so far of outstanding requests and also the latency of completed requests.  If either of those go above 1 second, it will limit the number of concurrent requests to a nominal amount to allow recovery.

This does include health checks, so in this state it might fail healthchecks and be taken out of service anyway.  This is intentional as the box isn't coping, but hopefully it can recover in time for the next health check.

Latency is only a proxy for how well the box is coping, as if the upstream has latency issues our boxes will stop serving some requests.  This might need tweaking so it doesn't cause too big problems, maybe 1 second should be longer.

I thought about limiting concurrent requests or by CPU, but since our apps are so different I would have to choose appropriate values for other services.

We could terminate in progress requests if they take too long, this might help the instance become healthy.

None of this detracts from the fact that our apps shouldn't use so many resources and also new instances should come up quicker.  But at least they can come up one at a time without dying again.

The benefit of this is the system will be able to come up under load.  This is especially important when we have had issues which have lost the instances.

@cb372 @markjamesbutler @adamnfish @TBonnin 

I'm going to do a little more testing with JMeter to make sure it behaves as I want before I put it out, as it could all go horribly wrong in prod!

### DIFF
--- a/common/app/conf/Filters.scala
+++ b/common/app/conf/Filters.scala
@@ -1,12 +1,16 @@
 package conf
 
+import akka.agent.Agent
 import cache.SurrogateKey
 import common.ExecutionContexts
 import filters.RequestLoggingFilter
-import org.apache.commons.codec.digest.DigestUtils
-import play.api.mvc.{EssentialFilter, Result, RequestHeader, Filter}
-import play.filters.gzip.GzipFilter
 import implicits.Responses._
+import org.joda.time.DateTime
+import play.api.mvc.{EssentialFilter, Filter, RequestHeader, Result}
+import play.filters.gzip.GzipFilter
+import play.api.mvc.Results.ServiceUnavailable
+
+import scala.collection.immutable.Queue
 import scala.concurrent.Future
 
 object Gzipper extends GzipFilter(shouldGzip = (_, resp) => !resp.isImage)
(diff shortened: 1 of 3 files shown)


TBonnin:
Adding @alexduf to get more eyes on this PR

rich-nguyen:
I guess there could be three scenarios where this system impacts our service:

1. The server recovers, and responds healthily.
2. The server does not recover, and the healthchecks mark the server as unhealthy.
3. The server somewhat recovers, and oscillates around the threshold, throwing a mix of 200s and 500s.

The third state is usually handled by a half-open circuitbreaker-like model, do you think we should consider that here? Otherwise, getting the threshold value absolutely correct is difficult with a two-state model like this one. I like the idea, but I wonder if we get enough benefit for the cost (the cost being complexity). I'm most keen on the warmed-up, server-in-waiting idea.

johnduffell:
Thanks for the comments Rich, I was mainly trying to minimise the complexity while solving the primary problem of getting boxes up under high load.  Certainly consider this a first iteration though.

I'll come and talk to you about half open because I'm not sure what a circuit breaker is in this context.  I was just using the threshold as a way to get at least some requests through so it can recover.  After a couple of seconds of fast requests it should be able to open up again.

I'm happy for it to oscillate for long enough for the ASG to scale up, the main worries are that the latency will be quick enough not to cause a scale up (due to "fast" 50x errors) or that it will still be marked unhealthy due to the percentage of failing healthchecks.

alexduf:
What about tweaking the [netty server](https://www.playframework.com/documentation/2.4.x/ProductionConfiguration#server-configuration-options) rather than adding it at a filter level? You could for instance measure how many connections you can reasonably expect / process per server and set the [netty backlog](http://stackoverflow.com/questions/14075688/what-does-channeloption-so-backlog-do) size accordingly.
You mentioned all micro-services have different needs, equally, this probably doesn't need to be tweaked on all the services but could be done on the most critical one?

I do think the intention behind the PR is more than legitimate but I'm worried it would have uncontrollable side effects like the inability to scale up as you mentioned.

johnduffell:
I had a chat with Rich and now understand the issues and circuit breakers better, at the proposal we really only have an open and closed state for the system, and the opening would work fine, but the closing has issues in that the healthcheck might fail too many times to keep the server alive.
As a result it would be worth having a half/partial open state where it lets through healthchecks plus a proportion of requests.
This might improve latency enough to stop the group auto scaling, so we will need to scale on connections shedded as well as latency in future.

johnduffell:
closing this until I make a better implementation

PR #12836, opened 2016-05-10
Remove video duration metric

jamesgorrie:
## What does this change?

Removes the video duration event from Omniture. We aren't using it and it costs us a bit.

We stop 💸 

Not yet, but they will b e doing the same thing.
## Request for comment

@akash1810 @ScottPainterGNM @mkopka


akash1810:
👍

ScottPainterGNM:
Yesssss!

johnduffell:
I've shipped this to prod, you might want to check it's working right!

PR #12837, opened 2016-05-10
Test Facebook Native Ads

regiskuckaertz:
We're currently testing Facebook Native Ads (FNAs) to see if we can create ads similar to the one we did for Canon.

**This is just a test**, but to get through the approval process it needs to run on theguardian.com. Hence it is hidden behind an `adtest` cookie value of `fbnative`. When this cookie is enabled, we'll insert their SDK and ad units all over the page... and yes, we're making sure the SDK [is not inserted twice](https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/common/modules/identity/facebook-authorizer.js#L95).

Ultimately, if we decide to move forward with it, we'll be using the official application id and introduce some kind of policy around where and in which circumstances FNAs should be inserted, also trying to see if there is a way to throw DFP in the process.

cc @jbreckmckye @uplne @rich-nguyen


regiskuckaertz:
Actually you can add the `localhost` domain and it will work, so I'll raise another PR when I have everything properly set up 😝

static/src/javascripts/projects/common/modules/commercial/third-party-tags/facebook.js
```diff
@@ -0,0 +1,81 @@
...
+        });
+    }
+
+    function insertFbScript() {
+        var ref = document.scripts[0];
+        var script = document.createElement('script');
+        script.id = scriptId;
+        script.src = scriptSrc + '&appId=' + appNetworkId;
+        fastdom.write(function () {
+            ref.parentNode.insertBefore(script, ref);
```
jbreckmckye: I sometimes see async script loaders do this - inserting the new tag before the first script in the document. Any idea why they always insert the script in that position (e.g. other than as the first child of the body)? Just a curiosity.

static/src/javascripts/projects/common/modules/commercial/third-party-tags/facebook.js
```diff
@@ -0,0 +1,81 @@
...
+            adUnit.markup = facebookTpl(adUnit);
+        });
+        fastdom.write(function () {
+            adUnits.forEach(function (adUnit) {
+                adUnit.referenceNode.insertAdjacentHTML(
+                    adUnit.where,
+                    adUnit.markup
+                );
+            });
+            adUnits = null;
```
jbreckmckye: Why do we need to null this? Won't it get GCed automatically?

Calanthe:
Can you also include some screens next time?

static/src/javascripts/projects/common/modules/commercial/third-party-tags/facebook.js
regiskuckaertz: I'm not sure it still applies but appending to the body may result in [an abort exception in IE](https://www.nczonline.net/blog/2008/03/17/the-dreaded-operation-aborted-error/). I like that version because if the script runs then there **must** be a `script` element in the page, it's like _to be or not to be_ in code 😅

static/src/javascripts/projects/common/modules/commercial/third-party-tags/facebook.js
regiskuckaertz: Nope, the module itself stays in memory and so do its local vars. This is fairly insignificant in this case though, but it's a habit I like to keep.

regiskuckaertz:
I will, but in that case the PR was meant to get approved by Facebook via
theguardian.com and nothing would be displayed before this approval.

On Tue, 10 May 2016 at 14:20 Zofia Korcz <notifications@github.com> wrote:

> Can you also include some screens next time?
>
> —
> You are receiving this because you modified the open/close state.
> <https://github.com/guardian/frontend/pull/12837#issuecomment-218154607>
>

-- 

in England Number 908396

PR #12838, opened 2016-05-10
quick switch logic and description change and

Calum-Campbell:
## What does this change?

Quick change to High-Merch-Slot pul request that changes the logic and description of switch. Switch logic is now reversed so that feature is switched ON with the switch turning on.

### DIFF
--- a/common/app/conf/switches/CommercialSwitches.scala
+++ b/common/app/conf/switches/CommercialSwitches.scala
@@ -296,7 +296,7 @@ trait CommercialSwitches {
   val highMerchandisingComponentSwitch = Switch(
     SwitchGroup.Commercial,
     "render-commercial-high-slot-always",
-    "If on, the commercial-high slot will always render on pageload",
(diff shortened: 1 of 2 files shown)


common/app/conf/switches/CommercialSwitches.scala
```diff
@@ -296,7 +296,7 @@ trait CommercialSwitches {
   val highMerchandisingComponentSwitch = Switch(
     SwitchGroup.Commercial,
     "render-commercial-high-slot-always",
-    "If on, the commercial-high slot will always render on pageload",
+    "If on, wil be a server side check of page keywords with high-merchandising keywords, before high-merch slot render.",
```
JonNorman: There is a little typo there, but it might be slightly clearer to reword that as "If on, server will check tags for high-merchandising target before rendering high-merch slot."

JonNorman:
Apart from that small comment (which you can address in GitHub directly if you want to) :+1:

Calum-Campbell:
@JonNorman that is a bit clearer yeah, cheers.

PR #12840, opened 2016-05-10
Remove out-of-page-slot switch

JonNorman:
Adops no longer server surveys through this out-of-page slot and so we no longer need this switch.


jbreckmckye:
👍 

Excellent. Do you know how much money this will end up saving?

PR #12842, opened 2016-05-10
Fabric: Default position of fixed backgrounds

jbreckmckye:
On Android / iOS tablets, the vertical position of a fixed / parallax scroll background is defaulted to the middle. It seems we expect it to default to the top instead, so that's what I'm setting it to, until someone can suggest some instance where we want otherwise.

Before:
image

After:
image


static/src/javascripts/projects/common/modules/commercial/creatives/fabric-expanding-v1.js
```diff
@@ -186,9 +186,11 @@ define([
             showPlus: (this.params.showMoreType === 'plus-only' || this.params.showMoreType === 'plus-and-arrow') ?
             '<button class="ad-exp__close-button ad-exp__open">' + svgs('closeCentralIcon') + '</button>' : ''
         };
+        var parallaxScrollType = this.params.backgroundImagePType;
+        var scrollbgDefaultY = '0%';
         var scrollingbg = {
-            scrollbg: (this.params.backgroundImagePType !== '' || this.params.backgroundImagePType !== 'none') ?
-            '<div class="ad-exp--expand-scrolling-bg" style="background-image: url(' + this.params.backgroundImageP + '); background-position: ' + this.params.backgroundImagePPosition + ' 50%; background-repeat: ' + this.params.backgroundImagePRepeat + ';"></div>' : ''
+            scrollbg: (parallaxScrollType !== '' || parallaxScrollType !== 'none') ?
```
JonNorman: Should this not be an `&&`?
(I'm assuming that the empty string and 'none' are synonymous)

jbreckmckye:
@JonNorman Your comment prompted me to take a look at the actual DFP creative template - I just made the value mandatory, so we needn't test for an empty string in the first place.

PR #12843, opened 2016-05-10
Move interactive bootstraps to the top of the JS app.

gklopper:
## What does this change?

Interactives are content. This pull request prioritises their rendering which should do two things.
1. Make them more stable. The earlier they run the less likely it is something else will interfere with them.
2. Make them render quicker.

This is not quite as ambitious as I hoped, and at one point I had the rendering down to < 1000ms, but I'm not confident that would work properly in PROD.

On my local machine the bootstraps are initialised in 1/2 the time once they are in main.js. Measured by taking the time in a script at the top of the page, and then again immediately after the `interactive.boot(...)` function is called.

boot

No

### DIFF
--- a/static/src/javascripts/bootstraps/standard/main.js
+++ b/static/src/javascripts/bootstraps/standard/main.js
@@ -110,6 +110,27 @@ define([
             }
         });
 
+        /*
+         *  Interactive bootstraps.
+         *
+         *  Interactives are content, we want them booting as soon (and as stable) as possible.
+         */
+
+        if (
+            config.switches.bootInteractivesFromMain &&
+            /Article|Interactive|LiveBlog/.test(config.page.contentType)
+        ) {
+            $('figure.interactive').each(function (el) {
+                require($(el).attr('data-interactive'), function (interactive) {
+                    fastdom.defer(function () {
+                        robust.catchErrorsAndLog('interactive-bootstrap', function () {
+                            interactive.boot(el, document, config, mediator);
+                        });
+                    });
+                });
+            });
+        }
+
         //
         // A/B tests
         //
--- a/common/app/conf/switches/PerformanceSwitches.scala
+++ b/common/app/conf/switches/PerformanceSwitches.scala
@@ -24,6 +24,15 @@ trait PerformanceSwitches {
     exposeClientSide = true
   )
 
+  val BootInteractivesFromMainSwitch = Switch(
+    SwitchGroup.Performance,
+    "boot-interactives-from-main",
+    "If this switch is on then interactive bootstraps will be booted from main.js for perf and stability reasons",
+    safeState = Off,
+    sellByDate = new LocalDate(2016, 6, 8),
+    exposeClientSide = true
+  )
+
   val TagPageSizeSwitch = Switch(
     SwitchGroup.Performance,
     "tag-page-size",
--- a/static/src/javascripts/bootstraps/enhanced/trail.js
+++ b/static/src/javascripts/bootstraps/enhanced/trail.js
@@ -125,7 +125,10 @@ define([
     }
 
     function augmentInteractive() {
-        if (/Article|Interactive|LiveBlog/.test(config.page.contentType)) {
+        if (
+            !config.switches.bootInteractivesFromMain &&
+            /Article|Interactive|LiveBlog/.test(config.page.contentType)
+        ) {
             $('figure.interactive').each(function (el) {
                 fastdom.defer(function () {
                     enhancer.render(el, document, config, mediator);


gklopper:
cc @desbo - had to water it down a bit...

gklopper:
cc @wpf500

wpf500:
This is brilliant. Out of curiosity, what was the sub-second plan that you dropped? Is there any changes we can make at our end to facilitate it?

gklopper:
I had a "require script" inline under each figure...

    <script>
        require('.....', function(){ interactive.boot(...); }
    </script>

Unfortunately this only works in DEV. In PROD we don't have `require` on the page (I know that is a bit of a crappy situation).

sndrs:
one of the reasons they were loaded from `enhanced` before, afaik, is that it means we've done the `ismodern` checks. `standard` js is expected to run everywhere there's a runtime (i.e. manage its own fallbacks etc). 

not a blocker, any maybe intentional, but this will mean all interactives will try to run on all browsers, whereas before, they didn't...

static/src/javascripts/bootstraps/standard/main.js
```diff
@@ -110,6 +110,27 @@ define([
...
+         *  Interactive bootstraps.
+         *
+         *  Interactives are content, we want them booting as soon (and as stable) as possible.
+         */
+
+        if (
+            config.switches.bootInteractivesFromMain &&
+            /Article|Interactive|LiveBlog/.test(config.page.contentType)
+        ) {
+            $('figure.interactive').each(function (el) {
```
sndrs: could this boot code become a module, now it's used in 2 places?

rich-nguyen: We do have `enhancer`, but that's DOM agnostic. Do you mean a frontend module?

sndrs: just meant it's the same code, twice. it's if just a short test and the switch will go, one will win, then np.

gklopper: Once its proved to work in `main.js` it gets deleted from the other module. They will not co-exist for long. Just keeping it there for now so we can switch between implementations.

gklopper: As far as `ismodern` goes, Interactives should run on all browsers, we cannot have blank pages on older browsers. @wpf500 is it the case interactives only work on newer "modern" browsers?

wpf500:
What is the definition of "modern"? I doubt that most of our interactives have ever done any checks, so to be honest some of them probably do just fail with a blank page.

wpf500:
Looking at your docs, looks like `querySelector` is considered a "modern" feature. I'm pretty sure every  interactive I've ever seen uses `querySelector`.

How would this have been failing in the past? I'm guessing if it was in enhanced the interactive never even tried to load?

gklopper:
@wpf500 

https://github.com/guardian/frontend/blob/2e00099b6509024fd5a9f04aa7daea03e08281ac/common/app/templates/inlineJS/blocking/config.scala.js#L7-L16

    var isModernBrowser =
        "querySelector" in document
        && "addEventListener" in window
        && "localStorage" in window
        && "sessionStorage" in window
        && "bind" in Function
        && (
            ("XMLHttpRequest" in window && "withCredentials" in new XMLHttpRequest())
            || "XDomainRequest" in window
        );

Surely you test in IE6 :trollface:

sndrs:
currently, it's this:

```js
var isModernBrowser =
    "querySelector" in document
    && "addEventListener" in window
    && "localStorage" in window
    && "sessionStorage" in window
    && "bind" in Function
    && (
        ("XMLHttpRequest" in window && "withCredentials" in new XMLHttpRequest())
        || "XDomainRequest" in window
    );
```
this is really an implementation detail of the current front end though, it's not a spec for 'guardian javascript'.

that said, we've seen interactives break in the past because they expected the presence of a non-AMD spec implementation detail of `curl` to, which is another detail of frontend. 

i'm sure there will be some that will 'break' as a result of this (where "break" really means "run when they wouldn't have"), was just flagging in case it wasn't anticipated...

wpf500:
I think generally we try to support IE9, but I can't guarantee that everyone does rigorous testing. I'm guessing from a user point-of-view this change doesn't affect what they see (i.e. still nothing)?

gklopper:
`robust.catchErrorsAndLog` will actually tell us where we do have these problems and we can then do something about it (or carry on ignoring it).

I'm going to ship this afternoon, switch on and see what happens.

sndrs:
yeah it think so – well, at worst i guess you go from nothing to something that's half broken. at best though, maybe from nothing to something which works :)

wpf500:
Where does `robust.catchErrorsAndLog` report errors? You got a dashboard somewhere?

All our iframed interactives are loaded using http://interactive.guim.co.uk/embed/iframe-wrapper/0.1/boot.js, which I've just noticed uses some "modern" features pointlessly. I'll look at changing it so at least iframes have got a chance!

wpf500:
Although I'd guess that `window.postMessage` is still not "core" JS? In which case most iframed interactives will look terrible because it won't resize properly anyway.

sndrs:
according to [caniuse](http://caniuse.com/#search=postMessage), guardian is on 98.34%, so should be ok

gklopper:
Went and tried some interactives on IE8 before rolling out new code.

None of them I tried currently work, so I can hardly make it worse (and niether does our image upgrades...)

sndrs:
ie8's not modern, so we won't be trying to load them

wpf500:
@gklopper can you let me know when its out? Thanks again!

gklopper:
@wpf500 have switched it on in PROD. Will take a few minutes to work its way through caches.

PR #12845, opened 2016-05-10
More in series container

jamesgorrie:
## What does this change?

After running [this test](https://github.com/guardian/frontend/pull/12688/files) it seems people really like to go to next in the video series, so we're making it a permanent feature.

People watch way more video and have a better, more informative time on the site.

Nope
## Request for comment

@akash1810 @harrysalmon

### DIFF
--- a/onward/app/controllers/MostViewedVideoController.scala
+++ b/onward/app/controllers/MostViewedVideoController.scala
@@ -16,8 +16,8 @@ object MostViewedVideoController extends Controller with Logging with ExecutionC
 
     getResponse(ContentApiClient.search(edition)
       .tag(series)
-      .showTags("all")
-      .showFields("all")
+      .showTags("series")
+      .showFields("headline")
       .page(page)
       .pageSize(6)
     ).map { response =>
--- a/onward/app/controllers/MostViewedVideoController.scala
+++ b/onward/app/controllers/MostViewedVideoController.scala
@@ -1,6 +1,6 @@
 package controllers
 
-import common.{Edition, ExecutionContexts, JsonComponent, Logging}
+import common._
 import feed.MostViewedVideoAgent
 import model.{Content, Video, Cached}
 import play.api.mvc.{Action, Controller}
@@ -13,8 +13,6 @@ object MostViewedVideoController extends Controller with Logging with ExecutionC
   def renderInSeries(series: String) = Action.async { implicit request =>
     val page = (request.getQueryString("page") getOrElse "1").toInt
     val edition = Edition(request)
-    val nextPage = Some(page + 1)
-    val prevPage = if (page > 1) Some(page - 1) else None
 
     getResponse(ContentApiClient.search(edition)
       .tag(series)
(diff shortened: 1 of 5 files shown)


paperboyo:
> people really like to go to next in the video series

Original Star Trek established that pretty clearly (never saw it). Or The Pickwick Papers (never read it) - before it.

Is this goodness coming also to gallery series? And too, you know, article series?

Also, me and the reader inside me, would really rather know what order are the episodes presented in*. We assume: chronological. Reverse-chronological. But we can't be sure. What if we are in the middle? We like to be sure, otherwise we feel lost. Grumble-grumble.

*If only because we were never sure if we should read comics like
```
A B C
D E F
```
or, rather
```
A C E
B D F
```

jamesgorrie:
@paperboyo we want to get this in as we have baseline for stats.
Then we can test layouts. We're _almost_ certain that a more playlisty type style would be better - but let's test.

paperboyo:
Sure! I was just for consistency of experience and information architecture (and against fixing stuff later, meaning: never :wink:). Whatever makes dents in the high walls we’ve built in between items is a friend of mine!

jamesgorrie:
> and against fixing stuff later, meaning: never :wink:

I am sad that that is the sentiment. We have proven that this is a better experience for our audience, and, thus, us. We should definitely continue to evolve the way we can make sure people get more relevant video content that they would want to watch.

paperboyo:
@jamesgorrie There is nothing in what you've said that I could possibly disagree with (and I love disagreeing!). My, personal, isolated, emoticonicised point was just to say that we do have serialised content of other types that could, much probably, gain from all/most of that research. And then - treatment. Peace!

jamesgorrie:
> we do have serialised content of other types that could, much probably, gain from all/most of that research. And then - treatment. Peace!

As I mentioned  to @zeftilldeath and @NataliaLKB the research is interesting, but shouldn't be applied to other pieces of content without it's own testing. The reason we used most popular in the first place was it performed well with articles. With videos it's another story.

Perhaps with galleries it might be galleries with similar colour tones? Basically the media types should be a bit more decoupled due to different consumption habits.

paperboyo:
Probably maybe. Worth testing. I wouldn't agree that the media types should be decoupled, though. Definitely not the ones that you know I strongly feel should be more coupled 😆

Series mechanics, while driven by possibly, maybe, probably, different drivers should work and look consistently for all the series though. If it turns out it shouldn't for some - they should not be called series, then.

jamesgorrie:
>  I strongly feel should be more coupled

I understand this - but think that a media type, video for example, should have it's own treatment, functionality and surroundings related to video, but also be able to be embedded into multi-media-type environments (perhaps with images, articles or something like that 😉?).

> consistently for all the series though

I disagree - video, audio, article, insert-new-media-type-here should be tested separately as they have been, and one thing works for articles, the other for video.

> hijacking

Discussions = good.

akash1810:
👍

paperboyo:
Hello,

To fight our always-too-poor information infrastructure, I would like to propose for the name in the series in the new container to be a link to the series tag page (nobody goes to, I know).

link more

As a minor point, on pagination action the whole container redraws, it would be nicer for just the contents of it to redraw (and for the pagination to provide a sense of place and breadth, although that may be to cluttered...).

pagination

Regards
Mateusz

PR #12850, opened 2016-05-11
Send CSP reports to elasticsearch instead of DynamoDB

desbo:
A second attempt at storing these. Dynamo was too restrictive for the type of queries we want to do.

@TBonnin

### DIFF
--- a/diagnostics/app/controllers/DiagnosticsController.scala
+++ b/diagnostics/app/controllers/DiagnosticsController.scala
@@ -43,7 +43,7 @@ object DiagnosticsController extends Controller with Logging {
   }
 
   def csp = Action(jsonParser) { implicit request =>
-    if (conf.switches.Switches.CspReporting.isSwitchedOn && r.nextInt(100) == 1) {
+    if (conf.switches.Switches.CspReporting.isSwitchedOn) {
       CSP.report(request.body)
     }
 
--- a/diagnostics/app/model/diagnostics/csp/CSP.scala
+++ b/diagnostics/app/model/diagnostics/csp/CSP.scala
@@ -1,24 +1,42 @@
 package model.diagnostics.csp
 
-import model.diagnostics.CSPDynamoDbReport
 import play.api.libs.functional.syntax._
 import play.api.libs.json._
+import common.Logging
+import net.logstash.logback.marker.Markers._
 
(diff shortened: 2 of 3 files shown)


diagnostics/app/controllers/DiagnosticsController.scala
```diff
@@ -43,7 +43,7 @@ object DiagnosticsController extends Controller with Logging {
   }
 
   def csp = Action(jsonParser) { implicit request =>
-    if (conf.switches.Switches.CspReporting.isSwitchedOn && r.nextInt(100) == 1) {
```
TBonnin: Let's check how many reports are ingested into ES when we turn this on. 
I don't think the volume of reports would be big deal but it worths checking

diagnostics/app/model/diagnostics/csp/CSP.scala
```diff
@@ -1,24 +1,42 @@
...
     )(CSPReport.apply _)
+
+  implicit val jsonWrites: Writes[CSPReport] = (
+    (__ \ "document-uri").write[String] and
+      (__ \ "referrer").write[String] and
+      (__ \ "blocked-uri").write[String] and
+      (__ \ "violated-directive").write[String] and
+      (__ \ "effective-directive").write[String] and
+      (__ \ "original-policy").write[String]
+    )(unlift(CSPReport.unapply))
```
TBonnin: I wonder if there is a more concise way to do this :)

TBonnin:
👍

diagnostics/app/model/diagnostics/csp/CSP.scala
desbo: https://www.playframework.com/documentation/2.1.1/ScalaJsonInception 8)

TBonnin: Yeah I know but here we are specifying every single field because our class parameters doesn't spell the same as the json fields

desbo: Ah yeah. I guess there's no way to do it unless you have matching field names

TBonnin:
When you enable this in prod @desbo, can you monitor the ELK kinesis stream to see if it impacts the write throughput? thx

PR #12851, opened 2016-05-11
Remove Ad Feature MPU

regiskuckaertz:
Confirmed with Adops UK/US/AU that these creatives are not used anymore (there are some remaining items in DFP; they're all inactive and it's their job to get rid of them).

@JonNorman @uplne @jbreckmckye @Calanthe


uplne:
👍

PR #12852, opened 2016-05-11
Move omniture to single call

dominickendrick:
## What does this change?

This PR removes the top up call from omniture 
- all existing data is passed as part of the first call
- apart from MVT data which is no longer passed
- ad blocking is also removed
- video tracking is now inlined into the template rather than relying on the top up call
- fixes broken/obtuse omniture tests !

It saves £7k per month of omniture fees 

No
## QA steps
- [x] video tracking
  check localhost:9000/embed/video/politics/video/2016/apr/11/dennis-skinner-suspended-for-dodgy-cameron-comments-video
- [x] registered / guest tracking : prop2, eVar2, prop31 , eVar31
- [x] logged in 
- [x] logged out 
- [x] nav interaction : eVar24, eVar37
- [x] sponsored content : prop38 
## Request for comment

@sndrs @rich-nguyen


static/src/javascripts/components/analytics/analytics.js
```diff
@@ -182,39 +181,71 @@ try {
...
-} catch(e) {
-    (new Image()).src = '@{Configuration.debug.beaconUrl}/count/omniture-pageview-error.gif';
-}
-
-
+    } catch(e) {
+        (new Image()).src = guardian.config.page.beaconUrl + '/count/omniture-pageview-error.gif';
+    }
+
+    return {
```
gtrufitt: If these are exclusively for testing, it might be worth identifying them as such. I don't think there is a standard approach, some return a test object that reveal the functions](https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/common/utils/cookies.js#L108), others use `_funcName`. @regiskuckaertz did we ever decide on an approach to revealing test-only methods?

common/app/model/content.scala
```diff
@@ -552,7 +552,7 @@ object Video {
 
 final case class Video (
   override val content: Content,
-  source: Option[String] ) extends ContentType {
+  source: Option[String] ) extends ContentType with Page {
```
rich-nguyen: Don't think we want to combine the concept of a `ContentType` with a `Page`, because we shouldn't always assume a capi item is being used in the context of a page. It could be a trail, for example, or a related item, etc.

See above...

applications/app/views/videoEmbed.scala.html
```diff
@@ -124,5 +124,10 @@
                 </script>
             }
         }
+
+        @{
+            page.model.map(videoPage => fragments.analytics.base(videoPage))
```
rich-nguyen: Shouldn't `page` be passed to `base`? To make that possible, `EmbedPage` could inherit from either `StandalonePage` or `ContentPage`. I notice `EmbedPage` has an optional model, if we could tweak this to be non-optional, we could easily write `EmbedPage extends ContentPage`

static/src/javascripts/components/analytics/analytics.js
regiskuckaertz: Nope, not really. I started using `@@tests` but I'm not sure this is the best approach.

jamesgorrie:
Less 💸 = More 👍

static/src/javascripts/components/analytics/analytics.js
```diff
@@ -1,17 +1,48 @@
-@()
-@import conf.Static
-@import org.joda.time.DateTime
-
-try {
+//This uses the named module pattern : https://github.com/umdjs/umd/blob/master/templates/amdWebGlobal.js
```
sndrs: `components` is where bower components are extracted to (unclear I know), probably an unsafe place to keep this file

static/src/javascripts/components/analytics/analytics.js
sndrs: @piuccio always makes a good case for breaking them out into modules, I think, if you need to test them

sndrs: either that or consider these methods internal implementation details, and test the overall effect of the module?

static/src/javascripts/components/analytics/analytics.js
jamesgorrie: Good catch - should it be in modules?

rich-nguyen:
Server changes look good 👍

sndrs:
👍

PR #12853, opened 2016-05-11
Re-enable single-campaign commercial container

kelvin-chappell:
Now using the new model and template.

This automates what we've been doing to fix single-campaign containers.
Still to establish definitive business rules around these containers.

/cc @JonNorman


JonNorman:
:+1:

PR #12854, opened 2016-05-11
Update ab-testing.md

joelochlann:
When writing an ab test, it's incredibly easy to be tripped up by the undocumented fact that if you don't follow a certain naming convention for your switch id and test id,  then [this code](https://github.com/guardian/frontend/blob/164820e060efa6df45c92c79f6bb3a2a6ff611f1/static/src/javascripts/projects/common/modules/experiments/ab.js#L92-L109) will stop anyone from participating in the test.

So I thought I would put some notes in the `ab-testing.md` docs to flag up this fact. A better option in the future might be to make the naming requirements less strict.

cc @annebyrne

### DIFF
--- a/docs/ab-testing.md
+++ b/docs/ab-testing.md
@@ -27,7 +27,10 @@ val ABFontDelaySwitch = Switch("A/B Tests", "ab-web-fonts-delay",
     safeState = Off)
 ```
 
-The only convention is that the test id has to start with the characters _'ab-'_.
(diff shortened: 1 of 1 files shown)


gtrufitt:
👍

rich-nguyen:
Thanks for updating this. Agreed, maybe the test configuration should require an explicit reference to the switch that's controlling the test.

PR #12855, opened 2016-05-11
Refactor CachedHealthCheck

TBonnin:
## What does this change?

While looking into applying cached healthcheck to the other services, I
figured out I could simplify the way CachedHealthCheck is structured.
See benefits below

A PR that moves all services to cached healthcheck will follow

This patch has the following benefits:
- Create abstract Trait so services like facia-press can define their own healthcheck custom logic
- Avoid overriding trait vals
- Easier to define service's own healthcheck controller object

No
## Request for comment

@johnduffell

### DIFF
--- a/common/app/conf/HealthCheck.scala
+++ b/common/app/conf/HealthCheck.scala
@@ -81,49 +81,64 @@ private[conf] trait HealthCheckCache extends HealthCheckFetcher {
   protected val cache = AkkaAgent[List[HealthCheckResult]](List[HealthCheckResult]())
   def get() = cache.get()
 
-  def allSuccessful: Boolean = {
-    get() match {
-      case Nil => false
-      case nonEmpty => nonEmpty.forall(_.recentlySucceed)
-    }
-  }
-  def anySuccessful: Boolean = get().exists(_.recentlySucceed)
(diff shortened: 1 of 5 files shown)


common/app/conf/HealthCheck.scala
```diff
@@ -81,49 +81,64 @@ private[conf] trait HealthCheckCache extends HealthCheckFetcher {
...
     fetchResults(testPort, paths:_*).map(allResults => cache.send(allResults.toList))
   }
-}
 
-trait CachedHealthCheckController extends Controller with Results with ExecutionContexts with Logging {
+}
 
-  val paths: Seq[String]
-  val testPort: Int
+sealed trait Policy
```
jamespamplin: Good work using Types instead of a boolean for this. Although, `Policy` will be defined on the package here, which is quite a generic name. May be slightly better to call it `HealthCheckPolicy`

jamespamplin:
:+1: small suggestion, but otherwise good work

common/app/conf/HealthCheck.scala
TBonnin: good point. I will change

PR #12857, opened 2016-05-11
Clever friend brexit

annebyrne:
## What does this change?

We are testing a new format as part of the Clever Friend membership proposition. They are manually embedded via Composer as Interactives into EU Referendum articles. Screenshot below:

picture 1

We want to limit exposure to a small subset of users (~ 5%) before we begin testing impact & engagement. Simplest way to do that is to segment using the A/B testing framework.

Since it is inserted by Composer it will be in the page by default, and so we need to remove it for the majority of users. We have set it to 0% of users at present, and will go back and amend after testing it live.

No it only affects web.

cc @joelochlann


static/src/javascripts/projects/common/modules/experiments/tests/clever-friend-brexit.js
```diff
@@ -0,0 +1,31 @@
...
+        this.canRun = function () {
+            return true;
+        };
+
+        this.variants = [{
+            id: 'remove-embed',
+            test: function() {
+                var companions = document.querySelectorAll('figure[data-canonical-url^="https://interactive.guim.co.uk/2016/05/brexit-companion/"]');
+                for (var i = 0; i < companions.length; i++) {
+                    companions[i].remove();
```
piuccio: `remove` is [not supported by IE](https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/remove). You might want to use `companions[i].parentNode.removeChild(companions[i])`

joelochlann:
@piuccio is this OK to merge now?

PR #12861, opened 2016-05-11
Fix potential race condition in CachedHealthCheckTest

TBonnin:
## What does this change?

Tests wait for the Future returned by `runChecks` to be completed before
to execute the tests. However since the cache value is dispatched to the
AkkaAgent, it was possible for the initial Future to complete without
the AkkaAgent to be fully updated
This patch ensures this doesn't happen

Make sure CachedHealthCheckTest do not randomly fail

No
## Screenshots

No
## Request for comment

@alexduf


alexduf:
Looks good to me :+1:

PR #12862, opened 2016-05-11
Aus live blog background

blongden73:
Adds the background to the live blog when the tag is australia election this is for both live and deadblogs.
Also makes the logo specific to the aus election rather than using the css for the us election. 

Request feedback
@NataliaLKB @rich-nguyen


common/app/views/fragments/containers/facia_cards/standardHeaderMeta.scala.html
```diff
@@ -29,7 +29,7 @@
                                 </div>
                             }
                             @if(isAUSElection) {
-                                <div class="badge-slot badge-slot--us-election">
+                                <div class="badge-slot badge-slot--aus-election">
```
sammorrisdesign: Do we need a separate elements for this? Couldn't we just have a generic `badge-slot` and change the image inside it?

static/src/stylesheets/module/content/_badging.scss
```diff
@@ -61,3 +88,39 @@
     }
 }
 
+/* AUS Election Liveblog Image
+   ========================================================================== */
+.content__head--aus-election {
```
sammorrisdesign: Yeah most of this is duplicating what we have above. Not sure we need to create a variant especially as this looks like it's the same but with a different image

blongden73: Agreed I wanted to just get something up then de dupe and make more generic in a separate thing but i guess i can do it here.

static/src/stylesheets/module/content/_badging.scss
```diff
@@ -61,3 +88,39 @@
...
+    .content__header .gs-container {
+        background-image: url('https://uploads.guim.co.uk/2016/05/11/Election_liveblog.jpg');
```
desbo: Can you remove the images from this commit if you're using this URL?

common/app/model/Badges.scala
```diff
@@ -0,0 +1,16 @@
+package model
+
+import conf.{Configuration, Static}
+import layout.FaciaContainer
+
+case class Badge(seriesTag: String, imageUrl: String)
+
+object Badges {
+  val usElection = Badge("us-news/us-elections-2016", s"${Configuration.static.path}/sys-images/Guardian/Pix/pictures/2016/2/2/1454424596176/USElectionlogooffset.png")
+  val ausElection = Badge("australia-news/australian-election-2016", Static("images/AUSElectionBadge.png").toString)
```
janua: `toString` !?

common/app/views/fragments/meta/metaInline.scala.html
```diff
@@ -1,49 +1,37 @@
...
+<div class="content__labels @if(item.content.tags.isUSElection) {content__labels--us-election} @if(item.content.tags.isAusElection) {content__labels--aus-election} @if(item.content.showNewGalleryDesign) {content__labels--gallery}">
+    @if(USElectionSwitch.isSwitchedOn || AUSElectionSwitch.isSwitchedOn) {
 
-        @if(isAUSElection) {
-            <div class="badge-slot badge-slot--us-election">
+        @badgeFor(item).map { badge =>
+            <div class="badge-slot badge-slot--election">
                 @if(isAmp) {
-                    <amp-img class="badge-slot__img" src="@Static("images/AUSElectionBadge.png")" layout="fixed" height="33" width="33"></amp-img>
+                    <amp-img class="badge-slot__img" src="@badge.imageUrl" layout="fixed" height="33" width="33"></amp-img>
```
janua: Why move the `Static` to `Badge` when you could just leave it here?

janua: Oh, it's because you have two different image locations.

common/app/model/Badges.scala
janua: Why can't the image of these badges come from the same location?

common/app/views/fragments/meta/metaInline.scala.html
desbo: Yeah, the stuff I did here is so we can add new badges without having to change templates.

janua:
Looks good! :+1:

PR #12864, opened 2016-05-11
Play 2.5 environment

alexduf:
## What does this change?

This change is a first step towards play 2.5. It gets rid of `play.current.mode` as this will be deprecated.
The implementation works around the problem by introducing a global state, please let me know if that's causing you any worry.

The value is an up-to-date codebase, no measure of success

No
## Screenshots

N/A
## Request for comment

@rich-nguyen @johnduffell @mchv @TBonnin @jamespamplin

### DIFF
--- a/common/app/common/ApplicationMode.scala
+++ b/common/app/common/ApplicationMode.scala
@@ -15,5 +15,6 @@ trait ApplicationMode extends GlobalSettings with Logging {
   override def onStart(app: Application): Unit = {
     log.info(s"Setting the application mode to ${app.mode}")
     ApplicationMode.setMode(app.mode)
+    super.onStart(app)
   }
 }
--- a/common/app/common/ApplicationMode.scala
+++ b/common/app/common/ApplicationMode.scala
@@ -1,0 +1,19 @@
+package common
+
+import play.api.{Application, GlobalSettings}
+import play.api.Mode._
+
+object ApplicationMode {
+  private var theMode = Prod
+  def mode: Mode = theMode
+  def setMode(theMode: Mode) = {
+    this.theMode = theMode
+  }
+}
+
+trait ApplicationMode extends GlobalSettings with Logging {
+  override def onStart(app: Application): Unit = {
+    log.info(s"Setting the application mode to ${app.mode}")
+    ApplicationMode.setMode(app.mode)
+  }
+}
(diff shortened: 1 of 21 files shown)


common/app/common/ApplicationMode.scala
```diff
@@ -0,0 +1,19 @@
+package common
+
+import play.api.{Application, GlobalSettings}
+import play.api.Mode._
+
+object ApplicationMode {
+  private var theMode = Prod
```
johnduffell: var? can you just inline this anyway?

johnduffell: ok I understand this now 😛 , although could we make this an option to avoid a default value?

johnduffell: then we can stop it being set more than once too (although it's all getting a bit javay at that point!)

johnduffell:
yep looks good to me! 👍 🐎  thanks

TBonnin:
@johnduffell made good comments (no default value and set only once). 
LGTM though 👍

mchv:
@alexduf approach seems very sensible to me!

common/app/common/ApplicationMode.scala
alexduf: What would be the behaviour if the mode is requested but no value is found?

rich-nguyen: I don't think Option works, here. Surely we need to simplify things, so that any code that references the mode, now has access to the application?

alexduf: I agree on not using an option with a defensive default value. It keeps the logic simple and ensure if we ever run into a problem, it'll be either when testing or coding locally (but not in prod)

You're suggesting we keep a global reference to a play.api.Application instead of the Mode? That would simplify the migration but I believe play 2.5 is getting away from that object that rules them all

rich-nguyen: it could be a global object, that supports DI-injected Application. Slightly perverse, but it means we can actually make a choice about how we want to use DI, rather than have it forced upon us.

johnduffell: I just preferred no default value because it shouldn't be used anyway.  e.g. if onStart doesn't get called, local dev will be running in prod mode and will start sending metrics to cloudwatch.  I'm not sure "Prod" could be classed as defensive.

alexduf: I see, an option defaulted to None would make sense then

alexduf:
There's a flaw in that solution: the tests.
Since there's not central place that starts the test I would have to manually set the mode of the app in every test that indirectly uses the mode. Plus it's a global state so I could fix it somewhere and given the order of the tests that may hide the problem somewhere else.

Back to square one.

johnduffell:
😢

PR #12866, opened 2016-05-11
Extending headline ab test for another month

dominickendrick:
## What does this change?

Extending headline ab test for another month

The team in the US are garenteeing more testing data for this, so i'm extending this for now
## Request for comment

@desbo


desbo:
👍

PR #12867, opened 2016-05-11
Cached healthcheck on all services

TBonnin:
## What does this change?

Follow-up on #12826
Caching healthCheck results has been running on Facia PROD for 2 days without issues.
This PR enables the caching mechanism for all services healthcheck 

Less load when ELBs are requesting /_healthcheck since we don't inline the internal request but serve the cached results (which are pulled every 5 seconds)

No
## Request for comment

@johnduffell


admin/app/controllers/HealthCheck.scala
```diff
@@ -1,17 +1,9 @@
...
 
-  // Unlike other health checks, Admin is initially set to healthy, and may move into
-  // an unhealthy state, which it will remain in until terminated.
-  status.set(true)
-
-  override def healthcheck() = if (!isOk) {
-    Action(InternalServerError("JAXP00010001"))
-  } else {
-    super.healthcheck()
-  }
```
TBonnin: The infamous `JAXP00010001` has been fixed: https://bugs.openjdk.java.net/browse/JDK-8028111

Note: we are running JDK 8u77

admin/test/services/AdminHealthcheckTest.scala
```diff
@@ -1,16 +0,0 @@
...
-import scala.concurrent.Await
-import scala.concurrent.duration._
-
-@DoNotDiscover class AdminHealthCheckTest extends FlatSpec with Matchers with ConfiguredTestSuite with TestWsConfig {
-
-  "Healthchecks" should "pass" in goTo("/admin"){ browser =>
-
-    Await.result(WS.clientUrl(withHost(s"/_healthcheck"))(longTimeoutConfig).get(), 10.seconds).status should be (200)
-  }
-}
```
TBonnin: All previous HealthCheckTest are redundant, we are already testing the underlying CachedHealthCheck https://github.com/guardian/frontend/blob/master/common/test/conf/CachedHealthCheckTest.scala

article/conf/routes
```diff
@@ -5,9 +5,9 @@
 # For dev machines
 GET     /assets/*path               dev.DevAssetsController.at(path)
 
-GET     /_healthcheck               conf.HealthCheck.healthcheck()
+GET     /_healthcheck               conf.HealthCheck.healthCheck()
 
-GET     /_cdn_healthcheck           controllers.CdnHealthcheckController.healthcheck()
+GET     /_cdn_healthcheck           conf.HealthCheck.healthCheck()
```
TBonnin: Since we cache the internal healthcheck results, we can use the same method for normal and cdn healthcheck. Eventually we can get rid of the /_cdn_healthcheck endpoint

preview/app/Global.scala
```diff
@@ -0,0 +1,3 @@
+import controllers.PreviewHealthCheckLifeCycle
+
+object Global extends StandaloneGlobal with PreviewHealthCheckLifeCycle
```
TBonnin: Necessary to give preview and training-preview a different HealthCheckLifeCycle trait

TBonnin:
I concede this is a very boring PR to review :)

article/conf/routes
johnduffell: might be worth putting that comment in the file itself as if someone else comes across this file they won't find this PR

johnduffell:
good work👍 go for it! 🚤

PR #12869, opened 2016-05-12
Extending the expired ab-minute test

NataliaLKB:
Extending `ab-minute` test which is currently breaking the build. I extended it until Monday.

cc @jamesgorrie


katebee:
👍

PR #12832, opened 2016-05-09
Hosted content

uplne:
## What does this change?

New format. We are going to try some new things. It will involve a lot of AB testing etc. Awesome times ahead.
## What is the value of this and can you measure success?

It will eventually replace @regiskuckaertz 
## Does this affect other platforms - Amp, Apps, etc?

Eventually will

@Calanthe @lps88

PR #12872, opened 2016-05-12
The .adverts class renaming

Calanthe:
## What does this change?

The hosted page (https://github.com/guardian/frontend/pull/12832) is entirely blocked by an adblock because we named the main css class as `adverts`. My proposition is to rename the `adverts` into `commodel` (because it's a box model for the commercial content). The test page, glabs page and unit tests seem to work fine. I don't know what else should I test.  

The hosted page won't be blocked by adblock.

no
## Request for comment

@uplne @regiskuckaertz @jbreckmckye

### DIFF
--- a/commercial/app/views/books/booksStandard.scala.html
+++ b/commercial/app/views/books/booksStandard.scala.html
@@ -11,7 +11,7 @@
                   dataLinkName = omnitureId.map(id => s"merchandising | books standard | ${id}"),
                   optActions = Some(actions)){
 
-    <a class="adverts__logo" href="@{clickMacro}http://www.guardianbookshop.co.uk/" data-link-name="merchandising-bookshop-v2_0_2014-10-15-medium-visit-shop">
+    <a class="commodel__logo" href="@{clickMacro}http://www.guardianbookshop.co.uk/" data-link-name="merchandising-bookshop-v2_0_2014-10-15-medium-visit-shop">
         @fragments.inlineSvg("marque-54", "icon")
         @fragments.inlineSvg("logo-guardian", "logo")
         @fragments.inlineSvg("logo-bookshop", "commercial", List("inline-commercial-brand"))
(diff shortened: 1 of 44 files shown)


regiskuckaertz:
Ace! What about `.advert` without an `s`? Shouldn't it be renamed too?

uplne:
I still think that `.thor` would be the best name

Calanthe:
@regiskuckaertz I was thinking about this, but I don't know. Also, should I rename the files too?

This seems like a bigger work and I really would like to focus on the Glabs stuff.

regiskuckaertz:
The thing is `.advert` is used in glabs containers.

On Thu, 12 May 2016 at 12:15 Zofia Korcz <notifications@github.com> wrote:

> @regiskuckaertz <https://github.com/regiskuckaertz> I was thinking about
> this, but I don't know. Also, should I rename the files too?
>
> This seems like a bigger work and I really would like to focus on the
> Glabs stuff.
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/guardian/frontend/pull/12872#issuecomment-218728334>
>

-- 

in England Number 908396

jbreckmckye:
One advantage of this change is that the classnames and templates are now much easier to grep.

In the long term, I think we will have to find a way to style these elements without content-derived classes - likely inline styles generated by JavaScript. Otherwise all it takes to hide these elements again is someone using ABP's block element dialog to generate a new EasyList rule:

image

regiskuckaertz:
This game of cat and play with adblockers is really tiring 😭 

Here is an evil solution: Let's find out a class name used for some very important content on another high-reputation site and use it for our adverts.

regiskuckaertz:
@uplne I'm sure there is a super villain in the marvel universe that would be a good fit for adverts 😆

uplne:
@regiskuckaertz that is great idea

uplne:
`.gsfi` google search box

jbreckmckye:
@regiskuckaertz Sadly, EasyList rules are scoped by domain!

There are alternatives, though. We could use atomic / functional CSS classes, for example, which would foil adblockers as soon as the message's design changed. We could also style the component entirely through JavaScript.

The important thing is that there's no route to the element with content-specific attributes. Unfortunately, this subverts a lot of the practices we've come to rely upon - like `js-` classes for DOM selection and `data-link`s for analytics. Even without classes and IDs, adblockers could still target elements via XPath, but we could at least randomise the DOM structure - or make it so shallow that any ABP selectors would sooner or later hit other content as collateral damage.

Calanthe:
Ok, I will rename it properly with the `.advert` class as well. But I will do it when the test page will be working correctly again.

PR #12873, opened 2016-05-12
remove minute test

jamesgorrie:
Remove the minute test - we have no stats as it was a 0% test to check viability.

sorry @NataliaLKB 
cc @akash1810


akash1810:
why so early?

jamesgorrie:
It's been running for a month.
Best we decide what to do with it, We can add it back if we need it.

static/src/javascripts/projects/common/modules/experiments/ab.js
```diff
@@ -61,10 +59,14 @@ define([
         new LiveBlogChromeNotificationsInternal(),
         new LiveBlogChromeNotificationsProd(),
         new LoyalAdblockingSurvey(),
+<<<<<<< HEAD
```
akash1810: the merge didn't work properly :'(

akash1810:
other than a bad merge, 👍

akash1810:
👍

PR #12874, opened 2016-05-12
Set the correct audience segment size for hiding clever friend

joelochlann:
Turning on the "test" (more of soft launch really) added in #12857, exposing the clever friend interactive to just 10% of people by removing it for 90%. Also amended the description to make it clearer what we're doing here.

### DIFF
--- a/static/src/javascripts/projects/common/modules/experiments/tests/clever-friend-brexit.js
+++ b/static/src/javascripts/projects/common/modules/experiments/tests/clever-friend-brexit.js
@@ -6,13 +6,13 @@ define([
         this.start = '2016-05-09';
         this.expiry = '2016-07-31';
         this.author = 'Anne Byrne';
-        this.description = 'Segmentation to target users with Clever Friend';
-        this.audience = 0.0;
+        this.description = 'Only expose the clever friend embed to 10% of people, by removing it for 90%';
+        this.audience = 0.9;
         this.audienceOffset = 0.0;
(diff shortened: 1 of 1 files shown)


static/src/javascripts/projects/common/modules/experiments/tests/clever-friend-brexit.js
```diff
@@ -6,13 +6,13 @@ define([
         this.start = '2016-05-09';
         this.expiry = '2016-07-31';
         this.author = 'Anne Byrne';
-        this.description = 'Segmentation to target users with Clever Friend';
-        this.audience = 0.0;
+        this.description = 'Only expose the clever friend embed to 10% of people, by removing it for 90%';
```
jbreckmckye: Your description implies you're now showing the companion to 90%. Is that incorrect?

joelochlann: No, we're showing it to 10%. It's an interactive embedded via composer so it will be in the article by default. We have to remove it for 90% of people. A bit confusing, I agree, but seemed like the simplest way to achieve what we wanted

joelochlann: Amended wording of my comment above, I can see it was misleading!

jbreckmckye:
👍

Let's ship it.

joelochlann:
🙌 🙌 🙌

PR #12875, opened 2016-05-12
request video content type for more in series

akash1810:
## What does this change?

More in series playlist only shows video.

The "How South Wales" item without a thumbnail below links to this article - http://www.theguardian.com/commentisfree/2012/may/03/credit-unions-evangelical-finance

screen shot 2016-05-12 at 12 34 32

Better experience.

No.
## Screenshots

n/a
## Request for comment

@jamesgorrie


jamesgorrie:
I was just about to do this. 👍

PR #12876, opened 2016-05-12
Show capi-sourced logos in branded containers

kelvin-chappell:
Eg.

image

When switch is on, logo will be served directly in template.  Otherwise it will be served from DFP as now.

/cc @steppenwells

### DIFF
--- a/common/app/common/commercial/BrandHunter.scala
+++ b/common/app/common/commercial/BrandHunter.scala
@@ -24,13 +24,11 @@ object BrandHunter {
     findBranding(frontProps.activeBrandings, publicationDate = None, edition)
   }
 
-  // difficult to find content's section at the moment so this param is temporarily optional
-  def findContentBranding(section: Option[Section],
-                          tags: Tags,
-                          publicationDate: Option[DateTime],
-                          edition: Edition): Option[Branding] = {
-    lazy val brandingBySection = section flatMap (findSectionBranding(_, publicationDate, edition))
-    lazy val brandingByTags = tags.tags.flatMap(findTagBranding(_, publicationDate, edition)).headOption
-    brandingBySection orElse brandingByTags
+  def findContentBranding(content: ContentType, edition: Edition): Option[Branding] = {
+    // TODO: incorporate branding by section - needs capi change to get section element in content results
+    val tags = content.tags
+    val publicationDate = Some(content.trail.webPublicationDate)
+    val brandingByTags = tags.tags.flatMap(findTagBranding(_, publicationDate, edition)).headOption
+    brandingByTags
   }
 }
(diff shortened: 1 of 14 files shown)


common/app/common/commercial/BrandHunter.scala
```diff
@@ -24,13 +24,11 @@ object BrandHunter {
...
-                          tags: Tags,
-                          publicationDate: Option[DateTime],
-                          edition: Edition): Option[Branding] = {
-    lazy val brandingBySection = section flatMap (findSectionBranding(_, publicationDate, edition))
-    lazy val brandingByTags = tags.tags.flatMap(findTagBranding(_, publicationDate, edition)).headOption
-    brandingBySection orElse brandingByTags
+  def findContentBranding(content: ContentType, edition: Edition): Option[Branding] = {
+    // TODO: incorporate branding by section - needs capi change to get section element in content results
+    val tags = content.tags
+    val publicationDate = Some(content.trail.webPublicationDate)
```
JonNorman: does this need to be an `Option`? It looks like we only call `findTagBranding` here.

kelvin-chappell: It's also used by tag pages, where there's no publication date: https://github.com/guardian/frontend/blob/master/common/app/model/Tag.scala#L292

common/app/common/commercial/ContainerModel.scala
```diff
@@ -23,7 +30,7 @@ case class ContainerContent(
 
 object ContainerModel {
 
-  def fromPressedCollection(collection: PressedCollection): ContainerModel = {
+  def fromPressedCollection(edition: Edition)(collection: PressedCollection): ContainerModel = {
```
JonNorman: Is there any reason to have two parameter lists here rather than one? (this is more me being curious than suggesting we merge them)

common/app/model/facia/PressedCollection.scala
```diff
@@ -29,7 +31,22 @@ case class PressedCollection(
...
 
-  lazy val curatedPlusBackfillDeduplicated = (curated ++ backfill).distinctBy(c => c.properties.maybeContentId.getOrElse(c.card.id))
+  lazy val curatedPlusBackfillDeduplicated = (curated ++ backfill).distinctBy { c =>
+    c.properties.maybeContentId.getOrElse(c.card.id)
+  }
+
+  def branding(edition: Edition): Option[Branding] = {
+
+    // TODO: replace this placeholder with value from fapi
+    val shouldBeBranded = true
```
JonNorman: Yes! The last piece of the puzzle!

common/app/common/commercial/ContainerModel.scala
kelvin-chappell: It's so that I can turn it into a function that can be passed into a map in frontBody.scala.html line 40.

JonNorman:
Looks good to me! Can't wait for when we can finally flip this switch on and dispense with all the DFP nonsense! :+1:

PR #12878, opened 2016-05-12
video embeds obey advertising rules of parent video

akash1810:
## What does this change?

Before:
Video is published with pre-roll=true
Video is embedded
Video pre-roll is then changed to false
Embed still has pre-roll=true

(and vice-versa for video published with pre-roll=false)

After:
I've added a field to /info.json which is used to determine if advertising should be blocked on the video. Therefore, similar to how expiry is calculated, the embed follows the parent video.

Video embeds obeys parent video. In order to achieve this before, the video had to be re-embedded.

No.
## Screenshots

n/a
## Request for comment

@jamesgorrie

### DIFF
--- a/static/src/javascripts/bootstraps/enhanced/media/main.js
+++ b/static/src/javascripts/bootstraps/enhanced/media/main.js
@@ -179,20 +179,20 @@ define([
         var mediaType = el.tagName.toLowerCase(),
             $el = bonzo(el).addClass('vjs vjs-tech-' + videojs.options.techOrder[0]),
             mediaId = $el.attr('data-media-id'),
-            blockVideoAds = $el.attr('data-block-video-ads') === 'true',
             showEndSlate = $el.attr('data-show-end-slate') === 'true',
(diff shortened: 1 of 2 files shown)


static/src/javascripts/bootstraps/enhanced/media/main.js
```diff
@@ -179,32 +179,36 @@ define([
...
-                    resolve(videoInfo.expired);
+                    resolve(videoInfo);
                 });
             }
         });
 
+        isPlayerExpired = videoInfo.expired;
+        blockVideoAds = videoInfo.shouldHideAdverts;
+        withPreroll = shouldPreroll && !blockVideoAds;
+
```
jamesgorrie: ^ These aren't available at this point - they are part of the promise, so you would need to refer to them in the resolution.

akash1810: updated, thanks

jamesgorrie:
👍

PR #12879, opened 2016-05-12
Release the Guardian Frame

regiskuckaertz:
Feedback from the creative team led to the creation of a new template in DFP to handle the recently announced Guardian Frame.

picture 1

I have left `gu-comcontent` (its original name) there for now because it is used in another context; I just made minor tweaks regarding the AD popup. I also refactored how we deal with the breakout iframe alg, to make sure the overarching promise resolved when the creative was rendered.

cc @uplne @Calanthe @jbreckmckye

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/dfp/breakout-iframe.js
+++ b/static/src/javascripts/projects/common/modules/commercial/dfp/breakout-iframe.js
@@ -34,48 +34,40 @@ define([
      * This can be set on the DFP creative.
      */
     function breakoutIFrame(iFrame, $slot) {
-        return new Promise(function (resolve, reject) {
-            /*eslint-disable no-eval*/
-            var $iFrame = bonzo(iFrame);
-            var iFrameBody = iFrame.contentDocument.body;
-            var $breakoutEl;
+        /*eslint-disable no-eval*/
+        var $iFrame = bonzo(iFrame);
+        var iFrameBody = iFrame.contentDocument.body;
+        var $breakoutEl;
 
-            if (!iFrameBody) {
-                reject();
-            }
+        $breakoutEl = $('.breakout__html, .breakout__script', iFrameBody);
 
-            $breakoutEl = $('.breakout__html, .breakout__script', iFrameBody);
-
-            if ($breakoutEl.hasClass('breakout__html')) {
-                resolve(fastdom.write(function () {
-                    $iFrame.hide();
-                    $breakoutEl.detach();
-                    $slot.append($breakoutEl[0].innerHTML);
-                }));
-            } else if ($breakoutEl.hasClass('breakout__script')) {
-                fastdom.write(function () {
-                    $iFrame.hide();
-                }).then(function () {
-                    var breakoutContent = $breakoutEl.html();
-                    if ($breakoutEl.attr('type') === 'application/json') {
-                        var creativeConfig = JSON.parse(breakoutContent);
-                        if (creativeConfig.name === 'fluid250-v4' || creativeConfig.name === 'fluid250-v3') {
-                            creativeConfig.name = 'fluid250';
-                        }
-                        require(['common/modules/commercial/creatives/' + creativeConfig.name], function (Creative) {
-                            new Creative($slot, creativeConfig.params, creativeConfig.opts).create();
-                        });
-                        resolve(creativeConfig.params.adType || '');
-                    } else {
-                        // evil, but we own the returning js snippet
-                        eval(breakoutContent);
-                        resolve();
+        if ($breakoutEl.hasClass('breakout__html')) {
+            return fastdom.write(function () {
+                $iFrame.hide();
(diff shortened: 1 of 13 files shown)


static/src/javascripts/projects/common/modules/commercial/dfp/breakout-iframe.js
```diff
@@ -34,48 +35,40 @@ define([
...
-            var $breakoutEl;
+        /*eslint-disable no-eval*/
+        var $iFrame = bonzo(iFrame);
+        var iFrameBody = iFrame.contentDocument.body;
+        var $breakoutEl;
 
-            if (!iFrameBody) {
-                reject();
-            }
+        $breakoutEl = $('.breakout__html, .breakout__script', iFrameBody);
```
jbreckmckye: Do we not still need the check to `iFrameBody`?

static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
```diff
@@ -407,7 +407,9 @@ define([
     }
 
     function shouldRenderLabel($adSlot) {
-        return $adSlot.data('label') !== false && qwery('.ad-slot__label', $adSlot[0]).length === 0;
+        return !$adSlot.hasClass('ad-slot--frame') &&
```
jbreckmckye: Where should we define whether an advert does or does not need a label added?

static/src/javascripts/projects/common/modules/commercial/dfp/breakout-iframe.js
```diff
@@ -34,48 +35,40 @@ define([
...
+                    var creativeConfig = JSON.parse(breakoutContent);
+                    if (creativeConfig.name === 'fluid250-v4' || creativeConfig.name === 'fluid250-v3') {
+                        creativeConfig.name = 'fluid250';
+                    }
+                    return new Promise(function(resolve) {
                         require(['common/modules/commercial/creatives/' + creativeConfig.name], function (Creative) {
-                            new Creative($slot, creativeConfig.params, creativeConfig.opts).create();
+                            resolve(new Creative($slot, creativeConfig.params, creativeConfig.opts).create());
                         });
-                        resolve(creativeConfig.params.adType || '');
```
jbreckmckye: I never much liked this as a return signature.

static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
regiskuckaertz: Perhaps via a config object somewhere? The thing is more complicated than it looks, some creative I saw already had a label, which presumably is why this function existed in the first place.

static/src/stylesheets/module/commercial/_frame.scss
```diff
@@ -0,0 +1,144 @@
+.frame {
```
jbreckmckye: Really clear style code, this.

static/src/javascripts/projects/common/modules/commercial/dfp/breakout-iframe.js
regiskuckaertz: There already is a test in `checkForBreakout` so it seemed redundant.

static/src/stylesheets/module/commercial/_frame.scss
regiskuckaertz: Thanks 😋

static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
jbreckmckye: I'd be tempted to add a flag to the creative object itself and return that from breakout-iframe, but I'd be a little worried about exposing those objects to other code, because they handle everything through mutable public fields. I guess you could have a static `needsLabel` field on the constructor itself, but it'd be a bit of an odd thing for the breakout script to return.

In any case, though, only the creative template itself knows whether it can handle its own labels, and so that's what should probably control whether labels are added further down the line.

regiskuckaertz: I can see a refactoring of our code where every creative is confined to its own module and deals with everything that is its own business ... indeed this has nothing to do in dfp-api.

static/src/javascripts/projects/common/modules/commercial/dfp/breakout-iframe.js
jbreckmckye: Fair enough.

static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
jbreckmckye: Well - we can worry about that after we get this thing shipped.

jbreckmckye:
👍

PR #12883, opened 2016-05-12
Make all CSP fields optional

desbo:
We don't have any control over these JSON submissions so it's safest not to assume anything will exist.
@TBonnin


TBonnin:
👍  🚀

PR #12887, opened 2016-05-12
check for video element before continuing

jamesgorrie:
## What does this change?

Does a check on the Bonzo element length.

No more console noise

Nope
## Screenshots

No
## Request for comment

@Calanthe @akash1810


jamesgorrie:
👍 from @akash1810

PR #12891, opened 2016-05-12
Make travel offer IDs keep their order

JonNorman:
## What does this change?

Currently, when travel offer IDs are provided within the DFP travel offer creative, the ordering is ignored and the merchandising container will render them (probably) alphabetically. This normally doesn't matter, but in the case of a _prominent_ travel offer component, it is necessary to be able to specify which offer should be prominent.

Ad Ops and others that use this component can specify the order by which travel offers are rendered in a merchandising container.

No
## Screenshots

**before -> after**
image

### DIFF
--- a/commercial/app/controllers/commercial/package.scala
+++ b/commercial/app/controllers/commercial/package.scala
@@ -20,7 +20,7 @@ package object commercial {
   }
 
   def specificId(implicit request: RequestHeader): Option[String] = request.queryString.get("t").map(_.head)
-  def specificIds(implicit request: RequestHeader): Seq[String] = request.queryString.getOrElse("t", Nil).reverse
+  def specificIds(implicit request: RequestHeader): Seq[String] = request.queryString.getOrElse("t", Nil)
 
   trait Relevance[T] {
     def view(ads: Seq[T])(implicit request: RequestHeader): Html
--- a/commercial/app/model/commercial/travel/TravelOffersAgent.scala
+++ b/commercial/app/model/commercial/travel/TravelOffersAgent.scala
@@ -15,14 +15,7 @@ object TravelOffersAgent extends MerchandiseAgent[TravelOffer] with ExecutionCon
   }
 
   def specificTravelOffers(offerIdStrings: Seq[String]): Seq[TravelOffer] = {
-
-    def sortByRequestOrder(a: TravelOffer, b: TravelOffer) = {
-
-      def getOriginalPositionOfId(offer: TravelOffer) = offerIdStrings.indexOf(offer.id)
-      getOriginalPositionOfId(a) < getOriginalPositionOfId(b)
-    }
-
-    available filter (offer => offerIdStrings contains offer.id) sortWith sortByRequestOrder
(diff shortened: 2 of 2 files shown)


commercial/app/controllers/commercial/package.scala
```diff
@@ -20,7 +20,7 @@ package object commercial {
   }
 
   def specificId(implicit request: RequestHeader): Option[String] = request.queryString.get("t").map(_.head)
-  def specificIds(implicit request: RequestHeader): Seq[String] = request.queryString.getOrElse("t", Nil).reverse
```
JonNorman: @kenlim do you know why you chose to reverse this list [originally](https://github.com/guardian/frontend/commit/0793d4ee905bd7dfc34855cd65c229b8aa99fa43#diff-80eb57197976cd6ca6bb91adc994aee3R19)?

commercial/app/model/commercial/travel/TravelOffersAgent.scala
```diff
@@ -15,7 +15,14 @@ object TravelOffersAgent extends MerchandiseAgent[TravelOffer] with ExecutionCon
...
   def specificTravelOffers(offerIdStrings: Seq[String]): Seq[TravelOffer] = {
-    available filter (offer => offerIdStrings contains offer.id)
+
+    def sortByRequestOrder(a: TravelOffer, b: TravelOffer) = {
+
+      def getOriginalPositionOfId(offer: TravelOffer) = offerIdStrings.indexOf(offer.id)
+      getOriginalPositionOfId(a) < getOriginalPositionOfId(b)
+    }
+
+    available filter (offer => offerIdStrings contains offer.id) sortWith sortByRequestOrder
```
regiskuckaertz: Wouldn't it be more elegant to just `offerIdStrings map (offerId => available filter (offer => offerId == offer.id))`?

kelvin-chappell: Or `sortBy(offerIdStrings.indexOf(_.id))`

commercial/app/controllers/commercial/package.scala
```diff
@@ -20,7 +20,7 @@ package object commercial {
...
+  def specificIds(implicit request: RequestHeader): Seq[String] = request.queryString.getOrElse("t", Nil)
```
kelvin-chappell: Is this guaranteed to get the values in the right order?  It probably does.

JonNorman: I believe so; the values associated with `t` are added to the request in the order they are specified in DFP. Playing around with various IDs seemed to confirm it.

commercial/app/model/commercial/travel/TravelOffersAgent.scala
JonNorman: Both good suggestions!

In just sorting/augmenting what was there I completely missed the opportunity to simplify, I think I'm going to go for:

`offerIdStrings flatMap (offerId => available find (_.id == offerId))`

It is succinct, keeps the order and doesn't require any sorting.

kelvin-chappell:
👍

PR #12892, opened 2016-05-12
Add low friction participation AB tests (initial segmentation)

gtrufitt:
## What does this change?

Adds A/B tests for Music (album reviews), Tv (episode by episode articles), all Recipes and all Fashion  **(FASHION FAB or FASHION FAIL)** articles. Initially testing the commenting so that we can put out at the right percentage.
## Request for comment

@crifmulholland @nicl

### DIFF
--- a/static/src/javascripts/projects/common/modules/experiments/tests/participation-low-fric-fashion.js
+++ b/static/src/javascripts/projects/common/modules/experiments/tests/participation-low-fric-fashion.js
@@ -1,0 +1,74 @@
+define([
+    'common/utils/config',
+    'common/utils/mediator',
+    'Promise'
+], function (
+    config,
+    mediator,
+    Promise
+) {
+
+    var omniture;
+
+    /**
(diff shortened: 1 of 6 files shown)


static/src/javascripts/projects/common/modules/experiments/tests/participation-low-fric-fashion.js
```diff
@@ -0,0 +1,74 @@
...
+    var omniture;
+
+    /**
+     * The omniture module depends on common/modules/experiments/ab, so trying to
+     * require omniture directly inside an AB test gives you a circular dependency.
+     *
+     * This is a workaround to load omniture without making it a dependency of
+     * this module, which is required by an AB test.
+     */
+    function getOmniture() {
```
gtrufitt: This could totally be a util but considering we're removing omniture I don't want to spread it.

crifmulholland:
👍

PR #12896, opened 2016-05-13
add timeout to contrib-ads on video to be larger than IMA timeout

jamesgorrie:
## What does this change?

This helps negate against the play => content-play => ad-play situation we are in at the moment.

We have two timeouts, one being `adtimeout` & `prerollTimeout` which are set by [videojs-contrib-ads](https://github.com/videojs/videojs-contrib-ads) and [videojs-ima](https://github.com/googleads/videojs-ima) respectively. 

At the moment we have `adsready?` => `adtimeout` => `contentplay` => `adsplay` as the adtimeout is set to the default (100ms) which is less than the ima `prerollTimeout`.

Now we will have `adready` => `adplay` => `contentplay`.

**Next**
Find out how to serve a blank frame as the first frame as we will currently serve the first frame before the ad, which is the same as the [IMA example](https://googleads.github.io/videojs-ima/examples/simple/).

People stop reporting the crap UX we give them.

No
## Request for comment

@akash1810

### DIFF
--- a/static/src/javascripts/bootstraps/enhanced/media/main.js
+++ b/static/src/javascripts/bootstraps/enhanced/media/main.js
@@ -293,8 +293,14 @@ define([
                                         player.ima({
                                             id: mediaId,
                                             adTagUrl: getAdUrl(),
-                                            prerollTimeout: 1000
+                                            prerollTimeout: 1000,
+                                            contribAdsSettings: {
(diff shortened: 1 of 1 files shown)


jamesgorrie:
As a note this mainly affects videos where we autoplay on landing on the page as we are still waiting for the IMA script to do it's business (which is quite a lot).

akash1810:
🎉 👍

PR #12899, opened 2016-05-13
Redisplay membership banners

paulbrown1982:
## What does this change?

This change triggers a re-display of the membership supporter banners (to people who have previously closed it). These appear on the footer of articles for any non-member who visits at least 10 different articles.

The Membership marketing team ask us to do this approximately twice per quarter.

There is usually an uplift of supporter signups when this is done.

No, only theguardian.com web users.
## Screenshots

picture 87


adamnfish:
Would it be clearer to use a date string (e.g. 2016-05) instead of a number? Means you'd be able to see at a glance when it was changed and it would better show if some regions have been changed at different times (at the moment it looks like UK is newer because its increment is higher).

paulbrown1982:
That's a good idea, I've submitted a change to include that.

jamespamplin:
:+1: code changes look fine to me

rtyley:
:+1:

PR #12900, opened 2016-05-13
Interactives don't have a tonal header, so they need some help

NataliaLKB:
After #12706  interactives with tones applied appeared to be without a headline. 

It turns out that interactives with tones already have some overrides to help them out. They have them because even though they have the tones, they don't have the tonal header. 

I have deleted the unnecessary overrides, and included a couple that are needed for headlines to appear correctly.
## Screenshots

Before:

image

image

After:

image

image

cc @sndrs

### DIFF
--- a/static/src/stylesheets/module/content/tones/_tone-special-report.scss
+++ b/static/src/stylesheets/module/content/tones/_tone-special-report.scss
@@ -18,17 +18,15 @@
 .tonal--tone-special-report {
     // Interactives need help because their headers are not tonal
     &.content--interactive {
-        .commentcount i {
-            @extend %i-comment--tone-neutral2;
-
-            @if $svg-support {
-                .svg & {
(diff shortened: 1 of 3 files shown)


static/src/stylesheets/module/content/tones/_tone-special-report.scss
```diff
@@ -18,17 +18,15 @@
...
-            @if $svg-support {
-                .svg & {
-                    @extend %svg-i-comment--tone-neutral2;
-                }
-            }
+        .content__series-label__link {
+            color: $neutral-2;
         }
 
-        .tone-colour,
```
NataliaLKB: Removing `.tone-colour` made sure that the comments showed up as grey rather than yellow. I find this more readable, but I can switch it back to yellow if needed.

sndrs:
👍 great  - thank you :)

PR #12902, opened 2016-05-13
Update prebid.js with custom build

regiskuckaertz:
This change brings v0.8.1 of prebid.js, built with the only adapter we're using right now, i.e. Rubicon

cc @jbreckmckye


jbreckmckye:
👍

PR #12903, opened 2016-05-13
Add contentType to standard js configuration

kelvin-chappell:
This should expose config.page.contentType and add it to hits request.


JonNorman:
:+1:

PR #12904, opened 2016-05-13
Fabric: replace top slot adverts with an inline fabric ad

jbreckmckye:
Fabric takeovers include full-width adverts in the navigation top slot, but mobile breakpoints don't include such a position. In these cases, the advert should appear in the first inline position instead. This pull request makes this work on articles.

Fabric top slots are requested under the size 88x71, so all we need to do is request that size in addition to ordinary MPUs in the first article body advert. If the page has a takeover, its sponsorship priority will force the slot to show the fabric creative. If not, we fall back to a standard 300x250 MPU.

Existing styles then make this replacement ad display in a flush layout:

image


Calanthe:
nice 👍

JonNorman:
Nice :+1:

PR #12905, opened 2016-05-13
move onward journey for videos to own component

jamesgorrie:
## What does this change?

While trying to figure out what the devil is slowing things down with ads, I keep coming up against not really knowing what's going on in this file.

I am going to try break it into a more coherent structure and flow.

We will need to break this up at some stage as having a new player would not remove the need to have all the furniture, so this isn't wasted effort (I hope 🙏).

I can actually do some work on the player.

Nope
## Screenshots

See site, they're the same
## Request for comment

@akash1810

### DIFF
--- a/static/src/javascripts/bootstraps/enhanced/media/main.js
+++ b/static/src/javascripts/bootstraps/enhanced/media/main.js
@@ -22,6 +22,7 @@ define([
     'common/modules/video/tech-order',
     'common/modules/video/video-container',
     'common/modules/video/onward-container',
+    'common/modules/video/more-in-series-container',
     // This must be the full path because we use curl config to change it based
     // on env
     'bootstraps/enhanced/media/video-player',
@@ -50,6 +51,7 @@ define([
     techOrder,
     videoContainer,
     onwardContainer,
+    moreInSeriesContainer,
     videojs,
     loadingTmpl
 ) {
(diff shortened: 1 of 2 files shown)


static/src/javascripts/projects/common/modules/video/more-in-series-container.js
```diff
@@ -0,0 +1,26 @@
...
+    'common/modules/component',
+    'common/utils/mediator'
+], function(
+    Component,
+    mediator
+) {
+
+    function init(el, mediaType, section, shortUrl, series) {
+        var component = new Component();
+        var endpoint  = '/' + mediaType + '/section/' + section +
```
akash1810: :'( string interpolation

akash1810:
👍

static/src/javascripts/bootstraps/enhanced/media/main.js
```diff
@@ -362,43 +364,31 @@ define([
...
-        }
-
-        section.endpoint = endpoint;
-
-        section.fetch(attachTo).then(function () {
-            mediator.emit('page:media:moreinloaded', attachTo);
-            mediator.emit('page:new-content', attachTo);
-        });
+        var el  = $('.js-onward')[0];
+        moreInSeriesContainer.init(
```
johnduffell: are you sure it's more in series? I had it as more in section (that's specifically *not* in the series)

johnduffell: the more in series one goes in the normal onward trails.js - https://github.com/guardian/frontend/blob/master/static/src/javascripts/bootstraps/enhanced/trail.js#L94

PR #12906, opened 2016-05-13
Add contentType to standard js configuration

kelvin-chappell:
Puts content-type value from metadata into js config by default, to avoid having to always add it with js config overrides.

### DIFF
--- a/commercial/app/controllers/commercial/CreativeTestPage.scala
+++ b/commercial/app/controllers/commercial/CreativeTestPage.scala
@@ -1,9 +1,9 @@
 package controllers.commercial
 
-import model.{MetaData, GuardianContentTypes, NoCache, Cached}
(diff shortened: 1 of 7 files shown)


rich-nguyen:
Can we remove all the overrides?

kelvin-chappell:
@rich-nguyen Yes, I'll do that.

kelvin-chappell:
@rich-nguyen Overrides all gone now.

rich-nguyen:
Great. I guess the only reason it wasn't there was that not all things Content-related have a Metadata representation. For example, all pages have Metadata, but not all those pages have CAPI content associated (discussion/identity pages, that kind of thing). The field can still make sense here though, so I think it's ok, especially because its not an Option. 👍

PR #12907, opened 2016-05-13
Allow the Member's Area redirect to apply on all parts of the site

rtyley:
## What does this change?

Previously, only articles that had `/membership/` in the url - like this one:

http://www.theguardian.com/membership/2015/apr/27/the-spin-london-cycle-festival-save-mop

...would actually have the [Member's Area restriction](https://docs.google.com/a/guardian.co.uk/document/d/1oQhe3PwhcJdmmm_q6_9issBqOA69vE0hMcT28WztQsA/edit?usp=sharing) take effect - the setting wouldn't affect articles under theguardian.com/film, for instance. This change gets rid of that `/membership/` restriction.

Currently, these are the only three articles that would be affected by this change:
- https://www.theguardian.com/news/2016/feb/12/guardian-bookshop-an-exclusive-offer
- https://www.theguardian.com/news/2015/oct/07/exclusive-to-guardian-members-buy-half-price-tickets-american-idiots-mop
- https://www.theguardian.com/news/2016/apr/29/looking-back-archaeology - marked as _paid-members-only_

We're planning to use the Member's Area on a wider basis around the site, and also apply it to the Mobile Apps premium content, so this '/membership/' restriction (initially in place to reduce possible editorial accidents) needs to go.

Eventually the Mobile Apps will apply the Member's Area restriction as well, but currently the dotcom web site is the only place where this takes effect.

cc @DiegoVazquezNanini @rcrphillips @mchv @paulbrown1982

### DIFF
--- a/common/app/model/meta.scala
+++ b/common/app/model/meta.scala
@@ -284,7 +284,7 @@ final case class MetaData (
   val isSurging: Seq[Int] = SurgingContentAgent.getSurgingLevelsFor(id)
 
   val requiresMembershipAccess: Boolean = {
-    conf.switches.Switches.MembersAreaSwitch.isSwitchedOn && membershipAccess.nonEmpty && url.contains("/membership/")
+    conf.switches.Switches.MembersAreaSwitch.isSwitchedOn && membershipAccess.nonEmpty
   }
 
   val hasSlimHeader: Boolean = contentType == "Interactive" || section == "identity" || (galleryRedesign.isSwitchedOn && contentType.toLowerCase == "gallery")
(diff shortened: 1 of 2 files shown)


paulbrown1982:
👍

mchv:
@rtyley my understanding that we will as well add the restriction for the `looking back` tag?

rtyley:
> my understanding that we will as well add the restriction for the looking back tag?

After discussion with Subhajit Banerjee, @rcrphillips and @DiegoVazquezNanini, we decided _not_ to do this.

Instead content under 'Looking back' will *also* have `paid-members-only` set on it in Composer. This means that Premium content will be 'blocked' on the web by the existing 'Members Area' functionality.

rtyley:
Confirmed in PROD that the redirect takes place for these test articles:

* https://www.theguardian.com/news/2016/feb/12/guardian-bookshop-an-exclusive-offer
* https://www.theguardian.com/news/2015/oct/07/exclusive-to-guardian-members-buy-half-price-tickets-american-idiots-mop
* https://www.theguardian.com/news/2016/apr/29/looking-back-archaeology - marked as *paid-members-only*

PR #12908, opened 2016-05-13
get frontend.properties from s3

johnduffell:
For all those who had to copy and paste a file from google drive and hated it, now setup.sh will do it for you magically.

Further benefits include access control being managed through janus rather than by me manually.

@adamnfish 🎉

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -30,6 +30,16 @@ create_install_vars() {
   fi
 }
 
+install_awscli() {
+  if ! installed aws; then
+    if linux; then
+      sudo apt-get install -y awscli
+    elif mac; then
+      brew install awscli
+    fi
+  fi
+}
+
 create_frontend_properties() {
   local path="$HOME/.gu"
   local filename="frontend.properties"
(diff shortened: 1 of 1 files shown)


setup.sh
```diff
@@ -58,22 +57,6 @@ region = eu-west-1" > "$path/$filename"
   fi
 }
 
-create_aws_credentials() {
```
johnduffell: this was pretty out of date anyway in these days of janus

setup.sh
```diff
@@ -39,8 +39,7 @@ create_frontend_properties() {
       mkdir "$path"
     fi
 
-    touch "$path/$filename"
-    EXTRA_STEPS+=("Ask a colleague for frontend.properties and add the contents to $path/$filename")
+    aws s3 cp --profile frontend s3://aws-frontend-store/template-frontend.properties "$path/$filename"
```
jamespamplin: Is this bucket versioned? If not, can we make it please :smile:

jamespamplin: Also can we please have an "installed" check for the `aws` cli

johnduffell: already is versioned by the looks

johnduffell: ok done that installing the aws cli

jamespamplin:
:+1: great idea, thanks for this

PR #12910, opened 2016-05-13
Fix for dev js booting

rich-nguyen:
There are some intermittent issues with the dev js loading. This was caused by using the prefetch system in dev, which was supposed to align prod and dev behaviour. The best fix is to remove the prefetch system, because it was pre-emptively downloading some bootstraps in the order of a few milliseconds (9ms when I measured). To balance out any performance risk, the "optimised" `enhanced-vendor` js has been removed, which was a separate 2KB download.

I kept the user timing marking to the standard app, it's still useful.

@jamesgorrie


static/src/javascripts/boot.js
```diff
@@ -12,20 +12,19 @@ Now we can safely require one of those modules.
...
 
-Bundles we need to run: commercial + (enhanced-vendor + enhanced)*
-* Only if we detect we should run enhance
+Bundles we need to run: commercial + enhanced
+
+Only if we detect we should run enhance.
  */
 
 define([
     'Promise',
```
sndrs: do these methods all really need to be wrapped in promises? there's not much going on now, feel a bit like overkill?

sndrs:
👍 other than comment - nice one

static/src/javascripts/boot.js
rich-nguyen: Yeah, the thought did cross my mind. If you agree it's worth doing, I'll open another pr for that?

sndrs: okedoke

static/src/javascripts/boot.js
```diff
@@ -39,14 +38,7 @@ define([
...
-        return domReadyPromise.then(function () {
-            return promiseRequire(['bootstraps/standard/main'])
-                .then(function (boot) { boot(); });
-        });
-    };
-
-    var bootEnhanced = function () {
-        return promiseRequire(['bootstraps/enhanced/main'])
+        return promiseRequire(['bootstraps/standard/main'])
             .then(function (boot) { boot(); });
```
regiskuckaertz: `.then(boot)`

rich-nguyen: Yeah, still a few of those around in this file. We may un-promise this library soon

PR #12913, opened 2016-05-16
Revert "add timeout to contrib-ads on video to be larger than IMA timeout"

jamesgorrie:
Reverts guardian/frontend#12896

It seems tracking might have taken a hit, although I can't see as to why.


akash1810:
👍

PR #12916, opened 2016-05-16
Remove eTags switch

johnduffell:
Etags are working really well, and noone has complained of caching issues since we shipped them.  As such it seems sensible to remove the switch rather than making it unlimited, as there's not a high chance we'll need to turn it off.  Leaving switches unnecessarily would clutter up the switchboard.


JonNorman:
"Yay" for fewer switches! :+1:

johnduffell:
thanks! we should all have a "delete a switch" day to clean up sometime 🎉 although the merge conflicts would be frightening!

PR #12917, opened 2016-05-16
Fix IE11 display bugs in the soulmates component

regiskuckaertz:
Fixes two display bugs where IE11 doesn't handle (1) the intrinsic height of wrapper elements and (2) the text wrapping in flex containers.

unnamed

cc @JonNorman


JonNorman:
:+1:

PR #12918, opened 2016-05-16
Update run-checks and config for us minute and us briefing

gtrufitt:
## What does this change?

Show the campaign minute sign-up on all election articles and the us briefing on the rest of the us news articles.

### DIFF
--- a/static/src/javascripts/projects/common/modules/email/email-article.js
+++ b/static/src/javascripts/projects/common/modules/email/email-article.js
@@ -46,11 +46,15 @@ define([
                 description: 'Sign up and we\'ll send you the Guardian US Campaign Minute, once per day.',
                 successHeadline: 'Thank you for signing up to the Guardian US Campaign minute',
                 successDescription: 'We will send you the biggest political story lines of the day',
-                modClass: 'post-article',
+                modClass: config.page.isMinuteArticle ? 'post-article' : 'end-article',
                 insertMethod: function () {
-                    return function ($iframeEl) {
-                        $iframeEl.insertAfter('.js-article__container');
-                    };
+                    if (config.page.isMinuteArticle) {
+                        return function ($iframeEl) {
+                            $iframeEl.insertAfter('.js-article__container');
+                        };
+                    } else {
+                        return insertBottomOfArticle();
+                    }
                 }
             },
             theFilmToday: {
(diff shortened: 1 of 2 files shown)


gtrufitt:
Any chance of a look-see at this from one of you? @crifmulholland @desbo @janua 

Wendy has asked for these rules for showing the email sign-up:

> Guardian US briefing tag - always show briefing sign-up
> US News section - show briefing sign-up unless it has the US Elections 2016 tag 
> in which case...
> US Elections 2016 tag - show campaign minute sign-up

There's no 'Guardian US briefing tag' but it is a series, hence using the series instead of keywords check from config.

And updating the us briefing sign-up text.

static/src/javascripts/projects/common/modules/email/email-article.js
```diff
@@ -46,11 +46,15 @@ define([
...
                 insertMethod: function () {
-                    return function ($iframeEl) {
-                        $iframeEl.insertAfter('.js-article__container');
-                    };
+                    if (config.page.isMinuteArticle) {
+                        return function ($iframeEl) {
+                            $iframeEl.insertAfter('.js-article__container');
+                        };
+                    } else {
+                        return insertBottomOfArticle();
```
desbo: did you mean to call this or just return the function itself?

gtrufitt: Yeah, gotta call it here as I am returning a function from `insertBottomOfArticle` (which is the one executed on insert).

desbo: ah, shouldn't the other `insertMethod`s here call it as well then?

gtrufitt: Nope because the call is `insertMethod()($iframeEl);` so on the others they call `insertBottomOfArticle()` as the first execution which will immediately return the function whereas in this one, because there is a function assigned to the property the first execution calls that function containing the if statement which returns either that function in the `if` or calls `insertBottomOfArticle()` which returns the same function as the other calls. There's that extra level so we can do the if check.

I mean, it's a bit of a head****. :/

static/src/javascripts/projects/common/modules/email/run-checks.js
```diff
@@ -97,7 +98,8 @@ define([
             return page.keywordExists(['Football']) && allowedArticleStructure();
         },
         usBriefing: function () {
-            return config.page.section === 'us-news' && allowedArticleStructure();
+            return (config.page.section === 'us-news' || config.page.series === 'Guardian US briefing') &&
```
janua: Where does `config.page.series` come from? This `'Guardian US briefing'` seems a bit arbitrary, but maybe it's not such a big deal if `config.page.series` is coming from something static, like a tag.

gtrufitt: Yeah, it's from the series tag in composer: 
image

static/src/javascripts/projects/common/modules/email/email-article.js
desbo: I don't think you need to return a function from `insertBottomOfArticle` – it can just take $iframeEl as a parameter and you can get rid of the first invocation on line 150, and not call it when you return on line 56

gtrufitt: Yeah, good point, there used to be a reason for the extra step but it's not used anymore... I'll update.

gtrufitt: Cheers @desbo , I've removed that insanity now.

desbo:
👍  🐸

PR #12919, opened 2016-05-16
Increase all silent participation ab tests to 100%

gtrufitt:
## What does this change?

Because we are trying to check what % our A/B test should go out to, we need to check against a high percentage (everyone) and then work out what the lowest number of users will be for statistical significance (as 10% was not giving us enough traffic to work that out).


nicl:
:+1: :foggy: 

@gtrufitt this is just to give a sense of numbers right? We're not at the stage of actually showing stuff to users?

gtrufitt:
@nicl Correct, cheers!

PR #12860, opened 2016-05-11
Hosted video

Calanthe:
## What does this change?

This PR is the continuation of the https://github.com/guardian/frontend/pull/12832. The hosted page will include at least one video with a specific behaviour (fade in and out of control buttons) and styling (bigger control buttons, different icon's colour based on the brand). I created a new, much simplified video module, since we don't need now any additional features like playlists or video pre-rolls. 
## What is the value of this and can you measure success?

We will see if people will watch the video.
## Does this affect other platforms - Amp, Apps, etc?

At some point it will be.
## Screenshots

hosted_video
## Request for comment

@uplne

PR #12920, opened 2016-05-16
Reduce the size of the commercial js asset

rich-nguyen:
The changes in https://github.com/guardian/frontend/pull/12860 mean that we build the javascript bootstrap for commercial, and include all the modules used by the `media/video-player` module. This meant commercial had ballooned to 455KB, and was acting in a similar way to the old `enhanced-vendor` optimisation, holding module definitions for lots of downstream bootstraps.

This modifies the behaviour of hosted-video so that this no longer happens.

Aside:
I've investigated the impact this issue had on my previous change, which removed `enhanced-vendor`, claiming it was only 2KB [here](https://github.com/guardian/frontend/pull/12815). Despite the misunderstanding, I still think this is a valid thing to do. The `enhanced-vendor` file hides all our videojs problems. If I use this PR as a baseline, and reintroduce `enhanced-vendor`, here are the asset build sizes:
1. `enhanced-vendor`, with all third-party js: 266KB. This leaves `media` bootstrap at 10KB.
2. `enhanced-vendor`, with most third-party js, excluding video: 2KB. This leaves `media` bootstrap at 273KB.

So it is still wise to consider `enhanced-vendor` a 2KB optimisation file. This means that we defer loading any video content until we actually choose to execute the media bootstrap.

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/hosted-video.js
+++ b/static/src/javascripts/projects/common/modules/commercial/hosted-video.js
@@ -8,7 +8,6 @@ define([
     'common/utils/defer-to-analytics',
     'common/modules/video/events',
     'common/modules/video/supportedBrowsers',
-    'bootstraps/enhanced/media/video-player',
     'text!common/views/ui/loading.html'
 ], function (
     bean,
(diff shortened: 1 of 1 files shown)


Calanthe:
Good work 👍  Thanks for looking into that Rich!

PR #12922, opened 2016-05-16
Test Facebook Native Ads

regiskuckaertz:
Will insert a Facebook HTML ad on mobile devices before the first container when the adtest cookie is set to `fbnative`.

Works locally, now needs to go online for stakeholders to see.

picture 1

_It's ugly but it's not the point_

@jbreckmckye @uplne @JonNorman @rich-nguyen @Calanthe


static/src/stylesheets/module/commercial/_facebook.scss
```diff
@@ -0,0 +1,32 @@
+.fb-ad-slot {
```
jbreckmckye: I guess all these styles are temporary?

static/src/javascripts/projects/common/modules/commercial/third-party-tags/facebook.js
```diff
@@ -0,0 +1,87 @@
...
+            referenceNode: '#top > .l-side-margins > :first-child',
+            where: 'beforebegin',
+            placementId: '978824118882656_978826235549111',
+            adId: 'ad_root'
+        }
+    ];
+    var facebookTpl;
+
+    return {
+        load: init
```
jbreckmckye: I _think_ I can see why we've named this method this way - `load` has more meaning for consumers of the module; `init` has more meaning inside the module - but couldn't we expose just a singular `loadFacebook` function?

static/src/javascripts/projects/common/modules/commercial/third-party-tags/facebook.js
```diff
@@ -0,0 +1,87 @@
...
+    'common/utils/cookies',
+    'common/utils/template',
+    'text!common/views/commercial/facebook.html'
+], function(fastdom, detect, cookies, template, facebookStr) {
+    var scriptId = 'facebook-jssdk';
+    var scriptSrc = '//connect.facebook.net/en_US/sdk/xfbml.ad.js#xfbml=1&version=v2.5';
+    var appNetworkId = '978824118882656';
+    var adUnits = [
+        {
+            referenceNode: '#top > .l-side-margins > :first-child',
```
jbreckmckye: Looking at this I don't think I get any sense that this is the first container

static/src/javascripts/projects/common/modules/commercial/third-party-tags/facebook.js
```diff
@@ -0,0 +1,87 @@
...
+            where: 'beforebegin',
+            placementId: '978824118882656_978826235549111',
+            adId: 'ad_root'
+        }
+    ];
+    var facebookTpl;
+
+    return {
```
jbreckmckye: I'd probably just shove this to the bottom.

static/src/javascripts/projects/common/modules/commercial/third-party-tags/facebook.js
```diff
@@ -0,0 +1,87 @@
...
+
+        // we need the SCRIPT element to have an ID so we won't be using curl.js
+        // this time, sorry guys
+        insertFbScript();
+    }
+
+    function insertAdUnits() {
+        facebookTpl = template(facebookStr);
+        adUnits.forEach(function (adUnit) {
+            adUnit.referenceNode = document.querySelector(adUnit.referenceNode);
```
jbreckmckye: I'm not sure if I'm keen on this kind of object mutation...

static/src/stylesheets/module/commercial/_facebook.scss
regiskuckaertz: Yep

static/src/javascripts/projects/common/modules/commercial/third-party-tags/facebook.js
regiskuckaertz: This is temporary too, I didn't want to think about it too much 😅

static/src/javascripts/projects/common/modules/commercial/third-party-tags/facebook.js
```diff
@@ -0,0 +1,87 @@
...
+                }
+            );
+        };
+
+        insertAdUnits();
+
+        // we need the SCRIPT element to have an ID so we won't be using curl.js
+        // this time, sorry guys
+        insertFbScript();
+    }
```
jbreckmckye: Maybe `init()` should take responsibility for freeing up memory.

static/src/javascripts/projects/common/views/commercial/facebook.html
```diff
@@ -0,0 +1,9 @@
+<div class="fb-ad" data-placementid="<%=placementId%>" data-format="native" data-nativeadid="<%=adId%>" data-testmode="false"></div>
+<div class="fb-ad-slot" id="<%=adId%>">
+    <a class="fbAdLink">
```
jbreckmckye: What does this class do?

static/src/javascripts/projects/common/modules/commercial/third-party-tags/facebook.js
jbreckmckye: Fair enough!

static/src/javascripts/projects/common/modules/commercial/third-party-tags/facebook.js
regiskuckaertz: This is to mimic the signature of other 3rd parties (in `third-party-tags`), although I'm not sure yet this module will remain there for long... we need to address questions such as which is the fallback of which: DFP or Facebook?

static/src/javascripts/projects/common/modules/commercial/third-party-tags/facebook.js
regiskuckaertz: Oh, I always think `forEach` implies some kind of mutation.

static/src/javascripts/projects/common/views/commercial/facebook.html
regiskuckaertz: It is used by the FB lib, I can't change these unfortunately.

static/src/javascripts/projects/common/modules/commercial/third-party-tags/facebook.js
regiskuckaertz: Bah, this is a small thing I can improve later, I don't want to chain promises for now. I'll probably remove it at some point.

static/src/javascripts/projects/common/modules/commercial/third-party-tags/facebook.js
jbreckmckye: Yeah, could be some interesting questions to resolve. I think this is fine as is.

jbreckmckye:
Progress marches on!

👍

PR #12923, opened 2016-05-16
Depromisify boot

sndrs:
## What does this change?

remove module dips from boot js, and makes it a bit easier to read
## Request for comment

@rich-nguyen


static/src/javascripts/boot.js
```diff
@@ -2,75 +2,38 @@
...
-                            }
-                        ));
+            // 'raven' is included in standard already, but we need a reference to it
+            require(['raven'].concat(commercialBootstrap, enhancedBootstrap), function (raven, commercialBootrstrap, bootEnhanced) {
+                if (bootEnhanced) {
+                    bootEnhanced();
+                }
+                if (config.switches.commercial) {
+                    raven.wrap({tags: { feature: 'commercial' }}, function () {
+                        commercialBootrstrap.init();
```
rich-nguyen: is this typo deliberate?  could we call it `commercialBootstrap`, and remove the `var commercialBootstrap`?

static/src/javascripts/boot.js
```diff
@@ -2,75 +2,38 @@
...
-    // https://github.com/cujojs/curl/issues/293
-    var promiseRequire = function (moduleIds) {
-        return Promise.resolve(require(moduleIds));
-    };
-
     var guardian = window.guardian;
     var config = guardian.config;
 
-    var domReadyPromise = new Promise(function (resolve) { domReady(resolve); });
+    var standardBootstrap = ['bootstraps/standard/main'];
```
rich-nguyen: Can we inline this array? It's only used once

rich-nguyen:
👍  apart from the two comments about inline arrays

static/src/javascripts/boot.js
```diff
@@ -2,75 +2,36 @@
...
-                            function (commercial) {
-                                commercial.init();
-                            }
-                        ));
+            // 'raven' is included in standard already, but we need a reference to it
+            require(['raven', 'bootstraps/commercial'].concat(enhancedBootstrap), function (raven, commercialBootstrap, bootEnhanced) {
+                if (bootEnhanced) {
+                    bootEnhanced();
+                }
+                if (config.switches.commercial) {
```
regiskuckaertz: Shouldn't this be treated the same way as `enhancedBoostrap`?

static/src/javascripts/boot.js
```diff
@@ -2,75 +2,36 @@
...
+                    raven.wrap({tags: { feature: 'commercial' }}, function () {
```
regiskuckaertz: or `raven.wrap({ ... }, commercial.init)`

(I know some browsers optimise tail calls now, but hey why not?)

static/src/javascripts/boot.js
sndrs: enhanced is built minus the commercial modules, so it need commercial to run, whereas commercial does not need enhanced. it's a bit convoluted, but it makes enhanced as light as possible, given we almost never want to _not_ run commercial. 

that switch is just there in case i guess – but disabling it won't stop enhanced running.

PR #12926, opened 2016-05-16
"Hosted By" Video Thrasher

lps88:
## What does this change?

New Thrasher for Hosted By campaigns, which looks like video content (may or may not play actual videos in it in future)
## Screenshots

?adtest=zosia01
image
## Request for comment


static/src/javascripts/projects/common/modules/commercial/creatives/hosted-thrasher.js
```diff
@@ -0,0 +1,41 @@
...
+        }
+
+        var domPromise = fastdom.write(function () {
+            this.$adSlot.append(hostedVideoTpl({ data: this.params}));
+        }.bind(this));
+
+        domPromise.then(function () {
+            hostedVideo.init();
+        });
+
```
regiskuckaertz: You can simplify a little here:

```
fastdom.write(function () { ... }, this).then(hostedVideo.init);
```

(Fastdom accepts a second parameter)

static/src/javascripts/projects/common/modules/commercial/creatives/hosted-thrasher.js
```diff
@@ -0,0 +1,41 @@
...
+        if (this.params.trackingPixel) {
```
regiskuckaertz: You can move that into the call to `fastdom.write`.

static/src/javascripts/projects/common/views/commercial/creatives/hosted-thrasher.html
```diff
@@ -0,0 +1,38 @@
+<div class="creative--hosted-video l-side-margins fc-container__inner" style="
+        background-color: <%=data.backgroundColour%>;
+    ">
+    <div class="fc-container__header">
+        <a href="/guardian-labs" class="fc-container__header__title">
```
Calanthe: We need to add `data-link` properties to all of the links.

static/src/javascripts/projects/common/modules/commercial/creatives/hosted-thrasher.js
```diff
@@ -0,0 +1,41 @@
...
+    'common/modules/commercial/creatives/add-tracking-pixel',
+    'text!common/views/commercial/creatives/hosted-thrasher.html'
+], function (
+    fastdom,
+    template,
+    hostedVideo,
+    addTrackingPixel,
+    hostedVideoStr
+) {
+    var hostedVideoTpl;
```
jbreckmckye: Eh, I'm not a huge fan of these kinds of abbreviations.

static/src/javascripts/projects/common/modules/commercial/creatives/hosted-thrasher.js
```diff
@@ -0,0 +1,41 @@
...
+
+    var HostedThrasher = function ($adSlot, params) {
```
jbreckmckye: Why declare this as a variable?

static/src/javascripts/projects/common/modules/commercial/creatives/hosted-thrasher.js
```diff
@@ -0,0 +1,41 @@
+define([
```
jbreckmckye: Doesn't this need to be bundled into breakout-iframe before it'll work outside of dev?

static/src/javascripts/projects/common/views/commercial/creatives/hosted-thrasher.html
```diff
@@ -0,0 +1,38 @@
...
+                        "><%=data.header1%></p>
+                    <h1 class="hosted__heading" style="
+                            color: <%=data.header2colour%>;
+                        "><%=data.header2%></h1>
+                </div>
+            </a>
+            <div class="faux-controls">
+                <div class="play-button"></div>
+                <div class="time">
+                    0:00 / <span class="length">1:26</span>
```
jamesgorrie: Are we sure we want to hardcode this?

static/src/javascripts/projects/common/modules/commercial/creatives/hosted-thrasher.js
```diff
@@ -0,0 +1,41 @@
...
+) {
+    var hostedVideoTpl;
+
+    var HostedThrasher = function ($adSlot, params) {
+        this.$adSlot = $adSlot;
+        this.params = params;
+    };
+
+    HostedThrasher.prototype.create = function () {
+        if (!hostedVideoTpl) {
```
jbreckmckye: You could simplify to a single line -

```
hostedVideoHtml = hostedVideoHtml || template(hostedVideoTemplate);
```

static/src/javascripts/projects/common/modules/commercial/creatives/hosted-thrasher.js
```diff
@@ -0,0 +1,41 @@
...
+            hostedVideoTpl = template(hostedVideoStr);
+        }
+
+        var domPromise = fastdom.write(function () {
+            this.$adSlot.append(hostedVideoTpl({ data: this.params}));
+        }.bind(this));
```
jbreckmckye: I never know how I feel about this public-all-the-things strategy for these components. I mean, I know it's the idiom for _this specific part of the codebase_, and I guess they never get exposed to any foreign code, but... would it be nicer if we just passed these values straight into static closure functions? We could keep all this stuff truly private (so no-one working on this code for the first time need worry about access control), and we wouldn't need to mess about binding `this` everywhere.

I'm just thinking aloud here.

static/src/javascripts/projects/common/modules/commercial/hosted-video.js
```diff
@@ -19,8 +19,7 @@ define([
     videojs,
     loadingTmpl
 ) {
-
-    var $videoEl = $('.vjs-hosted__video');
+    var $videoEl;
```
jbreckmckye: Why not local scope?

static/src/javascripts/projects/common/modules/commercial/hosted-video.js
```diff
@@ -45,7 +44,9 @@ define([
     }
 
     function init() {
-        if (!$videoEl) {
+        $videoEl = $('.vjs-hosted__video');
+
+        if (!$videoEl || !$videoEl.length) {
```
jbreckmckye: I'm pretty sure qwery always returns an array, so you could just check its length.

static/src/stylesheets/module/commercial/glabs/_hosted.scss
```diff
@@ -155,7 +155,7 @@ $hosted-stripe: 28px;
         padding-top: 21px;
     }
 
-    &.vjs .vjs-play-control {
+    &.vjs .vjs-button.vjs-play-control {
```
jbreckmckye: What's causing the specificity bump?

static/src/stylesheets/module/commercial/glabs/_hosted.scss
```diff
@@ -312,3 +307,93 @@ $hosted-stripe: 28px;
...
+            display: inline-block;
+            vertical-align: top;
+            line-height: 36px;
+            margin-left: $gs-gutter/2;
+            @include mq(desktop) {
+                line-height: 49px;
+            }
+        }
+        .length {
+            color: #bdbdbd;
```
jbreckmckye: Is there an existing var for this?

static/src/javascripts/projects/common/modules/commercial/hosted-video.js
Calanthe: The `hosted-video.js` is loaded with the whole commercial module. But because the thrasher is loaded from the DFP,  during the page load there is no `.vjs-hosted__video` yet in the DOM. So I needed to move it to the `init()` method so it will be run in Promise, after being attached to the body.

static/src/stylesheets/module/commercial/glabs/_hosted.scss
Calanthe: I was overwriting the styles of the video buttons (because at some point there will be video in the thrasher). Originally I was doing that only on the hosted page (http://m.code.dev-theguardian.com/commercial/guardian-hosted) but it seems that on fronts (where thrasher will be located) there is a different order of css files.

static/src/stylesheets/module/commercial/glabs/_hosted.scss
```diff
@@ -312,3 +307,93 @@ $hosted-stripe: 28px;
...
+        position: absolute;
+        background-image: linear-gradient(to bottom, rgba(238, 238, 238, 0), rgba(0, 0, 0, .4));
+        border-top: 2px solid transparent;
+    }
+
+    .faux-controls  {
+        position: absolute;
+        bottom: 0;
+        left: $gs-gutter/2;
+        .time {
```
Calanthe: Remove those `.time` and `.length` from html and css.

static/src/javascripts/projects/common/modules/commercial/creatives/hosted-thrasher.js
jamesgorrie: As it's having it's prototype extended?

static/src/javascripts/projects/common/modules/commercial/creatives/hosted-thrasher.js
jamesgorrie: It'd be great to figure out a good idiom for a component in general for the codebase - [we did a while ago](https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/common/modules/component.js) <= but that is a terrible execution (I can say so as I did a lot of it).

[Oliver tried another way which is nicer](https://github.com/guardian/frontend/blob/2e00099b6509024fd5a9f04aa7daea03e08281ac/static/src/javascripts/projects/common/utils/create-store.js), but I'm not sure it has buy in from others.

Shall we have a chat about this - or does it potentially fit into docom-future?

regiskuckaertz: Luckily, the objects created from these prototypes get GC'd pretty quickly (they're never assigned); thus, even though this is terrible API design, there is no memory leak. Oliver's design keeps internal state, so there is room for improvement.

static/src/javascripts/projects/common/modules/commercial/creatives/hosted-thrasher.js
jbreckmckye: @jamesgorrie You can extend the prototype of any function - all JavaScript functions are objects:

```javascript
function Foo() {
    this.x = 1;
}

Foo.prototype.y = 2;
```

static/src/javascripts/projects/common/modules/commercial/hosted-video.js
jbreckmckye: If it's just being used in `init`, though, perhaps it makes sense to keep the var declaration scoped there? Just to keep the outer scope as simple as possible.

static/src/javascripts/projects/common/modules/commercial/creatives/hosted-thrasher.js
jbreckmckye: Yeah, this isn't really something I think needs to be fixed here and now. It's just a bit of a bee in my bonnet - my rationale has always been that as soon as a symbol is public, when I maintain the code I now have to worry about whether anything is using it downstream.

Might be worth discussing over on dotcom-future, as James says.

static/src/javascripts/projects/common/modules/commercial/creatives/hosted-thrasher.js
```diff
@@ -0,0 +1,35 @@
...
+    };
+
+    HostedThrasher.prototype.create = function () {
+        hostedThrasherTemplate = hostedThrasherTemplate || template(hostedThrasherStr);
+
+        fastdom.write(function () {
+            this.$adSlot.append(hostedThrasherTemplate({ data: this.params}));
+        }, this).then(hostedVideo.init);
+
+        if (this.params.trackingPixel) {
```
jbreckmckye: Maybe do this in the fastdom call, too?

Other than that, though, it all looks good.

uplne:
👍

jbreckmckye:
Great - 👍

PR #12927, opened 2016-05-16
Update space-filler to use the queueAsync module

jbreckmckye:
And make our tests a little more terse whilst we're at it.


static/src/javascripts/projects/common/modules/article/spacefinder.js
```diff
@@ -16,7 +16,7 @@ define([
     curry
 ) {
     // total_hours_spent_maintaining_this = 56
```
uplne: You should update this

static/src/javascripts/projects/common/modules/article/space-filler.js
```diff
@@ -2,47 +2,57 @@ define([
...
+        function onSpacesFound(paragraphs) {
             return fastdom.write(function () {
-                writer(slots);
+                writer(paragraphs);
                 return true;
             });
         }
+
+        function onNoSpacesFound(ex) {
+            if (ex instanceof spacefinder.SpaceError) {
```
regiskuckaertz: 👍  much better

static/src/javascripts/projects/common/modules/article/space-filler.js
```diff
@@ -2,47 +2,57 @@ define([
     'common/utils/fastdom-promise',
     'Promise',
     'raven',
-    'common/modules/article/spacefinder'
+    'common/modules/article/spacefinder',
+    'common/modules/commercial/dfp/QueueAsync'
```
regiskuckaertz: I thought about this. Have you checked if memory is released in chained promises, once the promise is resolved? Might be good to have a look at the polyfill, as chaining many times can have bad results. Alternatively a simple FIFO queue where you override the `push` method might be a better alternative, especially since we're not really interested in the resolved value of the queueing process itself.

jbreckmckye: This is a good question, and seemed worth investigating.

I created a simple test 'rig' that I could call from the console -

```javascript
window.test = {};
window.test.count = 0;
window.test.qa = QueueAsync;
window.test.op = function () {
    return new Promise(function (resolve, reject) {
        var collection = [];
        window.test.count++;
        for (var i = 0; i < 999; i++) {
            collection.push({});
        }
        window.setTimeout(resolve, 0);
    });
};
window.q = new QueueAsync();
window.test.lotsOps = function () {
    var operations = [];
    for (var i = 0; i < 4999; i++) {
        operations.push(window.q.add(window.test.op))
    }
    return Promise.all(operations);
};
```

Then, in Chrome, I

1. Started recording for a few seconds;
2. Ran `window.test.lotsOps()`;
3. Waited for a console.log set to fire when the lotsOps promise resolved;
4. Waited another few seconds;
5. Triggered a hard GC;
6. Recorded a few seconds afterwards for good measure

...And then looked at my memory allocation graph:

image

You can see the heap being thrashed as `test.op` generates its large arrays and incremental GC reclaims the memory. When `lotsOps` completes, there are two sets of 'steps', before I manually trigger garbage collection and heap usage plummets to levels comparable to the start of the run.

On this basis, it seems that if there is memory leakage in our Promise shim, then it doesn't seem to be substantial enough to at least notice on this kind of scale. QueueAsync doesn't seem to leak references to its historical promises, or to the closures around those promises.

regiskuckaertz:
Now that the maintenance count has been update 😆  👍

PR #12928, opened 2016-05-16
Only allow http(s) URLs in CSP reports

desbo:
This prevents things like `about:blank`,`safari-extension://` getting into the reports.


TBonnin:
👍

PR #12929, opened 2016-05-16
Tidy high merch report

rich-nguyen:
Tidy of the high merchandising line item admin report page.

screen shot 2016-05-16 at 16 33 00


Calum-Campbell:
Looks much better 👍 !

PR #12933, opened 2016-05-17
Extend commercial high server side check

Calum-Campbell:
## What does this change?

Optimises the rendering of commercial-high slots so that they only render when there is a match based on both the page tags and adUnits with the known commerical-high tags and keywords, which can be seen on the frontend.gutools.co.uk/commercial/high-merchandising page. 
Previously had only been checking based on tags. This was leading to a large number of empty ad units being served.

Should see a significant drop in empty ad units being served and reported to Sentry.

No
## Request for comment

@kelvin-chappell @JonNorman @rich-nguyen


common/app/common/dfp/DfpAgent.scala
```diff
@@ -37,7 +40,10 @@ object DfpAgent
   protected def tagToSponsorsMap = tagToSponsorsMapAgent get()
   protected def tagToAdvertisementFeatureSponsorsMap = tagToAdvertisementFeatureSponsorsMapAgent get()
   protected def inlineMerchandisingTargetedTags: InlineMerchandisingTagSet = inlineMerchandisingTagsAgent get()
-  protected def highMerchandisingTargetedTags: HighMerchandisingTargetedTagSet = highMerchandisingTargetedTagsAgent get()
+  protected def highMerchandisingTargetedTags: HighMerchandisingLineItemTargetsSeq = highMerchandisingTargetedTagsAgent get()
+
+  protected def highMerchandisingTargets: HighMerchandisingLineItemTargetsSeq = highMerchandisingTargetedLineItemsAgent get()
```
JonNorman: Do we still need this?

common/app/common/dfp/DfpAgent.scala
```diff
@@ -113,12 +129,18 @@ object DfpAgent
...
+        if(freshData.lineItems.nonEmpty) freshData else oldData
       }
     }
 
+//    def updateHighMerchandisingLineItemTargets(freshData: HighMerchandisingLineItemTargetsSeq): Unit ={
+//      highMerchandisingTargetedLineItemsAgent sendOff { oldData =>
+//        if(freshData.lineItems.nonEmpty) freshData else oldData
+//      }
+//    }
+
```
JonNorman: Comment sneak-in!

common/app/common/dfp/TagSponsorship.scala
```diff
@@ -77,13 +92,38 @@ object HighMerchandisingTargetedTagSet {
   }
 }
 
+
 case class HighMerchandisingTargetedTagSet(items: Set[String] = Set.empty){
 
   def hasTag (tag: Tag): Boolean = items.exists(item => tag.id.endsWith(item))
 
+  def hasAdUnit (adUnitSuffix : String): Boolean = {
```
kelvin-chappell: This implementation looks a bit odd.

common/app/common/dfp/DfpAgent.scala
```diff
@@ -28,7 +28,10 @@ object DfpAgent
   private lazy val tagToSponsorsMapAgent = AkkaAgent[Map[String, Set[String]]](Map[String, Set[String]]())
   private lazy val tagToAdvertisementFeatureSponsorsMapAgent = AkkaAgent[Map[String, Set[String]]](Map[String, Set[String]]())
   private lazy val inlineMerchandisingTagsAgent = AkkaAgent[InlineMerchandisingTagSet](InlineMerchandisingTagSet())
-  private lazy val highMerchandisingTargetedTagsAgent = AkkaAgent[HighMerchandisingTargetedTagSet](HighMerchandisingTargetedTagSet())
+  private lazy val highMerchandisingTargetedTagsAgent = AkkaAgent[HighMerchandisingLineItemTargetsSeq](HighMerchandisingLineItemTargetsSeq())
+
+  private lazy val highMerchandisingTargetedLineItemsAgent = AkkaAgent[HighMerchandisingLineItemTargetsSeq](HighMerchandisingLineItemTargetsSeq())
```
JonNorman: Do we still need this?

common/app/common/dfp/DfpAgent.scala
Calum-Campbell: We do not, fixed

common/app/common/dfp/TagSponsorship.scala
Calum-Campbell: @kelvin-chappell Ah yes left over from before, thanks, removed now

common/app/model/meta.scala
```diff
@@ -103,13 +103,13 @@ final case class Commercial(
...
-  def hasHighMerchandisingTarget: Boolean = {
-    DfpAgent.hasHighMerchandisingTarget(tags.tags)
+  def hasHighMerchandisingAdUnitAndTag: Boolean = {
+    DfpAgent.hasHighMerchAdAndTag(metadata.adUnitSuffix,tags.tags)
   }
 
   def conditionalConfig: Map[String, JsValue] = {
-    val highMerchansisingMeta = if (hasHighMerchandisingTarget) {
-      Some("hasHighMerchandisingTarget", JsBoolean(hasHighMerchandisingTarget))
+    val highMerchansisingMeta = if (hasHighMerchandisingAdUnitAndTag) {
```
kelvin-chappell: Typo here.

common/app/common/dfp/TagSponsorship.scala
```diff
@@ -77,13 +92,38 @@ object HighMerchandisingTargetedTagSet {
...
+    println(adUnitSuffix)
+    true
+  }
+
 
   def nonEmpty = items.nonEmpty
+}
+
+case class HighMerchandisingLineItemTargetsSeq(lineItems : Seq[HighMerchandisingLineItem] = Seq.empty) {
```
kelvin-chappell: Best to avoid default values for case class properties, if possible.

common/app/common/dfp/TagSponsorship.scala
```diff
@@ -77,13 +92,38 @@ object HighMerchandisingTargetedTagSet {
...
+
+  def hasAdUnitAndTag (adUnitSuffix: String, tags:Seq[Tag]): Boolean = {
+
+    val tagSeq = tags.map(tag => tag.name)
+    val tagSeqModify = tagSeq.map(oldTag => oldTag.replaceAll(" ","-").toLowerCase)
```
kelvin-chappell: Could do these two steps in one:
`tags.map(_.name.replaceAll(" ","-").toLowerCase)`

common/app/conf/switches/CommercialSwitches.scala
```diff
@@ -286,7 +286,7 @@ trait CommercialSwitches {
 
   val highMerchandisingComponentSwitch = Switch(
     SwitchGroup.Commercial,
-    "render-commercial-high-slot-always",
+    "optomise-High-Merchandising",
```
kelvin-chappell: Typo here.

rich-nguyen: And lower case all parts

common/app/views/fragments/contentFooter.scala.html
```diff
@@ -19,7 +19,7 @@
     } {
         <div class="js-fronts-network-2" data-link-name="fronts-on-articles | network-front | block-2"></div>
     } {
-        @if(content.commercial.hasHighMerchandisingTarget) {
+        @if(content.commercial.hasHighMerchandisingAdUnitAndTag) {
```
rich-nguyen: It still has an odd name, for someone reading the content footer. A higher-level name would hide the implementation, and reveal the intent. For example `needsHighMerchandisingSlot`

Calum-Campbell: @rich-nguyen agreed, changed now.

common/app/common/dfp/TagSponsorship.scala
Calum-Campbell: @kelvin-chappell @rich-nguyen have now included all in HighMerchandisingLineItem case class rather than have a separate target case class.

common/app/common/dfp/TagSponsorship.scala
```diff
@@ -100,10 +83,18 @@ case class HighMerchandisingLineItem(
...
-case class HighMerchandisingLineItems(items: Seq[HighMerchandisingLineItem]) {
-  val sortedItems = items.sortBy(_.name)
+
+  def matchesAdUnitAndTag (adUnitSuffix: String, pageTags:Seq[Tag]): Boolean = {
+
+    def tagNames = pageTags map (_.name) map (_.replaceAll(" ","-").toLowerCase)
+
+    def matchesTag: Boolean = tagNames.exists(tags.contains)
+
+    def matchesAdUnit: Boolean = adUnits.exists(_.path contains adUnitSuffix)
```
kelvin-chappell: These could all be `vals` as they're only evaluated once and don't have any parameters.

Calum-Campbell: ok, is that the best practice in general ? If something is evaluated multiple time it will be required to be a def, but if only once it is best to set as val to be more explicit/clear?

kelvin-chappell: Yes, in general.  
You could argue in this case that `matchesAdUnit` doesn't need to be evaluated if `matchesTag` is false so it could be a `lazy val` instead.  But the benefit would be very small.

JonNorman: Good to know - I thought that since these methods will only be called once (and so the cost of calling them will be the same) that, using a `def`, there will be no subsequent val-storage cost. I suppose given the piddly amount of space we are talking, it makes more sense to be clearer about intent rather than worrying about a few bytes.

Calum-Campbell: Thanks Kelvin, good to know what the best practices are. Have updated.

kelvin-chappell:
👍

PR #12934, opened 2016-05-17
Remove dev-build route for sponsored-containers

scottp719:
## What does this change?

Removes missed route on dev-build to fix local

Stops things breaking

n/a
## Request for comment

@johnduffell @JonNorman


JonNorman:
:+1:

PR #12935, opened 2016-05-17
Link preconnect

rich-nguyen:
We've got pretty good coverage of `<link rel="preconnect">` now, which should allow us to hint to the browser that we want to open a connection. The aim is to reduce delays caused by opening connections during the critical path.

Currently, we use `dns-prefetch`, which has marginally better browser support, but doesn't pre-emptively perform the tcp connection (in orange, below), or the ssl negotiation (in purple). I want to see if `preconnect` saves us time, as increasingly move more things to TLS.

I've also made omniture always-ssl. The server was assuming this, but the client needs to be specifically targeting ssl. Without always-ssl, the dns-prefetch/preconnect was redundant.

screen shot 2016-05-17 at 11 56 35

More info [here](https://www.igvita.com/2015/08/17/eliminating-roundtrips-with-preconnect/)

### DIFF
--- a/common/app/conf/switches/PerformanceSwitches.scala
+++ b/common/app/conf/switches/PerformanceSwitches.scala
@@ -224,4 +224,12 @@ trait PerformanceSwitches {
     exposeClientSide = false
   )
 
+  val UseLinkPreconnect = Switch(
+    SwitchGroup.Performance,
+    "use-link-preconnect",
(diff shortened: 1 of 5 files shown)


gtrufitt:
Nice! 👍

johnduffell:
good idea! 👍

PR #12936, opened 2016-05-17
Simplify badges

superfrank:
Change the switch to be more generic, so that if we wish to turn off a badge we can remove it from Badges.scala

cc @desbo @blongden73


common/app/conf/switches/FeatureSwitches.scala
```diff
@@ -342,19 +342,11 @@ trait FeatureSwitches {
...
-    safeState = On,
-    sellByDate = new LocalDate(2017, 1, 5),
-    exposeClientSide = false
-  )
-  // Owner: Ben Longden 
-  val AUSElectionSwitch = Switch(
-    SwitchGroup.Feature,
-    "aus-election",
-    "When ON, items tagged with australia-news/australian-election-2016 will have visual elements added",
+    "eu-referendum",
```
desbo: needs a better name here

superfrank: You need a better name

desbo:
👍

PR #12937, opened 2016-05-17
Use fapi metadata to decide if container branded

kelvin-chappell:


common/app/model/pressedContent.scala
```diff
@@ -51,7 +53,9 @@ final case class CollectionConfig(
   excludeFromRss: Boolean,
   showTimestamps: Boolean,
   hideShowMore: Boolean
-)
+) {
+  def showBranding = metadata contains Branded
```
kelvin-chappell: There's an obvious bug here.

JonNorman:
The end is nigh! :+1:

PR #12939, opened 2016-05-17
Re-arrange user features module

jbreckmckye:
I thought I might improve the readability of `user-features.js` by grouping up private functions and using prototype assignments to clearly mark public methods.

I also got rid of the single-var declaration.


johnduffell:
not sure grouping by visibility makes sense in the general case, better to group together in cohesive units, however it's not a big file so any way is ok! 👍 
also bear in mind the git blame will be overwritten with your name on almost the whole file, which makes it harder to trace the origin of code.

jbreckmckye:
Well, it's not so much grouping by visibility, as by code path.

I know what you mean about Git blame, but I'm not sure we should avoid rearranging code just to keep annotations - we'd never be able to fix stuff, and a lot of them aren't that helpful anyway.

johnduffell:
yeah agreed! ship it 💨 😉

PR #12941, opened 2016-05-17
Serve hosted by container logo from dfp, and adjust the text positioning

lps88:
## What does this change?

Use a logo from dfp instead of the static one. Move the text a bit.


uplne:
👍

PR #12943, opened 2016-05-17
Add Facebook Native Ads to our creative portfolio

regiskuckaertz:
Will allow adops to make Facebook creatives compete with other programmatic networks via DFP. I styled the ad to look like one of our own.

picture 5

NB: ultimately, we'll use the official Facebook App ID, this one is temporary.

cc @jbreckmckye @uplne @JonNorman @kenlim


uplne:
👍

PR #12945, opened 2016-05-17
Update us briefing check so it doesn't look at the article structure

gtrufitt:
## What does this change?

Updates the us briefing sign-up page so it doesn't check the structure on the us briefing pages because they always end in a paragraph.


sndrs:
👍 :goldendiamond:

jbreckmckye:
👍

gtrufitt:
Double thumbs, hit the big time! Thanks :D

PR #12947, opened 2016-05-17
Make passbacks work with fabric ads

jbreckmckye:
When we're on a mobile article, there's no top nav slot, so we need to make the first inline slot request a Fabric 88x71 advert. To do this, I created a new adslot definition, _inline-fabric-top_.

However, this breaks passbacks, which it appear require a particular set of names ('inline', 'inline1' ... 'inline8') for retargeting. It may also break inreads, which need a `dfp-ad--inline1` slot ID.

To fix this in the short term, I've added a `name` override for the fabric definition.

A longer-term fix might be to change the `inline-1` definition to target 88x71s in mobile specifically, and fix dfp-api's `buildSizeMapping` to only use slot sizes on the largest matching breakpoint. I'd need to do a little sniffing to see if this would break anything else in turn first, though.

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/article-body-adverts.js
+++ b/static/src/javascripts/projects/common/modules/commercial/article-body-adverts.js
@@ -98,7 +98,7 @@ define([
         function insertInlineAd(paras) {
             bodyAds += 1;
             var isFabricTopReplacement = config.switches.fabricAdverts && (bodyAds === 1) && detect.isBreakpoint({max: 'phablet'});
-            var adDefinition = isFabricTopReplacement ? 'inline-fabric-top' : ('inline' + bodyAds);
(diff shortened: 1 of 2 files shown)


static/src/javascripts/projects/common/modules/commercial/article-body-adverts.js
```diff
@@ -98,7 +98,10 @@ define([
         function insertInlineAd(paras) {
             bodyAds += 1;
             var isFabricTopReplacement = config.switches.fabricAdverts && (bodyAds === 1) && detect.isBreakpoint({max: 'phablet'});
-            var adDefinition = isFabricTopReplacement ? 'inline-fabric-top' : ('inline' + bodyAds);
+
+            // Note that the names 'inline1' to 'inline8' are used by passbacks for retargeting -
+            // check with adOps before changing these labels
```
regiskuckaertz: 👍

regiskuckaertz:
👍

PR #12948, opened 2016-05-17
Eu ref badge

superfrank:
Adding the EU SVG badge to articles for both EU reality check and EU referendum (ignore the fact it's The Fiver in the screenshot)

image

@desbo @blongden73


desbo:
👍

PR #12950, opened 2016-05-17
Add to Homescreen: use proper roundel

OliverJAsh:
It's me again!

When one adds dotcom to their homescreen, the browser will use the square icon as per the manifest. This looks really bad against other icons which are not square!

Could we use the pure roundel instead, like the native app does?

Also I'm intrigued to know why we serve a 114px version?
# Before

image
# After

image


jamesgorrie:
> why we serve a 114px version?

[According to this, retina iPad](http://www.kylejlarson.com/blog/creating-retina-images-for-your-website/).

jbreckmckye:
👍 

This seems a reasonable change to me. Oliver, do you still have merge permissions?

OliverJAsh:
Looks like I do, but of course I won't be able to see it out :-(

sndrs:
nice one!

PR #12951, opened 2016-05-17
Badge - Update to https asset

mchv:
US elections liveblogs should not see a mixed-content warning when moving to https


TBonnin:
👍

PR #12952, opened 2016-05-17
Article  - use secureFile instead of file if available for tweet

mchv:
This is a follow-up of [Natalia changes](https://github.com/guardian/frontend/pull/11676/files) and should prevent mixed content error for `http://pbs.twimg.com` host (`805` today).

@desbo


jamesgorrie:
If we are going to serve over https, should we just not serve the file?
👍

mchv:
@jamesgorrie if you mean having all `file` being `https` in composer the API directly,  I think that is sensible but should be a second step.

PR #12953, opened 2016-05-17
resolve with object that has properties

jamesgorrie:
## What does this change

Being more explicit about the object that we can't et as we don't always have the canonical URL.
Doing this because we potentially have some tracking issues. 

There were no errors thrown as, apparently, you can have `false.whatever` or even `false.true` in JS and  all is well.

Nope
## Request for comment

@akash1810


akash1810:
oh wow, ❤️ js!

👍

PR #12954, opened 2016-05-17
Only send video id to player if mediaType is present on fronts

jamesgorrie:
## What does this change?

At present we send the article ID through to the `video` snippet, which is wrong as we expect the video page ID. 

Unfortunately this isn't present on the `ContentCard`, so we don't send it through unless we have `mediaType` proving that this is a card for video.

This should stop the current click-twice-to-play video, but also start pushing through stats, which are not working at the moment.

Click once - get video.

Nope
## Request for comment

@akash1810


akash1810:
👍

PR #12956, opened 2016-05-18
Update us campaign minute text

gtrufitt:
Showing different text for the minute email sign-up when on election pages


NathanielBennett:
:+1:

PR #12957, opened 2016-05-18
Fix Page Skins not being rendered when they should be

JonNorman:
## How we determine whether to render a Page Skin

In order to determine whether a page has a Page Skin, we look at the `PageSkinSponsorship`s that we have; we look for the **first** `PageSkinSponsorship` that matches the Ad Unit and Edition of the page. We then check to see whether we are running in production and whether this `PageSkinSponsorship` is behind a test cookie.
## The problem

When we have multiple Page Skin sponsorships for the same Ad Unit (some behind Ad Tests, some not), then if we only look at the first matching Ad Test, we may end up inspecting the Ad Test sponsorship with the test cookie, rather than the pure, puckka production-ready sponsorship, and falsely concluding that we shouldn't render a Page Skin on this page.
## The fix

Just look at all the `PageSkinSponsorship`s instead of the first one that matches our page criteria.

Page Skins should work again.

No.


jbreckmckye:
Cool 👍

PR #12958, opened 2016-05-18
eu-referendum tag disabled

superfrank:
Editorial want to hold off on the badging on all EU content for now and just have EU reality check tagged.

@desbo


gtrufitt:
👍

PR #12959, opened 2016-05-18
spring cleaning: remove unused dependencies

alexduf:
## What does this change?

It removes the unused dependencies I identified. If you know one of these is used or other should be removed, please let me know.
Some of these dependencies (gdata) are conflicting with play 2.5

faster compilation time and upgrade to play2.5 less painful

No
## Screenshots

N/A
## Request for comment

@TBonnin @johnduffell


TBonnin:
if stuff compiles, go for it 👍

johnduffell:
brilliant, excellent idea! 👍

johnduffell:
also the benefit would be smaller artifacts so faster deploys!

PR #12962, opened 2016-05-18
Revert "Revert "add timeout to contrib-ads on video to be larger than IMA timeout""

jamesgorrie:
Reverts guardian/frontend#12913

We've confirmed that this doesn't affect the tracking, and does help with the problem.

@akash1810


akash1810:
👍

PR #12964, opened 2016-05-18
Revert "Revert "Revert "add timeout to contrib-ads on video to be larger than IMA timeout"""

jamesgorrie:
Reverts guardian/frontend#12962

Seen a drop in traffic - reverting so as not to cause an issue.

Apologies - saying that, this doesn't seem to fix the issue everytime anyway.


akash1810:
👍

PR #12965, opened 2016-05-18
add classes to explain video cards better

jamesgorrie:
## What does this change?

This will help us differentiate between cards to video pages and articles with main media that is video.

It will help us with some tests we'd like to run on fronts.

Nope
## Screenshots

fact
## Request for comment

@akash1810

### DIFF
--- a/common/app/views/fragments/items/facia_cards/contentCard.scala.html
+++ b/common/app/views/fragments/items/facia_cards/contentCard.scala.html
@@ -44,7 +44,6 @@
             @kicker(item.header, List("fc-item__kicker--dreamsnap", "fc-item__kicker--dreamsnap-list"))
         }
 
-
         @item.displayElement.filter(const(item.showDisplayElement)) match {
             case Some(InlineVideo(videoElement, title, endSlatePath, fallback)) if item.cardTypes.showVideoPlayer => {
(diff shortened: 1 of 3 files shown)


common/app/views/fragments/items/facia_cards/contentCard.scala.html
```diff
@@ -59,7 +58,7 @@
                     // ID is actually pointing to the article, which is wrong. It would be nice to have the ID on the
                     // ContentCard, but we don't for now. Annoyingly this doesn't allow us to check for if we should
                     // play ads or if the video is expired, but as fronts change really often, this is a non-problem.
-                    item.mediaType.flatMap(_ => item.id)
+                    if (item.isMediaLink) item.id else None
```
akash1810: what are the effects of using `None`?

jamesgorrie: Same as before, the embed path is not added to the video (as it shouldn't be with article links).

jamesgorrie: It's the same as the map, but I felt it's more explanitory.

akash1810:
:+1:

PR #12966, opened 2016-05-18
Removed margin

zeftilldeath:
A margin over ride is making things tight and ugly, simple little change.

From this:
screen shot 2016-05-18 at 14 58 34

To this:
screen shot 2016-05-18 at 14 58 42


jamesgorrie:
I wonder why the override was in there in the first place?
👍

PR #12968, opened 2016-05-18
Refactor LiveEvents parsing and Agent

JonNorman:
## What does this change?

Nothing really, just tidier; there was some date filtering done on the parsing that should be done on tha Agent and is otherwise quite hard to detect.

Easier to read and reason about

No


johnduffell:
7 days old 😨 
get it shipped 🐎 
👍

PR #12970, opened 2016-05-18
Extend welcome banner switch date

marialivia16:
## What does this change?

Extends the expiry date for the welcome banner A/B test switch. It expires tonight and not changing it will break build. We are still keeping the switch because we will use it for the actual test (so far it has been used for a pretest).

@stephanfowler

### DIFF
--- a/common/app/conf/switches/ABTestSwitches.scala
+++ b/common/app/conf/switches/ABTestSwitches.scala
@@ -118,7 +118,7 @@ trait ABTestSwitches {
     "ab-play-video-on-fronts",
     "Don't play video on fronts, but auto play when in article",
     safeState = Off,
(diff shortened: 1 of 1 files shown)


common/app/conf/switches/ABTestSwitches.scala
```diff
@@ -118,7 +118,7 @@ trait ABTestSwitches {
     "ab-play-video-on-fronts",
     "Don't play video on fronts, but auto play when in article",
     safeState = Off,
-    sellByDate = new LocalDate(2016, 5, 25),
```
stephanfowler: 👍 - though there's no rush on that date...

marialivia16: I can extend more it if necessary in the next PR with the actual test.

PR #12971, opened 2016-05-18
Upgrade capi client to 8.10

TBonnin:
## What does this change?

Upgrading capi client to 8.10
Upgrading sbt to 0.13.11
Removing facia-json dependency that is not required anymore
Fixing failing tests after refereshing the test data file
## What is the value of this and can you measure success?

Frontend is throwing out thrift related errors http://kibana.gu-web.net/goto/16cb09613e0643ce1d95e00816f14db4 which should be fixed in recent capi client versions according to @mchv 
## Does this affect other platforms - Amp, Apps, etc?

No
## Request for comment

@johnduffell @rich-nguyen @jamespamplin

PR #12938, opened 2016-05-17
Remove IntersperseMultipleStoryPackages a/b test

TBonnin:
Remove the IntersperseMultipleStoryPackages A/B testing
We are going forward with intermixing stories when there are more than
one package.
## Does this affect other platforms - Amp, Apps, etc?

No
## Request for comment

@piuccio @johnduffell

PR #12972, opened 2016-05-18
Intermix packages stories only if more than one packages

TBonnin:
## What does this change?

Second try at #12938. After deploying it the application and article instances latency went to the roof
The issue didn't seem to be with those change directly but with a problem with the capi client when hitting this line https://github.com/guardian/frontend/blob/tbonnin-remove-ABIntersperseMultipleStoryPackagesStories/common/app/model/RelatedContent.scala#L49
Logs of the issue can be found at http://kibana.gu-web.net/goto/16cb09613e0643ce1d95e00816f14db4

Before this change, the faulty code was only executed for people in the A/B test group which was 1% of readers. I guess the servers were able to recover since it was not happening this often.

This PR is dependent on capi client upgrade: https://github.com/guardian/frontend/pull/12971 :)

I also added an improvement where the zipping, sorting, removing duplicates only happens where there are more than one packages

Removing a/b test

No
## Request for comment

@johnduffell

### DIFF
--- a/common/app/model/RelatedContent.scala
+++ b/common/app/model/RelatedContent.scala
@@ -15,26 +15,9 @@ case class RelatedContentItem (
   faciaContent: PressedContent
 )
 
-case class RelatedContent ( items: Seq[RelatedContentItem], storyPackagesCount: Int = 0) {
+case class RelatedContent (items: Seq[RelatedContentItem]) {
   val hasStoryPackage: Boolean = items.nonEmpty
   val faciaItems: Seq[PressedContent] = items.map(_.faciaContent)
-}
-
-object RelatedContent {
-  def apply(parent: ContentType, response: ItemResponse): RelatedContent = {
-    // It's misleading to use storyPackage here rather than relatedContent. A tidy up should rename this object
-    val storyPackagesContent = response.packages.map { packages =>
-      packages.flatMap { p =>
-        p.articles.map(_.content)
-      }
-    }.getOrElse(List.empty)
-
-    val items = storyPackagesContent.map { item =>
-      val frontendContent = Content(item)
-      RelatedContentItem(frontendContent, FaciaContentConvert.contentToFaciaContent(item))
-    }
-    RelatedContent(items.filterNot(_.content.metadata.id == parent.metadata.id))
-  }
 }
 
 object StoryPackages {
@@ -52,6 +35,6 @@ object StoryPackages {
       val frontendContent = Content(item)
       RelatedContentItem(frontendContent, FaciaContentConvert.contentToFaciaContent(item))
     }
-    RelatedContent(items.filterNot(_.content.metadata.id == parent.metadata.id), response.packages.fold(0)(_.size))
+    RelatedContent(items.filterNot(_.content.metadata.id == parent.metadata.id))
   }
 }
--- a/common/app/model/RelatedContent.scala
+++ b/common/app/model/RelatedContent.scala
@@ -28,8 +28,10 @@ object StoryPackages {
       if(packages.size > 1) { //intermix packages only if more than one
         allContentsPerPackage
          .flatMap(_.zipWithIndex) // zip content with its position
+         .groupBy(_._1.id).map(_._2.head).toSeq // remove duplicates (Note: not using `distinct` here that would require
+                                                // hashing/comparing the whole ApiContent object but instead grouping based on `id`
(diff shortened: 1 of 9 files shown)


common/app/model/RelatedContent.scala
```diff
@@ -39,13 +39,17 @@ object RelatedContent {
 
 object StoryPackages {
   def apply(parent: ContentType, response: ItemResponse): RelatedContent = {
-    val storyPackagesContent = response.packages.map { packages =>
-      packages.map { p =>
-        p.articles.map(_.content)
+
+    val storyPackagesContent: Seq[ApiContent] = response.packages.map { packages =>
+      val allContentsPerPackage: Seq[Seq[ApiContent]] = packages.map(_.articles.map(_.content))
+      if(packages.size > 0) { //intermix packages only if more than one
```
gtrufitt: Isn't `> 0` 1 or more rather than if more than one?

TBonnin: it should be `> 1`. Thanks for catching this

johnduffell:
wow that's a good spot, interesting that distinct really messed it up although.  distinct on an unsorted list should only have n log n complexity, but maybe the content was long enough to make comparisons slow?

common/app/model/RelatedContent.scala
```diff
@@ -39,13 +39,17 @@ object RelatedContent {
...
+        allContentsPerPackage
+         .flatMap(_.zipWithIndex) // zip content with its position
+         .sortBy(_._2).map(_._1)  // sort by position and extract content
+         .distinct                // remove duplicates
```
johnduffell: is this distinct happening on a list of apicontent?  what does content.equals do? can you distinct-by(id) instead of the whole thing?

johnduffell:
the capi client upgrade might help, but I think we can save effort on the distincting assuming my assumptions are correct

TBonnin:
@johnduffell good point. I am gonna ditch distinct (which is hashing the whole object) and use `.groupBy(_.id).map(_._2.head).toList` that would not require hashing the whole object :) Thanks

johnduffell:
that's a really smart solution, I like it!

common/app/model/RelatedContent.scala
```diff
@@ -39,13 +39,19 @@ object RelatedContent {
...
-      packages.map { p =>
-        p.articles.map(_.content)
+
+    val storyPackagesContent: Seq[ApiContent] = response.packages.map { packages =>
+      val allContentsPerPackage: Seq[Seq[ApiContent]] = packages.map(_.articles.map(_.content))
+      if(packages.size > 1) { //intermix packages only if more than one
+        allContentsPerPackage
+         .flatMap(_.zipWithIndex) // zip content with its position
+         .sortBy(_._2).map(_._1)  // sort by position and extract content
+         .groupBy(_.id).map(_._2.head).toList // remove duplicates (not using `distinct` here that would require
```
johnduffell: I feel like doing groupBy since it produces a map, will lose the ordering created in the previous step

TBonnin: You are right. I guess I would need to sort after removing duplicates

TBonnin:
@johnduffell https://github.com/guardian/frontend/pull/12972/commits/78c8436a4843b0e31bcfafd5243ed33d41d84ea4#diff-9b3d36cd982208831cf027d918e19a9cR48 :)

common/app/model/RelatedContent.scala
```diff
@@ -15,43 +15,32 @@ case class RelatedContentItem (
...
+
+    val storyPackagesContent: Seq[ApiContent] = response.packages.map { packages =>
+      val allContentsPerPackage: Seq[Seq[ApiContent]] = packages.map(_.articles.map(_.content))
+      if(packages.size > 1) { //intermix packages only if more than one
+        allContentsPerPackage
+         .flatMap(_.zipWithIndex) // zip content with its position
+         .groupBy(_._1.id).map(_._2.head).toSeq // remove duplicates (Note: not using `distinct` here that would require
+                                                // hashing/comparing the whole ApiContent object but instead grouping based on `id`
+                                                // and then collecting the first element of each group of duplicates)
+         .sortBy(_._2).map(_._1)  // sort by position and extract content
```
johnduffell: 🎉 
+thank goodness for your commenting

johnduffell:
great thanks! 👍

PR #12973, opened 2016-05-18
Try and make dfp-api.js less rubbish

jbreckmckye:
I am trying to turn...

image

into...

image

...by reordering the functions inside `dfp-api` so that procedures in particular code paths are at least a little better grouped.

What does this mean specifically?
- functions are grouped into blocks concerning particular parts of the ad lifecycle
- functions are followed directly by their (unique) dependencies
- whilst we're thrashing our `git blame` annotations anyway, replace the old single-var-declaration style


static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
```diff
@@ -247,169 +241,93 @@ define([
...
-    function allAdsRendered(adSlotId) {
-        if (adverts[adSlotId] && !adverts[adSlotId].isRendered) {
-            adverts[adSlotId].isLoading = false;
-            adverts[adSlotId].isRendered = true;
-        }
-
-        if (every(adverts, 'isRendered')) {
-            userTiming.mark('All ads are rendered');
-            mediator.emit('modules:commercial:dfp:alladsrendered');
+    function setPublisherProvidedId() {
```
regiskuckaertz: Perhaps move this one in the previous block?

static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
```diff
@@ -424,19 +342,61 @@ define([
...
+
+    /**
+     * ADD SLOT
+     */
+
+    function addSlot(adSlot) {
+        var $adSlot = bonzo(adSlot);
+        var slotId = $adSlot.attr('id');
+
+        function displayAd ($adSlot) {
```
regiskuckaertz: 😭  this is so much better

static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
```diff
@@ -118,49 +119,64 @@ define([
...
-            });
-        }
-    };
     var renderStartTime = null;
-    var recordFirstAdRendered = once(function () {
-        beacon.beaconCounts('ad-render');
-    });
     var prebidEnabled = config.switches.headerBiddingUs && config.page.edition === 'US';
 
+    var windowResize = debounce(
```
regiskuckaertz: I'd move this one in the "refresh on window resize" block

static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
```diff
@@ -452,6 +412,149 @@ define([
...
+        return prebidEnabled && shouldLazyLoad() && !contains(excludedSlotIds, advert.adSlotId);
+    }
+
+    /**
+     * REFRESH ON WINDOW RESIZE
+     */
+
+    function refresh(breakpoint, previousBreakpoint) {
+        googletag.pubads().refresh(
+            chain(slotsToRefresh)
```
regiskuckaertz: I'd move the decl of `slotsToRefresh` in this block too, if possible

static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
jbreckmckye: Unfortunately, that will run afoul of eslint's `no-use-before-define` restriction, because `postDisplay` passes the var to a callback. Any suggestions?

static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
jbreckmckye: Good spot.

static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
regiskuckaertz: hmm, you could probably define `windowResize` as just being the inner func and use `debounce` when registering the listener inside `postDisplay`?

jbreckmckye: Yeah, but then I'd have to wipe all the listeners in the `dfp.reset` call, which feels like a kludge. I guess that's the most pragmatic option, though.

jbreckmckye: I ended up replacing the `postDisplay` call with a `refreshOnResize` function that handles the debouncing privately.

regiskuckaertz: 👍

static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
```diff
@@ -783,7 +781,7 @@ define([
             rendered = false;
             adverts = {};
             slotsToRefresh = [];
-            mediator.off('window:resize', windowResize);
+            mediator.removeAllListeners('window:resize');
```
regiskuckaertz: Isn't there a risk of removing the listeners set elsewhere? why not keep the `windowResize` but assign it inside `refreshOnResize`?

regiskuckaertz:
Aside from minor comment, it's really great to find some order in all that mess 👍

static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
jbreckmckye: Yeah, I did see this as a bit of a kludge. Think I've now found something that both we and the linter are happy with, though.

PR #12974, opened 2016-05-18
add baseline for showing controls on main media

jamesgorrie:
## What does this change?

Running a baseline to see how many people we need to run a test with showing the video controls to on the main media.

It appears we actually set a class on it called `gu-media--show-controls-at-start`, which does nothing, but instead of just adding it, I wouldn't mind testing it.

🔢 
## Request for comment

@akash1810 @harrysalmon

### DIFF
--- a/static/src/javascripts/projects/common/modules/experiments/tests/video-controls-on-main-media.js
+++ b/static/src/javascripts/projects/common/modules/experiments/tests/video-controls-on-main-media.js
@@ -1,0 +1,37 @@
+define([
+    'qwery',
+    'common/utils/config'
+], function (
+    qwery,
+    config
+) {
+    return function () {
+        this.id = 'VideoControlsOnMainMedia';
+        this.start = '2016-05-19';
+        this.expiry = '2016-05-26';
+        this.author = 'James Gorrie';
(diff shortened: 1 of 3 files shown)


static/src/javascripts/projects/common/modules/experiments/tests/video-controls-on-main-media.js
```diff
@@ -0,0 +1,37 @@
...
+        this.description = 'Showing controls on main media.';
+        this.audience = 1;
+        this.audienceOffset = 0;
+        this.successMeasure = '';
+        this.audienceCriteria = 'Landing on an article with main media';
+        this.dataLinkNames = '';
+        this.idealOutcome = '';
+
+        this.canRun = function () {
+            return config.page.contentType === 'Article' && qwery('[data-component-type="main video"]').length > 0;
```
akash1810: can you have more than one main video?

akash1810:
👍

static/src/javascripts/projects/common/modules/experiments/tests/video-controls-on-main-media.js
jamesgorrie: Nope.

PR #12975, opened 2016-05-19
Use https for old interactives without a body

mchv:
2 years ago @theefer [added some logic to use a prefixed host](https://github.com/guardian/frontend/pull/2935/files) for some interactives. I don't know which interactives go to this code path but the host should be updated. However before merging I need to check that this is not breaking them.

@gklopper @JustinPinner do you know any interactives (or way to find them) I could test with those changes?

### DIFF
--- a/common/app/common/configuration.scala
+++ b/common/app/common/configuration.scala
@@ -348,7 +348,7 @@ class GuardianConfiguration(val application: String, val webappConfDirectory: St
   }
 
   object interactive {
-    lazy val url = "http://interactive.guim.co.uk/next-gen/"
+    lazy val url = "https://interactive.guim.co.uk/next-gen/"
   }
 
   object javascript {


mchv:
Interestingly [the test](https://github.com/guardian/frontend/blob/3ef8f2c450dc0b2896b2cd2dfad79f9f096db237/common/test/assets/javascripts/spec/Interactive.spec.js) which was used to test this feature does not seem to on `master`.

JustinPinner:
It looks like all interactives that are handled by next-gen pass through this code. In terms of testing, I can send you a list of URLs that _shouldn't_ pass through here if that helps?

johnduffell:
@mchv this one's getting on a bit, are you waiting for anything?

mchv:
@johnduffell yes we discussed with @JustinPinner and we think this is only used for r2 interactives that have been archived. So we should only apply the changes when those r2 archives will have been migrated as a last step to make them work.

humaislam:
@mchv @JustinPinner hello - what shall we do with this PR?

sndrs:
47 days…

mchv:
@humaislam @sndrs @JustinPinner I think we simply wait r2 interactives to be ready and when they are we merge. If we don't plan to migrate them, or or don't mind to break those we care while migrating we could merge now.

johnduffell:
any updates? close and reopen when ready?

PR #12976, opened 2016-05-19
Extend commercial high server check further

Calum-Campbell:
## What does this change?

Adds one more filter into server side check for commercial-High adSlot. Now filter by Edition as well as adUnit and tags.

Should further decrease errors reported to Sentry regarding empty ad slots being served. 

No
## Request for comment

@kelvin-chappell @JonNorman @rich-nguyen

### DIFF
--- a/common/app/common/dfp/HighMerchandiseComponentAgent.scala
+++ b/common/app/common/dfp/HighMerchandiseComponentAgent.scala
@@ -7,8 +7,8 @@ trait HighMerchandiseComponentAgent {
 
   protected def targetedHighMerchandisingLineItems: Seq[HighMerchandisingLineItem]
 
-  def hasHighMerchAdAndTagandEdition(adUnitSuffix:String, tags: Seq[Tag],edition:String) = {
+  def isTargetedByHighMerch(adUnitSuffix:String, tags: Seq[Tag],edition:String) = {
     highMerchandisingComponentSwitch.isSwitchedOff ||
-    targetedHighMerchandisingLineItems.exists(_.matchesAdUnitAndTagAndEdition(adUnitSuffix, tags,edition))
+    targetedHighMerchandisingLineItems.exists(_.matchesPageTargeting(adUnitSuffix, tags, edition))
   }
 }
--- a/common/app/common/dfp/HighMerchandiseComponentAgent.scala
+++ b/common/app/common/dfp/HighMerchandiseComponentAgent.scala
@@ -1,5 +1,6 @@
 package common.dfp
 
+import common.Edition
 import model.Tag
 import conf.switches.Switches._
 
--- a/common/app/common/dfp/HighMerchandiseComponentAgent.scala
+++ b/common/app/common/dfp/HighMerchandiseComponentAgent.scala
@@ -8,8 +8,8 @@ trait HighMerchandiseComponentAgent {
 
   protected def targetedHighMerchandisingLineItems: Seq[HighMerchandisingLineItem]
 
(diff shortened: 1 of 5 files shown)


common/app/common/dfp/HighMerchandiseComponentAgent.scala
```diff
@@ -7,8 +7,8 @@ trait HighMerchandiseComponentAgent {
 
   protected def targetedHighMerchandisingLineItems: Seq[HighMerchandisingLineItem]
 
-  def hasHighMerchAdAndTag(adUnitSuffix:String, tags: Seq[Tag]) = {
+  def hasHighMerchAdAndTagandEdition(adUnitSuffix:String, tags: Seq[Tag],edition:String) = {
```
JonNorman: Now that we are checking three things, enumerating each check in the method name becomes a bit untenable; maybe consider renaming the method to something like `isTargetedByHighMerch` and then ensure that the subsequent code is clear as to what that means i.e. a combination of `matchesAdUnit`, `matchesTag` and `matchesEdition` etc

Calum-Campbell: Agreed it was getting a bit long.

common/app/common/dfp/TagSponsorship.scala
```diff
@@ -81,19 +81,24 @@ case class HighMerchandisingLineItem(
...
+
   val customTargets = customTargetSet.map(_.targets)
-  val editions = customTargets.flatMap(sequence => sequence.filter((target) => target.name == "edition")).map(target => target.values)
+  val editions = customTargets.flatMap(_.filter( _.name == "edition")).map(_.values).distinct
 
+  def matchesPageTargeting (adUnitSuffix: String, pageTags:Seq[Tag], edition:String): Boolean = {
 
-  def matchesAdUnitAndTag (adUnitSuffix: String, pageTags:Seq[Tag]): Boolean = {
+    // edition comes in the form "UK edition" and "US edition"
+    val cleansedPageEdition = edition.split(" ", 2).head.toLowerCase
```
kelvin-chappell: Could use `edition.id` instead

common/app/common/dfp/TagSponsorship.scala
```diff
@@ -81,19 +81,24 @@ case class HighMerchandisingLineItem(
...
 
-    val matchesTag: Boolean = tagNames.exists(tags.contains)
+    val matchesTag: Boolean = cleansedPageTagNames.exists(tags.contains)
 
     lazy val matchesAdUnit: Boolean = adUnits.exists(_.path contains adUnitSuffix)
 
-    matchesTag && matchesAdUnit
+    lazy val matchesEdition: Boolean = editions.isEmpty || editions.flatten.contains(cleansedPageEdition)
+
+    matchesTag && matchesAdUnit && matchesEdition
```
kelvin-chappell: You should probably have some tests for this logic.

common/app/views/fragments/contentFooter.scala.html
```diff
@@ -19,7 +20,7 @@
     } {
         <div class="js-fronts-network-2" data-link-name="fronts-on-articles | network-front | block-2"></div>
     } {
-        @if(content.commercial.needsHighMerchandisingSlot) {
+        @if(content.commercial.needsHighMerchandisingSlot(Edition(request).displayName)) {
```
kelvin-chappell: Better to pass through whole `Edition` instance so that you get the benefit of knowing the type in the method signature

kelvin-chappell:
You should check that there's no fall in ad impressions in the high-relevance slot, as well as the fall in Sentry errors.  You can probably get this from DFP, or possibly the data lake.

PR #12981, opened 2016-05-19
fix typo in selector

jamesgorrie:
## What does this change?

A string

It works

No
## Request for comment

@akash1810


akash1810:
👍

PR #12982, opened 2016-05-19
Add check for clashing ab tests to email sign-up

gtrufitt:
## What does this change?

Adds a check to the email sign-up to ensure that we're not in a test (initially star ratings) that will clash with the insertion of the form.
## Request for comment

@regiskuckaertz @sndrs Hi - I wondered if you two had a thought on this - seems a little kludgy and specific to add a check to the email sign-up module to check if we're in a particular ab tests.

The high-level problem here is having components, that are async lazy loaded taking priority over other components that are async lazy loaded - if we don't know when either component is going to be loaded the component that is the lower priority needs to know whether another component is ever going to load (probably with definitely, maybe, never values) - this may be via an ab test participation or it may be a proximity load or it may be based on a user's logged in status. I've done this with outbrain and email sign-up by exposing the the [email sign-ups](https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/common/modules/email/run-checks.js) `run-checks` so they can be [checked ahead of time by the Outbrain module](https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/common/modules/commercial/third-party-tags/outbrain.js#L141).

Would creating some kind of generic `canRun` function with different levels of priority be sensible? Am I over thinking this completely? 

Anyway, this technique works here but something to think about.

### DIFF
--- a/static/src/javascripts/projects/common/modules/email/run-checks.js
+++ b/static/src/javascripts/projects/common/modules/email/run-checks.js
@@ -11,6 +11,7 @@ define([
     'lodash/collections/contains',
     'common/modules/user-prefs',
     'common/modules/identity/api',
+    'common/modules/experiments/ab',
     'Promise'
 ], function (
     $,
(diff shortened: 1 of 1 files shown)


regiskuckaertz:
That's a very good question! I can't give a very detailed answer now but I'll think about it over the week-end. There's potentially a substantial performance cost trying to orchestrate modules around milestones that could be mitigated with a few heuristics. For instance, we're running tests on the server side that feed flags in the `config` object (e.g. `hasInlineMerch`) which give us a pretty good hint if a process needs to happen or not.

gtrufitt:
Cheers @regiskuckaertz. Yeah, that's my concern, I don't want to shove a load of `canRun` checks for different modules into standard js but as much as we can learn about what will happen in the page that can be passed from the server the better. I shall have a  think too, I have some ideas and I'm not sure how much of a major issue it is right now, but I do remember you mentioning a similar problem at one point.

I'll probably merge this as-is if someone can +1 me and then we can have a think and I can refactor.

GHaberis:
:+1: looks good to me @gtrufitt

gtrufitt:
Thanks @GHaberis. I am going to merge this out for now, but really interested to hear the overall feelings about this when you guys have time, feel free to carry on the discussion in here - or maybe I should open an issue.

PR #8102, opened 2015-01-30
Localstorage access denied solved

uplne:
This PR is solving issue when localStorage is not accessible. It can happen on the rare occasions when LocalFlow folder in Windows doesn't have proper permissions. It will break JS execution so the the images won't load.

Before:
screen shot 2015-01-29 at 11 37 09

After:
screen shot 2015-01-29 at 12 13 16

CC @ironsidevsquincy

PR #12983, opened 2016-05-19
Ad-free experience for new users' first 3 days

davidfurey:
## What does this change?

Adds a test disabling ads for the first 3 days a new user visits the site

Hypothesis is that it will reduce churn

no
## Request for comment

https://docs.google.com/document/d/1MsBeGxzllnvX1L1FDaKMhAKIC8ZsHqA3c_Ct_4GDqbQ/edit


common/app/templates/inlineJS/blocking/applyAdfreeRenderCondition.scala.js
```diff
@@ -0,0 +1,119 @@
+@()
```
jbreckmckye: We've stopped using the single-`var` style, because it's a little unwieldy to edit. But that's a minor stylistic point.

common/app/templates/inlineJS/blocking/applyAdfreeRenderCondition.scala.js
```diff
@@ -0,0 +1,119 @@
...
+    if (isFirst3VisitDays() && isInVariantGroup()){
+        commercial.showingAdfree = true;
+        return 'is-adfree';
+    } else {
+        commercial.showingAdfree = false;
+        return '';
+    }
+
+    function isFirst3VisitDays() {
+        var firstVisit = readLocalStorage('gu.firstVisitTime') || 0;
```
jbreckmckye: You should probably refer to your constant again, here.

Also, when will this read return a null? And in that case, is it appropriate to act as though the user's first visit was at the start of the Unix Epoch?

gtrufitt:
😃 😃 😃

common/app/templates/inlineJS/blocking/applyAdfreeRenderCondition.scala.js
```diff
@@ -0,0 +1,119 @@
...
+            }];
+            testVariantIndex = mvtCookieId % variants.length;
+            variants[testVariantIndex].test();
+        } else {
+            writeCookie(AD_FREE_COOKIE, 'notintest');
+        }
+    }
+
+    function readLocalStorage(key) {
+        if ("localStorage" in window) {
```
gtrufitt: We have a [utility for storage](https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/common/utils/storage.js) which also has a method called `isStorageAvailable`

jbreckmckye: Will that be accessible here?

gtrufitt: Although, maybe you've done this so you don't have to pull a util into blocking JS :)

gtrufitt: @jbreckmckye Yeah, realised that after 😞

common/app/templates/inlineJS/blocking/applyAdfreeRenderCondition.scala.js
```diff
@@ -0,0 +1,119 @@
...
+
+    function isInVariantGroup() {
+        var cookie = readCookie(AD_FREE_COOKIE);
+        if (cookie == null) {
+            assignUser();
+        }
+        return readCookie(AD_FREE_COOKIE) == "variant";
+    }
+
+    function assignUser() {
```
jbreckmckye: It's annoying that we have to copy the A/B scripts into our preflight code, but I can't see an alternative, because we have to choose what the user will see before their first pageview renders: if the user is subject to the test, we must not show them any ads; if the user is not subject to the test, we cannot default them to seeing no adverts without compromising our revenue.

I think this is fair enough for an A/B test - it's not that much code, after all - but I think we'll want to take another look at this if and when this experiment graduates into something long-term.

davidfurey: I completely agree.

common/app/templates/inlineJS/blocking/applyAdfreeRenderCondition.scala.js
davidfurey: Thanks, I'll fix the constant.

The gu.firstVisitTime is a new cookie, so won't exist for non-new users.  This test is only interested in users who are new so using unix epoch instead of unknown is fine.

common/app/templates/inlineJS/blocking/applyAdfreeRenderCondition.scala.js
regiskuckaertz: All the modules required here are loaded in standard _before_ commercial, so there is no reason why you couldn't disable commercial at this stage. All you need to do is [switch it off](https://github.com/guardian/frontend/blob/master/static/src/javascripts/boot.js#L51).

kenlim:
I've had a chat with @davidfurey and since we are not going to promote this "feature" at all, and since it has approval from a product manager, and the test is double switched, I'm ok with it running. 

👍

davidfurey:
This PR will now just test that the sampling is working as expected.  Subsequent PR will enable the ad-free behaviour.

@jbreckmckye or @gtrufitt - can you have another look please?

static/src/javascripts/projects/common/modules/experiments/tests/new-user-adverts-disabled.js
```diff
@@ -30,14 +30,14 @@ define([
...
             test: function () {
-                cookies.add('gu_adfree_test', 'variant');
+                cookies.add('gu_adfree_test', 'sampletest-variant');
             }
         }, {
-            id: 'control',
+            id: 'sample-testcontrol',
             test: function () {
-                cookies.add('gu_adfree_test', 'control');
+                cookies.add('gu_adfree_test', 'sample-testcontrol');
```
jbreckmckye: Not `sampletest-control`?

jbreckmckye:
👍

static/src/javascripts/projects/common/modules/experiments/tests/new-user-adverts-disabled.js
davidfurey: Thanks

common/app/templates/inlineJS/blocking/applyAdfreeRenderCondition.scala.js
```diff
@@ -0,0 +1,133 @@
...
+        } else {
+            writeCookie(AD_FREE_COOKIE, 'notintest');
+        }
+    }
+
+    function readLocalStorage(key) {
+        if ("localStorage" in window) {
+            var data;
+            var dataParsed;
+            data = localStorage.getItem(key);
```
piuccio: I know this is already merged, but `"localStorage" in window` is not enough to guarantee that calls to `setItem` and `getItem` work without throwing. See for instance #8102 
If you can't use the utility storage I'd advise to copy paste the isStorageAvailable method or wrap everything in try / catch

jbreckmckye: This might also be an issue in OS X Safari - in incognito mode, some versions will report the presence of localStorage, but throw an exception when you try and write. There's also the marginal possibility of someone's localStorage filling up completely, in which case, any write will also throw an error. Unfortunately, localStorage errors aren't standardized, so you may have to resort to _pokemon exception handling_ - 'cos you gotta catch em all.

PR #12984, opened 2016-05-19
Hosted content - images and copy updates

uplne:
Just small copy, images and elements positions update

@regiskuckaertz @lps88


lps88:
👍

PR #12989, opened 2016-05-19
Labs v media

zeftilldeath:
Labs pages that also contain the media tone have some pretty weird styling. Some of this was due to new galleries and some elements were inheriting whatever tone the page was underneath. 

I have removed new gallery sharing tones and margins from labs pages. Also standardised some elements like save for later. These are a few changes and many more _could_ have been made if I had the time. 

Before:
screen shot 2016-05-19 at 16 08 53

After:
screen shot 2016-05-19 at 16 15 59

### DIFF
--- a/applications/app/views/fragments/galleryBody.scala.html
+++ b/applications/app/views/fragments/galleryBody.scala.html
@@ -4,7 +4,7 @@
 @import views.support.TrailCssClasses.toneClass
 @import views.support.{ImgSrc, RenderClasses}
 
-<div class="l-side-margins l-side-margins--media">
+<div class="l-side-margins @if(!page.gallery.commercial.isAdvertisementFeature){l-side-margins--media}">
 
     <article id="article" class="@RenderClasses(
             Map("content--advertisement-feature" -> page.gallery.commercial.isAdvertisementFeature,
--- a/applications/app/views/fragments/mediaBody.scala.html
+++ b/applications/app/views/fragments/mediaBody.scala.html
@@ -6,7 +6,7 @@
 @import views.support.{RenderClasses, SeoOptimisedContentImage, StripHtmlTags, Video640}
 
 @defining(page.media, page.media match { case _: Audio => "audio" case _ => "video" }){ case (media, mediaType) =>
-<div class="l-side-margins l-side-margins--media">
+<div class="l-side-margins @if(!page.media.commercial.isAdvertisementFeature){l-side-margins--media}">
 
     <article id="article" class="@RenderClasses(
             Map(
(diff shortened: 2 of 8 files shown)


applications/app/views/fragments/galleryBody.scala.html
```diff
@@ -4,7 +4,7 @@
 @import views.support.TrailCssClasses.toneClass
 @import views.support.{ImgSrc, RenderClasses}
 
-<div class="l-side-margins l-side-margins--media">
+<div class="l-side-margins @if(!page.gallery.commercial.isAdvertisementFeature){l-side-margins--media}">
```
zeftilldeath: Will be looking at this properly with Long John Longden @blongden73.

applications/app/views/fragments/mediaBody.scala.html
```diff
@@ -6,7 +6,7 @@
 @import views.support.{RenderClasses, SeoOptimisedContentImage, StripHtmlTags, Video640}
 
 @defining(page.media, page.media match { case _: Audio => "audio" case _ => "video" }){ case (media, mediaType) =>
-<div class="l-side-margins l-side-margins--media">
+<div class="l-side-margins @if(!page.media.commercial.isAdvertisementFeature){l-side-margins--media}">
```
jbreckmckye: Something is clearly going wrong if a class that ostensibly just fixes margins for media breaks GLabs skins.

zeftilldeath: Yeah, same with my css changes. There should really be a glabs media tone created separately, rather than loads of overrides, as glabs media pages are a different layout and completely different colour scheme to media.

static/src/stylesheets/module/_adslot.scss
```diff
@@ -454,35 +454,38 @@
...
+            @include mq(desktop) {
+                top: 100%;
+            }
 
-        @include mq(wide) {
-            left: -240px;
-        }
-    }
+            @include mq(leftCol) {
+                left: -160px;
```
sammorrisdesign: I know you're only indenting here. But shouldn't these all be like  `-(gs-span(2) + $gs-baseline)`

static/src/stylesheets/module/_adslot.scss
```diff
@@ -454,35 +454,38 @@
...
+            }
 
+            @include mq(wide) {
+                left: -240px;
```
sammorrisdesign: Shouldn't these reference `-(gs-span(3) + $gs-gutter)`?

I know you're only indenting but while you're here...

PR #12990, opened 2016-05-19
Test hiding comments

NathanielBennett:
## What does this change?

This is to test the value of comments to the business. On roughly half of content ( and front items ) we'll hide comment, and comment counts. Initially this will be 0% so we can verify that the tracking works correctly

We want to find out how valuable comments are to us. This enables us to determine whether being able to see comments effects the number of articles users read and the length of  time they spend of them

No
## Screenshots

I can make a screenshot of what an article looks like without comments if anyone wants


static/src/javascripts/projects/common/modules/experiments/tests/participation-hide-half-of-comments.js
```diff
@@ -0,0 +1,91 @@
...
+            this.audience = 0.0;
+            this.audienceOffset = 0.0;
+            this.successMeasure = 'We want to guage how valuable comments actually are to us';
+            this.audienceCriteria = 'All users';
+            this.dataLinkNames = '';
+            this.idealOutcome = 'DO we want to turn comments up or down';
+
+            this.canRun = function () {
+                var testAuthor = config.page.author || '';
+                var canRunOnBlog = doesNotContain(blogIds, config.page.blogIds || '');
```
gtrufitt: Sorry to be a pain, I mentioned this on Slack a couple of days ago but I think this should almost definitely be opt-in rather than opt-out because otherwise we're going to almost definitely going to accidentally hide comments where we don't want to.

common/app/views/fragments/contentMeta.scala.html
```diff
@@ -42,7 +42,7 @@
             <div class="meta__numbers modern-visible">
                 <div class="u-h meta__number js-sharecount">
                 </div>
-                <div class="u-h meta__number" data-discussion-id="@item.content.discussionId" data-commentcount-format="content" data-discussion-closed="@{
+                <div class="u-h meta__number js-commentcount" data-discussion-id="@item.content.discussionId" data-commentcount-format="content" data-discussion-closed="@{
```
gtrufitt: Minor but usually hyphenate between words in classes `comment-count`

static/src/javascripts/projects/common/modules/discussion/comment-count.js
```diff
@@ -68,10 +72,15 @@ define([
                 var format,
                     $node = bonzo(node),
                     url = $node.attr('data-discussion-url') || getContentUrl(node),
+                    shortUrl = $node.attr('data-loyalty-short-url').replace('/p/','') || '',
                     $container,
                     meta,
                     html;
 
+                if (shortUrl && ab.isInVariant('ParticipationHideHalfOfComments', 'hide-comments') &&  CommentBlocker.hideComments(shortUrl) ) {
```
gtrufitt: We shouldn't have checks outside of the ab test `canRun` function that change whether the ab test variant functionality is fired - ie: the check for `shortUrl` will modify whether we run the behaviour in the variant and will therefore skew the data so that users who are in the variant (because `canRun` has passed) will still be shown the comments.

You should move or duplicate the check for `shortUrl` into the `canRun` function so that **all** users in the variant will have hidden comments.

static/src/javascripts/projects/common/modules/experiments/tests/participation-hide-half-of-comments.js
```diff
@@ -0,0 +1,91 @@
...
+
+            this.variants = [
+                {
+                    id: 'hide-comments',
+                    test: function(){
+                        var isContent = !/Network Front|Section|Tag/.test(config.page.contentType),
+                            shortUrlSlug = (config.page.shortUrl || '').replace('http://gu.com/p/', ''),
+                            hide = CommentBlocker.hideComments(shortUrlSlug);
+
+                        if(isContent && hide) {
```
gtrufitt: We should move the `isContent` and `hideComments` logic into the canRun so we don't skew the test data. The variant should literally only make the changes

gtrufitt: Actually, sorry, `isContent` should be in `canRun` but obviously the hide comments needs to be in here - it should be that every user that lands in here has half the comments turned off though (as you have).

That way your segmentation is exactly defined by people you *know* only saw half comments and then the control group that had no changes

static/src/javascripts/projects/common/modules/experiments/tests/utils/comment-blocker.js
```diff
@@ -0,0 +1,21 @@
...
+], function(
+    config,
+    reduce
+) {
+
+    var CommentBlocker = {};
+
+    CommentBlocker.hideComments = function(shortUrlSlug) {
+
+        var sUrlInt = shortUrlSlug !== null ? reduce(shortUrlSlug.split(''), function(sum, ch) {
```
gtrufitt: This makes me crosseyed haha. Interesting check! Maybe we could hook in a custom event somewhere and see what the actual difference between shown and not shown was in the end - not important for this test though!

static/src/javascripts/projects/common/modules/experiments/tests/participation-hide-half-of-comments.js
```diff
@@ -0,0 +1,91 @@
...
+            this.idealOutcome = 'DO we want to turn comments up or down';
+
+            this.canRun = function () {
+                var testAuthor = config.page.author || '';
+                var canRunOnBlog = doesNotContain(blogIds, config.page.blogIds || '');
+                var canRunOnSeries = doesNotContain(seriesIds, config.page.seriesId || '');
+                return testAuthor !== dontRunOnAuthor && canRunOnBlog && canRunOnSeries;
+            };
+
+            this.isContent = !/Network Front|Section|Tag/.test(config.page.contentType);
```
gtrufitt: You might need to check it does exactly what you need but the `config.page` object has a `isContent` property

static/src/javascripts/projects/common/modules/discussion/comment-count.js
NathanielBennett: @gtrufitt  Yeah. That's not possible in this case. This is the code that renders comment counts on container and front items and it runs way after the ab-test has run - injecting each of the comment counts into the dom. I guess we could use mediator and remove the  counts from the selected shortUrls, once commentcouints.js has run. This would mean that the user saw rendered comment counts and then saw some of them disappear :( Surely it's a better solution if the comment counts which are to be hidden don't appear and then disappear ..  You can't duplicate the shortUrl test in the canrun function because in this case, you're checking the shortUrl of each item in a container, or front. Come chat to me if this doesn't make sense

static/src/javascripts/projects/common/modules/experiments/tests/participation-hide-half-of-comments.js
NathanielBennett: Yeah I knew that

static/src/javascripts/projects/common/modules/experiments/tests/utils/comment-blocker.js
NathanielBennett: @gtrufitt Yeah a played with various fronts for a few hours and it loooked to come out about even ...

static/src/javascripts/projects/common/modules/discussion/comment-count.js
gtrufitt: Ah ok. That makes sense yeah, thanks for explaining, that's fine then as we're not skewing any data :)

common/app/views/fragments/contentMeta.scala.html
NathanielBennett: @gtrufitt fine. Although checkout 'js-sharecount' a couple lines above  :)

static/src/javascripts/projects/common/modules/experiments/tests/utils/comment-blocker.js
gtrufitt: Haha, nice

static/src/javascripts/projects/common/modules/experiments/tests/participation-hide-half-of-comments.js
gtrufitt: 😉

common/app/views/fragments/contentMeta.scala.html
gtrufitt: Heh, yay for consistency!

gtrufitt:
This is a super interesting test! A couple of comments around making sure we're getting the right data but looks great, nice one! 👍

common/app/views/fragments/contentMeta.scala.html
NathanielBennett: I'm leaving it for consistency

static/src/javascripts/projects/common/modules/experiments/tests/participation-hide-half-of-comments.js
NathanielBennett: @gtrufitt Annoyingly, it the isContent property is undefined on fronts. I think this is why save-for-later does the same thing. https://github.com/guardian/frontend/blob/2e00099b6509024fd5a9f04aa7daea03e08281ac/static/src/javascripts/projects/common/modules/save-for-later.js#L62

jamesgorrie:
@NathanielBennett Have we run a baseline on how much data we would need?

I know @dominickendrick and @harrysalmon have been trying to get this into the process as it makes analysing the data a lot easier and accurate.

gtrufitt:
@jamesgorrie I think Jack P in the data team is crunching the numbers for us to work out what we would need for statistical significance

static/src/javascripts/projects/common/modules/experiments/tests/participation-hide-half-of-comments.js
gtrufitt: What are you trying to match here then? If it's undefined on fronts, that's fine, you know it's not content if it's undefined then right?

static/src/javascripts/projects/common/modules/experiments/tests/participation-hide-half-of-comments.js
gtrufitt: As we discussed, I guess we leave this as opt-out and see how we get on - just gotta make sure central prod, user help and certain editors are aware when and where it's running

static/src/javascripts/projects/common/modules/experiments/tests/participation-hide-half-of-comments.js
NathanielBennett: OK, I'm awake this week. undefined == false. God I hate javascript. Can I write some scala this week, please?

PR #12991, opened 2016-05-19
Hosted by - fixes

lps88:
## What does this change?

Just css changes and open thrasher link in same tab.


uplne:
Awesome 👍

uplne:
There is some switch thing going on, so you need to get the latest master and merge it here. That should fix it

PR #12993, opened 2016-05-19
Update hosted campaign metadata

kelvin-chappell:
Using static values for this test.


uplne:
Magic 👍 It makes me want to share it :-)

PR #12994, opened 2016-05-19
add mediaId to omniture video tracking

jamesgorrie:
## What does this change?

Adding the mediaId to Omniture media tracking to correspond with Ophan.

This was we can tell where a piece of content was played (`pageId`) and a unique ID to the content (`mediaId`).

We can measure a piece of contents performance across all the places it can be embedded.

No, but we will have to see what they are doing and get them to correspond.
## Screenshots

omniture-media-id
## Request for comment

@akash1810 @mkopka


akash1810:
🙌 👍

PR #12995, opened 2016-05-19
Add opengraph og:video:secure_url if url is https

TBonnin:
## What does this change?

Adding og:video:secure_url if content url is https
## Request for comment

@JustinPinner


JustinPinner:
LGTM 👍

PR #12949, opened 2016-05-17
Branded images on Facebook shares

katebee:
## What does this change?

This adds an optional AB test that modifies the Facebook share url on all Facebook buttons on Guardian articles. Users with the modified url will post a different image to Facebook; one with the Guardian logo overlaid (see screenshots).
## What is the value of this and can you measure success?

This currently has zero customer participation to allow testing.
Hypotheses to be tested are that the branded overlay will improve brand recognition, click-through, and revisits.
Will track Facebook referrals to the modified url, and seven day returns for these users.
## Does this affect other platforms - Amp, Apps, etc?

This affects organic shares to Facebook.
Potential to affect any other sites that obtain Open Graph metadata from Guardian pages (Telegram messenger, Whatsapp?)
## Screenshots

**BEFORE**
before: normal facebook share

**AFTER**
after: facebook share with branded overlay
## Request for comment

@rich-nguyen @sndrs @gklopper

PR #12996, opened 2016-05-19
Modify the canonical URL that will be scraped by Open Graph

katebee:
## What does this change?

This currently does nothing observable since tests are all inactive. See the [docs](https://docs.google.com/document/d/1q4BCIsvze7tQ3OMENFnfQGkPbJuVqRJ856FON5_qKgo/edit) if you want to try modifying your Facebook share button on the Guardian.
Follow up to #12949. This adds a second modification to the overrides, and changes the canonical URL (fetched by the crawler and stored in Facebook's database).

The `og:url` is a unique key in the Facebook database to which all of the other Open Graph metadata is associated. To be able to store different images for our (Control vs. Variant) Facebook posts, we need to create separate entries. 

This will affect organic, article shares to Facebook, but should only affect Facebook posts generated by the share button on the Guardian webpage.
Storing separate, canonical URLs in the Facebook database will mean that the Facebook share count will be divided between the variants.
## Screenshots
#12949
## Request for comment

@dominickendrick


common/app/facebookimages/openGraphOverlay.scala
```diff
@@ -11,7 +11,8 @@ object ArticleWithOpenGraphOverlay {
         metadata = article.content.metadata.copy(
           opengraphPropertiesOverrides = {
             val openGraph = article.content.metadata.opengraphProperties
-            openGraph + ("og:image" -> ImgSrc(article.content.rawOpenGraphImage, FacebookOpenGraphImage, true))
+            openGraph ++ Map("og:image" -> ImgSrc(article.content.rawOpenGraphImage, FacebookOpenGraphImage, true),
+              "og:url" -> s"${openGraph.getOrElse("og:url", "")}?page=facebookOverlayVariant")
```
gtrufitt: Do you not also need the control variant so that you can differentiate people returning from the control vs notintest?

dominickendrick: We are ok with serving the canonical meta data to facebook for the control, so we only need a unique url for the variant.

dominickendrick: We only need to change the canonical url for the variant, as the control group gets served the default image. The tracking is done using the `page` parameter, but in a different place

dominickendrick:
👍

PR #12997, opened 2016-05-19
controls attribute is boolean

akash1810:
## What does this change?

The presence of the attribute makes the value true, so we don't need to set it to anything, its presence is enough.

More valid html. Also makes it easier for me to debug a feature I'm building.

No.
## Screenshots

n/a
## Request for comment

@jamesgorrie


jamesgorrie:
Strangely got some breaking football tests

jamesgorrie:
👍

PR #12998, opened 2016-05-20
Fix results page title for football team results

TBonnin:
## What does this change?

Fix the results page title for football team results

Currently the page title (displayed in the browser tab) displays `Some(leicestercity) results | Football ...`
This patch change the page title to `Leicester results | Football ...`

Ex: https://www.theguardian.com/football/leicestercity/results
## Does this affect other platforms - Amp, Apps, etc..

No
## Request for comment

@JustinPinner

### DIFF
--- a/sport/app/football/controllers/ResultsController.scala
+++ b/sport/app/football/controllers/ResultsController.scala
@@ -37,7 +37,7 @@ object ResultsController extends MatchListController with CompetitionResultFilte
   private def page(tag: Option[String] = None): Option[FootballPage] = {
     def allPage = new FootballPage("football/results", "football", "All results", "GFE:Football:automatic:results")
     def competitionPage = (competition: Competition) => new FootballPage(s"${competition.url}/results", "football", s"${competition.fullName} results", "GFE:Football:automatic:competition results")
-    def teamPage = (team: FootballTeam) => new FootballPage(s"/football/$tag/results", "football", s"${tag} results", "GFE:Football:automatic:team results")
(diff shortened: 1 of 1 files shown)


sport/app/football/controllers/ResultsController.scala
```diff
@@ -37,7 +37,7 @@ object ResultsController extends MatchListController with CompetitionResultFilte
   private def page(tag: Option[String] = None): Option[FootballPage] = {
     def allPage = new FootballPage("football/results", "football", "All results", "GFE:Football:automatic:results")
     def competitionPage = (competition: Competition) => new FootballPage(s"${competition.url}/results", "football", s"${competition.fullName} results", "GFE:Football:automatic:competition results")
-    def teamPage = (team: FootballTeam) => new FootballPage(s"/football/$tag/results", "football", s"${tag} results", "GFE:Football:automatic:team results")
+    def teamPage = (team: FootballTeam) => new FootballPage(s"/football/${tag.getOrElse("")}/results", "football", s"${team.name} results", "GFE:Football:automatic:team results")
```
JustinPinner: What's the significance of not having a tag? That would produce a parameter to `FootballPage` that looks like `/football//results`. Maybe if we don't have a tag we shouldn't render the page?

TBonnin: In reality if you don't have a tag it would not be a team page results (or competition) so this code won't be executed. I guess that's the inconvenient in defining the logic before assessing what kind of results it would be

JustinPinner: Ok. +1 on its way.

JustinPinner:
👍

PR #12999, opened 2016-05-20
Fabric slice adverts should support 300x250 MPUs

jbreckmckye:
This is becoming a bit of a saga, now.

We want the Fabric slots we insert into fronts (on the first container) to support 300x250 MPUs. I've patched `slice-adverts` to use the `inline1-fabric` ad type, which handles both these sizes. It also uses the 'inline1' name, which means the slot should integrate with any existing passback behaviour.

The real solution to this problem, however, is to change `dfp-api` so that sizes specified for mobile breakpoints are not automatically reused in larger configurations. If that happens, we can remove the 'inline1-fabric' adtype entirely, and just specify 88x71 as a mobile size for `inline1`. This is something I'm doing in another branch.


kelvin-chappell:
👍

PR #13002, opened 2016-05-20
OpenGraph: pass video file info into the og:video parameters

TBonnin:
## What does this change?

Video shared on FB don't play currently. 
See https://www.facebook.com/annemarie.conlon.75?pnref=story  and try to play the video

So far we were passing the page url into og:video
This patch passes the video asset/file url in the og:video as well as adding og:video:width,
og:video:height and og:video:type as recommended in opengraph documentation https://developers.facebook.com/docs/sharing/webmasters#markup

Video shared on FB are actually playable

No, it only affects video shared on FB
## Request for comment

@JustinPinner @katebee

### DIFF
--- a/common/app/model/content.scala
+++ b/common/app/model/content.scala
@@ -514,21 +514,27 @@ object Video {
     val elements = content.elements
     val section = content.metadata.section
     val id = content.metadata.id
-    val source: Option[String] = elements.videos.find(_.properties.isMain).flatMap(_.videos.source)
+    val mainVideo: Option[VideoElement] = elements.videos.find(_.properties.isMain)
+    val source: Option[String] = mainVideo.flatMap(_.videos.source)
 
     val javascriptConfig: Map[String, JsValue] = Map(
       "isPodcast" -> JsBoolean(content.tags.isPodcast),
       "source" -> JsString(source.getOrElse("")),
-      "embeddable" -> JsBoolean(elements.videos.find(_.properties.isMain).map(_.videos.embeddable).getOrElse(false)),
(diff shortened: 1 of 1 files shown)


common/app/model/content.scala
```diff
@@ -514,21 +514,27 @@ object Video {
...
-    val opengraphProperties = Map(
-      "og:type" -> "video",
-      "og:video:type" -> "text/html",
-      "og:video" -> content.metadata.webUrl,
-      "video:tag" -> content.tags.keywords.map(_.name).mkString(",")
-    ) ++ optionalOpengraphProperties
+      "embeddable" -> JsBoolean(mainVideo.map(_.videos.embeddable).getOrElse(false)),
+      "videoDuration" -> mainVideo.map{ v => JsNumber(v.videos.duration)}.getOrElse(JsNull))
+
+    val opengraphProperties = mainVideo.flatMap { _.videos.largestVideo }.map { videoAsset =>
```
TBonnin: That would share the largest video. We can probably find a better suited size to share on FB but I want to first make sure this is the correct solution before to introduce any complexity

JustinPinner: Agreed. Don't want to force massive downloads/streams if we don't need to.

JustinPinner:
Nice. Let's share some vids! 👍

common/app/model/content.scala
jamesgorrie: Are we sure that they don't proxy it?
A cursory glance says no, but if they do, we won't have to worry.

TBonnin: Good point @jamesgorrie . I am not sure. I will verify this.

PR #13003, opened 2016-05-20
add test to show controls on main media video

jamesgorrie:
## What does this change?

Adding the test to show media controls on main media video.

Either people will notice this more and watch more video, in which case success = success - or they won't, in which case failure = success.

No
## Screenshots

definitely-a-video
## Request for comment

@akash1810 @harrysalmon


ScottPainterGNM:
Nice!

akash1810:
bit sad the visibility of the controls are controlled by css rather than at the time the videojs is instantiated - but as its a test 👍

jamesgorrie:
You can definitely control via JS. I might have a go if it proves profitable.

harrysalmon:
Here's the SQL for analysis once the test is finished https://github.com/guardian/data-science/blob/master/mvt/video_starts_on_main_media.sql

PR #13004, opened 2016-05-20
Enchilada ads for hosted content updates

uplne:
Again. Fixing some stylings. The bigger change is that we are hiding label for now. There needs to be some updates but that is low priority for now so the quick hack is to just hide it. quick hacks ftw. what can go wrong.

screen shot 2016-05-20 at 12 32 17

@regiskuckaertz @lps88

### DIFF
--- a/static/src/stylesheets/module/commercial/glabs/_hosted.scss
+++ b/static/src/stylesheets/module/commercial/glabs/_hosted.scss
@@ -518,7 +518,7 @@ $hosted-stripe: 28px;
         bottom: 0;
         position: absolute;
         background-image: linear-gradient(to bottom, rgba(238, 238, 238, 0), rgba(0, 0, 0, .4));
-        border-top: 2px solid transparent;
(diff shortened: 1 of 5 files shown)


regiskuckaertz:
> what can do wrong

bad ass style, bro!

giphy

uplne:
slides-56

static/src/stylesheets/module/commercial/glabs/_hosted.scss
```diff
@@ -549,8 +549,8 @@ $hosted-stripe: 28px;
         right: 0;
         bottom: 0;
         position: absolute;
-        background-image: linear-gradient(to bottom, rgba(0, 0, 0, 0), rgba(0, 0, 0, .4));
-        border-top: 2px solid transparent;
```
lps88: gradient should be linear-gradient(to bottom, rgba(0, 0, 0, 0), rgba(0, 0, 0, .4)) (merge error?)

PR #13005, opened 2016-05-20
Faster boot time for commercials

regiskuckaertz:
cc @jbreckmckye @uplne @sndrs @gtrufitt 

We have a big problem. On an empty cache, commercials load very, very late. The GPT library necessary to make any ad request is only loaded around the 4s mark:

picture 3

Still after the load event on a primed cache:

picture 4

This delay is noticeable when you browse the site and focus on the ads (admittedly, nobody but us do). The problem is that this delay has a very tangible effect our the bottom line—which we are finally able to measure thanks to @JonNorman's and @kenlim's project Maguire.

Using performance timing on my local machine, I found that:
- js kicks in around the **1.5s** mark
- standard js starts **1.5s** after that and takes only 5–10ms to run
- commercial js starts **1.5s** after that and runs for **~400ms** 😭 

Even if we take away the cost of doing network requests in dev mode, it's still pretty obvious we have some contention points here:
- the time necessary to parse the DOM and run blocking JS is huge
- the time necessary to load and parse all the standard and commercial JS dependencies are huge
- commercial is f-ing slow

We have to do something about it, and this PR is the first attempt in a series at doing so. Essentially I think along those lines:
- html/js is optimized for current browsers first
- older browsers will incur additional http requests to get the polyfills/helpers they need, depending on which js they run (standard vs enhanced) — e.g. es5-shim for IE8

Some obvious culprits that I can think of:
- a lot of standard code is written with compatibility with IE8 in mind: polyfills and helpers that are not needed for most browsers. Worth noting that [only IE11 appears in the list of top browsers](https://frontend.gutools.co.uk/analytics/quality) that hit theguardian.com... We should definitely consider updating our "cut the mustard" policy.
- many lodash helpers have an enormous amount of dependencies. Even in an optimized build, that's a lot of code. Now, lodash is an open source library meant for rookie js devs, so it has lots of bells and whistles that _we don't need_. We can do better.
- some standard modules are really huge because of careless coding. e.g. the `survey-simple` AB test required the whole `svgs` module; that's 44% of boot.js. others included modules that were not even used.

picture 1

I have gone through all the modules loaded in standard and removed as much as I could (mostly `bonzo`, and `lodash` collections functions). That's 150k worth of work removed:

picture 2

This is obviously a lot in a single PR and I can of course split this into more manageable chunks. But this is a start; I also plan to spend time on
- doing the same in the commercial js
- precompiling lodash templates (still a big, unnecessary part of boot.js)
- optimising the commercial code

I also think we really should update our browser compatibility strategy, have lines in the sand for standard modules, and decide on acceptable coding patterns. @ScottPainterGNM is already thinking about this; ultimately that is a discussion we should all be part of.


uplne:
So basically you are 100% confident with native forEach etc right

jbreckmckye:
@uplne We don't run any of the commercial scripts in non-enhanced browsers (e.g. IE8) any longer, so forEach and friends should work just fine. If not, we'll spot any errors in Sentry pretty quickly.

static/src/javascripts/projects/common/utils/config.js
```diff
@@ -26,6 +21,17 @@ define([
         return Math.floor(Math.random() * 36).toString(36);
     })};
 
+    function assign(dest, source) {
```
regiskuckaertz: Should probably be in a separate polyfill

gtrufitt:
Cool! I am certainly guilty of using the libraries instead of the native methods where I have no reason to and we can definitely do better.

If we remove these helper modules/libraries they obv need to be removed everywhere to have the effect we want but the problem I have with the pure-JS approach is having to do things like this:

```
 var body = document.getElementsByTagName('body')[0];
                         if ('classList' in body) {
                             body.classList.remove('has-membership-access-requirement');
                         } else {
                             body.className = body.className.replace('has-membership-access-requirement', '');
                        }
```

It just feels kinda error prone and though the majority of us are perfectly capable of remembering to do this, this codebase has lots of different developers working on it.

I think we need to think of *how* we tackle this - can we add some JSHint rules against the use of things like `$`, `removeClass` etc and add helpful warning/error messages that suggest the approach? Prevent the use of lodash in standard scripts?

That said the `classList` example here is moot if we move IE9 out of modern browsers and there's not a whole load of other areas where that complexity rears its head - maybe the [assign](https://github.com/guardian/frontend/pull/13005/files#diff-7b9f143f8e033237f74cd5988d18713dR24) method. Maybe we could move these into our own utils instead.

Anyway, this is great work, looking forward to seeing how much we can do going forward!

@GHaberis - You've been thinking about the same thing in the apps too, right? You've removed your micro-libraries?

static/src/javascripts/projects/common/utils/config.js
gtrufitt: Ha, you totally preempted my comment there ;)

gtrufitt:
Actually, the `classLIst` example isn't moot is it because it's in main.js

regiskuckaertz:
@gtrufitt definitely agree with all you said. That is why I think there should be specific lines in the sand for standard vs enhanced. In standard, there should be an emphasis on being as close as possible to browser engines in order to reduce overhead and increase perfs. I believe we can avoid the kind of coding you outlined 99% of the time, but yeah this excerpt exemplifies the sacrifice we'd have to do _sometimes_. If IE9 is out of the equation, we can add a `classList` polyfill and using the standard API ... (this is the only reason we're still using `bonzo` anyway).

Pragmatically speaking, a line should be drawn somewhere as to when it is ok to use a microlib in standard. I believe this is where the emphasis on recent browsers can help us move the needle in the right direction.

GHaberis:
@gtrufitt I've wanted to remove micro-libraries within the app templates, but haven't gone all the way yet. I do remove them as and when I touch code where I deem them unnecessary.

I've wanted to remove them for a few reasons. Firstly for performance gains. In previous roles I've seen load times improve in hybrid apps by keeping templates as lightweight as possible by removing libraries and their dependencies. Secondly libraries are used everywhere in situations were vanilla native JS could've done the same without any additional overhead. I also happen to have come from an environment where native JS was strongly encouraged above libraries, so I'm happy using native js. However I've held back removing libraries for a few reasons...

- I don't intend to be the sole contributor to the codebase, therefore I want others input into whether removing qwery, bean, bonzo etc is something they'd be happy with if they have to start contributing. Especially as every other project I've seen here uses these libraries.
- I'd like to actually prove performance gains can be made by removing these libraries, however because they're used by virtually every module I'd need to make the changes across the entire codebase to see these gains.

On the flipside I can see the benefits of using a utility library like lodash and being able to code in a more declarative/functional way. Basically I'm at a point where I know there are certain libraries I definitely want to remove but I'm unsure about imposing a 'no libaries at all' policy!

regiskuckaertz:
I also think micro libs are very useful (I'm writing [my own](https://github.com/regiskuckaertz/seier)), in that they help write concise and maintainable code that is easier to read for everyone. I just don't think lodash is the right fit for us. Given its appeal as a plug-in library that _just works_ everywhere, it cannot make many assumptions about the data it will be fed or the environment it will run in. That's the trade-off many open source projects must make if they want to spread… As @gtrufitt suggests, we could have our own simile of lodash, optimised for our unique context.

Essentially, standard and enhanced are two diametrically opposed environments: in enhanced, we have much more leeway. Degrading performance (a little) for the sake of readability/maintainability is acceptable–in standard, it is much less so. We could make the parallel with writing a device driver and an enterprise app, the former in C and the latter in ~~Java~~ Scala. I'm counting the days until we can rewrite standard in [wasm](https://webassembly.github.io/) 👯.

gtrufitt:
Aside from the comments raised around polyfills, which you may or may not address in this PR, this looks good 👍

regiskuckaertz:
@gtrufitt thanks 😄  one question I have is, should standard run in IE8? if yes then I just need to include es5-shim somewhere.

sndrs:
sorry i've not given this the time, i will do now, but yes – current idea is that standard js will try to run in all envs, like the inline JS. obviously some things are not supported, but it will always try (i.e. not be guarded by an `isModern`-like flag.

the only difference between it and the inline JS is that inline directly affects rendering (either blocking, which we want to run before render starts, or non-blocking, which we want to runs asap on each page, but not block).

sndrs:
this lgtm, but my god it's a bruiser :) 

hugely applaud the work that's gone into this and the rationale, just a couple of qs:

1. be interested to know how much faster this runs, esp on an older device? _is_ is faster?
2. this definitely needs a response to the issue of standard running everywhere, es5-shim sounds perfectly reasonable to me.
3. this had a _lot_ of changes in it, which makes it hard to review (as you point out). smaller chunks would be useful?

in terms of standard/support etc, totally no problem with some code not running where it can't e.g. localstorage is not supported, but we need to make sure anything outside a baseline check fails gracefully, and also works if it can. i.e. you could use `for` loops in place of `forEach`, `map` etc.

an es5 baseline means ~2% of current users can't run it – it seems reasonable to me that we polyfil in those circumstances, but it's worth noting too that curl will try and load the app on IE6+ – we need to know what we'll do in those circumstances.

👎 till polyfils, but 👏 the effort!

regiskuckaertz:
Thanks Alex!

1. I'll grab comparison metrics on both prod and dev modes asap
2. Check out [this PR](https://github.com/guardian/frontend/pull/13031)
3. Yep I think it's better if I do it that way

regiskuckaertz:
@sndrs In response to 1. I ran tests on the network front and an article both in DEV and PROD and consistently got between 400ms–1000ms improvements in how fast standard kicked in.

I'll close this PR and will release fixes in smaller batches. I think it will need a bit of communication too (e.g. stop using `svgs` in standard).

PR #13007, opened 2016-05-20
update selector for play buttons on fronts

jamesgorrie:
## What does this change?

It has to be more specific to be more accurate...

It's more accurate

Nope

@akash1810


akash1810:
👍

PR #13008, opened 2016-05-20
Changing video og tags to 'og:type = article'

TBonnin:
## What does this change?

James Murray confirmed that video shared on FB used to behave like any other articles. ie: clicking on the thumbnail would open a guardian.com video page.

Possible assumption: Something might have changed on FB side and even though we were specifying it was a video (`og:type =  video`) they were still loading the html we passed in the (`og:video`) property. Maybe they are now more strict about that. ie: it is declared as a video (`og:type = video`) so they try to play the file supposedly passed in the `og:video` property. Since it is a html doc and not a mp4/swf file it fails.

With this patch, I am trying to recreate the previous behaviour by declaring video page to be normal article. 

I don't think we want to play the video directly on Facebook without evaluating the impact it could have on our traffic.

No more blank page when clicking on a shared video on FB

Good question. I am not sure. Does Twitter uses the og tags? This might have more impact than I am aware off. Let me know what do you think
## Request for comment

@JustinPinner @johnduffell

### DIFF
--- a/common/app/model/content.scala
+++ b/common/app/model/content.scala
@@ -524,10 +524,13 @@ object Video {
 
     val optionalOpengraphProperties = if(content.metadata.webUrl.startsWith("https://")) Map("og:video:secure_url" -> content.metadata.webUrl) else Nil
     val opengraphProperties = Map(
-      "og:type" -> "video",
-      "og:video:type" -> "text/html",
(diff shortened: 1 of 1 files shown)


JustinPinner:
👍 Where will it all end?! 😱

johnduffell:
great! thanks for that one!

PR #13009, opened 2016-05-20
Remove fixed top above slot (for now)

regiskuckaertz:
There is an ongoing discussion around getting rid of any 728x90 ad in the top slot and the expanding animation, but nothing concrete yet. Removing this feature for now but keeping the logic for later re-use though.

cc @jbreckmckye @JonNorman


JonNorman:
Until we meet again :+1:

PR #13010, opened 2016-05-20
Instances will stop trying as hard when they're overloaded

johnduffell:
Following on from https://github.com/guardian/frontend/pull/12834

This is a better implementation after testing with JMeter.  I have still avoided hard coding any limits so it should work well for boxes with varying load (facia being the heavy one)  I am a little concerned how diagnostics (low latency, high request count) will respond as I have intentionally not synchronised but that causes race conditions.  However I have a switch in case of issues..

This implementation still reduces the percentage of requests served based on the latency of completed requests.  However this time the weighting of the most recently served request is 1, the weighting of the one before that is 0.99, the one before that is 0.9801, etc etc.  This is to try to reduce oscillations without spending too long to come back into service after a blip.

I have removed the inflightlatency monitor as I don't think it's needed after adding the feature descibed below.

A new feature keeps track of the balance of requests received to responses served.  Again, this is weighted so if we receive more than around 25 requests, the server will wait until it responds to at least one of those before taking on any more.  This avoids the problem of opening the flood gates for a second and then taking on too many requests causing latency to end up at 10 seconds before we realise and causing a slow recovery.

Overall, the aim is to make things better than before (servers accepted all requests, latency easily exceeded 60 seconds, servers never recovered and couldn't serve health checks any more and were killed, then repeat until someone turns off traffic to the load balancer)

@TBonnin @markjamesbutler @cb372 @alexduf any comments would be appreciated


common/app/common/metrics.scala
```diff
@@ -99,6 +99,26 @@ object SystemMetrics extends implicits.Numbers {
     get = () => buildNumber,
     metricUnit = StandardUnit.None
   )
+
+}
+
+object RequestMetrics {
+
+  val PanicRequestsSurgeMetric = CountMetric(
+    name = "requests-shed",
```
alexduf: Is it intentional that your three metrics carry the same name?

johnduffell:
Facia (locally) - graph of successful responses - you can see it struggles and blips in and out as taking on around 25 concurrent requests is too much
image
image

johnduffell:
Article is a lot more stable in the same situation (64 concurrent threads)
image
image

johnduffell:
if I run article from master it doesn't cope as well
image
image

common/app/common/metrics.scala
johnduffell: 😜 whoops

johnduffell:
@mchv any comments would be great!

alexduf:
That looks good to me :+1: 
But more eyes on it would be a good idea as mistake could quickly go wrong here.

common/app/conf/Filters.scala
```diff
@@ -71,6 +75,151 @@ object AmpFilter extends Filter with ExecutionContexts with implicits.Requests {
...
 }
 
+// this turns requests away with 5xx errors if we are too busy
+object PanicSheddingFilter extends Filter with Logging {
+
+  import scala.concurrent.ExecutionContext.Implicits.global
+  import scala.concurrent.duration._
+
+  val ALL_200s_MAX_LATENCY = 1.second.toMillis
+  val PANICING_MIN_LATENCY = 3.seconds.toMillis
```
TBonnin: PANICKING? :D

kelvin-chappell:
Looks good 👍 
But feels like something the ELB should be doing.

johnduffell:
kelvin too true!  or even in nginx would be ok for us.  Hopefully the dream of serverless architecture will come true and we won't need to run our own servers at all 🎉

PR #13011, opened 2016-05-20
Hosted by fixes

lps88:
## What does this change?

fix some bugs ready for Renault release 
## Request for comment

@uplne


static/src/stylesheets/head.content.scss
```diff
@@ -36,6 +37,9 @@
 @import 'module/external/_from-content-api';
 @import 'module/content/_badging';
 
+// Needed for more seamless loading of Hosted By page (Renault release)
```
uplne: Just let's make sure that in the future we will create own head.commercial.scss file which will be loaded only on this hosted pages and not globally. These files needs to be kept minimal.

uplne:
👍

PR #13013, opened 2016-05-20
Hosted content - updated twitter text

uplne:
Yeah


lps88:
👍  better solution will be done soon 😄

PR #13014, opened 2016-05-20
add test for playing videos on fronts

jamesgorrie:
## What does this change?

Adding the test to play videos in article rather than on the front when an article has video main media.

There's a few changes to our main codebase, but just events being emitted to keep the code in the test.

What we are testing is: If we send you to the article and play the video there, are you more likely to get to 25%?

Success is construct.

No
## Screenshots

Not-so-good:
plomp
## Request for comment

@akash1810 @harrysalmon @dominickendrick

### DIFF
--- a/static/src/javascripts/bootstraps/enhanced/media/main.js
+++ b/static/src/javascripts/bootstraps/enhanced/media/main.js
@@ -146,7 +146,9 @@ define([
                         placeholder.removeClass('media__placeholder--active').addClass('media__placeholder--hidden');
                         player.removeClass('media__container--hidden').addClass('media__container--active');
                         $el.removeClass('media__placeholder--active').addClass('media__placeholder--hidden');
-                        enhanceVideo($('video', player).get(0), true);
+                        var enhancedPlayer = enhanceVideo($('video', player).get(0), true);
+
+                        mediator.emit('ab:PlayVideoOnFronts:front-player-created', enhancedPlayer);
                     });
                 });
                 fastdom.write(function () {
@@ -166,7 +168,8 @@ define([
 
         fastdom.read(function () {
             $('.js-gu-media--enhance').each(function (el) {
-                enhanceVideo(el, false, withPreroll);
+                var enhancedPlayer = enhanceVideo(el, false, withPreroll);
+                mediator.emit('ab:PlayVideoOnFronts:in-article-video-created', enhancedPlayer);
             });
         });
 
@@ -351,6 +354,8 @@ define([
                 });
             }
         });
+
+        return player;
     }
 
     function initEndSlate(player, endSlatePath) {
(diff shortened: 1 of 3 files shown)


static/src/javascripts/projects/common/modules/experiments/tests/play-video-on-fronts.js
```diff
@@ -20,17 +27,58 @@ define([
 
         this.canRun = function () {
             // Only videos that are links to video pages have data-embed-paths
-            return config.page.isFront && $('.fc-item--has-video-main-media .js-video-play-button').length > 0;
+            // Or article pages that have been linked to from test-enhanced cards
+            return (config.page.isFront && $('.fc-item--has-video-main-media .js-video-play-button').length > 0) ||
+                (config.page.contentType === 'Article' && hasAutoplayHash);
```
harrysalmon: Adding a second page to the canrun function is going to make this a bit more difficult to analyse. We'd have to include `and publication_date is null` to the query to try to get only fronts. This isn't too bad, but does mean this test won't be analysable in the standard way. Ideally this is moved outside of canrun. @desbo 

I'd go ahead with it, but we'll have to ignore the Ophan pages when looking at sample size

paperboyo:
Interesting. That's a much longer wait... Does it move main media video in view after loading the page automagically?

static/src/javascripts/bootstraps/enhanced/media/main.js
```diff
@@ -166,7 +168,8 @@ define([
 
         fastdom.read(function () {
             $('.js-gu-media--enhance').each(function (el) {
-                enhanceVideo(el, false, withPreroll);
+                var enhancedPlayer = enhanceVideo(el, false, withPreroll);
```
akash1810: `enhanceVideo` doesn't return anything, this will always be `undefined`

jamesgorrie:
> That's a much longer wait

On local

> Does it move main media video in view after loading the page automagically?

 No

static/src/javascripts/projects/common/modules/experiments/tests/play-video-on-fronts.js
jamesgorrie: This would never have been able to have been true as it's from the added functionality i.e. Adding the link to the play button, so it won't really affect the numbers.

static/src/javascripts/bootstraps/enhanced/media/main.js
jamesgorrie: [Say what]()https://github.com/guardian/frontend/pull/13014/files#diff-b226c7adcb4e3819c2d7fd8c6a997f8dR358?

akash1810: ah! 👍

PR #13015, opened 2016-05-20
Get correct id from shortUrl regardless of the scheme

TBonnin:
## What does this change?

Properly get the id from a short url regardless of the scheme

CAPI can start returning short urls with https

No
## Request for comment

@mchv @johnduffell @sndrs

### DIFF
--- a/common/app/model/meta.scala
+++ b/common/app/model/meta.scala
@@ -159,7 +159,14 @@ final case class Fields(
   sensitive: Option[Boolean],
   legallySensitive: Option[Boolean]
 ){
-  def javascriptConfig: Map[String, JsValue] = Map(("shortUrl", JsString(shortUrl)))
+  lazy val shortUrlId = shortUrl.replaceFirst("^[a-zA-Z]+://gu.com", "") //removing scheme://gu.com
+
+  def javascriptConfig: Map[String, JsValue] = {
+    Map(
+      "shortUrl" -> JsString(shortUrl),
+      "shortUrlId" -> JsString(shortUrlId)
+    )
+  }
 }
 
 object MetaData {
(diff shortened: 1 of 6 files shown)


common/app/model/meta.scala
```diff
@@ -159,7 +159,14 @@ final case class Fields(
   sensitive: Option[Boolean],
   legallySensitive: Option[Boolean]
 ){
-  def javascriptConfig: Map[String, JsValue] = Map(("shortUrl", JsString(shortUrl)))
+  lazy val shortUrlId = shortUrl.replaceFirst("^[a-zA-Z]+://gu.com", "") //removing scheme://gu.com
```
TBonnin: Maybe CAPI should return the short Id as a field 😉  😉  😉

identity/app/services/PlaySavedArticlesService.scala
```diff
@@ -35,9 +35,9 @@ class PlaySavedArticlesService @Inject()(api: IdApiClient) extends SafeLogging w
 
     val sanitizedArticles = savedArticles.articles map {
       article =>
-        val id = article.id.replace("http://www.theguardian.com/","")
-        val shortUrl = article.shortUrl.replace("http://gu.com","")
-        SavedArticle(id, shortUrl, article.date, article.read, article.platform)
+        val id = article.id.replaceFirst("^[a-zA-Z]+://www.theguardian.com/","")
+        val shortUrlId = article.shortUrl.replaceFirst("^[a-zA-Z]+://gu.com","")
```
TBonnin: I wish I could centralize the replace in one place but it looks like SavedArticle have their own model

common/app/model/meta.scala
mchv: @TBonnin agreed. I am planning to do this

mchv:
👍

johnduffell:
great good work! 👍

TBonnin:
Added a test https://github.com/guardian/frontend/pull/13015/files#diff-9dc855c91aacd716cf340f723deb3571R132 🎯

PR #13019, opened 2016-05-23
Lower healthcheck timeout

TBonnin:
## What does this change?

Lower the healthcheck timeout from 10 seconds to 4s

Services are fetching healthcheck every 5 seconds.
It doesn't make sense to have timeout to be greater than this value.
If it fails it would be retried soon after

No
## Request for comment

@johnduffell


johnduffell:
great 👍

PR #13021, opened 2016-05-23
Revert "Instances will stop trying as hard when they're overloaded"

johnduffell:
The instances are having latency issues, reverting as a precaution.


johnduffell:
@TBonnin

PR #13023, opened 2016-05-23
fix switch date

NathanielBennett:
## What does this change?

Changes the expirry date of the switch introduced here: https://github.com/guardian/frontend/pull/12990 to match the expiry date of the abtest (also created by 12980)


gidsg:
👍  📅

PR #13026, opened 2016-05-23
Changing image overlay

dominickendrick:
## What does this change?

This changes the image overlay for #12949 

before:
screen shot 2016-05-23 at 15 56 21

after:
screen shot 2016-05-23 at 15 55 54

@zeftilldeath @gtrufitt


gtrufitt:
👍

PR #13029, opened 2016-05-23
Precompile lodash templates

regiskuckaertz:
Pfew! After many failed attempts at making this happen, I'm now confident this is working.

While it worked perfectly from the beginning in DEV, there was a reference error when the module was compiled: `_.escape` was not defined. This call was in the source generated by `template` so [I had to stub it in my loader](https://github.com/guardian/frontend/compare/precompile-lodash-tpls?expand=1#diff-cf46ea90955a09f806da77adb726507bR86).

cc @sndrs @gtrufitt 

Thanks @rich-nguyen for making the `useHashedBundles` flag.


regiskuckaertz:
@sndrs @gtrufitt @jbreckmckye

If you can have a look guys. Sorry for the large amount of changed files, more than 50% is actually the move to lodash 2.4.2, which was necessary to remove the _only_ reference to `window` (this broke the grunt compilation).

static/src/javascripts/bootstraps/enhanced/notifications.js
```diff
@@ -7,13 +7,12 @@ define([
...
     'common/utils/storage',
     'common/utils/ajax',
-    'common/utils/template',
     'common/utils/robust',
     'common/views/svgs',
     'common/modules/user-prefs',
-    'text!common/views/ui/notifications-follow-link.html',
-    'text!common/views/ui/notifications-explainer.html',
-    'text!common/views/ui/notifications-permission-denied-message.html',
+    'tpl!common/views/ui/notifications-follow-link.html',
```
jbreckmckye: Why not `template`?

jbreckmckye:
So how exactly does this work?

static/src/javascripts/bootstraps/enhanced/notifications.js
regiskuckaertz: admittedly I got bored after typing it 400 times 😭

regiskuckaertz:
Exactly like the `text` loader plugin. In DEV mode, it will fetch the html file over the network and return the compiled template. In PROD mode, it inlines the compiled template in a module.

jbreckmckye:
@regiskuckaertz Why does the dev build behave differently to the prod build - does precompiling the templates add a lot of time to the `make` task? I feel prod and dev should really have the same behaviour, at least until there's a compelling reason to distinguish them.

regiskuckaertz:
By that I meant that it will use the overarching mechanism that loads modules. If you use the `useHashedBundles` option, it will behave in DEV as it would in PROD.

gtrufitt:
Just so I make sure I understand this - before we inlined the templates as is using !text and ran them through the template util and now we inline the function returned from lodash template util so we skip running through the template util at runtime and call the compile functions directly?

regiskuckaertz:
@gtrufitt yes, the compilation happens during the build.

static/src/javascripts/bootstraps/enhanced/notifications.js
jbreckmckye: I'm just not certain about these kind of NGW-specific abbreviations. It seems to be adding proprietary knowledge to a codebase that really doesn't need it. It's not a point I'd live or die on, but this is probably not something I would choose myself.

jbreckmckye:
@regiskuckaertz - essentially, then, the precompile step turns each template into an AMD module, which will then be fetched in exactly the same way as any other module in the current environment. Is that correct? That would sound sensible.

static/src/javascripts/bower.json
```diff
@@ -6,7 +6,7 @@
     "domready": "ded/domready#v0.3.0",
     "eventEmitter": "~4.2.6",
     "fastdom": "0.8.5",
-    "lodash-amd": "~2.4.1",
+    "lodash-amd": "regiskuckaertz/lodash-amd#2.4.2",
```
jbreckmckye: Wait, are we using your fork, now?

jbreckmckye:
As I understand it, Regis, the main benefit of this change is that it means `commercial.js` will execute faster when it is fetched by the client. Do we have any figures that help quantify that the script is now faster?

gtrufitt:
Looks good to me 👍 Did you run any speed tests on this (just interested not a requirement for a +1 :D)?

jbreckmckye:
@gtrufitt I don't believe we should pass a performance fix without verifying that things really are faster. Whilst I can't see any way that they wouldn't - I don't think there's any cost to proving it.

gtrufitt:
@jbreckmckye You're right. The 'value' of this PR is shown via the numbers. I'll revert my thumb for a min if you can show us some numbers @regiskuckaertz

regiskuckaertz:
I don't think I need to run tests. A close look at the code in `template` will reveal this function is both time- and space- hungry:

- many regexps. Regular expressions are PSPACE-complete, and we have sometimes fairly large templates. This is compounded by the fact that the automata is fairly complicated.
- many string concatenation. The stack frame of this function is huge just because of that, it's no wonder the GC goes loco.
- slow meta-programming. No matter if we pass data or not, it will build a function, using the `Function` constructor. This is a slow code path and a potential security leak (`eval`'d code has access to the global context).

The hyper cool thing with the loader plugin is that there is no meta-programming happening on the client-side, since the generated source will be inline as a module. An inlined template is therefore just a string concatenation — O(n) in space and time.

jbreckmckye:
Everything you write makes perfect sense, Regis, but there is a non-zero cost to this code - it may have to be maintained - and it would be nice to quantify what we get in return.

For one thing - how much faster is `commercial.js` since we measured its execution time in #13005? Has this made any impact, or is there still much to be done?

jbreckmckye:
I'm going to give this code a 👍 based on its own merits, but I do think it would a shame if we didn't capture the performance improvements. For one thing - that knowledge makes it easier to justify similar refactoring in future.

regiskuckaertz:
You're right, I'll run a few tests when I have time. Re your first point, I don't think there is any maintenance cost, it is more or less a copy of the `text!` loader plugin. Thanks for the +1 😃

regiskuckaertz:
@jbreckmckye hi Jimmy. I ran a sample (small - manual-inline) template though compilation with and without the loader. It runs 30x faster on average, although on a modern browser the time savings won't be noticeable — once can expect the gap to increase polynomially with the length of the template.

static/src/javascripts/bower.json
regiskuckaertz: yes, back then Olly said that was ok - lodash-amd is not maintained anymore.

sndrs:
just to chime in on this v interesting discussion: in my experience, fixing a bugbear can sometimes feel like it's a real improvement, and they in turn can end up just being micro-optimisations (not saying this is one, speaking generally). 

I once suggested we remove all `.bind` calls for `that = this`. they're ugly, but they're massively faster, esp in older browsers. i forget the exact numbers, but on a jsperf test it was something like 30k/sec vs 5m/sec i.e. the worst possible scenario was still 30k/sec. there was just no need to make the change.

again, not saying that is or is not the case here, but I think its important we always understand the value of any change. proof of it's value is never a bad thing

regiskuckaertz:
I do agree with your arguments, guys. It's not like we're running algorithms on huge datasets, the computational complexity doesn't matter that much here. Given the results I'm seeing, I would consider this a micro-optimisation (disappointingly). The only impediment left is everything runs in a single thread and, while the ROI for these changes taken in isolation is negligible (3ms vs 0.1ms), they might have a non-negligible impact taken as a whole — think compound interest.

Happy to close this PR, I've just found [LinkedIn ran some experiments](https://engineering.linkedin.com/frontend/client-side-templating-throwdown-mustache-handlebars-dustjs-and-more) in the past and underscore/lodash performs really well.

jbreckmckye:
Interesting findings, Regis.

static/src/javascripts/bower.json
piuccio: As discussed at lunch, this is the line that worries me most. I see two solutions

1. Fork into `guardian/lodash-amd#2.4.2` at least if you leave, someone else will be able to maintain that fork
1. Use `lodash 4`. It's modular so you can include only the bits you need without wasting too many bytes

the second approach could be used as a starter in the lodash migration. Last time it didn't work because we were changing the whole website at once, which come with a huge risk. Using two versions at the same time we can migrate files at a slower pace.

regiskuckaertz: I like approach nb 2, however lodash 4 is not compatible with IE9- so it won't be possible until we upgrade our cut the mustard test :-/

johnduffell:
sounds like this one faltered, is it best to close and rethink?

regiskuckaertz:
@johnduffell yep.

@sndrs @jbreckmckye @gtrufitt I'm not totally sold on the idea that it's not worth it. We're already asking for _a lot_ on the client-side, so if this doesn't make developer's lives harder then why hesitate? Also, @piuccio 's idea to progressively introduce lodash 3 is interesting.

PR #13031, opened 2016-05-23
Add es5-shim

regiskuckaertz:
In preparation of our setting the baseline to es5 support.

cc @sndrs


regiskuckaertz:
Note that it has a shim for `Function.prototype.bind` so this is potentially something to remove from `polyfill.scala.js` too.

common/app/views/fragments/inlineJSBlocking.scala.html
```diff
@@ -5,6 +5,7 @@
 
 <!--[if lt IE 9]>
     <script src="@Static("javascripts/components/html5shiv/html5shiv.js")"></script>
+    <script src="@Static("javascripts/components/es5-shim/es5-shim.js")"></script>
```
sndrs: should these be merged?

sndrs:
>Note that it has a shim for Function.prototype.bind so this is potentially something to remove from polyfill.scala.js too.

if it's not being used, should remove it. if it is, i reckon it's maybe case for `that = this`?

regiskuckaertz:
@sndrs we use it extensively (at least I do, that's a native currying method given for free), but yeah whenever `that` makes sense. Support is widespread on desktop and mobile so I guess the polyfill can be safely removed.

common/app/views/fragments/inlineJSBlocking.scala.html
regiskuckaertz: I can add a grunt concat task, easy peasy.

sndrs:
> we use it extensively

in the inline stuff? sorry, that's where i meant.

but thinking about it, it'll be polyfilled now anyway. so that comment is pointless ⤵︎

common/app/views/fragments/inlineJSBlocking.scala.html
```diff
@@ -5,6 +5,7 @@
...
 <![endif]-->
```
jbreckmckye: Should we add the `polyfills()` call to this block, for good measure?

static/public/javascripts/components/es5-shim/es5-shim.js
```diff
@@ -0,0 +1,2051 @@
...
+        }
+    } else {
+        m = s + m;
+    }
+
+    return m;
+};
+defineProperties(NumberPrototype, { toFixed: toFixedShim }, hasToFixedBugs);
+
+v
```
jbreckmckye: We can now remove the `String.prototype.trim` shim from `polyfills.scala.js`.

sndrs:
both `.bind` and `.trim` can be removed with this. if you remove them and concat, 👍

sndrs:
gah sorry, github doesn't update any more! @jbreckmckye already pointed that out

sndrs:
> Should we add the polyfills() call to this section of the code, for good measure?

i think that's a good idea

sndrs:
we should also move the loader for enhanced to below the css stuff to i think, while you're in there.

jbreckmckye:
This is awesome!

sndrs:
i wonder if this is a good case for public CDN? seems v likely IE8 browsers would come across this regularly?

jbreckmckye:
@sndrs It couldn't hurt, but I doubt IE8 clients will encounter the shim often enough to keep it in cache. Most of the web is, after all, still built on jQuery 1.x, and a little Angular 1.x - and these will provide their own compatibility layers for IE8.

sndrs:
@jbreckmckye yeah deleted my comment. i think it's fine, and getting round that point of failure just adds complexity

regiskuckaertz:
@sndrs what do you mean by moving the enhanced stuff? `inlineJSBlocking` is already last in the `HEAD`, isn't it?

common/app/views/fragments/inlineJSBlocking.scala.html
```diff
@@ -4,7 +4,8 @@
 @import templates.inlineJS.blocking.js._
 
 <!--[if lt IE 9]>
-    <script src="@Static("javascripts/components/html5shiv/html5shiv.js")"></script>
+    <script src="@Static("javascripts/ie8polyfills.js")"></script>
+    <script>@Html(polyfills().body)</script>
```
sndrs: don't think the polyfills can be scoped to this comment e.g. `raf` is needed by IE9, opera mini, android < 4.4 etc. but good to group them up here.

grunt-configs/concat.js
```diff
@@ -9,6 +9,14 @@ module.exports = function (grunt, options) {
...
             ],
             dest: options.staticTargetDir + 'javascripts/app.js'
+        },
+
+        ie8: {
+            src: [
+                options.staticPublicDir + 'javascripts/components/es5-shim/es5-shim.js',
+                options.staticPublicDir + 'javascripts/components/html5shiv/html5shiv.js'
+            ],
+            dest: options.staticPublicDir + 'javascripts/ie8polyfills.js'
```
sndrs: sorry to be a pain, but how about calling this `es-html-5-shiv` or something like that? not a blocker, but I wonder if this would be more explicit to any interested readers?

```html
<!--[if lt IE 9]>
<script src="@Static("javascripts/es5-html5-shivs.js")"></script>
```

common/app/views/fragments/inlineJSBlocking.scala.html
jbreckmckye: @sndrs - yeah, but you could at least place `polyfills()` straight after the conditional comment. It's all render-blocking code in any case, so the order shouldn't change when the browser will get to paint anything.

regiskuckaertz: Doing it right now. Also I think JSON is also an IE7- issue so I'll move it into ie8polyfills.

sndrs: yep, totally agree.

sndrs: might as well move json to its own conditional comment? seeing as we're in there tidying it all up?

```html
<!--[if lt IE 8]>
    <script src="@Static("javascripts/components/JSON-js/json2.js")"></script>
<![endif]-->
```

PR #13033, opened 2016-05-24
Make 88x71 fluid on mobile

regiskuckaertz:
cc @jbreckmckye 

Before:
picture 2

After:
picture 1

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
+++ b/static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
@@ -486,8 +486,21 @@ define([
                     bonzo(geoMostPopular.elem).remove();
                 });
             });
-        }
+        },
+        '88,70': isFluid250('ad-slot--top-banner-ad'),
+        '88,71': isFluid250('ad-slot--top-banner-ad')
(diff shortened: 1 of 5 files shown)


static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
```diff
@@ -486,9 +486,22 @@ define([
...
                 });
             });
-        }
+        },
+        '88,70': isFluid250('ad-slot--top-banner-ad'),
+        '88,71': isFluid250('ad-slot--top-banner-ad')
+        '88,88': isFluid250('ad-slot--commercial-component'),
     };
 
+    function isFluid250(className) {
```
jbreckmckye: I'm not quite sure I follow this - are these adverts actually fluid250s? If so, can we not just make `fluid250.js` add the class to its container?

regiskuckaertz:
Nope 😭  There are other creatives that need it. This is fairly urgent but obviously the whole naming/logic should be rethought.

jbreckmckye:
👍

jbreckmckye:
👍 if it's actually needed - there's an outstanding question whether our existing Fabric templates could handle this requirement.

PR #13038, opened 2016-05-24
We don't need the inline1-fabric ad type

jbreckmckye:
The first inline slot now always takes a fabric advert on mobile, and `dfp-api.js` creates a sizemapping that lets us target ad dimensions to particular breakpoints, but not outside. This means we can simplify the way we add Fabric ads, inserting ordinary inline1s that have extra 88x71 mappings on sub-tablet breakpoints.


regiskuckaertz:
Yeah for simplicity 👍

PR #13040, opened 2016-05-24
Fabric: enable fixed/parallax background on iOS

regiskuckaertz:
cc @jbreckmckye


jbreckmckye:
I believe that we need to enable the scripted fixed background for Android, and use background-attachment: fixed for iOS Safari.

regiskuckaertz:
@jbreckmckye done! thanks for the advice, I tested the peekaboo template on my old samsung and it worked like a charm, so I don't see why this one couldn't 🎸

static/src/javascripts/projects/common/modules/commercial/creatives/fabric-v1.js
```diff
@@ -80,13 +80,13 @@ define([
                 bonzo(this.layer2).css('background-position', '');
             }
 
-            if (this.scrollType === 'fixed' && hasScrollEnabled) {
+            if (this.scrollType === 'fixed') {
```
jbreckmckye: Might want to add a `hasBackgroundFixedsupport` check to this, too - otherwise, if an Android phone starts supporting the property, every scroll will shift the background twice, and your fixed backgrounds will suddenly become parallax backgrounds!

regiskuckaertz: You're right!

static/src/javascripts/projects/common/modules/commercial/creatives/fabric-v1.js
```diff
@@ -107,6 +107,12 @@ define([
                     .addClass('ad-scrolling-bg-parallax')
                     .css('background-position', '50% ' + scrollAmount + '%');
             }, this);
+        } else if (this.scrollType === 'fixed' && !hasBackgroundFixedSupport) {
+            var adRect = this.$adSlot[0].getBoundingClientRect();
+            var vPos = (window.innerHeight - adRect.bottom + adRect.height / 2) / window.innerHeight * 100;
```
jbreckmckye: Could we just pass this as `px`?

jbreckmckye:
Excellent! 👍

static/src/javascripts/projects/common/modules/commercial/creatives/fabric-v1.js
regiskuckaertz: Hmm, probably. The nice thing with % is that 50% aligns the background perfectly when the slot is right in the middle of the viewport.

jbreckmckye: Then - let's keep it.

PR #13041, opened 2016-05-24
Support branding of individual cards in containers

kelvin-chappell:
This enables capi-sourced logos in container cards.


regiskuckaertz:
Let's launch that little buddy 🚀  👍

Issue #12678, opened 2016-04-26
Comment permalink urls don't always jump to the correct comment


TBonnin:
# ISSUE
Permalink urls don't always jump to the correct comment, especially on the first load 

## Steps to Reproduce
1. Load permalink to a specific comment ex: `https://www.theguardian.com/technology/2016/apr/12/how-we-analysed-70m-comments-guardian-website#comment-72211137`

## Actual Results (include screenshots)
The page loads but doesn't jump to the comment

## Expected Results (include screenshots)
The page load and jump to the comment section

## URL
https://www.theguardian.com/technology/2016/apr/12/how-we-analysed-70m-comments-guardian-website#comment-72211137

## Note: 
If you hit return it tends to work the 2nd time, so probably a race condition related to how javascript loads

Thanks to @gidsg for reporting the issue.

paperboyo:
Duplicate of https://github.com/guardian/frontend/issues/10471 and, probably, many more before it...

gidsg:
Seems to be completely broken in Safari (not intermittently working)
This was reported by users and I have reproduced

nicl:
@TBonnin any progress/update on this? User Help are still getting a lot of emails about this.

TBonnin:
I haven't made this my priority at all to be honest. I am happy to revisit though

ScottPainterGNM:
@gtrufitt is this closable now or still an issue?

gtrufitt:
Still getting some reports, but I'm no longer able to replicate. Apparently on `http` pages the comments jump doesn't work for some users.

This url works for me on Safari: https://discussion.theguardian.com/comment-permalink/75097820

Does it work for you @ScottPainterGNM?

These specific urls were reported:

http://www.theguardian.com/us-news/2016/may/27/chelsea-manning-six-years-prison-whistleblowing
http://www.theguardian.com/commentisfree/2016/may/27/acned-letter-teenage-self-acne-philippa-perry
http://www.theguardian.com/commentisfree/2016/may/27/obama-apologise-hiroshima-lessons-bombed
http://www.theguardian.com/cities/2016/may/27/isis-palmyra-3d-technology-copy-rebuild-city-venice-biennale

ScottPainterGNM:
That url you sent me didn't work at all on Safari, it just took me straight to the article

gtrufitt:
What Safari?

ScottPainterGNM:
9.0.1

gtrufitt:
[...]

Issue #10471, opened 2015-09-10
firefox - link to comment does not take you to the comment


michaelwmcnamara:
links to comment are not working for any page type.
eg: http://www.theguardian.com/crosswords/quick/14146#comment-59152565

The page loads but scroll does not jump to the linked comment
![image](https://cloud.githubusercontent.com/assets/7239344/9792991/16142640-57db-11e5-9fe7-d0994171bbc2.png)

Severity of this issue is increased as it also applies to the "view discussion" link in commnets and replies on your profile page.
eg a user who wants to view the discussion around the comment he made would click on this link to go straight there, however because this is broken he has to search through the  hundreds of comments that have since been added in order to deliver his brilliant riposte.

crifmulholland:
Which browser did you manage to recreate this on?

ScottPainterGNM:
I tried on firefox and if you click the direct link like the one linked above it didn't scroll down but if I click on a time to create a comment permalink then the scroll worked?

![permalink firefox](https://cloud.githubusercontent.com/assets/11922253/9793235/51e77c3e-57dc-11e5-90f6-94065650cf5a.gif)

ScottPainterGNM:
I'm having issues with the permalink not jumping to pages on non crosswords with Firefox as well, going to flag up for sustain rather than loyalty.

michaelwmcnamara:
Issue found on firefox 40.0.3 running on OSX Yosimite

paperboyo:
Reproduced now on 40.0.3 on 10.9.5, but observed for long time on different OSes/FF versions. It couldn't possibly be related to https://github.com/guardian/frontend/issues/10331 (and, then, to https://github.com/guardian/frontend/issues/9927 & https://github.com/guardian/frontend/issues/8338), could it? :laughing:

paperboyo:
Can’t reproduce any more. Possibly fixed by https://github.com/guardian/frontend/pull/10829? Safe to close?

paperboyo:
Doesn’t work on IE11 & Edge on Windows 10 either. Sometimes just clicking reload helps, sometimes it doesn’t. Chrome & Opera unaffected.

gidsg:
Closing as duplicate of #12678

PR #13042, opened 2016-05-24
Discussion bug #12678 #10471 - Comment Permalinks

gtrufitt:
## What does this change?

Fixes the issue where if a user follows a permalink to a comment, the page doesn't jump to the comment.

Issues #12678 #10471
## Request for comment

@nicl @gidsg


ScottPainterGNM:
Boom! 
image

gidsg:
Cool tricks @gtrufitt, have you tested this in Safari (as many of the reported users were on that browser?)

gtrufitt:
@gidsg Yeah, fixes Safari too. We were seeing it in Chrome too so not sure on the reason behind most reports being safari.

nicl:
👍 😲

PR #13045, opened 2016-05-24
Styling issues in CAPI components on mobile

regiskuckaertz:
Before:
picture 6

After:
picture 7

(images distorted and text size larger than supposed to)


Calanthe:
👍

PR #13046, opened 2016-05-24
Further Commercial Admin Page Cleanup

scottp719:
## What does this change?

Further removal of defunct elements of the commercial tools page, Slot Load times relied on Ophan data that we no longer receive and adunits for approval is now a function that dfp can perform itself so nobody uses this anymore.

Less code/only keeping tools running that are used

No
## Screenshots

n/a
## Request for comment

@kelvin-chappell


regiskuckaertz:
yeah 👍

PR #13047, opened 2016-05-24
Renault native cards - more fixes

lps88:
## What does this change?
- Unlink the Renault logo and remove the field from the template in DFP
- Make the link open in the same tab rather than opening in a new tab
- Ad button needs to be a perfect circle
## Request for comment

@Calanthe


Calanthe:
👍

PR #13049, opened 2016-05-24
make video ad call https

jamesgorrie:
## What does this change?

As @akash1810 was perusing the solution to another problem - this came up.
Seems like we should just switch it to HTTPS.

Nothing happens, and we still rake in the cash from video ads.
## Request for comment

@akash1810 @JustinPinner

### DIFF
--- a/static/src/javascripts/bootstraps/enhanced/media/main.js
+++ b/static/src/javascripts/bootstraps/enhanced/media/main.js
@@ -68,7 +68,7 @@ define([
             unviewed_position_start: 1
         };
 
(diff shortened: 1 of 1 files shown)


akash1810:
provided dfp support it 👍

humaislam:
Double checking the dfp support now

JustinPinner:
Seems sensible. Is there any advantage to making it protocol-neutral (`//` instead of `https://`)? Assuming that works in this context.

jamesgorrie:
@JustinPinner - [It's considered by some to be an anti-pattern](http://www.paulirish.com/2010/the-protocol-relative-url/) by some. I personally agree, why allow the ability to be worse?

PR #13050, opened 2016-05-24
add ec2 instance and request info to logs

TBonnin:
## What does this change?

Add additional fields to server-side logs (ec2 instance id, request url, request method, request id, request latency)

We can look into the logs by ec2 instance as well as querying by those new request fields

No
## Request for comment

@johnduffell @alexduf

### DIFF
--- a/common/app/common/Logback/Logstash.scala
+++ b/common/app/common/Logback/Logstash.scala
@@ -1,6 +1,7 @@
 package common.Logback
 
 import com.amazonaws.auth.AWSCredentialsProvider
+import com.amazonaws.util.EC2MetadataUtils
 import common.ManifestData
 import conf.switches.Switches
 import conf.Configuration
@@ -29,7 +30,8 @@ object Logstash {
     "app" -> Configuration.environment.projectName,
     "stage" -> Configuration.environment.stage.toUpperCase,
     "build" -> ManifestData.build,
-    "revision" -> ManifestData.revision
+    "revision" -> ManifestData.revision,
+    "ec2_instance" -> Option(EC2MetadataUtils.getInstanceId).getOrElse("Not running on ec2")
   )
 
   val config = for {
--- a/common/app/filters/RequestLoggingFilter.scala
+++ b/common/app/filters/RequestLoggingFilter.scala
@@ -1,59 +1,112 @@
 package filters
 
+import java.util.UUID
+
 import common.{ExecutionContexts, Logging, StopWatch}
-import play.api.mvc.{Result, RequestHeader, Filter}
+import play.api.mvc.{Filter, RequestHeader, Result}
 
 import scala.concurrent.Future
 import scala.util.{Failure, Random, Success}
 import conf.switches.Switches
+import net.logstash.logback.marker.LogstashMarker
+import net.logstash.logback.marker.Markers._
+import play.api.Logger
+
+import scala.collection.JavaConverters._
 
 class RequestLoggingFilter extends Filter with Logging with ExecutionContexts {
+
+  case class RequestLogger(rh: RequestHeader)(implicit internalLogger: Logger, stopWatch: StopWatch) {
+    private lazy val pseudoId = Random.nextInt(Integer.MAX_VALUE)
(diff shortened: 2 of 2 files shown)


common/app/filters/RequestLoggingFilter.scala
```diff
@@ -1,59 +1,114 @@
...
+        requestLog.warn(s"${rh.method} ${rh.uri} failed after ${stopWatch.elapsed} ms", error)
         log.warn(s"${rh.method} ${rh.uri} failed after ${stopWatch.elapsed} ms", error)
     }
-
     result
   }
 }
 
 object DiscussionRequestLoggingFilter extends Filter with Logging with ExecutionContexts {
+
```
TBonnin: Code below is duplicate I will remove once we figure out what is going on with discussion

common/app/filters/RequestLoggingFilter.scala
```diff
@@ -1,59 +1,112 @@
...
+import net.logstash.logback.marker.LogstashMarker
+import net.logstash.logback.marker.Markers._
+import play.api.Logger
+
+import scala.collection.JavaConverters._
 
 object RequestLoggingFilter extends Filter with Logging with ExecutionContexts {
+
+  case class RequestLogger(rh: RequestHeader)(implicit internalLogger: Logger, stopWatch: StopWatch) {
+    private lazy val pseudoId = Random.nextInt(Integer.MAX_VALUE)
```
johnduffell: isn't the default limit for nextInt max-value anyway?

johnduffell:
instance_id! 🎉 great thanks 👍 🚤 🐎

common/app/common/Logback/Logstash.scala
```diff
@@ -29,7 +30,8 @@ object Logstash {
     "app" -> Configuration.environment.projectName,
     "stage" -> Configuration.environment.stage.toUpperCase,
     "build" -> ManifestData.build,
-    "revision" -> ManifestData.revision
+    "revision" -> ManifestData.revision,
+    "ec2_instance" -> Option(EC2MetadataUtils.getInstanceId).getOrElse("Not running on ec2")
```
alexduf: Just a note, I don't think it matters here but ```EC2MetadataUtils.getInstanceId``` is a blocking call

common/app/filters/RequestLoggingFilter.scala
```diff
@@ -1,59 +1,112 @@
...
 
 object DiscussionRequestLoggingFilter extends Filter with Logging with ExecutionContexts {
+
+  case class RequestLogger(rh: RequestHeader)(implicit internalLogger: Logger, stopWatch: StopWatch) {
+    private lazy val uuid = UUID.randomUUID()
+    private def customFieldsMarkers(): LogstashMarker = {
+      val fields = Map(
+        "req.method" -> rh.method,
+        "req.url" -> rh.uri,
+        "req.uuid" -> uuid.toString,
```
alexduf: I don't mind the use of a UUID or random, but why using something different here than above?

alexduf:
Already looks good to me, there's a minor comment in there :+1:

common/app/common/Logback/Logstash.scala
johnduffell: good spot there!

common/app/filters/RequestLoggingFilter.scala
TBonnin: `nextInt()` returns values between minInt and maxInt. `nextInt(n)` returns values between 0 and n
I wanted to avoid negative values here

common/app/filters/RequestLoggingFilter.scala
TBonnin: Oops forget to change this to `Random.nextInt()`. Will fix
`UUID` might have some performance implications I didn't want to introduce on every single request

common/app/common/Logback/Logstash.scala
TBonnin: Good point. This is only executed once when the app starts and the logger is setup so it is not a big deal here. However I am happy to improve if you have a better idea

PR #13051, opened 2016-05-24
Remove the handling of filters from the "global" of each individual micro service

alexduf:
## What does this change?

This change extracts the filters out of `Global`s states of each micro service.
It is back ported from my play-2.5.3 branch and adapted to play 2.4
Filters are transformed from object to classes as they will need to be injected with a akka.stream.Materializer in play 2.5

It keeps the play-2.5.3 branch to human readable size and prepares the project to support dependency injection

No
## Request for comment

@johnduffell @TBonnin

### DIFF
--- a/discussion/app/Global.scala
+++ b/discussion/app/Global.scala
@@ -17,12 +17,12 @@ class DiscussionFilters extends HttpFilters {
   // NOTE - order is important here, Gzipper AFTER CorsVaryHeaders
   // which effectively means "JsonVaryHeaders goes around Gzipper"
   lazy val filters: List[EssentialFilter] = List(
+    new PanicSheddingFilter,
     new JsonVaryHeadersFilter,
     new Gzipper,
     new BackendHeaderFilter,
     new DiscussionRequestLoggingFilter,
     new SurrogateKeyFilter,
-    new AmpFilter,
-    new PanicSheddingFilter
+    new AmpFilter
   )
 }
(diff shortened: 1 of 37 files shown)


alexduf:
I just rebased and forced push as I didn't see I had a merge conflict

johnduffell:
seems a step backwards we have to create instances of a class just to keep play happy.  also it's a shame we're moving a compile time dependency to a string in a conf file.

alexduf:
@johnduffell that's temporary as when I have gotten rid of the ```Global``` objects I will be able to define an ```ApplicationLoader``` where we can override a ```filters``` attribute.
It's a temporary state that allows me to cut the migration in smaller chunks

johnduffell:
fair enough! sounds good then... may be worth adding a comment in the code if it would be convenient
👍

discussion/app/Global.scala
```diff
@@ -1,28 +1,28 @@
...
-    AmpFilter,
-    PanicSheddingFilter
+  lazy val filters: List[EssentialFilter] = List(
+    new JsonVaryHeadersFilter,
+    new Gzipper,
+    new BackendHeaderFilter,
+    new DiscussionRequestLoggingFilter,
+    new SurrogateKeyFilter,
+    new AmpFilter,
+    new PanicSheddingFilter
```
johnduffell: this should really be first, I thought I changed it but didn't spot here

alexduf: Good point, I missed that when rebasing, thanks.

TBonnin:
👍  Thanks for doing this @alexduf

PR #13052, opened 2016-05-25
Delete polling soft purge job.

gklopper:
## What does this change?

Gets rid of the polling job that does soft purge for content. This will be replaced by a listener on the event stream from CAPI.

Purging will be more effective and we can start pushing up cache times. Once that is done then cache hit rate should go up and result in origin bandwidth and CPU savings.

Depends on https://github.com/guardian/fastly-cache-purger/pull/3


johnduffell:
great! ship it 👍 🚤

johnduffell:
ship to prod @gklopper ?

PR #13054, opened 2016-05-25
Support unique content in shared hosted template

kelvin-chappell:
This enables us to use the same template for different videos, so everything for episode 1 should be there.
Hopefully I haven't broken the teaser.

/cc @Calanthe @lps88

### DIFF
--- a/commercial/app/controllers/commercial/HostedContentController.scala
+++ b/commercial/app/controllers/commercial/HostedContentController.scala
@@ -10,14 +10,28 @@ import views.html.hosted.{episode1Video, guardianHostedPage, teaserVideo}
 object HostedContentController extends Controller {
 
   def renderHostedPage(pageName: String) = Action { implicit request =>
-    val page = HostedPage(pageName)
+    val pageUrl = controllers.commercial.routes.HostedContentController.renderHostedPage(pageName).absoluteURL
     pageName match {
 
       case "design-competition-teaser" =>
+        val page = HostedPage(
+          pageUrl,
+          pageName,
+          pageTitle = "Advertiser content hosted by the Guardian: Designing the car of the future - video",
+          videoTitle = "Designing the car of the future",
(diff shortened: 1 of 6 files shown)


Calanthe:
Maybe you should ask a backend dev for a review as well, but it's 👍  from me.

commercial/app/controllers/commercial/HostedContentController.scala
```diff
@@ -10,14 +10,28 @@ import views.html.hosted.{episode1Video, guardianHostedPage, teaserVideo}
 object HostedContentController extends Controller {
 
   def renderHostedPage(pageName: String) = Action { implicit request =>
-    val page = HostedPage(pageName)
+    val pageUrl = controllers.commercial.routes.HostedContentController.renderHostedPage(pageName).absoluteURL
```
regiskuckaertz: Isn't it too expensive to call this function all the time, just to get an URL?

kelvin-chappell: I don't think so.  This is only going to be used for 2 or 3 pages, before we build a capi-based solution.  It gets used a lot in views to resolve URLs.

regiskuckaertz:
Not a backend but it all looks good 👍

commercial/app/controllers/commercial/HostedContentController.scala
janua: You could make it a `lazy`!

kelvin-chappell: It's always going to be used though for every request of this page.

janua: No it's not. It's only going to be needed when the `pageName` matches the `case matches` :smile:

kelvin-chappell: Oh yes.  You're right!

kelvin-chappell: Please see #13058

PR #13057, opened 2016-05-25
Various fixes on the hosted pages

Calanthe:
## What does this change?

This PR introduces a few changes to the hosted pages:
1. The video dimensions are now 960px x 540px on desktop and 100% of width on tablet & mobile.
2. The accessibility of the fading in-out animation is improved by changing the `visibility` property. (Inspired by the https://medium.engineering/five-goofy-things-medium-did-that-break-accessibility-3bc804ae818d#.834wsbg6w) 
3. The whole page is now wider - we have more space on the left column. Also the bottom car banner is wider.
4. The header on the video is now smaller and aligned properly to the new column dimensions.

The most important value is that the video is above the fold now.
## Screenshots

Before:
screen shot 2016-05-25 at 13 50 10

After:
screen shot 2016-05-25 at 13 50 29
## Request for comment

@lps88 @regiskuckaertz @jbreckmckye @blongden73 

TODO in another PR:
- add an animated 'Watch now' button in the middle of the video layer
- hide control buttons on the mouseout event
- change The Guardian logo


rich-nguyen:
👍

static/src/stylesheets/module/commercial/glabs/_hosted.scss
```diff
@@ -1,13 +1,17 @@
 $hosted-stripe: 28px;
+$hosted-video-width: 960px;
+$hosted-video-height: 540px;
 
 @mixin hosted-fade-in {
+    visibility: visible;
     opacity: 1;
-    transition: opacity 1s linear;
+    transition: opacity 1s linear, visibility 0s;
```
jbreckmckye: Do we actually want the things we're making invisible to be hidden from screen readers, and if we do, should we flag them with ARIA live regions?

Calanthe: Which ARIA live attribute do you have in mind? I can use the `aria-live="polite"` on the header, so the screen reader will speak it's value whenever the user is idle.

And yes, I want to make the header completely invisible, because after changing only the `opacity: 0`, the text layer is still there and it's impossible to click on the 'play' overlay.

jbreckmckye: It depends on the design of your page. Screen readers won't - unless coaxed otherwise - tell users that a piece of content has been removed from the page. It looks as though we're using this mixin on the video player's controls. If I'm using a screen reader, start a video, and then need to stop it, will I still be able to operate those controls in the way I originally would have?

Calanthe: I assume that yes, but I need to test it properly. I will add this task to the Glabs board.

PR #13058, opened 2016-05-25
Small optimization

kelvin-chappell:
No longer unnecessarily building a URL.


janua:
I was slightly tongue in cheek :laughing: :+1:

PR #13061, opened 2016-05-25
Stop duplication of styles

kelvin-chappell:
This is to stop same style being applied twice in a hosted page.

/cc @Calanthe @lps88


Calanthe:
👍

PR #13062, opened 2016-05-25
Control player components in JS

jamesgorrie:
## What does this change?

We control the components in JS. This has the added benefit that the DOM of all the other gumpf is not created.

The fullscreen button is also where it should be, especially on fronts where it is hidden on mobile, making our teeny tiny videos massively unappealing.

@paperboyo gives me something.

No
## Screenshots

**NOW!**
now-player

**THEN!**
then-player
## Request for comment

@akash1810 @paperboyo


akash1810:
👍

PR #13066, opened 2016-05-25
Stop tracking external links and download links as exit links

dominickendrick:
## What does this change?

This stops us tracking external links and download links as exit links, which reduces the number of server calls we make to omniture

Saves around £5k (£1K a month)
## Request for comment

@mkopka @desbo


desbo:
👍

PR #13068, opened 2016-05-25
Unleash the fluid ads

regiskuckaertz:
Google has added us to their alpha testers of fluid ads. Quite simply:
- we advertise our slots accept fluid ads using the `data-` attributes
- fluids ads returned by DFP use the 0x0 size

cc @jbreckmckye @kenlim @kelvin-chappell @Calanthe


rich-nguyen:
👍

PR #13073, opened 2016-05-25
Separate A/B test completion event from "can run" criteria

desbo:
## What does this change?

The "complete" event listeners for A/B tests are now registered on every page load, which means they don't depend on the test's `canRun` function being true.

We need to be able to fire complete events outside of `canRun`, for example on a different page to the one the test actually changed.

This is a web exclusive!!!
## Screenshots

Not a chance.
## Request for comment

@dominickendrick

### DIFF
--- a/static/test/javascripts/spec/common/experiments/ab.spec.js
+++ b/static/test/javascripts/spec/common/experiments/ab.spec.js
@@ -280,8 +280,7 @@ define([
 
                 test.one.variants[0].success = spy;
 
-                ab.segment();
-                ab.run();
+                ab.registerCompleteEvents();
 
                 expect(spy).toHaveBeenCalled();
             });
--- a/static/test/javascripts/spec/common/experiments/ab.spec.js
+++ b/static/test/javascripts/spec/common/experiments/ab.spec.js
@@ -273,13 +273,22 @@ define([
                 });
             });
 
-            it('should fire the success function', function() {
+            it('should fire the success function when canRun is true', function() {
                 var spy = sinon.spy();
 
                 ab.addTest(test.one);
-
                 test.one.variants[0].success = spy;
+                ab.registerCompleteEvents();
 
+                expect(spy).toHaveBeenCalled();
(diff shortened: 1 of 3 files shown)


gtrufitt:
🍠 Aces. Some tests to make pass but 🆒 👍

dominickendrick:
👍

jamespamplin:
Thanks for this @desbo can we possibly have a unit test for this if possible?

desbo:
@jamespamplin yep, this broke a few already which I'll fix, and add a new one if it's not already covered.

desbo:
@jamespamplin that's the tests all working, having looked at them the segmentation code is already covered pretty thoroughly, I don't think there would be anything to gain from adding another test since I have only really rearranged code and added memoization.

static/test/javascripts/spec/common/experiments/ab.spec.js
```diff
@@ -273,15 +273,23 @@ define([
...
-
                 test.one.variants[0].success = spy;
+                ab.registerCompleteEvents();
 
-                ab.segment();
-                ab.run();
+                expect(spy).toHaveBeenCalled();
+            });
+
+            it('should fire the success function when canRun is false', function() {
```
jamespamplin: Thanks for this @desbo

jamespamplin:
:+1: thanks!

desbo:
no worries 😄

gtrufitt:
I've realised this is going to skew my tests I think because previously listening to an event on a comment submission would only happen on pages where my test runs (which is what I want), whereas now, if a user comments on *any* page, it'll fire my control completion.

What's the solution here? Check my canRun in my success function too? I wonder whether the default of a completion should have the same restrictions as the test init and then this can be overridden with a setting on the prototype?

Thoughts @desbo @dominickendrick @jamespamplin ?

desbo:
Yeah, I'd add the `canRun` criteria to your control group's success function.

PR #13074, opened 2016-05-25
Video container will match "video" as well as "videos"

sammorrisdesign-zz:
## What does this change?

Makes it so the video container also matches the word "video" as well as "videos" to put the custom logo in place.

Nah
## Request for comment

@jamesgorrie


jamesgorrie:
Which one should it be? I would image "video" as that's what we use elsewhere.
👍

PR #13075, opened 2016-05-25
Fix immersive article headers

sammorrisdesign-zz:
## What does this change?

When adding immersive headers it will check for both classes separately, as per jsoup's `hasClass` function.


jbreckmckye:
👍

PR #13076, opened 2016-05-26
DeploysRadiator: fix when github return commits info without author login

TBonnin:
## What does this change?

If author of a commit is not a known github account (happens if your gitconfig email doesn't match your github email for instance), github api return `null` for commit login. DeploysRadiator was not handling this case very well

https://frontend.gutools.co.uk/deploys-radiator doesn't return a blank page in the case described above

No
## Request for comment

@NataliaLKB @lps88

### DIFF
--- a/static/src/deploys-radiator/app/api.ts
+++ b/static/src/deploys-radiator/app/api.ts
@@ -43,7 +43,7 @@ export const getDifference = (base: string, head: string): Promise<List<GitHubCo
                         {
                             url: gitHubCommitJson.html_url,
                             authorName: gitHubCommitJson.commit.author.name,
-                            authorLogin: gitHubCommitJson.author.login,
+                            authorLogin: (gitHubCommitJson.author) ? gitHubCommitJson.author.login : 'Unknown',
                             message: gitHubCommitJson.commit.message
                         }
                     ))));
@@ -51,7 +51,7 @@ export const getDifference = (base: string, head: string): Promise<List<GitHubCo
                 return response.clone().json<GitHubErrorJson>()
                     // We need to re-assert the type for rejected promises
                     // https://github.com/Microsoft/TypeScript/issues/7588
-                    .then(json => Promise.reject<List<GitHubCommit>>(new Error('foo')));
+                    .then(json => Promise.reject<List<GitHubCommit>>(new Error(json.message)));
             }
         })
 );


static/src/deploys-radiator/app/api.ts
```diff
@@ -43,15 +43,15 @@ export const getDifference = (base: string, head: string): Promise<List<GitHubCo
...
+                            authorLogin: (gitHubCommitJson.author) ? gitHubCommitJson.author.login : 'Unknown',
                             message: gitHubCommitJson.commit.message
                         }
                     ))));
             } else {
                 return response.clone().json<GitHubErrorJson>()
                     // We need to re-assert the type for rejected promises
                     // https://github.com/Microsoft/TypeScript/issues/7588
-                    .then(json => Promise.reject<List<GitHubCommit>>(new Error('foo')));
+                    .then(json => Promise.reject<List<GitHubCommit>>(new Error(json.message)));
```
TBonnin: Took advantage of this commit to return real error message 👊

NataliaLKB:
lgtm 👍

static/src/deploys-radiator/app/api.ts
```diff
@@ -43,15 +43,15 @@ export const getDifference = (base: string, head: string): Promise<List<GitHubCo
                         {
                             url: gitHubCommitJson.html_url,
                             authorName: gitHubCommitJson.commit.author.name,
-                            authorLogin: gitHubCommitJson.author.login,
+                            authorLogin: (gitHubCommitJson.author) ? gitHubCommitJson.author.login : 'Unknown',
```
OliverJAsh: This error slipped through the static type system because, currently in TypeScript, each type can be itself or `undefined`, e.g. `Author = Author | undefined`. This means you sometimes see `foo is not a {property,method} of undefined` errors slipping through the net. TypeScript doesn't know whether `gitHubCommitJson.author` is `Author` or `undefined`.

TypeScript 2 has a feature called control-flow based analysis which fixes this. It will follow the control flow of your program to determine the type. The above (`gitHubCommitJson.author.login`) would throw a compile error, but `gitHubCommitJson.author && gitHubCommitJson.author.login` would not. It forces you to guard these type of errors.

Another cool feature of TypeScript 2 is non-nullable types, which means `Author = Author`, instead of `Author = Author | undefined`.

static/src/deploys-radiator/app/api.ts
OliverJAsh: Oops :-P

static/src/deploys-radiator/app/api.ts
TBonnin: Thanks for the explanation. Looks like typescript 2 brings a lot of improvements. :)

PR #13077, opened 2016-05-26
Native card fixes -

lps88:
## What does this change?
- fix the position of the logo if the text is quite long (was abs positioned against bottom)
- make sure every part is a link
# Before

image
# After

image


Calanthe:
👍

PR #13080, opened 2016-05-26
introduce http handlers

alexduf:
## What does this change?

This follows my previous PR (https://github.com/guardian/frontend/pull/13051) where I was removing filters from the `Global`s.
This PR removes error handling and http request handling out of `Global`s in an effort to keep a future PR (getting rid of `Global`s) human sized.
This means a few things in the present PR are temporary, including the use of `@Inject` and the modifications of the configuration files that I will gladly remove as soon as I technically can.
The next step is to refactor the lifecycles, only then will I be able to delete the `Global`s

Human sized PRs

No
## Screenshots

No
## Request for comment

@johnduffell @TBonnin


johnduffell:
brilliant 👍 also good work on the splitting things down

PR #13082, opened 2016-05-26
Clean up AB test module

regiskuckaertz:
This time with a vengeance.

@sndrs

### DIFF
--- a/static/src/javascripts/projects/common/modules/experiments/ab.js
+++ b/static/src/javascripts/projects/common/modules/experiments/ab.js
@@ -5,6 +5,8 @@ define([
     'common/utils/mediator',
     'common/utils/storage',
     'common/modules/analytics/mvt-cookie',
+    'lodash/utilities/noop',
+
     'common/modules/experiments/tests/fronts-on-articles2',
     'common/modules/experiments/tests/live-blog-chrome-notifications-internal',
     'common/modules/experiments/tests/live-blog-chrome-notifications-prod',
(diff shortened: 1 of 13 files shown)


static/src/javascripts/projects/common/modules/experiments/ab.js
```diff
@@ -443,9 +418,8 @@ define([
          */
         testCanBeRun: function (test) {
             if (typeof test === 'string') {
-                return testCanBeRun(find(TESTS, function (t) {
-                    return t.id === test;
-                }));
+                var testIndex = TESTS.map(getId).indexOf(test);
+                return testIndex !== -1 && testCanBeRun(TESTS[testIndex]);
```
desbo: I wouldn't say this indexOf approach is cleaner than `find` (or `filter`), but you could make use of `getTest` here in any case.

desbo:
other than the comment, looks great to me 👍

static/src/javascripts/projects/common/modules/experiments/ab.js
```diff
@@ -161,30 +145,35 @@ define([
                 && test.canRun() && !expired && isTestSwitchedOn(test));
     }
 
+    function getId(test) {
```
sndrs: wat 👀 !

static/src/javascripts/projects/common/modules/experiments/ab.js
```diff
@@ -276,15 +265,13 @@ define([
...
-                return variant.id;
-            });
+            var variantIds = test.variants.map(getId);
 
             return variantIds[mvtCookieId % variantIds.length];
         } else {
             return 'notintest';
         }
-    }, function(test) { return test.id; }); // use test ids as memo cache keys
+    }, getId); // use test ids as memo cache keys
```
sndrs: ah ha

sndrs:
👍 lgtm, is @desbo is ok

static/src/javascripts/projects/common/modules/experiments/ab.js
regiskuckaertz: I would go as far as creating a functional because that pattern can be seen across the whole codebase. So:

```
function getProp(prop) {
  return function (_) {
    return _[prop];
  }
}

var getId = getProp('id');
...
```

static/src/javascripts/projects/common/modules/experiments/ab.js
regiskuckaertz: oh, good catch! thanks

PR #13083, opened 2016-05-26
Second Renault Video needs to have a different poster image

lps88:
## What does this change?

The latest Renault "Hosted By" video has a different poster image than the first one

Also changed the opengraph properties to use static image file instead of hardcoded s3 link, but need to test it works.
## Request for comment

@Calanthe @kelvin-chappell


commercial/app/views/hosted/episode1Video.scala.html
```diff
@@ -8,7 +8,7 @@
   preload="auto"
   style="width: 100%;"
   data-block-video-ads="true"
-  poster="@Static("images/commercial/renault-video-poster.jpg")"
+  poster="@Static("images/commercial/renault-video-poster-ep1.png")"
```
kelvin-chappell: Could be `@page.poster` if page is passed through as parameter

kelvin-chappell: Same for `teaserVideo.scala.html`

kelvin-chappell:
Apart from repeating URL, looks good 👍

paperboyo:
Whoa, stop the press! This image is 2MB 😆 Here it is a bit nicer (let me know if that's unacceptable) and 10 times smaller (images should go through the resizer, but these cannot, right?).

Also, we have removed all poster frames and rely on trailpics overlaid on videojs instead. Again, not sure if that would apply here...

renault-video-poster-ep1

lps88:
@paperboyo thanks! can't believe I didn't notice the file size, oops. We're looking to use video atoms soon, and these pages are static prototypes for now so we'll have a better solution for the video/posters before long (hopefully) 😸

PR #13085, opened 2016-05-26
Rename participation comments test

NathanielBennett:
## What does this change?

Turns out that the original hide comments test (https://github.com/guardian/frontend/pull/12990) proved a little controversial - the 'email user help' link dumps ab participations in the footer of the email, making it obvious what we were doing. Also, someone complained to user help that they couldn't see a comment thread that they'd contributed to earlier ( presumably on a different machine ). To mitigate these things, this re-implements the test - for the same audience and time frame - with a less obvious name and excludes signed-in users

We want to measure the effect comments has  on ( non-commenters ) browsing habitd
## Request for comment

@gtrufitt @gidsg @nicl


gtrufitt:
👍

PR #13086, opened 2016-05-26
Fixed height top slot

regiskuckaertz:
Says Pat:

> we have the nod to remove 728x90s from the Tech section in the next week or so as an initial trial which will inform removing them from the site

This revives the fixed height top slot, only this time the height is fixed at 250px everywhere and the ad slot takes care of its own height.

cc @jbreckmckye @kelvin-chappell

### DIFF
--- a/static/src/stylesheets/module/_adslot.scss
+++ b/static/src/stylesheets/module/_adslot.scss
@@ -115,14 +115,7 @@
         z-index: $zindex-popover;
     }
 }
-// Keeping this code for now since we'll be running another similar
-// experiment in the near future:
-// .top-banner-ad-container--small {
-//     min-height: ($gs-row-height * 3) + 18;
-// }
-// .top-banner-ad-container--responsive {
-//     min-height: ($gs-row-height * 7) + 16;
-// }
+
 .top-banner-ad-container--above-nav {
     @include mq(tablet) {
         .js-on:not(.is-adfree) & {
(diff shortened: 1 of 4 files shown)


jbreckmckye:
I notice we also have an extant 940x230 size - what happens to them?

static/src/stylesheets/module/_adslot.scss
```diff
@@ -161,6 +154,13 @@
         padding-bottom: $gs-row-height / 2;
         padding-top: $gs-row-height / 2;
 
+        /* Oh horrible code, I'm looking at you. One day, you'll disappear.
+           One day, I'll fucking rip you apart.
+         */
+        &.ad-slot--h250 {
+            height: 250px;
```
jbreckmckye: If the ad has a height of 230, are we happy showing an extra 20px gap underneath it? (AFAIK, the ad will just align to the top of its container)

regiskuckaertz:
James sayz we must keep it

static/src/stylesheets/module/_adslot.scss
regiskuckaertz: That's a very good question. As the intent is to eventually remove the top slot animation, I would say yes.

jbreckmckye:
Do we need to add an extra condition to `sticky-nav-banner.js`?

regiskuckaertz:
Yes, adding it. Thanks.

jbreckmckye:
👍

PR #13087, opened 2016-05-26
Remove server-side A/B tests variants

TBonnin:
## What does this change?

We used to have 10 variants (each of the representing 1% of our traffic)
which you can assign server-side a/b tests to.
However this was creating 10 CDN 'buckets' for each url and therefore
reduced our cache/hit ratio and cost us a lot of money.
We are moving away from this system: this patch is removing the
remaining variants and refactor slightly the MultiVariateTesting file

I guess this server-side code can disappear fully once the current tests has been run and the new a/b test framework Dom, Sam, etc... are working on

A PR to fastly-edge-cache will follow. :)

Removing all variants CDN-side as well as making the code less confusing regarding what it is allowed and what it is not.

No
## Request for comment

@johnduffell 
FYI @desbo @dominickendrick

### DIFF
--- a/common/app/common/Edition.scala
+++ b/common/app/common/Edition.scala
@@ -165,9 +165,3 @@ object InternationalEdition {
     }
   }
 }
-
-object InternationalEditionVariant {
-
-  def apply(request: RequestHeader): Option[String] = request.headers.get("X-GU-International")
-
-}
--- a/common/app/mvt/MultiVariateTesting.scala
+++ b/common/app/mvt/MultiVariateTesting.scala
@@ -1,7 +1,5 @@
 package mvt
 
-import MultiVariateTesting._
-import common.InternationalEditionVariant
 import conf.switches.{SwitchGroup, Switch}
 import org.joda.time.LocalDate
 import play.api.mvc.RequestHeader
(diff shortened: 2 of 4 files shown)


common/app/common/Edition.scala
```diff
@@ -165,9 +165,3 @@ object InternationalEdition {
     }
   }
 }
-
-object InternationalEditionVariant {
-
-  def apply(request: RequestHeader): Option[String] = request.headers.get("X-GU-International")
```
TBonnin: This test has been revoked long ago https://github.com/guardian/fastly-edge-cache/commit/a4a9244 :D

common/app/mvt/MultiVariateTesting.scala
```diff
@@ -103,40 +90,5 @@ case class TestDefinition (
...
-  private[mvt] val allVariants = List[Variant]()
-
-  def getVariant(request: RequestHeader): Option[Variant] = {
-    val cdnVariant: Option[String] = request.headers.get("X-GU-mvt-variant")
-
-    cdnVariant.flatMap( variantName => {
-      allVariants.find(_.name == variantName)
-    })
-  }
+  def isParticipating(implicit request: RequestHeader): Boolean = switch.isSwitchedOn && ServerSideTests.isSwitchedOn
```
johnduffell: if we're expected to override something it should really be abstract  to make it easier to follow what the code is doing, also make sure people don't forget to implement it
consider to pass in some code rather than overriding though (composition > inheritance)

common/app/mvt/MultiVariateTesting.scala
```diff
@@ -19,78 +17,67 @@ import conf.switches.Switches.ServerSideTests
...
 
 object ABOpenGraphOverlay extends TestDefinition(
-  variants = Nil,
   name = "ab-open-graph-overlay",
   description = "Add a branded overlay to images cached by the Facebook crawler",
   sellByDate = new LocalDate(2016, 6, 29)
 ) {
   override def isParticipating(implicit request: RequestHeader): Boolean = {
-    request.queryString.get("page").exists(_.contains("facebookOverlayVariant")) && switch.isSwitchedOn && ServerSideTests.isSwitchedOn
+    request.queryString.get("page").exists(_.contains("facebookOverlayVariant")) && super.isParticipating(request)
```
johnduffell: also ref my previous comment people will surely forget to call super.isParticipating sometimes

johnduffell:
based on the mvt id we actually only use 90% of the space for the client side ab tests (to avoid people clashing client and server side tests without realising) - we should change that back to 100% just to get a higher number of samples for the client side tests

johnduffell:
apart from that thanks! 👍

common/app/mvt/MultiVariateTesting.scala
TBonnin: overriding is not mandatory. If you want to run an a/b test everywhere, you won't need to override isParticipating.

TBonnin:
Good point about the client side mvt id range https://github.com/guardian/frontend/pull/13087/commits/2c4212418e8a2ebd6729fd471c52c49013d5ae7e

static/src/javascripts/projects/common/modules/analytics/mvt-cookie.js
```diff
@@ -6,10 +6,7 @@ define([
     var MULTIVARIATE_ID_COOKIE = 'GU_mvt_id',
         VISITOR_ID_COOKIE = 's_vi',
         BROWSER_ID_COOKIE = 'bwid',
-        // The full mvt ID interval is [1, 1000000]
-        // Server side mvt IDs occupy [9000000, 1000000)
-        // So the client-side mvt interval is [1, 899999]
-        MAX_CLIENT_MVT_ID = 899999;
+        MAX_CLIENT_MVT_ID = 1000000; // The full mvt ID interval is [1, 1000000]
```
TBonnin: @dominickendrick Would this change affect client-side tests currently running? :)

dominickendrick: Yes, this will increase the size of each allocation, and people in one set of tests could be removed or pushed across to another test.

TBonnin:
Nvm I checked with @dominickendrick and he advises me not to change the client side range since it could move some people from one test bucket to another.

TBonnin:
@johnduffell I addressed your concern about possible oversight when overriding `isParticipating` https://github.com/guardian/frontend/pull/13087/commits/7eaceaf77d063d0bcf213baf05579da24e7aa03b 😁

johnduffell:
also there's a load of logic about server side tests in here: https://github.com/guardian/frontend/blob/32f16818651e2c423ddd1a2367ac5a0f915ac921/static/src/javascripts/projects/common/modules/experiments/ab.js#L327 😝

TBonnin:
@johnduffell Yep that's the piece that pushes the running tests info to Omniture. I am not removing server-side tests. I am only removing the variants

johnduffell:
@dominickendrick even if I create a completely new incognito window and hit any page, I still get variants stored in local storage, so it wouldn't affect existing bucketing
image
see it in the code:
https://github.com/guardian/frontend/blob/32f16818651e2c423ddd1a2367ac5a0f915ac921/static/src/javascripts/projects/common/modules/experiments/ab.js#L276

having said all that, it would be possible for someone to get in two tests that are only just not overlapping if they started at different times and the user visited when one had started and then we changed the allocations and the other started, and then they visited again

johnduffell:
oh yeah good point thomas, I realised that and then forgot straight away 😨

johnduffell:
👍 this one is fine, go for it! we should come back to doing the mvt ids though as I'm sure it would be fine

PR #13088, opened 2016-05-26
Merge separate hosted video templates

kelvin-chappell:
All differences are now in the controller, but should have no effect on rendered output.

Depends on #13083 

/cc @lps88 @Calanthe


lps88:
This works much better 👍

PR #13089, opened 2016-05-26
Add a messenger share button

mchv:
## What does this change?

This adds a share on messenger social button.

Facebook want us to test out their new messenger share option - it's currently in beta. We're trying to build a trustful relationship with them and being early adopters of their platform and new services should help with that.  

We also don't know, how much of our audience uses Facebook Messenger. We would like to use the data from this social sharing option as a proxy to answer this question.
## Screenshots

screen shot 2016-05-26 at 16 03 25


cb372:
Why isn't the Pinterest button shown in your screenshot?

paperboyo:
https://youtu.be/VfcBxu101qM?t=2

Maybe we could hide them all behind one button?

ScottPainterGNM:
@cb372 I think we only show pinterest on galleries and whatsapp is mobile breakpoint only, I'm guessing the latter would be appropriate for messenger as well?

LATaylor-guardian:
@mchv Can we see how many people are clicking this button as soon as this change happens without further changes elsewhere?

mchv:
@LATaylor-guardian  I think I need first to make a chance to the fastly cache to redirect `sme` to a campaign code. 
@cb372 I have been playing with removing `pinterest` button, but I think it is not necessary a first step.

mchv:
@LATaylor-guardian change was in frontend and I just made it 😃

static/src/stylesheets/module/_social.scss
```diff
@@ -159,6 +165,13 @@ category: Common
     }
 }
 
+.social-icon--messenger {
+    background-color: $social-messenger;
+    &:hover {
+        background-color: darken($social-messenger, 10%);
```
gtrufitt: Is this, erm, official Facebook brand hover colour..? Also, you might want to add `:focus` and `:active` states here.

common/app/model/ShareLinks.scala
```diff
@@ -28,6 +28,14 @@ object Facebook extends SharePlatform {
   override val css = "facebook"
   override val userMessage = "Share on Facebook"
 }
+
+object Messenger extends SharePlatform {
+  override val campaign = Some("sme")
+  override val text = "Messenger"
+  override val css = "messenger"
+  override val userMessage = "Share on Messenger"
```
gtrufitt: (minor) 'Facebook Messenger' rather than just 'messenger'?

mchv: The app is called `Messenger` (see https://www.messenger.com/about)

static/src/stylesheets/module/_social.scss
mchv: `10%` is what we have for `twitter`, `whatsapp`, `gplus`, so I don't think we should be more specific for messenger here.

johnduffell:
ok it's great to share, but surely we can't just keep adding one more share button? I'm sure the optimal number of share buttons must be around 2.

static/src/stylesheets/module/_social.scss
mchv: Thanks for `:focus` and `:active` states but I think, if we add them, we should add them for all social icons, and so in a separate pr.

mchv:
@johnduffell Yep. We are [monitoring the shares](https://docs.google.com/spreadsheets/d/1HNGgHdfYKPUhhzwt3uqVkLFFuXePSMRcML1RlsSSHX0/edit?ts=5729b71e#gid=1074393378) and we will in the future remove the sharing that are less used. Please note as well that like `whatsapp`, `messenger` only show on mobile 😃

static/src/stylesheets/module/_social.scss
gtrufitt: Fair enough

common/app/model/ShareLinks.scala
gtrufitt: Yeah, noticed that after, trust the big F to TM such a generic name :D

johnduffell:
this is making the amp pages invalid so they're not appearing in search any more

PR #13090, opened 2016-05-26
Expect explicit targeted high merch

rich-nguyen:
Tweaks the high merch slot decision making, so that we only render a high merch slot when we find a line item that explicit matches the page requested. This means that if a line item does not target anything, it doesn't match all the pages on our site. DFP would consider this situation to be run-of-network, but we don't roll high merch across the whole site.

Also updated the admin page so we can easily spot high merch line items that don't have any targeting.


Calum-Campbell:
👍

PR #13092, opened 2016-05-26
hide cursor when inactive over video

akash1810:
## What does this change?

cursor can be distracting when watching video

less distraction = better UX.

No.
## Screenshots

Before
before

After
after
## Request for comment

@jamesgorrie


jamesgorrie:
👍

PR #13093, opened 2016-05-26
AB Test: Football ratings for match reports

gtrufitt:
## What does this change?

This adds an AB test for match reports, using tiny football icons.

We're testing 5% of the audience over this weekend because we got some big games like the CHAMPION'S LEAUGE FINAL and the SSE AIRTRICITY PREMIER DIVISION. Success will be vs a baseline of comments on the match reports.
## Screenshots

football-ratings
## Request for comment

@guardian/participation


NathanielBennett:
:+1: Ship it

PR #13094, opened 2016-05-26
Change `req.latency` log field type to integer

TBonnin:
## What does this change?

Change `req.latency` log field type to integer. (it is a string right now)

so we can do percentile aggregations in Kibana/ES

No
## Request for comment

@johnduffell

### DIFF
--- a/common/app/filters/RequestLoggingFilter.scala
+++ b/common/app/filters/RequestLoggingFilter.scala
@@ -23,7 +23,7 @@ class RequestLoggingFilter extends Filter with Logging with ExecutionContexts {
         "req.method" -> rh.method,
         "req.url" -> rh.uri,
         "req.id" -> pseudoId.toString,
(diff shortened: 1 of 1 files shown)


common/app/filters/RequestLoggingFilter.scala
```diff
@@ -23,7 +23,7 @@ class RequestLoggingFilter extends Filter with Logging with ExecutionContexts {
         "req.method" -> rh.method,
         "req.url" -> rh.uri,
         "req.id" -> pseudoId.toString,
-        "req.latency_ms" -> stopWatch.elapsed.toString
+        "req.latency_millis" -> stopWatch.elapsed
```
TBonnin: Renaming the field because change a field type in ES requires a re-indexing which is quite a costly operation. The field is not used anywhere yet so it doesn't cost anything to add a new field.

johnduffell:
great! 👍 do it!

PR #13095, opened 2016-05-26
Improve omnitureID on CAPI containers

regiskuckaertz:
This change will align as much as possible the omniture ID of our CAPI commercial containers with the ones produced for glabs containers.

cc @kelvin-chappell


kelvin-chappell:
👍

PR #13100, opened 2016-05-27
Passing healthcheck refresh interval as config param

TBonnin:
## What does this change?

Passing healthcheck refresh interval as config param

Healthcheck is only used by ELB and it could be annoying when debugging
in dev locally.
Passing the interval value as a config variable allows to have different
value depending on the STAGE and even disabling it if necessary

No
## Request for comment

@alexduf


alexduf:
:+1: if the compilation passes

PR #13102, opened 2016-05-27
simplify asset finder

johnduffell:
tldr - The asset finding code scared me and @sndrs so I've simplified it.

The back story is Alex is working on something to make the CSS really small which needs a mappings file available on the server side.  When I tried to understand the code in order to add the mappings loaded I eventually realised it's doing something simpler than it looked.  I think it's just a side effect of not having been refactored and been edited by many people.

There shouldn't be any actual change in functionality, but it should be slightly simpler to maintain.

The critical changes are in [assets.scala diff-aef18c02671cecab73288a55f9311a39L1](https://github.com/guardian/frontend/pull/13102/files#diff-aef18c02671cecab73288a55f9311a39L1)

I've tested locally with and without hashed bundles, but as it's a mildly risky change I will also test on code before I ship to prod just in case.

@rich-nguyen any chance you could have a look through please?


common/app/assets/assets.scala
```diff
@@ -1,160 +1,135 @@
...
 import play.api.libs.json._
 import play.api.{Mode, Play}
-import play.api.Play.current
 
 import scala.collection.concurrent.{Map => ConcurrentMap, TrieMap}
-import scala.util.control.NonFatal
+import scala.util.{Failure, Success, Try}
 
-case class Asset(path: String) {
-  val asModulePath = path.replace(".js", "")
```
johnduffell: nothing seemed to use these fields except a test

common/app/assets/assets.scala
```diff
@@ -1,160 +1,135 @@
...
-  override def toString = path
-}
-
-class AssetMap(base: String, assetMap: String) extends Logging {
+  lazy val lookup = Get(assetMap(mapResource))
 
-  def apply(path: String): Asset = try {
-    assets(path)
-  } catch {
-    case NonFatal(e) => {
```
johnduffell: play seems to log the exception anyway if it happens

common/app/assets/assets.scala
```diff
@@ -1,168 +1,135 @@
...
-  val asModulePath = path.replace(".js", "")
-  lazy val md5Key = path.split('/').dropRight(1).last
+// turns an unhashed name into a name that's hashed if it needs to be
+class Assets(base: String, mapResource: String, useHashedBundles: Boolean = Configuration.assets.useHashedBundles) extends Logging {
 
-  override def toString = path
-}
-
-class AssetMap(base: String, assetMap: String) extends Logging {
+  lazy val lookup = Get(assetMap(mapResource))
```
TBonnin: Can we be explicit about the type of `lookup`? It is not obvious when you read the code

johnduffell: good idea, done

common/app/assets/assets.scala
```diff
@@ -1,168 +1,135 @@
...
+  private def cssHead(project: String): String =
+    project match {
+      case "footballSnaps" => "head.footballSnaps"
+      case "facia" => "head.facia"
+      case "identity" => "head.identity"
+      case "football" => "head.football"
+      case "index" => "head.index"
+      case "rich-links" => "head.rich-links"
+      case "email" => "head.email"
+      case _ => "head.content"
```
TBonnin: nitpick: to avoid repetition why not:
`
val cssHeads = List("footballSnaps", "facia", ..., email)
val headSuffix = if (cssHeads.contains(project)) project else "content"
s"head.${headSuffix}"
`

common/app/assets/assets.scala
```diff
@@ -1,168 +1,135 @@
...
 }
 
-case class AssetNotFoundException(assetPath: String) extends Exception {
-  override val getMessage: String = if (Configuration.assets.useHashedBundles) {
-    s"Cannot find asset $assetPath. You should run `make compile`."
-  } else {
-    s"Cannot find asset $assetPath. You should run `make compile-dev`."
-  }
-}
+case class AssetNotFoundException(assetPath: String) extends Exception(s"Cannot find asset $assetPath. You should run `make compile`.")
```
TBonnin: Is the `You should run `make compile`` advice only valid when running services locally?

TBonnin:
Looks good to me 👍

common/app/assets/assets.scala
johnduffell: kind of, well the build should have done it on the real built artifacts, so it shouldn't really happen!  If it did happen we'd be in trouble.

common/app/assets/assets.scala
johnduffell: yeah good point, actually I didn't change that code I just left it as-is, maybe I'll do it another time!

rich-nguyen:
looks good to me. A lot of the code hasn't been sensibly designed, it's just grown incrementally when someone needs to add behaviour.

PR #13103, opened 2016-05-27
Move the HealthChecks in the "controllers" package

alexduf:
## What does this change?

This PR moves all the healthchecks in a consistent place. I chose the `controllers` package as it sounded appropriate.
A lot of the files named `context.scala` are left with only the HealthCheckLifcycle remaining in the file. This is intentional as another PR to come gets rid of these files completely.

Consitency across all the modules

No
## Screenshots

Nope.
## Request for comment

@TBonnin @johnduffell


johnduffell:
yep defainitely makes more sense in the existing structure 👍   although having said that a controllers package only makes sense if you're looking at the world through an MVC lens, which I'm not really sure if it's the best way

TBonnin:
👍

PR #13104, opened 2016-05-27
set videoDuration on article config

akash1810:
## What does this change?

Currently, we serve ads on embedded videos regardless of their duration. This allows the `vl` value to be sent to the ad-server to decide if pre-roll should be served on video.
- No pre-roll on embedded videos < 30 seconds = better UX.
- We obey our own ad-rules.

No.
## Screenshots

screen shot 2016-05-27 at 14 53 53
## Request for comment

@kenlim @jamesgorrie

### DIFF
--- a/common/app/model/content.scala
+++ b/common/app/model/content.scala
@@ -350,6 +350,10 @@ object Article {
     val fields = content.fields
     val bookReviewIsbn = content.isbn.map { i: String => Map("isbn" -> JsString(i)) }.getOrElse(Map())
 
+    // we don't serve pre-roll if there are multiple videos in an article
+    // `headOption` as the video could be main media or a regular embed, so just get the first video
+    val videoDuration = content.elements.videos.headOption.map { v => JsNumber(v.videos.duration) }.getOrElse(JsNull)
+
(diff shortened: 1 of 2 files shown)


common/app/model/content.scala
```diff
@@ -359,7 +363,8 @@ object Article {
       ("lightboxImages", lightbox.javascriptConfig),
       ("hasMultipleVideosInPage", JsBoolean(content.hasMultipleVideosInPage)),
       ("isImmersive", JsBoolean(content.isImmersive)),
-      ("isSensitive", JsBoolean(fields.sensitive.getOrElse(false)))
+      ("isSensitive", JsBoolean(fields.sensitive.getOrElse(false))),
+      ("videoDuration" -> videoDuration)
```
johnduffell: will this conflict with the existing one here? https://github.com/guardian/frontend/blob/1cd351a59b6c4cfd5f50bc3ac267e84cef7c40b1/common/app/model/content.scala#L524

johnduffell:
good idea 👍

common/app/model/content.scala
akash1810: Nope - that one is for Video pages, this is for Article pages

PR #13105, opened 2016-05-27
Respect order of tags when selecting a badge

desbo:
## What does this change?

Article badges will be selected according to the order of tags, so the first tag with a matching badge is used.

fixes a bug 
## Screenshots

before: 
screen shot 2016-05-27 at 14 59 39

after:
screen shot 2016-05-27 at 14 58 39
## Request for comment

@stephanfowler

### DIFF
--- a/common/app/model/Badges.scala
+++ b/common/app/model/Badges.scala
@@ -17,6 +17,9 @@ object Badges {
     else Seq()
   }
 
-  def badgeFor(c: ContentType) = allBadges.find(badge => c.tags.tags.exists(tag => tag.id == badge.seriesTag))
+  def badgeFor(c: ContentType) = c.tags.tags.map(_.id).foldLeft(None: Option[Badge]) { (maybeBadge, tag) =>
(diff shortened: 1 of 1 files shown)


common/app/model/Badges.scala
```diff
@@ -17,6 +17,9 @@ object Badges {
     else Seq()
   }
 
-  def badgeFor(c: ContentType) = allBadges.find(badge => c.tags.tags.exists(tag => tag.id == badge.seriesTag))
+  def badgeFor(c: ContentType) = c.tags.tags.map(_.id).foldLeft(None: Option[Badge]) { (maybeBadge, tag) =>
+      maybeBadge orElse allBadges.find(b => b.seriesTag == tag)
```
johnduffell: clever way of doing it!  now you'll loop through the whole list of tags even if you already found a badge, but it will be a short list so not a problem

johnduffell:
great! thanks 👍

PR #13106, opened 2016-05-27
Hosted by page - new video poster

lps88:
## What does this change?

different image file and some css fixes
## Request for comment

@Calanthe


Calanthe:
👍

PR #13109, opened 2016-05-27
Stars: Bugfix and multi item highlight

gtrufitt:
## What does this change?

Small positioning bugfix and adds multi-item highlight to star ratings and football icons.

Makes the UX nicer on desktop.
## Screenshots

hover highlight

### DIFF
--- a/static/src/javascripts/projects/common/modules/experiments/low-friction-participation.js
+++ b/static/src/javascripts/projects/common/modules/experiments/low-friction-participation.js
@@ -192,23 +192,24 @@ define([
         fastdomPromise.read(function() {
             itemLength = e.currentTarget.getAttribute('data-item-id');
             $lowFricButtons = $('.js-participation-low-fric--button');
-        });
+        }).then(updateIcons);
 
-        fastdomPromise.write(function() {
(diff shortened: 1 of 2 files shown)


static/src/javascripts/projects/common/modules/experiments/low-friction-participation.js
```diff
@@ -185,9 +185,41 @@ define([
...
+
+        fastdomPromise.read(function() {
+            itemLength = e.currentTarget.getAttribute('data-item-id');
+            $lowFricButtons = $('.js-participation-low-fric--button');
+        });
+
+        fastdomPromise.write(function() {
+            $lowFricButtons.removeClass('participation-low-fric--button__is-highlighted');
+
+            if (itemLength > -1) {
```
desbo: feel like you might have a race condition here since (afaik) the fastdom `read` isn't guaranteed to finish before the `write`

NathanielBennett:
@gtrufitt @desbo Yeah I noticed that too. Although I'm guessing you've tested that.
Apart from that LGTM
:+1:

gtrufitt:
Yeah, could happen on slower browsers though so will use promise chain

gtrufitt:
Cheers!

PR #13110, opened 2016-05-27
Track ads redux

regiskuckaertz:
There currently is no tracking of the lifecycle of adverts on the website, except for a flimsy event forgotten as soon as it is dispatched. This is problematic because a few modules rely on this information, and depending on when they get executed, they may never know whether an ad has ever been rendered/loaded.

This is a first batch at trying to solve this. It is incomplete still, in that (1) there still is a chance the event for an advert is dispatched before someone calls `trackAdLoad` and (2) there is no tracking of the ad _rendering_ (our homegrown creatives that break out of iframes, for example, render much later than when DFP dispatches `isSlotRendered`). But I need it for something else more urgent right now. 

Eventually, the state of all ads will be kept in memory, and tracked as soon as they are added.

cc @jbreckmckye


JonNorman:
Looks good to me :+1: - will be nice when we can have a single in-memory state of where each ad-slot is at!

Calanthe:
👍

PR #13111, opened 2016-05-27
Fix dev environment when adblocking

jbreckmckye:
If you use an adblocker whilst developing frontend, commercial.js may blow up on a `require` call. This means we cannot test adblock messaging.

image

This happens because when bundled with `make compile-dev`, the scripts fetch in modules on a file-by-file basis, and any adblocker then has the opportunity to individually bounce the request. Because our module is called `advert`, it fouls fowl of the AdblockPlus' detection heuristics.

The first part of the solution is to rename the file. This solves the immediate the problem.

The second part of the solution is to warn developers that any adblocker they're running might the cause of current issues, or exceptions logged to the console. This also reminds them that they're not actually testing the page as it will actually appear to the user.

Now, when a developer views the console with an adblocker enabled, they receive a warning:

image


piuccio:
👍

johnduffell:
> If you use an adblocker whilst developing frontend
😱 
...for testing purposes

PR #13112, opened 2016-05-27
Safe line item reads

rich-nguyen:
This modifies the way that we construct the `GuLineItem` object. It's nearly parameterless now, since most of the params were proxying to agent data, which I've relocated into service objects that can be called within the implementation.

The aim of this is to add a way that we can reject line items that appear invalid. This may happen because adunit IDs were not being resolved, since this resolution requires a populated agent. Ultimately, I did this in a filter after the line item object is built, but I still kept the other changes because they reduce the passing of static functions aliased as call-by-name params.

Also removes a few unused objects.

Hard to check this, I've seen the adunits failing intermittently. It would certainly nuke a lot of line items that arent targeting `theguardian.com`.

@kelvin-chappell


admin/app/dfp/AdUnitAgent.scala
```diff
@@ -36,4 +36,13 @@ object AdUnitAgent extends DataAgent[String, GuAdUnit] {
 
     maybeData getOrElse Map.empty
   }
+
+}
+
+object AdUnitService {
+
+  def adUnit(adUnitId: String): Option[GuAdUnit] = {
+    AdUnitAgent.get.data.get(adUnitId)
```
rich-nguyen: Each service-named object has been moved from dfpapi.scala

admin/app/dfp/CustomTargetingKeyAgent.scala
```diff
@@ -14,3 +14,15 @@ object CustomTargetingKeyAgent extends DataAgent[Long, String] {
     maybeData getOrElse Map.empty
   }
 }
+
+object CustomTargetingKeyService {
+
+  def targetingKey(session: SessionWrapper)(keyId: Long): String = {
+    lazy val fallback = {
+      val stmtBuilder = new StatementBuilder().where("id = :id").withBindVariableValue("id", keyId)
+      session.customTargetingKeys(stmtBuilder).head.getName
```
JonNorman: Do we need to use `headOption` here, in case we request a key ID that doesn't exist?

admin/app/dfp/CustomTargetingValueAgent.scala
```diff
@@ -14,3 +14,15 @@ object CustomTargetingValueAgent extends DataAgent[Long, String] {
     maybeData getOrElse Map.empty
   }
 }
+
+object CustomTargetingValueService {
+
+  def targetingValue(session: SessionWrapper)(valueId: Long): String = {
+    lazy val fallback = {
+      val stmtBuilder = new StatementBuilder().where("id = :id").withBindVariableValue("id", valueId)
+      session.customTargetingValues(stmtBuilder).head.getName
```
JonNorman: `headOption` question again (as with `targetingKey`)

admin/app/dfp/CustomTargetingKeyAgent.scala
rich-nguyen: Seems valid. I just moved this code from the location below. I'll add a get or else.

JonNorman:
Looks good to me - a lot simpler and cleaner :+1:

PR #13115, opened 2016-05-27
Lifecycle batch 1

alexduf:
## What does this change?

This PR starts to refactor our lifecycle-related events (onStart, onStop) into injectable objects. These injectable object are necessary to be able to migrate to play 2.5.
Given the sheer size of the full change (600+ lines) I started re-deviding it in 3 PR. This is the first batch, expect two more to come on that topic only.
Because of that, a temporary class `BackwardCompatibleLifecycleComponents` makes the link between the old world of `Global`s and the world of `ApplicationLoaders`.

Smooth migration towards play2.5, human size PR
## Request for comment

@TBonnin @johnduffell

### DIFF
--- a/admin/app/Global.scala
+++ b/admin/app/Global.scala
@@ -1,24 +1,31 @@
-import common.CloudWatchApplicationMetrics
+import common.{LifecycleComponent, BackwardCompatibleLifecycleComponents, CloudWatchApplicationMetrics}
 import common.Logback.Logstash
 import common.dfp.DfpAgentLifecycle
+import conf.InjectedCachedHealthCheckLifeCycle
 import conf.switches.SwitchboardLifecycle
-import controllers.AdminHealthCheckLifeCycle
+import controllers.HealthCheck
 import dfp.DfpDataCacheLifecycle
 import model.AdminLifecycle
 import ophan.SurgingContentAgentLifecycle
-import play.api.mvc.{RequestHeader, Results}
+import play.api.GlobalSettings
+import play.api.inject.ApplicationLifecycle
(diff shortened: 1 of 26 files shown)


archive/app/services/ArchiveMetrics.scala
```diff
@@ -1,32 +1,32 @@
...
+  appLifecycle.addStopHook{ () => Future{
+    Jobs.deschedule("ArchiveSystemMetricsJob")
+  }}
 
   private def report() {
     CloudWatch.putMetrics("ArchiveMetrics", List(GoogleBotMetric.Googlebot404Count), List.empty)
   }
 
-  override def onStart(app: PlayApp) {
+  override def start():Unit = {
```
johnduffell: v minor but I'd have added a space before Unit after the :

johnduffell:
looks good, well done on going through all that, pretty mechanical stuff 👍

admin/app/Global.scala
```diff
@@ -1,24 +1,31 @@
...
   with SurgingContentAgentLifecycle
   with DfpAgentLifecycle
-  with DfpDataCacheLifecycle
-  with Logstash
-  with AdminHealthCheckLifeCycle {
+  with Logstash {
 
   override lazy val applicationName = "frontend-admin"
+
+  def lifecycleComponents(appLifecycle: ApplicationLifecycle)(implicit ec: ExecutionContext): List[LifecycleComponent] = List(
```
TBonnin: missing override?

TBonnin:
👍  thanks again for doing this

PR #13122, opened 2016-05-31
Clean up config and url

regiskuckaertz:
@gtrufitt @sndrs 

Also introducing a new `assign` that will use the native implementation if it is available.


regiskuckaertz:
(I just had a look at `createIterator` which forms the heart of lodash `assign`/`defaults`/`forOwn`, and it's 😨 😱 )

static/src/javascripts/projects/common/utils/config.js
```diff
@@ -34,11 +30,11 @@ define([
             return (this.page.series || '').indexOf(name) > -1;
         },
         referencesOfType: function (name) {
-            return chain(this.page.references || []).and(filter, function (reference) {
+            return (this.page.references || []).filter(function (reference) {
```
piuccio: do we have a `filter` polyfill somewhere? That method was introduced in IE9

static/src/javascripts/projects/common/utils/url.js
```diff
@@ -16,9 +11,15 @@ define([
             // eg ?foo=bar&fizz=buzz returns {foo: 'bar', fizz: 'buzz'}
             getUrlVars: function (options) {
                 var opts = options || {};
-                return chain((opts.query || model.getCurrentQueryString()).split('&')).and(compact).and(map, function (query) {
+                return (opts.query || model.getCurrentQueryString()).split('&')
+                    .filter(identity)
```
piuccio: you can remove a dependency on `identity` and write `.filter(Boolean)`

regiskuckaertz: Ooh, this one has been bugging me a lot. Thanks!

static/src/javascripts/projects/common/utils/config.js
regiskuckaertz: Yep, we're loading `es5-shim` for IE8-.

static/src/javascripts/projects/common/utils/url.js
```diff
@@ -1,8 +1,7 @@
 define([
     'common/utils/detect',
-    'common/utils/mediator',
-    'lodash/utilities/identity'
-], function (detect, mediator, identity) {
+    'common/utils/mediator'
```
piuccio: unrelated to your changes, but I was intrigued by a dependency on `mediator`. I think it's a wrong design decision for a utility to listen to events. I tried to search for the subscribed event and found nothing. Wondering if that push state still used?

piuccio:
👍

static/src/javascripts/projects/common/utils/url.js
regiskuckaertz: It is indeed not a great design. Looks like it isn't used anywhere. Let's scratch it!

PR #13125, opened 2016-05-31
More EU referendum badges 

desbo:
These are some other tags that Polly Curtis has asked the EU referendum badge to be added to. Also gets rid rid of a redundant switch.

@stephanfowler


stephanfowler:
👍

PR #13127, opened 2016-05-31
bump the healthcheck interval to 5 minutes to fix CdnHealthCheckTest

alexduf:
## What does this change?

This fixes the default value of the configuration so that the tests run locally

@TBonnin


TBonnin:
👍  sorry about that

PR #13129, opened 2016-05-31
Remove viewability switch

regiskuckaertz:
@rich-nguyen @JonNorman


static/test/javascripts/spec/common/commercial/dfp/dfp-api.spec.js
```diff
@@ -212,6 +212,10 @@ define([
         });
 
         it('should display ads', function (done) {
+            config.page.hasPageSkin = true;
```
rich-nguyen: why has this test been changed to pageskins?

rich-nguyen:
👍  apart from comment on test

static/test/javascripts/spec/common/commercial/dfp/dfp-api.spec.js
regiskuckaertz: Because previously the behaviour in this test [was triggered by the `viewability` switch being off](https://github.com/guardian/frontend/pull/13129/files#diff-844a43280e1947d8dd7cc7a480fd0551R257)

regiskuckaertz:
Thanks 😄

PR #13130, opened 2016-05-31
Remove Fabric switch

regiskuckaertz:
Fabric ads are here to stay

@rich-nguyen @JonNorman


JonNorman:
Looks good! :+1:

As discussed, it would be marvellous to have single objects in JS and Scala respectively that alias specific sizes to their usage as `fabricAdvertsTop` does here. One for the backlog!

PR #13132, opened 2016-05-31
Rollout facebook overlay to food and drink section

dominickendrick:
## What does this change?

This rolls out the Facebook share image with overlay to the food and drink section of the site.
Once we are happy with the results (ie facebook don't get mad at us overlaying our logo on images) we'll roll it out across the whole site.
Follow on from : #13026

We will measure share clickthrough and return rate for Facebook referral users

Facebook and other sites that use og:image properties

@desbo @stephanfowler


paperboyo:
I know it says fb, but wanted to know: it doesn't go to twitter too, does it? Because there is this trick to keep images there free from heavy JPG artifacting. I think.

If they get mad - they can call me on my landline to discuss.

NataliaLKB:
👍

dominickendrick:
@paperboyo 
> I know it says fb, but wanted to know: it doesn't go to twitter too, does it? Because there is this trick to keep images there free from heavy JPG artifacting. I think.

This doesn't show up on twitter, because they use their own `twitter:image` meta tag, which we are not modifying. But any other site that scrapes the `og:image` property would see this new image!

PR #13134, opened 2016-05-31
Remove gallery switch

NataliaLKB:
The verdict for the new gallery design is that the new design is a keeper :+1:. This PR removes the switch and makes the new gallery design permanent! 

Unfortunately I couldn't delete all the old code because paid advertisements ([like this one](http://www.theguardian.com/renault-on-road-revolution/gallery/2016/may/19/clerkenwell-design-week-preview-in-pictures)) are still using it. 

However, I believe @blongden73 is working on a new design for them, so hopefully the rest of the old gallery code can be deleted soon.

Because I renamed some files, this would probably be easier to review commit by commit. 

Nope!


static/src/stylesheets/module/content/_gallery.scss
```diff
@@ -1,11 +1,301 @@
...
+    .meta__numbers {
+        border: 0;
+        flex-shrink: 0;
+        position: static;
+        min-width: gs-span(1);
+    }
 }
 
+// TODO:
+// When paid advertisement features are redesigned, delete below styling
```
NataliaLKB: @blongden73 all this code should be deleted with the new design for paid advertisements that you are coming up with! 😉

static/src/stylesheets/module/content/_gallery.scss
```diff
@@ -1,11 +1,301 @@
...
+        transition: border-color .15s ease-out;
+
+        &:hover {
+            border-bottom: 1px solid $neutral-7;
+            text-decoration: none;
+        }
+    }
+
+    // This is the gallery caption title
+    > strong:first-child {
```
gtrufitt: Can this and the `& + br` be replaced with classes by adding them in body cleaner?

applications/app/views/fragments/newGalleryBody.scala.html
```diff
@@ -1,115 +0,0 @@
-@(page: model.GalleryPage)(implicit request: RequestHeader)
```
dominickendrick: If you do a `git mv` on these files it would be able to keep track of any changes regardless of the file rename

static/src/stylesheets/module/content/_gallery.scss
NataliaLKB: Unfortunately I don't think so, but open to trying if you have an idea. See comments here: 
https://github.com/guardian/frontend/pull/12576#discussion_r61089324

applications/app/views/fragments/newGalleryBody.scala.html
NataliaLKB: That is a good tip! Thank you.

static/src/stylesheets/module/content/_gallery.scss
gtrufitt: We should be able to use the `htmlCleaner` here I believe. We already have something for [figCaptions](https://github.com/guardian/frontend/blob/cf339e646fad8af1db3dab2dd7073d4bdc870960/common/app/views/support/HtmlCleaner.scala#L492) - don't know if you should create a new one for galleryFigCaptions though. It uses jSoup, so you should be able to use pretty much the same selectors to get the title and add a class that you can target within the css instead of this selector?

gtrufitt: Oh, and instead of `display: none;` on the `<br>` you can just target it with jsoup and obliterate it from the DOM.

*but* maybe we shouldn't be messing with this mark-up at all because CAPI now gives us 'blocks' which should allow you to use the data directly see this [CAPI response](https://content.guardianapis.com/technology/gallery/2016/may/31/12-visually-arresting-games-playstation-4?format=json&api-key=teleporter-view&show-elements=all&show-rights=all&show-fields=all&show-tags=all&show-blocks=all&show-references=all) - you can see `imageTypeData` which has all of the caption data, without any of the HTML.

NataliaLKB: Yes I think you are right and we can add another cleaner to do this as well. I am happy to do that, but I am unsure which approach is better. It seems to be that this would be simply moving the complexity.

gtrufitt: IMO have a complex selector in the CSS that then needs more complex selectors in the future to override or refactor this to avoid higher selector overrides is going to lead to higher future tech debt  than delivering the mark-up we expect to the browser :)

NataliaLKB: Thanks @gtrufitt! Updated now.

gtrufitt: Ace!

gtrufitt:
👍

PR #13135, opened 2016-05-31
Don't add MPU under a thrasher

regiskuckaertz:
On mobile, we must not add an MPU right under a thrasher.

Before:
picture 2

After:
picture 1

cc @JonNorman @kenlim @rich-nguyen


JonNorman:
:+1:

PR #13137, opened 2016-05-31
Use option for getInventoryTargeting and getTargetedAdUnits api calls

rich-nguyen:
Fix for null pointer exception in admin.


kenlim:
👍

PR #13138, opened 2016-05-31
Run welcome header A/B test again

marialivia16:
## What does this change?

The result of the previous run of this test was very positive, but we found a 15% difference between the variant and the control group.
As a result we'll be running this test again with a 0.6 offset for 6 days.
## Request for comment

@stephanfowler @harrysalmon


harrysalmon:
Let me check what's up the "CleverFriendBrexit" test

PR #13128, opened 2016-05-31
Lifecycle batch 2

alexduf:
This PR is the second part, here [is part one](https://github.com/guardian/frontend/pull/13115)

PR #13139, opened 2016-05-31
Lifecycle batch 2 fixed

alexduf:
## What does this change?

This is following the first batch of lifecycle modifications. It is the same PR as #13128 but with an extra commit fixing the healthcheck issues.
## Request for comment

@TBonnin @johnduffell


johnduffell:
good luck!! 👍

TBonnin:
LGTM 👍

PR #13140, opened 2016-05-31
Remove latency metrics

johnduffell:
Now that we know the panic shedding is working ok, I have removed the latency metrics and replaced with a single metric logged when latency is too high (1 second) and a request comes in.

This can be used for scaling - if it's recorded over 2 minutes it's probably time to scale up.  At the moment the (quick) 500 errors are reducing the latency at the ELB and stopping us scaling correctly.

I'm sure there are better ways of scaling, but we don't want to solve everything in one go.

@JustinPinner simpler for now, any thoughts?


JustinPinner:
Looking good. Still have to be careful not to fall down that rabbit hole of writing your own ELB code! 👍

PR #13141, opened 2016-05-31
Adding icon, background and theme color

dominickendrick:
## What does this change?

Updates our manifest file to be inline with Google App Manifest guidelines.

@desbo @OliverJAsh


desbo:
👍 
 
`display: standalone` might be pretty cool eventually

PR #13142, opened 2016-05-31
Fix R2 interactive page presser

JustinPinner:
## What does this change?

Creates a new page presser for re-pressing R2 interactives on the /ng-interactive/ path. Optionally convert all outgoing paths to protocol-relative.

Also enables conversion to protocol-relative paths for all pressed types.

Currently these interactives don't serve correctly. After re-pressing they should at least work, even though they'll still use the old R2 styling. As a bonus, all pressed pages can be re-pressed and made ready for https using a checkbox on the pressing form;
screen shot 2016-06-01 at 15 59 48

No
## Screenshots

**Before:**
screen shot 2016-05-31 at 18 14 01

**After:**
screen shot 2016-05-31 at 18 10 13


admin/app/pagepresser/NextGenInteractiveHtmlCleaner.scala
```diff
@@ -0,0 +1,24 @@
...
+object NextGenInteractiveHtmlCleaner extends HtmlCleaner with implicits.WSRequests {
+
+  override def canClean(document: Document): Boolean = {
+    document.getElementById("interactive-content") != null &&
+      document.getElementsByAttributeValue("rel","canonical").attr("href").toLowerCase.contains("/ng-interactive/")
+  }
+
+  override def clean(document: Document, convertToHttps: Boolean) = {
+    universalClean(document)
+    //removeScripts(document)
```
johnduffell: delete completely?

johnduffell:
looks good! 👍 r2 will live on in spirit 😄

admin/app/pagepresser/NextGenInteractiveHtmlCleaner.scala
JustinPinner: I thought it made it a little more obvious what the main difference is between this and the other interactive cleaner. Might also be able to revisit once we've got a few working examples. But could go for now I guess.

johnduffell: ok change to an actual comment in that case ;)

PR #13146, opened 2016-06-01
Update frontend-notifications switches.

janua:
Push these switches back to end start of July.

@NathanielBennett @samdesborough


desbo:
👍

PR #13150, opened 2016-06-01
No ads unless certain

jamesgorrie:
## What does this change?

Use the ajax URL for ajax requests, we missed this when we first did it. It's also creating havoc with https.

Videos read this URL from the ajax URL.

Nope
## Request for comment

@akash1810


akash1810:
👍

PR #13151, opened 2016-06-01
Add try/catch around localstorage calls

davidfurey:
## What does this change?

Adds try/catch around localstorage calls in the new user ad-free test

Stops javascript crashing in Safari private browsing mode

no
## Request for comment

@piuccio @jbreckmckye


piuccio:
👍

PR #13152, opened 2016-06-01
remember to put cross origin request in video info

jamesgorrie:
## What does this change?

I forgot to specify that it's a crossOrigin request, which it isn't on dev.

It works.

No
## Request for comment

@akash1810


akash1810:
d'oh! 👍

PR #13153, opened 2016-06-01
Remove play on fronts test

jamesgorrie:
## What does this change?

Completely remove the test that @gklopper so kindly extended for me.

Makes grants mood from really bad to bad.

No
## Request for comment

@akash1810


akash1810:
👍

PR #13155, opened 2016-06-01
Facia shouldn't 503 when a collection is Not found

johnduffell:
I want to add something to scale (and alert probably) on 5xx errors.
The reason I want to scale on 5xx errors is the panic shedding functionality reports 5xx errors when the server is overloaded rather than just going slow and falling over.  This reduces the average latency measured by the ELB, which means it stops us autoscaling.

At the moment if you get /container/int-alpha/features/feature-stories.json you get a 503 error.  This turns out it's because the collection is not found.  This is adding a lot of noise (120 a minute) to the 5xxs and it should probably be a 404 not found.

This PR changes it to be a 404.

I've also updated the frontend client side code to grab the right container id when it's the international edition, thanks to @piuccio for helping out there!

### DIFF
--- a/facia/app/controllers/FaciaController.scala
+++ b/facia/app/controllers/FaciaController.scala
@@ -214,8 +214,8 @@ trait FaciaController extends Controller with Logging with ExecutionContexts wit
           if (request.isJson)
             Cached(60) {JsonCollection(html)}
           else
-            Cached(60)(WithoutRevalidationResult(NotFound))
-      }.getOrElse(ServiceUnavailable)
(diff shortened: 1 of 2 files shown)


piuccio:
makes perfect sense 👍

JustinPinner:
4️⃣0️⃣4️⃣ is the way to go! 👍

johnduffell:
📉 
image

johnduffell:
I did cheat becuase I turned the 503 to a 404 at the same time I removed it, here's the 404s graph
image

johnduffell:
still getting a generally higher level of 404s than before but generally calm again 
image

PR #13157, opened 2016-06-01
Enable Clever Friend for all users and start tailoring test

rtyley:
## What does this change?

Enable Clever Friend for all users (rather than restricting it to 10% of users) - and test tailoring for 50% of users. So the Explainer will start showing content depending on how many `politics/eu-referendum` articles the user has read.

We're hoping to see better user engagement with the tailored content.

No
## Screenshots

image
## Request for comment

@joelochlann @SiAdcock

### DIFF
--- a/static/src/javascripts/projects/common/modules/experiments/tests/clever-friend-brexit.js
+++ b/static/src/javascripts/projects/common/modules/experiments/tests/clever-friend-brexit.js
@@ -1,31 +1,48 @@
 define([
 
 ], function() {
+    function noop() {}
+
+    function companionInteractives() {
+        return document.querySelectorAll('figure[data-canonical-url^="https://interactive.guim.co.uk/2016/05/brexit-companion/"]');
+    }
+
     return function () {
         this.id = 'CleverFriendBrexit';
-        this.start = '2016-05-09';
+        this.start = '2016-06-01';
         this.expiry = '2016-07-31';
-        this.author = 'Anne Byrne';
-        this.description = 'Only expose the clever friend embed to 10% of people, by removing it for 90%';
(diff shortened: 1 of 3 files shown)


static/src/javascripts/projects/common/modules/experiments/tests/clever-friend-brexit.js
```diff
@@ -1,31 +1,48 @@
...
 
 ], function() {
+    function noop() {}
+
+    function companionInteractives() {
+        return document.querySelectorAll('figure[data-canonical-url^="https://interactive.guim.co.uk/2016/05/brexit-companion/"]');
+    }
+
     return function () {
         this.id = 'CleverFriendBrexit';
```
johnduffell: I think you need to change the id, otherwise people who were already put into buckets will stay there so you will get 50% of users who didn't come to the site since you started the test (and the existing 90% from before)

joelochlann:
If we do decide to run a tailoring test against the 10% of people who do see the atom, we could potentially leave this test intact and create a separate one with the audience offset at 0.9 and an audience of 0.1. Then we'd just be splitting the final 10% who aren't in the "test" which removes the atom.

static/src/javascripts/projects/common/modules/experiments/tests/clever-friend-brexit.js
rtyley: Good catch! Addressed in ca7a5b6

gtrufitt:
Getting a bit old this one, is it stale or waiting on something?

rtyley:
I think we can close this one for now.

gtrufitt:
😎

PR #13159, opened 2016-06-01
Krux key value report

rich-nguyen:
Adds another admin commercial report page to show the DFP key values and their internal dfp ids.


JonNorman:
Looks good :+1:

PR #13161, opened 2016-06-01
fix video teaser canRun

akash1810:
## What does this change?

bye bye `$`

`canRun` was always returning false and so the baseline wasn't run... use `document.getElementByClassName` because it the alternative was returning `undefined`

baselining a test

no
## Screenshots

n/a
## Request for comment

@jamesgorrie @harrysalmon


jamesgorrie:
👍

PR #13162, opened 2016-06-01
Video carousel lazy load images

jamesgorrie:
## What does this change?

I've added lazy loading to the video carousel as, on occasion, we can have 5-10 items that bring all their images with - which is a lot.

This container is now on the UK front, so definitely needs trimming, this was the first step. @paperboyo knows exactly how much.

The UK front, and any other lucky front to container this container is slimmer and users have a faster experience.

No
## Request for comment

@akash1810 @sammorrisdesign @paperboyo


static/src/javascripts/projects/common/modules/video/video-container.js
```diff
@@ -67,6 +76,47 @@ define([
         });
     }
 
+    // iniLazyLoadImages is used for when a user scrolls i.e. thin viewport
+    // and fetchImage is used for when the carousel controls are used i.e. fat viewport
+    function initLazyLoadImages() {
```
jamesgorrie: If I could find a way to do `element.on('transform')` we wouldn't have to have the two different ways of doing this.

Or we could have a way of [`element.scrollIntoView({ behaviour: 'smooth' })`](https://developer.mozilla.org/en/docs/Web/API/Element/scrollIntoView#Parameters). 

Writing either of those seemed harder than this.

static/src/javascripts/projects/common/modules/video/video-container.js
```diff
@@ -36,6 +44,7 @@ define([
         $('.video-playlist__item--active').removeClass('video-playlist__item--active');
 
         if (direction === 'next') {
+            fetchImage(containerPos + preloadImageCount);
```
akash1810: worth wrapping this in a promise?

jamesgorrie: What for?

akash1810: to advance only when the image has loaded

jamesgorrie: Na, just do it at the same time... on a slow connection pressing next, then waiting (with no load), to then be scrolled, I would be pissed. Probably best to add a loading image, but I can do that after.

If we went with the next track paradigm, you could definitely do loading, but you would actually be loading the video, which to me would make sense and actually eradicate the need for holding images.

paperboyo: > pressing next, then waiting (with no load), to then be scrolled, I would be pissed
+1

static/src/javascripts/projects/common/utils/element-inview.js
```diff
@@ -0,0 +1,64 @@
+define([
+    'bean',
+    'common/utils/$',
```
akash1810: we don't appear to be using `$` `mediator` or `debounce` so they can be removed

akash1810:
👍

static/src/javascripts/projects/common/utils/element-inview.js
jamesgorrie: We use `debounce` - good catch though.

PR #13163, opened 2016-06-01
Run the custom targeting key-value job on startup

rich-nguyen:
I forgot to add the new admin job to the startup lifecycle.


JonNorman:
👍

PR #13164, opened 2016-06-01
fix video embed page

akash1810:
## What does this change?

Update video-embed to use `videojsOptions` which sets default options. Previously, we were hiding some elements using css, however using `videojsOptions` means they are not even added to the DOM (which should also increase performance).

better rendering of content

FIA
## Screenshots

from
screen shot 2016-06-01 at 18 15 11

to
screen shot 2016-06-01 at 18 15 17
## Request for comment

@jamesgorrie @mchv


mchv:
👍  great investigation work!

PR #13165, opened 2016-06-01
remove questions when quiz has a result

johnduffell:
when people do personality quizzes, it shows a submit button and posts back, then showing the answers you put in with the result below.  This confuses people because they hit submit and expected to see their result not the questions they answered.

I was going to add an anchor to jump to, but you can't really do that with form actions (according to google) so I've just made it hide the questions when someone has a result.  There don't seem to be any personality quizzes done on the main site so far, so it's not going to break existing stuff as far as I can tell!

@rich-nguyen not quite what we discussed but probably ok. @sndrs just tagging you so you see this when you're around next


johnduffell:
ok closing this because it's going to end up inconsistent with apps and personality quizzes are so rarely used it doesn't make sense to do things with them

PR #13166, opened 2016-06-01
Lifecycle batch 3

alexduf:
## What does this change?

Last bit of the lifecycle changes. It brings the project in a state where no lifecycle is triggered through the cake-pattern on the Global object. (yay!)

Same as batch 1 and batch 2

human sized PR => path to play 2.5
## Request for comment

@TBonnin @johnduffell


TBonnin:
looks good to me 👍  💃

PR #13167, opened 2016-06-01
Just adding the link to the title too

kenlim:
Because single char links are really hard to click on.


jamesgorrie:
They sure are 👍

PR #13170, opened 2016-06-02
Fix newRequestData bug

JonNorman:
We have the following error in prod:

image

This PR fixes that error - when `userFeatures` is called, no object is being passed in; now we pass in `UserFeatures` so that `this` references correctly.

Thanks for @regiskuckaertz for working that out!

### DIFF
--- a/static/src/javascripts/bootstraps/enhanced/common.js
+++ b/static/src/javascripts/bootstraps/enhanced/common.js
@@ -420,7 +420,7 @@ define([
                 ['c-save-for-later', modules.saveForLater],
                 ['c-show-membership-messages', modules.showMembershipMessages],
                 ['c-email', modules.initEmail],
-                ['c-user-features', userFeatures.refresh],
+                ['c-user-features', userFeatures.refresh, userFeatures],
                 ['c-headlines-test-analytics', modules.headlinesTestAnalytics],
                 ['c-hosted-about-lightbox', modules.initHostedAboutLightbox]
             ]), function (fn) {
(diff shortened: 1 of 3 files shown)


regiskuckaertz:
After a second read, I'm genuinely surprised it's working as the `this` value should not be carried over in the `catchErrors` call. Might also be worth amending line 31:

```
var error = catchErrors(fn.bind(this));
```

JonNorman:
After adding the `binding` parameter to our `catchErrorsAndLog`, I looked to see whether any calls to `log` or `catchErrorsAndLog` actually provide a `reporter`; none do, apart from `robust.spec.js`, so I've removed that.

static/src/javascripts/bootstraps/enhanced/common.js
```diff
@@ -420,7 +420,7 @@ define([
                 ['c-save-for-later', modules.saveForLater],
                 ['c-show-membership-messages', modules.showMembershipMessages],
                 ['c-email', modules.initEmail],
-                ['c-user-features', userFeatures.refresh],
+                ['c-user-features', userFeatures.refresh, userFeatures],
```
jbreckmckye: Actually, come to think of it...

Could we not just replace this whole PR with -

```javascript
['c-user-features', userFeatures.refresh.bind(userFeatures)],
```

...?

JonNorman: That does seem far more straightforward and sensible!

JonNorman:
Closing this PR, will open a new one with fix suggested by @jbreckmckye (and will leave removal of `reporter` for a different PR)

PR #13171, opened 2016-06-02
Hosted by page - fix css imports

lps88:
## What does this change?

We had the hosted css imported in the head.content.scss file, it's now in a commercial-specific one.

We were also importing it in _commercial.scss which meant the styles were duplicated.

Some of the selectors had to be made stronger, as they appear in a different order now.

Also changed text positioning in the hosted CTA banner (at the bottom) and corrected borders in mobile layout (see images)
## Screenshots

image

image
## Request for comment

@Calanthe @regiskuckaertz


static/src/stylesheets/module/commercial/glabs/_hosted.scss
```diff
@@ -106,7 +106,7 @@ $renaultLight: #fff9e9;
     color: rgba(255, 255, 255, .5);
 }
 
-.hosted__container {
+section.hosted__container {
```
regiskuckaertz: I'm curious, why this?

regiskuckaertz:
Looks good to me, although if specificity is a problem it may be worth investigating what is causing it and work from there down. 👍

PR #13172, opened 2016-06-02
Removes all breaks in gallery captions

NataliaLKB:
After a change I made yesterday in #13134 this started happening:

image
(notice the extra space between the caption title and the caption body)

This is because in composer you have a free text field for gallery captions. This has resulted in inconsistencies in the html we receive for them. Yesterday we started cleaning the html to make it a bit better. 

The best solution though would be to get a caption title and a caption body from composer though, rather than just a caption. I talked to them yesterday, but will follow up today!

With this PR, we only have one break between the title and the body like so:
image

Nope 😄 
## Request for comment

@gtrufitt


gtrufitt:
👍 


👍

PR #13173, opened 2016-06-02
Rich links clashing with inline ads

regiskuckaertz:
This shouldn't happen, right?
picture 3

Well it did. Automated rich links are loaded in by `enhanced`, so they should load _after_ inline ads have loaded; well, it didn't, and it was the combination of several problems:
- adverts were loading too late because the promise around `dfp.init` was resolving once DFP had exhausted its command queue—we don't need to wait that long and can resolve as soon as the queue is filled
- automated rich links did not wait for ads to load before running—therefore nuking any spacefinder rule related to said ads
- the `dfp:rendered` events were dispatched much too soon, before the creative had any chance of being rendered—therefore nuking any chance of getting their geometry right when calling `getBoundingClientRect`
- the algorithm inserting inline ads was using a recursive promise chain on `spaceFiller`! That doesn't seem problematic, but as promises are asynchronous, there is a chance another module inserts a task in the space-filler queue between the algorithm dies—and by _another module_ I really mean rich links. This turned out as a kind of race condition where the latter would wait for ads from the former to load, but they didn't because they queued after; hence a timeout would let rich links load without waiting anymore.

So, to get around all this, I
- reorganized in which order and how fast commercial modules load
- added the ability to wait for ads in the spacefinder, which rich links could then use
- dispatched the `dfp:rendered` event once `parseAd` resolved
- replaced the recursive algorithm with a single-pass call to space-filler, using the much more efficient `filter` option

And _voila_
picture 4

cc @JonNorman @rich-nguyen @Calanthe @lps88 @gtrufitt


static/src/javascripts/projects/common/modules/article/spacefinder.js
```diff
@@ -1,19 +1,21 @@
...
     bean,
     Promise,
+    fastdom,
     config,
     mediator,
-    curry
+    trackAdLoaded,
+    memoize
 ) {
     // total_hours_spent_maintaining_this = 60
```
gtrufitt: You didn't update this.

regiskuckaertz: 😆  will do!

regiskuckaertz: done. although in all fairness to the spacefinder, the biggest problem lied in article-body-adverts 😜

static/src/javascripts/projects/common/modules/article/spacefinder.js
```diff
@@ -1,21 +1,23 @@
...
     Promise,
+    fastdom,
     config,
     mediator,
-    curry
+    trackAdLoaded,
+    memoize
 ) {
-    // total_hours_spent_maintaining_this = 60
+    // total_hours_spent_maintaining_this = 64
```
rich-nguyen: !!!

rich-nguyen:
the code looks fine. I will spend the rest of the day looking at spacefiller to understand the behaviour, though! Good friday reading.

👍

regiskuckaertz:
Yes, that is one piece of code! Once you understand what `minAbove` and `minBelow` stand for, you have most of the job done. I can walk you through it if sth is not clear.

PR #13174, opened 2016-06-02
Add switch owner and improve format of the Expiring switches email

TBonnin:
## What does this change?

Add the concept of switch owner as well as improving the formatting of the expiring switch email

I first thought about using email addresses as switch owners, however because this is a public repo we cannot expose emails directly. It would then require to keep a map of name -> email in the config. That might be an overkill.
I gave up on this idea then and only add the concept of owner, which can be identify by a name (ex: `health team`) and/or a github account)
I used `git blame` to assign already existing switches. I might have made some mistakes (:wink: @johnduffell)
Possible improvement: when a switch is about to expire, automatically open a github issue and assign it to the owner. :)

See the screenshot below for the change regarding formatting

Let me know if you think this could be useful. Thank you

Easier to spot which switch is about to expire
Easier to spot who is responsible for a given switch that is about to expire

No
## Screenshots

screen shot 2016-06-02 at 12 16 40
## Request for comment

@johnduffell @regiskuckaertz @jamespamplin @JustinPinner @alexduf @rich-nguyen

### DIFF
--- a/admin/app/views/email/expiringSwitches.scala.html
+++ b/admin/app/views/email/expiringSwitches.scala.html
@@ -1,25 +1,54 @@
-@(expiringImminently: Seq[conf.switches.Switch], expiringSoon: Seq[conf.switches.Switch])
+@import conf.switches.Switch
+@import _root_.jobs.ExpiringSwitches
 
-@showSwitch(switch: conf.switches.Switch) = {
-    <li>
-        <div>@{switch.name}</div>
-        <div>expires @{switch.sellByDate.get.toString("E dd MMM")} 23:59 UTC</div>
+@(switches: Seq[Switch])
+
+@showSwitch(switch: conf.switches.Switch, showExpirationDate: Boolean) = {
+    <li style="padding-bottom: 6px">
+        <div>
+            <strong>@{switch.name}</strong>
+            - @showOwners(switch.owners)
+            @if(showExpirationDate) { - <span style="color:grey;font-style: italic;">expires @{switch.sellByDate.get.toString("E dd MMM")} at 23:59 (London time)</span>}
+        </div>
         <div>@{switch.description}</div>
     </li>
 }
 
-@if(expiringImminently.nonEmpty) {
-    <div>
-        <p style="color:red">These switches are going to expire imminently!</p>
-        <ul>@for(switch <- expiringImminently) {@showSwitch(switch)}</ul>
-    </div>
+@emergencyColor(index: Int) = @{
+    index match {
+        case 0 => "red"
+        case 1 => "orangeRed"
+        case 2 => "orange"
+        case _ => "black"
+    }
 }
 
-@if(expiringSoon.nonEmpty) {
-    <div>
-        <p style="color:orange">These switches are going to expire soon:</p>
-        <ul>@for(switch <- expiringSoon) {@showSwitch(switch)}</ul>
-    </div>
+@showOwners(owners: Seq[conf.switches.Owner]) = {
+    Owners: @for(owner <- owners) { @showOwner(owner) }
+}
+@showOwner(owner: conf.switches.Owner) = {
+    @if(owner.github.isEmpty) {
+       @owner.name
+    } else {
+        <a href="https://github.com/@{owner.github.get}">@{owner.name.getOrElse(owner.github)}</a>
+    }
+
 }
 
-<p>See https://frontend.gutools.co.uk/dev/switchboard</p>
+<html>
+    <head></head>
+    <body>
+        @defining(ExpiringSwitches(switches).groupByPriority) { allGroups =>
+            @for(group <- allGroups) {
+                @if(group.switches.nonEmpty) {
+                    <div>
+                        <h1 style="font-size:1.3em;color:@emergencyColor(allGroups.indexOf(group))">@group.description.capitalize</h1>
+                        <ul>@for(switch <- group.switches) {@showSwitch(switch, (allGroups.last == group))}</ul>
+                    </div>
+                }
+            }
+        }
+
+    <p>See https://frontend.gutools.co.uk/dev/switchboard</p>
+    </body>
+</html>
--- a/admin/app/views/email/expiringSwitches.scala.html
+++ b/admin/app/views/email/expiringSwitches.scala.html
@@ -27,12 +27,15 @@
     Owners: @for(owner <- owners) { @showOwner(owner) }
 }
 @showOwner(owner: conf.switches.Owner) = {
-    @if(owner.github.isEmpty) {
-       @owner.name
+    @if(!owner.github.isEmpty) {
+       <a href="https://github.com/@{owner.github.get}">@{owner.name.getOrElse(owner.github)}</a>
     } else {
-        <a href="https://github.com/@{owner.github.get}">@{owner.name.getOrElse(owner.github)}</a>
+        @if(!owner.email.isEmpty) {
+            <a href="mailto:@{owner.email.get}">@{owner.name.getOrElse(owner.email)}</a>
+        } else {
+            @owner.name
+        }
     }
-
 }
 
 <html>
--- a/common/app/conf/switches/CommercialSwitches.scala
+++ b/common/app/conf/switches/CommercialSwitches.scala
@@ -9,24 +9,27 @@ trait CommercialSwitches {
     SwitchGroup.Commercial,
     "dfp-caching",
     "Have Admin will poll DFP to precache adserving data.",
+    owners = Seq(Owner.withName("commercial team")),
     safeState = On,
     sellByDate = never,
     exposeClientSide = false
   )
 
   val HeaderBiddingUS = Switch(
-     SwitchGroup.Commercial,
-     "header-bidding-us",
-     "Auction adverts on the client before calling DFP (US edition only)",
-     safeState = Off,
-     sellByDate = never,
-     exposeClientSide = true
+    SwitchGroup.Commercial,
+    "header-bidding-us",
+    "Auction adverts on the client before calling DFP (US edition only)",
+    owners = Seq(Owner.withGithub("regiskuckaertz ")),
+    safeState = Off,
+    sellByDate = never,
+    exposeClientSide = true
   )
 
   val CommercialSwitch = Switch(
     SwitchGroup.Commercial,
     "commercial",
     "If this switch is OFF, no calls will be made to the ad server. BEWARE!",
+    owners = Seq(Owner.withName("commercial team")),
     safeState = On,
     sellByDate = never,
     exposeClientSide = true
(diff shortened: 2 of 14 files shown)


common/app/conf/switches/CommercialSwitches.scala
```diff
@@ -9,24 +9,27 @@ trait CommercialSwitches {
...
-     safeState = Off,
-     sellByDate = never,
-     exposeClientSide = true
+    SwitchGroup.Commercial,
+    "header-bidding-us",
+    "Auction adverts on the client before calling DFP (US edition only)",
+    owners = Seq(Owner.withGithub("regiskuckaertz ")),
+    safeState = Off,
+    sellByDate = never,
+    exposeClientSide = true
```
TBonnin: Removed extra space :(

common/app/conf/switches/Switches.scala
```diff
@@ -118,7 +128,7 @@ object Switch {
 
   case class Expiry(daysToExpiry: Option[Int], expiresSoon: Boolean, hasExpired: Boolean)
 
-  def expiry(switch: Switch, today: LocalDate = new DateTime().toLocalDate) = {
+  def expiry(switch: Switch, today: LocalDate = new DateTime(DateTimeZone.forID("Europe/London")).toLocalDate) = { // We assume expiration datetime is set to London time
```
TBonnin: Making the timezone explicit 😬

admin/app/views/email/expiringSwitches.scala.html
```diff
@@ -1,25 +1,54 @@
...
+@(switches: Seq[Switch])
+
+@showSwitch(switch: conf.switches.Switch, showExpirationDate: Boolean) = {
+    <li style="padding-bottom: 6px">
+        <div>
+            <strong>@{switch.name}</strong>
+            - @showOwners(switch.owners)
+            @if(showExpirationDate) { - <span style="color:grey;font-style: italic;">expires @{switch.sellByDate.get.toString("E dd MMM")} at 23:59 (London time)</span>}
+        </div>
         <div>@{switch.description}</div>
```
jamesgorrie: Is this needed? I know it's not part of the PR - but seems like excess information when all we want to know is who and when?

jamesgorrie:
I like the idea of multiple owners, or even team owners. Great.

The definition of "about to expire" is quite integral here, as I have written tests that I expect to expire tomorrow, but I leave till the end of the day for data collections (but have also forgotten :rage4:). 

In this instance we could say that you should always have your A/B test switches expire a day late, or "about to expire" means 4pm before the 12pm deadline is hit - which goes back to someone's point of making the switches expire on the day (i.e. before 4ish).

The github issue generator is great, but it relies on people's workflow taking assigned issues as urgent (as with the email).

This on a dashboard would be great as you can catch it before it expires (i.e. I walk past at 4pm and see my (or someone else's) switch is expiring) and fix it.

👍

common/app/conf/switches/CommercialSwitches.scala
```diff
@@ -234,6 +259,7 @@ trait CommercialSwitches {
     SwitchGroup.Commercial,
     "krux-video-tracking",
     "If this switch is ON, there will be a Krux pixel fired to track particular videos",
+    owners = Seq(Owner.withGithub("johnduffell")),
```
johnduffell: commercial team

common/app/conf/switches/CommercialSwitches.scala
```diff
@@ -225,6 +249,7 @@ trait CommercialSwitches {
     SwitchGroup.Commercial,
     "adblock",
     "Switch for the Adblock Message.",
+    owners = Seq(Owner.withGithub("johnduffell")),
```
johnduffell: commercial team

common/app/conf/switches/MonitoringSwitches.scala
```diff
@@ -10,6 +10,7 @@ trait MonitoringSwitches {
     SwitchGroup.Monitoring,
     "offline-page-view",
     "If this switch is on, views of the offline page will be beaconed to cloudwatch when you come online again",
+    owners = Seq(Owner.withGithub("sndrs")),
     safeState = Off,
     sellByDate = new LocalDate(2016, 6, 3),
```
johnduffell: ooh good spot - @sndrs won't be able to do anything with that, we should extend until next wednesdayish!

common/app/conf/switches/PerformanceSwitches.scala
```diff
@@ -55,6 +60,7 @@ trait PerformanceSwitches {
     SwitchGroup.Performance,
     "server-side-buckets",
     "When this switch expires, remove the remaining predefined server side testing buckets",
+    owners = Seq(Owner.withGithub("johnduffell")),
     safeState = Off,
     sellByDate = new LocalDate(2016, 6, 22),
```
johnduffell: this can go away completely

common/app/conf/switches/MonitoringSwitches.scala
```diff
@@ -37,6 +40,7 @@ trait MonitoringSwitches {
     SwitchGroup.Monitoring,
     "google-analytics",
     "If this switch is on, then Google Analytics is enabled",
+    owners = Seq(Owner.withGithub("Grant Klopper")),
```
johnduffell: should be github name or id?

admin/app/views/email/expiringSwitches.scala.html
TBonnin: I feel like it is sometimes useful to have additional info, especially when the switch name is not very explicit. If a majority of people think it is useless I am happy to remove it though 😁

common/app/conf/switches/MonitoringSwitches.scala
TBonnin: github id. Will fix

common/app/conf/switches/MonitoringSwitches.scala
TBonnin: I will extend to next week

TBonnin:
@jamesgorrie The notion of expiring is actually quite confusing since it doesn't have impact on the value of the switch (ie: expired switch keep its set value), even though when switches where introduced switches used to go off when expired.
Currently the expiration date is more like a reminder for devs to re-evaluate the switch usefulness.

@johnduffell Having a list of owner gives you the flexibility to assign both a team and an individual (ex: `johnduffell`, `health team`)
In practice we should encourage team ownership in most cases

jamesgorrie:
@TBonnin 100% agree, dead code should not exist in the codebase.

I am more wondering what the most effective way of notifying people is, before breaking the build, because that is very much the most effective way.

admin/app/views/email/expiringSwitches.scala.html
regiskuckaertz: But probably the info hierarchy could be tweaked so that the description has less visual weight than the expiry date.

regiskuckaertz:
@johnduffell actually we thought you'd make a great choice for all the commercial switches 😜

johnduffell:
🎉 🚤 I'm happy anyway! 👍

TBonnin:
Thank you all.
I am off for a week starting tonight. Keep the feedback coming. I will tweak/merge this PR when I am back.

admin/app/views/email/expiringSwitches.scala.html
gklopper: You might appreciate the description while trying to figure out which switch to turn off in the middle of the night (I certainly do)

common/app/conf/switches/CommercialSwitches.scala
```diff
@@ -9,24 +9,27 @@ trait CommercialSwitches {
...
-     "Auction adverts on the client before calling DFP (US edition only)",
-     safeState = Off,
-     sellByDate = never,
-     exposeClientSide = true
+    SwitchGroup.Commercial,
+    "header-bidding-us",
+    "Auction adverts on the client before calling DFP (US edition only)",
+    owners = Seq(Owner.withGithub("regiskuckaertz ")),
+    safeState = Off,
+    sellByDate = never,
```
gklopper: Switches that never expire mean something (and are probably overused). The default behaviour should probably be to delete the switch (but leave the code in place) apart from a few switches that are used in panic situations.

If you plan on switching something off it should have a sensible `sellByDate`.

Maybe create a `NeverExpiresSwitch(...)` instead of `sellByDate = never` and at most these should belong to a team and not an individual.

johnduffell: good point about the overused thing, we even have a reminder from the venerable nick haley about death by switches and it's becoming true.  The expansion of the switches is probably a proxy for the expansion of the codebase in general though.  That's another PR though.

Also, never used to defined as 31st december 2100 I think, so it was just about possible!  I like the NeverExpiresSwitch idea.

common/app/conf/switches/CommercialSwitches.scala
alexduf: you can look at the diff [without the spaces](https://github.com/guardian/frontend/pull/13174/files?w=1)

common/app/conf/switches/CommercialSwitches.scala
regiskuckaertz: Indeed a few weeks ago I thought I'd go through all the commercial switches and remove all those without an expiration date, except perhaps the master `commercial` switch.

sndrs: maybe these are basically `SafetySwitch`es? like discussion etc, there for emergencies? they could have no sell-by date, while all others must?

alexduf: I feel like high-level switches like that (disable commercial or discussion) should be handled by a circuit breaker (and they probably are already). If we want to completely turn off comments on the website, setting the ASG size to 0 should be enough?
That being said, maybe there are features that aren't mapped to a microservice and that would need a flag as you're suggesting.

kelvin-chappell:
Good idea.  But it might work better at the [switch group](https://github.com/guardian/frontend/blob/master/common/app/conf/switches/Switches.scala#L16) level.

Perhaps the switches should be reorganised so that each switch group belongs to a team, and that team is shown as the owner of the switch group in the switchboard.  Then the email would show expiring switches arranged by group.

TBonnin:
@kelvin-chappell Good idea. The problem is that in some case the functional grouping of switches (like we have now) matches perfectly with the concept of team ownership (ex: commercial or identity) but for the other groups it is far from being the case. 
Also while we should encourage switch to belong to a team, I can think of a lot of cases where a temporary switch might be created/owned by an individual or small group of individuals with no formal team name :)

kelvin-chappell:
I see what you mean, but this seems like an edge case.  You could always create a temporary switch group owned by an individual if there was really no defined team that it seemed to belong to.  But it has a bad smell.

TBonnin:
@kelvin-chappell Exactly. The solution described in this PR has the advantage to give full flexibility in term of what an owner is but at the expense of an extra line for each single switch.

kelvin-chappell:
👍

TBonnin:
Thank you all for your feedback. There are obviously a lot more we could do, especially about switch that never expires. I am gonna merge this PR though and we can keep discussing improvements.

PR #13175, opened 2016-06-02
show video controls on videos with modifier set

jamesgorrie:
## What does this change?

I noticed we had the class `.gu-media--show-controls-at-start` a while back, [ran a test on putting it back](https://github.com/guardian/frontend/pull/13003) which proved to increase plays significantly, so we're adding it back.

People notice, thus watch more video.

No
## Screenshots

screencapture
## Request for comment

@akash1810 @blongden73

### DIFF
--- a/static/src/stylesheets/module/content/_media-player.scss
+++ b/static/src/stylesheets/module/content/_media-player.scss
@@ -1105,3 +1105,22 @@ $ima-controls-height: 70px;
 .vjs-modal-dialog-description {
     display: none;
 }
+
+// We have to use vjs classes here, and not BEM modifiers as videojs
+// changes the classes names via JS, and last time we messed with that,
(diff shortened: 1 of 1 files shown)


akash1810:
👍 

can we get rid of the grey, unhovered state of the big play button now too then?

jamesgorrie:
No, as this is only on main media and video pages. Fronts will still have said grey.
Write a test to prove people want it.

paperboyo:
> can we get rid of the grey, unhovered state of the big play button now too then

But how to get rid of it on front's cards without making the button-on-my-face problem much worse?

akash1810:
> without making the button-on-my-face problem much worse

If the test proves an increase in plays, I'd argue the button-on-my-face problem doesn't exist for readers...

PR #13176, opened 2016-06-02
DFP commercial line items should be validated against active adunits

rich-nguyen:
This fixes an issue where line items targeting inactive/archived ad units were being rejected from our system. 

The correct behaviour is to reject unknown adunits, instead of rejecting adunits which are not active.

Note to self: This will invalidate the format for: line items report, high merch report, and paid for tags report. In the long term, its better to do this invalidation than use an `Option` for something that will always be there.

### DIFF
--- a/admin/app/dfp/AdUnitAgent.scala
+++ b/admin/app/dfp/AdUnitAgent.scala
@@ -1,7 +1,6 @@
 package dfp
 
 import com.google.api.ads.dfp.axis.utils.v201508.StatementBuilder
-import com.google.api.ads.dfp.axis.v201508.InventoryStatus
 import common.dfp.GuAdUnit
 import conf.Configuration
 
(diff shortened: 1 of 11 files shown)


admin/app/dfp/AdUnitAgent.scala
```diff
@@ -41,8 +38,27 @@ object AdUnitAgent extends DataAgent[String, GuAdUnit] {
 
 object AdUnitService {
 
+  // Retrieves the ad unit object if the id matches and the ad unit is active.
```
JonNorman: Would it be worth calling this `activeAdUnit` just to be explicit (and complete with `archivedAdUnit` to make it clear that they are non-intersecting sets)?

admin/app/dfp/DataValidation.scala
```diff
@@ -16,7 +16,12 @@ object DataValidation {
         toSeq(inventoryTargeting.getTargetedAdUnits).map(_.getAdUnitId()
       )).getOrElse(Nil)
 
-    dfpAdUnitIds.forall( adUnitId => {
+    // The validation should not account for inactive or archived ad units.
+    val activeDfpAdUnits = dfpAdUnitIds.filterNot { adUnit =>
```
JonNorman: Perhaps change the variant name to `adunitId` to be consistent with that below

JonNorman:
Apart from the tiny comments, looks good to me :+1:

admin/app/dfp/AdUnitAgent.scala
rich-nguyen: yeah good idea

PR #13180, opened 2016-06-02
Robustify live events

JonNorman:
## What does this change?

The LiveEvents feed was failing to parse (and has done for about a week) as there is one badly formed LiveEvent in the feed. Now, when an element in the Events array fails the validation, it is discarded and the parsing can continue with the remaining LiveEvents in the feed.

In addition to this, there was a date filter on the LiveEvents that meant Events happening within the next two weeks couldn't be targeted. This seems odd and has been removed.

@regiskuckaertz


regiskuckaertz:
👍

PR #13181, opened 2016-06-02
Cache content for longer

gklopper:
## What does this change?

I want to slowly ramp up the time we cache content for. This pushes the vast majority of content up to 3800 seconds (over an hour) at which point our CDN writes it to disk.

This is too long to cache in the browser, so the CDN uses `Surrogate-Cache` and we limit `Cache-Control` to 60 seconds ([see docs](https://docs.fastly.com/guides/tutorials/cache-control-tutorial#surrogate-control))

This should mean a higher hit rate on our CDN (I'm hoping this pushes us to 75%) and a corresponding drop in origin CPU & hits.

It should also mean higher availability when we are in technical difficulties and (slightly) better performance for our end users (but our time to last byte is pretty good as it is).

No


gklopper:
cc @rich-nguyen

common/app/model/Cached.scala
```diff
@@ -80,19 +82,32 @@ object Cached extends implicits.Dates {
...
+    val surrogateCacheControl = s"max-age=$maxAge, stale-while-revalidate=$staleWhileRevalidateSeconds, stale-if-error=$tenDaysInSeconds"
 
-    // see
-    // http://tools.ietf.org/html/rfc5861
-    // http://www.fastly.com/blog/stale-while-revalidate
-    // http://docs.fastly.com/guides/22966608/40347813
-    val staleWhileRevalidateSeconds = math.max(maxAge / 10, 1)
-    val cacheControl = s"max-age=$maxAge, stale-while-revalidate=$staleWhileRevalidateSeconds, stale-if-error=$tenDaysInSeconds"
+    val cacheControl = if (LongCacheSwitch.isSwitchedOn) {
+      val browserMaxAge = min(maxAge, 60)
```
rich-nguyen: Why make the browser's max age smaller than the surrogate max age? Is it because this approach supports purging whilst giving longer CDN-level cache times? Also, when the system switches from long-cache = off to long-cache = on, browsers could hold on to the bigger max-age for a bit.

rich-nguyen:
👍

common/app/model/Cached.scala
gklopper: Yeah, we are moving more towards a "purge" strategy than a let it expire strategy, but there is no way for me to purge a bunch of browser caches (although our E-Tags partly solve that problem). 

>when the system switches from long-cache = off to long-cache = on, browsers could hold on to the bigger max-age for a bit.

Yeah, it does not matter

PR #13182, opened 2016-06-02
Update test page Live Event that won't expire for a while

JonNorman:
This one almost defeated me...
@regiskuckaertz


regiskuckaertz:
😭  🔫  👍

PR #13184, opened 2016-06-02
Show different message when you can't skip an ad.

jamesgorrie:
## What does this change?

Apologies for the double whammy.
- Show prerolls on video pages again as we were just resolving the promise with `shouldHideAdverts: true`, whereas we should have been looking it up.
- Only add the skip timer when the ad is longer than the time to skip.
- We get no more complaints saying how can I skip if the ad ends before I can skip.
- We serve preroll again.

No
## Request for comment

@akash1810

### DIFF
--- a/static/src/javascripts/bootstraps/enhanced/media/main.js
+++ b/static/src/javascripts/bootstraps/enhanced/media/main.js
@@ -185,6 +185,7 @@ define([
             embedPath = $el.attr('data-embed-path'),
             // we need to look up the embedPath for main media videos
             canonicalUrl = $el.attr('data-canonical-url') || (embedPath ? '/' + embedPath : null),
+            shouldHideAdverts = $el.attr('data-block-video-ads') === 'false' ? false : true,
             techPriority = techOrder(el),
             player,
             mouseMoveIdle,
@@ -197,7 +198,7 @@ define([
             // These are set to the safest defaults that will always play video.
             var defaultVideoInfo = {
                 expired: false,
-                shouldHideAdverts: true
+                shouldHideAdverts: shouldHideAdverts
             };
 
             if (!canonicalUrl) {
@@ -295,13 +296,19 @@ define([
                                 raven.wrap(
                                     { tags: { feature: 'media' } },
                                     function () {
-                                        player.adSkipCountdown(15);
+                                        var skipAdTime = 15;
                                         player.ima({
                                             id: mediaId,
                                             adTagUrl: getAdUrl(),
                                             prerollTimeout: 1000
                                         });
                                         player.ima.requestAds();
+                                        player.on('adstart', function() {
+                                            var adDuration = player.ima.getAdsManager().getCurrentAd().getDuration();
+                                            if (adDuration > skipAdTime) {
+                                                player.adSkipCountdown(skipAdTime);
+                                            }
+                                        });
 
                                         // Video analytics event.
                                         player.trigger(events.constructEventName('preroll:request', player));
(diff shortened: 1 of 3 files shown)


static/src/javascripts/bootstraps/enhanced/media/main.js
```diff
@@ -185,6 +185,7 @@ define([
             embedPath = $el.attr('data-embed-path'),
             // we need to look up the embedPath for main media videos
             canonicalUrl = $el.attr('data-canonical-url') || (embedPath ? '/' + embedPath : null),
+            shouldHideAdverts = $el.attr('data-block-video-ads') === 'false' ? false : true,
```
akash1810: this value isn't always correct, as embeds don't follow parent updates 😢 I guess its a good default value though...

static/src/javascripts/bootstraps/enhanced/media/main.js
```diff
@@ -295,13 +296,19 @@ define([
...
                                     function () {
-                                        player.adSkipCountdown(15);
+                                        var skipAdTime = 15;
                                         player.ima({
                                             id: mediaId,
                                             adTagUrl: getAdUrl(),
                                             prerollTimeout: 1000
                                         });
                                         player.ima.requestAds();
+                                        player.on('adstart', function() {
```
akash1810: should the event handler be before requesting the ad?

akash1810:
when ads are < 15 seconds long, should there be a count down? Such as "the video will play in 5 seconds"?

static/src/javascripts/bootstraps/enhanced/media/main.js
jamesgorrie: Yea, so, look it up, if you can't find it, set to true.
With embeds we look it up though.

static/src/javascripts/bootstraps/enhanced/media/main.js
jamesgorrie: I hate that that matters, but you're probably right. Changed.

jamesgorrie:
> when ads are < 15 seconds long, should there be a count down? Such as "the video will play in 5 seconds"?

There's a progress bar.

static/src/javascripts/bootstraps/enhanced/media/main.js
jamesgorrie: Order:

* has attribute? read attribute => set `shouldHideAdverts`
* has canonical URL? look it up via `info.json` => set `shouldHideAdverts`

So we should be right nearly all of the time except, like the other day, the `info.json` borks.

akash1810:
👍

jamesgorrie:
> when ads are < 15 seconds long, should there be a count down? Such as "the video will play in 5 seconds"?

@akash1810 maybe that is nicer. Updated:
timer

static/src/javascripts/projects/common/modules/video/events.js
```diff
@@ -233,10 +233,15 @@ define([
...
+                    var adDuration = this.ima.getAdsManager().getCurrentAd().getDuration();
+
+                    var skipButton = template(adsSkipOverlayTmpl, {
+                        adDuration: adDuration,
+                        skipTimeout: skipTimeout
+                    });
 
                     $(this.el()).append(skipButton);
-                    intervalId = setInterval(events.update.bind(this), 250);
+                    intervalId = setInterval(events.update.bind(this), 500);
```
jamesgorrie: This makes it seem more synced.

akash1810:
Nice! 👍

PR #13185, opened 2016-06-02
Pageskin breakpoint issue

Calum-Campbell:
## What does this change?

Will not lazyload adverts on any pages with pageSkins regardless of size.
This ensures ads are requested with a single request when there should be pageskin on the page. 
This fixes problem of pageskin top adverts not appearing when device is between mobile and 'wide' desktop.
## Request for comment

@kenlim @regiskuckaertz 

before
image

After
image


regiskuckaertz:
You need to pull master, the viewability switch was removed a few days ago. Otherwise, yeah 👍

PR #13188, opened 2016-06-03
Call Id.init() before using Id in standard

niklasvincent:
## What does this change?

Makes sure `Id.init()` gets called in the e-mail subscribe module run-checks. Otherwise `Id.ApiRoot` and `Id.idUrl` will be `null`.

The below code from `static/src/javascripts/projects/common/modules/identity/api.js` outlines why that is:

```
    Id.idApiRoot = null;
    Id.idUrl = null;

    Id.init = function () {
        Id.idApiRoot = config.page.idApiUrl;
        Id.idUrl = config.page.idUrl;
        mediator.emit('module:identity:api:loaded');
    };
```

Currently on certain pages that try to check the e-mail subscription status of the user, the code in `getUserEmailSubscriptions()` in `static/src/javascripts/projects/common/modules/email/run-checks.js` fails because the `Id.ApiRoot` is `null` and the wrong Identity API URL gets constructed:

image

Should make the e-mail subscribe module work on all pages and the console error about the failed request should go away.

No
## Request for comment

@gtrufitt

### DIFF
--- a/static/src/javascripts/bootstraps/enhanced/identity-common.js
+++ b/static/src/javascripts/bootstraps/enhanced/identity-common.js
@@ -11,8 +11,7 @@ define([
     robust,
     Id
 ) {
-    function setCssClass(config) {
-        Id.init(config);
+    function setCssClass() {
         // Used to show elements that need signin. Use .sign-in-required
         if (Id.isUserLoggedIn()) {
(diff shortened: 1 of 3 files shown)


gtrufitt:
👍

PR #13191, opened 2016-06-03
use the right healthcheck object

alexduf:
## What does this change?

Fixes a deployment error: there was two similar healtcheck, it needs to be consistent to work
## Request for comment

@johnduffell


johnduffell:
thanks! 👍

PR #13192, opened 2016-06-03
Hosted page - about lightbox

Calanthe:
## What does this change?

The 'about' text on hosted page is now shown in lightbox instead of toggled div with white background.

The 'about' text is displayed in a better way.
## Screenshots

Before:
screen shot 2016-06-03 at 11 25 47

After:
desktop:
screen shot 2016-06-03 at 11 11 59

mobile:
screen shot 2016-06-03 at 11 11 50

I also removed redundant svg icons and live-events modules which we don't use anymore.


commercial/app/views/hosted/guardianHostedPage.scala.html
```diff
@@ -6,12 +6,7 @@
...
-                <button class="u-button-reset paidfor-label__btn popup__toggle hosted__label-btn" data-toggle="js-hosted-popup" data-link-name="about-adertiser-content">About @fragments.inlineSvg("arrow-down", "icon")</button>
-                <div class="popup popup--paidfor hosted__popup is-off js-hosted-popup">
-                    <p class="hosted__popup-text">Advertiser content is used to describe advertisement features that are paid for, produced and controlled by the advertiser rather than the publisher​.</p>
-                    <p class="hosted__popup-text">They​ are subject to regulation by the Advertising Standards Authority in the UK, the Federal Trade Commission in the US and the Advertising Standards Bureau in Australia.</p>
-                    <p class="hosted__popup-text">This content is produced by the advertiser and does not involve GNM staff.</p>
-                </div>
+                <button class="u-button-reset paidfor-label__btn popup__toggle hosted__label-btn js-hosted-about" data-link-name="about-advertiser-content">About @fragments.inlineSvg("arrow-down", "icon")</button>
```
regiskuckaertz: The "down-arrow" somehow now feels counter-intuitive

static/src/javascripts/projects/common/modules/commercial/hosted-about.js
```diff
@@ -0,0 +1,31 @@
...
+            header: 'Advertiser content',
+            paragraph1: 'Advertiser content is used to describe advertisement features that are paid for, produced and controlled by the advertiser rather than the publisher​.',
+            paragraph2: 'They​ are subject to regulation by the Advertising Standards Authority in the UK, the Federal Trade Commission in the US and the Advertising Standards Bureau in Australia.',
+            paragraph3: 'This content is produced by the advertiser and does not involve GNM staff.',
+            showCloseBtn: true
+        });
+
+        survey.attach();
+
+        bean.on(document, 'click', $('.js-hosted-about'), function (e) {
```
regiskuckaertz: Will there be a lot of these `.js-hosted-about`? I think it's better to add the event to the element itself, document feels like too wide a net to cast.

regiskuckaertz:
Aside questions, looks good to me 👍

commercial/app/views/hosted/guardianHostedPage.scala.html
Calanthe: Yeah, I asked @blongden73 about that. Will do it in another PR.

static/src/javascripts/projects/common/modules/commercial/hosted-about.js
Calanthe: There will be at least 2 `.js-hosted-about` on the page and those elements don't have any common parent. There is a lot `bean.on(document ...` examples in our codebase. If I understood the bean documentation correctly, there can be only one element passed to the `bean.on` method (without delegation). I assume that using the `bean.on(document ...` method is the best way to attach the same event to multiple objects.

PR #13193, opened 2016-06-03
restore css import that is needed for hosted thrasher!

lps88:
## What does this change?

Fix the css for thrasher - isn't styled at all right now!


Calanthe:
👍

PR #13194, opened 2016-06-03
Include the run-of-network adunit in the adunit agent

rich-nguyen:
This updates the adunit agent on the admin server so that it can resolve the id of the ad unit which corresponds to the run-of-network. This is sometimes used to target the whole inventory.


regiskuckaertz:
Yeah! That was fun 😭  

Well done 👍

PR #13168, opened 2016-06-02
inject curl config to video embed page

akash1810:
## What does this change?

configures paths correctly so that work can be done in DEV
## What is the value of this and can you measure success?

can work in DEV
## Does this affect other platforms - Amp, Apps, etc?

no
## Screenshots

n/a
## Request for comment

@samdesborough

PR #13196, opened 2016-06-03
revert changes to js di on video embed page

akash1810:
## What does this change?

backs out a change introduced in #13164 as the embed page is currently borking 😢 need to work out why https://github.com/guardian/frontend/pull/13168 is actually needed but until then, lets fix the embed page!

embed page works again

no
## Screenshots

n/a
## Request for comment

@jamesgorrie


jamesgorrie:
👍

PR #13197, opened 2016-06-03
always ensure https urls for video pages

akash1810:
## What does this change?

We're getting some video urls as http - lets just upgrade them on frontend rather than forcing capi to update itself. We do similar things on embeds already.

https 🔒 

Amp - only because of a code abstraction.
## Screenshots

n/a
## Request for comment

@jamesgorrie @fredex42


jamesgorrie:
@JustinPinner 
👍

PR #13198, opened 2016-06-03
Track excluded adunits in dfp data extraction

rich-nguyen:
Currently, a line item's targeting only contains the adunits it includes. However, there are occasions when the line item excludes adunits too. If we track these, we get the full targeting picture:

commercial_high_slot


kenlim:
👍

kenlim:
👍

PR #13199, opened 2016-06-03
Cleanup standard

regiskuckaertz:
@gtrufitt


gtrufitt:
Other than our quick outer-screen discussion - 👍

PR #13200, opened 2016-06-03
New header resize relatively

NataliaLKB:
@zeftilldeath designed a beautiful new nav, but it is perfect on 320px wide devices. To keep the perfection across all devices this PR makes sure that everything in the new nav resizes proportionally.
## TODO in later PRs:
- [ ] Only resize until mobile landscape
- [ ] Only on mobile show the new-nav (show the old nav when resized to desktop)
- [x] Make links that work (home should go to home) (https://github.com/guardian/frontend/pull/13202)

Not yet 😉


jamesgorrie:
So many magic numbers!
viewport units aren't supported by opera mini if that matters.

👍

NataliaLKB:
Thanks @jamesgorrie!

For the moment it should be okay (it is still only an opt-in test), but the next step would be to add fallback numbers  and see if I can use percentages when needed.

static/src/stylesheets/layout/_ab-new-header-test-variant.scss
```diff
@@ -41,9 +41,19 @@
         // Removes extra spacing
         display: block;
 
+        // Aspect ratio: 16:3
+        @include mq($until: phablet) {
+            width: 50vw;
+            // height: auto doesn't work in Safari
+            height: calc(3/16 * 46vw);
```
mchv: any link to webkit bugzilla? so we could keep track of the issue.
why `calc(3/16 * 46vw)` and not `calc(3/16 * 50vw)`?

PR #13202, opened 2016-06-03
Links should work new header

NataliaLKB:
When testing the new nav, the first thing you would probably notice is that none of the buttons work. As a start, this PR adds some functionality to the links, even if some of it is temporary.

This also cleans up some of the svgs in the template. The wold icon has now moved to the `inline-svg` directory and the hamburger menu is done with css instead. This is because of browser support. We still want a menu even if ie8 doesn't support svgs 😄 

Nope!


gtrufitt:
👍

PR #13204, opened 2016-06-03
Height fallback immersive media

NataliaLKB:
The main media immersive on galleries uses `object-fit: cover;` to make sure that the image expands to fill up the viewport space.

Originally the fallback for this was to just show the image as is, without any sizing based on viewport. Unfortunately this resulted in some galleries looking like this on mobile safari:
image

Though this isn't terrible, most pictures don't look great like this. 

To solve this problem I wrote a simple polyfill so now it does this:
image

I think this is a bit of a better solution, but feedback appreciated!

Nope 😄


static/src/javascripts/projects/common/modules/ui/objectfit-polyfill.js
```diff
@@ -0,0 +1,42 @@
...
+                        .css('backgroundImage', 'url(' + src + ')')
+                        .addClass('immersive-main-media--fallback');
+                }
+            });
+        });
+    }
+
+    function init() {
+        return fastdomPromise.read(function () {
+            return document.getElementsByClassName('js-immersive-main-media__img')[0];
```
jamesgorrie: If this doesn't exists, won't it blow up?

paperboyo:
Amazing! This will help more than just Safari! ( 👀 at you, Microsoft!)

sndrs:
does this need JS? can you apply the fallback CSS by default and override/reset etc as necessary inside the `supports` block?

static/src/javascripts/projects/common/modules/ui/objectfit-polyfill.js
NataliaLKB: Thanks @jamesgorrie. Good catch! 

If it is a really old gallery it may not have any an immersive-main-media. If that is the case, this line returns `undefined`. I have added a check to make sure the picture is there before getting the style.

NataliaLKB:
@sndrs as discussed in person, I have switched out `getComputedStyle` with using `CSS.Supports()` api. Definitely a better solution.

Thanks 😄

common/app/views/main.scala.html
```diff
@@ -28,12 +28,12 @@
 }
 
 @immersiveMainMedia(page: GalleryPage) = {
-    <div class="new-gallery content tonal--tone-media immersive-main-media">
+    <div class="new-gallery content tonal--tone-media immersive-main-media js-background-image">
```
sndrs: i wonder if `js-background-image` is too generic? maybe `js-objectfit-pf` or something?

static/src/stylesheets/module/content/_article-immersive.scss
```diff
@@ -22,17 +22,32 @@
     width: 100%;
     // Remove inline spacing
     display: block;
+    // This is for the fallback;
+    min-height: 400px;
```
sndrs: is this duplicated on line 43?

static/src/javascripts/projects/common/modules/ui/objectfit-polyfill.js
```diff
@@ -0,0 +1,39 @@
...
+
+    function setBackgroundImage(src) {
+        return fastdomPromise.write(function () {
+            $('.js-background-image').each(function () {
+                var $container = $(this);
+
+                if (src) {
+                    $container
+                        .css('backgroundImage', 'url(' + src + ')')
+                        .addClass('immersive-main-media--fallback');
```
sndrs: i wonder, as it's a polyfill, if you should set this css in the JS?

sndrs:
👍 apart from comments - nice one :)

static/src/stylesheets/module/content/_article-immersive.scss
NataliaLKB: Great catch! It is on two different elements, but I only needed to put it on this one.

common/app/views/main.scala.html
NataliaLKB: Thats fair enough. I can imagine that being confusing in the future. Changed!

static/src/stylesheets/module/content/_article-immersive.scss
```diff
@@ -22,17 +22,31 @@
...
+        min-height: none;
         height: 100%;
         width: 100%;
         z-index: 0;
     }
 }
 
+.immersive-main-media--fallback {
+    background-size: cover;
+    background-position: center center;
```
gtrufitt: Only need the one `center` here I think?

static/src/javascripts/projects/common/modules/ui/objectfit-polyfill.js
```diff
@@ -0,0 +1,39 @@
...
+                }
+            });
+        });
+    }
+
+    function init() {
+        return fastdomPromise.read(function () {
```
gtrufitt: I don't think you need to use fastdom here because  you're only getting the element, not querying anything that would force layout. You're also tying your polyfill directly to the immersive functionality here, maybe it would be better to pass a class name to the init and loop over the elements with that class name, inserting a wrapping element directly (rather than applying a class to something else)?

gtrufitt:
Great to polyfill this into IE! Nice work :)

NataliaLKB:
@gtrufitt @sndrs thanks for reviewing! I don't have time to finish this PR this week so I will close down and reopen when it is done.

PR #13206, opened 2016-06-06
extend offline page view swtich

sndrs:
will remove it entirely later on, this will just stop it breaking for now

will remove later on today


desbo:
👍

JonNorman:
Does this not need to be extended to at least the 7th? It looks to be the 4th, which was Saturday...

sndrs:
that is a very good point!

JonNorman:
:+1:

PR #13207, opened 2016-06-06
Add preconnect link for DFP library

regiskuckaertz:
@rich-nguyen @JonNorman

### DIFF
--- a/common/app/views/fragments/metaData.scala.html
+++ b/common/app/views/fragments/metaData.scala.html
@@ -7,7 +7,7 @@
 @import play.api.Play
 @import play.api.Play.current
 @import views.support.{SeoThumbnail, StripHtmlTags}
-@import conf.switches.Switches.{AmpSwitch, UseLinkPreconnect, GoogleAnalyticsSwitch}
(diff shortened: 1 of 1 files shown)


piuccio:
`Configuration.googletag.jsLocation` suggest a javascript file location. `preconnect` works with origins, not with specific files. Well maybe browser are smart enough to extract the origin out of any path, is that the case?
If that's an actual file maybe you can use [prefetch](http://caniuse.com/#feat=link-rel-prefetch) that seems to have better support than [preconnect](http://caniuse.com/#feat=link-rel-preconnect)

regiskuckaertz:
you can put a fully qualified URL there, the browser will figure out the domain

piuccio:
I didn't think browsers were that smart 👅

regiskuckaertz:
😆 

`prefetch` is just a hint, the browser decides whether or not to follow it—`preconnect` is an order.

piuccio:
go for the order 👍

rich-nguyen:
👍

PR #13211, opened 2016-06-06
Remove offline pages

sndrs:
## What does this change?

removes the offline page view metrics switch (and page) plus the dev blog crossword one

the page view test is over, so that can go. 

the crossword page on the other hand, while only being viewable from the devblog, requires you to download a crossword and its assets on any visit to an https page in a service-worker capable browser, regardless of whether you visit a crossword page. this includes, for example, react, and a load of crossword-specific CSS. as we move nearer to https everywhere, this doesn't really seem ok...

the SW will also not install until all this has arrived, in theory meaning that a network error for an asset of a page you haven't visited, which you will only use if you try and visit the dev blog offline, could prevent the SW installing and denying all other stuff we use it for (or will) e.g. notifications. 

(it also adds a fair degree of complexity to maintenance)
## Request for comment

@rich-nguyen


rich-nguyen:
👍  did you get any numbers on the offline page views metric?

sndrs:
> did you get any numbers on the offline page views metric?

0.2% of people who could install the SW attempted to view an https page while offline and then returned (to any page)

theefer:
:cry:

sndrs:
something will take its place, just not this...(for now, anyway)

PR #13212, opened 2016-06-06
Hosted By page - css improvements

lps88:
## What does this change?

Simplify the new head.commercial.scss file

This file is just meant to hold commercial-specific styles (at the moment just hosted by stuff) but as it replaces head.content.scss in some cases, we now just import that whole file.

This should fix the styling issues we have with the commercial test page on Code at the moment
## Request for comment

@regiskuckaertz does this work okay/follow the normal conventions of the project?


regiskuckaertz:
@lps88 do you need all that stuff? it all depends on whether you want the hosted page CSS to be a 1:1 copy of content pages—think future changes, additions and removals—or if you want to be able to cherry pick the styles you really need.

If yes, then go for it 👍

lps88:
@regiskuckaertz I did cherry pick what I thought I needed, but I think the bug with http://m.code.dev-theguardian.com/commercial/test-page is because it's picking up head.commercial instead of head.content (probably here common/app/assets/assets.scala:87) and I don't want any other pages to break!

lps88:
But can't be sure that this is what's caused the test page to break as it works fine locally...

PR #13214, opened 2016-06-06
Hosted Video - fix mime type

lps88:
## What does this change?

We were using the same video url in all the <source> tags, they should each specify the correct mime type.

This was causing the video to not load properly in ios (it doesn't support webm type)

### DIFF
--- a/commercial/app/views/hosted/guardianHostedPage.scala.html
+++ b/commercial/app/views/hosted/guardianHostedPage.scala.html
@@ -29,10 +29,10 @@
                       data-block-video-ads="true"
                       poster="@page.video.posterUrl"
                       class="vjs-hosted__video hosted__video gu-media--video vjs vjs-tech-html5 vjs-paused vjs-controls-enabled vjs_video_1-dimensions vjs-user-active">
-                        <source type="video/mp4" src="@page.video.srcUrl">
-                        <source type="video/webm" src="@page.video.srcUrl">
-                        <source type="video/ogg" src="@page.video.srcUrl">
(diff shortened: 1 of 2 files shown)


commercial/app/views/hosted/guardianHostedPage.scala.html
```diff
@@ -29,10 +29,10 @@
...
                       poster="@page.video.posterUrl"
                       class="vjs-hosted__video hosted__video gu-media--video vjs vjs-tech-html5 vjs-paused vjs-controls-enabled vjs_video_1-dimensions vjs-user-active">
-                        <source type="video/mp4" src="@page.video.srcUrl">
-                        <source type="video/webm" src="@page.video.srcUrl">
-                        <source type="video/ogg" src="@page.video.srcUrl">
-                        <source type="video/m3u8" src="@page.video.srcUrl">
+                        <source type="video/mp4" src="@page.video.srcUrl&format=video/mp4&maxbitrate=2048">
+                        <source type="video/webm" src="@page.video.srcUrl&format=video/webm&maxbitrate=2048">
+                        <source type="video/ogg" src="@page.video.srcUrl&format=video/ogg&maxbitrate=2048">
+                        <source type="video/m3u8" src="@page.video.srcUrl&format=video/m3u8&maxbitrate=2048">
```
jamesgorrie: We could do this in the scala - potentially creating something like a:
`@page.video.srcUrlFor("mp4", 2048)`... not a biggy though.

jamesgorrie:
👍

PR #13216, opened 2016-06-06
Video nav test

jamesgorrie:
## What does this change?

Adding a test to see what happens when we add video to the main nav.
Does it cannibalise other clicks, or just interest people in video.

We will be making the video front a little better before sending this out.

People want video more, but not at the expense of others.

This isn't really an A/B per-say - more a test about, if we did this, what would the affect be? We can then make more informed decisions around it later. 

No
## Screenshots

video-nav
## Request for comment

@akash1810 @harrysalmon


static/src/javascripts/projects/common/modules/experiments/tests/video-nav.js
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+
+    return function () {
+        this.id = 'VideoNav';
+        this.start = '2016-06-06';
+        this.expiry = '2016-06-13';
+        this.author = 'James Gorrie';
+        this.description = 'Use video carousel as video nav';
+        this.audience = .2;
```
harrysalmon: You mentioned changing this

static/src/javascripts/projects/common/modules/experiments/tests/video-nav.js
```diff
@@ -0,0 +1,47 @@
+define([], function () {
+    function setVideoOnNav(text) {
+        var videoIcon = document.querySelector('.inline-video-icon').cloneNode(true);
```
akash1810: will this work if there's no video on the front?

static/src/javascripts/projects/common/modules/experiments/tests/video-nav.js
jamesgorrie: Looking at the a/b dashboard - we're good between 10 & 12%?

static/src/javascripts/projects/common/modules/experiments/tests/video-nav.js
jamesgorrie: Good catch, changed.

static/src/stylesheets/module/nav/_nav.scss
```diff
@@ -119,3 +119,12 @@
         }
     }
 }
+
+.video-nav-test-icon {
+    svg {
+        //fill: #ffbb00;
```
akash1810: not needed?

akash1810:
👍

PR #13218, opened 2016-06-06
Correctly identify https fronts regardless of your edition

dominickendrick:
## What does this change?

If you were on a `uk/` edition page, but had a cookie set for `au/` region, we would **not** report the `canonical` link as being on `https`. This now makes sure all sections of the site are reported to be on `https` regardless of the edition the user is visiting us from.

This manifested itself with Google when looking at the `rel=alternative` and `rel=canonical` link tags. It would be visiting the site with the default edition (`uk`) but be looking at the `us/sport` page, in this scenario the `canonical` url was set to `http` but the `alternative` url was using `https`.

Correctly report alternative urls in Google

Google Search Console
## Request for comment

@gklopper @johnduffell


common/app/common/LinkTo.scala
```diff
@@ -63,18 +64,25 @@ trait LinkTo extends Logging {
...
+
+    // check if the url has _any_ edition prefix (/au/, /us/ ... )
+    // as users can have au edition cookie but be on a us edition url
+
+    val hasEditionPrefix = Edition.all.exists{ edition =>
+      path.startsWith(edition.id.toLowerCase + "/")
     }
-    httpsEnabledSections.exists(noEdPath.startsWith)
+
+    val sectionPath = path.replaceFirst(s"^(${editionRegex})/", "")
```
johnduffell: make it def?

johnduffell:
great! well worked out, that's a trickey one 👍

dominickendrick:
@annebyrne

annebyrne:
👍

PR #13221, opened 2016-06-06
[test] always showcase main media videos

akash1810:
## What does this change?

Testing if large main media video leads to more plays.

⬆️ video plays.

No.
## Screenshots

screen shot 2016-06-06 at 18 31 05
## Request for comment

@jamesgorrie @mr-mr

### DIFF
--- a/static/src/javascripts/bootstraps/enhanced/media/main.js
+++ b/static/src/javascripts/bootstraps/enhanced/media/main.js
@@ -462,6 +462,17 @@ define([
         }
     }
 
+    function showcaseMainMedia() {
+        $('.content__meta-container').addClass('content__meta-container--showcase');
+        $('.media-primary').addClass('media-primary--showcase');
+    }
+
+    function initTests() {
+        if(ab.isInVariant('VideoMainMediaAlwaysShowcase', 'variant')) {
+            showcaseMainMedia();
+        }
+    }
+
     function init() {
         // The `hasMultipleVideosInPage` flag is temporary until the # will be fixed
         var shouldPreroll = commercialFeatures.videoPreRolls &&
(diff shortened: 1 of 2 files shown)


static/src/javascripts/bootstraps/enhanced/media/main.js
```diff
@@ -407,6 +409,17 @@ define([
         }
     }
 
+    function showcaseMainMedia() {
+        $('.content__meta-container').addClass('content__meta-container--showcase');
```
akash1810: if test succeeds, we'll do this sever side

jamesgorrie:
:

jamesgorrie:
👍

akash1810:
this'll go out next week when there's more audience availability

akash1810:
updated with real audience numbers after baselining - ping @harrysalmon

paperboyo:
Just wanted to mention that the weightings in Composer were thought as a Editorial tool to indicate the importance of (some!) content. So much so, that some are even behind a switch available only to some (designers). If Showcase for videos is not used like that, I would educate rather than just bluntly set all to that... Just my 2gr, obv!

paperboyo:
OTOH, if the numbers from this test prove conclusive, there is going to be a good incentive to use video weightings more!

akash1810:
> I would educate rather than just bluntly set all

Absolutely.

If the test proves positive, we'd likely default the weighting to showcase, but still allow the weighting to be changed. This is an assumption; lets wait for the test to complete before thinking too much on it 😄 .

harrysalmon:
Looks good to me! We'll be able to detect anything above a 2% difference in the number of 25% video completes in 5 days time. This'll fill the AB test space too

PR #13223, opened 2016-06-07
AB Test: Match reports - Stars vs Footballs

gtrufitt:
## What does this change?

This adds an AB test designed to prove the hypothesis: "Using a rating system consistent with others on the Guardian - the star ratings - will increase the amount of ratings".

We were underwhelmed by the performance of football icon ratings vs comments on articles (even though actual CTR of the low friction was level with film rating) so want to test whether the icons make a difference.

Success is: The recognisable ratings system, ie: stars, is 50% more clicked than the subject specific icon ratings system - ie: footballs
## Screenshots

screen shot 2016-06-07 at 11 16 39

**VS**

screen shot 2016-06-07 at 11 17 06
## Request for comment

@guardian/participation


NataliaLKB:
👍

PR #13224, opened 2016-06-07
Hosted By update video image

lps88:
## What does this change?
- new poster image for third renault video
- add description to pinterest share

cc @Calanthe


Calanthe:
👍

PR #13227, opened 2016-06-07
Ability to add multiple video playlists

jamesgorrie:
## What does this change?

We were adding no context to the video playlists, so adding multiple borked with events etc.

Using the content store as it's a nice way to deal with state, and this quite stateful. Less mutation = more 💪 

We can add more than one carousel (which we need to as we're planning on changing the video front).

No
## Request for comment

@akash1810


jamesgorrie:
This should also make it easier to test, although I will add those once this is in as I need this for our front tests

akash1810:
👍

PR #13228, opened 2016-06-07
extend highmerch switches for a month

Calum-Campbell:
## What does this change?

extends the high-merchandising-optomisation switches for another month.

@rich-nguyen


JonNorman:
:+1:

PR #13232, opened 2016-06-07
Move classlist polyfill to head

sndrs:
## What does this change?

moves the classlist polyfill to inline at the top

the conditional comments only served it to IE9 and below, but it's needed by opera mini too.
## Request for comment

 @regiskuckaertz


regiskuckaertz:
Great! 👍

PR #13235, opened 2016-06-07
Support capi-sourced section branding

kelvin-chappell:
Capi now has a `show-section` parameter, which includes a section element in an item result.
See https://github.com/guardian/content-api/pull/1567
This allows us to support section-level branding.

The main changes in this PR are:
- the `section` property of `Metadata` is now a `SectionSummary` rather than a section ID.
- `item` requests use the `show-section` parameter by default
- `BrandHunter` decides which logo to show for content based first on its section and then on its tags.

/cc @steppenwells


jbreckmckye:
What exactly do all these `/data/database` files actually do? I thought they were generated locally?

kelvin-chappell:
@jbreckmckye They're cached capi results that teamcity uses for test fixtures.

applications/app/model/TagIndexListingMetaData.scala
```diff
@@ -3,7 +3,7 @@ package model
 case class SubjectsListing() extends StandalonePage {
   override val metadata = MetaData.make(
     id = "index/subjects",
-    section = "Index",
+    sectionSummary = Some(SectionSummary.fromId("Index")),
```
JonNorman: In `TagIndexPageMetadata` the sectionSummary string has been extracted into `private val tagType: String` - could do the same here for consistency.

common/app/model/meta.scala
```diff
@@ -259,6 +260,8 @@ final case class MetaData (
   opengraphPropertiesOverrides: Map[String, String] = Map(),
   twitterPropertiesOverrides: Map[String, String] = Map()
 ){
+  // todo rename sectionId
```
JonNorman: Does this refer to the `sectionId` above?

common/app/common/commercial/BrandHunter.scala
```diff
@@ -6,29 +6,25 @@ import org.joda.time.DateTime
...
-    findBranding(tag.properties.activeBrandings, publicationDate, edition)
-  }
+    val brandingBySection = content.metadata.section.flatMap { section =>
+      findBranding(section.activeBrandings, edition, publicationDate)
+    }
 
-  def findFrontBranding(frontProps: FrontProperties, edition: Edition): Option[Branding] = {
-    findBranding(frontProps.activeBrandings, publicationDate = None, edition)
-  }
+    val brandingByTags = content.tags.tags.flatMap { tag =>
```
JonNorman: Not related to this PR but is there a reason why `Tags` doesn't just extend `Traversable[Tag]`? Calling `Tags.tags` smells funny.

common/app/views/fragments/headTonal.scala.html
```diff
@@ -85,10 +84,7 @@ <h1 class="content__headline js-score" itemprop="headline">
                     </div>
                 }
 
-                @if(
-                    (showBadge && !item.commercial.isAdvertisementFeature) ||
-                        (staticBadgesSwitch.isSwitchedOn && !item.tags.isVideo && !item.tags.isAudio)
-                ) {
+                @if(showBadge && !item.commercial.isAdvertisementFeature) {
```
JonNorman: Do we no longer care about whether there are video/audio tags on the page before we show the badge?

JonNorman:
Apart from minor comment above, looks good to me :+1:

common/app/common/commercial/BrandHunter.scala
kelvin-chappell: The frontend content model goes through occasional clean ups but there's a lot of scope for more work.  It would be less confusing if it followed the capi result more closely.

common/app/views/fragments/headTonal.scala.html
kelvin-chappell: This is to avoid logos appearing twice in articles.  They appear in a different place on audio and video pages.  Without following the logic through completely I've just done what seems to make logos appear in the right places.

PR #13236, opened 2016-06-07
set contrib-ads timeuot slightly higher that ima timeout

jamesgorrie:
## What does this change?

We did this before but it corresponded with a dip in video traffic which was just a dip in traffic.
Stops the video playing before the ad.

No more complaints

No

@akash1810


akash1810:
👍

PR #13237, opened 2016-06-07
remove darkened sides on video front for carousel

jamesgorrie:
## What does this change?

We're going to be putting more than one carousel on the video front so it starts getting a bit dark.
This is just first pass, we'll play with it as we go on.

No
### dark before

dark-side
### light now

light-sight
## Request for comment

@akash1810 @blongden73


akash1810:
👍

PR #13238, opened 2016-06-07
Aren't email headers great?

superfrank:
Lab notes email header added
image

@desbo Do I need to edit/add anything else, or we good?


gtrufitt:
👍

PR #13241, opened 2016-06-07
Refactor hosted page content

kelvin-chappell:
This moves all the hosted content into `HostedPage` so that it's easier to work out.
It should have no effect on what's rendered.

/cc @Calanthe @lps88


lps88:
Cool 👍

PR #13210, opened 2016-06-06
Facebook overlay everywhere!

dominickendrick:
## What does this change?

This rolls out the Facebook overlay across the entire site.
## Does this affect other platforms - Amp, Apps, etc?

Facebook
## Screenshots

screen shot 2016-06-06 at 11 45 43
## Request for comment

@stephanfowler @katebee

PR #13242, opened 2016-06-07
Twitter share image resize

katebee:
## What does this change?

The end result is a change to the content of the twitter:image meta-tag. 

There is currently no image profile for twitter images; this PR reduces the size of the image served to the Twitter scraper.

Twitter uses Open Graph tags as fallbacks when Twitter Card meta-tags are missing or not usable. After creating custom images for Open Graph (#13210), it was discovered that Twitter sometimes uses the Open Graph image, despite a twitter:image meta-tag existing on our page.

The cause appears to be that we are serving a raw image to Twitter, however, images must be less than 1MB in size for Twitter Cards ([docs](https://dev.twitter.com/cards/overview)), and our raw images sometimes exceed this.

This PR should ensure that Twitter will always use the image from the twitter:image meta-tag, and will prevent the overlay from appearing on Twitter.

Twitter
## Screenshots

**Twitter post that is using an Open Graph image:**

Affected article with an overlay on Twitter

**Twitter post that is using a Twitter Card image:**

Intended appearance of an article on Twitter
## Request for comment

@dominickendrick

### DIFF
--- a/article/test/ArticleFeatureTest.scala
+++ b/article/test/ArticleFeatureTest.scala
@@ -549,7 +549,7 @@ import collection.JavaConversions._
 
         // at the time of writing, Twitter does not like i.guim.co.uk
         // will see if I can get that fixed, but in the meantime this must be static.guim.co.uk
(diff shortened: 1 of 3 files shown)


NataliaLKB:
Other than the test that needs updating, it lgtm 👍

dominickendrick:
👍

PR #13244, opened 2016-06-07
Lydia hosted social share

lps88:
## What does this change?
- Pinterest share button to use different url so that the description field works
- Page titles on all pages need to be prefixed with "Advertiser content hosted by the Guardian: "
- All share buttons should use the full (prefixed) page title in subject/description field
- remove unnecessary display: block css which broke the layout on iphone

Depends on: https://github.com/guardian/frontend/pull/13241

image


kelvin-chappell:
👍 
I haven't merged my branch yet, so you can deploy this and I'll close mine.

PR #13249, opened 2016-06-08
Quizzes: put the personality message before the answers

rrees:
## What does this change?

To avoid making the user scroll through all the answers to see what kind of thing they were this creates the concept of a "header" section that has a summary of the personality and the share buttons.

Social referral for the quiz.

No
## Screenshots

I can't get this running locally, sorry.
## Request for comment

The styling class `atom-quiz__end-message` looks like it might be wrong now. But I'm not sure what the styling intent around this was. I'm not sure if it is really important to differentiate the start and end notes or whether it would be better to have a generic note class and then additional modifier classes of start and end.


common/app/views/fragments/atoms/quiz.scala.html
```diff
@@ -102,6 +103,26 @@
...
+@renderHeader(maybeResults: Option[QuizResults]) = {
+<div class="atom-quiz__end-message">
+    @maybeResults.map { results =>
+        @if(results.isPersonality) {
+
+        <div class="atom-quiz__score-message">
+            <div class="atom-quiz__personality-title">@results.resultBucket.map(_.title)</div>
+            <div class="atom-quiz__personality-description">@results.resultBucket.map(_.description)</div>
+        </div>
+        }
```
sndrs: the indenting's a bit off here, i missed this `}` at first.

common/app/views/fragments/atoms/quiz.scala.html
```diff
@@ -102,6 +103,26 @@
     }
 }
 
+@renderHeader(maybeResults: Option[QuizResults]) = {
+<div class="atom-quiz__end-message">
```
sndrs: maybe this should be `atom-quiz__response` or `atom-quiz__message`, or something similar, since it's not really at the end now?

common/app/views/fragments/atoms/quiz.scala.html
```diff
@@ -102,6 +103,26 @@
     }
 }
 
+@renderHeader(maybeResults: Option[QuizResults]) = {
```
sndrs: same with name here actually (which I'm writing after the classname one – sorry if that's confusing!)

sndrs:
👍 other than comments

rrees:
@sndrs can review and release if acceptable, I can't get this to work locally. Just seeing the atom content with no styling or behaviour 😞

sndrs:
screen shot 2016-06-09 at 14 13 50

PR #13250, opened 2016-06-08
Log object dimensions in the spacefinder

regiskuckaertz:
Temporary log to debug a case that is not reproducible on DEV.

cc @JonNorman


JonNorman:
:( :+1:

PR #13251, opened 2016-06-08
Remove show-mpu-in-all-containers config

rich-nguyen:
This testing feature isn't being used. @kelvin-chappell


kelvin-chappell:
Nice to see old stuff going. 👍

PR #13255, opened 2016-06-08
Remove data-component attribute from links

philwills:
## What does this change?

Removes `data-component` attribute from in body and in comment links.

Changes `data-link-name` from comment links to be more honest.

Less verbose HTML, less noise reported to Ophan.

No
## Screenshots

N/A
## Request for comment

This is not used by Omniture and as far as Ophan is concerned doesn't make sense on these elements.


jamesgorrie:
Bonza 👍

PR #13256, opened 2016-06-08
Decouple source of page branding and container branding 

kelvin-chappell:
So that they can be changed independently.

This is because there are changes in the fronts tool to be communicated before container branding goes live, and it would be wise to decouple this from changes in the tag manager.


JonNorman:
:+1:

PR #13259, opened 2016-06-08
Hosted By - extract video content

lps88:
## What does this change?

Move all the video page content into a video template, so that gallery page can use the Hosted By header.

Merging this in straightaway so that there aren't merge nightmares later.

cc @kelvin-chappell

### DIFF
--- a/commercial/app/controllers/commercial/HostedContentController.scala
+++ b/commercial/app/controllers/commercial/HostedContentController.scala
@@ -4,13 +4,13 @@ import common.commercial.HostedPage
 import model.Cached.RevalidatableResult
 import model.{Cached, NoCache}
 import play.api.mvc.{Action, Controller}
-import views.html.hosted.guardianHostedPage
+import views.html.hosted.{guardianHostedPage, guardianHostedVideo}
 
 object HostedContentController extends Controller {
 
   def renderHostedPage(pageName: String) = Action { implicit request =>
     HostedPage.fromPageName(pageName) match {
-      case Some(page) => Cached(60)(RevalidatableResult.Ok(guardianHostedPage(page)))
+      case Some(page) => Cached(60)(RevalidatableResult.Ok(guardianHostedPage(page, guardianHostedVideo(page))))
       case None => NoCache(NotFound)
     }
   }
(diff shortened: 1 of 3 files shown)


commercial/app/controllers/commercial/HostedContentController.scala
```diff
@@ -4,13 +4,13 @@ import common.commercial.HostedPage
...
 import play.api.mvc.{Action, Controller}
-import views.html.hosted.guardianHostedPage
+import views.html.hosted.{guardianHostedPage, guardianHostedVideo}
 
 object HostedContentController extends Controller {
 
   def renderHostedPage(pageName: String) = Action { implicit request =>
     HostedPage.fromPageName(pageName) match {
-      case Some(page) => Cached(60)(RevalidatableResult.Ok(guardianHostedPage(page)))
+      case Some(page) => Cached(60)(RevalidatableResult.Ok(guardianHostedPage(page, guardianHostedVideo(page))))
```
kelvin-chappell: Could call `guardianHostedPage` template from inside `guardianHostedVideo` template, then this becomes:
`RevalidatableResult.Ok(guardianHostedVideo(page))`

commercial/app/views/hosted/guardianHostedPage.scala.html
```diff
@@ -1,5 +1,5 @@
 @import common.commercial.HostedPage
-@(page: HostedPage)(implicit request: RequestHeader)
+@(page: HostedPage, content: Html)(implicit request: RequestHeader)
```
kelvin-chappell: Would be easier to use if the `content` was in a separate argument list:
`@(page: HostedPage)(content: Html)(implicit request: RequestHeader)`

commercial/app/views/hosted/guardianHostedVideo.scala.html
```diff
@@ -0,0 +1,120 @@
+@import common.commercial.HostedPage
+@(page: HostedPage)(implicit request: RequestHeader)
+
```
kelvin-chappell: Could include surrounding page template here (so that the controller only needs to know about one template):
```
@guardianHostedPage(page) {
...
}
```

kelvin-chappell:
Good idea to share the surrounding page template.
Just a few comments to make the implementation easier.  Hope it makes sense.
After that: 👍

PR #13261, opened 2016-06-08
OMComments: Fixing permalinking and jumping to comments

gtrufitt:
I thought I would do a nice thing a couple of weeks ago and fix permalinking, I added this [little fix](https://github.com/guardian/frontend/pull/13042/files), thinking all was well. 
## You'll never believe what happens next

A 1:1 ration of lines-added to bugs-written. Good Job Gareth. I'd written an infinite loop, isn't that fun!? Needless to say, Chrome was not happy.
### Moving loading functionality to the right place

First I removed the last then from here

```
promise.then(this.renderComments.bind(this)).then(this.goToPermalink.bind(this, options.comment));
```

and let our goToComment function handle this, like the rest of the post-comment-load functionality.
### Removing an infinite loop / brittle listener

To avoid any kind of changing of the `hash`, calling of the `hashchange` event loop, I removed the listener entirely - it called `gotoComment` which changed the hash, which fired the hashchange which..etc.

```
bean.on(window, 'hashchange', function() {
        var commentId = this.getCommentIdFromHash();
        if (commentId) {
            this.gotoComment(commentId);
        }
    }.bind(this));
```
### Removing needless fetch

The listener on the _jump to comment_ button on staff picks **always** loaded comments, no matter whether the comment was there or not - a needless http request, so that changed to `goToComment` which then handles whether the comment thread:
- Needs to be shown and jumped to
- Needs to be fetched then jumped to
- Is already visible and can immediately be jumped to
### Removing a double jump

It also used scroller to go to the top of the comments which didn't seem needed and in cases where the comment was already there meant you clicked on it, it took you to the comment and then it took  you back to the top of comment thread - that was confusing.
### The end game.

This should mean that all cases where comments are jumped to, they _should_ load:
- Clicking through from profile page, adds a hash to the page url, this is read when comments are loaded
- Clicking on a comment permalink in a thread, reloads the page anyway and takes you back to that comment
- Clicking on the staff pick should in all cases, whether the comment is on the page, hidden or needs to be loaded take you to the comment

cc/ @rich-nguyen @ScottPainterGNM @jbreckmckye


gklopper:
👍

ScottPainterGNM:
Nice work truff!

static/src/javascripts/projects/common/modules/discussion/loader.js
```diff
@@ -405,15 +403,27 @@ Loader.prototype.initPagination = function() {
...
     if (comment.length > 0) {
-        window.location.replace('#comment-' + id);
-        return;
+        var commentsAreHidden = $('.js-discussion-main-comments').css('display') === 'none';
+        // If comments are hidden, lets show them
+        if (commentsAreHidden) {
+            fastdom.write(function(){
+                thisLoader.comments.showHiddenComments();
+                thisLoader.removeState('truncated');
+                $('.d-discussion__show-all-comments').addClass('u-h');
```
rich-nguyen: does this class definitely exist?

static/src/javascripts/projects/common/modules/discussion/loader.js
```diff
@@ -445,6 +455,9 @@ Loader.prototype.loadComments = function(options) {
         } else {
             this.removeState('pagesize-msg-show');
         }
+        if (options.comment) {
+            this.gotoComment(options.comment);
```
rich-nguyen: do you need to `removeState(`truncated`)` too?

rich-nguyen:
👍

static/src/javascripts/projects/common/modules/discussion/loader.js
gtrufitt: It's handled in the gotoComment if required

gtrufitt:
Cheers all

PR #13262, opened 2016-06-08
Update live blog notifications expiry and start

janua:
This changes the expiry on `live-blog-chrome-notifications-internal` to August 31st 2016.

Also changes the expiry on `live-blog-chrome-notifications-prod` to August 31st 2016, but brings back the **start** to June the 8th 2016 so that it will _actually run_.

@samdesborough @NathanielBennett @crifmulholland


desbo:
👍  📆

PR #13264, opened 2016-06-09
Only show new header mobile

NataliaLKB:
This PR does two things. 

The first is that we are only going to test the new header on mobile. 

The second is to change how the header looks on mobileLandscape and above. Since the new header was fist implemented on the site, the design has changed above mobile. Luckily this has lead to a fair amount of deletions 😄 

Nope!
## Screenshots

320px:
image

414px:
image

568px:
image

730px:
image


paperboyo:
Three-line-chinese-trigram-for-heaven-menu-icon-aka-the-hamburger does not look exactly vertically centred on the first (320px) screenshot... Just sayin’

static/src/stylesheets/layout/_ab-new-header-test-variant.scss
```diff
@@ -5,55 +5,40 @@
...
-        }
-    }
-
-    @include mq(phablet) {
-        align-items: center;
-        box-sizing: border-box;
-        padding-right: ($gs-gutter / 2);
-        padding-left: ($gs-gutter / 2);
-    }
+    //&::after {
```
jamesgorrie: Any need to keep this here?

jamesgorrie:
Does this mean we will have a hamburger solar eclipse?
safety-glasses

👍 to the code.

NataliaLKB:
@paperboyo 
> Three-line-chinese-trigram-for-heaven-menu-icon-aka-the-hamburger does not look exactly vertically centred on the first (320px) screenshot... Just sayin’

Should be better now! I updated the picture in the description

static/src/stylesheets/layout/_ab-new-header-test-variant.scss
NataliaLKB: Thank you! Deleted.

NataliaLKB:
@jamesgorrie 
> Does this mean we will have a hamburger solar eclipse?

I will make a fix (add more space between the top of the page and the nav) in another PR

PR #13268, opened 2016-06-13
Phase out the ad-free test

davidfurey:
## What does this change?

Stops adding new users to the ad-free test variant group, but continues 
to serve ad-free to users already allocated.  These users will start seeing 
ads three days after their first visit, and then we can turn the test off
completely.

Part of the ad-free test

No


NataliaLKB:
👍

PR #13270, opened 2016-06-13
Hosted By videos - stop errors

lps88:
## What does this change?

Change the order of the video player initialisation steps in hosted-video.js

Should hopefully stop the hosted videos erroring when playing/pausing/scrubbing.
"A network error caused the media download to fail part-way."
## Request for comment

@ScottPainterGNM @kelvin-chappell


Calanthe:
👍  When I created that module I copied the initialisation order from the existing video module (afair). I am not sure if this will fix all of the issues with network errors but it's worth a try.

kelvin-chappell:
👍

PR #13246, opened 2016-06-08
Hosted thrasher - multi

Calanthe:
## What does this change?

This PR introduces a new hosted thrasher traffic driver with multiple elements (up to 4). At some point the current hosted thrasher (with single element) will be removed and replaced by multiple one.
## What is the value of this and can you measure success?

With the new thrasher we will be able to set up links to multiple hosted pages.
## Screenshots

One element on desktop:
screen shot 2016-06-08 at 10 32 00

One element on mobile:
screen shot 2016-06-08 at 10 32 12

Two elements on desktop:
screen shot 2016-06-08 at 10 28 22

Two elements on mobile:
screen shot 2016-06-08 at 10 12 33

Three elements on desktop:
screen shot 2016-06-08 at 10 22 10

Four elements on desktop:
screen shot 2016-06-08 at 10 12 15

The mobile version for 3 and 4 elements will always show only 2 first elements.
The play icon and video times will be empty, when no video length is specified in DFP.
Template in DFP: https://www.google.com/dfp/59666047#delivery/CreateCreativeTemplate/creativeTemplateId=10058007

Example ad is defined here: https://www.google.com/dfp/59666047#delivery/PreviewCreative/orderId=344882487&lineItemId=122823927&creativeId=92401152687

And can be seen after adding the `zosia05` adtest cookie: `/uk/technology?adtest=zosia05`

PR #13272, opened 2016-06-13
Sponsored thrasher bugfix on CODE

Calanthe:
## What does this change?

This PR fixes the bug related to the lack of hashed file on CODE after adding the `hosted-thrasher-multi.js` file in https://github.com/guardian/frontend/pull/13246


kelvin-chappell:
👍 
Perhaps this needs to be documented somewhere if it works.  Or it should break in dev if it won't work in prod, somehow.

johnduffell:
"it should break in dev if it won't work in prod" is always the best idea.  There is a useHashedBundles option to make things use the hashed files which might catch this kind of issue, but noone would want to in dev due to the speed of compilation.

PR #13273, opened 2016-06-13
Remove some complete A/B tests, extend switch and add A/B Audience test

gtrufitt:
## What does this change?
- Removes 2 un-used A/B tests
- Extends switch on low friction sport V2 test to get more data
- Adds an audience test A/B test to help us test the A/B test audience coming from the data lake

@samdesborough


desbo:
👍

PR #13275, opened 2016-06-13
Remove race condition from comment loading

gtrufitt:
## What does this change?

I realised there was another race condition bug where we would try to do stuff with comments before they were inserted into the DOM. 
## Request for comment

@jbreckmckye @rich-nguyen 

View this with [whitespace ignored](https://github.com/guardian/frontend/pull/13275/files?w=1)

### DIFF
--- a/static/src/javascripts/projects/common/modules/discussion/comments.js
+++ b/static/src/javascripts/projects/common/modules/discussion/comments.js
@@ -13,7 +13,8 @@ define([
     'common/modules/discussion/whole-discussion',
     'common/modules/ui/relativedates',
     'common/modules/user-prefs',
-    'common/views/svgs'
+    'common/views/svgs',
+    'Promise'
 ], function(
     bean,
     bonzo,
@@ -29,7 +30,8 @@ define([
     WholeDiscussion,
     relativedates,
     userPrefs,
-    svgs
+    svgs,
+    Promise
 ) {
 'use strict';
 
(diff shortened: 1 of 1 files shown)


jamesgorrie:
Don't you need to use this promise somewhere to take affect?
Or is that in another PR?

static/src/javascripts/projects/common/modules/discussion/comments.js
```diff
@@ -198,26 +200,29 @@ Comments.prototype.fetchComments = function(options) {
 
 
 Comments.prototype.renderComments = function(resp) {
-
     // The resp object received has a collection of rendered html fragments, ready for DOM insertion.
     // - commentsHtml - the main comments content.
     // - paginationHtml - the discussion's pagination based on user page size and number of comments.
     // - postedCommentHtml - an empty comment for when the user successfully posts a comment.
+    return new Promise(function (resolve) {
```
piuccio: or you could use `fastdom-promise` so that your DOM manipulations are inside a fastdom.write

jbreckmckye:
I'm not really clear how this PR changes any behaviour. And isn't the DOM operation synchronous already?

jbreckmckye:
I'm not clear how this PR fixes any behaviour. The DOM operation was always synchronous - the only thing I can see happening is the `then` in `discussion.js:loadComments` (maybe?) executing a tick later, because that (might?) be the behaviour of whatever promise implementation is in play (but maybe not the native constructor? I don't know).

Can you provide more details about exactly which functions are in contention with each other?

gtrufitt:
@jbreckmckye You're right, something else is at play here. BRB.

gtrufitt:
> then in discussion.js:loadComments (maybe?) executing a tick later, because that (might?) be the behaviour of whatever promise implementation is in play

This could be it, I need to test though

PR #13276, opened 2016-06-13
Refactor manual inline

regiskuckaertz:
This change builds on previous work to rid ourselves of redundant and inefficient code in ads creatives. This time it tackles inline components, the first of which is the manual one.

picture 1

Still to come: books and events.

@Calanthe @JonNorman @rich-nguyen


static/src/stylesheets/module/commercial/_adverts.scss
```diff
@@ -362,6 +362,50 @@
...
+        > .inline-icon {
+            margin-bottom: $gs-baseline / 2;
+        }
+    }
+
+    .adverts__title {
+        @include fs-header(1);
+    }
+
+    .advert > *:not(.advert__image-container) {
```
sndrs: don't need the `*`

sndrs: do you need the `.advert >`?

regiskuckaertz: Fair enough. The `.advert` is necessary in that case yes.

Calanthe:
👍  Although the `template.js` file is getting bigger and bigger. Maybe at some point it would be nice to reorganise it.

regiskuckaertz:
Right. A looooooot of it I added to normalise the names of properties in the JSON object that is sent from DFP. Sometimes it's `baseUrl`, sometimes `base_url`, sometimes `base__url`. That is maddening 😠 😭

PR #13277, opened 2016-06-13
Fix newRequestData bug

JonNorman:
We are making a call to a prototype method on `userFeatures` without providing the `userFeatures`. This PR fixes that.

@jbreckmckye @regiskuckaertz


jbreckmckye:
👍

PR #13279, opened 2016-06-13
Default comment ordering to oldest first

nicl:
## What does this change?

The default sort ordering for comments -> oldest first.

The hope is that it will reduce abusive behaviour as the oldest comments are more likely to be moderated/decent. **We have no way to easily measure this as things stand.**

No.
## Request for comment

@guardian/discussion @crifmulholland @alastairjardine let's discuss this with the Editoria/Mod people and decide whether to deploy.

### DIFF
--- a/common/app/conf/switches/FeatureSwitches.scala
+++ b/common/app/conf/switches/FeatureSwitches.scala
@@ -316,6 +316,16 @@ trait FeatureSwitches {
     exposeClientSide = true
   )
 
+  val DiscussionOrderByOldest = Switch(
+    SwitchGroup.Feature,
+    "discussion-order-by-oldest",
+    "If on, comments default sort ordering will be oldest first. If off, then the default is newest first",
+    owners = Seq(Owner.withGithub("nicl")),
+    safeState = Off,
+    sellByDate = never,
+    exposeClientSide = true
+  )
+
   val MissingVideoEndcodingsJobSwitch = Switch(
     SwitchGroup.Feature,
     "check-for-missing-video-encodings",
--- a/static/src/javascripts/projects/common/modules/discussion/comments.js
+++ b/static/src/javascripts/projects/common/modules/discussion/comments.js
@@ -70,7 +70,7 @@ Comments.prototype.defaultOptions = {
     discussionId: null,
     showRepliesCount: 3,
     commentId: null,
-    order: 'newest',
+    order: config.switches.discussionOrderByOldest ? 'oldest' : 'newest',
     threading: 'collapsed'
 };
 


jamesgorrie:
I wonder if you could have some sort of moderation line...
Similar to when you have new messages. "Moderated till here".
moderated-till-here

It's also interesting that when we say "collapsed" we mean we'll show the first three - I wonder if replies are not moderated as quickly, so a way of saying I just want meaningful comments related to the article is to say don't show me replies?

jamesgorrie:
Code wise 👍

nicl:
@jamesgorrie thanks! I've added a switch to allow for quick toggling.

Re: moderation line, it's a really interesting idea. It might be an option for watched discussions, but not for other things. Another idea that's been mentioned is segregating comments into groups - the first for 'trusted' users and the second for everyone else. And then perhaps defaulting to showing just trusted comments.

'Trust' might be a function of email validation, membership, account lifetime, commenting history, phone validation?, etc....

jamesgorrie:
Yea - trust does definitely sound like a good thing to follow up on.

It's a personal opinion - but, as the Guardian, we have a remit to help inform, educate, empower etc - for discussion conversations - potentially social media is the place to be. 

So trust is important - but we have to ask trust to what? Trust to be a real person? Trust to help inform? Trust to :trollface:

common/app/conf/switches/FeatureSwitches.scala
```diff
@@ -316,6 +316,16 @@ trait FeatureSwitches {
...
+  val DiscussionOrderByOldest = Switch(
+    SwitchGroup.Feature,
+    "discussion-order-by-oldest",
+    "If on, comments default sort ordering will be oldest first. If off, then the default is newest first",
+    owners = Seq(Owner.withGithub("nicl")),
+    safeState = Off,
+    sellByDate = never,
+    exposeClientSide = true
+  )
+
```
jamesgorrie: With this switch - is the intention of this to make a final decision?
If so - it might be worth making it a test switch so that we don't permanently add this complexity to the code.

nicl: @jamesgorrie where is a good place to do this? It is intended for testing, but it's difficult to formally A/B test so we were just going to try it and get feedback from Mods.

jamesgorrie: I would still make it an [A/B test switch](https://github.com/guardian/frontend/blob/master/common/app/conf/switches/ABTestSwitches.scala) because it's an A test of sorts.

The only difference is that it makes it clear that this isn't a feature, just something you're testing (which I think is a valuable distinguisher).

gtrufitt:
What's this waiting for @nicl ? Did we need moderator go-go still?

nicl:
@gtrufitt I need to add it as a feature switch. But also, I've been swayed a bit by the arguments around how we actually test the impact of this. I think I'll push it out anyway for a brief period!

rich-nguyen:
as far as I can remember, the default ordering option was as divisive as brexit. I foresee approx 50% of our commenters liking/disliking this change. :)

johnduffell:
I think it's worth trying, I personally use oldest first anyway, and people can always change it themselves if they prefer newest first.  might be worth thinking about how it might affect participation though.  In some ways I feel like discussion gives people too many options, and we should just standardise on one way to read (e.g. oldest first, threaded) which should help with caching too.

jamesgorrie:
> should help with caching too

Not really as it's just two cached reaults.

Although too many options is potentially right and has been a bit of a problem with the discussion platform since it has been reshaped many time.

I feel the one question that would be great to have an answer to is: what is the value the discussion platform is meant to bring?

If we decided it was for social interaction between users, newest first would make sense. If extra journalistic value, then potentially most related to the story, if it was opinion, I have no idea, but potentially different.

I am not too close to the project - so this is just my opinion.

piuccio:
how about default sorting by most upvoted?

johnduffell:
yeah agreed with james, I think we need to start by knowing what discussion can do for our journalism and then work backwards from here.  It just popped into my mind because I removed the oldest first option for live blogs about 3/4 months ago and saved lots of code (and caching), and I noticed some feedback today asking for oldest first, which made me realise how little it added to the journalism.

gtrufitt:
Damnit @piuccio

johnduffell:
@piuccio watch out for people fishing for upvotes! 😝

humaislam:
What next for this PR @nicl?

nicl:
Closing as there are lots of questions here and I'm not convinced the turn-on-and-see approach is adequate to answer them!

PR #13281, opened 2016-06-13
Application loaders batch 1

alexduf:
## What does this change?

This PR removes the `Global` object of the Admin application. Doing it for only one app also prepares the path for all the others, so I expect the following PRs to be mostly about creating Application loaders rather than touching all the lifecycles again.

As far as I'm aware ( @cb372 or @mchv could confirm maybe?) in Play 2.5 there is now proper way to trigger a processing when the application starts other than using an application loader, or relying on a class to be loaded by the class loader (which isn't always deterministic). I opted for the first one: I trigger my start events in the `load` method of an application loader. At that stage, `Play.current` isn't set, so any piece of code relying on its import or explicit usage will fail.

I had to modify `AkkaAsync`, `Jobs` and `LogstashLifecycle` as they were all having a direct or indirect dependency on a running app (`Play.current`). For each of those I created the appropriate companion object that keeps the same behaviour as previously so we can move away from it smoothly. Dependency injected applications are using the injected class, and the application that are still using `Global`s can use these companion objects.

_Nota: I recommend to read that PR commit per commit as I grouped modifications together for readability_

Less `Global` objects => another step towards Play 2.5
## Request for comment

@TBonnin @johnduffell


common/app/common/akka.scala
```diff
@@ -3,31 +3,38 @@ package common
...
 
 object AkkaAgent extends ExecutionContexts {
   def apply[T](value: T): Agent[T] = Agent(value)
 }
 
-object AkkaAsync extends ExecutionContexts {
+class AkkaAsync(env: Environment, actorSystem: ActorSystem) {
+  implicit val ec: ExecutionContext = actorSystem.dispatcher
 
   def apply(body: => Unit): Unit = after(1.second){ body }
```
TBonnin: Is the intend to rename `apply` into `after1s` and delete it eventually?

alexduf: Yes I found that using the apply method on a class doesn't bring much readability.
`AkkaAsync { blah }` => `akkaAsync { blah }`
For some reason it just looks dodgy, especially as you can name your object as you please.

Also, "akkaAsync.apply" isn't really self-explanatory about what it does, so yes I think the apply function should be removed. I would like to mark it as deprecated but that would raise a compilation error given our compilation options.

TBonnin:
Looks good to my novice eyes :) 👍

common/app/common/akka.scala
johnduffell: that's great for readability, scala can really let you write confusing code otherwise - I'd probably have made the old "apply" call the new after1s method, and left a comment to say it's going away, but that's because I might forgot 😛

project/Dependencies.scala
```diff
@@ -42,6 +42,7 @@ object Dependencies {
   val jodaConvert = "org.joda" % "joda-convert" % "1.2"
   val jSoup = "org.jsoup" % "jsoup" % "1.8.3"
   val liftJson = "net.liftweb" %% "lift-json" % "2.6-RC2"
+  val macwire = "com.softwaremill.macwire" %% "macros" % "2.2.3" % "provided"
```
johnduffell: what is "provided"? I can't find any clear info online

johnduffell:
great, thanks for splitting it up, you will win the git power users award now! 👍

project/Dependencies.scala
alexduf: It means it's not included in the classpath. In theory it's made to be used when you create a library and you don't want to bring transient dependencies. Now to be completely honest I copied and pasted from it the [macwire documentation :)](https://github.com/adamw/macwire#installation-using-with-sbt)

It makes sense as that library is only going to be used at compile time to modify the syntax tree (through macros), and is therefore useless at run time.

mchv: Why do you need `macwire ` ?
What is the benefit of 
```scala
lazy val jobScheduler = wire[JobScheduler]
```
compared to 
```scala
lazy val jobScheduler = new JobScheduler
```
?

alexduf: Technically I don't _need_ it, but it does bring automatic wiring. In the case you selected it's straightforward to create the `JobScheduler`.

It gets a lot more messier when you need to inject components like `WsClient`, `ExecutionContext` and such.
A good example is `AdminHttpErrorHandler` or `AdminLifecycle`, or even the `Router`
modifying the route file (even the order of lines) can modify the order of the parameters in the constructor, or add new parameters, which will trigger a compilation error (that won't be caught by Intellij) whereas macwire is handling that wiring for you at compile time.

So do we need it? no, is it nice? in my opinion yes, but I'm open to hear more about why we shouldn't use it

common/app/common/akka.scala
alexduf: Good point, let me fix that

project/Dependencies.scala
mchv: > A good example is AdminHttpErrorHandler or AdminLifecycle, or even the Router
modifying the route file (even the order of lines) can modify the order of the parameters in the constructor, or add new parameters, which will trigger a compilation error (that won't be caught by Intellij) whereas macwire is handling that wiring for you at compile time.

I really don't get this point. It it triggers a `compilation` error you should see it immediately with `sbt`. If you don't see a compilation error with `Intellij` that you see with `sbt` then `Intellij` is broken.

mchv: > why we shouldn't use it

Only my opinion:
 - increase your compile time
 - add [accidental complexity](https://en.wikipedia.org/wiki/No_Silver_Bullet) (new dsl, macro, new library) that does not solve any problem (you **should** see any compilation error without it)

alexduf: >  If you don't see a compilation error with Intellij that you see with sbt then Intellij is broken.

That wouldn't be the first time :smile: implicits and managed code (like the router) aren't handled nicely. (which is fine really, it's more of an edge case)

> I really don't get this point.

I'm saying as the needs of a component evolves and its dependencies change (all the time), it's nice to have the wiring done automatically, especially for components that have many dependencies, or repetitive dependencies. I will soon have to add `Environment` to pretty much every Controller of the application, and I would appreciate for my own sanity to have that done magically :smiley:.

> increase your compile time

I don't think it's significant enough to be a problem

> add accidental complexity

True, it's not a perfect solution

> that does not solve any problem (you should see any compilation error without it) 

It doesn't solve all wiring situations but it simplifies a whole bunch of it, which is an improvement in my opinion.

I'll give it a go without it for this PR, but if find myself in wiring hell during the following work I may re-introduce it.

mchv: @alexduf  Thanks for your reply!
Feel free to merge with your changes 😃 as this is only my opinion.
From you response I understand better the benefit which is to avoid you to write boilerplate code of constructor initialisations with dependencies correctly passed from instances, which is faire enough!

alexduf: Cheers, If I see I'm getting too many questions or sceptical reactions about macwire I'll reconsider its usage, nothing is set in stone :)

admin/app/AdminAppLoader.scala
```diff
@@ -0,0 +1,68 @@
...
+    components.application
+  }
+}
+
+trait AppComponents extends BuiltInComponents with NingWSComponents {
+  lazy val appIdentity = ApplicationIdentity("frontend-admin")
+  lazy val appMetrics = ApplicationMetrics()
+  implicit lazy val executionContext: ExecutionContext = actorSystem.dispatcher
+  // this is a workaround to make wsapi available to the injector
+  override lazy val injector: Injector = new SimpleInjector(NewInstanceInjector) + router + crypto + httpConfiguration + wsApi
```
cb372: I don't understand why you need this workaround. I thought mixing in `NingWSComponents` did this for you.

common/app/common/akka.scala
```diff
@@ -22,7 +22,9 @@ object AkkaAgent extends ExecutionContexts {
 class AkkaAsync(env: Environment, actorSystem: ActorSystem) {
   implicit val ec: ExecutionContext = actorSystem.dispatcher
 
-  def apply(body: => Unit): Unit = after(1.second){ body }
+  // "apply" isn't expressive and doesn't explain what it does.
+  // If you were considering using that function, use after1s instead as it doesn't leave any ambiguity.
+  def apply(body: => Unit): Unit = after1s(body)
```
cb372: Could add a `@deprecated` as well

cb372:
I'm late to the party, but this is great! It all looks fine to me.

>As far as I'm aware ... in Play 2.5 there is no proper way to trigger a processing when the application starts other than using an application loader,

Yes, any processing that you want to run at app startup should be run in the application loader's `load` method.

common/app/common/akka.scala
alexduf: I would if I could, but the project adds the option -Xfatal-warnings in the [compiler configuration.](https://github.com/guardian/frontend/blob/master/project/Prototypes.scala#L23)

admin/app/AdminAppLoader.scala
alexduf: Because the code still uses `Ws.url` which is then making use of the injector to resolve `wsClient`. So as long as _some_ components aren't passed upon creation, we need that thing here.

To be honest I think it's almost a bug in play 2.4, but I don't see them fixing it now they're working on play 2.6, even if I report it.

alexduf:
Great, thanks @cb372 !

PR #13284, opened 2016-06-14
Turn press cross account storage on by default

piuccio:
## What does this change?

Mostly nothing, the switch was on for a month. Just change the default

You name it


piuccio:
@janua mind having a look? The switch expires next week

janua:
:+1:

PR #13126, opened 2016-05-31
Do not add a messenger share button on amp

mchv:
## What does this change?

Following the add of [messenger share button](https://github.com/guardian/frontend/pull/13089) AMP pages does not validate anymore 😞 
For instance [this page](https://amp.theguardian.com/football/blog/2016/may/31/milan-daniele-massaro-barcelona-golden-goal#development=1), show the following error:

```
https://amp.theguardian.com/football/blog/2016/may/31/milan-daniele-massaro-barcelona-golden-goal:1873:0 Invalid URL protocol 'fb-messenger:' for attribute 'href' in tag 'a'.   
```

This seems to be an `AMP` issue, as it should in theory allow `fb-messenger` protocol.
With the associated changes we do not display the button for `AMP` requests.

@johnduffell @NataliaLKB @stephanfowler

PR #13285, opened 2016-06-14
Revert "Do not add a messenger share button on amp"

mchv:
`fb-messenger` support has now been released ([see in the validator file](https://cdn.ampproject.org/v0/validator.js)) so we could remove workaround of #13126


johnduffell:
great! thanks! 👍

NataliaLKB:
👍 Awesome!

PR #13287, opened 2016-06-14
Comments prevent loop

gtrufitt:
## What does this change?

There is a [discussion API bug](https://github.com/guardian/discussion-api/issues/395) that means we do not return the correct results when requesting some comments - which meant there was a loop that when checking for a particular comment, if it didn't exist, we would send the same request.

This prevents that loops, shows an error message to the user and pings Sentry.

### DIFF
--- a/static/src/javascripts/projects/common/modules/discussion/loader.js
+++ b/static/src/javascripts/projects/common/modules/discussion/loader.js
@@ -401,7 +401,7 @@ Loader.prototype.initPagination = function() {
     });
 };
 
-Loader.prototype.gotoComment = function(id) {
+Loader.prototype.gotoComment = function(id, fromRequest) {
     var comment = $('#comment-' + id, this.elem);
     var thisLoader = this;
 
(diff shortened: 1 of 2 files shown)


static/src/javascripts/projects/common/modules/discussion/loader.js
```diff
@@ -421,9 +421,25 @@ Loader.prototype.gotoComment = function(id) {
...
+        scroller.scrollToElement(qwery('.js-discussion-toolbar'), 100);
+        fastdom.write(function(){
+            $('.js-discussion-main-comments').prepend('<div class="d-discussion__message d-discussion__message--error">The comment you requested could not be found.</div>');
+        });
+
+        // Capture in sentry
+        raven.captureMessage('Comment doesn\'t exist in response', {
+            level: 'error',
+            extra: {
+                commentId: id
```
gidsg: Wonder if there's any more context we can include here like sort options, or the URL requested in NextGen API?

gidsg:
LGTM 🚨 👍

static/src/javascripts/projects/common/modules/discussion/loader.js
gtrufitt: I agree, I am going to merge this so we are no longer loopdalooping, then I will add that and I also want to add that same info to a contact us link in context. Cheers.

nicl:
Are we sure that the error can't occur when going via top comments (aka highlights/staff picks)? I.e. at the moment the logic is:

    goto comment
      if found -> show + focus
      else if !fromRequest (i.e. permalink) -> loop
      else -> error message

Instead, we could do:

    goto comment
      if found -> ...
      else -> attempt load and if fails error

I can't think why the issue/error should only affect permalinks and not staff picks?

PR #13288, opened 2016-06-14
Add headers in request logs

TBonnin:
## What does this change?

Adding headers info into request logs

More info in logs (like user-agent, referrer, cookie, etc...) that are very useful when debugging

No
I am going to monitor how this change affects the ELK cluster storage
## Screenshots

screen shot 2016-06-14 at 14 42 36
## Request for comment

@johnduffell @jamespamplin


johnduffell:
why not 👍

PR #13289, opened 2016-06-14
Track ad state

regiskuckaertz:
Currently, it is impossible to know when house creatives are rendered. This is problematic because, between the time when DFP triggers the `slotRenderEnded` and the time the creative is actually rendered, a lot can happen: an API call, some javascript setup, etc. Ideally we want to know exactly when all the DOM mutations have occurred, so that we can confidently report rendering to the outside world.

A second problem is that the means by which outside modules know when an ad is rendered is via the `dfp:rendered` event. Events are transient and so, if one module happens to bind a listener too late, then it will wait forever. We tried to circumvent this in the past by memoizing a promise around the event, but that is only carrying the problem over to the point when the promise is created. A third problem related to this one is that it was impossible to query the state of an ad, because everything about it is destroyed once it is rendered.

This PR addresses all these things and then some:
- all creatives now return a promise that resolves when they all DOM mutations have occurred
- adverts and their status are kept in memory _ad infinitum_ (although I take care of empty slots by removing references to objects that are not necessary anymore)
- there are now two separate flags, `isLoading` and `isRendering`, to query the current state of an advert (also, `isHidden`, because even hidden ads are tracked)
- both flags have a corresponding promise, `whenLoaded` and `whenRendered`, that are used by external APIs `trackAdLoad` and `trackAdRendered`

Also
- I move the sponsorship code out of dfp-api, nothing to do there
- Nodes are nodes, not `Bonzo` instances, and are passed as is across functions: most of the time we operate on a single node at a time, standard APIs are much more efficient in this cae
- as a consequence, a lot of lodash routines were not necessary anymore
- **do not lazyload ads that are too far up the viewport**
- use native events rather than those proxied via `fastdom`. This feels like a wrong idea given the rendering pipeline is fed by input events _and then possibly_ an animation frame.

cc @jbreckmckye @rich-nguyen


static/src/javascripts/projects/common/modules/commercial/creatives/commercial-component.js
```diff
@@ -167,32 +169,32 @@ define([
...
+                resolve(true);
+            }
 
-        function onError() {
-            fastdom.write(function () {
-                this.adSlot.style.display = 'none';
-            }, this);
-        }
+            function onError() {
+                resolve(false);
```
jbreckmckye: Should this be a `rejection`?

rich-nguyen:
There's quite a lot of `resolve(false)` in here. Is that normal? Why don't we use `reject`, which seems a more idiomatic way to resolve a promise with failure?

regiskuckaertz:
There was no error detection previously, that is why I am not rejecting the promise. I wouldn't throw an exception if an ad isn't rendered, but if this information is interesting to report then I can definitely amend the `catch` handler.

static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
```diff
@@ -671,37 +622,61 @@ define([
...
+            var advert = getAdvertById(id);
+            if (withHidden || !advert.isHidden) {
+                advertsById[id] = advert;
             }
+            return advertsById;
+        }, {});
+    }
+
+    function initAdvert(adSlotNode) {
+        return Object.seal({
```
rich-nguyen: Could this catch people out? new property changes will fail silently. I think it's a good intro to immutablility, but it's really hard to know when I'm dealing with a sealed or non-sealed object.

rich-nguyen:
That's great, much more sane. 👍 

A few small points:

- Removing bonzo instances could benefit perf, I'm not sure by how much. As developers, when should we use bonzo? We need a consistent approach. If we need to avoid it during perf-sensitive parts of the codebase, we should think of a way to make this clear to the reader.
- Some of the `Advert` objects' state management has now moved outside the function into [`dfp-api`](https://github.com/guardian/frontend/pull/13289/files#diff-844a43280e1947d8dd7cc7a480fd0551R154), is that going to be maintainable? It makes the lifecycle of an `Advert` spread wider over multiple files.

static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
regiskuckaertz: We are not running in strict mode? 😲  This is a perf killer.

regiskuckaertz:
Thanks 😃 

* I've been doing a lot of research on the topic. Without macros, a monadic signature is the cleanest and most concise one can get (I tried implementing Clojure's `->>` using sweet.js, this is so much better). Not many people like monoids because they compose from right to left (see my refactoring of `detect` for examples). So, I think an API like bonzo has good form and we should keep it. What I don't like is that it forces you to write things like `$adSlot[0]` all the time. That's why, when I operate on one node at a time, I prefer straight Web API.

* That is the bane of event-based programming 😭  I don't know any good solution for that except observables or channels. Or perhaps, I could do sth like this:

```
function initAdvert(... params...) {
  var slot, sizes, etc.

  return Object.freeze({
    setSizes: function() { ... },
    updateState: function() { ... },
    etc.
  });
}
```

jamesgorrie:
> That is the bane of event-based programming

Rx?

regiskuckaertz:
@jamesgorrie yes 😄

static/src/javascripts/projects/common/modules/commercial/dfp/breakout-iframe.js
```diff
@@ -27,51 +25,43 @@ define([
...
-    config,
     fastdom
 ) {
 
     /**
      * Allows ad content to break out of their iframes. The ad's content must have one of the above breakoutClasses.
      * This can be set on the DFP creative.
      */
-    function breakoutIFrame(iFrame, $slot) {
+    function breakoutIFrame(iFrame, slot) {
```
jbreckmckye: This file no longer exists...

regiskuckaertz: done, yippie!

PR #13292, opened 2016-06-14
Add referring section to video-start tracking properties

kelvin-chappell:
Tracking referring section for video plays, to measure effectiveness of traffic drivers.

image


jamesgorrie:
Just wondering who requested this as we're trying to keep our metrics tidy?
Also trying to keep them consistent between Ophan and Omniture.

jamesgorrie:
> who requested this

Should read - what report is this going to be used for.

kelvin-chappell:
@jamesgorrie: This was requested by Jono Caseley for hosted campaign reports.  I could restrict it to hosted content and put it behind a switch, especially as this new content type is all experimental at the moment so requirements are in flux.

jamesgorrie:
I don't think it matters, we just need to make sure that we don't rely on it in non-hosted world, which, I suppose the best way to do that, is not have it there in the first place.

👍  for now - but if there's more it might be good to make it more explicit.

PR #13293, opened 2016-06-14
Rename ad slots on mobile fronts for Fabric 

rich-nguyen:
This attempts to resolve a problem where fabric creatives are alternating on mobile fronts. This happens when a slot has multiple sizes, and a line item has several candidate creatives for that slot. We do not know which creative will appear in that slot.

The best solution is to have consistent slot names so that the ad ops team can create line items and creatives which target specific ad slots. To do this, mobile and desktop must use a consistent slot naming system. Right now, the slot `inline1` has multiple meanings. In articles, this slot is used in the body, inline. On fronts, `inline1` is the slot after the top slot in desktop. Unfortunately, there is no top slot on mobile fronts. Hence, this code introduces a top slot on mobile fronts, pushing the existing slots one position down.

**Before**
mobile fronts slot order = `[inline1, merchandising-high/middle, inline2, merchandising/bottom]`

**After**
mobile fronts slot order = `[top-above-nav, merchandising-high/middle, inline1, merchandising/bottom]`


static/src/javascripts/projects/common/modules/commercial/close-disabled-slots.js
```diff
@@ -0,0 +1,58 @@
...
+    qwery,
+    Promise,
+    $css,
+    fastdom,
+    commercialFeatures
+) {
+    var adSlotSelector = '.js-ad-slot';
+
+    return {
+        init: init
```
regiskuckaertz: For what purpose did you move this in a separate module? This feels over-engineered, especially since this module will linger in run once and then linger in memory forever.

fwiw, I keep a log of all incoming slots and mark them as hidden when necessary, in [my refactoring of DFP](https://github.com/guardian/frontend/pull/13289/files#diff-844a43280e1947d8dd7cc7a480fd0551R336). From there, we could reassign/refresh an existing slot manually, e.g. `top-above-nav'—removing it from the DOM does not remove it from memory, although for now I have taken care of it.

regiskuckaertz:
Aside from my question, this is top notch 👍

static/src/javascripts/projects/common/modules/commercial/close-disabled-slots.js
rich-nguyen: This was moved out of the `dfp-api` module to fix a race condition. I think this issue affects your [PR](https://github.com/guardian/frontend/pull/13289/files#diff-844a43280e1947d8dd7cc7a480fd0551R334) too, in `queueAdvert`. If we initialise `dfp-api`, and begin defining slots, then we must wait for all visually hidden slots to be removed from the dom before we use the gpt library. For most situations, the removed slots do not interfere with the new defined slots. However, this change introduces `top-above-nav` to mobile for fabric. So now the removed slots contains a `top-above-nav` (the old one rendered by the server as a top banner), and the new slots also contains a `top-above-nav` (the one on mobile fronts under the first container).

So the `close-disabled-slots` module has the same functionality as before, just wrapped in a promise. This means we can relocate this system further up commercial boot's `then` chain.

PR #13295, opened 2016-06-14
Hosted thrasher gradient removal

Calanthe:
This PR removes the gradient on the hosted thrasher images, because the colours in the image will be adjusted manually.
## Screenshots

Before:
screen shot 2016-06-14 at 18 03 20

After:
screen shot 2016-06-14 at 18 03 57


lps88:
👍

PR #13296, opened 2016-06-14
Refactor commercial metrics to retain granularity of feed reporting

JonNorman:
## What does this change?

This PR precedes a future PR that will report/chart commercial feed fetching/parsing failures. Currently there is logic to report this in aggregate (i.e. number number of failed fetches/parses). This has been refactored such that the feed-level information is kept and can then be used to build pretty charts (in a separate PR).

Supports future work for having a pretty dashboard or alerts if we want to set them up

No


JonNorman:
I've decided to use Kinbana to chart this stuff which affects slightly the approach set up here, closing for now.

PR #13300, opened 2016-06-15
Don't have an immersive main media on interactives

NataliaLKB:
This pull request fixes a bug where an immersive interactive looks like this:
image

It is supposed to look like this:
image


katebee:
👍  looks good!

PR #13301, opened 2016-06-15
Admin fix: Add the actor system to the injector

alexduf:
## What does this change?

It fixes the admin module.

We use WS.url and Akka.system everywhere and both these call are using the internal injector of the application. I can't reasonably fix the use of WS.url and Akka.system everywhere so I decided to add it to the injector.
Thanks @mchv for your brain :)
## Request for comment

@TBonnin @johnduffell @piuccio


TBonnin:
👍

PR #13302, opened 2016-06-15
Log dapi request latency as its own field

TBonnin:
## What does this change?

Log dapi request latency as its own field...

...so we can graph it in Kibana 😀 

No
## Request for comment

@johnduffell @alexduf


alexduf:
Looks good to me :+1:

PR #13303, opened 2016-06-15
Add window.guardian object setup for adblocker detection

jbreckmckye:
Currently, Jasmine tests are cluttered with exceptions raised by `detect.js`:

image

These exceptions don't fail any tests, but they do add noise and make real errors harder to find.

This happens because the code in question expects an array on the `guardian` object, normally created by a server-inserted script. This doesn't run in the test environment, so the array never exists, and calls to manipulate it through fatal errors.

As it turns out, though, there's a very straightforward solution: a simple `setup` script at the root of the `spec` folder can be wrangled to create the same config object we'd expect when running the code on an actual Guardian page.

### DIFF
--- a/static/test/javascripts/setup.js
+++ b/static/test/javascripts/setup.js
@@ -35,7 +35,10 @@ window.guardian = {
         }
     },
     css: {},
-    adBlockers: {}
+    adBlockers: {
+        active: undefined,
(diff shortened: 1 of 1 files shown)


JonNorman:
image

:+1:

PR #13304, opened 2016-06-15
Immersive articles should always have white headlines

NataliaLKB:
Originally I was under the impression that immersive articles were always long reads, and always feature tone. However, I was very mistaken! Apparently neither of those things are true.

Because of thinking, currently on the site if someone makes an immersive article that is news tone, the headline will appear grey, not white. This PR decouples that and makes sure the headlines are always white.
## Screenshots

What could happen:
image

How it should be:
image
## Request for comment

@na7hangood


sammorrisdesign:
👍

PR #13305, opened 2016-06-15
Refreshing the ad-block message at bottom of page.

paulbrown1982:
## What does this change?

This change updates the ad-block message which at the bottom of an article page for non-paying members who block ads.

This change will redisplay the message banner to any users who have previously closed it.

This only affects desktop-sized screens.
## Screenshots

picture 110
## Request for comment

cc @tomverran @dominickendrick @nlindblad


dominickendrick:
👍

PR #13307, opened 2016-06-15
add video-inbody-embed component

jamesgorrie:
## What does this change?

Adding a component to inbody videos

This will allow us to measure the effectiveness of adding video links to this - it is an editorial ask, and if it is super effective, we could automate it, if it's not, we can save some time not doing it.

Nope.
## Screamshots

🙀 
## Request for comment

@akash1810


jamesgorrie:
👍 from @akash1810

PR #13310, opened 2016-06-15
Whitelist headers log

TBonnin:
## What does this change?

I bundled 2 changes in this patch maybe I should have made 2 PRs :/
1. Refactor Logging custom fields to allow sending fields with multiple types (Int, Long, Double, String). Before this patch you could only send fields of the same type.
2. Limit logging whitelisted headers and Guardian specific headers

Logging all the headers leads to log a lot of crap🔪  that take a lot of space

No
## Request for comment

@alexduf

### DIFF
--- a/common/app/filters/RequestLoggingFilter.scala
+++ b/common/app/filters/RequestLoggingFilter.scala
@@ -14,27 +14,26 @@ class RequestLoggingFilter extends Filter with Logging with ExecutionContexts {
 
   case class RequestLogger(rh: RequestHeader)(implicit internalLogger: Logger, stopWatch: StopWatch) {
     private lazy val pseudoId = Random.nextInt(Integer.MAX_VALUE)
-    private def customFieldsMarkers(): LogstashMarker = {
+    private lazy val customFields: Map[String, Any] = {
       val headersFields = rh.headers.toMap.map {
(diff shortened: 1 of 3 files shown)


common/app/filters/RequestLoggingFilter.scala
```diff
@@ -4,37 +4,50 @@ import common.{ExecutionContexts, Logging, StopWatch}
...
-
 class RequestLoggingFilter extends Filter with Logging with ExecutionContexts {
 
   case class RequestLogger(rh: RequestHeader)(implicit internalLogger: Logger, stopWatch: StopWatch) {
     private lazy val pseudoId = Random.nextInt(Integer.MAX_VALUE)
-    private def customFieldsMarkers(): LogstashMarker = {
-      val headersFields = rh.headers.toMap.map {
-        case (headerName, headerValues) => (s"req.header.$headerName", headerValues.mkString(","))
+    private val headersFields: List[LogField] = {
+      val whitelistedHeaderNames = List(
```
alexduf: If you use a Set here the filter that happen line 31 will go a lot faster.

TBonnin: valid point. will change

alexduf:
aside from small optimisation, all good :+1: !

PR #13311, opened 2016-06-15
Hosted By gallery prototype

lps88:
## What does this change?
- Scrolling on the images will fade them in/out (page is behind a switch still so yet unfinished)
- Improve page styling/layout

image
## Request for comment

@Calanthe


static/src/javascripts/projects/common/modules/commercial/hosted-gallery.js
```diff
@@ -0,0 +1,38 @@
...
+    bonzo,
+    fastdom,
+    $
+) {
+
+
+    function init() {
+        var $scrollEl = $('.hosted-gallery__scroll-container');
+        var $images = $('.hosted-gallery__image');
+        var noImages = $images.length;
```
jamesgorrie: noImages is a really confusing name - given that we only use it once, we could just use it there and make it explicit? or `numberOfImages`?

static/src/javascripts/projects/common/modules/commercial/hosted-gallery.js
```diff
@@ -0,0 +1,38 @@
...
+        if (!$scrollEl.length) {
+            return;
+        }
+
+        bean.on($scrollEl[0], 'scroll', function (e) {
```
jamesgorrie: You'd really want to `throttle` this.
You could use the [`window:throttledScroll`](https://github.com/guardian/frontend/blob/6af3c99aeee98156685296b817cb3ec1599e8bbb/static/src/javascripts/bootstraps/standard/main.js#L197) event too.

static/src/javascripts/projects/common/modules/commercial/hosted-gallery.js
```diff
@@ -0,0 +1,38 @@
...
+            var scrollTop = e.target.scrollTop;
+            var scrollHeight = e.target.scrollHeight;
+            var opacity = noImages * (scrollTop/scrollHeight) + 1;
+            fastdom.write(function () {
+                $images.each(function (im, index) {
+                    bonzo(im).css('opacity', opacity - index);
+                });
+            });
+        });
```
jamesgorrie: I don't know if it would help but I wrote an [inview method that could help with this](https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/common/utils/element-inview.js).

jamesgorrie:
Is this literally for the aesthetic?

Normally this type of behaviour is for lazy loading images - although this won't as transparent images load anyway.

Potential browser perf hit.

This wol

lps88:
@jamesgorrie Yeah, it the behaviour is something like this: http://benlongdendesign.co.uk/hosted-gallery/

I do need to look at performance/lazy loading, just wanted to get something onto code so that the team can look at it, very much prototyping stage still. Thanks for the tips!

static/src/javascripts/projects/common/modules/commercial/hosted-gallery.js
```diff
@@ -0,0 +1,39 @@
...
+
+
+    function init() {
+        var $scrollEl = $('.hosted-gallery__scroll-container');
+        var $images = $('.hosted-gallery__image');
+        if (!$scrollEl.length) {
+            return;
+        }
+
+        bean.on($scrollEl[0], 'scroll', debounce(function (e) {
```
Calanthe: Specify the number of milliseconds to delay in `debounce` so the function won't be fired that often and performance will be improved.

Calanthe:
👍

PR #13312, opened 2016-06-15
Request logging: use lazy val

TBonnin:
## What does this change?

Request logging: use lazy val so log field values like latency are computed when accessed and not
when Request logger instance is created

Fix bug where request latency is always 0

No
## Request for comment

@alexduf


alexduf:
Looks good :+1:

PR #13313, opened 2016-06-16
Improve Facebook creatives design

regiskuckaertz:
Pat wants them to look as much as possible to Guardian Frame ads, so I added the "Ad" pill on the top right and changed the typography accordingly.

picture 1

cc @JonNorman @jbreckmckye @rich-nguyen @Calanthe


static/src/javascripts/projects/common/modules/commercial/creatives/facebook.js
```diff
@@ -1,63 +1,84 @@
...
-            this.$adSlot[0].insertAdjacentHTML('beforeend', markup);
+                FB.Event.subscribe(
+                    'ad.error',
+                    function(errorCode, errorMessage, placementID) {
+                        reportError(new Error('Facebook returned an empty ad response'), {
+                            feature: 'commercial',
+                            placementID: placementID,
+                            errorMessage: errorMessage
+                        }, false);
+                        resolve();
```
jbreckmckye: Shouldn't this `reject`, so we can short circuit the `addLabel()` and `callback[...]` flows?

regiskuckaertz: Yep, good point.

static/src/javascripts/projects/common/modules/commercial/creatives/facebook.js
```diff
@@ -1,63 +1,84 @@
...
+                    }
+                );
+            });
 
-            if (this.params.trackingPixel) {
-                addTrackingPixel(this.$adSlot, this.params.trackingPixel + this.params.cacheBuster);
-            }
+            var markup = facebookTpl(Object.assign({ externalLink: svgs('externalLink') }, adUnits[this.params.placement]));
```
jbreckmckye: `Object.assign` is ES6, so it won't work on IE11 - amongst others.

static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
```diff
@@ -454,6 +454,8 @@ define([
             if (mobileAdSizes && mobileAdSizes.indexOf(adSizes.stickyMpu) > -1) {
                 stickyMpu($adSlot);
             }
+        } else if ($adSlot.hasClass('ad-slot--facebook')) {
+            $adSlot.addClass('ad-slot--fluid');
```
jbreckmckye: Given the way that `then` operations are run a tick after the promise they listen to resolves, will there be a frame where the advert appears without its requisite styling? Perhaps we could just use `ad-slot--fluid` right from the start?

static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
```diff
@@ -621,6 +623,7 @@ define([
     function shouldRenderLabel($adSlot) {
         return !$adSlot.hasClass('ad-slot--frame') &&
```
jbreckmckye: Is there a better way to communicate whether a component handles its own labelling? We are having to swap between different classes on the same advert just to control this behaviour.

static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
regiskuckaertz: Good point, I had to verify but no. The promise resolves when the callback is called, which can only happen after the elements has been added to the DOM.

static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
regiskuckaertz: Yes there is, but this is a bigger change that will have ramifications in other creatives. If you want take a stab at it 😃

static/src/javascripts/projects/common/modules/commercial/creatives/facebook.js
regiskuckaertz: damn, good spot. I meant to change that.

static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
jbreckmckye: That's the problem, though - if the elements are added to the DOM with `ad-slot--facebook`, they won't be full width on their initial render. I'm not sure if users will be able to perceive this in practice, though.

regiskuckaertz: The whole `callback` thing needs to be rethought, that's for sure.

jbreckmckye:
Great stuff 👍

PR #13314, opened 2016-06-16
Remove fluid adverts switch

regiskuckaertz:
@guardian/commercial-dev

### DIFF
--- a/common/app/views/support/Commercial.scala
+++ b/common/app/views/support/Commercial.scala
@@ -4,7 +4,7 @@ import common.Edition
 import common.commercial.{Branding, CardContent, ContainerModel, PaidContent}
 import common.dfp.AdSize.responsiveSize
 import common.dfp._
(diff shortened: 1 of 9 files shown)


common/app/views/support/Commercial.scala
```diff
@@ -35,7 +35,7 @@ object Commercial {
 
     def adSizes(metaData: MetaData, edition: Edition): Map[String, Seq[String]] = {
       val fabricAdvertsTop = Seq("88,71")
-      val fluidAdvertsTop = if (FluidAdverts.isSwitchedOn) Some("fluid") else None
+      val fluidAdvertsTop = Seq("fluid")
```
kelvin-chappell: Is this meant to have changed from an `Option` to a `Seq`?

regiskuckaertz: Yes, we just concat it a few lines below.

kelvin-chappell:
👍

PR #13315, opened 2016-06-16
tweak panic shedding sensitivity and better logging

johnduffell:
This increases the response time of latency shedding so it will start taking account of the latency of the requests starting with the very first one.  In the previous implementation it would start off from cold allowing all requests, and would take about 150 requests to get to 75% sensitivity.
I have also tweaked the sensitivity to a spike in latency, it will only take 13 requests to get 75% sensitive to the change.  That means if latency was zero for all requests, and suddenly became 1 second, after 13 requests at 1 second latency it would consider the latency to be 750ms.

Another tweak is to make it always allow healthchecks through.  This means that as long as the healthcheck is ok on the job (i.e. within the 4 second timeout) then the healthcheck will respond OK.  This gives a small chance that the latency monitor could get "stuck" rejecting all requests, but the healthcheck would be passing, but as it is machines are getting marked unhealthy just because they are overloaded.  If we get autoscaling right (should scale on high latency metric) then it should scale up anyway in that situation.

@TBonnin @alexduf @jamespamplin any comments?  I'll monitor to make sure it's not firing off too often.

### DIFF
--- a/common/app/conf/Filters.scala
+++ b/common/app/conf/Filters.scala
@@ -76,11 +76,11 @@ object Filters {
   // NOTE - order is important here, Gzipper AFTER CorsVaryHeaders
   // which effectively means "JsonVaryHeaders goes around Gzipper"
   def common: List[EssentialFilter] = List(
+    new RequestLoggingFilter,
     new PanicSheddingFilter,
     new JsonVaryHeadersFilter,
     new Gzipper,
     new BackendHeaderFilter,
-    new RequestLoggingFilter,
     new SurrogateKeyFilter,
     new AmpFilter
   )
--- a/common/app/conf/PanicSheddingFilter.scala
+++ b/common/app/conf/PanicSheddingFilter.scala
@@ -41,8 +41,20 @@ class PanicSheddingFilter extends Filter with Logging {
     val surgeFactor = surgeFactorCounter.get.ratio
     lazy val loadAverage = ManagementFactory.getOperatingSystemMXBean.getSystemLoadAverage
 
+    def logIt(message: Option[String]) = {
+      val debugMessage = s"$requestsInProgress requests in progress, averageLatency = ${latency}ms, surgeFactor = $surgeFactor (range is -$RANGE,$RANGE, over $WORST_ALLOWABLE_SURGE is a spike), loadAverage = $loadAverage"
+      logInfoWithCustomFields(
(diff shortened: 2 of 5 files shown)


common/app/conf/Filters.scala
```diff
@@ -76,11 +76,11 @@ object Filters {
   // NOTE - order is important here, Gzipper AFTER CorsVaryHeaders
   // which effectively means "JsonVaryHeaders goes around Gzipper"
   def common: List[EssentialFilter] = List(
+    new RequestLoggingFilter,
```
johnduffell: I moved this because it wasn't actually logging/timing the whole request/response chain.  Perhaps that was intended, but it wasn't quite realistic.

alexduf: :+1:

alexduf:
It looks good to me :+1: 
There are failing test but it looks like it's unrelated

common/app/conf/Filters.scala
TBonnin: 👏

common/app/conf/PanicSheddingFilter.scala
```diff
@@ -41,40 +41,51 @@ class PanicSheddingFilter extends Filter with Logging {
     val surgeFactor = surgeFactorCounter.get.ratio
     lazy val loadAverage = ManagementFactory.getOperatingSystemMXBean.getSystemLoadAverage
 
+    def logIt(message: Option[String]) = {
```
TBonnin: Can we find better than `It`? `logPanic` or something like this

TBonnin: Would adding a default parameter value `...message: Option[String] = None...` make call without message easier to read and write: `logIt()` instead of `logIt(None)`

jamespamplin:
:+1: looks good to me @johnduffell, thanks!

johnduffell:
test failure may be because the safe state is on again, so I've turned it off again

common/app/conf/PanicSheddingFilter.scala
johnduffell: I changed the name, but I prefer to err on the side of explicit parameters

johnduffell: "It" was a terrible name 😆

TBonnin: Sounds good. Default param also has some culprits

PR #13316, opened 2016-06-16
Remove sponsored badges from Immersive template

JonNorman:
## What does this change?

Currently a badge is included in both the immersive template as well as the article/long-read/etc template beneath. This results in having two badges on the page, which looks bad and introduces further problems of DFP getting confused and throwing a hissy fit.

Badges look better and ads are served to the page.
## Screenshots

**before**
image

**after**
image


jbreckmckye:
If we're confident that whatever the immersive header is paired with can provide its own badges (and it sounds like we are), then this seems good to release!

👍

kelvin-chappell:
Well spotted.  Presumably this is the first sponsored immersive.

JonNorman:
I believe so; we couldn't find any examples of a sponsored immersive on anything other than a gallery though,

PR #13317, opened 2016-06-16
Fixing images on the Minute

NataliaLKB:
## What does this change?

Due to a combination of changes a month or two ago the images on the US minute now appear broken.

This was the simplest fix I could think of.
## Screenshots

Before:
image

After:
image
## Request for comment

@gtrufitt can you take a look?


gtrufitt:
👍

PR #13318, opened 2016-06-16
Remove padding on inline ads

regiskuckaertz:
Before
picture 2

After
picture 4

@JonNorman @rich-nguyen @jbreckmckye @Calanthe


JonNorman:
:+1:

PR #13319, opened 2016-06-16
Rolex/Wimbledon ad only on desktops with a fixed height

regiskuckaertz:
Pat asked that (1) the Rolex ad does not show up on tablets and (2) the top slot has a fixed height on desktops.

(1) can be achieved in targeting in DFP, but (2) can only be done via a media query. The only solution I found was to therefore delay the ad until the `leftCol` breakpoint has been crossed—i.e. screens smaller than 1140px wide won't see the ad, even on desktops.

I'm happy to improve if someone has a better idea.

cc @guardian/commercial-dev

### DIFF
--- a/common/app/views/support/Commercial.scala
+++ b/common/app/views/support/Commercial.scala
@@ -33,8 +33,8 @@ object Commercial {
       metaData.id == "uk/technology"
     }
 
-    private def isWimbledonEnabled(metaData: MetaData) = {
-      metaData.id == "sport/tennis" || metaData.id == "sport/wimbledon"
+    private def isWimbledonEnabled(metaData: MetaData, edition: Edition) = {
+      (metaData.id == "sport/tennis" || metaData.id == "sport/wimbledon") && edition.id == "UK"
     }
 
     def adSizes(metaData: MetaData, edition: Edition): Map[String, Seq[String]] = {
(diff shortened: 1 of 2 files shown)


rich-nguyen:
👍  can't think of a better way :)

common/app/views/support/Commercial.scala
```diff
@@ -33,28 +33,36 @@ object Commercial {
       metaData.id == "uk/technology"
     }
 
+    private def isWimbledonEnabled(metaData: MetaData, edition: Edition) = {
+      (metaData.id == "sport/tennis" || metaData.id == "sport/wimbledon") && edition.id == "UK"
```
kelvin-chappell: Better if edition check is `edition == Uk`

common/app/views/support/Commercial.scala
```diff
@@ -33,28 +33,36 @@ object Commercial {
...
+      if (WimbledonTopAd.isSwitchedOn && isWimbledonEnabled(metaData, edition)) {
+          topSlotAdSizes + wimbledonLeftCol
+      } else {
+          topSlotAdSizes
+      }
     }
 
     // The sizesOverride parameter is for testing only.
     def cssClasses(metaData: MetaData, edition: Edition, sizesOverride: Seq[AdSize] = Nil): String = {
+      var wimbledonClasses = if (WimbledonTopAd.isSwitchedOn && isWimbledonEnabled(metaData, edition)) Seq("top-banner-ad-container--wimbledon") else Nil
```
kelvin-chappell: There only seems to be one css class here, so could be `wimbledonClass: String`.
And should be a `val`.

common/app/views/support/Commercial.scala
```diff
@@ -33,28 +33,36 @@ object Commercial {
...
       val classes = Seq(
         "top-banner-ad-container",
         "top-banner-ad-container--desktop",
         "top-banner-ad-container--above-nav",
         "js-top-banner-above-nav",
         "top-banner-ad-container--reveal"
-      )
+      ) ++ wimbledonClasses
```
kelvin-chappell: But then this becomes less neat.  Would have to use a fold or map/getOrElse.  So feel free to ignore last comment.  Apart from val.

kelvin-chappell:
👍

PR #13321, opened 2016-06-16
Reset omniture prop71 cookie

kelvin-chappell:
Keep persistent value of referred section property for different events on same page.

### DIFF
--- a/static/src/javascripts/projects/common/modules/analytics/omnitureMedia.js
+++ b/static/src/javascripts/projects/common/modules/analytics/omnitureMedia.js
@@ -13,6 +13,12 @@ define([
             return player.el().getAttribute(attributeName);
         }
 
+        function resetProp71Cookie() {
+            if (config.switches.hostedContentTracking && config.page.toneIds == 'tone/hosted-content') {
+                s.getPreviousValue(s.prop71, 's_prev_ch');
+            }
+        }
+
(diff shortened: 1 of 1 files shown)


static/src/javascripts/projects/common/modules/analytics/omnitureMedia.js
```diff
@@ -13,6 +13,12 @@ define([
             return player.el().getAttribute(attributeName);
         }
 
+        function resetProp71Cookie() {
+            if (config.switches.hostedContentTracking && config.page.toneIds == 'tone/hosted-content') {
+                s.getPreviousValue(s.prop71, 's_prev_ch');
```
rich-nguyen: what does this actually do? and would a test help spec the behaviour in future? i'm not sure if `s_pre_ch` makes sense alone, a comment might help

rich-nguyen:
👍  aside from comment

kelvin-chappell:
This is a really ugly horrible thing that I'm sure can be done much better, which is why I've put a switch around it to remind me either to remove it or improve it.  The actual behaviour is so weird I'm not sure I can spec it in a way that would reflect how it behaves in prod.
I'll put a comment in though to that effect, and to explain that `s_prev_ch` is omniture's way of keeping a reference to the referring section (channel).

PR #13324, opened 2016-06-17
And finally, split up dfp-api.js

jbreckmckye:
It seems, sadly, that this could be my last pull request to this project. As such I'd like to at least use it to resolve one of the grievances I have with the ball of mud that is `dfp-api.js`.

This isn't a comprehensive refactor. In an ideal world, I'd probably have split the file into three modules: 
- an `advertisingService` that does nothing except dispatch requests and set up the targeting and dependencies that requires.
- a `renderAds` module that breathes behaviour, interactivity and styles into the creatives that come back;
- something responsible for setting up our lazy loaded ads (and dispatching whatever is already in view). Maybe I'd call it `requestAdverts` or something.

Mindful of economy of time, however, I've just done the second - pull out the `renderAds` stage from `dfp-api` into its own dedicated module. This would be the largest of the three modules, and seems to offer the greatest opportunity for tangential concerns like ad styling to 'leak into' `dfp-api`.

`dfp-api` is still something of an nebulous 'bag of holding', but my hope is that this change makes it a _little_ easier to navigate and a _little_ more focused in its concerns.

@guardian/commercial-dev @uplne

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/dfp/render-advert.js
+++ b/static/src/javascripts/projects/common/modules/commercial/dfp/render-advert.js
@@ -1,0 +1,176 @@
+define([
+    'bonzo',
+    'qwery',
+    'raven',
+    'common/utils/$',
+    'common/utils/fastdom-promise',
+    'common/modules/commercial/ad-sizes',
+    'common/modules/commercial/ads/sticky-mpu',
+    'common/modules/commercial/dfp/apply-creative-template',
+    'common/modules/onward/geo-most-popular'
+], function (
+    bonzo,
+    qwery,
+    raven,
+    $,
+    fastdom,
+    adSizes,
+    stickyMpu,
+    applyCreativeTemplate,
+    geoMostPopular
+) {
+    /**
+     * ADVERT RENDERING
+     * ----------------
+     *
+     * Most adverts come back from DFP ready to display as-is. But sometimes we need more: embedded components that can share
+     * Guardian styles, for example, or behaviours like sticky-scrolling. This module helps 'finish' rendering any advert, and
+     * decorating them with these behaviours.
+     *
+     * This module is also responsible for adding advert labels.
+     */
+
+    var sizeCallbacks = {};
+
+    /**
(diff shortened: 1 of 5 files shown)


static/src/javascripts/projects/common/modules/commercial/dfp/render-advert.js
```diff
@@ -0,0 +1,161 @@
...
+     * @param adSlotId - DOM ID of the rendered slot
+     * @param slotRenderEvent - GPT slotRenderEndedEvent
+     * @returns {Promise} - resolves once all necessary rendering is queued up
+     */
+    function renderAdvert(adSlotId, slotRenderEvent) {
+        var size;
+        var $adSlot = $('#' + adSlotId);
+
+        // remove any placeholder ad content
+        var $placeholder = $('.ad-slot__content--placeholder', $adSlot);
```
regiskuckaertz: What do you think of moving these next 10 lines in a separate function?

static/src/javascripts/projects/common/modules/commercial/dfp/render-advert.js
```diff
@@ -0,0 +1,161 @@
...
+        };
+    }
+
+    /**
+     * @param adSlotId - DOM ID of the rendered slot
+     * @param slotRenderEvent - GPT slotRenderEndedEvent
+     * @returns {Promise} - resolves once all necessary rendering is queued up
+     */
+    function renderAdvert(adSlotId, slotRenderEvent) {
+        var size;
```
regiskuckaertz: You can move this var into the promise handler.

static/src/javascripts/projects/common/modules/commercial/dfp/render-advert.js
jbreckmckye: Into a separate `removePlaceholderContent`? I think it's a good idea.

regiskuckaertz:
Hurray for some clarity ✌️  Come on you have time for another PR 👯 

👍

paperboyo:
:cry:

PR #13326, opened 2016-06-17
Inject admin router

alexduf:
## What does this change?

This refactors the application loader of the admin project to extract the shared bits between projects.
This PR also uses an injected router as it is the new default in play 2.5. This means all controllers aren't objects anymore but classes. I grouped repetitives changes in the same commit, so please read that PR commit per commit as I believe it will make more sense and be more readable.

Good old play 2.5 at some point
## Request for comment

@TBonnin @johnduffell 

Apologies for a bigger than usual PR, but I couldn't make it smaller. (plus it's very repetitive)


piuccio:
this time let's make sure we don't break scheduled front press 😉

common/app/app/FrontendApplicationLoader.scala
```diff
@@ -0,0 +1,65 @@
...
+  lazy val jobScheduler = new JobScheduler(environment)
+  lazy val akkaAsync = new AkkaAsync(environment, actorSystem)
+
+  // this is a workaround to make wsapi and the actorsystem available to the injector.
+  // I'm forced to do that as we still use Ws.url and Akka.system(app) *everywhere*, and both directly get the reference from the injector
+  override lazy val injector: Injector = new SimpleInjector(NewInstanceInjector) + router + crypto + httpConfiguration + wsApi + actorSystem
+
+  // here are the attributes you must provide for your app to start
+  def appIdentity: ApplicationIdentity
+  def appMetrics: ApplicationMetrics
```
TBonnin: Is appMetrics not different from one app to the other? If not why requiring every app to provide it?

TBonnin:
You might want to get more eyes 👀  on it but it look good to me? 👍

common/app/app/FrontendApplicationLoader.scala
alexduf: Some modules have more metrics than the "standard" one. But you're right, maybe I should set it to default and only override when necessary.

TBonnin: Got it. Up to you

johnduffell:
have a go, see how it goes! 👍 can't see anything obvious but I'm getting experienced at quick rollbacks....

alexduf:
@johnduffell ahah me too, more experienced than I'm willing to admit :)

PR #13329, opened 2016-06-17
Add sharing comments functionality

marialivia16:
## What does this change?

Adds the sharing functionality for comments.
TODO before merging:
- [x] Tweak animation
- [x] Add fb permalink's domain to App Domains field in app settings

No.
## Screenshots

Gif of showing buttons animation:
comment

Comment shared on Facebook:

Comment shared on Twitter:

## Request for comment

@guardian/participation - I created this PR so we can discuss about it, there are still changes that need to be done if we decide to put this in production.

### DIFF
--- a/common/app/conf/switches/FeatureSwitches.scala
+++ b/common/app/conf/switches/FeatureSwitches.scala
@@ -474,4 +474,15 @@ trait FeatureSwitches {
     exposeClientSide = true
   )
 
+  // Owner: Maria Livia Chiorean
+  val SharingComments = Switch(
+    SwitchGroup.Feature,
+    "sharing-comments",
+    "When ON, the user will be able to share comments",
+    owners = Seq(Owner.withGithub("marialivia16")),
+    safeState = Off,
+    sellByDate = never,
+    exposeClientSide = true
+  )
+
 }
--- a/discussion/app/views/fragments/commentShare.scala.html
+++ b/discussion/app/views/fragments/commentShare.scala.html
@@ -1,34 +1,37 @@
 @(comment: Comment)
 @import org.jsoup.Jsoup
 @import org.jsoup.safety.Whitelist
+@import conf.switches.Switches.SharingComments
 
-<div class="sharing-component d-comment__action">
+@if(SharingComments.isSwitchedOn) {
+    <div class="sharing-component d-comment__action">
 
-    <div class="sharing-text">@fragments.inlineSvg("share", "icon") Share</div>
+        <div class="sharing-text">@fragments.inlineSvg("share", "icon") Share</div>
 
-    <div class="block-share block-share--gallery sharing-buttons" data-link-name="block share">
-    @defining(s"https://gu.com${comment.discussion.key}#comment-${comment.id}") { permalink =>
-        @defining(Jsoup.clean(comment.body, Whitelist.none())) { commentBody =>
-            <a class="social__action social-icon-wrapper"
-            href="@ShareLinks.createShareLinkForComment(platform = Facebook, href = permalink, text = "", quote = Some(s"${comment.profile.displayName} commented: '$commentBody'")).href"
-            target="_blank"
-            data-link-name="social facebook">
-                <span class='inline-icon__fallback button share-modal__item share-modal__item--facebook'>Share on Facebook</span>
-                @fragments.inlineSvg("share-facebook", "icon", List("rounded-icon social-icon social-icon--facebook"))
-                <span class="u-h">Facebook</span>
-            </a>
+        <div class="block-share block-share--gallery sharing-buttons" data-link-name="block share">
+        @defining(s"https://gu.com${comment.discussion.key}#comment-${comment.id}") { permalink =>
+            @defining(Jsoup.clean(comment.body, Whitelist.none())) { commentBody =>
+                <a class="social__action social-icon-wrapper"
+                href="@ShareLinks.createShareLinkForComment(platform = Facebook, href = permalink, text = "", quote = Some(s"${comment.profile.displayName} commented: '$commentBody'")).href"
+                target="_blank"
+                data-link-name="social-comment facebook">
+                    <span class='inline-icon__fallback button share-modal__item share-modal__item--facebook'>Share on Facebook</span>
+                    @fragments.inlineSvg("share-facebook", "icon", List("rounded-icon social-icon social-icon--facebook"))
+                    <span class="u-h">Facebook</span>
+                </a>
 
-            <a class="social__action social-icon-wrapper"
-            href="@ShareLinks.createShareLinkForComment(platform = Twitter, href = comment.webUrl, text = s"'${commentBody.take(111)}...'").href"
-            target="_blank"
-            data-link-name="social twitter">
-                <span class='inline-icon__fallback button share-modal__item share-modal__item--twitter'>Share on Facebook</span>
-                @fragments.inlineSvg("share-twitter", "icon", List("rounded-icon social-icon social-icon--twitter"))
-                <span class="u-h">Twitter</span>
-            </a>
+                <a class="social__action social-icon-wrapper"
+                href="@ShareLinks.createShareLinkForComment(platform = Twitter, href = comment.webUrl, text = s"'${commentBody.take(111)}...'").href"
+                target="_blank"
+                data-link-name="social-comment twitter">
+                    <span class='inline-icon__fallback button share-modal__item share-modal__item--twitter'>Share on Facebook</span>
+                    @fragments.inlineSvg("share-twitter", "icon", List("rounded-icon social-icon social-icon--twitter"))
+                    <span class="u-h">Twitter</span>
+                </a>
+            }
         }
-    }
+        </div>
     </div>
+}
 
-</div>
 
--- a/discussion/app/views/fragments/commentShare.scala.html
+++ b/discussion/app/views/fragments/commentShare.scala.html
@@ -1,0 +1,25 @@
+@(comment: Comment)
+@import org.jsoup.Jsoup
+@import org.jsoup.safety.Whitelist
+
+<div class="block-share block-share--gallery inactive-share-buttons" data-link-name="block share">
+    @defining(s"https://gu.com${comment.discussion.key}#comment-${comment.id}") { permalink =>
+        @defining(Jsoup.clean(comment.body, Whitelist.none())) { commentBody =>
+            <a class="rounded-icon block-share__item block-share__item--facebook js-blockshare-link"
+            href="@ShareLinks.createShareLinkForComment(platform = Facebook, href = permalink, text = "", quote = Some(s"${comment.profile.displayName} commented: '$commentBody'")).href"
+            target="_blank"
+            data-link-name="social facebook">
+                @fragments.inlineSvg("share-facebook", "icon")
+                <span class="u-h">Facebook</span>
+            </a>
+
+            <a class="rounded-icon block-share__item block-share__item--twitter js-blockshare-link"
+            href="@ShareLinks.createShareLinkForComment(platform = Twitter, href = comment.webUrl, text = s"'${commentBody.take(111)}...'").href"
+            target="_blank"
+            data-link-name="social twitter">
+                @fragments.inlineSvg("share-twitter", "icon")
+                <span class="u-h">Twitter</span>
+            </a>
+        }
+    }
+</div>
--- a/discussion/app/views/fragments/commentShare.scala.html
+++ b/discussion/app/views/fragments/commentShare.scala.html
@@ -2,24 +2,33 @@
 @import org.jsoup.Jsoup
 @import org.jsoup.safety.Whitelist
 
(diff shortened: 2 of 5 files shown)


gtrufitt:
This is great, I really like the animation you prototyped there so you should definitely do that! :)

I am not sure whether you need 'Gelion commented:' when sharing a quote on Facebook? At least not in the quote, it feels like the quote should literally be the text you're sharing. Maybe the status text could be something related to it?

Again, I don't know if you need an author on the Twitter message, others may disagree, but when I share a quote on Twitter from an article I generally just add the quote in quotation marks with the link - I don't think the actual commenter name adds that much value to the tweet.

Are you limiting this to Facebook & Twitter? You could also add email and pass the article title as the subject and the quote as the body? Again, opinions may vary :)

A couple of other notes before moving to production:

- [ ] The share icon and text should have a hover state that includes `cursor: pointer`
- [ ] The twitter share text should use double quotes `"` instead of single to be consistent with existing sharing (text highlight)
- [ ] The icons should be visible if JS is off and then hidden when JS is on so that they can be used with no js

static/src/stylesheets/module/_discussion.scss
```diff
@@ -840,13 +840,27 @@ $comment-recommend-button-size: 19px;
...
     position: absolute;
     bottom: $gs-baseline;
+    opacity: .7;
     .discussion--closed & {
-        display: none;
+        .d-comment__action--reply, .d-comment__action-separator, .d-comment__action-pick {
+            display: none;
+        }
+        .sharing-component {
+            margin: 0 !important;
```
gtrufitt: We shouldn't be using `!important` - what's the reason? May need a refactor

static/src/stylesheets/module/_discussion.scss
```diff
@@ -840,13 +840,27 @@ $comment-recommend-button-size: 19px;
...
+
+            .sharing-text .inline-icon {
+                padding-left: 0 !important;
```
gtrufitt: This is also pretty heavily nested, we should try and keep CSS flat, so styling on a new class is preferable

static/src/stylesheets/module/_discussion.scss
```diff
@@ -903,6 +917,54 @@ $comment-recommend-button-size: 19px;
...
+    margin-left: 5px;
+
+    .sharing-buttons {
+        opacity: 0;
+        cursor: pointer;
+        transition: opacity 0s ease-in-out;
+        display: inline-block;
+
+        .inline-icon {
+            height: 0rem;
```
gtrufitt: `0` values don't need units.

static/src/stylesheets/module/_discussion.scss
```diff
@@ -903,6 +917,54 @@ $comment-recommend-button-size: 19px;
...
+    .sharing-text, .inline-icon {
+        cursor: default;
+        display: inline-block;
+        float: left;
+    }
+
+    .sharing-text {
+        .inline-icon {
+            padding: 1px 5px;
+            fill: #005689;
```
gtrufitt: This should come from a colour variable

static/src/stylesheets/module/_discussion.scss
```diff
@@ -903,6 +917,54 @@ $comment-recommend-button-size: 19px;
...
+        }
+    }
+
+    a:first-child {
+        margin-left: 5px;
+    }
+}
+
+.sharing-component:hover {
```
gtrufitt: It's best to try and keep the 'state' styles with your component - so you can nest `&:hover` inside your sharing component block above

static/src/stylesheets/module/_discussion.scss
```diff
@@ -840,13 +840,27 @@ $comment-recommend-button-size: 19px;
 .d-comment__actions {
     position: absolute;
     bottom: $gs-baseline;
+    opacity: .7;
     .discussion--closed & {
-        display: none;
+        .d-comment__action--reply, .d-comment__action-separator, .d-comment__action-pick {
```
gtrufitt: Each selector should be on a new line

discussion/app/views/fragments/commentShare.scala.html
```diff
@@ -0,0 +1,34 @@
...
+            data-link-name="social facebook">
+                <span class='inline-icon__fallback button share-modal__item share-modal__item--facebook'>Share on Facebook</span>
+                @fragments.inlineSvg("share-facebook", "icon", List("rounded-icon social-icon social-icon--facebook"))
+                <span class="u-h">Facebook</span>
+            </a>
+
+            <a class="social__action social-icon-wrapper"
+            href="@ShareLinks.createShareLinkForComment(platform = Twitter, href = comment.webUrl, text = s"'${commentBody.take(111)}...'").href"
+            target="_blank"
+            data-link-name="social twitter">
```
gtrufitt: We should try to associate this `data-link-name` to comment social: eg:  `social-comment : twitter`

static/src/stylesheets/module/_discussion.scss
NataliaLKB: I agree with @gtrufitt, if you find yourself using `!important` it is likely some code should be refactored. Ideally we shouldn't be "overriding" any styles.

static/src/stylesheets/module/_discussion.scss
NataliaLKB: In this case, all inline icons can be given other classes so maybe give these icons a specific class and then remove the nesting?

gtrufitt:
Apologies there are quite a few comments, feel free to give me a shout if you want to sit down and discuss this in person :)

static/src/stylesheets/module/_discussion.scss
```diff
@@ -903,6 +917,54 @@ $comment-recommend-button-size: 19px;
...
+    a:first-child {
+        margin-left: 5px;
+    }
+}
+
+.sharing-component:hover {
+    .sharing-buttons {
+        opacity: 1;
+
+        .inline-icon {
```
NataliaLKB: Same as above, probably doesn't need to be nested

discussion/app/views/fragments/commentShare.scala.html
```diff
@@ -0,0 +1,34 @@
+@(comment: Comment)
+@import org.jsoup.Jsoup
+@import org.jsoup.safety.Whitelist
+
+<div class="sharing-component d-comment__action">
+
+    <div class="sharing-text">@fragments.inlineSvg("share", "icon") Share</div>
+
+    <div class="block-share block-share--gallery sharing-buttons" data-link-name="block share">
```
NataliaLKB: Not sure, but should this data-link-name be something line `comment social`?

marialivia16:
@gtrufitt These are reasonable comments, I'll find some time to take a look this afternoon and let you know if I need any help.

gtrufitt:
@marialivia16 Aces :)

nicl:
@marialivia16 looks great!

Realise it's more work, but can we add a feature switch for this? 

To try and make this happen, I'll organise a catchup with editorlal folks as I know there are a few concerns, especially around having the actual comment body shown in the shared tweet/fb post.

marialivia16:
@nicl I can easily add a switch, no problem. I'm currently going through the comments and tidying up my CSS. A catch up would be great, thanks!

static/src/stylesheets/module/_discussion.scss
```diff
@@ -903,6 +915,55 @@ $comment-recommend-button-size: 19px;
...
+    cursor: pointer;
+    display: inline-block;
+}
+
+.comment-facebook-icon,
+.comment-twitter-icon {
+    height: 0;
+    min-width: 0;
+    margin: .6rem;
+    transition: all .5s ease-in-out;
```
marialivia16: This is only working on hover in, but not on hover out. Can anyone help me please?

gtrufitt: Do you need to use `width` instead of `min-width` so it resets?

marialivia16: If I'm using `width` the button doesn't expand from a dot in the middle but from a horizontal line. And it still doesn't work on hover out.

johnduffell:
this looks great, I like it!  surely needs to be an a/b test though, as otherwise we'll not know if it was a good idea.  If people hardly use it it's difficult to know if the extra code was worth it?

jamesgorrie:
> surely needs to be an a/b test though

As it's just adding functionality - wouldn't it just be an A test? If we just add tracking and make sure to check it, we should be good.

Thinking about it - is this what you're suggesting, add tracking, add a switch that expires that reminds us to do the analysis?

johnduffell:
@jamesgorrie ideally before adding complexity especially user facing, we should know what we expect to change.  The only way to measure the change would be to leave the feature off for some users.  If you add tracking and find that it gets used 2000 times over the week, will that be enough to justify the feature?  What about if the "normal" share count goes down by 2000 i.e. overall shares are the same?

I'm not saying I think it's a bad idea, actually personally I like it, I just want to be as rigorous adding features as we'd be taking them away.  In fact maybe even more rigorous, as complexity is our enemy!

ScottPainterGNM:
It might be worth querying this with the choccy...I mean MVT team to see what they think might make a worthwhile test around this.

ScottPainterGNM:
That being said I think it's great!

static/src/stylesheets/module/_discussion.scss
gtrufitt: Hey @marialivia16 - did you work this out? I can take another look if not :)

discussion/app/views/fragments/commentShare.scala.html
```diff
@@ -0,0 +1,38 @@
...
+    <div class="sharing-text">@fragments.inlineSvg("share", "icon", List("comment-share-icon")) Share</div>
+
+    <div class="sharing-buttons" data-link-name="comment social">
+    @defining(s"https://gu.com${comment.discussion.key}#comment-${comment.id}") { permalink =>
+        @defining(StringEscapeUtils.unescapeHtml(Jsoup.clean(comment.body, Whitelist.none()))) { commentBody =>
+            <a href="@ShareLinks.createShareLinkForComment(platform = Facebook, href = permalink, text = "", quote = Some(s"""${comment.profile.displayName} commented: "$commentBody"""")).href"
+            target="_blank"
+            class="social__action social-icon-wrapper"
+            data-link-name="social-comment : facebook">
+                <span class="inline-icon__fallback button share-modal__item share-modal__item--facebook">
```
gtrufitt: The `share-modal` classes suggest to me that we're re-using styles here that are designed for modals, can you pull them out and make them more generic?

static/src/stylesheets/module/_discussion.scss
marialivia16: @gtrufitt Hey! Still not working. Do you think we can look at this together sometime today?

gtrufitt: @marialivia16 Definitely - give me a shout when you're free! I pulled out the CSS to make a minimum example so I could get my head around it - is this working as you expect? http://codepen.io/anon/pen/jAwyLG

marialivia16: @gtrufitt Yes, I'll come over after lunch. Thanks!

marialivia16:
@nicl I'm currently not passing any campaign codes because the permalink doesn't allow parameters.

common/app/conf/switches/FeatureSwitches.scala
```diff
@@ -463,4 +463,15 @@ trait FeatureSwitches {
...
   )
 
+  // Owner: Maria Livia Chiorean
+  val SharingComments = Switch(
+    SwitchGroup.Feature,
+    "sharing-comments",
+    "When ON, the user will be able to share comments",
+    owners = Seq(Owner.withGithub("marialivia16")),
+    safeState = Off,
+    sellByDate = never,
```
NataliaLKB: This discussion may have already happened, but how will we determine this is a good thing? How can we test this? Should we maybe put in a `sellByDate` as a reminder to review this feature?

sndrs:
what's the status of this? can we close it if it's not going to ship? (can always reopen again later when ready)

marialivia16:
@gtrufitt @sndrs @NataliaLKB @stephanfowler 

This PR finally got approval to be merged and tested on CODE.

Anyone wants to have a final look and give a thumbs up?

common/app/conf/switches/FeatureSwitches.scala
marialivia16: @nicl @crifmulholland Any thoughts on how we can test this? One of the things we will be looking at is number of shares. We can track the number of clicks using the data-link-names.

crifmulholland:
👍  

This doesn't need to be an ab test, but should be a switch so that we review after a month. 

We'll look at number of clicks on the links and also traffic direct to comments referred to from facebook and twitter and determine whether this has made a significant impact.

User feedback on comments is important for us, and so it is also going to help our moderators see what's 'hot' and not.

NataliaLKB:
@marialivia16 lgtm 👍 

@crifmulholland thanks for the explanation!

PR #13332, opened 2016-06-17
Measure each time notifications are granted or denied

janua:
## What does this change?

This adds tracking to the liveblog notification signup flow; we now track if the user `grants` or `denies` permission to the browser notifications.

A big blocker in this flow is the browser step of asking the user permission to show notifications from our domain; it will be extremely useful to know if the drop-off happens here and if it is causing any issues.

No
## Request for comment

@gtrufitt @NathanielBennett @crifmulholland

### DIFF
--- a/static/src/javascripts/bootstraps/enhanced/notifications.js
+++ b/static/src/javascripts/bootstraps/enhanced/notifications.js
@@ -16,7 +16,8 @@ define([
     'text!common/views/ui/notifications-permission-denied-message.html',
     'lodash/collections/some',
     'lodash/arrays/uniq',
-    'lodash/arrays/without'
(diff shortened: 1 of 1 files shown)


gtrufitt:
This looks fine to me from a general code-style and a high-level understanding of what's going on but @NathanielBennett should probably review properly! 👍

janua:
@gtrufitt I do find it strange that when `denied` happens it comes back as a failed promise rather than a completed promise; which means we have to use `catch`.

gtrufitt:
Indeed - I was confused why you would have to check the `Notifications` object directly - it would seem that the param in the `catch` function should return the information about the permission directly

PR #13335, opened 2016-06-20
Increase ad load timeout to 10s

regiskuckaertz:
I introduced a new API to wait until an ad has rendered/loaded last friday, with a timeout of 500s... I thought that was more than enough but lo and behold: no it wasn't.

Pushing it to 10s 😭 

@ScottPainterGNM


ScottPainterGNM:
👍

PR #13336, opened 2016-06-20
Svg classes

NataliaLKB:
@sndrs is experimenting with a system for making the css in the head of the page smaller. We want to start experimenting with this on new features. Since I am working on the new header, and it doesn't inherit styling from anything else, it seems like a good place to start.

While looking at the code on Thursday we discovered one blocker. None of our inline svgs have classes. Since this new system doesn't support nesting, our inline svg helpers need to be updated.

This Pr updates the Scala one. Each class that the span is given the svg is also given but with `__svg` on the end of it. For example, if the span is given the classnames of `inline-icon inline-comment` than the svg is given the classnames `inline-icon__svg inline-comment__svg`. 

image
## Future Pr
- [ ] Update JavaScript Inline Svg helper

This does affect Amp, and anything else that uses this helper, but it is just adding classes so the pages still validate.


common/app/views/fragments/inlineSvg.scala.html
```diff
@@ -1,4 +1,12 @@
 @(file: String, path: String = "", classes: Seq[String] = List(), title: Option[String] = None)
+
+@import views.support.setSvgClasses
+
 <span @title.map{ t => title="@t" } class="inline-@file @if(path){inline-@path} @classes.mkString(" ")">
-    @Html(common.Assets.inlineSvg(path + "/" + file + ".svg"))
+    @defining((
+        path + "/" + file + ".svg",
+        classes ++ Seq("inline-" + file, "inline-" + path)
```
sndrs: does this mean both the svgs and the wrapper will get `.inline-@file` and `.inline-path`?

NataliaLKB: They will get `inline-@file__svga and `inline-@path__svg`: https://github.com/guardian/frontend/pull/13336/files/75a1c3ba7dbd31553bd583ee9402797e80f5f91e#diff-a78830a0e574fa7224f8e492f8ac2c9aR601

sndrs: ah `_.concat("__svg")`, sorry, missed it. great!

sndrs:
i'm pretty sure the reason that wrapper is there is because i didn't have the bright idea of using the html cleaner on the svgs, really nice. we should definitely look into removing it if possible

👍

NataliaLKB:
I have looked into refactoring the JS helper and I think it will be harder. Hopefully I will have time to do that soon!

dominickendrick:
👍

PR #13337, opened 2016-06-20
Add a quick survey regarding visit frequency

katebee:
## What does this change?

Show a one question survey to 'new' visitors to The Guardian, to see how frequently they come to the site. The survey is targeted at people who appear to be 'first-time' visitors according to their alreadyVisited count.

The survey inserts into the first article opened, above the first paragraph of the article.

The survey will only be seen once by each visitor; a cookie is set once it has appeared, regardless of interaction.

Investigation into how many visitors to the site are genuine 'first-time' visitors. The ideal outcome will be an understanding of what percentage visitors

No
##### Survey inserted after the first paragraph:

quick survey

##### Fade transition reveals thank you message after interaction:

quick survey thank you

## Request for comment

@dominickendrick


static/src/javascripts/projects/common/modules/experiments/tests/visitor-frequency-quick-survey.js
```diff
@@ -0,0 +1,108 @@
...
+            bean.on(surveyQuestion, 'click', function () {
+                fastdom.write(function () {
+                    disableRadioButtons('fi-survey__button');
+                    surveyFadeOut();
+                    thankyouFadeIn();
+                });
+            });
+        }
+
+        function checkCookies(cookieName) {
```
dominickendrick: You should explicitly `return false` when they fail the check

static/src/javascripts/projects/common/modules/experiments/tests/visitor-frequency-quick-survey.js
```diff
@@ -0,0 +1,108 @@
...
+            var article = document.getElementsByClassName('content__article-body')[0];
+            var firstChild = article.firstChild;
+            var survey = document.createElement('div');
+            survey.innerHTML = quickSurvey;
+            article.insertBefore(survey, firstChild);
+        }
+
+        function disableRadioButtons(buttonClassName) {
+            var radioButtons = document.getElementsByClassName(buttonClassName);
+            for (var i = 0; i < radioButtons.length; i++) {
```
dominickendrick: you could use a `forEach()` here

static/src/javascripts/projects/common/modules/experiments/tests/visitor-frequency-quick-survey.js
```diff
@@ -0,0 +1,108 @@
...
+                    radioButtons[i].disabled=true;
+                }
+        }
+
+        function surveyFadeOut() {
+            var surveyTextbox = document.getElementById('surveyTextbox');
+            fastdom.write(function () {
+                surveyTextbox.style.cssText += 'visibility:hidden;opacity:0;transition:visibility 0s ease-in 0.5s,opacity 0.5s linear';
```
dominickendrick: This css could this be moved to `quick-survey.html`, and you could just toggle a class on the element with js when you need to apply the style, alternatively you could also use the `.class-name[selected]` and `not:(.class-name[selected])` attribute to apply the styles once a checkbox has been selected to do this without js.

static/src/javascripts/projects/common/modules/experiments/tests/visitor-frequency-quick-survey.js
```diff
@@ -0,0 +1,108 @@
...
+            });
+        }
+
+        function thankyouFadeIn() {
+            var surveyThanks = document.getElementsByClassName('impressions-survey__thanks');
+            fastdom.write(function () {
+                surveyThanks[0].style.cssText += 'visibility:visible;opacity:1;transition:visibility 0s ease-in 0.5s,opacity 0.8s linear';
```
dominickendrick: same as above

common/app/conf/switches/ABTestSwitches.scala
```diff
@@ -74,6 +74,16 @@ trait ABTestSwitches {
     exposeClientSide = true
   )
 
+  val ABVisitorFrequencyQuickSurvey = Switch(
+    SwitchGroup.ABTests,
+    "ab-visitor-frequency-quick-survey",
+    "Add a single question survey to the submeta section of article pages",
+    owners = Seq(Owner.withGithub("katebee")),
+    safeState = Off,
+    sellByDate = new LocalDate(2016, 6, 18),
```
gtrufitt: This is the Saturday just gone :)

static/src/javascripts/projects/common/modules/experiments/tests/visitor-frequency-quick-survey.js
```diff
@@ -0,0 +1,108 @@
...
+        this.description = 'Add a single question survey to the submeta section of article pages';
+        this.audience = 0;
+        this.audienceOffset = 0;
+        this.successMeasure = 'Obtain a data-set on how often users visit the Guardian';
+        this.audienceCriteria = 'All users';
+        this.dataLinkNames = 'impressions frequency survey';
+        this.idealOutcome = '';
+
+        this.canRun = function () {
+            return config.page.contentType == 'Article' && checkCookies('GU_FI');
```
gtrufitt: Minor but should use `===`

dominickendrick: You could make the cookie name 'GU_FI' a property ie `this.cookieName = 'GU_FI'` then you wouldn't need to pass it as part of `checkCookies` and can use it when you set the cookie here  -> `cookies.add('GU_FI', 1, 5)`

static/src/javascripts/projects/common/modules/experiments/tests/visitor-frequency-quick-survey.js
dominickendrick: Could be named better ? `isFirstTimeUser` ?

static/src/javascripts/projects/common/views/experiments/quick-survey.html
```diff
@@ -0,0 +1,92 @@
...
+
+
+<div class="impressions-survey__body" data-link-name="impressions-frequency-survey">
+
+    <div id="surveyTextbox">
+        <h3 class="impressions-survey__question">
+            How often do you read the Guardian in a digital format?</h3>
+        <div id="impressions-survey__select">
+            <label class="fi-survey__label"><input type="radio" class="fi-survey__button" data-link-name="frequency_5" value="frequency_5">
+                Every day/most days</label><br>
```
dominickendrick: I would not use a `<br>` for formatting the labels. You could just set the display type on `fi-survey__label` to be `block` ie `.fi-survey__label { display:block }`

static/src/javascripts/projects/common/modules/experiments/tests/visitor-frequency-quick-survey.js
```diff
@@ -0,0 +1,108 @@
...
+        };
+
+        function renderQuickSurvey() {
+            // this cannot be wrapped in a fastdom function; the disable function breaks
```
gtrufitt: The reason the disableRadioButtons breaks is because you will be calling it before the HTML is inserted. I think instead of adding fastdom inside your `handleSurveyResponse` which calls `disableRadioButtons` you can move the fastdom to wrap all the function calls in test *or* even better return a fastdom promise from renderQuickSurvey and add the `disableRadioButtons` call inside the `.then()`

static/src/javascripts/projects/common/views/experiments/quick-survey.html
```diff
@@ -0,0 +1,92 @@
...
+    position:absolute;
+    top:20%;opacity:0;
+    visibility:collapse;
+}
+
+</style>
+
+<div class="impressions-survey__heading">
+    <div class="impressions-survey__title">Survey</div>
+    <div class="impressions-survey__title"><a href="https://www.theguardian.com/help/terms-of-service" target="_blank" class="impressions-survey__about">About</a></div>
```
dominickendrick: Can we indent the `<a>` so the line length isn't too long

static/src/javascripts/projects/common/modules/experiments/tests/visitor-frequency-quick-survey.js
```diff
@@ -0,0 +1,108 @@
...
+
+        function surveyFadeOut() {
+            var surveyTextbox = document.getElementById('surveyTextbox');
+            fastdom.write(function () {
+                surveyTextbox.style.cssText += 'visibility:hidden;opacity:0;transition:visibility 0s ease-in 0.5s,opacity 0.5s linear';
+            });
+        }
+
+        function thankyouFadeIn() {
+            var surveyThanks = document.getElementsByClassName('impressions-survey__thanks');
```
gtrufitt: Try to use `js-` prefixed classes for getting elements as it separates style from behavioural classes.

static/src/javascripts/projects/common/views/experiments/quick-survey.html
```diff
@@ -0,0 +1,92 @@
+<style>
```
gtrufitt: This should be in a CSS file I think - unless someone has mentioned to do it this way? If it is in scss we can make use of the variables for colours, margins, fonts etc - I guess it is embedded in the html as a way of not delivering it to the user?

static/src/javascripts/projects/common/views/experiments/quick-survey.html
```diff
@@ -0,0 +1,92 @@
...
+.fi-survey__button, .fi-survey__label {
+    margin: 0 0.2em 0 0.2em;
+}
+
+.impressions-survey__thanks {
+    color:#fff;
+    font-size:2em;
+    font-weight:bold;
+    position:absolute;
+    top:20%;opacity:0;
```
gtrufitt: Should be 1 declaration per line

static/src/javascripts/projects/common/views/experiments/quick-survey.html
```diff
@@ -0,0 +1,92 @@
...
+                Fortnightly</label><br>
+            <label class="fi-survey__label"><input type="radio" class="fi-survey__button" data-link-name="frequency_2" value="frequency_2">
+                Monthly or less</label><br>
+            <label class="fi-survey__label"><input type="radio" class="fi-survey__button" data-link-name="frequency_1" value="frequency_1">
+                This is my first visit</label>
+        </div>
+    </div>
+
+    <div class="impressions-survey__thanks">
+        Thank you for helping us<br>improve The Guardian
```
gtrufitt: It's a lower-case the - the Guardian :)

static/src/javascripts/projects/common/views/experiments/quick-survey.html
dominickendrick: By inlining it, we can avoid having to load an extra css file just for this fragment. Is there another way we could do this ?

gtrufitt: These [JS templates get bundled in with the JS anyway](https://assets.guim.co.uk/javascripts/bootstraps/enhanced/181c49fefa2c68adf4c9a2717551e64e/main.js) (search for `<iframe` for example), so you're not saving anything by putting the CSS in the template - you might as well have a seperate CSS file and make use of SCSS, variables, linting etc

dominickendrick: Ah ok !

gtrufitt: Granted, this maybe shouldn't be the case for *all* templates!

katebee: Cheers! That is good to know :smile:

gtrufitt: No problem! :) For an example - there is an experiments folder for stylesheets (same as JS) here - https://github.com/guardian/frontend/tree/master/static/src/stylesheets/module/experiments - if you put your styles in there, you can then just import them into [common](https://github.com/guardian/frontend/blob/master/static/src/stylesheets/_common.scss)

static/src/javascripts/projects/common/modules/experiments/tests/visitor-frequency-quick-survey.js
```diff
@@ -0,0 +1,112 @@
...
+                    thankyouFadeIn();
+                });
+            });
+        }
+
+        function isFirstTimeUser(cookieName) {
+            var surveyCookie = cookies.get(cookieName);
+            if (!(surveyCookie) && storage.local.isStorageAvailable()) {
+                var alreadyVisited = storage.local.get('gu.alreadyVisited');
+                if (!alreadyVisited || alreadyVisited < 1) {
```
dominickendrick: as this already a boolean statement you could rewrite as 
```
function isFirstTimeUser(cookieName) {
    var surveyCookie = cookies.get(cookieName);
    if (!(surveyCookie) && storage.local.isStorageAvailable()) {
          var alreadyVisited = storage.local.get('gu.alreadyVisited');
          return !alreadyVisited || alreadyVisited < 1;
    } 
    return false;
}
```

static/src/stylesheets/module/experiments/_survey.scss
```diff
@@ -0,0 +1,74 @@
...
+    color: #ffffff;
+    font-size: 2em;
+    font-weight: bold;
+    position: absolute;
+    top: 20%;
+    opacity: 0;
+    visibility: collapse;
+}
+
+.js-impressions-survey__fadein {
```
dominickendrick: 👏

dominickendrick:
👍 Great stuff : )

PR #13341, opened 2016-06-20
Add a switch for the Apple smart banner

marialivia16:
## What does this change?

Adds a switch for using the Apple's [smart banner](https://developer.apple.com/library/ios/documentation/AppleApplications/Reference/SafariWebContent/PromotingAppswithAppBanners/PromotingAppswithAppBanners.html). If the switch is on and it can be showed (iOS device and Safari) then we do not show the custom made banner we have. This work is part of this [idea](https://docs.google.com/document/d/1-01OozcbnIVQY8yh37jTnbNbtkGjTjy5zULEQskxKqs/edit#).
## Request for comment

@stephanfowler


common/app/views/fragments/metaData.scala.html
```diff
@@ -98,6 +98,10 @@
 <meta name="theme-color" content="#005689">
 <meta name="msapplication-TileImage" content="@Static("images/favicons/windows_tile_144_b.png")" />
 
+@if(SmartAppBanner.isSwitchedOn) {
+    <meta name="apple-itunes-app" content="app-id=409128287, app-argument=@page.metadata.webUrl">
```
TBonnin: I think there is a configuration value for the app id.

TBonnin:
The branch is probably not up-to-date with master.
Other than that and a tiny comment, lgtm 👍

stephanfowler:
👍 LGTM

common/app/views/fragments/metaData.scala.html
marialivia16: Good call, I changed it and I'm going to merge this.

PR #13343, opened 2016-06-20
Send fine-grain logging to Log Stash for commercial feeds

JonNorman:
## What does this change?

Whenever each commercial feed is parsed/fetched, it is logged with a number of key fields such that this can be visualised in Kibana.

We don't currently want to set an alarm on each action performed on each feed as this would be ~20 alarms, all of which would be fairly low priority.


kelvin-chappell:
Looks good.
Good to get rid of random starting intervals as well.
👍

TBonnin:
I know it has been merged already. Sorry for the late review.
You kept both reporting mechanisms (cloudwatch and logstash): is the goal to remove one of them at some point? Thx

PR #13344, opened 2016-06-20
Add logstash lifecycle to run in dev-build

JonNorman:
This just adds the Log Stash lifecycle to `dev-build` setup such that logging can be tested locally (given correct credentials added to `frontend.properties`).

**cc** @TBonnin


TBonnin:
👍  are the logstash properties in the default dev properties file stored on S3?

PR #13345, opened 2016-06-20
For krux in AMP: checking if local storage exists before using it

NataliaLKB:
AMP pages will soon be available within Newsstand. In the test our ads aren't loading. The only information I know for sure:
- Ads load in Guardian AMP pages for newsstand for IOS
- On android other third party embeds, like twitter, seem to work
- Some of the calls to Krux and DFP are working, just not the final ones

While talking to @maxspencer and @ChrisOwen101 I learnt that even though it is a webview, localstorage is not supported, and might not be handled by this app. Therefore I am adding a check to the amp remote file.

Only affects AMP
## Request for comment

@rich-nguyen

### DIFF
--- a/facia/public/amp_remote.html
+++ b/facia/public/amp_remote.html
@@ -73,8 +73,23 @@
             return config;
         }
 
+        function isLocalStorageReallyAvailable () {
+            var testKey = 'local-storage-module-test';
+            var data = 'test';
+
+            try {
+                // to fully test, need to set item
+                // http://stackoverflow.com/questions/9077101/iphone-localstorage-quota-exceeded-err-issue#answer-12976988
+                localStorage.setItem(testKey, data);
+                localStorage.removeItem(testKey);
+                return true;
+            } catch (e) {
+                return false;
+            }
+        }
+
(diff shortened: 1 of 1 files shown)


facia/public/amp_remote.html
```diff
@@ -74,7 +74,7 @@
         }
 
         function attachKruxSegments(config) {
-            if (localStorage.getItem('kxsegs') !== null) {
+            if (localStorage && localStorage.getItem('kxsegs') !== null) {
```
rich-nguyen: [Here](https://github.com/guardian/frontend/pull/12983#discussion_r64717316) is a thread about local storage's wild behaviour.

rich-nguyen: Yes, I have a made 2-chain pr reference.

NataliaLKB:
@rich-nguyen if you get a chance, can you take another look? I have put in a more comprehensive check based on our `storage.js` file.

Even though there are potentially more things to fix to getting our ads working in web views, I think fixing things one at a time is best.

rich-nguyen:
👍

PR #13346, opened 2016-06-20
Dimensionless inline merch slots

regiskuckaertz:
## What does this change?

Gone with the distracting indent that appears when an inline merchandising ad should load but (1) either DFP returns no creative or (2) the JSON endpoint returns a 404.

Before:
inline4

After:
inline2

On mobile too, rendering the creative is now one step:
inline1

cc @guardian/commercial-dev @sndrs @stephanfowler @rich-nguyen


static/src/javascripts/projects/common/modules/commercial/dfp/apply-creative-template.js
```diff
@@ -75,10 +75,8 @@ define([
         var creativeConfig = fetchCreativeConfig();
 
         if (creativeConfig) {
-            return Promise.all([
```
jbreckmckye: When the consumer of the returned promise sees it resolve, what should they be able to presume about the state of the adSlot?

regiskuckaertz: That the creative has been rendered, which is why the promise has to resolve with whatever `renderCreative` returns, instead of an array. The iframe is not really part of the rendering, is it?

jbreckmckye: Fine, but then we beg the question: why exactly are we waiting for this operation to complete, what's going to potentially run next, and what side effects do we want to settle before then?

In the general case, for an async operation to only finish half of its work before resolving its promise seems a bitty and booby-trapped interface to me.

regiskuckaertz: This case is actually the most important one. Contrary to other DFP line items, our own components require additional processing (network, DOM construction, logic) before calling an ad rendered. Other modules rely on this to query the geometry of the creative, eg the spacefinder. That's why it needs to wait.

I'm just wondering, do we really need to hide the iframe? It contains a JSON payload and nothing else.

jbreckmckye: If the iframe doesn't need to be hidden, this operation is just wasting a frame, really. So we can happily leave it be.

regiskuckaertz: Alright then! Thanks for your advice.

static/src/javascripts/projects/common/modules/commercial/dfp/render-advert.js
```diff
@@ -128,14 +130,25 @@ define([
...
+                var size = slotRenderEvent.size.join(',');
+                if (sizeCallbacks[size]) {
+                    sizeCallbacks[size](slotRenderEvent, advert);
+                }
             }
 
-            return isRendered;
+            function addRenderedClass() {
+                return isRendered ? fastdom.write(function () {
+                    advert.node.classList.add('ad-slot--rendered');
```
jbreckmckye: Yes, let's start pushing our codebase towards native DOM. The more of it we have, the less inclined others will be to use Bonzo et al.

static/src/stylesheets/module/_adslot.scss
```diff
@@ -413,11 +413,12 @@
 }
 .ad-slot--im {
     float: left;
-    margin: 5px $gs-gutter $gs-baseline 0;
+    width: 0;
+    height: 0;
 
-    @include mq($until: mobileLandscape) {
```
jbreckmckye: We don't need these margins any more?

static/src/stylesheets/module/_adslot.scss
```diff
@@ -413,11 +413,12 @@
...
-        margin-bottom: $gs-baseline / 2;
-        margin-right: $gs-gutter / 2;
+    &.ad-slot--rendered {
```
jbreckmckye: If adverts don't even appear in IE8, could we just use the `:not` selector to hide un-rendered IM slots?

static/src/stylesheets/module/commercial/_brandbadge.scss
```diff
@@ -61,24 +61,6 @@
...
-        margin-bottom: $gs-baseline / 2;
-        margin-right: $gs-gutter / 2;
-    }
-}
-.ad-slot--fobadge.ad-slot--im {
-    min-width: gs-span(2) + $gs-gutter;
-
-    .brandbadge__inner {
-        padding: $gs-baseline / 2 0 $gs-baseline;
-        border-top: 1px solid $neutral-3;
```
jbreckmckye: We no longer need these border styles?

static/src/javascripts/projects/common/modules/commercial/dfp/render-advert.js
rich-nguyen: I'm happy to do one or the other, but not both, which I fear this approach is leading us to. I'll raise it in the eng meeting to gather the thoughts of the rest of the team. Hi Jimmy!

regiskuckaertz: Happy to talk about this. I agree in certain circumstances that code readability (via a single api) trumps performance, but reading the codebase I feel bonzo is just encouraging sloppy coding practices.

static/src/stylesheets/module/_adslot.scss
regiskuckaertz: Spot on! Thanks for the tip 👯 🍭

static/src/stylesheets/module/commercial/_brandbadge.scss
regiskuckaertz: These styles are already in `_adslot.scss`, that's why I removed them.

static/src/javascripts/projects/common/modules/commercial/dfp/render-advert.js
regiskuckaertz: This discussion is worth having but unrelated to this PR, so I reverted to using bonzo for DOM manipulations.

humaislam:
What's the latest on this PR?

regiskuckaertz:
@humaislam It's ready to go, I just need approval.

static/src/javascripts/projects/common/modules/commercial/dfp/private/render-advert.js
```diff
@@ -47,13 +49,14 @@ define([
...
+        if ($node.hasClass('ad-slot--right')) {
             var mobileAdSizes = advert.sizes.mobile;
             if (mobileAdSizes && mobileAdSizes.some(function (size) { return size[0] === 300 && size[1] === 251; })) {
-                stickyMpu(bonzo(advert.node));
+                stickyMpu($node);
             }
-        } else if (advert.node.classList.contains('ad-slot--facebook')) {
-            advert.node.classList.add('ad-slot--fluid');
+        } else if ($node.hasClass('ad-slot--facebook')) {
+            $node.addClass('ad-slot--fluid');
```
sndrs: does this need to be fast-dom'd? if it's called from fd, there's a lot of calc in there that doesn't need to be?

static/src/javascripts/projects/common/modules/commercial/dfp/private/render-advert.js
```diff
@@ -63,11 +66,11 @@ define([
      */
     sizeCallbacks[adSizes.outOfPage] = function (event, advert) {
         if (!event.slot.getOutOfPage()) {
-            advert.node.classList.add('u-h');
-            var parent = advert.node.parentNode;
+            bonzo(advert.node).addClass('u-h');
```
sndrs: fd?

sndrs:
👍 apart from comments

static/src/javascripts/projects/common/modules/commercial/dfp/private/render-advert.js
regiskuckaertz: Yep, doing that. Thanks!

PR #13347, opened 2016-06-20
Remove welcome header A/B test

marialivia16:
## What does this change?

Removes the welcome header A/B test. We could see no increase in returning visitors after running the test for 6 days and analysing returning rate for the next 7 days.


regiskuckaertz:
Conflicts [my fault](https://github.com/guardian/frontend/pull/13350), sorry 😅  Happy to see this go 👍

PR #13350, opened 2016-06-20
Shaving 140KB off boot.js

regiskuckaertz:
It might be a good idea to spread the tip that AB tests that require SVGs should directly call `inlineSVG` rather than import the whole `svgs` module, which is huge. I extracted the function from it, so that it is still usable without the overload.

Before:
picture 2

After:
picture 1

@sndrs


sndrs:
💥 👀

static/src/javascripts/projects/common/views/svg.js
```diff
@@ -0,0 +1,24 @@
+define(function () {
+    return svg;
+
+    function svg(markup, classes, title) {
+        // Only mess with classes if we actually need to.
+        if (classes) {
+            if (Array.isArray(classes)) {
```
piuccio: I've probably asked this many times. Do we include a polyfill for `Array.isArray` in IE8? can't remember

piuccio:
looks really good. The only minor comment is that you've mixed some unrelated changes (remove `$`, `reduce`).

If they're not risky or hard to test, go for it! 👍

static/src/javascripts/projects/common/views/svg.js
regiskuckaertz: This is part of `es5-shim` yes

regiskuckaertz:
It's kind of related, I should have mentioned that the SVG thing is just one aspect of this PR 😅 

Thanks bro!

PR #13352, opened 2016-06-21
Remove completed AB test

gtrufitt:
## What does this change?

Removes the stars vs footballs test.

cc/ @guardian/participation


NataliaLKB:
👍

PR #13353, opened 2016-06-21
Remove DFP timeout and speed up adverts init

regiskuckaertz:
Multiple modules are waiting for an ad to have rendered before performing an action. This is done via `dfp.trackAdRender` which returns a promise that resolves when the corresponding creative has been rendered. Obviously, this presumes the promise exists in the first place, which, given the asynchronous nature of our code, cannot be assured.

To mitigate this, I introduced [a routine](https://github.com/guardian/frontend/compare/rk-remove-dfp-timeout?expand=1#diff-844a43280e1947d8dd7cc7a480fd0551L476) that _waits_ for an ad slot to have been registered. I optimistically introduced an 500s timeout (I know right? 😆 ) that I then upgraded to a whopping 10s. But this even wasn't enough, in one particular case: if the GPT library takes too much time to load, all our init code is delayed as much. Indeed, [everything is wrapped in a `googletag.cmd.push` call](https://github.com/guardian/frontend/compare/rk-remove-dfp-timeout?expand=1#diff-844a43280e1947d8dd7cc7a480fd0551L1880).

With that in mind, the PR introduces two fundamental changes:
- the decision to set timeout or not is now entirely up to the `dfp.trackAdRender` calling site: it feels arbitrary to unilaterally set a 10s timeout anyway.
- the initialisation of the adverts data structure is done "synchronously": we no longer wait for the GPT library before searching the DOM. This seems reasonable because there is nothing GPT-related during this process. _It is worth pointing out that this removes the reliance on GPT being loaded, but is still asynchronous in nature since we run everything in promises chains._

cc @ScottPainterGNM @rich-nguyen

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
+++ b/static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
@@ -472,7 +472,7 @@ define([
     }
 
     var waitForAdvert = memoize(function (id) {
-        return new Promise(function (resolve, reject) {
+        return new Promise(function (resolve) {
             checkAdvert();
             function checkAdvert() {
                 var advert = getAdvertById(id);
--- a/static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
+++ b/static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
@@ -187,21 +187,12 @@ define([
     function loadAdvertising() {
         createAdverts();
         googletag.cmd.push(
-            defineAdverts,
+            queueAdverts,
             setPublisherProvidedId,
             shouldLazyLoad() ? displayLazyAds : displayAds,
             // anything we want to happen after displaying ads
             refreshOnResize
         );
-    }
-
-    /**
-     * Loop through each slot detected on the page and define it based on the data
-     * attributes on the element.
-     */
-    function defineAdverts() {
-        // queue ads for load
-        adverts.forEach(queueAdvert);
     }
 
     function setPublisherProvidedId() {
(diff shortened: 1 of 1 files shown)


static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
```diff
@@ -199,9 +200,6 @@ define([
      * attributes on the element.
      */
     function defineAdverts() {
```
rich-nguyen: some of this terminology, particularly the verb usage, seems conflated now. `defineAdverts` just runs `queueAdvert` for every advert. Ideally, it should be consistent, like `createAdverts` and `createAdvert`, and companion functions like this might be more readable if they are adjacent to each other?

static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
```diff
@@ -549,16 +558,6 @@ define([
     function initAdvert(advert) {
         advert.sizes = getAdBreakpointSizes(advert);
```
rich-nguyen: Could we just inline the remaining body of `initAdvert` into `queueAdvert` now?

rich-nguyen:
👍  great idea. The comments are mainly about reducing the number of similar-sounding `[verb]Advert` function names. Seems like a good chance to tidy these.

static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
regiskuckaertz: Yep, totally agree. Thanks!

PR #13355, opened 2016-06-21
Add custom dimensions to the Google Analytics pageview event

cb372:
## What does this change?

Adds the following custom dimensions to the GA pageview event. 
1. Ophan pageview ID
2. Ophan browser ID
3. Platform (hardcoded to "theguardian.com")
4. Section ID
5. Content type
6. Commissioning desk
7. Content ID
8. Author name(s)
9. Keyword tag IDs
10. Tone tag IDs
11. Series tag ID

Of course, some of these are only relevant to content pages. On other pages the relevant JS variable will be `undefined` and so the custom dimension will not be set.

If this works as expected, I'll make a separate PR to remove the 1% sampling and switch to the new 'proper' GA account.

More useful reporting in GA.

No. GA integration for AMP will need to be done separately. Apps will also implement separately, using the same custom dimensions as the site.
## Screenshots

screen shot 2016-06-21 at 10 46 45

Not sure what to do about the keyword tag IDs list being too long to fit into a parameter. Will need some thought, but this is good enough for now.
## Request for comment

@gklopper

### DIFF
--- a/common/app/views/fragments/analytics/google.scala.html
+++ b/common/app/views/fragments/analytics/google.scala.html
@@ -5,6 +5,7 @@
 
 @if(GoogleAnalyticsSwitch.isSwitchedOn) {
     <script id='google_analytics'>
+
         @***************************************************************************************
         * Standard copy 'n paste Google Analytics code                                         *
         ***************************************************************************************@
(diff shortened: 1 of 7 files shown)


cb372:
Aaah, the tests failed. That means I need to get them running on my machine...

gklopper:
👍 are you working with the data team to keep a sensible mapping of dimensions to their uses?

cb372:
Yep, this is done in conjunction with Matt K on the data team and Alexis who is designing the custom dimensions. There may be some tweaking later e.g. to send tag names instead of/as well as tag IDs.

common/app/views/fragments/analytics/google.scala.html
```diff
@@ -22,10 +23,29 @@
...
+        * Some of these will be undefined for non-content pages, but that's fine.              *
+        ***************************************************************************************@
+        ga('set', 'dimension4', guardian.config.page.section);
+        ga('set', 'dimension5', guardian.config.page.contentType);
+        ga('set', 'dimension6', guardian.config.page.commissioningDesk);
+        ga('set', 'dimension7', guardian.config.page.contentId);
+        ga('set', 'dimension8', guardian.config.page.author);
+        ga('set', 'dimension9', guardian.config.page.keywordIds);
+        ga('set', 'dimension10', guardian.config.page.toneIds);
+        ga('set', 'dimension11', guardian.config.page.seriesId);
```
TBonnin: It is probably a question for the data team but is there a reason to give dimensions a meaningful name like `dimensionSection` rather than `dimension4`?

static/src/javascripts/projects/common/utils/config.js
```diff
@@ -15,13 +15,6 @@ define([
         config.page.adUnit = ['/', config.page.dfpAccountId, '/', adUnitOverride].join('');
     }
 
-    // This is duplicated from
-    // https://github.com/guardian/ophan/blob/master/tracker-js/assets/coffee/ophan/transmit.coffee
-    // Please do not change this without talking to the Ophan project first.
-    config.ophan = {pageViewId: new Date().getTime().toString(36) + 'xxxxxxxxxxxx'.replace(/x/g, function () {
-        return Math.floor(Math.random() * 36).toString(36);
-    })};
-
```
TBonnin: I think `config.ophan.pageViewId` is accessed from a few places in the codebase.

common/app/views/fragments/analytics/google.scala.html
gklopper: I don't think you have a choice with the names (if I remember correctly from docs). You do however map them to more sensible names in the GA tool.

TBonnin: Got it

TBonnin:
:+1:

PR #13356, opened 2016-06-21
Update live blog notifications audience

janua:
## What does this change?

This increases the audience for `live-blog-chrome-notifications` to 50%.
## Request for comment

@NathanielBennett @crifmulholland @gtrufitt


NathanielBennett:
:+1: Ship IT!

johnduffell:
you will also need to rename the test, otherwise you will only get 50% of the people who haven't visited since you started the test.  the rest will be opted 'notintest' already

NathanielBennett:
@johnduffell Yeah good point

janua:
I am currently reassessing what sort of stats we would need to be statistically significant.

PR #13357, opened 2016-06-21
Allow the app to serve files from /node_modules in dev

sndrs:
## What does this change?

requests to `/assets` if `useHashedBundles` is false with also check in `/node_modules` 

lays ground for removal of bower components

just checking this is a good way to do it, not safety hole etc
## Request for comment

@johnduffell @rich-nguyen

### DIFF
--- a/common/app/dev/DevAssetsController.scala
+++ b/common/app/dev/DevAssetsController.scala
@@ -20,6 +20,7 @@ object DevAssetsController extends Controller with ExecutionContexts {
     case path if new File(s"static/src/$path").exists() => s"static/src/$path"
     case path if new File(s"static/public/$path").exists() => s"static/public/$path"
     case path if new File(s"static/target/$path").exists() => s"static/target/$path"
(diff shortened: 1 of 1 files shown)


johnduffell:
should only be run in dev, plus explicitly exposes things in node_modules (which presumably is just copies off the public repo) so we're doubly protected 👍

desbo:
I don't know about this, it means you can run anything you like in dev and there's nothing to say it'll make its way to prod. Shouldn't we avoid things that make the difference between dev and prod so big?

sndrs:
it's so we can move to getting all our modules from node_modules, rather than using bower to install to `/src/javascripts/bower_components` and then copy them to `/src/javascripts/components`. 

in prod, the bundles will also bundle from `node_modules`. this PR is only so that we can load directly in the browser from the dev require config, as we do currently from `/src/javascripts/components` i.e. its only where they come from that's changing

PR #13358, opened 2016-06-21
set success=true when the fetch succeeds

JonNorman:
## What does this change?

It sets `success=true` when the feed fetch succeeds, as opposed to currently when it is set to `false`.

Sigh.


kelvin-chappell:
I like this change! 👍

PR #13359, opened 2016-06-21
extend test for another month

NathanielBennett:
## What does this change?

Extends the test introduced here https://github.com/guardian/frontend/pull/12990, for another month

As before

N/A
## Request for comment

@gtrufitt @crifmulholland


gtrufitt:
This is going to fail tests because the 24th July is a Sunday.

NathanielBennett:
@gtrufitt I have so got caught by that before. Hows that?

gtrufitt:
👍 cool

PR #13360, opened 2016-06-21
Add us-news, commentisfree and australia-news as secure LinkTo targets

JustinPinner:
## What does this change?

Makes LinkTo-generated links use https for us-news, commentisfree and australia-news. Existing tests pick up these new values so no new ones have been added.

Prevents an unnecessary redirect from http to https when following links in those sections.

No.


katebee:
👍

PR #13363, opened 2016-06-21
Improve layout of manual inline component on mobile

regiskuckaertz:
Before:
picture 3

After:
picture 2

And:
picture 1

cc @guardian/commercial-dev


JonNorman:
C'est bon :+1:

PR #13365, opened 2016-06-21
Remove password strength check update password

sndrs:
## What does this change?

removes the password strength check. 

apparently it's no longer needed, so we can remove it, and the zxcvbn module

shouldn't do!
## Request for comment

@nlindblad


desbo:
\o/

can't wait for the dictionary to stop showing up in `git grep`

nlindblad:
👍

PR #13367, opened 2016-06-21
Fabric: Fix wrong slot name in articles

regiskuckaertz:
It was decided that the first inline slot should be named `top-above-nav` on smartphones, so that Fabric creatives can be delivered.

**This is quite urgent** as a running campaign is under-delivering.

cc @rich-nguyen


lps88:
👍

rich-nguyen:
👍  great catch, thanks

PR #13370, opened 2016-06-21
init media tests in correct place

akash1810:
## What does this change?

there was no guarantee that the video was enhanced previously; moving initTests to initWithRaven does guarantee the player has finished initializing

videojs elements are on the DOM before a test that manipulates them is run

no
## Screenshots

💻
## Request for comment

@jamesgorrie


dominickendrick:
👍

jamesgorrie:
:rage4: this file.

akash1810:
we don't have any tests running at the moment, so no need to merge this atm - will close and make a not on it for when we add a new test.

PR #13371, opened 2016-06-21
Immersive article headline tweaks

NataliaLKB:
This PR does two things.

https://github.com/guardian/frontend/commit/83793b0277a63e183cda3196647c76bc5e432d8a: The interactive team expect certain behaviour of interactive articles. The behaviour that I accidentally removed (but makes sense!) was if there is no main media than to move the headline to the top of the page.

[this article](https://www.theguardian.com/technology/2016/apr/12/the-dark-side-of-guardian-comments) Currently looks like this:
image

But they would expect it to look like this:
image

Note that the interactive is still broken, but they will fix that!

https://github.com/guardian/frontend/commit/d0e4f5866b3c23ca7cc4fb5b8c001cc705daccd6: The second thing is I removed a few pattern matches and separated out the immersive media template into galleries and articles. There still may be more work to do to make it even better, but that can come in later PRs.

Nope!
## Request for comment

@dominickendrick 🐼

### DIFF
--- a/common/app/views/fragments/immersiveGalleryMainMedia.scala.html
+++ b/common/app/views/fragments/immersiveGalleryMainMedia.scala.html
@@ -1,0 +1,37 @@
+@(page: ContentPage)(implicit request: RequestHeader)
+
+@import layout.ContentWidths.MainMedia
+@import model.ContentPage
+@import views.support.TrailCssClasses.toneClass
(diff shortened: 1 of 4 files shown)


dominickendrick:
👍 It would be good to see how we can clean this up further : )

PR #13374, opened 2016-06-21
AB: Participation - emotional responses

gtrufitt:
## What does this change?

This adds an A/B test for emotional responses. Currently 0% as we need to work out where exactly to test this. It also adds some minor enhancements to the low friction participation JS to allow multiple SVGs and allows better tracking via data-link-names on buttons and confirmations.

Measuring against a baseline of comments on the same pages that icons show, we want to see a 100% higher click-rate on the icon completion.
## Screenshots

emotional
## Request for comment

@crifmulholland @guardian/participation @superfrank


static/src/javascripts/projects/common/modules/experiments/tests/participation-low-fric-emotion.js
```diff
@@ -0,0 +1,129 @@
+define([
+    'common/utils/config',
+    'common/utils/mediator',
+    'common/utils/detect',
+    'Promise',
+    'common/modules/experiments/low-friction-participation',
+    'inlineSvg!svgs/icon/emotion-happy',
```
gtrufitt: @regiskuckaertz Just to check, I am doing inlineSvg right here now?

dominickendrick:
EMOTIONS!!

static/src/javascripts/projects/common/modules/experiments/tests/participation-low-fric-emotion.js
regiskuckaertz: Yes, very good! and if you ever need to decorate the `SPAN` with classes, you can use `common/views/svg` (without s) 👍

regiskuckaertz:
huhuhu, me likey

static/src/javascripts/projects/common/modules/experiments/tests/participation-low-fric-emotion.js
gtrufitt: Super duper.

nicl:
Looks great! Do we need the 'Confirm' dialogue? Feels like it adds unnecessary friction.

nicl:
Also, the emoticons look great, but what does the guy with glasses represent - interested/quizzical?

gtrufitt:
He's confused :/ ;) Maybe these need tooltips to describe the emotion, though I quite like the user being able to interpret it - doesn't help us identify the sentiment very accurately though. 

RE: confirmation - it's been brought up a couple of times - @superfrank is actually looking at how we move this into production so as a part of that we can review that choice and whether it's needed (probably not)

nicl:
@gtrufitt okay thanks! I agree that the user shouldn't require tooltips to understand what they represent. Re: Mr/Ms Confused, I *think* I'd get that. I guess we'll see if people are confused by it (haha) or not once more have seen it!

humaislam:
What needs doing next on this @gtrufitt

gtrufitt:
Hopefully will find out if our team is proceeding this week... Closing this for now, will re-open if and when.

PR #13377, opened 2016-06-22
Remove manual inline switch

regiskuckaertz:
The test has now been running for 1 week without problem.

cc @JonNorman @Calum-Campbell @ScottPainterGNM


JonNorman:
:+1:

Issue #13333, opened 2016-06-17
Missing itemprop meta-data on some pages


katebee:
# ISSUE

Performing a Google search for 'Guardian' returns page previews under the 'News' tab (see image). Some previews are using the OpenGraph image with an overlay.

If the main media of a page is a video, or `page.contentType` is video, Google is taking the OpenGraph image as thumbnail.  This appears to be a fallback behaviour, either because:
- There is no `itemprop="image"` meta-tag specified on the page. 
- There is no `itemprop="thumbnail"` meta-tag specified on the page.

https://support.google.com/news/publisher/answer/13369?hl=en
http://schema.org/

## Steps to Reproduce

Go to Google, search for 'Guardian' or 'Guardian news' and click onto the 'News' tab. Look at the image thumbnails, some will have the branded overlay.

##### Results of searching 'Guardian'
<img width="657" alt="picture 65" src="https://cloud.githubusercontent.com/assets/8607683/16158981/02680ffc-34b8-11e6-8595-31a5c5018ffc.png">

##### Results of searching 'Guardian Video'
<img width="660" alt="" src="https://cloud.githubusercontent.com/assets/8607683/16158196/1add0f78-34b4-11e6-986e-e025701b1238.png">

## Expected Results (include screenshots)

Here is a page that shows up correctly in Google: https://www.theguardian.com/world/video/2016/jun/17/canadian-mp-breaks-down-in-tears-tribute-jo-cox-video

<img width="597" alt="correct thumbnail" src="https://cloud.githubusercontent.com/assets/8607683/16158536/77b74640-34b5-11e6-91a0-0a0ce9e63f6c.png">

## URL

https://www.google.co.uk/webhp#q=guardian+video&tbm=nws

http://www.theguardian.com/commentisfree/2016/jun/16/the-guardian-view-on-jo-cox-an-attack-on-humanity-idealism-and-democracy

TBonnin:
Does this mean there is also a problem with the AMP validator when there is not itemprop="image"?

katebee:
AMP appears to have it's own standard, so will not be affected.

AMP uses this type of tagging system to get content:
`<amp-img src="welcome.jpg" alt="Welcome" height="400" width="800"></amp-img>`

https://www.ampproject.org/docs/reference/spec.html

From what I can understand, the headers of article pages include:
`<link rel="amphtml" href="https://amp.theguardian.com/......`
This takes the AMP bot to a page with AMP tagged content, including amp-img.

[...]

PR #13378, opened 2016-06-22
Give crawlers more choice in meta-tags #13333

katebee:
## What does this change?

Details of issue in #13333 

The change to mainMedia.scala.html adds the missing `itemprop="image"` that is not generated when the main media is video. I believe this is why Google is falling back to the Open Graph image.

This should fix the meta tags on pages like this:
https://www.theguardian.com/commentisfree/2016/jun/16/the-guardian-view-on-jo-cox-an-attack-on-humanity-idealism-and-democracy

Google and AMP should preferentially use Schema.org tags (itemprop); some content types are missing certain meta-tags. Adding them should change how Google and AMP display our content, and remove the banner overlay from the image on links.

AMP
Google search results
Any other web crawlers that use Schema to scrape our site
##### Before:

missing meta-tags
##### With generated meta-tags:

new meta-tags generated
## Request for comment

@dominickendrick @stephanfowler


dominickendrick:
👍

katebee:
Because video.content.twitterCardImage now has a possible overlay, I'm adding a tiny, further change to use Item700 for the `itemprop="image"` rather than either of the Open Graph or Twitter profiles.

PR #13379, opened 2016-06-22
DFP: ring-fence access to iframe document

regiskuckaertz:
Same-origin security policy prevent cross-frame access and may break the whole ad rendering process.

cc @jbreckmckye @rich-nguyen @JonNorman

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/dfp/apply-creative-template.js
+++ b/static/src/javascripts/projects/common/modules/commercial/dfp/apply-creative-template.js
@@ -84,8 +84,12 @@ define([
         }
 
         function fetchCreativeConfig() {
-            var breakoutScript = iFrame.contentDocument.body.querySelector('.breakout__script[type="application/json"]');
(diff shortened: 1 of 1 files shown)


Calanthe:
👍  I am not sure if we care enough but if we do, we can send those errors to sentry https://github.com/guardian/frontend/blob/2e00099b6509024fd5a9f04aa7daea03e08281ac/static/src/javascripts/projects/common/utils/report-error.js.

regiskuckaertz:
@Calanthe thanks Zofia! this is just to prevent same-origin policy to throw an exception at us, nothing very fancy but it can break MPUs.

jbreckmckye:
:+1: - awesome.

It's good to see this part of the codebase getting some polish. One of the reasons I pulled this flow out of dfp-api was to try and encourage that - it's always much easier to see issues and push improvements on small modules, I feel.

regiskuckaertz:
Thanks, Jimmy. I agree, we need to do that more often (addSlot and track* in dfp-api come to mind).

PR #13380, opened 2016-06-22
Fix advert.sizes reference

regiskuckaertz:
Small bug that has been preventing the MPU on the right to be sticky.

cc @rich-nguyen @JonNorman @ScottPainterGNM

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/dfp/render-advert.js
+++ b/static/src/javascripts/projects/common/modules/commercial/dfp/render-advert.js
@@ -48,8 +48,8 @@ define([
      */
     sizeCallbacks[adSizes.mpu] = function (_, advert) {
         if (advert.node.classList.contains('ad-slot--right')) {
-            var mobileAdSizes = advert.sizes('data-mobile');
-            if (mobileAdSizes && mobileAdSizes.indexOf([300, 251]) > -1) {
+            var mobileAdSizes = advert.sizes.mobile;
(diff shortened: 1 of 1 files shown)


Calanthe:
👍

static/src/javascripts/projects/common/modules/commercial/dfp/render-advert.js
```diff
@@ -48,8 +48,8 @@ define([
      */
     sizeCallbacks[adSizes.mpu] = function (_, advert) {
         if (advert.node.classList.contains('ad-slot--right')) {
-            var mobileAdSizes = advert.sizes('data-mobile');
-            if (mobileAdSizes && mobileAdSizes.indexOf([300, 251]) > -1) {
+            var mobileAdSizes = advert.sizes.mobile;
+            if (mobileAdSizes && mobileAdSizes.some(function (size) { return size[0] === 300 && size[1] === 251; })) {
```
rich-nguyen: This function reads a bit implicitly because of the lack of structured typing, which appears to come from `dfp-api`, in the local function `createSizeMapping`. Maybe if `createSizeMapping` returned an object with `width` and `height`, the code would be more readable and debuggable? Not a big issue.

rich-nguyen:
👍

static/src/javascripts/projects/common/modules/commercial/dfp/render-advert.js
regiskuckaertz: Yes, very good idea. Ultimately everything that concerns ad sizes should be exposed through the `adSizes` API.

PR #13381, opened 2016-06-22
AB: Music star ratings - Ready for Glasto reviews

gtrufitt:
## What does this change?

Re-runs the music star-ratings test over Glastonbury weekend so we can get some good statistically significant results over a small period of time with small audience.

Comments on articles vs User clicking on star ratings

image

cc @crifmulholland


static/src/javascripts/projects/common/modules/experiments/low-friction-participation.js
```diff
@@ -45,6 +45,7 @@ define([
         itemIcon: star, // SVG icon
         buttonTextArray: [], // An array of strings to use as the button text, if array is empty will use current iteration value+1
         templateVars: { // Variables that will be passed through to all views
+            typeId: 'stars',
```
NathanielBennett: This is a bit Simply Red

NathanielBennett:
:+1: LGTM. A thought though - do we need some kind of guard to make sure that the cohort aren't in the hide comments test? If we're comparing participations?

gtrufitt:
@NathanielBennett Should be fine as you're only running that 50-60% so there's no overlap in the tests?

PR #13383, opened 2016-06-22
Removing space in amp cors origin list

NataliaLKB:
Some requests are failing for `amp-lists` due to a CORS error.

This is an AMP list:
image

This is what the headers should look like if you are looking at the page on the Google cache:
image

This is what they are currently looking like in some situations:
image

Sometimes this error shows up, and sometimes it doesn't. When I test CORS locally everything seems to work. I think potentially this space in the prod properties might be the cause of this problem. Other ideas welcome 😄 

The Newsstand team at Google want to launch support for AMP. They are waiting for our pages to work perfectly before this.

This should only affect AMP


dominickendrick:
👍

PR #13384, opened 2016-06-22
Bye bye bower pt1

sndrs:
## What does this change?

Begins removing bower from the app, to use only npm. In this case, from `static/src/javascripts` and `static/public/javascripts`.

simplification and hopefully more immediately apprendable codebase

no
## Request for comment

@rich-nguyen


TBonnin:
screen shot 2016-06-22 at 14 36 28 💃 💃 💃

grunt-configs/copy.js
```diff
@@ -20,6 +20,16 @@ module.exports = function (grunt, options) {
                 },
                 {
                     expand: true,
+                    cwd: 'node_modules',
+                    src: [
+                        'react/dist/react-with-addons.js',
+                        'JSON/json2.js'
+                    ],
```
rich-nguyen: would it be possible to locate all the vendor based dependencies from npm? then we could remove `static/src/javascripts/vendor`, and have `formstack`, `prebid` and `stripe` installed from `node_modules`

rich-nguyen:
“apprendable”, the neologist’s solution to a readable codebase born from an unreadable pr (due to github crashing). 👍  Great stuff

grunt-configs/copy.js
sndrs: yeah very happy to do that in subsequent PRs, was trying to limit this to just changes that fell out of the first bower going

jamespamplin:
:tada: thanks @sndrs

kaelig:
w00t

sndrs:
this is too big i think, and doesn't help us move on. closing for now, will implement it even more piecemeal as we move away from require

PR #13386, opened 2016-06-22
Remove omniture requests from AB test

gtrufitt:
## What does this change?

Removes our calls to omniture for AB tests as it tries to call a non-existent file on PROD. I need to look into why it used to work and why it doesn't now, but for now we will have to rely on the data lake to give us everything.


NathanielBennett:
:+1: LGTM

PR #13387, opened 2016-06-22
Dummy a/b test to check that the sampling works as expected

davidfurey:
## What does this change?

Adds an a/b test to check that sampling is working

There is some uncertainty about whether sample sizes are the same for the control and test group with an audience offset of zero.  This test should give us enough data to decided whether or not there is a problem

No


katebee:
👍

PR #13388, opened 2016-06-22
Update LiveEvents component link and images

JonNorman:
### Context

On [some articles](http://www.theguardian.com/lifeandstyle/2016/jun/19/sosharu-restaurant-review-jay-rayner) we embed inline Guardian Live Events components in order to advertise live events that readers of the article may find interesting.
image

We ultimately populate this component from the contents an Eventbrite feed, however not _all_ the information we require is in this feed; Eventbrite do not provide a link to the images. Fortunately, membership, who also work with live events, have [their own stripped down feed](https://membership.theguardian.com/events.json) which _does_ provide a suitable image, so (after parsing the Eventbrite feed), we can read the membership feed in order to get an image for each of our events. Yay.
### What does this change?
1. Currently, clicking the component links through to the URL provided in the Eventbrite feed, which links to an Eventbrite page; however we want it to link to a guardian membership page. Fortunately, the membership page URL is specified in the membership feed (see above) and given that we are already parsing that, we can just grab the URL in addition to the image link we are already getting.
2. Currently the image being served (taken form the membership feed) can be huge; the one linked above is a whopping **60MB!?1**, which isn't ideal. So this PR also uses the image resizing capabilities in `frontend` to request a more palatable image:

**before**
image

**after**
image

Users will be taken to the correct page and it won't use up 10% of their data plan to do so.


commercial/app/model/commercial/events/LiveEvent.scala
```diff
@@ -28,24 +24,16 @@ case class LiveEvent(eventId: String,
...
       name = event.name,
       date = event.startDate,
-      eventUrl = event.url,
+      eventUrl = maybeEventInternalFeed map {_.url} getOrElse event.url,
       description = event.description,
       status = event.status,
       venue = event.venue,
       tickets = event.tickets,
-      imageUrl = maybeImageUrl
+      imageUrl = maybeEventInternalFeed map {_.mainImageUrl}
```
Calum-Campbell: Should the image be optional here?

JonNorman: Good point - images are no longer optional; when parsing our EB feed, if we do not find an associated image from the membership feed, we don't add the LiveEvent to our agent.

commercial/app/model/commercial/events/LiveEvent.scala
```diff
@@ -2,20 +2,27 @@ package model.commercial.events
...
+
+object LiveEventMembershipInfo {
+  implicit val format = Json.format[LiveEventMembershipInfo]
+
+  // based on play.api.libs.json.LowPriorityDefaultReads.traversableReads
+  implicit val formats = new Reads[Seq[LiveEventMembershipInfo]] {
+    override def reads(json: JsValue): JsResult[Seq[LiveEventMembershipInfo]] = {
+      json match {
+        case JsArray(jsValues) => JsSuccess(jsValues.flatMap(_.asOpt[LiveEventMembershipInfo]))
+        case _ => JsError(Seq(JsPath() -> Seq(ValidationError("error.expected.jsarray"))))
```
kelvin-chappell: This is a strange error message.  Could it be more informative?

commercial/app/model/commercial/events/LiveEventAgent.scala
```diff
@@ -37,24 +36,23 @@ object LiveEventAgent extends ExecutionContexts with Logging {
...
+        (json \ "events").as[Seq[LiveEventMembershipInfo]]
       }
 
-      requestEventImages map { response =>
-        if (response.status == 200) {
-          parseEventImages(response.body)
-        } else {
+      requestLiveEventsMembershipInfo map { response =>
+        if (response.status == 200)
+          parseLiveEventsMembershipInfo(response.body)
```
kelvin-chappell: `response.json` will give you the body as a `JsValue`.

commercial/app/model/commercial/events/LiveEventAgent.scala
```diff
@@ -64,10 +62,12 @@ object LiveEventAgent extends ExecutionContexts with Logging {
 
     val liveEvents =
       for{
-        eventImages <- fetchAndParseLiveEventImages
+        eventMembershipInfo <- fetchAndParseLiveEventsMembershipInfo
         ParsedFeed(events, _) <- Eventbrite.parsePagesOfEvents(feedMetaData, feedContent)
-      } yield events
-            .map ( event => LiveEvent(event, eventImages find (_.eventId == event.id)) )
+      } yield events flatMap { event =>
```
kelvin-chappell: Could this flatMap be done as part of the for-comprehension?  I'm not sure.  
`events <- event`

commercial/app/model/commercial/events/LiveEvent.scala
JonNorman: Agreed, I was going off the example of https://github.com/guardian/frontend/blob/master/commercial/app/model/commercial/soulmates/Member.scala#L34

Ultimately, if we do enter that block then we were expecting to parse a JSON array but were not provided one - I suppose I can specify the expected content of the array i.e. LiveEventMembershipInfo

commercial/app/model/commercial/events/LiveEventAgent.scala
```diff
@@ -64,10 +62,12 @@ object LiveEventAgent extends ExecutionContexts with Logging {
...
+        val matchingMembershipInfo = eventMembershipInfo find (_.id == event.id)
+        matchingMembershipInfo map (LiveEvent(event, _) )
```
kelvin-chappell: Could convert both lists into hashmaps and then use `merged` to combine them.  But it would possibly be less readable.

kelvin-chappell:
Looks good. I like the solution to the image problem.
A few minor comments, but nothing important.
👍

commercial/app/model/commercial/events/LiveEvent.scala
kelvin-chappell: Yes, that would be better.  So if you read it in the logs you know immediately what's gone wrong.

commercial/app/model/commercial/events/LiveEventAgent.scala
JonNorman: Not that I could see; if we can't find a matching image and URL for a given `LiveEvent` then we want to discard it the `LiveEvent` completely.

If I add in `event <- events` to the comprehension, then I wouldn't be able to filter those events out unless I used a guard on the for comprehension. However, in order to use a guard, I'd have to also perform the matching in the comprehension (to then use the result in the guard), which cannot be done since I am iterating over `Futures` and the matching returns an `Option`, and I cannot easily convert between the two.

JonNorman: I could do it if I repeat the guard condition, but not sure it is much of an improvement:

```
val liveEvents =
      for{
        eventMembershipInfo <- fetchAndParseLiveEventsMembershipInfo
        ParsedFeed(events, _) <- Eventbrite.parsePagesOfEvents(feedMetaData, feedContent)
        event <- events
        if eventMembershipInfo.exists(_.id == event.id)
      } yield eventMembershipInfo.find(_.id == event.id).map(LiveEvent(event, _))
```

kelvin-chappell: No.  Best left as is.

PR #13389, opened 2016-06-22
Standalone controller migration

alexduf:
## What does this change?

This is a very repetitive PR that transforms all the controllers used by standalones apps (preview and training-preview) into classes.
It also creates a singleton for each one of them in order to be backward compatible for the time being

This is a necessary step to be able to migrate the standalone apps away from using globals, which is itself a necessary step towards play 2.5
## Request for comment

@TBonnin @johnduffell

### DIFF
--- a/admin-jobs/app/controllers/NewsAlertController.scala
+++ b/admin-jobs/app/controllers/NewsAlertController.scala
@@ -53,10 +53,12 @@ trait NewsAlertController extends Controller with AuthenticationSupport with Exe
   }
 }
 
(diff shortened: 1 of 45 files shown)


TBonnin:
e07b807b1cdc72b66c65c88e2ff2a5549b2809f3b380676d20c32e1028c9cd32

I admire your effort though :)

👍

alexduf:
Ahah, wait to see the PR related to dev-build -__-

I'll wait for the referendum to pass before merging anything

PR #13390, opened 2016-06-22
Standalone lifecycle migration

alexduf:
## What does this change?

This is another very repetitive PR that modifies all the lifecycle objects used by standalones apps (preview and training-preview) in order to stop depending on Play.current.
It also provides default values for each one of them in order to be backward compatible for the time being

This is a necessary step to be able to migrate the standalone apps away from using globals, which is itself a necessary step towards play 2.5
## Request for comment

@TBonnin @johnduffell


TBonnin:
lgtm 👍

PR #13392, opened 2016-06-22
In homepage weather widget search, show 10 results

TBonnin:
## What does this change?

Weather widget: Accuweather API returns 10 results when searching for a city, however we only display the first 5 items at the moment. 
=> if the city you are looking for is not in the first 5 items it is not displayed :(

No more user complaining they cannot find their city

No
## Request for comment

@sndrs 
Still looking for a way to test this locally :/

### DIFF
--- a/static/src/javascripts/projects/facia/modules/onwards/search-tool.js
+++ b/static/src/javascripts/projects/facia/modules/onwards/search-tool.js
@@ -148,7 +148,7 @@ define([
                     type: 'json',
                     crossOrigin: true
                 }).then(function (positions) {
-                    this.renderList(positions, 5);
+                    this.renderList(positions, 10);
                     oldQuery = newQuery;
                 }.bind(this));
             },
--- a/static/src/stylesheets/module/_search-tool.scss
+++ b/static/src/stylesheets/module/_search-tool.scss
@@ -44,7 +44,7 @@
     box-sizing: border-box;
     color: $neutral-1;
     padding-top: $gs-baseline / 3;
-    padding-bottom: $gs-baseline;
+    padding-bottom: $gs-baseline / 3;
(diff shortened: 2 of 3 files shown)


static/src/javascripts/projects/facia/modules/onwards/search-tool.js
```diff
@@ -215,9 +215,8 @@ define([
                 $input.val(inputValue);
             },
 
-            renderList: function (results, numOfResults) {
-                var docFragment   = document.createDocumentFragment(),
-                    resultsToShow = results.length - numOfResults;
```
TBonnin: Not sure what was the intention here 🔪

static/src/stylesheets/module/_search-tool.scss
```diff
@@ -44,7 +44,7 @@
     box-sizing: border-box;
     color: $neutral-1;
     padding-top: $gs-baseline / 3;
-    padding-bottom: $gs-baseline;
+    padding-bottom: $gs-baseline / 3;
```
TBonnin: George advice was to reduce the padding. It is odly unbalanced in prod currently
screen shot 2016-06-22 at 17 34 33

sndrs:
does this scale well? was the original cut to 5 to handle long lists in the response?

TBonnin:
I am not sure I understand what you mean by `scale`
Thinking about it anyway, a better way is `resultsToShow = min(results.length, 10)`

sndrs:
scale to long lists :)

TBonnin:
Scratch that. If the container is too short, the list is truncated. :( I am not sure there is a clean solution here

sndrs:
how about a max-height on it, with `overflow: auto`?

PR #13393, opened 2016-06-22
add data-link to the quiz questions

NathanielBennett:
## What does this change?

We wanna track quiz usage. This adds data-link-names in quiz answers ( on a per question basis ) so that those clicks end up in omniture in lieu of the datalake work being completed

Quiz answer clicks are recorded in omnture, that enables us to begin the value of quizzes

No
## Screenshots

N/A
## Request for comment

@nicl @gidsg 
@crifmulholland this would have taken 5 minutes if play didn't suck


gidsg:
👍 🐻

PR #13394, opened 2016-06-22
Hosted gallery prototype for mobile

lps88:
## What does this change?
- add swiping left-right behaviour for touch screens
- lazy load the images
- use events for triggering moving between images, etc.
- add captions (toggle on/off using 'i' button)

Still a bit of work to do around the call to action banner for the last slide.

NB - doesn't affect anything on prod, would like to release on Thursday ready for a demo 😺 
## Screenshots

image

image
## Request for comment

@Calanthe @regiskuckaertz


paperboyo:
Super-humble-personal-opinion-ignore-at-will:

Our Lightbox (and carousels) work horizontally as do similar devices on the wider internets. When I first saw this here prototype, I thought "nice, but why vertical?!", but haven't asked for a reason of this inconsistency, thinking that these are "hosted", so maybe that's the way (strange, IMHO) to get them distinguished. Or, maybe, the car was asian...

But I can see now, that on mobile - they are suddenly back to being horizontal. Can't contain myself: why vertical at all then?!

lps88:
Yeah good question. I suppose the images themselves aren't moving, and scrolling between them is easier vertically (at least on a mouse wheel).

We've got a demo today at 2pm in MCR if you want to come and ask your question to the group, otherwise I'll bring it up  😸

ScottPainterGNM:
It's a really good question and one worth discussing I think as it may prove to be beneficial going forward as you won't need to implement different actions per breakpoint.

I'm happy to bring up in stand up as not around this afternoon for demo alas.

paperboyo:
I'm on :palm_tree:, sadly...

But I do think actions (and underlying model that suggests structure relationships) should not change on different breakpoints. Facebook natives seem to prefer vertical scrolling, probably liking Lightbox experience to feed one. That is IMHO wrong, because even there the vertical feed is traversing different sources and to watch a progression of items from one source you can enter Facebook's own Lightbox (incl. real fullscreen!) where stuff happens horizontally (that in this mode one can't go down/up too to change sources as well while still in fullscreen is a failure on FB's part)...

Our main Lightbox uses up/down keys to show/hide captions (in addition to `i` key). There is no reason that the mousewhell could not be hooked up to work in a horizontal gallery (it works like that in our Win10 app, for example).

I strongly feel all Lightboxes should follow the model that's most familiar to billions of users - that of a Camera Roll-like apps on mobiles - horizontal swipe. And that's what a huge majority of them do, I think.

Until, that is, we start publishing in Chinese and then Hebrew and the whole site's reading/structure flow will react to reader's native tongue writing direction :wink:

commercial/app/views/hosted/guardianHostedGallery.scala.html
```diff
@@ -7,32 +7,36 @@
...
+                    <div class="hosted-gallery__scroll-container">
+                        @for(i <- page.images.indices) {
+                            <div class="hosted-gallery__image--placeholder"></div>
+                        }
                     </div>
-                    <div class="hosted-gallery__image-count">
-                        1/@page.images.length
+                    <div class="hosted-gallery__progress">
+                        @fragments.inlineSvg("arrow-up", "icon")
+                        <div class="hosted-gallery__progress--wrapper">
```
Calanthe: I would probably name it `hosted-gallery__progress-wrapper`, because (afaik) `wrapper` is not a modifier. But maybe ask someone who understands BEM more than I do ;)

commercial/app/views/hosted/guardianHostedGallery.scala.html
```diff
@@ -7,32 +7,36 @@
...
+                            <div class="hosted-gallery__progress--border">
```
Calanthe: Same here, I would name it `hosted-gallery__progress-border`

static/src/javascripts/projects/common/modules/commercial/hosted-gallery.js
```diff
@@ -3,79 +3,386 @@ define([
...
+        bean.on(document.body, 'keydown', this.handleKeyEvents.bind(this));
+
+        bean.on(this.nextBtn, 'click', this.scrollTo.bind(this, 1));
+        bean.on(this.prevBtn, 'click', this.scrollTo.bind(this, -1));
+
+        bean.on(this.$scrollEl[0], 'scroll', throttle(this.fadeContent.bind(this), 20));
+    };
+
+    HostedGallery.prototype.initSwipe = function () {
+
```
Calanthe: [minor] remove whitespace?

lps88:
Hmm. Well it would be easy to swap the arrow buttons to left/right, alter the key bindings so they match the lightbox, and leave the scrolling as it is. I might try that out.

If it makes any difference, at the moment choosing between the scrolling/swiping behaviour is decided by detecting touch screen, not by breakpoints. All this is still in prototype stage so isn't final!

static/src/javascripts/projects/common/modules/commercial/hosted-gallery.js
```diff
@@ -3,79 +3,386 @@ define([
...
+        var threshold, ox, dx, touchMove,
+            updateTime = 20; // time in ms
+        this.swipeContainerWidth = this.$galleryEl.dim().width;
+        this.$imagesContainer.css('width', this.$images.length + '00%');
```
Calanthe: Are you sure there will be always less than 10 images?

static/src/javascripts/projects/common/modules/commercial/hosted-gallery.js
```diff
@@ -3,79 +3,386 @@ define([
...
+
+        bean.on(this.nextBtn, 'click', this.scrollTo.bind(this, 1));
+        bean.on(this.prevBtn, 'click', this.scrollTo.bind(this, -1));
+
+        bean.on(this.$scrollEl[0], 'scroll', throttle(this.fadeContent.bind(this), 20));
+    };
+
+    HostedGallery.prototype.initSwipe = function () {
+
+        var threshold, ox, dx, touchMove,
```
Calanthe: What does `ox` and `dx` mean?

static/src/javascripts/projects/common/modules/commercial/hosted-gallery.js
```diff
@@ -3,79 +3,386 @@ define([
...
+                if (res) {
+                    gallery.index = parseInt(res[1], 10);
+                }
+            }
+        });
     }
 
     return {
-        init: init
+        init: bootstrap,
```
Calanthe: why not `init`? ;)

Calanthe:
I like the swipe drawing you made ;)
Good job with the gallery! Maybe ask also another front-end dev for feedback.

static/src/javascripts/projects/common/modules/commercial/hosted-gallery.js
lps88: It will still work fine, this is to set the width of the containing element. Then the images all take up equal space inside, so they're each 100% width (of the screen).

commercial/app/views/hosted/guardianHostedGallery.scala.html
lps88: Yes, I have got mixed up in my css conventions, I will go through them properly soon! This stuff is changing fast though! 😖

paperboyo:
> choosing between the scrolling/swiping behaviour is decided by detecting touch screen, not by breakpoints

I can imagine (a totally contrived!) scenario, where one has two displays of same size and resolution connected to one PC but only one having a touchscreen. Gallery would then render differently depending on where the browser window would be placed (assuming OS+browser correctly emits touchscreen availability/changes). Awkward, no?

Obviously, this isn't really this-PR problem as such, just my (poor) attempt to prove a more general point. Also: sorry, @lps88, for hijacking your PR for such a discussion. Not a best place for it, I know.

kelvin-chappell:
👍

PR #13401, opened 2016-06-23
Logic added for krux off fronts

Calum-Campbell:
## What does this change?
- Do not load Krux on fronts
- Krux tag firing when users visit fronts costs a lot of money and produces minimal useful data.
## Request for comment

@rich-nguyen @JonNorman @kenlim


JonNorman:
:+1:

PR #13403, opened 2016-06-23
Move badging back to separate files to avoid flash

sammorrisdesign-zz:
## What does this change?

This moves the reference to the `badging.scss` partial back to the `head.content` and `head.facia` files rather than `common`. The reason for this is to avoid a repaint where the badges are massive before being resized by the parent container found in `head.content` or `head.facia`. 💥
## Request for comment

@dominickendrick


dominickendrick:
👍

PR #13404, opened 2016-06-23
Change tracking to track granted only when it's a new state

janua:
This fixes the tracking so that it will only count `granted` state if it has not been granted before.

@samdesborough @NathanielBennett

### DIFF
--- a/static/src/javascripts/bootstraps/enhanced/notifications.js
+++ b/static/src/javascripts/bootstraps/enhanced/notifications.js
@@ -109,10 +109,11 @@ define([
         },
 
         subscribeHandler: function () {
-            var alreadyGranted = Notification.permission === 'granted';
+            var wasNotGranted = Notification.permission !== 'granted';
             modules.subscribe().then(modules.follow)
                 .then(function() {
-                    if (!alreadyGranted && Notification.permission === 'granted') {
+                    var isNowGranted = Notification.permission === 'granted';
+                    if (wasNotGranted && isNowGranted) {
                         omniture.trackLinkImmediate('browser-notifications-granted');
                     }
                 }) .catch( function () {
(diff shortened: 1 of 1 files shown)


static/src/javascripts/bootstraps/enhanced/notifications.js
```diff
@@ -107,9 +107,10 @@ define([
         },
 
         subscribeHandler: function () {
+            var alreadyGranted = Notification.permission === 'granted';
             modules.subscribe().then(modules.follow)
                 .then(function() {
-                    if (Notification.permission === 'granted') {
+                    if (!alreadyGranted && Notification.permission === 'granted') {
```
sndrs: is `Notification.permission` mutated by subscribing/following?

janua: Yes... Is that going to be a problem?

michaelwmcnamara:
Worth holding off till Monday?

janua:
@michaelmcnamara No one will be deploying anything today.

static/src/javascripts/bootstraps/enhanced/notifications.js
desbo: it does make this code a bit of a weird read,
I had a quick look if there's an event that's fired when `Notification.permission` changes, but there doesn't seem to be one. 
I'd probably just add a comment around here that explains it gets changed at some point during `modules.subscribe` or `modules.follow`

sndrs: is it `if (wasNotGranted && isNowGranted) {}`, is that the logic?

janua:
@sndrs That's exactly the logic. I renamed the `var`s to make it more clear; maybe I shouldn't have the superfluous `var` though?

janua:
Also added a comment closer to where the permission gets asked.

sndrs:
👍 :beautiful:

PR #13407, opened 2016-06-24
Redisplay engaged reader banner

paulbrown1982:
## What does this change?

It redisplays the Membership engaged reader banner to those who may have dismissed it before.

We typically do this once every 2 months and it usually results in a sizeable uplift of new supporters, on today's busy traffic day it should be even more.

Only desktop web.
## Screenshots

picture 130
## Request for comment

cc @dominickendrick


jamespamplin:
:+1:

Issue #8119, opened 2015-01-30
‘More from this series’ video endslate fail


paperboyo:
Hello,

All desktop platforms/browsers tested.
- some videos from the series show the end slates while embedded on fronts  - they should not (e.g. http://www.theguardian.com/lifeandstyle/video/2015/jan/30/get-over-ex-boyfriend-agony-aunt-problem-video)
- other do not display the end slate at all, which is wrong as well (‘Film show’ - a lot of available content)
- if there is just one more piece of content from a specific series, like in a link above, end slate looks wrong

Screenshots:

![series slate fail 1](https://cloud.githubusercontent.com/assets/6032869/5985268/0a043c6c-a8d8-11e4-93cb-602674ebd5f7.PNG)

![endslate-fail-2](https://cloud.githubusercontent.com/assets/6032869/5985274/0df233f6-a8d8-11e4-9dc6-6d00f9db0f0f.jpg)

![endslate fail 3](https://cloud.githubusercontent.com/assets/6032869/5985275/10fb0b22-a8d8-11e4-818e-37d58292ad5d.PNG)

Regards
Mateusz

johnduffell:
I have fixed the first point:
- some videos from the series show the end slates while embedded on fronts - they should not (e.g. http://www.theguardian.com/lifeandstyle/video/2015/jan/30/get-over-ex-boyfriend-agony-aunt-problem-video)

paperboyo:
Thx!

johnduffell:
second point, I can't reproduce?  If I skip to the end of this, http://www.theguardian.com/film/video/2015/jan/30/the-guardian-film-show-kingsman-the-secret-service-inherent-vice-trash-big-hero-6
then I can see some related content
![image](https://cloud.githubusercontent.com/assets/7304387/6108219/2120168a-b069-11e4-8d3c-4aace388136e.png)
is it still broken?

paperboyo:
Hi,

Ok, that's going to be embarassing... AdBlock Plus is preventing the end slate from showing up in both Chrome and Firefox on both Mac and Windows (default extension filters).

Hmm... I though I do not miss any of the good stuff... It's a pity it gets blocked, shouldn't, right?

:facepalm:
![embarrassed](https://cloud.githubusercontent.com/assets/6032869/6108432/0d080d4a-b06b-11e4-85ed-118636619f9c.gif)


Regards
Mateusz

johnduffell:
I think that's still kind of a bug, I'll have a look.

[...]

PR #13408, opened 2016-06-24
show empty end slate if there are no results

jamesgorrie:
## What does this change?

@TBonnin noticed that we 404 if there is no results from CAPI on the endslate. This isn't a great response, we should probably just say, "Sorry, no videos". This does that.

It also stops the 404 not caching, and thus doing a CAPI call everytime.

If deemed a simple enough change, I'll push this out.

CAPI loses some load.

Nope
## Request for comment

@akash1810


akash1810:
👍

alexduf:
I had a look as well, it looks good. :+1: 
I think we should only deploy onwards though

TBonnin:
👍

paperboyo:
Related to (3rd point of) https://github.com/guardian/frontend/issues/8119.

PR #13409, opened 2016-06-24
Extend tech top slot switch

JonNorman:
Extends a switch on behalf of @regiskuckaertz


janua:
This seems to do a lot more than just extend a switch!

JonNorman:
Woah! Yep, they are changes I made for a PR that I raised right before this one and clearly forgot to switch back to master, I'll re-raise this later!

PR #13410, opened 2016-06-24
Notification tracking changes

janua:
## What does this change?

This adds a `data-link-name` to the notifications explainer in order to track who thinks it is a CTA.

This also moves the `data-link-name` in the signup button for notifications to the `button` parent; it is currently possible to sign up (and unsubscribe) to notifications but side step the tracking by clicking the button just outside the sign-up text in the DOM. This makes it harder to track who has clicked the button as they have to click exactly on the text to have it tracked.

This is valuable to know which parts of the notification workflow are working and not working. This will also help with tracking users expectations of the explainer.

No
## Request for comment

@NathanielBennett @gtrufitt @crifmulholland


gtrufitt:
👍

PR #13411, opened 2016-06-24
Remove FB overlay on paid content

katebee:
## What does this change?

The branded overlay on Open Graph will no longer appear on images when paid content is shared to other sites. 

Swapping to using one of the other image profiles for the `og:image` paid content

Success will be if we exclude the banner from 'paid for' content posts.

All platforms scraping our Open Graph images (Facebook, WhatsApp, etc.)
##### BEFORE:

Before
##### AFTER:

After
## Request for comment

@stephanfowler @dominickendrick


dominickendrick:
👍

PR #13413, opened 2016-06-27
the test was meant to run for a week rather than 5 days

akash1810:
## What does this change?

we've lost a few hours of data - we may have to run the test again 😢 

correct test duration

no
## Screenshots

n/a
## Request for comment

@jamesgorrie


jamesgorrie:
👍

PR #13414, opened 2016-06-27
Add custom dimensions to Google Analytics pageview event (again)

cb372:
This is the second attempt. The first one (#13355) had to be reverted because it broke Ophan. The issue was that I moved the Ophan pageview ID generation to the server side, totally forgetting that Fastly would cache it, thus causing thousands of page views to be assigned the same pageview ID.

Adds the following custom dimensions to the GA pageview event:
1. Ophan pageview ID
2. Ophan browser ID
3. Platform (hardcoded to "theguardian.com")
4. Section ID
5. Content type
6. Commissioning desk
7. Content ID
8. Author name(s)
9. Keyword tag IDs
10. Tone tag IDs
11. Series tag ID

This required moving the Ophan pageViewId generation out of an AMD module and into `inlineBlockingJS` so it is available when the GA code runs.


cb372:
/cc @gklopper @obrienm 

I'm more confident that this will work as expected now, after some help from @philwills. But I'm happy to hold off on merging this for the next few days, given that we have a semi-moratorium on Ophan-related changes at the moment.

common/app/views/fragments/inlineJSBlocking.scala.html
```diff
@@ -59,6 +58,11 @@
     }
 
 
+    // ************* ANALYTICS *************
+
+    // Ophan pageview ID and browser ID are needed by Google Analytics, which runs sync at the bottom of the page
+    @InlineJs(ophanConfig().body, "ophanConfig.js")
```
cb372: @sndrs Can you see any problem with moving the Ophan pageview ID generation from the `utils/config` AMD module to the inline blocking JS section like this? The reason for doing so is to make it available at page load time so the GA code can read it.

obrienm:
Can we hold off until next week? Chris Moran is really very nervous of any change.

common/app/views/fragments/inlineJSBlocking.scala.html
sndrs: not with with being inlined, but can it go to the non-blocking? we don't do anything with ophan until domcontentlaoded fires anyway, so it should be fine. blocking is really for stuff we need done before paint can begin.

sndrs: we also have `nonBlocking/getUserData.scala.js` down there, which you should be able to re-purpose/combine with the cookie reading code?

cb372: I see, I thought "non blocking" meant async, but on closer inspection it actually means "stick it at the bottom of the `body`". In that case I'll move it there. Thanks.

sndrs: yeah maybe we should rename it `render-blocking` and `non-render-blocking`

jbreckmckye: @sndrs Or perhaps `before-render` and `after-render`?

cb372:
Got the verbal 👍  from @obrienm today to deploy this, so I'm going to 👍  myself and merge!

common/app/views/fragments/inlineJSBlocking.scala.html
cb372: Sounds like a sensible rename. But I wanted to keep this PR as minimal as possible, so I've left the names alone for now.

PR #13415, opened 2016-06-27
enable Wimbledon ad for live blogs and wimbledon-2016 section

JonNorman:
## What does this change?

The Wimbledon top-slot ad - set to run from now until the end of the tournament - is intended to run on Wimbledon live blogs as well as the **wimbledon-2016** section. This changes makes it so.

It isn't nice to do pattern matching, but this is a fix to a temporary piece of functionality that is well tested and signed off.
## Request for comment

@regiskuckaertz @Calum-Campbell


Calum-Campbell:
👍

PR #13416, opened 2016-06-27
Change updateSubscription to handle cases of the subscription turning to

janua:
## What does this change?

This changes the call to `updateSubscription` to check if the `subscription` is `null`. This can happen when a user changes their notification preferences for a domain; and probably other reasons I am unaware of.

If a user is subscribed and changes their notification preferences for the domain, the status of the subscription (in the browser) and stay stuck in subscribed as this will try to get a `endpoint` of a `null` and fail the chain of calls in `JS` to fully unsubscribe.

This makes notifications more reliable.

No
## Request for comment

@gtrufitt @NathanielBennett @crifmulholland


gtrufitt:
👍

PR #13417, opened 2016-06-27
Turn on quick survey regarding visit frequency

katebee:
## What does this change?

Turns on the survey added in #13337
The aim is to run the survey over a 24hr period, and receive a minimum of 1000 responses.
Audience participation will be 25%
## Request for comment

@dominickendrick


dominickendrick:
👍

PR #13418, opened 2016-06-27
Standalone app loader

alexduf:
## What does this change?

This removes the global from the previews apps (preview and training-preview)

One step closer to Play 2.5
## Request for comment

@TBonnin @JustinPinner


TBonnin:
lgtm 👍

PR #13419, opened 2016-06-27
pass tags through the templates to reliably determine Wimbledonness

JonNorman:
This is a followup to #13415 - instead of inspecting the URL/ID which is fraught with annoying edge cases (and requires that the articles are published with the phrase "Wimbledon 2016"), we now pass in the keywords to the top banner template and inspect them for "sport/wimbledon-2016".


kelvin-chappell:
👍

Issue #10775, opened 2015-10-05
grunt-sass and phantom versions incompatible with latest node version (4.1.1)


joelochlann:
When running `npm install` (on its own or as part of `setup.sh`), these dependencies in package.json fail to install under node 4.1.1:

```
    "grunt-sass": "1.0.0"
    "phantom": "0.6.5"
```
See log below for details.

Downgrading the node version to 0.12.* avoids the problem. However, the README setup instructions say to run `brew install node` which as of recently will install node 4.1.1. So the long-term solution is probably to bump the package versions of grunt-sass and phantom to ones which are compatible with newer versions of node.

## Steps to reproduce
Run `npm install` with node version 4.1.1 installed.

## Environment
**Node version:** 4.1.1
**OS:** OS X Yosemite. Darwin Kernel Version 14.5.0: Wed Jul 29 02:26:53 PDT 2015; root:xnu-2782.40.9~1/RELEASE_X86_64

## Log output:
```
npm WARN engine karma@0.12.31: wanted: {"node":"~0.8 || ~0.10"} (current: {"node":"4.1.0","npm":"2.14.4"})
npm WARN peerDependencies The peer dependency karma@>=0.11.14 included from karma-coffee-preprocessor will no
npm WARN peerDependencies longer be automatically installed to fulfill the peerDependency
npm WARN peerDependencies in npm 3+. Your application will need to depend on it explicitly.
npm WARN peerDependencies The peer dependency grunt@~0.4.0 included from grunt-contrib-copy will no
npm WARN peerDependencies longer be automatically installed to fulfill the peerDependency
npm WARN peerDependencies in npm 3+. Your application will need to depend on it explicitly.
npm WARN peerDependencies The peer dependency grunt@~0.4.0 included from grunt-contrib-clean will no
npm WARN peerDependencies longer be automatically installed to fulfill the peerDependency
npm WARN peerDependencies in npm 3+. Your application will need to depend on it explicitly.
npm WARN peerDependencies The peer dependency karma@>=0.9 included from karma-spec-reporter will no
npm WARN peerDependencies longer be automatically installed to fulfill the peerDependency
npm WARN peerDependencies in npm 3+. Your application will need to depend on it explicitly.
npm WARN peerDependencies The peer dependency grunt@~0.4.2 included from grunt-autoprefixer will no
npm WARN peerDependencies longer be automatically installed to fulfill the peerDependency
npm WARN peerDependencies in npm 3+. Your application will need to depend on it explicitly.
[...]

PR #13420, opened 2016-06-27
Make node 5 a requirement for frontend

sndrs:
## What does this change?
- adds npmrc and declare 5 a requirement
- manages minor updates to sass packages

gets everyone in the same boat


setup.sh
```diff
@@ -83,17 +83,16 @@ install_jdk() {
...
       sudo apt-get install -y curl
     fi
 
     if linux; then
-      curl -sL https://deb.nodesource.com/setup | sudo bash -
-      sudo apt-get install -y nodejs
+      curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.31.2/install.sh | bash
     elif mac && installed brew; then
-      brew install node
+      brew install nvm
```
desbo: could be worth adding `nvm install 5`  and `nvm use`

setup.sh
```diff
@@ -83,17 +83,16 @@ install_jdk() {
   fi
 }
 
-install_node() {
+install_nvm() {
   if ! installed node || ! installed npm; then
```
desbo: check for `nvm` here?

desbo:
👍  apart from the minor setup bits

setup.sh
sndrs: `nvm install 5` is a great call - does `use` feel like it's taking over your machine a bit?

static/src/stylesheets/base/_type.scss
```diff
@@ -56,7 +56,7 @@ h6 {
 
 
 // SerifText
-.type-5 { @extend %type-5; }
+.type-5 { @extend %type-5 !optional; }
```
jamesgorrie: Why do we need all the optionals? As it's our codebase can't we ensure they're there?

setup.sh
desbo: i don't think it's too bad since nvm makes switching version super easy, as long as we're explicit that it's happening. people are gonna need to run that eventually for nvmrc to be useful anyway. 

not too fussed though, i guess without it we should mention that you need to do `nvm use` manually at some point for anyone not familiar with nvm

static/src/stylesheets/base/_type.scss
sndrs: yeah certainly, but we weren't before, and the older version of node-sass was incorrectly skipping over these bugs, thats' all

setup.sh
sndrs: added link to https://gist.github.com/sndrs/5940e9e8a3f506b287233ed65365befb in extra steps?

desbo: nice!

sndrs:
going to close this, move to 6 (only had to update karma). will close to keep pr clean

joelochlann:
🙌   Was finally able to close [this issue](https://github.com/guardian/frontend/issues/10775) that I opened in my first week here!

PR #13425, opened 2016-06-28
Hosted Gallery - restyle/refactor progress circle & images

lps88:
## What does this change?
- Simplified css and new look for progress circle:

Before:
image

After:
image
- Align the text properly with the images:

Before:
image

After:
image
- Use img tags with alt text instead of background images

CC @Calanthe


Calanthe:
I definitely prefer the one Before...

Calanthe:
👍

PR #13426, opened 2016-06-28
Make Node 6 a requirement for frontend 

sndrs:
## What does this change?
- adds `.npmrc` and declares 6 a requirement
- installs `nvm` instead of `node` in `setup.sh`
- tidies up `package.json`
- manages minor updates to sass packages
- updates karma and phantom

gets everyone in the same boat

teamcity will need to be updated to install node 6
## Request for comment

@samdesborough


rich-nguyen:
Shouldn't we be targeting the active LTS version? That's consistent with the rest of our software images.

sndrs:
if it builds, it seems ok? it's the road to being LTS, and you get near-es6 with 6?

desbo:
i also think it's fine to use 6, it's due for LTS in october
if we have a known working version (i.e. the one @sndrs tested here) we could set that specific version in `nvmrc`

setup.sh
```diff
@@ -84,17 +84,19 @@ install_jdk() {
...
     if linux; then
-      curl -sL https://deb.nodesource.com/setup | sudo bash -
-      sudo apt-get install -y nodejs
+      curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.31.2/install.sh | bash
     elif mac && installed brew; then
-      brew install node
+      brew install nvm
     fi
+
+    nvm install 5
```
desbo: correcting myself but this could be `nvm install` which'll use the version `nvmrc` #decoupling

rich-nguyen:
Ok, it seems like we are favouring new features over stability and security. Which I do think is fine. Whilst having it building and locking it to a minor semver seems like appropriate mitigation, I still think it's easier to just accept that less stable versions have a larger level of risk which we are willing to fix reactively, and get the new features in our codebase ahead of October.

> Note that while v6 will eventually transition into LTS, until it does, we will still be actively landing new features (semver-minor). This means that there is an increased chance for regressions to be introduced.

sndrs:
> Ok, it seems like we are favouring new features over stability and security.

i guess i'm assuming that any instability will be low as its on the way anyway, and that effects will be minimal and/or caught (i wouldn't do this for a production server, for instance)

the brexit argument

rich-nguyen:
yep makes sense. 👍

jamespamplin:
Great stuff @sndrs :tada: :+1:

sndrs:
this is now waiting on updating team city build agents to node 6, which will do tomorrow

johnduffell:
big thumbs up for consistency, this has caused no end of questions from people setting up

cb372:
Once this is merged, please update the [Installation wiki page](https://github.com/guardian/frontend/wiki/Installation-steps). It contains multiple references to the Node version.

sndrs:
this is changing slightly, will reopen later, more cleanly

PR #13429, opened 2016-06-28
Enable https for everything except politics and standalone interactives

JustinPinner:
## What does this change?

Switches Link-to links to https by default **except** politics and standalone interactives.
Enable users to opt out (for the time being).

Limits the number of redirects we have to do to from http to https.

No.
## Screenshots

N/A
## Request for comment

@desbo @johnduffell @TBonnin


common/app/common/LinkTo.scala
```diff
@@ -20,19 +20,17 @@ trait LinkTo extends Logging {
 
   lazy val host = Configuration.site.host
 
-  private val ProdURL = "^http://www.theguardian.com/.*$".r
-  private val GuardianUrl = "^(http://www.theguardian.com)?(/.*)?$".r
+  private val ProdURL = "^http[s]?://www.theguardian.com/.*$".r
+  private val GuardianUrl = "^(http[s]?://www.theguardian.com)?(/.*)?$".r
```
johnduffell: actually (s)? should be enough, square is for a set of characters wheras () is for grouping

common/app/common/LinkTo.scala
```diff
@@ -20,19 +20,17 @@ trait LinkTo extends Logging {
...
   private val RssPath = "^(/.+)?(/rss)".r
   private val TagPattern = """^([\w\d-]+)/([\w\d-]+)$""".r
+  private val InteractiveUrl = """(/(ng-)?interactive/.*)$""".r
 
   /**
     * email is here to allow secure POSTs from the footer signup form
```
johnduffell: 😛  not any more!

common/app/common/LinkTo.scala
desbo: without a group the question mark just covers the preceding character, so even `https?` would do it :)

applications/app/controllers/OptInController.scala
```diff
@@ -6,16 +6,28 @@ import play.api.mvc.{Action, Controller, Cookie, DiscardingCookie}
...
+    case "in" => optIn()
+    case _ => optOut()
   }
+  def optIn() = SeeOther("/").withCookies(Cookie(cookieName, "true", maxAge = Some(lifetime)))
+  def optOut() = SeeOther("/").discardingCookies(DiscardingCookie(cookieName))
+}
+
+case class OptInFeature(cookieName: String, lifetime: Int = 90.days.toSeconds.toInt) extends OptInOutFeature
+
+class HttpsOptOutFeature extends OptInOutFeature {
```
desbo: might be better as an `object`

common/app/common/LinkTo.scala
johnduffell: true, but you have to watch a little for readability...  or use parser combinators for extra clarity ;)

johnduffell:
looks good to me! this is a great place to be 👍

applications/app/controllers/OptInController.scala
```diff
@@ -6,16 +6,28 @@ import play.api.mvc.{Action, Controller, Cookie, DiscardingCookie}
...
-object OptInController extends Controller {
-  case class OptInFeature(cookieName: String, lifetime: Int = 90.days.toSeconds.toInt) {
-    def opt(choice: String) = choice match {
-      case "in" => optIn()
-      case _ => optOut()
-    }
-
-    def optIn() = SeeOther("/").withCookies(Cookie(cookieName, "true", maxAge = Some(lifetime)))
-    def optOut() = SeeOther("/").discardingCookies(DiscardingCookie(cookieName))
+trait OptInOutFeature extends Controller {
```
johnduffell: not 100% sure it was the right thing to add a trait like that and subclasses
the code might have remaned simpler if we just treated it as opting in to http only, from a coding point of view?

common/app/common/LinkTo.scala
JustinPinner: Group with () screws up the match such that the expected value that is checked later needs additional work to cope with its absence. I could drop the square brackets completely but I think isolating the 's' makes it clearer what we are expecting.

applications/app/controllers/OptInController.scala
JustinPinner: Hmm. You're right but I was aiming to keep the semantics clear - you're opting _out_ of https, rather than opting _in_ to http. Could change it though if we feel it makes more sense.

johnduffell: depends how you look at it! after all https is default now, so maybe you're opting into insecure content?  either way I feel it's more a naming/terminology problem than a problem needing more complexity in the code, but I appreciate there's urls to consider too.

JustinPinner:
Sorry @johnduffell @desbo I've re-done the https opt in/out behaviour. 

We now have a three-state option for https: 
`/opt/in/HTTPS` = ON for everything 
`/opt/out/HTTPS` = OFF (http) for everything
`/opt/delete/HTTPS` = remove the cookie (https for everything except politics and interactives) 

What to you think?

applications/app/controllers/OptInController.scala
```diff
@@ -26,7 +36,7 @@ object OptInController extends Controller {
     }))
   }
 
-  val HTTPS = OptInFeature("https_opt_in")
+  val HTTPS = HttpsOptFeature("https_opt_in")
```
TBonnin: nitpicking here 😁 : why not Camel case here?

TBonnin:
👍

johnduffell:
sounds good to me! I am also a bit weirded out by the 3 different naming conventions on the vals, but it's not a disaster 👍

applications/app/controllers/OptInController.scala
JustinPinner: Fair comment. I ignored the casing of the variable name from the original I think as it made sense when I read it, being an abbreviation and all. I'll change it when I re-merge master in a few minutes :)

desbo:
looks good, nice one! 👍

PR #13430, opened 2016-06-28
Remove thrasher test

blongden73:
Removes all mention of video thrasher test from ABswitches, ab.js and the video-thrasher-test file. 

@NataliaLKB - could you just have a look?
@jamesgorrie - I think this is all that was added for this test?


NataliaLKB:
👍

akash1810:
was just about to raise a PR for this! 👍

PR #13431, opened 2016-06-28
kill controls on expired videos

jamesgorrie:
## What does this change?

Kills the control bar so you can't play the video.
As before, we know this isn't the real way to expire videos. Roll on atoms.

People can't play video

No
## Screenshots

👾 
## Request for comment

@akash1810


akash1810:
👍

PR #13432, opened 2016-06-29
Show series containers on paid content pages

kelvin-chappell:
We're already showing story package containers.

This shows content from same series as well (behind a switch just in case):
image


common/app/views/fragments/onwardPlaceholder.scala.html
```diff
@@ -1,5 +1,7 @@
-@()(implicit request: RequestHeader)
-<aside id="onward" class="onward js-onward facia-container facia-container--layout-content"
-role="complementary"></aside>
-<aside id="more-in-section" class="more-in-section js-more-in-section facia-container facia-container--layout-content"
-role="complementary"></aside>
+@(isAdvertisementFeature: Boolean)(implicit request: RequestHeader)
+<aside id="onward"
+  class="onward js-onward facia-container facia-container--layout-content facia-container--layout-content @if(isAdvertisementFeature){ paid-content--advertisement-feature}"
```
JonNorman: Looks like `facia-container--layout-content` is duplicated here.

common/app/views/support/ContentFooterContainersLayout.scala
```diff
@@ -23,7 +23,8 @@ object ContentFooterContainersLayout {
     val htmlBlocks = if (isAdvertisementFeature) {
 
       // majority of footer components we don't want to appear on advertisement feature articles
-      Seq(storyPackagePlaceholder)
+      if (showPaidSeriesContainer.isSwitchedOn) Seq(storyPackagePlaceholder, onwardPlaceholder)
+      else Seq(storyPackagePlaceholder)
```
JonNorman: Could reduce duplication of `Seq` initialisation:
```
Seq(storyPackagePlaceholder)  :+
if (shouldPaidSeriesContainer.isSwitchedOn) onwardPlaceholder else None
```
Subjective as to whether that is more readable though...

JonNorman:
Very minor comments, otherwise looks good :+1:

common/app/views/fragments/onwardPlaceholder.scala.html
kelvin-chappell: Oops! Thanks.

common/app/views/support/ContentFooterContainersLayout.scala
kelvin-chappell: Wouldn't that give you a `Seq` of `Option[Html]`?

JonNorman: Hmm, yes I think you're right. If we used the `++` operator instead then it would auto serialise the `None` to `Seq.empty`, but not sure what would happen to the HTML object.

Ignore me!

PR #13437, opened 2016-06-29
Stabilise preview

alexduf:
## What does this change?

This forces the lifecycles to start only when the application has been started by play.
It should fix the deployment instability we observed on preview and should allow me to carry on the migration.
## Request for comment

@johnduffell @TBonnin


johnduffell:
fair enough, good debugging, good commenting too 👍

PR #13438, opened 2016-06-29
Remove Twitter overlay on paid content

katebee:
## What does this change?

Just like #13411, but for Twitter! 🎉 

The branded overlay will no longer appear on images for paid content when articles are tweeted.

Swapping to using one of the other image profiles for the `twitter:image` meta-tag on paid content.

Success will be if we exclude the banner from 'paid for' content when it is tweeted (see image below).

Twitter
Anywhere else that uses Twitter Card meta-data (none known at this date)
Information on Twitter Cards: https://dev.twitter.com/cards/overview
## Screenshots

Paid content currently has the overlay appearing:
paid content

### DIFF
--- a/common/app/model/content.scala
+++ b/common/app/model/content.scala
@@ -8,7 +8,7 @@ import com.gu.facia.client.models.TrailMetaData
 import common._
 import common.dfp.DfpAgent
 import conf.Configuration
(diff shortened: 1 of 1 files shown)


johnduffell:
great! 👍

PR #13439, opened 2016-06-29
Further restrict when Plista should be served

JonNorman:
## What does this change?

We have a switch that determines whether we serve Outbrain or Plista at the bottom of pages. Up until now, if the switch is enabled and we are on the AU edition, then Plista will be served, otherwise Outbrain.

At the request of the guys in AU, we are changing the policy slightly to the following:
- mobile/tablet: all sections have outbrain 100% if the time
- desktop: world news has outbrain 100% of the time
- everything else controlled by the switch (including requirement to be on the AU edition)

I _think_ I'm right in thinking that we no longer support IE8, and thus I'm fine to use `Array.prototype.indexOf`...

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/third-party-tags.js
+++ b/static/src/javascripts/projects/common/modules/commercial/third-party-tags.js
@@ -48,7 +48,11 @@ define([
             $(documentAnchorClass).append(externalTpl({widgetType: widgetType}));
         }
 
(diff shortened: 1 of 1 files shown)


regiskuckaertz:
> I think I'm right in thinking that we no longer support IE8, and thus I'm fine to use Array.prototype.indexOf...

That is correct for everything in commercial. Although FYI, `es5-shim` shims these functions for IE8 and below.

regiskuckaertz:
Very good 👍

PR #13440, opened 2016-06-29
Remove the quick survey

katebee:
## What does this change?

This removes the survey added in #13337. The survey has been run and the responses have been recorded.

Leaving private-browsing.js (which was also created in #13337) for now; it may be useful as a util.

Removes some dead code 💀 sitting in the AB tests.
Pre-empts the expiring switch warnings.

No.


dominickendrick:
👍

PR #13442, opened 2016-06-30
Dev build controllers

alexduf:
## What does this change?

Another boring commit where I transform more controller into classes.
Coupled with a simplification of our GuardianConfiguration object.

Both these commits are on my way to remove the Global object of dev-build
## Request for comment

@TBonnin @johnduffell


johnduffell:
😴 👍

PR #13443, opened 2016-06-30
Fix headline colour for dark containers on immersive articles

jfsoul:
## What does this change?

Immersive articles with dark headline containers now will not have their headline colour overridden by the the tone of the article.
## Request for comment

@NataliaLKB @dominickendrick

### DIFF
--- a/static/src/stylesheets/module/content/_article-immersive.scss
+++ b/static/src/stylesheets/module/content/_article-immersive.scss
@@ -112,10 +112,16 @@
     font-weight: 200;
 }
 
+.content--immersive-article-without-main-media,
+.immersive-main-media__headline-container--dark {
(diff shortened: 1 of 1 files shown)


static/src/stylesheets/module/content/_article-immersive.scss
```diff
@@ -112,10 +112,16 @@
     font-weight: 200;
 }
 
+.content--immersive-article-without-main-media,
```
NataliaLKB: Don't need this line, but otherwise good!

NataliaLKB:
One comment, but lgtm 👍

jfsoul:
Thanks, amended as per suggestion.  Have also run through the various immersive types with @NataliaLKB, which should be overridden by more specific styling.

PR #13444, opened 2016-06-30
Dev build lifecycle / Autorefresh

alexduf:
## What does this change?

These autorefresh need to be instantiated to create the app, and they were depending on a started app. This PR passes an already instantiated actor system so that the akka.system() doesn't throw an exception on startup.
It also injects more lifecycle objects for the same reason.

next PR is about removing Global on dev-build, this is a way to make the next PR less overwhelming.
## Request for comment

@TBonnin @johnduffell


TBonnin:
👍

PR #13445, opened 2016-06-30
Fix Plista switch logic

JonNorman:
Don't need to call `.name` on the return of `getBreakpoint` is it already returned in the method.

@regiskuckaertz


regiskuckaertz:
👍

PR #13446, opened 2016-06-30
Labour and Liverpool immersive article

SiAdcock:
## What does this change?

Adds the Labour & Liverpool series template. This can be applied to an article by selecting the "Labour & Liverpool (membership series)" tag in Composer.

This template forms part of a twelve week Membership Explore test to find new ways to increase membership sign-up. We are monitoring traffic and other key measures on a weekly basis, including number of page views, median attentions time, number of browsers to visit each page of the series, percentage of page views from regular readers and the size of the mailing list accompanying this series.

Data for this test is being collected in [this spreadsheet](https://docs.google.com/spreadsheets/d/1SLFQMQksBP58kfMK5UxlCXyxNSNL_APXYbHeLAo7D84/edit#gid=673848165)

How we measure success is described in [this document](https://docs.google.com/document/d/1nZco0YbRZNbDE83GO-3wWJAGuaE_B8UfPvW_o4EaKEY/edit)

No
### Mobile

picture 37
### Desktop

picture 36
## Request for comment

@NataliaLKB


sndrs:
before getting too far into this, I don't feel at all good about referring to specify tags in content templates...

common/app/views/fragments/immersiveMainMedia.scala.html
```diff
@@ -8,11 +8,13 @@
 @defining((
     page.item.tags.isUSMinuteSeries,
     page.item.elements.hasMainEmbed,
-    page.item.fields.main.nonEmpty
-)) { case (isMinuteSeries, hasEmbed, hasMainMedia) =>
+    page.item.fields.main.nonEmpty,
+    page.item.tags.isLabourLiverpoolSeries
```
NataliaLKB: As discussed in person, this maybe should be its own template, not added to this one.

rich-nguyen:
We've never really clarified this, but as a rule, bespoke content work is typically short-term, maybe a visual tweak for the olympics, or some minor change for a general election, that kind of thing. Most of our tools have been developed to absorb most of these requests (badges now come from capi, bespoke containers can be embedded into fronts as 'thrashers'). The general aim being to minimise bespoke templating/styling maintenance in this codebase.

Whilst we do try to foster innovation through our templates (thanks @sammorrisdesign), there's a general "value added per template" that we all have some consideration for. Unfortunately, I don't think a single tag meets that threshold (as a counterexample, a content type or tone would pass that threshold). Happy to chat more in person.

SiAdcock:
Thanks for everyone's input, ideas and clarifications. I've made a number of changes to my original approach as a result.

1. I've extracted the main markup changes out of the Immersive Template and into a new type of template. I've called this the Heroic Template because it replaces the Guardian header and main navigation with a large hero banner. I didn't want to couple the name of the template type to this particular series of articles. Naming is hard. I'm happy to take suggestions for alternatives.
2. For styles and markup that is specific to the "Labour & Liverpool" series, I've created fragments and a stylesheet.
3. I've added a feature switch for the Heroic Template that is disabled by default. This is used to set a property on the Content model, which determines whether the Heroic Template should be applied to articles tagged with `membership/series/labour-liverpool`
4. The feature switch has a sell-by date of 29th September, the day after the Labour Party Conference. After determining whether this series has achieved its goals, we can decide whether to rollback this change or indeed whether we can find a more permanent home for this template. We could also come up with a new process for building templates that will allow us to develop more custom templates without putting maintenance pressure on `frontend`

Please let me know if you have any further suggestions

@sndrs @rich-nguyen @NataliaLKB

common/app/views/main.scala.html
```diff
@@ -68,12 +68,25 @@
                 </div>
             }
             case page: model.ContentPage if (page.item.content.isImmersive && !page.item.content.tags.isInteractive) => {
-                <div class="@if(page.item.fields.main.nonEmpty) { immersive-header-container }">
+                <div class="@RenderClasses(Map(
+                    ("immersive-header-container", page.item.fields.main.nonEmpty),
+                    ("immersive-header-container--labour-liverpool", page.item.content.tags.isLabourLiverpoolSeries)))">
```
SiAdcock: Whoops, missed a Labour & Liverpool reference

sammorrisdesign:
That new process for building new templates without adding more to frontend is interactives with html on the page at load... But this is probably isn't the place to do that.

Isn't this template specifically for Membership? If so, shouldn't we have a `tone/membership` tag and use this template when that tag is present?

Personally, I'm not entirely convinced we want to go down the standard < immersive < heroic < super-heroic < ultra-heroic < immersively super heroic path 😄

NataliaLKB:
I agree with @rich-nguyen and to further it I think we should come up with more concrete guidance to adding a new template for the future. So sorry that all this discussion has arisen from your PR @SiAdcock. It may be worth moving some of the "why" off this PR? I believe we have a meeting on Monday that might be good for coming up with a good plan for this and future features.

@sammorrisdesign I believe @blongden73 was building these as interactive pages but that isn't sustainable for him as this test will go on for 12 weeks - you are right that it isn't the place to do that.

@SiAdcock it would be great if you could answer some of these in your PR description:
What tests have been done on this? Has there been any UX testing or otherwise? What have the results been? Why is this worth building? Basically, since this is probably not going to be an a/b test, what other tests have been done? I seem to remember Mario suggesting you guys had data already, and it would be good to know what that is 😄

static/src/stylesheets/module/content/_article-labour-liverpool.scss
```diff
@@ -0,0 +1,138 @@
...
+        }
+
+        .content__standfirst--labour-liverpool-article {
+            @include fs-headline(3);
+            color: $labour-liverpool-standfirst-font-colour;
+            margin-top: 10px;
+
+            p::before {
+                float: left;
+                content: url('https://s3.amazonaws.com/gdn-cdn/membership-article-two/img/quote_marks.svg');
```
NataliaLKB: Should this be an inline-svg, like how we do it with comment pieces?

static/src/stylesheets/module/content/_article-labour-liverpool.scss
```diff
@@ -0,0 +1,138 @@
...
+        }
+    }
+}
+
+.content--labour-liverpool-article {
+    .heroic-main-media__headline-container {
+        position: absolute;
+        top: 0;
+        width: 100%;
+        background-image: url('https://s3.amazonaws.com/gdn-cdn/membership-article-two/img/membership-background-two-simple.png');
```
NataliaLKB: Is this an icon? Should this be an inline-svg?

sammorrisdesign:
@NataliaLKB But there are ways to "template" interactives by using spreadsheets. The Visuals team do it all the time rather than build from scratch every time. The reason we don't use that tool too much for testing templates like these, hypothetically before they get added to frontend, is because of the horrible loading flash and drop off on mobile (because of said loading).

NataliaLKB:
@sammorrisdesign that makes sense! Thanks for that - it is good to know!

ScottPainterGNM:
Sorry to add another thing but looking at the stand first font colour on a white background it doesn't meet colour contrast accessibility requirements. I'm also not sure about the dark shade of Ewan MacAskill against that dark membership header background either.

common/app/conf/switches/FeatureSwitches.scala
```diff
@@ -10,6 +10,16 @@ trait FeatureSwitches {
     "If this switch is on, main media embeds won't be iframed",
     owners = Seq(Owner.withGithub("sammorrisdesign")),
     safeState = Off,
+    sellByDate = new LocalDate(2016, 9, 29),
```
NataliaLKB: Did you mean to extend @sammorrisdesign 's switch?

common/app/conf/switches/FeatureSwitches.scala
```diff
@@ -10,6 +10,16 @@ trait FeatureSwitches {
...
+    exposeClientSide = false
+  )
+
+  val HeroicTemplateSwitch = Switch(
+    SwitchGroup.Feature,
+    "heroic-main-media",
+    "If this switch is on, Heroic template will be applied to heroic articles",
```
NataliaLKB: Is it worth mentioning in the description that this is a membership test?

common/app/views/fragments/labourLiverpoolStandfirst.scala.html
```diff
@@ -0,0 +1,25 @@
...
+            val trailText = views.support.StripHtmlTags(t)
+            // description has to end in a full stop in the metadata to match google linked data standards, but
+            // editorial style doesn't allow it in trail text.
+            s"$trailText${if (item.tags.isReview) "." else ""}"
+        }){ trailText =>
+            <meta itemprop="description" content="@trailText" />
+        }
+    }
+    
+    <img class="membership-cutout" src="https://i.guim.co.uk/img/static/sys-images/Guardian/Pix/contributor/2015/8/18/1439913873894/Ewen-MacAskill-R.png?w=300&q=55&auto=format&usm=12&fit=max&s=2c8dc813795a8c11dd98e28377a5893f">
```
NataliaLKB: Is this source in CAPI?

common/app/conf/switches/FeatureSwitches.scala
SiAdcock: Good catch 😅

common/app/conf/switches/FeatureSwitches.scala
SiAdcock: Good call!

static/src/inline-svgs/membership/email.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="32" height="32" viewBox="-2 -2 32 32"><path fill="#fff" d="M20.2 11.7l-.4-.4-5.7 2.7-5.7-2.7-.3.4 6 4.9 6.1-4.9zM21.4 7H6.5L5 9.5v9.3L6.6 20h14.9l1.5-1.2V9.5L21.4 7zm-.5 10.8H7.1V9.2h13.8v8.6z"/></svg>
```
NataliaLKB: Thanks for making this inline svgs! One thing, `xmlns="http://www.w3.org/2000/svg"` will break IE8: https://github.com/guardian/frontend/blob/master/docs/inline-svgs.md#image-preparation.

Mind removing?

common/app/conf/switches/FeatureSwitches.scala
sammorrisdesign: Go for it though ;)

SiAdcock:
Hi all, thanks for everyone's help with this PR. Assuming the build passes, would anyone be able to give us another code review? 

@NataliaLKB @sndrs

static/src/stylesheets/head.content.scss
```diff
@@ -24,6 +24,7 @@
 @import 'module/_chapters';
 @import 'module/content/_article-immersive';
 @import 'module/content/_article-minute';
+@import 'module/content/_article-labour-liverpool';
```
sndrs: this feels like the CSS should be inlined like the story package css on facia - i.e. not included in general content head css, which is really too big already...

SiAdcock: Good shout. Would this involve adding something to this object:

https://github.com/guardian/frontend/blob/master/common/app/assets/assets.scala#L51

sndrs: yeah that's right. maybe a public method that exposes the inline method? we don't really want to add a new variable for each new design i don't think? @johnduffell might have some good advice on this

johnduffell:
the inline stylesheet commit looks good to me!

SiAdcock:
@sndrs 

With @johnduffell's guidance, I copied the pattern for the story-package stylesheet and defined a new public method that inlines the Labour Liverpool stylesheet. We decided to go with the Rule of Three. If we need to do this again, we'll implement a more generic solution.

common/app/conf/switches/FeatureSwitches.scala
```diff
@@ -14,6 +14,16 @@ trait FeatureSwitches {
     exposeClientSide = false
   )
 
+  val HeroicTemplateSwitch = Switch(
```
sndrs: maybe a comment in here referring to this PR for anyone removing the switch. there's a quite a bit of bespoke stuff going in, which if left behind and if you don't what it is, you wouldn't want to remove later

NataliaLKB:
This looks much better. I think this is fine to go out for the test 👍 .

Please keep the web team up to date with how the test goes and if new templates will be needed. If that is the case we likely need a different way of adding new templates to make the process easier/quicker/cleaner. Something to think about anyway 😄

NataliaLKB:
Nit: Can you update the description of this PR to explain that this is only a 12 week test? There is lots of useful insights that we have gotten from you and the team in person, but not a lot of it is in this PR. Not a blocker though, it would just be good to have it for hindsight.

static/src/stylesheets/inline/article-labour-liverpool.scss
```diff
@@ -0,0 +1,261 @@
...
+
+.series-identity__title, .series-identity__byline {
+    @include fs-header(2);
+    line-height: 19px;
+    color: #ffffff;
+}
+
+.content--labour-liverpool-article__headline-background {
+    position: absolute;
+    //TODO: Need a designer to fix the SVG so that we don't need to position it in the CSS
```
NataliaLKB: Not a blocker, but make sure you talk to @blongden73 or another designer to fix these svgs!

SiAdcock: Sure thing, we're looking at it this afternoon 😄

SiAdcock:
Thanks for all your help @NataliaLKB. I've updated the PR description.

sndrs:
👍 nice one, cheers for being so patient!

SiAdcock:
Thanks for taking the time to help me get this up to scratch, @sndrs 😄 

I'll merge this morning

PR #13447, opened 2016-06-30
migrate dev-build away from globals

alexduf:
## What does this change?

It looks big but half of it is repetitive.

This is removing the `Global` object from dev-build. It now uses an application loader.
## Request for comment

@TBonnin @johnduffell


TBonnin:
lgtm 👍

PR #13448, opened 2016-06-30
Simple survey page

janua:
## What does this change?

This adds initial pages used to gauge interest in features we may build. Initially this just adds 4 simple pages that display:
- a landing page
- a page asking if the user would take a quick survey
- a survey page
- a thank you page.

The landing page and the survey page use forms built using `formstack`.

This will help us gauge interest in whether the feature is worth building. We will measure success by the amount of feedback we receive from users.

No
## Screenshots

surveypageone

surveypagetwo
## Request for comment

@jamespamplin


static/src/stylesheets/module/survey/_inline-email-form.scss
```diff
@@ -0,0 +1,86 @@
...
+            $guardian-brand,
+            #ffffff,
+            $guardian-brand
+        );
+        display: inline-block;
+        vertical-align: top;
+        width: auto;
+        font-weight: bold;
+        text-decoration: none;
+        border-radius: 1000px;
```
jamespamplin: You could probably use `100%` here

static/src/stylesheets/module/survey/_survey-page.scss
```diff
@@ -0,0 +1,71 @@
...
+    .fsError {
+        color: $error;
+        background-color: unset;
+        border: 0;
+        text-align: left;
+        padding-left: 0;
+    }
+
+    .fsSubmitButton {
+        @include fs-textSans(2);
```
jamespamplin: Can we possibly avoid the repetition of some of these class, and also can we remove the `!important` rules if possible?

static/src/stylesheets/module/survey/_survey-page.scss
```diff
@@ -0,0 +1,71 @@
...
+        color: $guardian-brand;
+        margin-bottom: $gs-baseline/2;
+    }
+
+    p {
+        @include fs-bodyCopy(2, size-only = true);
+        margin-bottom: $gs-baseline;
+    }
+
+    .fsError:before {
```
jamespamplin: Should use `::before` and `::after` for pseudo-element classes

applications/app/views/survey/emailDigest404.scala.html
```diff
@@ -0,0 +1,45 @@
...
+            <div class="wrapper">
+
+                <div class="wrapper-inner">
+
+                    <h1>Thank you for your interest in personalised email</h1>
+
+                    <p>This feature is not yet available as we are still testing demand for it. Tell us your email address and we'll let you know when it is available (we won't use it for any other purposes).</p>
+
+                    <div class="inline-email-form">
+                        <script type="text/javascript" src="https://@{formStackAccountName}.formstack.com/forms/js.php/email_digest_email?no_style_strict=1"></script><noscript><a href="https://guardiannewsampampmedia.formstack.com/forms/email_digest_email" title="Online Form">Online Form - Email digest email</a></noscript>
```
jamespamplin: `guardiannewsampampmedia` is also hardcoded on this line

applications/app/views/survey/emailDigest404.scala.html
```diff
@@ -0,0 +1,45 @@
+@(surveyPage: Page)(implicit request: RequestHeader)
+
+@import model.Page
+@import conf.Configuration.Survey.formStackAccountName
```
jamespamplin: Not a blocker, but is it possible to include the `formStackAccountName` and some of the other form stack related endpoints as part of the model, rather than injecting the config directly into the view, and concatenating strings in the view. May require creating another Model type for `SurveyPage`.

eg: `https://@{formStackAccountName}.formstack.com/forms/js.php/@formName`, will be slightly cleaner if provided by the model

applications/app/views/survey/thankyou.scala.html
```diff
@@ -0,0 +1,21 @@
+@(surveyPage: Page)(implicit request: RequestHeader)
```
jamespamplin: Is / Can we `noindex` the thankyou pages too?

applications/app/views/survey/emailDigest404.scala.html
janua: The idea was to pass an arbitrary string from the endpoint to the controller to the view, meaning we could view any form. Could do this, but then it requires a release for new forms? At least then we could see what forms are available.

applications/app/views/survey/thankyou.scala.html
janua: All the survey pages have `<meta name="robots" content="noindex">` because of `shouldGoogleIndex` on the `Survey` page being set to `false`.

static/src/stylesheets/module/survey/_survey-page.scss
janua: We can't remove the `!important` rules as lots of rules come from form stack and I couldn't see an easy way to override them other than `important`.

janua: I'm not sure what repetition you are talking about!

jamespamplin: Spoke with @janua - no need to worry about the `!important`s if they cannot be reduced - these are for form stack overrides, so should be isolated enough.

applications/app/views/survey/emailDigest404.scala.html
jamespamplin: Spoke with @janua - this isn't necessary for now, we can revise how we do these things later when we potentially separate it into another app

static/src/stylesheets/module/survey/_inline-email-form.scss
janua: Tried this, doesn't get the same effect.

PR #13449, opened 2016-06-30
Fix edition switch when https-enabled

JustinPinner:
## What does this change?

We should only compare hosts when assessing valid redirect targets in edition switch.

Fixes edition switching.

No.
## Request for comment

@alexduf @johnduffell


alexduf:
:+1:

johnduffell:
good one! 👍

PR #13451, opened 2016-06-30
filter on item not being a video on carousel

jamesgorrie:
## What does this change?

As we rely on the index for the JS to work, best we filter out the junk first.

The carousel works again.

No
## Screenshots

🍇 
## Request for comment

@akash1810


jamesgorrie:
[Better looking PR](https://github.com/guardian/frontend/pull/13451/files?w=1)

TBonnin:
👍

akash1810:
👍

PR #13456, opened 2016-07-01
Hosted video fails fix

Calanthe:
## What does this change?

A fix for the hosted videos fails issues. The videos' URLs link directly to https://cdn.theguardian.tv now instead of http://multimedia.guardianapis.com which is much faster.
Also all links are https now.

I didn't change the `M3u8` link because it's only a playlist text file and it's downloaded not opened by the browser.

No video errors on the hosted pages. Also videos load now much faster.


akash1810:
> I didn't change the M3u8 link because it's only a playlist text file and it's downloaded not opened by the browser.

The m3u8 encoding is used in Safari/iOS (m3u8 is the HLS flavour of the video). I'd convert that too then 👍

kelvin-chappell:
👍

Calanthe:
Thanks @akash1810 It's good to know that. I just changed the links for the m3u8 as well.

PR #13457, opened 2016-07-01
Split DFP API into small modules

regiskuckaertz:
Last week, the guys from @guardian/membership-and-subscriptions encountered a problem with our `dfp-api` module. @jamespamplin discovered the cause was a circular reference between `build-page-targeting`, `ab` and `dfp-api`.

This is one symptom among many where the sheer size of the `dfp-api` module makes it hard to reason about and debug. @jbreckmckye kickstarted the process of splitting it into smaller modules, which I will complete in this PR. In essence:
- DFP-related modules share an environment where we store the state of our ad stack. This is the `dfp-env` module.
- All private variables and functions have been split into modules stored in a `private` folder
- All public functions add themselves to the global `dfp` public object

There's still some work to do. I'm not very happy about the `create-advert` module and the `dfpEnv.fn` map. Also, I broke the tests and need to fix them... mostly it's just changing the paths of the modules. I just want to raise the PR now because ~~it's Fridayyyyyyyy~~ I want other people's feedback.

cc @rich-nguyen @uplne @Calanthe


static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js
```diff
@@ -2,701 +2,50 @@
...
+     *      data-desktop="300,250"
+     *      data-refresh="false"
+     *      data-label="false">
      * </div>
      *
      * You can set the set which size ad(s) should be loaded at which breakpoint by using the
-     * data attribute. The available breakpoints and their sizes are listed in the config below.
-     * You do not need to specify all of these. If you set a mobile size, then that size will be used
-     * for all ads in that slot until another breakpoint is detected, in the above case, that's desktop.
+     * data-[breakpoint] attributes (see detect). It works like a min-width media query.
```
JonNorman: minor comment miswording: "You can set the set"

static/src/javascripts/projects/common/modules/commercial/dfp/get-adverts.js
```diff
@@ -0,0 +1,18 @@
+define([
+    'common/modules/commercial/dfp/private/dfp-env',
+    'common/modules/commercial/dfp/private/get-advert-by-id',
+    'common/modules/commercial/dfp/dfp-obj'
+], function (dfpEnv, getAdvertById, dfp) {
+    dfp.getAdverts = getAdverts;
+    return getAdverts;
+
+    function getAdverts(isWithAllAds) {
```
JonNorman: What does `isWithAllAds` actually refer to here? What does it mean for an ad to be "with all ads"?

regiskuckaertz: Good point. `adverts` may contain ads that are empty, i.e. when DFP returned no creative. They used to be removed from the array, but instead I keep them so we can track the lifecycle of all ad slots, no matter what. Indeed, the parameter name is misleading.

static/src/javascripts/projects/common/modules/commercial/dfp/private/dfp-env.js
```diff
@@ -0,0 +1,45 @@
+define([
+    'common/utils/config'
+], function (config) {
+    var dfpEnv = {
+        /* renderStartTime: integer. Point in time when DFP kicks in */
+        renderStartTime: -1,
+
+        /* displayed: boolean. Set to true when the first advert is getting displayed (via googletag.display) */
+        displayed: false,
```
JonNorman: Could this just be `firstAdDisplayed`?

static/src/javascripts/projects/common/modules/commercial/dfp/add-slot.js
```diff
@@ -0,0 +1,37 @@
...
+    'common/modules/commercial/dfp/private/queue-advert',
+    'common/modules/commercial/dfp/private/lazy-load',
+    'common/modules/commercial/dfp/dfp-obj',
+    'common/utils/mediator'
+], function (dfpEnv, createAdvert, queueAdvert, lazyLoad, dfp, mediator) {
+    dfp.addSlot = addSlot;
+    return addSlot;
+
+    function addSlot(adSlot) {
+        adSlot = adSlot instanceof HTMLElement ? adSlot : adSlot[0];
```
JonNorman: Is this just differentiating between HTML and bonzo objects?

static/src/javascripts/projects/common/modules/commercial/dfp/dfp-obj.js
```diff
@@ -0,0 +1,4 @@
+define(function () {
```
rich-nguyen: I'm guessing this deliberately has no dependencies, but I'm not sure why? The reset function in `dfp-api`, and the decomposed function modules all extend `dfp-obj`, why is this not done here? Is is possible for a client to require `dfp-obj` and attempt to invoke a function on an empty object?

static/src/javascripts/projects/common/modules/commercial/dfp/add-slot.js
regiskuckaertz: Yes

static/src/javascripts/projects/common/modules/commercial/dfp/private/lazy-load.js
```diff
@@ -0,0 +1,67 @@
...
+                        var rect = advert.node.getBoundingClientRect();
+                        // load the ad only if it's setting within an acceptable range
+                        return (1 - depthOfScreen) * viewportHeight < rect.bottom && rect.top < viewportHeight * depthOfScreen;
+                    })
+                    .forEach(loadAdvert);
+            });
+        }
+    }, nbOfFrames * durationOfFrame);
+
+    dfpEnv.fn.shouldLazyLoad = shouldLazyLoad;
```
rich-nguyen: ...

static/src/javascripts/projects/common/modules/commercial/dfp/dfp-obj.js
regiskuckaertz: I'm not sure I see what your point is here, but I could move `dfp-obj` in `private` if it's clearer that way. All public modules will add themselves to the `dfp` map, but the `dfp` map itself doesn't have any dependency.

static/src/javascripts/projects/common/modules/commercial/dfp/private/lazy-load.js
```diff
@@ -0,0 +1,67 @@
...
+    dfpEnv.fn.enableLazyLoad = enableLazyLoad;
+    dfpEnv.fn.disableLazyLoad = disableLazyLoad;
+    return dfpEnv;
```
rich-nguyen: it's a bit odd that we return the module object of another module, modified with additional functions! maybe we can get around this. Could we return a module of functions from `lazy-load`?

rich-nguyen: ... things that need `shouldLazyLoad` should then require the `lazy-load` module, rather than `dfp-env`, think that makes sense to new readers?

regiskuckaertz: YES! I didn't like that part at all, I can just add this functions to the `lazyLoad` function itself. Thanks!

static/src/javascripts/projects/common/modules/commercial/dfp/dfp-obj.js
rich-nguyen: I think it's harder to understand. So if I want to use the `dfp` map, it's not clear what the module definition is, because the map assignements are spread across a number of modules. This makes it hard to read. Also, there is no rjs compile-time guarantees that the `dfp` is actually populated when a client uses it. That's not clear from the code structure either.

rich-nguyen: So my original question asks why can't we do something like: 
```
// this is dfp-obj.js

define([ 'track-ad-load.js' ],
function(trackAdLoad){

}
 )

```

regiskuckaertz: That's a fair point, populating the `dfp` map should be the responsibility of the `dfp-api` module, not each submodule. In reality, the `dfp` map is completely useless. I can make without it. Let's see.

rich-nguyen:
I like the structure of this. I think if we can get all functions out of `dfp-env`, and some rationale behind the purpose of `dfp-obj`, which seems will be redundant in the near future (ie. when all calls are made directly to the modules, rather than by alias), it'll all be great 👍

regiskuckaertz:
@rich-nguyen thanks for your feedback 😄  I removed both.

PR #11254, opened 2015-12-02
Fronts on Articles 

holder-josh:
The initial AB test went well, so this is a V2 of that. Much more of a 'production ready' test that obeys the outbrain rules and injects a mixture of section and network fronts.

Set up as a 0% test for now.

To do in another pull request:
- Sort out the 'Jump to comments' button's behaviour (currently jumps a lot due to AJAXing components in)
- Sort tracking

PR #13461, opened 2016-07-04
Removing fronts on articles test

NataliaLKB:
## What does this change?

Removing the fronts on Article test. It was first created in December here: https://github.com/guardian/frontend/pull/11254.

When this was tested it was actually proven successful, but it conflicted with our agreement with Outbrain and was never able to make it to production. Since it has been about 6 months I think it is time to remove it.

Nope!
## Screenshots

Everything should look the exact same
## Request for comment

@Jholder112233 @johnduffell @stephanfowler


static/src/javascripts/bootstraps/enhanced/trail.js
```diff
@@ -37,14 +35,8 @@ define([
     Popular,
     Related,
     TonalComponent,
-    FrontsContainers,
-    shareCount,
-    ab
+    shareCount
 ) {
-    // messy but removed within a week
```
johnduffell: 🎉 I always thought that comment could do with updating 😆

johnduffell:
great thanks for that natalia! 👍

PR #13463, opened 2016-07-04
refactor wiring

alexduf:
## What does this change?

This refactors the changes I've applied recently so that application share their wiring. It eases the maintenance of "aggregated" applications such as dev-build or standalone.

cleaner code base

no
## Request for comment

@TBonnin @johnduffell


johnduffell:
👍

PR #13469, opened 2016-07-04
move commercial away from Global and introduce app loader

alexduf:
## What does this change?

Remove the Global object, introduces DI with an application loader
## Request for comment

@TBonnin @johnduffell


johnduffell:
👍

PR #13470, opened 2016-07-04
move diagnostics away from Global and introduce app loader

alexduf:
## What does this change?

Remove the Global object, introduces DI with an application loader
## Request for comment

@TBonnin @johnduffell


johnduffell:
you can have a prize for most PRs in a day! 👍

alexduf:
Ahah, I may be able to squeeze one or two more :)

PR #13434, opened 2016-06-29
remove completed yellow button test

akash1810:
## What does this change?

removes test as its done
## What is the value of this and can you measure success?

clean code = clean mind ✌️ 
## Does this affect other platforms - Amp, Apps, etc?

no
## Screenshots

n/a
## Request for comment

@jamesgorrie

PR #13471, opened 2016-07-04
change colour of play button from grey to yellow

akash1810:
## What does this change?

We ran a [test](https://github.com/guardian/frontend/pull/13434/files) against this with the hypothesis that a yellow button helps to distinguish the element as a video rather than just another image and thus increases video starts.

We saw a 4.1% increase in video starts in the test, so making it a permanent change.

video starts ⬆️ 

no
## Screenshots

💛 

screen shot 2016-07-04 at 16 21 44
## Request for comment

@jamesgorrie (or anyone as James is in Spain this week!)

### DIFF
--- a/static/src/stylesheets/module/content/_media-player.scss
+++ b/static/src/stylesheets/module/content/_media-player.scss
@@ -119,11 +119,7 @@ $ima-controls-height: 70px;
             @include mq(mobileLandscape) {
                 @include video-play-button-size($vjs-large-button-size);
             }
-            background-color: rgba($media-mute, .6);
-            .gu-media-wrapper:hover &,
-            .gu-media__fallback:hover & {
-                background-color: $media-default;
-            }
+            background-color: $media-default;
         }
 
         &:after {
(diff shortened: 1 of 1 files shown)


static/src/stylesheets/module/content/_media-player.scss
```diff
@@ -157,10 +153,6 @@ $ima-controls-height: 70px;
     }
 }
 
-.vjs-big-play-button--yellow .vjs-big-play-button .vjs-control-text:before {
```
akash1810: left over from the test 😊

paperboyo:
Not even a touch of transparency for the faces to show? 😢

akash1810:
> Not even a touch of transparency for the faces to show?

Nope- wasn't in the original test 😜 .

johnduffell:
great use of data, but to be on the safe side did you check with any design people?

johnduffell:
code looks good! 👍 and great PR description

PR #13472, opened 2016-07-04
HealthCheck: introduce the notion of non expiring internal healthcheck

TBonnin:
## What does this change?

The goal is to be able to have healthcheck endpoints that would be
checked only once (ex: anything that involves upstream services)
Doing so would limit dependencies on upstream services while still
verifying that calling those upstream services is correclty configured at startup

This PR doesn't change the current behaviour of healthchecks since they are all still happening every 5s in the background. Introducing non-expiring healthcheck could be done in a following PR

@johnduffell I looked at having different refresh intervals depending on the healthcheck endpoint but this would introduce multiple refreshing jobs, a complexity I don't really want to add in the codebase

Avoid marking a frontend instance as unhealthy when an upstream service is temporarily unresponsive or slow, which trigger ec2 instances unnecessary restarts and potentially cascading issues 

No
## Request for comment

@johnduffell @alexduf

### DIFF
--- a/admin-jobs/app/controllers/HealthCheck.scala
+++ b/admin-jobs/app/controllers/HealthCheck.scala
@@ -1,8 +1,8 @@
 package controllers
 
-import conf.AllGoodCachedHealthCheck
+import conf.{AllGoodCachedHealthCheck, ExpiringSingleHealthCheck}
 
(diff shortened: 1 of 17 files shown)


johnduffell:
fair enough! 👍

alexduf:
it looks good to me too!

PR #13475, opened 2016-07-04
Move create-ad-slot along with other DFP-related modules

regiskuckaertz:
It seems logical that `create-ad-slot` goes with other DFP-related modules such as `add-slot`.

cc @guardian/commercial-dev @rich-nguyen


Calum-Campbell:
👍  makes sense!

PR #13477, opened 2016-07-05
Remove Aus campaign email sign-up

gtrufitt:
## What does this change?

Removes the Australia campaign catch-up email sign-up.


NataliaLKB:
👍

PR #13478, opened 2016-07-05
Hide iframe when rendering breakout creative

regiskuckaertz:
Turns out GTP will automatically assign the iframe `width` and `height` attributes with the returned creative dimensions:

picture 1

Therefore we must hide it:

picture 2

@guardian/commercial-dev


JonNorman:
:+1:

PR #13480, opened 2016-07-05
Omniture - Remove old R2 storage

mchv:
## What does this change?

Remove code that store tag information for Omniture which I think is not needed to run anymore as we do not serve anything from `r2`.

cc @cb372


johnduffell:
anyone who deletes stuff automatically gets a +1 from me! 👍

PR #13481, opened 2016-07-05
nicer error message

alexduf:
## What does this change?

It gives a nicer error message when starting the dev-build

less confused people :)

No
## Request for comment

@johnduffell cc @marialivia16


johnduffell:
yep brilliant, an error message that's useful can save hours of fun! 👍 thanks

PR #13486, opened 2016-07-05
Discussion: add ping and do not expires original internal healthcheck

TBonnin:
## What does this change?

I am introducing a non expiring internal healthcheck into the discussion app. If this goes well I will extend it to all the apps relying on an upstream service. :)

The real user-facing endpoint which rely on an upstream service (DAPI) would only be checked until it becomes successful.
Then a 200 response from /_healthcheck will reveal that the app is running propely

Make healthcheck less dependent on upstream services while still making sure upstream service are reachable at startup

No
## Screenshots

No
## Request for comment

@alexduf @johnduffell

### DIFF
--- a/discussion/app/controllers/HealthCheck.scala
+++ b/discussion/app/controllers/HealthCheck.scala
@@ -1,5 +1,5 @@
 package controllers
 
-import conf.{AllGoodCachedHealthCheck, ExpiringSingleHealthCheck}
+import conf.{AllGoodCachedHealthCheck, NeverExpiresSingleHealthCheck}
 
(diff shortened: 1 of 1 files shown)


johnduffell:
what does it add? by the fact that someone can call _healthcheck and get an OK, we don't need that in turn to call something else that just gets an OK.  We don't want to end up with a crazy complicated healthcheck system - the original changes to cache them were necessary to avoid overloading the servers, but I really feel simplicity is important here

TBonnin:
Fair point you are right you don't need a ping endpoint, responding to _healthcheck achieve the same thing :)

alexduf:
:+1:

PR #13487, opened 2016-07-05
Commercial app loader

alexduf:
## What does this change?

Removes the global object and introduces DI with the application loader.
It is the same as https://github.com/guardian/frontend/pull/13469 but with an extra commit to fix the deployment issue
## Request for comment

@TBonnin @johnduffell


commercial/conf/application.conf
```diff
@@ -46,11 +46,7 @@ play {
     compressionEnabled: true
   }
 
-  http {
-    filters: "conf.CommonFilters"
-    requestHandler: "dev.DevParametersHttpRequestHandler"
-    errorHandler: "http.CorsHttpErrorHandler"
-  }
+  application.loader = AppLoader
```
TBonnin: 😊

PR #13488, opened 2016-07-05
Hardcode hosted campaign content

kelvin-chappell:
This includes all the content I'm aware of.

Just need images for logo and cta banner to be included, and a css class for the campaign.
And the cta url.

/cc @Calanthe


Calanthe:
👍

PR #13489, opened 2016-07-05
extended switch and removed unneccassery code

Calum-Campbell:
## What does this change?

Extend HighMerch switch and removed some javascript that is not being used. 
## Request for comment

@rich-nguyen


NataliaLKB:
lgtm 👍 .

PR #13490, opened 2016-07-05
move discussion away from Global and introduce app loader

alexduf:
## What does this change?

Same Old: Removes the global object and introduces DI with the application loader.
## Request for comment

@TBonnin @johnduffell


TBonnin:
👍

alexduf:
this is about to be reverted here: https://github.com/guardian/frontend/pull/13514/files

PR #13491, opened 2016-07-05
Standardise on node 6 (take 2)

sndrs:
## What does this change?

This is a reanimation of #13426.
- adds `.npmrc` and declares 6 a requirement
- installs `nvm` instead of `node` in `setup.sh`
- updates karma and phantom
- tidies up `package.json`
- accommodates minor updates to sass packages – _compiled output is identical_
- refactors spricon to work with new phantom 
  - it runs faster
    screen shot 2016-07-05 at 11 51 29
  - the sprites are not identical, presumably because of the different engine, but its only anti-aliasing – here's a visual diff (changes are darker)
    screen shot 2016-07-06 at 13 10 33
- gets everyone in the same boat
- ES6
- [x] teamcity will need to be updated to install node 6

_the disabled `ip-10-248-169-45` agent has node 6 on it, and this PR, plus all other projects have green builds using it_
## Request for comment

@samdesborough, @TBonnin, @rich-nguyen – sorry it's the 3rd time of asking!


rich-nguyen:
Not sure why spricon needed to be updated for v6?

desbo:
I just noticed that the `installed` function didn't work for `nvm` (it's an alias), which made the setup script die if you already had `nvm` installed. I've added a different function to check that `nvm` exists.

sndrs:
> Not sure why spricon needed to be updated for v6?

talked about it offline, but for paper trail it node 6 needed to update phantom, which comes with a new api that is not backwards compatible. it's basically the same code, just wrapped differently

sndrs:
> I've added a different function to check that nvm exists.

nice one, thanks sam

tools/sprites/spricon.js
```diff
@@ -5,190 +5,174 @@
...
-    }
-
-    cssDestDefer.promise.then(function () {
-        phantom.create(function (ph) {
-            ph.createPage(function (sprite) {
-                sprite.set('viewportSize', { width: 600, height: 1 });
-                sprite.set('content', '<html><body id="sprite" style="margin: 0" /></html>');
+const fs = require('fs');
+const path = require('path');
+const glob = require("glob");
```
regiskuckaertz: double quotes here :snoot:

tools/sprites/spricon.js
```diff
@@ -5,190 +5,174 @@
...
+        return `
+                %svg-i-${name},
+                .svg-i-${name} {
+                    background-image: url(data:image/svg+xml;base64,${btoa(fileData.data)});
+                    background-position: 0 0;
+                }
+                .svg .i-${name} {
+                    @extend %svg-i-${name} !optional;
+                }
+        `.replace(/                /g, '');
```
regiskuckaertz: nice touch!

PR #13492, opened 2016-07-05
Add comscore analytics

cb372:
## What does this change?

Adds a Comscore tag to the analytics section at bottom of the page. Hidden behind a switch for now.

Currently Comscore reporting is implemented by piggybacking on top of Omniture (the Omniture server sends a request to Comscore for every request it receives). With the move to Google Analytics, we need to start sending requests directly to Comscore.

We don't need to worry about double-counting during the migration period: Comscore have assured us that they will de-dupe at their end.

Removes dependency on Omniture for Comscore reporting.

No. We will need to do this on those platforms as well.
## Screenshots

n/a
## Request for comment

@gklopper @sndrs @rich-nguyen

### DIFF
--- a/common/app/views/fragments/analytics/comscore.scala.html
+++ b/common/app/views/fragments/analytics/comscore.scala.html
@@ -1,0 +1,21 @@
+@(page: model.Page)
+@import conf.switches.Switches.ComscoreSwitch
+
+@if(ComscoreSwitch.isSwitchedOn) {
+    <script id='comscore'>
+        var _comscore = _comscore || [];
+        _comscore.push({ c1: "2", c2: "6035250" });
+        (function() {
+            var s = document.createElement("script"), el = document.getElementsByTagName("script")[0];
+            s.async = true;
+            s.src = "https://sb.scorecardresearch.com/beacon.js";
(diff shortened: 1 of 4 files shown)


common/app/views/fragments/analytics/comscore.scala.html
```diff
@@ -0,0 +1,21 @@
...
+@import conf.switches.Switches.ComscoreSwitch
+
+@if(ComscoreSwitch.isSwitchedOn) {
+    <script id='comscore'>
+        var _comscore = _comscore || [];
+        _comscore.push({ c1: "2", c2: "6035250" });
+        (function() {
+            var s = document.createElement("script"), el = document.getElementsByTagName("script")[0];
+            s.async = true;
+            s.src = "https://sb.scorecardresearch.com/beacon.js";
```
cb372: Do I need to add this domain to a CSP whitelist anywhere?

desbo: No, the current policy only covers protocols.

desbo:
looks fine to me 👍

sndrs:
yep 👍

PR #13493, opened 2016-07-05
Refactor badges

regiskuckaertz:
Lots of redundant code removed plus the fact that three sets of promises were run sequentially, rather than all in parallel, for no obvious reason.

cc @guardian/commercial-dev

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/badges.js
+++ b/static/src/javascripts/projects/common/modules/commercial/badges.js
@@ -134,22 +134,12 @@ define([
 
         // add a badge to a container (if appropriate)
         add: function (container) {
-            container = container instanceof Element ? container : container[0];
+            if (container) {
+                container = container instanceof Element ? container : container[0];
 
-            if (
-                bonzo(container).hasClass('js-sponsored-container') &&
-                qwery('.ad-slot--paid-for-badge', container).length === 0
-            ) {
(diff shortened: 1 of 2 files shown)


static/src/javascripts/projects/common/modules/commercial/badges.js
```diff
@@ -5,168 +5,145 @@ define([
...
-            if (sponsor) {
-                $adSlot.append(template(
-                    badgeTpl,
-                    {
-                        header:  header,
-                        sponsor: sponsor
-                    }
-                ));
-            }
+        sponsoredfeatures: {
```
JonNorman: Is there a reason that this is not quoted but the other two are?

JonNorman:
Looks good to me! :+1:

static/src/javascripts/projects/common/modules/commercial/badges.js
regiskuckaertz: Actually the opposite: in JS you don't have to quote object keys unless they contain a special character, such as `-`

JonNorman: Ah good to know, thanks! A minor point would then be why we have `sponsoredFeatures` and not `sponsored-features` but I don't think it will break the internet.

PR #13495, opened 2016-07-05
removing old cloudwatch feed metrics

Calum-Campbell:
## What does this change?

Removes old feed fetching/parsing metrics recording that was being used to measure total feed/parse failures/success. 
These are no longer being used and will be replaced by a Kibana log that will record the success/failure of each feed individually.  
## Request for comment

@JonNorman


JonNorman:
:+1:

PR #13496, opened 2016-07-05
Hosted Gallery more fixes

lps88:
## What does this change?
- add tracking when user moves between images
- add gradient behind gallery title & cta
- twitter share button should now work (error due to unescaped #)
- move call to action markup into a separate template as we use it twice
- better text sizes, and new footer styling for mobile breakpoint
  before:
  image
  after:
  image
- info button turns into close button when caption is open on mobile
  before:
  image
  after:
  image
- new styling for up/down arrows
  before:
  image
  after:
  image
## Request for comment

@Calanthe


static/src/javascripts/projects/common/modules/commercial/hosted-gallery.js
```diff
@@ -43,7 +45,7 @@ define([
         // ELEMENT BINDINGS
         this.$galleryEl = $('.js-hosted-gallery-container');
         this.$imagesContainer = $('.js-hosted-gallery-images', this.$galleryEl);
-        this.$captionContainer = $('.js-hosted-gallery-captions');
+        this.$captionContainer = $('.hosted-gallery__lower-bar');
```
Calanthe: Why did you remove the `js-` part?

commercial/app/views/hosted/guardianHostedGallery.scala.html
```diff
@@ -10,34 +11,27 @@
                 <div class="js-hosted-gallery-container hosted-gallery__gallery-container">
                     <div class="js-hosted-gallery-images hosted-gallery__images-container">
                         @guardianHostedGalleryImage(page.images.head, {
+                            <div class="hosted-gallery__image-gradient"></div>
```
regiskuckaertz: Do you really need an extra DIV for this? Elements can have multiple background images in CSS.

commercial/app/views/hosted/guardianHostedGalleryCta.scala.html
```diff
@@ -0,0 +1,10 @@
+@(ctaText: String, ctaLink: String, dataLinkName: String, cssClass: String = "")
+    <div class="hosted-gallery__cta @{cssClass}">
+        <div class="hosted-gallery__cta-title">
+            @{ctaText}
+        </div>
+        <a class="hosted-gallery__cta-link" href="@{ctaLink}" target="_blank" data-link-name="@{dataLinkName}">
+            Visit OMGB now @fragments.inlineSvg("arrow-right", "icon")
```
regiskuckaertz: Shouldn't this text be abstracted too?

commercial/app/views/hosted/guardianHostedGallery.scala.html
lps88: I hadn't thought about that, I'll update it 😺

commercial/app/views/hosted/guardianHostedGalleryCta.scala.html
lps88: Yes........ oops

static/src/javascripts/projects/common/modules/commercial/hosted-gallery.js
```diff
@@ -275,7 +277,7 @@ define([
                 if(this.useSwipe){
                     this.translateContent(this.index, 0, 100);
                 }
-
+                omniture.trackLinkImmediate(config.page.analyticsName + ' - image ' + this.index);
```
rich-nguyen: can you use a `data-link-name` attribute instead? that still ends up calling `trackLink`, but normalises the name

commercial/app/views/hosted/guardianHostedGallery.scala.html
lps88: Ah I think it's because it needed to be on top of the `img` tag but underneath the text, so didn't really fit anywhere else

regiskuckaertz: Hmm, I see. Can you add a class to `hosted-gallery__cta` that adds the gradient?

static/src/javascripts/projects/common/modules/commercial/hosted-gallery.js
lps88: This is called when changing between images (can be by scrolling, swiping or arrow buttons) which is all controlled by the javascript. So, nowhere to really put that attribute. Maybe trackLinkImmediate isn't the right omniture method to call? I'm not familiar at all with omniture! What do you think is best?

commercial/app/views/hosted/guardianHostedGallery.scala.html
lps88: Yeah I think I've found a way round it 🐻

Calanthe:
👍

PR #13497, opened 2016-07-05
Refactor 3rd party scripts

regiskuckaertz:
Every 3rd party module is basically doing the same thing many times, so I figured we could do that same thing only once instead.

cc @guardian/commercial-dev

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/third-party-tags.js
+++ b/static/src/javascripts/projects/common/modules/commercial/third-party-tags.js
@@ -97,9 +97,7 @@ define([
             var service = services.shift();
             var script = document.createElement('script');
             script.src = service.url;
-            if (service.onLoad) {
-                script.onload = service.onLoad;
-            }
+            script.onload = service.onLoad;
             frag.appendChild(script);
         }
         ref.parentNode.insertBefore(frag, ref);
--- a/static/src/javascripts/projects/common/modules/commercial/third-party-tags.js
+++ b/static/src/javascripts/projects/common/modules/commercial/third-party-tags.js
@@ -69,13 +69,6 @@ define([
             return false;
         }
 
-        switch (config.page.edition.toLowerCase()) {
-            case 'uk':
-                audienceSciencePql.load();
-                audienceScienceGateway.load();
-                break;
-        }
-
         // Outbrain/Plista needs to be loaded before first ad as it is checking for presence of high relevance component on page
         loadExternalContentWidget();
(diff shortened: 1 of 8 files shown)


static/src/javascripts/projects/common/modules/commercial/third-party-tags.js
```diff
@@ -84,12 +77,30 @@ define([
...
     function loadOther() {
-        imrWorldwide.load();
-        remarketing.load();
+        var services = [
+            audienceSciencePql,
+            audienceScienceGateway,
+            imrWorldwide,
+            remarketing,
+            krux
+        ].filter(function (_) { return _.shouldRun; });
```
sndrs: I think we should use fully descriptive variables where possible – where here, `_` is really `service`?

regiskuckaertz: Generally I agree, it is especially important when going through long procedural processes. But in this case, I feel like the important information `shouldRun` should be the focus of this step of the process (the filtering) and we don't really care about what should run.

static/src/javascripts/projects/common/modules/commercial/third-party-tags/audience-science-pql.js
```diff
@@ -1,71 +1,72 @@
...
-                        return placement[1]['default']; //eslint-disable-line
-                    }).and(map, function (placement) {
-                        return ['pq_' + placement[0], 'T'];
-                    }).and(zipObject).valueOf();
-                // set up the global asiPlacements var in case dfp returns before asg
-                window.asiPlacements = storedSegments[section];
-            }
-            return segments;
-        };
+    function onLoad() {
```
JonNorman: What are we actually doing here/ why? Can pop over and discuss if it is easier.

JonNorman:
Apart from the slightly confusing (to me at least) flow of `audience-science-pql`, looks good to me :+1:

static/src/javascripts/projects/common/modules/commercial/third-party-tags/audience-science-pql.js
regiskuckaertz: TBH I don't know, seems we swap two values: one we receive from the 3rd party, one we store in local storage.

JonNorman: 😂  it makes none of the sense.

PR #13498, opened 2016-07-05
Switch to production Google Analytics account

cb372:
## What does this change?
- Switch to universal GA account
- Remove sampling - send all traffic to GA
- Update the GA feature switch

👋  Omniture -->💰 

No
## Screenshots

n/a
## Request for comment

@gklopper @johnduffell

### DIFF
--- a/common/app/conf/switches/MonitoringSwitches.scala
+++ b/common/app/conf/switches/MonitoringSwitches.scala
@@ -31,8 +31,8 @@ trait MonitoringSwitches {
     "google-analytics",
     "If this switch is on, then Google Analytics is enabled",
     owners = Seq(Owner.withGithub("gklopper")),
-    safeState = Off,
-    sellByDate = new LocalDate(2016, 8, 26),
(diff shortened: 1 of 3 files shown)


common/app/conf/switches/MonitoringSwitches.scala
```diff
@@ -31,8 +31,8 @@ trait MonitoringSwitches {
     "google-analytics",
```
johnduffell: delete the switch completely?

johnduffell:
looks good to me, fwiw 👍

common/app/conf/switches/MonitoringSwitches.scala
cb372: I think I'll leave it for a little longer, just in case the change to send 100% of requests to GA has a disastrous effect on site performance.

johnduffell: fair enough in that case leave the expiry for now

PR #13501, opened 2016-07-06
Stop multiple URLs for hosted content

kelvin-chappell:
It's possible to serve hosted content on 
`/commercial/advertiser-content/...`
and 
`/advertiser-content...`

With this change, each hosted page is only available on a single path.
The content that's already been published is considered to be legacy in this fast-moving world.

/cc @Calanthe @lps88


lps88:
👍

PR #13503, opened 2016-07-06
Control the margins of the inline slot

regiskuckaertz:
The inline event and book components still use the old code. I will migrate them but until then the margins need to be controlled by the slot itself:

picture 4

Now:

picture 5


JonNorman:
:+1:

PR #13504, opened 2016-07-06
Fixes lightbox bug

### LINKED ISSUE #13452: Lightbox opening icon broken
Hello,

Helmut Kohl does not seem entirely happy about the extra dark innercircle in the below icon that appeared recently on the site (Brexit is enough of a worry):
kohl not happy
[all platforms, URL: https://www.theguardian.com/world/2016/jun/30/slovakia-foreign-minister-eu-support-measure-stop-brexit]

Maybe we could take this opportunity to revisit the icon itself? It recently disappeared from gallery pages. While I totally understand that it obstructs images, unfortunately the lack of it also makes the Lightbox less discoverable. And that goes against my sick plan to turn the Lighbox into an app that will provide a whole new parallel way to browse all Guardian content through multimedia and make us loadsamoney.

No rush on making money, though, we can just remove the extra dark circle :wink:

Regards
Mateusz

NataliaLKB:
Fixes: https://github.com/guardian/frontend/issues/13452.

It looks like this problem has been around for a while! Rather than a list of multiple strings, this is a list of one string. Fixing that fixes the lightbox problem 😄 .

Nope:
## Screenshots

Before:
image

After:
image

### DIFF
--- a/common/app/views/fragments/imageFigure.scala.html
+++ b/common/app/views/fragments/imageFigure.scala.html
@@ -89,7 +89,7 @@
         @lightboxIndex.map { index =>
         <a href="#img-@index" class="article__img-container js-gallerythumbs" data-link-name="Launch Article Lightbox" data-is-ajax>
             @imageHtml
-            @fragments.inlineSvg("expand-image", "icon", List("centered-icon rounded-icon article__fullscreen modern-visible"))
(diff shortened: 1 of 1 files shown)


rich-nguyen:
👍

PR #13502, opened 2016-07-06
make yellow play button accessible

akash1810:
## What does this change?

the current yellow and white fails accessibility tests, so making the button yellow and grey, with a darkened hover state (per social buttons)

original change https://github.com/guardian/frontend/pull/13471
## What is the value of this and can you measure success?

n/a
## Does this affect other platforms - Amp, Apps, etc?

no
## Screenshots

yellow
## Request for comment

@sammorrisdesign @zeftilldeath

PR #13505, opened 2016-07-06
hosted content play button back to neutral-7

akash1810:
## What does this change?

[recently changed](https://github.com/guardian/frontend/pull/13502) the play button to be neutral-1 by default - this was a global change but should not have touched hosted content 😶

branding

no
## Screenshots

screen shot 2016-07-06 at 13 20 06
## Request for comment

@ScottPainterGNM @Calanthe


Calanthe:
👍  Thanks for fixing it!

PR #13506, opened 2016-07-06
move facia away from Global and introduce app loader

alexduf:
## What does this change?

Removes the global object and introduces DI with the application loader.

after this one: 5 more to go.
## Request for comment

@TBonnin @johnduffell


TBonnin:
👍

PR #13508, opened 2016-07-06
Add prominent layout to manual multiple

regiskuckaertz:
Here it is:

picture 6

cc @guardian/commercial-dev


Calanthe:
Does it mean that in this layout, the caption for the first image will be always on the right? If so, there is a big gap under the CTA button and it looks ugly ;/
👍  or the code

regiskuckaertz:
Good spot! I will do that 😄

PR #13509, opened 2016-07-06
Use megalog to show extra steps in setup.sh

desbo:
## What does this change?

You get a more noticeable prompt if there are extra steps after running `setup.sh`.
## Screenshots

screen shot 2016-07-06 at 16 41 55
## Request for comment

@sndrs

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -132,8 +132,7 @@ compile() {
 
 report() {
   if [[ ${#EXTRA_STEPS[@]} -gt 0 ]]; then
-    echo -e
-    echo "Remaining tasks: "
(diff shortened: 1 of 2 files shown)


sndrs:
is there any way we could pass them into the body of megalog?

desbo:
we could add a flag to `messages.js` to let you pass in any text you want instead of one of the `case` names. not sure it's worth the effort though, the message will still be easy enough to find by showing the generic megalog and once you've found it the steps will be directly underneath

sndrs:
fair enough 👍

PR #13510, opened 2016-07-06
Removing AMP CORS headers logic as these are handled by fastly

jfsoul:
## What does this change?

Removes logic and config for adding AMP CORS headers in the backend as these are now added by fastly to avoid caching.

See associated trello card for more detail. See also https://github.com/guardian/fastly-edge-cache/pull/655

Avoids duplication of the logic for adding AMP CORS headers

AMP
## Request for comment

@TBonnin @NataliaLKB


NataliaLKB:
👍

PR #13511, opened 2016-07-06
Remove aus election badging on liveblogs

sndrs:
## What does this change?

Removes the badging hack that inserted the image of Katherine Murphy on AUS election liveblogs 

editorial request

no
## Screenshots

before:

screen shot 2016-07-06 at 16 45 56

after:

screen shot 2016-07-06 at 16 45 40
## Request for comment

@sammorrisdesign @blongden73 or anyone in design who has the time to look


sammorrisdesign:
👍

PR #13512, opened 2016-07-06
Leffe hosted page styling fixes

Calanthe:
## What does this change?

A few styling and content fixes for the Leffe hosted videos pages:
- poster and banner images added
- custom twitter and email text added
- brand styling (temporary is yellow - as the Renault campaign - because of the contrast issues)
## Screenshots

screen shot 2016-07-06 at 17 03 36


lps88:
It looks really nice! The only thing I would suggest is to put the images into a folder in this s3 bucket:
`https://console.aws.amazon.com/s3/home?region=eu-west-1#&bucket=static-theguardian-com&prefix=commercial/hosted/`
instead of into the repository. Then you can use a url something like `http://static.theguardian.com/commercial/hosted/leffe/some-pretty-image.jpg`

regiskuckaertz:
I agree with @lps88 but just want to 👍  because belgian beer.

On a side note, if the hosted CSS grows, it might be interesting to create a separate stylesheet that is loaded only on hosted articles, doesn't it?

Calanthe:
Good ideas @lps88 and @regiskuckaertz . I will do that in another PR.

PR #13513, opened 2016-07-06
Allow forcing of A/B complete events in dev mode

desbo:
## What does this change?

It's now possible to trigger the complete event listeners for an A/B test with the same anchor method you use to force yourself into a test, also adds a bit to the docs about this.

Easier debugging of tests

No
## Request for comment

@dominickendrick @gtrufitt @markjamesbutler

### DIFF
--- a/docs/ab-testing.md
+++ b/docs/ab-testing.md
@@ -144,8 +144,8 @@ Add #ab-<TestName>=<VariantName> to the end of your URL (in dev or prod) to forc
 e.g. www.theguardian.com/news#ab-MyGreatTest=GreenButton
 
 #### Firing complete events in dev mode
-In prod, the completion events are fired based on the MVT ID cookie. This doesn't exist in dev, so if you need to test a complete event, follow the #ab- pattern above.
-Whichever test you set in the hash will have its completion function called.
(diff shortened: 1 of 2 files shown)


gtrufitt:
Crackin 👍

docs/ab-testing.md
```diff
@@ -139,6 +139,14 @@ define([
     })
 ```
 
+### Forcing yourself into a test
+Add #ab-<TestName>=<VariantName> to the end of your URL (in dev or prod) to force yourself into a test.
+e.g. www.theguardian.com/news#ab-MyGreatTest=GreenButton
+
+#### Firing complete events in dev mode
+In prod, the completion events are fired based on the MVT ID cookie. This doesn't exist in dev, so if you need to test a complete event, follow the #ab- pattern above.
```
dominickendrick: > This doesn't exist in dev, so if you need to test a complete event, follow the #ab- pattern above.

Could this be slightly clearer ? Maybe
`This doesn't exist in dev, so if you need to test a complete event, force yourself into the test using the #ab-<TestName>=<VariantName> pattern described above.`

PR #13515, opened 2016-07-06
Quiz omniture record clicks

NathanielBennett:
## What does this change?

Turns out that because the quiz questions aren't buttons or anchors, the clickstream doesn't  automatically record participation in omniture. We want this for the next for weeks ( or whatever it is )

We see quiz clicks in omniture

N/A
## Request for comment

@gtrufitt @crifmulholland

### DIFF
--- a/static/src/javascripts/projects/common/modules/atoms/quiz.js
+++ b/static/src/javascripts/projects/common/modules/atoms/quiz.js
@@ -2,12 +2,14 @@ define([
     'bean',
     'common/utils/$',
     'common/utils/fastdom-promise',
-    'lodash/collections/toArray'
+    'lodash/collections/toArray',
+    'common/modules/analytics/omniture'
 ], function (
     bean,
     $,
     fastdom,
-    toArray
+    toArray,
+    omniture
 ) {
     return {
         // find a bucket message to show once you finish a quiz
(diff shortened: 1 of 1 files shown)


gtrufitt:
👍 Code good but probably need to check the event string with @crifmulholland

static/src/javascripts/projects/common/modules/atoms/quiz.js
```diff
@@ -21,10 +23,12 @@ define([
 
                 if ($quizzes.length > 0) {
                     bean.on(document, 'click', toArray($quizzes), function (e) {
-                        var quiz = e.currentTarget;
+                        var quiz = e.currentTarget,
+                            total =  $(':checked + .atom-quiz__answer__item', quiz).length;
+
+                        omniture.trackLinkImmediate('quiz-question-answered-'+total);
```
rich-nguyen: Any reason not to use `data-link-name` attribute?

NathanielBennett: @rich-nguyen I think ( looking at clickstream.js, and the fact that i spent about 2 hours failling to get a data-link-name to fire and event, it's because there's no surrounding anchor or button in the html here, so that doesn't work)

PR #13517, opened 2016-07-07
extend dpf empty ad repsonse switch

Calum-Campbell:

Extend the switch for reporting dfp empty ad responses.
## Request for comment

@rich-nguyen


rich-nguyen:
I'm going to remove the switch entirely, #13516

Calum-Campbell:
Ah fair enough!

PR #13518, opened 2016-07-07
Remove redundant setup step and file

desbo:
There's no need for `install-dependencies.sh` since it only calls `make install`. I've also removed  the install dependencies step from the script, since `compile` calls `make install` as well.

@sndrs


rich-nguyen:
👍

PR #13520, opened 2016-07-07
Fix libpng check in setup.sh

desbo:
Another fix to stop the script crashing. If you try to `brew install` something that already exists, the script will terminate.


sndrs:
👍

PR #13524, opened 2016-07-07
Hosted about overlay mobile fixed

Calanthe:
## What does this change?

The hosted 'about' overlay is scrollable on mobile now.


lps88:
👍

PR #13525, opened 2016-07-07
Pass event emitter to ophan so it can use throttledScroll

philwills:
## What does this change?

Makes Ophan use the event emitter for measuring scroll activity when measuring component attention.

Reduces the load on reader's browsers. No easily defined metric to determine success.

No
## Screenshots

N/A


philwills:
Build failure doesn't appear to be related at first sight.

dominickendrick:
👍

PR #13526, opened 2016-07-07
Fix double badges in containers

regiskuckaertz:
Sponsored fronts will add a badge to the first container, but if this container appears to be sponsored too, another badge will be added.

This change makes sure the front case is handled before the containers one, to avoid the double badge.

Before:
picture 8

After:
picture 7

cc @JonNorman @rich-nguyen


JonNorman:
c'est bon :+1:

PR #13527, opened 2016-07-07
Add missing thumbnailUrl for video elements

katebee:
## What does this change?

meta itemprop="thumbnailUrl" is missing from VideoObject.

Yes, having thumbnail and thumbnailUrl may seem redundant. There is an open issue on this with schemaorg: https://github.com/schemaorg/schemaorg/issues/239

Source for articles with video as the main-media ([example](http://www.theguardian.com/politics/video/2016/jun/21/brendan-cox-jo-cox-was-killed-because-of-her-political-views-video)) contain no `thumbnailUrl` for the video.

Adding one removes an error reported by the structured data testing tool: https://search.google.com/structured-data/testing-tool#

The Google search result for this article is currently using the OpenGraph image, possibly as a fallback due to missing meta data?

picture 102

Anything that uses Schema (http://schema.org/) structured data to parse or scrape our pages.
Google search, Bing, etc.

Oddly, Google+ appears to be giving OpenGraph higher ranking when parsing pages (over Schema). Google Search currently favours Schema.
#### BEFORE:

picture 101
#### AFTER:

picture 100

### DIFF
--- a/applications/app/views/fragments/mediaBody.scala.html
+++ b/applications/app/views/fragments/mediaBody.scala.html
@@ -70,7 +70,9 @@
                                         }
                                         <meta itemprop="requiresSubscription" content="false" />
                                         <meta itemprop="image" content="@Html(video.content.signedArticleImage)" />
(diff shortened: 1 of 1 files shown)


dominickendrick:
👍

PR #13528, opened 2016-07-07
Revert party

alexduf:
## What does this change?

It turns out play 2.4.8 breaks our admin login.
So I'm reverting that change and the discussion change that was dependent on it as well.
## Request for comment

@TBonnin


TBonnin:
👍  😢

PR #13529, opened 2016-07-07
Hosted pages - changes in the twitter and email shared text 

Calanthe:
## What does this change?

"Advertiser content hosted by the Guardian" removed from the Twitter and share via email text.


lps88:
👍 we probably need to fix this for galleries too then. I'll make a note!

Calanthe:
In the galleries that text is taken from the page title. I am not sure if it's so crucial to change it. I made this specific for the Leffe campaign because they specified it in docs.

PR #13530, opened 2016-07-07
Upadated EU ref email header to Brexit

zeftilldeath:
Cos it isn't a referendum now. It's a brexit.


desbo:
👍

dominickendrick:
Nice reuse of the old design! Would have been a little more work if it was remain!

zeftilldeath:
@dominickendrick It's Chris Clarke's design anyway but that's a really good point!

paperboyo:
Wider population always votes for less design-disrupting options. Pragmatism.

PR #13532, opened 2016-07-08
run krux on section fronts but not network fronts

Calum-Campbell:
## What does this change?

Previously we had not run the Krux script on any fronts. 

Now we will run it on all section fronts but **not** on Network Fronts.
## Request for comment

@regiskuckaertz  @guardian/commercial-dev


regiskuckaertz:
👍

PR #13533, opened 2016-07-08
Bigger images for image objects

katebee:
## What does this change?

Swaps the 300px wide image URL in the `itemprop=url` to a 700px wide one.

Shares a larger image (>696px) with crawlers reading our Schema microdata/metadata (http://schema.org/)

> ImageObject, required
> 
> The representative image of the article. Only a marked-up image that directly belongs to the article should be specified.
> 
> Images should be at least 696 pixels wide

https://developers.google.com/search/docs/guides/mark-up-content

Anything that uses Schema (http://schema.org/) structured data to parse or scrape our pages.
Google search, Bing, etc.

AMP will be unaffected, the AMP html is found on a separate page (`<link rel="amphtml"`) and the images for AMP are marked up using AMP tags: `<amp-img src="example.jpg"`
#### BEFORE

picture 104
#### AFTER

picture 105

### DIFF
--- a/common/app/views/fragments/imageFigure.scala.html
+++ b/common/app/views/fragments/imageFigure.scala.html
@@ -36,7 +36,7 @@
         @if(isMain) {
             <meta itemprop="representativeOfPage" content="true">
         }
-        <meta itemprop="url" content="@ImgSrc.getFallbackUrl(picture)">
+        <meta itemprop="url" content="@ImgSrc.findNearestSrc(picture, Item700)">
         <meta itemprop="width" content="@ImgSrc.getFallbackAsset(picture).fold(0)(_.width)">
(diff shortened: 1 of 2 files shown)


common/app/views/fragments/imageFigure.scala.html
```diff
@@ -36,7 +36,7 @@
         @if(isMain) {
             <meta itemprop="representativeOfPage" content="true">
         }
-        <meta itemprop="url" content="@ImgSrc.getFallbackUrl(picture)">
+        <meta itemprop="url" content="@ImgSrc.findNearestSrc(picture, Item700)">
```
katebee: getFallbackUrl is found in Profile.scala. It returns the result of findNearestSrc, but with Item300 being passed in, just removing the middleman to get a bigger image

janua:
:+1:

PR #13534, opened 2016-07-08
hosted video pages - twitter txt fix

Calanthe:
## What does this change?

The twitter message on hosted video pages includes `#ad` now and its properly encoded.


regiskuckaertz:
👍

PR #13536, opened 2016-07-08
Remove no ad on SSL pages flag and fix adverts promises

regiskuckaertz:
cc @Calum-Campbell


Calum-Campbell:
👍  looks good to me.

PR #13537, opened 2016-07-08
Fix manual multiple: Add hide-until-tablet class after mapping

regiskuckaertz:
Before:
picture 9

After:
picture 10

cc @Calum-Campbell


Calum-Campbell:
👍

PR #13519, opened 2016-07-07
bump to play 2.4.8

alexduf:
## What does this change?

This should fix the CSRF problem encountered yesterday with https://github.com/guardian/frontend/pull/13490
## Request for comment

@TBonnin @johnduffell

PR #13540, opened 2016-07-08
Discussion csrf play

alexduf:
## What does this change?
1. https://github.com/guardian/frontend/pull/13519 (commit eb3d86a; already reviewed)
2. https://github.com/guardian/frontend/pull/13490 (commit ea8e68a; already reviewed)
3. A fix related to Play's 2.4.6 version: https://github.com/playframework/playframework/issues/5829 (commit 3ae31b2). Play 2.4.6 only trusts X-Forwarded headers if the requests comes through a trusted proxy
4. A fix to use the right CSRFConfig in the CommentsController (38047cc)
5. A workaround a Play bug related to CSRF: helper.CSRF.formField is hardcoded to use the global CSRFConfig. That means using compile time dependency injection leads that part of the code to throw an exception. This has been fixed in play 2.5 so that workaround should be enough. (commit 03acac1)

Here is the chain of event that lead to that PR:
- Discussion was migrated to remove Globals => that lead to CSRF errors. (fix number 4 and 5)
- Bump play version to play 2.4.8 => that lead to request.secure returning false, locking people out of admin (fix number 3)

This PR should address the previously discovered issues.
I can only test point number 3 in CODE as it is related to our infrastructure.
## Request for comment

@TBonnin @johnduffell


TBonnin:
👍  Hopefully the last attempt 😁

PR #13541, opened 2016-07-08
Leffe branding fixes

Calanthe:
## What does this change?

A few Leffe branding fixes:
- colours changed from yellow to gold
- one common banner at the bottom
- the banner bottom link will have an extra button but this will be finished in another PR, right now it looks like Renault
## Screenshots

screen shot 2016-07-08 at 14 29 42
## Request for comment

@lps88 @kelvin-chappell


lps88:
👍

PR #13543, opened 2016-07-08
HSTS opt in/out

JustinPinner:
## What does this change?

Adds a new opt in/out feature that allows us to begin tests with HSTS headers. The HSTS header would be set for anybody opting in, regardless of whether the base url is served over https. 

This is to allow https opt in/out to be unaffected by our HSTS tests, but it's likely that you would opt in to both https and HSTS to get meaningful results. 

We can determine what issues remain prior to moving to 100% https.

No. Web only.
## Request for comment

@desbo @mchv


desbo:
👍

PR #13544, opened 2016-07-08
Visit britain galleries

lps88:
## What does this change?
- Add the rest of the gallery pages for visit britain campaign
- Use images from grid so that browser can request appropriately sized image

CC @Calanthe @kelvin-chappell

### DIFF
--- a/common/app/common/commercial/hosted/HostedGalleryPage.scala
+++ b/common/app/common/commercial/hosted/HostedGalleryPage.scala
@@ -13,7 +13,7 @@ case class HostedGalleryPage(
   ctaText: String,
   ctaLink: String,
   ctaButtonText: String,
-  ctaIndex: Integer,
+  ctaIndex: Integer = 0,
   images: List[HostedGalleryImage]
 ) extends HostedPage {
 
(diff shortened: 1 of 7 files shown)


static/src/javascripts/projects/common/modules/commercial/hosted-gallery.js
```diff
@@ -151,22 +151,16 @@ define([
...
-                $img.attr('src', imageContent[i]);
-
-                bean.one($img[0], 'load', function () {
-                    that.imageRatios[i] = this.naturalWidth/this.naturalHeight;
-                    that.resizeImage.call(that, i);
-                });
-            } else {
-                that.resizeImage.call(that, i);
+            if (!this.imageRatios[i]) {
+                that.imageRatios[i] = $img[0].naturalWidth/$img[0].naturalHeight;
```
Calanthe: Maybe its worth checking if the `$img[0].naturalHeight` value is not 0 or null? Maybe if both of them are numbers.

common/app/common/commercial/hosted/HostedGalleryPage.scala
```diff
@@ -13,7 +13,7 @@ case class HostedGalleryPage(
   ctaText: String,
   ctaLink: String,
   ctaButtonText: String,
-  ctaIndex: Integer,
+  ctaIndex: Integer = 0,
```
kelvin-chappell: Would be better to make this an `Option[Int]` if it's not always known at create time.

static/src/javascripts/projects/common/modules/commercial/hosted-gallery.js
lps88: if they're not numbers, the value will end up being NaN and next time it tries to load it will go through this if block again. I do need to put back the onload handler though...

kelvin-chappell:
👍

paperboyo:
> Use images from grid so that browser can request appropriately sized image

Woohoo! :-)

PR #13545, opened 2016-07-08
Remove top below nav slot

regiskuckaertz:
No longer necessary

@guardian/commercial-dev


kelvin-chappell:
👍  
I think it was only used for Apple campaigns.
If I mention them in a public repo do I win something?

jbreckmckye:
Good riddance.

regiskuckaertz:
Yeah! Thanks guys.

@kelvin-chappell I will ask Pat 😃

PR #13547, opened 2016-07-08
Remove redirect during Facebook share, and update tests

katebee:
## What does this change?
- Removes the redirect from the Facebook share button
- Allows content hosted on external sites (youtube) to be shared on Facebook, from the Guardian
- Stops the user from having two identical tabs open after using a FB share button
- Makes the FB share button have the same behaviour as the Twitter share button
- Doesn't affect the FB campaign codes (which are appended before the `redirect_url`)

When you try and share a livestream video ([example](https://www.theguardian.com/membership/2016/jun/28/brexit-livestream-what-happens-next-eu-referendum)) to Facebook, using the share buttons underneath the video, you get the error:

picture 108

This is because the `redirect_url` being sent to FB is for **youtube** (it is passed in from the `item.href`). To solve this we need to redirect back to the Guardian.

However, the redirect is not necessary! :sunglasses: FB shares work better without it:

All our FB sharing buttons already open the share dialog in a new tab (using `target="_blank"`). Without the `redirect_url`, the default behaviour for the facebook.com/dialog/share? tab is to close when the user clicks the POST button. This returns the user to the last active tab (generally, the page of the Guardian that they were trying to share).

Current behaviour means that the user will have two identical tabs open after sharing, although the redirected one will have `?CMP=share_btn_fb#_=_` in the URL

Affects the stages in sharing content to Facebook
## Request for comment

@mchv @rich-nguyen


katebee:
Thoughts on relying on `target=_blank` to open new tab, and allowing Facebook to control the final stage of the process (closing the tab)?

mchv:
👍

PR #13550, opened 2016-07-08
Sticky top banner race conditions

rich-nguyen:
There are some errors being thrown in Sentry which look like sticky top banner is operating on a dom tree without a `top-above-nav` slot. This is likely to be a race condition, where the commercial app is in the process of closing (ie removing) server-rendered slots. So we need to wait for commercial dom writes to finalise the slot structure on the page before running any sticky top banner code. I've moved the whole module into the commercial app.

Also, the logic to decide if we run with a sticky top banner, a top banner, or no top banner and header, was all a bit crazy. I've moved it to the server and tried to break it down, but it's still fundamentally complicated.

### DIFF
--- a/static/src/javascripts/bootstraps/commercial.js
+++ b/static/src/javascripts/bootstraps/commercial.js
@@ -17,7 +17,8 @@ define([
     'common/modules/commercial/slice-adverts',
     'common/modules/commercial/third-party-tags',
     'common/modules/commercial/paidfor-band',
-    'common/modules/commercial/paid-containers'
+    'common/modules/commercial/paid-containers',
+    'common/modules/commercial/sticky-top-banner'
 ], function (
     Promise,
     config,
@@ -37,7 +38,8 @@ define([
     sliceAdverts,
     thirdPartyTags,
     paidforBand,
-    paidContainers
+    paidContainers,
+    stickyTopBanner
 ) {
     var modules = [
         ['cm-dfp', dfpInit],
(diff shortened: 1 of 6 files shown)


static/src/javascripts/bootstraps/enhanced/common.js
```diff
@@ -133,27 +131,6 @@ define([
...
-                if (!(config.switches.disableStickyAdBannerOnMobile && detect.getBreakpoint() === 'mobile')
-                    && !config.page.shouldHideAdverts
-                    && config.page.section !== 'childrens-books-site'
-                    && !config.tests.abNewHeaderVariant
-                    && (config.page.hasSuperStickyBanner
-                        || config.page.contentType !== 'Interactive'
-                        && config.page.contentType !== 'Crossword'
-                        && config.page.contentType !== 'Hosted'
-                        && !config.page.isImmersive
-                        && !config.page.isUsMinute
```
rich-nguyen: i don't think this key ever exists, @gtrufitt, its called `isMinuteArticle` client-side?

static/src/javascripts/projects/common/modules/commercial/sticky-top-banner.js
```diff
@@ -211,7 +211,9 @@ define([
     };
 
     var initialise = function () {
-        if (detect.isBreakpoint({ min: 'desktop' })) {
+        // Although we check as much config as possible to decide whether to run sticky-top-banner,
+        // it is still entirely possible for the ad slot to be closed.
+        if (detect.isBreakpoint({ min: 'desktop' }) && $adBannerInner[0]) {
```
rich-nguyen: Hopefully this is the catch-all line which stops all the errors flooding sentry. But the rest of the PR is about making sure the ad slot really is there.

static/src/javascripts/bootstraps/commercial.js
```diff
@@ -46,12 +50,36 @@ define([
         ['cm-sliceAdverts', sliceAdverts.init],
         ['cm-frontCommercialComponents', frontCommercialComponents.init],
         ['cm-closeDisabledSlots', closeDisabledSlots.init]
+    ],
+    secondaryModules = [
```
JonNorman: I've been told we are moving away from single-var declaration, but I don't feel strongly either way :)

JonNorman:
Looks good to me, pending whether `isMinuteArticle` is still a thing :+1:

static/src/javascripts/bootstraps/enhanced/common.js
gtrufitt: Yeah, `isMinuteArticle`, probably a missed refactor somewhere along the line

rich-nguyen:
it's a thing, i used the server-side equivalent.

PR #13551, opened 2016-07-08
Reverting prebid upgrade

rich-nguyen:
There is some bizarre problem with prebid which manifests itself in an issue with our `create-store` module, so i'm rolling back.

@regiskuckaertz


regiskuckaertz:
😢

Issue #13177, opened 2016-06-02
Cmd+click on a Football result opens both the page in the current tab and a new tab


TBonnin:
# ISSUE
Cmd+click on a Football result opens **both** the page in the current tab and a new tab

## Steps to Reproduce
1. Go to https://www.theguardian.com/football/results
2. Cmd+click (or ctrl+click) on a result

## Actual Results (include screenshots)
The match page is  opened in both the current tab and the new tab

## Expected Results (include screenshots)
Nothing happens in the current tab
The match page is  opened in the new tab

## URL
https://www.theguardian.com/football/results

## OS and Browser details

Tried on Mac OS X 10.5 
Chrome 50 and Firefox 44.0.2

PR #13552, opened 2016-07-08
Fix football link issue #13177

katebee:
## What does this change?
- Remove the event setting `window.location = this.getAttribute('data-link-to');` as it is appending a click event to a table containing hyperlinks, this was causing issue #13177
- Removes an unused code block (as instructed by the comment); the switch `worldCupWallchartEmbed` appears to no longer exist
- CMD+click on a football result will no longer open the match page in the current tab ⚽
- Removes some dead code 💀 (lines 286-316)

No
## Screenshots

N/A
## Request for comment

@TBonnin

### DIFF
--- a/static/src/javascripts/bootstraps/enhanced/football.js
+++ b/static/src/javascripts/bootstraps/enhanced/football.js
@@ -13,8 +13,7 @@ define([
     'common/modules/sport/football/match-list-live',
     'common/modules/sport/football/tag-page-stats',
     'common/modules/sport/score-board',
-    'common/modules/ui/rhc',
-    'lodash/functions/debounce'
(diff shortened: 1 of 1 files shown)


TBonnin:
🔪 💀 
👍 

Since the `worldCupWallchartEmbed` switch doesn't exist anymore, do we even need to keep this file https://github.com/guardian/frontend/blob/6c1938d5ff86fd476dfab26d7433350ded680fea/sport/app/football/views/wallchart/embed.scala.html

katebee:
@TBonnin Probably not? I can go check then delete it too.

TBonnin:
Thank you @katebee

PR #13553, opened 2016-07-10
Use a sans-serif font for business stocks price

mchv:
### What does this change?

Currently business stocks price use guardian font where several numbers characters (`6`, `8` ⬆️  and `5` ⬇️ ) are not vertically aligned which render the reading difficult. I added a bit of space between letters to improve the reading.
#### before

capture d ecran 2016-07-10 a 10 59 40
#### after

capture d ecran 2016-07-10 a 11 24 34

### DIFF
--- a/static/src/stylesheets/module/business/_stocks.scss
+++ b/static/src/stylesheets/module/business/_stocks.scss
@@ -106,6 +106,7 @@ $stocks-icon-height: 10px;
 .stocks__price {
     margin-right: .25em;
     font-family: sans-serif;
+    letter-spacing: .05em;
 }
 
 .stocks__closed,
--- a/static/src/stylesheets/module/business/_stocks.scss
+++ b/static/src/stylesheets/module/business/_stocks.scss
@@ -105,6 +105,7 @@ $stocks-icon-height: 10px;
 
 .stocks__price {
     margin-right: .25em;
+    font-family: sans-serif;
 }
 
 .stocks__closed,


katebee:
👍

paperboyo:
This shouldn't be using `sans-serif`! It isn't used anywhere else on the site and brakes the beauty of the design. To attain lining figures (instead of old style figures), I recommend using Guardian Text Sans Regular instead.

This may be one of the rare places where the upcoming stripping of tabular lining (among other OT features) from fonts may hurt us (would we decide that we would like to have these stocks written in tabular lining)... But the weight/performance reductions will be worth it, I think. (cc/ @sndrs)

sndrs:
agree this should not be generic sans serif, we don't use that anywhere else and i'm certain design would not be ok with it

know it's a bit late but  👎 

@zeftilldeath @superfrank

mchv:
> This shouldn't be using sans-serif! It isn't used anywhere else on the site and brakes the beauty of the design. To attain lining figures (instead of old style figures), I recommend using Guardian Text Sans Regular instead.

I am fine `Guardian Text Sans` if this is readable (numbers should be vertical aligned). With default font it was so unreadable, that I bother created a PR over the week-end!

Anyway I have another PR to adjust the price change and fixed the broken color feature

@paperboyo happy to discuss with you about it 😃

paperboyo:
> numbers should be vertical aligned

Yep. Default for Guardian Text Sans is lining figures:

picture 35

sndrs:
can we revert this? it's still live, it looks wrong and afaik, was not a request?

zeftilldeath:
Yep, or as @mchv is planning another PR you could switch them to fs-textSans ?

mchv:
@sndrs I am fixing, please do not revert!

mchv:
@paperboyo @sndrs with https://github.com/guardian/frontend/pull/13559 it should used the `Guardian Text Sans`. still working on my other PR

mchv:
I have just created the PR to revert as I don't think it is useful to spend more time on this if we don't agree that this component need specific styling.

PR #13554, opened 2016-07-11
Leffe cta fix

Calanthe:
## What does this change?

The whole CTA banner is clickable now. Also, if there is a button text specified (like in the Leffe campaign) it looks little different than the version without that button (like Renault).
## Screenshots

Renault CTA:
screen shot 2016-07-11 at 12 20 07

Leffe CTA:
screen shot 2016-07-11 at 12 20 12
## Request for comment

@lps88


lps88:
👍

PR #13556, opened 2016-07-11
use correct omniture code on embed pages

akash1810:
## What does this change?

We're currently tracking embed pages to the `guardiangu-network` bucket when it should be the `guardiangu-thirdpartyapps` bucket.

This corrects that by pattern matching on `Page` and if `EmbedPage` use the correct code, else perform existing logic.

Correct omniture tracking.

No.
## Screenshots

n/a
## Request for comment

@jamesgorrie @mkopka

### DIFF
--- a/applications/app/views/videoEmbed.scala.html
+++ b/applications/app/views/videoEmbed.scala.html
@@ -1,4 +1,4 @@
-@(page: EmbedPage)(implicit request: RequestHeader)
+@(page: model.EmbedPage)(implicit request: RequestHeader)
 
 @import model.{Video, VideoPlayer}
 @import views.support.Video640
--- a/common/app/views/support/OmnitureAnalyticsData.scala
+++ b/common/app/views/support/OmnitureAnalyticsData.scala
@@ -3,20 +3,26 @@ package views.support
 import java.net.URLEncoder._
 
 import conf.Configuration
-import model.{ContentPage, MetaData, Page}
+import model.{ContentPage, MetaData, Page, EmbedPage}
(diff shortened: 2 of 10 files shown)


common/app/views/support/OmnitureAnalyticsData.scala
```diff
@@ -3,20 +3,26 @@ package views.support
...
+import model.{ContentPage, MetaData, Page, EmbedPage}
 import play.api.libs.json.{JsString, JsValue, Json}
 import play.api.mvc.RequestHeader
 import play.twirl.api.Html
 
+
 object OmnitureAnalyticsAccount {
-  def apply(page: MetaData): String = {
-    val sectionSpecficAccounts = Map(
-      ("guardian-masterclasses", "guardiangu-masterclasses"),
```
akash1810: these should really be config properties too...

applications/app/views/videoEmbed.scala.html
```diff
@@ -1,4 +1,4 @@
-@(page: EmbedPage)(implicit request: RequestHeader)
+@(page: model.EmbedPage)(implicit request: RequestHeader)
```
jamesgorrie: Why?

jamesgorrie:
👍

applications/app/views/videoEmbed.scala.html
akash1810: moving `EmbedPage` into `meta`? Following the current pattern of objects that extend `Page` living there e.g
 https://github.com/guardian/frontend/blob/master/common/app/model/meta.scala#L437

PR #13558, opened 2016-07-11
New header menu

NataliaLKB:
## What does this change?

This is the first PR to introduce the proposed new nav menu. I have left lots of comments for myself in the code for me to iterate on. This is by no means a finished product, but simply a starting point. 

To see this header in PROD go to:
https://www.theguardian.com/uk/opt/in/header

Please note the header only shows on Mobile at the moment.
## What needs to be done in subsequent PRs
- [ ] top level menu items should close when the another opens
- [ ] "amp five" links should open up the menu
- [ ] Need a polyfill/fallback for the `details` element: http://caniuse.com/#search=details
- [ ] editionalisation (?)
- [ ] @zeftilldeath to tweak styling 😉 
- [ ] Need "core" / no JS version of the nav

Not yet 😄 
## Screenshots

navopening
## Request for comment

Thank you @samdesborough for helping with the Scala!!!!!! 👏


common/app/common/NewNavigation.scala
```diff
@@ -0,0 +1,83 @@
+package common
+
+sealed class TopLevelSection(val name: String)
+
+case object News extends TopLevelSection("news")
+case object Opinion extends TopLevelSection("opinion")
+case object Sports extends TopLevelSection("sports")
+case object Arts extends TopLevelSection("arts")
+case object Life extends TopLevelSection("life")
```
TBonnin: nitpick: I usually like to wrap data type objects/classes like above in an object in order to make the structure and relation between them clearer. 
It also prevents congesting the namespace (especially in common package). Here objects like `News` or `Opinion` for instance are introduced and accessible from everywhere in common package. Wrapping them together would add context.
```
object TopLevelSection {
    case object News extends TopLevelSection("news")
    case object Opinion extends TopLevelSection("opinion")
    case object Sports extends TopLevelSection("sports")
    case object Arts extends TopLevelSection("arts")
    case object Life extends TopLevelSection("life")
}
```

common/app/views/fragments/nav/newHeaderMenu.scala.html
```diff
@@ -0,0 +1,67 @@
+@import common.NewNavigation
+
+@import common.TopLevelSection
+
+@sectionList(topLevelSection: TopLevelSection) = {
+    <li class="main-navigation__item navigation-border">
+            <!-- TODO: Firefox -->
+        <details>
```
regiskuckaertz: Kudos for using `DETAILS` 😃 

Are you polyfilling its behaviour somewhere?

common/app/views/fragments/nav/newHeaderMenu.scala.html
```diff
@@ -0,0 +1,67 @@
...
+            Sign in/up
+        </div>
+
+        <ul class="main-navigation">
+            @NewNavigation.topLevelSections.map { section =>
+                @sectionList(section)
+            }
+        </ul>
+        <ul class="navigation-group navigation-border">
+            <li class="navigation-group__item"><a href="#" tabindex="0">Jobs</a></li>
```
regiskuckaertz: Why are you adding `tabindex` here? By default, `A` is focusable and counted in the tab order. The only use case that I know of would be if you want to disable tab order but want programmatic focus with `tabindex="-1"`

static/src/javascripts/projects/common/modules/navigation/newHeaderNavigation.js
```diff
@@ -0,0 +1,77 @@
...
+            burgerLink.attr('href', mainMenuId);
+
+            // TODO: Support browsers that don't have transitions
+            // We still want to hide this
+            if (mainMenuEl.length > 0) {
+                mainMenuEl[0].addEventListener('transitionend', function handler() {
+                    return fastdomPromise.write(function () {
+                        mainMenuEl.removeClass('off-screen');
+                        mainMenuEl.removeClass('shown');
+                        mainMenuEl[0].removeEventListener('transitionend', handler);
```
regiskuckaertz: You can actually move that statement before the `fastdom` call 😄

regiskuckaertz:
Great! Looking forward to seeing it live.

common/app/common/NewNavigation.scala
NataliaLKB: @TBonnin yes that makes sense. So in the end it would look like this: 

```

sealed class TopLevelSection(val name: String)

object TopLevelSection {
  case object News extends TopLevelSection("news")
  case object Opinion extends TopLevelSection("opinion")
  case object Sports extends TopLevelSection("sports")
  case object Arts extends TopLevelSection("arts")
  case object Life extends TopLevelSection("life")
}
```

common/app/views/fragments/nav/newHeaderMenu.scala.html
NataliaLKB: No yet! That will have to happen in a separate PR. I will mention it in this PR's description.

common/app/common/NewNavigation.scala
TBonnin: Yep. And you access them like `TopLevelSection.News`, ...

NataliaLKB: Thanks! All updated

common/app/views/fragments/nav/newHeaderMenu.scala.html
NataliaLKB: good catch! My mistake. All updated now!

common/app/views/fragments/nav/newHeaderMenu.scala.html
```diff
@@ -0,0 +1,64 @@
...
+                    </li>
+                }
+            </ul>
+        </details>
+    </li>
+}
+
+<div class="main-menu-container" id="main-menu">
+        <!-- TODO: Provide tabbable alternative -->
+    <a href="#" class="main-menu-container__overlay"></a>
```
regiskuckaertz: can't you use a `DIV` instead?

static/src/javascripts/projects/common/modules/navigation/newHeaderNavigation.js
```diff
@@ -0,0 +1,80 @@
...
+            return fastdomPromise.write(function () {
+                mainMenuEl.removeClass('off-screen');
+            });
+        }).then(function () {
+            var firstButton = $('.main-navigation__item__button')[0];
+            if (firstButton) {
+                firstButton.focus();
+            }
+            // Prevents scrolling on the body
+            html.css('overflow', 'hidden');
```
regiskuckaertz: Should probably be wrapped in `fastdom`?

static/src/javascripts/projects/common/modules/navigation/newHeaderNavigation.js
```diff
@@ -0,0 +1,80 @@
...
+
+                    mainMenuEl[0].removeEventListener('transitionend', handler);
+
+                    return fastdomPromise.write(function () {
+                        mainMenuEl.removeClass('off-screen');
+                        mainMenuEl.removeClass('shown');
+                    }).then(function() {
+                        $('.new-header__nav__menu-button').focus();
+                        // Users should be able to scroll again
+                        html.css('overflow', '');
```
regiskuckaertz: ditto

common/app/views/fragments/nav/newHeaderMenu.scala.html
regiskuckaertz: scratch that. I see why 😃  That's elegant. Because it is a visual feature, I would then add the `aria-hidden="true"` attribute.

regiskuckaertz:
Minor comments but it all looks good to me 👍

PR #13559, opened 2016-07-11
business -  use guardian sans-serif font

mchv:
## What does this change?

Use guardian `sans-serif` font to not break design instead of the custom one.
See #13553


static/src/stylesheets/module/business/_stocks.scss
```diff
@@ -105,7 +105,7 @@ $stocks-icon-height: 10px;
 
 .stocks__price {
     margin-right: .25em;
-    font-family: sans-serif;
+    font-family: $f-sans-serif-text;
     letter-spacing: .05em;
```
sndrs: i think was added before? it's also non-standard

PR #13560, opened 2016-07-11
Revert "Use a sans-serif font for business stocks price"

mchv:
Reverts guardian/frontend#13553

This is not consistent with guardian style, so reverting.


sndrs:
👍

Issue #12044, opened 2016-02-19
Linking to comments is horrible user experience because page jank


michaelwmcnamara:
On mobile devices, when we link directly to the comments section - ie. as we do from an Amp page when the user selects to view comments. We are taken to a perfectly positioned view of the comments.
A few seconds later our view is jumped back into the article body.
This is incredibly awkward and looks borked!

The culprit for this is items things being slotted into the page above the comments.
Either these should not be loaded in,  or the page should know ahead of time how much space they will need.

## Steps to Reproduce
On a mobile device,
Load an Amp page
Scroll to the bottom
Select to view comments

or

On a mobile device, 
log in (or be already logged in)
open user profile page
select comments
select "view discussion" for a comment made on a long-ish article (to maximise jump effect)

Result:
You are first taken to comments, and are then jumped away - leaving you lost and alone.

![jankycomment](https://cloud.githubusercontent.com/assets/7239344/13183681/1414dd9c-d730-11e5-8383-64991db113bf.gif)

michaelwmcnamara:
Note while I am using Amp as an example, this is a problem with the www site.

michaelwmcnamara:
This is still an issue - given the excitement over [that comment](https://www.theguardian.com/politics/live/2016/jun/25/brexit-live-emergency-meetings-eu-uk-leave-vote#comment-77205935)
and its coverage in multiple outlets:
http://metro.co.uk/2016/06/26/someones-summed-up-why-boris-johnson-is-screwed-5967397/
http://www.huffingtonpost.co.uk/entry/guardian-comment-boris-johnson_uk_576faf9be4b0d2571149c8b9
http://qz.com/717182/a-brexit-conspiracy-theory-nails-the-no-win-situation-boris-johnson-now-finds-himself-in/
http://www.thepoke.co.uk/2016/06/30/amazing-theory-cameron-snookering-boris-turned-true-didnt/

We are missing an opportunity by not making this work correctly.

michaelwmcnamara:
part of a larger piece of work. Closing to remove redundancy

PR #13562, opened 2016-07-12
Calm the jumpy comments on page load

gtrufitt:
## What does this change?

This PR adds a utility for loading elements above the current scroll position by measuring the inserted height and adjusting the scroll position to avoid jolting the user down the page. This is the same functionality as is used in the sticky header.

I've initially applied this to the two biggest offenders - outbrain and adverts (on mobile), also applying a min-height to outbrain so we know the height before images are loaded.

There are a few things that it doesn't fix such as the insertion of rich links which re-flow content so cannot be measured directly by their inserted height.

A discussion on how this could be done would be welcome - my thoughts would be using an 'anchoring' element such as the submeta that measures the distance moved and adjusts the scroll position accordingly.

This aims to go some way to helping issue #12044

_Note:_ [Scroll anchoring ](https://developers.google.com/web/updates/2016/04/scroll-anchoring?hl=en)is in beta in Chrome which would fix some of these issues. I need to test whether this would then break this technique.

This is one of the major UX pains currently at this begins to look at fixing that.
## Screenshots

As you can see, there is more work to do, but the comments are _still on the screen_ when jumping to them!
### BEFORE

jumping comments - before
### AFTER

jumping comments - after - fast

| Before | After |
| :-: | :-: |
| jumping comments -mob - actually before | jumping comments -mob - before |

| Before | After |
| :-: | :-: |
| jumping comments - comment block - mob - before | jumping comments - comment block - mob - after |
## Request for comment

@sndrs @regiskuckaertz @stephanfowler @crifmulholland  @JonNorman @rich-nguyen  @NataliaLKB @desbo @michaelmcnamara @ScottPainterGNM 
## Todo
- [x]  Write tests
- [x]  Finish writing comments and updating jsdoc
- [x]  Move new queue utility and write tests
- [x]  Add switch that passes through and immediately call callback
- [x]  Fix broken ad tests


static/src/javascripts/projects/common/utils/steady-page.js
```diff
@@ -0,0 +1,115 @@
...
+    };
+
+    /*
+     * Height with margins
+     */
+    function outerHeight(el) {
+        var height = el.offsetHeight;
+        var style = getComputedStyle(el);
+
+        height += parseInt(style.marginTop) + parseInt(style.marginBottom);
```
desbo: easy to avoid this bit of mutation by just returning `el.offsetHeight + ...`

static/src/javascripts/projects/common/utils/steady-page.js
```diff
@@ -0,0 +1,115 @@
...
+
+            // Empty the container array once the loop is done
+            currentFrameContainers = [];
+        });
+    }
+
+    /*
+     * The callback should contain the functionality
+     * to display the element (insertion, change of display).
+     */
```
desbo: I don't think this comment should be here

static/src/javascripts/projects/common/utils/steady-page.js
```diff
@@ -0,0 +1,115 @@
...
+        });
+    }
+
+    /*
+     * Add the height of each inserted element within this animation frame
+     * together so we know what the total height change required for
+     * the scroll adjustment is
+     */
+    function getHeightAdjustment () {
+        return fastdom.read(function(){
```
desbo: does this need to be wrapped in another `fastdom.read` when its caller is already inside one?

static/src/javascripts/projects/common/utils/steady-page.js
```diff
@@ -0,0 +1,115 @@
...
+     * @param {HTMLElement} insertionContainer The element that the component is being inserted into
+     * @param {Function} insertionCallback Should contain all functionality that displays and lays-out the element
+     */
+    function insert (insertionContainer, insertionCallback) {
+        return fastdom.read(function() {
+
+            currentScrollPos = window.scrollY;
+
+            // Push all the containers in the current fastdom read frame
+            currentFrameContainers.push(insertionContainer);
```
desbo: I'd try to avoid this approach of adding to the global array and clearing it in the other function.

Can you pass `insertionContainer` to `getHeightAdjustment` and only have it operate on its parameter, and have it return a height instead of altering `heightToAdd`?

I think we should avoid mutating global variables as much as possible here since fastdom might cause things to happen in an unpredictable order.

static/src/javascripts/projects/common/utils/steady-page.js
gtrufitt: It needs to be read -> write -> read -> write because we can't read the heights until after we've inserted (and the initial read is to get the previous scroll position so it needs to be before the write)

\- *Read* the current window scroll position
-- *Write* the elements into the DOM
--- *Read* the heights
---- *Write* the new scroll position

static/src/javascripts/projects/common/modules/commercial/article-body-adverts.js
```diff
@@ -108,7 +110,16 @@ define([
 
     function insertAdAtPara(para, name, type) {
         var ad = createSlot(name, type);
-        para.parentNode.insertBefore(ad, para);
+        if (detect.isBreakpoint({max: 'tablet'})) {
+            // Insert ad using steady page
+            // To avoid jumping the user
+            steadyPage.insert(ad, function() {
```
gtrufitt: @regiskuckaertz Thinking about this more, I think I'm messing up the use of space-filler to insert the ad - i.e: Spacefiller handles the queuing of element insertion, then uses fastdom write to insert which calls this, which then does the read, write part. Spacefiller specifically states not to use fastdom for the writer, which is precisely what I'm doing. Maybe the utility needs to play nicer with spacefiller to avoid jumping animation frames. Thoughts on the approach here?

regiskuckaertz: Hmm. We could abstract in spacefiller. The "DOM writer" would be fastdom by default but we could override it with one of our own at the calling site. Would need to provide a `write` function thought. WDYT?

johnduffell:
this is fantastic!! well done 🎉 

>insertion of rich links which re-flow content so cannot be measured directly by their inserted height

Regarding that, the submeta would be a good improvement over the current state, but for a more accurate but expensive approach consider measuring the following n elements, and finding the subset that change height and steadying any view below the bottom of those.

e.g. this case, the first rich link only affects one paragraph, but you can see how far below the submeta is https://www.theguardian.com/world/2016/jul/12/philippines-wins-south-china-sea-case-against-china

-----

image

paperboyo:
This deserves a separate :gem:!

static/src/javascripts/projects/common/utils/steady-page.js
```diff
@@ -0,0 +1,115 @@
...
+    var heightToAdd = 0;
+    var options = {
+        pixelGrace: 100 // Give the scroll some distance as it's likely the user has scrolled past the top of the screen
+    };
+
+    /*
+     * Height with margins
+     */
+    function outerHeight(el) {
+        var height = el.offsetHeight;
```
sndrs: does this need to be done in `fastdom.read`?

sndrs:
got any tests guv?

static/src/javascripts/projects/common/utils/steady-page.js
sndrs: that's not really clear, is there any advantage to that? is it any worse to wrap the specific code that needs fastdoming so it's explicit?

static/src/javascripts/projects/common/utils/steady-page.js
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    /*
+     * The callback should contain the functionality
+     * to display the element (insertion, change of display).
+     */
+    function displayAndLayout (insertionCallback) {
+        return fastdom.write(function(){
+            insertionCallback();
+        });
```
sndrs: `fastdom.write(insertionCallback);`?

static/src/javascripts/bootstraps/enhanced/trail.js
```diff
@@ -102,8 +102,13 @@ define([
 
     function repositionComments() {
         if (!identityApi.isUserLoggedIn()) {
+            var $comments = $('.js-comments');
```
jamesgorrie: When you assign a val to a DOM element - do you need to use `fastdom.read`?
Or is that only if you read an attribute / property of it?

jamesgorrie: @sndrs ^

static/src/javascripts/projects/common/modules/commercial/third-party-tags/outbrain.js
```diff
@@ -72,7 +74,8 @@ define([
             breakpoint: breakpoint
         });
         widgetHtml = build(widgetCodes, breakpoint);
-        return fastdom.write(function () {
+
+        return steadyPage.insert($container[0], function() {
```
jamesgorrie: Is `$container[0]` 100% always there?

static/src/javascripts/bootstraps/enhanced/trail.js
sndrs: yeah only for anything that will/could trigger layout. getting a reference to dom object is ok

static/src/javascripts/projects/common/utils/detect.js
```diff
@@ -256,6 +256,14 @@ define([
...
     }
 
+    /**
+     *     Usage:
+     *     detect.isBreakpoint({min: 'tablet', max: 'leftCol'}) // Will return true for tablet, desktop, leftCol
+     *     detect.isBreakpoint({min: 'tablet'}) // Will return true for tablet, desktop, leftCol, wide
+     *     detect.isBreakpoint({max: 'tablet'}) // Will return true for mobile, mobileLandscape, tablet and phablet
+     *
+     *     
+     */
```
jamesgorrie: If only we had types... 🙍

static/src/javascripts/projects/common/utils/steady-page.js
jamesgorrie: Agreed with @sndrs - this would require you to know that this lib reads the DOM.

static/src/javascripts/projects/common/utils/steady-page.js
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    /*
+     * Add the height of each inserted element within this animation frame
+     * together so we know what the total height change required for
+     * the scroll adjustment is
+     */
+    function getHeightAdjustment () {
+        return fastdom.read(function(){
+            for (var i = 0; i < currentFrameContainers.length; i++) {
```
jamesgorrie: `forEach`?

static/src/stylesheets/module/facia/_containers.scss
```diff
@@ -42,6 +42,20 @@
...
     background-color: $neutral-8;
     display: none;
+
+    .fc-container__inner {
+        @include mq($from: mobile) {
+            min-height: 621px;
+        }
+
+        @include mq($from: tablet) {
+            min-height: 455.17px;
```
jamesgorrie: Magic numbers? Why the `.17`?

gtrufitt: I'll add a comment but the reason is that this is the natural height of a fully-loaded (inc images) outbrain widget at that breakpoint - I'm just ensuring that the widget already has that height when it's inserted into the page.

johnduffell:
this would be fantastic for live blog auto updates, there's a gigantic jump when you scroll up when updates are waiting. Once you've shipped this I'm going to be all over it!

static/src/javascripts/projects/common/utils/Queue.js
```diff
@@ -0,0 +1,42 @@
+define([], function (
```
sndrs: does this need to be abstracted?

regiskuckaertz: We use it in several places (space filler, prebid)

static/src/javascripts/projects/common/utils/steady-page.js
```diff
@@ -0,0 +1,193 @@
...
+                while (!q.empty()) {
+                    // Take the current queue items and add them to the batch array
+                    var insertion = q.dequeue();
+                    batch.push(insertion);
+                }
+
+                return batch;
+            })
+            .then(function(){
+                getHeightOfAllContainers(batch);
```
desbo: missing `return`

static/src/javascripts/projects/common/utils/steady-page.js
```diff
@@ -0,0 +1,193 @@
...
+     */
+    function calculateScrollY (state) {
+        if (q.empty()) {
+            // If the queue is empty (no more elements need to be added to the page) we immediately scroll
+            return state.newHeight + state.prevHeight + state.scrollY;
+        } else {
+            // If there are elements waiting to be added to the page we take the previous container's heights
+            // and recursively call the function so that we only scroll the page once the queue is empty -
+            // this prevents excessive and jarring scrolling
+            return go(assign(state, {
```
desbo: I don't know if moving this function out of the promise chain made things any clearer, I think the recursive call to `go` again should be inside `go`

static/src/javascripts/projects/common/utils/Queue.js
desbo: helps with encapsulation as well. in the steady page example we're working on a queue with defined operations instead of messing about with an array declared in the module

static/src/javascripts/projects/common/utils/steady-page.js
gtrufitt: Ya, agreed, have moved back in

static/src/javascripts/projects/common/utils/steady-page.js
gtrufitt: :/ Fixed

gtrufitt:
I've made all the changes and updates so could do with some 👀 on it again. Hopefully just the one test left to fix 😓

static/src/javascripts/projects/common/modules/commercial/article-body-adverts.js
```diff
@@ -116,7 +120,15 @@ define([
...
-        para.parentNode.insertBefore(ad, para);
+
+        function insertion () {
+            para.parentNode.insertBefore(ad, para);
+        }
+
+        // If on mobile we will
+        // insert ad using steady page
+        // to avoid jumping the user
+        detect.isBreakpoint({max: 'tablet'}) ? steadyPage.insert(ad, insertion): insertion();
```
sndrs: i'd go with an `if()` here i think?

static/src/javascripts/projects/common/utils/steady-page.js
```diff
@@ -0,0 +1,188 @@
...
+ inserted element.
+ Note: Inserted elements must have a fixed height when rendered
+ (i.e: no changing the height by loading an image, lazy-loading content etc)
+ */
+
+define([
+    'common/utils/fastdom-promise',
+    'Promise',
+    'lodash/objects/assign',
+    'common/utils/Queue',
```
desbo: even though this is a constructor I think we should only use lowercase for consistency. [require is case sensitive](https://github.com/requirejs/requirejs/issues/270) so let's not add a variation for this one module. I'd make the filename lowercase as well.

static/src/javascripts/projects/common/utils/steady-page.js
```diff
@@ -0,0 +1,188 @@
...
+
+    /**
+     * Process the insertion operation:
+     *
+     *   1. Calculate the original height of the container
+     *   2. Apply the insertion functions for all inserted elements
+     *   3. Calculate the new height of the container
+     *   4. Adjust the scroll position to account for the new container height
+     *
+     * @param  {Number} state.scrollY The original scroll position
```
desbo: this isn't just a number now

static/src/javascripts/projects/common/utils/steady-page.js
```diff
@@ -0,0 +1,188 @@
...
+    var running = false;
+    var promise;
+
+    /**
+     * Given a batch and a previous currentBatchHeight, measure the height of each container
+     * in the batch
+     *
+     * @param  {Array} batch
+     */
+    function getHeightOfAllContainers (batch) {
```
desbo: I'd move `insert` and `go` to the top of the file for readability, and put this one at the end.

static/src/javascripts/projects/common/utils/steady-page.js
```diff
@@ -0,0 +1,188 @@
...
+    function go(state) {
+        running = true;
+
+        var batch = [];
+        var batchHeightsBeforeInsert;
+
+        promise = fastdom.read(function(){
+                while (!q.empty()) {
+                    // Take the current queue items and add them to the batch array
+                    var insertion = q.dequeue();
```
desbo: no need for this var

regiskuckaertz: unrelated but `dequeue` means double-ended queue, this method should probably be called `pop` or `shift`.

desbo: `dequeue` as an operation is a different thing to a double ended queue (I think some people call them `deque`s). 
there's plenty of examples of it out there, e.g. [scala](http://www.scala-lang.org/api/2.7.7/scala/collection/mutable/Queue.html#dequeue%28%29), [wikipedia](https://en.wikipedia.org/wiki/Queue_(abstract_data_type))

static/src/javascripts/projects/common/utils/Queue.js
regiskuckaertz: (Actually we don't use that one, we use `QueueAsync`)

static/src/javascripts/projects/common/utils/steady-page.js
gtrufitt: Hmm, it should be, why do you say that?

static/src/javascripts/projects/common/utils/steady-page.js
gtrufitt: Possibly, but I am not sure about abstracting before it's proven it's required (even though I kinda did that with the queue util). If it proves we need a util at some point, then we can abstract?

static/src/javascripts/projects/common/utils/steady-page.js
```diff
@@ -0,0 +1,187 @@
...
+            })
+            .then(function(state){
+                if (q.empty()) {
+                    // If the queue is empty (no more elements need to be added to the page) we immediately scroll
+                    return state.newHeight + state.prevHeight + state.scrollY;
+                } else {
+                    // If there are elements waiting to be added to the page we take the previous container's heights
+                    // and recursively call the function so that we only scroll the page once the queue is empty -
+                    // this prevents excessive and jarring scrolling
+                    return go(assign(state, {
```
piuccio: I've got the feeling that this recursion is not tested in your specs. I might be wrong 😆

gtrufitt: Yeah, you're right actually! Thanks

static/test/javascripts/spec/common/utils/steady-page.spec.js
```diff
@@ -0,0 +1,264 @@
...
+    bonzo,
+    $,
+    steadyPage,
+    fixtures,
+    Promise,
+    Injector,
+    fastdom
+) {
+
+    fdescribe('Steady Page Utility', function () {
```
piuccio: forgot the `f` 😉

gtrufitt: Oh crap. Need to add a linter for that!

gtrufitt:
Would ❤️ to get this in soon if anyone can take a look?

static/src/javascripts/bootstraps/enhanced/trail.js
regiskuckaertz: The definitive references would be [this](https://gist.github.com/paulirish/5d52fb081b3570c81e3a) and [that](https://csstriggers.com/)

static/src/javascripts/projects/common/utils/steady-page.js
```diff
@@ -0,0 +1,187 @@
...
+ moving the current scroll position by the same distance as the height of the
+ inserted element.
+ Note: Inserted elements must have a fixed height when rendered
+ (i.e: no changing the height by loading an image, lazy-loading content etc)
+ */
+
+define([
+    'common/utils/fastdom-promise',
+    'Promise',
+    'lodash/objects/assign',
```
regiskuckaertz: Better to use our own polyfill which falls back to `Object.assign` if available

static/src/javascripts/projects/common/utils/steady-page.js
```diff
@@ -0,0 +1,187 @@
...
+     *
+     * @param  {Number} state.scrollY The original scroll position
+     */
+    function go(state) {
+        running = true;
+
+        var batch = [];
+        var batchHeightsBeforeInsert;
+
+        promise = fastdom.read(function(){
```
regiskuckaertz: As far as I can see, there is no DOM-related ops in this routine, so you could create a `Promise` instead of calling `fastdom.read`, no?

static/src/javascripts/projects/common/utils/steady-page.js
```diff
@@ -0,0 +1,187 @@
...
+            // removing the current height
+            return batch.filter(elementIsAbove).reduce(function(height, insertion) {
+                return height + readHeight(insertion.container);
+            }, 0);
+        });
+
+        function elementIsAbove(el) {
+            var parentElArray = [];
+            var parentEl = el.container;
+            var viewportAdjustment = Math.max(document.documentElement.clientHeight, window.innerHeight || 0) / 2;
```
regiskuckaertz: You can move this line into the `fastdom.read` callback so it is computed only once.

static/src/javascripts/projects/common/utils/steady-page.js
```diff
@@ -0,0 +1,187 @@
...
+            // Push the container's parents into an array so we can calculate
+            // the position of the elements all the way until the body
+            while (parentEl) {
+                parentElArray.push(parentEl);
+                parentEl = parentEl.offsetParent;
+            }
+
+            // Get the top position of the element + half of the viewport (to account for an element loading
+            // just below the top of the viewport)
+            var elTopPos = parentElArray.reduce(function(topPos, parentEl){
```
regiskuckaertz: Out of curiosity, if you want the top position of this element related to the viewport, why don't you just:

`var elTopPost = el.getBoundingClientRect().top - viewportAdjustment`

?

static/src/javascripts/projects/common/utils/steady-page.js
```diff
@@ -0,0 +1,187 @@
...
+            batch.forEach(function (insertion) {
+                insertion.cb();
+            });
+        });
+    }
+
+    function scrollThePage (scrollY) {
+        if (scrollY) {
+            window.scrollTo(0, scrollY);
+            running = false;
```
regiskuckaertz: I think this should be moved into the `go` routine, since there is a chance a task might be queued *after* the go routine has finished and *before* this callback is called, and then it would never get picked up. What do you think?

static/src/javascripts/projects/common/utils/steady-page.js
```diff
@@ -0,0 +1,187 @@
...
+            })
+            .then(function(){
+                return getHeightOfAllContainers(batch).then(function(newHeights) {
+                    return assign(state, {
+                        newHeight: newHeights - batchHeightsBeforeInsert
+                    });
+                });
+            })
+            .then(function(state){
+                if (q.empty()) {
```
regiskuckaertz: I think this is where you should the `running` flag to `false`

static/src/javascripts/projects/common/utils/steady-page.js
gtrufitt: I just love complexity 💥 Seriously though, for some reason I thought I needed to get the element from the top of the page, but yes, I only care about the viewport. Doh. Updated.

static/src/javascripts/projects/common/utils/steady-page.js
gtrufitt: Updated!

static/src/javascripts/projects/common/utils/steady-page.js
gtrufitt: Good shout 👍

gtrufitt:
Ta @regiskuckaertz, have made those changes

static/src/javascripts/projects/common/utils/steady-page.js
```diff
@@ -0,0 +1,171 @@
...
+        return fastdom.read(function() {
+            viewportHeight = Math.max(document.documentElement.clientHeight, window.innerHeight || 0);
+            // Add all the heights of the passed in batch
+            // removing the current height
+            return batch.filter(elementIsAbove).reduce(function(height, insertion) {
+                return height + readHeight(insertion.container);
+            }, 0);
+        });
+
+        function elementIsAbove(el) {
```
regiskuckaertz: Nice 😃

regiskuckaertz:
I'll be trying this on my phone tonight 👍

gtrufitt:
Gonna merge first thing as don't want 24/7 shouting at me because I broke ads or something 😆

static/src/javascripts/projects/common/utils/steady-page.js
desbo: the `state` param is an `object`, you can use [this syntax](http://usejsdoc.org/tags-property.html) to document properties

PR #13563, opened 2016-07-12
Perform ophan config assignment in blocking js

rich-nguyen:
I've moved the assignment of the ophan config object, so that the `pageViewId` field is definitely available.

I think there's a wider question about the timing behaviour of our app, using the polyfill async approach. It's promised on dom ready, but it appears like inline scripts were not fully executed...


sndrs:
👍

Issue #13286, opened 2016-06-14
Greedy CSS rule


regiskuckaertz:
# ISSUE

## Steps to Reproduce

The `.from-content-api h2` rule has side-effects in components it should not touch, one of which is [the inline merchandise](http://m.code.dev-theguardian.com/politics/2016/apr/12/dont-judge-politicians-by-their-tax-returns-says-william-hague?google_preview=hLruxlaTqk4YgLv_ugUwgNe0wgWIAYCAgKCzqKnqbQ&iu=59666047&gdfp_req=1&lineItemId=123216807&creativeId=93009006567).

I suggest using the child selector instead, `.from-content-api > h2`.

## Actual Results (include screenshots)

![picture 2](https://cloud.githubusercontent.com/assets/629976/16039856/340002d6-3224-11e6-9d3a-3a2f0969ea45.jpg)

## Expected Results (include screenshots)

![picture 3](https://cloud.githubusercontent.com/assets/629976/16039872/50960f62-3224-11e6-83cb-eea875b2a35b.jpg)

sndrs:
child selector massively ties it the to the underlying HTML structure. ideally, you could move the logic to the template and give it a unique class

regiskuckaertz:
The thing is `.from-content-api` is added to the `DIV` surrounding all the
content, giving it a large scope. As a consequence, shouldn't it take care
not to target unwanted elements?

On Tue, 14 Jun 2016 at 11:53 Alex Sanders <notifications@github.com> wrote:

> child selector massively ties it the to the underlying HTML structure.
> ideally, you could move the logic to the template and give it a unique class
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/guardian/frontend/issues/13286#issuecomment-225846893>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAmc2OfY4P0NIFWgsIChuEiTge-bGR_jks5qLogdgaJpZM4I1LzW>
> .
>

-- 

------------------------------
This e-mail and all attachments are confidential and may also be 
privileged. If you are not the named recipient, please notify the sender 
and delete the e-mail and all attachments immediately. Do not disclose the 
contents to another person. You may not use the information for any 
purpose, or store, or copy, it in any way.  Guardian News & Media Limited 
is not liable for any computer viruses or other material transmitted with 
or as part of this e-mail. You should employ virus checking software.
 
[...]

PR #13564, opened 2016-07-12
Refactor inline components to use the new styles

JonNorman:
## What does this change?

Reworking the last two inline components - books and live events - to use the new styles.

We can get rid of a load of the old templates and styles.

No
## Screenshots

No point - looks the same.


regiskuckaertz:
Let's not deploy this until #13286 is fixed.

rich-nguyen:
hopefully fixed now

JonNorman:
Yep, just checked:

image
image

regiskuckaertz:
Yep, I just added some emphasis in live events when there's not much space left. I would give you a +1 but as we worked on this together, let's have someone else review it.

kelvin-chappell:
👍

PR #13565, opened 2016-07-12
Catch errors reported by IE due to Iframe

Calum-Campbell:
## What does this change?

Catches the errors being thrown by IE and reported to Sentry. This is du to IE dealing with iFrames differently do other browsers. 

Should see a reduction in issue  :  https://app.getsentry.com/the-guardian/client-side-prod/issues/141001393/
## Request for comment


rich-nguyen:
👍

PR #13566, opened 2016-07-12
Don't request all the blocks for liveblog main page

johnduffell:
This PR changes the CAPI query for liveblog page 1 (the live page) and the ajax "get updates" call to only retrieve the necessary blocks, rather than the whole liveblog.

This should save a lot of bandwidth as liveblogs get bigger, and also reduce the CPU usage both in CAPI and frontend.  We noticed things were particularly bad when there were many big liveblogs running after brexit.  This is a problem particularly for those two endpoints because they only have a 5 second cache time.  Archive pages of a live blog cache for a few minutes.

This is part 2 of the series - part 1 was https://github.com/guardian/frontend/pull/13458

@TBonnin @cb372 @JustinPinner any comments or +1s please!


article/app/controllers/ArticleController.scala
```diff
@@ -234,15 +250,25 @@ class ArticleController extends Controller with RendersItemResponse with Logging
...
-  def apply(input: String): Option[String] = {
-    def withParser: Parser[Unit] = "with:" ^^ { _ => () }
-    def block: Parser[Unit] = "block-" ^^ { _ => () }
-    def id: Parser[String] = "[a-zA-Z0-9]+".r
-    def expr: Parser[String] = withParser ~> block ~> id
+
+  def withParser: Parser[Unit] = "with:" ^^ { _ => () }
+  def block: Parser[Unit] = "block-" ^^ { _ => () }
+  def id: Parser[String] = "[a-zA-Z0-9]+".r
+  def blockId = block ~> id
```
TBonnin: Would it make sense to make those private if they are not supposed to be exposed outside of this object?

article/app/controllers/ArticleController.scala
```diff
@@ -234,15 +250,25 @@ class ArticleController extends Controller with RendersItemResponse with Logging
...
       case _ => None
     }
   }
+
+  def fromBlockId(input: String): Option[SinceBlockId] = {
+    parse(blockId, input) match {
+      case Success(matched, _) => Some(SinceBlockId(matched))
+      case _ => None
+    }
+  }
```
TBonnin: `fromBlockId` and `fromPageParam` have similar name and input param but different output types. Can we make them have a similar and consistant signature so when using either of them you don't have to think to much or transform the output?

article/app/controllers/ArticleController.scala
```diff
@@ -32,11 +32,15 @@ class ArticleController extends Controller with RendersItemResponse with Logging
...
 
 
-  private def renderNewerUpdates(page: PageWithStoryPackage, lastUpdateBlockId: String, isLivePage: Option[Boolean])(implicit request: RequestHeader): Result = {
-    val newBlocks = page.article.fields.blocks.takeWhile(block => s"block-${block.id}" != lastUpdateBlockId)
+  private def renderNewerUpdates(page: PageWithStoryPackage, lastUpdateBlockId: SinceBlockId, isLivePage: Option[Boolean])(implicit request: RequestHeader): Result = {
+    val newBlocks = page.article.fields.blocks.toSeq.flatMap {
+      _.requestedBodyBlocks.getOrElse(lastUpdateBlockId.around, Seq())
+    }.takeWhile { block =>
+      block.id != lastUpdateBlockId.lastUpdate
+    }
```
TBonnin: I assume blocks are always sorted from the most recent. right?

article/app/controllers/ArticleController.scala
```diff
@@ -70,10 +74,10 @@ class ArticleController extends Controller with RendersItemResponse with Logging
 
   private def blockText(page: PageWithStoryPackage, number: Int)(implicit request: RequestHeader): Result = page match {
     case LiveBlogPage(liveBlog, _, _) =>
-      val blocks = liveBlog.blocks.collect {
+      val blocks = liveBlog.blocks.toSeq.flatMap{_.requestedBodyBlocks.get(Canonical.firstPage).toSeq.flatMap(_.collect {
```
TBonnin: A lot of things happen on this line. Would it make sense to break it down in multiple variables (or multiple lines with comment) so it is more self-explanatory

article/app/controllers/ArticleController.scala
```diff
@@ -117,16 +121,27 @@ class ArticleController extends Controller with RendersItemResponse with Logging
 
   def renderLiveBlog(path: String, page: Option[String] = None) =
     Action.async { implicit request =>
-      mapModel(path, blocks = true, page) {// temporarily only ask for blocks too for things we know are new live blogs until until the migration is done and we can always use blocks
-        render(path, _)
+      val range = page.map(ParseBlockId.fromPageParam) match {
+        case Some(Some(id)) => Left(PageWithBlock(id)) // we know the id of a block
+        case Some(None) => Right(NotFound) // page param there but couldn't extract a block id
+        case None => Left(Canonical) // no page param
```
TBonnin: Some(Some(..)) and Some(None) sounds weird to me. Would it make sense to create classes that encapsulate the different states?

article/app/controllers/ArticleController.scala
johnduffell: yes

article/app/controllers/ArticleController.scala
johnduffell: good idea, every line should do some specific thing

article/app/controllers/ArticleController.scala
```diff
@@ -117,16 +121,27 @@ class ArticleController extends Controller with RendersItemResponse with Logging
...
+      }
+      range.left.map { range =>
+        mapModel(path, range = Some(range)) {// temporarily only ask for blocks too for things we know are new live blogs until until the migration is done and we can always use blocks
+          render(path, _)
+        }
+      } match {
+        case Left(f) => f
+        case Right(status) => Future.successful(Cached(10)(WithoutRevalidationResult(status)))
```
TBonnin: This look confusing to me. We map page, match it, do something with the left side and match it again. Why not having only one match. Something like this:
```
      def renderWithRange(range: BlockRange) = mapModel(path, Some(range)) {
        render(path, _)
      }
      page.map(ParseBlockId.fromPageParam) match {
        case None => // no page param
          renderWithRange(Canonical)
        case Some(Some(id)) => // we know the id of a block
          renderWithRange(PageWithBlock(id))
        case Some(None) => // page param there but couldn't extract a block id
          Future.successful(Cached(10)(WithoutRevalidationResult(NotFound)))
      }
    }
```

article/app/controllers/ArticleController.scala
```diff
@@ -117,16 +121,27 @@ class ArticleController extends Controller with RendersItemResponse with Logging
...
       }
     }
 
   def renderLiveBlogJson(path: String, lastUpdate: Option[String], rendered: Option[Boolean], isLivePage: Option[Boolean]) = {
     Action.async { implicit request =>
-      mapModel(path, blocks = true) { model =>
-        (lastUpdate, rendered) match {
-          case (Some(lastUpdate), _) => renderNewerUpdates(model, lastUpdate, isLivePage)
+      val range = lastUpdate.flatMap(blockId => ParseBlockId.fromBlockId(blockId)).orElse(rendered.flatMap(r => if (r == false) Some(Canonical) else None))
```
TBonnin: breaking down this line in two might help making it easier to read

article/app/controllers/ArticleController.scala
johnduffell: good idea

article/app/liveblog/LiveBlogCurrentPage.scala
```diff
@@ -1,10 +1,71 @@
...
+      case ArticleBlocks => None
+    }
+  }
+
+  // filters newer blocks out of the list
+  def updates(blocks: Blocks, sinceBlockId: SinceBlockId) = {
+    val bodyBlocks = blocks.requestedBodyBlocks.get(sinceBlockId.around).toSeq.flatMap { bodyBlocks =>
+      bodyBlocks.takeWhile(_.id != sinceBlockId.lastUpdate)
+    }
+    Some(LiveBlogCurrentPage(FirstPage(bodyBlocks), None))// just pretend to be the first page, it'll be ignored
```
TBonnin: I don't understand the comment here. Do you mind explaining?

TBonnin:
Thanks for doing this @johnduffell 
A lot of going on in this PR. I would be happy if more people are reviewing it.

article/app/controllers/ArticleController.scala
johnduffell: big improvement, thanks

article/app/controllers/ArticleController.scala
johnduffell: actually it had a gap in the logic, I'm rewriting in line with your above suggestion

article/app/liveblog/LiveBlogCurrentPage.scala
```diff
@@ -1,10 +1,71 @@
...
+  // just get them all, the caching should prevent too many requests, could use "around"
+  val query = None
+}
+case object ArticleBlocks extends BlockRange {
+  // we currently only use this for emails, we just want all the blocks
+  val query = None
+}
+case class SinceBlockId(lastUpdate: String) extends BlockRange {
+  val around = s"body:around:$lastUpdate:5"
+  // more than 5 could come in (in one go), but unlikely and won't matter as it'll just fetch again soon
```
johnduffell: @TBonnin what I mean is the endpoint is checked every 5 seconds when you are looking at the live blog, but if you leave it in the background it can increase as far as 20 minutes.  It's possible the blogger would add more than 5 posts within that time.
```
current view:

vvv-newest          oldest-vvv
"i, j, k, l, m, n, o, p, q, r"

stuff that exists in capi
a, b, c, d, e, f, g, h, i, j, k, l, m, n, o, p, q, r

next update w'nt get us up to date:
"d, e, f, g, h"

subsequent update gets us up to date:
"a, b, c"
```
ideally it would get all 8 in the update, but it's a tradeoff between requesting a lot of blocks and re-requesting.

Also, in a susequent PR I'm going to make it cache responses once they become non-empty, so it would probably be more like:
```
update 1:
h
update 2:
g
update 3:
f
....etc....
```
however at that point I'll probably make it short circuit the delay when it gets a non empty response

article/app/liveblog/LiveBlogCurrentPage.scala
johnduffell: ah ok that's just because of the code structure, we have to turn a capi response into a frontend model in order to render it.  Normally this would be all the info needed for the article, liveblog, gallery, front, etc.  However now that  the liveblog model includes knowledge about which page we're on, and only includes a subset of the blocks, the liveblog model doesn't quite work for "updates since block-x" as there's no concept of "which page we're on".
I could have added a model type for LiveBlogUpdatesSince, but I thought for simplicity I'd just call it a FirstPage and embed the blocks in there.  The fact that there's no pagination and various other things doesn't matter to the code that eventually produces the json.

johnduffell: it's not nice, but at least the comment is there to excuse myself not doing more refactoring ;)

johnduffell:
thanks for picking through that one @TBonnin it can't have been fun, and more reviews are better.  I did a lot of things I wasn't 100% happy with the result, I feel like the code ended up slightly more complicated in the end.

cb372:
I'm going to take a look at it now.

article/app/controllers/ArticleController.scala
cb372: Actually CAPI doesn't guarantee that. We just give you the blocks in whatever order we get them from Composer. Might be worth sorting them somewhere on your side if you're not already.

article/app/controllers/ArticleController.scala
```diff
@@ -167,7 +194,7 @@ class ArticleController extends Controller with RendersItemResponse with Logging
       .showReferences("all")
       .showAtoms("all")
 
-    val capiItemWithBlocks = if (blocks) capiItem.showBlocks("body") else capiItem
+    val capiItemWithBlocks = range.map(r => capiItem.showBlocks(r.query.map(_.mkString(",")).getOrElse("body"))).getOrElse(capiItem)
```
cb372: Maybe split this line up for readability?

article/app/controllers/ArticleController.scala
```diff
@@ -234,15 +260,29 @@ class ArticleController extends Controller with RendersItemResponse with Logging
 }
 
 object ParseBlockId extends RegexParsers {
-  def apply(input: String): Option[String] = {
-    def withParser: Parser[Unit] = "with:" ^^ { _ => () }
-    def block: Parser[Unit] = "block-" ^^ { _ => () }
-    def id: Parser[String] = "[a-zA-Z0-9]+".r
-    def expr: Parser[String] = withParser ~> block ~> id
+
```
cb372: By the way, parsing with regex like this is pretty slow. Hard to say without profiling whether it's worth worrying about, but rewriting this parser using [parboiled](https://github.com/sirthias/parboiled2) might be a quick performance win.

article/app/liveblog/KeyEventData.scala
```diff
@@ -0,0 +1,32 @@
...
+  def apply(maybeBlocks: Option[Blocks], timezone: DateTimeZone): Seq[KeyEventData] = {
+
+    val blocks = maybeBlocks.toSeq.flatMap(blocks => blocks.requestedBodyBlocks.getOrElse(Canonical.timeline, blocks.body))
+
+    apply(blocks, timezone)
+  }
+
+  def apply(blocks: Seq[BodyBlock], timezone: DateTimeZone): Seq[KeyEventData] = {
+
+    val TIMELINE_MAX_ENTRIES = 7
```
cb372: Convention is to write constants like this in UpperCamelCase: `val TimelineMaxEntries = 7`

article/app/controllers/ArticleController.scala
jamesgorrie: 👍 to parboiled!
[We've use it on the grid too](https://github.com/guardian/grid/blob/master/media-api/app/lib/querysyntax/QuerySyntax.scala) (perhaps some refinements needed)

article/app/liveblog/KeyEventData.scala
jamesgorrie: Is it? Is this not just a `val` (`maxTimelineEntries`)?

jamesgorrie: `const` implies other things mutate no? Asking as I might be missing something completely.

article/app/liveblog/KeyEventData.scala
```diff
@@ -0,0 +1,32 @@
...
+
+    blocks.foldLeft((false, Nil: List[BodyBlock])) {
+      case ((summaryFound, soFar), nextBlock) if !summaryFound && nextBlock.eventType == SummaryEvent => (true, nextBlock :: soFar)
+      case ((summaryFound, soFar), nextBlock) if nextBlock.eventType == KeyEvent => (summaryFound, nextBlock :: soFar)
+      case ((summaryFound, soFar), _) => (summaryFound, soFar)
+    }._2.reverse.take(TIMELINE_MAX_ENTRIES).map { bodyBlock =>
+      KeyEventData(bodyBlock.id, bodyBlock.publishedCreatedDate(timezone), bodyBlock.title)
+    }
```
cb372: This foldLeft is clever, but I had to read it a few times before I understood it. It might be kinder to your fellow devs (and your future self) to write it as:

1. Find the latest summary
2. Find the key events
3. Merge them together and take the latest 7

e.g.

```scala
val latestSummary = blocks.find(_.eventType == SummaryEvent)
val keyEvents = blocks.filter(_.eventType == KeyEvent)
(latestSummary.toSeq ++ keyEvents).sortBy(_.lastModified).reverse.take(7)
```

article/app/liveblog/LiveBlogCurrentPage.scala
```diff
@@ -1,10 +1,71 @@
 package liveblog
 
-import model.liveblog.BodyBlock
+import model.liveblog.{Blocks, BodyBlock}
+
+sealed trait BlockRange { def query: Option[Seq[String]] }
+case object Canonical extends BlockRange {
+  val firstPage = "body:latest:30"
+  val oldestPage = "body:oldest:1"
```
cb372: Why only 1?

article/app/liveblog/KeyEventData.scala
jamesgorrie: Sorry for blurbing - it is true for uppcaseCamel for constants.
But this is just a method value?

cb372: Good point, as this is a constant it should really be moved outside the method as well as being renamed.

cb372:
Good stuff! I hope all this work pays off in bandwidth savings!

I'm going to work this afternoon on adding the support for "blocks created/modified since" that we discussed.

article/app/controllers/ArticleController.scala
johnduffell: yeah in the grand scheme of this PR it won't be much (it happens once per request) but Phil W pointed me at parboiled for something else so I'll be using it in future.  I'll add it to my list of tweaks to do once I've got the big perf stuff out of the way as it's a good idea.

article/app/liveblog/KeyEventData.scala
johnduffell: I just wanted to label the number without using a comment, happy to go with whatever naming convention but I'd rather keep it as closely scoped as possible

article/app/liveblog/KeyEventData.scala
johnduffell: oh yeah that's actually old code I wrote, maybe I was trying to be more efficient by only looping through once, yours is much better

article/app/liveblog/LiveBlogCurrentPage.scala
johnduffell: just need to get any block id in the last page so we can work out the link for the navigation bar when you click "oldest"

article/app/controllers/ArticleController.scala
johnduffell: what should we sort by? not sure which fields are guaranteed to be there as they're all options when they get to me - published, first published, last modified, created

johnduffell: I'm guessing based on your other PR comment I can sort on published date - I'll do that

johnduffell:
ok I've managed to get the tests passing, can anyone take a look and give me a +1 please?  Next stage after this is to increase the cache time for "updates since block X" requests where block X isn't the latest block.  That should leave only one html page and one json page with a 5 second cache time

johnduffell:
after that change all the perf updates should be complete, then I can start making it use published-since to get edits as well as new blocks and also fix the timeline to update again.

TBonnin:
👍  thank you very much @johnduffell

article/app/controllers/ArticleController.scala
johnduffell: fyi I'm now sorting it in https://github.com/guardian/frontend/pull/13566/commits/58ab38dd1368e9f4f627c21e1506c28d6ecb4060 and https://github.com/guardian/frontend/pull/13566/commits/a0ecfd0fe9dcc5ac3d4527b02c15eb3c3112f1e0

PR #13568, opened 2016-07-12
Do not share articles using short urls

mchv:
## What does this change?

The associated changes remove the usage of short urls when a user click on the social share buttons. Instead standard long urls are used.

We have used short urls in the past because it was allowing us to embed urls more easily in twitter. However [twitter now shortened every urls (even the short ones)](https://support.twitter.com/articles/78124) which means short urls only add `3` redirections that increase the latency before accessing the article. The latency increase is more important with `https`.

See related changes:
- [Twitter and RSS integration changes](https://github.com/guardian/content-api-transformer/pull/8)
- [Android app changes](https://github.com/guardian/android-news-app/pull/3088)
- [iOs app changes](https://github.com/guardian/ios-live/pull/2477)
- [Editorial sharing url tool](https://github.com/guardian/gdn-short-code/pull/2)

The code is used by `Amp`. 
## Audience tracking

Short campaign codes are translated to long campaign codes when frontend redirects a short url to a long url, so only long campaign codes are sent to ophan. The [ophan short url resolver](https://github.com/guardian/ophan/blob/master/shared-lib/src/main/scala/ophan/ShortUrlResolver.scala) simply follows redirect, so should continue to work fine. Therefore I think ophan should not be affected (may you confirm @obrienm @lindseydew ?)
## Communication to editorial

We have already communicate to editorial that we are removing the usage of short url, but I will write a specific email with a map between sort campaign codes and long campaign codes, so that  people editing manually the campaign code before are not lost

### DIFF
--- a/applications/test/ShareLinksTest.scala
+++ b/applications/test/ShareLinksTest.scala
@@ -23,13 +23,13 @@ import org.scalatest.concurrent.{Futures, ScalaFutures}
 
         pageShares.map(_.text) should be (List("Facebook", "Twitter", "Email", "LinkedIn", "Google plus", "WhatsApp", "Messenger"))
         pageShares.map(_.href) should be (List(
-          "https://www.facebook.com/dialog/share?app_id=202314643182694&href=http%3A%2F%2Fgu.com%2Fp%2F4gc8j%2Fsfb&redirect_uri=http%3A%2F%2Fgu.com%2Fp%2F4gc8j%2Fsfb",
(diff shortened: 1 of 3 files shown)


mchv:
@katebee we will have to sync our changes

johnduffell:
looks great to me, my only concern is that the work for the audience team will change from making a one character edit to a whole string change.  However without changing the tracking code I'm not sure we can do it without a redirect!  I think the ophan/etc tracking runs as JS in the page and inspect the ?CMP part of the url so it should be ok for ophan. 👍 
great work though, it'll be really positive for url recognition 👍

katebee:
@mchv thanks, will update my PR with your changes once they hit master.

mchv:
@johnduffell @katebee thanks!

I am waiting for editorial green light before releasing this. The plan is to do it tomorrow morning

PR #13570, opened 2016-07-12
Commercial leffe thrasher fix

Calanthe:
## What does this change?

The background color of brand name can be overwritten by DFP settings now.
## Screenshots

Before (Renault style was hardcoded):
screen shot 2016-07-12 at 14 24 51

After:
screen shot 2016-07-12 at 14 23 40
## Request for comment

@lps88


lps88:
👍

PR #13572, opened 2016-07-12
Remove babel

sndrs:
## What does this change?

remove babel from `package.json` as it was only being used in the `watch` task and is no longer needed

no one should have to see this:

screen shot 2016-07-12 at 15 03 38

no
## Screenshots

screen shot 2016-07-12 at 15 02 05


rich-nguyen:
👍

PR #13576, opened 2016-07-12
Rename LiveBlogChromeNotificationsProd

janua:
We ran `LiveBlogChromeNotificationsProd` for a while under a 1% test. Bumping the test to 10% only got new users that never seen the test before into the 10%, which wasn't as much as we would like.

This renames the test therefore rebooting it to 10% of all users.

@NathanielBennett @crifmulholland @gtrufitt


JonNorman:
:+1:

static/src/javascripts/projects/common/modules/experiments/tests/live-blog-chrome-notifications-prod.js
```diff
@@ -7,7 +7,7 @@ define([
 ) {
     return function () {
 
-        this.id = 'LiveBlogChromeNotificationsProd';
+        this.id = 'LiveBlogChromeNotificationsProd2';
         this.start = '2016-06-08';
```
JonNorman: I suppose it doesn't matter but might you want to alter the start time?

gtrufitt:
I think you should also update the switch name

static/src/javascripts/projects/common/modules/experiments/tests/live-blog-chrome-notifications-prod.js
NathanielBennett: I think you'll also need to change this line:
https://github.com/guardian/frontend/blob/master/static/src/javascripts/bootstraps/enhanced/liveblog.js#L71

NathanielBennett:
:+1: Apart from comments

janua:
@NathanielBennett `liveblog.js` has been updated.
I have also removed the `internal` test, we no longer use it or need it.

@gtrufitt We don't need to update the switch name.

gtrufitt:
Sure. Just from a cleanliness point of view than a requirement. 👍

PR #13476, opened 2016-07-04
Contributions - Article CTA

markjamesbutler:
## What does this change?

Add article inline component.  
## What is the value of this and can you measure success?

Increase in CTR to [/contribute](https://membership.theguardian.com/contribute) compared to merchandising component ads.  
## Does this affect other platforms - Amp, Apps, etc?

No
## Screenshots

screen shot 2016-07-04 at 18 40 58

PR #13577, opened 2016-07-12
Giraffe article inline styling update

markjamesbutler:
## What does this change?

Amends the styling of the Giraffe article inline component to differentiate it from the article itself.

As stated in #13476

As stated in #13476
## Screenshots

screen shot 2016-07-12 at 18 34 20


common/app/conf/switches/ABTestSwitches.scala
```diff
@@ -79,7 +79,7 @@ trait ABTestSwitches {
     "ab-giraffe",
     "Test effectiveness of inline CTA for contributions.",
     owners = Seq(Owner.withGithub("markjamesbutler"), Owner.withGithub("AWare")),
-    safeState = Off,
+    safeState = On,
```
jamesgorrie: Should `safeState` still be off? Mot ABs shouldn't run in `safeState`.

jamesgorrie:
👍

PR #13580, opened 2016-07-13
Change the GA test property

cb372:
I'm not sure what it was pointing at before. Probably an account that Grant set up and only he had access to. Updated it to use the "official" GA test account, which is set up with the same custom dimensions, etc. as the production account.


TBonnin:
go for it 👍

PR #13571, opened 2016-07-12
Identity app loader

alexduf:
## What does this change?

This removes the global object and uses an application loader for the identity module.
Identity was already using dependency injection via Guice, to make it consistent with the rest of the app I also moved it to compile time DI.
Removing the `@Inject` and `@Singleton` will come soon via another cleanup PR.

As usual this PR is easier to read on a commit per commit basis.
## Request for comment

@TBonnin @johnduffell

PR #13581, opened 2016-07-13
Identity app loader rebased

alexduf:
## What does this change?

This brings back this PR https://github.com/guardian/frontend/pull/13571 but reformulate the tests to hopefully make them more stable in our TC build.

Commits that are different are 503b1bd (melded two commits) and 1176515 which reformulate the tests

_The number of line change is high because of indetation change, add `?w=1` at the end of the url to ignore white-spaces while reading the PR_
## Request for comment

@TBonnin @johnduffell


alexduf:
I ran 10 consecutive builds on TC, it seems stable now

PR #13585, opened 2016-07-13
Migrate Onward away from Global object

alexduf:
## What does this change?

this removes the global object and introduce DI with an application loader
## Request for comment

@TBonnin @johnduffell


TBonnin:
👍

PR #13587, opened 2016-07-13
Fabric Video: position label and constrain video

regiskuckaertz:
Positions the label along the grid when a fabric advert is delivered in the top slot.

fluid-video

Although this is not shown here, the advert has a fixed 250 height and the video will cover this height and get clipped if the window gets resized horizontally, thanks to the `object-fit` property.

cc @guardian/commercial-dev


static/src/javascripts/projects/common/modules/commercial/dfp/private/render-advert.js
```diff
@@ -31,12 +31,28 @@ define([
      *
      */
 
+    function addClassIfHasClass(newClassNames) {
```
JonNorman: would it be worth naming each of these functions to make it clearer what each is doing?

static/src/stylesheets/module/_adslot.scss
```diff
@@ -135,12 +135,46 @@
         padding-top: $gs-row-height / 2;
 
         /* Oh horrible code, I'm looking at you. One day, you'll disappear.
-           One day, I'll fucking rip you apart.
+           One day, I'll [redacted].
```
JonNorman: 😂

static/src/javascripts/projects/common/modules/commercial/dfp/private/render-advert.js
regiskuckaertz: sure!

JonNorman:
Can't speak too strongly for the CSS but everything else looks good to me :+1:

(might want to look at that failing test though...)

PR #13588, opened 2016-07-14
upgrade fapi-client plus aws client

johnduffell:
Upgrade the fapi client (and as a result the aws client) to https://github.com/guardian/facia-scala-client/pull/169

This is to fix comment counts, which don't work because fapi-client assumed short urls started with http://gu.com.

There were other recent changes in the fapi client that we hadn't picked up.  As a result I have had to upgrade aws client and effectively revert this commit as it was reverted in fapi-client https://github.com/guardian/frontend/commit/d7556898e38ca428e589adbc8ddf2024b5019393

@janua @piuccio +anyone else


common/app/awswrappers/cloudwatch.scala
```diff
@@ -13,8 +13,8 @@ import scala.concurrent.Future
   */
 object cloudwatch {
   implicit class RichAsyncCloudWatchClient(client: AmazonCloudWatchAsyncClient) {
-    def putMetricDataFuture(request: PutMetricDataRequest): Future[Void] =
-      asFuture[PutMetricDataRequest, Void](client.putMetricDataAsync(request, _))
+    def putMetricDataFuture(request: PutMetricDataRequest): Future[PutMetricDataResult] =
+      asFuture[PutMetricDataRequest, PutMetricDataResult](client.putMetricDataAsync(request, _))
```
johnduffell: seems in AWS client the signature has changed from void to result

project/Dependencies.scala
```diff
@@ -7,8 +7,8 @@ object Dependencies {
   val identityLibVersion = "3.52"
   val seleniumVersion = "2.44.0"
   val slf4jVersion = "1.7.5"
-  val awsVersion = "1.10.58"
-  val faciaVersion = "1.6.2"
+  val awsVersion = "1.11.7"
+  val faciaVersion = "2.0.1-SNAPSHOT"
```
johnduffell: tbc once it's shipped

common/app/model/pressedContent.scala
```diff
@@ -481,13 +480,6 @@ object Replace {
 }
 final case class Replace(imageSrc: String, imageSrcWidth: String, imageSrcHeight: String) extends Image
 
-object ImageReplace {
-  def make(imageReplace: fapi.ImageReplace): ImageReplace = ImageReplace(
-    src = imageReplace.src)
-
-}
-final case class ImageReplace(src: String) extends Image
-
```
TBonnin: Image replace doesn't exist anymore?

TBonnin:
lgtm 👍

common/app/model/pressedContent.scala
piuccio: yeah, it was added some time ago but never really worked, so I've removed it

PR #13589, opened 2016-07-14
Article is heroic if content is heroic

SiAdcock:
## What does this change?

The ArticleController determines whether the Heroic Article view should be loaded by checking the `isHeroic` flag on the `Article` object. Currently the `isHeroic` flag is set if and only if the content is tagged as being part of the Labour & Liverpool series, without taking into consideration whether the Heroic Template Feature Switch is on.

This change will ensure that the `isHeroic` flag on the `Article` is set if the `isHeroic` flag is set on the `Content` object, which itself is set if the Heroic Template Feature Switch is on.
## Request for comment

@NataliaLKB


NataliaLKB:
👍

PR #13590, opened 2016-07-14
Commercial perf metrics

rich-nguyen:
These changes prepare the commercial system for more extensive logging of ad unit behaviour at a page view level. The aim is to move away from the ophan-based ad timings system, towards a data-lake driven detailed breakdown of the commercial behaviour.

Summary of changes:
- commercial modules must return a `Promise`
- `Advert` objects set `isEmpty` when they receive a response for dfp, instead of defaulting to a value. This means we can send accurate information about `Advert` objects at any time.
- use `updateAdvertMetric` to track when the internal state of an `Advert` object has changed. Currently, we don't measure this. We only measure dfp state changes, but it is often the case that our code is not performant!
- use  `updateAdvertMetric` to track pertinent details about `Advert` that will affect the performance of the load. Most importantly, we track the loading method, which can be one of four methods:
  - instant - these adverts are created and loaded as fast as possible, and have no deliberate wait periods.
  - lazy-load - these adverts are created as fast as possible, but do not load until they are near the viewport.
  - add-slot-instant - these ad slots are conditionally created, but loaded instantly.
  - add-slot-lazy - these ad slots are conditionally created, and loaded when they are near the viewport.
- Add a performance log to the `OphanTracking` module, so that we can capture the key metrics for:
  -  adverts - individual adverts loaded into ad unit slots
  - modules - the execution time of each commercial module
  - baselines - the milestones where the commercial system moves from phase to phase. We load commercial in two batches of modules, so we need to time module execution and advert performance based on the correct baseline. I've called them primary and secondary.
- Change `userTiming` module so that it uses the performance api where possible.


static/src/javascripts/bootstraps/commercial.js
```diff
@@ -79,25 +83,46 @@ define([
...
-                robust.catchErrorsAndLog(pair[0], function () {
-                    modulePromises.push(pair[1]());
+                    primaryModulePromises.push(modulePromise);
                 });
             });
 
-            Promise.all(modulePromises).then(function () {
-                robust.catchErrorsAndLogAll(secondaryModules);
+            Promise.all(primaryModulePromises).then(function () {
+                ophanTracking.addBaseline(ophanTracking.secondaryBaseline);
```
JonNorman: Given that the only change in this block to the block beginning at line `93` is the `ophanTracking` object, could it be wrapped into a single function that is called with a given `ophanTracking` object?

static/src/javascripts/projects/common/modules/commercial/dfp/private/ophan-tracking.js
```diff
@@ -29,6 +46,12 @@ define([
                     adTimings[slotId] = adTimings[slotId] || {};
                     timingAttr = lifecycleIdToTimingAttr[lifecycleId];
                     adTimings[slotId][timingAttr] = new Date().getTime();
+
+                    var advert = getAdvertById(slot.getSlotElementId());
```
JonNorman: Can you just use `slotId` here?

static/src/javascripts/projects/common/modules/commercial/hosted-video.js
```diff
@@ -82,101 +84,105 @@ define([
...
+                        if ($hostedNext.length && ab.getParticipations().HostedAutoplay
+                            && (ab.getParticipations().HostedAutoplay.variant === 'variant1' || ab.getParticipations().HostedAutoplay.variant === 'variant2')) {
+                            if (ab.getParticipations().HostedAutoplay.variant === 'variant2') {
+                                fastdom.write(function () {
+                                    $hostedNext.addClass('hosted-next-autoplay--variant2');
+                                });
+                            }
                             fastdom.write(function () {
-                                $hostedNext.addClass('hosted-next-autoplay--variant2');
+                                $('.js-hosted-fading').addClass('hosted-autoplay-ab');
```
JonNorman: Should this be in this PR?

JonNorman:
Minor comments, otherwise looks good to me :+1:

static/src/javascripts/projects/common/modules/commercial/hosted-video.js
rich-nguyen: I think that's the diff gone funny. Add `?w=1` to ignore whitespace

PR #13591, opened 2016-07-14
Hosted Gallery improve tracking

lps88:
## What does this change?
- Now able to track how the user navigates through the gallery, i.e. clicks or scrolls etc.
- Simplify the CTA/OJ, so there's only one version (of each) in the html
- allow key-press navigation in touch-screen mode
- change CTA button to external link icon instead of arrow

image
## Request for comment

@Calanthe


Calanthe:
👍

PR #13592, opened 2016-07-14
migrate RSS away from global and introduce App loader

alexduf:
## What does this change?

migrates RSS away from Global and introduce DI with an application loader
## Request for comment

@TBonnin @johnduffell


TBonnin:
:+1

PR #13593, opened 2016-07-14
Sport app loader

alexduf:
## What does this change?

(last one of that long series of boring PR :tada: )
Migrate Sport away from using Global and use AppLoader instead
## Request for comment

@TBonnin @johnduffell


TBonnin:
👍

PR #13594, opened 2016-07-14
Track content type and tone of hosted content

kelvin-chappell:
This change gives hosted content an identifying tone and reuses existing content types.
These are hardcoded in frontend at the moment, but their values for tracking should remain the same when we move to the scalable solution.

Eg. 
hosted video:
c9:Video
c10:Hosted

hosted gallery:
c9:Gallery
c10:Hosted

/cc @Calanthe @lps88


lps88:
👍

PR #13595, opened 2016-07-14
Hosted next video ab fixes

Calanthe:
## What does this change?

A few changes to the auto redirect AB test on hosted video pages:
- the test will run now only on hosted pages,
- proper styling on Leffe campaign
- success events registered in AB framework
- on first load, the layer does not slide out - it's hidden by default
## Screenshots

screen shot 2016-07-14 at 15 11 53
## Request for comment

@lps88 @samdesborough


lps88:
👍

static/src/javascripts/projects/common/modules/experiments/tests/hosted-autoplay.js
```diff
@@ -13,26 +23,55 @@ define([], function () {
...
+                    mediator.on('hosted video: autoredirect', complete);
                 }
             },
             {
                 id: 'variant2',
-                test: function () {
-
+                test: function () {},
+                success: function (complete) {
+                    bean.on(qwery('.hosted__next-video--tile')[0], 'click', function (){
```
desbo: Just to avoid some noise here, I think it would be worth changing these calls to `bean.on(qwery('.hosted__next-video--tile')[0], 'click', complete)`

desbo:
Minor comment aside, looks good 👍 
@harrysalmon, any thoughts on the different completion rates?

harrysalmon:
Happy with the multiple ways to have an onwards journey. Looks good to me!

PR #13596, opened 2016-07-14
Giraffe article additional styling updates

markjamesbutler:
## What does this change?

Even more the styling updates for Giraffe article inline component to differentiate it from the article itself.

As stated in #13476

As stated in #13476
## Screenshots

screen shot 2016-07-14 at 16 21 19

### DIFF
--- a/static/src/javascripts/projects/common/views/giraffe-message.html
+++ b/static/src/javascripts/projects/common/views/giraffe-message.html
@@ -1,7 +1,9 @@
 <div class="giraffe">
-    <h2 class="giraffe--heading"><%=linkText%></h2>
-    <p class="giraffe--copy">
-    <%= copy %>
-    </p>
-    <a id="giraffe-contribute" href="<%=linkHref%>" class="button button--primary button--large button--giraffe" data-link-name="<%=linkName%>">Contribute <%= svg %></a>
+    <div class="giraffe--message"/>
+        <h2 class="giraffe--heading"><%=linkText%></h2>
+        <p class="giraffe--copy">
+            <%= copy %>
+        </p>
+    </div>
+    <a id="giraffe-contribute" href="<%=linkHref%>" class="button button--primary button--large button--giraffe giraffe-button" data-link-name="<%=linkName%>">Contribute <%= svg %></a>
 </div>
--- a/static/src/javascripts/projects/common/views/giraffe-message.html
+++ b/static/src/javascripts/projects/common/views/giraffe-message.html
@@ -1,9 +1,9 @@
 <div class="giraffe">
-    <div class="giraffe--message"/>
-        <h2 class="giraffe--heading"><%=linkText%></h2>
-        <p class="giraffe--copy">
+    <div class="giraffe__message"/>
+        <h2 class="giraffe__heading"><%=linkText%></h2>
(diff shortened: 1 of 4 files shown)


static/src/stylesheets/module/experiments/_giraffe.scss
```diff
@@ -1,20 +1,32 @@
...
+    display: inline;
+}
+
+.giraffe-button {
+    background: $maps-main-2;
+    margin-top: $gs-baseline;
+    border: 0;
+
+    &:hover, &:focus {
+        background-color: darken(#328a3e, 10%);
```
jamesgorrie: Is there reasons we're not using vars here?
Just because it's a test?

jamesgorrie: By here - I mean this whole chunk of CSS.

jamesgorrie:
It looks very squashed, but that's subjective, and we should be using SASS vars, but if it's jus tfor the test 👍

static/src/javascripts/projects/common/views/giraffe-message.html
```diff
@@ -1,7 +1,9 @@
 <div class="giraffe">
-    <h2 class="giraffe--heading"><%=linkText%></h2>
-    <p class="giraffe--copy">
-    <%= copy %>
-    </p>
-    <a id="giraffe-contribute" href="<%=linkHref%>" class="button button--primary button--large button--giraffe" data-link-name="<%=linkName%>">Contribute <%= svg %></a>
+    <div class="giraffe--message"/>
```
regiskuckaertz: Minor point. We're using `__` for sub-elements: `giraffe__message`, `giraffe__contribute`, etc. and `--` for modifiers: `button--primary`.

static/src/stylesheets/module/experiments/_giraffe.scss
```diff
@@ -1,20 +1,32 @@
 .giraffe {
     @include f-header;
     width: 100%;
-    display: block;
-    margin-bottom: 1em;
-    border-top: 1px dotted $neutral-4;
+    padding: 5px 5px 12px;
```
regiskuckaertz: `padding: $gs-baseline / 2 $gs-gutter / 4 $gs-baseline`? I believe as much as possible we should tie layout to the grid vars, i.e. `$gs-baseline` for vertical rythm and `$gs-gutter` and `gs-span` for horizontal rythm.

markjamesbutler:
@jamesgorrie @regiskuckaertz Does this look any better?

static/src/stylesheets/module/experiments/_giraffe.scss
```diff
@@ -1,20 +1,34 @@
...
-    color: $neutral-2;
-    margin-bottom: $gs-baseline;
+
+.giraffe__copy {
+    @include fs-header(5);
+    color: #ffffff;
+    display: inline;
 }
+
+.button--primary {
```
regiskuckaertz: I think this class is used throughout the whole site, so you'd better change it to `button--giraffe` to avoid side-effects.

regiskuckaertz:
Minor comment but otherwise looks good to me 👍

jamesgorrie:
👍 bonza!

PR #13599, opened 2016-07-15
fix bug with message position and increase audience of test to 50%

GHaberis:
## What does this change?

Fixes this issue...

picture 18

It worked fine when polls were 100% width in liveblog blocks but now they're used in articles and potentially floated left the original implementation falls over...adding the message as a child of the figure rather than a sibling fixes it.


GHaberis:
@crifmulholland @gtrufitt

gtrufitt:
👍

PR #13602, opened 2016-07-15
Limit scope of CAPi styling for h2, h3 

rich-nguyen:
This addresses #13286 , the greedy css rules based on `from-content-api`. There isn't much consensus about this, as we can see the majority of rules in `from-content-api` are child-based, but some are general descendant. This does fix issues with body-based ad-slots, so I think it's worth doing. I can't see any reason to target elements we don't want to style, I haven't managed to find a counter-example.


regiskuckaertz:
It's a matter of style, but I find it's just good stewardship to limit the scope of a selector such as `.from-content-api` rather than have all the other selectors which work in the same space try to get around it. I'll give it a 👍, especially since it's blocking #13564

PR #13603, opened 2016-07-15
Internal healtchecks don't expire

TBonnin:
## What does this change?

This is a following this PR https://github.com/guardian/frontend/pull/13486 which has been live for 10 days without any issue.

Internal healthcheck are being executed only once to verify the service
can serve a real request but once it is successful the internal
healthcheck is not executed anymore. The goal is to minimize the impact
of any upstream services on the health of frontend services

See above. Frontend health doesn't depend too much on the health of upstream services

No
## Request for comment

@johnduffell @alexduf


alexduf:
Cool :+1:

PR #13604, opened 2016-07-15
Fix deploys radiator

sndrs:
## What does this change?

linting webpack i introduced a bug to building of deploys radiator, this fixes it

cc @samdesborough


johnduffell:
cool, is that why the radiator is switched off recently? 👍

admin/app/views/deploysRadiator/main.scala.html
```diff
@@ -11,7 +11,6 @@
     <script>
         window.gitHubAccessToken = '@conf.Configuration.github.token';
     </script>
-    <script src="https://cdnjs.cloudflare.com/ajax/libs/es6-promise/3.1.2/es6-promise.min.js"></script>
-    <script src="https://cdnjs.cloudflare.com/ajax/libs/fetch/0.11.0/fetch.min.js"></script>
+    <script src="https://cdn.polyfill.io/v2/polyfill.min.js?features=fetch,Promise"></script>
```
rich-nguyen: did those old polyfills not work?

sndrs:
> cool, is that why the radiator is switched off recently? 

no, not related. it should be working again now though

sndrs:
> did those old polyfills not work?

yeah they did (afaik), but both were unnecessary in the chrome instance it was running in before. seemed a good place to try polyfillio out?

PR #13606, opened 2016-07-15
Fixup isHeroic flag in JS config

SiAdcock:
## What does this change?

Update the isHeroic flag on the JS config to point to the `isHeroic` variable, rather than the `isImmersive` variable 😅 
## Request for comment

@NataliaLKB


NataliaLKB:
Haha! My bad too. 👍

SiAdcock:
^ Business case for pair programming as standard right there 😄

PR #13609, opened 2016-07-15
Update switch around LiveBlogChromeNotifications

janua:
@gtrufitt Yeah yeah... :smirk:


gtrufitt:
I would change the val - just for consistency with the rest of the test name references (unless you find a secret place that is used too...)

gtrufitt:
Other than that 👍

NathanielBennett:
:+1: How about changing the 'Owners' value as well. Anyone would think you're frightened of responsibility ...

janua:
@NathanielBennett What are you on about? Nobody owns this anymore, and you **are** the person who worked on it.

PR #13610, opened 2016-07-15
Better media tracking

jamesgorrie:
## What does this change?

I am trying to decouple our media tracking from videojs.
It's all a little complicated at the moment - this hopefully simplifies it a little bit.

If another player comes into place - we would just have to change the `media-events` to hook into the right thing and then all other analytics would work using mediator.

GA is going to be a good test for this.

I have left the preroll events in the very strange `video/events` so as not to change too much.

Nothing breaks, devs life becomes easier.

Nope
## Screenshots

🏩 
## Request for comment

@akash1810 @gidsg


static/src/javascripts/projects/common/modules/video/media-events.js
```diff
@@ -0,0 +1,67 @@
...
+
+        // This is to ensure the ready event fires, this is only in enhanced mode.
+        // We might want to restrict this more to desktop for data reasons, but we would have to figure what we do about
+        // the analytics.
+        el.preload = 'metadata';
+
+        // Here we are making sure that we only fire these events once so we aren't tracking things a million times.
+        // General UX events e.g. pause, play, fullscreen, will be tacked elsewhere.
+        var fireOnceEventMapping = {
+            canplay: 'ready',
```
akash1810: worth having the same key and value values?

static/src/javascripts/projects/common/modules/video/media-events.js
```diff
@@ -0,0 +1,67 @@
...
+            bean.one(el, eventName, function() {
+                bean.fire(el, fullEventName, eventProps);
+                mediator.emit(fullEventName, eventProps);
+            });
+        });
+
+        bean.on(el, 'timeupdate', throttle(function() {
+            var percent = Math.round((el.currentTime / el.duration) * 100);
+
+            if (percent >= 25) {
```
akash1810: should these be ranges? eg `>= 25 && < 50`

static/src/javascripts/projects/common/modules/video/ophan-media.js
```diff
@@ -0,0 +1,64 @@
...
+                ophanRecord(mediaProps.mediaId, mediaProps.mediaType, typedEventType);
+            });
+        });
+    }
+
+    function ophanRecord(mediaId, mediaType, eventType) {
+        if (mediaId) {
+            getOphan().then(function(ophan) {
+                var eventObject = {};
+                eventObject[mediaType] = {
```
akash1810: 😢

akash1810:
have we checked the embed page behaves normally?

static/src/javascripts/projects/common/modules/video/media-events.js
jamesgorrie: Nope. That's why I've use `passed` - if you're `59%` of the way you've passed `25` and `50%`. This happens every second.

the logic dealing with it happening once is the other part above.

jamesgorrie: I could do ranges and have something like:

`if (percent >= 25 && percent < 50) { bean.fire(el, 'position:1quarter') }`
`if (percent >= 50 && percent < 75) { bean.fire(el, 'position:2quarter') }`
`if (percent >= 75 && percent < 100) { bean.fire(el, 'position:3quarter') }`

Would that make more sense?

jamesgorrie:
> have we checked the embed page behaves normally?

It doesn't.

static/src/javascripts/projects/common/modules/video/media-events.js
akash1810: My reason for asking was for fear `passed25` would fire at 51%, but 

> This happens every second. the logic dealing with it happening once is the other part above.

The current implementation looks better 👍

akash1810:
> It doesn't.

Can it 😄

static/src/javascripts/projects/common/modules/video/media-events.js
jamesgorrie: `canplay` is from the native API - doesn't make that much sense and is fired loads of times. This fires once once we're ready.

static/src/javascripts/projects/common/modules/video/ophan-media.js
jamesgorrie: Why the 💧 ?

akash1810: I started writing a comment asking why not do this at time of variable declaration, e.g

```js
var eventObject = {
  mediaType: {
    id: mediaId,
    eventType: eventType
  }
}
```

Then remembered its js 😢

static/src/javascripts/projects/common/modules/video/ophan-media.js
```diff
@@ -0,0 +1,64 @@
...
+            'media:content:firstplay': 'media:content:play',
+            'media:content:watched25': 'media:content:25',
+            'media:content:watched50': 'media:content:50',
+            'media:content:watched75': 'media:content:75',
+            'media:content:firstended': 'media:content:end'
+        };
+
+        forOwn(eventMappings, function(ophanEventType, originalEventType) {
+            mediator.on(originalEventType, function(mediaProps) {
+                // We should have `mediaType` as part of the resord, rather than inferred by the `eventType` prefix.
```
TBonnin: tiny typo on `resord` 😝

static/src/javascripts/projects/common/modules/video/media-events.js
jamesgorrie: Basically I'm trying to come up with a set of events that make more sense to our tracking.

static/src/javascripts/projects/common/modules/video/media-events.js
jamesgorrie: > passed25 would fire at 51%

It would also fire at 27%, 53%, 90% etc. As the event name suggests, you're passed that %.

static/src/javascripts/projects/common/modules/video/ophan-media.js
jamesgorrie: es6:
```JavaScript
var eventObject = {
  [mediaType]: {
    id: mediaId,
    eventType: eventType
  }
}
```

akash1810:
if it works on the embed page too 👍

jamesgorrie:
@akash1810 @gidsg I wouldn't mind a once over again as I have changed quite a lot.

static/src/javascripts/projects/common/modules/video/media-events.js
```diff
@@ -0,0 +1,67 @@
...
+
+    function addContentEvents(el) {
+        var mediaType = el.tagName.toLowerCase();
+        // TODO: fastdom
+        var mediaId = el.getAttribute('data-media-id');
+
+        // This is to ensure the ready event fires, this is only in enhanced mode.
+        // We might want to restrict this more to desktop for data reasons, but we would have to figure what we do about
+        // the analytics.
+        el.preload = 'metadata';
```
akash1810: so we don't obey the element's attribute?

akash1810:
👍

static/src/javascripts/projects/common/modules/video/media-events.js
jamesgorrie: No.

PR #13611, opened 2016-07-15
Reduce commercial CSS complecting: top slot

regiskuckaertz:
Refactoring commercial CSS looks like a party going all wrong:

giphy 2

... but if we're looking to use [Atomic CSS](https://github.com/guardian/frontend/pull/13549), we'd better start somewhere. One thing we can do is reduce the "cyclomatic complexity" in our selectors. This will improve the overall _aesthetic_ of our stylesheet but also stop the sheer number of times we need to override styles in specificity wars.

I start slow, with the top slot, and am voluntarily spartan about removing styles for now—it will become easier as all selectors converge towards the smallest specificity.

cc @guardian/commercial-dev @sndrs


regiskuckaertz:
(This also fixes the FOOSA which happens for some of our home-fledged creatives)

*FOUSA = Flash Of Overlapping Sticky Ad

common/test/views/support/CommercialTest.scala
```diff
@@ -42,13 +42,4 @@ class CommercialTest extends FlatSpec with Matchers with OptionValues with Befor
   // they should "be responsive for 88x70 ad on US business front" in {
   //   topAboveNavSlot.cssClasses(metaDataFromId("us/business"), defaultEdition, Seq(responsiveSize)) should endWith("top-banner-ad-container--responsive")
   // }
-
-  they should "be default for any other page" in {
```
rich-nguyen: Should this test be removed? If the test was edited to check `endWith("js-top-banner-desktop")`, would it pass?

static/src/javascripts/projects/common/modules/commercial/dfp/create-slot.js
```diff
@@ -94,16 +94,6 @@ define([
                     adSizes.fluid250,
                     adSizes.fabric,
                     adSizes.fluid
-                ),
-                desktop: compile(
-                    adSizes.outOfPage,
-                    adSizes.fluid250,
-                    adSizes.leaderboard,
-                    adSizes.cascade,
-                    adSizes.superHeader,
```
rich-nguyen: can you also remove some of the entries from `ad-sizes.js`, like 'superHeader'?

rich-nguyen:
👍  nice

static/src/javascripts/projects/common/modules/commercial/dfp/create-slot.js
regiskuckaertz: Yep, nice catch!

common/test/views/support/CommercialTest.scala
regiskuckaertz: Makes sense, yes.

PR #13614, opened 2016-07-15
Fix: adverts rows don't wrap by default

regiskuckaertz:
This caused a layout bug in some flavours of webkit:

picture 1

cc @guardian/commercial-dev


rich-nguyen:
👍

PR #13615, opened 2016-07-15
Hosted gallery hash

lps88:
### What does this change?

Adds hash to url so you can reload on a particular image
### What is the value of this and can you measure success?

Want to be able to share specific images
### Screenshots

image
### Request for comment

@guardian/labs-beta

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/hosted-gallery.js
+++ b/static/src/javascripts/projects/common/modules/commercial/hosted-gallery.js
@@ -6,6 +6,7 @@ define([
     'common/utils/$',
     'qwery',
     'common/utils/config',
+    'common/utils/url',
     'common/utils/detect',
     'common/utils/fsm',
     'common/utils/mediator',
@@ -22,6 +23,7 @@ define([
              $,
              qwery,
              config,
+             url,
              detect,
              FiniteStateMachine,
              mediator,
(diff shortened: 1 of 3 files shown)


static/src/javascripts/projects/common/modules/commercial/hosted-gallery.js
```diff
@@ -299,6 +301,8 @@ define([
                     bonzo(this.$galleryEl).toggleClass('show-oj', this.index === this.$images.length);
                     bonzo(this.$galleryEl).toggleClass('show-cta', this.index === config.page.ctaIndex + 1);
                 }
+
+                url.pushUrl({}, document.title, config.page.pageName + '#img-' + this.index, true);
```
jamesgorrie: Could we not just use `window.location.hash = 'blahblah'`? This also ads it to the back button - which might be a consideration as if I have gone through 20 images, I have to hit back 20 times before actually going back/

jamesgorrie:
Just a question around the UX of the browser history.
👍

static/src/javascripts/projects/common/modules/commercial/hosted-gallery.js
lps88: The pushUrl function does a replaceState if you pass true in the 4th param, so you don't have to press back 20 times. The window.location.hash thing would be simpler though..

jamesgorrie: Ah, I didn't realise - I assumed by `push` it meant append as it would in array land.
👍  for clarifying.

PR #13616, opened 2016-07-15
Work around intermittent hosted video button bug

kelvin-chappell:
This reverts the part of https://github.com/guardian/frontend/pull/13590 that applies to hosted video.
It seems to be causing an intermittent bug here.


lps88:
This does fix the play button! But not the 'about' button. Which I guess is still an improvement, and enough of a fix until Monday  👍

PR #13618, opened 2016-07-15
Don't run Giraffe test on sensitive articles or live blog

markjamesbutler:
## What does this change?

Prevents the Giraffe ab test running on sensitive articles and live blogs.

None, this is a functional change to ensure our test isn't shown on sensitive articles.

No
## Screenshots

None

### DIFF
--- a/static/src/javascripts/projects/common/modules/experiments/tests/giraffe.js
+++ b/static/src/javascripts/projects/common/modules/experiments/tests/giraffe.js
@@ -33,7 +33,8 @@ define([
         this.dataLinkNames = '';
         this.idealOutcome = '';
         this.canRun = function () {
-             return window.guardian.config.page.edition == 'UK';
+             return !window.guardian.config.page.isSensitive &&
+                 !window.guardian.config.page.isLiveBlog && window.guardian.config.page.edition == 'UK';
         };
 
         var writer = function (linkText, linkHref, copy) {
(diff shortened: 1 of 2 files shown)


static/src/javascripts/projects/common/modules/experiments/tests/giraffe.js
```diff
@@ -33,7 +33,8 @@ define([
         this.dataLinkNames = '';
         this.idealOutcome = '';
         this.canRun = function () {
-             return window.guardian.config.page.edition == 'UK';
+             return !window.guardian.config.page.isSensitive &&
```
jamesgorrie: Not sure that it matters, but wondering if `!(isSensitive || isLiveBlog) && edition === 'UK'` - but that's subjective.

jamesgorrie:
:+1:

static/src/javascripts/projects/common/modules/experiments/tests/giraffe.js
JonNorman: So, in `commercial` we already have [commercial-features.js](https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/common/modules/commercial/commercial-features.js), which has `sensitiveContent` and other definitions defined already. Feel free to use!

markjamesbutler:
@jamesgorrie Does my change read better to you?

PR #13620, opened 2016-07-18
Increase onward endpoints caching from 1s to 900s

TBonnin:
## What does this change?

Some onward endpoints (most viewed on social,  series trails, media in section) were only cached 1s 
This patch increases the cache time to 900.

More caching, Less load for the onward service
[Out of the 10 most requested urls for the onward services](http://kibana.gu-web.net/goto/f43653282dfccf59e4571535388e67a4), 9 are /series/.... urls (first one being weather api widget) 

I don't think there is any good reason to cache those endpoints for only 1s. Also all other onward endpoints seem to be cached for 900 ([example](https://github.com/guardian/frontend/blob/01176a76e5122b65eec87df4cc35212e50561167/onward/app/controllers/TopStoriesController.scala#L77))

No
## Request for comment

@johnduffell

### DIFF
--- a/onward/app/controllers/MediaInSectionController.scala
+++ b/onward/app/controllers/MediaInSectionController.scala
@@ -89,7 +89,7 @@ class MediaInSectionController extends Controller with Logging with Paging with 
       ).withTimeStamps,
       FrontProperties.empty
     )(request)
-    renderFormat(response, response, 1)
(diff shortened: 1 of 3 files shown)


johnduffell:
that's crazy! great spot! 👍 🚤

PR #13623, opened 2016-07-18
Hosted Video 

rich-nguyen:
I'm reverting #13616 because we don't think this is the root cause. And moving `host-about` init to commercial from enhance, and tweaking the init test so it checks the tone, not the content type. 

This fixes the about button, and shouldn't affect the video upgrade.


kelvin-chappell:
As far as I can tell! 👍

static/src/javascripts/projects/common/modules/commercial/hosted-video.js
```diff
@@ -85,101 +87,105 @@ define([
...
-                    if (vol) {
-                        player.volume(0);
-                        player.volume(vol);
+        return new Promise(function (resolve) {
+            require(['bootstraps/enhanced/media/main'], function () {
+                require(['bootstraps/enhanced/media/video-player'], function (videojs) {
+                    var $videoEl = $('.vjs-hosted__video');
+
+                    if (!$videoEl.length) {
+                        return;
```
regiskuckaertz: Should be `resolve()` instead.

static/src/javascripts/projects/common/modules/commercial/hosted-video.js
```diff
@@ -85,101 +87,105 @@ define([
...
-                    upgradeVideoPlayerAccessibility(player);
-
-                    // unglitching the volume on first load
-                    vol = player.volume();
-                    if (vol) {
-                        player.volume(0);
-                        player.volume(vol);
+        return new Promise(function (resolve) {
+            require(['bootstraps/enhanced/media/main'], function () {
+                require(['bootstraps/enhanced/media/video-player'], function (videojs) {
```
regiskuckaertz: Man this function is looooooooooong. Might be worth abstracting a little in another PR 😮

regiskuckaertz:
Aside from minor comment, gooooo 👍

static/src/javascripts/projects/common/modules/commercial/hosted-video.js
rich-nguyen: yep good spot

Calanthe:
👍 
I will make a card for myself to extract some of the code in the`hosted-video.js` file.

PR #13625, opened 2016-07-18
Make a skeletal hosted article

kelvin-chappell:
This sets up the skeleton for the first campaign with a hosted article.

/cc @Calanthe


Calanthe:
👍

PR #13627, opened 2016-07-18
Facia - Remove internal content code when calling content API

mchv:
## What does this change?

Remove internal content code retrieval when calling content API. This field will be removed from content API response today.


piuccio:
👍  `internalContentCode` was replaced by `internalPageCode` some time ago, all clients are already up to date.

TBonnin:
Does it worth regenerating all the test data file that includes this field?

mchv:
@TBonnin I think it will be good when we will have release the change on content API side

TBonnin:
@mchv not asking for the field should be enough for not having it in the returned payload. right?

It is not a big deal here because I don't think this field is used anywhere but if you want to be more confident about this change, running all the tests again with data that doesn't contain this field could be useful.

No big deal though

mchv:
@TBonnin absolutely.

PR #13631, opened 2016-07-18
Fix short URL lookups for capi merchandise component

kelvin-chappell:
This fixes a bug that was introduced in move to https, where short URLs stopped being recognised.

I'm going to make a separate PR to be able to use standard URLs as well as short ones.

/cc @guardian/commercial-dev


rich-nguyen:
👍

PR #13632, opened 2016-07-18
Use permalink for sharing comments links

marialivia16:
## What does this change?

Uses the permalink instead of the direct link for both FB and Twitter sharing links.

No.
## Request for comment

@mchv @nicl


mchv:
👍  sorry for the late review!

nicl:
(retrospective 👍 - looks good to me)

PR #13634, opened 2016-07-18
Show all liveblog updates

johnduffell:
At the moment, if multiple liveblog updates come in while the window is in the background, all but the first are lost.

This bug was introduced when I made it not re-fetch all the unread blocks every 5 seconds.

There was also a small bug with the number in the status bar.

I have fixed both.

@TBonnin


TBonnin:
👍

PR #13635, opened 2016-07-18
Commercial top slot height for Fabric

regiskuckaertz:
There already is a `min-height` rule on the top slot for fabric adverts, so no need to add this class.

Some creatives like Fabric lie about their height 250 because they add their own "Advertisement" label and therefore increase their inner height to 268.

cc @JonNorman @rich-nguyen


rich-nguyen:
👍

PR #13636, opened 2016-07-18
Fixing the new-header in safari

NataliaLKB:
## What does this change?

Currently the new header on the smallest breakpoint doesn't look right on Safari. This just deletes the line of code that is causing the problem.

Nope!
## Screenshots

**before:**
image

**After:**
image


zeftilldeath:
Have a 🎅 👍

PR #13637, opened 2016-07-18
Add back Ad sizes used in PrebidService

regiskuckaertz:
I got overzealous and removed ad sizes that were still in use in our PrebidService module

cc @rich-nguyen


rich-nguyen:
👍

PR #13639, opened 2016-07-19
Fix top mobile slot height and z-index issues

regiskuckaertz:
So, the fluid250 creative, despite its name, is only 100px high on mobile. I [changed the `min-height` to accommodate for that oddity](https://github.com/guardian/frontend/compare/rk-fluid250-mobile?expand=1#diff-080ab32aa8365abdced6a7d722b4d6a5R115).

picture 3

The z-index issue is also fixed:
picture 4

And as a side bonus there was a missing rule that did not reset the top slot correctly for fluid ads:
picture 6

This is also fixed:
picture 5

cc @guardian/commercial-dev

### DIFF
--- a/static/src/stylesheets/module/_adslot.scss
+++ b/static/src/stylesheets/module/_adslot.scss
@@ -71,14 +71,13 @@
 
 .sticky-top-banner-ad {
     overflow: hidden;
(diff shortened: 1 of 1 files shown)


Calanthe:
👍

PR #13642, opened 2016-07-19
Add a noscript for GA

cb372:
## What does this change?

Adds a `<noscript>` tag below the GA snippet. For now it doesn't record the pageview with GA, it only contains a confidence beacon.

To investigate whether the lack of a `noscript` is the cause of the low GA confidence (currently around 90%).

No
## Screenshots

n/a
## Request for comment

@dominickendrick


jamesgorrie:
:+1:

PR #13643, opened 2016-07-19
Make share icons for comments bigger

marialivia16:
## What does this change?
- Increases share icons' size to make them easily tappable on mobile.
- Removes opacity.

No.
## Screenshots

bigger icons
## Request for comment

@zeftilldeath


zeftilldeath:
There are some measurements that could be done with our $gs-gutter etc. After that, looks great 👍

PR #13644, opened 2016-07-19
Fix video dev deps

jamesgorrie:
## What does this change?

Having a `video-player-dev` and `video-player` is strange.
This move the logic into our build systems (which is where some of it was too).

There was also some changing of the module names to match the module names given by the components, making it even simpler.

No
## Screenshots

👍 
## Request for comment

@akash1810 @gidsg


akash1810:
💥 👍

PR #13647, opened 2016-07-19
hide all video elements when fc-item--has-sublinks

akash1810:
## What does this change?

Fixes a rare bug of the phantom floating play button when a card has multiple sublinks and is showing video. This is caused by us hiding the video element (well, its parent) but not the play button. Targeting `fc-item__video` instead of `fc-item__media-wrapper` as `fc-item__video` is present on all elements that should be hidden.

pretty front

no
## Screenshots

Before
screen shot 2016-07-19 at 12 17 53

After
screen shot 2016-07-19 at 12 19 06
## Request for comment

@jamesgorrie @gidsg

### DIFF
--- a/static/src/stylesheets/module/facia/_items.scss
+++ b/static/src/stylesheets/module/facia/_items.scss
@@ -146,7 +146,7 @@
 
         &[class*='fc-item--has-sublinks'] {
             .fc-item__standfirst,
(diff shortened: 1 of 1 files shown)


jamesgorrie:
🍌 👍

PR #13649, opened 2016-07-19
Hosted content selective loading

rich-nguyen:
This addresses the following:
- we are currently loading the media js app on every page!
- there is an initialisation race condition between enhanced media, and hosted video media. This means that hosted video often fails to upgrade, and appears unskinned.
- hosted pages fail to run the `showUserName` inline script without throwing the error: `Cannot set property 'innerHTML' of undefined`

### DIFF
--- a/static/src/javascripts/bootstraps/commercial.js
+++ b/static/src/javascripts/bootstraps/commercial.js
@@ -60,7 +60,6 @@ define([
         ['cm-load', dfpLoad],
         ['cm-thirdPartyTags', thirdPartyTags.init],
         ['cm-sponsorships', sponsorships.init],
-
         ['cm-paidforBand', paidforBand.init],
         ['cm-paidContainers', paidContainers.init],
         ['cm-ready', function () {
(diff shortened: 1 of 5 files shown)


kelvin-chappell:
Hurray! 👍

Calanthe:
👍

static/src/javascripts/bootstraps/commercial.js
```diff
@@ -72,6 +69,13 @@ define([
         }]
     ];
 
+    if (config.isHosted) {
+        secondaryModules.unshift(
```
regiskuckaertz: Do we want these module to boot before `dfpLoad`?

regiskuckaertz:
Yeah 👍

static/src/javascripts/bootstraps/commercial.js
rich-nguyen: It's fine, there aren't any dependencies on `dfpLoad`.

PR #13650, opened 2016-07-19
map on container elements to ensure existance

jamesgorrie:
## What does this change?

Running a foreach on the collection ensures that we definitely send an element over to the container.

At the moment we send undefined errors.

No more error

Nope
## Screenshots

🍭


jamesgorrie:
@akash1810 @gidsg

akash1810:
👍

PR #13652, opened 2016-07-19
Commercial: remove commercial-v--high

regiskuckaertz:
Since we finally the two remaining inline components to the new markup, we no longer need these classes.

(There's a load more we don't need but I'll do each in a separate PR.)

cc @guardian/commercial-dev


rich-nguyen:
👍

PR #13654, opened 2016-07-19
Fix short URL lookups for capi merchandising component

kelvin-chappell:
I didn't quite fix it in #13631!

This should do it.

There's also a slight improvement to the agent update, to avoid potential dirty reads.

/cc @guardian/commercial-dev


regiskuckaertz:
👍

PR #13655, opened 2016-07-19
add media analytics to build

jamesgorrie:
## What does this change?

We need to add this to create the bootstrap file.

It works

No
## Screenshots

:house:
## Request for comment

@akash1810


akash1810:
👍

PR #13657, opened 2016-07-19
sloppy way of mediaType checking

jamesgorrie:
## What does this change?

Fixes things

Things are fixed

No
## Request for comment

@akash1810


akash1810:
👍

PR #13659, opened 2016-07-19
Habit Digest snap banner

katebee:
## What does this change?

Adds an **inactive** test (currently off and at zero audience) for the Habit Digest promo. This test will show users a banner offering a curated digest.

The banner co-opts the Membership Message banner to display one of three variant messages to the visitor (see screenshots). It has a lower priority than the Breaking News banner, and will not appear again once a visitor has dismissed it.
- Users will only see one snap variant
- Users will only see the snap once
- Breaking news gets priority

The sign-up button is a false door / 404 test that will open a info page.
- Discover if there is demand for various Guardian digest offerings
- See which version of the copy generates greatest engagement

Ideally, visitors will click on the CTA and demonstrate demand for the feature.

No
## Screenshots

picture 126
## For a subsequent PR:
- ~~Check the links in the [banner variants](https://github.com/guardian/frontend/blob/2920e52b7a8c1cbb1a6204f9964211f6a846103c/static/src/javascripts/projects/common/modules/experiments/tests/habit-forming-digest-promo.js#L96): are they correct and do the landing pages exist?~~ **done**
- Run the test (turn AB switch on and set audience percentage)
## Request for comment

@joelochlann

### DIFF
--- a/static/src/javascripts/projects/common/modules/experiments/tests/habit-forming-digest-promo.js
+++ b/static/src/javascripts/projects/common/modules/experiments/tests/habit-forming-digest-promo.js
@@ -35,8 +35,7 @@ define([
         this.idealOutcome = 'Visitors click on the CTA and demonstrate demand for the feature';
 
         this.canRun = function () {
-            return !(config.page.isAdvertisementFeature) &&
-            config.page.contentType === 'Article'
+            return !config.page.isAdvertisementFeature && config.page.contentType === 'Article';
         };
 
         var defaultData = {
--- a/static/src/javascripts/projects/common/modules/experiments/tests/habit-forming-digest-promo.js
+++ b/static/src/javascripts/projects/common/modules/experiments/tests/habit-forming-digest-promo.js
@@ -36,20 +36,12 @@ define([
 
         this.canRun = function () {
             return !(config.page.isAdvertisementFeature) &&
-            config.page.contentType === 'Article' &&
-            hasSeenMembership();
+            config.page.contentType === 'Article'
         };
 
         var defaultData = {
             arrowRight: svgs('arrowRight')
         };
-
-        // check that the membership banner has been seen and dismissed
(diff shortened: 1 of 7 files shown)


static/src/javascripts/projects/common/modules/experiments/tests/habit-forming-digest-promo.js
```diff
@@ -0,0 +1,122 @@
...
+
+        var defaultData = {
+            arrowRight: svgs('arrowRight')
+        };
+
+        // check if the user is one of the target audience
+        function isInfrequentVisitor() {
+            if (storage.local.isStorageAvailable()) {
+                var alreadyVisited = storage.local.get('gu.alreadyVisited') || 0;
+                if (alreadyVisited > 3 && !userFeatures.isPayingMember()) {
```
joelochlann: Hmmmm... this seems to be doing the opposite of what it says! It's looking for people who have visited the site more than 3 times.

joelochlann: or maybe it's the function name that's misleading

janua: @joelochlann I have also read it as the opposite! The double negative is confusing.

joelochlann: from Monica:

> Re audience, we will reverse analyse this afterwards, so of the people who see & interact with this, which fall into the non-regular / regular buckets of users etc as there is no good way to target those audiences at the moment.

so all we should be doing is:
* ensuring users only see the snap once
* ensuring we don't conflict with the membership snap

so it looks to me like this whole condition is irrelevant.

SiAdcock:
LGTM 👍

PR #13660, opened 2016-07-19
listen to event before being able to play on video

jamesgorrie:
## What does this change?

We need to attach the event listeners before setting videojs off as that could play the video before we're listening.


jamesgorrie:
👍

PR #13669, opened 2016-07-20
Cache auto update endpoints for liveblogs

johnduffell:
For liveblog updates currently we can ask for all blocks since block-nnnnn.  Since there are many blocks in a liveblog, that is many unique requests.

Previously all those were kept up to date by caching only 5 seconds.  This will cache all but the latest one for a longer time (1 hour) and the latest one only will be cached for 5 seconds.

This has the side effect of making browsers have to fetch multiple times to get all the updates, although with a 10 second cycle time it's unlikely to be a problem.  However the update frequency increases when you switch away from the tab, so I have made it short cycle on the client side if there's an update when we go back to the browser from an other tab.

Overall we should save money!

@TBonnin take a look


article/app/controllers/ArticleController.scala
```diff
@@ -238,8 +238,19 @@ class ArticleController extends Controller with RendersItemResponse with Logging
...
+          if (liveBlog.fields.isLive && pageModel.currentPage.blocks.isEmpty)
+            liveBlog.metadata.cacheTime
+          else
+            CacheTime.NotRecentlyUpdated
+        val liveBlogCache = liveBlog.copy(
+          content = liveBlog.content.copy(
+            metadata = liveBlog.content.metadata.copy(
+              cacheTime = cacheTime)))
+        Left(LiveBlogPage(liveBlogCache, pageModel, StoryPackages(liveBlog, response)))
+      case (Some(pageModel), _) =>
```
TBonnin: Looks like `liveBlogCache` definition is duplicated below. Can we only to match against the `blockRange` when defining cacheTime and keep the `liveBlogCache` in one place.

Also can we abstract away the `cacheTime` logic because it is getting quite involve at this point

static/src/javascripts/projects/common/modules/ui/autoupdate.js
```diff
@@ -164,8 +165,13 @@ define([
...
             }).always(function () {
-                updateDelay(currentUpdateDelay);
-                updateTimeoutId = setTimeout(checkForUpdates, currentUpdateDelay);
+                if (count == 0 || currentUpdateDelay > 0) {
+                    updateDelay(currentUpdateDelay);
+                    updateTimeoutId = setTimeout(checkForUpdates, currentUpdateDelay);
+                } else {
+                    // might have been cached so check straight away
+                    updateTimeoutId = setTimeout(checkForUpdates, 1);
+                }
```
TBonnin: I know this is not related to this PR specifically but I usually found all this states management (ifs, state vars) quite error prone. Not much you can do at this stage but to keep in mind if the liveblog logic keep growing.

johnduffell:
ok I've refactored that lump of code @TBonnin hopefully it's more intelligible now

TBonnin:
👍

johnduffell:
Requests per minute for the AJAX updates endpoint halved 🎉 :
image

johnduffell:
although they did close one of the live blogs around 16:33 so the actual improvement will be smaller

TBonnin:
Worth looking at data over a longer period of time (ex: one full day last week vs one full day after this merge)

johnduffell:
The last week graph by hour of number of 304 Not Modified responses per hour (i.e. unnecessary GETs) for the "get updates" endpoint.  I switched the lastest bit on between the 25 and the 13 of `2016-07-25 13:00`
image

johnduffell:
final graph with the time of deploy marked subtly, still seems to be doing fine
image

PR #13670, opened 2016-07-20
Commercial: remove topBannerMobile

regiskuckaertz:
This slot should no longer exist since we recently move to using `top-above-nav` as the first slot on mobile.

cc @guardian/commercial-dev


rich-nguyen:
👍  cool

PR #13672, opened 2016-07-20
Fixing bug in new header menu

NataliaLKB:
## What does this change?

This fixing a problem that sometimes arises in the new header menu.

The bug seems to happen when the function `animateMenuClose` runs when the page loads. Even though there is no transition yet, [this event listener](https://github.com/guardian/frontend/blob/52756757d4c9b942c1cb6ad357081d787ded0760/static/src/javascripts/projects/common/modules/navigation/newHeaderNavigation.js#L49) is set. When you then open the menu the class `shown` gets added to the menu container. Then, the transition ends which triggers the callback and removes the class `shown`. This bug looks like this:
navbug

Now everything is fine!
navworking

Nope!
## Request for comment

@SiAdcock @zeftilldeath 

❄️ 🐱 🐀 ⚡ 🐴 🐣 🐝 🔥 🐟 🐙 🐍 🐉 🍃 🌸 🌺 🍄 🌱 💧

### DIFF
--- a/static/src/javascripts/projects/common/modules/navigation/newHeaderNavigation.js
+++ b/static/src/javascripts/projects/common/modules/navigation/newHeaderNavigation.js
@@ -13,8 +13,7 @@ define([
 
     function animateMenuOpen() {
         return fastdomPromise.write(function () {
-            mainMenuEl.addClass('shown');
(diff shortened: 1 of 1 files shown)


SiAdcock:
Lovely stuff 💎  👍

PR #13673, opened 2016-07-20
ensure hard disk is encrypted

johnduffell:
Everyone's hard disk should be encrypted, this just adds a check to make sure it's so before downloading the properties file from S3.

Not sure how to do it on linux - someone who knows might update that side of the script?

@TBonnin 
@adamnfish @janua you are linux guys, what do you think?

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -56,8 +56,13 @@ create_frontend_properties() {
     if [[ ! -d "$path" ]]; then
       mkdir "$path"
     fi
-
-    aws s3 cp --profile frontend s3://aws-frontend-store/template-frontend.properties "$path/$filename"
+    
+    if [[ "$(fdesetup status)" != "FileVault is On." ]]; then
+      echo your hard disk is not encrypted!
+    else
+      aws s3 cp --profile frontend s3://aws-frontend-store/template-frontend.properties "$path/$filename"
+    fi
+ 
   fi
 }
 
--- a/setup.sh
+++ b/setup.sh
@@ -57,12 +57,16 @@ create_frontend_properties() {
       mkdir "$path"
     fi
     
-    if [[ "$(fdesetup status)" != "FileVault is On." ]]; then
-      echo your hard disk is not encrypted! follow these instructions: https://support.apple.com/en-gb/HT204837
-    else
-      aws s3 cp --profile frontend s3://aws-frontend-store/template-frontend.properties "$path/$filename"
+    if linux; then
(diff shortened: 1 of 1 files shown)


dominickendrick:
Great stuff !

setup.sh
```diff
@@ -56,8 +56,13 @@ create_frontend_properties() {
     if [[ ! -d "$path" ]]; then
       mkdir "$path"
     fi
-
-    aws s3 cp --profile frontend s3://aws-frontend-store/template-frontend.properties "$path/$filename"
+    
+    if [[ "$(fdesetup status)" != "FileVault is On." ]]; then
+      echo your hard disk is not encrypted!
```
TBonnin: I will also tell people what to do to solve the problem ex: `If Mac, go to Preferences > Security > FileVault ... If Linux...`

johnduffell: no I'm hoping the linux people will  let me know what command to add for that =)

adamnfish:
:-1: needs a `uname contains darwin` check. This will break it for everyone not on OSX!

johnduffell:
@adamnfish 
`if uname != darwin; then echo "please write the correct script to check encryption is on before continuing"; exit 1; fi`

janua:
Yeah definitely needs a check for linux at least, warning that it isn't checked.

Maybe there is a more OS free way of checking it? I know when you list mounts they say if they are encrypted or not in linux, not sure about mac though.

desbo:
the script already has OS checks, you can just reuse them with `if linux` or `if mac`

setup.sh
```diff
@@ -56,8 +56,13 @@ create_frontend_properties() {
     if [[ ! -d "$path" ]]; then
       mkdir "$path"
     fi
-
-    aws s3 cp --profile frontend s3://aws-frontend-store/template-frontend.properties "$path/$filename"
+    
+    if [[ "$(fdesetup status)" != "FileVault is On." ]]; then
+      echo your hard disk is not encrypted! follow these instructions: https://support.apple.com/en-gb/HT204837
```
desbo: you can use the `EXTRA_STEPS` thing here, have a look at the other examples in the file

sndrs:
:floppy_disk:

johnduffell:
ok thanks for the comments, I guess there's no simple linux check so I'll just skip that bit, someone can add it later!  Take another look 😛

desbo:
lgtm 👍

PR #13674, opened 2016-07-20
Commercial: remove obsolete ad sizes

regiskuckaertz:
We're getting rid of 300x251, an old ad size we used to make an MPU sticky, because the MPU in the right column always sticks.

Also a few remnants of 300x50 and 320x50 that are long dead.

cc @guardian/commercial-dev


rich-nguyen:
👍

PR #13542, opened 2016-07-08
Track header bidding events with GA

regiskuckaertz:
This will enable events to be collected when the following events occur:
- a bid request is sent
- a bid response is received
- some network failed to return bids within allocated time
- a winning bid got displayed

More information here: http://prebid.org/dev-docs/analytics-ga.html

cc @guardian/commercial-dev @cb372

PR #13675, opened 2016-07-20
Add more Google Analytics trackers

cb372:
## What does this change?

Add a couple more GA trackers:
- One for the dedicated header bidding property used by #13542
- One for the test GA account, making it easy for people to send events to the test account until they're sure everything works as expected

Unblocks #13542

No
## Screenshots

n/a
## Request for comment

@regiskuckaertz


regiskuckaertz:
Yeah 👍

PR #13676, opened 2016-07-20
New header dummy test

NataliaLKB:
## What does this change?

This adds two serverside dummy tests. It is in preparation for the new header tests that I hope to get out soon. 

Corresponding [fastly PR](https://github.com/guardian/fastly-edge-cache/pull/659) which will set the amounts for the buckets.
## Purposes of the dummy tests:
- Check if the control and variant get equal sized buckets
- See if the data we get back is enough to start testing
- Help us estimate how many people should be in each bucket

Nope!
## Request for comment

@stephanfowler


johnduffell:
👍

PR #13677, opened 2016-07-20
Removes blockquotes from comment share body

marialivia16:
## What does this change?

Removes blockquotes from the body of the shared comment because there is no delimitation between the quote and the actual comment

### DIFF
--- a/discussion/app/views/fragments/commentShare.scala.html
+++ b/discussion/app/views/fragments/commentShare.scala.html
@@ -9,7 +9,7 @@
 
     <div class="sharing-buttons" data-link-name="comment social">
     @defining(s"${comment.webUrl}") { permalink =>
-        @defining(StringEscapeUtils.unescapeHtml(Jsoup.clean(comment.body, Whitelist.none()))) { commentBody =>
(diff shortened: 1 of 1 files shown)


johnduffell:
not 100% keen on regexing html, (not that I'm keen on the jsouping we already do in the cleaners either)
If we require JS for this anyway, could we use JS to grab the right comment text from the dom when they hit share? This would let us grab the right bit (and also save sending it 1.5 times extra to the client)

marialivia16:
@johnduffell 
Do you mean by this doing something similar with the [selection sharing](https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/common/modules/ui/selection-sharing.js) implementation and not use the `ShareLinks.createShareLinkForComment(...)` at all?

johnduffell:
oh yeah I did mean that but I didn't know about that code!  It was just a thought that popped into my mind that might be an alternative approach - I didn't want to derail your PR.

I just realised you are trying to remove all blockquotes from the comment, not just one surrounding the whole comment, so my suggestion wouldn't really work easily anyway.

We had a similar problem with having to use JSoup to manipulate liveblog html - actually when CAPI started returning blocks in a structured way, we didn't need to do it any more.  DAPI could do something similar, although it would be overkill at the moment @guardian/discussion 

👍 go for it for now, but just beware of ending up with complicated manipulation with regexes

marialivia16:
@johnduffell Ok, got it! It's good to know.

PR #13679, opened 2016-07-20
Fix article body ads naming

regiskuckaertz:
Fixes an issue where two `top-above-nav` slots would be inserted into the body of articles on mobile, and other inline slots would start from 2 instead of 1.

picture 1

cc @guardian/commercial-dev


piuccio:
I'm not sure what would cause two `top-above-nav`in the old code

👍

regiskuckaertz:
The fact that we may call the function twice 😅

PR #13680, opened 2016-07-20
Fix ad size in comments slot

regiskuckaertz:
The slot should contain MPUs, not badges.

cc @guardian/commercial-dev


rich-nguyen:
👍

PR #13683, opened 2016-07-21
Commercial: fix false positive in lazy-loaded containers

regiskuckaertz:
`Element.getBoundingClientRect()` will return a zero-filled structure when the element is inside another element which is not displayed. This structure will go through [the test that tells if the ad is within the viewport range](https://github.com/guardian/frontend/compare/rk-inline2?expand=1#diff-b2831f831e4bbc7ec76981ed134c91cbL37) and succeed, hence implying the ad should be displayed.

This happens for ads inside containers that are lazy-loaded, i.e. containers starting from the 10th on fronts. To fix this issue, we just add another guard to detect the zero-filled structure.

cc @guardian/commercial-dev


rich-nguyen:
👍  do all browsers return 0s for hidden elements?

regiskuckaertz:
@rich-nguyen yes, this is [per spec](https://drafts.csswg.org/cssom-view/#dom-element-getboundingclientrect).

PR #13690, opened 2016-07-21
always show video over sublinks

akash1810:
## What does this change?

Currently, if a card is showing video and has two or more sublinks, the video is hidden (for reasons unknown).
This is different for images - cards with an image and two sublinks shows all elements.

This change brings video and images inline - videos are always shown regardless of the number of sublinks.

(originally raised by CP)

⬆️ videos on fronts and consequently ⬆️ plays on fronts

No.
## Screenshots

Before (image with two sublinks on left, video with two sublinks on right)
screen shot 2016-07-21 at 23 38 43

After (image with two sublinks on left, video with two sublinks on right)
screen shot 2016-07-21 at 23 50 13

NB - as you can see the before screenshot looks strange. Currently CP are advising to untick `show video` in the fronts tool as this causes the main media's poster image to be displayed, e.g.

screen shot 2016-07-21 at 23 57 05
## Request for comment

@jamesgorrie @gidsg

### DIFF
--- a/static/src/stylesheets/module/facia/_items.scss
+++ b/static/src/stylesheets/module/facia/_items.scss
@@ -145,8 +145,7 @@
         }
 
         &[class*='fc-item--has-sublinks'] {
-            .fc-item__standfirst,
-            .fc-item__video {
+            .fc-item__standfirst {
                 display: none;
             }
         }


jamesgorrie:
👍

akash1810:
tnx for merging and deploying 👍

PR #13692, opened 2016-07-22
Introducing Guardian Revealer

regiskuckaertz:
A new template for fixed background images on mobile:

revealer

This template is temporary. We will start moving all our templates to DFP fluid ads in this quarter.

cc @guardian/commercial-dev

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/creatives/revealer.js
+++ b/static/src/javascripts/projects/common/modules/commercial/creatives/revealer.js
@@ -1,0 +1,31 @@
+define([
+    'common/utils/fastdom-promise',
+    'common/utils/template',
+    'common/modules/commercial/creatives/add-tracking-pixel',
+    'text!common/views/commercial/creatives/revealer.html'
+], function(fastdom, template, addTrackingPixel, revealerStr) {
+    var hasBackgroundFixedSupport = !detect.isAndroid();
+    var revealerTpl;
+
+    function Revealer($adSlot, params) {
+        revealerTpl || (revealerTpl = template(revealerStr));
+
+        return Object.freeze({
+            create: create
+        });
+
+        function create() {
+            var markup = revealerTpl(params);
+
+            return fastdom.write(function () {
+                $adSlot[0].insertAdjacentHTML('beforeend', markup);
+                $adSlot.addClass('ad-slot--revealer');
+                if (params.trackingPixel) {
+                    addTrackingPixel($adSlot, params.trackingPixel + params.cacheBuster);
+                }
+            });
+        }
+    }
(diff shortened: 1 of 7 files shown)


static/src/javascripts/projects/common/modules/commercial/creatives/revealer.js
```diff
@@ -0,0 +1,30 @@
...
+    'common/utils/template',
+    'common/modules/commercial/creatives/add-tracking-pixel',
+    'text!common/views/commercial/creatives/revealer.html'
+], function(fastdom, template, addTrackingPixel, revealerStr) {
+    var revealerTpl;
+
+    function Revealer($adSlot, params) {
+        revealerTpl || (revealerTpl = template(revealerStr));
+
+        return Object.freeze({
```
janua: Why return this above `create`?

regiskuckaertz: Just a convention, I like to move declarations out of the actual logic of the function.

janua: I meant more; why depend on hoisting? If so you could just put this at the very top of the `function Revealer`?

regiskuckaertz: That's personal preference (and blatant laziness 😅 ). I prefer going top-down when reading code, to avoid unnecessary scrolling.

janua: I like it, maybe I will adopt the same style!

janua:
LGTM :+1: 
I don't really understand how the `creative__background` with a `position: fixed` gets fixed to the ancestor though, I thought it needs a transform or something else?

regiskuckaertz:
@janua Thanks! The coordinate space for fixed elements is relative to the viewport (unless you apply a transform to an ancestor elements, but this is a rare special case)

static/src/javascripts/projects/common/views/commercial/creatives/revealer.html
```diff
@@ -0,0 +1,6 @@
+<aside class="creative creative--revealer">
+    <a class="creative__cta" href="%%CLICK_URL_ESC%%<%=link%>" target="_blank">
+        <img class="creative__background" src="<%=backgroundImage%>" alt>
+        <img class="creative__button creative__button--<%=buttonVPos%> creative__button--<%=buttonHPos%>" src="<%=button%>" alt>
+    </a>
+</div>
```
Calanthe: Is it wrapped in `aside` or `div`?

PR #13693, opened 2016-07-22
Remove use of AkkaAsync object in Notification trait

TBonnin:
## What does this change?

Removing use of AkkaAsync object
I am gonna try to make small(ish) and digestable PR, so it is not too boring to review. Anyway it makes sense to review commit by commit 😬 

=> Play 2.5

The end goal is to get rid of the AkkaAsync object which rely on
Play.current. _Play.current_ being deprecated in Play 2.5
Instead using dependency injection to pass AkkaAsync instance to the
different functions where it is needed

No
## Request for comment

@alexduf

### DIFF
--- a/admin/app/AppLoader.scala
+++ b/admin/app/AppLoader.scala
@@ -16,11 +16,16 @@ import play.api.http.HttpErrorHandler
 import play.api.mvc.EssentialFilter
 import play.api.routing.Router
 import play.api._
-import services.ConfigAgentLifecycle
+import services.{ConfigAgentLifecycle, EmailService}
 import router.Routes
 
 class AppLoader extends FrontendApplicationLoader {
   override def buildComponents(context: Context): FrontendComponents = new BuiltInComponentsFromContext(context) with AppComponents
+}
+
+trait AdminServices {
+  def akkaAsync: AkkaAsync
+  lazy val emailService = wire[EmailService]
 }
 
 trait Controllers extends AdminControllers {
@@ -44,7 +49,7 @@ trait AdminLifecycleComponents {
   )
 }
 
-trait AppComponents extends FrontendComponents with AdminLifecycleComponents with Controllers {
+trait AppComponents extends FrontendComponents with AdminLifecycleComponents with Controllers with AdminServices {
 
   lazy val router: Router = wire[Routes]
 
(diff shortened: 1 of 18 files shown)


admin/app/controllers/SwitchboardPlistaController.scala
```diff
@@ -6,11 +6,11 @@ import conf.switches.{SwitchState, Switches}
 import controllers.AuthLogging
 import conf.Configuration
 import play.api.mvc._
-import services.Notification
+import services.SwitchNotification
```
TBonnin: Renamed it below to give it a more meaningful name

admin/app/AppLoader.scala
```diff
@@ -16,13 +16,18 @@ import play.api.http.HttpErrorHandler
...
+import services.{ConfigAgentLifecycle, EmailService}
 import router.Routes
 
 class AppLoader extends FrontendApplicationLoader {
   override def buildComponents(context: Context): FrontendComponents = new BuiltInComponentsFromContext(context) with AppComponents
 }
 
+trait Services {
+  self: AppComponents =>
+  lazy val emailService: EmailService = wire[EmailService]
```
TBonnin: Got the following error if not specifying the type
```
[error] ...frontend/admin/app/AppLoader.scala:28: recursive value emailService needs type
[error] lazy val emailService = wire[EmailService]
````

File: admin/app/controllers/AdminControllers.scala [hunk 3]
```diff
@@ -1,11 +1,14 @@
 package controllers
 
 import com.softwaremill.macwire._
+import common.AkkaAsync
 import controllers.admin._
 import controllers.admin.commercial.{DfpDataController, SlotController, TakeoverWithEmptyMPUsController}
 import controllers.cache.{ImageDecacheController, PageDecacheController}
 
 trait AdminControllers {
+  def akkaAsync: AkkaAsync
```
janua: Is this being mixed in from `FrontendComponents`?

janua:
LGTM :+1:

admin/app/controllers/AdminControllers.scala
TBonnin: Yep https://github.com/guardian/frontend/blob/8d202876eb0238c2b8ce9a0a8be4ba8486495abb/admin/app/AppLoader.scala#L26-L47

alexduf:
Same here, looks good to me :star:  :+1:  :star:

PR #13694, opened 2016-07-22
Vertical align sharing sharing comments icon

marialivia16:
## What does this change?

Noticed on my personal laptop that the icon was showing too high up, I'm unable to reproduce it on this machine, but this should solve the problem


janua:
:+1:

PR #13700, opened 2016-07-22
New media events

jamesgorrie:
## What does this change?

Adding new MediaEvents to follow more closely the model we have in Ophan and model we can use in GA.

Nothing is using these events at the moment - but the plan is to get them into GA, and then potentially port them to Ophan.

Things become simpler as time goes on.

No
## Screenshots

🎅 
## Request for comment

@akash1810 @gidsg

### DIFF
--- a/static/src/javascripts/bootstraps/enhanced/media/main.js
+++ b/static/src/javascripts/bootstraps/enhanced/media/main.js
@@ -166,6 +166,7 @@ define([
             withPreroll,
             blockVideoAds;
 
+
         var videoInfo = new Promise(function(resolve) {
             // We only have the canonical URL in videos embedded in articles / main media.
             // These are set to the safest defaults that will always play video.
@@ -200,6 +201,11 @@ define([
                 }
             }
         }));
+        events.addContentEvents(player, mediaId, mediaType);
+        events.addPrerollEvents(player, mediaId, mediaType);
+        if (window.location.hash === '#debugVideoEvents') {
+            events.debugEvents(player);
+        }
 
         videoInfo.then(function(videoInfo) {
             if (videoInfo.expired) {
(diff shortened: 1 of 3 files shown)


static/src/javascripts/bootstraps/enhanced/media/main.js
```diff
@@ -166,6 +166,7 @@ define([
             withPreroll,
             blockVideoAds;
 
+
```
gidsg: I do not feel this whitespace adds a huge amount of value.

jamesgorrie: Agreed. It has been terminated.

static/src/javascripts/projects/common/modules/video/events.js
```diff
@@ -36,6 +40,93 @@ define([
             'content:end'
         ];
 
+
+    /**
+     *
+     * @param mediaId {string}
+     * @param mediaType {string} audio|video
+     * @param eventType {string} e.g. firstplay, firstend
+     * @param preroll {boolean}
```
akash1810: minor: would `isPreroll` make the data-type more obvious?

static/src/javascripts/bootstraps/enhanced/media/main.js
```diff
@@ -200,6 +200,11 @@ define([
                 }
             }
         }));
+        events.addContentEvents(player, mediaId, mediaType);
+        events.addPrerollEvents(player, mediaId, mediaType);
+        if (window.location.hash === '#debugVideoEvents') {
```
akash1810: should this be an item in `frontend.properties` instead?

akash1810: would require a server restart for changes to be picked up, but would behave more like the `useHashedBundles` flag?

akash1810:
couple of minor comments, else 👍

static/src/javascripts/bootstraps/enhanced/media/main.js
jamesgorrie: Not 100% sure what you're talking about - here we're just looking at the hash?

PR #13702, opened 2016-07-25
Log all Guardian specific headers regardless of the case

TBonnin:
## What does this change?

Log all Guardian specific headers regardless of the case

Better logging and debugging 🎯 

No
## Request for comment

@johnduffell


johnduffell:
👍

PR #13705, opened 2016-07-25
initial implementation of AMP live blog

SiAdcock:
## What does this change?

This is a first pass at implementing Live Blogs using AMP. The intention is to get the simplest example of live blogs into production. 

As it stands, the live update feature will not work. It will not be indexed by Google.

As an MVP, this will provide a foundation upon which additional features (such as live updates) and design work can be built.

Existing non-AMP live blogs will not be affected. Existing AMP articles will not be affected.
## Screenshots

picture 44
## Request for comment

@sndrs

### DIFF
--- a/article/app/views/liveblog/liveBlogBody.scala.html
+++ b/article/app/views/liveblog/liveBlogBody.scala.html
@@ -41,9 +41,14 @@
                     <div class="content__main-column content__main-column--liveblog">
                         @fragments.standfirst(article)
 
-                        <div class="content__updated-container content__updated-container--liveblog">
-                            @if(article.fields.isLive) {<span class="blog__live live-pulse-icon flashing-image">LIVE</span> }Updated <time class="js-timestamp" datetime="@article.fields.lastModified" data-relativeformat="med"></time>
-                        </div>
+                        @if(!amp) {
+                            <div class="content__updated-container content__updated-container--liveblog">
+                                @if(article.fields.isLive) {
+                                    <span class="blog__live live-pulse-icon flashing-image">LIVE</span>
+                                }
+                                Updated <time class="js-timestamp" datetime="@article.fields.lastModified" data-relativeformat="med"></time>
+                            </div>
+                        }
                     </div>
                 </div>
             </div>
@@ -84,14 +89,16 @@
 
                     @fragments.dropdown("Live feed", Some("live-feed"), true) {
                         <div class="js-article__container" data-component="body">
-                            <div class="toast__space-reserver">
-                                <div id="toast__tofix" class="toast__container">
-                                    <button class="toast__button toast__button--closed button button--large">
-                                        <span class="toast__text"></span>
-                                        @fragments.inlineSvg("refresh", "icon", List(""))
-                                    </button>
+                            @if(!amp) {
+                                <div class="toast__space-reserver">
+                                    <div id="toast__tofix" class="toast__container">
+                                        <button class="toast__button toast__button--closed button button--large">
+                                            <span class="toast__text"></span>
+                                            @fragments.inlineSvg("refresh", "icon", List(""))
+                                        </button>
+                                    </div>
                                 </div>
-                            </div>
+                            }
                             @if(model.currentPage.currentPage.isArchivePage) {
                                 @fragments.liveBlogNavigation(article.content.metadata.id, model.currentPage)
                             }
--- a/article/app/views/liveblog/liveBlogBody.scala.html
+++ b/article/app/views/liveblog/liveBlogBody.scala.html
@@ -88,54 +88,10 @@
                     </div>
 
                     @if(amp) {
-                        <div class="js-article__container" data-component="body">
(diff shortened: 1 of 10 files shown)


article/app/views/liveblog/liveBlogBody.scala.html
```diff
@@ -108,6 +107,36 @@ <h1 itemprop="headline" class="content__headline js-score">@Html(article.trail.h
...
+                                @* Remove LiveFilter and the fragment once Toast comes out of AB test status *@
+                                @fragments.liveBlogNavigation(article.content.metadata.id, model.currentPage)
+                            </div>
+
+                            @fragments.submeta(article)
+
+                            <div class="after-article js-after-article"></div>
+
+                            <div class="js-bottom-marker"></div>
+                        }
```
TBonnin: There is a lot of code duplication in this `if/else`. Is there any way to avoid it?

SiAdcock: Good shout. I could probably extract everything inside the dropdown into a fragment

johnduffell:
great start! 👍 
@NataliaLKB and I had a few debates about whether it's best to pass the amp boolean around or use the requestHeader implicit when things are really nested.  I think we do different thing depending what feels most appropriate for that code.
Also I'd watch out for having big if (amp) {.......} blocks, I like that you split one of the views out to avoid that.

johnduffell:
I just noticed this isn't behind a switch or opt-in, be careful what google (and twitter uses amp) does, we don't want to accidentally send everyone who googles to our amp version if it doesn't provide the functionality we would expect.

SiAdcock:
Thanks @johnduffell, this is really helpful!

Passing that `amp` Boolean around felt a bit smelly. I shall have another pass and see if I can use `requestHeader.isAmp` directly anyway.

From what I have read, Google will not index the AMP version of the live blog as we do not add an `amphtml` link element to the canonical (non-AMP) version of the live blog. Therefore the AMP version [will not be discoverable by Google](https://support.google.com/webmasters/answer/6340290?hl=en#discovery). However, I can wrap this in a Feature Switch if you think it will be safer. It would also give us greater control, and enable us to quickly turn off the AMP live blog in the event of some catastrophe.

TBonnin:
+1 for a switch. It is easier to manage the release with a `off` default value and safer/quicker to disable if something goes wrong (even though nothing goes wrong ever of course 😬 )

SiAdcock:
After a conversation with @sndrs, I'm going to take another look at the CSS to remove all nested class selectors. This is because we would like the AMP live blog to eventually use [postcss-atomised](https://github.com/sndrs/postcss-atomised), which will not support nested classes for the time being.

SiAdcock:
I've made the following changes / excuses:

-  Added a feature switch for AMP live blogs. Prevents the `amphtml` link tag from being added to the canonical page while the switch is off. Switch is off by default
- De-nestified CSS in preparation for postcss-atomised
- The way the `amp` flag is being passed down the view tree is appropriate in all instances, as it is always being used within the views to determine which elements to show. If the flag was being passed to a view simply to allow that view to pass it to a child, in this instance referring to the `isAmp` flag on the `requestHeader` in the child view would be more appropriate. Happy to hear any views on this.

article/app/views/liveblog/liveBlogBodyContent.scala.html
```diff
@@ -0,0 +1,44 @@
+@import _root_.liveblog.KeyEventData
+@import model.Badges.badgeFor
+
+@import _root_.liveblog.Canonical
```
johnduffell: it probably wasn't yours originally, but the line spacing is a little wonky - and actually there's no need for any of the imports in the signature by the looks, so they could all be below (or all above if you prefer that way)

common/app/conf/switches/FeatureSwitches.scala
```diff
@@ -377,6 +377,16 @@ trait FeatureSwitches {
...
   )
 
+  val LiveBlogAmpSwitch = Switch(
+    SwitchGroup.Feature,
+    "live-blog-amp-switch",
+    "If this switch is on, link to amp pages will be in the metadata for live blogs",
+    owners = Seq(Owner.withGithub("SiAdcock")),
+    safeState = Off,
+    sellByDate = never,
+    exposeClientSide = true
```
johnduffell: is it really needed client side? also normally we don't use "switch" in the switch name, just in the name of the val itself

johnduffell: also it should have a sell by date, I don't think this switch will be around for ever

johnduffell:
looks great, just a couple of comments on the switch and then it can go as far as I'm concerned!  I edited the title to make it easier to read too if that's ok! 👓

SiAdcock:
Thanks for all your suggestions @johnduffell, I agree with all of them. I'll remove unnecessary imports, mark the switch for expiry (or review) one month from today, fix the switch name and prevent its exposure on the client.

Title update looks good too 😄

johnduffell:
👍

Issue #9898, opened 2015-07-22
Discussion Abuse Reports: Indicate success or failure of submission to user


gidsg:
We had some reports from legal of abuse reports not being received, this may be because the success or failure of the submission is not indicated to the user, so the report may 'silently fail.'

Steps to reproduce
================
Go to: http://www.theguardian.com/science/grrlscientist/2012/aug/07/3#comment-56181127
Turn off Wifi/Ethernet
Click report abuse on a comment and submit form.

Actual Results
============
Dialog closes with no indication of success or failure to the user

Expected Results
=============
Indicate the success or failure of an abuse report, as per posting a comment:

![screen shot 2015-07-22 at 12 39 04](https://cloud.githubusercontent.com/assets/1764158/8825698/f031010c-3078-11e5-8371-e3823a0ce66b.png)

CC @guardian/discussion @JustinPinner @OliverJAsh

JustinPinner:
@gidsg is this still an issue? @OliverJAsh referenced this issue in a PR merge in July.

gidsg:
Yes, @JustinPinner this is still an issue, happy to discuss in person if that'd be easier.

nicl:
@JustinPinner @gidsg yes, would be great to improve the UX here.

ScottPainterGNM:
@gtrufitt This was the one I mentioned earlier if you do look at discussion soon, bug is still reproduceable

NathanielBennett:
https://github.com/guardian/frontend/pull/13718
Done

PR #13706, opened 2016-07-25
Discussion abuse report error

NathanielBennett:
## What does this change?

Deals with this issue: https://github.com/guardian/frontend/issues/9898. When a user submits an abuse report on a comment, we check for success/failure and, in the case of failure, display an error message

Users will know if their report have not been submitted

N/A
## Screenshots

report-error-msg
## Request for comment

@gidsg @nicl @piuccio


static/src/javascripts/projects/common/modules/discussion/comments.js
```diff
@@ -422,20 +423,36 @@ Comments.prototype.reportComment = function(e) {
...
 };
 
+Comments.prototype.reportCommentSuccess = function(form) {
+    bonzo(form).addClass('u-h');
+};
+
+Comments.prototype.reportCommentFailure = function() {
+    var message = 'Sorry there was an error reporting this comment. Please try another browser or network connection.',
+        error = bonzo.create(
+        '<div class="d-discussion__error ' + this.getClass('error', true) + '">' +
```
piuccio: Can we put this into a `.scala.html` behind a hidden div and then simply toggle the visibility?

NathanielBennett: @piuccio Yeah could do. For this, i was just using the same functionality as we do for posting comments. But that's equally as do-able. What you think?

piuccio: I don't really know what's the most common approach in dotcom.  I'd advocate for splitting responsibilities, so HTML is rendered server side inside HTML, not in JS. But it would be good to know what the team says

@sndrs @NataliaLKB  maybe?

NathanielBennett:
@piuccio Hows that. Error messsage in scala

piuccio:
looks cleaner to me

static/src/javascripts/projects/common/modules/discussion/comments.js
```diff
@@ -413,7 +413,8 @@ Comments.prototype.replyToComment = function(e) {
 Comments.prototype.reportComment = function(e) {
     e.preventDefault();
 
-    var commentId = e.currentTarget.getAttribute('data-comment-id');
+    var self = this,
+        commentId = e.currentTarget.getAttribute('data-comment-id');
 
     $('.js-report-comment-form').first().each(function(form) {
         bean.one(form, 'submit', function(e) {
```
piuccio: kinda unrelated, but `bean.one` means that you'll be able submit only once. If there's an error, the second time I click on the submit button either nothing happens, or I'm redirected to the form location (hopefully that won't trigger other errors)

piuccio:
👍

nicl:
@NathanielBennett realise this has been merged, but only question is whether we should just say 'please try again' for the error message? Depends whether we think errors are more likely down to browser/client network or our own service being temporarily down/flaky.

gidsg:
@nicl +1 I do think it would be good to add 'please try again'

NathanielBennett:
k: i'm just shepherding something else out of the door. Then I can fix this.

PR #13689, opened 2016-07-21
Sport vs Sports

zeftilldeath:
US vs UK, Stephan vs The World
## What does this change?

The new header will read 'sport' rather than 'sports'.

screen shot 2016-07-21 at 17 04 04

@stephanfowler feel free to merge in my absence.

PR #13710, opened 2016-07-25
Depluralise sports text in AMP header

SiAdcock:
## What does this change?

As mentioned in #13689, I have updated the header in AMP articles to say "sport" rather than "sports"
## Screenshots

picture 45

@stephanfowler @johnduffell @NataliaLKB @zeftilldeath


stephanfowler:
👍

PR #13712, opened 2016-07-25
Move comment count to standalone widget

piuccio:
## What does this change?

Instead of having a giant repository of everything, take some dependencies out. The first one is the comment count widget, hosted [here](https://github.com/guardian/comment-count-widget).

It's a very simple component, given a custom tag, get the value from the server and update the DOM.

It's mostly refactor, the only behavioural change is that templates are now rendered server side.
Articles for instance will have a comments icon and 'comments' text link but the actual number is ajaxed in.

Developing should be easier. The widget was written using TDD and very quick iterations. It's also possible to use ES6 and transpile back to ES5 + AMD. No need to watch the whole frontend, recompile on html changes or run all the tests.

The widget is used in 

**fronts**
- front pages (inside any card)
- cards behind show more (comment count won't be displayed because they're ajaxed in)

**articles**
- comment count below the headline
- NOT the comment count near the discussion text area (that's another part of comments)
- comment count of story packages (if any)
- immersive articles

**live blog**
- story packages if any

**interactive**
- comment count below the headline
- NOT immersive interactives, the count is hidden

**galleries**
- comment count below the headline
- cartoons

AMP is not affected (I guess)

In terms of file size, the new widget is written in ES6 and transpiled, turns out that _`enhanced/main.js`_ is

|  | New | Old |
| --- | --- | --- |
| Minified | 87.604 | 87.985 |
| Gzipped | 26.241 | 26.105 |

file is slightly smaller, but once gzipped is marginally bigger (by 140 bytes)
## Request for comment

@sndrs

### DIFF
--- a/applications/app/views/fragments/crosswords/crosswordMetaHeader.scala.html
+++ b/applications/app/views/fragments/crosswords/crosswordMetaHeader.scala.html
@@ -37,7 +37,11 @@
                     <div class="meta__save-for-later js-save-for-later" data-position="top"></div>
                 }
                 <div class="meta__numbers modern-visible">
-                    <div class="u-h meta__number" data-discussion-id="@crosswordPage.content.content.discussionId" data-commentcount-format="content" data-discussion-closed="@{crosswordPage.content.trail.isCommentable}">
+                    <div class="meta__number">
+                        <a href="#comments" data-link-name="Comment count" class="commentcount2 tone-colour">
+                            <h3 class="commentcount2__heading">@fragments.inlineSvg("comment-16", "icon", List("inline-tone-fill")) Comments</h3>
+                            <comment-count class="commentcount2__value tone-colour" discussion="@crosswordPage.content.content.discussionId" @{if (!crosswordPage.content.trail.isCommentable) "closed"}></comment-count>
+                        </a>
                     </div>
                     <div class="u-h meta__number js-sharecount">
                     </div>
@@ -52,4 +56,3 @@
         </div>
     </div>
 </header>
-
--- a/common/app/templates/inlineJS/blocking/curlConfig.scala.js
+++ b/common/app/templates/inlineJS/blocking/curlConfig.scala.js
@@ -79,7 +79,10 @@ window.curlConfig = {
 
             // plugins
             text:         'components/requirejs-text/text',
-            inlineSvg:    'projects/common/utils/inlineSvg'
+            inlineSvg:    'projects/common/utils/inlineSvg',
+
+            // external guardian libraries
+            'comment-count':               '@{Configuration.assets.path}guardian-comment-count/dist/comment-count.amd.js'
         }
     }
 };
--- a/common/app/views/fragments/contentMeta.scala.html
+++ b/common/app/views/fragments/contentMeta.scala.html
@@ -42,9 +42,11 @@
             <div class="meta__numbers modern-visible">
                 <div class="u-h meta__number js-sharecount">
                 </div>
-                <div class="u-h meta__number" data-discussion-id="@item.content.discussionId" data-commentcount-format="content" data-discussion-closed="@{
-                    !item.trail.isCommentable
-                }">
+                <div class="meta__number js-commentcount">
+                    <a href="#comments" data-link-name="Comment count" class="commentcount2 tone-colour">
+                        <h3 class="commentcount2__heading">@fragments.inlineSvg("comment-16", "icon", List("inline-tone-fill")) Comments</h3>
+                        <comment-count class="commentcount2__value tone-colour" discussion="@item.content.discussionId" @{if (!item.trail.isCommentable) "closed"}></comment-count>
+                    </a>
                 </div>
             </div>
             @if(SaveForLaterSwitch.isSwitchedOn && !item.content.isImmersiveGallery) {
--- a/common/app/views/fragments/contentMetaImmersive.scala.html
+++ b/common/app/views/fragments/contentMetaImmersive.scala.html
@@ -15,7 +15,12 @@
 
     <div class="meta__numbers">
         <div class="u-h meta__number meta__number--sharecount js-sharecount js-sharecount-immersive"></div>
-        <div class="u-h meta__number meta__number--commentcount" data-discussion-id="@item.content.discussionId" data-commentcount-format="contentImmersive" data-discussion-closed="@{!item.trail.isCommentable}"></div>
+        <div class="meta__number meta__number--commentcount">
+            <a href="#comments" data-link-name="Comment count" class="commentcount2 tone-colour">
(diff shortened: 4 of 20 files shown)


static/src/javascripts/projects/common/modules/discussion/comment-count-widget.js
```diff
@@ -0,0 +1,79 @@
+/* globals guardian */
+define([
+    'comment-count',
+    'common/utils/get-property',
```
regiskuckaertz: You're removing this module in another PR 😆

static/src/javascripts/projects/common/modules/discussion/comment-count-widget.js
```diff
@@ -0,0 +1,79 @@
...
+
+    function init(overrideConfig, overrideAB, overrideBlocker) {
+        var abtest = overrideAB || ab;
+        var blocker = overrideBlocker || CommentBlocker;
+        var widgetConfig = assign({
+            apiBase: getProperty(guardian, 'config.page.ajaxUrl') + '/discussion/comment-counts.json',
+            apiQuery: 'shortUrls',
+            filter: function (node) {
+                var shortUrl = node.getAttribute('data-loyalty-short-url');
+                if (shortUrl && abtest.isInVariant('ParticipationDiscussionTest', 'variant-1') && blocker.hideComments(shortUrl.replace('/p/','')) ) {
```
NathanielBennett: You can get rid of this now, I've taken that a/b test down

static/src/javascripts/projects/common/modules/discussion/comment-count-widget.js
```diff
@@ -0,0 +1,79 @@
...
+    'common/utils/get-property',
+    'fastdom',
+    'common/modules/experiments/ab',
+    'common/modules/experiments/tests/utils/comment-blocker',
+    'common/utils/assign',
+    'common/utils/formatters',
+    'common/utils/mediator',
+    'common/utils/report-error'
+], function (
+    commentCountWidget,
```
NathanielBennett: Am I missing something here(I'm fairly sure I am)? Isn't this file comment count widget?

static/src/javascripts/projects/common/modules/discussion/comment-count-widget.js
```diff
@@ -0,0 +1,79 @@
+/* globals guardian */
+define([
+    'comment-count',
+    'common/utils/get-property',
+    'fastdom',
+    'common/modules/experiments/ab',
+    'common/modules/experiments/tests/utils/comment-blocker',
```
NathanielBennett: You can blat these two, too

static/src/javascripts/projects/common/modules/discussion/comment-count-widget.js
```diff
@@ -0,0 +1,79 @@
...
+    'common/utils/assign',
+    'common/utils/formatters',
+    'common/utils/mediator',
+    'common/utils/report-error'
+], function (
+    commentCountWidget,
+    getProperty,
+    fastdom,
+    ab,
+    CommentBlocker,
```
NathanielBennett: And these

applications/app/views/fragments/crosswords/crosswordMetaHeader.scala.html
```diff
@@ -37,7 +37,11 @@ <h1 itemprop="headline" class="content__headline js-score">@Html(crosswordPage.c
                     <div class="meta__save-for-later js-save-for-later" data-position="top"></div>
                 }
                 <div class="meta__numbers modern-visible">
-                    <div class="u-h meta__number" data-discussion-id="@crosswordPage.content.content.discussionId" data-commentcount-format="content" data-discussion-closed="@{crosswordPage.content.trail.isCommentable}">
+                    <div class="meta__number">
+                        <a href="#comments" data-link-name="Comment count" class="commentcount2 tone-colour">
+                            <h3 class="commentcount2__heading">@fragments.inlineSvg("comment-16", "icon", List("inline-tone-fill")) Comments</h3>
```
regiskuckaertz: Feels a bit unjustified to use a heading here. This is just a number after all, it's not really worth appearing in the document outline (and the accessibility tree, with potentially dozens of identical "Comments", will only slow down people browsing by heading)

applications/app/views/fragments/crosswords/crosswordMetaHeader.scala.html
```diff
@@ -37,7 +37,11 @@ <h1 itemprop="headline" class="content__headline js-score">@Html(crosswordPage.c
...
+                            <comment-count class="commentcount2__value tone-colour" data-discussion-id="@crosswordPage.content.content.discussionId" data-discussion-closed="@{crosswordPage.content.trail.isCommentable}"></comment-count>
```
regiskuckaertz: This is a custom element, so why bother using `data-*` attributes? You can just:

`<comment-count discussion="1234" closed></comment-count>`

piuccio:
thanks guys for the comments, I had another chat with @sndrs on the use of fetch. Closing this for now, I'll update and reopen

applications/app/views/fragments/crosswords/crosswordMetaHeader.scala.html
piuccio: Had a look at the [spec](https://www.w3.org/TR/custom-elements/), very informative, quote

> An autonomous custom element does not have any special meaning: it represents its children. A customized built-in element inherits the semantics of the element that it extends.

> Any namespace-less attribute that is relevant to the element's functioning, as determined by the element's author, may be specified on an autonomous custom element, so long as the attribute name is XML-compatible and contains no uppercase ASCII letters. The exception is the is attribute, which must not be specified on an autonomous custom element (and which will have no effect if it is).

> Customized built-in elements follow the normal requirements for attributes, based on the elements they extend. To add custom attribute-based behavior, use data-* attributes.



In short

```
<custom-element whatever="data" />
<button is="custom" data-whatever="data" />
```

piuccio:
I've rebased and fixed some things. The widget now uses dependency injection for `fetch` and `Promise`.
The only other dependency, `fastdom` is a peer dependency.

The size of __ enhanced/main.js__ changed a bit in the mean time, but the same considerations hold

|          | New    | Old    |
|----------|--------|--------|
| Minified | 89.352 | 89.730 |
| Gzipped  | 26.704 | 26.561 |

@michaelmcnamara can you show me how to test on IE please?

johnduffell:
@piuccio use browserstack if you just need to run it, otherwise I think mike has a windows box on the end of his bank

piuccio:
Tested on IE, it breaks because of #13818

piuccio:
#13818 merged and `comment-count-widget` updated.

Tested on IE and works fine now.

Only missing a +1

johnduffell:
1 :+1:great although your build is failing!

PR #13715, opened 2016-07-25
revert hide-comments-test

NathanielBennett:
## What does this change?

Takes down the ab-test which hid half of comments for a percentage of users. As created here: https://github.com/guardian/frontend/pull/12990

The test is done so we need to remove the code.

N/A
## Screenshots

N/A
## Request for comment

@piuccio : This is what we were discussing ( no pun intended ) earlier @crifmulholland


piuccio:
👍

PR #13716, opened 2016-07-25
keep panic shedding or not?

johnduffell:
extend or kill?  I'm tempted to kill it, but this PR was a simpler way to get the discussion going.
- panic shedding kicks in almost every day to take load off servers
- however most of those would not have been fatal even if it were left to serve all results
- we found that servers still couldn't come up from cold even when panic shedding was working (the whole server seemed to just go slow for a short time, and possibly get overwhelmed at a lower level than play)
- in the last month we haven't had any alerts (because of less traffic?)

My feeling is kill it and make a better solution e.g. only allow instance to pass health checks when there at least 3 servers ready, or wait for the long awaited ELB improvements to prevent this.

Just tagging all the people who commented on the original PRs in case you have opinions suggestions - @TBonnin @kelvin-chappell @rich-nguyen @alexduf @jamespamplin 

Pretty graph of requests made during high latency conditions in the last fortnight
image


TBonnin:
I would say `kill it` 🔪 . We are not sure of its usefulness and it make the code more complicated.

However if you are not 100% confident: what about disabling it and extending the switch for a couple of weeks? If we don't notice any issue during this period of time, it would mean it is not useful and you can then kill it for real.

kelvin-chappell:
I agree. Kill it.
It has a complicated effect and it's doing work that the elb should be doing.

rich-nguyen:
I would try another approach, yes. It's still an outstanding platform issue.

johnduffell:
Thanks for the comments, in the light of those I'm going to be conservative and leave it switched off for a short time and then remove assuming everything doesn't fall over every day. I think @TBonnin 's changes to the healthchecks (not to depend on upstream) probably had a bigger effect on reliability.  I'm extending for 4 weeks because that should give us enough time to know what happened, but I could go with a little less?

Anyone give a +1 to that course of action?

johnduffell:
they are off now! 😨 
image

johnduffell:
so the first lunch time after I turned it off, applications died in the classic fashion, so I've turned it on again

johnduffell:
ok after the lunch time outage I discussed with @TBonnin and the suggestion is to extend a month, and try turning it off at the end of the month for a week or so, and see if it happens again.  If so, and it's been generally fine otherwise, we'll change to "never" expires.

Anyone give us a +1 or another suggestion?

kelvin-chappell:
No better suggestion.  This will give thinking time so 👍

alexduf:
I'm joining the conversation a bit late, and agree with the general consensus. Let's see how it behave when turned off and decide from there.

TBonnin:
@alexduf it was turned OFF this morning and later at lunch a few app instances started to misbehave (discussion, applications and onward I believe).
Either we were extremely unlucky today or PanicShedding is actually useful.

@johnduffell and I decided to turn it ON again, wait a week or two to see if the same issue happens again. If not we will try to turn it off again for a final test.

PR #10068, opened 2015-08-07
Image cleanup

gklopper:
Deleting old image service code now that we have moved to the new service.

This **cannot be merged** till some config has rolled out for our CI servers.

Likewise devs now need the following config (it was optional till now)

```
images.media.token=XXX
images.static.token=YYY
```

PR #13719, opened 2016-07-25
clean ajax utility

piuccio:
`imgix` switch was deleted in August by #10068

`getProperty` is cool, but it was only used in one place


sndrs:
:kiss: :+1:

PR #13720, opened 2016-07-26
Extending new header dummy test for a week

NataliaLKB:
Extending the new header dummy test for one week. 
### Why a dummy test?

See original description here https://github.com/guardian/frontend/pull/13676
### Why is this being extended?

I believe the fastly change was undone last week while I was on holiday. 

This was the PR: https://github.com/guardian/fastly-edge-cache/pull/659

As far as I understand there was extra load on CAPI and an older version of fastly was deployed. @stephanfowler will possibly know more. 

Before I deploy the fastly change again (thus creating the buckets for the test) I will check with the data team and dotcom health to get a better understanding of the impact and whether this was the cause of the CAPI problems last week. 

This PR is really just to extend the switches so the build doesn't break tomorrow 😄


stephanfowler:
👍

PR #13726, opened 2016-07-26
Hosted CSS refactor

lps88:
## What does this change?
- try to use brand colour css classes, instead of applying colour to specific components
- remove brand colour references from gallery style sheet
- remove old thrasher css, html & js
- fix imports so we only import hosted css once, and only on hosted pages

Aiming to be able to pass the colour into the template, and not write custom css per campaign.
## Request for comment

@guardian/labs-beta


Calanthe:
I would prefer to have `border-top: 1px solid $omgbColor;` not only `border-color: $omgbColor;` in `_hosted-colours.scss` so we can include that file on top of other .scss files. But I know this is just a temporary solution so let's remember about that later while doing it the proper way - with composer.

👍

PR #13727, opened 2016-07-26
Fix: space between body adverts

regiskuckaertz:
This is not enough:

picture 1

That's because the height of the creative is not factored in `filter` when deciding whether there is enough space between two candidates.

cc @guardian/commercial-dev

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/article-body-adverts.js
+++ b/static/src/javascripts/projects/common/modules/commercial/article-body-adverts.js
@@ -4,6 +4,7 @@ define([
     'common/utils/config',
     'common/utils/detect',
     'common/modules/article/space-filler',
+    'common/modules/commercial/ad-sizes',
     'common/modules/commercial/dfp/add-slot',
     'common/modules/commercial/dfp/track-ad-render',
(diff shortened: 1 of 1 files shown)


static/src/javascripts/projects/common/modules/commercial/article-body-adverts.js
```diff
@@ -72,7 +74,7 @@ define([
         if (!longArticleRules) {
             longArticleRules = getRules();
             longArticleRules.selectors[' .ad-slot'].minAbove =
-            longArticleRules.selectors[' .ad-slot'].minBelow = 1300;
+            longArticleRules.selectors[' .ad-slot'].minBelow = detect.getViewport().height;
```
rich-nguyen: do you want to have a `min` on this?

rich-nguyen:
👍

static/src/javascripts/projects/common/modules/commercial/article-body-adverts.js
regiskuckaertz: I don't understand 😅

PR #13728, opened 2016-07-26
Removing WS.url #1

TBonnin:
## What does this change?

First batch of removing calls to WS.url()
I would strongly advise to review the commits one by one. They should be self contained.

No more WS.url which depends on `Play.current`. (`Play.current` being deprecated in Play 2.5)

No
## Request for comment

@alexduf


admin/app/controllers/cache/ImageDecacheController.scala
```diff
@@ -40,10 +39,11 @@ class ImageDecacheController extends Controller with Logging with AuthLogging wi
 
       val originUri = new URI(originUrl)
 
-      ImageServices.clearFastly(originUri)
-      ImageServices.clearImgix(originUri)
+      val imageServices = new ImageServices(wsClient)
```
alexduf: I don't know if decache is called intensibely, if it is it may be nicer to put the object creation outside the method (as an attribute of the controller for instance)

alexduf:
Out of curiosity, why do you use case classes for services or jobs ?

alexduf:
otherwise, looks good :+1:

admin/app/controllers/cache/ImageDecacheController.scala
TBonnin: Good point. I don't think it is used a lot but I will extract the creation outside of the method

TBonnin:
@alexduf I cannot remember to be honest. Let me not use case classes to see if I can recall why. 😁

TBonnin:
@alexduf there were no specific reason (other than me being lazy and avoiding using `new`)

I rebased master to take into account `AkkaAsync` changes and made 2 additional commits:
- Move the creation of the admin jobs into the apploader and then pass it down. Class like `VideoEncodingsJob` actually hold a state so we need to only use one.
- Move `EmailFormService` and `ImageServices` creation outside of a method

admin/app/model/AdminLifecycle.scala
```diff
@@ -10,13 +10,18 @@ import conf.switches.Switches._
...
 import scala.concurrent.{ExecutionContext, Future}
 
-class AdminLifecycle(appLifecycle: ApplicationLifecycle, jobs: JobScheduler, akkaAsync: AkkaAsync, emailService: EmailService, wsClient: WSClient)(implicit ec: ExecutionContext) extends LifecycleComponent with Logging {
+class AdminLifecycle(appLifecycle: ApplicationLifecycle,
+                     jobs: JobScheduler,
+                     akkaAsync: AkkaAsync,
+                     emailService: EmailService,
+                     fastlyCloudwatchLoadJob: FastlyCloudwatchLoadJob,
+                     r2PagePressJob: R2PagePressJob,
+                     videoEncodingsJob: VideoEncodingsJob)(implicit ec: ExecutionContext) extends LifecycleComponent with Logging {
```
alexduf: [nitpick] I usually like when parameters are indented like that
```scala
class AdminLifecycle(
  appLifecycle: ApplicationLifecycle,
  jobs: JobScheduler,
  akkaAsync: AkkaAsync,
  emailService: EmailService,
  fastlyCloudwatchLoadJob: FastlyCloudwatchLoadJob,
  r2PagePressJob: R2PagePressJob,
  videoEncodingsJob: VideoEncodingsJob
)(implicit ec: ExecutionContext)
```
that's entirely aesthetic no need to actually fix that.

alexduf:
It looks even better with the services injected rather than created by their client.
Looks great :+1:

admin/app/model/AdminLifecycle.scala
TBonnin: 👯  We have different taste. Probably some leftover from my Objective-C days.

PR #13729, opened 2016-07-26
Simplest ever fetch polyfill

piuccio:
## What does this change?

Introduce the simplest (most stupid) `fetch` polyfill. Internally it uses `reqwest` so there are no extra dependencies, but provides the same signature as the standard [fetch](https://developer.mozilla.org/en-US/docs/Web/API/GlobalFetch/fetch).

There's also a utility function to fetch JSON.

In the future, when we introduce a proper polyfill (maybe through @janua 's ESI) we can simply remove `common/utils/fetch` or change the require to `proper-polyfill/fetch`.

First step in removing `reqwest`. Success means weather widget still works. If positive I'll convert some other modules to `fetch` (maybe).

No 
## Request for comment

@sndrs @regiskuckaertz

### DIFF
--- a/static/src/javascripts/projects/common/utils/fetch.js
+++ b/static/src/javascripts/projects/common/utils/fetch.js
@@ -70,7 +70,7 @@ define([
 
         function text () {
             var result = bodyRead ? Promise.reject(new TypeError('Already read')) : Promise.resolve(body);
-            bodyRead = false;
+            bodyRead = true;
             return result;
         }
 
--- a/static/src/javascripts/projects/common/utils/fetch.js
+++ b/static/src/javascripts/projects/common/utils/fetch.js
@@ -1,0 +1,90 @@
+define([
+    'reqwest',
+    'Promise'
+], function (
+    reqwest,
+    Promise
+) {
+    /**
+     * Provide a minimal function equivalent to fetch. I don't dare calling it a
+     * polyfill but the signature is the same, albeit simplified.
+     *
+     * fetch (input, init): Promise
+     *
+     * Differences with the standard fetch are
+     *
+     * - fetch
+     * - - input can only be a string, Request is not supported
+     * - - does not support JSONP
+     * - - headers can't be specified
+     * - Request
+     * - - there's no way to send a body
+     * - Response
+     * - - blob is not supported
+     * - - formData is not supported
+     * - - headers are not populated
+     *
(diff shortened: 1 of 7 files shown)


static/src/javascripts/projects/common/utils/fetch.js
```diff
@@ -0,0 +1,90 @@
...
+        };
+    }
+
+    function createResponse (response) {
+        var bodyRead = false;
+        var body = response.responseText;
+
+        function text () {
+            var result = bodyRead ? Promise.reject(new TypeError('Already read')) : Promise.resolve(body);
+            bodyRead = false;
```
regiskuckaertz: `true`?

piuccio: 😲  the one thing I didn't write a test for, thanks for spotting

piuccio: fixed

regiskuckaertz:
Don't you want to update the 50 uses of the `ajax` module? 😅

static/src/javascripts/projects/common/utils/fetch.js
```diff
@@ -0,0 +1,90 @@
...
+     * If you're still wondering what it actually supports
+     * - CORS
+     * - credentials
+     * - response.text() and response.json()
+     * - response.ok .status .statusText
+     */
+    function fetch (input, init) {
+        return new Promise(function(resolve, reject) {
+            var req = buildRequest(input, init || {});
+            reqwest(req)
```
regiskuckaertz: Not wrapped with `raven`?

piuccio: I  don't think `fetch` should be wrapped, it's a polyfill afterall, maybe `fetch-json`but I don't really see the point. Aren't we reporting uncaught promises already using `when.js`?

regiskuckaertz: Fair enough, I also think this kind of error tracking should happen at a higher level.

regiskuckaertz:
👍  LGTM

sndrs:
> If positive I'll convert some other modules to fetch (maybe).

if we want to replace `ajax-promise` with `fetch`, we need to remove `ajax-promise`

👍 on the basis you'll do that 😉

PR #13731, opened 2016-07-26
Addition of paying member status to feedback email

scottp719:
## What does this change?

Adds the paying member status to the mailto footer of tech feedback and userhelp mails

Provides us with more information on the user at initial point of contact, should help reduce back and forth with users and flag paying member status early.

нет


katebee:
👍

PR #13687, opened 2016-07-21
My Digest and Weekend Reading

janua:
## What does this change?

This adds two pages/workflows; `weekend reading` and `my digest`.
### Weekend Reading

This page is a 404 with a sign up form to register interest in the feature.

The flow is `Weekend Reading with Signup` -> `Thank You`. There is an option to look at a sample from the page.
### My Digest

This is an explainer page, which then takes you to a 404 page to register interest.

The flow is as follows: `My Digest explainer` -> `404 with signup` -> `Survey on what the user would like` -> `Thank You`.
## What is the value of this and can you measure success?

The value of these pages is to gauge interest and to either validate or invalidate the concept. Success will be measured by how much interest we see from users.
## Does this affect other platforms - Amp, Apps, etc?

This is running from `applications`. But should not affect anything.
## Screenshots

weekendreading2

mydigest
## Request for comment
#13659 is dependant on this PR.

@superfrank @joelochlann @katebee

PR #13732, opened 2016-07-26
Inline survey CSS to prevent ugly flash of unstyled content

joelochlann:
The survey page template added in #13448 and used in #13687 doesn't inline all the styles needed to render the above-the-fold content, which means [this code](https://github.com/guardian/frontend/blob/2e00099b6509024fd5a9f04aa7daea03e08281ac/common/app/views/fragments/stylesheetLink.scala.html#L6) causes a perceptible delay before [this JS](https://github.com/guardian/frontend/blob/2e00099b6509024fd5a9f04aa7daea03e08281ac/common/app/templates/inlineJS/blocking/enableStylesheets.scala.js#L30) fires and styles the page properly.

I had this same issue with the membership tests in #12357 but didn't get to the bottom of it then. Seems likes it's quite easy for the `media="only x"` async CSS mechanism to backfire if a developer hasn't considered what's above & below the fold for a new page.
#### before

picture 234
#### after

picture 235

@janua 
@sndrs please check that I'm not abusing the CSS inlining mechanism here. also interested in any suggestions for how we could make it harder for developers not used to frontend to accidentally delay rendering of essential styles


joelochlann:
also @desbo because I see you dealt with an async CSS issue in the past: https://github.com/guardian/frontend/commit/0a59d0cca493354d4c6d1e14d327467237d6e77f

static/src/stylesheets/head.survey.scss
```diff
@@ -40,3 +40,9 @@
    ========================================================================== */
 
 @import 'module/atoms/_quiz';
+
+/* ==========================================================================
+   Survey
+   ========================================================================== */
+
+@import 'module/survey/survey';
```
janua: @joelochlann Does placing it here now mean we get all these other styles we don't need?

joelochlann: No more than we were already getting. Previously because of [this code](https://github.com/guardian/frontend/blob/master/common/app/assets/assets.scala#L116) we would get [head.content.scss](https://github.com/guardian/frontend/blob/master/static/src/stylesheets/head.content.scss) inlined, as well as [survey.scss](https://github.com/guardian/frontend/blob/master/static/src/stylesheets/survey.scss) loaded asynchronously. Since `head.survey` is just a copy of `head.content` plus the line above, we're not getting anything extra (in fact we're getting less because we're not getting [this stuff](https://github.com/guardian/frontend/blob/master/static/src/stylesheets/survey.scss#L1))

There is a question about how much of the stuff copied from `head.content` we actually need. We could try chopping out lines to see if anything's affected on the survey pages. I'd rather leave that for a separate PR though

sndrs:
that looks ok, as long as it's only going on survey pages.

as for:

> Seems likes it's quite easy for the media="only x" async CSS mechanism to backfire if a developer hasn't considered what's above & below the fold for a new page.

a developer _should_ consider that...

joelochlann:
@sndrs yep you're right. I guess I mean it's easy for a developer to add styles to this file assuming that they *will* block rendering. I've added some comments that will hopefully make people think twice

PR #13733, opened 2016-07-26
Give site messages an optional data-component

katebee:
## What does this change?
- Adds an option to the Message prototype so that a `data-component` can be passed as an option
- The `data-component` encapsulates all the elements and `data-link-names` associated with a site message. Different site messages can have different strings for the `data-component`
- The habit-forming-digest message is now using the cool new data-component option
- Existing site messages are not affected; the `data-component` is an optional param

The outer-most DIV of a site message can have a data-component! 🎉 

If you use `data-component` alongside `data-link-name` then queries for the data can be much simpler.

No
## Screenshots

picture 137

Change to the query string parameters sent to Ophan:

picture 138
## Request for comment

Thanks to @janua :clap:


janua:
Yes!! :+1:

PR #13734, opened 2016-07-26
Adding edition picker to the new nav

NataliaLKB:
## What does this change?

Making the edition picker in the new nav work. It is important that this works without JavaScript, as this will also be available to core users and no js users.

My [first attempt](https://github.com/guardian/frontend/pull/13617) at this was using the `<select>` and `<option>` elements. However I ran into some [accessibility concerns with IE](https://github.com/guardian/frontend/pull/13617#discussion_r70998183). 

I am aiming to run an a/b test for this in the next couple weeks!

Nope
## Screenshots

editionpickerstyled


static/src/stylesheets/layout/_ab-new-header-test-variant.scss
```diff
@@ -241,6 +204,93 @@ $gutter-medium: 37px;
...
+
+    &:hover {
+        fill: #ffffff;
+    }
+}
+
+.edition-picker__checkbox {
+    display: none;
+
+    &:checked {
```
regiskuckaertz: This is smart! To make this link explicit, you can also add the `aria-controls` attribute and then add/remove the `hidden` attribute on the list.

NataliaLKB: Yes that makes sense! I will do that

NataliaLKB: Changes we discussed have been made. Thanks for those suggestions @regiskuckaertz

static/src/javascripts/projects/common/modules/navigation/edition-picker.js
```diff
@@ -0,0 +1,67 @@
+define([
+    'common/utils/fastdom-promise',
+    'common/utils/$'
+], function (
+    fastdomPromise,
+    $
+) {
+    var dropdown = $('.js-edition-picker-dropdown');
+
+    function enhanceToButton() {
```
gtrufitt: Well, this is just beautiful, my only bother when reading the mark-up that it wasn't a button, but lo-and-behold you have it covered ✊

static/src/javascripts/projects/common/modules/navigation/edition-picker.js
```diff
@@ -0,0 +1,67 @@
...
+    function openMenuOnClick() {
+        var button = $('.js-click');
+
+        if (button.length > 0) {
+            button[0].addEventListener('click', function() {
+
+                fastdomPromise.write(function() {
+                    if (button.hasClass('open')) {
+                        button.removeClass('open');
+                        dropdown.attr('hidden');
```
gtrufitt: Does this set it as expected? I thought a single value in `attr` was a getter only but maybe it does something special with boolean attributes

NataliaLKB: Fixed @gtrufitt 😄

static/src/javascripts/projects/common/modules/navigation/edition-picker.js
```diff
@@ -0,0 +1,69 @@
...
+        fastdomPromise.write(function() {
+            checkBox.remove();
+            container.prepend(button);
+
+            openMenuOnClick();
+        });
+    }
+
+    function openMenuOnClick() {
+        var button = $('.js-click');
```
SiAdcock: Maybe this class could be more descriptive, or more certain to be unique to the `edition-picker`?

NataliaLKB: good shout! I will change it 😄

static/src/javascripts/projects/common/modules/navigation/edition-picker.js
```diff
@@ -0,0 +1,69 @@
...
+            });
+        }
+    }
+
+    function clickMenuOnEnter() {
+        var label = document.getElementsByClassName('js-on-enter-click')[0];
+
+        if (label) {
+            label.onkeypress =  function clickOnEnter(event) {
+                if((event.keyCode ? event.keyCode : event.which) == 13) {
```
SiAdcock: Triple eqs all the way! `===`

static/src/stylesheets/layout/_ab-new-header-test-variant.scss
```diff
@@ -8,8 +8,9 @@ $mobile-medium: 375px; // Breakpoint for our most common device sizes
 }
 
 .new-header__logo-wrapper {
-    padding-top: $gs-baseline / 2;
-    padding-bottom: $gs-baseline / 2;
+    padding-top: 6px;
```
SiAdcock: I think `gs-baseline` is `12px`, so we can stick with `$gs-baseline / 2`

static/src/javascripts/projects/common/modules/navigation/edition-picker.js
```diff
@@ -0,0 +1,69 @@
...
+                        button.attr('aria-expanded', 'true');
+                        dropdown.removeAttr('hidden');
+                    }
+                });
+            });
+        }
+    }
+
+    function clickMenuOnEnter() {
+        var label = document.getElementsByClassName('js-on-enter-click')[0];
```
SiAdcock: Is it possible for an element with this class to *not* be present on the page? If so, this line will fall over (whereas the pattern used on lines 29 - 32 above is more defensive)

NataliaLKB: It is always possible! I made some amendments. Thanks 😄

SiAdcock:
Very cool stuff 👍

PR #13735, opened 2016-07-27
Fix: Missing trait added to dev-build AppComponents

TBonnin:
## What does this change?

Fix issue Introduced by #13693

dev-build works again

No
## Request for comment

@alexduf


alexduf:
:+1:

PR #13736, opened 2016-07-27
remove join discussion after poll test

GHaberis:
## What does this change?

Removes join discussion after poll test


GHaberis:
Hi @crifmulholland @gtrufitt, this participation test expires today so I'm removing it, this is all I need to do right?

gtrufitt:
Yup, think so 👍

PR #13738, opened 2016-07-27
Add WithTestWsClient trait and fix HealthCheckTest

TBonnin:
## What does this change?
- Add WithTestWsClient trait so tests can access an instance of `WSClient` rather than calling `WS.url`

=> Play 2.5
Cleaning more call to `WS.url()`

NO
## Request for comment

@alexduf


common/test/package.scala
```diff
@@ -59,7 +60,11 @@ trait TestSettings {
   ContentApiClient.jsonClient._http = toRecorderHttp(ContentApiClient.jsonClient._http)
 }
 
-trait ConfiguredTestSuite extends ConfiguredServer with ConfiguredBrowser with ExecutionContexts {
+trait WithTestWsClient {
+  lazy val wsClient = NingWSClient(NingWSClientConfig(maxRequestRetry = 0))
```
alexduf: Two things here

- NingWsClient will need to be replaced by AhcWsClient when bumping to play 2.5. There's no way around it.
- When creating a wsClient yourself, you need to handle its destruction.

Here's an extract of the docs

> If you create a WSClient manually then you must call client.close() to clean it up when you’ve finished with it.

you can find it here https://www.playframework.com/documentation/2.5.x/ScalaWS#Using-WSClient

TBonnin:
@alexduf made the change to close the wsClient. Feedback welcome :)

common/app/common/Lazy.scala
```diff
@@ -0,0 +1,20 @@
+package common
+
+import scala.language.implicitConversions
+
+object Lazy {
+  def lazily[A](f: => A): Lazy[A] = new Lazy(f)
```
alexduf: I would rename this one `apply`

common/app/common/Lazy.scala
```diff
@@ -0,0 +1,20 @@
...
+  implicit def evalLazy[A](l: Lazy[A]): A = l()
+}
+
+class Lazy[A] private(f: => A) {
+  private var option: Option[A] = None
+
+  def apply(): A = option match {
```
alexduf: but this one `.get` :)

alexduf:
The two comments I left are really minor, please feel free to merge as it looks good! Good idea to wrap this in a type :+1:

common/app/common/Lazy.scala
TBonnin: Make sense

common/app/common/Lazy.scala
TBonnin: Good point

alexduf:
:star: :+1: :star:

PR #13739, opened 2016-07-27
Remove "hide" toggle on multi campaign containers 

regiskuckaertz:
Paid containers should never have a "hide" toggle, yet multi campaigns erroneously do:

picture 1

The root of this comes from two small errors:
1. [each card](https://github.com/guardian/frontend/compare/rk-multi-campaign-hide?expand=1#diff-cba913f7f475ab49435b6f99beffaabfL39) in the container has the `js-container__header` class, which is a hook used to insert the toggle
2. the container itself has the `js-container--toggle` [although it shouldn't](https://github.com/guardian/frontend/blob/master/common/app/views/support/GetClasses.scala#L117). This is because currently [multi campaigns are "empty" commercial containers](https://github.com/guardian/frontend/blob/master/common/app/layout/Front.scala#L258).

After fixing these two issues:

picture 2

cc @kelvin-chappell @JonNorman

### DIFF
--- a/common/app/layout/Front.scala
+++ b/common/app/layout/Front.scala
@@ -120,6 +120,7 @@ object ContainerCommercialOptions {
         isSponsored = dfpTag.paidForType == Sponsored,
         isPaidContainer = dfpTag.paidForType == AdvertisementFeature,
         isFoundationSupported = dfpTag.paidForType == FoundationFunded,
+        hasPaidForContent = false,
         sponsor,
         sponsorshipTag = Some(SponsorshipTag(dfpTag.tagType, capiTagId)),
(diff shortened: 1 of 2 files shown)


kelvin-chappell:
Good detective work! 👍 
The multi-campaign container should be going soon, when we turn it on as a standard feature of all paid containers.

regiskuckaertz:
Thanks @kelvin-chappell! I'll leave it up to you to decide then whether to keep the `hasPaidForContent` flag or not.

kelvin-chappell:
I think we should be able to get rid of `ContainerCommercialOptions` altogether in the bright new world.

PR #13740, opened 2016-07-27
Remove !important from AMP styles

SiAdcock:
## What does this change?
1. Prevent liveblog CSS from being included in non-liveblog articles
2. Remove `!important` from AMP liveblog stylesheet, as these break the AMP validator
## Request for comment

@NataliaLKB


NataliaLKB:
👍

johnduffell:
oooh good spot 👍  and the CSS packaging/building is non existent mostly because I hadn't a clue how to do any of that and we never got round to refactoring it, so feel free to make the system better

PR #13744, opened 2016-07-27
Pipe correct badges behaviour in multi-campaigns

regiskuckaertz:
Logos are not appearing following [the removal of the `js-container__header` class](https://github.com/guardian/frontend/pull/13739), but that was the wrong class anyway. Instead, we're using card-related class names to feed badges.

cc @kelvin-chappell


kelvin-chappell:
👍

PR #13746, opened 2016-07-27
Add empty r2.press queues url properties in dev

TBonnin:
## What does this change?

Add mising properties for r2.press queues in dev (empty)

Dev-build doesn't throw error
## Request for comment

@NataliaLKB


NataliaLKB:
👍

PR #13747, opened 2016-07-27
Hosted Gallery/Video pages - social sharing fixes

lps88:
## What does this change?
- pull out the social buttons html from the gallery/video page templates
- put the relevant fields into HostedPage class, instead of defining them differently per page type
- properly encode the share urls

cc @guardian/labs-beta


Calanthe:
👍  great! I need to apply those changes in my article page now.

PR #13748, opened 2016-07-27
Removing call to WS.url (Continue #3)

TBonnin:
## What does this change?

Removing call to WS.url in sport app:
- Football PaFeed
- Cricket PaFeed
- Rugby Feed

=> Play 2.5

No
## Request for comment

@alexduf

### DIFF
--- a/sport/app/rugby/feed/OptaFeed.scala
+++ b/sport/app/rugby/feed/OptaFeed.scala
@@ -1,14 +1,9 @@
 package rugby.feed
 
 import common.{ExecutionContexts, Logging}
-import play.api.Play.current
-import play.api.libs.ws.{WSRequest, WS}
-import rugby.jobs.RugbyStatsJob
+import play.api.libs.ws.WSClient
 import rugby.model._
-import conf.Configuration
-
 import scala.concurrent.Future
-import scala.xml.XML
 
 sealed trait OptaEvent {
   def competition: String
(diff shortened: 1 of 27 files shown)


sport/app/rugby/feed/OptaFeed.scala
```diff
@@ -24,7 +19,7 @@ case object WorldCup2015 extends OptaEvent {
 
 case class RugbyOptaFeedException(message: String) extends RuntimeException(message)
 
-object OptaFeed extends ExecutionContexts with Logging {
+case class OptaFeed(wsClient: WSClient) extends ExecutionContexts with Logging {
```
alexduf: I think a `class` would be more appropriate than a `case class` here as its only state is the `wsClient.

sport/app/rugby/jobs/RugbyStatsJob.scala
```diff
@@ -2,27 +2,24 @@ package rugby.jobs
...
 import rugby.model._
 import rugby.feed.{MatchNavigation, OptaFeed, OptaEvent, RugbyOptaFeedException}
 import scala.concurrent.Future
 import scala.util.Failure
 import scala.util.Success
 
-object RugbyStatsJob extends RugbyStatsJob
 
-trait RugbyStatsJob extends ExecutionContexts with Logging {
+case class RugbyStatsJob(optaFeed: OptaFeed) extends ExecutionContexts with Logging {
```
alexduf: same here `case class` => `class`

sport/test/package.scala
```diff
@@ -83,11 +82,11 @@ object FeedHttpRecorder extends HttpRecorder[WSResponse] {
 }
 
 // Stubs data for Football stats integration tests
-object TestHttp extends Http with ExecutionContexts {
+case class TestHttp(wsClient: WSClient) extends Http with ExecutionContexts {
```
alexduf: and here

alexduf:
:+1:

PR #13749, opened 2016-07-27
use configuration magic

johnduffell:
This changes frontend to use https://github.com/guardian/configuration-magic instead of https://github.com/guardian/guardian-configuration

In itself it's not really a useful change (it's part 1 of 3ish), but config magic has libraries to load config from dynamodb rather than off disk.
## Why?

We always encrypt our data, but the first rule of [securing data at rest](https://www.owasp.org/index.php/Cryptographic_Storage_Cheat_Sheet) is actually "don't store it in the first place"

> Rule - Only store sensitive data that you need

Hence, I think we should store all prod (and dev?) keys in one secure location, with other copies only in memory while they are needed.
## next step

In future we can use config magic's facility to load prod (and dev?) config from AWS dynamodb instead of from local disk.  Using AWS to store config is a good thing because AWS access is well controlled.

I also considered using S3 instead of dynamodb, but it's much easier to edit using AWS console.
The downside of not having a copy (of dev credentials) is it's harder to spin up the system when you're developing with no internet access, or didn't grab janus credentials, but the default situation should be to be secure and necessary keys can be added for offline usage.  Most keys you'd need to be online to actually use them anyway.
Also there's a downside that there's no built in approval or versioning in dynamodb, so we don't know who/why things were changed.

@TBonnin @adamnfish @rich-nguyen @alexduf  as I say this PR itself isn't too exciting, but I'd love some pointers to refine the idea - especially I need to think about the dev side of it


common/app/common/configuration.scala
```diff
@@ -21,11 +24,42 @@ class GuardianConfiguration extends Logging {
...
   case class OAuthCredentialsWithMultipleCallbacks(oauthClientId: String, oauthSecret: String, authorizedOauthCallbacks: List[String])
 
-  protected val configuration = ConfigurationFactory.getNonLoggingConfiguration("frontend", "env")
+  import com.gu.cm.{Configuration => CM}
+  import com.typesafe.config.Config
+  val configuration = {
 
-  private implicit class OptionalString2MandatoryString(conf: com.gu.conf.Configuration) {
-    def getMandatoryStringProperty(property: String) = configuration.getStringProperty(property)
+    val cmLogger = new Logger {
```
alexduf: There is a class that provides that implementation in the play module: `PlayDefaultLogger`. It's named play-24 but don't be afraid to use it as it works with play 2.5

common/app/common/configuration.scala
```diff
@@ -21,11 +24,42 @@ class GuardianConfiguration extends Logging {
...
+      override def warn(message: => String): Unit = log.warn(message)
+      override def error(message: => String): Unit = log.error(message)
+      override def error(message: => String, exception: => Throwable): Unit = log.error(message, exception)
+      override def info(message: => String): Unit = log.info(message)
+    }
+
+    val stage = new CM(List(FileConfigurationSource(s"/etc/gu/install_vars")), cmLogger).load.getStringProperty("STAGE").getOrElse("DEV")
```
alexduf: I recommend using `Identity.whoAmI` here as it will automatically use our convention to discover where is the machine running. Provided the instance is tagged correctly, it will populate all the fields of the `Identity` case class.

alexduf:
Let's talk about it IRL, I think it should be done at application startup rather than here.

alexduf:
Also I was planning on creating a module to integrate S3, but as you said it's better to use dynamodb for all the reasons you mentioned. When we have enough people using the library we can add a generic UI on top of the dynamodb table

TBonnin:
Regarding loading the config from dynamoDB:
Your concern about developing without internet access is already valid currently. Without Janus access you cannot get fresh aws credentials that would prevent a bunch of features not to work.

I am 100% for a centralized "service" that provide config regardless of the stage (DEV, TEST, CODE, PROD) and the app that is being run. It would be a plus security-wise but also in term of development process.
There is a few questions worth asking:
- Do we want the tests running on TeamCity to depends on config fetched from the network 
- How can we make it easy to add/remove/monitor config keys (for instance how do we ensure a key has values for each necessary environment for instance, etc...)

common/app/common/configuration.scala
johnduffell: yes good spot thanks for the pointer - I'd definitely like to use the facilities like that in the medium term, this first PR is just literally to switch the libraries over and use exactly the same config sources - then I will start changing the place it picks things up from

alexduf:
> Do we want the tests running on TeamCity to depends on config fetched from the network

The default behaviour of configuration magic is to load the configuration from the classpath when in `Test` mode. So it needs a configuration file with the test values (which should probably be in the repo)

> How can we make it easy to add/remove/monitor config keys (for instance how do we ensure a key has values for each necessary environment for instance, etc...)

configuration-magic doesn't ensure that, but I'd be happy to talk about it if we can find a way to implement it.

johnduffell:
yes agreed I think the teamcity tests should be testing the code using recorded data so the keys should be irrelevant, hopefully they should be deterministic and self contained therefore repeatable

as for required keys, at present if they are missing we just throw a runtime error, which we still have the same problem even after my changes.  As Alex says it'd be great to improve that, but that wasn't something I was specifically aiming for in this particular piece of work.

common/app/common/configuration.scala
```diff
@@ -1,33 +1,71 @@
...
+  lazy val configuration = {
+
+    val stage = new CM(List(FileConfigurationSource(s"/etc/gu/install_vars")), PlayDefaultLogger).load.getStringProperty("STAGE").getOrElse("DEV")
+    lazy val userPrivate = FileConfigurationSource(s"${System.getProperty("user.home")}/.gu/frontend.properties")
+    lazy val opsPrivate = FileConfigurationSource(s"/etc/gu/frontend.properties")
+    lazy val public = ClassPathConfigurationSource(s"env/$stage.properties")
+    new CM(List(userPrivate, opsPrivate, public), PlayDefaultLogger).load
+  }
+
+  implicit class ScalaConvertProperties(conf: Config) {
```
alexduf: I think you don't need any of these if instead of exposing a `Config` you expose a `play.api.Configuration` by doing `playConf = new Configuration(config)`

That should give you all the scala helpers you need.

admin/app/conf/AdminConfiguration.scala
```diff
@@ -1,14 +1,14 @@
...
 import conf.Configuration.OAuthCredentialsWithMultipleCallbacks
 import pa.PaClientConfig
 
 case class OmnitureCredentials(userName: String, secret: String)
 
 object AdminConfiguration {
-
-  val configuration = ConfigurationFactory.getConfiguration("frontend", "env")
+  import GuardianConfiguration._
+//  val configuration = ConfigurationFactory.getConfiguration("frontend", "env")
```
alexduf: did you voluntarily keep this one?

project/Dependencies.scala
```diff
@@ -25,6 +25,7 @@ object Dependencies {
   val awsElasticloadbalancing = "com.amazonaws" % "aws-java-sdk-elasticloadbalancing" % awsVersion
   val closureCompiler = "com.google.javascript" % "closure-compiler" % "v20150901"
   val commonsHttpClient = "commons-httpclient" % "commons-httpclient" % "3.1"
+  val commonsLang = "commons-lang" % "commons-lang" % "2.4"
```
alexduf: I couldn't find where this one was used, did I miss anything?

rich-nguyen:
I can see pros and cons, but at least with S3, you get versioning and does _feel_ simpler, and somehow adding a dependency on dynamo db to start up (we already have a dep on S3 to boot our servers) feels unwieldy. I think dynamo is the most flexible solution, but also the most complex. It would be good for adding metadata to keys too (stage, mandatory, article-server-only, etc).  

@alexduf hopefully the process is easy enough that we don't need a UI to update the props. Maybe something like goo could be used for get/set keys. The dynamo db console is probably less user-friendly than an s3 update.

johnduffell:
just testing it for a while

project/Dependencies.scala
johnduffell: that's used in a couple of places, forgot where, but it was relying on the transitive dependency brought in from guardian configuration

common/app/common/configuration.scala
johnduffell: discussed in person - the idea is to use CM in the old guardian conf way, just so we can get all the dynamo (or possibly s3) loading ability.  Perhaps later we can restructure so it fits in with the CM way of doing things, but for now I am happy with keeping the existing config file structure and just moving the minimal to AWS.

johnduffell:
I think based on the discussion s3 is probably slightly preferable if the tooling is there, but dynamo db has a built in key/value editor and already has the configuration magic loader so would be easier to get running.  Versioning is I think fairly important though.  I'll have a think before I proceed one way or another.

can I get a +1 on this preparatory work as it's ready to go?

common/app/common/configuration.scala
```diff
@@ -1,33 +1,76 @@
...
+        None
+    }
+
   }
 
+}
+
+class GuardianConfiguration extends Logging {
+  import GuardianConfiguration._
+  implicit private lazy val app = Play.current
```
TBonnin: I am in the process to remove all reference to the global object `Play.current` which is deprecated in Play 2.5.
Can you inject it as a class parameter? Thx

johnduffell: sure, good spot, I'll do that as a second PR

TBonnin:
go for it 👍

cb372:
The problem with storing secrets on files in S3 is that you have to download and re-upload the files to update them.

* The secrets that were supposed to be only stored in AWS are suddenly on your dev machine as well, which defeats the point. It's easy to forget to delete a file after updating it.

* When you are downloading, editing and re-uploading config files for multiple environments, it's quite easy to get confused and upload the wrong file. I think this kind of mistake is less likely (but still possible) with the Dynamo console.

So +1 for Dynamo. I can see the reason for wanting versioning of your config, but in practice I've never found that much need for it.

rich-nguyen:
@cb372 Is the ease of editing a high priority? The properties file does not go through much churn at all. And if you use the S3 api, which is probably best for a key value tool, you don't need to download to a file, or upload a file.

philwills:
I prefer Dynamo too, but I'm sufficiently cheered by the general thrust of this that I'm happy just to 👍 💯

PR #13751, opened 2016-07-27
Track clicks using GA as well as Omniture

cb372:
## What does this change?

Adds support for tracking the following events using GA:
- in-page link clicks
- internal link clicks
- external link clicks
- 'interactions' (e.g. printing the page)

As far as I know I haven't broken the existing Omniture code (🙏) so all these events are also tracked with Omniture as before.

Also refactored the existing code (which was heavily coupled to Omniture), introducing a click tracking layer to abstract over Omniture and GA.

Sending the GA events to the test tracker for now.

The 3 types of click events are handled as follows.
### In-page links

Easy: just fire off an Omniture/GA event
### Internal links

Write the necessary data to session storage, send a click event on next page load.

This is done in a slightly clunky way: first GA loads, finds the data in session storage, sends an event. Then Omniture loads, also finds the data in storage, sends an event, then deletes the data from storage. When we get rid of Omniture, we'll need to do the deletion elsewhere.
### External links

Fire off Omniture and GA events, and rely on Omniture's magic to delay navigation for 500ms to give the events time to complete.

Not sure exactly how Omniture does this, but when we remove Omniture we'll need to re-implement this behaviour.

One more step along the road to switching off Omniture

No
## Screenshots

n/a
## Request for comment

@janua @sndrs please mock my awful JS


crifmulholland:
Does this in any way relate to ophan tracking? Is there a reason that's not in same structure?

static/src/javascripts/projects/common/modules/analytics/interaction-tracking.js
```diff
@@ -0,0 +1,82 @@
...
+
+    function trackExternalLinkClick(spec) {
+        // Execute the GA and Omniture tracking in parallel
+        // and rely on Omniture to provide a 500 ms delay so they both get a chance to complete.
+        // TODO when Omniture goes away, implement the delay ourselves.
+        google.trackExternalLinkClick(spec.tag);
+        omniture.trackExternalLinkClick(spec.target, spec.tag, {customEventProperties: spec.customEventProperties});
+    }
+
+    addHandlers();
```
janua: Why aren't we using an `init` function in this module instead of depending on implicit running of this method?

cb372: I thought if I added one it would need to be called in a zillion different places, but it looks like that's not the case. `omniture.go()` is only called from `enhanced/common.js`.

I'll fix this.

cb372:
@crifmulholland Ophan's click-tracking is implemented independently [here](https://github.com/guardian/ophan/blob/master/tracker-js/assets/coffee/ophan/click-path-capture.coffee). It would be pretty tricky to integrate it.

common/app/views/fragments/analytics/google.scala.html
```diff
@@ -52,6 +52,30 @@
...
+            *@
+
+            try {
+                var NG_STORAGE_KEY = 'gu.analytics.referrerVars';
+                var referrerVarsData = window.sessionStorage.getItem(NG_STORAGE_KEY);
+                var referrerVars = JSON.parse(referrerVarsData);
+                if (referrerVars && referrerVars.value) {
+                    var d = new Date().getTime();
+                    if (d - referrerVars.value.time < 60 * 1000) { // One minute
+                        ga('guardianTestPropertyTracker.send', 'event', 'Click', 'Internal', referrerVars.value.tag, {
```
janua: Is it possible to use `interactionTracking` from here?

cb372: We can't reference `interactionTracking` from here, because this is happening in inline JS, outside of module-land. But there's no reason I need to fire the event this early. I could potentially do it in the init of `interactionTracking`.

The problem is that I need it to run before the data gets deleted from session storage [here](https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/common/modules/analytics/analytics.js#L202), but I can't work out when that code actually runs. `analytics.js` seems to be a bit of a weird module.

janua:
Looks good to me :+1:

janua:
(You might want to see if @sndrs has any reservations)

cb372:
He's now had 20 hours to

PR #13753, opened 2016-07-28
Hosted Page fixes

lps88:
## What does this change?
- Gallery: only use swipe mode on mobile/tablet (not on desktop touchscreens)
- Disable next video autoplay AB test for Zootropolis campaign (only 1 video)
- extend AB test switch

@Calanthe


Calanthe:
looks good 👍  just remember to extend the AB test too

PR #13754, opened 2016-07-28
Ab disussion ordering

NathanielBennett:
## What does this change?

Give different test cohorts comments ordered differently
## What is the value of this and can you measure success

We want to determine what effect, if any, the ordering of comments has on their consumption. Here. we're going to compare whether different orderings of comment ('Oldest' - ie, likely to have been moderated and therefore potentially of higher quality, 'Recommended', higher quality because of positive reactions' and so forth) has an an effect on the number or users who click 'View more comments'

N/A
## Screenshots

Does the value of the dropdown at the top have an effect on the number of users who click the button.

view-more-commens
## Request for comment

@nicl @crifmulholland @piuccio @desbo 

@gtrufitt I might have borrowed some of your code. Like T.S. Eliot and Bob Dylan, i'm basically a plagiarist, but my sources are impeccable ..


piuccio:
Nice,

how about `getMoreReplies`, is it a success if the user shows more replies of the two comments that are showed by default to her?

Slightly tangent, is there a way (maybe server access logs) to check what's the most popular sorting oder among the people who change it?

NathanielBennett:
@piuccio hows that - or are you saying rename the 'complete' callback? I think it may need to be called complete for abacus. 

Tangential. Possibly. Although, i don't see how you'd be able to measure people who changed to 'oldest' or 'newest' ( the default value depending on whether or not the discussion is closed or not ) from something else. Will have a dig

piuccio:
sorry I didn't mean to change the event name, but to emit the event also when you call the function `getMoreReplies`.

`removeTruncation` is called when you click on the blue `+view more comments`
`getMoreReplies` when you call the white `+show 2 more replies`

Do we care about the latter? @crifmulholland 

screen shot 2016-07-28 at 11 35 14

NathanielBennett:
@piuccio Gotcha. @crifmulholland I'd suggest that was your call.  Seems to be that asking to see more comments and asking to see replies to a top comment are kind of different.

crifmulholland:
Yeah, good shout with the 'show more replies'; I didn't think about that. Success is when we're sure they're interested in reading the comments so both clicks are valid and should be added to the success metric.

As supporting metrics, I'll look at number of replies too. 

@piuccio - getting the most selected order by user pref is on my todo list!

NathanielBennett:
@piuccio Turns out the code already functions as requested: https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/common/modules/discussion/loader.js#L101. I've now changed it so it doesn;t fire on get next page or view contributor. Make sense?

piuccio:
👍

PR #13755, opened 2016-07-28
Show adverts in AMP live blog

SiAdcock:
## What does this change?

This change displays adverts in AMP live blog. The rules:
- ads appear every 5 blog posts
- no more than 8 ads to a page

In order to build the data for the `<amp-ad>` element, I have extracted the logic previously in `AmpAdCleaner` into a [helper object](https://github.com/guardian/frontend/blob/sa-amp-live-blog-ads/common/app/views/support/AmpAd.scala).
## Screenshots

picture 49
## Request for comment

@NataliaLKB @johnduffell @TBonnin

### DIFF
--- a/article/app/views/liveblog/liveBlogBodyContent.scala.html
+++ b/article/app/views/liveblog/liveBlogBodyContent.scala.html
@@ -23,10 +23,11 @@
         }" data-test-id="live-blog-blocks"
         itemprop="@if(article.tags.isReview) {reviewBody} else {articleBody}">
             @if(amp) {
-                @model.currentPage.currentPage.blocks.sliding(5, 5).map { blockGroup =>
+                @model.currentPage.currentPage.blocks.sliding(5, 5).zipWithIndex.map { case (blockGroup, index) =>
                     @views.html.liveblog.liveBlogBlocks(blockGroup, article, timezone, amp)
 
-                    @if(blockGroup.length == 5) {
+                    @* Add advert every 5 blog posts, up to a maximum of 8 *@
+                    @if(blockGroup.length == 5 && index < 8) {
                         <div class="block amp-ad-container amp-ad-container--live-blog">
                             <amp-ad width="300" height="250" type="doubleclick"
                             json=@AmpAd.buildJsonString(article, request.path, Edition(request).id.toLowerCase())
--- a/article/app/views/liveblog/liveBlogBodyContent.scala.html
+++ b/article/app/views/liveblog/liveBlogBodyContent.scala.html
@@ -14,7 +14,7 @@
         @if(blockGroup.length == 5 && index < 8) {
             <div class="block amp-ad-container amp-ad-container--live-blog">
                 <amp-ad width="300" height="250" type="doubleclick"
-                json=@AmpAd(article, request.path, Edition(request).id.toLowerCase()).toJson().toString()
+                json=@AmpAd(article, request.path, Edition(request).id.toLowerCase()).toString()
                 data-slot=@AmpAdDataSlot(article).toString()>
                 </amp-ad>
             </div>
--- a/article/app/views/liveblog/liveBlogBodyContent.scala.html
+++ b/article/app/views/liveblog/liveBlogBodyContent.scala.html
@@ -1,7 +1,26 @@
+@import model.Article
+
 @(model: LiveBlogPage, amp: Boolean = false)(implicit request: RequestHeader)
 
 @import common.Edition
 @import views.support.AmpAd
+@import org.joda.time.DateTimeZone
+
+@liveBlogBlocksAMP(article: Article, timezone: DateTimeZone) = {
+    @model.currentPage.currentPage.blocks.sliding(5, 5).zipWithIndex.map { case (blockGroup, index) =>
+        @views.html.liveblog.liveBlogBlocks(blockGroup, article, timezone, amp)
+
+        @* Add advert every 5 blog posts, up to a maximum of 8 *@
+        @if(blockGroup.length == 5 && index < 8) {
+            <div class="block amp-ad-container amp-ad-container--live-blog">
+                <amp-ad width="300" height="250" type="doubleclick"
+                json=@AmpAd.adTargetingJson(article, request.path, Edition(request).id.toLowerCase()).toString()
+                data-slot=@AmpAd.buildDataSlot(article)>
+                </amp-ad>
+            </div>
+        }
+    }
+}
 
 @defining((model.article, Edition(request).timezone)) { case (article, timezone) =>
     <div class="js-article__container" data-component="body">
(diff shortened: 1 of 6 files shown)


common/app/views/support/AmpAd.scala
```diff
@@ -0,0 +1,43 @@
...
+        "section" -> article.metadata.sectionId
+      )
+    ).toString()
+  }
+
+  def buildDataSlot(article: Article): String = {
+    val section = article.metadata.sectionId
+    val contentType = article.metadata.contentType.toLowerCase
+
+    s"/59666047/theguardian.com/$section/$contentType/amp"
```
TBonnin: where does `59666047` coming from?

common/app/views/support/AmpAd.scala
```diff
@@ -0,0 +1,43 @@
...
+import play.api.libs.json.Json
+
+object AmpAd {
+  private def grabLastFragmentOfId(items: Seq[Tag]) = {
+    items.map { item =>
+      if (item.id == "uk/uk") {
+        item.id
+      } else {
+        val keyword = item.id.split("/").last
+        keyword.replaceAll("""/[+s]+/g""", "-").toLowerCase()
```
TBonnin: I am not sure to understand what you are trying to do here?

common/app/views/support/AmpAd.scala
```diff
@@ -0,0 +1,43 @@
...
+      }
+    }
+  }
+
+  def buildJsonString(article: Article, uri: String, edition: String): String = {
```
TBonnin: would it make sense for this method to return `json` and to stringify it where you need it to be a string

article/app/views/liveblog/liveBlogBodyContent.scala.html
```diff
@@ -21,7 +22,23 @@
             _root_.liveblog.LatestBlock(article.blocks)
         }" data-test-id="live-blog-blocks"
         itemprop="@if(article.tags.isReview) {reviewBody} else {articleBody}">
-        @views.html.liveblog.liveBlogBlocks(model.currentPage.currentPage.blocks, article, timezone, amp)
+            @if(amp) {
```
johnduffell: would it be clearer to separate the amp case out into its own function, so you can see the if and the else more easily?  doesn't necessarily have to be its own file

article/app/views/liveblog/liveBlogBodyContent.scala.html
```diff
@@ -21,7 +22,23 @@
...
+                @model.currentPage.currentPage.blocks.sliding(5, 5).zipWithIndex.map { case (blockGroup, index) =>
+                    @views.html.liveblog.liveBlogBlocks(blockGroup, article, timezone, amp)
+
+                    @* Add advert every 5 blog posts, up to a maximum of 8 *@
+                    @if(blockGroup.length == 5 && index < 8) {
```
johnduffell: max per page is only 30 or so for sport blogs I think, so wouldn't get to 8

common/app/views/support/AmpAd.scala
johnduffell: could we name it more along the lines of why it's doing it rather than what?  something like `adTargetingJson` if that's what it is?

common/app/views/support/AmpAd.scala
```diff
@@ -0,0 +1,43 @@
+package views.support
+
+import model.Tag
+import model.Article
+import play.api.libs.json.Json
+
+object AmpAd {
+  private def grabLastFragmentOfId(items: Seq[Tag]) = {
```
johnduffell: would be `grabLastFragmentsOfIds` but as below I'm guessing it is more like `turnTagsIntoAdFlags` or something?

johnduffell:
looks great, I mostly made some comments about naming which I realised mostly apply to code you moved, so maybe you're no wiser than me about what they should be called?

common/app/views/support/AmpAd.scala
SiAdcock: Copied from [AmpAdCleaner](https://github.com/guardian/frontend/blob/master/common/app/views/support/cleaner/AmpAdCleaner.scala#L104)

Originally, I'm not sure. Perhaps @johnduffell knows?

common/app/views/support/AmpAd.scala
SiAdcock: Again, extracted from [AmpAdCleaner](https://github.com/guardian/frontend/blob/master/common/app/views/support/cleaner/AmpAdCleaner.scala#L90)

common/app/views/support/AmpAd.scala
SiAdcock: Good shout ^ and ^^

article/app/views/liveblog/liveBlogBodyContent.scala.html
SiAdcock: Perhaps we don't need a maximum then? I was using the rules defined in [liveblog-adverts.js](https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/common/modules/commercial/liveblog-adverts.js#L24)

common/app/views/support/AmpAd.scala
SiAdcock: Has a nice ring to it 😄

common/app/views/support/AmpAd.scala
TBonnin: Do we need to duplicate it if it already exists?

common/app/views/support/AmpAd.scala
```diff
@@ -0,0 +1,43 @@
+package views.support
+
+import model.Tag
+import model.Article
+import play.api.libs.json.Json
+
+object AmpAd {
```
TBonnin: Can we make this `AmpAd` classes instead of a singleton object?
ex: 
```
case class AmpAd(article: Article, uri: String, edition: String) {
    def toJson(): Json = ...
}
case class AmpAdDataSlot(article: Article) {
    def toString(): String = ...
```

common/app/views/support/AmpAd.scala
johnduffell: not sure maybe our account id in dfp? 💭

common/app/views/support/AmpAd.scala
SiAdcock: Sounds good. Would it also be appropriate to write tests for these classes, do you think?

TBonnin: It would 😍

SiAdcock:
I have made the suggested changes, and also added tests for `AmpAd.scala`. Note these are my first ever Scala tests, so any pointers will be appreciated!

common/app/views/support/AmpAd.scala
```diff
@@ -4,7 +4,7 @@ import model.Tag
 import model.Article
 import play.api.libs.json.{JsObject, Json}
 
-object AmpAd {
+case class AmpAd(article: Article, uri: String, edition: String) {
```
TBonnin: I guess you can also add the `toString` in this class, so you don't need to explecitely do `toJson().toString` every time you need a string

article/app/views/liveblog/liveBlogBodyContent.scala.html
```diff
@@ -14,8 +14,8 @@
         @if(blockGroup.length == 5 && index < 8) {
             <div class="block amp-ad-container amp-ad-container--live-blog">
                 <amp-ad width="300" height="250" type="doubleclick"
-                json=@AmpAd.adTargetingJson(article, request.path, Edition(request).id.toLowerCase()).toString()
-                data-slot=@AmpAd.buildDataSlot(article)>
+                json=@AmpAd(article, request.path, Edition(request).id.toLowerCase()).toJson().toString()
```
TBonnin: I guess you can also add the `toString` method in the AmpAd class, so you don't need to explecitely do `toJson().toString` every time you need a string

TBonnin:
👍  Great job @SiAdcock

johnduffell:
fab, yeah looks great, go for it! 👍 🚤

SiAdcock:
Thanks @johnduffell and @TBonnin, I'll merge when the build succeeds 😄

PR #13756, opened 2016-07-28
Removing call to WS.url (continue #5)

TBonnin:
## What does this change?

Removing call to `WS.url` in Discussion app. 🔪 

🏃  => Play 2.5

No
## Request for comment

@alexduf

### DIFF
--- a/discussion/app/discussion/DiscussionApi.scala
+++ b/discussion/app/discussion/DiscussionApi.scala
@@ -202,7 +202,7 @@ trait DiscussionApiLike extends Http with ExecutionContexts with Logging {
   }
 }
 
-case class DiscussionApi(val wsClient: WSClient) extends DiscussionApiLike {
+class DiscussionApi(val wsClient: WSClient) extends DiscussionApiLike {
   override protected val apiRoot =
     if (Configuration.environment.isProd)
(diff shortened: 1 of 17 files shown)


discussion/app/discussion/DiscussionApi.scala
```diff
@@ -183,8 +185,34 @@ trait DiscussionApi extends Http with ExecutionContexts with Logging {
   }
 
   private def guClientHeader = ("GU-Client", clientHeaderValue)
+
+  def abuseReportToMap(abuseReport: DiscussionAbuseReport): Map[String, Seq[String]] = {
```
alexduf: :+1: great to move the logic in the service. Makes more sense!

discussion/test/package.scala
```diff
@@ -84,7 +84,4 @@ class DiscussionTestSuite extends Suites (
   new ProfileTest
   ) with SingleServerSuite {
   override lazy val port: Int = new HealthCheck().testPort
-
-  // Inject stub api.
-  controllers.delegate.api = new DiscussionApiStub()
```
alexduf: And I'm happy to see that disappear :+1:

alexduf:
Looks good :star: :+1: :star:

PR #13757, opened 2016-07-28
Commercial ad reporting

rich-nguyen:
Adds a beacon endpoint that receives performance logs from the commercial js app. This data will eventually be passed to data-lake, through a redis server. For now, the data gets as far as an S3 bucket.


common/app/mvt/MultiVariateTesting.scala
```diff
@@ -49,6 +49,18 @@ object ABHeadlinesTestControl extends TestDefinition(
   }
 }
 
+object CommercialClientLoggingVariant extends TestDefinition(
+  "commercial-client-logging",
+  "",
```
TBonnin: No description?

rich-nguyen: Oh yes, I will add one

diagnostics/app/model/diagnostics/commercial/CommercialReport.scala
```diff
@@ -0,0 +1,48 @@
...
+
+  implicit val moduleFormat = Json.format[Module]
+  implicit val advertFormat = Json.format[Advert]
+  implicit val baselineFormat = Json.format[Baseline]
+  implicit val reportFormat = Json.format[Report]
+
+  def report(requestBody: JsValue): Unit = {
+    requestBody.validate[Report] match {
+      case JsSuccess(report, _) => println("ok")
+      case JsError(e) => println("not ok")
```
TBonnin: println?

TBonnin:
What the expected extra traffic to Diagnostics?

diagnostics/app/model/diagnostics/commercial/CommercialReport.scala
```diff
@@ -0,0 +1,49 @@
...
+
+  implicit val moduleFormat = Json.format[Module]
+  implicit val advertFormat = Json.format[Advert]
+  implicit val baselineFormat = Json.format[Baseline]
+  implicit val reportFormat = Json.format[Report]
+
+  def report(requestBody: JsValue): Unit = {
+    requestBody.validate[Report] match {
+      case JsSuccess(report, _) => RedisReport.report(report)
+      case _ =>
```
janua: Get rid of the underscore and use `JsError` instead to avoid having a wildcard.. Also maybe we want logging here? If it falls into this case you are going to have to end up adding logging anyway!

rich-nguyen:
it should be 0.01% of traffic, so I'm testing the water. Will tweak the numbers as appropriate, we don't need a huge proportion. Kind of like sentry really. but cheaper

diagnostics/app/model/diagnostics/commercial/RedisReport.scala
```diff
@@ -0,0 +1,79 @@
...
+
+  def sub(channels: List[String]) = {
+    subscriber ! Subscribe(channels.toArray)
+  }
+
+  def callback(pubsub: PubSubMessage) = pubsub match {
+    case M(`expiredKeyEventChannel`, id) if !id.endsWith("-data") => {
+      uploadReportToS3(id)
+    }
+    case _ =>
```
janua: What are the alternatives here? Is it possible to get rid of the `_`?

diagnostics/app/model/diagnostics/commercial/RedisReport.scala
```diff
@@ -0,0 +1,79 @@
...
+
+/*
+ExpiredKeyEventSubscriber listens to Redis notifications. When a key expires, it gathers
+all the known data for that surrogate key from Redis, and writes a single loggable report object into S3.
+*/
+class ExpiredKeyEventSubscriber(client: RedisClient) {
+
+  val expiredKeyEventChannel = "__keyevent@0__:expired"
+
+  val subscriber = Akka.system().actorOf(Props(new Subscriber(client)))
```
TBonnin: Can we inject the actorSystem? `Akka.system` is deprecated in Play 2.5

diagnostics/app/model/diagnostics/commercial/RedisReport.scala
```diff
@@ -0,0 +1,79 @@
...
+  // Make a client for each usage, otherwise there may be protocol errors.
+  def redisClient: Option[RedisClient] = {
+    try {
+      Configuration.redis.endpoint.map(new RedisClient(_, 6379))
+    }
+    catch {
+      case e: Exception =>
+        log.logger.error(e.getMessage)
+        None
+    }
```
janua: Why not use the lovely `scala.util.Try`?

TBonnin:
What are you trying to achieve @rich-nguyen we already have a Elasticsearch cluster for logging. Why not scale it up enough to support client side logging?

janua:
Nice and clean :+1:

diagnostics/app/model/diagnostics/commercial/RedisReport.scala
rich-nguyen: The other alternatives are `U` `M` and `E`. I could add them

diagnostics/app/model/diagnostics/commercial/RedisReport.scala
rich-nguyen: Just to log the error, really.

rich-nguyen:
@TBonnin it's not just logging and system health for engineers, its processing our reports in a way that means we can track the performance trend over a long period of time, and give a level of business insight. This architecture allows us to send aggregated logs to data lake, in future. Historically, ad perf went to ophan, and then data lake. The cost was high enough that the ophan team decided to bypass elasticsearch, and go direct to data lake. The information is currently left in their avro logs.

Rather than add the data to elasticsearch once again, and use kibana to track commercial performance, we want to cut data in commercial-specific ways. For example, I cannot find out the ad slot performance of safari users. I cannot find out the average number of adverts successfully loaded per page, and compare that number between different loyalty segments. And most importantly, I currently do not have the ability to assess the impact of ab test variants on commercial performance.

In short, if performance really affects business, then this data flow will prove it. #perfmatters

aside:
we could have an elasticsearch instance, in future. But it will blend the ophan metrics we need with the commercial data being tracked here. So, instead of browser->elasticsearch->data-lake, we would expect a system like browser->redis->data-lake->elasticsearch. The latter is possible using Hive.

TBonnin:
@ok got it.

janua:
:+1:

PR #13760, opened 2016-07-28
Removing call to WS.url (Continue #6)

TBonnin:
## What does this change?

Removing call to `WS.url` in Facia

=> Play 2.5

No
## Request for comment

@alexduf


alexduf:
Looks good :+1:

PR #13761, opened 2016-07-28
Hosted Video - fix control bar alignments

lps88:
## What does this change?
- The timestamp is now lined up with the title. Bug probably introduced by CSS refactor (new order of css imports)
- Share by email button doesn't need encoded subject line
### Before

image
### After

image

@Calanthe


Calanthe:
👍

PR #13763, opened 2016-07-28
Removing call to WS.url (Continue #8)

TBonnin:
## What does this change?

Removing call to WS.url in HealthCheck

=> Play 2.5

No
## Request for comment

@alexduf


alexduf:
It looks good :+1:

PR #13764, opened 2016-07-28
Zootropolis article fixes

Calanthe:
## What does this change?

A few fixes:
- proper social sharing text
- 'download link' tracking with data-attributes
- caption under the image is removed
- info about the asset sizes added next to the download buttons
## Screenshots

screen shot 2016-07-28 at 15 47 57
## Request for comment

@lps88


lps88:
👍

PR #13765, opened 2016-07-28
Update the Foresee survey definition to support GA

cb372:
## What does this change?

Just copy-pasting some code we received from Foresee. 

The code extracts the Google Analytics client ID from one of the GA trackers. We have 3 trackers but I checked that they all return the same value, so it doesn't matter which one we use.

Needed to track Foresee survey completions using GA

No
## Screenshots

n/a
## Request for comment

anyone


janua:
:+1:

PR #13767, opened 2016-07-28
Zootropolis hosted article page - CTA image optimized

Calanthe:
## What does this change?

I updated a new much smaller CTA banner image in AWS (kudos to @paperboyo) but had to change the file title as well because couldn't see those changes due to (probably) catching.


kelvin-chappell:
👍

PR #13768, opened 2016-07-28
Add server-side test to test list

rich-nguyen:
forgot to add test


JonNorman:
Assuming you do indeed wish to run the test... :+1:

PR #13769, opened 2016-07-28
Commercial: add a generic cross-frame message handler

regiskuckaertz:
We're going to offload a lot of our code into DFP iframes, which will ultimately reduce the footprint of our commercial code front- and back-end. But in order to do this, we need a way to communicate between our ads and the parent page, so that important interactions can occur (send click events to analytics, get webfonts, etc.)

This is a starting point and there's a few things I need to do before merging:
- [x] write unit tests
- [x] improve incoming message handling (i.e. avoid false positives and security flaws)

cc @guardian/commercial-dev @piuccio


static/src/javascripts/projects/common/modules/commercial/sticky-top-banner.js
```diff
@@ -46,26 +48,17 @@ define([
...
-                if (isRubiconAdEvent && isEventForTopAdBanner) {
-                    fastdom.read(function () {
-                        var padding = parseInt($adBannerInner.css('padding-top'))
-                            + parseInt($adBannerInner.css('padding-bottom'));
-                        var clientHeight = parseInt(data.value.height) + padding;
-                        resolve(clientHeight);
-                    });
-                }
+        messenger.register('set-ad-height', function onSetAdHeight(data) {
+            var $iframe = getAdIframe();
```
jbreckmckye: Which iframe do I grab if there are multiple instances of this advert on a page?

I only remember this hazily, but I'm sure that when I was thinking about writing a replacement to Google's Safeframes implementation, that this was an awkward problem. You can get the window of the origin frame, but not its actual container in the host page.

Instead - and I might be misremembering the matter entirely - what you had to do was have each frame send a ping to its host as it loaded, at which point the host would send each a message containing it's container's identifier. Future messages from the frame would contain this ID, so you could reliably pluck its container from the DOM and manipulate it as needs be.

But my memory is very foggy.

static/src/javascripts/projects/common/modules/commercial/dfp/messenger.js
```diff
@@ -0,0 +1,92 @@
+define([
+    'Promise',
+    'common/utils/report-error'
+], function (Promise, reportError) {
```
jbreckmckye: Is the Grauniad STILL writing ES5?

static/src/javascripts/projects/common/modules/commercial/sticky-top-banner.js
regiskuckaertz: hey @jbreckmckye ! Good to hear from you 😃  Indeed that is something I started thinking about too. I wrote this on the condition that our ads are running on the same domain than the page they're embedded in. From there, inside the frame you have access to `window.frameElement`. From the parent page, you can loop over `window.frames`. But there are other areas I am exploring:

- apparently the `name` attribute on the `iframe` element is passed down in the context of the embedded document, but I need to test this assertion
- DFP assigns `id`s to all those iframes (which is what @OliverJAsh used here)
- this module is intended to be used only by our in-house templates and so we can imagine some kind of shared secret (would also be useful to authenticate the origin of a message)

static/src/javascripts/projects/common/modules/commercial/dfp/messenger.js
regiskuckaertz: We're [getting there](https://github.com/guardian/commercial-templates) 🎆

static/src/javascripts/projects/common/modules/commercial/sticky-top-banner.js
regiskuckaertz: addendum: so indeed, the `name` attribute is accessible as `window.name` and the good news is that DFP uses it. yiihaaa!

jbreckmckye: Those name attributes often get munged by DFP itself, though. A lot of legacy tech used it in the era before `window.postMessage`, and some of these scripts might still be in play here and there.

static/src/javascripts/projects/common/modules/commercial/dfp/messenger.js
```diff
@@ -0,0 +1,149 @@
...
+        register: register,
+        unregister: unregister
+    };
+
+    function register(type, callback, _window) {
+        if( registeredListeners === 0 ) {
+            on(_window || window);
+        }
+
+        listeners[type] || (listeners[type] = []);
```
piuccio: quite clever, I think I prefer the more explicit

```js
if (!listeners[type]) {
   listeners[type] = [];
}
```

static/src/javascripts/projects/common/modules/commercial/dfp/messenger.js
```diff
@@ -0,0 +1,149 @@
...
+            return;
+        }
+
+        try {
+            // Even though the postMessage API allows passing objects as-is, the
+            // serialisation/deserialisation is slower than using JSON
+            // Source: https://bugs.chromium.org/p/chromium/issues/detail?id=536620#c11
+            var data = JSON.parse(event.data);
+        } catch( ex ) {
+            respond(error418, null);
```
piuccio: yes!

static/src/javascripts/projects/common/modules/commercial/dfp/messenger.js
```diff
@@ -0,0 +1,149 @@
...
+
+        var promise = listeners[data.type]
+        // We offer, but don't impose, the possibility that a listener returns
+        // a value that must be sent back to the calling frame. To do this,
+        // we pass the cumulated returned value as a second argument to each
+        // listener. Notice we don't try some clever way to compose the result
+        // value ourselves, this would only make the solution more complex.
+        // That means a listener can ignore the cumulated return value and
+        // return something else entirely—life is unfair.
+        .map(function (listener) {
```
piuccio: how about mergin `.map` and `.reduce` into

```js
.reduce(function (promise, listener) {
    return promise.then(function (ret) {
        var thisRet = listener(data.value, ret);
        return thisRet === undefined ? ret : thisRet;
    };
}, Promise.resolve(true));
```

static/src/javascripts/projects/common/modules/commercial/dfp/messenger.js
```diff
@@ -0,0 +1,149 @@
...
+    function isValidPayload(payload) {
+        return 'type' in payload &&
+            'value' in payload &&
+            payload.type in listeners;
+    }
+
+    // Cheap string formatting function. It accepts as its first argument
+    // a string where each occurence of %% will be replaced by the following
+    // arguments
+    function formatError() {
```
piuccio: Kinda complex for what it does (your errors have only one `%%`).

Tiny correction on your comment, what you say is true if your String.replace is called with a string. Works differently with a regex.

For instance you could write

```js
function formatError () {
    if (arguments.length < 2) {
        return arguments[0] || '';
    }

    var replacements = [].slice.call(arguments, 1);
    return arguments[0].replace(/%%/g, function () {
        return replacements.shift();
    });
}
```
Not that it's much easier to read

static/src/javascripts/projects/common/modules/commercial/dfp/messenger.js
```diff
@@ -0,0 +1,149 @@
...
+        }
+
+        // If there is no routine attached to this event type, we just answer
+        // with an error code
+        if (!listeners[data.type].length) {
+            respond(formatError(error405, data.type), null);
+            return;
+        }
+
+        var promise = listeners[data.type]
```
piuccio: on a second look, this is slightly dangerous because you're calling 

```js
messenger.register('set-ad-height', function onSetAdHeight(data) {
    if (maybe) {
        messenger.unregister('set-ad-height', onSetAdHeight);
    }
});
```
that is modifying the listeners array while iterating on it.


It looks like this code
```js
[1,2,3].map(function (value, position, array) {
   console.log('val', value);
   if (position === 0) { array.splice(position, 1) }
})
val 1
val 3
```
as you can see, the value `2` is skipped.

What saves you is that you're wrapping everything in a `Promise`, very sensible. Promises are resolved in the next tick.
However that depends on the implementation and considering we changed polyfill already in the past, you might end up having random behaviours in the future.

Usually in these cases I clone the array and iterate on that
`listeners[data.type].slice().map`

static/test/javascripts/spec/common/commercial/dfp/messenger.spec.js
```diff
@@ -0,0 +1,134 @@
...
+            var payload = { type: 'this', value: 'hello' };
+            messenger.register('this', routines.thrower, mockWindow);
+            onMessage({ origin: 'http://localhost:9876', data: JSON.stringify(payload), source: mockFrame })
+            .then(function () {
+                expect(mockFrame.postMessage).toHaveBeenCalled();
+                expect(response.error.code).toBe(500);
+                expect(response.error.message).toBe('Internal server error\n\ncatch this if you can!');
+                messenger.unregister('this', routines.thrower, mockWindow);
+            })
+            .then(done);
```
piuccio: you can add a `.catch(done.fail)` so if your expectation fails your test end immediately instead of timing out

piuccio:
I like your tests!

static/src/javascripts/projects/common/modules/commercial/dfp/messenger.js
regiskuckaertz: yes I like that!

static/src/javascripts/projects/common/modules/commercial/dfp/messenger.js
regiskuckaertz: Nice. I'll rework the second part to avoid array mutations.

static/src/javascripts/projects/common/modules/commercial/dfp/messenger.js
regiskuckaertz: Yes! Thanks for spotting, second time I'm caught making the same mistake.

static/src/javascripts/projects/common/modules/commercial/dfp/messenger.js
```diff
@@ -126,23 +126,23 @@ define([
...
-        var i = 1;
-        var ii = arguments.length;
-
-        while (i < ii) {
-            // Keep in mind that when the pattern is a string, String.replace
-            // only replaces the first occurence
-            error.message = error.message.replace('%%', arguments[i]);
-            i += 1;
-        }
+        Array.prototype.slice.call(arguments, 1).forEach(function (arg) {
```
piuccio: even better than my suggestion!

static/src/javascripts/projects/common/modules/commercial/dfp/messenger.js
```diff
@@ -126,23 +126,23 @@ define([
     }
 
     // Cheap string formatting function. It accepts as its first argument
-    // a string where each occurence of %% will be replaced by the following
-    // arguments
+    // an object `{ code, message }`. `message` is a string where successive
+    // occurences of %% will be replaced by the following arguments. e.g.
+    //
+    // formatError({ message: "%%, you are so %%" }, "Regis", "lovely")
```
piuccio: 😆  great comment

static/src/javascripts/projects/common/modules/commercial/dfp/messenger.js
regiskuckaertz: This is an improved version of the expression `a = a || b`, which is kind of idiomatic javascript. You can find it in many OS libs.

piuccio:
👍

static/src/javascripts/projects/common/modules/commercial/dfp/messenger.js
```diff
@@ -0,0 +1,149 @@
...
+                listeners[type].splice(idx, 1);
+            }
+        }
+
+        if (registeredListeners === 0) {
+            off(_window || window);
+        }
+    }
+
+    function on(window) {
```
Calanthe: You can use `bean.on` and `bean.off` here

regiskuckaertz: Yeah, not a fan of `bean`, it does too much 😝

static/src/javascripts/projects/common/modules/commercial/dfp/messenger.js
```diff
@@ -0,0 +1,149 @@
...
+            payload.type in listeners;
+    }
+
+    // Cheap string formatting function. It accepts as its first argument
+    // an object `{ code, message }`. `message` is a string where successive
+    // occurences of %% will be replaced by the following arguments. e.g.
+    //
+    // formatError({ message: "%%, you are so %%" }, "Regis", "lovely")
+    //
+    // returns `{ message: "Regis, you are so lovely" }`. Oh, thank you!
```
Calanthe: :laughing:

PR #13771, opened 2016-07-29
Do not set config on hosted elasticache

rich-nguyen:
The set config command is not supported on aws elasticache


johnduffell:
👍 🐎

PR #13772, opened 2016-07-29
Removing new header dummy tests

NataliaLKB:
Removing the tests because we have had enough traffic now.

What to know why we did these dummy tests? Here is the original PR: https://github.com/guardian/frontend/pull/13676


johnduffell:
👍

PR #13773, opened 2016-07-29
Don't miss sections out of sport liveblogs

johnduffell:
After my previous PRs, now we only fetch as few blocks as possible when displaying the first page of a liveblog, in order to prevent overloading the backend.

Unfortunately sport liveblogs are 30 per archive page rather than a standard 10. That means the first page could be up to 59 blocks before a new page is split off.  Since we were only requesting the first 30, there could be up to 29 missing blocks from the view!

After chatting with @stephanfowler the decision is to change it to request 60 blocks which will be enough for sport.  At least it's not unbounded, so we should still get scalability for big blogs.

@TBonnin


TBonnin:
Other than adding a comment in the code to explain where does `60` come from.
 👍

PR #13774, opened 2016-07-29
Tweak GA custom dimensions

cb372:
## What does this change?

Just a couple of small changes to match what the mobile apps are planning to send:
- Articles can have multiple commissioning desks, so make it a comma-separated list
- Normalise the content type (no spaces, lowercase)

Also set the custom dimensions on both the prod and test trackers, so you get the same results no matter which tracker you are using.

GA data will be consistent across the frontend and the mobile apps.

Yes, apps
## Screenshots

n/a
## Request for comment

@marialivia16 @mohammad-haque

### DIFF
--- a/common/app/model/meta.scala
+++ b/common/app/model/meta.scala
@@ -632,7 +632,7 @@ final case class Tags(
 
   lazy val keywordIds = keywords.map { _.id }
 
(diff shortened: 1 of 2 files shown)


marialivia16:
Looks good to me 👍

PR #13775, opened 2016-07-29
Fix json parse for commercial reports

rich-nguyen:
The structure was expecting a string, but the js app reports a boolean


Calum-Campbell:
:+1:

PR #13777, opened 2016-07-29
Fix hosted thrasher return value

regiskuckaertz:
A creative is responsible for telling the commercial code whether it was rendered successfully or not. Failing to do that may have side effect like having the Outbrain widget appear where it should not:

Admittedly, failing to be displayed is exceptional and we should rather raise an exception for that. We'll change that in another PR.

cc @Calanthe @lps88


rich-nguyen:
👍

Calanthe:
👍

PR #13778, opened 2016-07-29
Move the client logging check to client-side

rich-nguyen:
Another bug. Fastly needs to interpret the cookie and set the mvt header as appropriate, but of course diagnostics does not run requests through Fastly. So the mvt participation check has to be done client-side.


JonNorman:
:+1:

PR #13779, opened 2016-07-29
Update failing commercial feeds link to use HTTPS

JonNorman:
Not going to change the world this one... but it does mean that the link to our Kibana dashboard on our admin page actually works (there isn't a redirect to HTTPS when accessing Kibana via HTTP so it currently just fails).

### DIFF
--- a/admin/app/views/commercial/commercialMenu.scala.html
+++ b/admin/app/views/commercial/commercialMenu.scala.html
@@ -111,7 +111,7 @@
                     <ul>
                         <li><a href="https://github.com/guardian/commercial-tools">Commercial tools repo</a></li>
                         <li><a href="https://dfp-playground.appspot.com/59739193">DFP API playground</a></li>
-                        <li><a href="http://kibana.gu-web.net/goto/1955d282a7b6cc452b673cdb9c27f1af">
+                        <li><a href="https://kibana.gu-web.net/goto/1955d282a7b6cc452b673cdb9c27f1af">
                             Failing Commercial Feeds
                         </a></li>
                     </ul>


regiskuckaertz:
:+1:

admin/app/views/commercial/commercialMenu.scala.html
```diff
@@ -111,7 +111,7 @@
                     <ul>
                         <li><a href="https://github.com/guardian/commercial-tools">Commercial tools repo</a></li>
                         <li><a href="https://dfp-playground.appspot.com/59739193">DFP API playground</a></li>
-                        <li><a href="http://kibana.gu-web.net/goto/1955d282a7b6cc452b673cdb9c27f1af">
+                        <li><a href="https://kibana.gu-web.net/goto/1955d282a7b6cc452b673cdb9c27f1af">
```
TBonnin: There is a property for the kibana url so you don't have to hardcode it. 💃 https://github.com/guardian/frontend/blob/master/common/app/common/configuration.scala#L544

JonNorman: But I'm linking to a specific Kibana dashboard here.
As far as I can tell, the `elk.kibana.url` property in [`frontend.properties`](https://github.com/guardian/platform/blob/master/provisioning/puppet/modules/frontend/templates/etc/gu/frontend.properties.erb#L247) relates to the "home screen" for Kibana.

TBonnin: You are right. we probably needs 2 properties (kibana.defaultUrl, kibana.rootUrl or something like this)

PR #13780, opened 2016-07-29
check whether complete works

NathanielBennett:
## What does this change?

Checks validity of ab success functions

Ab success functions should always run

N/A
## Request for comment

@jamespamplin


jamespamplin:
:+1: brilliant, thanks for fixing this!

jamespamplin:
For some more background on this, @NathanielBennett and I debugged an issue with complete events, to find if a test variant threw an error in its success function, it would prevent all subsequent tests from properly setting up their success functions. The problematic test is [hosted-autoplay](https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/common/modules/experiments/tests/hosted-autoplay.js#L35-L58)

This patch will not fix the problematic test, but will catch the errors that the test was throwing in its success function, allowing all tests to properly initialise their success functions.

PR #13786, opened 2016-08-01
Ignore protocol when deciding if a click is an internal link

cb372:
## What does this change?

When deciding whether a clicked link is an internal (same-host) link, we should be protocol-agnostic. e.g. a link from `https://www.theguardian.com/uk` to `http://www.theguardian.com/politics/foo` should be treated as an internal link, not an external one.

Click events are tracked correctly in GA and Omniture.

No
## Screenshots

n/a
## Request for comment

@mchv


mchv:
This looks good to me 👍

PR #13787, opened 2016-08-01
Remove cross account switch

piuccio:
## What does this change?

Remove an expiring switch. Always read fronts from cmsFronts bucket

Success when fronts work

No
## Request for comment

@janua


NathanielBennett:
:+1:

PR #13789, opened 2016-08-01
Remove ads from hosted articles

lps88:
These weirdly only appear when you use iPhone 5 device mode in chrome:

image


Calanthe:
👍

PR #13790, opened 2016-08-01
Updated names and dates for re-launch.

markjamesbutler:
## What does this change?

Nothing really, it just renames the giraffe ab test to enable tracking. 

As stated in #13476

As stated in #13476
## Screenshots

As stated in #13476


katebee:
👍

PR #13791, opened 2016-08-01
Ab rename discussion ordering

NathanielBennett:
## What does this change?

This renames the test created here: https://github.com/guardian/frontend/pull/13754. This is because the original test didn't have a complete function for the control group in the test, thus corrupting the data. This adds the missing function and renames the test ( so we can create a new test in abacus )

See: https://github.com/guardian/frontend/pull/13754

N/A
## Screenshots

N/A
## Request for comment

@crifmulholland @piuccio @nicl @DiegoVazquezNanini


jamespamplin:
:+1:

PR #13793, opened 2016-08-01
Add olympics (rio 2016) link to the uk sport nav

jfsoul:
## What does this change?

Adds an "olympics" link to the uk sport nav, which directs to the rio 2016 front.  We were given leeway to remove "US sports" if there wasn't enough room, but that doesn't appear to be a problem even at reduced widths.
## Screenshots

Before:
image

After:
image

image
## Request for comment

@SiAdcock @NataliaLKB


SiAdcock:
Assuming the build failed for no good reason!

Otherwise 👍

PR #13794, opened 2016-08-01
Discussion fix duplicate comment box

NathanielBennett:
## What does this change?

Fixes the duplicate comment box bug: https://trello.com/c/FgNzpcnn/53-double-comment-box-error-frontend

The duplicate comment box doesn't appear

N/A
## Screenshots

Before;

duplicate-comment-box

After: 
comment-box
## Request for comment

@nicl @piuccio @crifmulholland


piuccio:
👍  another way would be to avoid rendering entirely, but I reckon it's hard to change that logic

PR #13796, opened 2016-08-01
Don't send duplicate array indexes for email subscriptions

joelochlann:
Fixes a very long-standing bug in the email preferences page.

After a form POST, the state of all of the subscriptions is determined not by another request to ExactTarget but instead by the posted data. So each time, although the user is only changing the state of one subscription, all the other active subscriptions get posted and these are used to determine which buttons should be "on".

Unfortunately, because each section of the page is played out in a separate loop and we use a numerical index for a form field which becomes a `List`, different sections of the page can "overwrite" each other.

So, for example, "The Guardian Today - UK" at the top of the first section:

picture 237

gets given the same index as "Sleeve notes" which is at the top of the Culture section:

picture 236

so this gets sent to the server:

picture 239

and weirdness ensues. (The second `emailSubscription[1]` seems to be ignored, and "Sleeve notes" appears as unsubscribed as in the screenshot. On a GET reload of the page "Sleeve notes" reappears with its correct subscribed status).

@katebee

### DIFF
--- a/identity/app/views/profile/emailPrefs.scala.html
+++ b/identity/app/views/profile/emailPrefs.scala.html
@@ -11,7 +11,7 @@
                 "email-subscription" -> true,
                 "email-subscription--subscribed" -> subscription.subscribedTo,
                 "email-subscription--first" -> row.isFirst))">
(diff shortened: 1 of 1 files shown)


NathanielBennett:
:+1: Yup makes sense and looks good

PR #13798, opened 2016-08-01
Fix Google Analytics in video embeds

cb372:
## What does this change?

Embeds were throwing a JS error and failing to send the GA hit because they relied on some Ophan-related config code that was not being run.

Note: with this change, the Ophan pageview ID is now generated by frontend and passed to the Ophan JS via config, just like on normal frontend pages.

Track video embeds using GA

Affects anywhere we use video embeds, e.g. Facebook Instant Articles
## Screenshots

n/a
## Request for comment

@akash1810 @obrienm

### DIFF
--- a/applications/app/views/videoEmbed.scala.html
+++ b/applications/app/views/videoEmbed.scala.html
@@ -71,6 +71,18 @@
             config: @Html(templates.js.javaScriptConfig(model.SimpleContentPage(video)).body),
             adBlockers: { onDetect: [] }
         };
+        @* Decide the Ophan PV ID here so we can share it with Google Analytics *@
+        guardian.config.ophan = {
+            // This is duplicated from
+            // https://github.com/guardian/ophan/blob/master/tracker-js/assets/coffee/ophan/transmit.coffee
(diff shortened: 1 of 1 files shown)


applications/app/views/videoEmbed.scala.html
```diff
@@ -71,6 +71,18 @@
             config: @Html(templates.js.javaScriptConfig(model.SimpleContentPage(video)).body),
             adBlockers: { onDetect: [] }
         };
+        @* Decide the Ophan PV ID here so we can share it with Google Analytics *@
+        guardian.config.ophan = {
+            // This is duplicated from
+            // https://github.com/guardian/ophan/blob/master/tracker-js/assets/coffee/ophan/transmit.coffee
```
akash1810: I assume we're happy to reference a private repo from a public one?

akash1810:
👍

applications/app/views/videoEmbed.scala.html
cb372: We're already doing so, as I copy-pasted this from elsewhere in frontend 😄

PR #13800, opened 2016-08-01
Promote recommendation to its own module

piuccio:
## What does this change?
- refactor: move the recommendation logic in its own file and report errors
- feature: show a tooltip when anonymous users click on recommend

Maybe more logins? Actually it's just because a button doing nothing is stupid.

Nope
## Screenshots

screen shot 2016-08-01 at 17 32 23
## Request for comment

@NathanielBennett I've reuse the CSS of `report` comment, moving it to a generic `tooltip` module.

@sndrs not sure if I can use atomic CSS already

The wording might require some work, UX is on it.
## TODO
- [x] Unit tests
- [x] UX

### DIFF
--- a/discussion/app/views/fragments/recommendationTooltip.scala.html
+++ b/discussion/app/views/fragments/recommendationTooltip.scala.html
@@ -1,0 +1,16 @@
+<div class="tooltip-box tooltip-box--neutral tooltip-box--width4 tooltip-box-hidden js-rec-tooltip">
+    <button
+        class="tooltip-box__close js-rec-tooltip-close u-button-reset"
+        data-link-name="Close recommendation tooltip"
+        aria-label="Close tooltip"
+    >
+        <i class="i i-close-icon-dark-small"></i>
+    </button>
+
+    <p class="tooltip-box__p">
+        <a href="@Configuration.id.url/signin?INTCMP=DOTCOM_COMMENTS_SIGNIN">Sign in</a>
+        or
+        <a href="@Configuration.id.url/register?INTCMP=DOTCOM_COMMENTS_REG">create your Guardian account</a>
+        to recommend a comment
+    </p>
+</div>
--- a/discussion/app/views/fragments/recommendationTooltip.scala.html
+++ b/discussion/app/views/fragments/recommendationTooltip.scala.html
@@ -1,4 +1,4 @@
-<div class="tooltip-box tooltip-box--neutral tooltip-box--width4 tooltip-box-hidden js-rec-tooltip">
+<div hidden class="tooltip-box tooltip-box--neutral tooltip-box--width4 js-rec-tooltip">
     <button
         class="tooltip-box__close js-rec-tooltip-close u-button-reset"
         data-link-name="Close recommendation tooltip"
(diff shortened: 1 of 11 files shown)


NathanielBennett:
:+1: Code looks good. But I thought we were going to allow anonymous recommends. Which kind of changes the use case. @crifmulholland what you saying here?

discussion/app/views/fragments/recommendationTooltip.scala.html
```diff
@@ -0,0 +1,16 @@
+<div class="tooltip-box tooltip-box--neutral tooltip-box--width4 tooltip-box-hidden js-rec-tooltip">
```
regiskuckaertz: Generally speaking, every DOM fragment inside a page that is hidden should have either the `hidden` or `aria-hidden="true"` attributes. All these `*-hidden` classes are only for people with sight 😭

static/src/javascripts/projects/common/modules/discussion/upvote.js
```diff
@@ -0,0 +1,76 @@
...
+                unsetClicked(target);
+                reportError(ex, {
+                    feature: 'comments-recommend'
+                });
+            });
+        }
+    }
+
+    function isOpenForRecommendations (element) {
+        return !!element.querySelector('.d-discussion--recommendations-open');
```
regiskuckaertz: You could use `getElementsByClassName` which is much faster, but I don't know if there is a polyfill for IE8

regiskuckaertz:
Nice! Q: why is the upvote link even there when the user is not logged in?

piuccio:
Great question. The upvote link gives a context. I've tried removing the button but then it was hard to understand that `123` referred to the number of upvotes.

The arrow is not clickable but visually it still looks like a button, so it deserves some feedback when you click it.

discussion/app/views/fragments/reportComment.scala.html
```diff
@@ -1,7 +1,7 @@
-<form class="d-report-comment-form js-report-comment-form u-h">
-    <div class="d-discussion__error js-discussion__report-comment-error d-discussion__error--hidden">
+<form hidden class="tooltip-box tooltip-box--neutral tooltip-box--width4 js-report-comment-form">
+    <div hidden class="d-discussion__error js-discussion__report-comment-error">
         <i class="i i-alert"></i>
-        <span class="d-discussion__error-text">Sorry there was an error. Please try again later. If the problem persists, please <a data-link-name="report comment eeeor : mailto : userhelp" href="mailto:userhelp@@theguardian.com">Userhelp</a></span>
+        <span class="d-discussion__error-text">Sorry there was an error. Please try again later. If the problem persists, please contact <a data-link-name="report comment error : mailto : userhelp" href="mailto:userhelp@@theguardian.com">Userhelp</a></span>
```
NathanielBennett: Thanks for correcting my spelling for me. :)

regiskuckaertz:
Fair enough. I would hide the number of votes too 😝

piuccio:
@crifmulholland I've added tracking on `sign in` and  `register` and modified the `data-link-name` of the button when user is not logged in.

@NathanielBennett I've pushed some changes, it's not using `--hidden` classes anymore but the `hidden` attribute.

I've added some unit tests and UX is ok with the wording, let's get this out

NathanielBennett:
@piuccio I don't know enough CS to doubt you. Looks good. 
Lets ship
:+1:

PR #13801, opened 2016-08-01
Hide content footer in AMP live blog

SiAdcock:
## What does this change?

Discussion on AMP is superfluous and is dependant on JavaScript that is not available to AMP pages. This change will remove the content footer from the AMP live blog, and with it the discussion functionality, as well as other components that AMP documents do not support. This will achieve parity with the AMP version of the article body, which also excludes the `contentFooter` fragment.

The user will still see a "View comments" link that links to the canonical non-AMP live blog comments. 
### Before

picture 52
### After

picture 53
## Request for comment

@johnduffell @TBonnin @NataliaLKB

### DIFF
--- a/article/app/views/liveblog/liveBlogBody.scala.html
+++ b/article/app/views/liveblog/liveBlogBody.scala.html
@@ -123,7 +123,9 @@
         </div>
     </article>
 
-    @fragments.contentFooter(article, model.related)
+    @if(!amp) {
+        @fragments.contentFooter(article, model.related)
+    }
 
 </div>
 }


johnduffell:
👍 great!

TBonnin:
Does the contentFooter contain more than just comments?

SiAdcock:
@TBonnin fair comment, the footer contains a number of things that are not supported by AMP at the moment. The real purpose of this change was to achieve parity with the [AMP version of the article](https://github.com/guardian/frontend/blob/master/article/app/views/fragments/articleBody.scala.html#L105). I shall update the description to reflect this.

TBonnin:
got it. thx

PR #13802, opened 2016-08-01
messaging for geo-blocked video

akash1810:
## What does this change?

We can currently geo-block videos, this adds some messaging to the video element if you're in a blocked region.

The videos are blocked at the CDN level. Annoyingly, videojs doesn't make it easy to view the status code, so lets just make a HEAD request to get the response code. If its a 403, then we say the video isn't available. Any other status code is handled how they usually are.

Better UX as the user gets some feedback vs the current loading animation.

No.
## Screenshots

screen shot 2016-08-01 at 18 41 23
## Request for comment

@guardian/guardian-frontend as @gidsg is at a conference and (@jamesgorrie has gone somewhere...)

### DIFF
--- a/static/src/javascripts/bootstraps/enhanced/media/main.js
+++ b/static/src/javascripts/bootstraps/enhanced/media/main.js
@@ -170,6 +170,31 @@ define([
         });
     }
 
+    function isGeoBlocked(el) {
+        var source = el.currentSrc;
+
+        // we currently only block to the uk
+        // these files are placed in a special location
+        if (source.indexOf('/ukonly/') !== -1) {
+            return new Promise(function(resolve) {
+                ajax({
+                    url: source,
+                    crossOrigin: true,
+                    method: 'head'
+                }).then(function() {
+                    resolve(false);
+                }, function (response) {
+                    // videos are blocked at the CDN level
+                    resolve(response.status === 403);
+                });
+            });
+        } else {
+            return new Promise(function (resolve) {
+                resolve(false);
+            });
+        }
+    }
+
     function enhanceVideo(el, autoplay, shouldPreroll) {
         var mediaType = el.tagName.toLowerCase(),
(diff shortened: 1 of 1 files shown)


akash1810:
A nicer view of the change - https://github.com/guardian/frontend/pull/13802/files?w=1

akash1810:
cc @guardian/multimedia

akash1810:
Its easiest to use Charles Proxy to test this - setup a rewrite rule for the `cdn.theguardian.tv` domain to return a `403 Forbidden` status.

screen shot 2016-08-01 at 18 50 37

rich-nguyen:
Does this mean for the majority of video views, which are not geo-blocked, they have to make an unecessary HEAD request? Why doesn't videojs allow you to peek the return code?

akash1810:
> Why doesn't videojs allow you to peek the return code?

Not sure. It doesn't seem to fire any event or set any property that I could see. I didn't want to spend more than half a day debugging videojs too...

> Does this mean for the majority of video views, which are not geo-blocked, they have to make an unecessary HEAD request?

Yes... and when you put it that way, I realise this is bad! Geo-blocked videos are placed on a special path, so it would be best to regex the video source first and do a HEAD request if a match is found. Will update the PR.

johnduffell:
try raining it with the videojs guys, I think I did that before with something and they did help out

akash1810:
@rich-nguyen @johnduffell I've made some updates to only make a HEAD call when needed. I think this is ok as a first pass.

I'm working on a more [generic solution](http://jsbin.com/sividu/2/edit?js,console,output) however I'd like to get this out as soon as possible as its for Olympics.

Would you mind taking another look?

johnduffell:
yeah it's a bit rough with regexing urls but since you're working on a more general solution and need to get this out, 👍 
consider a switch or maybe a comment if you're planning to make it better later

PR #13804, opened 2016-08-02
Fix inline ads not waiting for merch

gtrufitt:
## What does this change?

Fixes doubling up of merch and normal ads:

image

I wasn't returning the promise so the rest of the inlineAds weren't waiting for the merch ads to finish up.

Fixes bad UX.
## Request for comment

@guardian/commercial-dev


rich-nguyen:
👍

PR #13805, opened 2016-08-02
Messenger.js: remove all responses to frames we're not sure of

regiskuckaertz:
The `messenger.js` module was too loquacious and eager to reply to any message coming its way, which is believed to slow down pages too much.

This fix removes all responses geared from messages we're not sure if they are bona fide messenger requests.

cc @guardian/commercial-dev


Calanthe:
👍

rich-nguyen:
👍

PR #13806, opened 2016-08-02
New nav veggie burger menu icon animation

zeftilldeath:
A burger to cross animation that fits with the reveal of the menu. 

spin

Spin animation is from here: https://jonsuh.com/hamburgers/

Hello @NataliaLKB 👋


common/app/views/fragments/newHeader.scala.html
```diff
@@ -17,10 +17,6 @@ <h1 class="u-h">The Guardian</h1>
         <nav class="new-header__nav">
             <a href="#main-menu" class="new-header__nav__menu-button js-change-link" tabindex="0">
                 <div class="new-header__burger-icon js-animate-menu">
```
NataliaLKB: Should this still be a div? Our current nav has `<i>`:
image

NataliaLKB:
One comment, but otherwise lgtm! 👍

paperboyo:
Since we do not have house animation style, when we use externally designed ones, I would prefer them to be less elaborate so that there is less risk of them clashing with each other. E.g. Squeeze looks better than Spin. This is entirely personal POV, to be discarded without giving reasons.

Also, I wrote the above just to provide a fake rationalisation against what I feel when seeing spin:
vertigo

zeftilldeath:
@paperboyo Thanks for sharing your fears. It may well be a bit too much, I'm gonna take a look today on devices. See how resource heavy it is too.

paperboyo:
Thanks! I wouldn’t have a clue, but there is [this technique](https://aerotwist.com/blog/flip-your-animations/) that @OliverJAsh was fond of. There was even a demo with this being applied to exactly the problem in question. [This](https://www.youtube.com/watch?v=thNyy5eYfbc), _I think_.

PR #13807, opened 2016-08-02
use unique magneto key name

johnduffell:
This changes the magneto key to have a unique name.  In java properties, dot is not a special charater.  In typesafe config it defines a hierarchy.  Unfortunately we had two keys where one was a parent of the other in the new world, this means there was a collision and it wasn't sure what to do, leading to a None.get in dev-build.

@NataliaLKB


NataliaLKB:
👍

PR #13808, opened 2016-08-02
Stop the page refresh and jump after each button click in Email Prefs

katebee:
## What does this change?

Runs some JavaScript over the Email Preferences page (https://profile.theguardian.com/email-prefs) to improve the UX:
- Switch email subscription buttons from type `submit` to type `button`
- Switch save button to  type `button` (for sending HTML/Text preference)
- Add an on-click event to each button, that makes a ajax request to the email controller
- Disables button while waiting for response and adds loading text
- Updates the button appearance (subscribe/unsubscribe) on success
- Prevents a full page refresh due to the form submission on any button press
- Still give the user an error message, but place it just below the button clicked (rather than at the top of the form) so it is visible without scrolling

Some kind of progressive enhancement was considered, however this page is already mostly non-functional without JS :fire: (due to the dropdowns) so for now JS is required for changing subscriptions.

No
## Screenshots

Example of the error message:
picture 147
## Request for comment

@joelochlann


static/src/javascripts/projects/common/modules/identity/email-preferences.js
```diff
@@ -0,0 +1,122 @@
...
+        'htmlPreference=' + encodeURIComponent(htmlPreference);
+    }
+
+    function renderErrorMessage(buttonEl) {
+        // appends an error message on the parent div of the button
+        var errorMessage = 'Sorry, an error has occurred, please refresh the page and try again';
+        return fastdom.write(function () {
+            var insertionPoint = $.ancestor(buttonEl, 'email-subscription');
+            // Only append an error message once for each email subscription DIV
+            if (qwery('.form__error', insertionPoint).length < 1) {
```
joelochlann: a better approach might be to remove any error messages before sending a new request? otherwise I think you could end up with stale error messages against other buttons

static/src/javascripts/projects/common/modules/identity/email-preferences.js
```diff
@@ -0,0 +1,122 @@
...
+                $.ancestor(buttonEl, 'email-subscription').classList.remove('email-subscription--subscribed');
+                buttonEl.disabled = false;
+            });
+        } else {
+            renderErrorMessage(buttonEl);
+        }
+    }
+
+    function getHTMLPref() {
+        if(qwery('#htmlPreference_HTML')[0].checked) {
```
joelochlann: I *think* you can just do `$('[name="htmlPreference"]:checked').val()` instead of all this

static/src/javascripts/projects/common/modules/identity/email-preferences.js
```diff
@@ -0,0 +1,122 @@
...
+          return 'HTML';
+      } else if(qwery('#htmlPreference_Text')[0].checked) {
+          return 'Text';
+      } else {
+          return '';
```
joelochlann: since they're radio buttons I'm not sure how this case could ever occur?

static/src/javascripts/projects/common/modules/identity/email-preferences.js
```diff
@@ -0,0 +1,122 @@
...
+            reqwest({
+                url: '/email-prefs',
+                method: 'POST',
+                data: formQueryString,
+                error: function () {
+                    renderErrorMessage(buttonEl);
+                },
+                success: function (response) {
+                    var subscriptionState;
+                    try {
```
joelochlann: rather than a try-catch, a better way to delve into an object that you're not sure has the structure you need is:

`if (response && response.subscriptions && response.subscriptions.length)`

However I'm not sure we actually need `subscriptionState` at all. Maybe we should just flip the button state if we get success, and leave it as is if we get an error? Rather than relying on the server response to determine whether we were subscribing or unsubscribing

static/src/javascripts/projects/common/modules/identity/email-preferences.js
katebee: I didn't think it could, until I got an error when testing the page and tracked it down to the data missing from the radio buttons... 😦 So added this for the weird edge case I seemed to run across

static/src/javascripts/projects/common/modules/identity/email-preferences.js
```diff
@@ -0,0 +1,122 @@
...
+    'fastdom',
+    'common/utils/$'
+], function (
+    bean,
+    qwery,
+    reqwest,
+    fastdom,
+    $
+) {
+    function useReqwest(buttonEl) {
```
joelochlann: Not sure about the name of this function. Might be clearer to have a `bindClickHandlers` function and separate `postFormData` function?

static/src/javascripts/projects/common/modules/identity/email-preferences.js
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    function updateButton(buttonEl, subscriptionState) {
+        // do this after successful POST request to update the appearance and value of the button
+        var buttonVal = buttonEl.value;
+        if (subscriptionState === true) {
+            fastdom.write(function () {
+                buttonEl.value = 'unsubscribe-' + buttonVal;
+                buttonEl.innerHTML = 'Unsubscribe';
+                $.ancestor(buttonEl, 'email-subscription').classList.add('email-subscription--subscribed');
```
joelochlann: `classList` won't work in IE9, and I *guess* we do want to aim for compatibility on this page since it's fundamental functionality. use bonzo's `addClass()`

static/src/javascripts/projects/common/modules/identity/email-preferences.js
```diff
@@ -0,0 +1,141 @@
...
+    reqwest,
+    fastdom,
+    $
+) {
+    function reqwestEmailSubscriptionUpdate(buttonEl) {
+        bean.on(buttonEl, 'click', function () {
+            buttonEl.disabled = true;
+            buttonEl.innerHTML = 'Loading...';
+            var formQueryString = generateFormQueryString(buttonEl);
+            reqwest({
```
joelochlann: out of interest what `Content-Type` header gets sent with this request?

paperboyo:
No easy-peasy way for these buttons to use our house round style? Like:
maybe next century

static/src/javascripts/projects/common/modules/identity/email-preferences.js
katebee: reqwest has contentType set in the defaultHeaders:
        `'contentType': 'application/x-www-form-urlencoded'`

So leaving it blank means the POST request contains:

picture 157

katebee:
@joelochlann I've made various changes, and hopefully improvements!
Thoughts? GTG?

static/src/javascripts/projects/common/modules/identity/email-preferences.js
```diff
@@ -0,0 +1,131 @@
...
+                '</div>'
+            ));
+            bonzo(errorMessage).insertAfter(buttonEl.parentNode);
+        });
+    }
+
+    function clearErrorMessages() {
+        if (qwery('.form__error')) {
+            $.forEachElement('.form__error', function (errorEl) {
+                errorEl.remove();
```
joelochlann: For some crazy reason `.remove()` is not available in IE. I would use `$().remove` (or you can do `errorEl.parentNode.removeChild(errorEl)` but that's a bit ugly).

static/src/javascripts/projects/common/modules/identity/email-preferences.js
```diff
@@ -0,0 +1,131 @@
...
+                data: formQueryString,
+                error: function () {
+                    renderErrorMessage(buttonEl);
+                },
+                success: function (response) {
+                    var subscriptionState;
+                    if (response && response.subscriptions && response.subscriptions.length) {
+                        subscriptionState = response.subscriptions[0].subscribedTo;
+                    }
+                    try {
```
joelochlann: why do we need the try/catch?

joelochlann:
👍  💎

static/src/javascripts/projects/common/modules/identity/email-preferences.js
```diff
@@ -0,0 +1,119 @@
...
+                '<div class="form__error">' +
+                    'Sorry, an error has occurred, please refresh the page and try again' +
+                '</div>'
+            );
+            $(errorMessage).insertAfter(buttonEl.parentNode);
+        });
+    }
+
+    function clearErrorMessages() {
+        if ($('.form__error')) {
```
joelochlann: you can actually just do `$('.form__error').each( ... )` here. if it's an empty set the function will be executed 0 times

static/src/javascripts/projects/common/modules/identity/email-preferences.js
```diff
@@ -0,0 +1,119 @@
...
+    function updateSubscriptionButton(buttonEl, isSubscribed) {
+        var buttonVal = buttonEl.value;
+        if (isSubscribed) {
+            fastdom.write(function () {
+                buttonEl.value = 'unsubscribe-' + buttonVal;
+                buttonEl.innerHTML = 'Unsubscribe';
+                $($.ancestor(buttonEl, 'email-subscription')).addClass('email-subscription--subscribed');
+                buttonEl.disabled = false;
+            });
+        } else if (!isSubscribed) {
```
joelochlann: this is actually redundant - `!isSubscribed` will always have the opposite value to `isSubscribed` so it's the same as `else`. I would ditch the `renderErrorMessage` bit and make this an `else`

joelochlann:
🚢  :shipit:

PR #13812, opened 2016-08-02
Redis message logs

rich-nguyen:
Add some logging to see how far the expired event key thing is getting. The data flow needs expired events to happen so we can pass the full log to data lake.


Calum-Campbell:
👍

diagnostics/app/model/diagnostics/commercial/RedisReport.scala
```diff
@@ -26,10 +26,12 @@ class ExpiredKeyEventSubscriber(client: RedisClient, system: ActorSystem) {
 
   def callback(pubsub: PubSubMessage) = pubsub match {
     case Message(`expiredKeyEventChannel`, id) if !id.endsWith("-data") => {
-      uploadReportToS3(id)
+        log.logger.info(s"expired key event received for $id")
+        uploadReportToS3(id)
     }
+    case Message(_, message) => log.logger.info(s"generic redis message received: $message")
     case Unsubscribed(_, _) =>
```
Calum-Campbell: Is there a reason you don't have any unsubscribed logging information?

rich-nguyen: Only because we don't unsubscribe

PR #13813, opened 2016-08-02
Commercial: fluid size everywhere

regiskuckaertz:
This PR adds the `fluid` keyword to all the slots (except badges and merchandising ones) so that we can deliver native ads in all places.

cc @guardian/commercial-dev


Calanthe:
👍

PR #13814, opened 2016-08-02
Add GA tracking to AMP article pages

cb372:
## What does this change?

Adds Google Analytics to AMP article pages.

See the [AMP docs](https://developers.google.com/analytics/devguides/collection/amp-analytics/) for details.

Another step towards switching off Omniture.

AMP accounts for about 2% of pageviews, so this should bump our [GA confidence](https://frontend.gutools.co.uk/analytics/confidence) by about 2%.

AMP, obvs.
## Screenshots

n/a
## Request for comment

Anyone who knows about AMP. @NataliaLKB ?

### DIFF
--- a/common/app/views/mainAMP.scala.html
+++ b/common/app/views/mainAMP.scala.html
@@ -3,7 +3,7 @@
 @import common.{AnalyticsHost, CanonicalLink, LinkTo}
 @import conf.Configuration
 @import conf.switches.Switches.LiveUpdateAmpSwitch
-@import views.support.OmnitureAnalyticsData
+@import views.support.{OmnitureAnalyticsData, GoogleAnalyticsAccount}
 @import views.support.FBPixel
 
 <!doctype html>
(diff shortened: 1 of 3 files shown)


common/app/views/mainAMP.scala.html
```diff
@@ -44,6 +44,38 @@
         <amp-pixel src="//www.facebook.com/tr?id=@FBPixel.account&ev=PageView&noscript=1"></amp-pixel>
         <amp-analytics config="https://ophan.theguardian.com/amp.json"></amp-analytics>
 
+        <amp-analytics type="googleanalytics" id="google-analytics">
```
NataliaLKB: Could this be its own fragment? Its a bit big to include in this file.

common/app/views/mainAMP.scala.html
```diff
@@ -44,6 +44,38 @@
...
+                    "keywordIds": "@{content.tags.keywords.map(_.id).mkString(",")}",
+                    "toneIds": "@{content.tags.tones.map(_.id).mkString(",")}",
+                    "seriesId": "@{content.tags.series.map(_.id).headOption.getOrElse("")}"
+                  }
+                }
+              }
+            }
+            </script>
+        </amp-analytics>
+        <amp-pixel src="//beacon.guim.co.uk/count/pvg.gif"></amp-pixel>
```
NataliaLKB: What is this for?

common/app/views/mainAMP.scala.html
cb372: OK, I'll split it out into a fragment

common/app/views/mainAMP.scala.html
cb372: It's for the [confidence graph](https://frontend.gutools.co.uk/analytics/confidence). 

The idea is that we download one pixel (`/count/pv.gif`) as soon as the page loads, and another pixel (`/count/pvg.gif`) after the GA request has successfully completed. That way we can measure how accurately GA is counting our pageviews. There is an equivalent pixel (`/count/pva.gif`) for Omniture as well.

At least that's how it works on normal pages - with AMP, there's no way to know when the GA call has successfully completed, so I had to just put the pixel directly after the `amp-analytics` tag.

Actually it looks like the initial `pv.gif` pixel is not being requested, so I'll add that as well.

NataliaLKB: Great! Thanks for the explanation. That is very helpful. Would you be able to add a comment explaining what it is?

cb372:
@NataliaLKB How does this look?

NataliaLKB:
Thanks for the changes @cb372. lgtm 👍

PR #13815, opened 2016-08-02
New header z-index fix

NataliaLKB:
## What does this change?

This fixes two bugs caused by the wrong z-index.

The first is that occasionally an ad would appear over the menu.

The second is that in Safari the overlay would appear over the menu. This meant you couldn't open anything in the menu
image

Nope!
## Request for comment

@gtrufitt @zeftilldeath


SiAdcock:
Look nice, good commenting! 👍

PR #13818, opened 2016-08-02
improve Object.assign polyfill

piuccio:
It's perfectly fine to pass undefined values to `Object.assign`, but our polyfill doesn't handle that very well.

Currently it breaks on IE if you `Object.assign({}, null);`


johnduffell:
looks better if you view with https://github.com/guardian/frontend/pull/13818/files?w=1
👍

PR #13820, opened 2016-08-02
Anonymous recommend comments behind switch

NathanielBennett:
## What does this change?

Allows comments to be recommended anonymously, behind a switch.

Enables us to guage whether anonymous upvoting is a significant form of participation on the site

N/A
## Screenshots

N/A
## Request for comment

@piuccio @nicl @crifmulholland


static/test/javascripts/spec/common/discussion/recommendation.spec.js
```diff
@@ -1,17 +1,28 @@
...
     describe('Recommendations of comments', function () {
         var discussionApi = {
             recommendComment: function () {}
-        };
+        },
+        injector = new Injector(), config, upvote;
+
+        beforeEach( function (done) {
+
+            injector.require(['common/modules/discussion/upvote', 'common/utils/config'], function() {
```
piuccio: I would have use dependency injection to pass `config` to the upvote module, which is a fancy way of saying

```js
function handle (target, container, user, discussionApi) {}
// becomes
function handle (target, container, user, discussionApi, config) {}
// or even
function handle (target, container, user, discussionApi, allowAnonymous) {}
// where
function allowAnonymous () { // defined in loader.js
   return config.switches.discussionAllowAnonymousRecommendsSwitch;
}
```
makes your tests straightforward. I've never liked `injector` but that's a personal thing, I don't understand how it works.

common/app/conf/switches/FeatureSwitches.scala
```diff
@@ -337,6 +337,17 @@ trait FeatureSwitches {
     exposeClientSide = true
   )
 
+  val DiscussionAllowAnonymousRecommendsSwitch = Switch(
+    SwitchGroup.Feature,
+    "discussion-allow-anonymous-recommends-switch",
+    "if this is switched on, comments can be recommended by signed out users",
+    owners = Seq(Owner.withGithub("NathanielBennett")),
+    safeState = Off,
+    sellByDate = never,
```
piuccio: this probably requires a `sellByDate` at least to remind us to re-evaluate this decision

@crifmulholland what do you think?

NathanielBennett: @piuccio . Yeah I was thinking this is a feature, rather than a test. If we don't want it we get rid of it.

piuccio:
the commit history includes an old commit of mine, maybe in this case it's best to rebase

static/test/javascripts/spec/common/discussion/recommendation.spec.js
NathanielBennett: @piuccio You're so right. I am so crap. Thanks.

PR #13821, opened 2016-08-02
Commercial: copy webfonts across frames

regiskuckaertz:
To make our [commercial templates](https://github.com/guardian/commercial-templates/pull/2) work without the breakout algorithm, we need them to get everything they need to be displayed correctly. One major aspect of that is using the grauniad's webfonts.

This is the first API built on top of `window.postMessage` that we will need to create in order to get rid of all the creatives within the commercial javascript.

cc @guardian/commercial-dev


paperboyo:
I genuinely have no idea, but could this mean that interactive iframes could use dotcom's fonts, too?

E.g. a map in https://www.theguardian.com/world/2016/aug/02/chlorine-attack-syria-dozens-ill-saraqeb-idlib

rich-nguyen:
no spec for  `get-stylesheets` module?

Just wondered, if origin policy permits, would a shared worker have been a possible solution too?

👍

regiskuckaertz:
@paperboyo yes, that's very easy, all it needs is to be more lax around the origin of the message

@rich-nguyen I wish I could use a shared worker, but for operations like this one which need access to the DOM, that's impossible 😞  specs: sure thing! I'll add them 😃

paperboyo:
> yes, that's very easy, all it needs is to be more lax around the origin of the message

Amazing, thanks! After upcoming work to webfonts is done, I will unleash my annoyance on the @guardian/interactive-team :laughing:

Jholder112233:
@paperboyo do! Would be good to find a good cross-platform solution so we don't have to double-load our fonts on iframed interactive content. Easy way to save 100s of KBs.

Whilst this would work for interactives dotCom, remember that we also iframe interactives in our apps and off-platform. So there's a lot more to it.

PR #13822, opened 2016-08-02
Add some logging to investigate s3 put problem

rich-nguyen:
For some reason, the commercial logs are being reported, but after one report, the system goes silent. I suspect something is going wrong in the handler.


JonNorman:
:+1:

PR #13823, opened 2016-08-02
Allow ads in the top slow to overflow their container

regiskuckaertz:
Unfortunately, some ads overflow their container:

picture 2

So I had to remove the `overflow` rule, without which the page flickers when some creatives load and expand the top slot:

picture 1

cc @guardian/commercial-dev


piuccio:
can the overflow auto be applied only when the ad is open?

regiskuckaertz:
@piuccio I will try to find a better solution but now this is quite urgent

rich-nguyen:
👍

PR #13824, opened 2016-08-02
Fix for commercial client logging

rich-nguyen:
The previous fix caused a Redis `val client` to be re-used out of scope


JonNorman:
:+1:

Issue #12782, opened 2016-05-04
Implement Facebook quote sharing on selection pop-up


gidsg:
# RFE

## Steps to Reproduce
Go to an article page
Highlight some text

## Expected Results 
Add option to share selection as a facebook quote:
see: https://developers.facebook.com/blog/post/2016/04/12/share-engage-and-measure-in-new-ways/
https://developers.facebook.com/docs/sharing/reference/share-dialog 'quote' parameter

PR #13825, opened 2016-08-02
Highlighter tool improvements

patstirling:
## What does this change?

Provides an updated highlighter tool which, while maintaining sharing functionality, answers a number of reader-based use cases including: 

1) flagging issues with the content (perceived spelling or factual errors) in a controlled way to Userhelp

2) allowing readers to quickly and easily look something up that they have highlighted on the page (initially through a google search but we can add further functionality if it proves valuable)

It also removes the use of short URLs in the highlighter sharing function.

Issues with the content should be more easily flagged with user help, and readers more easily able to look things up based on our content; we have tracking code in place for each use case.

No
## Screenshots

image
## Request for comment

This is my first pull request to the Guardian code-base - I  hope it's reviewed throughly but fairly!


paperboyo:
Woohoo! A hack-day project in production!
- how does the second-level sharing choice look like (it should mimic exactly what’s been recently done by @marialivia16 [for comment sharing](https://github.com/guardian/frontend/pull/13329) - including the animation, I should think)
- do these have explanations in tooltips before clicking (exclamation mark is not immediately obvious)
- they do not seem to be exactly (visually) centred
- these SVGs could be optimised (precision dropped to one decimal place), some may already be available as inline components (?)
- I always dreamt of them being ever-so-slightly transparent, so that the text below reads

In general, the search one seems to me a tad superfluous, offering what browsers already provide (although it’s one click less):

search-with

Was I fair? I hope I was fair! Or at least - thorough :stuck_out_tongue_winking_eye: Plus, I have absolutely no idea what most of these search engines are doing there... honest!

rich-nguyen:
Agree with @paperboyo 's comments.

Also, this new functionality appeals to the small number of feedback-driven users, and de-prioritizes the functionality that appeals to the larger number of unheard users; those who share. We should be careful not to make changes that work against the slient majority (i haven't checked the numbers recently, but select-sharing has been quite successful).

ScottPainterGNM:
Worth adding a facebook sharing option while this is being considered also, Gids added a Feature Request and as we now offer it on comments would it be sensible to do it for selection sharing also?

ScottPainterGNM:
See issue #12782

static/src/stylesheets/_vars.scss
```diff
@@ -136,6 +136,18 @@ $sport-away:             #003f64;
 
 // Stati
```
jbreckmckye: Small point, and not related to Pat's changes, but this isn't actually the correct pluralisation of "status" in Latin or otherwise. If we insist on using Latinate declension (wrongly - we only use foreign-language inflections for loan-words we'd encase in _italics_), then the genitive case of `status` is `statōrum`.

ScottPainterGNM: This might be my favourite pull request comment ever

image

static/src/javascripts/projects/common/modules/ui/selection-sharing.js
```diff
@@ -22,26 +25,50 @@ define([
...
+    resultsTemplate,
+    socialIconsTemplate,
     svgs,
     debounce,
     throttle,
     some
 ) {
 
     var $body = bonzo(document.body),
+        wikiIcon = svgs('searchWiki', ['icon']),
```
JonNorman: Do these need to be stored in variables? Looks like they are only used to populate the `selectionSharing` dict. It is probably clear enough to just put them straight into `selectionSharing`.

patstirling:
Thanks folks, this is great feedback and good questions. 

This was born ultimately out of my own frustration that there is no very quick way to do the thing I use highlighting for on the Guardian: find out more about a thing. I did some guerilla testing on my 1st hack effort on this which confirmed my suspicions that others felt the same - and found a new use case which cropped up numerous times, which is that sometimes people want to flag errors, and there is no obvious way to do so.

I'll have a dig into those sharing numbers Rich.

So if we are talking unheard users, we currently have absolutely no way of knowing whether either of those use cases are ones that lots of our users want (as opposed to the half a dozen or so I've talked to).

So I propose this: we test it! The simplest way would be to try for a week and see what happens. I'm open to other approaches.

On how the sharing functionality works, it works exactly as today for those buttons, but with an extra click on the share icon to unlock the share buttons:

image

I take the point that we should be doing the lovely animations and adding the FB button... and I certainly would like to do that as the next step once we've validated the above use cases.

I'd highlight (ha ha) that nothing here is worse than what we currently have (unless we see a significant drop in sharing, in which case test=fail).

I'll fix the SVGs; where they were available already I'm using them but let me know if I missed something.

On using search through browser highlighter: fair point, although the other advantage is that we gain more control; in theory we may want to do something as a next step to encourage people to stay on the Guardian (results in a search box, keyword-based results where we have them etc.)

paperboyo:
> in theory we may want to do something as a next step

Yeah. I was thinking if that shouldn't maybe be using our site-search. But given the sorry state of what we call our site-search... fixing it would be much higher priority IMHO (although, quite an undertaking).

re:SVGs
https://jakearchibald.github.io/svgomg/ <- great tool. Try to drop precision to 1. If vectors become too malfromed > drop it to 2 only (one can overcome this inability to drop to 1 by making SVG bigger and/or dropping the precision in Illustrator first, which seems... more precise).

static/src/javascripts/projects/common/modules/ui/selection-sharing.js
patstirling: Good spot Jon - thought I'd removed these (part of other experimentation)

patstirling:
I told a lie in the original outline of this - I need to add tracking to the search functionality

johnduffell:
@ScottPainterGNM just on your facebook sharing point - it's not possible to share with arbitrary text on facebook, only the whole article

johnduffell:
interestingly, the feedback I'd previously heard on the pop up share buttons was either "get rid of it completely" from those who want to mark their place while reading, and readers editors who wanted a report button for errors, as otherwise people just use discussion which noone sees.
Just shows you can't please everyone!  Maybe if we want to go crazy with buttons on highlight, we should make them appear on the edge of the screen.

Anyway, this has merge conflicts and a failing build, so it needs some love if it's going out?

ScottPainterGNM:
@johnduffell Gids advised me that they now do support it - https://developers.facebook.com/docs/plugins/quote looks like it got released earlier this year.

paperboyo:
> Just shows you can't please everyone!

You can! Make them part of User Preferences.

johnduffell:
@ScottPainterGNM oh yeah thanks, that's useful because we've hit this problem a few times!  must be a common problem then.

johnduffell:
@paperboyo then you don't please the developers!  we have to test and maintain all this stuff  😢

paperboyo:
True that! And User Help has it harder (maybe preferences states should be available in technical report?). Well, developers are there to suffer, no? This way they can refactor. 😉 Having more stuff available in User Preferences provides genuine and non-discriminatory reasons for signing up. As opposed to all the non-genuine and/or discriminatory reasons...

patstirling:
Hi folks, and thanks John for jumping on this - I'll work through the conflicts and build issues (this my first PR and quite a lot to learn, so thanks for bearing with me).

Data on the usage of this tool over last 90 days is available here:

https://docs.google.com/spreadsheets/d/112wBo6AAwjCDS_inBxKjhhAVgzwnBPVPMRMdxtjTvUw/edit#gid=0

Summary: it's fairly stable and contributing around 4,800 shares or 7.5% of all shares daily. So really is used a bit, which I wouldn't want to impact.

Before I go further, my proposal would be to trial in the AU edition for a couple of weeks and observe how the new options get used. Why AU? It's the smallest edition, and is representative of the whole - about 5% of shares come from AU and about 8% of those are from highlighter. 

As John points out, one of the primary asks from users around this has been to report errors; we have also seen people request being able to easily get translations or definitions, which the simplest proxy for is to offer a search functionality and see how often it gets used.

Success looks like no downturn in sharing and decent use by the end of the test period (I'd argue total 50+ a day of the new functions of search or flag based on usage of share on the page in AU).

Any objections? If not I'll resolve conflicts/build, make this edition-specific and get tracking in place where it's missing.

P.S. I want to be really clear that I'm not seeking to add complexity for the sake of it but am keen to understand in a test whether the use cases/need states we observed are real and whether we can increase people's engagement with the journalism through one of the most obvious interactions they can take; I'd want to include FB as next steps if the functionality is there on FB's side now.

johnduffell:
for me the key word in the whole of that is "test" as at the moment we're all just guessing whether it's a good idea.
I like that you have the existing stats and have thought about what success looks like, but you really need to talk to someone in the data team how to set up a proper test.  partitioning by any specific criteria (e.g. AU only) won't work, you need to randomly sample users and get a big enough sample size to detect the result you are after.  That would probably be relatively big (I guess in the area of 10% for a week or two) for something that's used relatively rarely that we need to detect a small change.
Once you have the results, you can check them against the success metrics you defined at the start and then you'll know whether it's a goer.  Or back to the drawing board!

johnduffell:
lots of old PRs around - close and reopen when ready?

static/src/stylesheets/_vars.scss
gtrufitt: Quod est incredibile

johnduffell:
clearing this off the list for now - please reopen when you are ready for more input!

PR #13826, opened 2016-08-03
Use existing logging bucket for commercial reports

rich-nguyen:
1. Uses the `aws-frontend-logs` to store log reports, which the server has write permissions for.
2. Filters out all the empty `csp report` log lines from diagnostics server log


JonNorman:
Looks good to me :+1:

PR #13828, opened 2016-08-03
AMP live blog pagination

SiAdcock:
## What does this change?

This change styles up the pagination controls on the AMP live blog. Clicking the "older" and "newer" buttons will display older and newer live blog content respectively, within the AMP live blog context.

The number of blog posts that will be displayed on one page follows the same logic of the non-AMP live blog.

The live updating functionality will only be enabled on the first page of the live blog (i.e. when viewing the most recently posted blog posts).
### Before 😢

picture 58
### After 🎉

picture 60
## Request for comment

@NataliaLKB @TBonnin @johnduffell @stephanfowler

### DIFF
--- a/article/app/views/fragments/liveBlogNavigation.scala.html
+++ b/article/app/views/fragments/liveBlogNavigation.scala.html
@@ -19,7 +19,7 @@
                     "arrow-left-double",
                     "icon",
                     List(
-                        "liveblog-navigation__link--secondary--newer__icon",
+                        "liveblog-navigation__link__icon--newer",
                         if(pagination.newest.isEmpty) "liveblog-navigation__link__icon--disabled" else "liveblog-navigation__link__icon--enabled"
                     )
                 )
@@ -71,7 +71,7 @@
                     "arrow-right-double",
                     "icon",
                     List(
-                        "liveblog-navigation__link--secondary--older__icon",
+                        "liveblog-navigation__link__icon--older",
                         if(pagination.oldest.isEmpty) "liveblog-navigation__link__icon--disabled" else "liveblog-navigation__link__icon--enabled"
                     )
                 )
--- a/article/app/views/fragments/liveBlogNavigation.scala.html
+++ b/article/app/views/fragments/liveBlogNavigation.scala.html
@@ -15,7 +15,7 @@
             }
             class="liveblog-navigation__link liveblog-navigation__link--secondary liveblog-navigation__link--secondary--newer @if(pagination.newest.isEmpty) { liveblog-navigation__link--disabled }"
             data-link-name="newest page">
-                @fragments.inlineSvg("arrow-left-double", "icon", List("liveblog-navigation__link--secondary--newer__icon"))
+                @fragments.inlineSvg("arrow-left-double", "icon", List("liveblog-navigation__link--secondary--newer__icon", if(pagination.newest.isEmpty) "liveblog-navigation__link--disabled" else "liveblog-navigation__link--enabled"))
                 Newest
             </a>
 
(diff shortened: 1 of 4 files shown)


jfsoul:
Font looks like Arial, and vertical alignment of the "Oldest"/"Newest" within the buttons is a little on the low side.  May be intentional or WIP, but just caught my eye 😃

NataliaLKB:
@zeftilldeath, what do you think?

common/app/views/fragments/amp/stylesheets/liveBlog.scala.html
```diff
@@ -207,3 +207,108 @@
...
+}
+
+.liveblog-navigation__link {
+    display: inline-block;
+    position: relative;
+    color: #333;
+    border: 0.0625rem solid #dcdcdc;
+    height: 2.25rem;
+    line-height: 2.375rem;
+    font-weight: bold;
```
NataliaLKB: Because it is a different font, I am not sure if bold looks the best with this design.

article/app/views/fragments/liveBlogNavigation.scala.html
```diff
@@ -15,7 +15,7 @@
             }
             class="liveblog-navigation__link liveblog-navigation__link--secondary liveblog-navigation__link--secondary--newer @if(pagination.newest.isEmpty) { liveblog-navigation__link--disabled }"
             data-link-name="newest page">
-                @fragments.inlineSvg("arrow-left-double", "icon", List("liveblog-navigation__link--secondary--newer__icon"))
+                @fragments.inlineSvg("arrow-left-double", "icon", List("liveblog-navigation__link--secondary--newer__icon", if(pagination.newest.isEmpty) "liveblog-navigation__link--disabled" else "liveblog-navigation__link--enabled"))
```
johnduffell: no idea if I'm on the right track here, but I think the `a` is the `liveblog-navigation__link` so is it right from a BEM point of view to use the class on the svg too?  if the way it works in the standard won't work in amp, maybe we need to change both to work in the amp way?

johnduffell:
the controls look great now! thanks!

article/app/views/fragments/liveBlogNavigation.scala.html
SiAdcock: Good shout @johnduffell, those names are not very BEM-y. I'll probably go with `liveblog-navigation__link__icon--disabled` and `liveblog-navigation__link__icon--enabled`

SiAdcock:
I have made a number of changes as a result of some excellent feedback from @zeftilldeath:

- Updated the main content background colour

picture 61

- disabled inline ads that were appearing in some live blog posts
- reinstated missing icons in the sharing block

picture 62

- hidden the labels within the navigation buttons on small screen devices

picture 63

NataliaLKB:
This looks much nicer!

article/app/views/fragments/liveBlogNavigation.scala.html
```diff
@@ -15,7 +15,14 @@
             }
             class="liveblog-navigation__link liveblog-navigation__link--secondary liveblog-navigation__link--secondary--newer @if(pagination.newest.isEmpty) { liveblog-navigation__link--disabled }"
             data-link-name="newest page">
-                @fragments.inlineSvg("arrow-left-double", "icon", List("liveblog-navigation__link--secondary--newer__icon"))
+                @fragments.inlineSvg(
+                    "arrow-left-double",
+                    "icon",
+                    List(
+                        "liveblog-navigation__link--secondary--newer__icon",
```
NataliaLKB: Should this be `liveblog-navigation__link__icon--newer`?

NataliaLKB:
Left one more comment, but after that 👍

article/app/views/fragments/liveBlogNavigation.scala.html
SiAdcock: It was pre-existing code, but you're right, it's messy and I can tidy it up as part of this change 😄

NataliaLKB: If you have time, this would be great!

PR #13831, opened 2016-08-03
Inject environment into Weather api (used to check the running Mode)

TBonnin:
## What does this change?

Inject environment into Weather api (used to check the running Mode)
This allows WeatherApi not to rely on Play.current anymore

=> Play 2.5

No
## Request for comment

@johnduffell @alexduf


johnduffell:
👍

PR #13832, opened 2016-08-03
Hoover state on the hosted CTA banner

Calanthe:
## What does this change?

The CTA banner is a bit darker while hovering. 
## Request for comment

@lps88


lps88:
:+1:

PR #13833, opened 2016-08-03
Inject Mode into S3BreakingNews class

TBonnin:
## What does this change?

Inject Mode into S3BreakingNews class

The goal is to remove dependency towards Play global object before to
upgrade to Play 2.5

No
## Request for comment

@johnduffell @alexduf


johnduffell:
:

PR #13835, opened 2016-08-03
Remove the global CTA for inline membership components

regiskuckaertz:
Otherwise the component won't display because of a reference error. Now:

picture 3

cc @guardian/commercial-dev


Calanthe:
👍

PR #13839, opened 2016-08-03
fetch timeout

piuccio:
## What does this change?

Get rid of `ajax-promise`. The last obstacle was `timeout`. Standard `fetch` implementation doesn't have a timeout feature yet (becuase of cancellable promises and all that) so I'm implementing this as another utility, `common/utils/timeout`.

None

No


static/test/javascripts/spec/common/utils/timeout.spec.js
```diff
@@ -0,0 +1,58 @@
...
+            setTimeout(resolve, time);
+        });
+    }
+
+    describe('Timeout utility', function () {
+        it('resolves with the promised value if the timeout doesn\'t fire', function (done) {
+            timeout(10, Promise.resolve(12))
+            .then(function (result) {
+                expect(result).toBe(12);
+                return wait(10);
```
regiskuckaertz: why this line?

regiskuckaertz:
Nice shot 👍

static/test/javascripts/spec/common/utils/timeout.spec.js
piuccio: internally the `timeout` function is still firing the `setTimeout` after 10ms, so I prefer to wait for that to happen before finishing the test

PR #13840, opened 2016-08-03
Bump facia cache age

jfsoul:
## What does this change?

Increases the `max-age`value in the `Cache-Control` header for all fronts (and rss/json variations).  The value has been bumped from 60 seconds to 5 minutes.

Note that the `CacheTime` objects replace calls to `Cached` with both a `Page` and an `Int`.  I believe this is safe because all the calls with `Page` would default to `CacheTime.Default` previously (see `PressedPage`).  If you have any suggestions for how we could tidy this up further I am open to suggestions.

This PR also introduces a _not found_ cache age which will be overwritten by fastly anyway (and if code comments are to be believed not passed by nginx).  Previously the value was hardcoded at 60 seconds in each call to `Cached` in the `FaciaController`.  It makes no difference other than being a bit tidier.

I also removed the redundant `EditionalisedKey`.

Now that the facia purger lambda is running we can safely turn down the cache age, as fronts will be purged immediately on changes.  This will save us bandwidth especially on fronts and at times that see fewer updates (but consistent page views).

Not that I am aware of.
## Request for comment

@johnduffell @TBonnin

### DIFF
--- a/common/app/model/Cached.scala
+++ b/common/app/model/Cached.scala
@@ -21,6 +21,8 @@ object CacheTime {
   object Default extends CacheTime(60)
   object LiveBlogActive extends CacheTime(5)
   object RecentlyUpdated extends CacheTime(60)
+  object Facia extends CacheTime(300)
+  object NotFound extends CacheTime(10) // This will be overwritten by fastly
 
   def LastDayUpdated = CacheTime(extended(60))
   def NotRecentlyUpdated = CacheTime(extended(300))
@@ -56,6 +58,10 @@ object Cached extends implicits.Dates {
 
   def apply(seconds: Int)(result: CacheableResult)(implicit request: RequestHeader): Result = {
     apply(seconds, result, request.headers.get("If-None-Match"))//FIXME could be comma separated
+  }
+
+  def apply(cacheTime: CacheTime)(result: CacheableResult)(implicit request: RequestHeader): Result = {
+    apply(cacheTime.cacheSeconds, result, request.headers.get("If-None-Match"))
   }
 
   def apply(duration: Duration)(result: CacheableResult)(implicit request: RequestHeader): Result = {
--- a/facia/app/controllers/FaciaController.scala
+++ b/facia/app/controllers/FaciaController.scala
@@ -20,9 +20,6 @@ import scala.concurrent.Future.successful
 
 trait FaciaController extends Controller with Logging with ExecutionContexts with implicits.Collections with implicits.Requests {
 
-  // This will be overwritten by fastly anyway for 404s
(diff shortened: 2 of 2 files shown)


facia/app/controllers/FaciaController.scala
```diff
@@ -97,14 +95,13 @@ trait FaciaController extends Controller with Logging with ExecutionContexts wit
 
   def redirectTo(path: String)(implicit request: RequestHeader): Future[Result] = successful {
     val params = request.rawQueryStringOption.map(q => s"?$q").getOrElse("")
-    Cached(60)(WithoutRevalidationResult(Found(LinkTo(s"/$path$params"))))
+    Cached(CacheTime.Facia)(WithoutRevalidationResult(Found(LinkTo(s"/$path$params"))))
```
johnduffell: although in reality, redirects aren't likely to change often so could be cached much longer - facia (and article) already return an excessive number of redirects

common/app/model/Cached.scala
```diff
@@ -21,6 +21,8 @@ object CacheTime {
   object Default extends CacheTime(60)
   object LiveBlogActive extends CacheTime(5)
   object RecentlyUpdated extends CacheTime(60)
+  object Facia extends CacheTime(300)
```
johnduffell: didn't go for the switch in the end? ;)

johnduffell:
yep certainly a neater way of doing it, good use of named variables 
👍 📉

jfsoul:
Going to keep an eye on the effect of this change but early signs are good:

(Blue is Facia NetworkIn, green Facia CPU and orange Facia NetworkOut, deploy was made where the cursor is hovering)

cloudwatch day one

TBonnin:
Very nice 👏 Great job
I wonder why NetworkOut doesn't fall as much as NetworkIn

jfsoul:
NetworkOut has already been optimised by @johnduffell's work to send 304 if front is unchanged?

johnduffell:
I still think it's worth trying for smaller machines as a next step, what do other people think? saving would be a few hundred a month based on jon's figures yesterday

jfsoul:
Update on CPU after this change.  The initial drop has been maintained since deploy, bringing up the question of whether we should drop instance size.  Second graph below shows free physical memory which varies between ~5-7GB, which suggests we might not want to drop down to <15GB total memory.  A switch to memory optimised instances might suit.

image

image

johnduffell:
fantastic, that's really interesting.  you can look at the heap usage as well if you like, we send some custom metrics under frontend-facia name
I'd definitely give a +1 for a trial of shrinking the instances, the savings would be significant

PR #13841, opened 2016-08-03
Anonymous recommend comments behind switch

NathanielBennett:
## What does this change?

This is basically a re-run of this: https://github.com/guardian/frontend/pull/13820. It provides a switch so that comments can be anonymously upvotes

We want to guage whether anonymous upvoting has an effect on participation

No
## Screenshots

N/A
## Request for comment

@piuccio - Hey when i deleted the original branch (having cherrypicked it) it blatted the original PR
@crifmulholland @nicl


piuccio:
👍

PR #13837, opened 2016-08-03
Revert "Merge pull request #13712 from guardian/comment-count"

TBonnin:
## What does this change?

Reverting commit breaking AMP validation
## What is the value of this and can you measure success?

AMP validation works
## Does this affect other platforms - Amp, Apps, etc?

Yes
## Request for comment

@piuccio @stephanfowler 

This reverts commit a0a4f30e5a7da66f648862b9b7285010f4e763a0, reversing
changes made to 7e2691fe444879b396f24e59fe9812b340c6c481.

PR #13847, opened 2016-08-03
Move comment count to standalone widget

piuccio:
## What does this change?

Take two of #13712 which was reverted in #13837

Turns out that custom elements break AMP. This PR wraps `.scala.html` into `!isAMP` so that the html is not included.

Same as the other PR, plus it doesn't break AMP .

The productivity benefit is controversial, I claim that
- changes are faster to implement
  - it's much quicker to work on a separate repo than inside frontend, especially if you're doing TDD
  - frontend is resistant to change, using ES6 is impossible, it's way easier to refactor from the bottom up than top down
- code is cleaner, it forces you to have a clear interface between units
- code can be reused
  - surely no one else is using `comment-count` at the moment, but it might be handy for apps in the future maybe?

Drawbacks highlighted so far
- Needs for 2 PR to make a single change
- Need to switch between 2 repos when reading code 
- Impossible to grep for something
- Hard to answer the question "what change in this feature?" since changes happened in multiple repos

Amp
## Request for comment

@TBonnin @guardian/guardian-frontend

### DIFF
--- a/applications/app/views/fragments/crosswords/crosswordMetaHeader.scala.html
+++ b/applications/app/views/fragments/crosswords/crosswordMetaHeader.scala.html
@@ -29,6 +29,7 @@
                 </div>
             }
 
+            @if(!requestHeader.isAmp) {
             <div class="meta__extras meta__extras--crossword">
                 <div class="meta__social" data-component="share">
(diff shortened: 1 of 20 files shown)


johnduffell:
is this still relevant?

PR #13848, opened 2016-08-03
Hosted native card mobile fix

Calanthe:
## What does this change?

A small fix of min-height value of native card on mobile breakpoint so the 'dead space' is reduced.
## Screenshots

Before:
screen shot 2016-08-03 at 16 14 12

After:
screen shot 2016-08-03 at 16 13 55
## Request for comment

@lps88


regiskuckaertz:
👍

PR #13850, opened 2016-08-03
Inject env mode in admin app (or removing use of app mode)

TBonnin:
## What does this change?

Inject env mode in admin app (or removing use of app mode if possible)

I recommend to review commit by commit.
Last commit (20e90a9) is a refactor of Admin to use the same authentication mechanism than preview (ie: authentication via Filters rather than authenticating every single endpoint). It would allow testing Admin controller without having to have a `testUser`, as well as removing some code duplication

No more `Play.isDev` or `Play.isTest` which are deprecated in Play 2.5

No
## Request for comment

@johnduffell @alexduf


johnduffell:
go for it! 👍

PR #13851, opened 2016-08-03
Remove useless background and border

regiskuckaertz:
`.ad-slot--container-inline` already has a background color and I replaced the border with a padding... more accurate

cc @guardian/commercial-dev


piuccio:
👍

PR #13852, opened 2016-08-03
Allow tracking display of Messages in Ophan; track engagement banner

rtyley:
## What does this change?

This allows you to track impressions of Messages displayed on dotcom frontend.

In particular, we want to track impressions of the engagement banner so we can evaluate user-conversion to membership.

Nope
## Request for comment

cc @markjamesbutler @AWare @philwills @joelochlann (thanks to @desbo for working out for me that ophan can't be included as a straight dependency!)

The `trackComponentAttention` method in Ophan was added quite recently, see https://github.com/guardian/ophan/pull/1522 .

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/membership-messages.js
+++ b/static/src/javascripts/projects/common/modules/commercial/membership-messages.js
@@ -107,6 +107,8 @@ define([
             pinOnHide: false,
             siteMessageLinkName: 'membership message',
             siteMessageCloseBtn: 'hide',
+            siteMessageComponentName: message.campaign,
+            trackDisplay: true,
             cssModifierClass: cssModifierClass
         }).show(template(messageTemplate, data));
     }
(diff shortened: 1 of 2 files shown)


gtrufitt:
Gotta bring master in but otherwise 👍

joelochlann:
Bit late on the bandwagon with this one but aren't these site messages `position: fixed` meaning they are always in the viewport? The component attention was intended to track attention on elements which are not necessarily in the viewport on page load and which the user might scroll past quickly or not scroll to at all. Not sure if this is the best mechanism to track impressions of banners like these?

rtyley:
> Not sure if this is the best mechanism to track impressions of banners like these?

I can see your point - the banners are indeed always on the page - but I did run this method past @philwills , so I think it's blessed by one of the surviving gods of Ophan :smile:

philwills:
I think @joelochlann's point is good though. Maybe it's worth having another chat to think about alternatives.

PR #13853, opened 2016-08-03
Remove useless code from admin

piuccio:
## What does this change?

I was on my quest of deprecating `reqwest` when I realised that some of the admin files are totally useless.

The path `/analytics/browsers` was removed by #4588 while the others by #12745

One fewer `common/utils/ajax` to fix

Admin
## Request for comment

@guardian/commercial-dev I'm not quite sure how your admin works, but I couldn't find any reference to the files I deleted. Do you mind double checking you're not using those files?


JonNorman:
I think that `AdTests.js` is used [here](https://frontend.gutools.co.uk/commercial/adtests), but I think the rest of it can go...

piuccio:
I tried that page but the `init` never gets called

JonNorman:
Ah okay, I think perhaps the code in `Admin` should have been updated when [these changes](https://github.com/guardian/frontend/commit/8527f43b4a0ceb01a2495eb523ab4adf7ce46df1#diff-5ac85536bd706648482d45335031b3faR90) were made to the admin routes file. That said, if the page still works once you've deleted `adTests.js` then my wires may just be crossed and I'll give you a :+1: 😺

piuccio:
I've also removed the button

JonNorman:
C'est magnifique :+1:

PR #13855, opened 2016-08-03
Removed unused switch to fix master build.

markjamesbutler:
## What does this change?

It removes an expired switch which is causing master to fail and blocking my deployment.

Fixes build.

No


common/app/conf/switches/ABTestSwitches.scala
```diff
@@ -43,17 +43,7 @@ trait ABTestSwitches {
     sellByDate = new LocalDate(2016, 8, 22),
     exposeClientSide = true
   )
-
-  val ABVideoCaption = Switch(
```
JustinPinner: Better to extend the switch than remove it? There will be other code running off of it that also needs to come out, but that'll be the call of the owner (Gideon in this case).

PR #13862, opened 2016-08-04
New header edition picker z-index bonanza

zeftilldeath:
Stops the edition picker hiding behind things, like the big scaredy🐈  it is.  

Hello @NataliaLKB


NataliaLKB:
👍

PR #13864, opened 2016-08-04
Remove API key from radiator

piuccio:
## What does this change?

RiffRaff only allows access from Guardian IPs, still api keys should be kept private. I've rotated the key and instead of hard coding it, I'm passing it from the controller.

Keys are private

Radiator only
## Request for comment

@adamnfish


piuccio:
@TBonnin I've rebased after your admin changes, should be good to go now

TBonnin:
I guess in dev mode the riffraff link won't be working (since it won't contain the api key)

PR #13865, opened 2016-08-04
Using aria-hidden instead of hidden

NataliaLKB:
## What does this change?

using `hidden` came with some default styling (`display:none !important;`) that I don't want. I think using aria-hidden is the best compromise. This way I can still be very explicit, while not having to override importants 😬 .

Nope!
## Request for comment

@zeftilldeath

### DIFF
--- a/common/app/views/fragments/nav/editionPicker.scala.html
+++ b/common/app/views/fragments/nav/editionPicker.scala.html
@@ -15,7 +15,7 @@
         @Edition(request).displayName
     </label>
 
-    <ul class="edition-picker__dropdown js-edition-picker-dropdown" id="edition-picker__dropdown" hidden>
+    <ul class="edition-picker__dropdown js-edition-picker-dropdown" id="edition-picker__dropdown" aria-hidden="true">
         @Edition.others(request).map { edition =>
             <li class="edition-picker__dropdown__item">
                 <a class="edition-picker__dropdown-item-link"
--- a/static/src/javascripts/projects/common/modules/navigation/edition-picker.js
+++ b/static/src/javascripts/projects/common/modules/navigation/edition-picker.js
@@ -35,11 +35,11 @@ define([
                     if (button.hasClass('open')) {
                         button.removeClass('open');
                         button.attr('aria-expanded', 'false');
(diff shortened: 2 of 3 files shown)


SiAdcock:
I love it. Just a thought, do you think we could / should use it is a CSS hook for making things hidden?

```css
.edition-picker__dropdown[aria-hidden="true"] {
  display: none;
}
  
.edition-picker__dropdown[aria-hidden="false"] {
  display: block;
}
```

Although, looking at the actual CSS for the edition picker, it is probably going to be slightly more complicated than that!

NataliaLKB:
> Although, looking at the actual CSS for the edition picker, it is probably going to be slightly more complicated than that!

I like that in principle, but it may actually lead to more confusing code. The edition picker needs to work without any JS. Unfortunately until the checkbox is enhanced to a button the aria attributes won't change. 

Open to change it anyway if you think its better.

NataliaLKB:
As far as I can tell, the change would be:

From this:
```
.edition-picker__checkbox, .edition-picker__button {
    display: none;

    &:checked, &.open {
        & ~ .edition-picker__dropdown {
            display: block;
            background-color: $guardian-brand-dark;
        }
        & ~ .edition-picker__current-edition {
            z-index: 3;
        }
    }
}
```

To This:
```
.edition-picker__checkbox {
    display: none;

    &:checked {
        & ~ .edition-picker__dropdown {
            display: block;
            background-color: $guardian-brand-dark;
        }
        & ~ .edition-picker__current-edition {
            z-index: 3;
        }
    }
}

.edition-picker__button {
    display: none;

    &.open ~ .edition-picker__current-edition {
        z-index: 3;
    }
}

.edition-picker__dropdown[aria-hidden=false] {
    display: block;
    background-color: $guardian-brand-dark;
}
```

SiAdcock:
Ah, okay, yeah I can't see how you'd make it work without JS. It's still a good change though, so 👍  💎  🎊

SiAdcock:
Actually, I like how well-factored the CSS is for the JS and no-JS versions at present. Totally cool with leaving it how it is! Just causing trouble 😄

PR #13867, opened 2016-08-04
Hosted logos clickable

Calanthe:
## What does this change?

The logos in the top left corner should be clickable. If no link is provided by the client (and currently none of them are) then the CTA link should be used. 
## Request for comment

@lps88


lps88:
Might be neater to use options instead of empty strings. But if it works! 👍

lps88:
👍  🐙

PR #13860, opened 2016-08-04
Inject env mode in admin app (or removing use of app mode) Try #2

TBonnin:
## What does this change?

Second attempt at https://github.com/guardian/frontend/pull/13850

`mode` in AdminJobsService was conflicting with `mode` in FrontendAppComponents added in a previous PR
## What is the value of this and can you measure success?

getting closer to upgrading to Play 2.5
## Does this affect other platforms - Amp, Apps, etc?

No
## Request for comment

@johnduffell (you already reviewed the previous PR, I only added 2a9874e)

PR #13868, opened 2016-08-04
Inject env mode in admin app (or removing use of app mode) Try #3

TBonnin:
## What does this change?

3rd attempt after https://github.com/guardian/frontend/pull/13860 🔁 

Admin endpoint are not AuthActions anymore (Authentication is handled via
app-wide Filters)
AuthLogging was only used in 3 controllers to log the authenticated user, nothing that
seems absolutely required.
Any action on switchboard is already recording the requester by using session identity.
## Request for comment

@johnduffell


johnduffell:
fair enough, no point going to all that effort for logging we're not really using 👍

PR #13870, opened 2016-08-04
Test demand for series reminder emails

joelochlann:
Tests demand for a reminder email about articles that are part of a series. Ultimately we're hoping that such a feature would increase visit frequency for our more infrequent visitors.

This reminder might be one-off or regular (e.g. weekly), and we may either mention the specific story coming up next week or provide a general description of the series - hence four test variants. Since we want to run this on both Blind Date and Experience articles and we have further copy variation between the two of these, there are actually eight possible permutations that a user might see:
#### Weekly reminder, series description

picture 257
#### Weekly reminder, coming up

picture 256
#### One-off reminder, series description

picture 255
#### One-off reminder, coming up

picture 258
#### Weekly reminder, series description

picture 261
#### Weekly reminder, coming up

picture 260
#### One-off reminder, series description

picture 259
#### One-off reminder, coming up

picture 258

After clicking the CTA button, the button is replaced with 404 text like so:

picture 263

@katebee

### DIFF
--- a/static/src/javascripts/projects/common/modules/experiments/tests/remind-me-email.js
+++ b/static/src/javascripts/projects/common/modules/experiments/tests/remind-me-email.js
@@ -30,8 +30,8 @@ define([
         this.author = 'Joseph Smith';
         this.description = 'Add a signup for a reminder email at the bottom of series articles';
         this.showForSensitive = false;
-        this.audience = 0;
-        this.audienceOffset = 0;
+        this.audience = 0.2;
+        this.audienceOffset = 0.7;
         this.successMeasure = 'Greatest clickthrough rate';
         this.audienceCriteria = 'All users';
         this.dataLinkNames = '';
--- a/static/src/javascripts/projects/common/modules/experiments/tests/remind-me-email.js
+++ b/static/src/javascripts/projects/common/modules/experiments/tests/remind-me-email.js
@@ -141,7 +141,7 @@ define([
 
         function completer(complete) {
             mediator.on('remind-me-email:insert', function () {
-                bean.on(qwery('#remind-me-signup')[0], 'click', function (){
+                bean.on(qwery('.js-remind-me-signup')[0], 'click', function (){
                     var html404 =
                         '<p class="remind-me__cta">' +
                             'Yessss! Another click! Thanks for your interest in this feature, we’re testing demand. ' +
--- a/static/src/javascripts/projects/common/modules/experiments/tests/remind-me-email.js
+++ b/static/src/javascripts/projects/common/modules/experiments/tests/remind-me-email.js
@@ -1,0 +1,194 @@
+define([
+    'bean',
+    'qwery',
+    'common/utils/$',
+    'common/utils/template',
+    'common/views/svg',
+    'common/utils/fastdom-promise',
+    'common/utils/mediator',
+    'text!common/views/experiments/remind-me-email-signup.html',
+    'inlineSvg!svgs/icon/envelope',
+    'lodash/objects/defaults'
+], function (bean,
+             qwery,
+             $,
+             template,
+             svg,
+             fastdom,
+             mediator,
+             remindMeEmailSignup,
+             envelope,
+             defaults
+) {
+    return function () {
+
+        this.id = 'RemindMeEmail';
+        this.start = '2016-07-18';
+        this.expiry = '2016-08-10';
+        this.author = 'Joseph Smith';
+        this.description = 'Add a signup for a reminder email at the bottom of series articles';
+        this.showForSensitive = false;
+        this.audience = 0;
+        this.audienceOffset = 0;
+        this.successMeasure = 'Greatest clickthrough rate';
+        this.audienceCriteria = 'All users';
+        this.dataLinkNames = '';
+        this.idealOutcome = '';
+        this.canRun = function () {
+            return isBlindDate() || isExperience();
+        };
+
+        var copy = {
+            oneOff: {
+                comingUp: {
+                    blindDate: {
+                        subtitle: 'Coming up in next week&rsquo;s Blind Date',
+                        title: '<b>&lsquo;She didn’t judge me for talking about giraffes&rsquo; necks&rsquo;</b>',
+                        description: 'Did digital communications consultant Lucy, 23, click with waitress Charlie, 20?',
+                        cta: 'Remind me',
+                        dataLinkName: 'one-off_coming-up_blind-date'
+                    },
+                    experience: {
+                        subtitle: 'Coming up in next week&rsquo;s Experience',
+                        title: '<b>I was stranded in the wilderness for nine days</b>',
+                        description: '&lsquo;I had lost so much weight that I looked like walking skeleton. I was ready to give up&rsquo;',
+                        cta: 'Remind me',
+                        dataLinkName: 'one-off_coming-up_experience'
+                    }
+                },
+                seriesDescription: {
+                    blindDate: {
+                        subtitle: null,
+                        title: 'What is <b>Blind Date?</b>',
+                        description: 'Every Saturday we fix up two Guardian readers',
+                        cta: 'Remind me next Saturday',
+                        dataLinkName: 'one-off_series-description_blind-date'
+                    },
+                    experience: {
+                        subtitle: null,
+                        title: 'What is <b>Experience?</b>',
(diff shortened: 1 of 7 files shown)


static/src/javascripts/projects/common/modules/experiments/tests/remind-me-email.js
```diff
@@ -0,0 +1,194 @@
...
+                        description: 'Every Saturday a Guardian reader shares their unique story',
+                        cta: 'Get a weekly reminder',
+                        dataLinkName: 'weekly_series-description_experience'
+                    }
+                }
+            }
+        };
+
+        function isBlindDate() {
+            return window.guardian.config.page.series === 'Blind date';
```
gtrufitt: You should import the config module, instead of looking at the window ` 'common/utils/config'`

static/src/javascripts/projects/common/modules/experiments/tests/remind-me-email.js
```diff
@@ -0,0 +1,194 @@
...
+        }
+
+        function isExperience() {
+            return window.guardian.config.page.series === 'Experience';
+        }
+
+        function insertSignupBox(templateData) {
+            var params = defaults({
+                svg: svg(envelope, ['submit-input__icon'])
```
gtrufitt: Try to import SVGs directly, as you don't need the entire module ` 'inlineSvg!svgs/icon/marque-36'`

static/src/javascripts/projects/common/modules/experiments/tests/remind-me-email.js
```diff
@@ -0,0 +1,194 @@
...
+            }, templateData);
+
+            var $signupBox = $.create(template(remindMeEmailSignup, params));
```
gtrufitt: I don't think you need to call create on this as you seem to just be inserting it directly into the DOM anyway and can just do that with the returned mark-up?

static/src/javascripts/projects/common/modules/experiments/tests/remind-me-email.js
```diff
@@ -0,0 +1,194 @@
...
+
+
+        function getTemplateData(data) {
+            return window.guardian.config.page.series === 'Blind date' ? data.blindDate : data.experience;
+        }
+
+
+        function completer(complete) {
+            mediator.on('remind-me-email:insert', function () {
+                bean.on(qwery('#remind-me-signup')[0], 'click', function (){
```
gtrufitt: As you're importing `$` anyway, you can just use `$('#remind-me-signup).on('click'...`. Also you should try and use `.js-` selectors in your JS instead of IDs (or non .js selectors)

static/src/javascripts/projects/common/modules/experiments/tests/remind-me-email.js
```diff
@@ -0,0 +1,194 @@
...
+                    var html404 =
+                        '<p class="remind-me__cta">' +
+                            'Yessss! Another click! Thanks for your interest in this feature, we’re testing demand. ' +
```
gtrufitt: This is suuuuuper colloquial language, not sure if it's too 'fun'!?

static/src/stylesheets/module/experiments/_remind-me.scss
```diff
@@ -0,0 +1,41 @@
...
+.remind-me {
+    background-color: $remind-me-color;
+    padding: $gs-baseline/4 $gs-gutter/2 $gs-baseline/2;
+    border-top: 3px solid $features-support-3;
+    color: #ffffff;
+}
+
+.remind-me__subtitle {
+    @include fs-textSans(5);
+    font-weight: bold;
```
gtrufitt: I don't think you should need to mess with font-weights? @superfrank ?

superfrank: We use a bold version of text sans. Or are you suggesting there's a better way of providing it in bold?

gtrufitt: No, I think this is ok actually - it's applied to here 
image but I thought 'subtitle' would be a heading and using the heading mixins but its not, so nevermind!

static/src/javascripts/projects/common/modules/experiments/tests/remind-me-email.js
joelochlann: I thought I was importing an SVG directly? I'm importing `'inlineSvg!svgs/icon/envelope'`

gtrufitt: Indeed you are - apologies, I thought you were importing /svgs 👍

static/src/javascripts/projects/common/modules/experiments/tests/remind-me-email.js
joelochlann: Celine from Editorial wrote this! I think it's OK - remember it's appearing on Blind Date and Experience stories, not ones about atrocities in Syria

static/src/stylesheets/module/experiments/_remind-me.scss
joelochlann: Yeah, subtitle might not be the best name but couldn't think of a better one... "surtitle" maybe??

static/src/javascripts/projects/common/modules/experiments/tests/remind-me-email.js
joelochlann: Ha! Yes I know, this syntax drives me nuts (it's even worse when you need to bind handlers to a set of elements) but `bonzo` doesn't have an `.on()` method - one of the many reasons why we should just use jQuery instead. (Or at least modify our custom `$` function so it can bind events)

👍  for the point about the `.js-` selector

static/src/javascripts/projects/common/modules/experiments/tests/remind-me-email.js
gtrufitt: 😆 Haha, yeah, fair enough, just wanted to check as not see much copy like this here!

static/src/stylesheets/module/experiments/_remind-me.scss
gtrufitt: It's all good, I did the same on email sign-up :shrug:

static/src/javascripts/projects/common/modules/experiments/tests/remind-me-email.js
joelochlann: There's no obvious parent to set `.innerHTML` on, so I'd have to use native `.insertBefore()` which requires a DOM element anyway (rather than markup). It would seem from [this](https://github.com/ded/bonzo#bonzoinsertbeforetarget) that I could pass markup to bonzo and it would automatically turn it into a DOM node for insertion, but having tried it just now, it didn't work (it's still a string by the time it's trying to put it into the DOM). And [looking at the bonzo code](https://github.com/ded/bonzo/blob/master/src/bonzo.js#L311-L322), it makes sense because I can't see any logic to create elements on the fly if you pass markup into bonzo. Guess the documentation is wrong.

So, yet another reason to hate `bonzo`.

static/src/javascripts/projects/common/views/experiments/remind-me-email-signup.html
```diff
@@ -0,0 +1,11 @@
+<div class="remind-me">
+    <% if (subtitle) { %>
+    <h2 class="remind-me__subtitle"><%= subtitle %></h2>
+    <% } %>
+    <h1 class="remind-me__title"><%= title %></h1>
```
gtrufitt: Why is this a `h1`?

gtrufitt:
One last query, but otherwise 👍

static/src/javascripts/projects/common/views/experiments/remind-me-email-signup.html
joelochlann: Yep fair point! I guess because it's the main title of this component but of course there are more important headings elsewhere in the page - kind of forgot I wasn't inside an iframe when writing that (did a lot of experiments with embeds previously). Fixed

joelochlann:
Thanks @gtrufitt

static/src/javascripts/projects/common/views/experiments/remind-me-email-signup.html
gtrufitt: Yeah, I did have to check you weren't in an iframe haha :)

PR #13871, opened 2016-08-04
AMP live blog key events

SiAdcock:
## What does this change?

Adds a Key Events navigation to the top of the AMP live blog.

Clicking a key event link will redirect the user to the non-AMP version of the live blog, anchored to the relevant key event blog post.

As part of this change, I have also updated the permalinks in the blog post timestamps to link to the non-AMP version of the live blog post.
## Screenshots

picture 65
## Request for comment

@NataliaLKB @zeftilldeath @jfsoul @johnduffell @TBonnin

### DIFF
--- a/article/app/views/liveblog/liveBlogBlocks.scala.html
+++ b/article/app/views/liveblog/liveBlogBlocks.scala.html
@@ -6,10 +6,12 @@
 
 @(blocks: Seq[BodyBlock], article: Article, timezone: DateTimeZone, amp: Boolean = false)(implicit request: RequestHeader)
 
+@import conf.Configuration
+
 @blocks.map { block =>
     <div id="block-@block.id" @if(amp && LiveUpdateAmpSwitch.isSwitchedOn){ data-sort-time="@block.publishedCreatedTimestamp()" } class="block block--content@block.eventClass" itemprop="liveBlogUpdate" itemscope="" itemtype="http://schema.org/BlogPosting">
         <p class="block-time published-time">
-            <a href="/@article.metadata.id?page=with:block-@block.id#block-@block.id" itemprop="url" class="block-time__link">
+            <a href="@if(amp){@Configuration.site.host}/@article.metadata.id?page=with:block-@block.id#block-@block.id" itemprop="url" class="block-time__link">
             @views.html.liveblog.dateBlock(block.publishedCreatedDate(timezone))
             </a>
         </p>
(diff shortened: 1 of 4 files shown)


article/app/views/liveblog/liveBlogBlocks.scala.html
```diff
@@ -6,10 +6,12 @@
 
 @(blocks: Seq[BodyBlock], article: Article, timezone: DateTimeZone, amp: Boolean = false)(implicit request: RequestHeader)
 
+@import conf.Configuration
+
 @blocks.map { block =>
     <div id="block-@block.id" @if(amp && LiveUpdateAmpSwitch.isSwitchedOn){ data-sort-time="@block.publishedCreatedTimestamp()" } class="block block--content@block.eventClass" itemprop="liveBlogUpdate" itemscope="" itemtype="http://schema.org/BlogPosting">
         <p class="block-time published-time">
-            <a href="/@article.metadata.id?page=with:block-@block.id#block-@block.id" itemprop="url" class="block-time__link">
+            <a href="@if(amp){@Configuration.site.host}/@article.metadata.id?page=with:block-@block.id#block-@block.id" itemprop="url" class="block-time__link">
```
SiAdcock: What do people think of this? I'm making the permalink in the timestamp an absolute URL in AMP. Should this be an absolute URL in non-AMP also?

johnduffell: for fully qualified urls you should probably use [LinkTo url builder](https://github.com/guardian/frontend/blob/master/common/app/common/LinkTo.scala)

johnduffell:
great, not sure if it should be folded up rather than keeping it open all the time.  I think the mobile version of the normal one used to be like that, I must have broken that when I made it work on the server side 😊  but noone has complained... other than that, great 👍

SiAdcock:
Thanks @johnduffell. I might try and experiment with `amp-accordion` to see how folding up the content would look.

SiAdcock:
I have made the Key Events navigation into an `amp-accordion` that is collapsed by default:

picture 66

I have also used the `LinkTo` URL builder to create URLs to key events.

johnduffell:
go for it, if it works already you can always make a second PR for changes later! 👍 ship early, ship often, right? 😛

SiAdcock:
I love it, @johnduffell. Apology > permission

PR #13874, opened 2016-08-04
Fix regression with rubicon ads

regiskuckaertz:
The cross-frame protocol I created requires a simili UUID in the JSON payload, but Rubicon ads do not send that information and therefore the top banner does not get resized anymore:

Now it does:
picture 1

cc @guardian/commercial-dev @piuccio


piuccio:
👍

PR #13878, opened 2016-08-04
Add 88x89 size for the high-merch component on ad features pages

regiskuckaertz:
The team needs the high-relevance commercial component on advertisement feature pages (articles, galleries, videos, etc.), but if we add it as-is it will be fed creatives that are all over the place (bookshop, masterclasses, etc.).

Now, we want some control over the creatives, and the fastest way to do that is to add a new custom size. This is what this PR does: the 88x89 size will only be used in advertisement features.

cc @guardian/commercial-dev

### DIFF
--- a/common/app/views/fragments/commercial/commercialComponentHigh.scala.html
+++ b/common/app/views/fragments/commercial/commercialComponentHigh.scala.html
@@ -1,4 +1,4 @@
-@(isAdFeature: Boolean = false)
+@(isAdFeature: Boolean)
 @import conf.switches.Switches._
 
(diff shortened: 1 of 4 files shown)


common/app/views/fragments/commercial/commercialComponentHigh.scala.html
```diff
@@ -1,10 +1,11 @@
+@(isAdFeature: Boolean = false)
```
rich-nguyen: Don't think the default value is needed here

rich-nguyen:
👍

PR #13879, opened 2016-08-04
Discover which ads don't meet quality expectations

JustinPinner:
## What does this change?

For (initially) 2% of readers, we'll make it possible to report poor-quality ads in order to build up a picture of where our current ad experience needs improvement.

By knowing which kinds of ads (and particularly specific instances) are problematic for users we can be in a better position to improve overall quality.

No.
## Screenshots

**Before;**
screen shot 2016-08-04 at 17 59 24

screen shot 2016-08-04 at 18 00 19

**During;**
screen shot 2016-08-04 at 18 01 42

screen shot 2016-08-04 at 18 02 32

**After;**
screen shot 2016-08-04 at 18 03 10

screen shot 2016-08-04 at 18 03 47
## Request for comment

@rich-nguyen @regiskuckaertz @sndrs


paperboyo:
💎
I suspect only a very low number of readers would miss the option of “Performance hog”, right? And there are already so many choices there...

static/src/javascripts/projects/common/modules/commercial/ads/user-ad-feedback.js
```diff
@@ -0,0 +1,48 @@
...
+    fetch,
+    config
+) {
+
+    function userAdFeedback(pagePath, adSlotId, creativeid, reason) {
+        // TODO: This. Better.
+        if (reason !== 'ad-feedback-menu-opened') {
+            return recordUserAdFeedback(pagePath, adSlotId, creativeid, reason);
+        } else {
+            return recordUserAdFeedback(pagePath, adSlotId, creativeid, 'ad-feedback-menu-opened');
```
regiskuckaertz: But if the code enters this branch, that means `reason === 'ad-feedback-menu-opened'` and so
```
return recordUserAdFeedback(pagePath, adSlotId, creativeid, 'ad-feedback-menu-opened');
```
is equivalent to
```
return recordUserAdFeedback(pagePath, adSlotId, creativeid, reason);
```
and thus you don't need the if at all. Consequently, you don't need `userAdFeedback` either.

static/src/javascripts/projects/common/modules/commercial/ads/user-ad-feedback.js
```diff
@@ -0,0 +1,48 @@
...
+        }
+    }
+
+    function recordUserAdFeedback(pagePath, adSlotId, creativeId, feedbackType) {
+        var feedbackUrl = 'https://j2cy9stt59.execute-api.eu-west-1.amazonaws.com/prod/adFeedback';
+        var stage = config.page.isProd ? 'PROD' : 'CODE';
+        var data = '{"stage":"'+stage+'", "adPage":"'+pagePath+'", "adSlotId":"'+adSlotId+'", "adCreativeId":"'+creativeId+'", "adFeedback":"'+feedbackType+'"}';
```
regiskuckaertz: Slightly easier to read:

`var data = JSON.stringify({
  stage: stage,
  adPage: pagePath,
  adSlotId: adSlotId,
  adCreativeId: creativeId,
  adFeedback: feedbackType
});
```

File: static/src/javascripts/projects/common/modules/commercial/ads/user-ad-feedback.js [hunk 3]
```diff
@@ -0,0 +1,48 @@
...
+        return fetch(feedbackUrl, {
+            method: 'post',
+            body: data,
+            mode: 'cors'
+        }).then(function() {    // ignore the response
```
regiskuckaertz: `Promise.then` accepts two callbacks: one when the promise is kept, one when the promise is broken (`.catch` is just sugar). You can avoid superfluous ticks this way:

```
fetch(...).then(onComplete, onComplete);
function onComplete() {
   if( feedbackType ...
}
```

static/src/javascripts/projects/common/modules/commercial/ads/user-ad-feedback.js
```diff
@@ -0,0 +1,48 @@
...
+        }).then(function() {    // we're complete - update the UI
+            if (feedbackType !== 'ad-feedback-menu-opened') {
+                var adSlot = bonzo(qwery('#' + adSlotId));
+                adSlot.each(function (ad) {
+                    for (var i = 0; i < ad.children.length; i++) {
+                        if (ad.children[i].classList.contains('ad-slot__label')) {
+                            ad.children[i].innerHTML = 'Advertisement (Thanks for your feedback)';
+                        }
+                    }
+                });
```
regiskuckaertz: Lines 34–41 can be replaced with `bonzo(qwery('#' + adSlotId + '>.ad-slot__label').text('Advertisement (Thanks for your feedback)');`

static/src/javascripts/projects/common/modules/commercial/dfp/private/render-advert.js
```diff
@@ -120,6 +124,8 @@ define([
 
         return applyCreativeTemplate(advert.node).then(function (isRendered) {
             return renderAdvertLabel(advert.node)
+                .then(addFeedbackDropdownToggle())
+                .then(applyFeedbackOnClickListeners(slotRenderEvent.creativeId))
```
regiskuckaertz: You will have to wrap that call in a lambda, otherwise the operation will be executed synchronously:
```
.then(function () { return applyFeedbackOnClickListeners(slotRenderEvent.creativeId); })
```

static/src/javascripts/projects/common/modules/commercial/dfp/private/render-advert-label.js
```diff
@@ -1,14 +1,40 @@
...
 ) {
     function renderAdvertLabel(adSlotNode) {
         if (shouldRenderLabel(adSlotNode)) {
+            var zIndexOverlay = 1050;
+            var labelInner = '';
+            if (config.switches.abAdFeedback) {
+                try {
+                    var abParticipant = store.local.get('gu.ab.participations')['AdFeedback'];
+                    if (abParticipant && abParticipant.variant === 'ad-feedback-variant') {
+                        var feedBackOpts = '<li class="popup__item"><a><div class="popup__item-problem--option" slot="' + adSlotNode.id + '" problem="distracting">Distracting</div></a></li>';
```
regiskuckaertz: Let's make it easier for you to change the markup:

```
var feedBackOpts = [
  ['distracting'],
  ['obscures', 'Obscures content'],
  ['inappropriate'],
  ['repetitive'],
  ['irrelevant']
].map(function(_) {
  return '<li class="popup__item"><a><div class="popup__item-problem--option" slot="' + adSlotNode.id + '" problem="' + _[0] + '">' + (_[1] || _[0].replace(/^[a-z]/, function($1) { return $1.toUpperCase(); }) + '</div></a></li>';
}).join('');
```

JustinPinner:
Thanks for the great feedback @regiskuckaertz I'll work through those changes ⭐

static/src/javascripts/projects/common/modules/commercial/dfp/private/render-advert.js
JustinPinner: Is that true for `.then(addFeedbackDropdownToggle())` too?

regiskuckaertz: Yes - I missed that one - although in that case you can just get rid of the parentheses: `.then(addFeedbackDropdownToggle)`

JustinPinner: Ok, thanks. I think I'll be reading about `Promise` and related things this afternoon!

static/src/javascripts/projects/common/modules/commercial/ads/user-ad-feedback.js
```diff
@@ -10,39 +10,27 @@ define([
...
-                    for (var i = 0; i < ad.children.length; i++) {
-                        if (ad.children[i].classList.contains('ad-slot__label')) {
-                            ad.children[i].innerHTML = 'Advertisement (Thanks for your feedback)';
-                        }
-                    }
-                });
-            }
-        });
-    }
+        }).then(onComplete(adSlotId, feedbackType), function(){});
```
regiskuckaertz: Nearly there! You need to pass the same callback twice, so that it is called no matter what happens, a rejection or a resolve. And you also need to wrap it in a lambda:
```
.then(
  function() { return onComplete(adSlotId, feedbackType); },
  function() { return onComplete(adSlotId, feedbackType); }
)
```

static/src/javascripts/projects/common/modules/commercial/ads/user-ad-feedback.js
```diff
@@ -10,39 +10,27 @@ define([
...
+    };
 
-    return userAdFeedback;
+    function onComplete(adSlotId, feedbackType) {    // we're complete - update the UI
+        if (feedbackType !== 'ad-feedback-menu-opened') {
+            bonzo(qwery('#' + adSlotId + '>.ad-slot__label')).text('Advertisement (Thanks for your feedback)');
```
regiskuckaertz: Now let's wrap this in a `fastdom.write` and we're done.

JustinPinner:
Hopefully that's all done now. It'd nice to get this deployed this afternoon so we can switch on on CODE and leave off on PROD. Then folks can have a last look / tweak here and there.

regiskuckaertz:
Let's get this ball rolling! great work! 👍

PR #13880, opened 2016-08-04
Added xmlns to fix asset build.

markjamesbutler:
## What does this change?

Fixes xml namespace ref to fix the asset parsing and ultimately the build. 

The build is fixed.

No


TBonnin:
:+1: as long as `grunt svgmin` doesn’t throw any error I am good

PR #13882, opened 2016-08-05
Remove code branch that should never be needed

gtrufitt:
## What does this change?

Removes some unneeded code - if email sign-up is in the page first, Outbrain loads the right widget, if Outbrain is in the page first, email sign-up won't load so this 'pre-check' in Outbrain is no longer required.

Tidy tidy.

@guardian/commercial-dev @guardian/dotcom-platform


rich-nguyen:
👍

PR #13883, opened 2016-08-05
Another attempt at link preconnect resource hint

rich-nguyen:
I don't think the [first attempt](https://github.com/guardian/frontend/pull/12935) at this potential performance optimisation worked as intended. 
### Before

Here is a waterfall without preconnect, just dns-prefetch:
before preconnect
### After

Here is preconnect and dns-prefetch:
after preconnect

I'm totally speculating, but maybe the browser has prioritised all the preconnect hints in front of the resource loading itself, thus extending the critical path.

I'm going to try again with a single preconnect, just to test the water.

I've extended the thrift switch too.


Calum-Campbell:
👍

PR #13884, opened 2016-08-05
Admin: Add exemption for /deploys-notify authentication

TBonnin:
## What does this change?

Fix /deploy-notify endpoint (used to post deploy notifications on slack)

Yesterday I change the way the Admin app handles authentication:
Instead of having each endpoint take care of authentication themselves, it is now handle app-wide via a Play filter.
I forgot the /deploy-notify endpoint, because it is accessed by non-human, is authenticated via api-key.
This PR adds a exemption for the app-wide Google authentication for this specific endpoint

Slack deploy notification works again
## Request for comment

@katebee


jfsoul:
👍  can't live without those slack notifications!

PR #13885, opened 2016-08-05
Rio Olympics surging content email

Calum-Campbell:
## What does this change?
- Adds an automatic email sent to sales team when surging content related to Rio Olympics appears
  - Checks surging content every 30 minutes and filters for sport or Rio2016 related content
  - Grabs content with over 500 pageView/min
## Screenshots

image
## Request for comment

@JonNorman 
@rich-nguyen


admin/app/jobs/SurgingSportEmailJob.scala
```diff
@@ -0,0 +1,80 @@
...
+  private val matchesKeyword = content.exists(_.tags.keywordIds.exists(keyword =>
+    sportsTargetedKeywords.contains(keyword)
+  ))
+
+  val isSportsSurgingContent: Boolean = matchesKeyword && pageViewCount > 500
+}
+
+case class SurgingSportEmailJob(emailService: EmailService) extends Logging with ExecutionContexts {
+
+  private val subject = "New surging sports content"
```
JonNorman: Would it be useful to add a timestamp to the subject? Or on the admin page itself so we know when it was last updated?

admin/app/jobs/SurgingSportEmailJob.scala
```diff
@@ -0,0 +1,80 @@
...
+
+  def run() : Future[Unit] = {
+    val futureEmail: Option[Future[Unit]] = for {
+      adTech <- adTechTeam
+      surgingContentEmail <- surgingContentTeam
+    } yield {
+
+      getAllSportsSurgingContent
+        .map( surgingContent => {
```
JonNorman: I think you can use `onSuccess` here since you want to ultimately return `Unit`, which means you can get rid of the final `.map(_ => ())`

JonNorman: It will require reworking the signature of this method a bit but since we just want to return a `Future[Unit]` it would make sense to me.

admin/app/jobs/SurgingSportEmailJob.scala
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  def getContentFromId(id: String, viewCount: Int): Future[SportsSurging] = {
+    getResponse(ContentApiClient.item(id, Edition.defaultEdition))
+        .map(response =>
+          SportsSurging(
+            id = id,
+            pageViewCount = viewCount,
+            content = response.content.map(Content(_))
+          )
```
JonNorman: At this point it isn't definitely sports content, so maybe once the `SportsSurging` has been made, you can filter on `_.isSportsSurgingContent`. Looks a bit odd given the name of the class.

JonNorman:
Looks good to me :+1: - some minor points about names/return types but given that this is a quick hacky thing not long for this world, it might not be worth it.

PR #13888, opened 2016-08-05
fix threshold number

Calum-Campbell:
## What does this change?
- Quick fix for wrong threshold...
## Request for comment

@JonNorman


JonNorman:
This is an outrageous attempt of shirking, up with which I shall not put, but have a :+1:

PR #13890, opened 2016-08-05
Refactor comment ordering test

NathanielBennett:
## What does this change?

This is a refactor of this: https://github.com/guardian/frontend/pull/13754/files, which tested the effect the ordering of comments had on their consumption. This time, we want to distinguish between live-blogs ( where repeat visitors may be returning to read the latest comments ) and other commentable content, by running two separate tests.

As before, we want do determine how we can maximise the consumption of comments.

No
## Screenshots

N/A
## Request for comment

@crifmulholland @piuccio @nicl


piuccio:
👍

PR #13893, opened 2016-08-05
Hosted content - adblock fix

Calanthe:
## What does this change?

Currently all of the hosted pages are almost not visible behind the adblock addon. This is due to the class names `adverts` that were copied from the commercial components. We have two possible ways to fix this now (that I can think of, let me know if you have better idea): 
1. Rename all of the `adverts` classes into something else on both hosted and commercial templates (and scss files). Then test all of the commercial components on the page.
2. Rename all of the `adverts` classes on hosted templates only, copy and rename those styles to the `hosted.scss` file.

I chose the second solution because I didn't want to mess around with commercial components (which can and should be blocked by adblockers). 

Changes from this PR make hosted content look fine but galleries and videos don't work on CODE because js files (which are not minified and compressed) like `Advert.js`, `get-ads-by-id.js`, `get-adverts.js` are blocked by the adblocker... This is never ending story and I am not even sure if it's worth start fixing...
## Request for comment


regiskuckaertz:
👍

PR #13894, opened 2016-08-05
Messenger: use the location protocol

regiskuckaertz:
Small fix: when the page is running on HTTPS, ads will be loaded from HTTPS too.

@guardian/commercial-dev


Calanthe:
👍

PR #13896, opened 2016-08-05
Hosted autoplay Ab switch extended

Calanthe:
## What does this change?

The hosted autoplay AB test is extended by one week.


gtrufitt:
👍 +1

PR #13897, opened 2016-08-05
Add fluid size to merchandising slots

regiskuckaertz:
Since they will eventually be migrated to native ads too, but also to allow @sammorrisdesign to experiment with his hosted content component.

cc @guardian/commercial-dev


sammorrisdesign:
👍

PR #13898, opened 2016-08-05
Change tracking of Message banners

rtyley:
## What does this change?

Tracked Messages (currently, only the engagement banner) will have only the fact they were recorded displayed, rather than their attention time.

We shouldn't record meaningless things!

Nope, just web.

cc @joelochlann @philwills 

Prompted by feedback in https://github.com/guardian/frontend/pull/13852#issuecomment-237837097

### DIFF
--- a/static/src/javascripts/projects/common/modules/ui/message.js
+++ b/static/src/javascripts/projects/common/modules/ui/message.js
@@ -5,6 +5,7 @@ define([
     'common/modules/user-prefs',
     'common/utils/mediator',
     'common/utils/detect',
+    'common/modules/analytics/register',
     'lodash/functions/debounce',
     'lodash/arrays/uniq'
 ], function (
@@ -14,6 +15,7 @@ define([
     userPrefs,
     mediator,
     detect,
+    register,
     debounce,
     uniq
 ) {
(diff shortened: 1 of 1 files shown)


joelochlann:
👍  🙌  👏  👌

rtyley:
You can see Ophan sending the new messages:

image

```
register:[{"name":"MEMBERSHIP_SUPPORTER_BANNER_UK","status":"unfinished"}]
```

The `"status":"unfinished"` refers to the part of [`analytics/register.js`](https://github.com/guardian/frontend/blob/8134b23ba/static/src/javascripts/projects/common/modules/analytics/register.js) that tracks the render time of the component - we don't really care about the render time of the component (because it's basically instant), so like many other users of this library we don't bother to call `end()`.

The component data is known as [`lazyComponents`](https://github.com/guardian/ophan-data-lake/search?utf8=%E2%9C%93&q=lazyComponents) within the Ophan data-lake.

tomseddon:
SELECT component
FROM raw.ophan_event LATERAL VIEW explode (lazycomponents.components) c as component
WHERE day = '2016/08/09/'
and component.name = 'MEMBERSHIP_SUPPORTER_BANNER_UK' LIMIT 10

PR #13899, opened 2016-08-06
Guard against undefined saved for later

piuccio:
## What does this change?

As of yesterday there's a spike of errors `Unable to get property 'addEventListener' of undefined or null reference`.

I noticed this while investigating an issue with the current skin, but that seems to be unrelated.

I tracked down the error to the code below, for whatever reason the link is undefined (switch turned off maybe or video container, who knows).

I don't fancy understanding why exactly we're calling the function with an undefined link, for now a simple guard seems enough.

Fix a sentry error reported 1.3k times in the last 24h

No
## Request for comment

@regiskuckaertz


regiskuckaertz:
:+1:

PR #13900, opened 2016-08-06
Fix two more addEventListener errors

piuccio:
## What does this change?

Turns out #13899 was not the only place adding event listeners on undefined elements. This fixes other 2 occurrences

Reduce sentry errors

No
## Request for comment

@regiskuckaertz @Calanthe @lps88


regiskuckaertz:
The problem is accessing element [0] of an empty array so you should test for that instead

piuccio:
like this?

regiskuckaertz:
:+1:

PR #13902, opened 2016-08-06
Fix another DOM access of undefined object

piuccio:
## What does this change?

It might happen that save for later elements on't have a link (why?), in which case all the fastdom manipulations are broken. Skip those

Fix this error

screen shot 2016-08-06 at 17 40 02

Happens on `/uk` front right now

No
## Screenshots

Above
## Request for comment

@NathanielBennett I've deleted a forEach loop that seems useless to me. However most likely it was supposed to do something, mind having a look?

It feels like there are cases where `save for later` link are not exactly what we expect, I didn't look into it too hard because it's Saturday, but we should understand the root of the problem.

### DIFF
--- a/static/src/javascripts/projects/common/modules/save-for-later.js
+++ b/static/src/javascripts/projects/common/modules/save-for-later.js
@@ -163,11 +163,7 @@ define([
     };
 
     SaveForLater.prototype.getElementsIndexedById = function (context) {
-        var elements = qwery('[' + this.attributes.containerItemShortUrl + ']', context);
-
-        return forEach(elements, function (el) {
-            return bonzo(el).attr(this.attributes.containerItemShortUrl);
-        }.bind(this));
+        return qwery('[' + this.attributes.containerItemShortUrl + ']', context);
     };
 
     SaveForLater.prototype.prepareFaciaItemLinks = function (signedIn) {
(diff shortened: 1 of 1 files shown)


static/src/javascripts/projects/common/modules/save-for-later.js
```diff
@@ -207,6 +203,10 @@ define([
                 id = item.getAttribute(this.attributes.containerItemDataId),
                 isSaved = signedIn ? this.getSavedArticle(shortUrl) : false;
 
+            if (!$itemSaveLink) {
```
regiskuckaertz: Should probably be `!$itemSaveLink.length` #bonzo

piuccio: thanks for spotting this

piuccio:
@NathanielBennett are you ok with the changes?

johnduffell:
ready to go? or no longer needed?

piuccio:
just need a +1

johnduffell:
sure have one from me 👍 😛

gtrufitt:
🚢 ⛵ 🐑

NathanielBennett:
:+1: I'm sorry @piuccio this slipped by me

PR #13903, opened 2016-08-06
Detect adblockers fails on IE8

piuccio:
## What does this change?

Don't run detect adblock on old browsers

Stop seeing [this error](https://app.getsentry.com/the-guardian/client-side-prod/issues/49237141/) on Sentry, the second most frequent error since the beginning of time.

No
## Screenshots

Damn IE8
## Request for comment

@Calanthe I'm sure there's a better way to prevent adblock detect from running on IE8, let's discuss on Monday


regiskuckaertz:
👍  our commercials don't run in IE8 anyway.

PR #13905, opened 2016-08-08
Amp five open menu

NataliaLKB:
## What does this change?

This adds the main functionality for opening the menu.

When you click any of the primary links (new, opinion, sport, arts, life), it should open the same menu the burger menu opens, but move the related link to the top of the list, and also open it.

Here is it in action:
openingandclosingexperiments

Nope!

### DIFF
--- a/static/src/javascripts/projects/common/modules/navigation/newHeaderNavigation.js
+++ b/static/src/javascripts/projects/common/modules/navigation/newHeaderNavigation.js
@@ -124,7 +124,7 @@ define([
         });
     }
 
-    function handlePrimaryItemClicks() {
+    function bindPrimaryItemClickEvents() {
         primaryItems.each(function (item) {
 
             item.addEventListener('click', closeAllOtherPrimaryLists.bind(null, item));
@@ -146,7 +146,7 @@ define([
         window.addEventListener('hashchange', handleHashChange);
         handleHashChange();
 
-        handlePrimaryItemClicks();
+        bindPrimaryItemClickEvents();
         openTargetListOnClick();
 
         editionPicker();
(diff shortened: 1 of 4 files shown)


static/src/javascripts/projects/common/modules/navigation/newHeaderNavigation.js
```diff
@@ -68,6 +84,53 @@ define([
...
+                       fastdomPromise.write(function () {
+                        menuToOpen.attr('open', '');
+                        return id;
+                    }).then(moveTargetListToTop.bind(id));
+                });
+            });
+        });
+    }
+
+    function handlePrimaryItemClicks() {
```
SiAdcock: Perhaps `bindPrimaryItemClickEvents` would be more semantic here

SiAdcock:
Looks proper sweet, some very nifty JS! Struggling to find anything to change other than pedantry! 👍

paperboyo:
Hmm, interesting! Good for discovery if a tad disrupting. One question. Never seen new menu working, but I presume, while browsing it, clicking on the top sections (news, opinion etc.) expands and collapses subsections. So what happens if one clicks sport on the site, menu opens with sport and its subsections expanded, but one would like to go to all sport content (not a specific subsection)?

paperboyo:
Maybe I assume wrong? Maybe to expand/collapse a section one _has to_ click/tap on chevrons (and tapping on a section name opens this section on the site, closing the menu)?

static/src/javascripts/projects/common/modules/navigation/newHeaderNavigation.js
NataliaLKB: Yes that makes sense! I will change it :)

static/src/stylesheets/layout/_ab-new-header-test-variant.scss
```diff
@@ -397,12 +397,15 @@ $gutter-large: 55px;
     // Unset ul from user agent stylesheet
     margin-bottom: 0;
     text-transform: lowercase;
+    display: flex;
```
NataliaLKB: I also should put a comment here to explain this is for the reordering which is added in JS.

NataliaLKB:
@paperboyo Good questions! Want to come over to the desks and see the prototype?

PR #13907, opened 2016-08-08
Fixing new header words wrapping on safari

NataliaLKB:
## What does this change?

Knocks the font-size down slightly for the new header primary links.

This is to prevent them from wrapping on Safari.

It looks better

Nope!
## Screenshots

Before:
image

After:
image
## Request for comment

@zeftilldeath


zeftilldeath:
thumb

PR #13908, opened 2016-08-08
Books logging

janua:
I am attempting to trace a bug with the books service. This just adds logging and some types.

@JonNorman @kelvin-chappell


kelvin-chappell:
👍

PR #13909, opened 2016-08-08
Fix Admin football fronts

TBonnin:
## What does this change?

Fix for https://frontend.gutools.co.uk/admin/football/fronts "No data available" error

I am about to open a ticket with PA support to understand why some competitions like the uefa champion's league or fa cup, etc... returns 'no data' errors when requesting the participating teams list

https://frontend.gutools.co.uk/admin/football/fronts works 😄 

No
## Request for comment

@johnduffell @jfsoul 

Do not fail when PA client returns an exception when 'no data available'
error.

### DIFF
--- a/admin/app/football/controllers/FrontsController.scala
+++ b/admin/app/football/controllers/FrontsController.scala
@@ -27,7 +27,14 @@ class FrontsController(val wsClient: WSClient, val mode: Mode.Mode) extends Cont
   def index = Action.async { implicit request =>
     for {
       competitions <- client.competitions.map(PA.filterCompetitions)
-      competitionTeams <- Future.traverse(competitions){comp => client.teams(comp.competitionId, comp.startDate, comp.endDate)}
+      competitionTeams <- Future.traverse(competitions){ comp =>
+        client.teams(comp.competitionId, comp.startDate, comp.endDate).recover {
+          case e: PaClientErrorsException =>
+            // 'No data' is returned as an error by PA API. Therefore we ignore exception and return an empty list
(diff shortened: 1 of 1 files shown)


admin/app/football/controllers/FrontsController.scala
```diff
@@ -27,7 +27,12 @@ class FrontsController(val wsClient: WSClient, val mode: Mode.Mode) extends Cont
   def index = Action.async { implicit request =>
     for {
       competitions <- client.competitions.map(PA.filterCompetitions)
-      competitionTeams <- Future.traverse(competitions){comp => client.teams(comp.competitionId, comp.startDate, comp.endDate)}
+      competitionTeams <- Future.traverse(competitions){ comp =>
+        client.teams(comp.competitionId, comp.startDate, comp.endDate).recover {
+          // 'No data' is returned as an error by PA API. Therefore we ignore exception and replace it by an empty list
+          case e: Exception => List()
```
jfsoul: Do we want to log/alert these exceptions somehow? I suppose we'll just get the heads up from central prod that there are competitions with empty lists anyway.

👍

jfsoul:
👍  let's get the tool working, football season starts soon!

admin/app/football/controllers/FrontsController.scala
johnduffell: which exception?

johnduffell: if you really want to swallow them at least log any that aren't the one we're expecting

TBonnin: You are right. let me limit those to the `PaClientException`

PR #13912, opened 2016-08-08
More book logging

janua:
Log out the reason as to why parsing is failing

@kelvin-chappell


kelvin-chappell:
Didn't realise this was unsafe.
👍

PR #13916, opened 2016-08-08
AMP live blogs have itemtype of NewsArticle

SiAdcock:
## What does this change?

This hardcodes the itemtype of AMP live blog article to be `NewsArticle`, and not `LiveBlogPosting`. The reason for this change is because Google's AMP carousel only supports articles of type `NewsArticle`.

Google's Sebastian Benz recommends:

> [...] the Top News Carousel currently supports only articles of the type NewsArticle - this is why your articles don't show up in the carousel. I'd recommend using NewsArticle markup for now and switch to LiveBlog once we've released the search integration, which will be released together with the amp-live-list component. I'll keep you updated.

No
## Request for comment

@johnduffell @jfsoul @NataliaLKB @TBonnin @gtrufitt

### DIFF
--- a/article/app/views/liveblog/liveBlogBody.scala.html
+++ b/article/app/views/liveblog/liveBlogBody.scala.html
@@ -1,7 +1,8 @@
 @import _root_.liveblog.KeyEventData
 @import model.Badges.badgeFor
-
+@import model.ArticleSchemas
 @import _root_.liveblog.Canonical
+
 @(model: LiveBlogPage, amp: Boolean = false)(implicit request: RequestHeader)
 
 @import common.{Edition, LinkTo}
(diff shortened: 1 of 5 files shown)


article/app/views/liveblog/liveBlogBody.scala.html
```diff
@@ -14,9 +14,15 @@
...
+
+    @*
+        TODO: We override itemtype if amp as Google's AMP Carousel currently only
+            supports articles of type NewsArticle. We should remove this conditional
+            when LiveBlogPosting support is implemented
+    *@
     <article id="live-blog" data-test-id="live-blog"
         class="content content--liveblog tonal tonal--@toneClass(article) blog @if(article.fields.isLive){is-live} section-@article.trail.sectionName.toLowerCase"
-        itemscope itemtype="@article.metadata.schemaType" role="main">
+        itemscope itemtype="@if(amp){http://schema.org/NewsArticle} else {@article.metadata.schemaType}" role="main">
```
TBonnin: Can we have a flag so we don't forget about it?

jfsoul: I know it should only be temporary, but it might be nice to define the different schema in one place and reference it here. `ArticleSchemas.NewsArticle` or similar.

SiAdcock:
Thanks guys. I have:

- exposed the `ArticleSchemas` as a public singleton. This is the source of schemas for article elements and is accessible from `PageWithStoryPackage`
- Added a feature switch so we can toggle AMP Live Blogs' `itemtype` between `NewsArticle` and `LiveBlogPosting`

johnduffell:
great, as long as the switch has an expiry we know to go back and do something with it! 👍

SiAdcock:
Sure thing, I've set it a month from tomorrow. 

Interesting related fact: the temporary helper function that sets article type to NewsArticle in AMP articles to support Google limitations was originally implemented in February 2015

johnduffell:
oh dear, I didn't think amp was even invented until summer last year!  anyway we should probably chase up google about when reviews etc will be supported

johnduffell:
actually when it says Feb 2015 that's just a typo.  it was feb 2016 😆

SiAdcock:
Good shout, I'll drop them a message on Slack

PR #13918, opened 2016-08-08
Add rounded corners and updating animation to buttons

katebee:
## What does this change?

Affects the email preferences page (https://profile.theguardian.com/email-prefs)
- Add rounded corners to the buttons
- Replace the 'loading...' text with the three dot animated gif (used in the profile pages when loading)
- Better visual feedback to user when ajax request is waiting for response to update the button
- Improved appearance of the email prefs page

No
## Screenshots

Rounded corners:

picture 171

Animation replaces button content:

picture 172
## Request for comment

@paperboyo
@joelochlann

### DIFF
--- a/static/src/javascripts/projects/common/modules/identity/email-preferences.js
+++ b/static/src/javascripts/projects/common/modules/identity/email-preferences.js
@@ -15,8 +15,7 @@ define([
 ) {
     function reqwestEmailSubscriptionUpdate(buttonEl) {
         bean.on(buttonEl, 'click', function () {
-            buttonEl.disabled = true;
-            buttonEl.innerHTML = 'Loading...';
+            addUpdatingState(buttonEl);
             var formQueryString = generateFormQueryString(buttonEl);
             reqwest({
                 url: '/email-prefs',
@@ -30,6 +29,7 @@ define([
                     if (response && response.subscriptions && response.subscriptions.length) {
                         isSubscribed = true;
                     }
+                    $(buttonEl).removeClass('js-button-updating');
                     updateButton(buttonEl, isSubscribed);
                 }
             });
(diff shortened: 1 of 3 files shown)


static/src/stylesheets/module/identity/_identity.scss
```diff
@@ -366,6 +366,9 @@
 .email-subscription__button {
     @include fs-textSans(2);
     background: $guardian-brand;
+    width: $gs-baseline * 9;
+    height: $gs-gutter * 2;
+    border-radius: 62.5rem;
```
joelochlann: I think we have a `px-to-rem` plugin in frontend so you can use pixels in sass files

static/src/stylesheets/module/identity/_identity.scss
```diff
@@ -377,11 +380,21 @@
...
+.save__button {
+    float: none;
+}
 
+.js-button-updating {
+    background-color: $neutral-5 !important;
+    background-image: url(/assets/images/auto-update-activity.gif);
+    background-repeat: no-repeat;
+    background-position: center;
+    background-size: 2.5rem;
```
joelochlann: think these `rems` should be `px`

static/src/javascripts/projects/common/modules/identity/email-preferences.js
```diff
@@ -30,6 +29,7 @@ define([
                     if (response && response.subscriptions && response.subscriptions.length) {
                         isSubscribed = true;
                     }
+                    $(buttonEl).removeClass('js-button-updating');
```
joelochlann: On membership we had a convention that state classes began with `.is-`, e.g. `.is-updating`. The `.js-` classes were to provide hooks for JS behaviour, so in this case `if ($(buttonEl).hasClass('save__button'))` is where you would use a `.js-` class to distinguish from classes used for styling. The idea being that you should be able to change a non `.js-` class without worrying that it would break some behaviour.

joelochlann:
💎  looks beautiful! Just a few minor stylistic points

static/src/stylesheets/module/identity/_identity.scss
joelochlann: (or maybe 100% for this, if that works)

paperboyo:
A beauty! Thank you!

joelochlann:
👍

PR #13922, opened 2016-08-09
Hosted By - remove old switches

lps88:
## What does this change?

Delete switches for campaigns that are already live.

cc @guardian/labs-beta


kelvin-chappell:
👍

PR #13926, opened 2016-08-09
AMP live blog byline contributor image

SiAdcock:
## What does this change?

Byline images in live blogs always use `` tags, even in AMP live blogs. Vanilla `` tags are not support in AMP articles, so the articles were not validating. This change introduces `<amp-img>` for byline contributor images in AMP live blogs.

Nope

### DIFF
--- a/common/app/views/fragments/amp/stylesheets/liveBlog.scala.html
+++ b/common/app/views/fragments/amp/stylesheets/liveBlog.scala.html
@@ -79,9 +79,6 @@
 
 .liveblog-block-byline__img {
     vertical-align: middle;
-}
-
-.liveblog-block-byline__img img {
     border-radius: 50%;
 }
 
--- a/common/app/views/fragments/amp/stylesheets/liveBlog.scala.html
+++ b/common/app/views/fragments/amp/stylesheets/liveBlog.scala.html
@@ -78,11 +78,11 @@
 }
 
 .liveblog-block-byline__img {
-    display: inline;
-    border-radius: 50%;
-    max-width: 2.25rem;
     vertical-align: middle;
-    margin-right: 0.25rem;
+}
+
+.liveblog-block-byline__img img {
+    border-radius: 50%;
 }
 
 .liveblog-block-byline__name {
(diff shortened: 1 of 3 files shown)


common/app/views/fragments/amp/stylesheets/liveBlog.scala.html
```diff
@@ -78,11 +78,11 @@
...
 
 .liveblog-block-byline__img {
-    display: inline;
-    border-radius: 50%;
-    max-width: 2.25rem;
     vertical-align: middle;
-    margin-right: 0.25rem;
+}
+
+.liveblog-block-byline__img img {
```
gtrufitt: Guessing this might be an amp thing, but should we be styling elements directly?

gtrufitt:
Minor comment but otherwise 👍

common/app/views/fragments/amp/stylesheets/liveBlog.scala.html
NataliaLKB: Ideally we shouldn't be... We should probably stick to styling `amp-img` instead.

SiAdcock: I could not find a way to style this element directly without nesting. We could always dispense with the rounded borders for the contributor image in AMP, I guess. The only classes we can hook into are generic classes provided by AMP.

NataliaLKB: Why not just style `liveblog-block-byline__img`? You can put the border-radius on that.

NataliaLKB: If you do that, you also lose a couple new lines of code 😉

SiAdcock:
Thanks all, I've applied the border radius to `amp-img` and it's all working. Diagnosis: the border radius wasn't being applied when emulating devices in Chrome's device emulator. You have to disable the emulator, reload the page, then re-enable the emulator to pick up the border radius 😖 

I've tested it on Chrome and Safari for iOS and it works fine in real life.

NataliaLKB:
Wow! Well done for figuring it out.

PR #13927, opened 2016-08-09
Hide "More on this story" on AMP live blog

SiAdcock:
## What does this change?

The "more on this story" section of the AMP live blog does not look too hot right now. It also prevents the AMP article from validating.

This change will prevent the section from appearing in the AMP live blog until we get it nicely styled for AMP.

Nope
## Screenshots

amp theguardian com-uk-news-live-2016-aug-08-southern-rail-strike-begins-live-updates iphone 5 2


katebee:
👍

PR #13928, opened 2016-08-09
use new facia client with special report

johnduffell:
use https://github.com/guardian/facia-scala-client/pull/170
@blongden73 @hoyla @TBonnin please can I get a +1?


TBonnin:
👍

PR #13932, opened 2016-08-09
Fix broken links to builds from radiator

kelvin-chappell:
Links from radiator to riff-raff builds seem to have got broken in #13864 

/cc @piuccio

### DIFF
--- a/static/src/javascripts/projects/admin/bootstraps/radiator.js
+++ b/static/src/javascripts/projects/admin/bootstraps/radiator.js
@@ -78,14 +78,14 @@ define([
                     var d  = latestDeployments[stage][deployment];
                     var nameAbbreviation = d.projectName.substr(7, 4); //start at 7 to drop 'dotcom: '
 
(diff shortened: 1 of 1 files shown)


piuccio:
funny

`target` is a `ul`
`link` is a `a`
`li` is a `li`

The generated markup is
```html
<ul>
   <a><li></li></a>
   <a><li></li></a>
</ul>
```

which feels wrong, but if that works 👍

gtrufitt:
It'll work but it's not valid, do with that what you will :)

kelvin-chappell:
I see what you mean.  I think the problem is the clickable area is empty and therefore unclickable.  I'll see if I can make it valid and working.

gtrufitt:
Not entirely sure what is happening, but could you just do ` link.innerHTML = nameAbbreviation + ' ' + d.build;` and append that to the li instead?

kelvin-chappell:
@gtrufitt I've taken your advice

PR #13933, opened 2016-08-09
added special badge to badges

blongden73:
Adds the special badge to the badges file. 

@johnduffell can you have a look.

### DIFF
--- a/common/app/model/Badges.scala
+++ b/common/app/model/Badges.scala
@@ -5,6 +5,7 @@ import conf.{Configuration, Static}
 import layout.FaciaContainer
 import java.security.MessageDigest
 import java.math.BigInteger
+import scala.util.control.NonFatal
 
 
 trait BaseBadge{
@@ -30,7 +31,7 @@ case class SpecialBadge(hashedTag: String, classModifier: Option[String] = None)
 
         Option(new BigInteger(1, digest.digest()).toString(16))
       } catch {
-        case _: Throwable => None
+        case NonFatal(_) => None
       }
     }
 }
--- a/common/app/model/Badges.scala
+++ b/common/app/model/Badges.scala
@@ -3,10 +3,40 @@ package model
 import conf.switches.Switches
 import conf.{Configuration, Static}
 import layout.FaciaContainer
+import java.security.MessageDigest
+import java.math.BigInteger
 
-case class Badge(seriesTag: String, imageUrl: String, classModifier: Option[String] = None)
+
+trait BaseBadge{
+  def imageUrl(id: String): String
+  def classModifier: Option[String]
+  def isBadgeTag(tag: String):Boolean
+}
+
+case class Badge(seriesTag: String, staticImageUrl: String, classModifier: Option[String] = None) extends BaseBadge {
+  def isBadgeTag(tag: String) = seriesTag == tag
+  def imageUrl(id: String) = staticImageUrl
+}
+case class SpecialBadge(hashedTag: String, classModifier: Option[String] = None) extends BaseBadge {
+  def isBadgeTag(tag: String) = md5(salt + tag).contains(hashedTag)
+  def imageUrl(id: String) = s"https://assets.guim.co.uk/special/$id/special-badge.svg"
+
+  private val salt = "a-public-salt3W#ywHav!p+?r+W2$E6="
+    private val digest = MessageDigest.getInstance("MD5")
+
+    private def md5(input: String): Option[String] = {
+      try {
+        digest.update(input.getBytes(), 0, input.length)
+
+        Option(new BigInteger(1, digest.digest()).toString(16))
+      } catch {
+        case _: Throwable => None
+      }
+    }
+}
 
 object Badges {
+
   private val euSvg = Static("images/badges/eu-ref.svg")
 
   val usElection = Badge("us-news/us-elections-2016", Static("images/badges/us-election.png"), Some("us-election"))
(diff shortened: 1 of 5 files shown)


common/app/model/Badges.scala
```diff
@@ -3,10 +3,40 @@ package model
...
+
+case class Badge(seriesTag: String, staticImageUrl: String, classModifier: Option[String] = None) extends BaseBadge {
+  def isBadgeTag(tag: String) = seriesTag == tag
+  def imageUrl(id: String) = staticImageUrl
+}
+case class SpecialBadge(hashedTag: String, classModifier: Option[String] = None) extends BaseBadge {
+  def isBadgeTag(tag: String) = md5(salt + tag).contains(hashedTag)
+  def imageUrl(id: String) = s"https://assets.guim.co.uk/special/$id/special-badge.svg"
+
+  private val salt = "a-public-salt3W#ywHav!p+?r+W2$E6="
```
TBonnin: Should this be a secret property?

common/app/model/Badges.scala
```diff
@@ -3,10 +3,40 @@ package model
...
+
+trait BaseBadge{
+  def imageUrl(id: String): String
+  def classModifier: Option[String]
+  def isBadgeTag(tag: String):Boolean
+}
+
+case class Badge(seriesTag: String, staticImageUrl: String, classModifier: Option[String] = None) extends BaseBadge {
+  def isBadgeTag(tag: String) = seriesTag == tag
+  def imageUrl(id: String) = staticImageUrl
```
TBonnin: `id` is not used. Should we make it a Optional then with default value set to `None`

common/app/model/Badges.scala
```diff
@@ -3,10 +3,40 @@ package model
...
+  private val salt = "a-public-salt3W#ywHav!p+?r+W2$E6="
+    private val digest = MessageDigest.getInstance("MD5")
+
+    private def md5(input: String): Option[String] = {
+      try {
+        digest.update(input.getBytes(), 0, input.length)
+
+        Option(new BigInteger(1, digest.digest()).toString(16))
+      } catch {
+        case _: Throwable => None
```
TBonnin: `NonFatal` instead of `Throwable`?

common/app/model/Badges.scala
johnduffell: no, it's just a salt

common/app/model/Badges.scala
```diff
@@ -3,10 +3,40 @@ package model
...
+  def isBadgeTag(tag: String):Boolean
+}
+
+case class Badge(seriesTag: String, staticImageUrl: String, classModifier: Option[String] = None) extends BaseBadge {
+  def isBadgeTag(tag: String) = seriesTag == tag
+  def imageUrl(id: String) = staticImageUrl
+}
+case class SpecialBadge(hashedTag: String, classModifier: Option[String] = None) extends BaseBadge {
+  def isBadgeTag(tag: String) = md5(salt + tag).contains(hashedTag)
+  def imageUrl(id: String) = s"https://assets.guim.co.uk/special/$id/special-badge.svg"
```
TBonnin: Should you check if the `tag` is matching the `hashedTag`?

common/app/model/Badges.scala
johnduffell: although admittedly it should really be different for each tag id

common/app/model/Badges.scala
johnduffell: good point, we copied this from elsewhere

common/app/model/Badges.scala
johnduffell: it's always passed in, just not always used

common/app/model/Badges.scala
johnduffell: by the time we call imageUrl we have the Badge instance and we're happy it is the right one

common/app/model/Badges.scala
TBonnin: That's my point. Signature conveys the idea the url depends on this string which is not always the case

common/app/model/Badges.scala
```diff
@@ -20,11 +50,13 @@ object Badges {
...
   val rio2016 = Badge("sport/rio-2016", Static("images/badges/rio-2016.svg"))
 
-  val allBadges = Seq(usElection, ausElection, voicesOfAmerica, rio2016, euElection, euRealityCheck, euBriefing, euSparrow)
+  val special = SpecialBadge("4dae5700e6b6fdf66d1567769b41c1c2")
+
+  val allBadges = Seq(usElection, ausElection, voicesOfAmerica, special, rio2016, euElection, euRealityCheck, euBriefing, euSparrow)
 
-  def badgeFor(c: ContentType) = c.tags.tags.map(_.id).foldLeft(None: Option[Badge]) { (maybeBadge, tag) =>
-      maybeBadge orElse allBadges.find(b => b.seriesTag == tag)
+  def badgeFor(c: ContentType) = c.tags.tags.map(_.id).foldLeft(None: Option[(BaseBadge, String)]) { (maybeBadge, tag) =>
```
johnduffell: we can change this tomorrow to return a `Badge` instead of a `(BaseBadge,String)` by turning SpecialBadge into a Badge, thanks to @TBonnin for the suggestion

TBonnin:
Since you are in a rush, feel free to merge. 
Although, as discussed face to face, you might simplify the code by transforming the `SpecialBadge` into a `Badge` once you have verified it. Then you won't have to carry away the tag and you can simplify the `BaseBadge` trait as well

PR #13936, opened 2016-08-09
don't compare boolean with string

johnduffell:
so much for static typing... doesn't work with ==
@TBonnin @blongden73 

All badges are missing from fronts atm 😭

### DIFF
--- a/common/app/model/Badges.scala
+++ b/common/app/model/Badges.scala
@@ -59,5 +59,5 @@ object Badges {
       maybeBadge orElse allBadges.find(b => b.isBadgeTag(tag)).map(badge => (badge, tag))
   }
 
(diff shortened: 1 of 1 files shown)


common/app/model/Badges.scala
```diff
@@ -59,5 +59,5 @@ object Badges {
       maybeBadge orElse allBadges.find(b => b.isBadgeTag(tag)).map(badge => (badge, tag))
   }
 
-  def badgeFor(fc: FaciaContainer) = fc.href.flatMap(href => allBadges.find(badge => href == badge.isBadgeTag(href)))
```
TBonnin: How does this even compile?

TBonnin:
👍

PR #13937, opened 2016-08-09
Amp article validity unit tests

jfsoul:
## What does this change?

Adds unit tests that test validity of various AMP-supported article content.  These tests use `HtmlUnit` to fetch content before passing that to the [official amp command line validator](https://github.com/ampproject/amphtml/tree/master/validator/nodejs).

Would appreciate thoughts on:
- Enabling all these tests immediately, effectively blocking changes until amp validation is fixed (this is kind of the point 😝).  ~~One concern would be that the validator is updated and previously valid content fails these tests (again, kind of the point).~~
- ~~The fact that these tests make a network request to fetch the validator from cdn.ampproject.org (alternative is to manage our own version, but we would need to ensure it is _always_ up to date)~~. ✅ 
- Whether there is any other content we should be testing at this stage (liveblogs omitted for now). ✅ 

Should stop majority of breaking changes to amp content by flagging them during the development process.

Amp 😁 
## Request for comment

@TBonnin @NataliaLKB @SiAdcock @stephanfowler @johnduffell @gtrufitt @sndrs


stephanfowler:
> Whether there is any other content we should be testing at this stage

No; for now only articles are being amp'ified. Add liveblogs later, once they're seemingly valid and making it into Google's amp cache.

common/test/common/AmpValidityTest.scala
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  private def writeToProcess(str: String)(out: OutputStream): Unit = {
+    out.write(str.getBytes)
+    out.close()
+  }
+
+  // This is too simplistic, but all we need for now.
+  // It should be replaced if support for urls with existing params is needed.
+  private def formatAmpUrl(url: String): String = {
```
TBonnin: not much formatting here. What about `amplifyUrl(url: String)`?

SiAdcock:
Just to get the ball rolling, thanks @jfsoul for this awesome change! There's probably a bunch of articles that are failing AMP validation right now that we don't even know about.

I have two things I'd like to consider:

## Special article types

We need to make sure we validate special article types, such as [Heroic articles](https://www.theguardian.com/membership/2016/jul/20/labour-liverpool-clp-constituency), [Immersive articles](https://www.theguardian.com/environment/ng-interactive/2015/nov/26/the-mekong-river-stories-from-the-heart-of-the-climate-crisis-interactive) and Immersive Galleries. There are probably others.

## Validator updates

I wonder if it is correct that the validator is always up to date. Picture a scenario in which someone makes a change that does not affect AMP validation at all, but is blocked because the validator gets updated and some unrelated article structure is no longer valid. It doesn't feel right to me that all changes are blocked until someone fixes AMP validation.

As an alternative, we could (at least in the short term) manually update the validator on a regular basis, and as part of the change that updates the validator, fix any new validation issues.

In the longer term, we could implement a bot that automatically raises a PR whenever there is a change to the AMP validator.

johnduffell:
enable immediately - I'd leave them running a short time if you're not confident, then if it's working well ship it as a failing test
network calls - avoid in DEVINFRA mode (when the tests are running in TC) ideally you want to do what the tests do for capi and save the content in the data directory during DEV mode and load it from there in DEVINFRA mode.

common/test/package.scala
```diff
@@ -65,6 +66,7 @@ trait TestSettings {
 trait ConfiguredTestSuite extends ConfiguredServer with ConfiguredBrowser with ExecutionContexts {
   this: ConfiguredTestSuite with org.scalatest.Suite =>
 
+  lazy val webClient = new WebClient(BrowserVersion.CHROME)
```
johnduffell: all we want is the html, seems very heavy duty to fire up a chrome just to get that.  we could use WS.url? or even just call the controller directly?

TBonnin:
To follow up on @SiAdcock comment, I don't think it is a good idea to inline a test that can fail because of an external dependencies (ex: validation rules change). It would halt all changes before those rules are implemented and the build fixed.

Is there a way store the `validator.js` file locally and pass it into the validation step? That would be ideal. Like @johnduffell mentioned you could fetch the real file when in dev mode and store it locally while CI will use the stored version.

common/test/common/AmpValidityTest.scala
```diff
@@ -0,0 +1,45 @@
...
+
+    s"The AMP page at $url" should "pass an AMP validator" in getContentString(ampUrl) { content =>
+
+      val commandInputWriter: OutputStream => Unit = writeToProcess(content)
+
+      // The process fails when not using stdout/stderr, but these may prove useful for debugging anyway
+      val io = new ProcessIO(commandInputWriter, BasicIO.toStdOut, BasicIO.toStdErr)
+
+      // Pass the content to the command line tool (external process) via stdin ('-' option)
+      val process = "node_modules/amphtml-validator/index.sh -".run(io)
```
johnduffell: just wondering how long it takes to run the validator?  presumably subsequent invocations are quicker than the first, but maybe a batch validation is possible if it's too slow?

common/test/package.scala
jfsoul: I can take a look at calling the controller directly - I ended up using HtmlUnit partly because my initial attempts needed to execute javascript.

HtmlUnit is just simulating Chrome (which I believe does not much more than changes to user-agent and javascript execution), so this implementation is pretty close to WS.url I think.

common/test/common/AmpValidityTest.scala
jfsoul: Just ran the tests a couple of times for anecdotal evidence.  First invocation of the validator seems to take about 30ms, subsequent around 5ms.  Batching probably not worth the investigation effort at that speed?

johnduffell: for speed doesn't seem a big problem then

common/test/package.scala
jfsoul: There seems to be no discernible difference in calling the controller directly or using WS.url in terms of execution speed.

Tempted to stick with HtmlUnit as the code will be very slightly more portable to future amp pages (as you just need a URL).

jfsoul:
@SiAdcock in regards to special article types, are we supporting these on amp yet?

common/test/package.scala
johnduffell: if it's just a simulation that's fine, I just thought it was an actual chrome which would have been very heavy!  sounds fine to me...the only benefit of hitting the controller would be the static typing rather than using the routes, although depending if you're hoping to test the routes as well, maybe you want that! 😀

SiAdcock:
@jfsoul after some investigation, it appears Immersive and Heroic articles are not AMP-supported right now. As discussed, let's not include examples of these in the AMP validation tests for now.

jfsoul:
@SiAdcock @TBonnin @johnduffell good suggestions regarding network requests and auto-update of validator. Thanks!

In this case we won't be auto-updating the validator yet.  Going forwards we either have the option of additional full-blown integration tests with latest validator outside the development workflow, or the auto-PR method suggested previously.

I'm now using `HttpRecorder` to fetch the validator from `data/amp-validator` if it exists, which should avoid the network requests on TC.  There's also a bonus refactor because I noticed we're duplicating code for the most basic usage of `HttpRecorder` (a request using `wsClient`).  The `ContentApiHttpRecorder` is not included in that refactor, but I have moved it out into its own file (the diff is not great because it has been replaced by the similar and new `DefaultHttpRecorder`.

johnduffell:
great, looks like a good way to do it, thanks! 🎉 👍

common/test/common/AmpValidityTest.scala
```diff
@@ -24,14 +33,21 @@ trait AmpValidityTest extends FlatSpec with Matchers with ConfiguredTestSuite {
...
         process.exitValue() should be(0)
       }
     }
   }
 
+  private def fetchValidator(): String = {
+    recorder.load(VALIDATOR_URI) {
+      wsClient.url(VALIDATOR_URI).get()
+    }
+    recorder.fileLocation(VALIDATOR_URI)
```
TBonnin: I am not sure I understand why `fileLocation` is useful here. Does the `load` not returning the filepath (wrapped in a future)?

SiAdcock:
Nicely done, man 👍

common/test/common/AmpValidityTest.scala
```diff
@@ -20,27 +22,27 @@ trait AmpValidityTest extends FlatSpec with Matchers with ConfiguredTestSuite wi
   def testAmpPageValidity(url: String): Unit = {
     val ampUrl = ampifyUrl(url)
 
-    s"The AMP page at $url" should "pass an AMP validator" in getContentString(ampUrl) { content =>
+    val future: Future[File] = recorder.loadFile(validatorUri) {
```
TBonnin: `file` rather than `future`?

common/test/recorder/HttpRecorder.scala
```diff
@@ -16,13 +16,16 @@ import scala.concurrent.Future
...
 
 
-trait
-HttpRecorder[A] extends ExecutionContexts {
+trait HttpRecorder[A] extends ExecutionContexts {
 
   def baseDir: File
 
+  final def load(url: String, headers: Map[String, String] = Map.empty)(fetch: => Future[A]): Future[A] =
+    loadFile(url, headers)(fetch).map(toResponse _ compose contentFromFile _)
```
TBonnin: I am not sure what `toResponse _ compose contentFromFile _` bring over `file => toResponse(contentFromFile(file))`
I found the latter easier to read :)

common/test/recorder/HttpRecorder.scala
```diff
@@ -32,37 +35,38 @@ HttpRecorder[A] extends ExecutionContexts {
...
-    }.getOrElse {
-      val response = fetch
-      response.foreach(r => put(fileName, fromResponse(r)))
-      response
-    }
+    get(fileName)
+      .map(Future(_))
+      .getOrElse {
+        val response = fetch
+        response.map(r => put(fileName, fromResponse(r)))
```
TBonnin: I don't think your need a local val here
```
fetch.map{ r => 
    put(fileName, fromResponse(r))
}
```

TBonnin:
👍

jfsoul:
Failing test is expected (catching these is the idea), fix has been merged to master.

PR #13938, opened 2016-08-10
video icon now white

blongden73:
Get some white on that Icon. 

@akash1810 @johnduffell

### DIFF
--- a/static/src/stylesheets/base/_inline-svgs.scss
+++ b/static/src/stylesheets/base/_inline-svgs.scss
@@ -52,7 +52,7 @@
     }
 
     .tonal--tone-special-report & {
-        fill: $news-support-6;
+        fill: #ffffff;
     }
 }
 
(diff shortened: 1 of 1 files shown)


static/src/stylesheets/base/_inline-svgs.scss
```diff
@@ -52,7 +52,7 @@
     }
 
     .tonal--tone-special-report & {
-        fill: $news-support-6;
+        fill: #ffffff;
```
akash1810: could we use a variable? [`neutral-7`](https://github.com/guardian/frontend/blob/master/static/src/stylesheets/pasteup/palette/_src.scss#L15) perhaps? or a declare a new one?

akash1810:
may be better to use a variable, other than that 👍

akash1810:
👍

johnduffell:
thanks ben!

PR #13939, opened 2016-08-10
darker colour for twitter badge on special reports media pages

akash1810:
## What does this change?

better contrast of twitter handle button on video page of special reports

less eye strain

no
## Screenshots

before
screen shot 2016-08-10 at 11 10 03

after
screen shot 2016-08-10 at 11 10 12
## Request for comment

@blongden73 @mr-mr

### DIFF
--- a/static/src/stylesheets/module/content/tones/_tone-special-report.scss
+++ b/static/src/stylesheets/module/content/tones/_tone-special-report.scss
@@ -85,5 +85,16 @@
             background-color: mix($news-support-6, $neutral-1, 70%);
             border-color: mix($news-support-6, $neutral-1, 70%);
         }
+
+        .button--secondary {
+            @include button-colour(
+                $neutral-2,
+                $neutral-1
+            );
+
+            @include button-hover-colour(
+                darken($neutral-2, 10%)
+            );
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


rich-nguyen:
the styles are fine, but might want to check with @sammorrisdesign or @sndrs about whether pasteup changes are ok directly in this repo (i think they are)

👍

sammorrisdesign:
So you can make changes to pasteup but they won't be released in any way. For all intents and purposes the pasteup folder is 100% frontends now. This is an interim while we work on an npm (non-bower) way of releasing a mono-pasteup.

But, I'd say for this change it should go in: https://github.com/guardian/frontend/blob/8ce085b1aae89a7823b7633679bec1f54f47c0b2/static/src/stylesheets/module/content/tones/_tone-special-report.scss

The colour styles in buttons are meant to be more generic when we thought it would be used that way. I'm not even sure if the button styles in here are used on the site (could be wrong here though). Either way I think it's best to keep all the special report tonal stuff together in that file.

What say you @sndrs?

johnduffell:
this is getting old and alex isn't available at the moment, is this PR blocked by that decision, if not ship it and revisit?

akash1810:
shoot sorry for letting this hang around for so long - I'll move the change into the correct place now

akash1810:
Moved rules to `_tone-special-report.scss` now. Merging.

PR #13940, opened 2016-08-10
simplify special badge code

johnduffell:
this follows on from @TBonnin comments on the special badging yesterday, now we always return a Badge to the templates, and only use SpecialBadge when we are working out whether it even applies.

cc @blongden73

### DIFF
--- a/common/app/model/Badges.scala
+++ b/common/app/model/Badges.scala
@@ -59,11 +59,12 @@ object Badges {
 
   def badgeForTags(tags: Traversable[String]) = {
 
-    val badgesForTags = tags.flatMap {
-      case tagId => allBadges.flatMap {
-        case baseBadge => baseBadge.maybeThisBadge(tagId)
-      }
-    }
+    val badgesForTags =
+      for {
+        tagId <- tags
+        baseBadge <- allBadges
+        maybeBadge <- baseBadge.maybeThisBadge(tagId)
+      } yield maybeBadge
     badgesForTags.headOption
   }
 
--- a/common/app/model/Badges.scala
+++ b/common/app/model/Badges.scala
@@ -53,10 +53,20 @@ object Badges {
 
   val allBadges = Seq(usElection, ausElection, voicesOfAmerica, special, rio2016, euElection, euRealityCheck, euBriefing, euSparrow)
 
-  def badgeFor(c: ContentType) = c.tags.tags.map(_.id).foldLeft(None: Option[Badge]) { (maybeBadge, tag) =>
-      maybeBadge orElse allBadges.flatMap(b => b.maybeThisBadge(tag)).headOption
+  def badgeFor(c: ContentType) = {
+    badgeForTags(c.tags.tags.map(_.id))
   }
 
-  def badgeFor(fc: FaciaContainer) = fc.href.flatMap(href => allBadges.flatMap(b => b.maybeThisBadge(href)).headOption)
+  def badgeForTags(tags: Traversable[String]) = {
+
+    val badgesForTags = tags.flatMap {
+      case tagId => allBadges.flatMap {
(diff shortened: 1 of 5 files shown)


common/app/model/Badges.scala
```diff
@@ -8,19 +8,17 @@ import java.math.BigInteger
...
+  def maybeThisBadge(tag: String): Option[Badge]
 }
 
-case class Badge(seriesTag: String, staticImageUrl: String, classModifier: Option[String] = None) extends BaseBadge {
-  def isBadgeTag(tag: String) = seriesTag == tag
-  def imageUrl(id: String) = staticImageUrl
+case class Badge(seriesTag: String, imageUrl: String, classModifier: Option[String] = None) extends BaseBadge {
+  def maybeThisBadge(tag: String) = if (seriesTag == tag) Some(this) else None
 }
 case class SpecialBadge(hashedTag: String, classModifier: Option[String] = None) extends BaseBadge {
```
TBonnin: do you need the `classModifier` param in `SpecialBadge`?

common/app/model/Badges.scala
```diff
@@ -8,19 +8,17 @@ import java.math.BigInteger
...
-trait BaseBadge{
-  def imageUrl(id: String): String
-  def classModifier: Option[String]
-  def isBadgeTag(tag: String):Boolean
+trait BaseBadge {
+  def maybeThisBadge(tag: String): Option[Badge]
 }
 
-case class Badge(seriesTag: String, staticImageUrl: String, classModifier: Option[String] = None) extends BaseBadge {
-  def isBadgeTag(tag: String) = seriesTag == tag
```
TBonnin: why have you not keep this `isBadgeForTag()` method? 
I found having a class method return `this` kind of weird. Also it mean using `flatmap` below when `find` is imho a better way to convey what the code is doing.

common/app/model/Badges.scala
johnduffell: good point, removed it

common/app/model/Badges.scala
```diff
@@ -53,10 +53,20 @@ object Badges {
...
-  def badgeFor(c: ContentType) = c.tags.tags.map(_.id).foldLeft(None: Option[Badge]) { (maybeBadge, tag) =>
-      maybeBadge orElse allBadges.flatMap(b => b.maybeThisBadge(tag)).headOption
+  def badgeFor(c: ContentType) = {
+    badgeForTags(c.tags.tags.map(_.id))
   }
 
-  def badgeFor(fc: FaciaContainer) = fc.href.flatMap(href => allBadges.flatMap(b => b.maybeThisBadge(href)).headOption)
+  def badgeForTags(tags: Traversable[String]) = {
+
+    val badgesForTags = tags.flatMap {
```
TBonnin: You didn't like the for comprehension? :wink:

common/app/model/Badges.scala
```diff
@@ -53,10 +53,20 @@ object Badges {
...
+      case tagId => allBadges.flatMap {
+        case baseBadge => baseBadge.maybeThisBadge(tagId)
```
TBonnin: Do you need cases here?
Also in most part of the code map/flatmap are formatted like below:
```
something.map { wrapped =>
    doingSomethingWith(wrapped)
    ...
}
```
Can we do the same here?

common/app/model/Badges.scala
johnduffell: oh I thought you didn't, and suggested I used flatmap =P

common/app/model/Badges.scala
johnduffell: true, I can revert to for comprehension anyway

TBonnin:
👍

PR #13941, opened 2016-08-10
New header menu styling tweaks

zeftilldeath:
## What does this change?

Sign in has been moved to sit alongside search
There is a font size increase on larger devices
I've tidied up the icon structure, and alignment to include less 🎩 🐇 ⭐  numbers

localhost-3000-sport-2016-aug-10-lilly-king-olympian-rio-2016 iphone 6

@stephanfowler @NataliaLKB


static/src/stylesheets/layout/_ab-new-header-test-variant.scss
```diff
@@ -361,34 +361,16 @@ $gutter-large: 55px;
         margin-right: $gutter-large + ($veggie-burger-medium / 2);
     }
 
+    @include mq($mobile-medium) {
+        font-size: 23px;
```
blongden73: 23px?

blongden73: If you look at making the type reduce/increase by percentage across the board then we can have a rule for this happening elsewhere! Refer to our conversation in person! Otherwise 👍

PR #13942, opened 2016-08-10
add minute JS to fronts (0% A/B test)

gidsg:
## What does this change?

Extends minute JS to run on fronts

Enable us to test effects on fronts with 3rd party vendor on production. (As part of opt-in 0% A/B test only)

No
## Screenshots

N/A
## Request for comment

CC @akash1810

### DIFF
--- a/static/src/javascripts/projects/common/modules/experiments/tests/minute.js
+++ b/static/src/javascripts/projects/common/modules/experiments/tests/minute.js
@@ -19,7 +19,7 @@ define([
         this.dataLinkNames = '';
         this.idealOutcome = '';
         this.canRun = function() {
-            return config.page.contentType === 'Article' && qwery('[data-component="main video"]').length > 0;
+            return (config.page.contentType === 'Section' || config.page.contentType === 'Network Front') || config.page.contentType === 'Article'  && qwery('[data-component="main video"]').length > 0;
         };
 
         this.variants = [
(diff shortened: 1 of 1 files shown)


static/src/javascripts/projects/common/modules/experiments/tests/minute.js
```diff
@@ -19,7 +19,7 @@ define([
         this.dataLinkNames = '';
         this.idealOutcome = '';
         this.canRun = function() {
-            return config.page.contentType === 'Article' && qwery('[data-component="main video"]').length > 0;
+            return (config.page.contentType === 'Section' || config.page.contentType === 'Network Front') || config.page.contentType === 'Article'  && qwery('[data-component="main video"]').length > 0;
```
akash1810: why `Section`?

gidsg: I thought because it is a type of Front e.g. https://www.theguardian.com/video

akash1810: would `config.page.isFront === true` be simpler?

gidsg: Ah cool, I didn't see that option, will amend.

akash1810:
👍

PR #13943, opened 2016-08-10
Change free become a member to become a supporter

AWare:
## What does this change?
- The "free" become a supporter link now says "become a member".
- The membership url is now set to the supporter page
- The members data api url doesn't attempt to read the membership URL from config and now has its own entry.

value: We get rid of the word "free" from every page on the guardian.
measure: 💰
## Screenshots

image


SiAdcock:
I support this change 👍

PR #9569, opened 2015-06-22
Fix SFL delete all button

sndrs:
- automatically reverts from the confirmation state to the standard state after 2 seconds
- updates the confirmation message to be a simple 'sure?' so it fits at all breakpoints

screen shot 2015-06-22 at 15 35 10

(if this works well, it seems like it would a useful behaviour to add to pasteup, so that, for example, by the addition of a simple `confirm` attribute to a button, you could co-opt this behaviour. but that doesn't need doing for this…

PR #13948, opened 2016-08-10
Give the Email Prefs page an 'Unsubscribe From All' button

katebee:
## What does this change?

Adds an `Unsubscribe From All` button to the [email preferences](https://profile.theguardian.com/email-prefs) page.

This button mimics the behaviour of the `Remove All` button on the [Save For Later](https://profile.theguardian.com/saved-for-later) page, but has the 2 second timeout removed (to ensure there are fewer accessibility issues).

Clicking the `Unsubscribe From All` button changes it to a confirmation button (to check the user really wants to do this). A second click within 2 seconds will unsubscribe the user from every email listed on the email preferences page.
- Shows our users we care, by minimising the number of clicks required to unsubscribe from all 29 of our current email offerings on this page
- Alternative to the current, nuclear option, of `unsubscribe from all` in ExactTarget that permanently removes the user from all emails 🔥 💀

No
## Screenshots

**Appearance of the button:**
picture 179

**After first click it requests confirmation:**
picture 180


static/src/stylesheets/module/identity/_identity.scss
```diff
@@ -386,15 +386,39 @@
...
         background-position: center;
         background-size: 40px;
     }
+
+    &.email-unsubscribe {
+        float: none;
+        color: $guardian-brand;
+        background-color: #ffffff;
+        border: 2px solid $guardian-brand;
+        min-width: $gs-baseline * 14;
```
SiAdcock: Better to user a `gs-span()` for widths than a large multiple `$gs-baseline` (likewise in `.save__button` on line 421)

SiAdcock:
Out of interest, why only give the user 2 seconds to hit the confirmation? Are there accessibility implications with this?

katebee:
@SiAdcock @ScottPainterGNM 
I based the 2 second wait on the user flow of the 'Saved For Later' page (which also uses a 2 second wait). I'm not hugely attached to the number and am happy to increment it!

SiAdcock:
Sure thing, I see it originally came from #9569, although I can't see any justification for it in the PR description. I'm not sold on the idea of a timeout at all, I can see it being frustrating for people who have reduced mobility. Happy to listen to opinions on why we should keep it, but my vote for now is to lose it.

static/src/javascripts/projects/common/modules/identity/email-preferences.js
```diff
@@ -29,22 +29,51 @@ define([
...
+                updateButton(buttonEl, true);
+            } else {
+                confirmUnsubscriptionFromAll(buttonEl);
+            }
+        });
+    }
+
+    function getAllSubscribedButtons() {
+        var buttons = [];
+        $.forEachElement('.js-subscription-button', function (buttonEl) {
```
joelochlann: I think you can just do `$('[value^="unsubscribe"'])` to get these elements

static/src/javascripts/projects/common/modules/identity/email-preferences.js
```diff
@@ -29,22 +29,51 @@ define([
...
-    function encodeFormData(csrfToken, buttonVal, htmlPreference) {
-        return 'csrfToken=' + encodeURIComponent(csrfToken) + '&' +
-        'addEmailSubscription=' + encodeURIComponent(buttonVal) + '&' +
-        'htmlPreference=' + encodeURIComponent(htmlPreference);
+    function unsubscribeFromAll(buttonEl) {
+        bean.on(buttonEl, 'click', function () {
+            if ($(buttonEl).hasClass('js-confirm-unsubscribe')) {
+                addUpdatingState(buttonEl);
+                resetUnsubscribeFromAll(buttonEl);
+                var subscribedButtons = getAllSubscribedButtons();
```
joelochlann: might be easier to work with a bonzo wrapped set here, so you can just do `$buttons.each(function() { ... } )` instead of `for` loops

static/src/javascripts/projects/common/modules/identity/email-preferences.js
```diff
@@ -29,22 +29,51 @@ define([
...
+                for (var i = 0; i < subscribedButtons.length; i++) {
+                    subscribedButtons[i].click();
```
joelochlann: as we discussed, I'm a bit wary of firing off loads of requests when we could actually batch them into one. I'll tweak the backend so it can accept multiple add/remove requests in one

gtrufitt:
omg, does this mean you can get rid of that crazy exacttarget unsubscribe from all (and never be able to re-subscribe to an email ever again) button? Whoopy de do.

joelochlann:
👍  Looks good. Let's 🚢 🚢 🚢

PR #13949, opened 2016-08-10
Change contributions coin header test so that coin icon only shows for signed in users

jranks123:
## What does this change?

This changes the nature of the contributions header test. Previously the contribute CTA was replacing the "become a supporter" button, so only non-signed in users would see it. Now it places itself before the subscriptions button, if and only if the user is signed in. 

It decides if a user is signed in by checking for the existence of the "GU_U" cookie. While this is a very low level way of determining if a user is logged in, it is adequate for the purpose of this small test. 

We can compare the conversion rate to the previous test with non signed in users and see what the difference in behaviour is. 

No
## Screenshots

screenshot at aug 10 17-26-35

### DIFF
--- a/static/src/javascripts/projects/common/modules/experiments/tests/contributions-header.js
+++ b/static/src/javascripts/projects/common/modules/experiments/tests/contributions-header.js
@@ -6,6 +6,7 @@ define([
     'common/views/svgs',
     'common/utils/fastdom-promise',
     'common/utils/mediator',
+    'common/utils/cookies',
     'text!common/views/contributions-header.html'
 ], function (bean,
              qwery,
@@ -14,6 +15,7 @@ define([
              svgs,
              fastdom,
              mediator,
+             cookies,
              contributionsHeader
 ) {
     return function () {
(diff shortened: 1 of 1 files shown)


jranks123:
@rtyley

static/src/javascripts/projects/common/modules/experiments/tests/contributions-header.js
```diff
@@ -43,8 +43,8 @@ define([
             }));
 
             return fastdom.write(function () {
-                var a = $('.brand-bar__item--register');
-                a.replaceWith($newThing);
+                var a = $('.brand-bar__item--profile.popup-container.has-popup.js-profile-nav.is-signed-in');
```
rtyley: So does there end up being any kind of race-condition, given that `is-signed-in` is [dynamically added]( https://github.com/guardian/frontend/blob/220c46ad/common/app/templates/inlineJS/nonBlocking/showUserName.scala.js#L15)?

static/src/javascripts/projects/common/modules/experiments/tests/contributions-header.js
```diff
@@ -32,7 +34,8 @@ define([
         this.idealOutcome = '';
         this.canRun = function () {
             var pageObj = window.guardian.config.page;
-            return !(pageObj.isLiveBlog || pageObj.isAdvertisementFeature) && pageObj.edition === 'UK';
+            var userCookie = cookies.get('GU_U');
```
rtyley: Oh! I didn't mean you should actually read the cookie directly- `window.guardian.config.user` probably _is_ the right thing to read, I just thought it was worth checking whether it was definitely completed before the AB javascript was run.

jranks123: Ok, thanks for all the feedback! For now, I'm happy with just checking for the existence of this cookie in the interest of getting it out today

johnduffell:
looks good to my eyes! 👍 🚤

PR #13950, opened 2016-08-10
Use AdminFilters in dev-build

desbo:
This fixes a bug where users aren't prompted to login on /admin in `dev-build`. 
@TBonnin


TBonnin:
Sorry that's an oversight on my part. 

Thanks for fixing it
👍

johnduffell:
now users have to log into everything 😢 and there's no clear  wayto know an admin url

PR #13958, opened 2016-08-11
Revert "Use AdminFilters in dev-build"

johnduffell:
Reverts guardian/frontend#13950

Unfortunately there changes made it so you have to log in to everything.  Since there's no clear url group not going to admin, the list of urls exempt wouldn't be possible.

Reverting so we can come up with another solution - in the mean time we can run admin directly to change switches.

@desbo @TBonnin


TBonnin:
👍

PR #13960, opened 2016-08-11
Remove use of Option.get in SwitchboardController

desbo:
Fixes a bug where you can't save switches in `dev-build` due to there being no google oauth.
@TBonnin @johnduffell


johnduffell:
great! thanks 👍 🚀

PR #13957, opened 2016-08-11
Make Membership engagement banner into an A/B test with Contributions

rtyley:
## What does this change?

50% of the audience will be shown a Contributions message on the engagement banner rather than the Membership message.

A fair bit of refactoring of the existing Membership engagement banner functionality was necessary, to make it fit into the AB testing framework - which in particualar has a synchronous `canRun` method, which is problematic given that the banner is not supposed to display at all if the asynchronous check for an adblocker comes back positive.
## What is the value of this and can you measure success?

We want to get a baseline the conversion rate of the Contributions message ("If you use it, if you like it, why not pay for it? It's only fair.") as used in the engagement banner, and also compare it against the Membership message.

https://trello.com/c/HlD5H68N/325-engagement-banner-a-b-test
## Does this affect other platforms - Amp, Apps, etc?

Nope.
## Screenshots

With the Contributions banner (`#ab-MembershipAndContributionsEngagementBanner=contributions`)
image

With the Membership banner (`#ab-MembershipAndContributionsEngagementBanner=membership`)
image
## Request for comment

@desbo

https://trello.com/c/HlD5H68N/325-engagement-banner-a-b-test

PR #13961, opened 2016-08-11
Kill unused deps to avoid silent RequireJS death on circular deps

rtyley:
## What does this change?

RequireJS is not good at telling you when you have a circular dependency (see https://github.com/requirejs/requirejs/issues/202 )

These pre-existing and unused deps were what were causing everything to (silently) die in https://github.com/guardian/frontend/pull/13957

I tried to use https://github.com/requirejs/xrayquire to find the circular deps, but I couldn't get it running before @desbo figured it out!


SiAdcock:
👍

PR #13962, opened 2016-08-11
Less panic shedding during medium latency conditions

johnduffell:
So due to a logic error, we weren't specifically allowing the minimum number of connections when we were only serving a percentage of responses.  Although this should still allow through enough to recover, it was not intentional, as it can happen during startup.

This fixes it so we will always allow the 4 connections.

@TBonnin have a look!


TBonnin:
lgtm. 
How did you find out and also how did that affects the apps so far?

TBonnin:
Good catch. @jfsoul noticed it after a CODE deploy I think. And because of caching we haven't noticed in PROD if it ever happen.
Thank you

jfsoul:
Good catch, symptoms definitely sound familiar 💯

PR #13966, opened 2016-08-11
Make it so snaps don't need images

sammorrisdesign-zz:
## What does this change?

**STORY TIME**
So I was testing some snaps in `CODE` where the fallback was a front and not an article. In `PROD` this works because they're pressed into the page. Turns out it's because the javascript snaps require an image for some unknown reason. So snaps now don't require images.


piuccio:
👍  doesn't look like the js code dealing with snaps expects an image

PR #13967, opened 2016-08-12
remove singapore f1 article image

lps88:
## What does this change?

remove an article image!

cc @kelvin-chappell


kelvin-chappell:
👍

PR #13968, opened 2016-08-12
Continue Remind Me email test for another week

joelochlann:
As of yesterday results from #13870 are:

```
 variant                    | unique browsers | which saw component     | clicks | conversion rate
                            | in test         | for more than a second  |        | 
----------------------------+-----------------+-------------------------+--------+----------------
 weekly-series-description  |            6169 |  1755                   | 2      | 0.11%
 one-off-series-description |            6204 |  1617                   | 2      | 0.12%
 weekly-coming-up           |            6001 |  1673                   | 3      | 0.18%
 one-off-coming-up          |            6082 |  1570                   | 3      | 0.19%
```

According to the data team we need 10,000 to have seen each variant for this to be a statistically significant demand test. Since this is five times as many saw the component in the past week, we're extending for another week and exposing to 100% of users. Given that this component only appears at the bottom of Blind Date and Experience articles (rather than on, say, a news article or a front) we don't think this is too intrusive.

### DIFF
--- a/common/app/conf/switches/ABTestSwitches.scala
+++ b/common/app/conf/switches/ABTestSwitches.scala
@@ -80,7 +80,7 @@ trait ABTestSwitches {
     "Sign up for an email to remind you about the next item in a series",
     owners = Seq(Owner.withGithub("joelochlann")),
     safeState = On,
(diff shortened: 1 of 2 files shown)


SiAdcock:
👍

PR #13969, opened 2016-08-12
Show "More on this story" in AMP articles 

SiAdcock:
## What does this change?

The "More on this Story" section is not appearing on AMP articles. This is because the article checks whether `Content.hasStoryPackage` is flagged as true, when it should be checking whether `RelatedContent.hasStoryPackage` is flagged as true. This reflects how the "More on this Story" section is [populated in non-AMP articles](https://github.com/guardian/frontend/blob/master/common/app/views/fragments/storyPackagePlaceholder.scala.html#L13).

We expect `Content.hasStoryPackage` to be set by CAPI, but as far as I can tell, this flag is not being passed from CAPI. This might be a bug their end, I'll chase them on it.

No
## Screenshots

localhost-3000-uk-news-live-2016-aug-08-southern-rail-strike-begins-live-updates-amp iphone 5

### DIFF
--- a/common/app/views/mainAMP.scala.html
+++ b/common/app/views/mainAMP.scala.html
@@ -77,7 +77,7 @@
                 </div>
 
             @if(content.metadata.sectionId != "childrens-books-site") {
-                @if(content.hasStoryPackage) {
+                @if(related.hasStoryPackage) {
                     @fragments.amp.storyPackageAmp(related)
                 }
                 @if(content.tags.series.nonEmpty) {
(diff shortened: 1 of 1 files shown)


SiAdcock:
Confirmed on CAPI, the [`hasStoryPackage`](https://github.com/guardian/content-api/blob/d414c13b50205ba6ff97b206db4389c53a3d8f63/porter/src/main/scala/com.gu.contentapi.porter/integration/flexible/FlexibleTransformer.scala#L177) field on `Content` is a legacy field and is not used.

jfsoul:
I had difficulty the other day finding working examples of this, but didn't get round to investigating! Nice work! 👍

TBonnin:
👍

PR #13971, opened 2016-08-12
Docker dev environment

TBonnin:
## What does this change?

First attempt at setting up a dev environment with Docker.

This patch allows to run a single app in the container.
See [setup instructions](https://github.com/guardian/frontend/blob/tbonnin-docker-dev-env/docs/01-start-here/01-installation-steps.md#docker).

Of course there is a lot of things that can be improved. (Ex: having a private image repo so each dev don't need to build their own image, speed, etc...)

This patch also remove Vagrant 🔪 🃏 

I recommend to start from a fresh repo. Especially if you are running Mac (node modules like node-sass compiled for your Mac won't play well within the container environment) 

Feedback welcome!!!! 🙏 

A single dev environment, no need for a setup.sh that must take into account all the subtleties that comes from different OS/systems.
Eventually all developers run the same environment in dev 
## Request for comment

@jamespamplin @desbo @johnduffell @jfsoul @dominickendrick  @guardian/dotcom-platform


gtrufitt:
🐳

Dockerfile-dev
```diff
@@ -0,0 +1,37 @@
...
+RUN rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+
+VOLUME /frontend
+
+WORKDIR /frontend
+
+EXPOSE 9000
+
+# Make prompt prettier
+RUN echo 'export PS1="[🐳 :\h]\w# "' >> /root/.bashrc
```
desbo: 😀

Dockerfile-dev
```diff
@@ -0,0 +1,37 @@
...
+  apt-get install sbt
+
+# Install Node.js
+RUN \
+    curl -sL https://deb.nodesource.com/setup_6.x | bash - && \
+    apt-get update && \
+    apt-get install -y --no-install-recommends build-essential nodejs
+
+# Install grunt
+RUN npm i -g grunt-cli
```
jamespamplin: Wish we could remove global grunt...

Dockerfile-dev
```diff
@@ -0,0 +1,37 @@
...
+
+# Create config files
+RUN mkdir -p /etc/gu/ && echo 'STAGE=DEV' > /etc/gu/install_vars
+
+# Cleaning up after apt-get
+RUN apt-get clean
+RUN rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
```
jamespamplin: You should do this as a single command after each (or the last) apt install, to take advantage of docker's cached layers. ie:

```
RUN \
    curl -sL https://deb.nodesource.com/setup_6.x | bash - && \
    apt-get update && \
    apt-get install -y --no-install-recommends build-essential nodejs && \
    apt-get clean && \
    rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
```

Or better yet, is there a no-cache style option for apt-get like in other package managers?

Dockerfile-dev
```diff
@@ -0,0 +1,37 @@
...
+VOLUME /frontend
+
+WORKDIR /frontend
+
+EXPOSE 9000
+
+# Make prompt prettier
+RUN echo 'export PS1="[🐳 :\h]\w# "' >> /root/.bashrc
+
+RUN echo ">>> Build completed: Start your dev environment with 'docker-compose run --service-ports dev'"
```
jamespamplin: Normally this wouldn't be very useful in a container, but I can see why you've added it here.

Also, might be worth adding `--rm` as an option. It will remove the container once it exits, not doing so will create a new container every time you run `docker-compose run`

docs/01-start-here/01-installation-steps.md
```diff
@@ -23,27 +23,49 @@ Follow [this link](https://www.google.co.uk) and enter the relevant search strin
...
+    - `git clone git@github.com:guardian/frontend.git && cd frontend`
+3. Get [janus credentials](https://janus.gutools.co.uk/) for frontend and CMS fronts 
+4. Get properties file
+    - `aws s3 cp --profile frontend s3://aws-frontend-store/template-frontend.properties "~/.gu/frontend.properties"`
+5. Build the container image
+    - `docker-compose build dev`
+6. Run the container (dev service)
+    - `docker-compose run --service-ports dev`
+7. Build and run the app within the container
+    - `make build compile && ./sbt`
```
jamespamplin: I'd still prefer running in non-attached mode with `docker-compose up`. Perhaps we could look into creating an entrypoint which negates the need for interactive mode in most cases, running `make watch` and `sbt "project dev-build" ~run` in one container.

TBonnin: Yep. Just discussed with @desbo about that. I am looking into it

desbo: I can see the appeal of non-interactive run commands in general, but here I think an interactive sbt prompt is way better than a forced `dev-build run`. We need to be able to use the sbt prompt for switching projects, running tests, etc.

jamespamplin:
Works for me locally, good stuff!

docs/01-start-here/01-installation-steps.md
SiAdcock: Off-topic, but should this line be `make install compile && ./sbt`?

TBonnin: Indeed. Thanks @SiAdcock

TBonnin:
@jamespamplin can you run `time dd if=/dev/zero of=test.dat bs=1024 count=100000` in the container and report the result.

It is quite slow on my computer and I suspect there is an issue with the mounted volume on Mac.

TBonnin:
On my computer:

Non-mounted volume
```
[🐳 :1434708d3240]/tmp# time dd if=/dev/zero of=test.dat bs=1024 count=100000
100000+0 records in
100000+0 records out
102400000 bytes (102 MB) copied, 0.186296 s, 550 MB/s
real	0m0.188s
user	0m0.010s
sys	0m0.170s
```
Mounted volume
```
[🐳 :1434708d3240]/frontend# cd /frontend/
[🐳 :1434708d3240]/frontend# time dd if=/dev/zero of=test.dat bs=1024 count=100000
100000+0 records in
100000+0 records out
102400000 bytes (102 MB) copied, 22.2325 s, 4.6 MB/s
real	0m22.248s
user	0m0.070s
sys	0m1.390s
```

😢 

Looks like it is a known issue though: https://forums.docker.com/t/file-access-in-mounted-volumes-extremely-slow-cpu-bound/8076

jamespamplin:
@TBonnin yeah I was hoping they'd fixed this, I thought it was a problem with the virtualbox volume mapping mechanism. In the past when I manually cranked virtualbox images, volume sharing was fastest over NFS shares. Does the new docker for mac still use virtualbox underneath? If you can use NFS for volume mounts it might be faster.

My values:
```
[🐳 :8f282124fba4]/frontend# time dd if=/dev/zero of=test.dat bs=1024 count=100000
100000+0 records in
100000+0 records out
102400000 bytes (102 MB) copied, 0.146257 s, 700 MB/s

real	0m0.163s
user	0m0.010s
sys	0m0.138s
[🐳 :8f282124fba4]/frontend# cd /tmp/
[🐳 :8f282124fba4]/tmp# time dd if=/dev/zero of=test.dat bs=1024 count=100000
100000+0 records in
100000+0 records out
102400000 bytes (102 MB) copied, 0.120671 s, 849 MB/s

real	0m0.122s
user	0m0.007s
sys	0m0.115s
```

TBonnin:
The problem is explained in details here btw https://forums.docker.com/t/file-access-in-mounted-volumes-extremely-slow-cpu-bound/8076/157 :(

TBonnin:
@jamespamplin I slightly change the approach:
- compile and run the project once during image building to populate caches. It makes the building of the image super long but building doesn't happen very often
- uploaded the image to ECR so devs don't have to build it themselves

let me know what you think

Dockerfile-dev
```diff
@@ -0,0 +1,43 @@
...
+
+VOLUME /frontend
+
+WORKDIR /frontend
+
+# Build app once to populate node and sbt cache
+ADD . /source
+RUN cd /source; make install compile
+RUN  cd /source; ./sbt compile 
+RUN rm -rf /source
```
jamespamplin: Make these a single `RUN` command, so that docker can cache it as a single layer.

Dockerfile-dev
jamespamplin: You could move this to `dev-help.sh` now too :whale:

Dockerfile-dev
```diff
@@ -0,0 +1,43 @@
...
+
+EXPOSE 9000
```
jamespamplin: Move the `EXPOSE` command above any `ADD` or `COPY` commands. Whenever the project sources change, it invalidates each layer under the associated `ADD` or `COPY`, `EXPOSE` is just config and can come earlier in the build

jamespamplin:
:+1: great work :tada:

PR #13973, opened 2016-08-12
New AB test for hosted galleries

lps88:

We want to find out where is the most effective placement of the 'call to action' text and button. Success is getting then most clicks on the button.
### Testing galleries of varying length

We have 4 galleries, which (if you exclude the first and last slide) have between 3 and 7 possible placements for the CTA. Can we test the first, the middle and the last possible placements, even though those will be at different indices for each? Can we test placing it at 1st, 2nd and 3rd available positions, and ignore the rest of the images in the longer galleries? Do we need different tests for different length galleries? Answers on a postcard...
### Current test

At the moment, I've excluded the shorter galleries, so it only shows up on these two: 
https://www.theguardian.com/advertiser-content/visit-britain/activities
https://www.theguardian.com/advertiser-content/visit-britain/coast
and places them on either 3rd, 4th, 5th or 6th image.
## Screenshots

image

### DIFF
--- a/static/src/javascripts/projects/common/modules/experiments/tests/hosted-gallery-cta.js
+++ b/static/src/javascripts/projects/common/modules/experiments/tests/hosted-gallery-cta.js
@@ -1,0 +1,84 @@
+define([
+    'bean',
+    'fastdom',
+    'qwery',
+    'common/utils/$',
+    'common/utils/config',
+    'common/utils/detect',
+    'lodash/collections/contains'
+], function (
+    bean,
+    fastdom,
+    qwery,
+    $,
+    config
+) {
+    return function () {
+        this.id = 'HostedGalleryCta';
+        this.start = '2016-08-12';
+        this.expiry = '2016-09-16';
+        this.author = 'Lydia Shepherd';
+        this.description = 'Tests which image in the gallery the call to action link is most effective on';
+        this.audience = 1;
+        this.audienceOffset = 0;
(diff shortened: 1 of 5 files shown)


dominickendrick:
👍

static/src/javascripts/projects/common/modules/experiments/tests/hosted-gallery-cta.js
```diff
@@ -0,0 +1,84 @@
...
+        };
+
+        this.variants = [
+            {
+                id: '3',
+                test: function () {
+                    config.page.ctaIndex = 2;
+                },
+                success: function (complete) {
+                    var ctaButton = qwery('.hosted-gallery__cta-link');
```
Calanthe: This block could be extracted into another function. Currently it's repeated in all four variants.

static/src/javascripts/projects/common/modules/experiments/tests/hosted-gallery-cta.js
```diff
@@ -0,0 +1,84 @@
...
+        this.dataLinkNames = 'Find more inspiration';
+        this.idealOutcome = 'People will click on the call to action link more often.';
+
+        this.canRun = function () {
+            return config.page.contentType === 'Gallery' && config.page.tones === 'Hosted' && config.page.images.length > 6;
+        };
+
+        this.variants = [
+            {
+                id: '3',
```
Calanthe: Where is variant 1 and 2? ;)

PR #13974, opened 2016-08-12
Refactor metadata of HostedVideoPages

kelvin-chappell:
This is enabling work to make it easier to start populating `HostedVideoPages` with data from capi.  It shouldn't have any effect on anything that's already in production.

/cc @lps88 @guardian/labs-beta


lps88:
👍

PR #13977, opened 2016-08-12
Add heatmap param to list allowed for local dev

philwills:
## What does this change?

Adds 'heatmap' to the list of locally allowed params

You can test the heatmap via this method locally

No
## Screenshots

image


dominickendrick:
👍

PR #13978, opened 2016-08-12
Remove Contributions from the Engagement banner

rtyley:
_Do not merge until Tuesday midday, when 4 days of Contributions data has been collected_
## What does this change?

Partially undoes the work of #13957 - ie this makes Membership messages the only messages shown on the engagement-banner. We're intending to use the Engagement banner for Contributions again in the near future, so I haven't converted all the code back from being an AB test, just made Membership 100%.

@desbo


desbo:
👍

PR #13979, opened 2016-08-12
Editionalising the new header menus with JS

NataliaLKB:
## What does this change?

One of the goals of the new header is to remove the need to cache four versions of every page. To do this we need to move editionalisation from happening on the server, to happening in JS.

So far the new header has been designed with that in mind. This PR adds in the correct editionalised lists.

Will be tested soon!

Nope
## Screenshots

UK:
image

US:
image

### DIFF
--- a/common/app/views/fragments/nav/newHeaderMenu.scala.html
+++ b/common/app/views/fragments/nav/newHeaderMenu.scala.html
@@ -1,8 +1,9 @@
 @()(implicit request: RequestHeader)
 
-@import common.{ NewNavigation, TopLevelSection, LinkTo }
+@import common.{ NewNavigation, LinkTo, Edition }
+@import views.support.RenderClasses
 
-@sectionList(topLevelSection: TopLevelSection) = {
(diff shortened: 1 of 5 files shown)


common/app/views/fragments/nav/newHeaderMenu.scala.html
```diff
@@ -1,22 +1,33 @@
...
-                @NewNavigation.sectionItems(topLevelSection).map { sectionItem =>
-                    <li class = "navigation-group__item" >
-                        <a href="@LinkTo { sectionItem.url }">@sectionItem.name</a>
-                    </li>
-                }
-            </ul>
+
+            @Edition.all.map { edition =>
+                <ul class="@RenderClasses(Map(
+                    "is-hidden" -> (edition.id.toLowerCase != "uk") // Our default edition is UK
```
piuccio: how about the `hidden` attribute?

NataliaLKB: Good call @piuccio !  I have updated the PR now.

piuccio:
looks good to me 👍

PR #13980, opened 2016-08-15
Fix: wait for ads to render before measuring their geometry

regiskuckaertz:
Spacefinder has an option where it can wait for ads in the root element to _load_ before performing its calculations. Thing is, an ad loaded may not have geometry yet and this is what happens:

picture 3

Here, a rich link is inserted after the inline merchandising slot has been loaded—but it shouldn't. Now, this is solved if instead we want for ads to be _rendered_:

picture 4

cc @guardian/commercial-dev @gtrufitt


rich-nguyen:
👍

gtrufitt:
🅱️ 🅾️ 🅾️ Ⓜ️

PR #13981, opened 2016-08-15
extend AB test until M Butler comes back

alexduf:
## What does this change?

extends the AB test duration by a week, so we don't block frontend. @markjamesbutler will have to remove that test when he comes back.

no build breaks because of that test
## Request for comment

@TBonnin @johnduffell


dominickendrick:
👍

johnduffell:
thanks! 👍

PR #13982, opened 2016-08-15
fixing the z-index problem for the edition picker when there is a thr…

NataliaLKB:
## What does this change?

When there is a thrasher, the new header edition picker becomes hidden. This PR make the z-index higher.

Makes sure people can always move onto the international edition

Nope
## Screenshots

Before:
image

After:
image
## Request for comment

@zeftilldeath


zeftilldeath:
✋ 😄 👍

PR #13983, opened 2016-08-15
Better error message when fetching S3 object fails

TBonnin:
## What does this change?

Add an error message when fetching S3 Object and hint about the most likely cause: expired AWS credentials.

imho it is better than throwing the exception and filling the sbt console with lots of stack traces. The error message tells WHAT is happening and HOW to fix it.

No
## Request for comment

@guardian/dotcom-platform @johnduffell @jfsoul @dominickendrick

### DIFF
--- a/common/app/services/S3.scala
+++ b/common/app/services/S3.scala
@@ -52,6 +52,13 @@ trait S3 extends Logging {
         log.warn("not found at %s - %s" format(bucket, key))
         None
       }
+      case e: AmazonS3Exception => {
+        val errorMsg = s"Unable to fetch S3 object (key: $key)"
+        val hintMsg =   "Hint: your AWS credentials might be missing or expired. You can fetch new ones using Janus."
+        log.error(errorMsg, e)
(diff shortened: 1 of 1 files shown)


NataliaLKB:
👍

common/app/services/S3.scala
```diff
@@ -52,6 +52,13 @@ trait S3 extends Logging {
         log.warn("not found at %s - %s" format(bucket, key))
         None
       }
+      case e: AmazonS3Exception => {
+        val errorMsg = s"Impossible to fetch S3 object (key: $key)"
```
jfsoul: Impossible -> Unable? It is possible if you remember to use Janus!

johnduffell:
I like the println, nice touch to make it obvious.  not sure if it's worth checking Play.isProd first 😆 (or whatever the new way is) and throwing if so 
👍

TBonnin:
@johnduffel I though about that but I don't see a huge difference between returning None or throwing it. Whatever use the None response will fail anyway.

PR #13984, opened 2016-08-15
Hide broken "more on this story" images

SiAdcock:
## What does this change?

For some articles, the thumbnail path for story package images is not being set. AMP articles on which this happens fail validation, causing them not to be returns in Google search results.

This initial fix will prevent an `<amp-img>` from being constructed without a `src` attribute in the story package template. I'll address the root cause in a separate PR.

Nope
## Screenshots

No images... but at least it validates

picture 85

### DIFF
--- a/common/app/views/fragments/amp/storyPackageAmp.scala.html
+++ b/common/app/views/fragments/amp/storyPackageAmp.scala.html
@@ -18,12 +18,15 @@
 </div>
 
 @onward(content: CuratedContent) = {
-    <div class="fc-item @TrailCssClasses.toneClass(content)">
-        @if(content.properties.maybeContent.flatMap(_.trail.thumbnailPath)) {
-            <div class="fc-item__image-container">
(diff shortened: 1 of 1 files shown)


common/app/views/fragments/amp/storyPackageAmp.scala.html
```diff
@@ -19,9 +19,11 @@
 
 @onward(content: CuratedContent) = {
     <div class="fc-item @TrailCssClasses.toneClass(content)">
-        <div class="fc-item__image-container">
-            <amp-img src="@content.properties.maybeContent.flatMap(_.trail.thumbnailPath)" layout="fixed" width=126 height=75></amp-img>
-        </div>
+        @if(content.properties.maybeContent.flatMap(_.trail.thumbnailPath)) {
```
jfsoul: shall we put `content.properties.maybeContent.flatMap(_.trail.thumbnailPath)` in a variable?

jfsoul:
works for me 👍

jfsoul:
👍

PR #13985, opened 2016-08-15
Use 88x89 size on advertisement feature fronts

regiskuckaertz:
Sets up everything so that ad feature fronts use 88x89 instead of 88x87 in the high-relevance merchandising slot.

cc @guardian/commercial-dev


janua:
LGTM :+1:

PR #13987, opened 2016-08-15
Small fixes to the email lists

katebee:
## What does this change?

Changes the email preferences page (https://profile.theguardian.com/email-prefs)
- Remove the duplicate listings for **Guardian Opinion** and **The Fiver**
- **The Fiver** now only appears in the 'Sport' section (no longer also in 'News')
- **Guardian Opinion** now only appears in the 'Comment' section (no longer also in 'News')
- Move **The Guardian Today - AUS** next to the other listings of **The Guardian Today**
- Gives some sanity to the order of emails in the big, hardcoded list, so it reflects the page
- Sanity! 🌷 
- Stop treating AUS emails as a addendum, instead list them next to their UK equivalent

Nope
## Screenshots

Moving **The Guardian Today - AUS** to be with the other **Guardian Today** emails:

picture 183

### DIFF
--- a/identity/app/model/EmailSubscriptions.scala
+++ b/identity/app/model/EmailSubscriptions.scala
@@ -16,29 +16,30 @@ case class EmailSubscription(
 object EmailSubscription {
   def apply(emailSubscription: EmailSubscription) = emailSubscription
 }
+
 object EmailSubscriptions {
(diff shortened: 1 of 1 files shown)


janua:
:+1:

PR #13988, opened 2016-08-15
Support capi-driven hosted video pages

kelvin-chappell:
Hosted video pages previously had to be hardcoded in the frontend codebase.
This change uses capi to get the content.

The existing hardcoded pages should continue to work.

There are some outstanding todos in the code.

/cc @Calanthe @lps88 @guardian/labs-beta


johnduffell:
awesome, well done to get rid of hard coded stuff!

common/app/common/commercial/hosted/HostedVideoPage.scala
```diff
@@ -82,6 +82,7 @@ object HostedVideoPage extends Logging {
           ),
           // todo: standardise css so that only colour varies
           cssClass = "renault",
```
Calanthe: we probably won't need this value

kelvin-chappell: No, just leaving it there so that the page works before colour has been isolated in the css.

PR #13990, opened 2016-08-15
Reset scroll position on re-ordering of menu (new-header)

NataliaLKB:
## What does this change?

This sets the scroll position to '0' when you open the menu. 

Much less annoying for users! They actually get to see what they want to see (if they click sports, they can now see the sports menu first)

Nope!
## Screenshots

Without the reset:
scroll-position-bug

Now:
scroll-position-fix
## Request for comment

@stephanfowler

### DIFF
--- a/static/src/javascripts/projects/common/modules/navigation/editionalise-menu.js
+++ b/static/src/javascripts/projects/common/modules/navigation/editionalise-menu.js
@@ -36,5 +36,5 @@ define([
         }
     }
 
-    return editionaliseMenu();
+    return editionaliseMenu;
 });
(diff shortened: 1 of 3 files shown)


static/src/javascripts/projects/common/modules/navigation/editionalise-menu.js
```diff
@@ -36,5 +36,5 @@ define([
         }
     }
 
-    return editionaliseMenu();
```
NataliaLKB: I realise this doesn't make sense to invoke here, its already being invoked in another file 😅

johnduffell:
👍

PR #13965, opened 2016-08-11
Recommended for you test

joelochlann:
Tests demand for a Recommended For You section on fronts. A dummy section is inserted above Opinion and the user can choose to "turn it on" or not. It's a 404 test: clicking "Yes Please" will give a 404 message and clicking "No Thanks" will remove the section.
### Desktop

picture 267
### Tablet

picture 266
### Mobile

picture 265
### 404 message after clicking "Yes Please" button

picture 268

@katebee

PR #13991, opened 2016-08-15
Expose 'Recommended For You' to 10% of audience

joelochlann:
Expose #13965 to 10% of our audience (5% per variant).

Also:
- Don't redisplay the section after the user has given feedback
- Track attention time of this section with Ophan
- Show just on network front
- Change the CTA for the "Yes" response to "Turn On" for the user history variant (because "Get Started" implies there's something that the user will have to give some manual input, which is not the case with the user history variant)

@katebee


katebee:
👍

PR #13992, opened 2016-08-15
Standalone comments - part 1

piuccio:
## What does this change?

As a first step, move the comment count of discussion module into a standalone React / preact application.

Discussion frontend is [developed here](https://github.com/guardian/discussion-frontend)

This first implementation doesn't include any CSS and is behind a 0 audience multivariant test.

The variants are:
- control: keep using the old bonzo code
- react: load react ~48kb minified gzipped
- preact: load preact + compat ~9kb minified gzipped

React / preact are loaded asynchronously by `enhanced`, so they won't impact load of other features.

`dev-build` and `article` poll the discussion frontend assets map every 5 seconds and inject that value in the curl config, all this is behind a switch in case it goes banana. The A/B test is clever enough to not run if the curl config is not populated correctly (the switch is off or something else didn't work).

For the curious minds, preact doesn't work nicely in dev mode because the `proptypes` module has a UMD definition that breaks our AMD require. Works nice once minified because proptypes are stripped off. I'm working on a fix but it's not a blocker.

At this point success means all versions work just fine and it's possible to deploy comments independently of frontend.
Users won't see any difference unless they opt in in a test

No, it's all loaded in JS, no html change
## Screenshots

screen shot 2016-08-15 at 16 06 20
## Request for comment

@guardian/guardian-frontend @nicl 

## TODO
- [x] Update platform
- [x] fastly in front of discussion frontend
- [x] readme on discussion frontend

### DIFF
--- a/common/app/assets/DiscussionAssets.scala
+++ b/common/app/assets/DiscussionAssets.scala
@@ -85,7 +85,7 @@ class DiscussionExternalAssetsLifecycle (config: GuardianConfiguration, wsClient
       "DiscussionRefreshAssetsMap",
       config.discussion.frontendAssetsMapRefreshInterval
     ) {
-      refresh()
+      refresh().map(_ => ())
     }
   }
 
--- a/common/app/assets/DiscussionAssets.scala
+++ b/common/app/assets/DiscussionAssets.scala
@@ -1,0 +1,99 @@
+package assets
+
+import java.net.URI
+
+import app.LifecycleComponent
+import common.{AkkaAgent, ExecutionContexts, GuardianConfiguration, JobScheduler, Logging}
+import conf.switches.Switches
+import play.api.libs.ws.{WSClient, WSResponse}
+
+import scala.concurrent.Future
+import scala.concurrent.duration._
+import scala.util.{Failure, Success, Try}
+
+/**
+  * External assets are not hosted by frontend but referenced in an assets map
+  *
+  * This class pull the assets map regularly and translates a generic name into a full URL
+  *
+  * The map is a JSON object looking like
+  * {
+  *   "name": "js/name.min.hash.js"
+  * }
+  * The path in the object value is relative to the assets map
+  */
+class DiscussionExternalAssetsLifecycle (config: GuardianConfiguration, wsClient: WSClient, jobs: JobScheduler)
+  extends LifecycleComponent with ExecutionContexts with Logging {
+
+  def refresh(): Future[Unit] = {
+    config.discussion.frontendAssetsMap match {
+      case Some(url) if Switches.DiscussionFetchExternalAssets.isSwitchedOn =>
+        fetchAssetsMap(url, wsClient).map(
+          parseResponse(_, url).map(DiscussionAssetsMap.alter(_, URI.create(url)))
+        )
+      case Some(_) =>
+        log.info("Fetching discussion external assets is switched off")
+        Future.successful(None)
+      case None =>
+        log.warn("External discussion frontend endpoint is not configured, you might want to update `discussion.frontend.assetsMap`")
+        Future.successful(None)
+    }
+  }
+
+  private def fetchAssetsMap(url: String, wsClient: WSClient): Future[WSResponse] = {
+    wsClient.url(url)
+      .withHeaders("User-Agent" -> "GU-ExternalAssets")
+      .withRequestTimeout(4.seconds.toMillis)
+      .get()
+  }
+
+  private def parseResponse(response: WSResponse, url: String): Option[Map[String, String]] = {
+    response.status match {
+      case 200 => asMap(response, url)
+      case statusCode =>
+        log.error(s"Cannot download discussion assets map from $url with status code $statusCode")
+        None
+    }
+  }
+
+  private def asMap(response: WSResponse, url: String): Option[Map[String, String]] = {
+    Try(response.json.as[Map[String, String]]) match {
+      case Success(assetsMap) => Some(assetsMap)
+      case Failure(error) =>
+        log.error(s"Invalid JSON assets map from $url", error)
+        None
+    }
+  }
+
+
+  def start(): Unit = {
+    descheduleJobs()
+    scheduleJobs()
+    refresh()
+  }
+
+  private def scheduleJobs(): Unit = {
+    jobs.scheduleEvery(
+      "DiscussionRefreshAssetsMap",
+      config.discussion.frontendAssetsMapRefreshInterval
+    ) {
+      refresh()
+    }
+  }
+
+  private def descheduleJobs(): Unit = {
+    jobs.deschedule("DiscussionRefreshAssetsMap")
+  }
+}
+
+object DiscussionAssetsMap {
+  private lazy val agent = AkkaAgent[Map[String, String]](Map.empty)
+
+  def alter (map: Map[String, String], baseURI: URI): Future[Map[String, String]] = {
+    agent.alter(map.mapValues(value => baseURI.resolve(value).toString))
+  }
+
+  def getURL (assetName: String): String = {
+    agent.get().getOrElse(assetName, "")
+  }
+}
--- a/common/app/assets/DiscussionAssets.scala
+++ b/common/app/assets/DiscussionAssets.scala
@@ -25,18 +25,26 @@ import scala.util.{Failure, Success, Try}
 class DiscussionExternalAssetsLifecycle (config: GuardianConfiguration, wsClient: WSClient, jobs: JobScheduler)
   extends LifecycleComponent with ExecutionContexts with Logging {
 
-  def refresh(): Future[Unit] = {
+  def refresh(): Future[Map[String, String]] = {
     config.discussion.frontendAssetsMap match {
       case Some(url) if Switches.DiscussionFetchExternalAssets.isSwitchedOn =>
         fetchAssetsMap(url, wsClient).map(
-          parseResponse(_, url).map(DiscussionAssetsMap.alter(_, URI.create(url)))
-        )
+          parseResponse(_, url) match {
+            case Some(parsed) => DiscussionAssetsMap.alter(parsed, URI.create(url))
+            case None =>
+              val errMsg = "Impossible to parse discussion assets map"
+              log.warn(errMsg)
+              Future.failed(new RuntimeException(errMsg))
+          }
+        ).flatMap(identity)
       case Some(_) =>
-        log.info("Fetching discussion external assets is switched off")
-        Future.successful(None)
+        val errMsg = "Fetching discussion external assets is switched off"
+        log.info(errMsg)
+        Future.failed(new RuntimeException(errMsg))
       case None =>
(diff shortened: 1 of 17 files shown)


common/app/assets/DiscussionAssets.scala
```diff
@@ -0,0 +1,99 @@
...
+  def start(): Unit = {
+    descheduleJobs()
+    scheduleJobs()
+    refresh()
+  }
+
+  private def scheduleJobs(): Unit = {
+    jobs.scheduleEveryNSeconds(
+      "DiscussionRefreshAssetsMap",
+      config.discussion.frontendAssetsMapRefreshInterval.toSeconds.toInt
```
johnduffell: can the toseconds.toint bit go in the config part?

piuccio: sure

johnduffell: or actually can the schedule every n seconds take an interval directly

static/src/javascripts/projects/common/modules/experiments/tests/discussion-external-frontend.js
```diff
@@ -0,0 +1,41 @@
+define([], function () {
+    return function () {
+
+        this.id = 'DiscussionExternalFrontend';
```
johnduffell: if you do increase the audience, make sure to rename the test id

johnduffell:
looks great, exciting stuff.  👍 for the server side stuff, client side looks ok but I might have missed something

common/app/assets/DiscussionAssets.scala
piuccio: that's what I wanted to do initially, but I felt incapable of changing that code 😁 

what should be the interface then, maybe `scheduleEvery(name: String, interval: Interval)`?

johnduffell: haha, I have changed it before and nothing broke ;) maybe don't go crazy at this stage, I'd generally use another PR for that kind of thing, otherwise people will not know what the primary point of this PR is...

johnduffell: but yeah that signature would be fine if you want to go for it! =)

common/app/assets/DiscussionAssets.scala
```diff
@@ -0,0 +1,99 @@
...
+    wsClient.url(url)
+      .withHeaders("User-Agent" -> "GU-ExternalAssets")
+      .withRequestTimeout(4.seconds.toMillis)
+      .get()
+  }
+
+  private def parseResponse(response: WSResponse, url: String): Option[Map[String, String]] = {
+    response.status match {
+      case 200 => asMap(response, url)
+      case _ =>
```
janua: Log the response code

common/app/assets/DiscussionAssets.scala
```diff
@@ -0,0 +1,99 @@
...
+        log.error(s"Cannot download discussion assets map from $url")
+        None
+    }
+  }
+
+  private def asMap(response: WSResponse, url: String): Option[Map[String, String]] = {
+    Try(response.json.as[Map[String, String]]) match {
+      case Success(assetsMap) => Some(assetsMap)
+      case Failure(_) =>
+        log.error(s"Invalid JSON assets map from $url")
```
johnduffell: also as per @janua  comment above, log the exception from the failure too

static/src/javascripts/projects/common/modules/discussion/loader.js
```diff
@@ -38,7 +40,9 @@ define([
     DiscussionApi,
     CommentBox,
     Comments,
+    frontend,
```
janua: Can we call this `discussion-frontend` or something else? Didn't initially realise it was the `discussion-frontend` module.

common/app/assets/DiscussionAssets.scala
janua: I actually ignored this because `response.json` will throw an exception saying the content type is not in the correct format; so you already know what the problem is going to be. Come to think of it, `.as` will also throw an exception too, so probably worth logging out the `Failure(_)`!

piuccio:
guys thanks for the feedback, I'll merge later on

common/app/assets/DiscussionAssets.scala
```diff
@@ -0,0 +1,99 @@
...
+  * }
+  * The path in the object value is relative to the assets map
+  */
+class DiscussionExternalAssetsLifecycle (config: GuardianConfiguration, wsClient: WSClient, jobs: JobScheduler)
+  extends LifecycleComponent with ExecutionContexts with Logging {
+
+  def refresh(): Future[Unit] = {
+    config.discussion.frontendAssetsMap match {
+      case Some(url) if Switches.DiscussionFetchExternalAssets.isSwitchedOn =>
+        fetchAssetsMap(url, wsClient).map(
```
rich-nguyen: This has huge implications. We will no longer be able to understand which version of code we have live in prod anymore. Riff Raff's deployment history would only indicate the server version, not the true client-side asset version. We need to be really sure that we want that.

piuccio: Yup, that's the whole point. I want to be able to deploy discussion independently of frontend.

Think of discussion as an interactive / embded / atom / whatever with its own lifecycle and needs

johnduffell: I think that's the idea of this PR, that discussion is decoupled from and independent of frontend with a fairly defined API.  I'm assuming discussion-new will be deployed with riffraff also, so hopefully we would have enough information to know what was deployed at a given time.
Personally i feel like frontend is very big and would benefit from being split down.  However as you say, with this change it makes it harder to make changes across all parts of the code base and deploy together, but I think the interface between frontend and discussion is small enough to make sense.  Maybe there's another way to do it that keeps the atomic deployment though (red/green canary stack perhaps?)

johnduffell: oh yeah, what @piuccio said 😊

rich-nguyen:
I think this is the beginning of a decoupled frontend, whereby developers can deploy some assets independently of the server. Whilst that has benefits, we should make sure we really think it through, because there are probably many subtle assumptions we have made that rely on lock-step server-and-asset deployment. The most obvious example is a PR which changes both server-side html and client-side js/css can no longer be deployed atomically.

From the proposal document, it seems like the rationale for decoupling is to allow the discussion team to move faster, and independently. This therefore seems like an approach which is to isolate a new repo to the parts that can be wholly owned by that team. The upside is better team throughput. The downside is that this reduces the ease of sharing, adds complexity at the point where the two codebases bridge together (this PR), and complicates the task of managing hard (javascript modules) and soft (html container markup) dependencies.

Are there any alternatives we have considered? Are there fundamental difficulties within Frontend that we should be fixing to benefit the wider team, who feel it is too wieldy to develop on?

common/app/assets/DiscussionAssets.scala
```diff
@@ -0,0 +1,99 @@
...
+  * The map is a JSON object looking like
+  * {
+  *   "name": "js/name.min.hash.js"
+  * }
+  * The path in the object value is relative to the assets map
+  */
+class DiscussionExternalAssetsLifecycle (config: GuardianConfiguration, wsClient: WSClient, jobs: JobScheduler)
+  extends LifecycleComponent with ExecutionContexts with Logging {
+
+  def refresh(): Future[Unit] = {
```
TBonnin: Nit: I know the value returned by this method is never used and I guess that's why you forced it to Future[Unit] but if at some point we need to access the returned value, its type would be quite hard to work with (ie: `Future[Option[Future[Map[String, String]]]]`. Would it be possible to make it return a `Future[Map[String, String]]`?

piuccio:
Great summary

> The downside is that this reduces the ease of sharing

I'm not sure I see this point. As a matter of facts, the new discussion-frontend can easily be loaded by mobile apps in a web view. We can more quickly provide the same features to 3 platforms

> adds complexity at the point where the two codebases bridge together

Absolutely right, but I'd argue the complexity is small. The contract is that the html includes a `<div>` and calls a JS function passing the element and a bunch of other information like discussion id, identity and maybe others.

There's no shared HTML / CSS / JS. Discussion frontend bundles all its dependencies and comes with its own style. At least in the final solution, the transition is painful, I'm well aware.

> Are there fundamental difficulties within Frontend that we should be fixing

Yes, and we are fixing them, but it'll take way longer than splitting discussion. I'm also quite confident that removing discussion from frontend will make fixing those problem easier.

> Are there any alternatives we have considered?

The only alternative to splitting is not doing it. In other words add a new build system inside the repo which will make testing and compiling assets harder. You'll have to synchronise two ways of building assets unless the alternative is doing nothing at all and continue using dead libraries and outdated codebase.

johnduffell:
>deploy some assets independently of the server

perhaps this is a slight difference in concept, AIUI this is a little more than just moving the assets, it's moving the whole job of knowing what discussion is to a separate place.  So frontend repo shouldn't have any concept of discussion on the server or client side, other than it needs to insert a placeholder and pull in a library to do something with the placeholder.

piuccio:
I've fixed the return type of `refresh` and configured fastly to serve statics with @jamespamplin 

Instead of modifying platform, I've put the properties inside this repo, they are not secrets, in fact they're public paths.

@rich-nguyen I forgot to mention it in this PR but `assets.json`is versioned. When the contract between frontend and discussion changes we can bump the version and be sure that breaking changes are deployed correctly.
I've also added the RiffRaff build id in `assets.json`, if needed I can add an endpoint somewhere that returns this value. This might help understand what's the current cached version. Or maybe simply `log.info` the build id on every refresh

gtrufitt:
I wonder whether we can allay some of the concerns of multiple repos by using something like https://lernajs.io/ to keep our Frontend '[monorepo](https://github.com/babel/babel/blob/master/doc/design/monorepo.md)' but still allow us to modularise. 

We would probably look at doing similar with pasteup - @sndrs and @sammorrisdesign have been looking at moving it to an NPM module *managed* from within Frontend I believe - so similar usecase. 

May not have legs but worth a look @piuccio?

cc/ @rich-nguyen

regiskuckaertz:
What about Git submodules? @piuccio

nicl:
Just to add my two cents to this, although I pretty much agree with @piuccio  :

* decoupling is definitely the aim here

This would be massive for my team. Having to grapple with Frontend really slows us down. Of course, we can't and shouldn't avoid this completely, but the majority of development and testing would be much easier for the team.

* we also cut out the Next Gen API middle-layer (this currently generates the HTML)

This service is an extra point of failure (and we've seen this quite often recently), costs money, and confused responsibility - as it is not really owned by or understood by the Discussion team.

* using the same code for web and apps is also a potential benefit, although there is more discussion to be had here with Mobile

* separate repos vs mono

I think this is largely orthogonal to the above, but a separate repo is easier for new people working on commenting to get started with, and also ensures no cross-deps, etc. There's debate about whether Fabio's work should include all CSS/JS itself or share it - from a decoupling point of view being self-contained feels better, but I know there are technical/perf challenges here.

Issue #13569, opened 2016-07-12
Ad overlaying header


akash1810:
# ISSUE
Ad overlaying header.

## Steps to Reproduce
- Visit URL provided (see below)
- Wait for page to finish loading (with lightboxed image)
- Press esc to close lightbox
- Scroll to top of page
- Witness ad displaying over Guardian header (see Actual screenshot below)

## Actual Results (include screenshots)
![screen shot 2016-07-12 at 13 17 27](https://cloud.githubusercontent.com/assets/836140/16766600/f1b4f38a-4832-11e6-9308-29be4afe49f4.jpeg)

## Expected Results (include screenshots)
![screen shot 2016-07-12 at 13 17 49](https://cloud.githubusercontent.com/assets/836140/16766607/fb818d24-4832-11e6-8eea-cac8083a51d1.jpeg)

## URL
May require a few page refreshes...

https://www.theguardian.com/world/2016/jul/12/several-killed-as-two-trains-collide-in-southern-italy-puglia#img-1

## OS and Browser details

- Device: macbook
- Browser (inc version): Chrome 51.0.2704.103 (64-bit)
- OS: Yosemite

*Please add relevant labels*

JonNorman:
This is because the `top-banner-ad-container` element is set to `display: none` when the `body` element has the `has-overlay` class, which, when loading a page initially with a lightbox, it does. This in turn means that when the top-banner slot (which contains the ad) is measured, the height is 0 and so no extra space is added to the element.

Upon closing the lightbox, the overlay is removed, the `top-banner-ad-container` becomes visible again but there is no space for it to sit in.

Is it necessary to set `display: none` on the `top-banner-ad-container' when loading the page straight to an image fragment?

akash1810:
Good to know this logic.

> Is it necessary to set `display: none` on the `top-banner-ad-container` when loading the page straight to an image fragment?

As the lightbox isn't transparent and assuming the ad doesn't set a high `z-index`, I would say its not necessary.

PR #13995, opened 2016-08-15
Prevent top banner ad overlaying header

annebyrne:
## What does this change?

This stops the ad overlay on the header when exiting from a Lightbox. 
See: https://github.com/guardian/frontend/issues/13569
## Request for comment

@regiskuckaertz


regiskuckaertz:
yeah! 👍

PR #13998, opened 2016-08-15
Fix amp-youtube embed validation

jfsoul:
## What does this change?

Fixes amp validation on articles with youtube embeds.  We were still using the deprecated `video-id` attribute on `<amp-youtube>`.  This was previously only a warning from the validator, but with the latest release (12/8/16) it is now a failure.  Updating the validator version we store that would have detected this error.

Articles with youtube embeds pass validation.
## Request for comment

@SiAdcock


SiAdcock:
Very cool stuff, man 👍  💯

PR #14001, opened 2016-08-15
Fix font colour in immersive template

janua:
This fixes `.paidfor-box--header` and `.paidfor-box--text` when they appear in immersive gallery pages.

This situation also occurs on campaign minute articles and "heroic" articles; I'm not sure if I need to exclude these. 
@regiskuckaertz What do you think?

machines for living modernist american architecture in pictures art and design the guardian
witharticletext


regiskuckaertz:
Nearly there! The "paid content" and "About" labels should also have that colour. Otherwise, 👍

janua:
Now with updated `.paidfor-meta`!

darkmeta

PR #14002, opened 2016-08-15
Schedule jobs by duration

piuccio:
## What does this change?

Replace `scheduleEveryNSeconds`with `scheduleEvery` which takes a `Duration`. It's more versatile than having `scheduleEveryNMinutes` and `scheduleEveryNSeconds`

I don't feel like changing all `scheduleEveryNMinutes` now, maybe in a separate PR?

Kinda useless 😄 

R2 press and healthcheck
## Request for comment

@johnduffell

### DIFF
--- a/common/app/common/jobs.scala
+++ b/common/app/common/jobs.scala
@@ -5,10 +5,12 @@ import java.util.TimeZone
 import org.quartz.impl.StdSchedulerFactory
 import org.quartz._
 import play.api.Mode.Test
+
 import scala.collection.mutable
 import play.api.{Environment, Play}
 
 import scala.concurrent.Future
+import scala.concurrent.duration.Duration
 import scala.util.{Failure, Success}
 
 object JobsState {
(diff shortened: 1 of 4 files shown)


common/app/common/jobs.scala
```diff
@@ -5,10 +5,12 @@ import java.util.TimeZone
 import org.quartz.impl.StdSchedulerFactory
 import org.quartz._
 import play.api.Mode.Test
+
 import scala.collection.mutable
 import play.api.{Environment, Play}
 
 import scala.concurrent.Future
+import scala.concurrent.duration.Duration
```
johnduffell: the only thing I don't like about .seconds is the magical import

johnduffell:
fantastic, much improved:tada:

> Kinda useless 😄

no it's super valuable cos it makes the code more maintainable and your other PR simpler 👯 

👍 

and yes the minutes one is more complex, but hopefully just mechanical changes again... see how we feel

common/app/common/jobs.scala
piuccio: Do you prefer `Duration.create(5, "seconds")` ?

johnduffell: good suggestion I didn't know about that way... if "seconds" was checked at compile time (i.e. a constant or something) i would, however as a magic string I dislike this even more! 😞

janua:
:+1:

PR #14003, opened 2016-08-15
Commercial performance dashboard

rich-nguyen:
A first stab at a live dashboard for showing the real-user application start times of our commercial js system. The data is already in ElastiCache, so this PR pulls the data out through admin.


diagnostics/app/model/diagnostics/commercial/RedisReport.scala
```diff
@@ -42,7 +44,10 @@ class ExpiredKeyEventSubscriber(client: RedisClient, system: ActorSystem) extend
         reportData <- redisClient.get(RedisReport.dataKeyFromId(id))
       } {
         log.logger.info(s"writing report to s3 bucket, view id: $id")
-        S3CommercialReports.putPublic(s"commercial-client-logs/$id", reportData, "text/plain")
+
+        if (Configuration.environment.isProd) {
```
janua: Why do this here? It's not really uploading to S3 anymore. Maybe do it in the `callback`?

Why do you even need this anyway? Does it listen on `CODE`?

janua:
Why have `Report` instead of `CommercialReport`? (I know it's not in the PR)

Looks good :+1:

diagnostics/app/model/diagnostics/commercial/RedisReport.scala
rich-nguyen: ah i see, i'll move it to `callback`

PR #14004, opened 2016-08-16
Fix wrong sticky ad dimension

regiskuckaertz:
Before (notice the ad overlaps content):
picture 7

After:
picture 6

cc @guardian/commercial-dev


rich-nguyen:
👍

PR #14005, opened 2016-08-16
Do the nvm install step when nvm is already installed

sihil:
## What does this change?

The command that installs the relevant version of node (`nvm install`) is only run just after `nvm` itself has just been installed. This changes it so that `install_node` will run `nvm install` when `nvm` in already installed.

If a user already has `nvm` but has not worked on Frontend before (or when the version of node for Frontend is changed) then the developer is unlikely to have the specified version of node. By running `nvm install`, `setup.sh` will install node if necessary.

No

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -18,10 +18,23 @@ installed() {
 }
 
 nvm_installed() {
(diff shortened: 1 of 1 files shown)


rich-nguyen:
👍

TBonnin:
👍

PR #14006, opened 2016-08-16
Use latest sbt-riffraff-artifact with new artifact format

sihil:
## What does this change?

This updates the frontend build to use the new riff-raff artifact format. The SBT build is now uses the latest `sbt-riffraff-artifact` plugin. I've also updated the custom `dotcom:static` build to use the new format. 

Whilst reviewing the custom build script I also switched the upload order so that `build.json` is uploaded last to eliminate a race condition.

The plugin now uses the new Riff-Raff artifact format (see https://github.com/guardian/riff-raff/pull/311 and https://github.com/guardian/sbt-riffraff-artifact/pull/25). I expect the build to be a little faster as it no longer creates archives for all of the projects.

In addition I expect that these continuous delivery failures will stop: https://riffraff.gutools.co.uk/deployment/history?projectName=dotcom%3Astatic&status=Failed&pageSize=20&page=1

No

### DIFF
--- a/dev/teamcity/dist-publish-assets-tc
+++ b/dev/teamcity/dist-publish-assets-tc
@@ -19,11 +19,6 @@ cp "static/deploy.json"      "$static_folder"
 cp -r static/hash/*          "$static_folder/packages/frontend-static"
 cp static/abtests.json       "$static_folder/packages/frontend-abtests"
 
-pushd $static_folder
-zip -r "../artifacts.zip" .
-mv ../artifacts.zip ./artifacts.zip
-popd
-
 echo "Uploading artifact and build manifest to S3"
 
 set +u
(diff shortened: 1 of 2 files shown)


piuccio:
👍

TBonnin:
👍

sihil:
Whilst checking I understood the TC build that the static artifact is a custom affair. I've updated it to use the new format. I've also switched the upload order so that `build.json` is uploaded last (this file is used by Riff-Raff to kick of continuous deployment so there is a race condition if this is uploaded first - explains these failures: https://riffraff.gutools.co.uk/deployment/history?projectName=dotcom%3Astatic&status=Failed&pageSize=20&page=1)

dev/teamcity/dist-publish-assets-tc
```diff
@@ -49,13 +44,8 @@ sed -e "s|<%build_number%>|$BUILD_NUMBER|" \
...
     static/build-template.json | tee $static_folder/build.json
 
+aws s3 cp --acl bucket-owner-full-control --region=eu-west-1 --recursive $static_folder s3://$RIFF_RAFF_ARTIFACT_BUCKET/dotcom:static/$BUILD_NUMBER
 aws s3api put-object --acl bucket-owner-full-control --region=eu-west-1 --bucket $RIFF_RAFF_BUILD_BUCKET --key dotcom:static/$BUILD_NUMBER/build.json  --body $static_folder/build.json
-aws s3api put-object --acl bucket-owner-full-control --region=eu-west-1 --bucket $RIFF_RAFF_ARTIFACT_BUCKET --key dotcom:static/$BUILD_NUMBER/artifacts.zip --body $static_folder/artifacts.zip
-
-set +o xtrace
-echo "##teamcity[publishArtifacts '$static_folder/artifacts.zip => static']"
-echo "##teamcity[publishArtifacts '$static_folder/build.json => static']"
-set -o xtrace
```
TBonnin: Removing those lines means we won't be able to see the artifacts in teamcity. Probably not a big deal though.

TBonnin:
How much slower it is to upload all files rather than the zipped archive?

dev/teamcity/dist-publish-assets-tc
sihil: I've done the same for all of the other artifacts (albeit through a change in behaviour in `sbt-riffraff-artifact` in 0.8.4 - https://github.com/guardian/sbt-riffraff-artifact/pull/20). I sent an e-mail earlier asking whether people wanted them - perhaps that wasn't the best channel for the question though.

TBonnin: Sorry, I missed the email. I am fine with this change, as long as it is done on purpose 😉

TBonnin:
👍

sihil:
I've not noticed a big difference when the latency to AWS is low (i.e. the build server is in the same region). My testing locally (where latency is comparatively high) indicates that it is slightly slower: 
Artifact.zip:
```
real   	0m3.953s
user   	0m1.978s
sys    	0m0.489s
```

Individual files:
```
real   	0m4.992s
user   	0m2.290s
sys    	0m0.902s
```

TBonnin:
Thanks. Not a huge difference.

PR #14008, opened 2016-08-16
Fall back to trailPicture if thumbnailPath is missing

SiAdcock:
## What does this change?

It seems that not all articles have a thumbnail field returned by CAPI. This may happen, for instance, when the trail image does not meet the [required minimum image dimensions](https://github.com/guardian/flexible-content/blob/master/flexible-content-api/src/main/scala/com/gu/flexiblecontent/api/event/stream/model/StreamableContent.scala#L686) of 500x300.

As a fallback, this changes uses the `largestImageUrl` property on the trail picture

No


NataliaLKB:
👍

PR #14010, opened 2016-08-16
Stop appearance of secondary scroll bar

scottp719:
## What does this change?

adds an overflow: hidden attribute to the page skin in adslot.css to stop the appearance of the nasty second scroll bar

Death to the second scroll

Nopers
## Screenshots

image
## Request for comment

@regiskuckaertz @JonNorman

### DIFF
--- a/static/src/stylesheets/module/_adslot.scss
+++ b/static/src/stylesheets/module/_adslot.scss
@@ -602,6 +602,7 @@
  */
 .ad-slot--page-skin {
     height: 0;
+    overflow: hidden;
 }
 @include mq(mobile, $until: wide) {
     .has-page-skin {


JonNorman:
I still don't fully understand how the page-skin ad slot causes this second scroll bar to appear. But this ensures that whatever happens when we load the page-skin, it won't mess up our scrolling.

LGTM :+1:

regiskuckaertz:
I think it has to do with the fact that it is the last visible element in normal flow, after the footer. If you move it before the footer, the scrollbar disappears.

regiskuckaertz:
Yeah!

regiskuckaertz:
(nb: I also think this element should be taken out of the normal flow with `position: absolute`, to avoid any potential side effect should another element ever be appended to the `BODY`)

PR #14011, opened 2016-08-16
Commercial: remove unused styles

regiskuckaertz:
We have a lot of styles hanging around that were used in the previous instantiation of our commercial components and we can get rid of. Two components still remain though, so some styles have to be there:
- `commercialContainers.scala.html` is for multi-campaigns containers and will soon go away as the GLabs stream is deploying branded containers (@kelvin-chappell can you please let me know when this is done?)
- `soulmatesTest.scala.html` is apparently not used anymore but this is pending confirmation from the merchandising team

cc @guardian/commercial-dev


JonNorman:
LGTM 👍

Calanthe:
👍

PR #14012, opened 2016-08-16
Use d3 version 3.x to support epoch-charting js library

rich-nguyen:
A fix for the admin dashboard which uses `epoch-charting`, which specifically needs d3 `3.x`, not `4.x`. This somehow worked locally.


janua:
:+1:

PR #14013, opened 2016-08-16
Update GA click events

cb372:
## What does this change?
- Send click events (in-page/internal/external link clicks) to the correct GA property: the prod one in prod, the test one in test
- Apply sampling to the test GA tracker (except in dev mode), so people can test new types of events in prod without sending huge numbers of events to the test tracker. This means we can upgrade the test tracker to a premium property, giving better parity with the prod property, without worrying about blowing our server call quota.

Send click events to the correct GA property

No
## Screenshots

n/a
## Request for comment

@dominickendrick @ajosephides


ajosephides:
👍

PR #14015, opened 2016-08-16
Increase the refresh rate of the commercial browser dashboard

rich-nguyen:
Tweak the refresh rate to every second. This is an internal dashboard, should be fine.


janua:
:+1:

PR #14016, opened 2016-08-16
Implement details element polyfill for new navigation

SiAdcock:
## What does this change?

The `<details>` element is not supported in Internet Explorer and older Firefox versions. This change introduces a JavaScript-based polyfill.

No

### DIFF
--- a/static/src/javascripts/projects/common/modules/navigation/newHeaderNavigation.js
+++ b/static/src/javascripts/projects/common/modules/navigation/newHeaderNavigation.js
@@ -2,7 +2,8 @@ define([
     'common/utils/fastdom-promise',
     'common/utils/$',
     'common/modules/navigation/edition-picker',
-    'common/modules/navigation/editionalise-menu'
+    'common/modules/navigation/editionalise-menu',
+    'common/utils/details-polyfill'
 ], function (
     fastdomPromise,
     $,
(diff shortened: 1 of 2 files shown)


piuccio:
screen shot 2016-08-17 at 10 04 54

it might affect amp

SiAdcock:
That's a good shout, thanks @piuccio. 

This polyfill won't be loaded in AMP pages because we can't add any JavaScript to AMP pages. However, we'll still have to bear this in mind when we build an AMP version of the navigation.

dominickendrick:
:+1:

rich-nguyen:
isn't @piuccio 's observation regarding AMP a bit of a deal-breaker for using `details`? Is there a way to implement the nav without it?

NataliaLKB:
>

With the current restrictions in AMP, I don't think we would ever have the same implementation of the nav for both our site and AMP. That said I think the plan is certainly to make them look the same. For AMP we would use the AMP component [amp-sidebar](https://github.com/ampproject/amphtml/blob/75d7352d573f3a55253deab4e4abf32bf8763aed/extensions/amp-sidebar/amp-sidebar.md)

paperboyo:
Even amp-sidebar does not have swipe-from-edge to reveal and swipe-left to hide functionality :sadpanda: Safari is still stupid here, hijacking this for OS-wide navigation, but maybe it could at least work in all proper browsers?

PR #14017, opened 2016-08-16
Add static/assets.json to .gitignore

sihil:
Add `static/assets.json` to `.gitignore`. This is generated by some of the scripts under `dev/teamcity` - which are not usually run by devs, but none the less we should ensure it is never checked in.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,6 +59,7 @@ developer.properties
 static/hash
 static/target
 static/riffraff
+static/abtests.json
 
 static/src/stylesheets/icons/*
 


.gitignore
```diff
@@ -59,6 +59,7 @@ developer.properties
 static/hash
 static/target
 static/riffraff
+static/abtests.json
```
piuccio: > Add static/assets.json to .gitignore

😁

piuccio:
👍

johnduffell:
get it shipped! 🚀

PR #14018, opened 2016-08-16
Hosted video pages - capi driven css

Calanthe:
## What does this change?

This is the follow up to what @kelvin-chappell did in the https://github.com/guardian/frontend/pull/13988. Hosted pages (right now only video) will have brand specific styles taken from capi (brand colour and font theme - black or white). The idea is to declare those styles on top of the hosted template and in the future remove the whole `_hosted-colours.scss` file together with brand specific class names in the templates (eg `hostedbadge.hosted-tone-bg--f1-singapore`).

Two values: brand colour and font theme will be specified in composer and based on the capi, proper css rules will be generated in templates.

Not yet.
## Screenshots

screen shot 2016-08-16 at 17 32 42
## Request for comment

@kelvin-chappell @guardian/commercial-dev

### DIFF
--- a/commercial/app/views/hosted/guardianHostedVideo.scala.html
+++ b/commercial/app/views/hosted/guardianHostedVideo.scala.html
@@ -44,27 +44,10 @@
     .hosted-page ~ .survey-overlay-simple .survey-text__header {
         background-color: @{page.campaign.brandColour};
     }
-
-    @if(page.campaign.brightFont) {
-        .hosted-tone-btn,
-        .hostedbadge.hosted-tone-bg .hostedbadge__info,
-        .hosted-page ~ .survey-overlay-simple .survey-text__header {
-            color: #ffffff;
-        }
-
-        .hosted-tone-btn svg,
-        .hosted-page ~ .survey-overlay-simple .survey-text__header .site-message__close-btn .inline-cross {
-            fill: #ffffff;
-        }
-
-        .hosted-page ~ .survey-overlay-simple .survey-text__header .site-message__close-btn {
-            border-color: #ffffff;
-        }
-    }
     </style>
     <!--<![endif]-->
-    @guardianHostedHeader("hosted-video-page", page.campaign.cssClass, page.campaign.logo.url, page.campaign.owner, page.campaign.logoLink.getOrElse(page.cta.url))
-    <div class="hosted-page l-side-margins hosted__side hosted-video-page @{page.campaign.cssClass}">
+    @guardianHostedHeader(if(page.campaign.brightFont) "hosted-video-page hosted-page--bright" else "hosted-video-page", page.campaign.cssClass, page.campaign.logo.url, page.campaign.owner, page.campaign.logoLink.getOrElse(page.cta.url))
+    <div class="hosted-page l-side-margins hosted__side hosted-video-page @{page.campaign.cssClass} @if(page.campaign.brightFont) {hosted-page--bright}">
         <section class="hosted-tone--dark">
             <div class="host hosted__container--full">
                 <div class="u-responsive-ratio u-responsive-ratio--hd">
--- a/commercial/app/views/hosted/guardianHostedVideo.scala.html
+++ b/commercial/app/views/hosted/guardianHostedVideo.scala.html
@@ -5,6 +5,17 @@
 @import views.html.hosted.guardianHostedCta
 
 @main(page, Some("commercial")) { } {
+    <!--[if (gt IE 9)|(IEMobile)]><!-->
+    <style>
+    .hosted-ton {
+        color: page.campaign.mainColour;
+    }
+
+    .hosted-tone-bg {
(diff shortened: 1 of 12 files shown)


commercial/app/views/hosted/guardianHostedVideo.scala.html
```diff
@@ -5,6 +5,64 @@
 @import views.html.hosted.guardianHostedCta
 
 @main(page, Some("commercial")) { } {
+    <!--[if (gt IE 9)|(IEMobile)]><!-->
```
regiskuckaertz: Would be great if this was minified. Perhaps in the grunt pipeline?

commercial/app/views/hosted/guardianHostedVideo.scala.html
```diff
@@ -5,6 +5,64 @@
...
+
+    .hosted-page ~ .survey-overlay-simple .survey-text__header {
+        background-color: @{page.campaign.brandColour};
+    }
+
+    @if(page.campaign.brightFont) {
+        .hosted-tone-btn,
+        .hostedbadge.hosted-tone-bg .hostedbadge__info,
+        .hosted-page ~ .survey-overlay-simple .survey-text__header {
+            color: #ffffff;
```
regiskuckaertz: Shouldn't all this go away, leaving only the variables there?

commercial/app/views/hosted/guardianHostedVideo.scala.html
```diff
@@ -5,6 +5,64 @@
...
+        }
+
+        .hosted-tone-btn svg,
+        .hosted-page ~ .survey-overlay-simple .survey-text__header .site-message__close-btn .inline-cross {
```
regiskuckaertz: now _that_ is a selector 😆

commercial/app/views/hosted/guardianHostedVideo.scala.html
```diff
@@ -5,6 +5,64 @@
...
+        background-color: @{page.campaign.brandColour};
+        border-color: @{page.campaign.brandColour};
+    }
+
+    .hosted-page .hosted-next-autoplay .hosted-next-autoplay__poster-timer,
+    .hosted-page .hosted-next-autoplay .hosted-next-autoplay__tile {
+        background-color: @{page.campaign.brandColour};
+    }
+
+    .hosted-page .hosted-next-autoplay .hosted-next-autoplay__video,
```
gtrufitt: There's some pretty deeply nested classes in this CSS - It seems if they're all prefixed with `.hosted-` you should be able to just use the last class? If you're overriding specificity, I'd try and revisit the classes you're overriding to see if you can reduce theirs.

common/app/common/commercial/hosted/HostedPage.scala
```diff
@@ -35,9 +35,8 @@ case class HostedCampaign(
   owner: String,
   logo: HostedLogo,
   cssClass: String,
-  mainColour: String,
-  // todo: this will come through from capi once worked out how it's going to be used
-  secondaryColour: Option[String] = None,
+  brandColour: String,
+  brightFont: Boolean,
```
kelvin-chappell: Maybe better to encapsulate these two properties into a case class, eg:
```
case class CampaignColour(hexCode: String) {
    val isBright: Boolean = ???
}
```

There's a possible approach to determining brightness here:
http://stackoverflow.com/a/4680309/593836

commercial/app/views/hosted/guardianHostedVideo.scala.html
Calanthe: Yes, good point. I will refactor those selectors

commercial/app/views/hosted/guardianHostedVideo.scala.html
Calanthe: I am not sure if I understand you. There will be two font colours schemes: white and black. User will only specify which scheme will look better for the campaign/brand colours.

commercial/app/views/hosted/guardianHostedVideo.scala.html
Calanthe: Good point! I will add this to the TODO list and implement it as one of the last step. Currently it's just a proof of concept to test this solution.

common/app/common/commercial/hosted/HostedPage.scala
Calanthe: Good idea! I will try to implement this in another PR.

commercial/app/views/hosted/guardianHostedVideo.scala.html
regiskuckaertz: forget it, I did not see the test above 🙈  but I think a small improvement would be to add a class to the BODY if `brightFont` is true and move all these styles into the CSS. but you can do this in another release.

regiskuckaertz:
Aside from selector complexity, looks good to m 👍

PR #14019, opened 2016-08-17
Now that's what I call graceful degradation 2016

zeftilldeath:
If people bump up their font size, or have phones the size of field mice - the new header will look A-ok 👍 🐭 

From this:
screen shot 2016-08-17 at 14 42 45

To this:
smaller

With added this (hit area as large as possible): 
screen shot 2016-08-17 at 14 41 42

@stephanfowler


duarte:
:+100

PR #14020, opened 2016-08-17
Hosted Gallery - loading optimisation

lps88:
## What does this change?

Use some funky css to approximate the width of the gallery before the js has loaded.

At the moment, while the page loads the header and footer elements sit at the far edges of the screen, then jump inwards to align to the image edges, which is a bit jarring.

Now, they should start quite close to their eventual (calculated by js) position, making it feel smoother!

NB: only works where vh & calc are supported, but is only an optimisation so not too worried
## Screenshots

Previous starting position of the header/footer:
image

Now:
image
## Request for comment

@guardian/labs-beta


Calanthe:
👍

PR #14021, opened 2016-08-17
Commercial: add resize message

regiskuckaertz:
## What does this change?

Next task in the pipe for our cross-frame communication protocol, is to allow the ad to ask the page to resize the slot to a specific size. In this iteration, we will deliberately resize the ad slot as specified, but in the future we might adopt an approach ala AMP, where a user has to advertise intent (e.g. click on a button) for the ad to get expanded.

For all our slots, we will try to set a fixed height according to the type of creative we expect back from DFP. But this size might not always be ideal across all breakpoints and creatives, so there is a need for dynamic resizing.

Nope, not yet.

cc @guardian/commercial-dev

### DIFF
--- a/common/app/templates/inlineJS/blocking/polyfills/matches.scala.js
+++ b/common/app/templates/inlineJS/blocking/polyfills/matches.scala.js
@@ -1,0 +1,21 @@
+@()
+
+/* Source: https://github.com/remy/polyfills/blob/master/classList.js */
+(function (Element) {
+
+    if (!Element.prototype.matches) {
+        Element.prototype.matches =
+            Element.prototype.matchesSelector || 
+            Element.prototype.mozMatchesSelector ||
+            Element.prototype.msMatchesSelector ||
+            Element.prototype.oMatchesSelector ||
+            Element.prototype.webkitMatchesSelector ||
+            function(s) {
+                var matches = (this.document || this.ownerDocument).querySelectorAll(s),
+                    i = matches.length;
+                while (--i >= 0 && matches.item(i) !== this) {}
+                return i > -1;
+            };
+    }
+
+})(window.Element);
--- a/common/app/templates/inlineJS/blocking/polyfills/matches.scala.js
+++ b/common/app/templates/inlineJS/blocking/polyfills/matches.scala.js
@@ -5,16 +5,19 @@
 
     if (!Element.prototype.matches) {
         Element.prototype.matches =
-            Element.prototype.matchesSelector || 
+            Element.prototype.matchesSelector ||
             Element.prototype.mozMatchesSelector ||
             Element.prototype.msMatchesSelector ||
             Element.prototype.oMatchesSelector ||
             Element.prototype.webkitMatchesSelector ||
             function(s) {
-                var matches = (this.document || this.ownerDocument).querySelectorAll(s),
-                    i = matches.length;
-                while (--i >= 0 && matches.item(i) !== this) {}
-                return i > -1;
+                var matches = (this.document || this.ownerDocument).querySelectorAll(s);
+                var ii = matches.length;
+                var i = 0;
+                while (i < ii && matches[i] !== this) {
+                    i += 1;
+                }
+                return i < ii;
             };
     }
 
--- a/common/app/templates/inlineJS/blocking/polyfills/matches.scala.js
+++ b/common/app/templates/inlineJS/blocking/polyfills/matches.scala.js
@@ -1,15 +1,15 @@
 @()
 
 /* Source: https://github.com/remy/polyfills/blob/master/classList.js */
-(function (Element) {
+(function (ElementPrototype) {
 
-    if (!Element.prototype.matches) {
-        Element.prototype.matches =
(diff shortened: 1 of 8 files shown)


static/src/javascripts/projects/common/utils/closest.js
```diff
@@ -0,0 +1,18 @@
+define(function () {
+    /**
+     * Adds a wrapper around Element.closest()
+     */
+    var matches = 'matches' in Element.prototype ? 'matches' : 'msMatchesSelector';
```
rich-nguyen: are we sure that all polyfilled browsers have either `matches` or `msMatchesSelector` in `Element` proto?

regiskuckaertz: I based this on the assumption that `Element.matches` exists for all browsers that are running commercial, except IE9 which uses `msMatchesSelector`. I may need to add the `webkit` prefix too [for older androids](http://caniuse.com/#feat=matchesselector) though. @ScottPainterGNM do you agree?

static/src/javascripts/projects/common/utils/closest.js
```diff
@@ -0,0 +1,18 @@
...
+    return 'closest' in Element.prototype ? closestNative : closestPolyfill;
+
+    function closestNative(element, selectors) {
+        return element.closest(selectors);
+    }
+
+    function closestPolyfill(element, selectors) {
+        while (element && !element[matches](selectors)) {
```
rich-nguyen: I think (not sure!) there is a bug here. on the first while iteration, we are running `matches` on the specified node, not the first ancestor. So if the specified node happens to contain descendant nodes that match the selector, the polyfill will return the specified node, even when no closest ancestor satisfies the selector.

static/src/javascripts/projects/common/utils/closest.js
rich-nguyen: Polyfill for `matches` here: https://developer.mozilla.org/en-US/docs/Web/API/Element/matches

static/src/javascripts/projects/common/utils/closest.js
regiskuckaertz: This is indeed counter-intuitive, but it is [how the `closest` method works](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest). ECMAScript authors would be lambasted by taxonomists, that's for sure 😆

rich-nguyen: Just tested it, it's fine, ignore the above. I thought `matches` was using a context, but it doesn't.

rich-nguyen:
👍  I wish there was a clear way to distinguish amd modules which are used for module initialisation (like registering handlers), and ones which return usable module objects.

common/app/templates/inlineJS/blocking/polyfills/matches.scala.js
```diff
@@ -0,0 +1,21 @@
+@()
+
+/* Source: https://github.com/remy/polyfills/blob/master/classList.js */
```
rich-nguyen: Is that the right source url?

rich-nguyen:
@regiskuckaertz do you think we should have some sort of convention for modules like `get-stylesheet` and `resize`? Otherwise it will soon be hard to see the messages available to be used in the protocol. Similarly for the `set-ad-height` message, i'm not sure where the best place is for that. it should probably be changed to use `resize` ?

static/src/javascripts/projects/common/utils/closest.js
ScottPainterGNM: Sorry I was late to the party on this but yes I think it's worth adding the `webkit` prefix as the uptake of new versions of Android is way slower than iOS and as such we can expect a fair amount of traffic across versions of Android 4.

regiskuckaertz:
@rich-nguyen 

- re AMD modules: indeed this distinction does not really exist and the problem won't go away as we switch to ES6 modules. We could come up with some nomenclature and also try to segregate the ones from the others.

- re convention: you mean, in terms of their internal structure? or naming? the `set-ad-height` is quite problematic, it is intricately related to the sticky banner; and the fact that there are many duplicate rubicon line items make it hard to update it. I can probably pipe it to `resize` though, very good idea!

rich-nguyen:
@regiskuckaertz I was thinking any simple convention which distinguishes modules which are protocol-message-listeners from the modules that make up the functional parts of the decomposed old `dfp-api`. Think this would help readers.

- Yes, maybe move `messenger.js`, and message-listener modules into a sub folder within `private`, like `cross-frame-messages`.
- Just naming. So, always names modules that listen to messages by the message name, eg. `get-stylesheet.js`  would be `get-styles.js`.

Piping to `resize` seems sensible, it feels like `set-ad-height` was the first instance of cross-frame comms, so now we have a general implementation (`resize`), they should work through one shared message handler, allowing us to safely migrate away from `set-ad-height` within the DFP scripts.

johnduffell:
conflicts! 😢 is this ready to go?

regiskuckaertz:
Closing this until the DFP team figures out a solution to the problem I've reported.

PR #14022, opened 2016-08-17
Remove browserslist file and move config into autoprefixer JS

gtrufitt:
## What does this change?

Moves autoprefixer configuration from seemingly random (or 'documentation') file to in the JS file that uses it.

We won't break the website by removing it accidentally in clean up.

@guardian/dotcom-platform 

Note: There may be tools that can use browserlist in the future (eg webpack), but even then, I'm super un-keen on this magic configuration file, it's something we can probably put in package.json or a dotfile or something.


johnduffell:
great great great! 👍 🚀

PR #14023, opened 2016-08-17
Rotated some arrows in the new nav

zeftilldeath:
Influenced by something called UX (No idea what that means; if it was User Experience it would be UE right?) @paperboyo suggested the chevrons change direction to be nice and clear.

And by gum it's worked.

screen shot 2016-08-17 at 16 01 23

As usual cc @stephanfowler


duarte:
_"Inspiring"_ — ★★★★✩

👍

paperboyo:
Everything’s proceeding according to my plan...

PR #14025, opened 2016-08-17
Ad perf dashboard visual tweaks

rich-nguyen:
Added a label for the y-axis, and an average start time component. Also filtered erroneous datapoints coming in, I'll investigate this later. It looks like some data is using the `Date.getTime` object incorrectly.

cursor_and_real-user_advert_performance


janua:
:+1:

PR #14028, opened 2016-08-17
Amp-validation utility for teamcity

jfsoul:
## What does this change?

Adds a node.js utility for running the official amphtml-validator against various AMP pages on www.theguardian.

The utility is intended to be run as a separate project on teamcity (probably on a schedule).  Right now this is ~~untested~~ tested on teamcity - at this stage it would just be good to get feedback on this ~~early (and slightly rough)~~ version.  ~~I'm guessing there might need to be tweaks to accommodate some things I haven't thought of.~~

The flow is as follows:
- Fetch 2 versions of the validator.js from cdn.ampproject.org
  - release
  - pre-release (this is fetched via addition of AMP_CANARY=1 cookie)
- For each validator.js version, run each page against the validator
- Output the results
- Exit the process with relevant exit code (i.e. 1 if any of the pages fail validation, 0 if not).

Comments on ES6-ness especially appreciated.  Dependencies are intentionally kept to a minimum at the moment, so it's quite bare-bones.

Our AMP validation process should be much tighter, as this will always validate pages against the latest released validator (which is not the case for scala tests), _**and**_ the pre-release version.
## Screenshots

image
## Request for comment

@SiAdcock @gtrufitt @TBonnin

### DIFF
--- a/article/test/ArticleAmpValidityTest.scala
+++ b/article/test/ArticleAmpValidityTest.scala
@@ -7,7 +7,8 @@ import org.scalatest.DoNotDiscover
 
   /*
       NB: these may be duplicated in tools/amp-validation - if you add or remove
-      a URL from here, think about whether it should be removed there too.
+      a URL from here, think about whether it should be changed there too. Not
+      centralising as running full suite here may be overkill as we add more tests
    */
   Seq(
     "/commentisfree/2016/aug/09/jeremy-corbyn-supporters-voters-labour-leader-politics", // Comment tone
--- a/article/test/ArticleAmpValidityTest.scala
+++ b/article/test/ArticleAmpValidityTest.scala
@@ -4,6 +4,11 @@ import org.scalatest.DoNotDiscover
 
 
 @DoNotDiscover class ArticleAmpValidityTest extends AmpValidityTest {
+
+  /*
+      NB: these may be duplicated in tools/amp-validation - if you add or remove
+      a URL from here, think about whether it should be removed there too.
+   */
   Seq(
     "/commentisfree/2016/aug/09/jeremy-corbyn-supporters-voters-labour-leader-politics", // Comment tone
     "/politics/2016/aug/09/tom-watson-interview-jeremy-corbyn-labour-rifts-hug-shout", // Feature tone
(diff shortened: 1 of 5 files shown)


article/test/ArticleAmpValidityTest.scala
```diff
@@ -4,6 +4,11 @@ import org.scalatest.DoNotDiscover
 
 
 @DoNotDiscover class ArticleAmpValidityTest extends AmpValidityTest {
+
+  /*
+      NB: these may be duplicated in tools/amp-validation - if you add or remove
+      a URL from here, think about whether it should be removed there too.
+   */
```
TBonnin: Can we consolidate all those urls in a single location? Like a file both tests would read to get all the urls.
I am not sure developers read comments 😜

tools/amp-validation/index.js
```diff
@@ -0,0 +1,72 @@
...
+const amphtmlValidator = require('amphtml-validator');
+const validatorJs = require('./validator-js');
+const fetchPage = require('./fetch-page');
+
+/*
+    NB: these may be duplicated elsewhere in scala tests
+    that extend AmpValidityTest - consider adding any new
+    URLs there as well.
+ */
+// TODO: should we centralise these?
```
TBonnin: Yes 🙏

article/test/ArticleAmpValidityTest.scala
jfsoul: As discussed, we're expecting the 2 sets of test endpoints to diverge over time.  Have updated comments to match (even if no one reads them 😁 )

jfsoul:
@dominickendrick this is the PR, if you have a time to have a glance it would be much appreciated!

TBonnin:
@jfsoul I am not an expert but LGTM 👍

tools/amp-validation/validator-js.js
```diff
@@ -0,0 +1,68 @@
+'use strict';
+
+const https = require('https'); // TODO - will we ever need http?
```
dominickendrick: isn't AMP https by default ?

dominickendrick:
👍

PR #14029, opened 2016-08-17
User Timings module should return relative time values

rich-nguyen:
For browsers that don't support the performance API, a `Date` object was used, but the values ought to be relative time periods (in milliseconds) to approximate the behaviour of performance API.


janua:
:+1:

PR #14030, opened 2016-08-17
Force pulling the docker image and make install at startup

TBonnin:
## What does this change?

Pull automatically the most recent version of the frontend-dev docker image and run make install at startup

Running the frontend-dev container will always mean running the most recent version of the frontend code. 

No
## Request for comment

@jamespamplin @jfsoul @dominickendrick


dev.sh
```diff
@@ -1,7 +1,14 @@
 #!/bin/sh
 
 # Always try to pull most recent image
-docker push 642631414762.dkr.ecr.eu-west-1.amazonaws.com/frontend-dev:latest
+ecrLogin=$(aws ecr get-login --profile frontend) 
+if [ $? -eq 255  ] 
+then
+    echo "Please use Janus to get credentials"
+    exit 1;
+fi
```
TBonnin: Forcing aws credentials to be valid. You need it to run the apps anyway so I don't think it is an overkill to ask for them when running the container.
It would only be an issue if you want to hack around without connection. Let me know if you think this is a problem

dev.sh
```diff
@@ -1,7 +1,14 @@
 #!/bin/sh
 
 # Always try to pull most recent image
-docker push 642631414762.dkr.ecr.eu-west-1.amazonaws.com/frontend-dev:latest
```
TBonnin: Was doing a push before. Clearly a mistake in my previous commit

docs/01-start-here/01-installation-steps.md
```diff
@@ -35,12 +35,9 @@ Prerequesites:
 2. Get AWS Credentials using [janus](https://janus.gutools.co.uk/) for frontend and CMS fronts 
 3. Get properties file
     - `aws s3 cp --profile frontend s3://aws-frontend-store/template-frontend.properties "~/.gu/frontend.properties"`
-4. Log into the EC2 Container Registry
-    - `aws ecr get-login --region eu-west-1 --profile frontend`
-    - Run the docker login command returned in the previous step
```
TBonnin: This step is not necessary anymore since it is part of running `./dev.sh`

Dockerfile-dev
```diff
@@ -37,7 +37,6 @@ RUN cd /source && \
     ./sbt compile && \
     rm -rf /source
 
-
 # Add useful info in .bashrc
-ADD dev-help.sh /root/.dev-help.sh
-RUN echo 'source /root/.dev-help.sh' >> /root/.bashrc
+ADD docker-dev-startup.sh /root/.startup.sh
```
TBonnin: rename `dev-help.sh` since it now contains more than just help info

dev.sh
```diff
@@ -1,7 +1,55 @@
...
+function showHelp {
+    echo ""
+    echo "Usage: ./dev.sh [OPTIONS]"
+    echo ""
+    echo "Run the Guardian frontend dev environment in a Docker container"
+    echo ""
+    echo "Options:"
+    echo "  -h, --help      Print usage"
+    echo "  -n, --nopull    Do not try to pull newest Docker image. Use in offline mode for example."
```
jfsoul: No pull option: 💯  👍

docker-dev-startup.sh
```diff
@@ -0,0 +1,34 @@
...
+white=`tput setaf 7`
+
+reset=`tput sgr0`
+
+# Make the prompt prettier
+export PS1="[🐳 :\h]\w# "
+
+# Update node dependencies
+echo "${background}${white}${bold}Updating node dependencies...${reset}"
+pushd /frontend; npm install && npm prune; popd
```
jfsoul: Is this equivalent to make install?

docs/01-start-here/01-installation-steps.md
TBonnin: Yep without the messages and such

jfsoul:
lgtm, usability improvements always welcome 👍

dev.sh
```diff
@@ -52,4 +65,9 @@ then
 fi
 
 # Run dev container
-docker-compose run --rm --service-ports dev
+if [ -z "$ports" ] #Ports option is empty
+then
+    ports="--service-port" #Publish all service's port as defined in docker-compose file
```
jfsoul: `--service-ports`?

dev.sh
```diff
@@ -35,9 +41,16 @@ case $key in
...
     ;;
+    -w|--watch)
+    serviceToRun="watch"
+    ;;
     -n|--nopull)
     shouldPullLatestImage=0
     ;;
+    -p|--port)
+    ports="$ports -p $2"
+    shift
```
jfsoul: you'll have to explain this one to me ✋

dev.sh
TBonnin: good spot

jfsoul:
later changes 👌  👍

PR #14032, opened 2016-08-18
Default style rich-links 

gtrufitt:
## What does this change?

Two things:
- Changes the default styling of rich-links to be nearer to their enhanced version
- Doesn't enhance rich-links on mobile 
- Having a more fully styled rich-link at first paint means that there is less jumping when the enhanced rich links load - especially on desktop where the default rich links are in the left column of the article instead of in-article
- Not enhancing rich-links on mobile means that we _do not jump the page at all_ 
- We're running as an A/B test to ensure total CTR and loyalty of user's is not affected
### Desktop

| Un-enhanced before | Un-enhanced after |
| --- | --- |
| image | image |
### Mobile

Full default rich-link (no enhancement on mobile):

image
## Request for comment

@guardian/dotcom-platform @stephanfowler @zeftilldeath


static/src/javascripts/projects/common/modules/experiments/tests/platform-dont-upgrade-mobile-rich-links.js
```diff
@@ -0,0 +1,32 @@
+define([], function () {
+    return function () {
+        this.id = 'DontUpgradeMobileRichLinks';
+        this.start = '2016-08-18';
+        this.expiry = '2016-09-14';
+        this.author = 'Gareth Trufitt';
+        this.description = 'Test whether the loyalty of users decreases with non-enhanced rich links';
+        this.audience = 0.5;
+        this.audienceOffset = 0;
+        this.successMeasure = 'Loyalty of users';
```
johnduffell: which definition are we using? might be worth chatting with the data team to make sure the test is valid

stephanfowler: It would be more accurate to say we're testing _engagement_, notably that there is no drop in overall CTR on the page, or article visits per browser per day.

johnduffell:
> Not enhancing rich-links on mobile means that we do not jump the page at all

I like that! we don't do much testing whether existing features on the site add value and removing if not, so this is a rare treat! 🎉 

👍

static/src/javascripts/projects/common/modules/experiments/tests/platform-dont-upgrade-mobile-rich-links.js
johnduffell: I think the main thing is to invent the test criteria precisely before you start, but I don't want to steal the data guys' thunder as they're way better at explaining it than me

stephanfowler:
👍 
Success criteria are: 
1. no drop in overall article page CTR,
2. no drop in article visits per browser per day.

johnduffell:
sounds good. only trouble is the size of the drop detectable and the number of samples needed are inversely related, so we do need to work out in advance what the smallest drop is that's acceptable, and calculate the sample size needed based on detecting any drop larger than that.  otherwise we'll be running the test for ever 😉

gtrufitt:
Aye, I emailed the MVT team, they mentioned to talk to the ABacus team so will do that

paperboyo:
> Not enhancing rich-links on mobile means that we do not jump the page at all

💯 Amazing! Personally, having no jumping is enough of a “metric”. Even if everything drops in numbers. Reason being: it’s impossible to test for the huge annoyance that the jumping brings and all the _accidental_ clicks that the jumping generates. I had occasions where I have tapped on something I didn’t want three times in a row! I would gladly part with all these extra “pageviews”!

Jholder112233:
> Not enhancing rich-links on mobile means that we do not jump the page at all

Unless things have changed, a big part of rich links being progressively enhanced was so that they abide by takedowns and always have the latest headline. For example, live blog headlines change often, yet if you're not progressively enhancing them on mobile they'll always show the headline that was baked in at embed time. In the case of breaking news, we could be displaying out of date information.

gtrufitt:
> For example, live blog headlines change often, yet if you're not progressively enhancing them on mobile they'll always show the headline that was baked in at embed time.

😭 Good point.

common/app/views/fragments/richLinkDefault.scala.html
```diff
@@ -0,0 +1,20 @@
...
+
+    <div class="rich-link__container">
+
+    <div class="rich-link__header">
+        <h1 class="rich-link__title">@title</h1>
+    </div>
+
+    <div class="rich-link__read-more">
+        <div class="rich-link__arrow">
+            <svg width="24" height="24" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg"><g fill="#005689" fill-rule="evenodd"><path class="rich-link__arrow-icon" d="m12 0c-6.627 0-12 5.373-12 12 0 6.627 5.373 12 12 12 6.627 0 12-5.373 12-12 0-6.627-5.373-12-12-12m.21 19l-.637-.668 4.888-6.326h-11.465v-1.01h11.465l-4.888-6.333.637-.668 6.79 7.158v.685l-6.79 7.157"/></g></svg>
```
regiskuckaertz: Shouldn't this be using `inlineSvg`?

gtrufitt:
Thanks all. 

Re: test sample size - I am using ABacus which means I don't need a specific sample size beforehand - will run at 40% audience and see what the sizes are looking like after 'pretest' phase. 

@regiskuckaertz I've updated to use inlineSvg in this new template and the original one that wasn't using it either. 

@Jholder112233 - Cheers for the info, we need to think about it a little more, going to run this test and see how it gets on and then figure out a solution for updating the headline.

And @paperboyo - completely agree.

regiskuckaertz:
@gtrufitt @Jholder112233 with the CSS `line-clamp` property, you could silently update the text of the link without changing its dimensions. Doing that in an idle callback would make sense, since this is a low priority task compared to other frontend stuff.

gtrufitt:
@regiskuckaertz Yeah, I'm thinking truncation of some kind, line-clamp is a good shout, especially as this is particularly painful on mobile where webkit browsers prevail

PR #14033, opened 2016-08-18
Commercial: relax constraint in ad featured fronts

regiskuckaertz:
## What does this change?

On fronts, at least 2 containers must exist before a merchandising component is inserted. Alas, ad featured tag fronts contain only one container most of the time. In that case, we relax the constraint and insert the component if at least 1 container exists.

mo' money

cc @guardian/commercial-dev


rich-nguyen:
👍

PR #13846, opened 2016-08-03
AB Test name fix for Giraffe

markjamesbutler:
## What does this change?

Fix test name to match js test.
## What is the value of this and can you measure success?

None, except allowing the test to function.
## Does this affect other platforms - Amp, Apps, etc?

No
## Screenshots

screen shot 2016-08-03 at 15 56 13
screen shot 2016-08-03 at 15 55 43
screen shot 2016-08-03 at 15 55 33

screen shot 2016-08-03 at 15 56 00

PR #14034, opened 2016-08-18
Contributions inline update messages

markjamesbutler:
## What does this change?

Adds new message variants to contributions inline article component.

See #13846

No

### DIFF
--- a/static/src/javascripts/projects/common/modules/experiments/tests/giraffe.js
+++ b/static/src/javascripts/projects/common/modules/experiments/tests/giraffe.js
@@ -65,7 +65,7 @@ define([
             {
                 id: 'about',
                 test: function () {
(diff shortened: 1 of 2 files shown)


static/src/javascripts/projects/common/modules/experiments/tests/giraffe.js
```diff
@@ -63,18 +63,18 @@ define([
 
         this.variants = [
             {
-                id: 'always',
+                id: 'about',
                 test: function () {
-                    writer('You’re always reading it. So make sure it’s always here. It’s only fair. ', 'Give to the Guardian', 'co_uk_inarticle_always');
+                    writer('Read all about it, knowing you\'re a part of it. ', 'Please give to the Guardian', 'co_uk_inarticle_about');
```
desbo: We should use the right apostrophe character here, `’`

katebee:
👍  cool

PR #14035, opened 2016-08-18
Remove Shoe package

gtrufitt:
## What does this change?

The shoe package was throwing a warning about `node-waf` but turns out we don't seem to use this package anywhere, so removing it.

Tidy.

@guardian/dotcom-platform 

@rich-nguyen - You mentioned previously you'd done stuff with sockets/ streams, don't know if this was related to that?


rich-nguyen:
I've never used shoe. or shoes.

👍

johnduffell:
👠

PR #14036, opened 2016-08-18
Remove crosswords thumbnails

piuccio:
## What does this change?

Thumbnail haven't worked in a while.

They used to look like this

screen shot 2016-08-17 at 12 41 41

but as o April they look like this

screen shot 2016-08-17 at 12 38 21

Not a huge difference, I'm removing the feature instead of fixing it because otherwise I'll have to implement XML parsing of fetch requests.

Less code and DOM fiddling for crossword

No
## Screenshots

Above


desbo:
👍

PR #14037, opened 2016-08-18
Only splice dashboard data when there are more than 1000 items

rich-nguyen:
Little bug here which caused the average to be miscalculated.


janua:
:+1:

PR #14038, opened 2016-08-18
stop tests needing secrets

johnduffell:
to see the changes without killing the diff screen with 553 files, view only the code changes: https://github.com/guardian/frontend/pull/14038/commits/d54425ec952144ffbdc0bea0b3be497b326176b4
## What does this change?

This enables the tests to run only dependent on the checked in data files and not on any secrets or network.
It also fixes a few other issues I found when trying to refresh the data files.
A notable change, is that in DEV mode it will always download the files from the network.  This stops tests becoming too out of date relative to the live servers, as hopefully it will be refreshed occasionally.  It would make the tests slower locally.

This is part 2 of https://github.com/guardian/frontend/pull/13749
This would enable the teamcity server not to have any secrets.  This means we're less likely to print secrets to the console by mistake during a build, and also means when we move the config into S3, teamcity doesn't need access.  It would also be handy for if we want to move the build to travis.

I have also gone through and fixed up various other issues with the tests of varying significance.  However perfection, I certainly haven't achieved!

@guardian/dotcom-platform @rich-nguyen any comments would be very helpful
@adamnfish go go security!


johnduffell:
look at commit https://github.com/guardian/frontend/pull/14038/commits/d54425ec952144ffbdc0bea0b3be497b326176b4
as the other is all the millions of data files - thanks @TBonnin for helping me split them out again

sport/test/FootballTestData.scala
```diff
@@ -7,12 +7,13 @@ import pa._
 import model.{Competition, Tag, TagProperties, TeamMap}
 import feed.CompetitionsService
 import org.scalatest.mock.MockitoSugar
-import play.api.libs.ws.WSClient
 
 trait FootballTestData extends MockitoSugar {
```
TBonnin: no need for MockitoSugar anymore

common/test/recorder/HttpRecorder.scala
```diff
@@ -15,31 +16,34 @@ import play.api.libs.ws.ning.NingWSResponse
...
+      get(fileName).map(Future(_)).getOrElse {
+        throw new IllegalStateException(s"Data file has not been checked in for: $url - $components, file: $fileName, headers: ${headersFormat(headers)}")
       }
+    } else {
+      // always get the new files, this means we'll find out fast when we've broken stuff
+      // otherwise it's impossible to regenerate things because everything's been running off the checked in file
+      // even when it was broken :(
+      fetch.map(r => put(fileName, fromResponse(r)))
+    }
+
```
TBonnin: As discussed IRL I don't think it is a good idea to force going to the network when running the tests locally.
While I agree it would be good to refresh the data files regularly, failing tests which are unrelated to what you are currently working on will hurt productivity and create confusion.

TBonnin:
👍

PR #14040, opened 2016-08-18
break out minute loading JS into separate AB test 

gidsg:
## What does this change?

Breaks out the loading of the minute.ly JS into a separate AB test in order to target include it on some content pages.

In order for minute.ly to work their code is required on article and video pages in order to track clickthrough and playback. The existing minute test runs alongside to the same audience but now only tracks the clickthrough from fronts.

I am aware this increases the payload for the test group but it is required to be able to test in order to assess the value of the minute.ly video teasers.

No
## Screenshots

See example video teasers on http://abcnews.go.com/
## Request for comment

CC @akash1810 @harrysalmon

### DIFF
--- a/common/app/conf/switches/ABTestSwitches.scala
+++ b/common/app/conf/switches/ABTestSwitches.scala
@@ -144,6 +144,16 @@ trait ABTestSwitches {
     exposeClientSide = true
   )
 
+  val ABMinuteLoadJs = Switch(
+    SwitchGroup.ABTests,
+    "ab-minute-load-js",
+    "Load JS for minute test participants on some content pages.",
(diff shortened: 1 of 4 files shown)


akash1810:
> The existing minute test runs alongside to the same audience but now only tracks the clickthrough from fronts.

Can't see this change in this PR... is there another one that I've just missed?

gidsg:
@akash1810 see changes in `minute.js` https://github.com/guardian/frontend/pull/14040/files#diff-17d1fb509ed3c571d84e9d8caa9071e1L49

akash1810:
👍

PR #14041, opened 2016-08-18
Hosted video pages - HEX into RGB module

Calanthe:
## What does this change?

The next video box on hosted video pages has slightly different background colour than the brand one. We don't want to use another colour variable taken from CAPI, so I created this module which translates HEX into RGB so we can change the transparency value of the background: `background: rgba(RGB, 0.1)` without introducing any new variable.
## Screenshots

Renault before:
screen shot 2016-08-18 at 14 25 17

Renault after (background colour calculated based on the brand colour):
screen shot 2016-08-18 at 14 26 14

Leffe before:
screen shot 2016-08-18 at 14 26 49

Leffe after:
screen shot 2016-08-18 at 14 27 14

Zootropolis before:
screen shot 2016-08-18 at 14 28 27

Zootropolis after:
screen shot 2016-08-18 at 14 29 20
## Request for comment

@guardian/commercial-dev

### DIFF
--- a/commercial/app/views/hosted/guardianHostedVideo.scala.html
+++ b/commercial/app/views/hosted/guardianHostedVideo.scala.html
@@ -34,7 +34,6 @@
 
     .hosted-page .hosted__next-video--tile {
         border-top-color: @{page.campaign.brandColour};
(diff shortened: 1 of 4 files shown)


rich-nguyen:
👍

PR #14043, opened 2016-08-18
Hosted video - autoplay AB test removal

Calanthe:
## What does this change?

The hosted autoplay AB test is removed.
## Request for comment

@guardian/commercial-dev


lps88:
What was the outcome of this test? Are we removing the autoplay completely? If so, we need to delete the markup for it in `guardianHostedVideo.scala.html`

Calanthe:
Good point @lps88 I forgot about the `guardianHostedVideo.scala.html` file.

I don't know yet what is the outcome of this test but if we will need that functionality in the future, everything is saved in the github history.

regiskuckaertz:
Yeah 👍

PR #14044, opened 2016-08-18
make panic shedding permanently on

johnduffell:
we tried turning this off last month for a day, and everything fell over.  We thought we'd try again, and this time it lasted a day:
Earlier applications had a wobble
image
and later discussion went completely down the pan and needed intervention
image

Interestingly, it turns out this pattern of failure has been around since the year dot (2013), at which point it was christened "thrashing" and the explanation degenerated into spiritual incantation in the absence of other explanation.
https://github.com/guardian/frontend/blob/2016df80df5e5cf21a165684ed65817427199648/docs/99-archives/incidents/2013-05-08.md

@guardian/dotcom-platform

### DIFF
--- a/common/app/conf/PanicSheddingFilter.scala
+++ b/common/app/conf/PanicSheddingFilter.scala
@@ -53,10 +53,6 @@ class PanicSheddingFilter extends Filter with Logging {
       )
     }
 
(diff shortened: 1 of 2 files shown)


TBonnin:
👍

PR #14047, opened 2016-08-19
Infer the status code from the server response

piuccio:
## What does this change?

Reqwest in Internet explorer 9 uses XDomainRequest which doesn't have a `status`

If status is missing, infer it from the `responseText`

Comments, breaking news notification and all things ajaxed in work in IE9

No


rich-nguyen:
👍

PR #14048, opened 2016-08-19
Fix: prevent duplicate Audience Science impressions

regiskuckaertz:
## What does this change?

Prevents multiple Audience Science impressions per page. Our contract with Audience Science only allows one impression per page, and it has been estimated we are currently losing 8K£/month because of that.

The problem: Audience Science-related targeting keys are set at the page-level, because any slot can receive an AS creative. And an AS ad is _only_ delivered if some targeting keys are present. Therefore, we need to make sure that those keys are removed as soon as an AS creative is delivered.

The solution: The sooner we know an AS creative is delivered is ... in the creative itself! This creative will call a piece of code we set up in the global context, which will then remove all AS-related targeting keys.

**Small caveat**: this solution is not 100% bulletproof. Because ads are loaded asynchronously based on the user scrolling behaviour, multiple ad calls with the AS keys may still be seen before the first AS ad is rendered. Still, this solution has been accepted by the stakeholders and if it still isn't sufficient, we will make sure DFP calls are made sequentially.

A drastic drop of Audience Science impressions and mo' money in the bank.

Nope
## Request for comment

@guardian/commercial-dev

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/dfp/init.js
+++ b/static/src/javascripts/projects/common/modules/commercial/dfp/init.js
@@ -73,10 +73,10 @@ define([
         window.asiDirectPrequal = function () {
             var pubads = window.googletag.pubads();
             Object.keys(audienceSciencePql.getSegments()).forEach(removeKey);
-            console.log('hey guys, I just removed all Audience Science keys. Regis, how do you do this??! I know, right?');
+
             function removeKey(key) {
                 pubads.clearTargeting(key);
             }
-        }
+        };
     }
 });
--- a/static/src/javascripts/projects/common/modules/commercial/dfp/init.js
+++ b/static/src/javascripts/projects/common/modules/commercial/dfp/init.js
@@ -73,10 +73,9 @@ define([
         window.asiDirectPrequal = function () {
             var pubads = window.googletag.pubads();
             Object.keys(audienceSciencePql.getSegments()).forEach(removeKey);
-            console.log('hey guys, I just removed all Audience Science keys. Regis, how do you do this??! I know, right?');
             function removeKey(key) {
                 pubads.clearTargeting(key);
             }
-        }
+        };
     }
 });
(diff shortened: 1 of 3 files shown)


static/src/javascripts/projects/common/modules/commercial/dfp/init.js
```diff
@@ -45,23 +48,34 @@ define([
...
-            window.googletag.pubads().setTargeting(key, targeting[key]);
+            pubads.setTargeting(key, targeting[key]);
         });
     }
+
+    // Remove all Audience Science related targeting keys as soon as we recieve
+    // an AS creative (will get called by the creative itself)
+    function setAudienceScienceCallback() {
+        window.asiDirectPrequal = function () {
+            var pubads = window.googletag.pubads();
```
rich-nguyen: could you add some tracking so that we know how many times `asiDirectPrequal` is called? Maybe using `beacon`? Also, the function name is a bit cryptic, could it be something like `audienceScienceCreativeLoaded`?

static/src/javascripts/projects/common/modules/commercial/dfp/init.js
```diff
@@ -45,23 +48,34 @@ define([
...
         Object.keys(targeting).forEach(function (key) {
-            window.googletag.pubads().setTargeting(key, targeting[key]);
+            pubads.setTargeting(key, targeting[key]);
         });
     }
+
+    // Remove all Audience Science related targeting keys as soon as we recieve
+    // an AS creative (will get called by the creative itself)
+    function setAudienceScienceCallback() {
+        window.asiDirectPrequal = function () {
```
rich-nguyen: should we call `setAudienceCallback` at all if `dfpEnv.shouldLazyLoad()` is false?

rich-nguyen:
super! 👍  

Do you think we should add a creative template? Or are adops ok with copy pasting? I don't mind either way, but I've realised how hard it is to audit dfp with copy pasted script code.

static/src/javascripts/projects/common/modules/commercial/dfp/init.js
regiskuckaertz: Sure thing!

static/src/javascripts/projects/common/modules/commercial/dfp/init.js
regiskuckaertz: You're right, I will make sure we don't.

regiskuckaertz:
I will ask Sian, I don't know if this is possible since they may be recycling the same line items.

regiskuckaertz:
Sian said they are indeed using the same 18 creatives and she can fly through them quickly. But indeed for Rubicon and Page skins, we should really think about it.

PR #14053, opened 2016-08-19
Fix facebook share count

TBonnin:
## What does this change?

Fix facebook share count

To obtain an article share count we call the FB graph API using[ the v1
format](https://developers.facebook.com/docs/apps/changelog) (non versioned), probably because it doesn't require authentication.
However the response returned by this endpoint seems to have been
following the v2.0 format, which has been retired on August 7th.
I guess the non versioned url we are using is now following the v2.1
format, which seems to be different from what we used to expect.
This patch is handling this new response format as specified at
https://developers.facebook.com/docs/graph-api/reference/v2.1/url

Note: https://developers.facebook.com/docs/apps/changelog

Share counts are not all displays as `0`

No
## Request for comment

@jfsoul @johnduffell @dominickendrick

### DIFF
--- a/static/src/javascripts/projects/common/modules/social/share-count.js
+++ b/static/src/javascripts/projects/common/modules/social/share-count.js
@@ -86,12 +86,12 @@ define([
             var url = 'http://www.theguardian.com/' + config.page.pageId;
             try {
                 ajax({
-                    url: 'https://graph.facebook.com/' + url,
+                    url: 'https://graph.facebook.com/' + url, //TODO: use recent Graph API endpoint format (versioned) https://developers.facebook.com/docs/graph-api/reference/v2.7/url
                     type: 'json',
                     method: 'get',
                     crossOrigin: true
                 }).then(function (resp) {
-                    var count = resp.shares || 0;
+                    var count = resp.share && resp.share.share_count || 0;
                     counts.facebook = count;
                     addToShareCount(count);
                     updateTooltip();


static/src/javascripts/projects/common/modules/social/share-count.js
```diff
@@ -86,12 +86,12 @@ define([
...
             try {
                 ajax({
-                    url: 'https://graph.facebook.com/' + url,
+                    url: 'https://graph.facebook.com/' + url, //TODO: use recent Graph API endpoint format (versioned) https://developers.facebook.com/docs/graph-api/reference/v2.7/url
                     type: 'json',
                     method: 'get',
                     crossOrigin: true
                 }).then(function (resp) {
-                    var count = resp.shares || 0;
+                    var count = resp.share.share_count || 0;
```
jfsoul: This is slightly more likely to throw if facebook change their API again. If we're happy for this to go to the `errorReporter` then it's fine, but if we really want it to default to 0 then `var count = resp.share && resp.share.share_count || 0` is an option.

jfsoul:
One comment that is your choice but otherwise 👍

static/src/javascripts/projects/common/modules/social/share-count.js
TBonnin: Fair point. Will be closer to actual behavior.
I still don't feel comfortable with it failing silently and us relying on CP to report this kind of issues.

johnduffell:
👍 its not a critical service so I don't mind it failing silently... however I think it should only appear on the screen when the count has come in, rahter than staying as 0.

Then, if it breaks and noone cares, we can delete it! 😆 

Having said all that, thanks for fixing!

PR #14055, opened 2016-08-19
Fix branding of paid interactives

kelvin-chappell:
When branding is driven by capi.
This shouldn't affect anything that's already in production.

It will need a bit of work to reposition the logo.

/cc @guardian/labs-beta


rich-nguyen:
👍

PR #14058, opened 2016-08-19
ensures canonical-url data attribute is on all video elements

akash1810:
## What does this change?

In order to have a clear event label in the video events for GA, adds `data-canonical-url` attribute to `video` elements on:
- [x] main media
- [x] article embed
- [x] video pages
- [x] video embed pages
- [x] video cards on fronts

We cannot add this new attribute to main media embeds on fronts as the video id is not available. Rather than blocking progress (as fronts would need updating) we'll fallback to `window.location.pathname` for now as we get little traffic to these videos.

Also using named parameters for explicitness.

Tracking.

No.
## Request for comment

@cb372 @guardian/multimedia

### DIFF
--- a/article/app/views/fragments/mainMedia.scala.html
+++ b/article/app/views/fragments/mainMedia.scala.html
@@ -26,10 +26,11 @@
                                 showControlsAtStart = true,
                                 endSlatePath = EndSlateComponents.fromContent(article.content).toUriPath,
                                 overrideIsRatioHd = None,
-                                embedPath = article.content.mainVideoCanonicalPath
+                                embedPath = Some(article.content.mainVideoCanonicalPath),
+                                path = article.content.mainVideoCanonicalPath
                             )
                         ) { player =>
-                            @fragments.media.video(player, true, amp = amp)
+                            @fragments.media.video(player, enhance = true, amp = amp)
                             <meta itemprop="name" content="@player.title">
                             <meta itemprop="image" content="@player.poster">
                             <meta itemprop="thumbnail" content="@player.poster">
--- a/article/app/views/fragments/mainMedia.scala.html
+++ b/article/app/views/fragments/mainMedia.scala.html
@@ -27,7 +27,7 @@
                                 endSlatePath = EndSlateComponents.fromContent(article.content).toUriPath,
                                 overrideIsRatioHd = None,
                                 embedPath = Some(article.content.mainVideoCanonicalPath),
-                                path = article.content.mainVideoCanonicalPath
+                                path = Some(article.content.mainVideoCanonicalPath)
                             )
                         ) { player =>
                             @fragments.media.video(player, enhance = true, amp = amp)
--- a/article/app/views/fragments/mainMedia.scala.html
+++ b/article/app/views/fragments/mainMedia.scala.html
@@ -26,8 +26,8 @@
                                 showControlsAtStart = true,
                                 endSlatePath = EndSlateComponents.fromContent(article.content).toUriPath,
                                 overrideIsRatioHd = None,
-                                embedPath = Some(article.content.mainVideoCanonicalPath),
-                                path = Some(article.content.mainVideoCanonicalPath)
+                                embedPath = article.content.mainVideoCanonicalPath,
+                                path = article.content.mainVideoCanonicalPath
                             )
                         ) { player =>
                             @fragments.media.video(player, enhance = true, amp = amp)
--- a/common/app/model/content.scala
+++ b/common/app/model/content.scala
@@ -8,7 +8,7 @@ import com.gu.facia.client.models.TrailMetaData
 import common._
 import common.dfp.DfpAgent
 import conf.Configuration
-import conf.switches.Switches.{FacebookShareUseTrailPicFirstSwitch, FacebookShareImageLogoOverlay, TwitterShareImageLogoOverlay, HeroicTemplateSwitch}
+import conf.switches.Switches.{FacebookShareImageLogoOverlay, FacebookShareUseTrailPicFirstSwitch, HeroicTemplateSwitch, TwitterShareImageLogoOverlay}
 import cricketPa.CricketTeams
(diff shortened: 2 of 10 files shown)


common/app/views/fragments/items/facia_cards/contentCard.scala.html
```diff
@@ -50,12 +50,13 @@
...
                     endSlatePath,
-                    Some(false),
+                    overrideIsRatioHd = Some(false),
                     // We only pass in the ID if this is a video or audio block, as if it's a mainMedia video, the
                     // ID is actually pointing to the article, which is wrong. It would be nice to have the ID on the
                     // ContentCard, but we don't for now. Annoyingly this doesn't allow us to check for if we should
                     // play ads or if the video is expired, but as fronts change really often, this is a non-problem.
-                    if (item.isMediaLink) item.id else None
+                    embedPath = if (item.isMediaLink) item.id else None,
+                    path = item.id.getOrElse("foo")
```
akash1810: need to chat to @piuccio about this - [fixing this](https://github.com/guardian/frontend/blob/master/common/app/views/fragments/items/facia_cards/contentCard.scala.html#L54-L57) would help massively here.

akash1810:
WIP because [`foo` is bad](https://github.com/guardian/frontend/pull/14058/files#diff-a2b2ea169f137c124dce5857de42402fR59).

common/app/model/content.scala
```diff
@@ -175,15 +175,15 @@ final case class Content(
...
 
-  lazy val hasMultipleVideosInPage: Boolean = mainVideoCanonicalPath match {
-    case Some(_) => numberOfVideosInTheBody > 0
-    case None => numberOfVideosInTheBody > 1
-  }
+  lazy val mainMediaVideo = Jsoup.parseBodyFragment(fields.main).body.getElementsByClass("element-video")
 
-  lazy val mainVideoCanonicalPath: Option[String] = Jsoup.parseBodyFragment(fields.main).body.getElementsByClass("element-video").headOption.map { v =>
-    new URL(v.attr("data-canonical-url")).getPath.stripPrefix("/")
+  lazy val hasMultipleVideosInPage: Boolean = mainMediaVideo.isEmpty match {
```
akash1810: doing a match on `mainMediaVideo.isEmpty` rather than the contents of `mainVideoCanonicalPath` feels a lot cleaner. It also means `mainVideoCanonicalPath` can become a `String` (previously an `Option[String]`).

article/app/views/fragments/mainMedia.scala.html
```diff
@@ -26,10 +26,11 @@
                                 showControlsAtStart = true,
                                 endSlatePath = EndSlateComponents.fromContent(article.content).toUriPath,
                                 overrideIsRatioHd = None,
-                                embedPath = article.content.mainVideoCanonicalPath
+                                embedPath = Some(article.content.mainVideoCanonicalPath),
```
gidsg: Just wanted to check what will happen if this field is `None`?

common/app/model/content.scala
```diff
@@ -175,15 +175,15 @@ final case class Content(
...
+    case true   => numberOfVideosInTheBody > 1
+    case false  => numberOfVideosInTheBody > 0
```
gidsg: could this just be `case false => _` ?

article/app/views/fragments/mainMedia.scala.html
akash1810: `article.content.mainVideoCanonicalPath` should never be `None` as it comes from CAPI.

common/app/model/content.scala
akash1810: not sure how that works 😢

akash1810:
out of WIP now, if you want to 👀 at it @gidsg @cb372 😄

gidsg:
Looks good to me 👍 🌐

common/app/model/content.scala
gidsg: Could you express the same thing just with
`lazy val hasMultipleVideosInPage: Boolean = mainMediaVideo.isEmpty && numberOfVideosInTheBody > 1`
?

akash1810: Don't think that'll work. The (rather convoluted) logic is counting the number of videos in an article:

if there is one in main media, then check if the body has at least 1 (so total is at least 2)
if there isn't one in main media, then check if the body has more than 1 video

article/app/views/fragments/mainMedia.scala.html
cb372: That sounds a little suspicious. Don't trust anything that comes from CAPI 😉

common/app/model/content.scala
```diff
@@ -175,15 +175,15 @@ final case class Content(
...
   }
 
+  lazy val mainVideoCanonicalPath: String = new URL(mainMediaVideo.attr("data-canonical-url")).getPath.stripPrefix("/")
```
cb372: This will blow up if `mainMediaVideo` is empty:

```
scala> new java.net.URL("")
java.net.MalformedURLException: no protocol:
  at java.net.URL.<init>(URL.java:586)
```

So maybe it should be `Option[String]`? e.g.

```scala
lazy val mainVideoCanonicalPath: Option[String] = mainMediaVideo.attr("data-canonical-url") match {
  case "" => None
  case url => Some(new URL(url).getPath.stripPrefix("/"))
}

common/app/model/content.scala
cb372: I don't understand why `mainVideoCanonicalPath` should not be an option. For an arbitrary piece of content, we don't know whether it has a main video at all, let alone whether that video has a canonical URL.

akash1810: all articles should have a main media video :trollface:

common/app/model/content.scala
akash1810: updated

cb372:
👍 thanks for doing this work!

PR #14059, opened 2016-08-19
Fix fixed positioned element overflow by clipping

regiskuckaertz:
## What does this change?

The Revealer template uses a fixed positioned element inside a canvas. The problem is, the container of fixed positioned elements is the viewport, and so the element overflows its container:

picture 2

Yeah, not cool. Although I really don't know why it worked [in the first instance](https://github.com/guardian/frontend/pull/13692). Anyway, by using the `clip` property on the parent of this element, we can achieve the desired result:

picture 4

And _voila!_

cc @guardian/commercial-dev


rich-nguyen:
👍

PR #14061, opened 2016-08-19
Remove remnant polyfill code from matches

rich-nguyen:
Slight tidy up for code which used to use `matches` or `msMatchesSelector`. This can now use `matches` thanks to the polyfill, but still checks the function is defined.


regiskuckaertz:
yeah! thanks 👍  (see the specs too)

PR #14062, opened 2016-08-19
Fabric width fix

janua:
## What does this change?

This changes the `fabric` creatives to go full width. This is done by moving the slot that will contain the creative to outside of the container; this means that `fabric` creatives will now appear between the top two containers on fronts.

These creatives and slots are unable to go full width on mobile when inside a container.

This still has spacing issues when moves out of the container. See screenshot.

No
## Screenshots

fabricfix
## Request for comment

@regiskuckaertz @rich-nguyen


regiskuckaertz:
👍

PR #14066, opened 2016-08-22
Wrap moved fabric in section

janua:
We want to make sure that the moved ad-slot is nested in a section, as the location it gets moved to has `section` as siblings.

@regiskuckaertz @rich-nguyen


regiskuckaertz:
👍

PR #14067, opened 2016-08-22
Commercial: revealer on desktop/tablet

regiskuckaertz:
So, the position of the revealer creatives on desktop is a little bit off:

picture 6

This is fixed by moving the fixed positioned image closer to the adslot expanded `DOMRect`.

picture 7

cc @guardian/commercial-dev


rich-nguyen:
👍

PR #14069, opened 2016-08-22
Paid interactives - logo position fixed

Calanthe:
## What does this change?

The position of the logo should be on the right column. Just like the logos served via DFP. 
## Screenshots

Before:
screen shot 2016-08-22 at 13 37 33

After:
screen shot 2016-08-22 at 13 37 00
## Request for comment

@kelvin-chappell


kelvin-chappell:
👍

PR #14070, opened 2016-08-22
Return error during get comment process

nicl:
Previously errors were simply swallowed, making it hard to debug prod issues.

(The specific motivation here is a prod issue that we have been unable to replicate locally - relating to braces in comments.)

@johnduffell

### DIFF
--- a/discussion/app/controllers/CommentsController.scala
+++ b/discussion/app/controllers/CommentsController.scala
@@ -144,11 +144,11 @@ class CommentsController(csrfConfig: CSRFConfig, val discussionApi: DiscussionAp
           RevalidatableResult.Ok(views.html.discussionComments.discussionPage(page))
         }
       }
+    } recover {
+      case NonFatal(e) =>
+        log.error(s"Discussion $key cannot be retrieved", e)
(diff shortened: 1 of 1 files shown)


discussion/app/controllers/CommentsController.scala
```diff
@@ -144,11 +144,10 @@ class CommentsController(csrfConfig: CSRFConfig, val discussionApi: DiscussionAp
           RevalidatableResult.Ok(views.html.discussionComments.discussionPage(page))
         }
       }
+    } recover {
+      case NonFatal(e) =>
+        InternalServerError(s"Discussion $key cannot be retrieved, has error: ${e.getMessage}")
```
johnduffell: I would be a little wary about letting an error message out, also it won't give all the info.  May be worth keeing the message the same but sending the exception to the logger, we can view the logs with kibana so it's quite easy.

nicl: @johnduffell ah okay. I didn't realise logs were so easy to view in Frontend. Okay great - will update.

johnduffell:
👍 thanks! 🚀

PR #14072, opened 2016-08-22
Adblocking block test

JustinPinner:
## What does this change?

Adds an ABC test to trial messaging to adblock users. Initially a zero percent test with a ten minute grace period for testing. 

We need to reach out to our readers who use adblockers effectively and without upsetting anyone. We're going to have to make sure this works as well as possible and with known detection limitations, hence the need to pre-test prior to activation.

No.
## Screenshots

**Non-dismissable**
adblocknondismissable

**Delayed dismiss**
adblockdelayeddismissable

**Immediate dismiss**
adblockimmediatelydismissable
## Request for comment

@regiskuckaertz @rich-nguyen @ScottPainterGNM


commercial/app/views/subscriberNumberPage.scala.html
```diff
@@ -0,0 +1,25 @@
...
+        <div class="l-side-margins">
+            <div class="live-events-survey-wrapper" data-link-name="subscriber number page">
+                <div class="live-events-survey-wrapper-inner">
+                    <h1>Please enter your current subscription number below:</h1>
+                    <div class="subscriber-number">
+                        <form class="subscriber-number-form js-subscriber-number-form">
+                            <input type="text" name="number" id="number" class="input-number"/>
+                            <button type="submit" name="submit" class="button--primary">Submit</button>
+                        </form>
+                        <p class="subscriber-number-correct standfirst js-subscriber-number-correct u-h">
```
regiskuckaertz: I think it's better if using `hidden` instead of `u-h` in terms of accessibility. The former won't be read by voice-based UA-s.

static/src/javascripts/projects/common/modules/experiments/tests/adblocking-response.js
```diff
@@ -0,0 +1,158 @@
...
+    return function () {
+        this.id = 'AdBlockingResponse';
+        this.start = '2016-08-17';
+        this.expiry = '2016-10-18';
+        this.author = 'Justin Pinner';
+        this.description = 'Adblocking response ZERO PERCENT TEST WITH 10 MINUTES GRACE';
+        this.audience = 0;
+        this.audienceOffset = 0;
+        this.successMeasure = 'Adblocking users will either pay, prove they have a subscription or turn off their adblocker.';
+        this.audienceCriteria = 'All non-Firefox users with active adblockers.';
```
jbreckmckye: Why exclude Firefox? The only issue with FF is that FF ABP can't be detected.

We had a means to validate that it was _installed_, but this would return true when the user had whitelisted us, so we made a clear distinction between `adblockInUse` and `adblockerInstalled`. However, the latter no longer exists as it was deemed a potential privacy encroachment, so we just let FF ABP users evade our detection.

Not sure if including FF users would give you better results / a larger sample, but thought it worth mentioning...

jbreckmckye:
Heads up: this won't work on pretty much any mobile adblocker AFAIK. Everything we tested - Peace, 1Blocker, etc. - wasn't detected by the synchronous blocker check.

Testing adblockers is not straightforward, because they are effectively isolated processes running in separate threads, and this is something commercial should really invest some time into if they aren't already.

commercial/app/views/subscriberNumberPage.scala.html
JustinPinner: Ok thanks. Is `hidden` a replacement for `u-h` across the board?

static/src/javascripts/projects/common/modules/experiments/tests/adblocking-response.js
JustinPinner: Thanks for the feedback! We'll undoubtedly expand this at some point, but one thing we don't want to get into is prompting when it's not needed.

paperboyo:
This copy is as verbose as the stuff I write! 😆 I would _at least_ drop “world-class” (not the time, nor the place). Also, the close button looks non-standard.

👋 Jimmy!

Sent from Firefox

gtrufitt:
oh goodness

commercial/app/views/subscriberNumberPage.scala.html
regiskuckaertz: No, they are two opposite thing. The former completely hides an element from the DOM, as in "it doesn't exist" in the rendering tree _and_ the accessibility tree. The latter, just hides it from the rendering tree (in practice, it is cleverly transformed in such a way as not to be visible, but that is just details)

paperboyo:
👍  https://webmasters.googleblog.com/2016/08/helping-users-easily-access-content-on.html

johnduffell:
I really like the principle, but there are a lot of words there which makes it difficult to work out what to do.  Given most people just want to read the article, can we make that flow easier?

paperboyo:
> Given most people just want to read the article, can we make that flow easier?

Agreed! Unfortunately, the whole point of this is that the commercial considerations go against this natural flow here. In the case of the very worst - the Non-dismissable variant - they dry the flow dead. I personally deserted more than couple of publications that blackmailed me in this way. I do hope this variant will die a quick death.

JustinPinner:
Ok, can we put copy issues aside for a moment, and say if we think the code is good enough to release as a zero percent test, so that we can get more eyes on it and fine-tune the words? If so, please can someone 👍  it so we can get it rolling? Big thanks for all the feedback given so far 🎸

static/src/javascripts/bootstraps/enhanced/common.js
```diff
@@ -177,9 +179,7 @@ define([
             },
 
             cleanupLocalStorage : function () {
-                var deprecatedKeys = [
-                    'gu.subscriber'
-                ];
+                var deprecatedKeys = [];    //TODO: reinstate gu.subscriber after completion of ab-adblocking-response test
```
johnduffell: add a link to this PR to the comment, just in case it gets left behind?

static/src/javascripts/projects/common/modules/experiments/subscriber-number-form.js
```diff
@@ -0,0 +1,58 @@
...
+        $numberInput.addClass('incorrect');
+        $numberInput.removeClass('correct');
+    }
+
+    function submitForm($form, event) {
+        event.preventDefault();
+
+        var $numberInput = $('.input-number', $form),
+            numberVal = $numberInput.val(),
+            isSubscriber = /^(?=\S{8,11}$)(00|GA|A-S)\S+/.test(numberVal),
```
johnduffell: 😆

static/src/javascripts/projects/common/modules/experiments/tests/adblocking-response.js
```diff
@@ -0,0 +1,156 @@
...
+) {
+    return function () {
+        this.id = 'AdBlockingResponse';
+        this.start = '2016-08-17';
+        this.expiry = '2016-10-18';
+        this.author = 'Justin Pinner';
+        this.description = 'Adblocking response ZERO PERCENT TEST WITH 10 MINUTES GRACE';
+        this.audience = 0;
+        this.audienceOffset = 0;
+        this.successMeasure = 'Adblocking users will either pay, prove they have a subscription or turn off their adblocker.';
```
johnduffell: that's not quite a measure, it's an objective

static/src/javascripts/projects/common/modules/experiments/tests/adblocking-response.js
```diff
@@ -0,0 +1,156 @@
...
+            ['user help email immediate-close'],
+            ['adblock whitelist delayed-close'],
+            ['adblock membership delayed-close'],
+            ['subscriber number delayed-close'],
+            ['user help email delayed-close']
+        ].map(function(_) {
+            return _[0];
+        }).join(', ');
+        this.dataLinkNames = 'adblock response overlay: '+variantDataLinkNames+', subscriber number page: user help email';
+        this.idealOutcome = 'Adblocking users will either pay or allow ads.';
```
johnduffell: should be more specific - if this is actually an AB test to check something (in a statistically useful way) you should talk to the data team.  If it's just an experiment where you're not going to use the stats just to get opinions, you needn't plan an AB test

static/src/javascripts/projects/common/modules/experiments/tests/adblocking-response.js
```diff
@@ -0,0 +1,156 @@
...
+
+        this.canRun = function () {
+            return !contains('firefox', detect.getUserAgent.browser.toLowerCase())
+                && contains(['desktop', 'leftCol', 'wide'], detect.getBreakpoint())
+                && config.page.edition === 'UK'
+                && !cookies.get('gu_abm_x');
+        };
+
+        this.variants = [{
+            id: 'no-close',
```
johnduffell: there's no control group, depending what you are baselining against you might need one, but that depends on the answers to the above (and I'm guessing this is more an experiment than an AB test)

static/src/javascripts/projects/common/modules/experiments/tests/adblocking-response.js
```diff
@@ -0,0 +1,156 @@
...
+            test: function () {
+                detect.adblockInUse.then(function (adblockUsed) {
+                    if (adblockUsed && !config.page.isFront &&
+                        !userFeatures.isPayingMember() &&
+                        config.page.webTitle !== 'Subscriber number form' &&
+                        config.page.webTitle !== 'How to disable your adblocker for theguardian.com' &&
+                        !storage.local.get('gu.subscriber')) {
```
johnduffell: be careful of users with no/broken local storage - will they see it every time?  we have this problem with breaking news banners showing up repeatedly so it does happen

static/src/javascripts/projects/common/modules/experiments/subscriber-number-form.js
JustinPinner: I know! Right?!

static/src/javascripts/projects/common/modules/experiments/tests/adblocking-response.js
gtrufitt: Indeed, before using `get` you should use `isStorageAvailable` and if `false`, don't do your stuff.

static/src/javascripts/projects/common/modules/experiments/tests/adblocking-response.js
regiskuckaertz: in all fairness, the folder is called `experiments`, and there's hardly any other way to experiment at that scale than with some form of AB testing.

static/src/javascripts/projects/common/modules/experiments/tests/adblocking-response.js
regiskuckaertz: I wonder if this is due to the vagaries of ad blocker detection. the "no control group" will contain lots of false negatives and thus their behaviour is not a safe baseline.

regiskuckaertz:
👍  for the code, what about the control group q?

static/src/javascripts/projects/common/modules/experiments/tests/adblocking-response.js
johnduffell: of course - I'm not arguing terminology, I mean to say by all means ship it, but you can't analyse the data to draw any meaningful conclusion.  Or to put it another way, this is an experiment, but not an AB test.
Having said that I went to talk to Justin in person to explain what I actually mean =)

PR #14074, opened 2016-08-22
Change colour of headline on immersive articles

sammorrisdesign-zz:
I'm pretty sure we would always want these headlines to be white on immersive articles.
## Screenshots

Before:
image

After:
image

### DIFF
--- a/static/src/stylesheets/module/content/_article-immersive.scss
+++ b/static/src/stylesheets/module/content/_article-immersive.scss
@@ -112,15 +112,10 @@
     font-weight: 200;
 }
 
-.immersive-main-media__headline-container--dark {
-    .content__headline--immersive-article {
-        color: #ffffff;
-    }
-}
-
 .content__headline--immersive-article {
     @include fs-headline(7, true);
     line-height: 1;
+    color: #ffffff;
 
     @include mq(desktop) {
         font-size: 4rem;


jfsoul:
Careful with this one, looks like it will undo a fix we put in a couple of months back.

https://github.com/guardian/frontend/pull/13443

jfsoul:
I think you might be right, but we explicitly specify the colour for `.immersive-main-media__headline-container--dark` to make sure it is specific enough to override the colour from any tone that is applied.  I also originally included the `.content--immersive-article-without-main-media` (which looks like the screenshots you have?) but @NataliaLKB recommended to leave out as it would normally be overridden by a tone (IIRC).

Which headlines are coming out dark on dark, out of interest?

jfsoul:
You should be able to recreate the original bug with this article http://www.theguardian.com/politics/commentisfree/2016/jun/24/lifelong-english-european-the-biggest-defeat-of-my-political-life-timothy-garton-ash-brexit.

Change the `tonal--tone-feature` on one of the headline containers to `tonal--tone-comment`.  Does it come out dark on dark with your changes?

sammorrisdesign:
Ah, so yeah I've tried that and I get dark text. However, seeing as we always want light text here regardless of `--dark` or the `tonal--tone`, this might be a case for important.

The only reason the previous fix worked was because of specificity. specificity that has excluded my original case.

static/src/stylesheets/module/content/_article-immersive.scss
```diff
@@ -112,15 +112,10 @@
     font-weight: 200;
 }
 
-.immersive-main-media__headline-container--dark {
-    .content__headline--immersive-article {
-        color: #ffffff;
-    }
-}
-
```
jfsoul: Can we just re-add this block, we'll get best of both worlds - no overriding via tone for this specific case, and a sensible default in the general case?  Don't know your case that well, so it's a bit of a guess.

sammorrisdesign: But as far as I can tell, we always want it to be white. So this fix doesn't work, as if something doesn't get `--dark` it breaks.

After doing some digging I'm now unsure if we even need `--dark`, it doesn't override or add anything that we wouldn't want all templates getting.

jfsoul: Yeah that's fair enough, feel free to take a closer look at this.  The thinking for keeping my PR super low-impact was that there were already plans to potentially refactor some of this so tones/immersives wouldn't override each other in this way.  That's probably fairly long-term though, so shouldn't stop you from going ahead.

PR #14076, opened 2016-08-22
Fix encoding issue with fb share links

nicl:
Hopefully fixes issue with Facebook social share links. I'm not sure the consequences of encoding the quote yet fully so would be good to get thoughts here.

@johnduffell @marialivia16


johnduffell:
> Hopefully

> I'm not sure

these words always worry me!  Since security depends on correct encoding, it would be nice to understand why it stops encoding at the quote.  Is a XSS attack possible by injecting some terminating character into a comment? @adamnfish do you have any thoughts re security?

nicl:
@johnduffell definitely! Which is why I used those words. This is not ready to be merged until we understand this better (although it is definitely an encoding issue). I wanted to get thoughts from yourself and also Maria.

common/app/model/ShareLinks.scala
```diff
@@ -93,7 +93,7 @@ object ShareLinks {
     lazy val facebookParams = List(
       Some("app_id" -> facebookAppId),
       Some("href" -> encodedHref),
-      quote.map(q => "quote" -> q),
+      quote.map(q => "quote" -> q.urlEncoded),
       mediaPath.map(path => "picture" -> path.urlEncoded)
```
marialivia16: The same encoding is necessary for the picture. Is this related?

nicl: Basically, if you don't urlencode this part of it, then the you get an error like:

    java.net.URISyntaxException: Illegal character in query at index 194: https://www.facebook.com/dialog/share?app_id=157221981065719&href=https%3A%2F%2Fdiscussion.code.dev-theguardian.com%2Fcomment-permalink%2F64283254&quote=niclong+commented:+%22breaking+things+%7B"

I've managed to replicate the error locally. 

So the bug happens:

- when social sharing links are ON for comments
- and the discussion contains at least one comment in the comments being requested that contains an invalid character

where 'invalid character' means anything which is not allowed in URLs and not encoded.

nicl:
@johnduffell Re security, you can put any valid (urlencoded) html in the text. But Facebook will HTML escape stuff before sharing it with their users.

common/app/model/ShareLinks.scala
johnduffell: something still isn't right.  We're calling a method on a uri builder to add parameters, and it's falling over because it's failing to escape them properly?
https://github.com/guardian/frontend/blob/bf6421b383c4eb7a0aba68a13e8bedfff35d0420/common/app/implicits/Strings.scala#L50
according to the docs, it should be escaping them
http://docs.oracle.com/javaee/7/api/javax/ws/rs/core/UriBuilder.html
> Builder methods perform contextual encoding of characters not permitted in the corresponding URI component following the rules of the application/x-www-form-urlencoded media type for query parameters and RFC 3986 for all other components. Note that only characters not permitted in a particular component are subject to encoding

nicl:
The quote is a red-herring. I think that is just how the URI library reports things when it gets abused.

common/app/model/ShareLinks.scala
johnduffell: so actually it's not to do with url encoding at all, actually { and } are special substitution characters for the uribuilder.  So those need to be escaped (not clear how though), not the " etc.  The UriBuilder will do the rest.  The escaping of `{` and `}` should be done in the `appendQueryParams` method, not in every piece of code that calls it.

common/app/implicits/Strings.scala
```diff
@@ -46,11 +46,12 @@ trait Strings {
     }
   }
 
+  // Note, this method encodes parameter values so you don't have to
```
johnduffell: 😆 technically it encodes UriBuilder substitutions i.e. { and } so you don't have to, and the rest of the url characters so UriBuilder doesn't have to

common/app/implicits/Strings.scala
```diff
@@ -46,11 +46,12 @@ trait Strings {
...
   implicit class String2Uri(uri: String) {
     def appendQueryParams(queryParams: Map[String, String]): String = {
       queryParams.foldLeft(uri)((currentUri, queryParam) => {
         javax.ws.rs.core.UriBuilder.fromUri(currentUri)
-          .queryParam(queryParam._1, queryParam._2)
+          .queryParam(queryParam._1, queryParam._2.urlEncoded)
```
johnduffell: great, thanks for that! 🚀

johnduffell:
👍 finally we can put this brackets issue to bed! thanks 🎉

nicl:
@johnduffell thanks. I've improved the comment to have a bit more detail.

johnduffell:
go for it! thanks =)

johnduffell:
I think your test is failing because the `:` in `with:2#2` is getting excaped twice

marialivia16:
Great work, thanks for investigating this! 👍

PR #14077, opened 2016-08-22
GA: send internal link click events to correct tracker

cb372:
## What does this change?

Sends internal link click events to the correct GA tracker. 

In-page links and external links are handled elsewhere and have already been switched over to the prod tracker in #14013

whatevs

no
## Screenshots

n/a
## Request for comment

@ajosephides @dominickendrick


ajosephides:
👍

PR #14081, opened 2016-08-23
Config from S3

johnduffell:
skip first section if you're just reviewing the PR
## What do I need to know as a dev?

If you're using frontend locally, your `~/.gu/frontend.properties` file will no longer work.  If you don't have any special properties, just delete it.  If you want to override things, create a [HOCON](https://github.com/typesafehub/config/blob/master/HOCON.md) file `~/.gu/frontend.conf` and give it the content as you wish, e.g.

```
devOverrides {
    switches.key=DEV/config/switches-yourname.properties
    facia.stage=CODE
}
```
## What does this change?

This makes frontend download config from S3 at runtime instead of from local files/copied from platform repo.
This follows on from the first PR which has more explanation of the why and motivations:
https://github.com/guardian/frontend/pull/13749
See the platform PR for other information and the config changes: https://github.com/guardian/platform/pull/1052

This means that we no longer have any secrets anywhere they shouldn't be.  This means in the event of a loss of data, unless we lose S3 data we don't need to worry.
This also makes it easier to add properties or rotate keys as we only need to do it in one place and it updates for everyone.

The downside is it's S3, although it's cheap and reliable, it's hard to edit files in place.  We would really do well to have a tool to edit stuff.

@guardian/dotcom-platform @alexduf @rich-nguyen comments if you can!


common/app/common/configuration.scala
```diff
@@ -20,21 +20,56 @@ import scala.util.{Failure, Success, Try}
...
+      val stack = p.getProperty("stack", "frontend")
+      // if got config at app startup, we wouldn't need to configure it
+      val app = p.getProperty("app", "dev-build")
+      val stage = p.getProperty("STAGE", "DEV")
+      val region = p.getProperty("region", "eu-west-1")
+      val configBucket = p.getProperty("configBucket", "aws-frontend-store")
+      if (stage == "DEV" && new File(s"${System.getProperty("user.home")}/.gu/frontend.properties").exists) {
+        throw new RuntimeException(
+          "\n\nYou have a file ~/.gu/frontend.properties with secrets - please delete that file and any copies as it is not needed.\n  " +
+            "All secrets are now stored in AWS, not on your laptop.\n  " +
```
TBonnin: ...now stored in *an AWS S3 bucket*, ...

common/app/common/configuration.scala
```diff
@@ -20,21 +20,56 @@ import scala.util.{Failure, Success, Try}
...
+            "Any non-secret preferences can be added to a new file ~/.gu/frontend.conf. \n" +
```
TBonnin: As discussed face to face, making this less encouraging for people to override property unless it is necessary for dev purpose.

common/test/common/AmpValidityTest.scala
```diff
@@ -30,16 +31,22 @@ trait AmpValidityTest extends FlatSpec with Matchers with ConfiguredTestSuite wi
...
 
         // Pass the content to the command line tool (external process) via stdin ('-' option)
         val process = s"node_modules/.bin/amphtml-validator --validator_js ${f.getAbsolutePath} -".run(io)
 
-        withClue(s"AMP validator for $url should pass.\nHint: Try checking your browser developer console for errors when appending '#development=1' to the failing URL.\nThe validator process exit value of ") {
-          process.exitValue() should be(0)
+        val exitValue = process.exitValue() // side effect - await for process to finish
+
+        withClue(s"AMP validator for $url should pass.\nHint: Try checking your browser developer console for errors when appending '#development=1' to the failing URL.\nOutput:\n---------\n${output.reverse.mkString("\n")}\n---------\nThe validator process exit value of ") {
+          exitValue should be(0)
```
TBonnin: Not sure why this change is in this PR

TBonnin:
Do we still need the properties files in the codebase (other than the DEVINFRA one)? https://github.com/guardian/frontend/blob/master/common/conf/env/PROD.properties

common/test/common/AmpValidityTest.scala
johnduffell: oh maybe I could pull that out, basically that test didn't report the stdout so I couldn't find the failure reason.

TBonnin: pinging @jfsoul 😁

jfsoul: What was failing? I turned off stdout/stderr logging for this because the logging was happening in the wrong order by default (and amp validation failures are easy to check in the browser or in various web UIs).  By including it in the clue you are avoiding the problem altogether though.

Wouldn't using a `StringBuffer` rather than a `List[String]` be a bit neater though?

TBonnin:
lgtm so far

We need to update the 'how to install steps' doc and `setup.sh` to reflect this change since downloading manually the properties file is not necessary anymore

alexduf:
I had a look, everything looks sensible so far! :+1:

docs/01-start-here/01-installation-steps.md
```diff
@@ -33,8 +33,6 @@ Prerequesites:
 1. Clone repository
     - `git clone git@github.com:guardian/frontend.git && cd frontend`
 2. Get AWS Credentials using [janus](https://janus.gutools.co.uk/) for frontend and CMS fronts 
-3. Get properties file
-    - `aws s3 cp --profile frontend s3://aws-frontend-store/template-frontend.properties "~/.gu/frontend.properties"`
 4. Run the container (dev service). The first time you do so the container image would be pulled from the EC2 Container Registry
     - ./dev.sh
 5. Build and run the app within the container
```
TBonnin: You'll need to decrement number for following bullet points 😁

TBonnin:
👍

rich-nguyen:
I've just realised a few things since using this system:
- Properties which aren't sensitive are no longer versioned in Git, so I can't see commit history anymore.
- It's not easy to see which properties are sensitive or not. 
- The HOCON syntax is ok, getting used to it, but there is far too much property for one file.

The last two issues are easy to fix (add a `sensitive` var to the s3 conf, split the conf into shared and per-stage files). Any thoughts on the first, @johnduffell ?

johnduffell:
@rich-nguyen thanks for the comments, that's very true.
@dominickendrick has already hit a versioning issue and he's going to make a versioning system and a script to edit and create a new version.
1. Indeed, in fact if it's in the repo with the code, it might as well be scala, because then at least we get compile time checking (and scala is good for DSLs).  Also we can group them by functional area in the right package just like the actual code.  So maybe I could have done a pre-PR to refactor the DEV.properties etc into scala rather than just pulling it into one external file.  It wouldn't be too hard to revert the dev/prod/code.properties files back into the repo if that's a better middle ground and it wouldn't go against the security objective.
2. Convention would be nice, indeed runtime parameters are now following the convention of being in the "parameters" block, sensitive could be similar, or could use a suffix instead of a prefix.
3. Definitely it's way to big, it needs refactoring/cleaning, it's just cut and pasted to get us going.  Once the sensible split is clear, we should definitely do so.  My preference would be to break it down by functional area, so config is grouped in the same way as the scala packages.  That assumes we have a cohesive enough package structure.

I created a trello card with potential future improvements/nice to have, everyone should feel free to update it (or just ship some PRs!) https://trello.com/c/BLiE3txV/72-config-in-s3-improvements

janua:
This is still a complete pain to work with. I have been working with @gidsg on some facia stuff. Looking up properties is a real faff and a complete time sink for just looking up a string, and when we finally got the properties file we thought was being used the property we wanted to change wasn't even in the file.

This isn't the only time I have had a bust up with this new method of doing properties. It's continually a faff, why can't non secrets just be in git? That way we don't need versions for a lot of the changes and the version is just a TeamCity build.

@johnduffell Have we thought about putting back non secrets as scala? That would be perfect. Maybe it's already done and that's why I couldn't find the property?

I think the points in the trello card are good. You could do point 1, 2, 3 and 4 if you changed it back into scala. With point 4 being that what is brought back into git the non-secret stuff, and point 1 being done by some nice deduplication whilst you do it.

Thoughts?

TBonnin:
@janua 
I agree that working with config properties today is painful. Ideally config values should be:
1. all in the same place (so you don't waste time looking up)
2. easy to access
3. easy to modify
4. store "securely" (ie: it is not required for devs to have a copy on their machine)

Today only points 1 and 4 are kinda fulfilled. Accessing and modifying them is painful and error-prone. 

@dominickendrick @stephanfowler: Since config management seems to become a significant problem (Francis is not the only one to point out this issue) we might need to bring it to the top of our list. 🎯

PR #14082, opened 2016-08-23
Enable code capi to be used in code environment

kelvin-chappell:
Use content that's in both code and prod capi for healthchecks, so that we can point the code environment at code capi occasionally for testing.

Unfortunately the onward app uses ophan popularity data for its healthcheck, so it's difficult to make this work with code capi.  We'll just have to accept that it will break on code temporarily.

/cc @johnduffell


article/app/controllers/HealthCheck.scala
```diff
@@ -6,5 +6,5 @@ import play.api.libs.ws.WSClient
 class HealthCheck(override val wsClient: WSClient) extends AllGoodCachedHealthCheck(
   wsClient,
   9004,
-  NeverExpiresSingleHealthCheck("/world/2012/sep/11/barcelona-march-catalan-independence")
```
johnduffell: I like the way you kept the same key words ;)

johnduffell:
great idea, although I guess you need to merge in master! 👍

PR #14083, opened 2016-08-23
Extended contribute in-article test

markjamesbutler:
## What does this change?

Extend test as we didn't get enough impressions to get results.
#13846

No


katebee:
👍  🚢

PR #14084, opened 2016-08-23
Amp liveblog - bug fix

gtrufitt:
## What does this change?

Fixes a validation error where the element with `on="tap:live-list.update"` is expected to have the semantics of a button. Misleadingly the validator suggests adding a role and tabindex to the element, but instead making the element a `button` fixes the validation error. This required some minor CSS fixes.

Fixes AMP liveblog validation errors.

AMP only.
## CC

@stephanfowler

@SiAdcock @NataliaLKB - Lets review this when you get back!


gtrufitt:
@guardian/dotcom-platform

johnduffell:
fair enough, good fixing 👍

SiAdcock:
LGTM dude 👍

PR #14086, opened 2016-08-23
YANK: Reserve space for comments and save for later

gtrufitt:
## What does this change?

It makes sure that comments, shares and SFL don't jump the page.

Instead of this:

jumping-meta-before

You get this:

jumping-meta

_Note:_ This does mean that no-js and non-modern browsers get some space that is never filled... but this is far better than the alternative.
## Request for comment

@guardian/dotcom-platform

### DIFF
--- a/common/app/views/fragments/contentMeta.scala.html
+++ b/common/app/views/fragments/contentMeta.scala.html
@@ -39,7 +39,7 @@
                 }
             </div>
 
-            <div class="meta__numbers modern-visible">
+            <div class="meta__numbers">
                 <div class="u-h meta__number js-sharecount">
                 </div>
                 <div class="u-h meta__number" data-discussion-id="@item.content.discussionId" data-commentcount-format="content" data-discussion-closed="@{
--- a/static/src/stylesheets/module/content/_content.global.scss
+++ b/static/src/stylesheets/module/content/_content.global.scss
@@ -113,8 +113,7 @@
 }
 
 .meta__save-for-later {
-    // Shown with test
-    display: none;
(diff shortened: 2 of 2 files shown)


johnduffell:
3 lines? now that's the low hanging fruit ;) 👍 🚀

ScottPainterGNM:
Sweet moves.....if the SFL icon was up next to the share buttons that would save EVEN more room, right gang?

gtrufitt:
@ScottPainterGNM Hmm, it would Scott, it would. Do you mean without the text? Gonna merge this but lets discuss that further.

ScottPainterGNM:
That's what I was thinking, lets discuss!

rich-nguyen:
This is on the boundary of acceptable progressive enhancement. Whenever possible, the page should look good without enhancement. That's particularly relevant for any yank-fixing work. And scale this solution to other situations, do we want to accept a general approach that means we should always leave whitespace for future components? Probably not.

The wider issue of an improved meta promises a more long term solution, however, so I think we should be in a better position eventually.

gtrufitt:
> This is on the boundary of acceptable progressive enhancement.

Agreed. This is the lesser of two evils, until we can revisit the meta - yank is far worse a curse than some extra space at this point and is *far* more annoying and affects *far* more users. BetaMeta is TBDiscussed.

> do we want to accept a general approach that means we should always leave whitespace for future components?

I definitely don't think it's a general solution. There are situations where, with the current design and layout, it's the most feasible solution but we're trying to look at things case by case and if things need a design re-think, that's a part of the process.

PR #14087, opened 2016-08-23
AMP: Bugfix: Check for amp for default rich-link styles

gtrufitt:
## What does this change?

Bugfix caused by #14032

Richlinks look right on AMP again.
## Screenshots

Broken: 

screen shot 2016-08-23 at 13 22 19

Fixed:

screen shot 2016-08-23 at 13 21 54
## Request for comment

@guardian/dotcom-platform

### DIFF
--- a/article/app/views/package.scala
+++ b/article/app/views/package.scala
@@ -55,7 +55,7 @@ object BodyCleaner {
       DropCaps(article.tags.isComment || article.tags.isFeature, article.isImmersive),
       ImmersiveHeaders(article.isImmersive),
       FigCaptionCleaner,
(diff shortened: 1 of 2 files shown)


johnduffell:
great 👍 although additionally it would be nice to make the amp links look as much like the normal ones as possible.  This might become even more doable if we're applying the same anti jumping techniques to normal pages!

jfsoul:
👍  might be worth adding some articles with rich links to the amp validation tests if we don't already have them

gtrufitt:
Sure thing - I'll merge this bugfix and then look at the default styling.

PR #14088, opened 2016-08-23
Update clashing test name given test name update.

markjamesbutler:
## What does this change?

Given the In Article test has been renamed I've updated the clashing tests name.  (I should have done this in my last PR.)

Nothing its a functional fix.

No


gtrufitt:
👍

PR #14090, opened 2016-08-23
Contributions embed abtest

jranks123:
## What does this change?

This is a test which adds the contributions interactive embed at the end of articles. This embed is inserted as an interactive. The way it is implemented is as follows:
1. If the user is in the test, insert the following <figure> element to the end of the article, which is invisible until explicitly rendered: 
2. Run a script which looks for this specific interactive and renders it. 

What makes this contributions ad different from those previous is that it asks the user to select their contribution amount in the ad itself, which propagates through to the contributions form 

The value is we can evaluate the effectiveness of asking for the contribution amount in the ad. We are measuring the success by looking at the impression to contribution rate. 

No
## Screenshots

The embed on desktop

screenshot at aug 23 13-55-02

The embed on mobile
screenshot at aug 23 13-59-13

Note: the design is set to change, but this doesn't affect this pull request as the interactive embed is built and deployed separately from frontend.

### DIFF
--- a/common/app/conf/switches/ABTestSwitches.scala
+++ b/common/app/conf/switches/ABTestSwitches.scala
@@ -64,6 +64,16 @@ trait ABTestSwitches {
     exposeClientSide = true
   )
 
+  val ABContributionsEmbed20160822 = Switch(
+    SwitchGroup.ABTests,
+    "ab-contributions-embed-20160822",
+    "Test contributions embed with amount picker.",
+    owners = Seq(Owner.withGithub("jranks123")),
+    safeState = Off,
+    sellByDate = new LocalDate(2016, 8, 26),
+    exposeClientSide = true
+  )
+
   val ABParticipationDiscussionOrderingLiveBlogs = Switch(
     SwitchGroup.ABTests,
     "ab-participation-discussion-ordering-live-blog",
--- a/common/app/conf/switches/ABTestSwitches.scala
+++ b/common/app/conf/switches/ABTestSwitches.scala
@@ -70,7 +70,7 @@ trait ABTestSwitches {
     "Test contributions embed with amount picker.",
     owners = Seq(Owner.withGithub("jranks123")),
     safeState = Off,
-    sellByDate = new LocalDate(2016, 8, 26),
+    sellByDate = new LocalDate(2016, 8, 27),
     exposeClientSide = true
   )
 
--- a/common/app/conf/switches/ABTestSwitches.scala
+++ b/common/app/conf/switches/ABTestSwitches.scala
@@ -64,13 +64,13 @@ trait ABTestSwitches {
     exposeClientSide = true
   )
 
-  val ABContributionsEmbed20160822 = Switch(
+  val ABContributionsEmbed20160823= Switch(
     SwitchGroup.ABTests,
-    "ab-contributions-embed-20160822",
+    "ab-contributions-embed-20160823",
     "Test contributions embed with amount picker.",
     owners = Seq(Owner.withGithub("jranks123")),
     safeState = Off,
-    sellByDate = new LocalDate(2016, 8, 27),
+    sellByDate = new LocalDate(2016, 8, 30),
     exposeClientSide = true
   )
 
--- a/static/src/javascripts/projects/common/modules/experiments/tests/contributions-embed.js
+++ b/static/src/javascripts/projects/common/modules/experiments/tests/contributions-embed.js
@@ -1,0 +1,95 @@
+define([
+    'bean',
+    'qwery',
+    'common/utils/$',
+    'common/utils/template',
+    'common/views/svg',
+    'common/utils/fastdom-promise',
+    'common/utils/mediator',
+    'text!common/views/contributions-embed.html',
+    'text!common/views/contributions-header.html',
+    'common/views/svgs',
+    'common/modules/article/space-filler'
+], function (bean,
+             qwery,
+             $,
+             template,
+             svg,
+             fastdom,
+             mediator,
+             contributionsEmbed,
+             contributionsHeader,
+             svgs,
+             spaceFiller
+) {
+
+
+    return function () {
+
+        this.id = 'ContributionsEmbed20160822';
+        this.start = '2016-08-22';
+        this.expiry = '2016-08-26';
+        this.author = 'Jonathan Rankin';
+        this.description = 'Test contributions embed with amount picker.';
+        this.showForSensitive = false;
+        this.audience = 0.13;
+        this.audienceOffset = 0;
+        this.successMeasure = 'Determine the effectivness of allowing user to pick contribution amount in the cta';
+        this.audienceCriteria = 'All users';
+        this.dataLinkNames = '';
+        this.idealOutcome = '';
+        this.canRun = function () {
+            var pageObj = window.guardian.config.page;
+            return !(pageObj.isSensitive || pageObj.isLiveBlog || pageObj.isFront || pageObj.isAdvertisementFeature) && pageObj.edition === 'UK';
+        };
+
+        var getSpacefinderRules = function () {
+            return {
+                bodySelector: '.js-article__body',
+                slotSelector: ' > p',
+                minAbove: 200,
+                minBelow: 150,
+                clearContentMeta: 50,
+                fromBottom: true,
+                selectors: {
+                    ' .element-rich-link': {minAbove: 100, minBelow: 100},
+                    ' > h2': {minAbove: 200, minBelow: 0},
+                    ' > *:not(p):not(h2):not(blockquote)': {minAbove: 35, minBelow: 200},
+                    ' .ad-slot': {minAbove: 150, minBelow: 200}
+                }
+            };
+        };
+
+        var writer = function () {
+            var $embed = $.create(template(contributionsEmbed, {
+                linkText : 'give to the Guardian',
+                linkHref : 'https://contribute.theguardian.com/uk?INTCMP=co_uk_head_give_coin',
+                coinSvg: svgs('contributionsCoin', ['rounded-icon', 'control__icon-wrapper'])
+            }));
+
+           return spaceFiller.fillSpace(getSpacefinderRules(), function (paras) {
+                    $embed.insertBefore(paras[0]);
+                    mediator.emit('contributions-embed:insert');
+            });
+
+        };
+
+        var completer = function (complete) {
+            mediator.on('contributions-header:insert', function () {
+                bean.on(qwery('#contributions__header-contribute-link')[0], 'click', complete);
+            });
+        };
+
+        this.variants = [
+            {
+                id: 'control',
+                test: function () {
+                    writer();
+                },
+                success: function (complete) {
+                    completer(complete);
+                }
+            }
+        ];
+    };
+});
--- a/static/src/javascripts/projects/common/modules/experiments/tests/contributions-embed.js
+++ b/static/src/javascripts/projects/common/modules/experiments/tests/contributions-embed.js
@@ -7,9 +7,8 @@ define([
     'common/utils/fastdom-promise',
     'common/utils/mediator',
     'text!common/views/contributions-embed.html',
-    'text!common/views/contributions-header.html',
-    'common/views/svgs',
-    'common/modules/article/space-filler'
+    'common/utils/robust'
+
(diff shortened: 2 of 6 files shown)


common/app/conf/switches/ABTestSwitches.scala
```diff
@@ -64,6 +64,16 @@ trait ABTestSwitches {
     exposeClientSide = true
   )
 
+  val ABContributionsEmbed20160822 = Switch(
+    SwitchGroup.ABTests,
+    "ab-contributions-embed-20160822",
+    "Test contributions embed with amount picker.",
+    owners = Seq(Owner.withGithub("jranks123")),
+    safeState = Off,
+    sellByDate = new LocalDate(2016, 8, 27),
```
markjamesbutler: Let's set this to expire on a work day.

static/src/javascripts/projects/common/modules/experiments/tests/contributions-embed.js
```diff
@@ -0,0 +1,95 @@
...
+
+    return function () {
+
+        this.id = 'ContributionsEmbed20160822';
+        this.start = '2016-08-23';
+        this.expiry = '2016-08-27';
+        this.author = 'Jonathan Rankin';
+        this.description = 'Test contributions embed with amount picker.';
+        this.showForSensitive = false;
+        this.audience = 0.13;
```
markjamesbutler: We need to get this value from Asia.  13% might not be the correct figure.

static/src/javascripts/projects/common/modules/experiments/tests/contributions-embed.js
```diff
@@ -0,0 +1,95 @@
...
+                $embed.insertBefore(a);
+                mediator.emit('contributions-embed:insert');
+            });
+
+        };
+
+        var completer = function (complete) {
+            mediator.on('contributions-embed:insert', function () {
+                qwery('figure.interactive.contribute-embed').forEach(function (el) {
+                    require([el.getAttribute('data-interactive')], function (interactive) {
```
markjamesbutler: Why do we need to load this inline?

jranks123: What do you mean exactly?

static/src/javascripts/projects/common/modules/experiments/tests/contributions-embed.js
jranks123: Asias response : ""If you are running a split a / b test -- which means showing the message to 50% and then donatum plus same message to the other 50% -- and you expect the component to produce about 3x increase over current approach, then expose each variant to 600,000 -- so, 1.2m total""

I think that this means we should have and equal (but distinct) audience share from the article test, hence the 0.13 offset and 0.13 audience

static/src/javascripts/projects/common/views/contributions-embed.html
```diff
@@ -0,0 +1,3 @@
+<figure class="element element-interactive contribute-embed interactive element--inline" data-interactive="https://interactive.guim.co.uk/embed/iframe-wrapper/0.1/boot.js" data-canonical-url="https://interactive.guim.co.uk/contributions-embeds/embed/embed.html" data-alt="Contribute">
```
johnduffell: it's only an AB test so less important if it's not going out for real, but fyi @gtrufitt is doing a lot of work to get rid of jumping (yank) in articles, so if possible specify a width and a height in the view that's rendered server side it should be done when it's non-ABified (don't let that stop it going out now though)

johnduffell:
the switch looks ok, but if you want a review of the JS, me and Thomas maybe aren't the best people

static/src/javascripts/projects/common/modules/experiments/tests/contributions-embed.js
```diff
@@ -0,0 +1,95 @@
...
+        this.expiry = '2016-08-30';
+        this.author = 'Jonathan Rankin';
+        this.description = 'Test contributions embed with amount picker.';
+        this.showForSensitive = false;
+        this.audience = 0.13;
+        this.audienceOffset = 0.05;
+        this.successMeasure = 'Determine the effectivness of allowing user to pick contribution amount in the cta';
+        this.audienceCriteria = 'All users';
+        this.dataLinkNames = '';
+        this.idealOutcome = '';
```
johnduffell: if you don't have an ideal outcome, you're not going to know if your test was a success

static/src/javascripts/projects/common/modules/experiments/tests/contributions-embed.js
```diff
@@ -0,0 +1,95 @@
...
+
+        this.id = 'ContributionsEmbed20160823';
+        this.start = '2016-08-23';
+        this.expiry = '2016-08-30';
+        this.author = 'Jonathan Rankin';
+        this.description = 'Test contributions embed with amount picker.';
+        this.showForSensitive = false;
+        this.audience = 0.13;
+        this.audienceOffset = 0.05;
+        this.successMeasure = 'Determine the effectivness of allowing user to pick contribution amount in the cta';
```
johnduffell: that's not a measure, you need to measure something like pv per visit, or I suppose average fall off rate between clicking the button and paying?

static/src/javascripts/projects/common/modules/experiments/tests/contributions-embed.js
```diff
@@ -0,0 +1,95 @@
...
+                bean.on(qwery('#giraffe__contribute-button')[0], 'click', function (){
+                    complete();
+                });
+
+            });
+        };
+
+        this.variants = [
+            {
+                id: 'control',
```
johnduffell: do you only have a control group? what about the variants you want to test?

jranks123: So basically, we have just run a test in same slot with a different ad component and we are comparing the results of this with the results of that test. We could add that variant to this test again, but there wouldn't be much to learn seeing as we have already collected the results.

static/src/javascripts/projects/common/views/contributions-embed.html
jranks123: thanks, once this is in i'll have a look at this

static/src/javascripts/projects/common/modules/experiments/tests/contributions-embed.js
johnduffell: you need unbiased samples to run ab tests, you can't run them consecutively as that's inherently biased (different news, people change their context, economy changes)  Talk to someone in the data team if you want help understanding how to design an AB test.

jranks123: thanks, i have added a proper control now

static/src/javascripts/projects/common/modules/experiments/tests/contributions-embed.js
```diff
@@ -0,0 +1,124 @@
...
+        this.description = 'Test contributions embed with amount picker.';
+        this.showForSensitive = false;
+        this.audience = 0.10;
+        this.audienceOffset = 0.13;
+        this.successMeasure = 'Impressions to number of contributions';
+        this.audienceCriteria = 'All users';
+        this.dataLinkNames = '';
+        this.idealOutcome = 'The interactive embed performs 3x better than the control';
+        this.canRun = function () {
+            var pageObj = window.guardian.config.page;
```
gtrufitt: There is a config util you should import instead of looking at the window.

static/src/javascripts/projects/common/modules/experiments/tests/contributions-embed.js
```diff
@@ -0,0 +1,124 @@
...
+                        var a = el.querySelector('a');
+                        var href = a && a.href;
+
+                        if (href) {
+                            ophan.trackComponentAttention(href, el);
+                        }
+                    });
+                });
+
+                bean.on(qwery('#giraffe__contribute-button')[0], 'click', function (){
```
gtrufitt: We genearlly try and us `js-` prefixed classes for js functionality

static/src/javascripts/projects/common/modules/experiments/tests/contributions-embed.js
```diff
@@ -0,0 +1,124 @@
...
+            });
+
+        };
+
+        var interactiveCompleter = function (complete) {
+            mediator.on('contributions-embed:insert', function () {
+                qwery('figure.interactive.contribute-embed').forEach(function (el) {
+                    require([el.getAttribute('data-interactive')], function (interactive) {
+                        robust.catchErrorsAndLog('interactive-bootstrap', function () {
+                            interactive.boot(el, document, window.guardian.config, mediator);
```
gtrufitt: If possible, same here with passing config object from util rather than window

static/src/javascripts/projects/common/modules/experiments/tests/contributions-embed.js
gtrufitt: Can be `bean.on(qwery('.js-***')[0], 'click', complete)`

static/src/javascripts/projects/common/modules/experiments/tests/contributions-embed.js
```diff
@@ -0,0 +1,124 @@
...
+                bean.on(qwery('#giraffe__contribute-button')[0], 'click', function (){
+                    complete();
+                });
+
+            });
+        };
+
+        var controlCompleter = function (complete) {
+            mediator.on('giraffe:insert', function () {
+                bean.on(qwery('#giraffe__contribute')[0], 'click', function (){
```
gtrufitt: Just checking your click target here is `#giraffe__contribute` instead of button as above?

static/src/javascripts/projects/common/modules/experiments/tests/contributions-embed.js
```diff
@@ -0,0 +1,124 @@
...
+                    var component = $.create(template(giraffeMessage, {
+                        linkText: 'If you use it, if you like it, why not pay for it? It’s only fair',
+                        linkName: 'contribute',
+                        linkHref: 'https://contribute.theguardian.com/?INTCMP=co_uk_cobed_like_control',
+                        copy: 'Give to the Guardian',
+                        svg: svg(arrowRight, ['button--giraffe__icon'])
+                    }));
+                    writer(component);
+                },
+                success: function (complete) {
```
gtrufitt: Can just be `success: controlCompleter` I think

static/src/javascripts/projects/common/modules/experiments/tests/contributions-embed.js
jranks123: Yes, it comes from here https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/common/views/giraffe-message.html

jranks123: and the other one comes from the interactive, the code for which isn't in frontend

static/src/javascripts/projects/common/modules/experiments/tests/contributions-embed.js
```diff
@@ -0,0 +1,124 @@
...
+            return !(pageObj.isSensitive || pageObj.isLiveBlog || pageObj.isFront || pageObj.isAdvertisementFeature) && pageObj.edition === 'UK';
+        };
+
+
+        var writer = function (component) {
+
+            return fastdom.write(function () {
+                var a = $('.submeta');
+                component.insertBefore(a);
+                mediator.emit('contributions-embed:insert');
```
gtrufitt: I think you can pass the component as the second argument:

`mediator.emit('contributions-embed:insert', component);`

and then you would just be able to do:

```
mediator.on('contributions-embed:insert', function (el) {
   require([el.getAttribute('data-interactive')], function 
```

below.

gtrufitt:
Some minor bits but lgtm 👍

gtrufitt:
You need to merge master in :)

static/src/javascripts/projects/common/modules/experiments/tests/contributions-embed.js
gtrufitt: Cool, I'd add a `.js-` class to the button so that you're not hanging off an ID (though probably not as major as we won't be styling off an ID but if you leave the ID then `document.getElementById` is maybe better...)

PR #14091, opened 2016-08-23
Native fluid ads behave like Fabrics

regiskuckaertz:
## What does this change?

When DFP returns a native/fluid ad, it needs to expand to fill its container. This is the current behaviour of the Fabrics format (with one insidious use case where we have to move the ad slot out of its `.fc-container`)

Anyway, this is so native ads are 100% by default.

cc @guardian/commercial-dev

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/dfp/private/render-advert.js
+++ b/static/src/javascripts/projects/common/modules/commercial/dfp/private/render-advert.js
@@ -50,6 +50,21 @@ define([
         };
     }
 
+    function onFluidAd(event, advert) {
+        var node = advert.node;
+        var closestFcContainer = closest(node, '.fc-container');
+        var sectionContainer = bonzo(bonzo.create('<section>'));
+
+        if (closestFcContainer) {
+            fastdom.write(function () {
+                sectionContainer.append(node);
+                sectionContainer.insertAfter(closestFcContainer);
+            });
+        }
+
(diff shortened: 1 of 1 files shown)


static/src/javascripts/projects/common/modules/commercial/dfp/private/render-advert.js
```diff
@@ -149,6 +151,9 @@ define([
 
             function callSizeCallback() {
                 var size = slotRenderEvent.size.join(',');
+                if (size === '0,0') {
```
rich-nguyen: why not add a `0,0` entry to `sizeCallbacks`?

rich-nguyen:
👍

static/src/javascripts/projects/common/modules/commercial/dfp/private/render-advert.js
regiskuckaertz: I wanted to keep `sizeCallbacks[adSize.fluid]` (which will become `sizeCallbacks.fluid`), but I agree this is ugly. Ideally I would like to find the correct element in `adSize` based on `slotRenderEvent.size`. Perhaps the whole `ad-sizes` module needs rethinking.

rich-nguyen: Yeah, neither option is ideal. I thought maybe for the purposes of clarity we could alias `0,0` to `adSize.oldFluid`, which callbacks to `onFluidAd`. Not great

PR #14092, opened 2016-08-23
make amp validator log out the results

johnduffell:
This logs the output from the amp validator so you know why the build is failing, if that's the problem.  Moved out from my other PR as it's not really related.

@TBonnin @jfsoul

### DIFF
--- a/common/test/common/AmpValidityTest.scala
+++ b/common/test/common/AmpValidityTest.scala
@@ -6,6 +6,7 @@ import org.scalatest.concurrent.ScalaFutures
 import org.scalatest.{BeforeAndAfterAll, FlatSpec, Matchers}
 import recorder.DefaultHttpRecorder
 
+import scala.collection.mutable
 import scala.concurrent.Future
 import scala.sys.process._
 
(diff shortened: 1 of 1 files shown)


common/test/common/AmpValidityTest.scala
```diff
@@ -30,16 +31,22 @@ trait AmpValidityTest extends FlatSpec with Matchers with ConfiguredTestSuite wi
...
 
         val commandInputWriter: OutputStream => Unit = writeToProcess(content)
+        var output = List[String]()
 
         // Generate a ProcessIO with desired input and no output (error or otherwise)
-        val io: ProcessIO = BasicIO(withIn = false, ProcessLogger((_) => ()))
+        val io: ProcessIO = BasicIO(withIn = false, ProcessLogger { line =>
+          output = line :: output // Mutate because we can't return!
+          ()
+        })
```
jfsoul: How about going with `StringBuilder` or `StringBuffer` instead?

    val output = new StringBuilder

    // Generate a ProcessIO with desired input and no output (error or otherwise)
    val io: ProcessIO = BasicIO(withIn = false, ProcessLogger(line => output.append(line + "\n")))
      .withInput(commandInputWriter)

Then it's just a case of `output.toString()` for the clue.

johnduffell:
@jfsoul enjoy 😝

jfsoul:
Sweet, sweet `StringBuilder` 👍  👍

PR #14093, opened 2016-08-23
Update capital one cup to efl cup

gtrufitt:
## What does this change?

The capital one cup is now called the EFL cup. Update our naming.

Once out, need to redirect with the redirect tool:

/football/capital-one-cup > /football/efl-cup/
/football/capital-one-cup/live > /football/efl-cup/live
/football/capital-one-cup/fixtures > /football/efl-cup/fixtures
/football/capital-one-cup/results > /football/efl-cup/results

There is no /table or /group/table for efl cup it seems.
## Request for comment

@guardian/dotcom-platform

### DIFF
--- a/admin/app/football/model/PA.scala
+++ b/admin/app/football/model/PA.scala
@@ -17,7 +17,7 @@ object PA extends Collections {
     ("123", "Scottish League Two"),
     ("300", "FA Cup"),
     ("320", "Scottish Cup"),
-    ("301", "Capital One Cup"),
+    ("301", "EFL Cup"),
     ("400", "Community Shield"),
     ("500", "Champions League"),
     ("510", "Europa League"),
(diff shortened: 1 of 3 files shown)


jfsoul:
🏈  👍

TBonnin:
👍 
I wonder if we can not hardcode the name and use the one provided by PA API

gtrufitt:
Maybe, but the name from PA is `EFL Cup 16/17` so may not be consistent with what we want to display. Also, things like 'Barclays Premier League 13/14' in there and doesn't seem we've ever had to [update that](https://github.com/guardian/frontend/commits/21ef5e445e7b5a948c1c466c31b7ddee2bcf33e9/admin/app/football/model/PA.scala) - it's always been 'Premier League', so maybe this league should have been the English football league (efl) from the beginning?

PR #14094, opened 2016-08-23
Extend the ad feedback test and rename the test js file.

JustinPinner:
## What does this change?

Extends the ad feedback test expiry date. Also renames a javascript file because it's currently breaking another test we're building.

We've been receiving good feedback, and want to collect more. The test should have been defined to run until September 14th, but the actual test file declared an earlier termination date.

No.
## Screenshots

As per original PR (https://github.com/guardian/frontend/pull/13879)
## Request for comment

@regiskuckaertz


regiskuckaertz:
👍

PR #14096, opened 2016-08-23
Fluid/native ads are 250px high by default

regiskuckaertz:
Until we find a reliable way to link an IFRAME to an incoming message, dynamic resizing of native ads will not happen. So we might provide a sensible default in the meantime.

cc @guardian/commercial-dev


JonNorman:
:+1:

PR #14099, opened 2016-08-24
Commercial: fine-tune the revealer background

regiskuckaertz:
These changes to improve performance on older devices and to make sure the background image spans the whole width should the user ever rotate the screen.

cc @guardian/commercial-dev


janua:
:+1:

PR #14102, opened 2016-08-24
Remove unused import

MatthewJWalls:
## What does this change?

Removes an unused import in the opengraph CardController (route tagged as experimental). Doing this as an example change to test out the deployment infra.

No value. Tests/compile works

no
## Screenshots

n/a
## Request for comment

n/a


dominickendrick:
👍

PR #14103, opened 2016-08-24
Commercial: temporary fix for native ads in moving slots

regiskuckaertz:
The `top-above-nav` slot gets removed at runtime, after a creative has been delivered by DFP. This is to account for Fabric ads expanding to fill the whole width of the screen (when inside `fc-container`, an `overflow` of `hidden` is applied and thus bars any possibility of the ad expanding). The same process must happen for native ads. Only apparently the gpt library is then unable to update the height of the IFRAME based on its content:

picture 1

We have a campaign that is due to go live now, so a quick fix is to manually set the height of the IFRAME to a sensible default, which is 250px. This works:

picture 2

@janua I also ring-fenced the "move the slot" thing because a native ad can go into any slot, not always the `top-above-nav` one.

cc @guardian/commercial-dev


janua:
:+1:

PR #14104, opened 2016-08-24
cAPI driven css on hosted video pages - correct font contrast

Calanthe:
## What does this change?

This is continuation of the https://github.com/guardian/frontend/pull/14018. Originally we planned to have two variables related to the custom colours on the page: `brandColour` and `brightFont` which had to be specified by the user. Ideally would be to have only one value - `brandColour`. 

This is the attempt to calculate the colour of the font based on the `brandColour` value. The implemented algorithm `isBright()` calculates if the `brandColour` is bright or not and returns a new variable `shouldHaveDarkBackground` which is used in the template to determine if we should use white font colour or not.

The algorithm works as expected but at the same time it indicates to use different (compared to what we specified manually in the hardcoded version) font colours, eg on the Renault page it used to be like this:

screen shot 2016-08-24 at 14 27 48

screen shot 2016-08-24 at 14 27 53

And now is white font colour:
screen shot 2016-08-24 at 14 28 10

screen shot 2016-08-24 at 14 28 15

On the Leffe pages it was:
screen shot 2016-08-24 at 14 31 14

screen shot 2016-08-24 at 14 31 18

And now looks like this:
screen shot 2016-08-24 at 14 30 44

screen shot 2016-08-24 at 14 30 50

@blongden73 Do you think that this is acceptable?
## Request for comment

@blongden73  @guardian/commercial-dev


common/app/common/commercial/hosted/HostedPage.scala
```diff
@@ -35,11 +36,24 @@ case class HostedCampaign(
...
-  brightFont: Boolean,
+  fontColour: FontColour,
   logoLink: Option[String] = None
 )
 
+case class FontColour(brandColour: String) {
+
+  lazy val shouldHaveDarkBackground = !isBright
+
+  lazy val isBright = {
```
rich-nguyen: can be `private`

rich-nguyen:
code looks good 👍

common/app/common/commercial/hosted/HostedPage.scala
```diff
@@ -35,11 +36,24 @@ case class HostedCampaign(
...
+    val hexColour = brandColour.stripPrefix("#")
+    val rgb = Integer.parseInt(hexColour, 16)
+    val c = new Color(rgb)
+    val hsb = Color.RGBtoHSB(c.getRed, c.getGreen, c.getBlue, null)
+    val brightness = hsb(2)
+    brightness < 0.5
```
blongden73: I'd check to see if this 0.5 is the right amount. I think the colour contrast on those examples is too little, I would expect the text to be black not white.

PR #14107, opened 2016-08-24
Commercial: revealer perfs on Android

regiskuckaertz:
Minor improvement of the scroll performance of the Reveal on Android

cc @guardian/commercial-dev


rich-nguyen:
👍

static/src/stylesheets/module/commercial/creatives/_revealer.scss
```diff
@@ -8,6 +8,7 @@
         width: 100%;
         clip: rect(0, auto, auto, 0);
         overflow: hidden;
+        z-index: 1;
```
rich-nguyen: comment? doesn't appear obvious what this does

regiskuckaertz: good call! will do.

PR #14109, opened 2016-08-24
Extend AMP live blog switch

SiAdcock:
## What does this change?

Extends the AMP live blog switch by two months. There are still a number of changes that Google have promised over the next few months that could break AMP live blogs or make them otherwise untenable. It is therefore desirable for us to continue to control AMP live blogs independently of AMP articles in general.


gtrufitt:
👍

PR #14110, opened 2016-08-24
Enable email format for all LiveBlogs, including Minute pages

katebee:
Alternate name: 404NoMore
## What does this change?

`/email` for all! 🚀 (or at least, all LiveBlogs)

**Changes some 404 pages to email formatted content**
Affects requests to LiveBlog pages when requested with `/email` appended.
All articles (including LiveBlogs) can now also have `?format=email` as the query param, which will also return the email format

The US Minute currently has an email format: 
https://www.theguardian.com/us-news/live/2016/aug/23/magical-missing-trump-voters/email

Other Minute pages do not have one; appending **/email** to the URL results in a 404:
https://www.theguardian.com/sport/live/2016/aug/20/the-olympic-games-minute-2016-bolts-triple-treble-triumph

This PR makes it possible to receive a rendered email page for all LiveBlogs, based on the template built for the _US Campaign Minute_. (Minute pages are a subset of LiveBlogs, they don't have their own `isMinute` flag)
- Able to email any Minute series, without giving the webpage all the _US Campaign Minute_ specific stylings
- Editors will be able to create Minute format emails (for scaping by ExactTarget)
- Ever want to email someone the Olympic Games Minute 2016? Now you can
- You can even email a whole LiveBlog, if you like long reads (e.g. this https://www.theguardian.com/world/live/2016/aug/24/italy-terremoto-earthquake-buildings-collapse-amatrice-rome-people-trapped/email)
- `?format=email` will also work: 
  https://www.theguardian.com/world/live/2016/aug/24/italy-terremoto-earthquake-buildings-collapse-amatrice-rome-people-trapped?format=email

Nope.
## Screenshots

The Euro Minute (left) 🏆  Random football LiveBlog (middle) ⚽  The Olympics Minute (Right)🏅
picture 199
## Request for comment

@samdesborough @joelochlann


article/app/controllers/ArticleController.scala
```diff
@@ -218,6 +218,9 @@ class ArticleController extends Controller with RendersItemResponse with Logging
     val content: Either[PageWithStoryPackage, Result] = supportedContentResult.left.flatMap {
       case minute: Article if minute.isUSMinute =>
         Left(MinutePage(minute, StoryPackages(minute, response)))
+        // Enable an email format for 'Minute' content (which are actually composed as a LiveBlog), without changing the non-email display of the page
+      case minute: Article if (minute.isLiveBlog && request.isEmail) =>
+        Left(MinutePage(minute, StoryPackages(minute, response)))
```
joelochlann: result is the same as the above case, so could combine the two conditions and remove this case - unless you find this version more readable

joelochlann: or.... would it be clearer to add in the `render()` function I wonder?

joelochlann:
💎  Looks good to me! Worth getting @desbo's 👀  on it too as he actually knows frontend

desbo:
looks good! 👍

johnduffell:
you may not know that purging in fastly only happens where the path matches item id in capi.  This means that <id>/email won't be purged when changes happen to the original, which might be worse for live blogs than for articles (articles are historically like that because we didn't realise that problem).  I would recommend if possible to make these use ?format=email or some other url parameter rather than appending /email.
Also we unfortunately have the same issue with /print on crosswords, so we need to avoid this habit as much as possible.

joelochlann:
oh wow, thanks @johnduffell! We were in the process of adding `/email` to fronts as well...

paperboyo:
Oh 😢 I just thought it’s such a nice convention /this thing. Our URLs have been historically best when it comes to navigating structure (UI rarely follows, see e.g. tag combiners). Of course I don’t care if it’s /this or ?format=this, I would just like it consistent.

katebee:
@johnduffell: thanks! Have taken a different approach. This now also renders an email format if the url has **?format=email** added (it will not work alongside pagination)

Another PR could be done to remove the **/email** route and move Articles over to (however, this will have to be coordinated with Editorial since they use **/email** for ExactTarget scraping...)

@paperboyo: yeah 😢 **/email** will still also work, but may be cached for longer. I don't currently see caching as a major issue since the email format is only intended for use by Editorial to send out various Minute content (which are published once and not often updated or changed). So the only *intended* user of the **?format=email** URL with be a robot... 🤖

katebee:
Right, feel this needs a thorough check again after the further commits; I've made changes to `routes` 😱 `Requests` 😱 and `ArticleController`😱

paperboyo:
That's great and good to know, thanks @katabee!

desbo:
Looks good to me 👍

johnduffell:
looks good, but you might need to tweak the VCL in fastly a little.
At the moment it only lets params through if you have page in the params.  This is to strip campaign codes out so we don't split the cache.  If you never need campaign codes, you can just whitelist format as well, however if you need format and a campaign code (I guess not) then you'll have to be more clever.
However, it's only a regex, so no biggie.

article/app/controllers/ArticleController.scala
```diff
@@ -128,20 +128,23 @@ class ArticleController extends Controller with RendersItemResponse with Logging
       renderFormat(htmlResponse, jsonResponse, article, Switches.all)
   }
 
-  def renderLiveBlog(path: String, page: Option[String] = None) =
-    Action.async { implicit request =>
+  def renderLiveBlog(path: String, page: Option[String] = None, format: Option[String] = None) =
+    if (format.contains("email"))
+      renderArticle(path)
```
johnduffell: if format isn't email, you should throw a 404 ideally, so we don't cache the same content on more than one url

PR #14111, opened 2016-08-24
Move mobile ad-slot to outside of containers

janua:
## What does this change?

This moves the ad-slots on mobile on fronts to appear outside the container instead of being the last child of the container. This uses `insertAfter` instead of `append` on the `anchor`.

This will also move all the `inline` ads to their own section in between containers.

This seems to causes spacing issues above the ad where a show more button would have appeared. We can fix these as we go along.

We no longer need `onFluidAd` as we are no longer manipulating the DOM after the ad-slot has been added.
### Fluid

fluidad
### Fabric

fabricad
### MPU

mpu
## Request for comment

@regiskuckaertz @rich-nguyen


regiskuckaertz:
Nice job dude! 👍

PR #14112, opened 2016-08-24
Cta Zootropolis AB test removal

Calanthe:
## What does this change?

We have enough data and also I don't want to stop the deploys because switch already expired ;).
## Request for comment

@gtrufitt @guardian/commercial-dev


gtrufitt:
👍

PR #14115, opened 2016-08-24
Clashing tests embed

markjamesbutler:
## What does this change?

Add ability to check multiple ab tests for clashes with email sign-in component.  

None, its a functional change to prevent test clashing.

No

### DIFF
--- a/static/src/javascripts/projects/common/modules/email/run-checks.js
+++ b/static/src/javascripts/projects/common/modules/email/run-checks.js
@@ -41,7 +41,7 @@ define([
             config.page.seriesId === 'world/series/guardian-morning-briefing';
     }
 
-    function userIsInAClashingAbTest(f) {
+    function userIsInAClashingAbTest() {
         return checkTestClash(ab.isInVariant);
     }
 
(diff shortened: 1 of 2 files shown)


static/src/javascripts/projects/common/modules/email/run-checks.js
```diff
@@ -194,6 +204,7 @@ define([
         getEmailInserted: getEmailInserted,
         allEmailCanRun: allEmailCanRun,
         getUserEmailSubscriptions: getUserEmailSubscriptions,
-        listCanRun: listCanRun
+        listCanRun: listCanRun,
+        checkTestClash: checkTestClash
```
gtrufitt: I guess this is just for a test (thanks btw!) - can you differentiate it? I think we normally use a `_test` property that has an object of the test-only methods.

markjamesbutler: Yes kind of.  Its actually used in userIsInAClashingAbTest line 45.  Maybe I should rename it to remove the word test although its only exposed outside this file so I can call it in the unit test.

gtrufitt: Yeah, using internally is fine, but if we're only exporting it for a unit test, differentiating is a good idea :)

gtrufitt:
Hey ho, need to review this approach as we're going to have more things clashing sooner or later, but for this PR 👍

static/src/javascripts/projects/common/modules/email/run-checks.js
markjamesbutler: Cool makes sense. 👍

PR #14117, opened 2016-08-25
Add list of article embed examples to guardian visual glossary

michaelwmcnamara:
Adds a link to a list of article embed examples to the page


gtrufitt:
Just checking there's nothing secret or anything about that url (not exposing anything that'd help a bad person from the root for example https://s3-eu-west-1.amazonaws.com/capi-wpt-querybot)?  - otherwise 👍 Thanks!

michaelwmcnamara:
Discussed directly with Gareth rather than going into detail online, but the upshot is, yes its safe and there is nothing publically available of any importance.

PR #14119, opened 2016-08-25
Removed unused test.

markjamesbutler:
## What does this change?

Removes ab test as testing this component has now finished.

None its a functional change.

No


gtrufitt:
👍

PR #14122, opened 2016-08-25
Move become a supporter copy into subscribe link don't show to members

AWare:
## What does this change?

This adds the become a supporter link to the subscribe button, so that both may be shown. Previously, the become a supporter link was only shown to signed out users. This also hides the supporter link if the user is a paying member. I don't think we can do the same with the subscriber link presently. (I don't think the content authorisation proxy would handle the load)

This should increase membership supporter sign ups without impacting subscriptions. 
## Screenshots

image
## Request for comment

@SiAdcock @samdesborough would appreciate comment on the js, and when we should handle hiding the membership link.
@tomverran for subs
@markjamesbutler for membership

### DIFF
--- a/static/src/javascripts/projects/common/modules/navigation/membership.js
+++ b/static/src/javascripts/projects/common/modules/navigation/membership.js
@@ -2,28 +2,14 @@ define(['common/modules/commercial/user-features', 'common/utils/fastdom-promise
     var LAST_CLASS = 'brand-bar__item--split--last';
 
     function init() {
-        var refreshUser = new Promise(function (resolve) {
-            userFeatures.refresh();
-            resolve(userFeatures.isPayingMember());
-        });
-        refreshUser.then(function (isMember) {
-            if (isMember) {
-                hideSupporterLink();
-            }
-       });
-    }
-
-    function hideSupporterLink() {
-
-        var $becomeMemberLink = $('.js-become-member');
-        var $subscriberLink = $('.js-subscribe');
-
-        fastdom.write(function () {
-            $becomeMemberLink.attr('hidden', 'hidden');
-            $subscriberLink.removeClass(LAST_CLASS);
-        });
-
-
+        if (userFeatures.isPayingMember()) {
+            var $becomeMemberLink = $('.js-become-member');
+            var $subscriberLink = $('.js-subscribe');
+            fastdom.write(function () {
+                $becomeMemberLink.attr('hidden', 'hidden');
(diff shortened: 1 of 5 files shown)


tomverran:
👍  though it would be nice if we could also show subscribe on tablet, since we sell a tablet edition

static/src/javascripts/projects/common/modules/navigation/membership.js
```diff
@@ -0,0 +1,31 @@
+define(['common/modules/commercial/user-features', 'common/utils/fastdom-promise', 'Promise', 'common/utils/$'], function (userFeatures, fastdom, Promise, $) {
+    var LAST_CLASS = 'brand-bar__item--split--last';
+
+    function init() {
+        var refreshUser = new Promise(function (resolve) {
```
desbo: there's no need to assign this `Promise` to a var really, I'd return the entire chain from the function instead

static/src/javascripts/projects/common/modules/navigation/membership.js
```diff
@@ -0,0 +1,31 @@
...
+            userFeatures.refresh();
+            resolve(userFeatures.isPayingMember());
```
desbo: it doesn't look like either of these `userFeatures` functions involve I/O; is there any need for them to be in the promise chain?

static/src/stylesheets/module/nav/_brand-bar.scss
```diff
@@ -43,6 +43,24 @@
...
+    left: 11px;
+    position: relative;
+    height: 24px;
+    bottom: -5px;
+    display: block;
+}
+
+.brand-bar__item--split--last {
+    padding-left: 11px;
+    height: 34px;
```
SiAdcock: Is it possible to use less magic numbers for these heights, positions and paddings? Perhaps some multiples of `$gs-gutter` (for horizontal spacing) `$gs-baseline` (for vertical spacing) and `$gs-height` (for element height)

static/src/javascripts/projects/common/modules/navigation/membership.js
AWare: refresh does some async stuff, but not in a way that's useful here, I'm not going to call refresh, and just not show the supporter link if we don't know membership status

SiAdcock:
Very nice, man 👍

desbo:
ANOTHER ONE 👍

PR #14123, opened 2016-08-25
Change ad-slot--container-inline for mobileLandscape to not have a

janua:
This reduces some of the spacing above `.ad-slot--container-inline` which is the container for ads on mobile that get inserted between the containers. There is still a bit of space but this lessens the gap.

@regiskuckaertz @rich-nguyen


rich-nguyen:
👍

PR #14124, opened 2016-08-25
Commercial: none of the creative layers should ever repeat in the fabric video

regiskuckaertz:
Before:

picture 3

After:

picture 4

cc @guardian/commercial-dev @tijanamiletic


rich-nguyen:
👍

PR #14125, opened 2016-08-25
Messenger: id calling iframe

regiskuckaertz:
## What does this change?

Sends a nonce to an embedded view as soon as it has loaded, with the ID of the iframe into which it is loaded—only if the embedded view is a native ad.

Native ads heavily rely on `window.postMessage`. Sadly, there is standard way to link an incoming message to the corresponding `iframe`. But we need this information though, as in some scenarios we might want to manipulate that frame. Fortunately, DFP gives us a means to know when the `load` event of a slot has fired. We will use that as our way in and send some information that will allow us to create that link later on.

cc @guardian/commercial-dev

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/dfp/private/on-slot-load.js
+++ b/static/src/javascripts/projects/common/modules/commercial/dfp/private/on-slot-load.js
@@ -2,11 +2,14 @@ define([
     'common/modules/commercial/dfp/private/get-advert-by-id',
     'common/modules/commercial/dfp/private/post-message'
 ], function (getAdvertById, postMessage) {
+    var nativeAdName = /^\d+-\d+-\d+;\d+;<!DOCTYPE/;
     return onLoad;
 
     function onLoad(event) {
         var advert = getAdvertById(event.slot.getSlotElementId());
         var iframe = advert.node.querySelector('iframe');
-        postMessage({ id: iframe.id }, iframe.contentWindow);
+        if (nativeAdName.test(iframe.name)) {
+            postMessage({ id: iframe.id }, iframe.contentWindow);
+        }
     }
 });
--- a/static/src/javascripts/projects/common/modules/commercial/dfp/private/on-slot-load.js
+++ b/static/src/javascripts/projects/common/modules/commercial/dfp/private/on-slot-load.js
@@ -1,0 +1,12 @@
+define([
+    'common/modules/commercial/dfp/private/get-advert-by-id',
+    'common/modules/commercial/dfp/private/post-message'
+], function (getAdvertById, postMessage) {
+    return onLoad;
+
+    function onLoad(event) {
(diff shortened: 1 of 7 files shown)


static/src/javascripts/projects/common/modules/commercial/dfp/private/on-slot-load.js
```diff
@@ -0,0 +1,18 @@
...
+    'common/modules/commercial/dfp/private/get-advert-by-id',
+    'common/modules/commercial/dfp/private/post-message'
+], function (getAdvertById, postMessage) {
+    var nativeAdName = /^\d+-\d+-\d+;\d+;<!DOCTYPE/;
+    return onLoad;
+
+    function onLoad(event) {
+        var advert = getAdvertById(event.slot.getSlotElementId());
+        var iframe = advert.node.querySelector('iframe');
+        /* We don't have (yet) a means to identify a native creative, so we
```
rich-nguyen: eeek!

regiskuckaertz:
@rich-nguyen I will reorganise our modules in a separate PR. It seems to me that most of what's in the dfp folder is private, and thus has nothing to do in the common folder.

static/src/javascripts/projects/common/modules/commercial/dfp/private/post-message.js
```diff
@@ -0,0 +1,7 @@
+define(function () {
+    return postMessage;
+
+    function postMessage(message, targetWindow) {
+        targetWindow.postMessage(JSON.stringify(message), location.protocol + '//tpc.googlesyndication.com');
```
rich-nguyen: I would reference `location.protocol + '//tpc.googlesyndication.com'` as a shared definition, so that the `postMessage` function in the `post-message` module and the `onMessage` function in `messenger` cannot deviate. It is crucial that they match.

static/test/javascripts/spec/common/commercial/dfp/messenger.spec.js
```diff
@@ -133,7 +133,7 @@ define([
         it('should respond to Rubicon messages with no IDs', function (done) {
             var payload = { type: 'set-ad-height', value: { id: 'test', height: '20px' } };
             messenger.register('set-ad-height', routines.rubicon, mockWindow);
-            onMessage({ origin: location.protocol + '//' + location.host, data: JSON.stringify(payload), source: mockFrame })
+            onMessage({ origin: 'http://tpc.googlesyndication.com', data: JSON.stringify(payload), source: mockFrame })
```
rich-nguyen: same as comment above

rich-nguyen:
👍

kelvin-chappell:
Interesting use of the word 'nonce'.  In British English it has a quite different meaning!

regiskuckaertz:
giphy

PR #14126, opened 2016-08-25
AMPify soundcloud embeds in AMP articles

SiAdcock:
## What does this change?

Vanilla Soundcloud embeds are not supported by AMP. Luckily, AMP provide an `amp-soundcloud` component that we can use instead.

This change adds an cleaner that finds Soundcloud embeds in AMP articles and replaces them with their `amp-soundcloud` counterparts

AMP articles that contain Soundcloud embeds will validate successfully

Only AMP articles
### News article

localhost-3000-healthcare-network-2016-aug-25-look-soundcloud-amp iphone 6
### Live blog

localhost-3000-sport-live-2016-aug-23-team-gb-return-home-from-rio-olympics-live-updates-amp iphone 6
## Request for comment

@guardian/dotcom-platform

### DIFF
--- a/common/app/views/support/cleaner/AmpEmbedCleaner.scala
+++ b/common/app/views/support/cleaner/AmpEmbedCleaner.scala
@@ -80,13 +80,12 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
       for {
         audioElement <- document.getElementsByClass("element-audio")
         iframeElement <- audioElement.getElementsByTag("iframe")
-        soundcloudUrl = iframeElement.attr("src")
-        trackId <- getTrackIdFromUrl(soundcloudUrl) filterNot { _.isEmpty }
-        soundcloudElement = createElement(document, trackId)
-      } yield (audioElement, iframeElement, soundcloudElement, {
+        trackId <- getTrackIdFromUrl(iframeElement.attr("src"))
+      } yield {
+        val soundcloudElement = createElement(document, trackId)
         audioElement.appendChild(soundcloudElement)
         iframeElement.remove()
-      })
+      }
     }
   }
 
--- a/common/app/views/support/cleaner/AmpEmbedCleaner.scala
+++ b/common/app/views/support/cleaner/AmpEmbedCleaner.scala
@@ -73,7 +73,7 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
       }.foreach { audioElementWithIframe: Element =>
         audioElementWithIframe.getElementsByTag("iframe").foreach { iframeElement: Element =>
           val soundcloudUrl = iframeElement.attr("src")
-          val pattern = ".*%2Ftracks%2F(\\d+).*".r
+          val pattern = ".*soundcloud.com%2Ftracks%2F(\\d+).*".r
           soundcloudUrl match {
             case pattern(trackId) =>
               val soundcloud = createElement(document, trackId)
--- a/common/app/views/support/cleaner/AmpEmbedCleaner.scala
+++ b/common/app/views/support/cleaner/AmpEmbedCleaner.scala
@@ -67,19 +67,20 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
       soundcloud.attr("data-visual", "true")
       soundcloud.attr("height", "300") // height is necessary if data-visual == true
     }
+
+    def getTrackIdFromUrl(soundcloudUrl: String): String = {
+      val pattern = "soundcloud.com%2Ftracks%2F(\\d+)".r
+      pattern.findAllIn(soundcloudUrl).matchData.map { matches => matches.group(1) }.mkString
+    }
+
     def clean(document: Document) = {
-      document.getElementsByClass("element-audio").filterNot { audioElement: Element =>
-        audioElement.getElementsByTag("iframe").isEmpty
-      }.foreach { audioElementWithIframe: Element =>
+      document.getElementsByClass("element-audio").foreach { audioElementWithIframe: Element =>
         audioElementWithIframe.getElementsByTag("iframe").foreach { iframeElement: Element =>
           val soundcloudUrl = iframeElement.attr("src")
-          val pattern = ".*soundcloud.com%2Ftracks%2F(\\d+).*".r
-          soundcloudUrl match {
-            case pattern(trackId) =>
-              val soundcloud = createElement(document, trackId)
-              audioElementWithIframe.appendChild(soundcloud)
-              iframeElement.remove()
-            case _ =>
+          Option(getTrackIdFromUrl(soundcloudUrl)) filterNot { _.isEmpty } foreach { trackId =>
+            val soundcloudElement = createElement(document, trackId)
+            audioElementWithIframe.appendChild(soundcloudElement)
+            iframeElement.remove()
           }
         }
       }
--- a/common/app/views/support/cleaner/AmpEmbedCleaner.scala
+++ b/common/app/views/support/cleaner/AmpEmbedCleaner.scala
@@ -71,7 +71,7 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
     def getTrackIdFromUrl(soundcloudUrl: String): Option[String] = {
       val pattern = ".*soundcloud.com%2Ftracks%2F(\\d+).*".r
       soundcloudUrl match {
-        case pattern(trackId) => Option(trackId)
+        case pattern(trackId) => Some(trackId)
         case _ => None
       }
     }
--- a/common/app/views/support/cleaner/AmpEmbedCleaner.scala
+++ b/common/app/views/support/cleaner/AmpEmbedCleaner.scala
@@ -66,14 +66,14 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
     }.foreach { audioElementWithIframe: Element =>
       audioElementWithIframe.getElementsByTag("iframe").foreach { iframeElement: Element =>
         val soundcloudUrl = iframeElement.attr("src")
-        val pattern = "%2Ftracks%2F(\\d+)".r
-        val trackId = pattern.findAllIn(soundcloudUrl).matchData.map { matches => matches.group(1) }.mkString
-        val soundcloud = document.createElement("amp-soundcloud")
-        soundcloud.attr("data-trackid", trackId)
-        soundcloud.attr("data-visual", "true")
-        soundcloud.attr("height", "300") // height is necessary if data-visual == true
-        audioElementWithIframe.appendChild(soundcloud)
-        iframeElement.remove()
+        val pattern = ".*%2Ftracks%2F(\\d+).*".r
+        soundcloudUrl match {
+          case pattern(trackId) =>
+            val soundcloud = createAmpSoundcloudElement(document, trackId)
+            audioElementWithIframe.appendChild(soundcloud)
+            iframeElement.remove()
+          case _ =>
+        }
       }
     }
   }
@@ -136,6 +136,13 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
       iframe.appendChild(overflowElem)
       interactive.appendChild(iframe)
     }
+  }
+
+  private def createAmpSoundcloudElement(document: Document, trackId: String): Element = {
+    val soundcloud = document.createElement("amp-soundcloud")
+    soundcloud.attr("data-trackid", trackId)
+    soundcloud.attr("data-visual", "true")
+    soundcloud.attr("height", "300") // height is necessary if data-visual == true
   }
 
   private def getVideoAssets(id:String): Seq[VideoAsset] = article.elements.bodyVideos.filter(_.properties.id == id).flatMap(_.videos.videoAssets)
--- a/common/app/views/support/cleaner/AmpEmbedCleaner.scala
+++ b/common/app/views/support/cleaner/AmpEmbedCleaner.scala
@@ -61,8 +61,8 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
   }
 
   def cleanAmpSoundcloud(document: Document) = {
-    document.getElementsByClass("element-audio").filter { audioElement: Element =>
-      audioElement.getElementsByTag("iframe").length != 0
+    document.getElementsByClass("element-audio").filterNot { audioElement: Element =>
+      audioElement.getElementsByTag("iframe").isEmpty
     }.foreach { audioElementWithIframe: Element =>
       audioElementWithIframe.getElementsByTag("iframe").foreach { iframeElement: Element =>
         val soundcloudUrl = iframeElement.attr("src")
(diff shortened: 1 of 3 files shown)


common/test/views/support/cleaner/AmpEmbedCleanerTest.scala
```diff
@@ -0,0 +1,57 @@
+package views.support.cleaner
```
johnduffell: 🎉 tests! 🌟

common/app/views/support/cleaner/AmpEmbedCleaner.scala
```diff
@@ -60,6 +60,24 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
 
   }
 
+  def cleanAmpSoundcloud(document: Document) = {
+    document.getElementsByClass("element-audio").filter { audioElement: Element =>
+      audioElement.getElementsByTag("iframe").length != 0
```
johnduffell: you can use isDefined instead of length != 0.  although it's a bit more tricky if it's a java list

common/app/views/support/cleaner/AmpEmbedCleaner.scala
```diff
@@ -60,6 +60,24 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
...
+    }.foreach { audioElementWithIframe: Element =>
+      audioElementWithIframe.getElementsByTag("iframe").foreach { iframeElement: Element =>
+        val soundcloudUrl = iframeElement.attr("src")
+        val pattern = "%2Ftracks%2F(\\d+)".r
+        val trackId = pattern.findAllIn(soundcloudUrl).matchData.map { matches => matches.group(1) }.mkString
```
johnduffell: do you really want to find them all, and then concatenate them together? or should you be finding the first, and only create the element if there's a match?

common/app/views/support/cleaner/AmpEmbedCleaner.scala
```diff
@@ -60,6 +60,24 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
...
+        val soundcloud = document.createElement("amp-soundcloud")
+        soundcloud.attr("data-trackid", trackId)
+        soundcloud.attr("data-visual", "true")
+        soundcloud.attr("height", "300") // height is necessary if data-visual == true
+        audioElementWithIframe.appendChild(soundcloud)
+        iframeElement.remove()
```
johnduffell: there's a lot of heavy lifting in this process that's a little distracting, so think about having a `getIdFromElement` and a `createAmpElement` function and calling one then the other and then do the append/remove as a final step

common/app/views/support/cleaner/AmpEmbedCleaner.scala
SiAdcock: We could use:
`!audioElement.getElementsByTag("iframe").isEmpty()`

johnduffell:
looks great, mainly the comment about concatenating that I don't think is quite right 👍

common/app/views/support/cleaner/AmpEmbedCleaner.scala
johnduffell: actually, I'm less of a fan of negating: `!.....................................................isEmpty` because we're splitting the "not" and the "empty"
some people even go for `!!!.................................................isEmpty` just to make the `!!!` more obvious

dominickendrick: or just `filterNot`

SiAdcock: As discussed, `.length != 0` is slightly clearer than the alternative 😄

SiAdcock:
@johnduffell I've made a few changes, mainly the simplification of that pattern matcher by using an extractor... what do you think?

common/app/views/support/cleaner/AmpEmbedCleaner.scala
```diff
@@ -113,12 +131,20 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
     }
   }
 
+  private def createAmpSoundcloudElement(document: Document, trackId: String): Element = {
```
johnduffell: this method should be within the cleanAmpSoundcloud one as it's not related to any of the others.  You could wrap them both in an object for readability.

common/app/views/support/cleaner/AmpEmbedCleaner.scala
```diff
@@ -60,6 +60,24 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
...
   }
 
+  def cleanAmpSoundcloud(document: Document) = {
+    document.getElementsByClass("element-audio").filterNot { audioElement: Element =>
+      audioElement.getElementsByTag("iframe").isEmpty
+    }.foreach { audioElementWithIframe: Element =>
+      audioElementWithIframe.getElementsByTag("iframe").foreach { iframeElement: Element =>
+        val soundcloudUrl = iframeElement.attr("src")
+        val pattern = ".*%2Ftracks%2F(\\d+).*".r
+        soundcloudUrl match {
```
johnduffell: I've not seen a pattern match used like this before!

common/app/views/support/cleaner/AmpEmbedCleaner.scala
```diff
@@ -60,6 +60,24 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
 
   }
 
+  def cleanAmpSoundcloud(document: Document) = {
+    document.getElementsByClass("element-audio").filterNot { audioElement: Element =>
+      audioElement.getElementsByTag("iframe").isEmpty
+    }.foreach { audioElementWithIframe: Element =>
+      audioElementWithIframe.getElementsByTag("iframe").foreach { iframeElement: Element =>
+        val soundcloudUrl = iframeElement.attr("src")
+        val pattern = ".*%2Ftracks%2F(\\d+).*".r
```
TBonnin: Why not checking this is a soundcloud url here?

common/test/views/support/cleaner/AmpEmbedCleanerTest.scala
TBonnin: 💯

common/test/views/support/cleaner/AmpEmbedCleanerTest.scala
```diff
@@ -0,0 +1,66 @@
...
+import com.gu.contentapi.client.model.v1.{Content => ApiContent}
+import model.{Article, Content}
+import org.jsoup.Jsoup
+import org.jsoup.nodes.Document
+import org.scalatest.{FlatSpec, Matchers}
+
+class AmpEmbedCleanerTest extends FlatSpec with Matchers {
+
+  private def clean(document: Document): Document = {
+    val cleaner = AmpEmbedCleaner(article())
```
TBonnin: Nit: you can pass a html string as argument and do the `Jsoup.parse()` in this method rather than repeating it in every single tests :)

common/app/views/support/cleaner/AmpEmbedCleaner.scala
```diff
@@ -60,6 +60,24 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
...
+        soundcloudUrl match {
+          case pattern(trackId) =>
+            val soundcloud = createAmpSoundcloudElement(document, trackId)
+            audioElementWithIframe.appendChild(soundcloud)
+            iframeElement.remove()
+          case _ =>
```
johnduffell: don't necessarily listen too much to my specific suggestions, but in general, I fund a useful principle for structuring code is think about how you would explain what you want the code to do to someone else, and then make sure the code as written matches that structure.
I don't want to pick apart too much such a small piece of code, but think about  considerations such as correctness, clarity and anything else you can think of.  Some may be competing, but not always.

SiAdcock: @johnduffell would you have time at some point to discuss this briefly? 😄

SiAdcock:
Behold, @johnduffell and @TBonnin, the majestic AMP Soundcloud cleaner! See its functional purity, marvel at its idiomatic use of monads and functions that don't have side effects.

common/app/views/support/cleaner/AmpEmbedCleaner.scala
```diff
@@ -60,6 +60,33 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
...
 
+  object AmpSoundcloud {
+    def createElement(document: Document, trackId: String): Element = {
+      val soundcloud = document.createElement("amp-soundcloud")
+      soundcloud.attr("data-trackid", trackId)
+      soundcloud.attr("data-visual", "true")
+      soundcloud.attr("height", "300") // height is necessary if data-visual == true
+    }
+
+    def getTrackIdFromUrl(soundcloudUrl: String): String = {
```
johnduffell: this should probably return an Option[String] in case it isn't a sound cloud url

common/app/views/support/cleaner/AmpEmbedCleaner.scala
```diff
@@ -60,6 +60,33 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
...
+      val pattern = "soundcloud.com%2Ftracks%2F(\\d+)".r
+      pattern.findAllIn(soundcloudUrl).matchData.map { matches => matches.group(1) }.mkString
+    }
+
+    def clean(document: Document) = {
+      document.getElementsByClass("element-audio").foreach { audioElementWithIframe: Element =>
+        audioElementWithIframe.getElementsByTag("iframe").foreach { iframeElement: Element =>
+          val soundcloudUrl = iframeElement.attr("src")
+          Option(getTrackIdFromUrl(soundcloudUrl)) filterNot { _.isEmpty } foreach { trackId =>
```
johnduffell: ah yes move this optiony stuff into the getTrackId... function

SiAdcock: Damn, that's how I originally implemented it. I should have trusted my instincts! 😄 

Thanks for your help, dude

common/app/views/support/cleaner/AmpEmbedCleaner.scala
```diff
@@ -60,6 +60,33 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
...
+            val soundcloudElement = createElement(document, trackId)
+            audioElementWithIframe.appendChild(soundcloudElement)
+            iframeElement.remove()
+          }
```
johnduffell: the structure is good, but to simplify this glue logic you can use a for comprehension I think - as all the things are seq/option.  Also you can split the pure stuff and the mutating a little more if you like.  Something along these lines perhaps:
```
// work out what we need to change
val toDo = for {
  audioElementWithIframe<- document.getElementsByClass("element-audio")
  iframeElement <- audioElementWithIframe.getElementsByTag("iframe")
  soundcloudUrl = iframeElement.attr("src")
  trackId <- getTrackIdFromUrl(soundcloudUrl)
} yield (audioElementWithIframe, iframeElement, createElement(document, trackId))
// go through and do the mutation to the jsoup dom
toDo.foreach { case (audioElementWithIframe, iframeElement, soundcloudElement) =>
  audioElementWithIframe.appendChild(soundcloudElement)
  iframeElement.remove()
}
```

common/app/views/support/cleaner/AmpEmbedCleaner.scala
```diff
@@ -60,6 +60,33 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
...
+    def createElement(document: Document, trackId: String): Element = {
+      val soundcloud = document.createElement("amp-soundcloud")
+      soundcloud.attr("data-trackid", trackId)
+      soundcloud.attr("data-visual", "true")
+      soundcloud.attr("height", "300") // height is necessary if data-visual == true
+    }
+
+    def getTrackIdFromUrl(soundcloudUrl: String): String = {
+      val pattern = "soundcloud.com%2Ftracks%2F(\\d+)".r
+      pattern.findAllIn(soundcloudUrl).matchData.map { matches => matches.group(1) }.mkString
```
TBonnin: Why not doing a simple match rather than a findAllIn -> map -> group(1) -> mkString?
Also should this method return a Option[String] to clearly signify there is no result rather than relying on an empty string

SiAdcock: The trouble with a simple match is that I want to capture the `(\\d+)` (which will be the track ID). I'm I missing something really obvious?

SiAdcock:
I have used for comprehension as suggested by @johnduffell for the DOM manipulation and extractor pattern as suggested by @TBonnin for the regex. Can I get some love please? 😄

common/app/views/support/cleaner/AmpEmbedCleaner.scala
```diff
@@ -69,10 +69,11 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
     }
 
     def getTrackIdFromUrl(soundcloudUrl: String): Option[String] = {
-      val pattern = "soundcloud.com%2Ftracks%2F(\\d+)".r
-      val trackId = pattern.findAllIn(soundcloudUrl).matchData.map { matches => matches.group(1) }.mkString
-
-      Option(trackId)
+      val pattern = ".*soundcloud.com%2Ftracks%2F(\\d+).*".r
+      soundcloudUrl match {
+        case pattern(trackId) => Option(trackId)
```
TBonnin: We used the more constrained `Some(x)` rather than `Option`

gtrufitt:
I preferred your first version 😁

johnduffell:
```

  def cleanAmpSoundcloud(document: Document) = {
    document.getElementsByClass("element-audio").filter { audioElement: Element =>
      audioElement.getElementsByTag("iframe").length != 0
    }.foreach { audioElementWithIframe: Element =>
      audioElementWithIframe.getElementsByTag("iframe").foreach { iframeElement: Element =>
        val soundcloudUrl = iframeElement.attr("src")
        val pattern = "%2Ftracks%2F(\\d+)".r
        val trackId = pattern.findAllIn(soundcloudUrl).matchData.map { matches => matches.group(1) }.mkString
        val soundcloud = document.createElement("amp-soundcloud")
        soundcloud.attr("data-trackid", trackId)
        soundcloud.attr("data-visual", "true")
        soundcloud.attr("height", "300") // height is necessary if data-visual == true
        audioElementWithIframe.appendChild(soundcloud)
        iframeElement.remove()
      }
    }
  }
```

or...

```

    def clean(document: Document) = {
      for {
        audioElement <- document.getElementsByClass("element-audio")
        iframeElement <- audioElement.getElementsByTag("iframe")
        trackId <- getTrackIdFromUrl(iframeElement.attr("src"))
      } yield {
        val soundcloudElement = createElement(document, trackId)
        audioElement.appendChild(soundcloudElement)
        iframeElement.remove()
      }
    }
  }
```
(with some helper methods above 😛 )

TBonnin:
👍

johnduffell:
well, I'm happy if everyone else is, although with code, you can't keep everyone happy! 👍 
🚀

SiAdcock:
Soooo much nicer, Scala is awesome! Thanks for your help guys 😄

PR #14127, opened 2016-08-26
Tiny Bugfix: Avoid displaying non-upgraded tag rich-links

gtrufitt:
## What does this change?

I bungled the selector in my previous change, this prevents showing non-upgraded rich-links.

Don't get an odd plain link.

@guardian/dotcom-platform


jfsoul:
💰 🔗 👍

johnduffell:
👍

PR #14130, opened 2016-08-26
Track web notifications follow button attention

janua:
## What does this change?

This adds attention tracking to the follow button for web notifications. Currently, we don't truly know how many people have seen the button as it depends on lots of different variables; On Chrome, on HTTPS, on a liveblog and in the AB test (for now).

This will allow us to track how many people actually seen the button, but also allow us to see how long the button was on screen for the user if we need that.

No
## Request for comment

@stephanfowler @zeftilldeath @joelochlann @gtrufitt

### DIFF
--- a/common/app/conf/switches/ABTestSwitches.scala
+++ b/common/app/conf/switches/ABTestSwitches.scala
@@ -8,9 +8,9 @@ trait ABTestSwitches {
     SwitchGroup.ABTests,
     "ab-live-blog-chrome-notifications-prod2",
     "Live blog chrome notifications - prod",
-    owners = Seq(Owner.withGithub("NathanielBennett")),
+    owners = Seq(Owner.withGithub("janua")),
     safeState = Off,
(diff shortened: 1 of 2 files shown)


common/app/conf/switches/ABTestSwitches.scala
```diff
@@ -8,9 +8,9 @@ trait ABTestSwitches {
     SwitchGroup.ABTests,
     "ab-live-blog-chrome-notifications-prod2",
     "Live blog chrome notifications - prod",
-    owners = Seq(Owner.withGithub("NathanielBennett")),
+    owners = Seq(Owner.withGithub("janua")),
```
gtrufitt: 😆 passed the baton

gtrufitt:
👍

common/app/conf/switches/ABTestSwitches.scala
janua: :smile:

PR #14131, opened 2016-08-26
Add pageview ID to prebid fastlane call

JonNorman:
## What does this change?

We pass a number of parameters to the Rubicon header bidding calls we make. This adds the `pageview-id` to this, which will then make it available in the logs we receive from Rubicon (which are then loaded and processed within the data lake).

See the documentation [here](http://prebid.github.io/dev-docs/bidders.html#rubicon)

It means that we should be able to track Rubicon header bidding calls from the logs that Rubicon do/will provide us.

Nope
## Request for comment

@regiskuckaertz - this is a rather trivial change, but I'd be glad to get your comments on the best way of testing this, what to look out for etc as I'm fairly green when it comes to header bidding.


regiskuckaertz:
I don't think you can test this, prebid.js handles everything. When we have some post-processing on the front-end can we verify parameters domains and control logic.

👍

PR #14132, opened 2016-08-26
Fluid ads in liveblogs

regiskuckaertz:
## What does this change?

In liveblogs, inserts slots during the commercial bootstrap rather than the enhanced one. Also harmonises the slot naming scheme with the rest of the site. And makes sure a fluid ad expands to fill the width of the screen on mobile.

picture 1

cc @guardian/commercial-dev

### DIFF
--- a/static/src/javascripts/projects/commercial/modules/liveblog-adverts.js
+++ b/static/src/javascripts/projects/commercial/modules/liveblog-adverts.js
@@ -1,6 +1,7 @@
 define([
     'bonzo',
     'common/utils/fastdom-promise',
+    'common/utils/detect',
     'common/utils/config',
     'common/utils/mediator',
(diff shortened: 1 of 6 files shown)


static/src/javascripts/projects/commercial/modules/liveblog-adverts.js
```diff
@@ -64,9 +66,14 @@ define([
 
     function insertAds(slots) {
         for (var i = 0; i < slots.length && slotCounter < MAX_ADS; i++) {
-            var $adSlot = bonzo(createSlot('inline1' + slotCounter++, 'liveblog-inline block'));
+            var slotName = isMobile && slotCounter === 0 ?
```
rich-nguyen: the mobile logic ruins everything. :(

rich-nguyen:
👍

PR #14133, opened 2016-08-26
GA discussion events

cb372:
## What does this change?

Send discussion events to GA as well as Omniture.

Also fix a bug that meant the "followed a permalink" event was never sent. The event was being emitted by `loader.js` before the analytics module had been initialised and started listening for it.

~~Note: we don't quite have parity with Omniture because we don't yet send the logged in user's ID as a custom dimension. That's a little involved, and not only relevant to discussion, so I'll do it in a separate PR.~~ This has been done in #14137.

~~Dunno. Waiting for @crifmulholland to come back from holiday so I can check which events are actually needed.~~ Checked with Chris M, decided we only need the "scrolled to comments" event. Other events are already covered by the generic click tracking.

No
## Screenshots

n/a
## Request for comment

@nicl @ajosephides


ajosephides:
👍

PR #14134, opened 2016-08-26
AB test: Remove share counts. 😨

gtrufitt:
## What does this change?

This adds an AB test to discover whether whether the social share counts make a difference to the number of clicks on social buttons.

The audience is 50% as the amount of share count clicks is fairly low vs PV and it'd be nice to finish the test this year. To test a 2% minimum detectable effect we need to run this high for a couple of weeks but we can rely on ABacus to do it's job RE samples.

Share counts take maintenance time and degrade UX due to lazy loading. We're re-visting meta and want to prove that share counts are worth the time and effort. Facebook is periodically retiring versions of it's API which make the functionality harder to do. Twitter's API changed months ago and we've not had Twitter share counts included since then (and no-one noticed/cared). It seems to be a largely internal vanity metric that can be retrieved from other sources.
## Request for comment

@guardian/dotcom-platform

### DIFF
--- a/static/src/javascripts/projects/common/modules/experiments/tests/no-social-count.js
+++ b/static/src/javascripts/projects/common/modules/experiments/tests/no-social-count.js
@@ -10,7 +10,7 @@ define([
 
         this.id = 'NoSocialCount';
         this.start = '2016-08-25';
-        this.expiry = '2016-09-07';
+        this.expiry = '2016-09-21';
         this.author = 'Gareth Trufitt';
         this.description = 'Test whether the social share counts make a difference to the number of clicks on social buttons';
         this.audience = 0.5;
--- a/static/src/javascripts/projects/common/modules/experiments/tests/no-social-count.js
+++ b/static/src/javascripts/projects/common/modules/experiments/tests/no-social-count.js
@@ -12,10 +12,10 @@ define([
         this.start = '2016-08-25';
         this.expiry = '2016-09-07';
         this.author = 'Gareth Trufitt';
-        this.description = 'Test whether the social share counts increase the number of shares an article has';
+        this.description = 'Test whether the social share counts make a difference to the number of clicks on social buttons';
         this.audience = 0.5;
         this.audienceOffset = 0.5;
-        this.successMeasure = 'The control group, that shows social shares, will have 2% more clicks on sharing buttons';
+        this.successMeasure = 'No significant difference in clicks between the variant and control';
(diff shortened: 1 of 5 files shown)


stephanfowler:
👍 - don't forget to warn Central Prod..!

static/src/javascripts/projects/common/modules/experiments/tests/no-social-count.js
```diff
@@ -0,0 +1,49 @@
...
+        this.start = '2016-08-25';
+        this.expiry = '2016-09-21';
+        this.author = 'Gareth Trufitt';
+        this.description = 'Test whether the social share counts make a difference to the number of clicks on social buttons';
+        this.audience = 0.5;
+        this.audienceOffset = 0.5;
+        this.successMeasure = 'No significant difference in clicks between the variant and control';
+        this.audienceCriteria = 'All users';
+        this.dataLinkNames = '';
+        this.idealOutcome = 'Social share counts *do not* cause a significant increase in sharing, so we can remove them';
```
johnduffell: the ideal outcome in statistical sense should be total clicks on all share buttons in the top article meta decreases by less than 2%?

johnduffell:
great! only what stephan said! 👍 exciting 🎉 🔪

johnduffell:
good work! 🎈

paperboyo:
I almost entirely agree with the reasons for removal given by @gtrufitt, especially the technical ones (yank!). Just wanted to indicate that there is a use case that won’t show in the metric, because it doesn’t involve any observable use: being a misanthrope, I have never shared a thing in my life, but I tend to observe indicators like this (or comment count etc.) passively to reinforce my sense of societal failure (when the counts are high on crap content) or, OTOH, to reinforce my sense of societal failure (when the counts are low on good content). From time to time, I’m also unpleasantly surprised.

The above does not in any way pertain only to the Guardian content, so it has nothing to do with vanity. Absolutely nothing.

gtrufitt:
@paperboyo You can still reinforce your sense of societal failure by using the debug tool or a chrome extension such as https://chrome.google.com/webstore/detail/buzzsumo/gedpbnanjmblcmlfhgfficjnglidndfo - that way you have the minimal cognitive load possible to reinforce your sense of societal failure, no matter what site you're on 💫

paperboyo:
Thanks, @gtrufitt, good advice even though I still am undecided if I hate extensions more than people or the other way around 😉

PR #14135, opened 2016-08-26
Remove spinny sharecount

gtrufitt:
## What does this change?

Doesn't do the ol' count up to the number re-paint re-paint re-paint malarkey. Just sticks the number in.

No re-paints. Uses fastdom. 
## Request for comment

@guardian/dotcom-platform


johnduffell:
fab, simplify 🔋

PR #14136, opened 2016-08-26
Messenger: Allow ads running on the local domain

regiskuckaertz:
## What does this change?

SafeFrame _and_ ads running on the local domain should be able to send messages.

cc @guardian/commercial-dev


JonNorman:
:+1:

PR #14137, opened 2016-08-26
GA: set identityId, isLoggedIn custom dimensions

cb372:
## What does this change?

Send the user's ID (extracted from the `GU_U` cookie) and their logged-in-ness to GA as custom dimensions.

Tested locally by copying my `GU_U` cookie from PROD and checking the GA pageview event's custom dimensions.

Moar tracking

No
## Screenshots

n/a
## Request for comment

@ajosephides @dominickendrick

### DIFF
--- a/common/app/views/fragments/analytics/google.scala.html
+++ b/common/app/views/fragments/analytics/google.scala.html
@@ -5,6 +5,20 @@
 
 (function() {
 
+    @*
+    TODO Because this JS lives outside of module-land,
+    we can't depend on any of the util modules and thus
+    end up with a lot of copy-pasta here.
+
+    It might be best to just include the GA snippet here
+    and move the pageview event into a module, as long as we
+    can ensure that it loads very early on (because we want to send the pageview asap).
+
+    However, it's tricky to do that while Omniture is still around.
+    I think we should endure the copy-pasta for now
+    and refactor after Omniture has been removed.
+    *@
+
     // Remove spaces from content type and convert it to lower case
     function convertContentType(contentType) {
         var result = contentType;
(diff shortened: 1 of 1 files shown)


common/app/views/fragments/analytics/google.scala.html
```diff
@@ -22,6 +36,55 @@
...
+            userData = cookieData ? JSON.parse(decodeBase64(cookieData.split('.')[0])) : null;
+        if (userData) {
+            return userData[0];
+        } else {
+            return null;
+        }
+    }
+
+    var identityId = getIdentityIdFromCookie();
+    var isLoggedIn = (identityId !== null);
```
cb372: @ajosephides Is this what you intended for the `isLoggedIn` dimension? Just a flag to say whether the cookie is set or not? Seems a little redundant.

ajosephides: isLoggedIn is a boolean regarding the users current logged in state and set on a hit level scope. The identityId is set on a user level scope. 
When looking at isLoggedIn it should be examined in conjunction with identityId. Just having identityId logged could give the impression that the user is permanently logged in (see user scope in GA) even though they may not be. 
Subscriptions and Membership have had a separate variables in place to isolate someone that has a login but is not currently logged in.
The approach above gives us a consistent way in which to approach this in Google Analytics across our properties.
Does that make sense?

cb372: Oh yeah, I forgot they were different scopes. So this code is correct?

ajosephides: 👍

common/app/views/fragments/analytics/google.scala.html
```diff
@@ -5,6 +5,20 @@
 
 (function() {
 
+    @*
```
dominickendrick: Luckly this code is already excuted earlier in the javascript loading cycle in https://github.com/guardian/frontend/blob/master/common/app/templates/inlineJS/nonBlocking/getUserData.scala.js and should be available in `window.guardian.config.user` if they are logged in!

cb372: Lovely! Thank you.

dominickendrick:
👍

PR #14139, opened 2016-08-30
Extend sports surging switch for one week

JonNorman:
This feature is no longer needed but @Calum-Campbell is currently on annual leave.

I'm extending it for a week when Calum will be back and can remove the feature (and the switch).


rich-nguyen:
👍

PR #14140, opened 2016-08-30
Re-organise and add more amp validation test endpoints

jfsoul:
## What does this change?

The main change is the addition of more AMP validation tests, focussing on all the different kind of embeds.  These have been added to the teamcity validation tests only, as many of the embeds are probably infrequently found in articles (this can be looked at again later if necessary).  For now, a number of these are commented out because they fail validation.  The intention is that these be enabled as the bugs are fixed.

I have also removed the pre-release validator functionality (in a non-destructive way).  The AMP developers have confirmed there is not currently a way to fetch this, but there is an open issue for this: https://github.com/ampproject/amphtml/issues/4663.

I've also moved the examples of different tones from the scala validation tests, as I don't think these endpoints are high-risk in terms of failing amp validation after changes.

Better amp validation test coverage, less redundant pre-release validator tests are run.
## Request for comment

@guardian/dotcom-platform


gtrufitt:
💎 Nice! Lgtm. 👍

PR #14141, opened 2016-08-30
make casing in controller match routes

gidsg:
## What does this change?

Make controller casing match casing in routes files

Should fix route on production

No
## Screenshots

N/A
## Request for comment

CC @gtrufitt @akash1810


gtrufitt:
👍

PR #14142, opened 2016-08-30
Update quiz question to inline-block instead of block style

gtrufitt:
## What does this change?

Fixes this in IE11:

image

@guardian/dotcom-platform


jfsoul:
seems reasonable to me 👍

PR #14143, opened 2016-08-30
Radiator tweak: change title and show message when no expiring switches

TBonnin:
## What does this change?

Mini tweak for radiator: show "None" rather than nothing when there is no expiring switch

Clearer message

No
## Request for comment

@jfsoul

### DIFF
--- a/admin/app/views/radiator.scala.html
+++ b/admin/app/views/radiator.scala.html
@@ -50,8 +50,9 @@
 
 
     <div class="expiring-wrapper">
-        <h2>Expiring features</h2>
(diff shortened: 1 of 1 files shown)


admin/app/views/radiator.scala.html
```diff
@@ -50,8 +50,9 @@
 
 
     <div class="expiring-wrapper">
-        <h2>Expiring features</h2>
+        <h2>Expiring switches</h2>
```
TBonnin: because a switch is not always a feature :)

jfsoul:
👍

PR #14145, opened 2016-08-30
Add retry to page fetching for amp validation

jfsoul:
## What does this change?

Adds retry functionality to the fetching of pages in the amp validator.  We were getting a couple of errors from the teamcity job that appear to just be occasional network issues.  They always clear up by the next time the job is run, so I've added a couple of retries on failure just to see if we can avoid the noisy warning email.

Less annoying emails for non-issues.
## Request for comment

@guardian/dotcom-platform

### DIFF
--- a/tools/amp-validation/fetch-page.js
+++ b/tools/amp-validation/fetch-page.js
@@ -1,27 +1,32 @@
 'use strict';
 
 const https = require('https');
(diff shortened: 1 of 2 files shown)


gtrufitt:
lgtm 👍

PR #14146, opened 2016-08-30
Move new navigation from Enhanced to Standard

SiAdcock:
## What does this change?

This change moves the new navigation module from the Enhanced JS bundle into Standard for the duration of the new navigation's AB test.

We have found that, if the user taps the menu icon when the page first loads, but before all DOM manipulation has completed, there is a ~500ms delay between the user interaction and the menu opening. Moving the navigation module from enhanced into standard reduces this delay to ~350ms. For slower connections, the improvement is more dramatic.

As part of the AB test, we will determine whether users tend to tap the menu icon while the DOM is still being loaded.

Finally, the Details polyfill is being loaded in the wrong place, and should be loaded inline. Since the new navigation is the only module that uses it, it is safe to remove the polyfill for now. In a future PR, I will add it to the correct place. 

Nope

### DIFF
--- a/static/src/javascripts/bootstraps/enhanced/common.js
+++ b/static/src/javascripts/bootstraps/enhanced/common.js
@@ -28,7 +28,6 @@ define([
     'common/modules/identity/autosignin',
     'common/modules/identity/cookierefresh',
     'common/modules/navigation/navigation',
-    'common/modules/navigation/newHeaderNavigation',
     'common/modules/navigation/profile',
     'common/modules/navigation/search',
     'common/modules/navigation/membership',
@@ -82,7 +81,6 @@ define([
     AutoSignin,
     CookieRefresh,
     navigation,
-    newHeaderNavigation,
     Profile,
     Search,
     membership,
@@ -129,7 +127,6 @@ define([
 
             initialiseNavigation: function () {
                 navigation.init();
-                newHeaderNavigation();
             },
 
             showTabs: function () {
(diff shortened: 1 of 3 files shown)


rich-nguyen:
> if the user taps the menu icon when the page first loads

How often do we think this is happening?

The difference between standard and enhanced, alongside the technical implications you have mentioned, is that things that are in standard are considered so important that they should run before our commercial system. We haven't communicated that well as a team, admittedly, but do you think we should? It would make it easy to have discussions about what should be in standard. Improving the response time of something which isn't in the critical path for rendering, should be considered enhanced, in my view.

SiAdcock:
Thanks @rich-nguyen 

The new header is not yet live, so no real users have experienced the delay yet. It was just something that came up in QA.

I agree it would be helpful to understand what constitutes a standard module and what is enhanced, to help us make better decisions about where to load modules.

It's hard for me to make a judgement call as to whether primary navigation should be loaded before the commercial system. In our tests, interacting with the menu on a device with a poor 3G connection had a 4 second response time. Poor response times would likely skew the success of the new navigation's A/B test. Happy to discuss further

johnduffell:
sounds like the debate is about order of loading, but we have to avoid just bumping things up arbitrarily.  Is this going back to the drawing board, if so it should probably close for now?

SiAdcock:
Yeah, a new plan for loading the navigation is currently being formalised. I'll close this for now.

PR #14149, opened 2016-08-30
Messenger: include iframe element in the callback arguments

regiskuckaertz:
Some cross-frame callbacks will need a reference to the `iframe` from which a message came from. This is achieved via a [simple handshake](https://github.com/guardian/frontend/pull/14125/files), after which each incoming message will provide the `iframeId` argument.

cc @guardian/commercial-dev


rich-nguyen:
👍

PR #14150, opened 2016-08-30
Extend AMP live blog switches

SiAdcock:
## What does this change?

Extend all AMP live blog switches until October. We are still unsure what the impact of future changes to `<amp-live-list>` and Google's AMP blue links will be, so we'd still like to have granular control over AMP support for live blogs for the foreseeable future.


jfsoul:
👍

PR #14151, opened 2016-08-30
Fix wrong clickMacro expansion

regiskuckaertz:
`%%CLICK_URL_UNESC%%` is the click macro in DFP, but we pass it as `clickMacro` in our custom templates.

cc @guardian/commercial-dev


JonNorman:
:+1:

PR #14153, opened 2016-08-30
Outbrain test clashes

markjamesbutler:
## What does this change?

Loads the correct outbrain component if the user is in a clashing ab test.  i.e. Contributions donatom.  

None, this is a functional change required for compliance with the Outbrain T&C's.

No

### DIFF
--- a/static/src/javascripts/projects/commercial/modules/third-party-tags/outbrain.js
+++ b/static/src/javascripts/projects/commercial/modules/third-party-tags/outbrain.js
@@ -121,7 +121,12 @@ define([
     function checkDependencies() {
         return Promise.all([checkEmailSignup(), checkClashingABTest()])
             .then(function(result) {
-                result.find('email');
+
+                function findEmail(value) {
+                    return value == 'email';
+                }
+
+                return result.find(findEmail);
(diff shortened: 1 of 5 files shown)


static/src/javascripts/projects/commercial/modules/third-party-tags/outbrain.js
```diff
@@ -114,6 +117,37 @@ define([
...
+            .catch(function () {
+                return 'email';
+            });
+    }
+
+    function checkClashingABTest() {
+        if (!clashingABTestPromise) {
+            clashingABTestPromise = new Promise(function (resolve) {
+                if (clash.userIsInAClashingAbTest()) {
+                    resolve('email');
```
gtrufitt: Could we possibly update this to a more generic `non-compliant` as it's not specific to email now?

gtrufitt: or `nonCompliant` as we're in JS 💃

gtrufitt:
👍 Minor comment, but looks like a good step to untangling things a bit, cheers!

PR #14154, opened 2016-08-30
Override User Agent style for lists

SiAdcock:
## What does this change?

User Agent styles for `<ul>` elements were sneaking into AMP live blogs. This change overrides those styles, removing the list style type and the default padding. 

The change makes lists look consistent with those on non-AMP live blogs.

Nope 
### Before

amp theguardian com-business-live-2016-aug-30-reports-to-offer-latest-clues-on-brex--business-live iphone 6
### After

localhost-3000-business-live-2016-aug-30-reports-to-offer-latest-clues-on-brex--business-live-amp iphone 6


gtrufitt:
👍

PR #14155, opened 2016-08-30
Messenger: scroll routine and persistent listeners

regiskuckaertz:
**Introducing persistent listeners.** Normal routines produce a single value and are wrapped in a Promise. Some other routines produce a stream of values and in that case a Promise is not a working abstraction. Persistent listeners answer that need by having direct write access to the iframe messaging pipe. They can send as many messages as necessary.

The first example of persistent listener is the scroll routine. It sends the coordinates of an ad slot to the corresponding embedded document as soon as it enters in the viewport.

cc @guardian/commercial-dev


static/src/javascripts/projects/commercial/modules/messenger.js
```diff
@@ -18,32 +19,47 @@ define([
...
             listeners[type].length = 0;
         } else {
-            var idx = listeners[type].indexOf(callback);
-            if (idx > -1) {
+            if (listeners[type] === callback) {
+                listeners[type] = null;
                 registeredListeners -= 1;
-                listeners[type].splice(idx, 1);
+            } else {
+                var idx = listeners[type].indexOf(callback);
```
rich-nguyen: `(listeners[type] || [ ]).indexOf(...`

static/src/javascripts/projects/commercial/modules/messenger.js
```diff
@@ -81,33 +97,39 @@ define([
...
-        }, Promise.resolve(true));
-
-        return promise.then(function (response) {
-            respond(null, response);
-        }).catch(function (ex) {
-            reportError(ex, { feature: 'native-ads' });
-            respond(formatError(error500, ex), null);
-        });
+        } else {
+            listeners[data.type](respond, data.value, iframe);
```
rich-nguyen: may be `undefined`?

static/src/javascripts/projects/commercial/modules/messenger/scroll.js
```diff
@@ -0,0 +1,125 @@
...
+
+    function onIntersect(changes) {
+        changes.forEach(function (_) {
+            var slot = slots[_.target.id];
+            listeners[slot.iframeId].visible = _.intersectionRatio > 0;
+        });
+    }
+
+    // Instances of classes bound to the current view are not serialised correctly
+    // by JSON.stringify. That's ok, we don't case if it's a DOMRect or some other
```
rich-nguyen: `care`

static/src/javascripts/projects/commercial/modules/messenger/scroll.js
```diff
@@ -0,0 +1,125 @@
...
+                observer.unobserve(listeners[id].slot);
+                slots[listeners[id].slot.id] = false;
+            }
+            listeners[id] = false;
+            listenerCounter -= 1;
+        }
+
+        if (listenerCounter === 0) {
+            window.removeEventListener('scroll', onScroll);
+            if (useIO) {
```
rich-nguyen: `(useIO && observer)`, for idempotence

static/src/javascripts/projects/commercial/modules/messenger/scroll.js
```diff
@@ -0,0 +1,125 @@
...
+
+    function onScroll() {
+        if (!taskQueued) {
+            taskQueued = true;
+            fastdom.read(function () {
+                taskQueued = false;
+
+                var slots = Object.keys(listeners)
+                    .filter(function (id) {
+                        return listeners[id].visible;
```
rich-nguyen: If `useIO` is `false`, won't all the slots be filtered here? It doesn't seem like you would ever get a call of `respond` without using `IntersectionObserver`.

rich-nguyen: A simple fix is to use `Object.keys(listeners)` when not using IOs

rich-nguyen:
super awesome 👍

static/src/javascripts/projects/commercial/modules/messenger/scroll.js
regiskuckaertz: If `useIO` is `false`, then [`visible` is always `true`](https://github.com/guardian/frontend/pull/14155/files#diff-2fb78cc4b5143edecc6410391f4cbd3eR43). When `useIO` is `true`, this flag will be set to `true` [only when a slot enters the viewport](https://github.com/guardian/frontend/pull/14155/files#diff-2fb78cc4b5143edecc6410391f4cbd3eR108) and will be set back to `false` when it leaves it.

But you're right, this is not really obvious by looking at the code, I'll rework that part. What do you mean by the fix?

static/src/javascripts/projects/commercial/modules/messenger/scroll.js
regiskuckaertz: thanks 😄

static/src/javascripts/projects/commercial/modules/messenger.js
regiskuckaertz: yes, I'll take care of that

static/src/javascripts/projects/commercial/modules/messenger.js
regiskuckaertz: yes indeed.

static/src/javascripts/projects/commercial/modules/messenger.js
regiskuckaertz: hmm actually that can't happen, but this is not obvious

static/src/javascripts/projects/commercial/modules/messenger.js
regiskuckaertz: I've guarded the whole process. If `listeners[type]` is defined, we know it is either a function or an array.

static/src/javascripts/projects/commercial/modules/messenger/scroll.js
rich-nguyen: Yes, that makes sense, i suspected it was because `visible` was set to `!useIO`. I thought you could rejig the code so that there is a single `filter` call, and inside the filter function, if `useIO` is `false`, you use a viewport-top check, and if `true`, it uses `visible`. The source of the filter would be `Object.keys(listeners)`, rather than [slots](https://github.com/guardian/frontend/pull/14155/files/0f2bc13a4cfd5e2e2375d2e7527f1e7c2f470328#diff-2fb78cc4b5143edecc6410391f4cbd3eR93)

regiskuckaertz: Ah, now I remember why I have two filters: because in the `useIO` case, I want to avoid the cost of computing the geometry of slots that are not within the viewport. So what I've done is clearly separate the two processes instead.

PR #14156, opened 2016-08-31
Update switch expiry dates.

markjamesbutler:
## What does this change?

Extends switch expiry to fix the build.  We will soon be re-launching these tests so I have not removed them in this PR.

None, its a functional test.

No


gtrufitt:
👍 🚢

PR #14158, opened 2016-08-31
Apply correct style to updated times in AMP live blog

SiAdcock:
## What does this change?

When a live blog post is updated after it is published, the time at which it was last updated is displayed at the bottom of the block. In the AMP live blog, the presentational style of the updated time is ugly and inconsistent with the non-AMP live blog. This change brings the styling of the AMP version into line with the non-AMP version

No
### Before

amp theguardian com-business-live-2016-aug-30-reports-to-offer-latest-clues-on-brex--business-live-amp iphone 6
### After

localhost-3000-business-live-2016-aug-30-reports-to-offer-latest-clues-on-brex--business-live-amp iphone 6 1

### DIFF
--- a/common/app/views/fragments/amp/stylesheets/liveBlog.scala.html
+++ b/common/app/views/fragments/amp/stylesheets/liveBlog.scala.html
@@ -50,6 +50,17 @@ ul {
     font-weight: bold;
 }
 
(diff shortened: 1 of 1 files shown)


gtrufitt:
👍

PR #14162, opened 2016-08-31
Add s3 config versioning

dominickendrick:
## What does this change?

This PR add support for versioning of our config file via the new updates in `configuration-magic`.
- Pin our config to different versions via the `s3ConfigVersion` variable
- Add a read me entry on the manual process for updating config
- There is a new updated config file at `eu-west-1-frontend.v1.conf` that adds Targetting tool config options:
  - `aws.composer.account.role`
  - `target.campaigns.bucket`
## What is the value of this and can you measure success?

We can now version our config! This allows us to:
- test our changes locally
- track and version changes in config, whilst keeping them private.
## Does this affect other platforms - Amp, Apps, etc?

No
## Request for comment

@TBonnin @currysoup @guardian/dotcom-platform

PR #14163, opened 2016-08-31
Targeting Client

itsibitzi:
## What does this change?

Adds an Akka agent that pulls in currently active campaigns and adds them to articles. The JSON for the matching campaigns will end up at the top of the article beside the `nonKeywordTagIds` etc.

This patch is unrelated to the rendering side of targeting, it simply delivers the latest data.

If campaigns with the correct tag rules end up in the JSON embedded in the HTML.

Currently isolated to `article` so I think this is isolated, but would like some feedback. We'll be adding features like this to the mobile app soon. I'm unsure about Amp.

This also depends on https://github.com/guardian/frontend/pull/14162
## Request for comment

@dominickendrick


common/app/common/configuration.scala
```diff
@@ -457,6 +457,31 @@ class GuardianConfiguration extends Logging {
...
+          log.error("amazon client cross account exception", ex)
+          if (Play.isProd) throw ex
+          None
+      }
+    }
+
+    lazy val crossAccountMandatoryCredentials = this.crossAccountCredentials
+      .getOrElse(throw new BadConfigurationException(s"AWS credentials for targeting cross account are not correctly configured"))
+
+    lazy val campaignsBucket = configuration.getMandatoryStringProperty("targeting.campaigns.bucket")
```
dominickendrick: This was added as `target.campaigns.bucket` in config as per instructions

common/app/model/meta.scala
```diff
@@ -391,7 +392,8 @@ trait ContentPage extends Page {
     item.commercial.javascriptConfig ++
     item.content.conditionalConfig ++
     item.content.javascriptConfig ++
-    metadata.javascriptConfigOverrides
+    metadata.javascriptConfigOverrides ++
+    Map("campaigns" -> JsArray(item.content.campaigns.map(Campaign.toJson)))
```
TBonnin: Can the formatting happen in `content.scala` (like the other configs)?

common/app/targeting/CampaignAgent.scala
```diff
@@ -0,0 +1,43 @@
...
+
+  def refresh() {
+    crossAccountClient.foreach(client => {
+      agent.alter { old =>
+        try {
+          old.update(client, Configuration.targeting.campaignsBucket)
+        } catch {
+          case NonFatal(e) => log.error("Failed to update campaign list.", e)
+        }
+        old
```
TBonnin: Why returning old? Since you don't seem to do anything with the resulting `Future[CampaignCache]` you might want to use `send` rather than `alter`

common/app/targeting/TargetingLifecycle.scala
```diff
@@ -0,0 +1,29 @@
...
+  jobs: JobScheduler,
+  akkaAsync: AkkaAsync)(implicit executionContext: ExecutionContext) extends LifecycleComponent {
+
+  appLifecycle.addStopHook { () => Future {
+    jobs.deschedule("TargetingCampaignRefreshJob")
+  }}
+
+  override def start(): Unit = {
+    jobs.deschedule("TargetingCampaignRefreshJob")
+    jobs.schedule("TargetingCampaignRefreshJob", "0 * * * * ?") {
```
TBonnin: I am curious: How often campaign changes / are added / removed?

TBonnin:
Other than the small comments I made, it lgtm. 👍

currysoup:
Composer credentials are required since we're using STS to access a S3 bucket in the composer account.

TBonnin:
@currysoup Ok. Might be a good idea to explain why the Composer credentials are required then. Also I (and most of frontend dev) don't have Composer access in Janus. What happen in this case?

currysoup:
If the STS client fails to setup you'll simply not get any campaigns, and it'll log the failure. I thought it reasonable to fail softly with this since it's non-critical.

TBonnin:
Ok fair enough

dominickendrick:
I've added `composer` permissions to `frontendDeveloper` in `Janus` here https://github.com/guardian/janus/pull/491

TBonnin:
@dominickendrick I hadn't see your janus PR. However I am quite concerned about giving access to devs for dev mode only. Is there any other options?

dominickendrick:
Yeah there are other options, but they may also be slightly convoluted :
1. We can invert the control, and make the `targeting` tool write into and s3 bucket in the `frontend` account, but `mapi` also needs access to the data, so they would need to write data into both accounts.
2. This data could be made available via a `http` API and the endpoint secured with a cert
This is the same scenario we have with the `cmsFronts` access, so there is a precedent, and we are only granting local devs `s3 Read` access.
@NickPapacostas or @currysoup may have other options !

currysoup:
@dominickendrick Hey! I think, like you said, the inversion of control would help frontend but not MAPI so that approach is a bit questionable.

I actually like the idea of a HTTP end point (stateless is good, right?) but we had some concerns with tools scaling in line with frontend. Though it would only be each frontend server hitting our end point once every minute, which isn't a lot at all. The precedent was a big driver behind the S3 decision, but I'm actually pretty eager to look into a HTTP version. That would allow local development to hit against `targeting.code.dev-gutools.co.uk` which would mean you could get real data too.

This would allow us to remove the STS stuff and would mostly just be a `targeting-client` update.

rich-nguyen:
Might be worth adding a comment somewhere which explains what a `campaign` is.

common/app/targeting/TargetingLifecycle.scala
johnduffell: better not to use schedule, as its is extremely difficult to check it's sensible. There is a scheduleEvery method that takes a duration so you can pass in 1.hour or something and everyone can see what it means by inspection

common/app/targeting/CampaignAgent.scala
```diff
@@ -0,0 +1,43 @@
...
+import scala.util.control.NonFatal
+
+object CampaignAgent extends Logging with ExecutionContexts {
+  val crossAccountClient: Option[AmazonS3Client] = Try({
+    val client = new AmazonS3Client(Configuration.targeting.crossAccountMandatoryCredentials)
+    client.setEndpoint(AwsEndpoints.s3)
+    client
+  }).toOption
+
+  private val agent = AkkaAgent[CampaignCache](new CampaignCache())
```
johnduffell: I don't seem to have access to the targeting client repo so I'm not sure what the behaviour of CampaignCache is

dominickendrick: @johnduffell You should have the permissions to read this now

dominickendrick: @rich-nguyen adding an issue regarding documentation here, then we can link through to it, rather than replicating it in frontend. https://github.com/guardian/targeting/issues/46

johnduffell: @dominickendrick read where?  I am searching the whole guardian organisation 😢

dominickendrick: @johnduffell https://github.com/guardian/targeting is the repo

common/app/targeting/CampaignAgent.scala
```diff
@@ -0,0 +1,43 @@
...
+
+  def refresh() {
+    crossAccountClient.foreach(client => {
+      agent.alter { old =>
+        try {
+          old.update(client, Configuration.targeting.campaignsBucket)
```
johnduffell: `update` (and in turn `refresh`) should return a Future so that the scheduler knows when it's completed (and we can check if it's successful too)

common/app/model/content.scala
```diff
@@ -82,6 +83,7 @@ final case class Content(
   lazy val isHeroic = HeroicTemplateSwitch.isSwitchedOn && tags.isLabourLiverpoolSeries
   lazy val isImmersive = fields.displayHint.contains("immersive") || isImmersiveGallery || tags.isUSMinuteSeries || isHeroic
   lazy val isAdvertisementFeature: Boolean = tags.tags.exists{ tag => tag.id == "tone/advertisement-features" }
+  lazy val campaigns: List[Campaign] = targeting.CampaignAgent.getCampaignsForTags(tags.tags.map(_.id))
```
johnduffell: This is happening on every origin request.  Is there some limit enforced on the number and complexity of campaigns existing in the s3 bucket?  as we're basically only using lists everywhere so the time complexity will be about O(nrules*(nlacking+nrequired)*ncontenttags*campaigns)
This is fine if we have enforced limits on all of those things, but if we want any of those to have arbitrary length, we need a better algorithm.

currysoup: There are no limitations currently, our main issue is we don't know how much the campaign system will be used. We currently have the concept of an 'active' campaign but there's nothing to stop anyone making a super complex campaign with loads of rules which is always active.

The likely outcome is most campaigns will be simple (only a few required/lacking tags). Any advice on latency requirements would be great so I could come up with some sensible limitations.

johnduffell: dom came and discussed this and other issues on friday, basically the limits don't have to be restrictive at this stage, just they have to exist so that people have to make a conscious decision to remove them.
just take whatever you expect to use in the next month or two and multiply them all by 3, so something like 12 rules with 12 lacking/required and 12 tags over 30 campaigns.

Think of it more like a switch that will expire when it's used too much seriously, only rather than being a time based switch it's a usage based one.

common/app/targeting/CampaignAgent.scala
```diff
@@ -0,0 +1,27 @@
...
+import scala.util.control.NonFatal
+import scala.concurrent.Future
+
+object CampaignAgent extends Logging with ExecutionContexts {
+  private val agent = AkkaAgent[CampaignCache](CampaignCache(List()))
+
+  def refresh(): Future[Unit] = {
+    Configuration.targeting.campaignsUrl.map(url => {
+      CampaignCache.fetch(url).map(agent.send)
+    }).getOrElse(Future.failed(throw new BadConfigurationException("Campaigns URL not configured")))
```
currysoup: @johnduffell is this what you meant wrt `Future`s?

I'm not 100% on the `Future.failed` part 😕

johnduffell: yep only either remove the word `throw` as you just want to put the exception in as a fail,not throw the exception from the code, or alternatively you could remove the `Future.failed` and let something else catch it.  The former is probably more functional programming compliant but I *think* the scheduler should cope with the latter too

johnduffell: I think agent.send is returning a Unit here?  you might want to use something that returns a future, and then do a Future.sequence on the result.  Otherwise the scheduler will know when the CampaignCache.fetch completes, but won't realise that the agent.send is still in progress.

currysoup: The `fetch` returns a `Future[CampaignCache]` which I map to `Future[Unit]` by sending it to the akka agent. The scheduler wants a `Future[Unit]` so this is convenient.

Thanks for the notice about the throw, totally a schoolboy error on my part 😞

johnduffell: send happens to return unit, but it's actually starting an asynchronous task, so although the types happen to line up it's still effectively hiding an orphaned Future from you.

So at present the scheduler will consider the job finished even if the send is still running.  And any error from the send (if that's possible) will go somewhere undefined.

All this is unlikely to break something in the successful case, but when I was investigating perf issues last year, this kind of lost Futures problem caused me a lot of hassle.

So I'd recommend using alter instead of send, and sequencing the future that's been returned so that the scheduler knows when they're both finished successfully and can log properly.

johnduffell:
thanks looks great! and sorry I meant `flatMap` and said `sequence` but it looks like you worked that out 😊

johnduffell:
I think the only worry I have left is about the limits on someone adding too many/too complex campaigns by mistake and killing frontend.  I had a quick look through the targeting client PRs but I can't see anything yet, any pointers?

currysoup:
I'm going to add a `.take(255)` to the active-only endpoint and make a warning in the frontend that will inform users that there are too many active campaigns.

I suspect this will need to be revisited, depending on how popular targeting becomes 😄 but for now just limiting + warning seems sensible.

johnduffell:
sounds good, although do that also for the rules and their include/excluded tags, just in case a rogue campaign gets in there with loads of tags

currysoup:
Note on limiting the number of rules and campaigns that can be delivered to the frontend.

We're currently limiting the number of rules to 20 tags:
https://github.com/guardian/targeting/pull/62/files

We're also limiting the number of active campaigns that can be delivered to the frontend to 100: 
https://github.com/guardian/targeting/pull/60

currysoup:
Added the ability for frontend to override the number of campaigns received - also a feature switch

common/app/targeting/CampaignAgent.scala
```diff
@@ -5,21 +5,28 @@ import com.gu.targeting.client.CampaignCache
...
+import conf.switches.Switches.Targeting
 
 object CampaignAgent extends Logging with ExecutionContexts {
-  private val agent = AkkaAgent[CampaignCache](CampaignCache(List()))
+  private val agent = AkkaAgent[CampaignCache](CampaignCache(Nil, None))
 
   def refresh(): Future[Unit] = {
     Configuration.targeting.campaignsUrl.map(url => {
-      CampaignCache.fetch(url).flatMap(agent.alter).map(_ => ())
+      CampaignCache.fetch(url, limit = 100).flatMap(agent.alter).map(_ => ())
```
johnduffell: does this limit the rules and tags as well as the campaigns?

project/Dependencies.scala
```diff
@@ -72,7 +72,7 @@ object Dependencies {
   val libPhoneNumber = "com.googlecode.libphonenumber" % "libphonenumber" % "7.2.4"
   val logback = "net.logstash.logback" % "logstash-logback-encoder" % "4.6"
   val kinesisLogbackAppender = "com.gu" % "kinesis-logback-appender" % "1.3.0"
-  val targetingClient = "com.gu" %% "targeting-client-play24" % "0.6.0"
+  val targetingClient = "com.gu" %% "targeting-client-play24" % "0.7.0"
```
johnduffell: when you bump the version, also link to the relevant PRs in targeting-client so frontend devs can easily review your changes there too

common/app/targeting/CampaignAgent.scala
```diff
@@ -5,21 +5,28 @@ import com.gu.targeting.client.CampaignCache
 import conf.Configuration
 import scala.util.control.NonFatal
 import scala.concurrent.Future
+import conf.switches.Switches.Targeting
 
 object CampaignAgent extends Logging with ExecutionContexts {
-  private val agent = AkkaAgent[CampaignCache](CampaignCache(List()))
+  private val agent = AkkaAgent[CampaignCache](CampaignCache(Nil, None))
```
TBonnin: I am not sure to understand why Nil is the default value for no campaign. Why not keeping an empty list? Same question for the total count. Why not a simple `Int`?

common/app/targeting/CampaignAgent.scala
```diff
@@ -5,21 +5,28 @@ import com.gu.targeting.client.CampaignCache
...
-  def getCampaignsForTags(tags: Seq[String]) = try {
-    agent().getCampaignsForTags(tags)
-  } catch {
-    case NonFatal(e) =>
-      log.error("Failed to get campaigns for tags.", e)
-      List()
+  def getCampaignsForTags(tags: Seq[String]) = {
+    if (Targeting.isSwitchedOn) {
+      try {
+        agent().getCampaignsForTags(tags)
```
TBonnin: How can `getCampaignsForTags` throw an exception?

common/app/targeting/CampaignAgent.scala
currysoup: Nil is the same as an empty list, I think it's considered more idiomatic. 

For the total count, we need a way to express a difference between there being 0 active campaigns and there being no data retrieved from the server. This could potentially be achieved with a boxed integer and using `null` but I think that would be bad practice.

TBonnin: From the client perspective (by client I mean, the piece of code which is using the campaign info), does it make a difference whatever no data has been retrieved or there is no active campaign?

common/app/targeting/CampaignAgent.scala
currysoup: It's more a case of defensive coding, right now the rules engine is very simple and I don't think there are any cases were it could throw an `Exception` in particular but in the future this could change. I'd say it's preferable that the agent itself handle the failure case rather than allowing the exception to propagate to the job scheduler which lacks the context to properly deal with it.

common/app/targeting/CampaignAgent.scala
TBonnin: Regarding Nil, ok fair enough. In this case let's also use `Nil` in the `catch` below

common/app/targeting/CampaignAgent.scala
TBonnin: I would say "Let's not solve a problem we don't have yet" 😁

common/app/targeting/TargetingLifecycle.scala
```diff
@@ -5,24 +5,32 @@ import common.{JobScheduler, AkkaAsync}
...
+        if (Targeting.isSwitchedOn) {
+          jobs.deschedule("TargetingCampaignRefreshJob")
+        }
+      }
     }
 
-    akkaAsync.after1s {
-      CampaignAgent.refresh()
+  override def start(): Unit = {
+    if (Targeting.isSwitchedOn) {
```
currysoup: Sorted, thanks! :)

currysoup:
Moving to version `0.8.0` of the targeting client to support https://github.com/guardian/targeting-client/pull/11 Which adds support for filtering out campaigns with too many rules or tags.

common/app/common/configuration.scala
```diff
@@ -474,6 +474,21 @@ class GuardianConfiguration extends Logging {
     lazy val config = configuration.getMandatoryStringProperty("front.config")
   }
 
+  object targeting {
+    lazy val campaignsUrl = configuration.getStringProperty("targeting.campaignsUrl")
+
+    // Total number of rules allowed per campaign, any campaigns with more than one rule will be filtered
+    lazy val ruleLimit: Int =
+      Try(configuration.getStringProperty("targeting.ruleLimit").get.toInt)
```
dominickendrick: Could this be done using `getIntegerProperty`

currysoup: Thanks! I've added that in my local branch, just trying to figure out why the tests are failing :)

johnduffell: do we even need the limits to be configured (are they different between code and prod?) or can they just be vals in the code?

currysoup: I figured if you wanted to change them it would be easier to override them as config instead of going through a code change hence I implemented them as config with a reasonable default

johnduffell: we already have a lot of config, in my opinion we should save it for private stuff and things that must differ between code and prod where possible, otherwise we end up writing the code in configuration to avoid compiling

currysoup: Moved into the agent itself

common/app/targeting/CampaignAgent.scala
```diff
@@ -10,9 +10,16 @@ object CampaignAgent extends Logging with ExecutionContexts {
...
+    val ruleLimit: Int = 1
+
+    // The combined number of tags allowed in the required and lacking fields of a rule.
+    // Any campaigns with rules with too many tags will be filtered
+    val tagLimit: Int = 20
+
     if (Targeting.isSwitchedOn) {
       Configuration.targeting.campaignsUrl.map(url => {
-        CampaignCache.fetch(url, 100, Some(Configuration.targeting.ruleLimit), Some(Configuration.targeting.tagLimit))
+        CampaignCache.fetch(url, 100, Some(ruleLimit), Some(tagLimit))
```
dominickendrick: Could we move the `fetch` limit into a `val` too for consistency ?

common/app/model/meta.scala
```diff
@@ -14,7 +15,7 @@ import ophan.SurgingContentAgent
 import org.apache.commons.lang3.StringUtils
 import org.joda.time.DateTime
 import org.scala_tools.time.Imports._
-import play.api.libs.json.{JsBoolean, JsString, JsValue}
+import play.api.libs.json.{JsArray, JsBoolean, JsString, JsValue}
```
TBonnin: Is this import necessary?

TBonnin (approved):

dominickendrick:
👍 Great work @currysoup ! Thanks again for being so patient with this ! 🎉

PR #14164, opened 2016-08-31
Update clashing tests with new ContributionsUserTesting test

markjamesbutler:
## What does this change?

Ensures that the email component doesn't run when the ContributionsUserTesting20160831 is running.

No


gtrufitt:
👍

PR #14166, opened 2016-09-01
Set a width on the toast container so we don't cover pagination buttons

gtrufitt:
## What does this change?

The pagination buttons are being covered by the update toast, preventing them from being clicked on > the first page of the liveblog.

Can navigate through the liveblog.

@guardian/dotcom-platform


jfsoul:
Couldn't find a live blog with any updates to see this 😞 

But looks good 👍

PR #14167, opened 2016-09-01
Remove unnecesary config file source.

TBonnin:
## What does this change?

See title ^

This config file doesn't exist anymore after we moved the config file to S3.

Less code 🔪 .  Easier to understand.
## Request for comment

@johnduffell @jfsoul


johnduffell:
😊 thanks 👍

PR #14168, opened 2016-09-01
Fix concat on undefined bug for coverage task

MatthewJWalls:
## What does this change?

I kept getting concat on undefined when running grunt:coverage, as the grunt config for reporters was never set. This just introduces the check

Can run coverage

no
## Screenshots

n/a


gtrufitt:
👍

gtrufitt:
Try merging master in to fix the failing build.

MatthewJWalls:
merged + now builds

PR #14169, opened 2016-09-01
Remove Prebid.js GA tracking

regiskuckaertz:
We've got enough data to assess the behaviour of Prebid.js, and in the meantime we're sending much more events than allowed in the GA free tier.

cc @cb372 @ajosephides


rich-nguyen:
Don't forget https://github.com/guardian/frontend/blob/master/common/app/views/fragments/analytics/google.scala.html#L42 

👍

regiskuckaertz:
Oh, thanks!

PR #13845, opened 2016-08-03
Outbrain email checks

gtrufitt:
## What does this change?
- Show the non-compliant outbrain widget on _any_ email sign-up shown
- Don't show email sign-up if outbrain has already loaded
## What is the value of this and can you measure success?

Doesn't break Outbrain contracts

@guardian/dotcom-platform

PR #14176, opened 2016-09-02
Contributions embed position test

jranks123:
## What does this change?

Sets up A/B tests comparing the contributions interactive embed at the bottom of the article with it inline and in-article.

For the inline in-article variant, the component is inserted programatically via the spacefinder tool, which has the parameters adjusted to make the component appear roughly towards the top of the article. 

We will learn how much the position change can affect impressions to contributions. 

No
## Screenshots

Bottom of article on mobile
bottommobile

Bottom of article desktop
bottomwide

Placement of inline in article component
inarticle
## Request for comment

@gtrufitt


static/src/javascripts/projects/common/modules/experiments/tests/contributions-embed.js
```diff
@@ -23,102 +25,106 @@ define([
...
 
-        this.id = 'ContributionsEmbed20160823';
-        this.start = '2016-08-23';
-        this.expiry = '2016-09-06';
+        this.id = 'ContributionsEmbed20160902';
+        this.start = '2016-09-02';
+        this.expiry = '2016-09-07';
         this.author = 'Jonathan Rankin';
-        this.description = 'Test contributions embed with amount picker.';
+        this.description = '    "Test whether contributions embed performs better inline and in-article than at the bottom of the article.",';
```
katebee: Is the nested quote intentional?

static/src/javascripts/projects/common/modules/experiments/embed.js
```diff
@@ -0,0 +1,49 @@
...
+            el.addEventListener('click', function () {
+                var element = event.currentTarget;
+                var amount = element.getAttribute('data-amount');
+                select(element);
+                setAmount(amount);
+            });
+        });
+    }
+
+    var getSpecifiedParent = function(element, parentClass) {
```
katebee: (subjective: feel free to ignore)
Would it be better to name these `getSpecifiedAncestor`  and `ancestorClass`?

$.js has a `ancestor` function that may be helpful

static/src/javascripts/projects/common/modules/experiments/tests/contributions-embed.js
```diff
@@ -23,102 +25,106 @@ define([
...
+                }
+            };
+        }
 
-        var writer = function (component) {
+        var bottomWriter = function (component) {
 
             return fastdom.write(function () {
                 var a = $('.submeta');
                 component.insertBefore(a);
```
katebee: Injecting things right before the submeta may cause issues with Outbrain, it may be worth checking that all is okay or adding to the Outbrain config: https://github.com/guardian/frontend/pull/13845

static/src/javascripts/projects/common/modules/experiments/tests/contributions-embed.js.orig
```diff
@@ -0,0 +1,138 @@
...
+            });
+
+        };
+
+        var interactiveCompleter = function (complete) {
+            mediator.on('contributions-embed:insert', function (el) {
+                if (el.length < 1 ) {
+                    return;
+                }
+<<<<<<< Updated upstream
```
gtrufitt: You've got merge conflicts here

static/src/javascripts/projects/common/views/contributions-embed.html
```diff
@@ -1,3 +1,39 @@
...
+            </div>
+
+            <div >
+                <p class="form-section">If you use it, if you like it, why not pay for it? It's only fair. Please give to the Guardian.</p>
+                <p class="form-section">Make a one-off contribution:</p>
+            </div>
+            <div class="amount-field <%=bottomClass%>">
+                <div>
+                    <div class="js-button-group button-group u-cf">
+                        <button type="button" tabindex="8" class="js-button option-button option-button--bold <%=bottomClass%>" data-amount=25>
```
gtrufitt: Why have we tabindex on these?

gtrufitt: The amount value isn't quoted here

static/src/javascripts/projects/common/modules/experiments/tests/contributions-embed.js
jranks123: No, thanks!

static/src/javascripts/projects/common/views/contributions-embed.html
```diff
@@ -1,3 +1,39 @@
...
+                        <button type="button" tabindex="8" class="js-button option-button option-button--bold <%=bottomClass%>" data-amount=100>
+                            <span class="currency js-currency">£</span>100
+                        </button>
+                        <button type="button" tabindex="8" class="js-button option-button option-button--bold <%=bottomClass%>" data-amount=250>
+                            <span class="currency js-currency">£</span>250
+                        </button>
+                    </div>
+                </div>
+                <div >
+                    <a class="js-submit-input option-button <%=bottomClass%> contribute action--contribute" href="https://contribute.theguardian.com/uk?INTCMP=co_uk_cobed_like_interactive&amount=50" target="_blank">
```
gtrufitt: Just checking that your JS updates the amount query param on this url?

static/src/javascripts/projects/common/modules/experiments/tests/contributions-embed.js.orig
jranks123: ah, this file is not needed i don't think, i'll delete it

static/src/javascripts/projects/common/modules/experiments/tests/contributions-embed.js.orig
```diff
@@ -0,0 +1,138 @@
+define([
```
jranks123: yep!

static/src/stylesheets/components/sass-mq/_mq.scss
```diff
@@ -33,6 +33,7 @@ $mq-responsive: true !default;
 /// @link https://github.com/sass-mq/sass-mq#seeing-the-currently-active-breakpoint Full documentation and examples
 $mq-breakpoints: (
     mobile:  320px,
+    phablet: 620px,
```
gtrufitt: You've changed a bower module here - we already have phablet set as a breakpoint width in `_vars.scss`, so you shouldn't need this, any reason why you added it?

static/src/stylesheets/module/experiments/_embed.scss
```diff
@@ -0,0 +1,171 @@
...
+
+
+
+.interactive-embed {
+    height: 100%;
+    width: 100%;
+    border-top: 1px solid #4bc6df;
+    background-color: #efefec;
+
+    p {
```
gtrufitt: We generally try and style on classes - I would try and add classes to your elements with the BEM element syntax `.interactive-embed__subtitle` and not nest at all

static/src/stylesheets/module/experiments/_embed.scss
```diff
@@ -0,0 +1,171 @@
...
+        line-height: 24px;
+        font-family: 'Guardian Egyptian Web';
```
gtrufitt: A couple of things throughout this file:

- we shouldn't be adding font-familes directly like this, we have have our [font mixins](https://github.com/guardian/frontend/blob/master/static/src/stylesheets/pasteup/typography/_src.scss#L68) - probably `fs-header` for this. 
- we should be trying to use the $gs-baseline and $gs-gutter variables for paddings and margin ensuring that our layouts are consistent.

static/src/stylesheets/module/experiments/_embed.scss
```diff
@@ -0,0 +1,171 @@
...
+        }
+
+        &.active {
+            background-color: $contribute-yellow;
+            color: $black;
+        }
+    }
+
+    .option-button {
+        @include button();
```
gtrufitt: There's not really any need to define this button as a mixin when you only use it here

static/src/stylesheets/module/experiments/_embed.scss
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+
+    @mixin circular {
+        border-radius: 1000px; // Android 2.3 cannot deal with '50%'
+    }
+
+    @mixin button {
+        @include fs-textSans(4);
+        @include circular;
```
gtrufitt: Probably no need to create a mixin for this

static/src/javascripts/projects/common/views/contributions-embed.html
jranks123: Does it have to be?

static/src/stylesheets/module/experiments/_embed.scss
```diff
@@ -0,0 +1,171 @@
+@import 'svg';
+
+
+$contribute-yellow: #ffbb00;
+$gu-blue: #005689;
```
gtrufitt: Shouldn't have to create variables for this, there are already variables for both - http://guardian.github.io/pasteup-palette/demo/

static/src/stylesheets/module/experiments/_embed.scss
```diff
@@ -0,0 +1,171 @@
...
+$gu-blue-dark: #003d70;
+$comment-support-1: #7d7569;
+$white: #ffffff;
+$black: #000000;
+$gs-gutter: 20px;
```
gtrufitt: I guess these things are being redefined because this was in an interactive, but probably need to clean up now

static/src/javascripts/projects/common/views/contributions-embed.html
jranks123: Is this necessary?

gtrufitt: Technically it isn't but there's nowhere else in the codebase where we don't so we should be trying to keep things consistent

static/src/javascripts/projects/common/views/contributions-embed.html
jranks123: Yes

static/src/stylesheets/module/experiments/_svg.scss
```diff
@@ -0,0 +1,40 @@
+//
+//  Function to create an optimized svg url
+//  See: http://codepen.io/tigt/post/optimizing-svgs-in-data-uris
```
gtrufitt: Interesting! Nice one

static/src/stylesheets/module/experiments/_embed.scss
```diff
@@ -0,0 +1,171 @@
...
+            height: 20px;
+            background-image: svg-url('<svg xmlns="http://www.w3.org/2000/svg" width="40" height="40" viewBox="0 0 40 40"><path fill="white" stroke="white" d="M22.8 14.6L15.2 7l-.7.7 5.5 6.6H6v1.5h14l-5.5 6.6.7.7 7.6-7.6v-.9"/></svg>');
+            background-size: 100%;
+            float: right;
+        }
+    }
+
+
+    @include mq($from: phablet) {
+        .bottomClass.option-button {
```
gtrufitt: Try and use BEM modifiers - e.g `.option-button--bottom` than chaining classes

static/src/javascripts/projects/common/modules/experiments/tests/contributions-embed.js
gtrufitt: Indeed, you'll need to also add it to the work @markjamesbutler did: https://github.com/guardian/frontend/blob/1b4b90f11a6554666b3b6b39c085b5727808ff51/static/src/javascripts/projects/common/modules/experiments/ab-test-clash.js

static/src/javascripts/projects/common/modules/experiments/embed.js
```diff
@@ -0,0 +1,49 @@
...
+
+    function select(el) {
+        $(getSpecifiedParent(el, 'js-button-group').querySelectorAll('.js-button')).removeClass(ACTIVE_CLASS);
+        $(el).addClass(ACTIVE_CLASS);
+    }
+
+    function setAmount(amount) {
+        $('input.' + AMOUNT_CLASS).val(amount);
+        var a = document.querySelector('.js-submit-input');
+        a.href = 'https://contribute.theguardian.com/uk?INTCMP=co_uk_cobed_like_interactive&amount=' + amount.toString();
```
gtrufitt: Ah yeah, cool

static/src/javascripts/projects/common/views/contributions-embed.html
jranks123: Ok cool. I was just checking for future reference whether this causes any problems! Will change to keep the style.

static/src/stylesheets/components/sass-mq/_mq.scss
jranks123: Actually that breakpoint is a different size: I need 620px rather than 660px. Is it ok to define a new 620px breakpoint with a non-conflicting name in my own cs file (embed.css)?

static/src/javascripts/projects/common/views/contributions-embed.html
gtrufitt: :) 👍

static/src/stylesheets/components/sass-mq/_mq.scss
gtrufitt: We tend to always design to the current breakpoint widths, as things can get difficult to understand quickly if the design of your component changes at different times to the rest of the site, in fact looking at the use of the mq mixin, there's only 2 cases of using anything other than the defined widths. What's your reasoning behind needing 620 instead of 660?

static/src/javascripts/projects/common/modules/experiments/embed.js
regiskuckaertz: Or you can use `common/utils/closest`: `closest(element, cssSelector)`

static/src/javascripts/projects/common/modules/experiments/embed.js
```diff
@@ -0,0 +1,49 @@
...
+            if (i > 10) {
+                throw 'You are either traversing a lot of elements! Is this wise? Or your $element argument is undefined';
+            }
+            i++;
+        } while (element && !element.classList.contains(parentClass));
+        return element;
+    };
+
+    function select(el) {
+        $(getSpecifiedParent(el, 'js-button-group').querySelectorAll('.js-button')).removeClass(ACTIVE_CLASS);
```
regiskuckaertz: minor speedup: `$(....querySelectorAll(ACTIVE_CLASS)).removeClass(ACTIVE_CLASS)`

static/src/stylesheets/module/experiments/_embed.scss
jranks123: So with this one i think this is neccassery. Due to the nature of this test, we have _basically_ the same component used in different places, but there are some subtle differences in how they behave at breakpoints. I could make 2 html templates and give them different classes, but that would mean double work whenever updating the html, so what I do is when I form the html, I optionally pass the bottom class as a parameter, which is added to the couple of elements that behave differently.

jranks123: So basically this test has 2 variations of the same element, the only difference being how 4 of the elements react at this breakpoint for one of the variants. So i pass in the bottomClass class as a parameter when forming the template, which means I can tag all the elements that need special treatment at this breakpoint for the variant. I could add .option-button--bottom and contribute--bottom and amount-field--bottom etc as parameters, but then I'm passing loads of parameters into the template. So thats why it is done this way, do you still think I should use BEM? Totally happy to do so, just wanted to explain the trade-off first

gtrufitt: I think what you're doing is replacing the BEM way of doing modifiers with a changing classes way - you shouldn't need to pass loads more parameters into the template, you can do instead of:

```
<div class="amount-field <%=bottomClass%>">
                 <div>
                     <div class="js-button-group button-group u-cf">
                         <button type="button" tabindex="8" class="js-button option-button option-button--bold <%=bottomClass%>" 
```

this: 

```
<div class="amount-field amount-field--<%=bottomClass%>">
                 <div>
                     <div class="js-button-group button-group u-cf">
                         <button type="button" tabindex="8" class="js-button option-button option-button--bold option-button--<%=bottomClass%>" 
```

And style on `.option-button--bottom` (probably don't need the word class and classes should be hyphenated btw) and `.amount-field--bottom`

Does that make sense?

jranks123: ah yes that looks great. Ok thanks!

static/src/javascripts/projects/common/modules/experiments/embed.js
jranks123: This didn't work :(

gtrufitt:
Based on making the changes on the comments 👍 - If you've got anything else to discuss we can pick it up Monday dude!

static/src/javascripts/projects/common/modules/experiments/tests/contributions-embed.js
jranks123: yep, dealt with this now

static/src/stylesheets/components/sass-mq/_mq.scss
jranks123: this wasn't actually being used anyway! deleted

static/src/stylesheets/module/experiments/_embed.scss
```diff
@@ -0,0 +1,144 @@
+@import 'svg';
+$news-main-1-dark: #003d70;
+
+
+
+.interactive-embed  {
```
gtrufitt: Is `interactive-embed` the right class here now? Don't think it's an interactive anymore?

katebee:
Not sure how common this is, but the mid-article variant is colliding with the rich-link: 

https://www.theguardian.com/uk-news/2016/sep/08/uk-workers-zero-hours-contracts-rise-tuc

picture 219

katebee:
Nevermind! I can't replicate it 😄

dominickendrick:
I can replicate it !! 
1. start on a small screen width
2. load the page
3. expand the page width
4. The items then overlap.

This may be a mega edge case tho!

overlap

gtrufitt:
I'm not sure we would want to be running spacefiller and moving stuff around on resize though?

@jranks123 I would maybe increase the distance from rich links in the spacefiller rules to prevent this happening though. What Kate is seeing might be explained by a rich link being included with spacefiller before enhancing and then changing height when it enhances - again this could be resolved by having more distance from rich links, probably reverting to the default of 250px (I think?)

katebee:
It's happened again, no resize, just opened an article in a new tab.

**After page load:**

picture 221

I think it is because **I was also being shown the email signup** at the bottom of the article (Which also has to check Outbrain version before running?). Once I closed this, the contributions embed moved to a better position.

**Email Sign-up:**

picture 222

**Contributions Embed after closing Email Sign-up:**

picture 223

PR #14177, opened 2016-09-02
Add snap banner to promote Weekend Reading

katebee:
## What does this change?

Return of the Habit Forming snaps! (see #13659)

Adds a test for the Weekend Reading promo. This test will show users a banner promoting the Weekend Reading email.

The banner co-opts the Membership Message banner to display a message to the visitor (see screenshot). It has a lower priority than the Breaking News banner, and will not appear again once a visitor has dismissed it.
- Users will only see the snap once
- Breaking news gets priority

The sign-up button will open a sign-up page in a new tab 
- Snap banner promotes the Weekend Reading email and leads to visitors signing up
- We can track interactions with the banner (through the CMP param on the URL) and see if these users complete the email sign-up process

No
## Screenshots

picture 207
## Request for comment

@joelochlann


static/src/javascripts/projects/common/modules/experiments/tests/weekend-reading-promo.js
```diff
@@ -0,0 +1,77 @@
...
+    uniq,
+    weekendReadingPromo
+) {
+    return function () {
+        this.id = 'WeekendReadingPromo';
+        this.start = '2016-09-01';
+        this.expiry = '2016-09-12';
+        this.author = 'Kate Whalen';
+        this.description = 'For just one pageview, show users a banner promoting the Weekend Reading email';
+        this.audience = 0;
```
jranks123: Should this be 0?

joelochlann:
👍  Looks good! Just wondering, should we really be using the a/b testing framework for this now? Isn't this now a "real thing"?

katebee:
@joelochlann good point; it could probably be moved to the feature switches 😅 

I wanted a quick way of checking if it should run, and an easy way of toggling it on/off
(when all you have is a :hammer:)

joelochlann:
The obvious problem I can foresee is that the switch will expire in a week - I imagine we probably don't want that

katebee:
That was actually my intention! I didn't know how long we were going to have the recruitment drive for, so decided to force myself to review it Monday, to ensure that we don't keep this live longer than needed

PR #14181, opened 2016-09-05
Remove Rio 2016 Olympics from nav

gtrufitt:
## What does this change?

Remove Rio 2016 Olympics from nav

@guardian/dotcom-platform


jfsoul:
👍

PR #14182, opened 2016-09-05
Extend link preconnect switch

rich-nguyen:
Extending this switch, we still haven't confirmed what the impact is of using `link-preconnect`


regiskuckaertz:
👍

PR #14184, opened 2016-09-05
Fix failing line of setup.sh script

MatthewJWalls:
## What does this change?

Fixes the setup.sh to allow it to run, by removing the accidentally left-in awscli install call (the function for which was removed in a previous commit)

success measured by the script not failing to run

no
## Screenshots

n/a
## Request for comment

The awscli install function was removed in a previous commit, but
the call was left in - and causes the script to fail. This change
removes the call

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -145,7 +145,6 @@ main() {
   check_encryption
   create_aws_config
   install_homebrew
-  install_awscli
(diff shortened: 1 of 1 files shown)


johnduffell:
thanks! 👍 😊

PR #14185, opened 2016-09-05
Remove dead 404 test pages and routes

katebee:
## What does this change?
- Removes the routes for old 404 tests
- Removes (the subsequently) unused functions from Controller class
- Removes dead html and css 💀 
- What is dead may never die 🐙 

Nope


janua:
:+1:

PR #14188, opened 2016-09-05
Better control of overflow for fabrics

regiskuckaertz:
This is ugly but we'll get rid of this when we move to native ads. Currently the Fabric templates overflow their container and this creates [a jarring and ugly flicker effect](https://www.theguardian.com/uk?adtest=demo_responsive).

Now:
fabric-jump

And:
fabric-jump-2


paperboyo:
May be totally not related, but I've just recently noticed that the ads on the top of the page do not work correctly in latest IE11 and Edge on latest Windows...

regiskuckaertz:
@paperboyo I don't think so, but please if it happens again grab me and I'll have a look 😃

janua:
:+1:

PR #14189, opened 2016-09-05
Add campaign code for the smart app banner

marialivia16:
## What does this change?

Adds a campaign code to the Apple's smart app banner.

We will see in iTunes analytics how many people install our app by clicking on the smart app banner.

No


gtrufitt:
👍

PR #14192, opened 2016-09-05
Update email config

kelvin-chappell:
Emails not being sent, and this error appears in log:
`email.adOpsTeam has type OBJECT rather than STRING`

Seems to be because in the HOCON format, these property names are seen as parts of the same object:
email.adOpsTeam
email.adOpsTeam.us
email.adOpsTeam.au

So I've done a bit of renaming.

/cc @guardian/commercial-dev


rich-nguyen:
👍

PR #14193, opened 2016-09-05
Fix sign in link when register with phone number is off

piuccio:
## What does this change?

Sign in currently looks like this

screen shot 2016-09-05 at 15 54 34

Links  are back to normal

No
## Request for comment

@NathanielBennett hopefully I didn't break the case `RegisterWithPhoneNumber.isSwitchedOn: true`


gtrufitt:
☎️ 👍

NathanielBennett:
:+1: Looks like i broke it. Thanks @piuccio !

PR #14194, opened 2016-09-05
Fix clashing tests bug for Contributions Embed test

jranks123:
## What does this change?

Fixes a bug where the name of the Contributions Embed test was not updated in the clashing tests when the test was launched.

It means that we will be contractually compliant with Outbrain.

No
## Screenshots

n/a
## Request for comment

@gtrufitt


gtrufitt:
👍

PR #14196, opened 2016-09-05
Point Football API at a CDN

gklopper:
## What does this change?

Has a switch which proxies the 3rd party football api through a CDN.

We are hitting the 3rd party too often and they are complaining. Our architecture means we need to make a lot of calls as we do not maintain our own database.

Success is a happy 3rd party.

No, but once it is proved working it could be used by apps too.

### DIFF
--- a/sport/app/conf/SportConfiguration.scala
+++ b/sport/app/conf/SportConfiguration.scala
@@ -8,7 +8,7 @@ object SportConfiguration {
   import GuardianConfiguration._
 
   object pa {
-    lazy val footballHost = PaClientConfig.baseUrl
+    lazy val footballHost = "http://football-api.gu-web.net/v1.5"
     lazy val footballKey = configuration.getMandatoryStringProperty("pa.api.key")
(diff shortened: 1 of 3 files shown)


sport/app/conf/SportConfiguration.scala
```diff
@@ -8,7 +8,7 @@ object SportConfiguration {
   import GuardianConfiguration._
 
   object pa {
-    lazy val footballHost = PaClientConfig.baseUrl
+    lazy val footballHost = "http://football-api.gu-web.net/v1.5"
```
gklopper: This host is used by our online docs and troubleshooter, so I'm not bothering to put a switch on it.

gklopper:
cc @johnduffell @jfsoul

johnduffell:
🚀 👍 if there might be significant traffic it might be worth letting stephen b know in case it affects his calculations

jfsoul:
Thanks for saving me the trouble! 👍 

I was toying with putting the `cachedBase` in s3 config, but probably overkill anyway.

PR #14198, opened 2016-09-05
Update outbrain codes for nonCompliant widget

gtrufitt:
## What does this change?

The non compliant widget promise was resolving with a `nonCompliant` key but wasn't matching for the outbrain codes side of things.

Don't serve non-compliant widgets.

@markjamesbutler @guardian/dotcom-platform


jfsoul:
👍

PR #14171, opened 2016-09-01
New Weekend Reading sign-up page

katebee:
## What does this change?
- Changes the [Weekend Reading 404 test](https://www.theguardian.com/survey/weekendreading) to an actual sign-up page 🎉 
- The path will now be: https://www.theguardian.com/signup/weekendreading  (see screenshot 3)
- The webtitle (page title) will be `Sign up for Guardian weekend reading` (see screenshot 3)
- Adds an AB test to the page so we can try out two different email formats
- Email form now sends sign-up data to ExactTarget, rather than FormStack
## What is the value of this and can you measure success?
- We can promote, and sign up people, to the weekend reading email
- Control and Variant groups go to two separate mailing lists in ExactTarget
- We can evalulate the CTO (click-to-open) rates of the different email formats
- We can track referrers to see which CTA led to a sign up
## Does this affect other platforms - Amp, Apps, etc?

Nope.
## Screenshots

Before:

weekend reading before

After:

weekend reading final

Routing and Webtitle change:

Routing and Webtitle change
## Request for comment

@joelochlann 
@superfrank 
@janua

PR #14199, opened 2016-09-05
Move 'Signup' out of 'Surveys' and separate the Controllers and routes

katebee:
## What does this change?

Separates **Signups** from **Surveys**; it makes more sense for them to have their own folders, templates, controllers and routes

(example signup page: https://www.theguardian.com/signup/weekendreading)
- Moves views and stylesheets relating to **Signup** out of `/surveys` and into new `/signup` folders
- Makes a new controller (SignupController), and moves function out of SurveyController
- Move `/signup` out of the `/survey` section of **Routes**

Bonus: continued tidying of the scss for weekend-reading and signup pages (see: #14171). 
- More logical structure and location for **Signup** pages
- Potential for future, customised, email signup pages
- Codebase and file locations (hopefully) becomes less confusing
## Screenshots

New `/signup` folders:
picture 215


static/src/stylesheets/_common.scss
```diff
@@ -19,6 +19,7 @@
 @import 'module/commercial/_commercial';
 @import 'module/crosswords/_main';
 @import 'module/email/_main';
+@import 'module/signup/_main';
```
janua: I thought we split out the CSS for these pages when it was survey? Can we do that for signup?

katebee:
@janua: Thanks, I've looked into how the CSS was separated for `survey` and hopefully done a better job of replicating it this time

janua:
Looks good! :+1:

PR #14201, opened 2016-09-06
made contributions epic element

jranks123:
## What does this change?

Creates the Contributions Epic test, which is testing a longer, editorial written message for the contributions ask. 

Test whether contributions embed performs better our previous in-article component tests.

No 
## Screenshots

screenshot at sep 06 12-04-24
## Request for comment

@gtrufitt

### DIFF
--- a/common/app/conf/switches/ABTestSwitches.scala
+++ b/common/app/conf/switches/ABTestSwitches.scala
@@ -64,6 +64,18 @@ trait ABTestSwitches {
     exposeClientSide = true
   )
 
+  val ABContributionsEpic20160906 = Switch(
+    SwitchGroup.ABTests,
+    "ab-contributions-epic-20160906",
+    "Test whether contributions embed performs better inline and in-article than at the bottom of the article.",
+    owners = Seq(Owner.withGithub("jranks123")),
+    safeState = Off,
+    sellByDate = new LocalDate(2016, 9, 12),
+    exposeClientSide = true
+  )
+
+
+
   val ABParticipationDiscussionOrderingLiveBlogs = Switch(
     SwitchGroup.ABTests,
     "ab-participation-discussion-ordering-live-blog",
--- a/common/app/conf/switches/ABTestSwitches.scala
+++ b/common/app/conf/switches/ABTestSwitches.scala
@@ -67,7 +67,7 @@ trait ABTestSwitches {
   val ABContributionsEpic20160906 = Switch(
     SwitchGroup.ABTests,
     "ab-contributions-epic-20160906",
-    "Test whether contributions embed performs better inline and in-article than at the bottom of the article.",
+    "Test whether contributions embed performs better than our previous in-article component tests.",
     owners = Seq(Owner.withGithub("jranks123")),
     safeState = Off,
     sellByDate = new LocalDate(2016, 9, 12),
--- a/static/src/javascripts/projects/common/modules/experiments/tests/contributions-epic.js
+++ b/static/src/javascripts/projects/common/modules/experiments/tests/contributions-epic.js
@@ -32,14 +32,14 @@ define([
         this.start = '2016-09-06';
         this.expiry = '2016-09-12';
         this.author = 'Jonathan Rankin';
-        this.description = 'Test whether the long ask performs better than _____.';
+        this.description = 'Test whether contributions embed performs better than our previous in-article component tests.';
         this.showForSensitive = false;
-        this.audience = 0.10;
+        this.audience = 0.05;
         this.audienceOffset = 0.33;
         this.successMeasure = 'Impressions to number of contributions';
         this.audienceCriteria = 'All users';
         this.dataLinkNames = '';
-        this.idealOutcome = 'The embed performs 20% better inline and in-article than it does at the bottom of the article';
+        this.idealOutcome = 'The embed performs at least as good as our previous in-article component tests';
         this.canRun = function () {
             var pageObj = config.page;
             return !(pageObj.isSensitive || pageObj.isLiveBlog || pageObj.isFront || obWidgetIsShown() || pageObj.isAdvertisementFeature) && pageObj.edition === 'UK';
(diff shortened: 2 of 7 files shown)


common/app/conf/switches/ABTestSwitches.scala
```diff
@@ -67,13 +67,25 @@ trait ABTestSwitches {
   val ABContributionsEmbed20160905= Switch(
     SwitchGroup.ABTests,
     "ab-contributions-embed-20160905",
-    "Test whether contributions embed performs better inline and in-article than at the bottom of the article.",
+    "Test whether contributions embed performs better our previous in-article component tests.",
```
gtrufitt: missing 'than'

common/app/conf/switches/ABTestSwitches.scala
```diff
@@ -67,13 +67,25 @@ trait ABTestSwitches {
...
     owners = Seq(Owner.withGithub("jranks123")),
     safeState = Off,
     sellByDate = new LocalDate(2016, 9, 9),
     exposeClientSide = true
   )
 
+  val ABContributionsEpic20160906 = Switch(
+    SwitchGroup.ABTests,
+    "ab-contributions-epic-20160906",
+    "Test whether contributions embed performs better our previous in-article component tests.",
```
gtrufitt: 'than' again :)

static/src/javascripts/projects/common/modules/experiments/tests/contributions-epic.js
```diff
@@ -0,0 +1,88 @@
+define([
+    'bean',
+    'qwery',
+    'common/utils/$',
+    'common/utils/template',
+    'common/views/svg',
+    'common/utils/fastdom-promise',
+    'common/utils/mediator',
+    'text!common/views/contributions-epic.html',
```
gtrufitt: Is there any way you can combine all the contributions templates into one and pass everything in as variables? The reason I ask is that AB tests are bundled into standard.js which is the absolute bottom baseline of JS - so AB test should be kept as small and lean as possible and reducing the number of included modules is an important part of that.

static/src/javascripts/projects/common/views/contributions-epic.html
```diff
@@ -0,0 +1,14 @@
+<div class="contributions__epic">
+    <div>
+        <h3 class="contributions__title contributions__title-epic">
+            Since you're here ...</h3>
+        <p class="contributions__paragraph contributions__paragraph-epic">... we have a small favour to ask. More people are reading the Guardian than ever. But far fewer are paying for it. And advertising revenues are falling
```
gtrufitt: Modifiers have double hyphens: `contributions__paragraph--epic`

static/src/stylesheets/module/experiments/_embed.scss
```diff
@@ -6,18 +6,36 @@
     padding: 0 $gs-gutter / 4 $gs-baseline $gs-gutter / 4;
 }
 
+.contributions__epic {
+    border-top: 1px solid $multimedia-main-2;
+    background-color: #efefec;
```
gtrufitt: This is `$comment-support-2`

static/src/stylesheets/module/experiments/_embed.scss
```diff
@@ -6,18 +6,36 @@
...
     font-size: 1em;
     line-height: 20px;
     font-family: $f-serif-headline;
     margin-bottom: $gs-baseline;
 }
 
+.contributions__paragraph-epic {
+    line-height: 1.5;
+    font-family: $f-serif-text;
+    -moz-osx-font-smoothing: grayscale;
```
gtrufitt: Shouldn't need any font-smoothing in here. Also, we should be using the mixins - did you work out what was different with the one yesterday?

gtrufitt:
Just wondering if Ben has taken a look at this? Feels pretty blocky and hard to read, like it could do with paragraphs (and the component was designed for a small line of text so not sure how well it lends to the large block of text). 

If he's happy with it then no worries.

static/src/javascripts/projects/common/modules/experiments/tests/contributions-epic.js
jranks123: I think giraffe-message.html and contributions-epic.html should be combined but if you look at contributions-embed, it is really quite different, for example I had to put a container div around the buttons and the contribution button because their orientation changed depending on which variant of the test it was running, plus it's wrapped in a figure tag, which is not necessary for the other 2. Could I just combine the first 2?

gtrufitt: Whatever you can do to reduce the size of the standard.js is good 👍

jranks123: Out of interest, what is a bigger bottleneck, the combined size of the files or the number of them?

gtrufitt: They're both issues for different reasons - the combined size will mean that the JS is larger coming down the wire and hence take longer and the more files there are, the more work the module loader has to do when it is processing the JS file.

jranks123: So for now, I do not believe that it is worth creating a multipurpose template as one of the tempaltes we would be combining is actually going to be retired very soon, possibly today or tomorrow (it's the giraffe-message.html). This is because the results show it isn't as effective as the other one. As we will be iterating on the contributions-epic.html template, what makes more sense is to make this more paramatised next time we iterate it.

jranks123:
Thanks @gtrufitt , Ben agreed with you on the design. I have included an updated screenshot. Do you agree with my comment on the refactor and if so, are you happy with the PR? Thanks

static/src/javascripts/projects/common/modules/experiments/tests/contributions-epic.js
gtrufitt: Fair enough, one to keep in mind for the future 👌

gtrufitt:
🆒 👍

PR #14202, opened 2016-09-06
Audience for comments in react

piuccio:
## What does this change?

Start showing the react/preact version of comment count to users.

The test is measuring the correct loading of comments count. I don't expect the variants to perform better, but I want to see if (p)react breaks on some of the browsers in the wild.

It's quite possible that the ratio of comments count loaded is lower for (p)react simply because it takes longer to load the assets and people leave the page in the mean time.
Assets are loaded asynchronously so they won't affect page load.

No


janua:
LGTM :+1:

PR #14203, opened 2016-09-06
Implement ignored sentry action tags

MatthewJWalls:
## What does this change?
- Implements a new configuration value for ignored sentry actions (action is a tag set on some errors before passing them to raven-js for logging out to sentry)
- Alters the raven-js shouldSendCallback hook to check the action tag of the exception before sending to Sentry.

This could also have been implemented as a special "ignored" tag which may have been cleaner as it would require no config changes on the scala side, but it would have meant this change would not be retroactive and we'd have to go back and put the ignored tag on all the existing ads code.

Should reduce the number of crashes being sent to sentry due to ad-blocker related exceptions by several hundred thousand a month.

no
## Screenshots

n/a

### DIFF
--- a/static/src/javascripts/bootstraps/standard/main.js
+++ b/static/src/javascripts/bootstraps/standard/main.js
@@ -75,9 +75,7 @@ define([
                 },
                 shouldSendCallback: function (data) {
                     var isDev = config.page.isDev;
-
-                    var isIgnored = typeof(data.tags.action) !== 'undefined' &&
-                        config.page.ignoredSentryActions.indexOf(data.tags.action) != -1;
+                    var isIgnored = typeof(data.tags.ignored) !== 'undefined' && data.tags.ignored;
 
                     if (isDev && !isIgnored) {
                         // Some environments don't support or don't always expose the console object
--- a/static/src/javascripts/bootstraps/standard/main.js
+++ b/static/src/javascripts/bootstraps/standard/main.js
@@ -75,7 +75,11 @@ define([
                 },
                 shouldSendCallback: function (data) {
                     var isDev = config.page.isDev;
-                    if (isDev) {
+
+                    var isIgnored = typeof(data.tags.action) !== 'undefined' &&
+                        config.page.ignoredSentryActions.indexOf(data.tags.action) != -1;
+
+                    if (isDev && !isIgnored) {
                         // Some environments don't support or don't always expose the console object
                         if (window.console && window.console.warn) {
                             window.console.warn('Raven captured error.', data);
@@ -84,6 +88,7 @@ define([
 
                     return config.switches.enableSentryReporting &&
                         Math.random() < 0.1 &&
+                        !isIgnored &&
                         !isDev; // don't actually notify sentry in dev mode
                 }
             }
--- a/common/app/views/support/JavaScriptPage.scala
+++ b/common/app/views/support/JavaScriptPage.scala
@@ -7,7 +7,7 @@ import conf.Configuration.environment
 import model._
 import play.api.Play
 import play.api.Play.current
-import play.api.libs.json.{JsValue, JsBoolean, JsString, Json}
+import play.api.libs.json.{JsBoolean, JsString, JsArray, JsValue, Json}
 import play.api.mvc.RequestHeader
 
 object JavaScriptPage {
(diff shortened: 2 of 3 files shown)


static/src/javascripts/bootstraps/standard/main.js
```diff
@@ -75,7 +75,11 @@ define([
                 },
                 shouldSendCallback: function (data) {
                     var isDev = config.page.isDev;
-                    if (isDev) {
+
+                    var isIgnored = typeof(data.tags.action) !== 'undefined' &&
+                        config.page.ignoredSentryActions.indexOf(data.tags.action) != -1;
```
jfsoul: `!==` on the second comparison here?

Also is undefined specifically a special case here?  If not: 

    var isIgnored = data.tags.action && ...

gtrufitt:
I could only find one location where there was an ['ads' action - in the enhanced media js](https://github.com/guardian/frontend/blob/master/static/src/javascripts/bootstraps/enhanced/media/main.js#L446)

```
raven.captureException(e, { tags: { feature: 'media', action: 'ads' } });
```

Exactly what you mentioned yesterday on the ima stuff which looks like it tries to run preroll and if the request is blocked, just plays the video.

I can see the value of the ignore functionality but it seems that in this case we'd only be removing this one require error, where it might as well fail silently and not log to sentry anyway? Am I missing something?

MatthewJWalls:
@gtrufitt Not missing anything, I was going to propose taking what we've done in the media js and making it the standard, this change is only intended to implement a way to ignore the crashes going into sentry, not to universally fix all of them - we still need to go back and update the existing error handling for the other cases.

static/src/javascripts/bootstraps/standard/main.js
MatthewJWalls: @jfsoul 

right on the != needing to be !==

As for just relying on the falsyness of undefined; it depends what we want to do. It is possible for the action field not to be set, and I've always considered that since JS has had some historically (currently?) weird rules for equality handling, that being explicit about the intent of the code trumps the extra characters, hence me putting in explicitly !== undefined since that carries with it the hint to the programmer that the value can actually be undefined as an expected case.

gtrufitt:
Ah ok, that's what I thought. So does raven capture errors globally right now, so we *have* to specifically capture the error and then ignore it to prevent it from going to sentry?

MatthewJWalls:
@gtrufitt yeah, raven installs a global handler to catch everything by default, but you can use the ravenjs's setUserContext, wrap or explicit try/catch to add extra tags (such as 'action') where appropriate.

I think this is correct; send everything to sentry unless you decide not to locally.

gtrufitt:
Riiight... makes sense now, cheers! Only thing I would maybe do as add something that specifically states it as an adblock ignore (which I know you addressed but think it would be clearer in the future if we could do that) but other than that 👍 

Here's to useful sentry! 📉

MatthewJWalls:
I was having a bit of analysis paralysis on this because of the different ways of doing it.

If we're going to use a more specific 'ignored' tag then there doesn't really need to be any scala side config, which is nice. And in fact I might as well really just delete the one line of code that raises an exception to raven in the media/main.js file - but I'm a bit hesitant about that since I don't really want to put in a standard that says "swallow the exception and don't rethrow it" in case we change our minds in the future.

So I think I'll just remove the scala side configuration, and have the shouldSendCallback simply check for a tag on the exception like doReport being true or something?

gtrufitt:
Indeed, apologies for adding to that...

> So I think I'll just remove the scala side configuration, and have the shouldSendCallback simply check for a tag on the exception like doReport being true or something?

Yeah, if we have to explicitly handle errors to avoid them being caught by raven's global error handling then dealing with them this way makes sense. Though the default being report and explicitly overriding to *not* report might be clearer?

static/src/javascripts/bootstraps/standard/main.js
```diff
@@ -75,7 +75,9 @@ define([
                 },
                 shouldSendCallback: function (data) {
                     var isDev = config.page.isDev;
-                    if (isDev) {
+                    var isIgnored = typeof(data.tags.ignored) !== 'undefined' && data.tags.ignored;
```
piuccio: do all `data` have `tags`? I've never seen `tags` on captured exceptions before

`data.tags ? data.tags.ignored === true : false` probably works better

MatthewJWalls: @piuccio I believe the Raven lib wraps it before being handed to the shouldSendCallback which adds on the extra data like tags, so I think tags will exist, but ignored is not necessarily set as one of those tags.

piuccio: I see

PR #14204, opened 2016-09-06
Send media and discussion events to the prod GA property

cb372:
## What does this change?

Send media events (video/audio playback events) and the "scrolled down to comments" event to the production GA property.

We have been sending these events to the test property for a while and they look fine.

Really running out of things to write here.

No
## Screenshots

n/a
## Request for comment

@ajosephides @akash1810 @dominickendrick


ajosephides:
👍

akash1810:
👍

PR #14206, opened 2016-09-06
add to to copy

jranks123:
## What does this change?

Fixes a copy typo in the Contributions Epic component 

The copy is now correct, meaning that our readers will better understand what we are trying to say. 

No
## Screenshots

screenshot at sep 06 16-18-50


akash1810:
👍

PR #14208, opened 2016-09-06
add "INTCMP" copy to href on contribute link

jranks123:
## What does this change?

Adds in the INTCMP part of the href on the contribute button in the Contribute Epic Test.

It means we can track people who came to the contributions page via the test. Without this we cannot!


desbo:
👍

PR #14211, opened 2016-09-06
Added width of 1 gs-span to date block from tablet upwards, to force …

Ap0c:
## What does this change?

Adds a width of 1 gs-span to the date block on a liveblog, to force long dates to wrap.

Fixes a problem where the full dates displayed on old liveblogs were overlapping with the content.

…wrap of full dates.


gtrufitt:
Gotta before/after screenshot? 📷

Ap0c:
before-after

Applies to tablet -> wide. Mobile is unchanged as layout is different (date shown above each item rather than to the side).

gtrufitt:
Ah yes, perfect 👍

PR #14212, opened 2016-09-06
Dropped margin rule that was resetting margin on focus, causing tag t…

Ap0c:
…o jump to the left.
## What does this change?

Removes a margin reset (to 0) when brand bar elements are focused.

Fixes a problem where link to "become a supporter" was jumping to the left when receiving focus.

### DIFF
--- a/static/src/stylesheets/module/nav/_brand-bar.scss
+++ b/static/src/stylesheets/module/nav/_brand-bar.scss
@@ -87,7 +87,6 @@
 
 .brand-bar__item--action,
 .brand-bar__item--action:focus {
-    margin: 0;
     cursor: pointer;
     color: $c-guardian-services-action;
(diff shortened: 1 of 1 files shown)


gtrufitt:
That `margin: 0` still applies if the element isn't focussed though, are you sure that's the cause?

Ap0c:
Yeah I double-checked. When the element isn't focused the `.brand-bar__item .brand-bar__item--split--first` rule (same file, line 46 in this commit) is more specific, and overrides the left margin set by the `margin: 0` shorthand.

When focused they have the same specificity, and the :focus rule takes precedence because it's further down the stylesheet, setting `margin-left` to 0 and causing the jump to the left.

gtrufitt (approved):
Cool, sorry lgtm

PR #14213, opened 2016-09-06
Docs: Add port and browsersync mention for make watch

gtrufitt:
As title.


desbo:
Should these docs mention `m.thegulocal.com` (if you use that, it'll try port 3000 falling back to 9000) and maybe how we use nginx in general?

gtrufitt:
Yeah... I have never used m.thegulocal.com 😆

PR #14215, opened 2016-09-07
Reqwest doesn't work with discussion API

piuccio:
## What does this change?

`reqwest` works now because we're proxying requests through `discussion` app. However `discussion-frontend` talks directly to discussion API which is not configured to handle `X-Requested-With`.
Until we fix the API, use global fetch when available.

I can start the AB test which would otherwise always fail loading comment count

No


johnduffell:
👍 but consider removing the commented out code and moving it to the dapi backlog

johnduffell:
just in case the dapi change is never done/changed mind about

PR #14220, opened 2016-09-07
Show more useful message for dev config problems

kelvin-chappell:
This is what you see in the dev environment if your credentials are out of date:

image

If app config fails it's probably because credentials need refreshing, but that's not immediately obvious.  So this change gives you a hint:

image


johnduffell:
a lot better message 👍 although I always dislike "probably" in errors!

PR #14221, opened 2016-09-07
Create new epic test with amount buttons

jranks123:
## What does this change?

Creates a new test running alongside the currently running epic test, testing the epic against the epic with the amount buttons in. We are running this test on a distinct audience segment from the previous epic test and they are not dependent on each other.

Also, remove ABContributionsArticle20160822 test

It should increase the rate of impressions to conversions, 

No. 
## Screenshots

control: 
screenshot at sep 07 12-59-28

buttons:
screenshot at sep 07 12-59-56


desbo:
👍

PR #14223, opened 2016-09-07
Remove misleading setup.sh message

rtyley:
The script no longer downloads a file from S3, no matter whether your disk is encrypted or not, so the message was misleading.

cc @johnduffell


johnduffell:
brill, thanks Roberto! 👍 🚀

PR #14224, opened 2016-09-07
Editionalisted nav footer

NataliaLKB:
## What does this change?

The "Nav footer" links (can you think of a better name?) in the new nav are editionalised too. This adds the correct links per each edition.

This is a picture of the links:
image

No styling is changed.

Nope!

@guardian/dotcom-platform


SiAdcock:
Very tidy change 👍

PR #14226, opened 2016-09-07
Fix liveblog updated timestamp

Ap0c:
## What does this change?

Reduces the scope of the width rule used to make timestamps wrap on liveblogs; now only applies to published timestamps. It was previously unintentionally also affecting updated timestamps.

Fixes a problem with liveblog updated timestamps caused by PR #14211 (sorry, my bad!).
## Screenshots

updated-before-after


gtrufitt:
👍

PR #14230, opened 2016-09-07
Ad-block pop-over edits

JustinPinner:
## What does this change?

The text of the adblock detection message was too long. Now it's a bit shorter.
We've added a simple graphic to quickly enable whitelisting in AdBlock/Plus.
We've made the background a little more opaque.
It won't show to first-time visitors. It will wait until a few pieces of content have been read.
It won't show on ad-prohibited content.
It will only activate for Chrome users. For now.
It's still a **zero percent test**. For now.
The data lake properties for the dismiss button have been changed so we can tell which is which.
The test definitions have been updated.

This is an iteration of the previous internal test version (https://github.com/guardian/frontend/pull/14072)
**We still have to define audience targeting with the data team.**

No.
## Screenshots

(latest screenshot)
screen shot 2016-09-08 at 18 32 01
## Request for comment

@regiskuckaertz (sorry to keep picking on you!) @rich-nguyen


static/src/javascripts/projects/common/modules/experiments/tests/adblocking-response.js
```diff
@@ -25,58 +25,58 @@ define([
...
                 detect.adblockInUse.then(function (adblockUsed) {
-                    if (adblockUsed && !config.page.isFront &&
-                        !userFeatures.isPayingMember() &&
-                        config.page.webTitle !== 'Subscriber number form' &&
-                        config.page.webTitle !== 'How to disable your adblocker for theguardian.com' &&
-                        storage.local.isStorageAvailable() &&
-                        !storage.local.get('gu.subscriber')
-                    ) {
+                    if (adblockUsed && isQualified()) {
                         var surveyOverlay = new SurveyAdBlock({
```
regiskuckaertz: Couldn't we have only one instance of this object and use it in all `new SurveyAdBlock()` calls?

static/src/javascripts/projects/common/views/commercial/survey/survey-adblock.html
```diff
@@ -24,33 +29,53 @@ <h3 class="survey-text__header">
...
+                    <p class="survey-text__strong">
+                        <%=surveyTextSecond%>
+                    </p>
                 <% } %>
                 <% if (surveyTextThird) { %>
                 <p class="survey-text__info">
                     <%=surveyTextThird%>
                 </p>
                 <% } %>
+                <% if (membershipLink || signupLink) { %>
```
regiskuckaertz: As a token of caution, be aware that this might break the built module --- at least it happened to me in the past, but I _think_ that was because of quotation marks.

regiskuckaertz:
Aside minor comments, looks good 👍

paperboyo:
10x smaller pngs (and better quality), but I haven't changed the main one so that the black borders in the above screenshot would disappear...

ad-block-instructions-chrome

ab-icon
abp-icon

static/src/javascripts/projects/common/views/commercial/survey/survey-adblock.html
JustinPinner: Hi @regiskuckaertz can you explain this a bit further? What might break and how? Thanks.

regiskuckaertz: Better test with `make compile` and `useAssetBundles=true`. In my previous experience, when a lodash template had a test like that, the `text!` plugin created a syntax error in the JS build. But again, I suspect this only happened when my test contained quotes, like `x === 'y'`.

JustinPinner: Thanks. I just ran `make compile useAssetBundles=true` and it looks like it completed ok. Should I expect a failure in the compile, or a runtime problem? I assume you meant to pass the `useAssetBundle=true` to the `make compile` command?

regiskuckaertz: Oh that was sth that used to be in `frontend.properties` but as we got rid of that I'm not sure it still is possible. A failure should occur at runtime, if there is a failure, which I doubt.

PR #14235, opened 2016-09-08
Retrieve original location href via iframe hash

janua:
## What does this change?

This changes the AMP Bluelinks AB test tracking so that we get the query parameter from the JSON object that is included on the `window.location.hash` of the iframe. This is in `JSON` format and includes the original context through `_context`.
## Request for comment

@NataliaLKB @JamieKMorrison

### DIFF
--- a/facia/public/amp_remote.html
+++ b/facia/public/amp_remote.html
@@ -127,13 +127,23 @@
         function attachQueryParams(config) {
             //Google AMP AB test query parameter
             var whitelist = ['0p19G'];
-            var paramsMap = getUrlVars(config.targeting.url);
+            var originalLocationHref = getOriginalLocationHref();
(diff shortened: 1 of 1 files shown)


NataliaLKB:
👍

facia/public/amp_remote.html
```diff
@@ -136,6 +137,15 @@
             return config;
         }
 
+        function getOriginalLocationHref() {
+            try {
+                var configString = window.location.hash.substr(1);
+                var config = JSON.parse(configString);
+                return config._context.location.href;
+            } catch(e){}
+            return "";
```
akash1810: do you need to return empty string?

janua: No, actually I don't.

PR #14236, opened 2016-09-08
Render fronts as email-friendly HTML

joelochlann:
## What does this change?

Extends @desbo's work on rendering articles with email-friendly HTML, which began in #11673. Now fronts can also be rendered as emails with either the `/email` slug or `?format=email` parameter. e.g. http://www.theguardian.com/politics?format=email

Currently, emails which consist mainly of links ("email collections") are generated by the [gu-email-renderer](https://github.com/guardian/gu-email-renderer). This has two main limitations:
- Emails are rendered in an old R2 design
- Changes to the structure & contents of certain emails require developer involvement

Rendering fronts as emails gives us an opportunity to reskin our link-based emails while also giving control back to editorial via the fronts tool. Editorial staff can now apply all the flexibility of fronts to emails, including:
- Determine what sections appear in an email
- Choose specific articles to appear in those sections
- Autopopulate sections with CAPI queries (via backfill)
- Override kickers, headlines, trail text etc if desired
- Choose external links to populate certain sections
- Add custom text to the top of the email
- etc...

The metrics that we're hoping to ultimately influence with all this control are:
- Click-To-Open rate of our link-based emails (proportion of people that read an email who click on at least one link in it)
- therefore web traffic from email products
- ...therefore visit frequency for non-regulars

Just web
### The front

screencapture-m-thegulocal-email-the-flyer-1473337874565
### The email

screencapture-m-thegulocal-email-the-flyer-1473337825739
## Request for comment

@desbo @katebee

### DIFF
--- a/article/app/views/fragments/emailArticleBody.scala.html
+++ b/article/app/views/fragments/emailArticleBody.scala.html
@@ -1,18 +1,18 @@
 @(page: PageWithStoryPackage)(implicit request: RequestHeader)
 
-@import views.support.EmailArticleImage
+@import views.support.EmailImage
 @import views.support.EmailHelpers._
 @import model.liveblog._
 @import model.EmailAddons.EmailContentType
 
 @defining(page.article) { article =>
     @fullRow {
-        <a href="@article.metadata.webUrl" @page.item.email.map { email => title="View @email.name online"}>
-            <img width="580" src="@article.banner" @page.item.email.map { email => alt="View @email.name online"}>
+        <a href="@article.metadata.webUrl" @page.email.map { email => title="View @email.name online"}>
+            <img width="580" src="@page.banner" @page.email.map { email => alt="View @email.name online"}>
         </a>
     }
 
-    @article.fallbackSeriesText.map { seriesName =>
+    @page.fallbackSeriesText.map { seriesName =>
         @paddedRow {
             <h3 class="text--brand">@seriesName</h3>
             <hr class="rule--compact" />
@@ -81,7 +81,7 @@
                     }
 
                     case ImageBlockElement(media, data, showCredit) => {
-                        @EmailArticleImage.bestFor(media).map { url =>
+                        @EmailImage.bestFor(media).map { url =>
                             @fullRow {
                                 <img width="580" class="media--main" src="@url" @data.get("alt").map { alt => alt="@alt" }>
                             }
@@ -107,7 +107,7 @@
                     case GuVideoBlockElement(video, media, data) => {
                         @data.get("url").map { url =>
                             @fullRowWithBackground(media) {
-                                <a class="play--button" href="@url"><img class="play--icon" src="@Images.play"></a>
+                                <a class="play--button" href="@url"><img class="float-left play--icon" src="@Images.play"></a>
                             }
                         }
                     }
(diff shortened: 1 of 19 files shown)


common/app/model/EmailAddons.scala
```diff
@@ -2,108 +2,122 @@ package model
 
 import conf.Static
 
-sealed trait EmailContent extends Product with Serializable {
+
+sealed trait EmailMetadata[T] extends Product with Serializable {
   def name: String
-  def banner: String
-  def test(c: ContentType): Boolean
+  def banner: Option[String]
```
desbo: v minor but you could provide a default of `None` here, as with `address`, to avoid having to specify it in the implementation

desbo:
This is awesome, great work! 👍

TBonnin:
pinging @johnduffell who has some insights about `/email` slug caching

desbo:
Is there a reason to keep `/email` now, or can we just update ExactTarget to use `format=email`? @joelochlann @crifmulholland

article/app/views/fragments/emailArticleBody.scala.html
```diff
@@ -1,18 +1,18 @@
...
 @import views.support.EmailHelpers._
 @import model.liveblog._
 @import model.EmailAddons.EmailContentType
 
 @defining(page.article) { article =>
     @fullRow {
-        <a href="@article.metadata.webUrl" @page.item.email.map { email => title="View @email.name online"}>
-            <img width="580" src="@article.banner" @page.item.email.map { email => alt="View @email.name online"}>
+        <a href="@article.metadata.webUrl" @page.email.map { email => title="View @email.name online"}>
+            <img width="580" src="@page.banner" @page.email.map { email => alt="View @email.name online"}>
```
desbo: what if `page.banner` is `None`?

crifmulholland:
This is AMAZING! 

/email is used in all the Exact Target configurations so we need to swap all of them over if that's not happened. We can switch over, and create, the rest of the legacy emails at the same time. Happy to help / do this.

Awesome work.

joelochlann:
@TBonnin am aware of the caching issues around `/email` and the preferred way of doing it now is `?format=email`. But @desbo @crifmulholland removing the `/email` route should be done in its own PR I think - currently the ExactTarget email templates are presumably relying on this so we should make sure we coordinate those changes.

joelochlann:
@crifmulholland snap! (crossed comments there). and thanks, hope this will be useful!

common/app/implicits/Requests.scala
```diff
@@ -24,7 +24,7 @@ trait Requests {
 
     lazy val isAmp: Boolean = r.getQueryString("amp").isDefined || (!r.host.isEmpty && r.host == Configuration.amp.host)
 
-    lazy val isEmail: Boolean = (r.getQueryString("format") == Some("email")) || r.path.endsWith(EMAIL_SUFFIX)
+    lazy val isEmail: Boolean = r.getQueryString("format").contains("email") || r.path.endsWith(EMAIL_SUFFIX)
```
johnduffell: personally I'd rather have check that the url is exactly format=email or empty, and return a 404 if it has any other parameter exists or other value.  That means that we can't ever be unknowningly caching the same content for format=blah, format=fish, format=email&CMP=fb_gu, and standard.

johnduffell: this could (probably should) be done as a filter I suppose, rather than in every controller

joelochlann: Actually @johnduffell only `format=email` is allowed through Fastly anyway: https://github.com/guardian/fastly-edge-cache/pull/673

johnduffell: oh yes!  didn't realise that so thanks for the pointer!  Although that doesn't stop my 3rd example above, although we could just hope that noone does that as it would be a little weird!

PR #14237, opened 2016-09-08
Add warning in docs about Docker for Mac

TBonnin:
## What does this change?

Adding warning to let know potential devs about the issue with mounted volumes in Docker for Mac.
Since most of Guardian devs are on Mac, I've reordered the install method to de-prioritize the Docker one.

Devs are aware of this known issue for Docker for Mac.
## Request for comment

@jfsoul

### DIFF
--- a/docs/01-start-here/01-installation-steps.md
+++ b/docs/01-start-here/01-installation-steps.md
@@ -24,36 +24,7 @@ Follow [this link](https://www.google.co.uk) and enter the relevant search strin
 
 # Local Test Server setup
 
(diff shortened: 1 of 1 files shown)


docs/01-start-here/01-installation-steps.md
```diff
@@ -175,6 +146,38 @@ After this, you can compile the assets:
 make compile
 ```
 
+## Docker
+
+Warning: Docker for Mac is suffering performance issues for directories mounted via osxfs, making the Guardian frontend setup on Mac painfully slow. We'll recommand that Mac users follow the instructions above to install and run frontend on their local machine.
```
TBonnin: This is the warning part. The rest is reordering paragraphs.

jfsoul:
👍  thanks for that.  Would have done it myself but didn't know the specifics (+lazy).

PR #14240, opened 2016-09-08
Fabric Video Expandable v2

regiskuckaertz:
Introduces breaking changes and new features:
- add a visible call-to-action over the video
- add an invisible call-to-action over the image area (when the video is playing)
- move the creative background image to the full-width container
- add the ability to specify where the cross icon needs to appear (either on the edge of the screen or the edge of the inner container)


janua:
LGTM :+1:

PR #14241, opened 2016-09-08
Config update: rotate PA api key

TBonnin:
## What does this change?

New config: PA api key rotation
## Request for comment

@jfsoul


jfsoul:
👍

PR #14245, opened 2016-09-08
Extended test expiry to ensure we get enough impressions.

markjamesbutler:
## What does this change?

Extend switches/tests as we need to run the tests for longer to get the correct number of impressions.

No its a config change.

No


gtrufitt:
Also extending the JS side? Other than that 👍

katebee:
👍

PR #14248, opened 2016-09-09
Small revamp of the ELB admin dashboard

TBonnin:
## What does this change?
1. Adding the number of healthy instances for each ELB (next to the name). _I was kind of tired to go to the AWS console and filter out the CODE elbs to know how many instances were up._
2. Grouping latency and 2xx/mins count for each ELB into a single graph. _Only one graph per ELB: Less graphs => less scrolling. I don't think having 2 lines per graph is worsening the graph readability and it makes easier to spot correlation (or non-correlation) between the two metrics._
3. Some small refactoring.

Dashboard is more useful.
## Screenshots

screen shot 2016-09-09 at 11 25 03
I am open to suggestion in term of color scheme :wink:
## Request for comment

@guardian/dotcom-platform

### DIFF
--- a/admin/app/tools/CloudWatch.scala
+++ b/admin/app/tools/CloudWatch.scala
@@ -6,10 +6,8 @@ import com.amazonaws.services.cloudwatch.model._
 import common.{ExecutionContexts, Logging}
 import conf.Configuration
 import conf.Configuration._
-import controllers.HealthCheck
 import org.joda.time.DateTime
 import services.AwsEndpoints
-
 import scala.collection.JavaConversions._
 import scala.concurrent.Future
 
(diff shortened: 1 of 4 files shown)


admin/app/tools/CloudWatch.scala
```diff
@@ -61,7 +58,6 @@ object CloudWatch extends Logging with ExecutionContexts {
     ("frontend-applications", ChartFormat(Colour.`tone-news-1`)),
     ("frontend-discussion",   ChartFormat(Colour.`tone-news-2`)),
     ("frontend-identity",     ChartFormat(Colour.`tone-news-2`)),
-    ("frontend-image",        ChartFormat(Colour.`tone-news-2`)),
```
TBonnin: frontend-image doesn't exist anymore 🔪

NataliaLKB:
lgtm 👍 

I also get a bit tired filtering through elbs on Amazon everytime I want to see how things are doing. I also like having both metrics on the same graph. Perfectly readable to me. Thanks for this!!

jamespamplin:
:+1: nice one @TBonnin!

PR #14249, opened 2016-09-09
Set top edge to 0 for fixed position banner ads.

Ap0c:
## What does this change?

Sets top edge to 0 on banner ads in immersive galleries.

It fixes a compatibility problem in IE11/Edge, where banner ads were appearing further down the page in immersive galleries and covering other content.
## Screenshots

**Edge Before**

edge-before

**Edge After**

edge-after
## Request for comment

@regiskuckaertz


static/src/stylesheets/module/_adslot.scss
```diff
@@ -165,6 +165,12 @@
     }
 }
 
+// Compatibility fix for top banner ads in IE11/Edge. Without this top banner
+// appears further down the page, and covers content, in immersive galleries.
+.immersive-header-container .top-banner-ad-container {
```
regiskuckaertz: Would it do any harm to add the `top` to the top banner everywhere, not only in immersive galleries? Let's try to keep the layout consistent.

Ap0c: Ok sure, tweaked that to cover all .top-banner-ad-container elements.

regiskuckaertz:
👍

regiskuckaertz:
You will need to merge master to remove expired switches.

paperboyo:
Thank you, Gentlemen!

PR #14250, opened 2016-09-09
Remove superfluous variable names from `ABTestSwitches`

rtyley:
The names don't help anything:
- no individual reference is made server-side to any of these A/B switches
- the `Switch()` constructor [automatically registers itself](https://github.com/guardian/frontend/blob/08649d62/common/app/conf/switches/Switches.scala#L107) in a full list of switches in the JVM, so the switch appears in https://frontend.gutools.co.uk/dev/switchboard

... and the additional Scala `val` identifiers duplicate the name already given within the switch, which just adds noise.

cc @desbo

### DIFF
--- a/common/app/conf/switches/ABTestSwitches.scala
+++ b/common/app/conf/switches/ABTestSwitches.scala
@@ -1,12 +1,13 @@
 package conf.switches
 
 import common.Edition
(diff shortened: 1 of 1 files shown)


desbo:
Good Housekeeping 👍

PR #14252, opened 2016-09-09
Render hosted article

kelvin-chappell:
This is the initial rendering of hosted article content.

/cc @guardian/labs-beta


Calanthe:
👍

PR #14254, opened 2016-09-09
cAPI driven hosted article - inline styles added

Calanthe:
## What does this change?

Similar to the hosted video pages https://github.com/guardian/frontend/pull/14018 we have to include inline styles with brand specific colour. This PR adds those styles into template.

I am aware that the red colour does not look right. The contrast font is calculated correctly, because based on the http://www.colorhexa.com/e31b22, the brightness value in the `HSB` format is equal to 89 therefore font should be dark. We may have to change that.

The work on this page is still in progress. We need to upload correct CTA and top banner images. We also have to think about adding some padding around the brand logo.
## Screenshots

screen shot 2016-09-09 at 17 14 18

screen shot 2016-09-09 at 17 14 29
## Request for comment

@kelvin-chappell

### DIFF
--- a/commercial/app/views/hosted/guardianHostedArticle2.scala.html
+++ b/commercial/app/views/hosted/guardianHostedArticle2.scala.html
@@ -3,8 +3,38 @@
 @import views.html.hosted._
 
 @main(page, Some("commercial")) { } {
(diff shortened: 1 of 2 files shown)


kelvin-chappell:
👍

PR #14218, opened 2016-09-07
Make Weekend Reading banner wait for the Membership banner

katebee:
## What does this change?

Our Weekend Reading Snap banner (#14177) is super speedy and races ahead of the Membership and Adblock banners to take the banner slot. 🏃 🏃 

(The Weekend Reading Banner has not been switched on yet)

Mediator is now used to make `renderWeekendReadingSnap` wait until the Membership banner has decided if it is going to display.
- Weekend Reading banner will not appear if the Adblock banner intends to show
- Weekend Reading banner waits for Membership banner to complete before it tries to show
- Banner queue: Breaking News > Cookies >Adblocker > Membership > WeekendReading
## What is the value of this and can you measure success?
- Give the Adblocker and Membership banners priority over Weekend Reading banner
- We can switch on the Weekend Reading banner for 50% of visitors
## Does this affect other platforms - Amp, Apps, etc?

No
## Screenshots

Weekend reading banner:

picture 216
## Request for comment

Thanks @joelochlann for recommending mediator 👏 

@markjamesbutler: I'm adding a `mediator.emit` to membership-messages.js

PR #14255, opened 2016-09-09
Add US Survey Banner

sammorrisdesign-zz:
## What does this change?

As part of an OKR we have here, we're sending out a survey link by using the banner area. The user will only see it once, if they're in the US.

No
## Screenshots

image


katebee:
This will probably beat the MembershipEngagementBanner and AdblockerBanner to the banner slot (see #14218) (and subsequently, the Habit Forming SnapBanner, but we can wait our turn! :smile:) 

It may be worth checking for adblockers, or letting the AdblockerBanner go first, in case users are in a private-browsing mode, or losing cookies/local storage?

static/src/javascripts/projects/common/modules/onward/us-survey-banner.js
```diff
@@ -0,0 +1,78 @@
...
+
+    function UsUser() {
+        return config.page.edition && config.page.edition === 'US';
+    }
+
+    function init() {
+        if (canShowPromo()) {
+            userPrefs.set('us-survey-banner', {
+                'seen': true
+            });
```
katebee: Message has a `message.remember()` function that will store the ID of the message ('mobile-labs-alert') in alongside the other 'messages' in local storage

static/src/javascripts/projects/common/modules/onward/us-survey-banner.js
```diff
@@ -0,0 +1,78 @@
...
+    detect,
+    template,
+    Message,
+    userPrefs,
+    messageTemplate
+){
+
+    var messageId = 'mobile-labs-alert',
+        messageOptions = {
+            siteMessageLinkName: 'mobile labs | message | presidential primary alerts',
```
katebee: It may be easier to get the interactions out of Ophan if you add a data-component

You could use the option `siteMessageComponentName` to add your preferred component name to the outermost div of the site message

katebee:
Banner looks good, However, if I clear my `prefs.messages` I see the Membership banner yet `gu.prefs.us-survey-banner` immediately gets set, even though the us-survey banner has not been displayed. This mean on subsequent page views I wont ever get the us-survey banner. 😭 

It may work better if you only set a value in local storage if Message.show() has run.

sammorrisdesign:
Thanks @katebee, I've updated the PR based on your feedback.

katebee (approved):
If you are planning to track interaction (with the banner) in Ophan, you may want to set siteMessageComponentName and siteMessageLinkName in the messageOptions.

Looks good :+1:

PR #14256, opened 2016-09-09
Fix the image references so that they work across environments.

JustinPinner:
## What does this change?

Makes image references on the adblock popover environment-compatible. Also opens clicked links in the same tab/window.

If you see images... SUCCESS and MUCH WOW!

No.
## Screenshots

Should go from this 😢 ;
screen shot 2016-09-09 at 19 25 42

to this 😄 ;
screen shot 2016-09-08 at 18 32 01

on CODE and PROD (and DEV)
## Request for comment

@Calanthe (thanks for the tips!) @regiskuckaertz


Calanthe:
👍

PR #14258, opened 2016-09-12
Add main image to hosted article

kelvin-chappell:
/cc @Calanthe

### DIFF
--- a/common/app/common/commercial/hosted/HostedArticlePage2.scala
+++ b/common/app/common/commercial/hosted/HostedArticlePage2.scala
@@ -1,6 +1,7 @@
 package common.commercial.hosted
 
-import com.gu.contentapi.client.model.v1.{Content, TagType}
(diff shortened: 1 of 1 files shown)


Calanthe:
👍

paperboyo:
Excuse the code-blind, but will it use mediaservice's `/master/` asset (as it always should if the images are to come from Grid)? Ignore me if these images aren't going through Grid/resizer route...

kelvin-chappell:
@paperboyo We're going to revisit the way we handle images after we've got the first composer-driven campaign live.  They should be handled in the same way as normal articles ultimately.  They are coming from the grid.

paperboyo:
Thanks, @kelvin-chappell!

> They should be handled in the same way as normal articles ultimately.

article-1326196173340-021e3f97000004b0-358038_568x386

PR #14259, opened 2016-09-12
cAPI driven hosted video page fix

Calanthe:
## What does this change?

The wrong cssClass value should be removed because it causes wrong styling.
## Screenshots

Before:
screen shot 2016-09-12 at 12 42 53

After:
screen shot 2016-09-12 at 12 43 07
## Request for comment

@kelvin-chappell


kelvin-chappell:
👍

PR #14260, opened 2016-09-12
Inject `wsClient` into `OphanApi` rather than calling `WS`

TBonnin:
## What does this change?

Inject `wsClient` into `OphanApi` rather than calling `WS` in preparation of Play 2.5 upgrade where global object `WS` has been deprecated

One caveat though: `OphanApi` object is called from `SurgingContentAgent` which is itself called from Page `meta` in order to populate some Javascript config value.
Until I find an elegant way to inject `OphanApi` into the `SurgingContentAgent`, I have kept the `OphanApi` object. 😢 

=> Play 2.5
## Request for comment

@alexduf

### DIFF
--- a/common/app/model/content.scala
+++ b/common/app/model/content.scala
@@ -13,7 +13,6 @@ import cricketPa.CricketTeams
 import layout.ContentWidths.GalleryMedia
 import model.content.{Atoms, Quiz}
 import model.pressed._
-import ophan.SurgingContentAgent
 import org.joda.time.DateTime
 import org.jsoup.Jsoup
 import org.jsoup.safety.Whitelist
@@ -69,7 +68,6 @@ final case class Content(
   showFooterContainers: Boolean = false
 ) {
 
-  lazy val isSurging: Seq[Int] = SurgingContentAgent.getSurgingLevelsFor(metadata.id)
   lazy val isBlog: Boolean = tags.blogs.nonEmpty
   lazy val isSeries: Boolean = tags.series.nonEmpty
   lazy val isFromTheObserver: Boolean = publication == "The Observer"
--- a/common/app/services/OphanApi.scala
+++ b/common/app/services/OphanApi.scala
@@ -6,7 +6,7 @@ import common.{BadConfigurationException, ExecutionContexts, Logging}
 import conf.Configuration._
 import play.api.Play.current
 import play.api.libs.json._
-import play.api.libs.ws.WS
+import play.api.libs.ws.{WS, WSClient}
 
 import scala.concurrent.Future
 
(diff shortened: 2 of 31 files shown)


common/app/services/OphanApi.scala
```diff
@@ -105,3 +105,6 @@ object OphanApi extends ExecutionContexts with Logging with implicits.WSRequests
   def getMostViewedVideos(hours: Int, count: Int): Future[JsValue] =
     getBody("video/mostviewed")(Map("hours" -> hours.toString, "count" -> count.toString))
 }
+
+object OphanApi extends OphanApi(WS.client) //Do not use. TODO: To delete once we find an elegant way to inject OphanApi into SurgingContentAgent
+
```
TBonnin: As explain in the PR description, I have not found a nice way to deal with the `SugingContentAgent` dependency within `meta`. 😞  Keeping this object for now

common/app/templates/inlineJS/blocking/config.scala.js
```diff
@@ -1,7 +1,6 @@
 @(page: model.Page)(implicit request: RequestHeader)
 @import common.{Edition, StringEncodings}
 @import conf.Static
-@import views.support.{JavaScriptPage, CamelCase}
```
TBonnin: Not use anywhere. 🔪

onward/app/AppLoader.scala
```diff
@@ -17,19 +17,36 @@ import play.api.routing.Router
...
+  lazy val weatherApi = wire[WeatherApi]
+  lazy val geoMostPopularAgent = wire[GeoMostPopularAgent]
+  lazy val dayMostPopularAgent = wire[DayMostPopularAgent]
+  lazy val mostReadAgent = wire[MostReadAgent]
+  lazy val mostPopularSocialAutoRefresh = wire[MostPopularSocialAutoRefresh]
+  lazy val mostViewedAudioAgent = wire[MostViewedAudioAgent]
+  lazy val mostViewedGalleryAgent = wire[MostViewedGalleryAgent]
+  lazy val mostViewedVideoAgent = wire[MostViewedVideoAgent]
+}
+
```
TBonnin: Moved from `OnwardControllers.scala`. See below

common/app/model/content.scala
```diff
@@ -69,7 +68,6 @@ final case class Content(
   showFooterContainers: Boolean = false
 ) {
 
-  lazy val isSurging: Seq[Int] = SurgingContentAgent.getSurgingLevelsFor(metadata.id)
```
TBonnin: Not called anywhere

common/app/services/OphanApi.scala
alexduf: I think it's sensible to use that approach until we have a solution for meta!

alexduf:
:+1:

PR #14261, opened 2016-09-12
Extend the expiry and audience of the WR snap

katebee:
- Extending the Weekend Reading snap banner (#14177), and the Weekend Reading email AB test (#14171) for a couple more weeks.
- Expanding the snap banner test to include 100% of the audience.

The snap banner still goes last, since it is waiting for the emitters to fire from other banners (#14218).


NataliaLKB:
👍

PR #14263, opened 2016-09-12
Fix match error in CSP reports

desbo:
## What does this change?

Fixes a non-exhaustive match

Less noise in Kibana

Nope
## Screenshots

Nope
## Request for comment

@TBonnin


TBonnin:
👍  Thank you very much @samdesborough

PR #14265, opened 2016-09-12
Fix missing member implementations in dev-build

TBonnin:
## What does this change?

Fix missing member implementations in dev-build

dev-build project compiles
## Request for comment

@JustinPinner @alexduf


JustinPinner:
Fair enough 👍

PR #14269, opened 2016-09-13
Inject http client into ContentApiClient

TBonnin:
## What does this change?

First PR to remove WS reference within ContentApiClient. 👊 
Create a contentApiClass but keep a ContentApiClient object around for legacy purpose. I started to move to using the class so I can inject the http client but I am far from being done. More PRs to come

=> Play 2.5
## Request for comment

@alexduf

### DIFF
--- a/common/app/contentapi/ContentApiClient.scala
+++ b/common/app/contentapi/ContentApiClient.scala
@@ -10,11 +10,10 @@ import common._
 import conf.Configuration
 import conf.Configuration.contentApi
 import conf.switches.Switches.{CircuitBreakerSwitch, ContentApiUseThrift}
-import metrics.{CountMetric, TimingMetric}
 import model.{Content, Trail}
 import org.joda.time.DateTime
 import org.scala_tools.time.Implicits._
-
+import play.api.libs.ws.WS
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.duration.{Duration, MILLISECONDS}
 import scala.concurrent.{ExecutionContext, Future}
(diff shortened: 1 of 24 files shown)


dev-build/app/AppLoader.scala
```diff
@@ -64,9 +64,8 @@ trait AppComponents
   with DiscussionServices
   with FapiServices
   with AdminJobsServices
-  with OnwardServices {
-
-  override lazy val ophanApi = wire[OphanApi] // ophanApi is already defined in AdminServices and OnwardServices. Overriding it here to avoid conflict
```
alexduf: How come ophanApi isn't needed anymore?

common/app/app/CommonServices.scala
```diff
@@ -0,0 +1,12 @@
+package app
+
+import com.softwaremill.macwire._
+import contentapi.{CapiHttpClient, ContentApiClient}
+import play.api.libs.ws.WSClient
+
+trait CommonServices {
```
alexduf: I'm not sure where I stand about that trait.
On one side `CommonServices` is generic enough to be dangerous as we may shove everything in here, on the other I'm not sure `ContentApiComponents` would be much better as it is very specific.
Do you think if you had a `ContentApiComponents` you would add more to it?

common/app/contentapi/ContentApiClient.scala
```diff
@@ -10,11 +10,10 @@ import common._
 import conf.Configuration
 import conf.Configuration.contentApi
 import conf.switches.Switches.{CircuitBreakerSwitch, ContentApiUseThrift}
-import metrics.{CountMetric, TimingMetric}
 import model.{Content, Trail}
 import org.joda.time.DateTime
 import org.scala_tools.time.Implicits._
-
+import play.api.libs.ws.{WS}
```
alexduf: [pedentic] `import play.api.libs.ws.WS` [/pedentic]

TBonnin:
@alexduf I implemented the changes we talked about. Feel free to review 😸  Thanks

alexduf:
Looks good :+1:

PR #14270, opened 2016-09-13
AMP - Increase Maximum No. of Ads to 8

Ap0c:
## What does this change?

Ups the maximum number of ads on AMP pages from 2 to 8. **Note**: 8 is the maximum, but the number of ads shown is proportional to the length of the article, so shorter pieces won't necessarily show as many as this.

Allows more ads to display on AMP versions of articles, bringing the tally closer to the number on the standard versions.

AMP.
## Request for comment

@kenlim

### DIFF
--- a/common/test/views/support/cleaner/AmpAdCleanerTest.scala
+++ b/common/test/views/support/cleaner/AmpAdCleanerTest.scala
@@ -39,8 +39,17 @@ class AmpAdCleanerTest extends FlatSpec with Matchers {
 
   }
 
-  "AmpAdCleaner" should "only add 8 ads in total" in {
+  "AmpAdCleaner" should "only add 2 ads for short article" in {
     val doc = s"""<html><body>${s"<p>${tenChars * 70}</p>" * 4}</body></html>"""
+    val document: Document = Jsoup.parse(doc)
+    val result: Document = clean(document)
+
+    result.getElementsByTag("amp-ad").size should be(2)
+
+  }
+
+  "AmpAdCleaner" should "only add 8 ads in total" in {
+    val doc = s"""<html><body>${s"<p>${tenChars * 70}</p>" * 25}</body></html>"""
     val document: Document = Jsoup.parse(doc)
     val result: Document = clean(document)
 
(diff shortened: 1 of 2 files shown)


common/test/views/support/cleaner/AmpAdCleanerTest.scala
```diff
@@ -48,6 +48,15 @@ class AmpAdCleanerTest extends FlatSpec with Matchers {
 
   }
 
+  "AmpAdCleaner" should "only add 8 ads in total" in {
```
johnduffell: no need for both tests, the intention was to test that it didn't add the (n+1)th  advert, so just update it to make sure there's space for 9, but it should only add 8.

common/test/views/support/cleaner/AmpAdCleanerTest.scala
```diff
@@ -39,7 +39,7 @@ class AmpAdCleanerTest extends FlatSpec with Matchers {
 
   }
 
-  "AmpAdCleaner" should "only add 2 ads in total" in {
+  "AmpAdCleaner" should "only add 2 ads for short article" in {
```
johnduffell: assuming this test still passes, something is broken, as the intention was to make sure it didn't add 3 adverts where there was space to.
So you might want to look into that too.

johnduffell:
see comments about tests, also make sure you test locally to make sure it looks sensible on longer articles.  I think liveblogs use different code, but satisfy yourself of that too.

common/test/views/support/cleaner/AmpAdCleanerTest.scala
```diff
@@ -39,12 +39,19 @@ class AmpAdCleanerTest extends FlatSpec with Matchers {
 
   }
 
-  "AmpAdCleaner" should "only add 2 ads in total" in {
-    val doc = s"""<html><body>${s"<p>${tenChars * 70}</p>" * 4}</body></html>"""
+  "AmpAdCleaner" should "include ads at defined, regular intervals" in {
```
johnduffell: might go for something like `"not have been changed without someone checking the tests are still valid"`

johnduffell:
thanks, looks good! 🚀 👍

PR #14271, opened 2016-09-13
Fix ad targeting bug with paid-content tags

kelvin-chappell:
Some tags that were previously keywords are now paid-content tags of subtype topic, and this meant some _k_ parameters were being missed out in ad calls.

Eg. 

```
id: "cook-with-leisure/cook-with-leisure",
type: "paid-content",
webTitle: "Cook with Leisure",
webUrl: "https://www.theguardian.com/cook-with-leisure/cook-with-leisure",
apiUrl: "https://content.guardianapis.com/cook-with-leisure/cook-with-leisure",
paidContentType: "Topic"
```


kenlim:
👍

PR #14273, opened 2016-09-13
Update Email Prefs page with Brexit Briefing and Weekend Reading

katebee:
## What does this change?
- Brexit briefing was missing from the email prefs page, it will now listed
- Guardian Opinion US was missing from the email prefs page, it will now listed
- Change the copy for the Weekend Reading email listing
- Ensures that users can fully unsubscribe from the Weekend Reading email, regardless of which variant they ended up in (AB test added in #14171)

If the ABWeekendReadingEmail is running, we need to ensure we unsubscribe the user from both the **control** (ListID = 3744) and the **variant** (ListID = 3743), otherwise they may still receive the email.

An additional query string will be sent when unsubscribing from Weekend Reading:

picture 232

Nope
## Screenshots

picture 231
## Request for comment

@joelochlann 😄


identity/app/model/EmailSubscriptions.scala
```diff
@@ -54,10 +54,10 @@ object EmailSubscriptions {
       "Weekend reading",
       "news",
       "News",
-      "Our editors' picks for the day's top news and commentary delivered to your inbox each weekday.",
+      "The best stuff you didn't have time to read during the week - from features and news analysis to lifestyle and culture",
       "Every Saturday",
       "3744",
-      subscribedTo = subscribedListIds.exists{ x => x == "3744" } || subscribedListIds.exists{ x => x == "3744" }
+      subscribedTo = subscribedListIds.exists{ x => x == "3743" } || subscribedListIds.exists{ x => x == "3744" }
```
joelochlann: I think this could be written `subscribedListIds.exists{ x => x == "3743" || x == "3744" }`

joelochlann:
👍  looks good! if we find ourselves doing a lot of this "double unsubscribe" because of A/B tests we could look at a better abstraction for it which would make it clear what's going on without the need for explanatory comments

PR #14275, opened 2016-09-13
Hosted by / Advertiser Content label

blongden73:
Before
picture 53
picture 52
After
picture 50
picture 49

@Calanthe @lps88

### DIFF
--- a/static/src/stylesheets/module/commercial/glabs/_hosted-article.scss
+++ b/static/src/stylesheets/module/commercial/glabs/_hosted-article.scss
@@ -189,6 +189,18 @@ $zootropolisColor: #2ec869; //need to remove references to brand colours here ev
             text-align: center;
         }
     }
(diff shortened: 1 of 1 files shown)


Calanthe:
This is fine but now I realised that in the last PR you added styles to every paragraph except `hosted__next-page-title` https://github.com/guardian/frontend/blob/2085e9574d76b906cbf6584f377875a095c22911/static/src/stylesheets/module/commercial/glabs/_hosted-article.scss#L177. Maybe it would be better to fix the previous PR by adding necessary changes to the paragraphs inside article only? So you won't have to add those fixes (like in this PR) to every other <p> element.

blongden73:
@Calanthe Cool, i'll add this so that its specifically for the <p> within the article. I'd still like to style these specifically though, as it makes the size of the two labels the same rather than just falling back to the default styling of the <p>.

Calanthe:
ok 👍

PR #14276, opened 2016-09-13
Font contrast added to the gallery page

Calanthe:
## What does this change?

The `hosted-page--bright` is added to the `hosted-gallery-page` `<div/>` so the correct font contrast is applied.
## Request for comment

@lps88


lps88:
👍

PR #14277, opened 2016-09-13
Fix image loading for instagram embeds

rjmk:
## What does this change?

Exempts children of `.instagram-media` elements from having text matching Guardian tags from being turned into links.  

Instagram seems to rely on the HTML it provides to be unchanged for loading the image in.

Am being taught tomorrow how to set things up so I can test that none of the other HTML cleaning can break Instagram embeds.

No, though it relies on (a) CAPI continuing to use 'instagram-media' as the class on embeds and other elements not having this class
## Screenshots

Before:

image

After (:tada:):

image
## Request for comment

@guardian/dotcom-platform


common/app/views/support/HtmlCleaner.scala
```diff
@@ -337,14 +337,18 @@ case class TagLinker(article: Article)(implicit val edition: Edition, implicit v
     s"""(.*)( |^)($tagName)( |,|$$|$dot|$question)(.*)""".r
   }
 
+  private def or[A](pred: A => Boolean, pred2: A => Boolean)(a: A) = pred(a) || pred2(a)
```
rjmk: Does this utility already exist? I had a quick look but couldn't find it

TBonnin:
It is written absolutely nowhere so you couldn't guess but most of us tend to prefix our branch name with our name or initial (ex: `tbonnin-fix-broken-thing`). It is easier to manage, search and avoid conflicting names. 🎯

common/app/views/support/HtmlCleaner.scala
```diff
@@ -337,14 +337,18 @@ case class TagLinker(article: Article)(implicit val edition: Edition, implicit v
...
 
-      // Get all paragraphs which are not contained in a pullquote
-      val paragraphs = doc.getElementsByTag("p").filterNot( p =>
-        p.parents.exists( ancestor =>
-          ancestor.tagName() == "aside" && ancestor.hasClass("element-pullquote")
+      // Get all paragraphs which are not contained in a pullquote or in an instagram caption
+      val paragraphs = doc.getElementsByTag("p").filterNot( p => p.parents.exists(
+          or((ancestor: Element) =>
+            ancestor.tagName() == "aside" && ancestor.hasClass("element-pullquote"),
+            _.hasClass("instagram-media")
```
TBonnin: I am not sure what does this bring vs a simple:
```
... 
   ancestor.tagName() == "aside" && ancestor.hasClass("element-pullquote") || _.hasClass("instagram-media")
...
```

Am I missing something?

rjmk: I wrote it that way as I thought it's nicer to separate the 2 checks rather than expanding one check to deal with both (a) the paragraph is not in a pullquote and (b) the paragraph is not in an instagram block. I suppose it's not such an issue with only the 2 checks

TBonnin: Makes sens to be explicit. What about?
```
val isPullquote = ancestor.tagName() == "aside" && ancestor.hasClass("element-pullquote")
val isInstagramBlock = ancestor.hasClass("instagram-media")
isPullquote || isInstagramBlock
````

Comment (rjmk) [inline common/app/views/support/HtmlCleaner.scala]: Sounds good

File: common/app/views/support/HtmlCleaner.scala [hunk 3]
```diff
@@ -337,19 +337,16 @@ case class TagLinker(article: Article)(implicit val edition: Edition, implicit v
...
-          or((ancestor: Element) =>
-            ancestor.tagName() == "aside" && ancestor.hasClass("element-pullquote"),
-            _.hasClass("instagram-media")
-          )
-        )
+      val paragraphs = doc.getElementsByTag("p").filterNot( p => p.parents.exists( ancestor => {
+          val inPullquote = ancestor.tagName() == "aside" && ancestor.hasClass("element-pullquote")
+          val inInstagramBlock = ancestor.hasClass("instagram-media")
+          inPullquote || inInstagramBlock
+        })
```
TBonnin: Regarding the curly braces, you would usually make the function passed to `exists` a block by itself (rather than a single expression (ie: **{** thing => expression **}** vs thing => **{** expression **}**
You would get something like this:
````
p.parents.exists( { ancestor => 
   val inPullquote = ancestor.tagName() == "aside" && ancestor.hasClass("element-pullquote")
   val inInstagramBlock = ancestor.hasClass("instagram-media")
   inPullquote || inInstagramBlock
} )
```

Then you would be able abbreviate and to omit the unnecessary parenthesis:
```
p.parents.exists { ancestor => 
   val inPullquote = ancestor.tagName() == "aside" && ancestor.hasClass("element-pullquote")
   val inInstagramBlock = ancestor.hasClass("instagram-media")
   inPullquote || inInstagramBlock
}
```

rjmk: That's cool! I'll get it done

rjmk:
:zap:

TBonnin:
Thank you very much. lgtm 👍

common/app/views/support/HtmlCleaner.scala
```diff
@@ -341,11 +341,12 @@ case class TagLinker(article: Article)(implicit val edition: Edition, implicit v
 
     if (article.content.showInRelated) {
 
-      // Get all paragraphs which are not contained in a pullquote
-      val paragraphs = doc.getElementsByTag("p").filterNot( p =>
-        p.parents.exists( ancestor =>
-          ancestor.tagName() == "aside" && ancestor.hasClass("element-pullquote")
-        )
+      // Get all paragraphs which are not contained in a pullquote or in an instagram caption
+      val paragraphs = doc.getElementsByTag("p").filterNot( p => p.parents.exists { ancestor =>
```
TBonnin: I am nitpicking a lot here: I would keep `p.parents...` on a new line. Feel free to ignore me though and push as it is

johnduffell: my "rule" is work out how you would describe what you want the code to do to someone.  Then make sure that you structure the code the same way, e.g. one line per each piece you describe, and functions for the bits you describe separately.  And try to minimise the plumbing parts.

rjmk:
@TBonnin Is this the sort of thing you meant?

TBonnin:
Yep. Perfect. Thank you for dealing with all my nitpicking

rjmk:
Not at all! Thanks for tips :)

rjmk:
Do either of you know why we need [this line](https://github.com/guardian/frontend/pull/14277/files#diff-a78830a0e574fa7224f8e492f8ac2c9aR363). It seems to me that using `map` and `forEach` deals with this for us?

TBonnin:
You are right, functionally it doesn't bring anything. The only advantage is not executing `keywordRegex`

PR #14278, opened 2016-09-13
Hosted pages - adding 'next page' links for chester zoo

lps88:
## What does this change?

Adding links between different pages of the campaign and the cta banner, neither are yet provided from capi so hardcoded in order to get the campaign out.
## Screenshots

image
image
## Request for comment

@Calanthe


Calanthe:
👍

PR #14280, opened 2016-09-13
Fix styling of paid gallery

kelvin-chappell:
All galleries should appear as immersive galleries apart from paid galleries.


Calanthe:
👍

PR #14285, opened 2016-09-13
Chester Zoo - add all the next page info

lps88:
## What does this change?

The 'More from Chester Zoo' links should go to appropriate other pages, with the right images (mostly)

image

@guardian/labs-beta


kelvin-chappell:
Good pragmatic solution that avoids having to hack around with capi for this campaign.  👍

PR #14288, opened 2016-09-14
Hosted video tone fix

Calanthe:
## What does this change?

Because of the wrong tone `guardian.config.page.tones` on the hosted video page, js wasn't initialised at all. Now everything works (including 'About' link and video) but this is temporary solution.
## Request for comment

@kelvin-chappell


kelvin-chappell:
👍

PR #14290, opened 2016-09-14
Hosted content - js error removed

Calanthe:
## What does this change?

Currently (and temporary) we don't have the next video autoplay cards in the end of the video. But the module raises a js error if there is no `.js-next-video` element. This PR fixes that.

## Request for comment

@lps88


lps88:
👍

PR #14291, opened 2016-09-14
Restrict all Contributions CTAs to not show to paying members

rtyley:
We weren't checking on the Contributions Epic/Embed CTAs if the users were paying members, which was obviously irritating for those users.

This is a small refactor to introduce `canReasonablyAskForMoney` and `canAskForAContribution` - which we should be able to apply to all our tests.

cc @markjamesbutler @jranks123

### DIFF
--- a/static/test/javascripts/spec/common/commercial/commercial-features.spec.js
+++ b/static/test/javascripts/spec/common/commercial/commercial-features.spec.js
@@ -315,7 +315,7 @@ define(['helpers/injector', 'Promise'], function (Injector, Promise) {
         describe('Membership messages', function () {
             it('Displays messages by default', function (done) {
                 features = new CommercialFeatures;
-                features.async.membershipMessages.then(function (flag) {
+                features.async.canDisplayMembershipEngagementBanner.then(function (flag) {
                     expect(flag).toBe(true);
                     done();
                 });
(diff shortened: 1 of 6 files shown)


static/test/javascripts/spec/common/commercial/commercial-features.spec.js
```diff
@@ -325,28 +325,16 @@ define(['helpers/injector', 'Promise'], function (Injector, Promise) {
...
-                detect.getBreakpoint = function () { return 'mobile'; };
-                features = new CommercialFeatures;
-                features.async.membershipMessages.then(function (flag) {
-                    expect(flag).toBe(true);
-                    done();
-                });
-            });
-
-            it('Does not display messages to existing members', function (done) {
+            it('Does not display messages to existing members because they are already giving us money', function () {
```
jranks123: I wonder if this should explicitly express that they are a paying member rather than "already giving us money", which could be confused for having already given a contribution, especially if we start limiting contribution ads for people who have already contributed in the future

jranks123:
👍  with comment. Also as a side note, I like the naming of 'canReasonablyAskForMoney', what a considerate function!

rtyley:
Looks good on PROD

Not logged in, having viewed 10 articles:
image

Logged in as a member, message disappears:
image

PR #14292, opened 2016-09-14
Inject contentApiClient PR #2 (mostly in controllers)

TBonnin:
## What does this change?

Inject contentApiClient in all necessary controllers (and one agent)
I am trying to keep the size of those PR manageable because it is quite repetitive at this point.

➡️ Play 2.5

## Request for comment

@alexduf

### DIFF
--- a/common/app/common/dfp/TagSponsorship.scala
+++ b/common/app/common/dfp/TagSponsorship.scala
@@ -141,7 +141,37 @@ case class PaidForTag(targetedName: String,
                       lineItems: Seq[GuLineItem]) {
 }
 
-object PaidForTag {
+object PaidForTagFormat {
+
+  implicit val pftWrites: Writes[PaidForTag] = new Writes[PaidForTag] {
+    override def writes(tag: PaidForTag): JsValue = {
+      Json.obj(
+        "targetedName" -> tag.targetedName,
+        "tagType" -> tag.tagType.name,
+        "paidForType" -> tag.paidForType.name,
+        "matchingCapiTagIds" -> tag.matchingCapiTagIds,
+        "lineItems" -> tag.lineItems
+      )
+    }
+  }
+
+  implicit val pftReads: Reads[PaidForTag] = (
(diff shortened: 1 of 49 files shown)


common/app/common/dfp/TagSponsorship.scala
```diff
@@ -141,30 +141,9 @@ case class PaidForTag(targetedName: String,
                       lineItems: Seq[GuLineItem]) {
 }
 
-object PaidForTag {
+object PaidForTagFormat {
```
alexduf: I think this one should remain a companion object for `PaidForTag`

standalone/app/controllers/ItemController.scala
```diff
@@ -1,9 +1,14 @@
...
+
 // If you add to this, don't forget the one in dev-build
-class ItemController(articleController: ArticleController, faciaDraftController: FaciaDraftController) extends ItemResponseController(
+class ItemController(articleController: ArticleController,
+                     faciaDraftController: FaciaDraftController,
+                     contentApiClient: ContentApiClient
+                    ) extends ItemResponseController(
   articleController,
-  GalleryController,
+  new GalleryController(contentApiClient),
```
alexduf: you could directly receive the controller as a parameter, as it is done for `articleController`, not that this has any sorts of impact as the controller shouldn't have any state

standalone/app/controllers/ItemController.scala
```diff
@@ -10,7 +10,7 @@ class ItemController(articleController: ArticleController,
   articleController,
   new GalleryController(contentApiClient),
   MediaController,
-  InteractiveController,
+  new InteractiveController(contentApiClient),
```
alexduf: same here, could be passed, but not absolutely necessary

alexduf:
All my comments are minor, that looks good :star: :+1: :star:

TBonnin:
Thank you @alexduf. I have addressed you comments.

PR #14294, opened 2016-09-14
AdBlocker popover inconsistency fix

JustinPinner:
## What does this change?

Asynchronous adblocker detection doesn't work reliably in a new tab.

We should be able to see more consistent behaviour of detection and messaging.

No (desktop only)
## Screenshots

screen shot 2016-09-14 at 12 26 17
## Request for comment

@regiskuckaertz


rich-nguyen:
presumably the sync method is slower? Looks fine code wise 👍

JustinPinner:
Thanks @rich-nguyen. It is a bit slower, but (so far, locally at least) works more reliably.

PR #14300, opened 2016-09-14
More injection on ContentApiClient #3

TBonnin:
## What does this change?

More injection of ContentApiClient.
I hope you are not bored yet because I have only done the easy part so far. 😜 

➡️ Play 2.5
## Request for comment

@alexduf

### DIFF
--- a/admin/app/AppLoader.scala
+++ b/admin/app/AppLoader.scala
@@ -30,7 +30,7 @@ class AppLoader extends FrontendApplicationLoader {
 trait AdminServices {
   def wsClient: WSClient
   def akkaAsync: AkkaAsync
-  lazy val capiHttpClient = wire[CapiHttpClient]
+  private lazy val capiHttpClient = wire[CapiHttpClient]
   lazy val contentApiClient = wire[ContentApiClient]
   lazy val ophanApi = wire[OphanApi]
   lazy val emailService = wire[EmailService]
--- a/admin/app/AppLoader.scala
+++ b/admin/app/AppLoader.scala
@@ -6,12 +6,12 @@ import dfp.{CapiLookupAgent, DfpAgentLifecycle}
 import conf.switches.SwitchboardLifecycle
 import conf.{AdminFilters, CachedHealthCheckLifeCycle, CommonGzipFilter}
 import controllers.{AdminControllers, HealthCheck}
-import _root_.dfp.{DfpDataCacheLifecycle, DfpDataCacheJob}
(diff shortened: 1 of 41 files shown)


admin/app/AppLoader.scala
```diff
@@ -30,7 +30,7 @@ class AppLoader extends FrontendApplicationLoader {
 trait AdminServices {
   def wsClient: WSClient
   def akkaAsync: AkkaAsync
-  lazy val capiHttpClient = wire[CapiHttpClient]
+  private lazy val capiHttpClient = wire[CapiHttpClient]
```
TBonnin: Making it private because it is only necessary for initializing the ContentApiClient (one line below) and therefore avoiding conflicting member with different type in preview when using a specific `trainingCapiHttpClient` 

@alexduf let me know if you can think of a more elegant way to solve this

admin/app/indexes/ContentApiTagsEnumerator.scala
```diff
@@ -39,7 +37,7 @@ object ContentApiTagsEnumerator extends Logging {
   }
 
   def enumerateTagType(tagType: String) = enumeratePages { page =>
-    getResponse(ContentApiClient.tags.tagType(tagType).pageSize(MaxPageSize).page(page))
+    contentApiClient.getResponse(contentApiClient.tags.tagType(tagType).pageSize(MaxPageSize).page(page))
```
TBonnin: @alexduf you've probably already noticed this pattern all over the place. I found it weird but I don't want to touch the internal of ContentApiClient at this stage. I will be happy to revisit later though

alexduf: I don't see the issue here. It's pretty much the same as before, it's just that removing the import makes it more visible.
Did I miss something?

TBonnin: I mean the `contentApiClient.getResponse(contentApiClient.someQuery)` rather than something like `contentApiClient.someQuery.getResponse`

alexduf: Oh I see, yes that would make sense to change it later, or at least to split it in two lines :
```scala
val query = ContentApiQuery()
  .tags.tagType(tagType)
  .pageSize(MaxPageSize)
  .page(page)
contentApiClient.execute(query)
```

but as you said that's unrelated

alexduf:
Looks good! :+1: It's good to see so many `object`s gone!

PR #14302, opened 2016-09-14
Add tags to the commercial report

rich-nguyen:
Adds some useful tags to help us distinguish commercial reports that are using different methods of ad delivery.

If we do this, we can easily bucket our client logs into categories, which makes makes for easy comparison of performance between "branches" of our system. For example, how is prebid performing compared to waterfall? 

I've identified a list of things which we can tag up later, to enable us to diagnose commercial sub-features:

**Interesting page-level commercial tags**
- Which programmatic services are being used?
  - Is header-bidding used (exposed in this pr)
- Lazy loaded?
- Has page skin?

**Interesting advert-level tags**
- Inside facia container?
- Fluid?
- Template-rendered? (ie, is it rendered with our js templating system)
- Was advert disabled/empty/hidden?
- Was the slot provided by the server or through JS?


janua:
:+1:

PR #14305, opened 2016-09-14
Surging level 5

Ap0c:
## What does this change?

Adds a new surging level (5) to articles, which will target 50-100 pageviews per minute. The page config object (`guardian.config.page.isSurging`) will now contain this new value as appropriate. Articles with less than 50 pv/min will still report as `0`.

Marking content getting more than 50 pv/min as surging will provide more inventory for ads.


TBonnin (approved):
👍

PR #14306, opened 2016-09-14
Make the Weekend Reading Signup sample link to the latest email

katebee:
## What does this change?

No longer link to the [example](https://www.theguardian.com/info/2016/jul/20/weekend-round-up-no10s-revolving-door/email) Weekend Reading email. Now that the email is being curated each week we can use the `/latest/email` convention to show a more [recent example](https://www.theguardian.com/membership/series/weekend-reading/latest/email)

Nope


SiAdcock (approved):
Very nice 👍

PR #14307, opened 2016-09-14
Enable users to run the app without `etc/gu/install_vars`

rjmk:
## What does this change?

Allows devs not to have a `etc/gu/install_vars` file (currently even if empty, it must exist)

The solution was reasonably constrained as ~23 files rely on `environment.stage` being in place, and I thought it better not to tweak each of them
## Request for comment

@guardian/dotcom-platform

### DIFF
--- a/common/app/common/configuration.scala
+++ b/common/app/common/configuration.scala
@@ -20,51 +20,59 @@ import scala.util.{Failure, Success, Try}
 
 class BadConfigurationException(msg: String) extends RuntimeException(msg)
 
-case class InstallVars(stack: String, app: String, guStage: String, awsRegion: String, configBucket: String)
+object InstallVars {
 
-object GuardianConfiguration extends Logging {
+  val installVars = new File("/etc/gu/install_vars") match {
+      case f if f.exists => IOUtils.toString(new FileInputStream(f))
+      case _ => ""
+    }
+
+  val properties = Properties(installVars)
+
+  def apply(key: String, default: String) = properties.getOrElse(key, default)
+
+  trait InstallationVars {
+    val stack = apply("stack", "frontend")
+    // if got config at app startup, we wouldn't need to configure it
+    val app = apply("app", "dev-build")
+    val stage = apply("STAGE", "DEV")
+    val awsRegion = apply("region", "eu-west-1")
+    val configBucket = apply("configBucket", "aws-frontend-store")
+    if (stage == "DEV" && new File(s"${System.getProperty("user.home")}/.gu/frontend.properties").exists) {
+    throw new RuntimeException(
+      "\n\nYou have a file ~/.gu/frontend.properties with secrets - please delete that file and any copies as it is not needed.\n  " +
+      "All secrets are now stored in S3 bucket aws-frontend-store, not on your laptop.\n\n  " +
+      "Should you need to override any properties in DEV, create a new file ~/.gu/frontend.conf. \n" +
+      "For an example see https://github.com/guardian/frontend/blob/master/common/app/common/configuration.scala#L48\n" +
+      "For details of the changes see https://github.com/guardian/frontend/pull/14081")
+    /*
+    ~/.gu/frontend.conf example file:
+
+    # local development (DEV stage) config overrides (not secrets)
+    devOverrides {
+      switches.key=DEV/config/switches-yournamehere.properties
+      facia.stage=CODE
+    }
+
+     */
+    }
+  }
+}
+
+object GuardianConfiguration extends Logging with InstallVars.InstallationVars {
 
   import com.gu.cm.{Configuration => CM}
   import com.typesafe.config.Config
+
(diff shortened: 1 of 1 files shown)


common/app/common/configuration.scala
```diff
@@ -115,7 +123,7 @@ devOverrides {
 
 class GuardianConfiguration extends Logging {
   import GuardianConfiguration._
-  implicit private lazy val app = Play.current
+  import play.api.Play.current
```
rjmk: The compiler decided it preferred this to what was in place

TBonnin:
cc'ing @johnduffell

TBonnin (commented):
When is InstallVars content loaded now? Am I missing something?

rjmk:
`apply` on line 32 is an alias for `properties.getOrElse(_, _)`, so all the variables that are declared will get the value from the file if present. Is that what you mean?

If you mean the `InstallVars` trait, it's mixed in to a couple of objects

johnduffell:
Thanks for doing the updates, and good first PR.
Personally I'd rather refactor 23 places than make configuration.scala more complicated, it's already not as cohesive as I'd like (you get all configuration from the config object regardless of whether the app needs it) so in some ways ideally we should be splitting installvars into a separate file with as few cross dependencies as possible.
I also prefer to avoid traits/inheritance if possible unless I'm making an algebraic data type (case classes/objects) as it can make things confusing
Don't be afraid to refactor in lots of places if the code needs it, as we need to constantly improve things just to keep standing still.  Since everything sits in the one repo, you can be confident not to break other people.

common/app/common/configuration.scala
```diff
@@ -136,17 +144,7 @@ class GuardianConfiguration extends Logging {
...
-      case f if f.exists => IOUtils.toString(new FileInputStream(f))
-      case _ => ""
-    }
-
-    private val properties = Properties(installVars)
-
-    def apply(key: String, default: String) = properties.getOrElse(key, default).toLowerCase
-
-    val stage = apply("STAGE", "unknown")
+  object environment extends InstallVars.InstallationVars {
```
johnduffell: so basically to be more concrete instead of the general comments I made, I'd suggest basically keep everything as you've done it is great, and just take this `extends InstallVars.InstallationVars` away and either make a lazy val stage to point to the InstallVars object, or make changes to all the 23 places if you prefer.  Then you've avoided having to extend other traits.

rjmk: Thanks for the feedback, John! One thing that's not quite clear to me is whether `stage` really is the only install variable that also belongs in the environment. If that's the case I'm happy to expose it as a val. Otherwise, I think it's a bit ad hoc

johnduffell: well, so far it's true that we only need one, so we don't need to worry about where the others belong until someone needs them.  We can revisit at that point.  It's a balance between bringing them in explicitly, or dragging them all in with a trait.  Or, to avoid the problem entirely, we can alter the 23 places to use installvars directly and remove from environment.

common/app/common/configuration.scala
```diff
@@ -20,64 +20,73 @@ import scala.util.{Failure, Success, Try}
 
 class BadConfigurationException(msg: String) extends RuntimeException(msg)
 
-case class InstallVars(stack: String, app: String, guStage: String, awsRegion: String, configBucket: String)
+object InstallVars {
```
rjmk: I decided against moving this to another file as I didn't want to add any more complication to the file structure. Lemme know if you think that's a bad idea

rjmk:
A bit better, @johnduffell?

common/app/common/configuration.scala
johnduffell: I'm happy!

johnduffell:
go for it and thanks for fixing all that up! 😛 🚀 🚤 👍

rich-nguyen:
What happens in prod, if the install_vars is missing? will the server run in dev mode? I'd rather the server doesn't come up at all.

rjmk:
Yeah, that's right. The title of this PR is a bit misleading, as it is fixing a bug in the code to allow no `install_vars`, rather than introducing the feature.

Looking at the `blame`, it seems not throwing without the file has been the behaviour in the past; the error is a relatively new behaviour. I don't know if that makes a difference, but thought it was worth mentioning

johnduffell:
@rich-nguyen yeah agreed and in fact @TBonnin and I did debate that a bit and we thought about making the health check always return an error in DEV mode.

TBonnin:
Not sure if disabling the healthcheck in dev mode is the best option here but I agree with @rich-nguyen that failing in prod when install_vars is missing would be the best outcome.

johnduffell:
I think there's no perfect way as we have conflicting desires, maybe in dev we can always use frontend profile credentials rather than iam roles to get the file from the bucket, then it won't be able to load it when it's started in AWS?

PR #14308, opened 2016-09-14
Amp local validation

gtrufitt:
## What does this change?

I added that developers should run  `make validate-amp` to the PR guidelines but then realised that just runs it on the live site, so added a local validation check to the tool to allow us to run it on our localhost urls.

@guardian/dotcom-platform


tools/amp-validation/index.js
```diff
@@ -26,10 +24,12 @@ function checkEndpoints(devChannel) {
 }
 
 function runValidator(validator, devChannel) {
+    const isDev = process.argv[2] || false;
```
TBonnin: not crucial but when I see `node index.js true` I have no idea what the second argument `true` means.
Can we make it more explicit? iex: `node index.js --dev`

tools/amp-validation/index.js
```diff
@@ -26,10 +24,12 @@ function checkEndpoints(devChannel) {
...
+    
     return endpoint => fetchPage
-        .get(endpoint)
+        .get(endpoint, isDev)
         .then(res => {
-            console.log(`Checking the AMP validity (${devChannel ? 'pre-release' : 'release'}) of the page at ${endpoint}, result is:`);
+            console.log(`Checking the AMP validity (${devChannel ? 'pre-release' : 'release'}) of the page at ${isDev ? 'localhost:9000' : ''}${endpoint}, result is:`);
```
TBonnin: Is it enough to run the `article` app locally or does the amp test require another app? 
Also how do we explain testers what needs to run locally?

tools/amp-validation/index.js
desbo: the convention is to use `NODE_ENV=dev` and `process.env`, I'd prefix the command with the env var

TBonnin: works for me

gtrufitt:
Thanks both, I'm now using node env and showing a big message telling people they should run dev-build or article

desbo (approved):
approved-29149_960_720

gtrufitt:
THANKS

johnduffell:
that's massive!

jfsoul:
Nice idea for local testing 👏 

Could have always used an npm script for `NODE_ENV=dev node index.js` (there is a `start` script already) and decoupled slightly more, but not worth worrying about now.

PR #14310, opened 2016-09-14
Hide publication date in old paid content

kelvin-chappell:


common/app/model/trails.scala
```diff
@@ -90,7 +93,13 @@ final case class Trail (
   lazy val showByline: Boolean = tags.isComment
 
   lazy val shouldHidePublicationDate: Boolean = {
-    commercial.isAdvertisementFeature && webPublicationDate.isOlderThan(2.weeks)
+    lazy val isPaidContentInDfp = staticBadgesSwitch.isSwitchedOff && commercial.isAdvertisementFeature
+    lazy val isPaidContentInCapi =
+      staticBadgesSwitch.isSwitchedOn && BrandHunter.findTrailBranding(this, defaultEdition).exists {
```
regiskuckaertz: Just thinking here. Might be worth abstracting this expression in something like `BrandHunter.isPaidContent(this, defaultEdition)`?

regiskuckaertz (approved):
LGTM 👍

common/app/model/trails.scala
kelvin-chappell: Yes.  I think you're right.

PR #14312, opened 2016-09-15
Remove deprecated keyIdentifier

mchv:
## What does this change?

Remove `keyIdentifier` as a property from `bean.js` as this is [now deprecated and will be removed soon](https://www.chromestatus.com/features/5316065118650368).

Currently console display a warning on every page:

```
'KeyboardEvent.keyIdentifier' is deprecated and will be removed in M53, around September 2016. See https://www.chromestatus.com/features/5316065118650368 for more details.
```

As `bean.js` seems to be copied, I don't know if we should diverge or wait for [the upstream issue](https://github.com/fat/bean/issues/128) being fixed and then update to the latest version.  

Prevent breaking frontend (throwing an client side error potentially breaking other features) when latest `chrome` will remove `keyIdentifier`.

`keyIdentifier` does not seems to be used on `frontend` so I don't think it should break anything or anyone.

### DIFF
--- a/static/src/javascripts/components/bean/bean.js
+++ b/static/src/javascripts/components/bean/bean.js
@@ -106,7 +106,7 @@
               'fromElement offsetX offsetY pageX pageY screenX screenY toElement'))
           , mouseWheelProps = mouseProps.concat(str2arr('wheelDelta wheelDeltaX wheelDeltaY wheelDeltaZ ' +
               'axis')) // 'axis' is FF specific
-          , keyProps     = commonProps.concat(str2arr('char charCode key keyCode keyIdentifier '          +
+          , keyProps     = commonProps.concat(str2arr('char charCode key keyCode '          +
               'keyLocation location'))
           , textProps    = commonProps.concat(str2arr('data'))
           , touchProps   = commonProps.concat(str2arr('touches targetTouches changedTouches scale rotation'))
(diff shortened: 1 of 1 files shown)


rich-nguyen:
👍

static/src/javascripts/components/bean/bean.js
```diff
@@ -106,7 +106,7 @@
               'fromElement offsetX offsetY pageX pageY screenX screenY toElement'))
           , mouseWheelProps = mouseProps.concat(str2arr('wheelDelta wheelDeltaX wheelDeltaY wheelDeltaZ ' +
               'axis')) // 'axis' is FF specific
-          , keyProps     = commonProps.concat(str2arr('char charCode key keyCode keyIdentifier '          +
+          , keyProps     = commonProps.concat(str2arr('char charCode key keyCode '          +
```
gtrufitt: Yes, I think until the (issue is fixed in Bean)[https://github.com/fat/bean/issues/128] this is the only thing we can do. Can you add a comment to this PR? 👍

gtrufitt (approved):

static/src/javascripts/components/bean/bean.js
mchv: Sorry updated the link in PR description

AWare: We use this library in membership as well, it might be worth pushing this upstream or just plain forking it before we diverge further.

PR #14314, opened 2016-09-15
Fetch hosted social share texts from capi

kelvin-chappell:
This change uses the socialShareText and shortSocialShareText fields in Composer to populate the twitter, facebook and email texts in hosted content.

/cc @guardian/labs-beta


Calanthe:
👍

PR #14316, opened 2016-09-15
Badges floating fix

Calanthe:
## What does this change?

Fix for the badges in the left column. They weren't clickable because of the `float: left` style.
Now this should be fixed. I also tested `/commercial-containers`, interactive and other article pages.

## Screenshots

Before:
screen shot 2016-09-15 at 11 44 41

After:
screen shot 2016-09-15 at 11 44 51
## Request for comment

@kelvin-chappell


kelvin-chappell:
👍

PR #14318, opened 2016-09-15
Admin->redirects page - change example urls on redirect page to be https

michaelwmcnamara:
This is a small cosmetic change. 
It changes the example urls on the admin->redirect page to be https instead of http.
no impact anywhere else


rich-nguyen:
👍

PR #14324, opened 2016-09-15
Hosted article inline videos

Calanthe:
## What does this change?

Hosted article inline videos look and work much better now.

## Screenshots

Before:
screen shot 2016-09-15 at 15 39 15

After:
screen shot 2016-09-15 at 15 39 29
## Request for comment

@lps88 @blongden73


lps88:
👍

PR #14325, opened 2016-09-15
Chester zoo - ability to switch links on/off

lps88:
## What does this change?

Make it possible to update which links are shown using switches - in case we dont launch all pages together


Calanthe:
👍

PR #14332, opened 2016-09-15
Inline the details polyfill 

SiAdcock:
## What does this change?

Currently the polyfill for `<details>` elements exists in two places:
1. as part of the [paid-containers](https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/commercial/modules/paid-containers.js) module
2. as an [independent module](https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/common/utils/details-polyfill.js) under `utils`. At the moment, this is required only by the [newHeaderNavigation](https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/common/modules/navigation/newHeaderNavigation.js) module

This change will move the polyfill to `common/app/templates/inlineJS/blocking/polyfills`, ensuring that the code is inlined, and is thus available as soon as the page loads.

As a result, I have had to remove the dependencies on `fastDomPromise` and `bean`, as these libraries are not available when the page loads.

I have also removed the polyfill code from the `paid-container` module as the inline polyfill will provide the functionality.

This change ensures `<details>` elements will work across browsers as soon as the page is loaded. The polyfill is very small with zero dependencies, and is loaded in a way that is consistent with other polyfills such as `classList`.

Nope
## Request for comment

@NataliaLKB @gtrufitt @regiskuckaertz


common/app/templates/inlineJS/blocking/polyfills/details.scala.js
```diff
@@ -0,0 +1,49 @@
...
+        }
+        if (targetNode.nodeName.toLowerCase() === 'summary') {
+            toggleDetailsState(targetNode.parentNode);
+        }
+    }
+
+    function bindEvents() {
+        document.addEventListener('click', handleEvent);
+        document.addEventListener('keypress', function (event) {
+            if (event.keyCode === 0x20 || event.keyCode === 0x0D) {
```
regiskuckaertz: `keyCode` is deprecated, has been for a long time, although browsers did not implement the standard `key` property until recently. Let's be defensive here and use `keyCode` only as a fallback, no?

common/app/templates/inlineJS/blocking/polyfills/details.scala.js
```diff
@@ -0,0 +1,49 @@
...
+        while (targetNode.nodeName.toLowerCase() !== 'summary' && targetNode !== document) {
+            targetNode = targetNode.parentNode;
+        }
+        if (targetNode.nodeName.toLowerCase() === 'summary') {
+            toggleDetailsState(targetNode.parentNode);
+        }
+    }
+
+    function bindEvents() {
+        document.addEventListener('click', handleEvent);
```
regiskuckaertz: I think in this case it makes sense to set `useCapture` to `true`. wdyt?

static/src/javascripts/projects/commercial/modules/paid-containers.js
```diff
@@ -4,38 +4,19 @@ define([
...
 
     function init() {
-        if (shouldWePolyfill) {
-            bean.on(document, 'click', 'summary', onClick);
-            bean.on(document, 'keypress', 'summary', onKeyPress(onClick));
-        }
-
         var showMores = qwery('.adverts__more > summary');
         bean.on(document, 'click', showMores, onOpenClick);
         bean.on(document, 'click', showMores, onKeyPress(onOpenClick));
```
regiskuckaertz: woops, I think I made a mistake here, this should really be `'keypress'` 😊

regiskuckaertz: I'll fix that.

regiskuckaertz (changes requested):
Minor comments, but otherwise lgtm 👍

regiskuckaertz:
Waw, I did not realise the `request changes` would actually produce so much red. Let's approve this little buddy.

regiskuckaertz (approved):

common/app/templates/inlineJS/blocking/polyfills/details.scala.js
SiAdcock: Yeah, that's a good shout. It would stop people accidentally breaking the `<details>` functionality if they use `stopPropagation`

SiAdcock:
Thanks @regiskuckaertz, I've made those changes

PR #14333, opened 2016-09-15
Remove circular dependencies in Commercial app

TBonnin:
## What does this change?

Remove circular dependencies in Commercial app
Before this patch:
MasterclassAgent depends on MasterclassTagsAgent and vice-versa
TravelOffersAgent depends on Countries and vice-versa
This patch removes the circular dependencies by moving the logic of one agent into the other ones.

When removing usage of ContentApiClient global object, those agents would need to become classes so the capi client can be injected. This circular dependencies is preventing such a refactoring.
## Request for comment

@JonNorman


TBonnin:
@JonNorman and @regiskuckaertz confirmed those agents are actually not in used anymore.
Removing it will be added to the commercial team backlog tasks 😉 
Still merging so I can move ahead with pre-Play 2.5 refactoring

JonNorman (approved):
Nice work - even if these automatic components are no longer used. 👍

PR #14334, opened 2016-09-15
Add self-validation route for adblocking contributors

JustinPinner:
## What does this change?

Contributor + AdBlocker = poor show. Let contributors validate themselves and carry on.

We don't want to interrupt anyone who's actively supporting us already. Let them tell us if they are a contributor (we already accommodate members and subscribers).

No.

## Screenshots

screen shot 2016-09-15 at 18 44 18

screen shot 2016-09-15 at 18 44 37

screen shot 2016-09-15 at 18 45 19
## Request for comment

Anybody?


paperboyo:
> Anybody?

You’ve asked for it 😉 +1 for the idea! But, judging by the screenshots, the flow seems very much disruptive. Ideally, it would be realised within the same overlay. Then, the fact that “Email” and “Submit” are written using wrong fonts and the button is wrong style would be non-problems (granted, even now they are tiny bit contrived ones).

JustinPinner:
Thanks @paperboyo. There's much to be done in this area yet so all comments duly noted. The final offering needs to be much more polished. And reliable in terms of detection.

kenlim:
👍

common/app/staticpages/StaticPages.scala
```diff
@@ -23,14 +23,34 @@ object StaticPages {
...
-      MetaData.make(
-        id = id,
-        section = Option(SectionSummary(id="global", activeBrandings=None)),
-        webTitle = webTitle,
-        analyticsName = "global",
-        // a currently running AB test is using `contentType = "survey"` change this to "signup" after 2016-09-13
-        contentType = "survey",
-        iosType = None,
-        shouldGoogleIndex = false))
+  def contributorEmailPage: SimplePage = SimplePage(
```
rich-nguyen: I don't really understand this file. If a static page object is created for use inside the `commercial` server, shouldn't the page definition go there too? Not sure why we want to put server-specific stuff in `common`.

static/src/javascripts/projects/common/modules/experiments/contributor-email-form.js
```diff
@@ -0,0 +1,4 @@
+define([], function () {
```
rich-nguyen: what does this do?

JustinPinner: A lot less that was initially intended! I kept the file in case it needed to be there, as everything else I looked at had a companion .js file.

rich-nguyen:
I agree with @paperboyo. I feel it's significantly different a flow to what we really want to do, that it would be hard to draw any conclusion from a test.

Code looks fine, I assume, the email passed in is ignored. 

👍  (technically)

static/src/javascripts/projects/common/modules/experiments/contributor-email-form.js
rich-nguyen: It should be removed, if it doesn't do anything!

common/app/staticpages/StaticPages.scala
JustinPinner: I'll look at relocating it when we go live with the feature. It's here because I followed the same pattern as other pages that I knew worked.

JustinPinner:
Thanks @rich-nguyen and @kenlim.

This test is supposed to examine the differences between behaviour depending on what the user can do with the pop-up. There's a lot more to be done before it can be used to determine much else.

PR #14335, opened 2016-09-16
Change order of precedence for branding

kelvin-chappell:
So that tag branding has precedence over section branding.


regiskuckaertz (approved):
Nice 👍

kelvin-chappell:
@guardian/mobile-server-side-staff This is a good test case: https://www.theguardian.com/global-development/2016/sep/14/africa-shortage-of-engineering-skills-and-female-students-will-stunt-its-growth

PR #14339, opened 2016-09-16
Move comments avatar to discussion frontend

piuccio:
## What does this change?

The user avatar is now part of discussion-frontend. It appears to me that the image is broken. It's always showing the anonymous image even when logged in.
There is some logic to render the actual avatar but it doesn't work. I've fixed that.

The code was implemented in guardian/discussion-frontend#8 and it's the first time I'm bundling CSS as well. Some of the styles are now in `discussion-frontend` and injected in the `head` during rendering. I'm using CSS modules so they shouldn't conflict with what's already in the page.

Part 3 of migrating to discussion frontend.

No, all checks `PASS`
## Screenshots

screen shot 2016-09-16 at 11 25 50

### DIFF
--- a/common/app/common/configuration.scala
+++ b/common/app/common/configuration.scala
@@ -363,7 +363,7 @@ class GuardianConfiguration extends Logging {
     lazy val d2Uid = configuration.getMandatoryStringProperty("discussion.d2Uid")
     lazy val frontendAssetsMap = configuration.getStringProperty("discussion.frontend.assetsMap")
     lazy val frontendAssetsMapRefreshInterval = 5.seconds
-    lazy val frontendAssetsVersion = "v1.1.0"
+    lazy val frontendAssetsVersion = "v1.2.0"
   }
 
   object witness {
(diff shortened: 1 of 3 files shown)


paperboyo:
These are a bit crazy. My Network monitor spends 50 seconds loading 750KB of them images even for collapsed discussion thread... just hitting the test URL  with loads of comments, e.g.: https://www.theguardian.com/us-news/2016/sep/16/donald-trump-campaign-admits-barack-obama-was-born-in-us

Both the loading logic and maybe the images (resizer?) should be rethought, maybe....

NathanielBennett:
:+1:

piuccio:
I've made a change that fixes avatars even outside of the AB test.

@paperboyo you're right, the new `discussion-frontend` eventually will fix that (I hope) because it won't inject comments in the page if they are hidden.

On the loading part, I'm actually confused.
Works like this on frontend:

1. get the user ID from https://discussion.theguardian.com/discussion-api/profile/me the response actually contains an avatarURL, which gets ignored.
2. get the avatar URL from https://avatar.theguardian.com/v1/avatars/user/me/active
3. if the previous avatar API responds correctly, use that value, otherwise build the URL as `https://avatar.guim.co.uk/user/` + userId

In discussion-frontend I'm simply building the URL as in 3, but maybe I can trust the value returned by `profie/me`?

Looping in @guardian/identity for clarification

PR #14340, opened 2016-09-16
Inject wsclient into capi client #4

TBonnin:
## What does this change?

More injection of ContentApiClient.
There is two more references to the ContentApiClient in `TodayCrosswordsGrid` and `TeamMap` but it will requires some additional refactoring.  

➡️  Play 2.5
## Request for comment

@alexduf


facia-press/app/AppLoader.scala
```diff
@@ -19,25 +20,18 @@ class AppLoader extends FrontendApplicationLoader {
...
 
-trait Controllers {
-  def wsClient: WSClient
-  def toolPressQueueWorker: ToolPressQueueWorker
-  def liveFapiFrontPress: LiveFapiFrontPress
-  def draftFapiFrontPress: DraftFapiFrontPress
-  lazy val healthCheck = wire[HealthCheck]
-  lazy val applicationController: Application = wire[Application]
-}
+trait AppComponents extends FrontendComponents {
```
TBonnin: I tried the approach of one single big `AppComponents` (not getting rid of `FrontendComponents` obviously) we talked about a few days ago. It works quite well for `facia-press` since it is not this big and dev-build/preview don't depend on it.
Let me know what you think. (Note: I have done the same thing with the Rss app for the same reason. See below)

I will try it on `article` or `applications` at some point in the future, that's where trouble will appear I think. 😉

alexduf: I think you're right, it does make it easier to digest. Let's see how it scales!

alexduf (approved):
It looks good :+1: !!

PR #14348, opened 2016-09-16
Removing TodayCrosswordGrid

TBonnin:
## What does this change?

TodayCrosswordGrid is not working and apparently never worked.

Removing unused feature and help removing intricate use of ContentApiClient :winwin:
## Request for comment

@gtrufitt


gtrufitt (approved):
Looks good.

PR #14350, opened 2016-09-16
Fix test page

Calum-Campbell:
## What does this change?
- Fix the commercial/test-page 404'ing problem
- Lowercase the if condition so as not to jump out of it and return not found. 
- possibly due to https://github.com/guardian/frontend/pull/14307
## Request for comment

@regiskuckaertz @rich-nguyen


rich-nguyen:
👍

PR #14353, opened 2016-09-16
Hide video across all Fabrics when ad collapsed

regiskuckaertz:
I previously added this fix only to the v2 template, but it applies to the v1 as well.


rich-nguyen:
👍

PR #14354, opened 2016-09-16
Each Header-Bidding test variant should be logging commercial reports

rich-nguyen:
These variants should be logging commercial reports.


janua:
:+1:

PR #14356, opened 2016-09-16
CSS-only header navigation

SiAdcock:
## What does this change?
- implemented a CSS-only version of the new header navigation. This is available as soon as the HTML and CSS has loaded
- stripped most of the JavaScript functionality of the header; JS now provides only accessibility and user experience enhancements. Also removed most of its third party dependency libraries, with the exception of `fastdom` and `qwery`
- moved the JavaScript from enhanced into standard

Currently, the new header navigation does not respond to user interaction until all enhanced JavaScript has been loaded. There is an unacceptable delay between user interaction and perceived response from the header UI. Implementing a CSS-only version of the navigation ensures there will be no such delay.

The CSS-only version of the navigation is not accessible, and the user experience is sub-optimal. Therefore we will still need to maintain minimal JavaScript that addresses these issues.

Nope
## Request for comment

@NataliaLKB @gtrufitt @rich-nguyen @TBonnin @zeftilldeath @regiskuckaertz

### DIFF
--- a/common/app/views/fragments/nav/editionPicker.scala.html
+++ b/common/app/views/fragments/nav/editionPicker.scala.html
@@ -4,7 +4,7 @@
 
 <div class="edition-picker js-edition-picker-container">
 
-    <input class="edition-picker__checkbox js-enhance-checkbox"
+    <input class="edition-picker__button js-enhance-checkbox js-edition-picker-button"
         type="checkbox"
         id="edition-picker"
         aria-controls="edition-picker__dropdown">
--- a/common/app/views/fragments/nav/editionPicker.scala.html
+++ b/common/app/views/fragments/nav/editionPicker.scala.html
@@ -4,7 +4,7 @@
 
 <div class="edition-picker js-edition-picker-container">
 
-    <input class="edition-picker__button js-enhance-checkbox js-edition-picker-button"
+    <input class="edition-picker__button js-enhance-checkbox"
         type="checkbox"
         id="edition-picker"
         aria-controls="edition-picker__dropdown">
--- a/common/app/views/fragments/nav/newHeaderMenu.scala.html
+++ b/common/app/views/fragments/nav/newHeaderMenu.scala.html
@@ -3,10 +3,10 @@
 @import common.{ NewNavigation, LinkTo, Edition }
 
 @sectionList(topLevelSection: NewNavigation.EditionalisedNavigationSection) = {
-    <li class="main-navigation__item navigation-border">
+    <li class="main-navigation__item navigation-border js-navigation-item">
             <!-- TODO: Firefox -->
         <details class="js-close-nav-list main-navigation__primary-list" id="primary-list-@topLevelSection.name" open>
-            <summary class="main-navigation__item__button">
+            <summary class="main-navigation__item__button js-navigation-button">
                 <i class="main-navigation__icon"></i>
                 @topLevelSection.name
             </summary>
--- a/common/app/views/fragments/nav/newHeaderMenu.scala.html
+++ b/common/app/views/fragments/nav/newHeaderMenu.scala.html
@@ -5,7 +5,7 @@
 @sectionList(topLevelSection: NewNavigation.EditionalisedNavigationSection) = {
     <li class="main-navigation__item navigation-border">
             <!-- TODO: Firefox -->
-        <details class="js-close-nav-list main-navigation__primary-list" id="primary-list-@topLevelSection.name">
+        <details class="js-close-nav-list main-navigation__primary-list" id="primary-list-@topLevelSection.name" open>
             <summary class="main-navigation__item__button">
                 <i class="main-navigation__icon"></i>
                 @topLevelSection.name
@@ -16,6 +16,7 @@
                     navigation-group
                     navigation-group--@{edition.id.toLowerCase}
                     js-editionlise-secondary-nav"
+                    data-edition="@{edition.id.toLowerCase}"
                     @if(edition.id.toLowerCase != "uk") { hidden }> @* Our default edition is UK *@
 
                     @topLevelSection.getEditionalisedNavLinks(edition).map { sectionItem =>
@@ -29,11 +30,12 @@
     </li>
 }
 
-<div class="main-menu-container" id="main-menu">
-        <!-- TODO: Provide tabbable alternative -->
-    <a href="#" class="main-menu-container__overlay" aria-hidden="true"></a>
+<div href="#" class="main-menu-container__overlay" aria-hidden="true"></div>
 
-    <div class="main-menu-container__menu js-reset-scroll-on-menu">
+<div class="main-menu-container">
+        <!-- TODO: Provide tabbable alternative -->
+
+    <div class="main-menu-container__menu js-reset-scroll-on-menu" id="main-menu" aria-hidden="true">
         <ul class="main-navigation">
             @NewNavigation.topLevelSections.map { section =>
                 @sectionList(section)
--- a/common/app/views/fragments/nav/newHeaderMenu.scala.html
+++ b/common/app/views/fragments/nav/newHeaderMenu.scala.html
@@ -30,9 +30,10 @@
     </li>
 }
 
-<div href="#" class="main-menu-container__overlay" aria-hidden="true"></div>
+<label for="main-menu-toggle" class="main-menu-container__overlay" aria-hidden="true"></label>
 
 <div class="main-menu-container">
+
         <!-- TODO: Provide tabbable alternative -->
 
     <div class="main-menu-container__menu js-reset-scroll-on-menu" id="main-menu" aria-hidden="true">
@@ -78,3 +79,4 @@
         </ul>
     </div>
 </div>
+
--- a/common/app/views/fragments/nav/newHeaderMenu.scala.html
+++ b/common/app/views/fragments/nav/newHeaderMenu.scala.html
@@ -34,9 +34,9 @@
 
 <div class="main-menu-container">
 
-        <!-- TODO: Provide tabbable alternative -->
+    <!-- TODO: Provide tabbable alternative -->
 
-    <div class="main-menu-container__menu js-reset-scroll-on-menu" id="main-menu" aria-hidden="true">
+    <div class="main-menu-container__menu js-main-menu" id="main-menu" aria-hidden="true">
         <ul class="main-navigation">
             @NewNavigation.topLevelSections.map { section =>
                 @sectionList(section)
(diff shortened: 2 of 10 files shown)


common/app/views/fragments/nav/newHeaderMenu.scala.html
```diff
@@ -5,17 +5,17 @@
 @sectionList(topLevelSection: NewNavigation.TopLevelSection) = {
     <li class="main-navigation__item navigation-border">
             <!-- TODO: Firefox -->
-        <details class="js-close-nav-list main-navigation__primary-list" id="primary-list-@topLevelSection.name">
+        <details class="js-close-nav-list main-navigation__primary-list" id="primary-list-@topLevelSection.name" open>
```
NataliaLKB: Note: `open` is added to all the details elements so they are all open by default

static/src/javascripts/projects/common/modules/navigation/edition-picker.js
```diff
@@ -1,84 +1,41 @@
...
-    $
+    fastdom,
+    qwery
 ) {
-    var dropdown = $('.js-edition-picker-dropdown');
-    var button;
+    function editionPickerClickHandler(event) {
+        event.stopPropagation();
+        var button = event.target;
+        var editionPickerDropdown = qwery('.js-edition-picker-dropdown')[0];
```
NataliaLKB: Will the edition picker dropdown always be there? If not, we should wrap line 25 and 34 in` if (editionPickerDropdown) ...` or use a try catch.

static/src/javascripts/projects/common/modules/navigation/newHeaderNavigation.js
```diff
@@ -1,159 +1,160 @@
...
+        if (!mainMenu) {
+            return;
+        }
+        if (menuIsOpen()) {
+            fastdom.write(function () {
+                var mainListItems = qwery('.main-navigation__item');
+
+                button.setAttribute('aria-expanded', 'false');
+                mainMenu.setAttribute('aria-hidden', 'true');
+                veggieBurgerLink[0].classList.remove('new-header__nav__menu-button--open');
```
NataliaLKB: Should this have a check that the `veggieBurgerLink` exists before trying to use it?

common/app/views/fragments/nav/newHeaderMenu.scala.html
regiskuckaertz: I think you still need to specify it for browsers that don't support `details` yet. The CSS rule will kick off and not hide its content.

static/src/javascripts/projects/common/modules/navigation/edition-picker.js
SiAdcock: Good call. Changed!

common/app/views/fragments/nav/newHeaderMenu.scala.html
```diff
@@ -29,11 +29,12 @@
     </li>
 }
 
-<div class="main-menu-container" id="main-menu">
+<div href="#" class="main-menu-container__overlay" aria-hidden="true"></div>
```
regiskuckaertz: Why isn't this an `A` element?

static/src/javascripts/projects/common/modules/navigation/newHeaderNavigation.js
SiAdcock: Sure thing. I've also brought the variable assignment into the handler function as it is not used elsewhere

common/app/views/fragments/nav/newHeaderMenu.scala.html
NataliaLKB: Sharp eyes @regiskuckaertz ! This was an `a` tag, but with the new and improved implementation it doesn't make sense for it to be that anymore. I think @SiAdcock is going to investigate making it either a label (so will uncheck the checkbox) or it will only be a functional way of closing the menu if Javascript is involved.

common/app/views/fragments/nav/editionPicker.scala.html
```diff
@@ -4,7 +4,7 @@
 
 <div class="edition-picker js-edition-picker-container">
 
-    <input class="edition-picker__checkbox js-enhance-checkbox"
+    <input class="edition-picker__button js-enhance-checkbox js-edition-picker-button"
```
gtrufitt: Is this JS class used anywhere?

gtrufitt: if so I think it should follow the BEM style `js-edition-picker__button` still (personal preference, don't think that's written down anywhere so up to you...)

SiAdcock: Nice one @gtrufitt, it's not being used anywhere so I'll just trash it.

static/src/stylesheets/layout/_ab-new-header-test-variant.scss
```diff
@@ -323,10 +302,10 @@ $gutter-large: 55px;
     }
 }
 
-.edition-picker__checkbox, .edition-picker__button {
+.edition-picker__button {
     display: none;
 
-    &:checked, &.open {
+    &:checked, &[aria-expanded='true'] {
```
gtrufitt: Minor, selector on a new line

common/app/views/fragments/newHeader.scala.html
```diff
@@ -26,16 +23,17 @@ <h1 class="u-h">The Guardian</h1>
...
+                <i class="new-header__veggie-burger-icon"></i>
                 <span class="u-h">Menu</span>
-            </a>
-            <ul class="new-header__nav__list">
-                @NewNavigation.topLevelSections.map { section =>
-                    @primaryLinks(section.name)
-                }
-            </ul>
+            </label>
+            <input type="checkbox" id="main-menu-toggle" class="new-header__nav__button js-enhance-checkbox" aria-controls="main-menu">
```
gtrufitt: Does this input need to close?

gtrufitt (commented):
Really great to see the reliance on JS get less, nice job!

static/src/javascripts/projects/common/modules/navigation/newHeaderNavigation.js
```diff
@@ -1,159 +1,160 @@
...
 ) {
-    var mainMenuId = '#main-menu';
-    var html = $('html');
-    var mainMenuEl = $(mainMenuId);
-    var veggieBurgerLink = $('.js-change-link');
-    var primaryItems = $('.js-close-nav-list');
+    var html = qwery('html')[0];
+    var menuItems = qwery('.js-close-nav-list');
+    var buttonClickHandlers = {
+        'main-menu-toggle': veggieBurgerClickHandler,
```
gtrufitt: Oh... err. Carry on then.

common/app/views/fragments/nav/newHeaderMenu.scala.html
```diff
@@ -29,11 +29,13 @@
     </li>
 }
 
-<div class="main-menu-container" id="main-menu">
+<label for="main-menu-toggle" class="main-menu-container__overlay" aria-hidden="true"></label>
```
gtrufitt: k remembered the form element is the tabbabable element so if you're avoiding tabbing to the checkbox this is cool innit.

SiAdcock: Since this test is for mobile only, I think tabbability or non-tabbability is out of scope, at least for this PR. What do you think?

gtrufitt: Yeah, fair enough at the mo

common/app/views/fragments/newHeader.scala.html
SiAdcock: As per [this comment](https://github.com/guardian/frontend/pull/14356#r79419807), we won't focus on tabbing in this PR.

I don't believe void tags like `<input>` need to be closed in HTML5, but I'll defer to any style rules that say otherwise 😄

common/app/views/fragments/nav/newHeaderMenu.scala.html
```diff
@@ -29,11 +29,12 @@
...
 
-<div class="main-menu-container" id="main-menu">
+<div href="#" class="main-menu-container__overlay" aria-hidden="true"></div>
+
+<div class="main-menu-container">
         <!-- TODO: Provide tabbable alternative -->
-    <a href="#" class="main-menu-container__overlay" aria-hidden="true"></a>
 
-    <div class="main-menu-container__menu js-reset-scroll-on-menu">
+    <div class="main-menu-container__menu js-reset-scroll-on-menu" id="main-menu" aria-hidden="true">
```
SiAdcock: I've updated the class name to `js-main-menu`. The JavaScript that does the scrolling now looks like:

```
var menuContainer = qwery('.js-main-menu')[0];

menuContainer.scrollTop = 0;
```

static/src/javascripts/projects/common/modules/navigation/newHeaderNavigation.js
```diff
@@ -1,159 +1,160 @@
...
+    function enhanceToButton() {
+        var checkboxes = qwery('.js-enhance-checkbox');
+        fastdom.read(function () {
+            var buttons = checkboxes.map(function (checkbox) {
+                var button = document.createElement('button');
+                var checkboxId = checkbox.id;
+                var checkboxControls = checkbox.getAttribute('aria-controls');
+                var forEachClass = Array.prototype.forEach.bind(checkbox.classList);
+
+                forEachClass(function (c) {
```
SiAdcock: TL;DR this doesn't quite work as one might expect!

Actually this won't work because the context gets broken by `bind`. Even if we rebind the correct context using `button.classList.add.bind(button.classList)`, the `forEach` callback passes the loop index and a reference to the checkbox classList array, which get stringified and added to the button classList.

gtrufitt: 😨 🎉

common/app/views/fragments/newHeader.scala.html
```diff
@@ -26,16 +23,17 @@ <h1 class="u-h">The Guardian</h1>
...
 
         <nav class="new-header__nav">
-            <a href="#main-menu" class="new-header__nav__menu-button js-change-link" tabindex="0">
-                <i class="new-header__veggie-burger-icon">
-                </i>
+            @NewNavigation.topLevelSections.map { section =>
+                @primaryLinks(section.name)
+            }
+            <label for="main-menu-toggle" class="new-header__nav__menu-button js-change-link" tabindex="0">
+                <i class="new-header__veggie-burger-icon"></i>
```
SiAdcock: The icon is pure HTML / CSS, which allows us to animate the elements using CSS transform and transition, magically transforming the three-line veggie burger into a X icon. 

I do agree that `<span>` is semantically more correct than `<i>`, so I will update.

PR #14362, opened 2016-09-19
Audience Science without localStorage

regiskuckaertz:
The guys from Audience Science told us we should never cache and reuse codes across page views; they're only valid for one impression.

This change removes all the `localStorage` logic and makes sure codes are added only when both DFP and AS are loaded.


static/src/javascripts/projects/commercial/modules/third-party-tags/audience-science-pql.js
```diff
@@ -28,36 +28,52 @@ define([
...
-                    return ['pq_' + placement, 'T'];
-                }).reduce(function (result, input) {
-                    result[input[0]] = input[1];
-                    return result;
-                }, {});
-            // set up the global asiPlacements var in case dfp returns before asg
-            window.asiPlacements = storedSegments[section];
+        var placements = window.asiPlacements || {};
+        return Object.keys(placements)
+        .filter(function (placement) {
```
rich-nguyen: indentation

static/src/javascripts/projects/commercial/modules/third-party-tags/audience-science-pql.js
```diff
@@ -28,36 +28,52 @@ define([
...
 
     function onLoad() {
-        var asiPlacements = window.asiPlacements;
-        var segments = storage.local.get(storageKey) || {};
-        // override the global value with our previously stored one
-        window.asiPlacements = segments[section];
-        segments[section] = asiPlacements;
-        storage.local.set(storageKey, segments);
+        isLoaded = true;
+        mediator.emit('commercial:audience-science:loaded');
```
rich-nguyen: why not call `setAudienceScienceCallback` and `setAudienceScienceKeys` here? then you don't need `mediator`, `isLoaded`, `init`, or `run`.

rich-nguyen:
👍

static/src/javascripts/projects/commercial/modules/third-party-tags/audience-science-pql.js
regiskuckaertz: We can't assume DFP will be loaded. But you're right, even if it is not loaded the googletag queue is already present. Good catch!

rich-nguyen: I think i've missed something.  I'm not sure how the code guarantees that `pql` is loaded, and targeting has been set, before the dfp requests are sent out?

static/src/javascripts/projects/commercial/modules/third-party-tags/audience-science-pql.js
regiskuckaertz: That's @piuccio's fault! 😆  I've come to like it actually, the `.` is enough for me to understand it is linked to the previous line. Also I had a teacher who used to write pseudo-code in algol with little to no indentation at all---and yet it was crystal clear.

piuccio: lol, I use this style because it's clear (for me) and because if you indent the `.filter` most editors mess up with indentation, so when you press enter they stay on the `.` level instead of going back to `return`. So yeah

static/src/javascripts/projects/commercial/modules/third-party-tags/audience-science-pql.js
regiskuckaertz: In the end I removed the `shouldLazyLoad` test entirely. AS also loads in single-call mode; the test was initially used to _not_ add the callback that removes keys, but I'm OK to let that go.

static/src/javascripts/projects/commercial/modules/third-party-tags/audience-science-pql.js
regiskuckaertz: Also when you work with D3 this is a life saver. I will probably give me a few additional years before having carpal tunnel syndrome 😆

static/src/javascripts/projects/commercial/modules/dfp/init.js
```diff
@@ -45,12 +47,7 @@ define([
                 require(['js!googletag.js']);
                 ophanTracking.addTag('sonobi');
             } else {
-                if (!window.googletag) {
-                    window.googletag = {cmd: []};
-
-                    // If we don't already have googletag, create command queue and load it async.
-                    require(['js!googletag.js']);
-                }
+                require(['js!googletag.js']);
```
rich-nguyen: looks like you can lift the `require` out of the if clause now, nice.

static/src/javascripts/projects/commercial/modules/third-party-tags/audience-science-pql.js
rich-nguyen: I think file consistency is pretty much the only style rule I follow, so whichever way works is fine, but not mixed in the same file!

regiskuckaertz: Where's that? 😮

rich-nguyen: you deleted it! :)

PR #14363, opened 2016-09-19
Fix related content container on paid content

kelvin-chappell:
The sponsorship type wasn't being set client-side since the logo migration, which meant the related content container was bringing in obscure content on paid pages.  This should fix it.

Eg. https://www.theguardian.com/discover-cool-canada/2016/sep/09/magnifique-montreal-at-a-glance

Before:
image

After:
image


regiskuckaertz:
👍

JonNorman:
:+1:

PR #14364, opened 2016-09-19
deploy.json: Set publicReadAcl to false and remove not needed parameters

TBonnin:
## What does this change?

Set publicReadAcl to false and remove not needed parameters in deploy.json files
- App is downloaded using `aws s3 cp` so no need for public read permission
- `port` and `healthcheck_paths` are not supported params of autoscaling
  deploy type (https://riffraff.gutools.co.uk/docs/magenta-lib/types#autoscaling)
## Request for comment

@sihil @guardian/dotcom-platform

### DIFF
--- a/admin-jobs/deploy.json
+++ b/admin-jobs/deploy.json
@@ -5,13 +5,10 @@
             "type":"autoscaling",
             "data":{
                 "secondsToWait":1200,
-                "port":18080,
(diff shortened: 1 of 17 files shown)


sihil (approved):
Good to see the unused parameters being removed too.

PR #14365, opened 2016-09-19
Immersive image of explore template. 

blongden73:
Adding the same css to this template as in the immersive template for getting the full screen image when the immersive switch is turned on. 

@SiAdcock @gtrufitt - Could you take a look please.


static/src/stylesheets/inline/article-labour-liverpool.scss
```diff
@@ -87,6 +92,14 @@ $labour-liverpool-header-element-opacity: .85;
     }
 }
 
+.heroic-immersive{
```
gtrufitt: Don't need to test here `.heroic-immersive .labour-livepool-headline` is fine

Also, in line with BEM you probably want to describe this a modification of headline like `.headline--labour-liverpool`

gtrufitt (approved):
Two minor changes but otherwise cool 👍

static/src/stylesheets/inline/article-labour-liverpool.scss
blongden73: @gtrufitt I'm going to untangle these styles and naming this week to be called "Explore" anyway so Labour and Liverpool will be gone. So I'll sort this out then, Simon set all these styles up so it will be a bit of work to untangle. What do you think?

gtrufitt: Yeah, ok, if that's your plan this week 👍

PR #14369, opened 2016-09-19
Tidy up hosted templates

lps88:
# What does this change?
- Move the brand-specific colour style rules into the templates
- delete `_hosted-colours.scss`
- Fix Zootropolis page styling
- make first-letter selector more specific (supposed to apply to drop cap only but was catching some other elements)

We no longer need the brand-specific css that was in `_hosted-colours.scss`, and also can delete corresponding css classes from the model and html. Now there is far less duplication, and no need to add new style rules for each campaign.
## Screenshots

Fixed:
image
## Request for comment

@Calanthe


Calanthe:
nice 👍

PR #14370, opened 2016-09-19
Regenerating test files in data/database and data/discussion

TBonnin:
## What does this change?

Regenerating test files in data/database and data/discussion

Those test data files are generated automatically in function of the endpoint url, headers and query string params. 
It is quite easy to add a new one but cleaning old ones that have been replaced and are not needed anymore is almost impossible to do since it's very hard to figure out which ones are in use.
This patch is removing unused test data files (7 files) as well as regenerating used ones (60 files)
## Request for comment

@guardian/dotcom-platform @johnduffell


NataliaLKB:
I can't see most of the files, but the tests pass so 👍 . Thanks for removing the unused files!

common/test/views/support/ContentFooterContainersLayoutTest.scala
```diff
@@ -81,7 +81,6 @@ class ContentFooterContainersLayoutTest extends FlatSpec with Matchers {
 
   it should "include story package placeholder even when there's no story package to show" in {
     val html = buildHtml(contentItem(showInRelatedContent = false), emptyRelatedContent)
-    println("'" + html + "'")
```
johnduffell: 😆

johnduffell: 😆

johnduffell:
thanks!

PR #14374, opened 2016-09-19
Remove immersiveMainMediaEmbedSwitch

sammorrisdesign-zz:
## What does this change?

This removes the `immersiveMainMediaEmbedSwitch` that is due to expire in a week or so.

Nah


rich-nguyen:
👍

PR #14375, opened 2016-09-19
Programmatic advert load times

rich-nguyen:
This changes the dashboard to measure the individual advert load times. It also backfills the data to increase the accuracy of the averages calculated.


janua:
:+1:

Issue #5237, opened 2014-07-28
[Video player] General critique (long)


paperboyo:
Hello!

I am not even sure it belongs on github, there are definitely some bugs here, but mostly these would be suggestions/UX musings, so feel free to close whenever. I wanted to reach all potentially interested, though. Also, it's all low priority and only my humble opinions, as always, so fell free to replace 'should' with 'could/would be nice' and 'shit' with 'subprime/award-winning'.

They are in no particular order, but I've tried to keep the "bugs" first...

1. Just before the video starts most browsers show their native controls for a moment and the poster frame is being resized/jumps - can we hide all that ugliness?

2. The interaction with the volume bar is unsatisfactory - it provides misleading feedback: the bar that signifies the current volume level can be dragged out of view at extremes.

3. At the end of the video when clicking on any item on an end plate while in full screen, video exits full screen, browser loads another video page and then the user has to click full screen again and press play to enjoy the next video. Instead, the next video should load and auto play all while in full screen mode and the underlying video page should (probably) change without the interruption in user’s experience. Maybe, it should work that way also when not in full screen?

4. When end plate is reached, the play icon should change to a replay icon.

5. If you are watching any of Popular videos, it’s also shown in Popular videos again (and sometimes is also shown in myriads of other containers:)

6. Currently in full screen, the control bar disappears when there is no user interaction wherever over the full screen, I think the control bar should stay open when the mouse is over the control bar (this is how Youtube works), but should close when the mouse is at the very bottom edge of the screen (not how Youtube works) - while I am not sure about this one, there are alternatives which are still better than the current behaviour (like the fade out to be MUCH longer while the cursor is over the control bar)…

7. With Youtube, BBC iPlayer, all browser’s native video players and many more, a video can be brought in and out of full screen by double clicking on it. I would then consider it a bug rather than the lack of a feature, that our videos do not work that way.

8. What’s the status of custom skins/controls on mobile? Isn’t Chrome on Android ready to use them?

9. Can we wrap all (also, third party, embedded) videos in our customized controls/skin?

10. In general, I think that for some types of content there are some metadata information that should be made available to the user as soon as possible - sometimes right along the thumbnails (like date for news, authorship for comment pieces etc.). That very important piece of info for videos is the length of the clip. I think it should be visible over video item thumbnails in all contexts. But at the very least, it should be available when user opens the video page. Currently, only the poster frame pic and the play icon is visible.

[...]

PR #14378, opened 2016-09-19
Adding duration to the video button - AB test

mr-mr:
## What does this change?

Adds duration to the video button across the site as an AB test.

This has been tested before by showing the control bar and designs were shown in the UX lab. This is the first (small) step in implementing a new style for video cards and embeds. We've learnt from lab  and user panel testing that showing certain information up front before playing a video makes the click more valuable. 

No (ran test, all pass)
## Screenshots

image

image
## Request for comment

@akash1810 @jamespamplin @gidsg @sndrs

### DIFF
--- a/common/app/views/fragments/media/video.scala.html
+++ b/common/app/views/fragments/media/video.scala.html
@@ -41,7 +41,8 @@
                 data-auto-play="@player.autoPlay"
                 data-show-end-slate="@(player.showEndSlate && showEndSlate)"
                 @if(showPoster) { poster="@player.poster" }
-                data-duration="@player.video.videos.duration.toString()"
+                data-duration="@player.video.videos.formattedDuration.toString()"
+                data-formatted-duration="@player.video.videos.formattedDuration.toString()"
                 data-media-id="@player.video.properties.id"
                 data-end-slate="@player.endSlatePath"
                 data-block-video-ads="@player.blockVideoAds"
--- a/common/app/views/fragments/media/video.scala.html
+++ b/common/app/views/fragments/media/video.scala.html
@@ -42,7 +42,6 @@
                 data-show-end-slate="@(player.showEndSlate && showEndSlate)"
                 @if(showPoster) { poster="@player.poster" }
                 data-duration="@player.video.videos.formattedDuration.toString()"
-                data-formatted-duration="@player.video.videos.formattedDuration.toString()"
                 data-media-id="@player.video.properties.id"
                 data-end-slate="@player.endSlatePath"
                 data-block-video-ads="@player.blockVideoAds"
--- a/common/app/views/fragments/media/video.scala.html
+++ b/common/app/views/fragments/media/video.scala.html
@@ -41,7 +41,8 @@
                 data-auto-play="@player.autoPlay"
                 data-show-end-slate="@(player.showEndSlate && showEndSlate)"
                 @if(showPoster) { poster="@player.poster" }
-                data-duration="@player.video.videos.formattedDuration.toString()"
+                data-duration="@player.video.videos.duration.toString()"
+                data-formatted-duration="@player.video.videos.formattedDuration.toString()"
                 data-media-id="@player.video.properties.id"
                 data-end-slate="@player.endSlatePath"
                 data-block-video-ads="@player.blockVideoAds"
--- a/static/src/javascripts/bootstraps/enhanced/media/main.js
+++ b/static/src/javascripts/bootstraps/enhanced/media/main.js
@@ -108,7 +108,7 @@ define([
             if (!isNaN(duration)) {
                 player.duration(duration);
                 player.trigger('timeupdate'); // triggers a refresh of relevant control bar components
-                // player.controlBar.addChild('button', { text: 'hello' });
+                // Sets the duration for the Video JS button element
                 $('.vjs-big-play-button')[0].setAttribute('data-duration', $(el).attr('data-duration'));
             }
             // we have some special autoplay rules, so do not want to depend on 'default' autoplay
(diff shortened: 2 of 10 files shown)


paperboyo:
Woohoo! Took us some time to test the obvious ([pt. 10](https://github.com/guardian/frontend/issues/5237#issue-38899877)). 😆 

I would only say (liking how it looks!):
- good thing: it won't cover faces that tend to be centrally cropped
- bad thing: because it makes play button bigger and is not transparent even by a touch - it will cover an even bigger portion of the image area
- style is not consistent with the only place where we show duration now - the carousel - so either one should change in the future
- why is it differently positioned vertically depending on the card size? (my subjective preference and consistency with apps would suggest bottom left like in the screenshot of the small card)

static/src/javascripts/bootstraps/enhanced/media/main.js
```diff
@@ -118,7 +118,15 @@ define([
 
     function initPlayButtons(root) {
         fastdom.read(function () {
-            $('.js-video-play-button', root).each(function (el) {
+            $('.js-video-play-button', root).each(function (el, options, duration) {
+                function initButtonDuration() {
```
akash1810: I would declare this function outside of the `.each` so that its declared once in total, rather than once per loop

static/src/javascripts/bootstraps/enhanced/media/main.js
```diff
@@ -118,7 +118,15 @@ define([
...
+                  el.getAttribute('data-duration', duration);
```
akash1810: not sure this'll do anything...

static/src/stylesheets/module/content/_media-player.scss
```diff
@@ -152,6 +152,76 @@ $ima-controls-height: 70px;
...
+.vjs-big-play-button__duration {
+    cursor: pointer;
+    display: inline-block;
+    z-index: 2;
+    padding: 0 10px;
+    position: absolute;
+    bottom: 8px;
+    left: 5px;
+    line-height: 28px;
+    padding-left: 41px;
```
akash1810: can we add this rule to the `padding` one above?

static/src/stylesheets/module/content/_media-player.scss
```diff
@@ -1095,7 +1165,7 @@ $ima-controls-height: 70px;
 // we had major refactoring issues.
 .gu-media--show-controls-at-start.vjs-paused {
     .vjs-control-bar {
-        bottom: 0;
+        bottom: -3.95em;
```
akash1810: this feels a little magical - should we add a comment against the value of this property?

common/app/views/fragments/media/video.scala.html
```diff
@@ -41,7 +41,8 @@
                 data-auto-play="@player.autoPlay"
                 data-show-end-slate="@(player.showEndSlate && showEndSlate)"
                 @if(showPoster) { poster="@player.poster" }
-                data-duration="@player.video.videos.duration.toString()"
+                data-duration="@player.video.videos.formattedDuration.toString()"
```
akash1810: I think we'd want to keep `data-duration` as the raw value, or remove `data-formatted-duration` as there isn't much use in them both being the same value?

static/src/javascripts/bootstraps/enhanced/media/main.js
```diff
@@ -95,16 +95,16 @@ define([
         $('.vjs-fullscreen-control', player.el()).attr('aria-label', 'video fullscreen');
     }
 
-    function createVideoPlayer(el, options) {
+    function createVideoPlayer(el, options, duration) {
```
akash1810: I think `duration` will be `undefined` here as we haven't updated how we [call](https://github.com/guardian/frontend/blob/master/static/src/javascripts/bootstraps/enhanced/media/main.js#L238) `createVideoPlayer` to pass in duration...

mr-mr:
@paperboyo 

good thing: it won't cover faces that tend to be centrally cropped
bad thing: because it makes play button bigger and is not transparent even by a touch - it will cover an even bigger portion of the image area
With the button being transparent it was constantly being missed. With this being a video the button should trump the image i.m.o. As this is an AB test I'm keen to see this further our previous yellow button test.

style is not consistent with the only place where we show duration now - the carousel - so either one should change in the future - agreed, a later change.

why is it differently positioned vertically depending on the card size? (my subjective preference and consistency with apps would suggest bottom left like in the screenshot of the small card) 
The control bar gets in the way on hover, and as such it was placed vertically and left. This is one part of updating the poster image for the video so I'd be happy to go ahead with this as a test, knowing that Youtube will render the current control bar redundant.

paperboyo:
> The control bar gets in the way on hover, and as such it was placed vertically and left.

@mr-mr Oh, yes: silly me. But it will get in the way on a small card, looking at the above screenshot, no?

> With this being a video the button should trump the image i.m.o.

Agreed. And it’s gonna be better now, as less people will have a button instead of a face 😄 

> This is one part of updating the poster image for the video

These video trail pics and the video poster images (different beasts) still could use some imgIX magic while we’re at it: [INTERNAL RESOURCE](https://trello.com/c/cWSiXgfV/87-p3-smallworks-all-imagery-to-properly-use-master-assets-and-resizer-responsively).

mr-mr:
@paperboyo The small card playing video is an issue in itself. I'm OK with this small bug for the weeks test if you are. The cards playing generally, especially the small cards is a longer conversation!

paperboyo:
> if you are

Oh, I’m just a bystander to be ignored at will!

I agree that videos playing in the small cards may seem... questionable. But at least now they have a fullscreen button visible (was lacking for a year or so) in addition to a standard doubleclick-to-fullscreen. Ideally, the new button would disappear if the control bar gets in the way (as control bar duplicates both of its functions: play and duration). Or at least sit comfortably _behind_ the control bar (can’t see how they interact from the screenshots).

akash1810 (approved):
LGTM 👍

PR #14379, opened 2016-09-19
Subscribe everyone to the WR variant, until AB buckets are balanced

katebee:
## What does this change?

Add all Weekend Reading subscriptions to `version 2` until balance is restored (and this PR can be reverted)

Temporarily puts all Weekend Reading subscriptions into ListID **3743**.

We a currently running a site-wide AB test to subscribe users to two different email formats. Randomisation seems to be favouring `version 1` (ListID **3744**) significantly.

(Randomisation occurs when the user visits the Guardian, irrespective of if they sign up to the email, or visit the email page)

We now have twice as many people signed up to `version 1` as to `version 2` of the Weekend Reading email. This PR aims to balance the groups.

Nope
## Request for comment

@joelochlann

### DIFF
--- a/applications/app/views/signup/weekendReading.scala.html
+++ b/applications/app/views/signup/weekendReading.scala.html
@@ -25,7 +25,7 @@
                     </div>
 
                     <div class="email-signup" data-component="weekend-reading">
-                        <form action="https://www.theguardian.com/email" method="post" class="email-sub__form js-email-sub__form email-sub__form--article" data-email-form-type="article" data-email-list-id="3744" data-link-name="referrer-unknown">
+                        <form action="https://www.theguardian.com/email" method="post" class="email-sub__form js-email-sub__form email-sub__form--article" data-email-form-type="article" data-email-list-id="3743" data-link-name="referrer-unknown">
                           <div class="email-sub__form-wrapper">
                             <div class="email-sub__inline-label js-email-sub__inline-label">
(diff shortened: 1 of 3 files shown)


joelochlann:
I don't think the problem is with the "randomisation" per se - looking at #14171 I can see that **3744** is the one which is in the original page that gets served, and it only gets changed to 3743 by JavaScript. So probably for the extra users going into the 3744 bucket, the JS is not executing - either JS is turned off, there's a bug or browser compatibility issue in the code, they submit the form before JS executes, or some combination of the above

katebee:
@joelochlann yeah! 😄 Sorry, I didn't mean that the AB randomisation was the cause. My main theory was that this is a symptom of JS not running. I thought that changing the default value in the html would be the most straight-forward approach to balancing the buckets.

joelochlann:
I'd suggest looking at the distribution of browsers in each test bucket to see whether, say, IE is absent from the less-favoured bucket - that would indicate a browser compatibility problem. (Also a quick IE test in BrowserStack maybe). It's so strongly skewed that I doubt it's due solely to users who have JS disabled

If we can't find the issue and, once the buckets are rebalanced, we need to continue the test in such a way that we won't start skewing in the other direction, we could perhaps have the form disabled & no list ids on page load, so only users for whom the test JS runs properly would actually be able to sign up

joelochlann (approved):

PR #14381, opened 2016-09-19
add encodings to in-body videos in a specific order

akash1810:
## What does this change?

Browsers attempt playback of video sources in order of declaration. VideoJS is loading metadata on page load to get duration information for the control bar. This involves a network request to the first playable source.

If you look at the network requests to this page https://www.theguardian.com/us-news/2016/sep/13/household-income-increase-poverty-us-census-report we can see 4.9MB webm file being requested on page load. This is the entire video being downloaded before user intention to play. This is expensive and wrong.

It looks like there is a general bug loading metadata for webm files - see here http://output.jsbin.com/juceda.

This PR always adds the mp4 encoding before the webm one, which means the article above has a smaller initial page weight. Smaller page weight = less data = happy readers.

Lighter pages.

Yes.
## Screenshots

Before
screen shot 2016-09-20 at 00 01 10

After
screen shot 2016-09-20 at 00 00 54
## Request for comment

@gidsg @michaelwmcnamara


common/app/views/support/cleaner/VideoEmbedCleaner.scala
```diff
@@ -111,9 +110,5 @@ case class VideoEmbedCleaner(article: Article) extends HtmlCleaner {
     document
   }
 
-  def getVideoAssets(id:String): Seq[VideoAsset] = article.elements.bodyVideos.filter(_.properties.id == id).flatMap(_.videos.videoAssets)
```
akash1810: these are no longer used

common/app/views/support/cleaner/VideoEmbedCleaner.scala
```diff
@@ -57,22 +57,21 @@ case class VideoEmbedCleaner(article: Article) extends HtmlCleaner {
...
-        element.append(sourceHTML)
-
         // add the poster url
         video.map(_.images).flatMap(Item640.bestFor).map(_.toString()).foreach { url =>
           element.attr("poster", url)
         }
 
-        findVideoApiElement(mediaId).foreach { videoElement =>
+        video.foreach { videoElement =>
+          videoElement.videos.encodings.map { encoding => {
```
akash1810: `encodings` are [sorted](https://github.com/guardian/frontend/blob/master/common/app/model/Element.scala#L91-L94)

TBonnin: What about adding a test to the `VideoMedia` model for instance to verify the expected encoding ordering? That would make it obvious that the ordering is actually set this way on purpose and modifying it by accident would break the test.

akash1810: @TBonnin a (slightly crude) [test](https://github.com/guardian/frontend/pull/14381/commits/ffe0439ba82bf64a5894df01ed020f7b57ed3cdd) has been added now.

TBonnin: I would have simply added an unit test for the `MediaVideo` class (to avoid the html parsing gymnastic) but yours does the job I guess.

TBonnin:
👍

PR #14382, opened 2016-09-20
Firfox fix heroic video

blongden73:
**Before**
picture 66

**After**
picture 67

@SiAdcock could you take a look...


SiAdcock:
👍

PR #14383, opened 2016-09-20
Removed unused contributions test

markjamesbutler:
## What does this change?

Removes an old ab test we are no longer using.

None

No


rich-nguyen:
👍

PR #14384, opened 2016-09-20
Displays saved crosswords on "saved for later" page

rjmk:
## What does this change?

Saved crosswords now appear in the "saved for later" list
## Request for comment

@guardian/dotcom-platform

### DIFF
--- a/identity/app/model/SaveForLaterPageData.scala
+++ b/identity/app/model/SaveForLaterPageData.scala
@@ -50,7 +50,8 @@ class SaveForLaterDataBuilder(contentApiClient: ContentApiClient, idUrlBuilder: 
         for {
           content <- Some(Content(result))
           faciaContent <- Some(FaciaContentConvert.contentToFaciaContent(result))
-          article <- articles.find( article => article.id == content.metadata.id)
+          // The `replace` is a bit of a hack -- it saves having to touch CAPI or pageId generation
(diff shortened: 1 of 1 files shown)


identity/app/model/SaveForLaterPageData.scala
```diff
@@ -50,7 +50,8 @@ class SaveForLaterDataBuilder(contentApiClient: ContentApiClient, idUrlBuilder:
         for {
           content <- Some(Content(result))
           faciaContent <- Some(FaciaContentConvert.contentToFaciaContent(result))
-          article <- articles.find( article => article.id == content.metadata.id)
+          // The `replace` is a bit of a hack -- it saves having to touch CAPI or pageId generation
+          article <- articles.find( article => article.id == content.metadata.id.replace("crosswords/", ""))
```
rjmk: I went through a few of our repos to try and find the ultimate source of this bug. I think it lies fairly deep, so thought this would be the least chaotic solution. Let me know if that's not the case

TBonnin:
Based on the description of the bug the crossword seems to be added to the list but cannot be accessed thereafter. Is it what's happening?

rjmk:
The crossword gets counted in "You have X saved items" but doesn't appear in the list. That's because the id is saved without the leading "crosswords/" from the slug, so when we try to match retrieved articles with the items that were saved for later, we skip the crosswords.

Gareth has suggested I try and resolve the missaved slug, so will close this PR for now

PR #14385, opened 2016-09-20
Promote comments in (p)react

piuccio:
## What does this change?

Remove control and split users 50-50 among react and preact

The initial test shows that external comments don't affect page load. As of this morning the complete ratio (number of people that stay long enough for the comments count to load) is:

| Variant | Completion |
| --- | --- |
| control | 61.61% |
| preact | 61.49% |
| react | 61.37% |

The test ran only on 5% so the results are not statistically significant and are subject to little variations but at least they show that there's no evident negative effect in rendering comments in (p)react.

This PR modifies the test to always use the external `discussion-frontend` and split users in two groups, react and preact. It should give statistical significance more easily.

Because of browsers constraints (you need `fetch` and `Promise`) some users will be `not-in-test`. This test should tell us how many they are and track the count over time.

No

### DIFF
--- a/common/app/conf/switches/ABTestSwitches.scala
+++ b/common/app/conf/switches/ABTestSwitches.scala
@@ -18,7 +18,7 @@ trait ABTestSwitches {
 
   Switch(
     ABTests,
-    "ab-discussion-external-frontend-count",
+    "ab-discussion-external-frontend-avatar",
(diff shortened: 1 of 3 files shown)


regiskuckaertz (approved):
LGTM

PR #14389, opened 2016-09-20
Update docs to allow successfully running `identity`

rjmk:
## What does this change?

Update the docs for the identity app. `frontend.properties` [is no longer used](https://github.com/guardian/frontend/blob/a61225df3b211b638aeef3d7bfd676d5456231ef/common/app/common/configuration.scala#L41-L47)

Devs get up and running more easily
## Request for comment

cc/ @guardian/dotcom-platform

### DIFF
--- a/identity/README.md
+++ b/identity/README.md
@@ -31,19 +31,6 @@ So the required steps are:
 
 * configure and run nginx (see `/nginx`)
 * update your hosts file (also described in `/nginx`)
-* make sure that your `~/.gu/frontend.conf` file contains
-
-
-
-```
-# ID
-id.apiRoot=https://id.code.dev-guardianapis.com
-id.apiClientToken=frontend-code-client-token
-```
-
-This configures the locally running identity module with the Identity API 
-on the CODE environment.
-
 * run the Identity subproject on port 9009
 
 With these in place, you'll be able to browse Identity on
@@ -57,6 +44,8 @@ following properties in `~/.gu/frontend.conf`.
 
 ```
 # ID
-id.apiRoot=https://idapi.thegulocal.com
-id.apiClientToken=frontend-dev-client-token
+devOverrides {
+  id.apiRoot=https://idapi.thegulocal.com
+  id.apiClientToken=frontend-dev-client-token
+}
 ```
--- a/identity/README.md
+++ b/identity/README.md
@@ -31,7 +31,7 @@ So the required steps are:
 
 * configure and run nginx (see `/nginx`)
 * update your hosts file (also described in `/nginx`)
-* make sure that your frontend.properties file contains
+* make sure that your `~/.gu/frontend.conf` file contains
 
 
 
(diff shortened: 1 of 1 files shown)


TBonnin:
👍  Thank you very much

TBonnin (approved):
👍  Thank you very much

identity/README.md
```diff
@@ -31,7 +31,7 @@ So the required steps are:
 
 * configure and run nginx (see `/nginx`)
 * update your hosts file (also described in `/nginx`)
-* make sure that your frontend.properties file contains
+* make sure that your `~/.gu/frontend.conf` file contains
```
gtrufitt: Minor, but I think we would need to quote the strings below now?

gtrufitt (approved):
👍 Other than minor comment

identity/README.md
TBonnin: Indeed I talked to fast. Do you need to wrap the dev param into 
```
devOverrides {
   id.apiRoot=https://id.code.dev-guardianapis.com
   id.apiClientToken=frontend-code-client-token		 
}
```

identity/README.md
```diff
@@ -31,7 +31,7 @@ So the required steps are:
...
```
johnduffell: no this should be removed completely, if it doesn't work with the default settings they should be added to the master file in s3.  The local file is only for user specific overrides.

identity/README.md
johnduffell: talked in person, but for future reference, no because it's not an override, it's just the normal parameter that everyone needs, so it should be in the config file that's in S3

identity/README.md
TBonnin: You are right. I talked to fast twice. 😖 

Whatever @johnduffell said 🎯

rjmk:
Having done some experimenting, looks like everything's in place. Will just update the docs by removing this section :tada:

gtrufitt:
💣 💥

johnduffell:
sorry, I may have left the team, but I can still cause chaos from outside 😆 🚀

rjmk:
@johnduffell Chaos that removes unnecessary config steps can only be good!

identity/README.md
```diff
@@ -53,10 +40,12 @@ With these in place, you'll be able to browse Identity on
 ## Configuration of local Identity API
 
 The Identity site can be configured to use the local Identity API with the
-following properties in `frontend.properties`.
+following properties in `~/.gu/frontend.conf`.
 
 ```
 # ID
```
johnduffell: minor suggestion, but I'd put the id comment inside devOverrides, as the person might have other overrides that apply to other things.

johnduffell:
thanks! 👍

identity/README.md
rjmk: :zap:

paperboyo:
> sorry, I may have left the team

One _never_ leaves the team, @johnduffell. Don’t fool yourself...

PR #14390, opened 2016-09-20
Hosted pages - choose contrasting colour using HSL

lps88:
## What does this change?

Instead of using the B (brightness) of HSB colour space to choose a contrasting colour to the brand colour for each Hosted By campaign, calculate the lightness (using HSL colour space) and use that instead.

image
### Interesting links on the topic:
- http://colorizer.org/
- https://en.wikipedia.org/wiki/HSL_and_HSV
- http://www.niwa.nu/2013/05/math-behind-colorspace-conversions-rgb-hsl/

We had to hard code some exceptions to the previous rule, it categorised Visit Britain and Chester Zoo reds as very bright, which forced the text colour to be black, which looked bad! Now the same algorithm works for all the current campaigns.

image
## Request for comment

@guardian/labs-beta

### DIFF
--- a/common/app/common/commercial/hosted/HostedPage.scala
+++ b/common/app/common/commercial/hosted/HostedPage.scala
@@ -51,13 +51,12 @@ case class FontColour(brandColour: String) {
     val hexColour = brandColour.stripPrefix("#")
     val rgb = Integer.parseInt(hexColour, 16)
     val c = new Color(rgb)
-    val hsb = Color.RGBtoHSB(c.getRed, c.getGreen, c.getBlue, null)
-    val brightness = hsb(2)
-    if(brandColour == "#E31B22" || brandColour == "#E41F13") {
-      false
-    } else {
-      brightness > 0.5
-    }
+    // the conversion in java.awt.Color uses HSB colour space, whereas we want HSL here
+    // see http://www.niwa.nu/2013/05/math-behind-colorspace-conversions-rgb-hsl/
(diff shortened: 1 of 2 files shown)


common/app/common/commercial/hosted/HostedPage.scala
```diff
@@ -51,13 +51,12 @@ case class FontColour(brandColour: String) {
...
-    if(brandColour == "#E31B22" || brandColour == "#E41F13") {
-      false
-    } else {
-      brightness > 0.5
-    }
+    // the conversion in java.awt.Color uses HSB colour space, whereas we want HSL here
+    // see http://www.niwa.nu/2013/05/math-behind-colorspace-conversions-rgb-hsl/
+    val min: Float = Math.min(Math.min(c.getRed, c.getGreen), c.getBlue)
+    val max: Float = Math.max(Math.max(c.getRed, c.getGreen), c.getBlue)
+    val lightness = (min + max) / 510
```
Calanthe: why 510? What does it mean?

Calanthe:
Thanks for fixing that! We had some issues with those colours with @kelvin-chappell.

common/app/common/commercial/hosted/HostedPage.scala
lps88: You convert the colours to be in the range 0..1 instead of 0..255 and then find the average. So divide by 255 and then also by 2. It's from here: http://www.niwa.nu/2013/05/math-behind-colorspace-conversions-rgb-hsl

kelvin-chappell:
Nice work!

PR #14391, opened 2016-09-20
Make Revealer full-width in articles

regiskuckaertz:
Although it is already full-width on fronts, revealer are not f-w on articles:

picture 4

Fixed:

picture 5


rich-nguyen:
👍

PR #14393, opened 2016-09-20
[not ready to merge] Always redirect to https for redirects defined in dynamo

mchv:
## What does this change?

Ensure that all redirection defined in dynamo table (so not the one in `fastly`) are done with `https` scheme.

This prevent additional roundtrips.

No

### DIFF
--- a/archive/app/controllers/ArchiveController.scala
+++ b/archive/app/controllers/ArchiveController.scala
@@ -64,7 +64,7 @@ class ArchiveController(dynamoDB: DynamoDB) extends Controller with Logging with
         Some(path)
       case _ => None
     }
-    s"http://${normalised.getOrElse(path)}"
+    s"https://${normalised.getOrElse(path)}"
   }
 
   def linksToItself(path: String, destination: String): Boolean = path match {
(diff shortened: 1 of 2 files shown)


archive/app/controllers/ArchiveController.scala
```diff
@@ -64,7 +64,7 @@ class ArchiveController(dynamoDB: DynamoDB) extends Controller with Logging with
         Some(path)
       case _ => None
     }
-    s"http://${normalised.getOrElse(path)}"
+    s"https://${normalised.getOrElse(path)}"
```
TBonnin: Does this mean we input and store urls with the scheme but actually ignores when returning the actual redirect? If yes this is quite confusing

mchv: Absolutely. I agree this is confusing and it looks like a side effect of the `normalisation`.
This probably need a bit of tidy up or even a redesign because the redirect table seems to be used for a lot of different concerns :/

TBonnin: Indeed. You are right, it might require some refactoring. For now I give you a 👍 💃

gklopper: I 'think' the above code is just to make it work on different environments. LinkTo will achieve the same thing, but be a bit cleaner maybe

jfsoul:
@mchv do you mind if I take this work off your hands?  Currently looking into making some improvements for the redirect tool and this would definitely be one of them.

mchv:
@jfsoul go ahead. I never merged as I was unsure.

dominickendrick:
@jfsoul says this can be closed

jfsoul:
I will make sure these changes are included in a future PR (imminent)

PR #14394, opened 2016-09-20
Remove cssClass field from HostedCampaign

lps88:
## What does this change?

Remove a field which was used to set a campaign-specific css class on hosted pages.

All the styling should be the same across the campaigns. The brand-specific colours are now driven from capi, so we don't need it any more.
## Request for comment

@guardian/labs-beta


kelvin-chappell:
👍

PR #14395, opened 2016-09-20
Define size mappings for out of page slots too

regiskuckaertz:
Although I added the 2x2 size to the pageskin slot, the information was not carried over to DFP. That is because out of page slots don't have a size mapping:

picture 2

Except that now they do:

picture 1


rich-nguyen:
Nice tidy up 👍

PR #14397, opened 2016-09-20
Remove StaticBadge and containerBrandingFromCapi switches

regiskuckaertz:
cc @kelvin-chappell


applications/app/views/fragments/galleryHeader.scala.html
```diff
@@ -9,7 +8,7 @@
     <h1 class="is-hidden" itemprop="headline">@Html(gallery.item.trail.headline)</h1>
 
     <div class="gs-container">
-        @if(gallery.showBadge || staticBadgesSwitch.isSwitchedOn) {
+        @if(gallery.showBadge) {
```
kelvin-chappell: I think this condition can be removed as it was having no effect when switch was on.

common/app/model/trails.scala
```diff
@@ -94,9 +93,8 @@ final case class Trail (
   lazy val showByline: Boolean = tags.isComment
 
   def shouldHidePublicationDate(implicit request: RequestHeader): Boolean = {
-    lazy val isPaidContentInDfp = staticBadgesSwitch.isSwitchedOff && commercial.isAdvertisementFeature
-    lazy val isPaidContentInCapi = staticBadgesSwitch.isSwitchedOn && BrandHunter.isPaidContent(this, Edition(request))
-    (isPaidContentInDfp || isPaidContentInCapi) && webPublicationDate.isOlderThan(2.weeks)
+    lazy val isPaidContent = BrandHunter.isPaidContent(this, Edition(request))
```
kelvin-chappell: This no longer needs to be lazy.

common/app/views/fragments/headDefault.scala.html
```diff
@@ -13,7 +12,7 @@ <h1 class="content__headline js-score" itemprop="headline">@Html(item.trail.head
                 @fragments.standfirst(item)
             }
 
-            @if(showBadge || staticBadgesSwitch.isSwitchedOn) {
+            @if(showBadge) {
```
kelvin-chappell: I think this `showBadge` param can go.

kelvin-chappell (approved):
Looks great!  This really simplifies the code.  👍 
[Just minor comments that could be fixed in this PR or later.]

PR #14398, opened 2016-09-20
New Header: search should search

NataliaLKB:
## What does this change?

Making search work on the new header. 

It works like this:

search-example

Without javascript it still searches the query, just not our site. 

People can find things via search, even though it takes you off our page, and onto Google.

Nope!
## Request for comment

cc @SiAdcock @stephanfowler


TBonnin:
Is there a way to come back once the reader is in Google interface?

NataliaLKB:
@TBonnin there are two.

1. Since all the links are guardian links if they click on one they will come back.
2. The back button

Other than that, none so far. 😞  Maybe @stephanfowler has some ideas?

NataliaLKB:
> In the screenshot above, there is a flash of a suggestion: do we have control over it? can we style it? Or is it browser/OS?

@paperboyo it is the browser. And it might be because I am actually on a desktop in that example

SiAdcock:
Code looks great 👍

PR #14400, opened 2016-09-20
Create a Soulmates JSON API endpoint

JonNorman:
The new DFP Native templates need to be able to request a list of Soulmates to render the component.

For now the old fragment-serving route is being left in; once the native template are reigning supreme, these can be ripped out.

@guardian/commercial-dev


commercial/app/controllers/commercial/SoulmatesController.scala
```diff
@@ -3,37 +3,42 @@ package controllers.commercial
...
 
-    sample.toList match {
+  def renderSoulmates(groupName: String) = Action { implicit request =>
+    soulmatesSample(groupName).toList match {
       case Nil => NoCache(jsonFormat.nilResult.result)
       case soulmates => Cached(componentMaxAge) {
         val clickMacro = request.getParameter("clickMacro")
         val omnitureId = request.getParameter("omnitureId")
-        jsonFormat.result(view(soulmates, omnitureId, clickMacro))
+        jsonFormat.result(views.html.soulmates.soulmates(soulmates, omnitureId, clickMacro))
```
TBonnin: Should caching headers be set here?

commercial/app/controllers/commercial/SoulmatesController.scala
```diff
@@ -3,37 +3,42 @@ package controllers.commercial
...
       }
     }
   }
 
-  def renderSoulmates(groupName: String): Action[AnyContent] = Action { implicit request =>
-    result(groupName, views.html.soulmates.soulmates(_, _, _))
+  def getSoulmates(groupName: String) = Action { implicit request =>
+    val sample: Seq[Member] = soulmatesSample(groupName)
+    val json = Json.toJson(sample)
+    Ok(json)
```
TBonnin: Same question about caching headers?

regiskuckaertz (approved):
Aside from @TBonnin 's q's, this is LGTM

commercial/app/controllers/commercial/SoulmatesController.scala
JonNorman: There are cache headers added [here](https://github.com/guardian/frontend/pull/14400#discussion-diff-79650215R31) which I think are sufficient. Thoughts?

commercial/app/controllers/commercial/SoulmatesController.scala
JonNorman: I've now added a 60 second cache to this response; thanks for explaining the mechanism 😄

JonNorman:
Hmm, I pushed too soon; still need to alter the `Result` type before this will work.

commercial/app/controllers/commercial/SoulmatesController.scala
TBonnin: my bad. I missed it

commercial/app/controllers/commercial/SoulmatesController.scala
```diff
@@ -3,37 +3,44 @@ package controllers.commercial
...
       }
     }
   }
 
-  def renderSoulmates(groupName: String): Action[AnyContent] = Action { implicit request =>
-    result(groupName, views.html.soulmates.soulmates(_, _, _))
+  def getSoulmates(groupName: String) = Action { implicit request =>
+
+    val json = Json.toJson(soulmatesSample(groupName))
+    Cached.explicitlyCache(60){
```
TBonnin: I don't think you need `explicitlyCache` here. It is usually used to force adding cache headers, even though the result is usually not cached.

JonNorman: Okay, in order to cache the result, it needs to be a `CacheableResult`, so would the following work, or is there a better/standard way of otherwise converting a `Result` to a `CacheableResult`?
```
Cached(60.seconds){
      WithoutRevalidationResult(Ok(json))
    }
```
I couldn't find an example where we are already serving cached JSON responses but perhaps my search-fu is just poor this morning. Know of anywhere in the code base we are doing that?

annebyrne:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;3xatfI3F&#x2F;147-migrate-capi-single"> Migrate CAPI single</a>

TBonnin:
👍
Don't forget to get rid of `forJsValue` in a subsequent PR

JonNorman:
Yep, got it right [here](https://github.com/guardian/frontend/pull/14407) :)

PR #14401, opened 2016-09-20
Fix radiator graphs (Fastly, googlebot-errors)

TBonnin:
## What does this change?

A recent change of configuration class modifed the current stage case ('prod' -> 'PROD')
This patch remove the hardcoded 'prod' and replaces them with the proper
configuraton stage value

Graph in admin are working 📈 
## Request for comment


gtrufitt (approved):
👍 💯

NataliaLKB (approved):

PR #14402, opened 2016-09-20
If markup is pasted into an immersive interactive, don't iframe it

sammorrisdesign-zz:
## What does this change?

This makes it so immersive interactives that are pasted markup in composer, aren't iframed. I've put a little switch around it for safety, but haven't found any issues with existing interactives. This is the start of improving the performance of our interactives for our developers and readers.

It will mean that the work by our interactive teams is visible by Google and it removes the lengthy 2nd request – where we see the most drop off on mobile.

Nope, no changes to them. any interactives supplied in this way are still iframed.
## Request for comment

@clloyd @gtrufitt @stephanfowler


paperboyo:
Woah! Does it mean that these interactives will (or could) use dotcom’s fonts instead of loading their own?

gtrufitt:
@paperboyo It means they can use Dotcoms everything - they become first-class content citizens, having their HTML a part of the initial page response

paperboyo:
Finally! Bless ’em. And every single one of ya.
💎 !

stephanfowler:
Very nice proof of concept. @clloyd : the outcome of this is broadly the same as what an interactive atom would achieve, right?

wpf500:
I'm don't think this is the same as an interactive atom. With this we'd have to copy & paste an interactive's HTML every time we want to make an update, and more worryingly, so would subs/editors/etc. who want to fix copy errors etc. which feels very breakable.

Maybe I'm misunderstanding the implications, but I'm a bit concerned about this change, can we have a chat about it first? Lots of problems spring to mind, apps being the biggest one.

Jholder112233:
Apart from the obvious complications of iframing content on one platform and not on another, I've just tested this and currently the app doesn't render anything for immersive interactives if you just supply raw html as opposed to a boot.js.

sammorrisdesign:
I've had a chat with @wpf500 and he's back on board.

Only issue is making sure this works in apps too. Weirdly this is something Composer has always supported but neither platform has in the immersive interactive template...

common/app/model/content.scala
```diff
@@ -840,6 +840,16 @@ final case class Interactive(
     }
   }
 
+  lazy val hasSrcdoc = {
+    val srcdoc = Jsoup.parseBodyFragment(fields.body).getElementsByTag("iframe").attr("srcdoc")
+
+    if (srcdoc.length > 0) {
```
sammorrisdesign: and you told me you didn't know Scala...

gtrufitt:
I'm going to pull in @guardian/content-platforms and @guardian/dotcom-apps-collab to give us some advice on where we should be doing this for Apps (MAPI?)

common/app/model/content.scala
gtrufitt: This bit is basically jQuery ;)

mchv:
This should not be done in Content API as this is a purely rendering concern

GHaberis:
@Jholder112233 Would you be able to show me an immersive interactive with raw html in the srcdoc attribute within the app? I've not found an example of how it does (or doesn't) render when coming via MAPI, but I have mocked an example and the html was rendering.

GHaberis:
OK regarding apps, we could ask MAPI to identify when we have an immersive article with an embedded-iframe within the body, if the iframe has a srcdoc attribute then they'll need to pull the markup out of the srcdoc attribute and replace the body value with that markup.

I'd like clarification around what scenarios this would be used. Would the markup in the srcdoc contain a script which strips down the template and completely replaces it with something else? Like the interactives which currently break out of the iframe? Or would the markup sit embedded within template's html? If so we'd need to think about the risk of style leakage.

Also the benefit of the interactive not loading in it's own fonts but using frontend's fonts has been mentioned.  However we'd need to look at whether this would work across platforms. eg. are the same font-families available across frontend/apps.

sammorrisdesign:
@GHaberis There isn't an example of an immersive interactive with raw html, because despite composer supporting it, frontend and the apps never have.

It could contain a script but more likely it will just sit embedded within the template's html (which for immersive is basically no html). Yes there is a risk of style leakage but immersive interactive already suffer from the issue when they inject html into the page. At least here, we can guarantee the user (and google) will see something on load.

gtrufitt (approved):
LGTM. If everyone is happy and had their concerns allayed, sounds like MAPI are on board (for now these will be iframed), if we get this merged we can take this forward. 

It gives us a good POC to start thinking about what works well and what doesn't work well when we come to do atoms proper.

🎈

PR #14403, opened 2016-09-21
Remove size argument in defineOutOfPageSlot

regiskuckaertz:
It turns out `defineOutOfPageSlot`, unlike `defineSlot`, does not accept a `size` parameter.


rich-nguyen:
👍

PR #14405, opened 2016-09-21
Momentous headline leading change

zeftilldeath:
Galleries and immersive templates should both have the same line-height. Now they do. 

tight


blongden73:
👍

PR #14406, opened 2016-09-21
Hot fix the video carousel button for test

mr-mr:
Fixes the carousel duration button as it was missing in scala land

image


akash1810 (approved):
👍

PR #14407, opened 2016-09-21
Remove redundant caching method

JonNorman:
A quick clean up from the work of #14400.

Most of this PR is minor edits. They all stem from the changes I made [here](https://github.com/guardian/frontend/pull/14407/files#diff-e677de1846e5477a72e51aa2889bc3bbR18). By creating a generic `JsonComponent` constructor, the many various calls to its constructor can be simplified.

I haven't exhaustively refactored this; there is doubtless a nicer way to make the calls more consistent and remove a bunch of the overloaded constructors. But this was just a quick edit off the back of my previous PR.

@TBonnin


common/app/common/JsonComponent.scala
```diff
@@ -12,6 +12,12 @@ import play.api.http.ContentTypes._
 
 object JsonComponent extends Results with implicits.Requests {
 
+  def withRefreshStatus(json: JsValue): JsValue =
+    json.as[JsObject] + ("refreshStatus" -> toJson(AutoRefreshSwitch.isSwitchedOn))
+
+  def apply[A](value: A)(implicit request: RequestHeader, writes:Writes[A]) =
+    resultFor(request, Json.stringify(withRefreshStatus(toJson(value))))
```
TBonnin: The problem with calling `withRefreshStatus` (and therefore value.as[JsObject]) is that it can blow up at runtime because you can have a value from a type with an implicit `Writes` but that cannot be transformed into a `JsObject` (ex: a `JsString`)

TBonnin (changes requested):

common/app/common/JsonComponent.scala
JonNorman: Good catch! I'm changing this to now check to see whether `value` is of type `JsObject` and, if so, the `refreshStatus` will be added to it, otherwise it won't be. I believe that this then leaves all behaviour unchanged.

JonNorman (commented):
I think I've covered all the bases now. Any calls to `JsonComponent` that are explicitly passed as `JsObject` have been restored and the `apply` method deals with the addition of the `RefreshStatus`.

I replaced `InstanceOf` with the JSON `as` methods where I saw it; neither are ideal but given that this is a refactor, I'm not going to start making type checks more robust all over the place.

common/test/common/JsonComponentTest.scala
```diff
@@ -55,7 +57,7 @@ class JsonComponentTest extends FlatSpec with Matchers with ExecutionContexts {
 
     val result = Future {
       val request = FakeRequest("GET", "http://foo.bar.com/data.json")
-      JsonComponent(obj("name" -> "foo"))(request).result
+      JsonComponent(obj("name" -> "foo"))(request, play.api.libs.json.Writes.JsValueWrites).result
```
TBonnin: Do you still need to specific the `JsValueWrites`? Also you might need `WithRefreshStatus()` here

TBonnin (changes requested):

common/test/common/JsonComponentTest.scala
JonNorman: I either need to provide the `Writes` explicitly or make the `request` implicit and not pass it explicitly. I'll opt for the latter as it is more representative of how it is used and have added the `withRefreshStatus`.

JonNorman:
@TBonnin, reckon these are good to go? I'm happy with them as an initial effort to simplify JSON responding.

TBonnin (approved):

PR #14408, opened 2016-09-21
background colour added, illustration extended

superfrank:
White BG is displaying so image looks broken at largest breakpoints

image

@katebee


katebee:
Can't spot the change in the SVG 😂 😭  Do you have an 'after' pic?

superfrank:
It's exactly the same! Just twice the width so SVG is now 5200, rather than 2600.

superfrank:
image

katebee (approved):
Woop! 👍

PR #14409, opened 2016-09-21
Add dimensions to logo amp-img elements

kelvin-chappell:
This change adds width and height parameters to amp-img element for branded logos.


Calanthe:
👍

PR #14410, opened 2016-09-21
Promote comments with a sticky banner

piuccio:
## What does this change?

Add a new AB test to measure the effect of a sticky banner prompting to join the discussion in the comments.

I'm trying not to be too annoying, so the banner is only visible while the main content is in view. (it doesn't appear immediately on page load and goes away when you scroll to related content, outbrain, comments, footer and so on).

Slightly contrary to what I just said, the banner is not dismissable, it comes in and goes away at its own will. Clicking on it takes you down to the comments without expanding it.

The related PR on discussion frontend is guardian/discussion-frontend#10

People should interact more with comments (go to next page, toggle more comments, change sort order, stuff like that). We don't care much about posting (at least in this test).

No

## Screenshots

out

### DIFF
--- a/static/src/javascripts/projects/common/modules/discussion/discussion-frontend.js
+++ b/static/src/javascripts/projects/common/modules/discussion/discussion-frontend.js
@@ -29,6 +29,13 @@ define([
             //     json: fetchJson
             // };
 
+            // Show the sticky banner only if we are in the AB test and other banners are not visible
+            var noOtherBanners = !otherBannersVisible();
+            opts.featureStickyBanner = ab.isInVariant('DiscussionPromoteComments', 'bottom-banner') && noOtherBanners;
+            opts.featureTopBanner = ab.isInVariant('DiscussionPromoteComments', 'top-banner') && noOtherBanners;
+            opts.featureStickyBadge = ab.isInVariant('DiscussionPromoteComments', 'bubble') && noOtherBanners;
+            opts.featureStickyBannerDismissable = true;
+
             frontend(opts)
             .then(function (emitter) {
                 emitter.on('error', function (feature, error) {
@@ -47,6 +54,20 @@ define([
         }, function (error) {
             reportError(error, { feature: 'discussion' });
         });
+    }
+
+    function otherBannersVisible () {
+        var siteMessage = document.getElementsByClassName('js-site-message');
+        if (siteMessage.length && !siteMessage[0].classList.contains('is-hidden')) {
+            // Contribution banner is visible
+            return true;
+        }
+        var breakingNews = document.getElementsByClassName('js-breaking-news-placeholder');
+        if (breakingNews.length && !breakingNews[0].classList.contains('breaking-news--hidden')) {
+            // Breaking news alert is visible
+            return true;
+        }
+        return false;
     }
 
     return {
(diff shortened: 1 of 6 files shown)


piuccio:
@paperboyo 😧

paperboyo:
giphy

ScottPainterGNM:
Just out of curiosity is there a particular reason as to why it's non dismissible? Could that be an alternative bucket of the test perhaps to see how many people do dismiss instead of following the banners suggestion?

piuccio:
Mostly because it looks better without the close icon and I guess we want people to click on it really badly.

piuccio:
I've update the PR to include two more variants

top banner 
screen shot 2016-09-22 at 12 57 55

floating bubble
screen shot 2016-09-23 at 15 32 51

The behaviour is the same, they appear when you're reading the content and disappear when you scroll past

katebee:
As much as we want people to interact with it, I really feel it should be dismissible. Most of our banners look cleaner without the close button, however, is it a bit of a dark pattern to say "we really want you to click this, so we wont let you hide it". 😱

How will we determine the number of clicks that were desperation ones; where users hoped that clicking on the banner would remove it?

paperboyo:
> How will we determine the number of clicks that were desperation ones; where users hoped that clicking on the banner would remove it?

We can just pretend all of these were excited about being able to go straight to comments fulfilling our objectives, while secretly knowing that ~97% were just really hoping to dismiss the bloody thing 😜

piuccio:
Personally I'm happy if people comment out of desperation. :trollface: 

Let me go back to the drawing board.

piuccio:
screen shot 2016-09-26 at 12 02 49

close button

ScottPainterGNM:
image

static/src/javascripts/projects/common/modules/discussion/discussion-frontend.js
```diff
@@ -49,6 +56,20 @@ define([
...
+            return true;
+        }
+        var breakingNews = document.getElementsByClassName('js-breaking-news-placeholder');
+        if (breakingNews.length && !breakingNews[0].classList.contains('breaking-news--hidden')) {
+            // Breaking news alert is visible
+            return true;
+        }
+        return false;
+    }
+
```
katebee: There are a couple of mediator events (one on the membership banner, one on the adblock banner) that may be worth waiting for, so that this banner doesn't beat them to the slot (they both have async requests).

How does it behave with the Weekend Reading banner (#14218) and the US survey banner (#14255)? Does it beat them to the slot?

piuccio: Fantastic comment! You've finally pointed out the madness of banners. So the logic to detect them is fine because they're all using `ui/message` but the race condition remains.

I claim that this bit of code is the last to run, because it has to asynchronously download react and make an API request before it can run, but I'll change the code to listen to those events.

Thanks

piuccio: The plot thickens

`banner-message:complete` is emitted (or supposed to be emitted) in two places:

1. `donot-use-adblock`
2. `membership-engagement-banner`

However,  (1) is loaded by `enhanced/main` before asynchronously requiring `enhanced/trail`. So when I initialize `discussion-frontend` I've missed already the first `emit`.
(2) is only fired if you're in one of the variants of an AB test. Grantend that the test has 100% audience, but if you're `notintest` the event is not fired.

So listening to two events that may or may not be fired seems like a very brittle solution.
The solution maybe should be to refactor that logic and store in `ui/message` information about pending scripts that are trying to use the slot.

Another approach is to check there are no banners every time I try to render one. The performance cost is higher and given the previous consideration it seems unnecessary.

previous consideration: Discussion frontend is loaded async by `trail` which is loaded async by `enhanced` and ask (p)react to render a banner after making an API request. Most likely it'll be the last in the queue.

It's a short lived AB test (my last famous word), so if you don't mind I'd merge it like this.

PR #14411, opened 2016-09-21
Remove commercial container types

kelvin-chappell:
They are now redundant because standard containers can be branded.


Calanthe:
Don't forget to remove them also fro the test page http://m.code.dev-theguardian.com/commercial/test-page

kelvin-chappell:
@Calanthe I don't think there are any in that test page.  They should all have been converted to standard containers now.

PR #14415, opened 2016-09-21
Only enhance menu checkboxes if menus not open

SiAdcock:
## What does this change?

Users can interact with the new header menu before JavaScript has loaded. It uses a CSS solution that tracks the state of a hidden checkbox.

If the user opens the menu before the JavaScript is loaded, then when the JavaScript loads, the checkbox is replaced with a button. As a result, the menu automatically closes itself as soon as the JavaScript loads, which is an annoying experience.

menu-close-enhancement

This change ensures the JavaScript enhancements are not applied while the menu is open. The enhancements are applied later, if the user closes the menu.

This functionality is applied to the main menu and the edition picker.

No
## Request for comment

@NataliaLKB @gtrufitt

### DIFF
--- a/static/src/javascripts/projects/common/modules/navigation/newHeaderNavigation.js
+++ b/static/src/javascripts/projects/common/modules/navigation/newHeaderNavigation.js
@@ -65,12 +65,11 @@ define([
     }
 
     function enhanceCheckboxesToButtons() {
-        var checkboxes = [
-            qwery('#main-menu-toggle')[0],
-            qwery('#edition-picker')[0]
-        ];
+        var checkboxIds = ['main-menu-toggle', 'edition-picker'];
 
-        checkboxes.forEach(function (checkbox) {
+        checkboxIds.forEach(function (checkboxId) {
+            var checkbox = document.getElementById(checkboxId);
+
             if (!checkbox) {
                 return;
             }
@@ -87,7 +86,7 @@ define([
 
     function veggieBurgerClickHandler(event) {
         var button = event.target;
-        var mainMenu = qwery('#main-menu')[0];
+        var mainMenu = document.getElementById('main-menu');
         var veggieBurgerLink = qwery('.js-change-link')[0];
 
         function menuIsOpen() {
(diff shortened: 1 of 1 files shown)


static/src/javascripts/projects/common/modules/navigation/newHeaderNavigation.js
```diff
@@ -3,18 +3,48 @@ define([
     'fastdom',
     'common/modules/navigation/edition-picker',
     'common/modules/navigation/editionalise-menu'
-], function (
-    qwery,
-    fastdom,
-    editionPicker,
-    editionaliseMenu
-) {
+], function (qwery,
```
NataliaLKB: IntelliJ strikes again! This indentation should be fixed to be the same as the other files in our codebase

static/src/javascripts/projects/common/modules/navigation/newHeaderNavigation.js
```diff
@@ -32,33 +62,24 @@ define([
...
-                button.setAttribute('aria-expanded', 'false');
-
-                return button;
-            });
-            fastdom.write(function () {
-                buttons.forEach(function (button, index) {
-                    var checkbox = checkboxes[index];
-                    var eventHandler = buttonClickHandlers[button.id];
+    function enhanceCheckboxesToButtons() {
+        var toggles = [
```
NataliaLKB: Nit (you don't have to listen to this): could we rename this variable `checkboxes`. As I am reading this code I keep forgetting what toggles are.

static/src/javascripts/projects/common/modules/navigation/newHeaderNavigation.js
```diff
@@ -32,33 +64,24 @@ define([
...
-
-                return button;
-            });
-            fastdom.write(function () {
-                buttons.forEach(function (button, index) {
-                    var checkbox = checkboxes[index];
-                    var eventHandler = buttonClickHandlers[button.id];
+    function enhanceCheckboxesToButtons() {
+        var checkboxes = [
+            qwery('#main-menu-toggle')[0],
```
gtrufitt: Might as well use `getElementById` here?

gtrufitt (approved):
👍

NataliaLKB (approved):
👍

PR #14418, opened 2016-09-21
Update email listings, add teasers and signupPages

katebee:
## What does this change?
- Use parameter naming for the various params that get passed into `EmailSubscription`
- Removes the unused `popularity` parameter from the class and each email listing
- Adds new `teaser` (intended to include shorter copy), `signUpPage` and `subheading` parameters, that will support upcoming work
- Switch all `exampleUrl` to use HTTPS 🌟 
- Move Weekend Reading and Long Read into their own little `Feature` huddle - they still get appended into the News dropdown on the Email Prefs page
- Copy and naming changes based on most recent documentation shared by Editorial ❤️ 
- Add the new Documentaries Update email to the emails list (so users can un/subscribe)
- ⭐ NAMED PARAMETERS ⭐ 
- 🍘 SANITY AND CLARITY 🍘
- 🔐 HTTPS 🔐

Apart from the above, this work precedes a larger project that will be making further use of the list from EmailSubscriptions.

Nope
## Screenshots

picture 256
#### Optional subheading on desktop:

new subheading on desktop
#### Optional subheading on mobile:

new subheading on mobile
## Request for comment

@guardian/guardian-frontend: 
A lot of copy update here! However, descriptions and teasers are ctrl+c, ctrl+v from editorial documentation so should be okay.
**Anything glaringly wrong with the Scala, or the Email Prefs page?**


identity/app/model/EmailSubscriptions.scala
```diff
@@ -3,13 +3,15 @@ package model
 case class EmailSubscriptions(subscriptions: List[EmailSubscription])
 case class EmailSubscription(
   name: String,
+  subheading: Option[String] = None,
```
janua: Would it be possible to put this at the end of the parameter list with the others or remove the default and fill it out in each case?
I would try and stick to defaults at the end of the parameter list of case classes as I've seen odd behaviour with inference/macros when done like this.

identity/app/views/profile/emailPrefs.scala.html
```diff
@@ -12,7 +12,9 @@
                 "email-subscription--subscribed" -> subscription.subscribedTo,
                 "email-subscription--first" -> row.isFirst))">
                 @if(subscription.subscribedTo){<input type="hidden" name="emailSubscription[]" value="@subscription.listId" />}
-                <div class="email-subscription__name">@subscription.name</div>
+                <div class="email-subscription__name">@subscription.name
+                    @if(subscription.subheading.isDefined){<br>@subscription.subheading}
```
janua: You could just `map` this to add the `<br>`.

janua: (If you wanted) :smiley:

janua:
LGTM :+1:

identity/app/views/profile/emailPrefs.scala.html
dominickendrick: I would also suggest not using a `<br>` tag if its just for formatting reasons. You could change the markup so `subscription.subheading` has its own tag (maybe use a `h2`, and `h3`)

identity/app/views/profile/emailPrefs.scala.html
```diff
@@ -12,7 +12,11 @@
                 "email-subscription--subscribed" -> subscription.subscribedTo,
                 "email-subscription--first" -> row.isFirst))">
                 @if(subscription.subscribedTo){<input type="hidden" name="emailSubscription[]" value="@subscription.listId" />}
-                <div class="email-subscription__name">@subscription.name</div>
+                <div class="email-subscription__name">@subscription.name
+                    @if(subscription.subheading.isDefined){
+                        <div class="email-subscription__name">@subscription.subheading</div>
```
dominickendrick: Hmm, both of these should really be `inline` elements, not `block` level, and should have different class names, maybe 
```
<h2 class="email-subscription__name">@subscription.name</h2>

@if(subscription.subheading.isDefined){
  <h3 class="email-subscription__subheading">@subscription.subheading</h3>
}
```

identity/app/model/EmailSubscriptions.scala
```diff
@@ -5,11 +5,13 @@ case class EmailSubscription(
   name: String,
   theme: String,
   about: String,
+  teaser: String,
   description: String,
   frequency: String,
   listId: String,
-  popularity: Int = 0,
   subscribedTo: Boolean = false,
+  subheading: Option[String] = None,
```
janua: :heart:

identity/app/views/profile/emailPrefs.scala.html
katebee: Thank you, that looks much nicer! I'm keeping the div as a wrapper to make a 'column' (to the left of the description) for the name and subheading:

picture 260

PR #14419, opened 2016-09-22
Send parent origin along with iframe id to native ads

regiskuckaertz:
Currently, if we want to test native ads in a local build, we have to resort to a creative variable (`DevMode`) and manually switch its value when we go from local to dev, to prod. This is highly inefficient and also polluting the creative schema—besides, we don't want people from merchandising, adops or richmedia to accidentally switch it on.

The solution: because of security constraints, it is impossible to know the origin of the parent frame. That's ok, we'll send this information along with the iframe ID when the ad has loaded.


JonNorman:
:+1:

PR #14420, opened 2016-09-22
Editionalise menu without JavaScript

SiAdcock:
## What does this change?

Depending on which edition the user has selected, we present different links within the menu. By default, we show links for the UK edition and then, if the user is looking at a non-UK edition, we unhide edition-specific links on the client using JavaScript.

There is a short delay between the initial render of the default UK links and the update to non-UK links as the JavaScript downloads and is executed on the client.

With this change, the logic to determine which links to show happens on the server, so the user instantly sees links relevant to the selected edition.
## Request for comment

@NataliaLKB @gtrufitt @TBonnin


SiAdcock:
Nope

common/app/views/fragments/nav/newHeaderMenu.scala.html
```diff
@@ -17,7 +17,7 @@
                     navigation-group--@{edition.id.toLowerCase}
                     js-editionlise-secondary-nav"
                     data-edition="@{edition.id.toLowerCase}"
-                    @if(edition.id.toLowerCase != "uk") { hidden }> @* Our default edition is UK *@
+                    @if(edition.id.toLowerCase != Edition(request).id.toLowerCase) { hidden }>
```
TBonnin: I thought the new header was not editionalized at all server-side. Did I understand it wrongly?

SiAdcock: Yeah, my bad, just been schooled by @NataliaLKB. I'll close this!

PR #14421, opened 2016-09-22
Update mobile apps api when changing competitions

davidfurey:
## What does this change?

When changes to competitions are required in frontend they are always required in mobile apps api too.  The change will be really similar - e.g. https://github.com/guardian/frontend/pull/14093 vs https://github.com/guardian/mobile-apps-api/pull/828.  This PR just adds a comment pointing this out.

Mobile football won't be updated 30 days after frontend, two developers won't have to make the same trivial change.  @rcrphillips will smile.

YES.  I've added the same note to mobile apps repo

## Request for comment

@gtrufitt

### DIFF
--- a/sport/app/football/feed/Competitions.scala
+++ b/sport/app/football/feed/Competitions.scala
@@ -89,6 +89,9 @@ object Competitions {
     val competitions = comps
   }
 }
+ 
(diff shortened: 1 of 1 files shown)


rcrphillips:
👍

gtrufitt (approved):
👯

PR #14423, opened 2016-09-22
Don't let imgix auto-choose format for email images

joelochlann:
Outlook wasn't displaying main image in articles like [this one](https://www.theguardian.com/media/2016/sep/02/itv-news-daily-mail-eamonn-holmes/email) because of imgix's automatic content negotiation. We were passing `auto=format` as a parameter to imgix, which according to [imgix's documentation](https://docs.imgix.com/tutorials/improved-compression-auto-content-negotiation) does the following:

> Behind the scenes, these are the steps we take during content negotiation:
> 1. Check if your browser supports WebP. If it does, we serve back a WebP image and return. 
> 2. Check if your browser supports JPEG XR. If it does, we serve back a JPEG XR image and return.
> 3. Check if the request has a fallback format. If it does, we serve back the image in the fallback format.
> 4. If we hit this point, serve back the image in its original format. 

When Outlook requests images for emails it sends the following User Agent string:

`User-Agent: Mozilla/4.0 (compatible; MSIE 7.0; Windows NT 6.3; WOW64; Trident/7.0; .NET4.0E; .NET4.0C; Microsoft Outlook 16.0.7167; ms-office; MSOffice 16)`

This causes imgix to return [JPEG XR](https://en.wikipedia.org/wiki/JPEG_XR) format with content-type `image/vnd.ms-photo`, which is [Microsoft-specific](http://caniuse.com/#feat=jpegxr). Ironically, Outlook can't render it!

So for email images, we now omit `auto=format` which means the image is always returned as a JPEG, which Outlook can render.

@desbo

### DIFF
--- a/common/app/views/support/Profile.scala
+++ b/common/app/views/support/Profile.scala
@@ -17,6 +17,7 @@ sealed trait ElementProfile {
   def hidpi: Boolean
   def compression: Int
   def isPng: Boolean
+  def autoFormat: Boolean
 
   def elementFor(image: ImageMedia): Option[ImageAsset] = {
     val sortedCrops = image.imageCrops.sortBy(-_.width)
@@ -38,7 +39,7 @@ sealed trait ElementProfile {
 
   // NOTE - if you modify this in any way there is a decent chance that you decache all our images :(
   val qualityparam = if (hidpi) {"q=20"} else {"q=55"}
-  val autoParam = "auto=format"
+  val autoParam = if (autoFormat) "auto=format" else ""
   val sharpParam = "usm=12"
   val fitParam = "fit=max"
   val dprParam = if (hidpi) {
@@ -66,7 +67,8 @@ case class Profile(
   override val height: Option[Int] = None,
   override val hidpi: Boolean = false,
   override val compression: Int = 95,
-  override val isPng: Boolean = false) extends ElementProfile
+  override val isPng: Boolean = false,
+  override val autoFormat: Boolean = true) extends ElementProfile
 
 object VideoProfile {
   lazy val ratioHD = new Fraction(16,9)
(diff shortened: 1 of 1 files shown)


desbo:
this was a tough one! great work joe 👍

joelochlann:
cc @davidfurey since he was instrumental in helping me track this one down and may want to see the fruits of our labours

paperboyo:
A good one!

> the image is always returned as a JPEG

Or a PNG (super-rarely for now): https://www.theguardian.com/lifeandstyle/lostinshowbiz/2016/aug/25/bonos-email-to-hillary-clinton-give-me-some-space/email

Couple of general points (not this PR):
- for all sorts of reasons these (all) resizes should always be created from `/master/` assets (compare image URL from [pure dotcom](https://i.guim.co.uk/img/media/618fd35e65705c736ab7f7f62d7d872d0495bbb6/115_34_4389_2633/master/4389.jpg?w=620&q=20&auto=format&usm=12&fit=max&dpr=2&s=b0373c71d4fdcc0ba70748f8b5afac56) vs. [/email](https://i.guim.co.uk/img/media/618fd35e65705c736ab7f7f62d7d872d0495bbb6/115_34_4389_2633/4389.jpg?w=640&q=55&usm=12&fit=max&s=93b21f1357c4efadfeac3d89b0d080a0))
- we may want to lower quality from 55 (maybe even to ~20-30, especially if we add `codec=mozjpeg`)
- we may want to differ the requested size depending on image size (especially for emails from fronts - smaller cards, but even main images in articles/email are shown at 580px but requested at 640px)
- there may be a way to make PNGs smaller, but I have to have a think about it. Would like to know (because they are very rare in articles): are there plans to make them work for front/email route (minority come from facia-tool image replacements, but some are in articles’ CAPI)?

“we may want” = much smaller and faster emails = more clicks = more 🎆

Obviously, we should be careful for any changes to only affect /email route...

PR #14428, opened 2016-09-22
re-add commercial badging tests

michaelwmcnamara:
This change adds tests to the amp-validation testing list.
This adds:

One page with commercial badges where the height and width of these badges is specified;
and One page with GLabs badging.


michaelwmcnamara:
I have also updated the url we are pointing to for the google maps test. However this remains commented out as the page in question currently fails amp validation

gtrufitt:
👍

PR #14429, opened 2016-09-22
Fix Immersive Interactives

sammorrisdesign-zz:
## What does this change?

Fixes immersive interactives. Imma just merge this in though because us interactives have some stuff to launch soon.

I'm sure there's a better way to do this though. Would love to know if there is.

### DIFF
--- a/common/app/model/content.scala
+++ b/common/app/model/content.scala
@@ -833,7 +833,15 @@ final case class Interactive(
     }
   }
 
-  lazy val hasSrcdoc = Jsoup.parseBodyFragment(fields.body).getElementsByTag("iframe").first().hasAttr("srcdoc")
+  lazy val hasSrcdoc = {
+    val iframe = Jsoup.parseBodyFragment(fields.body).getElementsByTag("iframe")
+
+    if (iframe.length > 0) {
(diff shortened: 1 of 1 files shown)


gtrufitt:
😵

common/app/model/content.scala
```diff
@@ -833,7 +833,15 @@ final case class Interactive(
     }
   }
 
-  lazy val hasSrcdoc = Jsoup.parseBodyFragment(fields.body).getElementsByTag("iframe").first().hasAttr("srcdoc")
+  lazy val hasSrcdoc = {
+    val iframe = Jsoup.parseBodyFragment(fields.body).getElementsByTag("iframe")
+
+    if (iframe.length > 0) {
```
rich-nguyen: You can achieve this in scala without length checking:

`val iframes = Jsoup.parseBodyFragment(fields.body).getElementsByTag("iframe")
iframes.headOption.exists(_.hasAttr("srcdoc"))
`

rich-nguyen:
👍

common/app/model/content.scala
sammorrisdesign: Ah sorry I just looked at the option above. That seems way cleaner though.

PR #14431, opened 2016-09-23
New navigation - inline editionalisation JS

SiAdcock:
## What does this change?

Depending on which edition the user has selected, we present different links within the menu. By default, we show links for the UK edition and then, if the user is looking at a non-UK edition, we unhide edition-specific links on the client using JavaScript.

There is a short delay between the initial render of the default UK links and the update to non-UK links as the JavaScript downloads and is executed on the client.

With this change, the JS to determine which links to show is inlined, making the delay between initial render and editionalisation imperceptible.
## Request for comment

@NataliaLKB


NataliaLKB:
Great stuff! 👍

NataliaLKB (approved):

PR #14432, opened 2016-09-23
Update google.scala.html

mkopka:
## What does this change?

Expands our campaign code data collection allowing for improved reporting of our various campaigns

Improved reporting

No 

## Request for comment

Adding in update to pass the 'CMP_TU' query param values into custom dimension23. Exact same thing we already do for other query params.


cb372:
👍

gtrufitt:
@mkopka Not sure if you're aware of the deployment process, but give me a shout today/when you want to merge and we can go through it if you need.

dominickendrick:
Oops, I was supposed to do this!

PR #14434, opened 2016-09-23
Remove dimensions from logos

kelvin-chappell:
They get used in different sizes on the site depending on context.

/cc @Calanthe


Calanthe:
👍

PR #14435, opened 2016-09-23
Radiator refresh

TBonnin:
## What does this change?

Slightly modify the radiator page:
- Move pingdom and deployments down
- Show router 50x rather than user facing 50x errors
- Ditch Teamcity widget and replace it with request to TC api and a simple html list. Also, adding the AmpValidation project build status
- Add links to deploys-radiator and loadbalancers dashboards

Better readibility of Teamcity build statuses.
Focus on the most important things: build status, switch status, pageviews and errors
## Screenshots

screen shot 2016-09-23 at 11 30 20
## Request for comment

### DIFF
--- a/admin/app/controllers/DeploysNotifyController.scala
+++ b/admin/app/controllers/DeploysNotifyController.scala
@@ -62,11 +62,8 @@ trait DeploysNotifyController extends Controller with ApiKeyAuthenticationSuppor
   //
 
   def notifyStep(number: String) = ApiKeyAuthenticatedAction.async(BodyJson[NotifyRequestBody]) { implicit request =>
-    teamcity.getTeamCityBuild(number).flatMap { teamcityBuild =>
-      val notices = for {
-        build <- teamcityBuild.right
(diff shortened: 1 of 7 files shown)


NataliaLKB:
Looks like all sensible changes 👍

rich-nguyen:
I think the user 500 graph is really useful, though. That's where you need to stop what you are doing, and investigate the issue. Router 500 is often a proxy for another issue, dotcom or non-dotcom related. Still useful though, alarm-worthy at least (I think it already has an alarm).

johnduffell:
I wouldn't want to wait until users are seeing errors, if router is 5xxing then we're already serving stale.  However the useful distinction is that user 5xx is only for content, wheras router 5xx is also for api calls.

TBonnin:
@rich-nguyen 
I am happy to bring user500 back. However I've noticed router is serving a fair amount of errors that are kept quiet and not investigated because nobody sees them. I though that maybe if they were visible, we would be more incline to look and fix them.

PR #14436, opened 2016-09-23
Add AMP data-loading-strategy attribute, "prefer-viewability-over-views"

rich-nguyen:
This adds an additional attribute to `amp-ad` tags on AMP pages. The data loading strategy affects how the amp platform loads adverts. More info [here](https://www.ampproject.org/docs/reference/extended/amp-ad.html).

Note, we expect viewability to increase, but as the docs state, we understand that ads may load so late they fail to generate a view (impression). Stephan says this is like moving from lazy-load to lazy-lazy-load.


rich-nguyen:
[Add 'prefer viewability over ads' to AMP ad calls](https://trello.com/c/iI5oM4PD/154-add-prefer-viewability-over-ads-to-amp-ad-calls)

janua:
:+1:

PR #14438, opened 2016-09-23
Radiator: Fix missing charts css class

TBonnin:
## What does this change?

Following up on https://github.com/guardian/frontend/pull/14435
Radiator: I forgot one of the css class which cause this layout issue:
screen shot 2016-09-23 at 13 19 54

Proper layout where pageviews and errors graphs sit side by side rather than on top of each other
## Request for comment

@NataliaLKB @gtrufitt


NataliaLKB:
👍

PR #14439, opened 2016-09-23
Removed Contributions CTA AB-test

markjamesbutler:
## What does this change?

Remove AB test as we now have the results.

None

No


kelvin-chappell:
👍

PR #14441, opened 2016-09-23
Badges markup refactor

Calanthe:
## What does this change?

In order to have badges' markup consistent everywhere on the page, I had to use code introduced by @regiskuckaertz in merchandising components (https://github.com/guardian/frontend/blob/master/common/app/views/commercial/containerWrapper.scala.html#L29 and https://github.com/guardian/frontend/blob/3be9765e2b52c887909ccb0b4dc11b6101c6ec30/commercial/app/views/contentapi/items.scala.html#L85)

This was more complicated that I assumed because we have those badges in a few different types of content pages. Most of those pages required changes in styling that's why I had to add a few conditions in the `contentLogo.scala.html` file.

We have those badges eg in:
1. Merchandising single and multiple - badge in the header on the left - done by Regis
2. Paid content containers - single branded - done by Regis
3. Paid content containers - multi branded - different logos in each card, they have to be always at the bottom of the card, that's why there is a `adverts__badge--card-branding` class added.
4. Interactive article - badge always on the right side, `adverts__badge--interactive` added.
5. Facia pages - badge always on the left but not inside header element. That's why I had to add `float: left`to the `adverts__badge--alt` class originally created by Regis.
6. On live blogs, badges are on the left column already so I had to remove the `float: left` style.
7. All of the non paid content: video, audio and gallery pages are on the black background so had to change the font colour.

Those changes also fixes the issue with 'jumping' `<h3/>` inline styles.

I still need to delete unused styles from the old markup eg `brandbadge--onside ad-slot--force-display ad-slot--paid-for-badge`. This will be done in another PR.

All of the pages from the https://docs.google.com/spreadsheets/d/17iRV-OFLC5dzLl-LgeZ_6Q9_ZGPNAVUbgg2jVbT_0fk/edit?ts=57e2a20b#gid=0 document have been tested against my changes.

Badges markup is consistent

no

## Screenshots

Merchandising single and multiple:
screen shot 2016-09-23 at 14 58 54

Paid content containers - single branded:
screen shot 2016-09-23 at 14 59 36 (logo is stretched but this will be fixed with https://github.com/guardian/frontend/pull/14434)

Paid content containers - multi branded:
screen shot 2016-09-23 at 15 01 48

Interactive article:
screen shot 2016-09-23 at 15 02 17

Facia page:
screen shot 2016-09-23 at 15 03 47

Live blog:
screen shot 2016-09-23 at 15 04 36

Video page:
screen shot 2016-09-23 at 15 05 13
## Request for comment

@kelvin-chappell @regiskuckaertz @lps88


common/app/views/commercial/cards/itemLargeCard.scala.html
```diff
@@ -27,7 +27,7 @@ <h2 class="advert__title">
         }
         @if(useCardBranding) {
             @for(branding <- item.branding) {
-                @contentLogo(branding)
+                @contentLogo(branding, isInteractivePage = false, isCardBranding = false, isOnTheLeft = false)
```
kelvin-chappell: Condition looks a bit strange.  This is a card.

common/app/views/commercial/cards/itemSmallCard.scala.html
```diff
@@ -18,7 +18,7 @@
     </p>
     @if(useCardBranding) {
         @for(branding <- item.branding) {
-            @contentLogo(branding)
+            @contentLogo(branding, isInteractivePage = false, isCardBranding = false, isOnTheLeft = false)
```
kelvin-chappell: This is also a card.

kelvin-chappell (approved):
Maybe rename `isCardBranding` to `isStandardSizeCardBranding` or something like that if that's the only place it applies.

I'll fix the conflict.

common/app/views/commercial/cards/itemCard.scala.html
```diff
@@ -16,7 +16,7 @@
 <div class="@CssClassBuilder.linkFromStandardCard(item, optAdvertClassNames, optClassNames, useCardBranding)">
     <a href="@item.targetUrl"
        data-link-name="@omnitureId"
-       class="advert">
+       class="advert advert--card-branding">
```
regiskuckaertz: Perhaps `advert--branded` would be better here. Would also solve @kelvin-chappell 's comments with `isCardBranding -> isBranded`

common/app/views/fragments/commercial/contentLogo.scala.html
```diff
@@ -1,33 +1,35 @@
...
-                }
-            } else {
-                <img src="@branding.sponsorLogo.url"
-                    @for(dim <- branding.sponsorLogo.dimensions) {
-                        width="@dim.width"
-                        height="@dim.height"
-                    }
-                    alt="@{branding.sponsorName}'s logo"
-                    class="brandbadge__logo">
+<div class="adverts__badge js-badge
```
regiskuckaertz: I wonder if we still need the `.js-badge` class...

regiskuckaertz: Also, because badges can appear in other contexts, what do you think of dropping the `adverts__`? It seems to me that it's a bona fide component.

regiskuckaertz (approved):
Great! Aside minor edits, looks great to me.

commercial/app/views/contentapi/item.scala.html
```diff
@@ -75,12 +75,12 @@
 @badge = {
 
     @optSponsorLabel
-    <a class="adverts__badge__link" href="@clickMacro@optCapiLink" data-link-name="logo link">
-        @for(logoUrl <- optLogo) {<img class="adverts__badge__logo" src="@logoUrl" alt="">}
+    <a class="badge__link" href="@clickMacro@optCapiLink" data-link-name="logo link">
```
regiskuckaertz: 🏆

regiskuckaertz (approved):

common/app/views/fragments/commercial/contentLogo.scala.html
Calanthe: Of course, I had to remove the `adverts__`part because otherwise, badges won't be visible behind adblockers.

Calanthe: Thanks for reminding me that ;)

PR #14443, opened 2016-09-23
Update AB tests doc and admin page

NataliaLKB:
## What does this change?

At the moment we have quite a lot of overlapping tests. Though we should always be testing lots of things, we should be conscientious other tests the same users might be in. 

**Good questions to consider:**
Are we showing the users more than one disruptive thing at a time?
Will this test conflict with other tests running on the same users?
And will any overlapping tests affect the data?

At the moment I think the biggest problem is that people are not aware of other tests running at the moment. This is what this PR aims to change. That is why I have added a comment to our PR template pointing to our helpful doc.

I have discussed with people whether we should make it a programatic rule that tests can't overlap, but I feel that may cause more problems then it will solve. We have lots of teams who should have the ability to run their tests, and run them quickly (may need a higher % of users to run certain tests). Limiting how many tests can run at one time will mean many tests which are valuable could have to wait for other tests to finish. 

Our approach so far is very much to trust our devs and their teams to use their best judgment. I would like to keep this approach and instead aim to make people more aware of our guidelines.

The value of these updates is that it is easier and quicker for people to see the audience breakdown.

We will know this is successful if we have less overlapping tests

Nope!
## Screenshots

Abtest on admin page:

image
## Request for comment

@guardian/dotcom-platform


admin/app/views/abtests.scala.html
```diff
@@ -25,12 +25,12 @@
         @fragments.abTestReport()
     </div>
 
-    <div class="abtests-audience"></div>
-
     <h3 class="abtests-expired-title">Expired tests <a href="#">show</a></h3>
 
     <div class="abtests-expired"></div>
 
+    @fragments.serverBasedTests()
```
rjmk: Should this be above expired tests?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -2,6 +2,10 @@
 
 ## What is the value of this and can you measure success?
 
+<!--
+If setting up an AB test, make sure you have read our [AB testing doc](https://github.com/guardian/frontend/blob/master/docs/03-dev-howtos/01-ab-testing.md).
```
rjmk: People may have been unable to set up an AB test without reading that doc in the first place; is it worth having a final section of the AB README that summarises things to double check (Including the new addition to the README)?

admin/app/views/abtests.scala.html
rjmk: Cool 👍

.github/PULL_REQUEST_TEMPLATE.md
NataliaLKB: > People may have been unable to set up an AB test

Do you mean able? 

I agree though. A summary is good. Maybe we can just make the first part of this better: https://github.com/guardian/frontend/blob/master/docs/03-dev-howtos/01-ab-testing.md#guide

What do you think?

rjmk: > Do you mean able?

No, though I agree my phrasing is confusing. I meant that in general, I would expect people who have set up an AB test to have read that doc at least once. But the way you would read that doc when trying to figure out how to create a working AB test would be different from the way you would read it to make sure you've done everything right. And a little paragraph is easy to miss even if you do a due diligence sweep of the doc afterwards.

> Maybe we can just make the first part of this better

That's a good idea. Maybe there's just an extra step after "Running the test" which is "Checking for test clashes"?

johnduffell:
please note that overlapping tests are ok in principle if we need to run too many - the idea is that we don't just use the space as a linear space but we use multiple dimensions to keep the groups in each test identical even if both groups include people in another test in even proportion. see my pretty diagram:
image

however if being in both tests would cause an invalid user experience, rather than just happening to both affect the same metrics, then this would not work.

rjmk:
Yes, @NataliaLKB is definitely aware of that issue. Do you think that the docs suggest that this is not allowed? I think the intention is to make sure people are at least *aware* of overlapping tests.

Also, I'm not entirely sure I follow

> we use multiple dimensions to keep the groups in each test identical even if both groups include people in another test in even proportion

I think the idea is that if fraction x of people are in the variant for test 1 and fraction y are in the variant for test 2, we know that x * y are in both, (1-x)*(1-y) are in neither, and so on. Is that right?

.github/PULL_REQUEST_TEMPLATE.md
NataliaLKB: Check out my last commit @rjmk

NataliaLKB:
Thanks @johnduffell and @rjmk . Yes I think in principle having a user in more than one test is completely fine, but not if both those tests are showing the user something which they have to close/dismiss. It is just too much noise for one user and potentially invalidates the tests.

I don't think people are aware of what other tests are running which is what I would like to help change.

johnduffell:
yes sorry didn't mean to derail things, I just wanted to remind people that overlapping isn't always bad if you do it right, keep going as you were!

paperboyo:
> both tests are showing the user something which they have to close/dismiss. It is just too much noise for one user

👍 Bad situation is permanently made worse if both tests, separately, prove successful. And both annoyances end up in production 😜

NataliaLKB:
@rjmk @johnduffell @paperboyo anymore changes that I should add?

johnduffell:
nothing from me!

docs/03-dev-howtos/01-ab-testing.md
```diff
@@ -8,13 +8,21 @@ Most tests can be written in JavaScript, although we can serve variants via Varn
...
+ - [Adding a switch to turn the test on & off](#adding-a-switch)
+ - [Writing a test, which is typically a simple AMD module](#writing-a-test)
+ - [Running the test](#running-the-test)
+ - [Analysis of the test data](#analysis-of-the-test-data)
+ - [Share your findings](#share-your-findings)
  - Delete the test
 
+## Quick Tips
+
+- Creating your switch: if its an ab test it should start with `ab-` (see more naming conventions in [Adding a switch](#adding-a-switch)
```
rjmk: `s/its/it's`

rjmk:
Spotted a typo, but :

docs/README.md
```diff
@@ -18,7 +18,7 @@
 - [Javascript](02-architecture/06-client-side-architecture.md)
 
 ##[Dev howtos](03-dev-howtos/)
-- [How to setup and run A/B tests](03-dev-howtos/01-ab-testing.md)
+- [](03-dev-howtos/01-ab-testing.md)
```
rjmk: Oops?

NataliaLKB: Haha good catch!

NataliaLKB: Made another edit and made sure I didn't have a space at the top of my file.

PR #14446, opened 2016-09-23
Avoid transition on closing ad

regiskuckaertz:
This will fix the issue where advert copy is overlaid with the CTA when the Fabric video is getting collapsed.


JonNorman:
:+1:

PR #14447, opened 2016-09-23
Native ads: resize iframe along with parent ad slot

regiskuckaertz:
Normally, DFP resizes the iframe of native ads so that the height matches the height of the creative. But, with some of our creatives injecting HTML at load time, this does not seem to work:

picture 6

This above is the result of us resizing the ad _slot_ to the size of the creative, and DFP not resizing the ad _frame_ at all. As a solution, our resize routine will now handle the frame too.

picture 5


JonNorman:
Tasty :+1:

PR #14448, opened 2016-09-26
Update grunt-svgmin

gtrufitt:
## What does this change?

This updates grunt-svgmin to the latest which uses the [latest svgo](https://github.com/svg/svgo/releases/tag/0.7.0)

Also added the `removeXMLNS` option which strips the many accidently left in xmlns attrs from inline-svgs.
## Request for comment

@paperboyo - What was it you were excited about? 
@guardian/dotcom-platform

### DIFF
--- a/grunt-configs/svgmin.js
+++ b/grunt-configs/svgmin.js
@@ -6,7 +6,14 @@ module.exports = function () {
                 cwd: 'common/conf/assets/inline-svgs',
                 src: ['**/*.svg'],
                 dest: 'common/conf/assets/inline-svgs'
-            }]
+            }],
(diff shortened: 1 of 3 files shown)


NataliaLKB:
great! 👍

paperboyo:
> What was it you were excited about?

E-ve-ry-thing. Just a lot of bugs were squashed in between 6 and 7. Plus, I'm an update junkie. 😄

gtrufitt:
Haha, fair enough - if you've got any specific plugins you think we should be using for this let me know (I already turned on auto-remove xmlns)

PR #14451, opened 2016-09-26
New header: fix Android tap highlight position

SiAdcock:
## What does this change?

Previously, when a user tapped an item or link within the new menu, tap highlights on Chrome for Android appeared slightly offset down and to the right. This turned out to be due to the combination of the `box-shadow` and `overflow`. 

I have moved the `box-shadow` from the menu up to the container, preserving the box shadow's visual effect, but preventing the tap highlight from being offset.

Nah
### Before

android-tap-highlight-before
### After

android-tap-highlight-after
## Request for comment

@NataliaLKB


NataliaLKB:
What a strange bug! Well done for find it 👍

PR #14452, opened 2016-09-26
Bump frontend configuration version to 5

regiskuckaertz:
The URL of the Jobs feed property is wrong in the current configuration for the DEV environment:

```
jobs.api.url.template="http://jobs.theguardian.com/feeds/GuardianLiveJobsForWebComponents_yyyy-MM-dd_.xml"
```

It should be identical to the PROD-CODE environment, i.e.

```
jobs.api.url = "http://jobs.theguardian.com/feeds/GuardianLiveJobsForWebComponents.xml"
```

... which is what I changed in v5.

cc @johnduffell


NataliaLKB:
👍

PR #14453, opened 2016-09-26
cApi merchandising spacing fixed

Calanthe:
## What does this change?

Fix for the merchandising items. There should be spacing (additional `advert--branded` class) only when there is badge inside item.

## Screenshots

Before:
screen shot 2016-09-26 at 14 32 53

After:
screen shot 2016-09-26 at 14 32 58
## Request for comment

@lps88 @kelvin-chappell


kelvin-chappell:
Will logos ever end?
👍

Calanthe:
@kelvin-chappell Probably not but I think that we can see some light in the end of the tunnel now ;)

PR #14454, opened 2016-09-26
New header: When you are signed in you should see your account links

NataliaLKB:
## What does this change?

For the new header, this adds the behaviour for detecting when you are signed in. 

If you are, then you should be shown your account links instead of still staying "sign in".

This also updates the styling, but I am sure @zeftilldeath will want to make it even better! 🎉 

For users in this test you can now access all the your my account links when you are signed in.

Nope
## Screenshots

And after you have signed in: 

## Request for comment

@zeftilldeath @SiAdcock


zeftilldeath:
Awesome! I will do my styling fiddles in a seperate PR post-merge. 

🐫

common/app/views/fragments/nav/userAccountLinks.scala.html
```diff
@@ -0,0 +1,62 @@
...
+                    <li class="user-account__signed-in__link">
+                        <a href="@Configuration.id.url/email-prefs"
+                        data-link-name="Email preferences">
+                            Email preferences
+                        </a>
+                    </li>
+                    <li class="user-account__signed-in__link">
+                        <a href="@Configuration.id.url/password/change"
+                        data-link-name="Change password">
+                            Change password</a>
```
dominickendrick: move `</a>` to new line for consistency ?

paperboyo:
Shouldn't "my account" read "[username]"? And maybe even we should swap a generic avatar for whatever avatar the user has set?

NataliaLKB:
@paperboyo 

> Shouldn't "my account" read "[username]"? And maybe even we should swap a generic avatar for whatever avatar the user has set?

We decided not to go that route. We don't think that it is as clear as "my account". However it is always possible this will change again in the future. @stephanfowler may be able to add more.

common/app/views/fragments/nav/userAccountLinks.scala.html
NataliaLKB: done!

stephanfowler:
@paperboyo : for the first public test, we're focusing on the utility of and performance of the nav. We need to just get this out - then we'll iterate separately on the best solution for each of signin/membership/everything else.

static/src/javascripts/projects/common/modules/navigation/is-signed-in.js
```diff
@@ -0,0 +1,36 @@
+define([
+    'fastdom',
+    'qwery',
+    'common/modules/identity/api'
+], function (
+    fastdom,
```
NataliaLKB: Haha good catch. I new I was missing something!

static/src/javascripts/projects/common/modules/navigation/newHeaderNavigation.js
```diff
@@ -178,6 +180,7 @@ define([
         enhanceCheckboxesToButtons();
         bindMenuItemClickEvents();
         bindPrimaryItemsClickEvents();
+        isSignedIn();
```
NataliaLKB: Great point! I renamed it to `user-account`

common/app/views/fragments/nav/userAccountLinks.scala.html
```diff
@@ -0,0 +1,62 @@
...
+                <ul class="user-account__signed-in__links">
+                    @if(SaveForLaterSwitch.isSwitchedOn) {
+                        <li class="user-account__signed-in__link">
+                            <a href="@Configuration.id.url/saved-for-later"
+                            data-link-name="Saved for Later">
+                                Saved for later
+                            </a>
+                        </li>
+                    }
+                    <li class="user-account__signed-in__link js-add-comment-activity-link" hidden>
```
NataliaLKB: Given there isn't any logic in this template (woohoo!), I don't think it would add much to abstract the links, classes, data names, and display names for these things. They won't change much and are not editionalised. That said, if it makes it more readable we should abstract!

static/src/javascripts/projects/common/modules/navigation/is-signed-in.js
NataliaLKB: This is sorted now, can you take another look?

SiAdcock:
💎 💎 💎 💎 💎 👍 💎 💎 💎 💎 💎
Love it!

PR #14457, opened 2016-09-26
Remove the annoying and redundant .json suffix from soulmates routes.

JonNorman:
Makes the Native Templates code a bit nicer as we can store the URL in the config file.

@regiskuckaertz


Calum-Campbell:
👍

PR #14458, opened 2016-09-26
Added Ophan attributes for new header

SiAdcock:
## What does this change?

Identifies the `nav` component and link names in the new header navigation.

This will allow us to identify in Ophan whether and how users are using the new menu.

No
## Request for comment

@NataliaLKB

### DIFF
--- a/common/app/views/fragments/nav/newHeaderMenu.scala.html
+++ b/common/app/views/fragments/nav/newHeaderMenu.scala.html
@@ -21,7 +21,7 @@
 
                     @topLevelSection.getEditionalisedNavLinks(edition).map { sectionItem =>
                         <li class="navigation-group__item">
-                            <a href="@LinkTo { @sectionItem.url }">@sectionItem.name</a>
+                            <a href="@LinkTo { @sectionItem.url }" data-link-name="nav : secondary : @sectionItem.name">@sectionItem.name</a>
                         </li>
                     }
                 </ul>
--- a/common/app/views/fragments/nav/newHeaderMenu.scala.html
+++ b/common/app/views/fragments/nav/newHeaderMenu.scala.html
@@ -21,7 +21,7 @@
 
                     @topLevelSection.getEditionalisedNavLinks(edition).map { sectionItem =>
                         <li class="navigation-group__item">
-                            <a href="@LinkTo { @sectionItem.url }" data-link-name="nav : secondary : @sectionItem.name">@sectionItem.name</a>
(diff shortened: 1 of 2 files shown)


NataliaLKB:
pinging @stephanfowler for this too.

common/app/views/fragments/nav/newHeaderMenu.scala.html
```diff
@@ -21,7 +21,7 @@
 
                     @topLevelSection.getEditionalisedNavLinks(edition).map { sectionItem =>
                         <li class="navigation-group__item">
-                            <a href="@LinkTo { @sectionItem.url }">@sectionItem.name</a>
+                            <a href="@LinkTo { @sectionItem.url }" data-link-name="nav : secondary : @sectionItem.name">@sectionItem.name</a>
```
NataliaLKB: I think we can actually make it clearer and call it just the `sectionItem.name`. The rest just adds to the over complicated nature of our current nav. 

Previously people could click on the nav links in the footer and the nav links in the header and we needed to differentiate. But we don't need to do that anymore with the new header so it should be okay to omit this.

stephanfowler:
Wondering whether data-component should be something like `nav2`, to differentiate from clicks on the old nav. The AB variants will differentiate the traffic initially, but it might just make analysis easier to have separate components from the start. Whaddyou think?

NataliaLKB:
@stephanfowler 
> Wondering whether data-component should be something like nav2, to differentiate from clicks on the old nav. The AB variants will differentiate the traffic initially, but it might just make analysis easier to have separate components from the start. Whaddyou think?

That makes sense. I initially advised to make it the same, but if it will make analysis easier then we should do it! 

What do you think should be a good name @SiAdcock ?

SiAdcock:
Sounds good. It's tempting to follow the convention we have used in the codebase and go with something like `new-header-menu`, but  the `new` part is troubling. Eventually the new header will no longer be new, but it will still be labelled as such.

While it's relatively easy to refactor the frontend codebase to remove the`new` part, it will probably be less desirable to do this in Ophan, where consistency of data over time is more important. With that in mind, shall we go with something like `main-navigation`?

SiAdcock:
@NataliaLKB @stephanfowler 

I've updated the `data-link-name` and `data-component` attributes, let me know if this is closer to the acceptable 😄

stephanfowler:
`main-navigation`is fine.

NataliaLKB:
👍

PR #14462, opened 2016-09-27
Show logos on immersive articles

kelvin-chappell:
Condition for showing logos is wrong on immersive articles.
This should fix it.

/cc @guardian/commercial-dev


regiskuckaertz (approved):
¡Ajiba!

PR #14463, opened 2016-09-27
[BONZO ALL OVER AGAIN!!!] Promote preact as default variant in discussion frontend

piuccio:
## What does this change?

Sorry for the caps.
After a long fiddling with (p)react and the nested dependencies, I've decided to use preact instead of react as our default library for discussion frontend.

I'm using `preact-compat` to maintain compatibility with react and `discussion-frontend` still generates both variants in case preact turns out to be sub-optimal in the future.

I managed to tree shake the following out of react
- Proptypes
- React.createClass
- Vendor prefixed `transition` and `animation` events
- React dev tools

but the result is:

| React | Preact |
| --- | --- |
| 48.2 KB | 15.6 KB |

(\* minified and gzipped)

`~32KB` bigger is a hard sell.

What's left inside React that doesn't seem very useful at the moment, but it's much harder to tree-shake is:
- Event normalization: I'm only using a click listener, but React adds support for touch events, transitions, scroll/wheel, audio/video playing events ...
- SVGs, they are XML so React can render them from code, fancy stuff but all my svgs are constant and injected as they are
- Refs to components, as in `<input ref={(input) => this._input = input} />`. I'm sure it's handy in some cases but I'm not using it and it takes space.

Preact doesn't even support them (well, maybe `ref`, I don't know)

**important**:
In order to reach all browsers, `discussion-frontend` allows to inject `Promise` as dependency, while it bundles `fetch` polyfill because it only contributes half a kilobyte gzipped.
The are better approaches but the reasoning is explained [here](https://github.com/guardian/discussion-frontend/pull/15).

Remove a 100% audience AB test and use a single code base instead of the 3 options react, preact, old.

No

## Request for comment

@guardian/guardian-frontend 
FYI @developit


static/src/javascripts/projects/common/modules/discussion/discussion-frontend.js
```diff
@@ -1,33 +1,39 @@
...
+                if (value === 0) {
+                    loader.setState('empty');
+                }
+                mediator.emit('comments-count-loaded');
+            });
+        }
+
+        return require('discussion-frontend-' + VARIANT, function (frontend) {
             // Preact works in a slightly different way
             // https://github.com/developit/preact-compat/issues/145
```
dominickendrick: This issue seems to be closed, does this code still need to be here ?

static/src/javascripts/projects/common/modules/discussion/discussion-frontend.js
```diff
@@ -1,33 +1,39 @@
...
 ], function(
-    fetchJson,
     mediator,
     reportError
 ) {
-    function canRun(ab, curlConfig) {
-        return (ab.isInVariant('DiscussionExternalFrontendAvatar', 'react') && curlConfig.paths['discussion-frontend-react']) ||
-            (ab.isInVariant('DiscussionExternalFrontendAvatar', 'preact') && curlConfig.paths['discussion-frontend-preact']);
-    }
+    var VARIANT = 'preact';
```
dominickendrick: Is it worth setting the constant as we've decided to go with `preact` ? 
Why not just `require('discussion-frontend-preact')`,

static/src/javascripts/projects/common/modules/discussion/discussion-frontend.js
piuccio: I didn't realize it was fixed, let me check.

static/src/javascripts/projects/common/modules/discussion/discussion-frontend.js
piuccio: just in case we want to use `'react'` in the future or you want to override it locally

developit:
Nice summary @piuccio! Just a note - inline svg and refs are both supported in Preact :tada:

piuccio:
@developit good to know, maybe I should try and tree-shake them 😁

piuccio:
@dominickendrick I've changed the code with your suggestions and removed some more dead code. Especially the template part, it now lives in `discussion-frontend`

desbo:
awesome work fabio

piuccio:
@desbo yet nobody gave a +1 😢

dominickendrick:
👍

PR #14465, opened 2016-09-27
Returing 404 when cannot retrieve a discussion from DAPI

TBonnin:
## What does this change?

Previous commit https://github.com/guardian/frontend/commit/a524158f7616b0a04e14f4e8e4614b4746d19a5d
explicitely changed 404 to 500 (in addition to add logging).
This patch revert the change of response type.

Stop returning 500 for non-existing discussions which should be used when an error happens and not when content doesn't exist
## Request for comment

@nicl @jfsoul


jfsoul (approved):
👍  sounds sensible to me

PR #14466, opened 2016-09-27
Updates routes for soulmates JSON and caches Nil rendered result.

JonNorman:
This brings the soulmates JSON endpoint into line with the convention of the other soulmates components. It also ensures we are caching the `Nil` result when we render it, which was previously being explicitly uncached...


janua:
:+1:

PR #14467, opened 2016-09-27
Head space

zeftilldeath:
A few tweaks to prevent page content from getting too intimate with the new header.

eg: 
screen shot 2016-09-26 at 14 16 00

screen shot 2016-09-27 at 11 29 38 

Sure there may be more out there but this is as much as I can find at the moment. 

CC @NataliaLKB


NataliaLKB:
lgtm 👍

PR #14469, opened 2016-09-27
Badges - bylines fixed on mobile

Calanthe:
## What does this change?

Fix for the mobile version of badges (in articles and commercial containers - in the left column).

Badges look better when not floated and without the redundant top border.

no

## Screenshots

Before:
screen shot 2016-09-27 at 12 52 45

After:
screen shot 2016-09-27 at 12 53 29
## Request for comment

@lps88


lps88 (approved):
👍

PR #14470, opened 2016-09-27
Removed expiring switch

markjamesbutler:
## What does this change?

Removed the switch that I missed when removing the test.

No


jfsoul:
👍

PR #14472, opened 2016-09-27
Sonobi 50 percent

janua:
This removes the multi variant tests around `Sonobi` which currently has a `sonobi` test, a `control` test and a `prebid` test group.
I have removed both `control` and `prebid` as they aren't currently used for anything.

We are leaving `sonobi` to be used as a switch with a new percentage of test users. See referenced PR.

@rich-nguyen

### DIFF
--- a/static/src/javascripts/projects/commercial/modules/dfp/ophan-tracking.js
+++ b/static/src/javascripts/projects/commercial/modules/dfp/ophan-tracking.js
@@ -173,10 +173,7 @@ define([
     }
 
     function reportTrackingData() {
-        if (config.tests.commercialClientLogging ||
-            config.tests.commercialHbSonobi ||
-            config.tests.commercialHbPrebid ||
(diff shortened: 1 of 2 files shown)


static/src/javascripts/projects/commercial/modules/dfp/ophan-tracking.js
```diff
@@ -173,10 +173,7 @@ define([
     }
 
     function reportTrackingData() {
-        if (config.tests.commercialClientLogging ||
-            config.tests.commercialHbSonobi ||
```
janua: @rich-nguyen Could have left this here, but didn't see the point. Let's just report all users that are in the logging group.

rich-nguyen: ok yep

rich-nguyen:
👍

PR #14473, opened 2016-09-27
JSON endpoint for CAPI merchandising component

annebyrne:
## What does this change?

Adds a new endpoint to populate the DFP native template with content from CAPI. The new function makes the controller rather verbose, but once the current templates have been deprecated the wordier method will be removed.

The new native templates will also auto-populate branding information so that has been included here.

@guardian/commercial-dev


commercial/app/controllers/commercial/ContentApiOffersController.scala
```diff
@@ -111,6 +115,64 @@ class ContentApiOffersController(contentApiClient: ContentApiClient, capiAgent:
...
+      val capiSingle = CapiSingle.fromContent(response)
+      Cached(60.seconds) {
+        JsonComponent(capiSingle)
+      }
+    })
+
+  }
+
+  def nativeJson = renderNative(jsonFormat, isMulti = false)
+  def nativeJsonMulti = renderNative(jsonFormat, isMulti = true)
```
rich-nguyen: do we need `nativeJsonMulti`?

rich-nguyen:
Ace 👍

commercial/app/controllers/commercial/ContentApiOffersController.scala
annebyrne: Not yet: i pre-empted that @Ap0c will need a different endpoint, and tried to combine the methods as much as possible.

commercial/app/controllers/commercial/ContentApiOffersController.scala
```diff
@@ -111,6 +115,64 @@ class ContentApiOffersController(contentApiClient: ContentApiClient, capiAgent:
     }
   }
 
+
+  case class CapiSingle(articleHeadline: String, articleUrl: String,
```
regiskuckaertz: Move this into model?

commercial/app/controllers/commercial/ContentApiOffersController.scala
```diff
@@ -111,6 +115,64 @@ class ContentApiOffersController(contentApiClient: ContentApiClient, capiAgent:
...
+      val response = content.head
+      val capiSingle = CapiSingle.fromContent(response)
+      Cached(60.seconds) {
+        JsonComponent(capiSingle)
+      }
+    })
+
+  }
+
+  def nativeJson = renderNative(jsonFormat, isMulti = false)
```
regiskuckaertz: `renderNative` and `renderItems` have a lot in common, you could take the data extraction part out of both so that they only take care of the rendering?

regiskuckaertz (changes requested):

commercial/app/controllers/commercial/ContentApiOffersController.scala
annebyrne: In the latest commit I've removed the common code, as agreed its v wordy!

annebyrne:
@Ap0c FYI

regiskuckaertz (approved):

commercial/app/controllers/commercial/ContentApiOffersController.scala
```diff
@@ -30,33 +34,39 @@ class ContentApiOffersController(contentApiClient: ContentApiClient, capiAgent:
...
 
-    futureContents.map(_.toList) map {
+    futureContents
+  }
+
+
+  private def renderItems(format: Format, isMulti: Boolean) = Action.async { implicit request =>
+
+    retrieveContent().map(_.toList) map {
       case Nil => NoCache(format.nilResult.result)
```
JonNorman: If you take a look at the `SoulmatesController` or equivalent for Books and Travel, we now cache the `Nil` responses for a short amount of time.

JonNorman: https://github.com/guardian/frontend/blob/master/commercial/app/controllers/commercial/BookOffersController.scala#L60

annebyrne: This should be solved with https://github.com/guardian/frontend/pull/14473/commits/af9fb3ae7012ce1cb73081b451cf348e866e6859

commercial/app/model/commercial/CapiSingle.scala
```diff
@@ -0,0 +1,24 @@
...
+
+case class CapiSingle(articleHeadline: String, articleUrl: String,
+                      articleText: Option[String], articleImage: Seq[ImageElement],
+                      audioTag: Boolean, galleryTag: Boolean,
+                      videoTag: Boolean, branding: Option[Branding])
+
+object CapiSingle {
+  import ElementsFormat._
+
+  def fromContent(contentType: ContentType): CapiSingle = {
```
TBonnin: should this def called `apply` so you can use it as auxiliary constructor like this `CapiSingle(someContentType)`?

annebyrne: It was originally. I changed so that it wouldn't impact on serializing the JSON.

JonNorman: You can, but then when it comes to using the Play JSON reads/writes macros, it can stumble when it tries to find the correct `apply`/`unapply` method to use.

JonNorman: Doh. Late as usual :(

annebyrne: Much more eloquent than me, cheers Jon! 👍

TBonnin: Oh I see. Nevermind then

PR #14474, opened 2016-09-27
New header: don't show the nav in the footer

NataliaLKB:
## What does this change?

This adds the class `hide-on-mobile` when you are in the new header test. What this means is that on mobile (when you also see the new header) you don't have access to the old navigation in the footer.

However, when you are on desktop you still do.

When you are in the new header test, you don't have access to the current header

Nope!
## Screenshots

Before:
image

After:
image
## Request for comment

@SiAdcock @zeftilldeath


SiAdcock:
👍

PR #14475, opened 2016-09-27
Sort out hosted metadata

kelvin-chappell:
This makes hosted metadata consistent, and it makes use of the `isHosted` flag that now comes from capi to mark hosted content.  

This means we don't need to use tone to identify hosted content, because the tone of hosted content is `advertisement-features`.

/cc @guardian/labs-beta


JonNorman:
Looks good to me :+1:

PR #12439, opened 2016-04-06
New header WIP: initial version (0%)

OliverJAsh:
**Please note that this is very much still WIP.**

This is an initial version of the new header. It will be served to 0% and accessible via a header (`X-GU-ab-new-header: variant`).

By starting to integrate the design, we can easily share it around, which will enable a faster feedback cycle with internal users whilst we continue to iterate.

For full details about the new header, please refer to the [rationale document](https://docs.google.com/document/d/1wuQ8Qd0PMT7FhThkHJdyDCANx_0crxH2Y8GgrosFVn0/edit#heading=h.t2w358cznb3).

/cc @zeftilldeath @stephanfowler @NataliaLKB @johnduffell
# To do
- [ ] [Update cache key with test request header](https://github.com/guardian/fastly-edge-cache/pull/599)
# Screenshots
## Mobile

image
## Desktop

image

PR #14478, opened 2016-09-27
Enable sticky ads when opted in to new header

SiAdcock:
## What does this change?

The sticky ad banner was disabled if the user participating in the header test. This was [originally disabled](https://github.com/guardian/frontend/pull/12439#r58707823) due to uncertainty around how the sticky banner would interact with the new header navigation.

However, in the meanwhile, it was decided that the new header navigation would not be implemented on desktop, meaning it will not interfere with the sticky banner, which applies to desktop only. 

Therefore the sticky banner can be enabled even when the user has opted in to the new header.
## Request for comment

@NataliaLKB


NataliaLKB:
👍

PR #14480, opened 2016-09-27
Add AB test for rich link no upgrade above viewport

gtrufitt:
## What does this change?

Adds a test so that we only upgrade rich links on mobile below the current viewport - avoiding yank. 

Our previous sledge hammer no upgrade test made too much of a difference to onward journeys from rich links.

 @stephanfowler  @guardian/dotcom-platform


static/src/javascripts/projects/common/modules/article/rich-links.js
```diff
@@ -24,37 +26,53 @@ define([
...
+        return fastdomPromise.read(function(){
+            var rect = el.getBoundingClientRect();
+            return rect.top > (window.innerHeight || document.documentElement.clientHeight);
+        });
+    }
+
     function upgradeRichLink(el) {
         var href = $('a', el).attr('href'),
-            matches = href.match(/(?:^https?:\/\/(?:www\.|m\.code\.dev-)theguardian\.com)?(\/.*)/);
+            matches = href.match(/(?:^https?:\/\/(?:www\.|m\.code\.dev-)theguardian\.com)?(\/.*)/),
```
dominickendrick: Separate `var`s please !

static/src/javascripts/projects/common/modules/experiments/tests/upgrade-mobile-rich-links-below-viewport.js
```diff
@@ -0,0 +1,36 @@
...
+        this.idealOutcome = 'Happier users & no major drop in article visits per browser per day';
+
+        this.canRun = function () {
+            return config.page.isContent && detect.isBreakpoint({max: 'mobile'}) ;
+        };
+
+        this.variants = [
+            {
+                id: 'control',
+                test: function () {}
```
dominickendrick: @desbo Are we able to add component level click through tracking in `abacus` ?

dominickendrick:
👍

static/src/javascripts/projects/common/modules/experiments/tests/upgrade-mobile-rich-links-below-viewport.js
desbo: Can you give me an example? It seems like it might need some extra paramaters passed through to the query so could be unlikely.

gtrufitt: Right now there is `onward-journeys` in Abacus but, in the example of this test, we want to track `onward journeys form rich links` so would need to pass the `rich-link` ID from `data-component` and measure clicks. We could do this manually by calling the complete method but it'd be good if it was built into abacus along with other component-level metrics (like visibility).

(PS: I don't think we'll hold up this test but just a query for future reference)

PR #14471, opened 2016-09-27
Adblocking test with ad-free offering. 

JustinPinner:
## What does this change?

This is the second iteration of the adblock engagement test (https://github.com/guardian/frontend/pull/14072). We're exploring the appetite for an ad-free offering, targeted initially at our ad-blocking readers.

**This is a ZERO PERCENT test** for validating everything prior to deploying the real deal.
## What is the value of this and can you measure success?

We're exploring the possibility that a significant proportion of ad-blocking readers are interested in paying for an ad-free experience.
## Does this affect other platforms - Amp, Apps, etc?

No.

## Screenshots

screen shot 2016-09-27 at 14 37 28

screen shot 2016-09-27 at 14 37 54

screen shot 2016-09-27 at 14 38 12
## Request for comment

@regiskuckaertz @rich-nguyen @kenlim

PR #14481, opened 2016-09-27
Set the adblock with ad-free test live.

JustinPinner:
## What does this change?

Set the ad block with ad-free experiment live. See https://github.com/guardian/frontend/pull/14471 for more info.

**We're planning to activate this tomorrow morning (28th)**.

As described in the previous PR (above). This now picks up an offset audience of 12% spread over 4 test categories.

No.

## Screenshots

See previous PR (https://github.com/guardian/frontend/pull/14471)
## Request for comment

@dominickendrick @desbo @ScottPainterGNM @regiskuckaertz

### DIFF
--- a/static/src/javascripts/projects/common/modules/experiments/tests/adblocking-response.js
+++ b/static/src/javascripts/projects/common/modules/experiments/tests/adblocking-response.js
@@ -16,13 +16,13 @@ define([
     SurveyAdBlock
 ) {
     return function () {
-        this.id = 'AdBlockingResponse30pc';
-        this.start = '2016-09-22';
+        this.id = 'AdBlockingResponse3';
+        this.start = '2016-09-28';
         this.expiry = '2016-10-14'; // this test will expire on Thurs 13th PM.
         this.author = 'Justin Pinner';
-        this.description = 'Adblocking response with ad-free 404 test ZERO PERCENT';
-        this.audience = 0;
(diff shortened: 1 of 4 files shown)


static/src/javascripts/projects/common/modules/experiments/tests/adblocking-response.js
```diff
@@ -73,13 +73,13 @@ define([
             id: 'control',
             test: function(){}
         }, {
-            id: '299',
+            id: '300',
             test: function () {
                 if (isQualified()) {
```
desbo: You might want `isQualified` to be a part of the `canRun` check rather than the `test` function. As it is now, you'll track impressions for people that don't qualify, so your results for the variants with `isQualified` in test will be lower. 

If you move `isQualified` to `canRun`, you'll only fire impressions for people that do qualify, so your results won't have the negative bias that you'd see for the variants in their current form.

desbo: p.s. this is exciting ❗

desbo: After chatting in person, what you've got makes sense, there's no need to change this.

desbo (approved):

dominickendrick:
👍

PR #14484, opened 2016-09-28
give ContributionsStory it's correct name in ab-test-clash.js

jranks123:
## What does this change?

Fixes a bug whereby ContributionsStory was incorrectly named in ab-test-clash.js. The result of this bug is that we may not be compliant with OutBrain


gtrufitt:
👍

PR #14486, opened 2016-09-28
Hosted article - multiple inline videos tracking fixed

Calanthe:
## What does this change?

Fix for the multiple inline videos tracking issue (only the latest one was tracked correctly).

We can track all of the videos now.

no

## Screenshots

screen shot 2016-09-28 at 10 24 41
## Request for comment

@lps88 @akash1810

### DIFF
--- a/static/src/javascripts/projects/commercial/modules/hosted/video.js
+++ b/static/src/javascripts/projects/commercial/modules/hosted/video.js
@@ -116,7 +116,8 @@ define([
 
                         player.ready(function () {
                             var vol;
-                            duration = parseInt(this.duration(), 10);
+                            var player = this;
+                            duration = parseInt(player.duration(), 10);
                             initLoadingSpinner(player);
                             upgradeVideoPlayerAccessibility(player);
 


lps88:
👍

static/src/javascripts/projects/commercial/modules/hosted/video.js
```diff
@@ -116,7 +116,8 @@ define([
 
                         player.ready(function () {
                             var vol;
-                            duration = parseInt(this.duration(), 10);
+                            var player = this;
```
akash1810: I don't think this variable is needed as we're assigning it [above](https://github.com/guardian/frontend/pull/14486/files#diff-547d3fcebab67a2a4a843b1d1c287f52R113).

Calanthe: It is needed here, this is what was broken. In this scope, `player` points to the wrong object (always the last video), so `this` is the correct reference.

akash1810 (approved):
👍

PR #14490, opened 2016-09-28
Update facia-press to read cross account off the correct bucket

janua:
`Facia-tool` is currently writing to two account buckets, the `tools` account bucket and the`frontend` account bucket. 

The tools team are looking to switch off writing to the `frontend` bucket. `Facia-press` still reads off the `frontend` bucket, but should be reading off the `tools` team bucket.

This adds a switch to allow switching over `facia-press` to read cross account off the tools bucket.

@Reettaphant @piuccio

### DIFF
--- a/common/app/conf/switches/FaciaSwitches.scala
+++ b/common/app/conf/switches/FaciaSwitches.scala
@@ -65,4 +65,14 @@ trait FaciaSwitches {
     exposeClientSide = false
   )
 
+  val FaciaPressCrossAccountSwitch = Switch(
+    SwitchGroup.Facia,
+    "facia-press-cross-account",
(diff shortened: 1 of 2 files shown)


piuccio:
I thought this was already done in #12677

Keep in mind that facia press has to read from one bucket but write on another. You shuold own the `pressed.json`files

janua:
@piuccio You are correct. It needs to write to our bucket. But this is still reading the collections and config from the wrong location. Will re-open when I have time!

It seems it wasn't done in the other PR, as that only did the `ConfigAgent`.

janua:
@piuccio I have checked and with this PR the changes are valid; we will still be writing to one bucket and reading from another when this is switched on. Don't go.

piuccio (approved):
I trust you

PR #14491, opened 2016-09-28
Point Redis Report logs to ophan data-lake bucket

rich-nguyen:
Points the commercial reports to ophan data lake. Pipes open, logs start flowing through.


Calum-Campbell:
👍

PR #14494, opened 2016-09-28
Fix race condition with Facia healthcheck

TBonnin:
## What does this change?

Facia doesn't server fronts before the Facia config has been downloaded
(done by an agent) and return a 200 with X-Acccel-Redirect (internal
redirect) in this case.
Fetching the Healthcheck endpoint before the config was downloaded was
therefore returned with a 200 and considered successful, which allows
ELB to mark instances as healthy even though there might be issue with
underlying fronts rendering
This PR checks the config has been fetched before to return proper healthcheck result

Note:
I didn't go for a different endpoint specific to healthcheck or adding a condition in the logic of front rendering because it defeats the principle that code executed by the healthcheck should be identical to the one executed to reply to real-user requests. 
Also, right now we are testing only one endpoint (`/uk/business`) but we might several in the future and I didn't want to have to modify multiple endpoints then.
Last, it keeps the additional logic self contained with the healthcheck code

Facia healtcheck is actually working and Facia instances are not marked as healthy and added in the rotation if there is an underlying problem with front rendering
## Request for comment

@jfsoul @janua

### DIFF
--- a/common/app/services/ConfigAgentTrait.scala
+++ b/common/app/services/ConfigAgentTrait.scala
@@ -23,6 +23,8 @@ trait ConfigAgentTrait extends ExecutionContexts with Logging {
   implicit val alterTimeout: Timeout = Configuration.faciatool.configBeforePressTimeout.millis
   private lazy val configAgent = AkkaAgent[Option[Config]](None)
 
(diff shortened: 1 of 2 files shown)


jfsoul (approved):
👍

PR #14495, opened 2016-09-28
Fix for sonobi in prebid environment

rich-nguyen:
Add a check to the `dfp env` to make sure that prebid is fully disabled when sonobi is active.


janua:
:+1:

PR #14499, opened 2016-09-28
Native capi multiple endpoint

Ap0c:
## What does this change?
- Extends the work done in #14473.
- Modifies the cAPI offers json endpoint to include the cAPI multiple data.
- Adds image source data to endpoint for both cAPI single and multiple.
- Adds new route for cAPI multiple endpoint.

This is needed for work being done on the [cAPI native templates](https://github.com/guardian/commercial-templates). It provides data for the cAPI multiple template, and image information for both the cAPI single and multiple.
## Request for comment

@regiskuckaertz @JonNorman, @annebyrne I'm afraid I had to modify some of your nice functional code in the controller, and replace it with some very boring imperative stuff to get access to the multiple data. `headOption` was unfortunately stripping out the additional content. If I've made a mess or if you think there's a better way of doing this please let me know! 😔


commercial/app/controllers/commercial/ContentApiOffersController.scala
```diff
@@ -123,17 +123,28 @@ class ContentApiOffersController(contentApiClient: ContentApiClient, capiAgent:
 
   private def renderNative(format: Format, isMulti: Boolean) = Action.async { implicit request =>
```
janua: What's this `format` used for? I know it's not apart of the PR, but maybe it's time to remove it.

commercial/app/controllers/commercial/ContentApiOffersController.scala
```diff
@@ -123,17 +123,28 @@ class ContentApiOffersController(contentApiClient: ContentApiClient, capiAgent:
...
-        case Some(content) => {
-          val capiSingle = CapiSingle.fromContent(content)
-          Cached(60.seconds) {
-            JsonComponent(capiSingle)
-          }
+    retrieveContent().map(result => {
+
+      val content = result.toList
+
+      if (!content.isEmpty) {
```
janua: You should put all of this into a pattern match.

commercial/app/model/commercial/CapiImages.scala
```diff
@@ -0,0 +1,58 @@
...
+        breakpointWidth.breakpoint.minWidth.getOrElse("0").toString,
+        breakpointWidth.width.toString,
+        ImgSrc.srcsetForBreakpoint(breakpointWidth, breakpointWidths, None,
+          imageData, hidpi = true),
+        ImgSrc.srcsetForBreakpoint(breakpointWidth, breakpointWidths, None,
+          imageData)
+      )
+    }
+
+    ImageInfo(sources, fallbackImageUrl.getOrElse(""))
```
janua: Do you really want an empty string here for the `backupSrc`? Maybe it would be better to just not have it by making it `Option`.

commercial/conf/routes
```diff
@@ -38,6 +38,7 @@ GET         /commercial/capi.json                                         contro
 GET         /commercial/capi                                              controllers.commercial.ContentApiOffersController.itemsHtml
 GET         /commercial/capi-single.json                                  controllers.commercial.ContentApiOffersController.itemJson
 GET         /commercial/api/capi-single                                   controllers.commercial.ContentApiOffersController.nativeJson
+GET         /commercial/api/capi-multiple                                 controllers.commercial.ContentApiOffersController.nativeJsonMulti
```
regiskuckaertz: Please do add this route to `dev-build` as well

commercial/app/controllers/commercial/ContentApiOffersController.scala
```diff
@@ -123,17 +123,28 @@ class ContentApiOffersController(contentApiClient: ContentApiClient, capiAgent:
...
+
+        val capiData = if (isMulti) {
+          JsonComponent(CapiMultiple.fromContent(content))
+        } else {
+          JsonComponent(CapiSingle.fromContent(content.head))
```
JonNorman: You could do something like
```
content.headOption map { head =>
    Cached(60.seconds, 
        if isMulti
            JsonComponent(CapiMultiple.fromContent(content))
        else
            JsonComponent(CapiSingle.fromContent(head)))
    )
} getOrElse (Cached(componentNilMaxAge, jsonFormat.nilResult))
```

commercial/app/model/commercial/CapiImages.scala
```diff
@@ -0,0 +1,58 @@
...
+object CapiImages {
+
+  // Puts together image source info using data from cAPI.
+  def buildImageData(imageData: Option[ImageMedia]): ImageInfo = {
+
+    val fallbackImageUrl = imageData flatMap ImgSrc.getFallbackUrl
+    val cardType = Standard
+
+    val breakpointWidths = FaciaWidths.mediaFromItemClasses(ItemClasses(
+      mobile = Standard,
```
JonNorman: You can use `cardType` here

JonNorman:
I think we might be able to keep the functional stuff in `ContentApiOffersController.scala` but apart from that and the comments from @janua, this looks good to me.

commercial/conf/routes
Ap0c: Shall I add the new route for single too?

commercial/app/controllers/commercial/ContentApiOffersController.scala
```diff
@@ -123,17 +123,22 @@ class ContentApiOffersController(contentApiClient: ContentApiClient, capiAgent:
...
 
-    retrieveContent().map(_.toList.headOption).map(
-      {
-        case None => Cached(componentNilMaxAge){ jsonFormat.nilResult }
-        case Some(content) => {
-          val capiSingle = CapiSingle.fromContent(content)
-          Cached(60.seconds) {
-            JsonComponent(capiSingle)
-          }
+    retrieveContent().map {content =>
```
janua: I was thinking more
```scala
retrieveContent().map {
  case Nil                => Cached(componentNilMaxAge){ jsonFormat.nilResult }
  case content if isMulti => Cached(60.seconds) {JsonComponent(CapiMultiple.fromContent(content))}
  case first :: _         => Cached(60.seconds) {JsonComponent(CapiSingle.fromContent(head))}   
}
```

`isMulti` is weird, where has it come from? You know if it is single or multi from the list...

annebyrne: IsMulti is a hangover from the previous methods. I suppose we could remove them now following all the changes

janua: Then you could have this without `isMulti`:

```scala
retrieveContent().map {
  case Nil          => Cached(componentNilMaxAge){ jsonFormat.nilResult }
  case first :: Nil => Cached(60.seconds) {JsonComponent(CapiSingle.fromContent(head))}   
  case content      => Cached(60.seconds) {JsonComponent(CapiMultiple.fromContent(content))}
}
```

Ap0c: The list is up to 4 regardless is it not? It initially looks up eight `lookup.latestContentByKeyword(keyword, 8)`, and then trims that to four `(specific ++ latestByKeyword.filter(_.trail.trailPicture.nonEmpty)).distinct take 4`.

So do we not need `isMulti` to decide whether to take all or just the head?

Ap0c: Updated to the pattern match in bf6af4f.

commercial/app/model/commercial/CapiImages.scala
Ap0c: Done a5b8d45.

commercial/conf/routes
Ap0c: Added both in 474d288.

commercial/app/controllers/commercial/ContentApiOffersController.scala
janua: You might need to convert to a `List` if it doesn't compile; `.map(_.toList)`

commercial/app/model/commercial/CapiImages.scala
Ap0c: Hmm, looking at that you've made me wonder whether I hardcoded something while I was building something else and then forgot about it. I'll double-check.

commercial/app/controllers/commercial/ContentApiOffersController.scala
Ap0c: It compiles ok 😄

commercial/app/model/commercial/CapiImages.scala
Ap0c: Turns out something I thought we'd be able to do client-side was actually a bit more complicated.

The image data depends on the number of cards in a couple different ways (didn't realise this before as it's buried quite deeply). Should only affect the occasions when there are fewer than 4 cards available, but need to handle it regardless: 47ac760.

commercial/conf/routes
```diff
@@ -34,15 +34,16 @@ GET         /commercial/liveevents/event.json                               cont
...
-GET         /commercial/capi-single-merch.json                              controllers.commercial.ContentApiOffersController.itemJson
-GET         /commercial/capi-single-merch                                   controllers.commercial.ContentApiOffersController.itemHtml
-GET         /commercial/paid.json                                           controllers.commercial.PaidContentCardController.cardJson
-GET         /commercial/paid                                                controllers.commercial.PaidContentCardController.cardHtml
+GET         /commercial/capi.json                                         controllers.commercial.ContentApiOffersController.itemsJson
+GET         /commercial/capi                                              controllers.commercial.ContentApiOffersController.itemsHtml
+GET         /commercial/capi-single.json                                  controllers.commercial.ContentApiOffersController.itemJson
+GET         /commercial/api/capi-single                                   controllers.commercial.ContentApiOffersController.nativeJson
```
annebyrne: The reason that the endpoints in PROD aren't working is because routes that hit the API router (which allow query params), require .json.

This means that for both CAPI single / multiple the endpoint needs to be changed to /commercial/api/capi-single.json to work in PROD.

annebyrne (changes requested):
The endpoints need to be changed for both cards to allow it to function in PROD.

commercial/app/controllers/commercial/ContentApiOffersController.scala
Ap0c: I've been asked not to tell you that you were right again, but... ba38517.

commercial/conf/routes
Ap0c: Ok, changes made for both 4b0d6b4.

annebyrne (approved):

PR #14500, opened 2016-09-28
Revert "Targeting Client"

itsibitzi:
Reverts guardian/frontend#14163

Going to add some config updates to the same branch rather than have two deploys.


dominickendrick:
👍

johnduffell:
😱 I was all excited it was going out in the end!

PR #14502, opened 2016-09-28
Show main images on audio page

MatthewJWalls:
## What does this change?

I have _massive_ misgivings about the content of this PR, but I thought it'd create it anyway as a way to get feedback.

This change is to begin showing main pictures on Audio pages. 

CAPI Now supply that image (conditionally) on audio content, so in theory this is just a template change to get it to show up.

Here is an audio item that has the main image enabled on the CAPI output: https://www.theguardian.com/science/audio/2016/sep/13/the-nature-of-intelligence-science-weekly-podcast

Now, some background.

The scala-capi-client library that we use is what talks to CAPI, consumes it's output, and spits us out an ItemResponse. Upon this ItemResponse is a Content object, and that Content object has various properties such as tags, fields, blocks and elements.

I am almost sure that the way it's SUPPOSED to work is that the scala capi client should give us the images on the ItemResponse.Content.elements property (this is, for example, how Video media elements work).

However, the scala client does not give us this, it only appears on the _blocks_ property of the Content class, not elements, which we actually something we've never had to use before. I talked to the CAPI people and they were adamant that I need to pick the images off of the Blocks property.

So if you look at my code you'll see I had to start passing through the Blocks element down to the Audio class, and introduce a weird one-off getMainSectionImages val. This feels pretty wonky but technically it works so I'll be interested in the feedback on this.

As another aside, had to put in a toggle for captions on the ImageFigure since they don't want it being shown, despite the image data actually having a caption. I could have fudged the ImageAsset classes to do this, but I decided since it was a presentation concern that the logic should be at the presentation level.

Shows images on media pages

May affect AMP - need to confirm.

### DIFF
--- a/applications/app/controllers/MediaController.scala
+++ b/applications/app/controllers/MediaController.scala
@@ -37,7 +37,6 @@ class MediaController(contentApiClient: ContentApiClient) extends Controller wit
     val response: Future[ItemResponse] = contentApiClient.getResponse(
       contentApiClient.item(path, edition)
         .showFields("all")
-        .showBlocks("main")
     )
 
     val result = response map { response =>
--- a/applications/app/controllers/MediaController.scala
+++ b/applications/app/controllers/MediaController.scala
@@ -37,6 +37,7 @@ class MediaController(contentApiClient: ContentApiClient) extends Controller wit
     val response: Future[ItemResponse] = contentApiClient.getResponse(
       contentApiClient.item(path, edition)
         .showFields("all")
+        .showBlocks("main")
     )
 
     val result = response map { response =>
--- a/common/app/model/content.scala
+++ b/common/app/model/content.scala
@@ -2,6 +2,7 @@ package model
 
 import java.net.URL
 
+import com.gu.contentapi.client.model.v1.{AssetType, Blocks, ElementType}
 import com.gu.contentapi.client.model.{v1 => contentapi}
 import com.gu.facia.api.{utils => fapiutils}
 import com.gu.facia.client.models.TrailMetaData
@@ -67,7 +68,8 @@ final case class Content(
   showByline: Boolean,
   hasStoryPackage: Boolean,
   rawOpenGraphImage: String,
-  showFooterContainers: Boolean = false
+  showFooterContainers: Boolean = false,
+  blocks: Option[Blocks] = None
 ) {
 
   lazy val isBlog: Boolean = tags.blogs.nonEmpty
@@ -346,6 +348,7 @@ object Content {
     val references: Map[String,String] = apiContent.references.map(ref => (ref.`type`, Reference.split(ref.id)._2)).toMap
     val cardStyle: fapiutils.CardStyle = fapiutils.CardStyle(apiContent, TrailMetaData.empty)
 
+
     Content(
       trail = trail,
       metadata = metadata,
@@ -389,7 +392,8 @@ object Content {
           .orElse(elements.mainPicture.flatMap(_.images.largestImageUrl))
           .orElse(trail.trailPicture.flatMap(_.largestImageUrl))
           .getOrElse(Configuration.images.fallbackLogo)
-      }
+      },
+      blocks = apiContent.blocks
     )
   }
 }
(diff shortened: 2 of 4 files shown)


paperboyo:
Yeah, something may be 🐟y. I'm dim, so I don't understand what you wrote above, but looking at CAPI responses for all other content types - main media is consistently in a different place: `response>content>elements>type: image; relation: main` and, slightly differently for video: `response>content>elements>type: image`. It's _never_ under `response>content>blocks>elements>type:image` like in the audio content you've linked above.

Why main media images are not of `relation: main` for video and audio and, more importantly, why audio has them under `blocks`? Composer team should know, I think: @na7hangood?

This is the flexible-content PR that introduced main media for audio content: [INTERNAL RESOURCE](https://github.com/guardian/flexible-content/pull/2625)

common/app/model/content.scala
```diff
@@ -66,7 +67,8 @@ final case class Content(
   showByline: Boolean,
   hasStoryPackage: Boolean,
   rawOpenGraphImage: String,
-  showFooterContainers: Boolean = false
+  showFooterContainers: Boolean = false,
+  blocks: Option[Blocks] = None
```
dominickendrick: @johnduffell Do you have any insight into `blocks` from the work you did on liveblogs ? Does this seem a logical use to you ?

na7hangood:
As far as I'm aware - the main block of both video and audio consists of either video encodings + images, or the audio file and images.
http://content.guardianapis.com/us-news/video/2016/sep/29/clinton-trump-colorado-us-election-anywhere-but-washington?format=json&api-key=teleporter-view&show-blocks=all&show-elements=all&show-rights=all&show-fields=all&show-tags=all&show-references=all
http://content.guardianapis.com/science/audio/2016/sep/13/the-nature-of-intelligence-science-weekly-podcast?format=json&api-key=teleporter-view&show-blocks=all&show-elements=all&show-rights=all&show-fields=all&show-tags=all&show-references=all
I think this does follow logically

applications/app/controllers/MediaController.scala
```diff
@@ -37,6 +37,7 @@ class MediaController(contentApiClient: ContentApiClient) extends Controller wit
     val response: Future[ItemResponse] = contentApiClient.getResponse(
       contentApiClient.item(path, edition)
         .showFields("all")
+        .showBlocks("main")
```
johnduffell: what are main blocks and how big are they?  I only dealt with body blocks before, and the problem there is we would fetch the body twice, once in the  blocks and once in the body field.  My worry here is you are adding more data to every response (for media) so we need to think about cost and whether we need to scale things up.

For live blogs (which have massive bodys) I got around that with a horrible hack that if we're really sure it is a live blog we only get the blocks, otherwise we only get the body.  This does mean some old live blogs don't really format properly as it now tries to render them as articles.

common/app/model/content.scala
```diff
@@ -386,7 +389,8 @@ object Content {
           .orElse(elements.mainPicture.flatMap(_.images.largestImageUrl))
           .orElse(trail.trailPicture.flatMap(_.largestImageUrl))
           .getOrElse(Configuration.images.fallbackLogo)
-      }
+      },
+      blocks = apiContent.blocks
```
johnduffell: for live blogs I converted the low level capi blocks into something more useful for live blogs here: https://github.com/guardian/frontend/blob/155f738f5b9e22cf8c894bc44527966e630a10cf/common/app/model/liveblog/BodyBlock.scala#L13

I'm not sure what the usual practice is regarding carring over capi models into the high level model, I think normally we rejig into a condensed format for rendering at this point.  Maybe canvas opinion on that one a bit.

MatthewJWalls:
Ok let me try and show it visually. Here's me debugging the ItemResponse for a *video element*

video debug

See the assets highlighted? There's 9 of them; a few video assets, and a few image assets.

Since CAPI now puts image assets in the CAPI JSON to be like video elements, I'd expect it to be the same. But lets try debugging the new Audio content:

audio debug

Only one asset: the audio, despite the fact that the CAPI content now looks like Video does

mchv:
@MatthewJWalls @LATaylor-guardian I think this is a bug in Content API. You should find the main image in elements

MatthewJWalls:
I've rebased/force pushed changes to this that remove my Blocks hack, now that we have proper images coming in from CAPI on the elements.

common/app/model/content.scala
```diff
@@ -346,6 +346,7 @@ object Content {
     val references: Map[String,String] = apiContent.references.map(ref => (ref.`type`, Reference.split(ref.id)._2)).toMap
     val cardStyle: fapiutils.CardStyle = fapiutils.CardStyle(apiContent, TrailMetaData.empty)
 
+
```
johnduffell: 😜

dominickendrick:
Great stuff ! 👍

PR #14503, opened 2016-09-28
Converting isLoggedIn dimension from boolean to string for consistency in reporting

dominickendrick:
## What does this change?

Converting `isLoggedIn` dimension from `Boolean` to `String` for consistency in reporting from Google Analytics

Clarity in reporting 🔮 

Google Anayltics

## Screenshots

`cd16` stands for `dimension16`
**before** 😢 
screen shot 2016-09-28 at 12 38 30
screen shot 2016-09-28 at 12 38 08

**after** 😀 
screen shot 2016-09-28 at 17 00 48
screen shot 2016-09-28 at 12 33 31
## Request for comment

@mkopka @guardian/dotcom-platform


common/app/views/fragments/analytics/google.scala.html
```diff
@@ -26,7 +26,7 @@
     if (guardian.config.user) {
         identityId = guardian.config.user.id;
     }
-    var isLoggedIn = !!identityId;
+    var isLoggedIn = (!!identityId).toString();
```
TBonnin: Will all booleans need to be reported as string?

dominickendrick: yes, but all the others are already `string`s

rjmk:
Seems crazy that GA only displays booleans as 0/1! Of course we should make this change so that we're consistent, seems a shame to lose the constraint on the possible values

dominickendrick:
@rjmk 
> Seems crazy that GA only displays booleans as 0/1! Of course we should make this change so that we're consistent, seems a shame to lose the constraint on the possible values

They are converted in the GA `js` to `1|0`

rjmk:
:(

👍

PR #14508, opened 2016-09-29
Refactor logo templates

kelvin-chappell:
This is an attempt to reduce the complexity of logo templates and isolate future problems a bit more.  Although a change to a css class will still affect them all.

/cc @Calanthe


common/app/views/fragments/commercial/cardLogo.scala.html
```diff
@@ -0,0 +1,10 @@
+@(branding: common.commercial.Branding, isStandardSizeCard: Boolean)(implicit request: RequestHeader)
+
+<div class="badge @if(!isStandardSizeCard) {badge--branded}">
+    @branding.label
+    <a class="badge__link" href="@branding.sponsorLink" data-link-name="logo link">
+        <img src="@branding.sponsorLogo.url"
+             alt="@{branding.sponsorName}'s logo"
```
regiskuckaertz: You can remove the `'s logo` part. The point of an alternative text is not to describe the image but to describe _the content_ of the image.

common/app/views/fragments/commercial/containerLogo.scala.html
```diff
@@ -0,0 +1,14 @@
+@(branding: common.commercial.Branding, isOnTheLeft: Boolean)(implicit request: RequestHeader)
+@import common.commercial.PaidContent
+
+<div class="badge @if(isOnTheLeft){badge--alt}">
+    @branding.label
+    <a class="badge__link" href="@branding.sponsorLink" data-link-name="logo link">
+        <img src="@branding.sponsorLogo.url"
```
regiskuckaertz: Shouldn't there be an AMP variant?

common/app/views/fragments/commercial/cardLogo.scala.html
kelvin-chappell: I see.  I always thought alt was to tell you what you were missing if you couldn't see the image.  I'll change that.

common/app/views/fragments/commercial/containerLogo.scala.html
kelvin-chappell: I thought amp was just for content pages.  But I'll add the amp variant.  So I suppose it should go back on the card logo as well.

PR #14509, opened 2016-09-29
Update Heroic switch name and extend expiry date

SiAdcock:
## What does this change?

Push the switch expiry to the week after the US elections, as it will continue to be used for Explore articles covering the election.

I've also renamed the switch to better suit its purpose.
## Request for comment

@gtrufitt @blongden73


gtrufitt:
👍 🚢 🚢 🚢

PR #14497, opened 2016-09-28
Hosted Article fix

lps88:
## What does this change?

The **More from** container shouldn't be on the page if there are no other pages in the series.
## Screenshots

image
## Request for comment

@guardian/labs-beta

PR #14468, opened 2016-09-27
Make onward component for hosted content

kelvin-chappell:
This provides a new endpoint for onward components on hosted pages.
The container is wrapped in json because it's designed to be called from the clientside.

Still to do in another PR:
- Make call from hosted page to insert onward container
- Remove existing onward components, apart from on hardcoded pages

/cc @guardian/labs-beta @DiegoVazquezNanini

PR #14510, opened 2016-09-29
Hosted Article fix

lps88:
## What does this change?

extract article onward journey to separate class, as we use it twice

a fix I put in yesterday (https://github.com/guardian/frontend/pull/14497) was accidentally only applied in one of the onward journey positions

@guardian/labs-beta


kelvin-chappell:
👍 
I'll need to get rid of the new template that's coming in via PR #14468

PR #14518, opened 2016-09-29
Native blended JSON endpoint

regiskuckaertz:
A lot of moving things around for essentially abstracting the data extraction part of the Blended component.

picture 4

cc @janua @JonNorman (thanks for your help, guys!)


commercial/app/controllers/commercial/Multi.scala
```diff
@@ -21,94 +24,95 @@ class Multi(bestsellersAgent: BestsellersAgent,
...
+      case ("Travel", None)           => Future.successful(travelOffersAgent.offersTargetedAt(segment))
+      case _                          => Future.successful(Nil)
+    }
 
-    val componentsAndSpecificIds = requestedContent zip slotIds
+    Future.sequence(samples) map { realSamples =>
+        realSamples.map(_.headOption)
+    } map { realSamples =>
+        realSamples.flatten
+    }
```
TBonnin: Would 
```
Future.sequence(samples) map { sample =>
   sample.map(_.headOption).flatten
}
```
lead to the same result? 😁

commercial/app/controllers/commercial/Multi.scala
```diff
@@ -21,94 +24,95 @@ class Multi(bestsellersAgent: BestsellersAgent,
...
+
+    val offerIds = request.getParameters("offerIds").map { slotId => if (slotId.trim.isEmpty) None else Some(slotId) }
+
+    val eventualContents = multiSample(offerTypes, offerIds, segment)
+
+
+    eventualContents map { contents =>
+      if (offerTypes.size == contents.size) {
+        Cached(componentMaxAge) {
+          JsonComponent(JsArray((offerTypes zip contents).map({ case (contentType, content) => Json.obj(
```
TBonnin: You can leave the parenthesis out if you have curly bracket, ie: `.map {`

commercial/app/model/commercial/events/LiveEventMembershipInfo.scala
```diff
@@ -0,0 +1,22 @@
...
+
+object LiveEventMembershipInfo {
+  implicit val format = Json.format[LiveEventMembershipInfo]
+
+  // based on play.api.libs.json.LowPriorityDefaultReads.traversableReads
+  implicit val formats = new Reads[Seq[LiveEventMembershipInfo]] {
+    override def reads(json: JsValue): JsResult[Seq[LiveEventMembershipInfo]] = {
+      json match {
+        case JsArray(jsValues) => JsSuccess(jsValues.flatMap(_.asOpt[LiveEventMembershipInfo]))
+        case _ => JsError(Seq(JsPath() -> Seq(ValidationError("error.expected.jsarray containing LiveEventMembershipInfo"))))
```
TBonnin: I am not sure I understand why having a custom pattern matching is necessary?

TBonnin (changes requested):

commercial/app/controllers/commercial/Multi.scala
janua: `.flatMap(_.headOption)`.

commercial/app/controllers/commercial/Multi.scala
```diff
@@ -21,94 +24,95 @@ class Multi(bestsellersAgent: BestsellersAgent,
...
+    val eventualContents = multiSample(requestedContent, slotIds, segment)
+
+    eventualContents map { contents =>
+      val content = contents map {
+        case b: Book        => views.html.books.booksBlended(b, clickMacro)
+        case j: Job         => views.html.jobs.jobsBlended(j, clickMacro)
+        case m: Masterclass => views.html.masterclasses.masterclassesBlended(m, clickMacro)
+        case p: MemberPair  => views.html.soulmates.soulmatesBlended(Random.shuffle(Seq(p.member1, p.member2)), clickMacro)
+        case t: TravelOffer => views.html.travel.travelBlended(t, clickMacro)
+        case _              => Html("")
```
janua: You can delete this case. If you used an ADT, you will never get a different type as it's impossible. And if you add a new type to the ADT it will warn you here that you haven't yet covered the case for the new type.

commercial/app/controllers/commercial/Multi.scala
```diff
@@ -21,94 +24,95 @@ class Multi(bestsellersAgent: BestsellersAgent,
...
-    val componentsAndSpecificIds = requestedContent zip slotIds
+    Future.sequence(samples) map { realSamples =>
+        realSamples.map(_.headOption)
+    } map { realSamples =>
+        realSamples.flatten
+    }
+  }
+
+  def renderMulti() = Action.async { implicit request =>
+    val requestedContent = request.getParameters("components").map { component => component match {
```
janua: You can do a direct `case` in here:
```scala
.map {
 case "books" =>
}
```

commercial/app/controllers/commercial/Multi.scala
```diff
@@ -21,94 +24,95 @@ class Multi(bestsellersAgent: BestsellersAgent,
...
+        case "books"  => "Book"
+        case "jobs"   => "Job"
+        case "travel" => "Travel"
+        case "masterclasses" => "Masterclass"
+        case "soulmates" => "Soulmates"
+        case _        => ""
+    }}
+
+    val slotIds = request.getParameters("slotIds").map { slotId => if (slotId.trim.isEmpty) None else Some(slotId) }
```
janua: Isn't this type going to be `Option[Option[String]]`? Is that what you want?

janua: Ah, my guess is that it's `Seq[Option[String]]`

commercial/app/model/commercial/events/LiveEventMembershipInfo.scala
JonNorman: This is a pattern we've used before to tolerate the presence of a malformed element.
1. We expect it to be a `JsArray`, if it isn't then error out
2. If it is a JsArray then try to parse each element - either we get a `Some[LiveEventMembershipInfo]` or we get a `None` at which point we can flatMap it.

commercial/app/model/commercial/events/LiveEventMembershipInfo.scala
```diff
@@ -0,0 +1,22 @@
...
+
+case class LiveEventMembershipInfo(id: String,
+                                   url: String,
+                                   mainImageUrl: String)
+
+object LiveEventMembershipInfo {
+  implicit val format = Json.format[LiveEventMembershipInfo]
+
+  // based on play.api.libs.json.LowPriorityDefaultReads.traversableReads
+  implicit val formats = new Reads[Seq[LiveEventMembershipInfo]] {
```
JonNorman: This is entirely my fault, but this should be called `readsLiveEventMembershipInfo`, not `formats` as it isn't a `Formats` 😕

commercial/app/model/commercial/Merchandise.scala
```diff
@@ -0,0 +1,370 @@
...
+  }
+
+  val profileUrl: String = profileId.map(id => s"https://soulmates.theguardian.com/landing/$id")
+    .getOrElse("http://soulmates.theguardian.com/")
+
+}
+
+case class MemberPair(member1: Member, member2: Member) extends Merchandise
+
+sealed case class Gender(name: String) {
```
janua: Can we move this to the location of the things that `extend` it?

Also it's worth considering why you need `name` here as a parameter; I think you probably don't need it. You might even just be able to get away with `sealed trait Gender`, because you usually know the type if you are matching it out, unless you are using this base type directly and don't care if it's `Woman` or `Man`.

commercial/app/model/commercial/Merchandise.scala
```diff
@@ -0,0 +1,370 @@
...
+  override def toString = name
```
janua: `toString` :scream:

commercial/app/controllers/commercial/Multi.scala
JonNorman: I think there is actually a type that isn't implemented at the moment (`LiveEvent`), hence the `_`. Given all the hard work @regiskuckaertz has done to let the compiler do its job, it might be worthwhile adding them in explicitly here, where each can still return `Html("")`...

commercial/app/model/commercial/Merchandise.scala
```diff
@@ -0,0 +1,370 @@
...
+  implicit val bookReads: Reads[Book] = (
+    (JsPath \ "name").read[String] and
+      authorReads and
+      (JsPath \ "isbn").read[String] and
+      stringOrDoubleAsDouble("regular_price_with_tax") and
+      stringOrDoubleAsDouble("final_price_with_tax") and
+      (JsPath \ "description").readNullable[String] and
+      (JsPath \ "images")(0).readNullable[String] and
+      (JsPath \ "product_url").readNullable[String] and
+      (JsPath \ "guardian_bestseller_rank").readNullable[String].map(_.map(_.toDouble.toInt)) and
```
JonNorman: Appreciate you didn't write this so maybe I can pick this up in a different PR, but it looks  a bit funny reading as a `String`, mapping it twice to a `Double` and then to an `Int`... Are we aiming to round down the ranking or are we just going really round the houses to read an `Int`? Might be worth checking the feed to see what is going on.

commercial/app/model/commercial/Merchandise.scala
```diff
@@ -0,0 +1,370 @@
...
+
+object TravelOffer {
+
+  def fromXml(xml: Node): TravelOffer = {
+
+    def text(nodeSelector: String): String = (xml \ nodeSelector).text.trim
+    def tagText(group: String): Seq[String] = (xml \\ "tag").filter(hasGroupAttr(_, group)).map(_.text.trim)
+
+    def parseInt(s: String) = try {
+      Some(s.toInt)
```
janua: `scala.util.Try` does exactly this, and it useful to avoid using the low level `try` keyword;

```scala
   Try(s.toInt).toOption
```

It only catches `NonFatal`.

commercial/app/model/commercial/Merchandise.scala
```diff
@@ -0,0 +1,370 @@
...
+      eventUrl = eventMembershipInformation.url,
+      description = event.description,
+      status = event.status,
+      venue = event.venue,
+      tickets = event.tickets,
+      imageUrl = eventMembershipInformation.mainImageUrl
+    )
+
+  implicit val writesLiveEvent: Writes[LiveEvent] = new Writes[LiveEvent] {
+    def writes(m: LiveEvent) = JsString("")
```
JonNorman: Maybe at least stick in a "//TODO" here 😂

commercial/app/model/commercial/events/LiveEventMembershipInfo.scala
janua: Can you not just try to cast it to the type you expect by using `json.validate[List[LiveEventMembershipInfo]]` ?
In fact if you do that you probably don't even need this `Reads`, you can just use the JSON macros.

commercial/app/controllers/commercial/Multi.scala
janua: `LiveEvent` will need to be added to the `ADT` for that to be exhaustive otherwise you will always need the `_`. You must not be matching on the `ADT` if this is compiling and there is a type outside of it.

janua:
You should type everything you can as it makes reading in github much easier! Even `String` or basic structures that might seem obvious; we can very quickly review things when there are types. (Although I am extreme when it comes to this, I even type `val b: Boolean = true`. :laughing: )

Another thing you might want to consider is to try using `Future.traverse` instead of `Future.sequence`, as lovely as `sequence` is, `traverse` performs better.

commercial/app/model/commercial/events/LiveEventMembershipInfo.scala
JonNorman: I'm not sure - it is definitely worth trying - but if I recall correctly, the macros have no tolerance for elements within the `JsArray` that are malformed and would throw a `JsError` only, hence this approach of mapping and manually transforming and flattening each one.

janua: Ah yes, if you want something more tolerant like that you probably want to do it yourself like this. You could just write a general tolerant `List` reads. There might even be one already!

commercial/app/controllers/commercial/Multi.scala
regiskuckaertz: `LiveEvent` and `Member`. Can I do `case _: LiveEvent | Member` ? I'll try

commercial/app/controllers/commercial/Multi.scala
regiskuckaertz: `Seq[Option[String]]` indeed.

regiskuckaertz: That is also why I need to change to pattern matching in `multiSample` so that the `Some(String)` falls back to the `None` case (now I remember why I needed `getOrElse`).

commercial/app/model/commercial/events/LiveEventMembershipInfo.scala
regiskuckaertz: ok, changing that.

commercial/app/model/commercial/Merchandise.scala
```diff
@@ -0,0 +1,370 @@
...
+    val path = JsPath \ value
+    path.readNullable[Double] orElse path.readNullable[String].map(_.map(_.toDouble))
+  }
+
+  implicit val bookReads: Reads[Book] = (
+    (JsPath \ "name").read[String] and
+      authorReads and
+      (JsPath \ "isbn").read[String] and
+      stringOrDoubleAsDouble("regular_price_with_tax") and
+      stringOrDoubleAsDouble("final_price_with_tax") and
```
regiskuckaertz: I don't know, this code was written long ago 😊

TBonnin (approved):

regiskuckaertz:
@janua `Future.traverse` done, me likey, thank you!

commercial/app/model/commercial/events/LiveEventMembershipInfo.scala
JonNorman: Sounds like a good idea - I think this crops up in at least two other places in frontend, would be good to extract the logic to a single Reads that can be implicitly scoped.

PR #14520, opened 2016-09-29
Batched commercial logging

rich-nguyen:
Things learnt this week: it is very bad to send performance reports to the data-lake, one page view at a time.

This PR stops that behaviour. It will log the performance reports as files, stored locally through logback. In a future PR, I'll make those files copy over to ophan's S3 bucket periodically, reducing the number of objects in the bucket.

Moving to a batched uploading of reports also removes the need to expire surrogate keys in Redis. There is no need for the pub/sub mechanism at all, so it's been removed.


gtrufitt:
LGTM 👍

TBonnin:
Would it make sense to make this whole reporting a independent Lambda (triggered on schedule) that read Redis and upload the generated report into the data lake?

rich-nguyen:
It concerns the delivery of adverts to web users. This job gets feedback about the performance of the commercial code in this frontend codebase. It could certainly be a lambda, again this goes back to whether people prefer seeing code in one or many repos. Either way, the functional dependencies would be the same. I do think this is a frontend concern.

TBonnin:
I do agree this is a frontend concern. I am just pointing out this is quite independent from delivering web content to the user and could be move to a Lambda.
I understand that, within Dotcom, there is no standardized way to create and easily maintain Lambda for running jobs like this and the more productive way to do so is to do exacatly what you did. 
However I would like to open a discussion in the near feature about managing those kind of asynchronous/batching jobs outside of this codebase. This is probably a task for dotcom-platform for the next quarter. Stayed tuned.

PR #14529, opened 2016-09-30
Blended: fallback values

regiskuckaertz:
In my previous PR, I mistakenly thought the `getOrElse` in `multiSample` would account for empty results. It would only if empty results were converted into `None`s, and `map` was replaced with `flatMap`.

Also, a few values needed to be changed in the Masterclass JSON writer.

cc @janua @JonNorman

### DIFF
--- a/commercial/app/controllers/commercial/Multi.scala
+++ b/commercial/app/controllers/commercial/Multi.scala
@@ -29,11 +29,9 @@ class Multi(bestsellersAgent: BestsellersAgent,
 
     val samples = Future.traverse(components) {
       case ("Book", Some(bookId))=>
-        bestsellersAgent.getSpecificBooks(Seq(bookId)) map { books =>
-          if(books.isEmpty)
-            bestsellersAgent.bestsellersTargetedAt(segment)
-          else
-            books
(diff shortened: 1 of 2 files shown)


commercial/app/controllers/commercial/Multi.scala
```diff
@@ -28,23 +28,30 @@ class Multi(bestsellersAgent: BestsellersAgent,
     val components = offerTypes zip offerIds
 
     val samples = Future.traverse(components) {
-      case ("Book", optId)        => optId.map { bookId =>
-        bestsellersAgent.getSpecificBooks(Seq(bookId))
-      }.getOrElse {
+      case ("Book", Some(bookId))=>
+        bestsellersAgent.getSpecificBooks(Seq(bookId)) map { books =>
+          if(books.isEmpty)
```
janua:
```scala
case Nil => bestsellersAgent.bestsellersTargetedAt(segment)
case books => books
```

janua:
Would be nice if it was indented more after the `=>`, as what appears after a `=>` doesn't need to be on the same line!
:+1:

janua:
:heart_decoration:

PR #14531, opened 2016-09-30
Use parameters to query soulmates API rather than path capturing.

JonNorman:
Yet more small Native Template changes.

The native template will now fetch soulmates via URL params, as opposed to the path. This will allow us to more easily expand the customised nature of the API call in future e.g. number of soulmates etc.


regiskuckaertz (approved):
Yihaa!

PR #14532, opened 2016-09-30
Bump config for switch to gu-web beacon urls

rich-nguyen:
This changes the config so that the beacon endpoint moves to `gu-web.net`. This means:

`beacon.www.code.dev-theguardian.com` becomes `beacon-code.gu-web.net` and
`beacon.guim.co.uk` becomes `beacon.gu-web.net`

The reason I'm doing this is so that I can move diagnostics into the VPC.


janua:
:+1:

PR #14533, opened 2016-09-30
Hosted youtube video - controls in the new template

Calanthe:
## What does this change?

Just a small change when controls are moved to the separate template, so the main one stays simple.

## Request for comment

@lps88 @akash1810


akash1810:
👍

PR #14535, opened 2016-09-30
Module loader nightmare take 3: rollup

piuccio:
## What does this change?

The brightest minds in this company have already tried to replace curl / requirejs with something better with little success. SystemJS / JSPM was a failure. Webpack is not going well. During my last day I thought it was a good idea to kill time trying rollup 🎉 

You can close the tab now if you think this is a joke. It gets boring.
### Javascript architecture

Currently we generate 3 root bundles: `boot`, `bootstraps/commercial`, `bootstraps/enhanced/main`. `boot` takes care of loading things, `commercial` is an independent feature that can be turned on or off, `enhanced/main` is a conditional script loader.
Based on things like page type (front / article), page id, section and so on, `enhanced/main` loads other bootstrap modules.

Bootstrap modules (like `facia`) may or may not load other bundles asynchronously at their will.

I'm going to call `leaf module` any bundle that doesn't load other bundles.
### AMD / RequireJS packaging

Al our files are written in AMD (`define([...deps])`) and bundled with `r.js` (part of `grunt-contrib-requirejs`). It works like this:

```
boot.js
- dep1.js
- dep2.js
   - dep2_1.js

facia.js
- dep1.js
- dep3.js
```

When bundling `boot.js`, `requirejs` finds all nested dependencies and generates a single file that looks like this:

``` js
define('dep2_1.js', [], function () {});
define('dep2.js', ['dep2_1.js'], function () {});
define('dep1.js', [], function () {});
define('boot.js', ['dep1.js', 'dep2.js'], function (d1, d2) {});
```

You can then bundle `facia.js` marking `boot.js` as external. `requirejs` will bundle together all `facia` dependencies excluding anything that was already included in `boot`. The resulting file is

``` js
define('dep3.js', [], function () {});
define('facia.js', ['dep1.js', 'dep3.js'], function (d1, d3) {});
// Note how `dep1.js` is not in the bundle because it lives in `boot.js`
```
### Enter rollup

The epiphany came when I realised that rollup generates AMD files as output. Roughly speaking one can:
1. Use a plugin to convert AMD to ES6 modules
2. Let rollup do the magic
3. Output an AMD bundle

This approach can be easily applied to `leaf modules` because from curl perspective they both look the same.

The internals however are quite different. This is what rollup generates for the sample`boot.js` from above

``` js
var dep2_1 = (function () {})();
var dep2 = (function () {})();
var dep1 = (function () {})();
define('boot.js', [], function () {});
```

and `facia.js`

``` js
var d3 = (function () {})();
define('facia.js', ['dep1.js'], function (d1) {});
```

In other words, dependencies are inlined in the module. This has a great potential of reducing the file size, because now we can get rid of lots of `define` calls and module names as strings.

This contrived example however shows the downside of rollup and why it might not work with root modules.
Because dependencies are inlined instead of `define`d, they are not added to the module registry. The next script that comes along and depends on `dep1` won't find it.

Nonetheless, leaves can be highly optimized.
### Pain points
- `rollup` externals work differently than `requirejs` in that they only exclude that module. I had to use `madge` to get the nested list of dependencies of our root modules. Even worse I'm using my forked version of `madge` that ignores lazy loaded dependencies. There's a PR open [here](https://github.com/pahen/madge/pull/105).
- Relative dependencies are hard. Even if `madge` internally uses `requirejs` to normalize paths, things like `./common` end up pointing to different files. I've rewritten one file that was causing troubles.
- Source maps are gone. The plugin that converts AMD to ES6 doesn't support them. Might not be missed.
- It's kinda hard to make sure files have the correct external dependencies. Took me a while. If you're trying to use rollup with other modules, run `DEBUG=rewire:external node rollup.js` it'll give you a list of modules that are not bundled. You can compare that to the occurrences of `define('moduleId',` in the bundle generated by requirejs

If you're still reading, you're probably asking yourself why this???
### File size

| requirejs | rollup |
| --- | --- |
| 7556 B | 7163 B |

*minified gzipped

Switching to `rollup` saves us ~0.5KB minified gzipped, that's about 5% of the requirejs file size.
### The future!

I've already included `rollup-plugin-buble` You can now use all the arrow functions you want! `buble` transpiles any ES6 down to ES5.
### Build time

Well this is arguable. On a single file grunt takes ~4s, while rollup takes ~6s, however `rollup` script can handle multiple files at the same time, while requirejs has to run in parallel processes. This means that requirejs is computing the nested dependencies of `boot`, `enhanced`, and `commercial` 21 times (one per bundle), while `rollup` can do it only once.

Who knows
## Screenshots

You should rather see me, I'm so tired
## Request for comment

FYI @Rich-Harris
## What next?
1. Probably rethink how we generate bundles. I have no idea on how to split root / leaf keeping shared files
2. Start using ES6 and imports. This would actually fix source maps
3. Figure out how to run tests when files are written in ES6 modules or include any ES6 feature


SiAdcock:
After running some benchmarks with rollup-bundled code, I've decided that there is enough justification to bring this branch up to date with `master`.

There are still a few more things we'll need to do to get standard bundling with rollup, should we decide to proceed:

- [ ] Some third party libraries we include in our codebase (qwery, bean, reqwest, bonzo, raven, EventEmitter) are written as UMD modules. We will need to convert these to AMD, or find an alternative way to include them.
- [ ] `rollup.js` as written is a demonstration of how we can bundle "leaf modules", while excluding dependencies that are bundled into the "root modules". This is great, but we will need define a new script that does the things we need rollup to do (e.g. bundle the standard module). 
- [ ] We'll need to find the appropriate place to slot this into our build process. Perhaps we could call it from [`bundle-standard.js`](https://github.com/guardian/frontend/blob/master/tools/assets/javascript/bundle-standard.js)

piuccio:
I've been thinking about core modules: 
- `standard/main` is sync except for ophan and interactives that however are not part of any bundle
- `commercial` is sync
- `enhanced/main` is async (hence the problem), however all it does is deciding which bundle to load

If we move the bundle selection logic to Scala (which kinda makes sense, you can use case matching) we can flatten the structure and inject the bundles that need loading inside the global `guardian` variable.

So `boot.js` instead of loading `enhanced/main` can call

```js
        if (guardian.isEnhanced) {
            return promiseRequire(guardian.enhancedBundles)
                .then(function (...boots) {
                    boots.forEach(_ => _());  // :grin:
                });
        }
```

With this approach everything becomes a leaf module, except for external modules that don't need bundling anyway.
The drawback is that shared dependencies between say commercial and enhanced gets duplicated. The solution would be to modify `rollup-plugin-amd` and allow some modules to be exported through `define` and mark them as external in the bundles.

paperboyo:
👋 from Poland, @piuccio!

SiAdcock:
Closing as it is unlikely we will use Rollup to bundle modules in the frontend

PR #14537, opened 2016-09-30
Add Edition to cAPI Endpoint for Native Templates

Ap0c:
## What does this change?
- Replaces placeholder edition used to generate branding in the endpoint.
- Passes edition information back to the client for use in cAPI native templates.

Needed to allow cAPI native templates to display edition and branding correctly.
## Request for comment

@annebyrne 
@regiskuckaertz decided against modifying `getIframeId` message passing; we were already using edition for branding in the endpoint (plus we only need it for these templates anyway) .


regiskuckaertz (approved):
Nice

PR #14538, opened 2016-09-30
Update references to beacon.guim.co.uk

rich-nguyen:
There were a number of hard-coded references to the beacon config. I've fixed all these up.


janua:
:+1:

PR #14539, opened 2016-09-30
Move some fluid styles to fabric

regiskuckaertz:
Last change before the end of the quarter 😆 

There were some overlap between the treatment of Fabrics and Fluid ads around the label. Fluid ads won't have label, so we can safely move that over to Fabrics only. Also, we seal the styling on Fabrics in its own modifier (`ad-slot--fabric`) to prevent future mixups.

Before (no label):

picture 2

After:

picture 1


janua:
:+1:

PR #14542, opened 2016-09-30
Display image with video icon correctly in emails

joelochlann:
Currently these clients don't support background images:
- all windows desktop versions of outlook
- gmail app (android)
- outlook.com

Unfortunately we use a background image to overlay a video "play" icon on top of an image as an enticement to click the link and see the video on theguardian.com. This doesn't look great in the above clients. So I've opted to use imgix magic to overlay the play icon on top of the main image and just serve up one image. This has other advantages, namely being able to center the play icon, and deleting some now unused markup & styles.
### Before

picture 304
### After

picture 306

@desbo


joelochlann:
cc @davidfurey @maxspencer in case you're interested!

desbo (approved):

PR #14544, opened 2016-10-03
Remove paddings on fluid ad slots

regiskuckaertz:
The top slot has a top padding to make space for the `Advertisement` label, which is positioned absolutely inside the ad slot box. This is handled by the parent frame.

Native/Fluids ads will take care of this themselves, inside the `iframe`. Thus, the space at the top becomes irrelevant. We discard it.


janua:
:+1:

PR #14545, opened 2016-10-03
Platform: Restore Jobs to Int'l edition header

gtrufitt:
## What does this change?

Re-adds jobs into the international header on desktop: 

image

Also, move all those links to https to avoid redirect.

@guardian/dotcom-platform


NataliaLKB:
👍

rjmk (commented):
👍

rjmk:
(That was supposed to be specifically on ab75c3b, but it looks like that didn't really read)

PR #14546, opened 2016-10-03
Removed all video button duration code

mr-mr:
## What does this change?

Strips out all the button duration test from PR https://github.com/guardian/frontend/pull/14378

cc @akash1810 @gidsg


akash1810:
👍

PR #14547, opened 2016-10-03
Add fluid inline merchandising

Calum-Campbell:
## What does this change?
- Allows fluid ad slots to be targeted at inline merchandising slots.
- Allows the migrated native manual inline template to be used in fluid advertising slots. 
- Manual inline migrated to native template! 
## Request for comment


regiskuckaertz (approved):
LGTM but please confirm these changes don't have any impact on regular inline merch components

PR #14519, opened 2016-09-29
New header styling

zeftilldeath:
Just when you thought I wouldn't make another pull request called new header styling...

Today's menu includes:
- Slightly smaller font size
- Contrasting font sizes
- Re-jigged the css structure to make it easier to add icons in the future/standardised the positioning rules
- Tweaked colours and spacing
- New icons added
- Search bar focus state

And much, much more.

b
a

screen shot 2016-09-30 at 14 30 27

@NataliaLKB @stephanfowler

PR #14548, opened 2016-10-03
Implementing tertiary nav for the new header test

NataliaLKB:
## What does this change?

This implements tertiary nav to the new header test.

This will allow super users (and some other users) to navigate to parts of the site that aren't in the nav. For example, match reports on football.

Nope!
## Screenshots

image
## Request for comment

@zeftilldeath @stephanfowler @SiAdcock

### DIFF
--- a/common/app/views/fragments/containers/facia_cards/container.scala.html
+++ b/common/app/views/fragments/containers/facia_cards/container.scala.html
@@ -91,11 +91,11 @@
 
                         <ul class="tertiary-navigation mobile-only">
 
-                            @NewNavigation.TertiaryNav.getTertiaryNavLinks(id).map { footballNavItem =>
+                            @NewNavigation.TertiaryNav.getTertiaryNavLinks(id).map { tertiaryLink =>
 
                                 <li class="tertiary-navigation__item">
-                                    <a class="tertiary-navigation__link" href="@LinkTo(footballNavItem.url)" data-link-name="@footballNavItem.name">
-                                        @Html(footballNavItem.name)
+                                    <a class="tertiary-navigation__link" href="@LinkTo(tertiaryLink.url)" data-link-name="@tertiaryLink.name">
+                                        @Html(tertiaryLink.name)
                                         <i class="tertiary-navigation__icon rounded-icon"></i>
                                     </a>
                                 </li>
--- a/common/app/views/fragments/containers/facia_cards/container.scala.html
+++ b/common/app/views/fragments/containers/facia_cards/container.scala.html
@@ -1,4 +1,5 @@
 @import common.commercial.ContainerModel
+@import common.{NewNavigation, LinkTo}
 @import layout.MetaDataHeader
 @import slices.{Dynamic, Fixed, MostPopular, NavList, NavMediaList, Video}
 @import views.html.commercial.containers.paidContainer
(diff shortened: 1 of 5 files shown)


static/src/stylesheets/layout/_ab-new-header-test-variant.scss
```diff
@@ -668,3 +668,75 @@ svg.main-menu__icon--social__svg {
 .no-text-transform {
     text-transform: none;
 }
+
+// Breathing space between header and content
+@include mq($until: tablet) {
```
NataliaLKB: This bit was removed accidentally in this PR: 
https://github.com/guardian/frontend/pull/14519/files#diff-a6a8d02d22743e9a6dc47e407c359fb6L677

Just adding it back.

common/app/views/fragments/containers/facia_cards/container.scala.html
```diff
@@ -79,6 +81,29 @@
...
+                @defining((
+                    containerDefinition.index == 0,
+                    frontId.getOrElse(pageId)
+                    )) { case (isFirstContainer, id) =>
+
+                    @if(mvt.ABNewHeaderVariant.isParticipating && isFirstContainer) {
+
+                        <ul class="tertiary-navigation mobile-only">
+
+                            @NewNavigation.TertiaryNav.getTertiaryNavLinks(id).map { footballNavItem =>
```
dominickendrick: Are these always going to be football related ? If not could `footballNavItem` be called `tertiaryNavItem` ?

common/app/views/fragments/containers/facia_cards/container.scala.html
```diff
@@ -79,6 +81,29 @@
...
+
+                                <li class="tertiary-navigation__item">
+                                    <a class="tertiary-navigation__link" href="@LinkTo(footballNavItem.url)" data-link-name="@footballNavItem.name">
+                                        @Html(footballNavItem.name)
+                                        <i class="tertiary-navigation__icon rounded-icon"></i>
```
dominickendrick: Could the icon be added as an `:after` on the `tertiary-navigation__link` instead of having its own empty element ?

dominickendrick:
👍 apart from comments ! Great stuff !

common/app/views/fragments/containers/facia_cards/container.scala.html
NataliaLKB: Yes! Thanks for spotting. That should definitely be something more generic. renamed to  `tertiaryLink`

dominickendrick: 🎉

common/app/views/fragments/containers/facia_cards/container.scala.html
NataliaLKB: Right now the `<i>` is actually the background and the `:before` is the icon. We would need a separate element then the `<a>` tag to ensure we could have a background and an icon. The link also needs to surround both the text and the icon.

dominickendrick: Ah ok !

paperboyo:
Hmm... Grumpy one has a general grumble: it’s kind of a pity, slightly, maybe, that these have to be presented outside of the nav proper. That way, they may brake the navigation flow if they are the intended final destination. They are also visually very disjointed from their desktop counterparts this way and from the main nav elements (+the look of their chevrons says “commercial” to me somehow). And will, probably, maybe, have, only natural, tendency to diverge more and more. We already have some other disjointed desktop-only navigation sprinkles here and there, eg:

nav sprinkles

I understand that they are designed this way because of our decision to make new nav elements be expand/collapse-only at the top level, but navigation-only at the second level (the other by-product of that decision is that to navigate to top-level element we have to repeat it as a first item of the second level which introduces confusion of its own: arts>culture home).

For a nav that separates expand/collapse and navigation functions consistently at all levels, see eg FT:

ft-anv

As my sick mind fantasises about fourth, fifth levels sometimes and because of my general allergy to sprinkles, I would urge a re-think (esp. because that’s the initial phase!). Of course, the model that FT adopted comes with its own set of problems. But the readers will get used to whatever we choose, provided we won’t be changing it every week.

Sorry for the length! Plus, I’m not really sure I have a clear idea of how best to sort this out :cry:

PR #14549, opened 2016-10-03
Extend expiry of Weekend Reading switches and JS

katebee:
## What does this change?

Extends the Weekend Reading promo and AB test until the end of the month.
- Keeps the tests running 
- Build doesn't break tomorrow.

Nope
## Request for comment

@joelochlann


dominickendrick:
👍 Would be good to see the test results too

PR #14550, opened 2016-10-03
Fix save for later for crosswords

rjmk:
## What does this change?

The `pageId` for crossword pages

The `pageId` is normally the path part of the URL. For crossword pages, it was missing the "crosswords" prefix. This meant that when saving crosswords for later (while logged in), the resulting saved entry would not appear in the list of saved items.

Nope
## Request for comment

@guardian/dotcom-platform


dominickendrick:
👍 👊

PR #14551, opened 2016-10-03
Hosted gallery - get content from CAPI

lps88:
## What does this change?

We've only ever used hard-coded content for hosted galleries, so this Dymocks one is the first served from CAPI/composer!

I've switched off the automatic cropping of landscape images (they should all be 5:3 ratio normally, but we can't crop these images because you lose the title text etc). This doesn't affect other galleries because we've cropped those images to the correct ratio in the grid.

There's a workaround in at the moment because there's an extra `/gallery` part in the URL which shouldn't be there. It will be fixed soon in [this PR](https://github.com/guardian/flexible-content/pull/2724).
## Screenshots

[http://localhost:9000/advertiser-content/kids--top-51-books **/gallery** /the-best-kids-books-of-all-time-revealed](http://localhost:9000/advertiser-content/kids--top-51-books/gallery/the-best-kids-books-of-all-time-revealed)
image
## Request for comment

@guardian/labs-beta


kelvin-chappell:
Nice! 👍

PR #14552, opened 2016-10-03
Use correct fetching time period

rich-nguyen:
A fix to make sure that the time period specified has data to read. The redis data is only kept for 20 minutes, so 600s \* 2 was probably missing the window of opportunity.


janua:
:+1:

PR #14553, opened 2016-10-03
Fix Explore series link

SiAdcock:
## What does this change?

Currently, the Labour Liverpool series identity link points to a non-existent "explore series" page. It looks like a case of find-replace gone awry. I have restored it to point to the Labour Liverpool series page.

Readers can navigate from Labour Liverpool articles to the Labour Liverpool series page.

No
## Request for comment

@guardian/dotcom-platform


rjmk:
👍

PR #14556, opened 2016-10-03
Supporting element from tablet onwards

zeftilldeath:
# From leftCol a supporting element looks like this:

screen shot 2016-10-03 at 16 16 18
# And on a tablet it looks like this:

screen shot 2016-10-03 at 16 16 41
## It makes no sense for a supporting element to take up 100% of the content area when it is hierarchically considered subordinate to the content. It also means interactive elements often sit uncomfortably _not-quite-full-width_ whilst not allowing text to flow around them*
# So let's make it look like this:

screen shot 2016-10-03 at 16 16 52

*This causes unneccasary suffering for people like @duarte

### DIFF
--- a/static/src/stylesheets/module/_layout-hints.scss
+++ b/static/src/stylesheets/module/_layout-hints.scss
@@ -38,7 +38,7 @@
 }
 
 figure.element.element--supporting {
-    @include mq(desktop) {
+    @include mq(tablet) {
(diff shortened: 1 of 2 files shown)


gtrufitt (approved):
That selector (`figure.element.element--supporting`) is bad. But this change is alright I suppose 👍

PR #14558, opened 2016-10-03
Fix text for share button being wrong

rjmk:
## What does this change?

Fallback text for share icons in comments

Previously the twitter button had the wrong text ("Share on Facebook") and lots of hardcoded properties that could be read off the available data. Now that happens, so we have reduced the number of sources of truth

No
## Request for comment

@guardian/dotcom-platform

### DIFF
--- a/discussion/app/views/fragments/commentShare.scala.html
+++ b/discussion/app/views/fragments/commentShare.scala.html
@@ -10,29 +10,27 @@
     <div class="sharing-buttons" data-link-name="comment social">
     @defining(s"${comment.webUrl}") { permalink =>
         @defining(StringEscapeUtils.unescapeHtml(Jsoup.clean(comment.body.replaceAll("<blockquote>.*</blockquote>", ""), Whitelist.none()))) { commentBody =>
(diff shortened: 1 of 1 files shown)


discussion/app/views/fragments/commentShare.scala.html
```diff
@@ -10,29 +10,27 @@
...
-                        Share on Facebook</span>
-                    @fragments.inlineSvg("share-twitter", "icon", List("rounded-icon", "social-icon", "social-icon--twitter", "comment-twitter-icon"))
-                    <span class="u-h">Twitter</span>
-                </a>
+                @shareLink(permalink, Twitter, commentText, None)
             }
         }
     }
+
+    @shareLink(permalink: String, platform: SharePlatform, text: String, quote: Option[String]) = {
```
rjmk: I tried making this `@shareLink(permalink: String)(platform: ....)`, but the templating didn't seem to like defining `@shareLinkArticle = { @shareLink(permalink) }` when the permalink was in scope

jfsoul (approved):
Looks good to me! 👍

PR #14559, opened 2016-10-03
Hosted galleries - remove workaround for routing problem

lps88:
## What does this change?

We had the wrong URLs for the hosted gallery content from CAPI, so there was a workaround for the frontend routing. Now it's fixed and can be removed.

Also some other minor fixes to remove some errors.

Clean up! 🗑 
## Request for comment

@guardian/labs-beta


static/src/javascripts/projects/commercial/modules/hosted/onward.js
```diff
@@ -10,11 +10,9 @@ define([
 
     function loadOnwardComponent() {
 
-        // todo: why is this necessary?  if it's not here get 404s all over the site
-        if (config.page.isHosted) {
-
-            var placeholders = document.querySelectorAll('.js-onward-placeholder');
+        var placeholders = document.querySelectorAll('.js-onward-placeholder');
 
+        if (config.page.isHosted && placeholders.length) {
```
kelvin-chappell: Is the `config.page.isHosted` condition still necessary?  This component is only supposed to be included in hosted content: https://github.com/guardian/frontend/blob/master/static/src/javascripts/bootstraps/commercial.js#L83

I think it was being triggered on every page because I put brackets into the init reference.
But we can easily test this.

Or if you want to get this out quickly we can get rid of it later.

kelvin-chappell (approved):

static/src/javascripts/projects/commercial/modules/hosted/onward.js
lps88: good point, I'll get rid of it

PR #14563, opened 2016-10-04
Styling tweaks to the tertiary nav

NataliaLKB:
## What does this change?

This improves the tertiary nav styling. 

When the display names for the links are too long they will be two lines. Currently, this doesn't look the best, but this PR changes it to look much better.

Also, on mobile it feels like a bit too small an area to tap, so making it bigger seems like a good idea.

Things look better!

Nope
## Screenshots

**Before:**
## image

image

**After:**
## image

image
## Request for comment

@SiAdcock @zeftilldeath

### DIFF
--- a/static/src/stylesheets/layout/_ab-new-header-test-variant.scss
+++ b/static/src/stylesheets/layout/_ab-new-header-test-variant.scss
@@ -708,7 +708,7 @@ svg.main-menu__icon--social__svg {
 
 .tertiary-navigation__item {
     width: calc(50% - #{$gs-gutter / 2});
-    height: $gs-row-height;
+    min-height: $gs-row-height;
     border-top: 1px solid $neutral-5;
     padding-top: 2px;
     padding-bottom: $gs-baseline / 2;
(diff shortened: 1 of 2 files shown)


SiAdcock:
👍  Lovely

static/src/stylesheets/layout/_ab-new-header-test-variant.scss
```diff
@@ -708,12 +708,20 @@ svg.main-menu__icon--social__svg {
...
 }
 
+.tertiary-navigation__text {
+    width: calc(100% - #{$gs-gutter});
+    float: left;
+}
+
 .tertiary-navigation__link {
     @include fs-textSans(5);
+    line-height: 18px;
```
gtrufitt: Minor, but could this be a proportion of the item height (i.e - you expect the item to be two lines so the line height is `$gs-row-height / 2`

gtrufitt (approved):
👍

static/src/stylesheets/layout/_ab-new-header-test-variant.scss
NataliaLKB: That is an interesting though @gtrufitt . I don't think the line height is related to the item height though. If that variable value was ever changed so that `$gs-row-height` was equal to 40px instead, we would still want the line-height to be 18px.

That said, I doubt we would change the value of  `$gs-row-height` anytime soon, if ever.

PR #14565, opened 2016-10-04
Creating the new header control for the test

NataliaLKB:
## What does this change?

We need a control to test the variant against. 

Please note that this PR does not turn on the test! That needs to be done via fastly. A PR for that will follow shortly

Better data!

Nope!
## Request for comment

@guardian/dotcom-platform


SiAdcock:
👍  Makes sense

PR #14566, opened 2016-10-04
Fix hosted onward component

kelvin-chappell:
Wasn't showing expected number of trails, because of a couple of bugs.

Test fixtures were a bit painful.  Should have used some stubs instead!

/cc @lps88


Calanthe:
👍

PR #14567, opened 2016-10-04
Add functionality for A/B tests to defer their Ophan impression event

desbo:
## What does this change?

If all of the variants in a test provide an `impression` function, they won't be registered Ophan as part of the [`trackEvent` call in `standard/main`](https://github.com/guardian/frontend/blob/master/static/src/javascripts/bootstraps/standard/main.js#L160). Instead, the `impression` function in the variant will need to call the function it receives as a parameter to register itself with Ophan. It works in the same way as the `success` function that you can already add to your variants.

This impacts the `completions` A/B metric, which compares the number of `complete = false` events against the number of `complete = true` for a test. `complete = false` is fired either at pageload (when `canRun` is true) or now by the `impression` function, if a test provides it.

Instead of counting impressions as soon as the page loads, this gives you extra control over when an impression is counted. It'll probably remove some noise from test results.

For example, if you're testing the behaviour of a component that's positioned low down on a page, you might only want to count impressions from when the element is actually in view, to rule out anyone who hasn't seen it. Here's how you'd do that:

``` javascript
this.variants = [
    {
        id: 'variantA',

        test: function () {
            // add some element to the page
        },

       impression: function (track) {
           // when the element is in view, call track()
        },

        success: function (complete) {
           // when the user clicks the element, call complete()
        }
    }
// more variants
```

No 
## Request for comment

@jranks123 @gtrufitt

### DIFF
--- a/docs/03-dev-howtos/01-ab-testing.md
+++ b/docs/03-dev-howtos/01-ab-testing.md
@@ -76,6 +76,10 @@ define(['bonzo'], function (bonzo) {
                 test: function (context, config) {
                 },
 
+                impression: function(track) {
+                    // call track() when the impression should be registered (e.g. your element is in view)
+                },
+
                 success: function(complete) {
                     // do something that determines whether the user has completed
                     // the test (e.g. set up an event listener) and call 'complete' afterwards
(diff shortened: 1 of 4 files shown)


jranks123:
Awesome, this is great! 👍

gtrufitt (approved):
Awesome, super useful!

PR #14568, opened 2016-10-04
Remove array mutation from loadAdvert

regiskuckaertz:
`loadAdvert` had side-effects that forced the developer to take extra measures at the calling site, but this was not obvious and already created a few bugs in the past.

I'm just making the mutation explicit by moving it out of `loadAdvert`. Also, I got rid of using `splice` and favored computing a new collection every time instead.


janua:
:+1:

PR #14569, opened 2016-10-04
Add discussion external assets lifecycle to applications AppLoader

jfsoul:
## What does this change?

The lifecycle component that populated the `DiscussionAssetsMap` agent was not added to the `applications` app.  This meant that content served from `applications` (e.g. cartoons) did not have the asset map populated and therefore would not load the comment count.  This adds the lifecycle component to `applications` (I don't think it needs to be added to any other apps).

Comment count works for content served from `applications`
## Screenshots

No screenshots as the comment section is still being hidden (but comment count not loading will no longer be the cause!)
## Request for comment

### DIFF
--- a/applications/app/AppLoader.scala
+++ b/applications/app/AppLoader.scala
@@ -1,4 +1,5 @@
 import app.{FrontendApplicationLoader, FrontendComponents}
+import assets.DiscussionExternalAssetsLifecycle
 import com.softwaremill.macwire._
 import common.dfp.DfpAgentLifecycle
 import common.{ApplicationMetrics, CloudWatchMetricsLifecycle, ContentApiMetrics, EmailSubsciptionMetrics}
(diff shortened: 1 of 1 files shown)


gtrufitt (approved):
@nicl

gtrufitt:
👍

NataliaLKB (approved):

nicl:
👍 thanks for the heads up @gtrufitt @jfsoul 

Couple of questions:

* are they any other content type exclusions we should be aware of here that still exist?
* why is the comment section still hidden? (I'm not sure I really understood this part though - the note in the 'Screenshots' section above)

jfsoul:
> are they any other content type exclusions we should be aware of here that still exist?

We had a think about other content types, but the conclusion was that `applications` covered everything else that contained comments.

> why is the comment section still hidden?

It's not, my original comment wasn't clear at all.  Locally the section was still hidden for some reason other than the comment count being unable to load.  When we rolled this change out, the comments became visible again on cartoons etc.


My only remaining reservation is whether we should be hiding the whole comment section if the comment count can't be loaded on closed-comment content.

PR #14570, opened 2016-10-04
Fix youtube URL

lps88:
## What does this change?

The origin parameter should be https instead of http. Quick fix to get the videos working, will clean it up later.

@guardian/labs-beta


Calanthe:
👍

PR #14573, opened 2016-10-04
Use cors mode to call hosted onward container

kelvin-chappell:
This should avoid the preflight options call that results in a 404.


regiskuckaertz (commented):
rock'n'roll 🎸 🎶

regiskuckaertz (approved):

PR #14578, opened 2016-10-04
Refactor hosted video sources

kelvin-chappell:
This change avoids having to specify specific video mime types in frontend.
We just render whatever is passed in through capi.

/cc @guardian/labs-beta


lps88:
Nice 👍

PR #14579, opened 2016-10-04
Fixing scrolling bug new header

NataliaLKB:
## What does this change?

At the moment if you are opted in to the new header and click on one of the primary links (life, art, news, etc) the menu opens and you can't scroll on the body. But, if you close the menu then you still can't scroll. 

This fixes it.

People can still scroll 😓 

Nope!
## Request for comment

@SiAdcock @zeftilldeath @stephanfowler


SiAdcock:
👍

PR #14580, opened 2016-10-04
Remove hardcoded tones from hosted content

kelvin-chappell:
We can take what capi says is the tone.


JonNorman:
I don't know any of the context here, but does the Zootropolis page still need to be hardcoded?

kelvin-chappell:
Unfortunately.  We're keeping the hardcoded campaigns until there's enough content in composer to show potential clients.

JonNorman:
Fair 'nuff :+1:

JonNorman (approved):

PR #14582, opened 2016-10-04
Fabrics sizes

regiskuckaertz:
Fixes two bugs and adds a use case in the Fabrics family:
- the Fabric Video has a max-width of 1920
- the Revealer should be styled as a Fabric

and
- when the video fails to load, Fabric Video should display the background poster image


JonNorman:
LGTM :+1:

PR #14585, opened 2016-10-05
Hosted onward journey more fixes

lps88:
## What does this change?

Get rid of flexbox styles in onward journey component, and separate the video/article styles a bit more. Should fix this issue:

image

@guardian/labs-beta


kelvin-chappell (approved):
👍

PR #14587, opened 2016-10-05
Enable liveblog paging in preview by adding article route to standalone

jfsoul:
## What does this change?

Adds a liveblog route to `standalone` (which is imported by preview) so that paging on liveblogs in preview works.

Still looking into why liveblog updates in preview are in reverse, but removing this annoyance might help.

Liveblog previews are pageable.

No

## Request for comment

@guardian/dotcom-platform @johnduffell


gtrufitt (approved):
👍

johnduffell:
thanks!
as said check what it's sorting on, you may find that field is a None on preview and therefore under a stable sort they're staying in order of first to last

PR #14590, opened 2016-10-05
Reduce hosted template duplication

lps88:
## What does this change?

Shoehorn the custom [Zootropolis article page](https://www.theguardian.com/advertiser-content/disney-zootropolis/meet-the-characters-of-zootropolis) into the standard article template, and remove its custom page class.

The page renders exactly the same except for the onward journey (see below) and means less maintenance for us.
## Screenshots

The page now has an article-style onward journey link instead of the video-style one it had originally:
image
## Request for comment

@guardian/labs-beta


kelvin-chappell (approved):
👍

PR #14594, opened 2016-10-05
Update CAPI models and client

wpf500:
CAPI models have undergone a major version upgrade, which I think changes a lot of internals, but nothing externally (@tomrf1 might know more, see https://github.com/guardian/content-api-models/pull/60)

I need a more recent version of CAPI models (`10.2`) for the interactive atom work, this PR is a stepping stone to get frontend synced to `10.x`. I'm not aware of any testing, but have spoken to @jfsoul who says he will check it!

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -29,7 +29,7 @@ object Dependencies {
   val commonsLang = "commons-lang" % "commons-lang" % "2.4"
   val commonsIo = "commons-io" % "commons-io" % "2.4"
   val cssParser = "net.sourceforge.cssparser" % "cssparser" % "0.9.18"
-  val contentApiClient = "com.gu" %% "content-api-client" % "9.5"
+  val contentApiClient = "com.gu" %% "content-api-client" % "10.2"
   val capiScalaModels = "com.gu" % "content-api-models-scala" % capiModelsVersion
   val capiJsonModels = "com.gu" % "content-api-models-json" % capiModelsVersion
   val dfpAxis = "com.google.api-ads" % "dfp-axis" % "2.8.0"
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -9,7 +9,7 @@ object Dependencies {
   val slf4jVersion = "1.7.5"
   val awsVersion = "1.11.7"
   val faciaVersion = "2.0.2"
-  val capiModelsVersion = "9.24"
+  val capiModelsVersion = "10.1"
 
   val akkaAgent = "com.typesafe.akka" %% "akka-agent" % "2.3.4"
   val akkaContrib = "com.typesafe.akka" %% "akka-contrib" % "2.3.5"


dominickendrick:
I think you accidentally deleted the PR guidelines before you wrote this description 😉 
Can you add a more detailed descriptions about why we need this change ?
Also is there any testing we need to do on this ?

project/Dependencies.scala
```diff
@@ -9,7 +9,7 @@ object Dependencies {
   val slf4jVersion = "1.7.5"
   val awsVersion = "1.11.7"
   val faciaVersion = "2.0.2"
-  val capiModelsVersion = "9.24"
+  val capiModelsVersion = "10.1"
```
dominickendrick: Why is this `10.1` and the client `10.2` ?

wpf500: The Guardian seems to like messing with people.

project/Dependencies.scala
```diff
@@ -29,7 +29,7 @@ object Dependencies {
   val commonsLang = "commons-lang" % "commons-lang" % "2.4"
   val commonsIo = "commons-io" % "commons-io" % "2.4"
   val cssParser = "net.sourceforge.cssparser" % "cssparser" % "0.9.18"
-  val contentApiClient = "com.gu" %% "content-api-client" % "9.5"
```
gtrufitt: If this is for the interactive atom, [isn't that in 10.3](https://github.com/guardian/content-api-scala-client/blob/96b8ad38ed3921e69992ee71eb0f14798ab973c1/CHANGELOG.md)?

wpf500:
@dominickendrick, added a better description

wpf500:
It seems the tests are failing but I don't have TeamCity access, @jfsoul happy to help if I can?

project/Dependencies.scala
wpf500: The fact that these version numbers is so similar is in fact coincidental I think, they aren't always necessarily released together. @mchv probably knows more

project/Dependencies.scala
gtrufitt: I commented on the wrong side

wpf500: I'm trying to separate my PRs, so bumping you up to the `10.x` series first, then will bump to the version with interactive atoms in my atom PR

project/Dependencies.scala
dominickendrick: Ah ok !

jfsoul:
@wpf500 Teamcity has 185 test failures - a lot of these can be put down to test fixtures needing to be updated/added, but there are also a lot that aren't related to this.

I'm not sure how much work there is in fixing all these yet, but it might not be unsubstantial.  Will take a further look tomorrow.

project/Dependencies.scala
dominickendrick: @mchv says the `content-api-models-scala` and `content-api-models-json` are included as a transitive dependency of the `content-api-client`, so I don't think we need to import these specifically, so we should be able to remove these

wpf500: I tried this but it failed at runtime, so there does seem to be some strange dependency somewhere

gtrufitt: It's worth pinging @kelvin-chappell (just did) as he recently added the `capiModelsVersion` I believe

kelvin-chappell:
The capi client no longer supports json from v10x, https://github.com/guardian/content-api-scala-client/pull/170, 
so any tests that depend on cached test capi fixtures will fail until the HttpRecorder is modified.

kelvin-chappell:
I need to use the latest models too so I'll get this sorted out.

kelvin-chappell:
Please see #14615

wpf500:
Closing this as fixed by #14615

PR #14595, opened 2016-10-05
Vagina dispatches 🌷 

zeftilldeath:
Email header:
vagina-dispatches

@desbo Is this ok? Will it work? How are you?


desbo:
this is ok,
it will work,
i'm fine

👍

PR #14596, opened 2016-10-05
Fix preview liveblog ordering

jfsoul:
## What does this change?

Fixes the order of liveblogs in preview by changing the sorting logic for blocks.  If there is no published date (as in the case of preview) then we default to using the created date for ordering.

Liveblogs are once again ordered correctly in preview.

No

## Screenshots

Before:
image

After:
image
## Request for comment

@guardian/dotcom-platform 

@johnduffell thanks for the tip!

### DIFF
--- a/common/app/model/liveblog/BodyBlock.scala
+++ b/common/app/model/liveblog/BodyBlock.scala
@@ -13,7 +13,8 @@ object Blocks {
   def make(blocks: ApiBlocks): Blocks = {
 
     def orderBlocks(blocks: Seq[BodyBlock]) =
-      blocks.sortBy(_.firstPublishedDate.map(_.getMillis).getOrElse(0L)).reverse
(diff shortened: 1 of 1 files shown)


common/app/model/liveblog/BodyBlock.scala
```diff
@@ -13,7 +13,8 @@ object Blocks {
   def make(blocks: ApiBlocks): Blocks = {
 
     def orderBlocks(blocks: Seq[BodyBlock]) =
-      blocks.sortBy(_.firstPublishedDate.map(_.getMillis).getOrElse(0L)).reverse
+      blocks.sortBy(-_.publishedCreatedTimestamp().getOrElse(0L)) // Negate rather than reverse result: leaves
```
johnduffell: ignore that, I just found the implementation does that anyway!

johnduffell:
great! thanks for that 🎉 🚀 👍

PR #14598, opened 2016-10-05
Use a specific search criteria to choose log files to upload 

rich-nguyen:
The commercial logs that are written to data lake suffered a problem where a log race condition was interfering with the upload. This was a clash between logback's max history setting and the upload step.

### DIFF
--- a/admin-jobs/conf/application-logger.xml
+++ b/admin-jobs/conf/application-logger.xml
@@ -20,7 +20,7 @@
 
         <rollingPolicy class="ch.qos.logback.core.rolling.TimeBasedRollingPolicy">
             <fileNamePattern>logs/frontend-commercial-client-side-archive/%d{yyyy-MM-dd--HH-mm}.log</fileNamePattern>
-            <maxHistory>20</maxHistory>
+            <maxHistory>40</maxHistory>
         </rollingPolicy>
 
         <encoder>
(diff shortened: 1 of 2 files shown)


admin-jobs/conf/application-logger.xml
```diff
@@ -20,7 +20,7 @@
 
         <rollingPolicy class="ch.qos.logback.core.rolling.TimeBasedRollingPolicy">
             <fileNamePattern>logs/frontend-commercial-client-side-archive/%d{yyyy-MM-dd--HH-mm}.log</fileNamePattern>
-            <maxHistory>20</maxHistory>
+            <maxHistory>40</maxHistory>
```
rich-nguyen: 40 minutely log items in the archive, maximum, or 40 minutes.

JonNorman (approved):
So to recap my understanding on this, there are two steps:

1) The commercial metrics are written to the logs every 5 minutes.
2) Every 15 minutes, the commercial uploader picks up the large log files (typically three of them, given every 5 minutes in a 15 minute window) and uploads them to S3.

Because of the times above, a race condition exists such that the third log block may be written at the same time as the commercial uploader collects the logs to send to S3. As a result, it may not get picked up until the subsequent 15 minute window. This means that it will be 20 minutes old by the time it is collected.

Prior to this change, the logs were automatically cleaned up every 20 minutes, leading to the race condition whereby a block of logs may be removed before ever being sent to S3. By setting the log cleaning frequency to 40 minutes, this second race condition should now be averted.

Sounds sensible to me!

rich-nguyen:
@JonNorman Correct! ⭐️

PR #14600, opened 2016-10-06
Fix the unstyled flash on Weekend Reading

katebee:
## What does this change?

Fix for https://www.theguardian.com/signup/weekendreading

Some critical styles have ended up in the below-the-line async rendering, leading an to ugly page before loading completes:

image
- Adds the required styles to the css head
- Improve the SVG usage on the Weekend Reading page
- Beautiful pages 🌈 

Bonus: It now looks good regardless of JavaScript

Nope
#### No JavaScript:

picture 281
#### With JavaScript:

picture 280
## Request for comment

@guardian/guardian-frontend


static/src/stylesheets/module/signup/_signup-page.scss
```diff
@@ -54,35 +54,58 @@
...
+    }
 
-            @include mq($until: tablet) {
-                width: 100%;
-                margin-left: 0;
-            }
+    .email-signup__svg--envelope__svg {
+        fill: #ffffff;
+        width: 28px;
+        height: 12px;
```
NataliaLKB: Why not user `$gs-baseline`?

static/src/stylesheets/module/signup/_signup-page.scss
```diff
@@ -54,35 +54,58 @@
...
+        pointer-events: none;
+    }
 
-            @include mq($until: tablet) {
-                width: 100%;
-                margin-left: 0;
-            }
+    .email-signup__svg--envelope__svg {
+        fill: #ffffff;
+        width: 28px;
```
NataliaLKB: I would probably use a value from the grid system unless that doesn't work visually. Then I would usually leave a note explaining that this value is important and why.

Em values are really useful for icons. Especially if you have text that changes size between breakpoints. You only have to set one em value for the icons and they will always look in proportion. If this is the case I would probably go back to using the `em` value, but if not then make a note :)

katebee: Ah, excellent! Didn't know that; I will change to use the em. Thanks!

NataliaLKB (approved):
Looks great! 👍

PR #14602, opened 2016-10-06
Auto-describe makefile in `make help`

sndrs:
## What does this change?
- auto-describes the `makefile` (lists all targets with any comments)
- sets `make help` to be the default target (`make` = `make help`)

we're planning to ditch grunt, which means probably adding more to the makefile. this will hopefully keep that as useful as possible, and no one will have to remember to update the help target when it changes (it's already out of date).

no
## Screenshots

before:

screen shot 2016-10-06 at 12 59 33

after:

screen shot 2016-10-06 at 12 59 47
## Request for comment

@guardian/dotcom-platform


gtrufitt:
👍 Yay, auto updating docs!

PR #14607, opened 2016-10-06
Give signup pages a slim header

katebee:
## What does this change?
- Gives the signup pages a slim header (again)
## Request for comment

@dominickendrick


SiAdcock:
👍

PR #14608, opened 2016-10-06
Create notion of ‘private’ targets in makefile

sndrs:
## What does this change?
- `make`/`make help` now only lists 'common' tasks, which means tasks not marked as `# PRIVATE`
- new `make list` target lists everything in the makefile

As tasks are pulled out of the grunt file, the makefile will fill up. the idea is hide the workings out if you just need to get the site up and running, but allow anyone who wants to to easily see what's going on.

no
## Screenshots

`make help`

screen shot 2016-10-06 at 17 00 55

`make list`

screen shot 2016-10-06 at 17 01 03
## Request for comment

@guardian/dotcom-platform

### DIFF
--- a/makefile
+++ b/makefile
@@ -1,30 +1,19 @@
-# Lists all `make` targets.
-help:
+# Targets marked '# PRIVATE' will be hidden when running `make help`.
+# They're helper targets that you probably only need to know about
+# if you've got as far as reading the makefile.
+
+# Lists common targets.
+# Also the default task (`make` === `make help`).
+help: # PRIVATE
 	@node tools/messages.js describeMakefile
 
-# ********************************************************
+# Lists *all* targets.
+list: # PRIVATE
+	@node tools/messages.js describeMakefile --all
 
-# Watch and automatically reload all JS/SCSS.
-# Uses port 3000 insead of 9000.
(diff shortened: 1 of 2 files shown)


makefile
```diff
@@ -1,20 +1,58 @@
-# Lists all `make` targets.
-help:
+# Targets marked '# PRIVATE' will be hidden when running `make help`.
+# They're helper targets that you probably only need to know about
+# if you've got as far as reading the makefile.
+
+# Lists common targets.
+# Also the default task (`make` === `make help`).
+help: # PRIVATE
```
rjmk: Is this better off public?

sndrs: the only way you could see it listed would be to run it :)

`private` doesn't stop tasks being run, just listed by `make help` - sorry if that wasn't clear

rjmk: No, that's clear! For me, I just would find it very strange to run `make help` and not see `help` as an option

jfsoul:
👍  Nice! Anything we can do to get more people using the makefile with ease 💯

makefile
sndrs: yeah that's a fair point. reinforcing it may also help people to remember it too

PR #14609, opened 2016-10-06
[WIP] Dirt simple flow setup

rjmk:
(**Please look through by commit, because of all the noise from the files**. The only thing of note that happens in cbff5655 is tweaking [this file](https://github.com/guardian/frontend/blob/ac9485d9b9644c27047f109ff6de6c61f504593b/static/src/flow/projects/common/utils/assign.js))

Yo, @guardian/dotcom-platform!

Here is a super simple way of setting up flow. I have included some bad annotations in `static/src/flow/projects/common/utils/assign.js` so you can check flow is working.

The approach is to work on the files in `flow` and then strip all the type annotations, putting the result into `javascripts`. Then, everything else works exactly as before. It's integrated with our make makefile as well


rjmk:
https://github.com/guardian/frontend/blob/ac9485d9b9644c27047f109ff6de6c61f504593b/static/src/flow/projects/common/utils/assign.js

jfsoul:
> [WIP][Strawman][DNM][ :hand: ]

I blame you for this abbreviation explosion @gtrufitt

sndrs:
is it possible to not move all the JS? little bit tricky to see what's actually changed...

rjmk:
@sndrs All the action is in f4c52a6 and 52a447b

sndrs:
i think it's generally good practise to PR as few changes as possible, so you can comment on the diff rather than individual commits, but here's some general thoughts:

- given we don't know what blockers we're going to hit with any of the stuff we're talking about, moving the entire app in one go seems a bit too dangerous
- re `static/src/flow` - without knowing the build process, what would you think think was in there? it's not really 'flow', it's javascript. but as part of a build step, we create a `javascript` directory next to it, which _is_ the javascript from `flow`, cleaned up. if i want to edit javascript, how do i know not to edit  the directory called `javascript` in `src`, but the one called `flow`? perhaps `target` might be better, or an un-tracked interstitial `flow` directory, like we do for the build dir of `r.js`.
- don't forget the watcher will need to track changes to the flow files, compile them and alert browser sync too

all these feel like it's adding an extra level of complexity, when what we want to do it reduce it. 

if we were not going to remove require any time soon, i'd actually be up for taking this a bit further and working on the same principle with babel, which could incorporate flow in its transformation. one of the reason's i think we'd do well to drop require is it requires you to perform this kind of i/o dance for whatever you want to do to your source.

i think by holding off till we can incorporate these transformations cleanly in a new build process, we should be able to adopt this and any other processing of the source simply and (hopefully) almost transparently

gtrufitt:
This is really great to see this progress but I share Alex's reservations with applying to our entire codebase and moving the JS - I know you've discussed this with Simon and Dom about it but lets catch-up as a team about our approach to this stuff so we're all focussed on the same goal. 

I've put 15 minutes to discuss the approach to implementing 'future' into our codebase.

gtrufitt (changes requested):
Gonna block merge on this for the moment, only to grey out that button from an accidental click 😨

rjmk:
@gtrufitt Thanks for the block! I will clean up the title now

gtrufitt:
As far as the strawman goes though, it's worked 😄

rjmk:
> i think it's generally good practise to PR as few changes as possible, so you can comment on the diff rather than individual commits

Agreed, but I think for this approach it sort of is the fewest changes possible (without merging a move of all the files). I'll add another commit to unmove lots of the files to glean up github's diff.

> re `static/src/flow` - without knowing the build process, what would you think think was in there? it's not really 'flow', it's javascript

Good point!

> if i want to edit javascript, how do i know not to edit the directory called javascript in src, but the one called flow

One part is documentation, but another is that it won't show up in your git changes

I think one thing I might not have been clear enough about why I went for this approach, is that it's quite important that the folder called `javascripts` is not renamed. This allows all the existing tooling to work without any issues. I would be pretty surprised if something was still able to break

> don't forget the watcher will need to track changes to the flow files, compile them and alert browser sync too

I'll work on that

> all these feel like it's adding an extra level of complexity, when what we want to do it reduce it

I actually have a very different take on this. I think that the discrete steps are simpler -- they're doing one thing and well. Ultimately, they're not the right choice because a bunch of them will collapse under their own weight, but I think we can start here quite sensibly

> i think by holding off till we can incorporate these transformations cleanly in a new build process, we should be able to adopt this and any other processing of the source simply and (hopefully) almost transparently

Fair enough. My opinion is that we should start incorporating the transforms which have value, and then when managing becomes a problem, we have nice high quality motivation for the much harder task of changing the build process

rjmk:
This is much easier to diff. Thanks for the suggestion, Alex

rjmk:
Watching: done

PR #14611, opened 2016-10-07
[PlatformFuture] Initial commit.

gtrufitt:
In the interest of small PRs, here is a `package.json` and a `readme` for static-future as a start to creating a sandbox (and fairly rapidly production build) for the future of our client side architecture.

### DIFF
--- a/static-future/README.md
+++ b/static-future/README.md
@@ -1,0 +1,9 @@
+# Future Client-Side Architecture
+
+## Current Status
+
+**ALPHA** - Initial investigation and implementation, do not use in production.
+
+## Client-Side DX JS
+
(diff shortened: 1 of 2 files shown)


gtrufitt:
@guardian/dotcom-platform

jfsoul:
Get that ball rolling 👍  ⚽️  🎾  🎱

sndrs:
one thing that came up in a discussion with @rich-nguyen about this kind of thing before was the value of making sure that everyone knows their environment

(appreciate this is a slightly different case and also not a final PR, so only mean this as a point of discussion more than anything else...)

but rich's thought, with which he persuaded me, was that there's a lot of overhead in needing to find out, for example, which version of lodash is installed for this file i'm working on. more than the value of being able to move forward more quickly in some areas of the codebase than in others

by having one set of packages, you can just _know_ 'we're using lodash 3' or 'we can now use `async`/`await` etc...

rjmk:
I'm not sure how to use the `static-future` folder. What's your vision for it?

PR #14612, opened 2016-10-07
Correct the frequency of two emails

katebee:
## What does this change?

Issue raised by User Help team
- Corrects the frequency of two emails


jfsoul:
👍

PR #14613, opened 2016-10-07
Messenger: viewport size persistent listener

regiskuckaertz:
Adds a new persistent listener that updates interested native ads every time the viewport gets resized.


rich-nguyen:
👍

PR #14614, opened 2016-10-07
This is not the greatest PR in the world, no. This is just a tribute.

zeftilldeath:
VW sizing behaves badly on ios, bumping it down prevents the word life from getting too friendly with the veggie burger.

imgres


duarte:
👍

duarte (approved):

PR #14615, opened 2016-10-07
Upgrade capi client

kelvin-chappell:
This bounces the capi client library to v10, which no longer supports json.

I've modified the implementation of HttpRecorder slightly to enable cached capi test fixtures to be in thrift format.  Hopefully teamcity will pass.

I've also removed the switch to choose between json and thrift capi responses.

I've also removed explicit dependency on capi models so that we allow the client to bring in the versions it needs.


mchv:
This looks good to me from a CAPI perspective

jfsoul:
Think I might be missing some files from the diff due to number of files changed.  I can't see `Dependencies.scala` at least.

I can see the `HttpRecorder`, switch removal and `ContentApiClient` changes - anything I'm missing?

💯  for getting green on teamcity.

kelvin-chappell:
Seems to be very difficult to filter particular files in a large PR.  It won't show more than 300 and the file dropdown isn't complete either.
I suppose I should have put the test fixture changes in a separate commit.
This is what Dependencies looks like in my branch anyway:
https://github.com/guardian/frontend/blob/kc-capi-client/project/Dependencies.scala

jfsoul:
Yeah I had the same difficulty.  If you do a branch comparison and then add `.patch` then I think you can get everything - but the test fixtures are so huge my browser has trouble rendering/scrolling.

Thanks for this @kelvin-chappell, looks good 👍

kelvin-chappell:
I'll deploy this on Monday.

PR #14619, opened 2016-10-07
Move AB tests into enhanced js

gtrufitt:
## What does this change?

AB tests were bundled into the standard JS but it's rare there is a requirement that an AB test runs before the enhanced JS.

Some implications of this:
- As we don't run enhanced js on ipads, we won't see AB tests there
- AB tests will run later in the day

The standard bundle will be much smaller as the dependency tree of ab tests can be pretty large. It also makes converting our standard JS for testing in a new architecture much easier.

In fact, with the current batch of A/B tests the app.js (bundled standard) is _308kb_, whereas removing the AB tests, the bundle is _89kb_. Obviously these kbs are moved elsewhere but that feels far more reasonable for **critical** JS..

@guardian/dotcom-platform

### DIFF
--- a/static/src/javascripts/bootstraps/enhanced/main.js
+++ b/static/src/javascripts/bootstraps/enhanced/main.js
@@ -8,6 +8,8 @@ define([
     'common/utils/detect',
     'common/utils/mediator',
     'common/utils/user-timing',
+    'common/utils/robust',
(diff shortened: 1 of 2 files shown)


sndrs (approved):
lgtm! 𐄷

stephanfowler:
This is long overdue!

PR #14620, opened 2016-10-07
Report clicks in native ads

regiskuckaertz:
Native ads are running inside iframes and so we need a programmable way of reproducing the `data-link-name` attribute reporting.

When a click happens in one of those ads, a message will be sent to the parent window with the content of the attribute. From there, our listener here takes over and just pipes this info into the correct tracking API.

NB: I also fixed a potential bug in the tracking where it was accessing a proprietary property that will be dropped some time in the future, in favour of the standard one, `textContent`.

cc @dominickendrick @ajosephides @gtrufitt 

There is a piece of information I am not sending right now, for lack of know how to do it: the name of the ad slot from where the click originated. Please tell me which of the `dimensionX` I can use to do that.


static/src/javascripts/projects/common/modules/analytics/google.js
```diff
@@ -34,9 +34,14 @@ define([
         });
     }
 
+    function trackNativeAdLinkClick(slotName, tag) {
+        ga(send, 'event', 'Click', 'Native Ad', tag);
+    }
```
ajosephides: Hi,
2 things.

I've created a new dimension for the Ad Slot Name (adSlotName). This can be referenced within here as 'dimension25'.

Also, if could add the non-interaction boolean - "nonInteraction: true"
Although, you could argue this isn't needed in this case (prevent bounce rate being affected by a click) as its an external click, in this case (I believe) the click opens a new tab and so we the user may well continue to browse theguardian.com

gtrufitt:
👍 lgtm code wise once you've added that dimension

static/src/javascripts/projects/common/modules/analytics/google.js
regiskuckaertz: Great! I'm doing that now, thanks 😄

PR #14621, opened 2016-10-07
Add verification code for https://www.theguardian.com

gklopper:
Allows us to verify the HTTPS with Google search console


gklopper:
cc @stephanfowler - we need to merge this before I can add HTTPS to the search console

stephanfowler:
:+1: thanks Grant

PR #14622, opened 2016-10-09
Clean up the commercial structure

JonNorman:
## What does this change?

_Big intake of breath_
This is the first PR that attempts to sort out the commercial directory/package structure.
This PR only concentrates on the directory structure mainly doing the following:

1) Removing pointless `commercial` folders within the commercial project
2) Creating a more 'MVC' structure that will suit the eventual update of package names

This does have the downside of making it harder to track changes across time as git doesn't track renaming of files, but better now than further down the line I reckon.

I'd highly recommend having a look commit-by-commit as each change should be fairly small/simple/atomic.

I'd initially planned to sort out the packages (and began so with the controllers), but will do the models in a separate PR as it may/will get horrible.

Navigating the project should be a lot easier and easier to pick up.

No
## Request for comment

@guardian/commercial-dev


JonNorman:
Whilst minor, it would be good to get a second opinion on this: a lot of the commercial packages are of the form `model.commercial` and `controllers.commercial`. This doesn't make sense to me, since if we want to add a further subpackage, e.g. `merchandise` then it would become `model.commercial.merchandise`, which seems disordered.

It would make more sense if there was some root project that imported models from all over the place and thus we had `model.commercial` and `model.facia` and `model.article` - but as I understand it, we explicitly don't do this as each project should be siloed; if there is any shared functionality then it should go in `common` at which point it should be `common.model.commercial` etc.

kelvin-chappell (changes requested):
Great idea!
Perhaps all the packages should start at `commercial` so that they're definitely unique and there's no chance of conflict with the `common` module. Eg. `commercial.model`

Also while you're at it I'd get rid of all the package objects as they're inherently evil. (Although that could be a separate PR.)

JonNorman:
@kelvin-chappell I am in utter agreement with you - I went to edit all the packages but it ballooned the size of this already sizeable PR, so I am planning to change all the packages following the successful merge of this one!

With respect to the `package.scala`, yeah I'll find a way to nicely squash that...

JonNorman:
Also, as a little trick - when I changed all the packages (on a separate branch), intelliJ complained about the folder structure not reflecting the package structure (i.e. no `app/commercial` folder), however on the Project Settings there is an option to add an implicit package name to a folder; by adding 'commercial' as an implicit package name, all the warnings fall away :)

kelvin-chappell (approved):

rich-nguyen:
👍

PR #14624, opened 2016-10-10
Extend switch for link preconnect

rich-nguyen:
I haven't got round to looking into the numbers for this yet, so extending. @gtrufitt


gtrufitt:
👍

PR #14625, opened 2016-10-10
Make the 'upgrade rich-links only below viewport' test permanent 

gtrufitt:
## What does this change?

We will no longer upgrade rich-links if they are in or above the current the viewport.

Rich links upgrades were a main contributor to yank and a UX issue. 

We tried to remove rich link upgrades from mobile devices completely but saw a 50% drop in rich-link referrals - showing that there is real value in upgrading rich links. 

To test a compromise we only upgraded rich links if they're not in or above the viewport - eliminating their yank - but only seeing a 4.5% drop in rich-link referrals which, considering the UX gains, we're happy with.

@guardian/dotcom-platform


stephanfowler:
👍

PR #14626, opened 2016-10-10
Extend the membership banner A/B test

paulbrown1982:
## What does this change?

The membership banner A/B test expired on Friday, we need to extend it for a short while while we think of new variants.

I also removed the not-so-well performing UK variant at Jesse's request.

I did not enable the re-display to people who have closed it, as that will happen when the new variants go live.

Increased  conversions / resumption of the existing conversion rate.

No, only web.

## Request for comment

cc @johnduffell @JustinPinner @rtyley

### DIFF
--- a/static/src/javascripts/projects/common/modules/experiments/tests/membership-engagement-banner.js
+++ b/static/src/javascripts/projects/common/modules/experiments/tests/membership-engagement-banner.js
@@ -24,15 +24,15 @@ define([
     var EditionTest = function (edition) {
 
         this.edition = edition;
-        this.id = 'MembershipEngagementBanner'+edition[0].toUpperCase() + edition.substr(1);
-        this.start = '2016-09-08';
-        this.expiry = '2016-10-08';
+        this.id = 'MembershipEngagementBannerExtended'+edition[0].toUpperCase() + edition.substr(1);
+        this.start = '2016-10-10';
+        this.expiry = '2017-10-10';
         this.author = 'Roberto Tyley';
-        this.description = 'Show contributions as well as membership messages for the ' + edition + ' edition.';
+        this.description = 'Show membership messages for the ' + edition + ' edition.';
         this.showForSensitive = false;
         this.audience = 1.0;
         this.audienceOffset = 0;
-        this.successMeasure = 'Conversion for contributions';
+        this.successMeasure = 'Conversion for membership';
         this.audienceCriteria = 'All users in the ' + edition + ' edition.';
         this.dataLinkNames = '';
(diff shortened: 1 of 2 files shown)


static/src/javascripts/projects/common/modules/experiments/tests/membership-engagement-banner.js
```diff
@@ -24,15 +24,15 @@ define([
     var EditionTest = function (edition) {
 
         this.edition = edition;
-        this.id = 'MembershipEngagementBanner'+edition[0].toUpperCase() + edition.substr(1);
-        this.start = '2016-09-08';
-        this.expiry = '2016-10-08';
+        this.id = 'MembershipEngagementBannerExtended'+edition[0].toUpperCase() + edition.substr(1);
```
akash1810: Per [the docs](https://github.com/guardian/frontend/blob/master/docs/03-dev-howtos/01-ab-testing.md#writing-a-test), you'll need to update the test name in `ABTestSwitches` to reflect this change in name.

> id: The unique name of the test. This must be TitleCased and correspond to the hyphen-separated portion of the switch id that follows ab-. e.g. if the switch id is ab-geo-most-popular, this id must be GeoMostPopular

static/src/javascripts/projects/common/modules/experiments/tests/membership-engagement-banner.js
```diff
@@ -24,15 +24,15 @@ define([
...
+        this.start = '2016-10-10';
+        this.expiry = '2017-10-10';
```
akash1810: probably want to match this to the `sellByDate` of `ABTestSwitches`

static/src/javascripts/projects/common/modules/experiments/tests/membership-engagement-banner.js
paulbrown1982: Updated.

NataliaLKB:
Does this test need to take 100% of the ab testing audience?

In fact, there is only one edition with a variant, and no control. Should `au` `int` and `uk` be switches now, not ab tests?

paulbrown1982:
Agreed @NataliaLKB. @rtyley / @JustinPinner's team will be taking on the work to revert this back into being a permanent frontend feature, to which A/B tests can be applied. That should mean there is never a conversion collapse when a test ends, nor is there be a requirement that tests span 100% of the dotcom audience.

NataliaLKB:
Wonderful! Thanks @paulbrown1982 .

So that we can get this back up along with our membership revenue: 👍

PR #14629, opened 2016-10-10
Add a better log messages for the FaciaPressCron

janua:
This adds better failure for log messages on `facia-press`.

@kelvin-chappell


kelvin-chappell (approved):
👍

PR #14630, opened 2016-10-10
Show light logo on dark backgrounds, if one is available

kelvin-chappell:
The tag manager now supports an optional extra logo for use on dark backgrounds.

For example:

Before:
image

After:
image

/cc @guardian/labs-beta 

@DiegoVazquezNanini @GHaberis Not sure if apps will need to do something similar?


janua:
LGTM :+1:

PR #14634, opened 2016-10-10
To avoid an error, encode url first

NataliaLKB:
## What does this change?

We had an image in an old article cause an internal server error not too long ago. It was brought up because the article had to be edited by legal. Even though embedding svgs isn't a normal procedure in composer, our code should handle this better.

Even though I can't replicate this error, some googling around seems to suggest that special characters causes this error:

```
java.net.URISyntaxException: Illegal character in path at index 67: https://static.guim.co.uk/ni/1399891635186/Conle-Buildings_1205.svg
at java.net.URI$Parser.fail(URI.java:2848)
at java.net.URI$Parser.checkChars(URI.java:3021)
at java.net.URI$Parser.parseHierarchical(URI.java:3105)
at java.net.URI$Parser.parse(URI.java:3053)
at java.net.URI.<init>(URI.java:588)
```

The advice is to encode the url first. Mind you, since we deal with images better now the chance of this error again is low. 

The bug is described more here: https://trello.com/c/L1d7z4rN/119-embedded-svg-causes-internal-server-error

This only protects against very old articles that are not embedding images correctly

Nope!
## Request for comment

@guardian/dotcom-platform

### DIFF
--- a/common/app/views/support/HtmlCleaner.scala
+++ b/common/app/views/support/HtmlCleaner.scala
@@ -1,7 +1,7 @@
 package views.support
 
 import java.text.Normalizer
-import java.net.URI
+import java.net.{URI, URLEncoder}
 import java.util.regex.{Matcher, Pattern}
 import common.{Edition, LinkTo}
 import conf.switches.Switches._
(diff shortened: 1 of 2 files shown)


gtrufitt:
👍

gtrufitt (approved):

johnduffell:
67 is the end, are you sure there's not a trailing control character?

NataliaLKB:
After taking with some people on the grid, we do encode urls already. This change isn't needed and I will close this PR.

I think the only time this will be an issue again (if ever) is on very old articles and if those articles contain images that are not in the grid and have to be edited.

johnduffell:
good detective work there! thanks! :female_detective:

PR #14636, opened 2016-10-10
load youtube js only when you need it

akash1810:
## What does this change?

to avoid a clash when double loading youtube api (e.g. documentary pages)

Lighter pages as only loading lib when needed, working documentary interactives.

No.
## Request for comment

@lps88 @gidsg


gidsg:
👍 💥 Nice to save 8kb on every page!

PR #14637, opened 2016-10-10
Refactor RequestLogger and add response status custom field

TBonnin:
## What does this change?

Refactoring RequestLogger and adding tests.
The goal was to be able to log response info rather than request info only

Being able to log response status so we can graph in kibana per status. Right now the status code is only part of the message, which we cannot parse or tokenize.

I want to use this new status field to investigate redirects loops (301 to same url) that have been recently reported  

## Request for comment

@jfsoul @guardian/dotcom-platform

### DIFF
--- a/common/app/common/RequestLogger.scala
+++ b/common/app/common/RequestLogger.scala
@@ -3,6 +3,8 @@ package common
 import play.api.mvc.{RequestHeader, Result}
 import common.LoggingField._
 
+import scala.util.Random
+
 case class RequestLoggerFields(request: Option[RequestHeader], response: Option[Result], stopWatch: Option[StopWatch]) {
 
   private lazy val requestHeadersFields: List[LogField] = {
@@ -36,7 +38,7 @@ case class RequestLoggerFields(request: Option[RequestHeader], response: Option[
       List[LogField](
         "req.method" -> r.method,
         "req.url" -> r.uri,
-        "req.id" -> r.id.toString
+        "req.id" -> Random.nextInt(Integer.MAX_VALUE).toString
       )
     }.getOrElse(Nil)
 
--- a/common/app/common/RequestLogger.scala
+++ b/common/app/common/RequestLogger.scala
@@ -1,0 +1,47 @@
+package common
+
+import play.api.mvc.RequestHeader
+import common.LoggingField._
+
+case class RequestLogger(rh: RequestHeader)(implicit stopWatch: StopWatch) extends Logging {
+  private lazy val headersFields: List[LogField] = {
+    val whitelistedHeaderNames = Set(
+      "Host",
+      "From",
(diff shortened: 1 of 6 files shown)


common/app/filters/RequestLoggingFilter.scala
```diff
@@ -1,71 +1,25 @@
...
     val result = next(rh)
-    val requestLogger = RequestLogger(rh)
+    val requestLogger = RequestLogger().withRequestHeaders(rh).withStopWatch(stopWatch)
     result onComplete {
       case Success(response) =>
-        response.header.headers.get("X-Accel-Redirect") match {
-          case Some(internalRedirect) =>
-            requestLogger.info(s"${rh.method} ${rh.uri} took ${stopWatch.elapsed} ms and redirected to $internalRedirect")
-          case None =>
-            requestLogger.info(s"${rh.method} ${rh.uri} took ${stopWatch.elapsed} ms and returned ${response.header.status}")
```
TBonnin: Removing redundant `elapsed time` and `response.status` in message since they are passed as custom fields

common/app/filters/RequestLoggingFilter.scala
```diff
@@ -1,71 +1,25 @@
...
+        val additionalInfo = response.header.headers.get("X-Accel-Redirect") match {
+          case Some(internalRedirect) => s" - internal redirect to $internalRedirect"
+          case None => ""
         }
-
+        requestLogger.withResponse(response).info(s"${rh.method} ${rh.uri}$additionalInfo")
       case Failure(error) =>
         requestLogger.warn(s"${rh.method} ${rh.uri} failed after ${stopWatch.elapsed} ms", error)
```
jfsoul: can we remove the stopwatch here too?

common/app/common/RequestLogger.scala
```diff
@@ -0,0 +1,92 @@
...
+    val whitelistedHeaders = allHeadersFields.filterKeys(whitelistedHeaderNames.contains(_))
+    val guardianSpecificHeaders = allHeadersFields.filterKeys(_.toUpperCase.startsWith("X-GU-"))
+    (whitelistedHeaders ++ guardianSpecificHeaders).toList.map(t => LogFieldString(s"req.header.${t._1}", t._2))
+  }
+  private lazy val customFields: List[LogField] = {
+    val requestHeaders: List[LogField] = request.map { r: RequestHeader =>
+      List[LogField](
+        "req.method" -> r.method,
+        "req.url" -> r.uri,
+        "req.id" -> Random.nextInt(Integer.MAX_VALUE).toString
```
jfsoul: Any reason not to prefer a `UUID` here?

jfsoul (approved):
good to go except for minor comments 👍

PR #14638, opened 2016-10-11
fixed video page bug with explore and immersive

blongden73:
Fixing this problem. ->

picture 127

Immersive also needs to otherwise it trys to use the immersive template second. 

@SiAdcock how does this look?


SiAdcock:
👍

PR #14639, opened 2016-10-11
add geo locaiton end point

jranks123:
## What does this change?

This PR adds an endpoint to the onward project which returns the geo location of whoever calls it. It is needed for filtering contributions ads to specific countries, as it has been decided that filtering based on edition is not good enough for requirements. 

It means that when, for example, we show a targeted ad to the US, we can be sure that everyone seeing that ad is from the US. A not tiny amount of people reading the US edition are not from the US, and also a not tiny amount of people from the US do not read the US edition, so by targeting on geo-location we can reach more of the right people with our ads. 

No

## Screenshots

geo
## Request for comment

@dominickendrick @stephanfowler @guardian/dotcom-platform

### DIFF
--- a/onward/app/controllers/GeoLocationController.scala
+++ b/onward/app/controllers/GeoLocationController.scala
@@ -1,0 +1,20 @@
+package controllers
+
+import common.JsonComponent
+import model.{Cached}
+import play.api.mvc.{Action, Controller}
+import scala.concurrent.duration._
+
+class GeoLocationController extends Controller {
+
+
+  def getGeoLocation() = Action { implicit request =>
+    val headers = request.headers.toSimpleMap
+    val countryCode = headers.getOrElse("X-GU-GeoLocation","country:row").replace("country:","")
+    Cached(4*7.days) {
+      JsonComponent(
+        "country" -> countryCode
+      )
+    }
+  }
+}
--- a/dev-build/conf/routes
+++ b/dev-build/conf/routes
@@ -101,6 +101,8 @@ GET            /most-read/*path.json                                            
 GET            /most-read-geo.json                                                                                               controllers.MostPopularController.renderPopularGeo()
 GET            /most-read-day.json                                                                                               controllers.MostPopularController.renderPopularDay(countryCode)
 
+GET            /user-geo-location.json                                                                                           controllers.GeoLocationController.getGeoLocation()
+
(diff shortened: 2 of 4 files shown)


johnduffell:
did you consider just doing synthetic responses from fastly? as here we are just getting some information from fastly and sending it back to the user.  Also mind your caching, as it would have to be split per country.  Also I'm sure a much longer cache time would be appropriate. Using synthetic responses would effectively be an infinite cache time, but >24 hours would seem more appropriate.

jranks123:
@johnduffell I have changed the cache to ```Cached(4*7*24*60*60)```. Is this OK?

johnduffell:
sure, I'd be tempted to put it in a constant called FOUR_WEEKS or something, but it's pretty much clear enough

TBonnin:
Synthetic response from Fastly is not a bad idea. I had not thought about that.

onward/app/controllers/GeoLocationController.scala
```diff
@@ -0,0 +1,19 @@
...
+import model.{Cached}
+import play.api.mvc.{Action, Controller}
+
+class GeoLocationController extends Controller {
+
+
+  def getGeoLocation() = Action { implicit request =>
+    val headers = request.headers.toSimpleMap
+    val countryCode = headers.getOrElse("X-GU-GeoLocation","country:row").replace("country:","")
+    Cached(4*7*24*60*60) {
```
TBonnin: You can use `scala.concurrent.duration` (ie: `4 * 7.days`)

TBonnin (changes requested):

jranks123:
@TBonnin I have made the change you requested, not sure what the process is with this new Changes requested thing now, do you have to re-review it? And @TBonnin @johnduffell , are we happy with this was of doing it or should we do synthetic response from Fastly? @TBonnin if I remember we were going to add something to Fastly that would have the same result of protecting our servers, so are there any other benefits to using the synthetic response method?

TBonnin:
Using a synthetic response means we don't do anything server side (ie: this PR is not necessary anymore). The request would be handled by Fastly directly without having to reach our servers.
Since we need to add some custom logic in the VCL anyway I think it makes sense to handle this endpoint entirely in the VCL

jranks123:
@TBonnin Thank you, yes that sounds like a great idea!

PR #14641, opened 2016-10-11
Edition wins over geoip for membership landing pages

mariogalic:
## What does this change?

Makes sure if the user selects edition (US/AU/Int), and then clicks on 'Become a supporter' in the header,  the corresponding Supporter landing page is selected on the basis of the edition as opposed to fastly geoip. 

See [membership-frontend: Info.scala#L22](https://github.com/guardian/membership-frontend/blob/c932329b57d19eee93bc522d6ebe2001e814f553/frontend/app/controllers/Info.scala#L22)
## Request for comment

@paulbrown1982 @JuliaBellis


paulbrown1982:
👍

PR #14643, opened 2016-10-11
Remove the obstructive adblocker popover.

JustinPinner:
## What does this change?

Removes the adblocker blocking test v3

Throwing out he code we don't need anymore. If everything still works - SUCCESS!

No. Web only.
## Screenshots

N/A
## Request for comment

@Calanthe @regiskuckaertz @kenlim and anyone else who's shown an interest in the past.


regiskuckaertz:
That's a biggie 👍

PR #14618, opened 2016-10-07
Add cleaner for youtube atoms to add ID and set params

gidsg:
## What does this change?

Adds the ability for frontend to read structured data from youtube media atoms
## What is the value of this and can you measure success?

Changes markup to enable us to utilise YouTube iframe API in future and adds player parameters required for YouTube Player For Publishers (PFP) 
There is an example of a youtube atom live at:
http://www.theguardian.com/science/grrlscientist/2012/aug/07/3
This PR is a prerequisite for enhancing YouTube Atoms in future using Javascript
## Does this affect other platforms - Amp, Apps, etc?

No

## Screenshots

N/A
## Request for comment

CC @akash1810 @markjamesbutler

PR #14644, opened 2016-10-11
Fix Media Atom rendering race condition

gidsg:
## What does this change?

The previous YouTube cleaner added in #14618 was intermittently failing due to a race condition whereby it relied on `AtomsCleaner` to have run prior to `YouTubeAtomCleaner` in order to find the DOM elements it depended on.

This PR fixes the issue by moving the rendering of Media atoms into a template called by `AtomsCleaner`. 

YouTube templating is more reliable. Template logic is better encapsulated.

No
## Screenshots

N/A
## Request for comment

@akash1810 @markjamesbutler @TBonnin


common/app/views/fragments/atoms/media.scala.html
```diff
@@ -0,0 +1,14 @@
+@import conf.Configuration.site.host
+@(media: model.content.MediaAtom)(implicit request: RequestHeader)
+@defining(media.assets.head.platform == "Youtube") { isYouTube =>
+    @if(isYouTube) {
+        @defining(if(!host.isEmpty) "&origin=" + host else "") { origin =>
+            <iframe id="youtube-@media.assets.head.id" width="420" height="315"
```
akash1810: can we width and height with css?

akash1810 (approved):

common/app/views/fragments/atoms/media.scala.html
```diff
@@ -0,0 +1,24 @@
+@import conf.Configuration.site.host
+@import views.support.RenderClasses
+
+@(media: model.content.MediaAtom)(implicit request: RequestHeader)
+
+@defining(media.assets.head.platform == "Youtube") { isYouTube =>
```
TBonnin: Can we differentiate YoutubeAtom and MediaAtom and extract this logic out of this template? For instance something like this in `atom.scala.html`
```
model match {
    case quiz: Quiz => views.html.fragments.atoms.quiz(quiz, maybeResults = None, false, Nil)
    case media: YoutubeAtom => views.html.fragments.atoms.youtube(media, origin)
    case media: MediaAtom => views.html.fragments.atoms.media(media)
    case _ =>		        
```

TBonnin (changes requested):

TBonnin (approved):

PR #14646, opened 2016-10-11
Include server side tests whose name starts with commercial

rich-nguyen:
Adding another clause to include tests that affect commercial. This will allow us to read these test participations from our viewability report


Calum-Campbell:
👍

PR #14648, opened 2016-10-11
Some small improvements to hosted code

kelvin-chappell:
- Got rid of some redundant properties
- Made hosted campaign builder to avoid repetition
- Moved all the code we can get rid of when hardcoded campaigns come to an end to the hardcoded package to make it easy to spot
- Made overridden properties easier to spot


lps88:
Looks good 👍

PR #14650, opened 2016-10-11
Prepare ten percent supporter message AB test

JustinPinner:
## What does this change?

Replaces the 100% messaging test that expired last week with a 10% relaxed constraints version, as per discussion with Jesse. This can run its course while we're shifting the original messaging test out of AB and into a feature.

By reaching a larger, and possibly more diverse audience segment we hope to measure an uplift in received contributions, even though the percentage conversion rate observed during the previous iterations will likely decrease.

No. Web only.
## Screenshots

N/A (same as original tests)
## Request for comment

@rtyley


JustinPinner:
Closing as we're going to have to extract the banner back out to a feature and _then_ apply a 10% test.

PR #14651, opened 2016-10-12
Boot: remove preferences test

regiskuckaertz:
All ads are https-compliant now, so there is no need for that test anymore. Also, minor refactoring.

cc @gtrufitt


gtrufitt (approved):
👌

PR #14653, opened 2016-10-12
Code cleanup: remove DFP-related badge/logo logic

regiskuckaertz:
Now that badges have been served via the Tag Manager for a few weeks, we can safely discard anything related to badges served via DFP.

Do @Calanthe and @kelvin-chappell agree with this?


kelvin-chappell (approved):
It looks redundant to me. 👍

PR #14654, opened 2016-10-12
Refactor and clean up commercial Merchandise models

JonNorman:
## What does this change?

This cleans out a bunch of our commercial merchandise models and removes generally smelly code.

Reduces the amount of code in our commercial models, leverages macros that Play provides for free,  cleans up the commercial codebase.
## What could be done that isn't?

The test for parsing our merchandise could definitely be more a) structured and b) comprehensive. I'll card that out and add it to the commercial Trello board, but it is lower priority than other things currently in play.
## Request for comment

@guardian/commercial-dev


commercial/app/model/merchandise/Merchandise.scala
```diff
@@ -89,26 +83,18 @@ case class TravelOffer(id: String,
...
 
   val profileId: Option[String] = profilePhoto match {
     case Member.IdPattern(id) => Some(id)
     case _ => None
   }
 
-  val profileUrl: String = profileId.map(id => s"https://soulmates.theguardian.com/landing/$id")
-    .getOrElse("http://soulmates.theguardian.com/")
-
+  val profileUrl: String = s"https://soulmates.theguardian.com/" + ( profileId.map(id => s"landing/$id") getOrElse "" )
```
regiskuckaertz: Can you get rid of the `s` here?

JonNorman: Yep - spotted that one but have just noticed that a bunch of files have stopped compiling after my merge so need to fix that first :(

commercial/app/controllers/Multi.scala
```diff
@@ -128,7 +128,7 @@ class Multi(bestsellersAgent: BestsellersAgent,
         Cached(componentMaxAge) {
           JsonComponent(JsArray((offerTypes zip contents).map { case (contentType, content) => Json.obj(
               "type" -> contentType,
-              "value" -> Json.toJson(content)(Merchandise.writes)
+              "value" -> Json.toJson(content)(Merchandise.merchandiseWrites)
```
kelvin-chappell: Do you need to specify the writes implementation?  It should be able to work it out.

commercial/app/model/merchandise/Merchandise.scala
```diff
@@ -89,26 +83,18 @@ case class TravelOffer(id: String,
...
 
   val profileId: Option[String] = profilePhoto match {
     case Member.IdPattern(id) => Some(id)
     case _ => None
   }
 
-  val profileUrl: String = profileId.map(id => s"https://soulmates.theguardian.com/landing/$id")
-    .getOrElse("http://soulmates.theguardian.com/")
-
+  val profileUrl: String = "https://soulmates.theguardian.com/" + ( profileId.map(id => s"landing/$id") getOrElse "" )
```
kelvin-chappell: Better to use string interpolator.

commercial/app/model/merchandise/Merchandise.scala
```diff
@@ -134,8 +120,29 @@ case class Job(id: Int,
...
+}
+
+case object Man extends Gender {
+  val name: String = "Man"
+}
+
 object Merchandise {
-  val writes: Writes[Merchandise] = new Writes[Merchandise] {
+
+  val merchandiseWrites: Writes[Merchandise] = new Writes[Merchandise] {
```
kelvin-chappell: This looks a bit odd.  I'd have thought books and jobs know how to represent themselves.

commercial/app/model/merchandise/Merchandise.scala
```diff
@@ -203,35 +210,27 @@ object Masterclass {
...
-      }
-    }
-
     extractGuardianUrl map { extractedUrl =>
 
       new Masterclass(
         id = event.id,
         name = event.name,
-        startDate = event.startDate,
+        startDate = event.start_date,
```
kelvin-chappell: Why has the property become snake-case?

commercial/app/model/merchandise/events/Eventbrite.scala
```diff
@@ -3,185 +3,108 @@ package model.commercial.events
...
-  case class EBResponse(pagination: EBPagination, events: Seq[EBEvent])
-
-  object EBResponse {
+  case class Response(pagination: Pagination, events: Seq[Event])
+
+  case class Pagination(page_number: Int, page_count: Int)
+
+  case class Event(id: String,
+                   name: String,
+                   start_date: DateTime,
```
kelvin-chappell: Some strange snake-cases here.

kelvin-chappell (changes requested):
Looks good but some strange snake-cases creeping in, which looks like forcing the code to fit in with a template.

commercial/app/model/merchandise/events/Eventbrite.scala
```diff
@@ -3,185 +3,108 @@ package model.commercial.events
...
 import scala.util.control.NonFatal
 
 object Eventbrite extends ExecutionContexts with Logging {
 
-  case class EBResponse(pagination: EBPagination, events: Seq[EBEvent])
-
-  object EBResponse {
+  case class Response(pagination: Pagination, events: Seq[Event])
+
+  case class Pagination(page_number: Int, page_count: Int)
```
TBonnin: no camelCase?

commercial/app/controllers/Multi.scala
JonNorman: Unfortunately, despite using a `sealed trait` - the compiler doesn't seem able to infer that there is a Json serializer for all implementations of `Merchandise`, thus it complains `No Json serializer found for type model.commercial.Merchandise.`.

It's a bit annoying since the whole point of sealed traits is to allow this kind of compilation inference :/

commercial/app/model/merchandise/events/Eventbrite.scala
```diff
@@ -199,21 +122,28 @@ object Eventbrite extends ExecutionContexts with Logging {
     }
   }
 
-  trait EBTicketHandler {
-    def tickets: Seq[EBTicket]
+  trait TicketHandler {
+    def tickets: Seq[Ticket]
+
+    lazy val displayPriceRange = {
```
TBonnin: why a lazy val here?

commercial/app/model/merchandise/Merchandise.scala
JonNorman: So, as you alluded to above, I altered the `Eventbrite` model to reflect the Eventbrite feed values  to allow us to utilise the Json macros. In the case of `Eventbrite.Event`, we have an explicit `Reads` method so it doesn't actually help here, but in other cases, such as `Eventbrite.Pagination` and `Eventbrite.Response` the snake-case helps us out by letting Play read it implicitly to our case class. I decided therefor to match all the Eventbrite case class variables to their real-feed names, type and all. 

If you reckon that this isn't worthwhile then I'm happy to take it out and write explicit reads, just thought I'd throw it in there :)

commercial/app/model/merchandise/events/Eventbrite.scala
JonNorman: See https://github.com/guardian/frontend/pull/14654#discussion_r82986121

Happy to reinstantiate the camel case; having looked at the way that Membership parse Eventbrite models, they do it with snake-case as I've done here. We can use camel case, but it does make sense to make our case classes reflect as closely as possible the actual feed, since they are inextricably linked. Happy to get rid of that though if it isn't deemed useful enough.

commercial/app/model/merchandise/events/Eventbrite.scala
JonNorman: I believe that this is a trait/instantiation thing wherein, but I can't actually remember a good reason for it. I'll dig in and get back to you!

JonNorman: Okay - I can't see/find a reason that we need this to be lazy, so I'm taking it out.

commercial/app/model/merchandise/Merchandise.scala
kelvin-chappell: I think if you're going to convert something into a Scala case class you should probably follow the Scala conventions.  Otherwise you could do [coast-to-coast json](https://www.playframework.com/documentation/2.5.x/ScalaJsonTransformers) instead.  But if there's a good reason for abandoning the convention you should put a comment on the field to avoid someone refactoring it later.

JonNorman: Okay I'll take another look at it and see if I can make it prettier/simpler.

JonNorman:
Okay - I've removed the lazy vals and I've reintroduced the camelCase and written explicit reads for the `Pagination` and `Response` Eventbrite models!

kelvin-chappell (approved):

PR #14632, opened 2016-10-10
Add hosted content dimension to GA

gtrufitt:
## What does this change?

Adds `isHostedFlag` or `dimension26` to the page level GA tracking.
## What is the value of this and can you measure success?

Hosted content filtering in GA.
## Does this affect other platforms - Amp, Apps, etc?

Yes, added to AMP.

@guardian/dotcom-platform @guardian/labs-beta @mkopka

PR #14655, opened 2016-10-12
fail gracefully if `isHosted` property doesn't exist

akash1810:
## What does this change?

Relates to https://github.com/guardian/frontend/pull/14632

`isHosted` doesn't seem to be set on interactive pages, this fixes the `TypeError: Cannot read property 'toString' of undefined` being thrown when `isHosted` is not on the page config e.g. here - https://www.theguardian.com/environment/ng-interactive/2015/nov/26/the-mekong-river-stories-from-the-heart-of-the-climate-crisis-interactive.

Fewer JS errors.

No.

## Screenshots

n/a
## Request for comment

@gtrufitt


gtrufitt:
Whoops &  Thanks! 👍

PR #14657, opened 2016-10-12
Cache clear tool improvements

jfsoul:
## What does this change?

Updates the cache clearing tool so that its intended use and functionality is more explicit.  Specifically the copy has been updated to include warnings and further explanation.  There is also a pre-purge test step, which makes a request to the specified path (directly on router), which will give a further warning if unsuccessful (but you can still perform the purge, e.g. for takedowns).

I did consider merging this tool with the troubleshoot tool, but I think that works well as it is, and this makes the dangers of using the purge tool more explicit.

Less likely to misuse the cache purging tool, and will at least get a warning before purging content that cannot be replaced.

No
## Screenshots

image
image
image
## Request for comment

@TBonnin @guardian/dotcom-platform


admin/app/controllers/cache/PageDecacheController.scala
```diff
@@ -9,31 +9,62 @@ import controllers.admin.AuthActions
...
+    }
+  }
+
+  private def renderPrePurgeTestResult(purgeUrl: String)(implicit request: Request[AnyContent]) = {
+    getRouterUrl(purgeUrl).map { routerUrl =>
+      wsClient.url(routerUrl)
+        .withRequestTimeout(2000)
+        .withVirtualHost("www.theguardian.com")
+        .get().map {
+          response => PrePurgeTestResult(purgeUrl, response.status == 200)
```
TBonnin: What if the response is some `2XX` or a `3XX`?

admin/app/controllers/cache/PageDecacheController.scala
```diff
@@ -9,31 +9,62 @@ import controllers.admin.AuthActions
...
+
+      } recoverWith {
+        case t: Throwable => successful(PrePurgeTestResult(purgeUrl, passed = false))
+      } map { result =>
+        NoCache(Ok(views.html.cache.pageDecache(Some(result))))
+      }
```
TBonnin: Formatting: I prefer not to mix together dot and space notations.
Also I also prefer to have one method call per line.
What do you think?

admin/app/controllers/cache/PageDecacheController.scala
```diff
@@ -9,31 +9,62 @@ import controllers.admin.AuthActions
...
     getSubmittedUrl(request).map(new URI(_)).map{ urlToDecache =>
 
       new CdnPurge(wsClient).hard(SurrogateKey(urlToDecache.getPath))
-      successful(NoCache(Ok(views.html.cache.pageDecacheForm())))
-    }.getOrElse(successful(BadRequest("No image submitted")))
-
+        .map(purgeSent => {
+          val message = if(purgeSent) "Purge request successfully sent" else "Purge request was not successful, please report this issue"
+          NoCache(Ok(views.html.cache.pageDecache(None, message)))
+        })
```
TBonnin: you can get rid of the "internal" `{` and replace the map `(` with `{`

admin/app/controllers/cache/PageDecacheController.scala
```diff
@@ -9,31 +9,62 @@ import controllers.admin.AuthActions
...
-  private def getSubmittedUrl(request: AuthenticatedRequest[AnyContent, UserIdentity]): Option[String] = request
-    .body.asFormUrlEncoded
-    .getOrElse(Map.empty)
-    .get("url")
-    .flatMap(_.headOption)
-    .map(_.trim)
+  private def getRouterUrl(url: String): Option[String] =
+    LoadBalancer("frontend-router")
+      .flatMap(_.url)
+      .map(router => url.replaceFirst("^https?:\\/\\/www\\.theguardian\\.com", s"http://$router"))
```
TBonnin: what about using `URL` type rather than a regex here?

TBonnin (changes requested):

johnduffell:
discussed in person - personally I think we should only give them soft purge, if anything at all (subject to client error 4xx not being classed as an backend error by fastly)

jfsoul:
Looks like we're safe on assuming 4xx aren't considered backend errors.

https://github.com/guardian/fastly-edge-cache/blob/6fbd328ed24a1abefc4de80d462618f1f8df2d72/theguardiancom/src/main/resources/varnish21/error.vcl#L14

admin/app/controllers/cache/PageDecacheController.scala
jfsoul: Yes, will change this - indentation is more natural that way as well

TBonnin:
I am fine with those change, as long as we confirm that CP never have to do a hard purge.

jfsoul:
Checked in with CP, and the opinion is that soft purge is enough and they are happy with the changes.

Thanks for comments @TBonnin @johnduffell, I should have incorporated them all.

TBonnin (approved):

PR #14658, opened 2016-10-12
Goodbye grunt pt1 (redux)

sndrs:
## What does this change?

Begins the process (again) of removing grunt, by moving the top-level compile tasks out of the `gruntfile` into `tools/assets/`.

asset pipeline tasks can now be defined in their own modules, as either:
- strings (interpreted as a commands)
- javascript functions
- arrays of subtasks

It also provides a very lightweight wrapper around [listr](https://github.com/SamVerschueren/listr) to manage running the tasks, and to create a cli UI, in lieu of grunt. 

 This is a diff of `static/hash` and `common/conf/assets/` in this branch and master, after running `make compile`:

:pray:

```
diff -r branch master
Only in branch/hash: .DS_Store
```

If this is accepted, I'll remove the rest of grunt in subsequent PRs
- it will remove our reliance on grunt plugins staying up to date. although this wraps the tasks in `listr`, it's not a requirement. 
- compiling without grunt ~~will~~ should be much faster (stats when complete)
- compilation steps are just commands, functions, or lists of either

no
## Request for comment

@guardian/dotcom-platform @rich-nguyen


SiAdcock:
I literally thought you had implemented `redux` in frontend for a second 😄

sammorrisdesign:
Can we add `Reticulating Splines` to the cli list of tasks?

gtrufitt:
This is really cool. I had an initial concern that we were going to be rebuilding grunt when I saw `runner.js` but like you say, the wrappers here are very slim and `Listr` is a really nice DX improvement (though sometimes it's 'nice' to see that stuff is happening and the spinner isn't just stuck (svgs)!).

Do you have any concerns around this growing to the point where we have built a task runner? I.e one day will we be like, hmm, it'd be nice to stream this task into the next task? And is that a big deal if it's still focussed on exactly our needs?

Otherwise, looks great man!

sndrs:
> Do you have any concerns around this growing to the point where we have built a task runner? I.e one day will we be like, hmm, it'd be nice to stream this task into the next task? And is that a big deal if it's still focussed on exactly our needs?

yeah i do, i guess this PR was designed to show what it would do, weigh up the options.

i think the fact that it's just commands or functions (or arrays of either) means it's trivial to strip them out if we hate it or it becomes too complex to manage. if this gets merged I'll remove all the other grunt plugins and tasks, so we are only building using the basic tools, deps are minimised. 

i'll be happy to pull it all back to the makefile if listr ends up getting the way or becomes too flaky etc

sndrs:
as an adendum – I guess there's a lot of value in grunt, especially the feedback in the console and ease of running tasks. 

the issues seem to me to be:
- needing to use it's own plugins rather than the actual thing they wrap
- the complexity of the config
- execution time

hopefully, this should address those while keeping reasonable feedback available...that's the plan anyway!

gtrufitt (approved):
Totally agree with the analysis of the problems with our Grunt set up. Sounds good anyway, I actually have no comments on code or structure, all lgtm! 👍

sndrs:
fyi @SamVerschueren

SamVerschueren:
Not sure if I have to respond to anything. Nice to see `Listr` being used though ;).

> Though sometimes it's 'nice' to see that stuff is happening and the spinner isn't just stuck (svgs)!).

If a task is able to emit output, you could look into `Observables` which can be used inside a listr task. Every value that the `Observable` emits, will get printed under the task description. More info in the [observable section](https://github.com/samverschueren/listr#observable).

PR #14660, opened 2016-10-12
Repackage Commercial

JonNorman:
Part II of https://github.com/guardian/frontend/pull/14622

_and breathe out..._

Following the rip-roaring success of changing our directory structure, breaking dev-build and causing myself merge-hell , this is the "final" PR that sorts out all the packaging in commercial and makes it a) reflect the directory structure and b) reflect a useful partitioning of our models.

Things to note:

a) When using commercial models in views, it is necessary to explicitly import the models as required. Without doing this, the automatic default imports (`models._` and `views.support._` can cause some conflicts. This is a bit annoying but it does mean that IntelliJ can do it's thing when looking at the template files - before the lack of an explicit import caused it to throw a wobbly).

**If this looks worse than what we currently have, then I'm happy to drop these changes, but they look a lot clearer to me**

@guardian/commercial-dev


commercial/app/views/books/book.scala.html
```diff
@@ -1,5 +1,5 @@
-@(book: model.commercial.Book, omnitureId: Option[String], clickMacro: Option[String])(implicit request: RequestHeader)
-@import views.support.{ImgSrc, Item300}
+@import _root_.commercial.model.merchandise.Book
```
janua: Why do you now need `_root_`, but didn't before?

JonNorman: Whenever I try to use `import commercial.model.merchandise.Book` I am informed by the compiler that there is an ambiguous reference to `commercial`. I believe that this is because. on compilation, all templates are compiled with [default imports](https://github.com/guardian/frontend/blob/master/project/Prototypes.scala#L75). I think that there may be a subpackage of `commercial` within either the `model` package or the `views.support` package.

It is immensely frustrating. Perhaps I am missing something but the only way to directly reference the commercial package I want was to add the `_root_` prefix.

janua: But if commercial has now been removed from `model`, then this won't happen? Did you verify by removing the import?

It might be because of the project `commercial` gets a namespace of it's own during compilation. I have come up against this before (although not the exact same situation), it's very frustrating to know where the other reference is coming from. You could print it out!

kelvin-chappell (approved):
👍

JonNorman:
I'm going to close this - there are other changes that need to be made and they can be made independently, in separate PRs, which is what I'll do now...

PR #14661, opened 2016-10-12
Fix if statement in article-aside-adverts

janua:
## What does this change?

This fixes a bug where we used the constrains of an immersive article on a normal article because of `if` logic.

This fixes the `if` logic to never use the immersive constraints on a normal article, and vice versa.

No

## Request for comment

@regiskuckaertz @Ap0c


janua:
['Most Popular' being cropped on short articles](https://trello.com/c/xcCEfylK/64-most-popular-being-cropped-on-short-articles)

regiskuckaertz (approved):

regiskuckaertz:
The test is broken because _now_ the `adType` value is correct 😆

Ap0c:
You fixed it, nice! 💥

PR #14663, opened 2016-10-12
Convert membership banner test back to a feature

JustinPinner:
## What does this change?

Removes the AB test for the membership engagement banner and makes it a feature again, using static message text but adhering to the original conditions that the AB test ran under.

This had been run as a 100% AB test (which is a little bit naughty) so this addresses that. We do want to run more tests with it, but we should do that over the feature.

No. Web only.
## Screenshots

I'll add a couple of screen caps soon.
## Request for comment

@NataliaLKB @rtyley


common/app/conf/switches/CommercialSwitches.scala
```diff
@@ -206,6 +206,16 @@ trait CommercialSwitches {
     exposeClientSide = false
   )
 
+  val MembershipEngagementBanner = Switch(
+    SwitchGroup.Commercial,
```
NataliaLKB: Is this a commercial switch?

NataliaLKB: And if so, why? I thought membership would be different from commercial

NataliaLKB:
Minor comments and questions, but super happy about the tests!

common/app/conf/switches/CommercialSwitches.scala
JustinPinner: I wondered that too, but a) membership doesn't have its own switch group I think and b) this whole thing is a subset of commercial, presumably because it's all about the revenue.

JustinPinner:
Just the merge conflict to fix now then 🙄

common/app/conf/switches/CommercialSwitches.scala
NataliaLKB: I just chatted with @SiAdcock and @gtrufitt. Personally I think it makes sense for membership to have their own switch group as it is one of the core things we are doing now. However, it is completely up to the membership team to decide this and implement it. 

Certainly I think it's outside the scope of this PR.

rtyley:
:+1: apart from the minor point on `code: messageCode`, ` minVisits: minVisits` getting duplicated. The next step is to reintroduce the A/B test for 10%, right?

JustinPinner:
@rtyley yes, this makes the default behaviour visible to 100% of users. Another PR will introduce a new AB test to change things for the 10 (or however many) percent.

NataliaLKB:
👍

PR #14667, opened 2016-10-12
Remove redundant logo styling CSS

scottp719:
## What does this change?

This temporary fix is no longer required as tag manager can now serve two logos, including a high contrast version for media tone pages.

Less code, no mess

Nopers

## Request for comment

@kelvin-chappell


kelvin-chappell (approved):
👍

PR #14668, opened 2016-10-12
Remove high merch targeting check

Calum-Campbell:
## What does this change?
- Remove code and switch for server side check of high merchandising slot.
- Ultimately it seems too difficult to identify wether or not the page will require a high merchandising slot or not server side.
- There are too many possible tag combinations for high merchandising slots to reliably decide wether or not to render a high-merchandising slot.  
- For example there may be a high merch line item that targets US edition and a certain URL, you would need to add this case to the checking rules which would not be possible for all possible cases. Or, you could set wide check and render a slot on nearly all pages and hope to not miss any cases and cause a high relevance merchandising slot not to render. This would not reduce the problem much and risk not rendering high merch slots when they should be. 
- Related to v.old PR https://github.com/guardian/frontend/pull/12795
- Remove redundant code!
## Request for comment

@rich-nguyen @guardian/commercial-dev

### DIFF
--- a/applications/app/views/fragments/crosswords/crosswordFooter.scala.html
+++ b/applications/app/views/fragments/crosswords/crosswordFooter.scala.html
@@ -9,7 +9,7 @@
 
     @fragments.onwardPlaceholder(isAdvertisementFeature = false)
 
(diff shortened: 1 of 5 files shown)


janua:
:+1:

JonNorman (approved):
Looks good to me! What was the main difficulty in detecting whether a page required a high-merchandise slot?

PR #14672, opened 2016-10-13
Add control test definition for Sonobi

rich-nguyen:
This allows us to report the viewability of the header bidding control group.


janua:
:+1:

PR #14674, opened 2016-10-13
Ignore ensime config files

kelvin-chappell:


JonNorman (approved):
Are you now using Atom to develop frontend with?

kelvin-chappell:
I'm going to keep an eye on it.

desbo:
Should `.ensime` be ignored? What if I want to try it?

kelvin-chappell:
It seems to be full of directory names etc so it's very machine and user-specific.  You'll have to generate your own!

JonNorman:
I managed to get mine to work (before my mac got nuked), it took a bit of faffing about but eventually I got go-to functionality in Atom with it.

PR #14675, opened 2016-10-13
Remove grunt require

sndrs:
## What does this change?

stops bundling the JS in grunt, does it directly with `r.js`.
- removes dependency on grunt requirejs plugin
- halves js compilation time on my machine (`46s` vs `22s`)
- since it bundles a newer version of uglify, bundles are fractionally smaller, but the unminified output is no different

no
## Screenshots

with grunt:
grunt

without grunt:
new
## Request for comment

@guardian/dotcom-platform


NataliaLKB:
Looks like you need to run `make validate`. The changes lgtm 👍

sndrs:
>Looks like you need to run make validate.

whoops - indeed i did #14678

PR #14678, opened 2016-10-13
missed this in the labyrinth

sndrs:
## What does this change?

fixes an oversight in #14658

temporary work around for `grunt validate:css`

no
## Request for comment

@NataliaLKB 

temporay workaround


NataliaLKB:
👍 !

PR #14683, opened 2016-10-13
A 24 hour AB test of immediate display of engagement banner

JustinPinner:
## What does this change?

Introduces a 24 hour AB test to probe the hypothesis that showing the engagement banner to more users sooner, rather than waiting until 10 pages had been read, can result in additional contributions being made.

This is aimed specifically at UK edition readers, with a 20/80 split (20% see the messaging immediately, 80% follow the normal 10 article rule). We've tagged each variant so we can meaningfully track the effects over this relatively short timeframe.

**We're aiming to activate this today**, so it can be switched off again before the weekend.

To prove the hypothesis that even with a reduction in percentage of conversions we would still see an uplift in raw contributions made.

No. Web only.
## Screenshots

N/A - uses standard messaging UI
## Request for comment

@rtyley @NataliaLKB

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/membership-engagement-banner.js
+++ b/static/src/javascripts/projects/common/modules/commercial/membership-engagement-banner.js
@@ -84,7 +84,11 @@ define([
 
     function isInTest(testId, variant) {
         var participations = storage.local.get('gu.ab.participations');
-        return variant === undefined ? participations[testId] : participations[testId].variant === variant;
+        if (participations) {
+            return variant === undefined ? participations[testId] : participations[testId].variant === variant;
+        } else {
+            return false;
+        }
     }
 
     function init() {
--- a/static/src/javascripts/projects/common/modules/commercial/membership-engagement-banner.js
+++ b/static/src/javascripts/projects/common/modules/commercial/membership-engagement-banner.js
@@ -82,16 +82,24 @@ define([
         return messageShown;
     }
 
+    function isInTest(testId, variant) {
+        var participations = storage.local.get('gu.ab.participations');
+        return variant === undefined ? participations[testId] : participations[testId].variant === variant;
+    }
+
     function init() {
+        // block default behaviour for participants of the MembershipEngagementImmediate AB test
         var edition = config.page.edition;
-        var message = messages[edition];
-        if (message) {
-            var userHasMadeEnoughVisits = (storage.local.get('gu.alreadyVisited') || 0) >= 10;
-            return commercialFeatures.async.canDisplayMembershipEngagementBanner.then(function (canShow) {
-                if (canShow && userHasMadeEnoughVisits) {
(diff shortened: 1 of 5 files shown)


static/src/javascripts/projects/common/modules/commercial/membership-engagement-banner.js
```diff
@@ -82,16 +82,24 @@ define([
         return messageShown;
     }
 
+    function isInTest(testId, variant) {
+        var participations = storage.local.get('gu.ab.participations');
```
NataliaLKB: I don't think you have check local storage yourself. You should just be able to do something like this:
`ab.getTestVariantId('nameOfTest') &&  ab.testCanBeRun('nameOfTest') `

JustinPinner: Doesn't that mean importing the whole of ab into this file?

static/src/javascripts/projects/common/modules/experiments/tests/membership-engagement-immediate.js
```diff
@@ -0,0 +1,114 @@
...
+                    siteMessageLinkName: 'membership message',
+                    siteMessageCloseBtn: 'hide',
+                    siteMessageComponentName: id,
+                    trackDisplay: true,
+                    cssModifierClass: cssModifierClass
+                }).show(renderedBanner);
+
+        };
+
+        this.variants = [
```
NataliaLKB: This is a lot of changes to make in this file, when it might be easier (and less code) to use [this](https://github.com/guardian/frontend/pull/10916/files) as an example of how to change your code by variant (if statements) instead of doing it in this file

JustinPinner:
Can I get a 👍 (for when the tests pass)? Thanks!

static/src/javascripts/projects/common/modules/experiments/tests/membership-engagement-immediate.js
JustinPinner: Noted for future reference. Thanks.

paulbrown1982:
👍 from a membership perspective.

rtyley:
:+1:

I'm happy for this to go out, though if this were code that needed to live longer than 24 hours, I think it would be worth removing a lot of the duplication - if you look at these two sets of lines:

* [membership-engagement-immediate.js#L72-L90](https://github.com/guardian/frontend/blob/dc9db489a/static/src/javascripts/projects/common/modules/experiments/tests/membership-engagement-immediate.js#L72-L90)
* [membership-engagement-immediate.js#L94-L111](https://github.com/guardian/frontend/blob/dc9db489a/static/src/javascripts/projects/common/modules/experiments/tests/membership-engagement-immediate.js#L94-L111)

...you'll see that they're exactly the same, and a refactoring could get rid of that!

PR #14684, opened 2016-10-13
Logging form submission user errors as log.info

TBonnin:
## What does this change?

Logging form submission user errors as log.info

There are user errors, not errors on our side.

Logging them as log.info so they doen't pollute our server-side error
logs

## Request for comment

@gtrufitt @janua


janua:
:+1:

PR #14686, opened 2016-10-13
Upgrade config version (training-preview: removing ophan.api key)

TBonnin:
## What does this change?

Upgrading config version

Removing Ophan.api keys in config for training preview, rather than empty string, so training-preview doesn't throw error when trying to fetch ophan url without a scheme and host

## Request for comment

@jfsoul


jfsoul:
👍

PR #14690, opened 2016-10-14
A 24 hour AB test of immediate display of engagement banner (revised)

JustinPinner:
## What does this change?

This is a revision to the previous PR https://github.com/guardian/frontend/pull/14683

Introduces a 24 hour AB test to prove the hypothesis that showing the engagement banner to users sooner, rather than waiting until 10 pages had been read, can result in additional contributions being made.

This is aimed specifically at UK edition readers, drawing from 40% of the audience, (20% see the messaging immediately while 20% follow the normal 10 article rule). We've tagged each variant so we can meaningfully track the effects over this relatively short timeframe.

To prove the hypothesis that even with a reduction in percentage of conversions we would still see an uplift in raw contributions made.

No. Web only.
## Screenshots

N/A - uses standard messaging UI
## Request for comment

@rtyley (I removed some duplication) @NataliaLKB (I modified the audience selection & dates)

### DIFF
--- a/common/app/conf/switches/ABTestSwitches.scala
+++ b/common/app/conf/switches/ABTestSwitches.scala
@@ -26,13 +26,13 @@ trait ABTestSwitches {
     exposeClientSide = true
   )
 
-  for (edition <- Edition.all) Switch(
+  Switch(
     ABTests,
-    "ab-membership-engagement-banner-extended-"+edition.id.toLowerCase,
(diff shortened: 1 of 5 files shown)


SiAdcock:
LGTM 👍

JustinPinner:
Thanks @SiAdcock! Time to merge.

PR #14694, opened 2016-10-14
Static deploy: don't prefix with stack

TBonnin:
## What does this change?

Static deploy: don't prefix with stack

Static deployment works :/

## Request for comment

@guardian/dotcom-platform

Riff-Raff uses the default stack to prefix the file key.
Since we don't want this prefix we need to set prefixStack to false


jfsoul:
👍

PR #14695, opened 2016-10-14
Check node version

sndrs:
## What does this change?

checks current node version matches `.nvmrc` when running `make` tasks.

stops issues where things don't seem to work because you're using the wrong version of node.

no
## Screenshots

screen shot 2016-10-14 at 12 05 58
## Request for comment

@guardian/dotcom-platform


gtrufitt (approved):
👍

PR #14697, opened 2016-10-14
Provide commercial support for native LiveEvent merchandise component

JonNorman:
### Background

We have a number of templates in frontend used to render specific merchandise components on the site. There is ongoing work to remove these templates and ultimately have all of these fragments stored directly in DFP. When being served from frontend, the fragments can call on the merchandise agents, grab a specific LiveEvent/Book/Job/Soulmate and use it/them to render a template in one fell swoop.

The new native merchandise component (that will sit in DFP) also needs to be able to render itself using specific Live Event information. Given that it will not have access to the frontend agents, it will have to fetch it client side; hence the new JSON route in this PR.

Eventually the `renderLiveEvent` method will be removed when we are happy with the native components as no rendering of components will be done within frontend.
### Changes

This tidies up the LiveEvent controller in general and provides a new method for getting specific live events back as raw JSON rather than renderable HTML.

This also updates the `LiveEvent` model to keep an `ImageInfo` rather than a simple Image URL. This allows the recipient of the LiveEvent JSON to use `srcsets` and be responsi(ve/ble).

Nope
## Request for comment

@guardian/commercial-dev

### DIFF
--- a/commercial/app/controllers/LiveEventsController.scala
+++ b/commercial/app/controllers/LiveEventsController.scala
@@ -33,4 +33,8 @@ class LiveEventsController(liveEventAgent: LiveEventAgent)
       Future.successful(NoCache(jsonFormat.nilResult.result))
     }
   }
+
+  def getLiveEvent = Action { implicit request =>
+    ???
+  }
 }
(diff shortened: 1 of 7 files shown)


regiskuckaertz:
Lovely 👍

JonNorman:
[Inline: Live Event](https://trello.com/c/uRleu5cE/38-inline-live-event)

PR #14702, opened 2016-10-14
Fix matching of retriveContent; it expects a List; make it so.

JonNorman:
We have converted a bunch of the `Seq`s to `List`s within the Merchandise component controllers. We missed one.
@Ap0c


Ap0c (approved):
Looks good, the old, pre-native routes don't seem to be affected (probably because the controller already maps them to a list anyway).

PR #14705, opened 2016-10-15
Fixed background in native ads

regiskuckaertz:
Fixed backgrounds inside iframes are impossible. Of course, because inside the iframe is another view and any fixed-positioned element will be relative to this view, not the browser viewport. Still, we could imagine an absolutely positioned element the size of the browser viewport, its top coordinate updated on every scroll; this turns out to create a subpar experience, even on desktop.

So, we have no other choice but to delegate the fixed positioned background of a native ad to the parent frame. This is the purpose of this PR.

fixed-background


regiskuckaertz:
[Fabric native: smooth fixed background effect](https://trello.com/c/YYevCWyH/85-fabric-native-smooth-fixed-background-effect)

JonNorman (approved):

PR #14706, opened 2016-10-15
Parallax background in native ads

regiskuckaertz:
On [the same note](https://github.com/guardian/frontend/pull/14705), here is the routine dealing with parallax background effects.

More effects can be added later, the possibilities, like the sky, is infinite 😆 

source


regiskuckaertz:
[Fabric native: smooth parallax background effect](https://trello.com/c/n37aasY3/86-fabric-native-smooth-parallax-background-effect)

JonNorman (approved):
Lovely stuff; the comments are really useful as well - thanks for putting them in.

PR #14710, opened 2016-10-17
Send iframe coordinates on connect

regiskuckaertz:
To mirror the behaviour of the `viewport` persistent listener, we send the iframe coordinates as soon as it starts listening. This is useful to bootstrap creatives on the opposite end.


regiskuckaertz:
[Migrate Fabric Video](https://trello.com/c/xaAMpcM9/84-migrate-fabric-video)

rich-nguyen:
👍

PR #14711, opened 2016-10-17
Move non-blocking JS and analytics from under head.scala.html

rjmk:
## What does this change?

Change the nesting of the list in the client-side architecture docs

I believe these fragments are loaded directly in `main.scala.html`, not in `head.scala.html`.

No
## Request for comment

@gtrufitt


gtrufitt (approved):
Thanks!

PR #14712, opened 2016-10-17
🔪 Play.current (part 1)

TBonnin:
## What does this change?

Removing Play.current (deprecated in Play 2.5)
There is still a few reference to `Play.current` which are more intricate to untangle. :( I will do those in subsequent PRs

Preparation for Play 2.5 upgrade

## Request for comment

@jfsoul @alexduf 
Note: probably worth reviewing commit by commit ;)


commercial/app/AppLoader.scala
```diff
@@ -77,4 +59,15 @@ trait AppComponents extends FrontendComponents with AppLifecycleComponents with
...
+  lazy val devAssetsController = wire[DevAssetsController]
+  lazy val healthCheck = wire[HealthCheck]
+
+  override lazy val lifecycleComponents = List(
+    wire[LogstashLifecycle],
+    wire[CommercialLifecycle],
+    wire[SwitchboardLifecycle],
+    wire[CloudWatchMetricsLifecycle],
+    wire[CachedHealthCheckLifeCycle]
+  )
```
TBonnin: @alexduf Not related to removing `Play.current` but as we mentioned earlier, simplifying AppLoader.
It makes sense to have traits for controllers and services since they are also used in Standalone app but `lifecycleComponents`, `devAssetsController` and `healthCheck` can be defined in `AppComponents`

common/test/common/TestWsConfig.scala
```diff
@@ -1,21 +0,0 @@
...
-    val c = WS.client.underlying.asInstanceOf[AsyncHttpClient].getConfig
-    val con = new Builder(c)
-      .setConnectTimeout(10000)
-      .setReadTimeout(10000)
-      .setRequestTimeout(10000)
-      .build()
-    new NingWSClient(con)
-  }
-
-}
```
TBonnin: Not used anywhere 🔪

commercial/app/AppLoader.scala
```diff
@@ -56,8 +50,10 @@ trait CommercialServices {
   lazy val feedsParser = wire[FeedsParser]
 }
 
-trait AppLifecycleComponents {
-  self: FrontendComponents with Controllers with CommercialServices =>
+trait AppComponents extends FrontendComponents with CommercialControllers with CommercialServices {
+
+  lazy val devAssetsController = wire[DevAssetsController]
+  lazy val healthCheck = wire[HealthCheck]
```
TBonnin: @alexduf Not related to removing Play.current but as we mentioned earlier, simplifying AppLoader.
It makes sense to have traits for controllers and services since they are also used in Standalone app but lifecycleComponents, devAssetsController and healthCheck can be defined in AppComponents

alexduf: You're right, I think it looks cleaner and is probably easier to read like that!

I got `trait` crazy on this one.

admin/app/AppLoader.scala
```diff
@@ -43,14 +46,12 @@ trait AdminServices {
   lazy val rebuildIndexJob = wire[RebuildIndexJob]
 }
 
-trait Controllers extends AdminControllers {
-  def wsClient: WSClient
+trait AppComponents extends FrontendComponents with AdminControllers with AdminServices with I18nComponents {
+
+  lazy val messages: Messages = Messages(langs.preferred(langs.availables), messagesApi)
```
TBonnin: The Langs interface looks weird to me:
```
/**
 * Manages languages in Play
 */
trait Langs {

  /**
   * The available languages.
   *
   * These can be configured in `application.conf`, like so:
   *
   * {{{
   * play.i18n.langs = ["fr", "en", "de"]
   * }}}
   */
  def availables: Seq[Lang]

  /**
   * Select a preferred language, given the list of candidates.
   *
   * Will select the preferred language, based on what languages are available, or return the default language if
   * none of the candidates are available.
   */
  def preferred(candidates: Seq[Lang]): Lang
}
```

Am I missing something?

TBonnin:
@alexduf I have updated this PR and use `environment` rather than `application`. Let me know what you think

standalone/app/controllers/StandaloneControllerComponents.scala
```diff
@@ -25,6 +25,7 @@ trait StandaloneControllerComponents
 
   def wsClient: WSClient
   def frontJsonFapiDraft: FrontJsonFapiDraft
+  val application: Application
```
alexduf: What is this one used for?

admin/app/AppLoader.scala
alexduf: I'm not sure about how it's defined in Play 2.4, but in Play 2.5 I'm fairly sure you can direcly use the `MessageApi` like that

```scala
class MyController(messageApi) extends Controller {
  val myMessage: String = messageApi("blah")
}
```

So I don't think you need to instantiate `Message` at all

alexduf (approved):
All the rest looks good now!

admin/app/AppLoader.scala
```diff
@@ -43,14 +46,12 @@ trait AdminServices {
   lazy val rebuildIndexJob = wire[RebuildIndexJob]
 }
 
-trait Controllers extends AdminControllers {
-  def wsClient: WSClient
+trait AppComponents extends FrontendComponents with AdminControllers with AdminServices with I18nComponents {
+
+  lazy val messages: Messages = Messages(Lang.defaultLang, messagesApi)
```
TBonnin: @alexduf 👯

alexduf: Nice! 🌟

alexduf (approved):
👍

PR #14713, opened 2016-10-17
Tidy up the commercial and common/commercial projects a bit

JonNorman:
At some point in the future, I plan to sort out the packaging namespaces used in commercial - they are kinda inconsistent and seemingly a bit backwards. However, that PR is not this PR.

This is just a small set of otherwise-unrelated changes that - in my view - clean up our commercial code base and will make it easier to repackage commercial at a later date.

@guardian/commercial-dev


rich-nguyen:
I don't think we should have an object called `Commercial` at all, especially in `views/support`. It's so misleading when you read the `import` lines in the other files. What do you think? Is that something you were considering for the future? Otherwise it's fine, it's just that you've exposed another latent issue, by moving `TrackingCodeBuilder` to the right place.

👍

JonNorman:
I agree - there doesn't seem to be a need for it to be an object and it is doing the same job as a package. The main blocker to doing the repackaging work is how the common modules are laid out, so I plan to take a look at this properly later on.

However, I might as well make this better here rather than some future PR. I propose that I move each of the commercial objects into their own file into either
a) `views/support/commercial`, with the package `views.support.commercial`, or
b) `views/support` with the package `views.support`.

I'm more for option **a**; nothing else in there seems to be organised into project-oriented packages/directories, but doesn't mean we can't be. What do you think?

kelvin-chappell:
I would move all the plain Scala files out of the views package, so that it's dedicated to templates.  And keep `commercial` as the root of all Scala classes in the commercial module.  Otherwise you're bound to get collisions with common sooner or later.

Brave work! 👍

rich-nguyen:
Agree with @kelvin-chappell `views` are for views. And that approach would indeed keep the root namespace consistent.

JonNorman:
Okay - I've pretty much completely redone, rebased and destroyed my original branch. I think this is a lot clearer now. The first 5 commits are basically odd bits and bobs, the remaining 12 commits are just moving each set of commercial models across one by one.

Things not covered:
1) I haven't sorted out the commercial namespace within common. This is a fairly large task and would basically require refactoring all of common.

2) I haven't gotten rid of the `package.scala` objects yet. To me they seem to have a decent use case (useful things we want in all our `commercial.model` and `commercial.controllers`) but I'm happy to be overruled on that!

JonNorman:
[Update commercial packages to reflect directory structure.](https://trello.com/c/E37n5Dot/76-update-commercial-packages-to-reflect-directory-structure)

kelvin-chappell (approved):
🚢  :roll_eyes: 🚀

PR #14601, opened 2016-10-06
Use the main template as a base for signup pages

katebee:
## What does this change?

Affects https://www.theguardian.com/signup/weekendreading

Remove the **signupMain template** and switch to using the **main template** for signup pages, so that Google Analytics works.
## What is the value of this and can you measure success?
- Google Analytics will run on the Weekend Reading page
- Consistency of the page contentType with the routing (survey >> signup) 
- Updates the AB test with the new contentType
- Added bonus - delete a subsequently unnecessary template
## Does this affect other platforms - Amp, Apps, etc?

Nope
## Request for comment

@guardian/guardian-frontend

PR #14507, opened 2016-09-28
[WIP/MVP] Give the Guardian an email-newsletters collection page

katebee:
## What does this change?

Adds a brand new email-newsletters page to the Guardian under the route `/email-newsletters`

Like [Email Prefs](https://profile.theguardian.com/email-prefs), this new page uses `EmailSubscriptions.scala` as the canonical reference for all the possible data fields that an email listing may have.
## What is the value of this and can you measure success?
- Email Prefs is currently behind a login process. The email-newsletters page is a MVP for a collections page that will show all of our email offerings
- Website visitors can subscribe without logging into the Guardian
- All of the Guardian emails in one place
- Supports a no-JS experience
## Does this affect other platforms - Amp, Apps, etc?

Does not affect AMP
Not an AMP page
#### NO JAVASCRIPT, ANON USER:

picture 334
#### ENHANCED JAVASCRIPT RUNNING, ANON USER:

picture 335

On clicking `Sign up`, the input appears:
picture 336

Submitting an email address gives a confirmation: 
picture 337
#### SIGNED IN USER (ENHANCED JS):

JS updates the (hidden) email input field with the `primaryEmailAddress`
⭐️ Single click sign-up:
picture 341
## Request for comment

@joelochlann @superfrank

PR #14717, opened 2016-10-17
Pages with signup contentType avoid the sticky banner once again

katebee:
## What does this change?

Adds a new rule to stop ads appearing on the `signup` pages. These do not come from the content API, they are custom standalone pages.

Signup pages used to be built with a custom `main` template, which avoided loading the top banner. These page(s) are now being built using the `main` template to ensure Google Analytics runs (#14601). 

This has had the side effect of the sticky ad banner now appearing; this PR resolves this.

Signup is a new contentType that is (currently) only used by the weekend reading [email signup page](https://www.theguardian.com/signup/weekendreading). The newsletters page (#14507) will also have the same type and this fix will extend to it.

No
## Screenshots

⚡️💌
## Request for comment

@regiskuckaertz @joelochlann

### DIFF
--- a/common/app/staticpages/StaticPages.scala
+++ b/common/app/staticpages/StaticPages.scala
@@ -19,7 +19,7 @@ object StaticPages {
       section = Option(SectionSummary(id="email-signup-page", activeBrandings=None)),
       webTitle = webTitle,
       analyticsName = "email-signup-page",
-      contentType = "signup",
+      contentType = "Signup",
       iosType = None,
       shouldGoogleIndex = false))
 }
--- a/common/app/views/support/Commercial.scala
+++ b/common/app/views/support/Commercial.scala
@@ -14,6 +14,7 @@ object Commercial {
   def shouldShowAds(page: Page): Boolean = page match {
     case c: model.ContentPage if c.item.content.shouldHideAdverts => false
     case p: model.Page if p.metadata.sectionId == "identity" => false
+    case s: model.SimplePage if s.metadata.contentType == "Signup" => false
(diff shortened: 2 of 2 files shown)


regiskuckaertz:
👍

common/app/staticpages/StaticPages.scala
```diff
@@ -19,7 +19,7 @@ object StaticPages {
       section = Option(SectionSummary(id="email-signup-page", activeBrandings=None)),
       webTitle = webTitle,
       analyticsName = "email-signup-page",
-      contentType = "signup",
+      contentType = "Signup",
```
joelochlann: why is this now capitalised? I would've thought to avoid tripping people up we should keep it consistent and `"survey"` above isn't capitalised

katebee: capitalisation appears to be the more common approach in other pages: 
`contentType = "Interactive"` so I was mostly going with the convention of the masses 😅 

in meta.scala the approach is to use `contentType.toLowerCase` when doing string comparisons, probably to avoid the CaSe issUes ¯\\\_(ツ)_/¯

joelochlann: In general stuff like this might be better done with a Sum Type (sealed trait extended by case objects) rather than an arbitrary string. Then you get type safety since the compiler can tell you what the possible values of `contentType` are. Case issues such as these become irrelevant because if you use the wrong case, the code won't compile in the first place. (There may be a valid reasons why that wouldn't work in this case though)

PR #14718, opened 2016-10-18
Put attributions back onto the audio main image

MatthewJWalls:
## What does this change?

This puts the attribution caption back below the main pictures on audio pages. It was mistakenly requested to be omitted under a previous change.

This is valuable so we actually comply with the photo usage license :) 

Should be ok


katebee:
:+1:

PR #14719, opened 2016-10-18
Fix explainer in multiple paid-for MPUs

janua:
When there is more than one `paidFor` MPU on the page, the about button breaks.

This fixes `getTarget` used to get the toggle target so that is searches from the parent only instead of the whole document.

@regiskuckaertz

### DIFF
--- a/static/src/javascripts/projects/common/modules/ui/toggles.js
+++ b/static/src/javascripts/projects/common/modules/ui/toggles.js
@@ -26,7 +26,7 @@ define([
 
             controls.forEach(function (control) {
                 if (!bonzo(control).hasClass(readyClass)) {
-                    var target = self.getTarget(control);
+                    var target = self.getTarget(parent, control);
 
                     if (target && !(!isSignedIn && control.getAttribute('data-toggle-signed-in') === 'true')) {
                         control.toggleTarget = target;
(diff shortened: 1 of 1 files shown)


janua:
[Multiple paid content MPUs broken about behaviour](https://trello.com/c/G5PtdpqY/90-multiple-paid-content-mpus-broken-about-behaviour)

static/src/javascripts/projects/common/modules/ui/toggles.js
```diff
@@ -63,10 +63,10 @@ define([
         });
     };
 
-    Toggles.prototype.getTarget = function (control) {
+    Toggles.prototype.getTarget = function (parent, control) {
         var targetClass = bonzo(control).data('toggle');
         if (targetClass) {
-            return document.body.querySelector('.' + targetClass);
+            return parent.querySelector('.' + targetClass);
```
regiskuckaertz: LOL, indeed.

static/src/javascripts/projects/common/modules/ui/toggles.js
```diff
@@ -26,7 +26,7 @@ define([
 
             controls.forEach(function (control) {
                 if (!bonzo(control).hasClass(readyClass)) {
-                    var target = self.getTarget(control);
+                    var target = self.getTarget(parent, control);
```
regiskuckaertz: Is it `parent` or `component`?

static/src/javascripts/projects/common/modules/ui/toggles.js
regiskuckaertz: (line 22)

regiskuckaertz (changes requested):

janua:
@regiskuckaertz Now using `component` in `getTarget` call.

regiskuckaertz (approved):

regiskuckaertz:
Yiihaaa

PR #14720, opened 2016-10-18
Remove implicit/global imports for all templates; explicitly state each

JonNorman:
In `Prototypes.scala`, we specify the following packages as _always_ being imported:
-      "common._",  
-      "model._",  
-      "views._",  
-      "views.support._"

This is _really_ annoying, because it means that any subpackage in any of these packages can cause a namespace conflict with something else and starts to require specifying modules with `_root_.` etc.

Making this change means that each project is able to manage it's own namespace and each template can specifically import whichever models/controllers/views/helpers it needs, without worrying about clashing namespaces within common / model / views / views.support.

This won't really add much of a burden on anyone developing new templates or editing current ones - it is just quite a large initial effort to do (and maybe review (sorry)).

**I'd suggest reviewing this ignoring whitespace...**

**Added bonus**
Play allows the specification of imports before specifying the parameters to a template, so our parameter lists can now be things like `page: Page` rather than `page: model.Page`, which is a bit cleaner.

### DIFF
--- a/common/app/views/fragments/atoms/interactive.scala.html
+++ b/common/app/views/fragments/atoms/interactive.scala.html
@@ -1,6 +1,8 @@
-@(interactive: _root_.model.content.InteractiveAtom, shouldFence: Boolean)
+@import common.InlineJs
+@import _root_.model.content.InteractiveAtom
+@import templates.inlineJS.nonBlocking.js.{interactiveFonts, interactiveResize}
 
-@import templates.inlineJS.nonBlocking.js._
+@(interactive: InteractiveAtom, shouldFence: Boolean)
 
 @iframeBody = {
     <!DOCTYPE html>
--- a/common/app/views/fragments/atoms/interactive.scala.html
+++ b/common/app/views/fragments/atoms/interactive.scala.html
@@ -1,5 +1,5 @@
 @import common.InlineJs
-@import _root_.model.content.InteractiveAtom
+@import model.content.InteractiveAtom
 @import templates.inlineJS.nonBlocking.js.{interactiveFonts, interactiveResize}
 
 @(interactive: InteractiveAtom, shouldFence: Boolean)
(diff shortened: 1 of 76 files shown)


common/app/views/fragments/atoms/interactive.scala.html
```diff
@@ -1,6 +1,8 @@
-@(interactive: _root_.model.content.InteractiveAtom, shouldFence: Boolean)
+@import common.InlineJs
+@import _root_.model.content.InteractiveAtom
```
rich-nguyen: is `_root_` needed here?

common/app/views/fragments/immersiveMainMedia.scala.html
```diff
@@ -1,12 +1,12 @@
-@(page: ContentPage)(implicit request: RequestHeader)
-
 @import common.{LinkTo, Localisation}
+@import conf.switches.Switches._
```
rich-nguyen: `. ArticleBadgesSwitch`

discussion/app/views/fragments/commentShare.scala.html
```diff
@@ -1,7 +1,10 @@
-@(comment: Comment)
+@import _root_.model.{Facebook, ShareLinks, SharePlatform, Twitter}
```
rich-nguyen: again, not sure if  `_root_` is really needed? I guess it is for some reason

rich-nguyen:
👍

discussion/app/views/fragments/commentShare.scala.html
JonNorman: In this case it is - the discussion project [has a number of default imports](https://github.com/guardian/frontend/blob/master/project/Frontend.scala#L104) that would otherwise cause `model._` to clash. I haven't changed this as each project can do what they want

PR #14722, opened 2016-10-18
Remove png sprites

sndrs:
## What does this change?

stop creating png sprite equivalents of the data-uri'd svgs.

we only serve them to `IE <= 8`, which according to GA is only now IE7, on 0.07%, and which we don't support in the rest of the CSS anyway.
- it removes complexity from the build
- removes a dependency on phantom 
- saves ~7 secs off build time.

no
## Screenshots

I couldn't spot any obvious issues, and it only removes the sprite CSS, but I could have overlooked something... (by no means the first time)

screen shot 2016-10-18 at 10 37 41
## Request for comment

@guardian/dotcom-platform @guardian/commercial-dev


regiskuckaertz:
Yihaa! Commercials don't run in IE < 9 anyway so we're happy to see this go 🎆

gtrufitt (approved):
image

🔥 👍 💯

sndrs:
fingers crossed...

PR #14723, opened 2016-10-18
Upgrade dfp axis library

rich-nguyen:
Google are deprecating the older namespaces soon, so this change updates the library and the imports to the latest version. I didn't see any problems, running locally to build the commercial json content.


rich-nguyen:
[DFP api deprecation notice](https://trello.com/c/ej4ctKEI/93-dfp-api-deprecation-notice)

regiskuckaertz:
woohoo 👍

kelvin-chappell:
If only it were possible to hold a central alias for all the versioned package names so it only needed to be done in one place.  But I don't think it is.

👍

PR #14727, opened 2016-10-18
Dev-build also needs to have access to play.api.i18n.Messages value

TBonnin:
## What does this change?

Dev-build also needs to have access to a play.api.i18n.Messages value.
Moving its definition to AdminServices which is extended by dev-build
AppComponents

dev-build compiles :/

## Request for comment

@rich-nguyen


jfsoul:
👍

rich-nguyen:
👍  thanks @TBonnin

PR #14728, opened 2016-10-18
Add clear to dateline for floated badges

rich-nguyen:
The new badges system floats the badge left, which needs clearing through dateline. The vertical alignment is a separate issue.

**Before**

before

**After**

after


rich-nguyen:
[Bug with click targets on badges](https://trello.com/c/8YcYpp05/94-bug-with-click-targets-on-badges)

regiskuckaertz:
👍

PR #14729, opened 2016-10-18
Call impression as well as complete events when forcing into an A/B test

desbo:
## What does this change?

When you force yourself into a test with the anchor link, it'll call the new `impression` function for your variant as well as `success`.

Easier (well, actually possible) to test the `impression` function during development!

@jranks123


jranks123:
👍

PR #14732, opened 2016-10-19
use config value rather than hard-coded string

akash1810:
## What does this change?

nothing really, just makes sense

removes a magic string

No.

## Screenshots

n/a
## Request for comment

@guardian/guardian-frontend


kelvin-chappell (approved):
👍

PR #14730, opened 2016-10-18
Nb add variant again

NataliaLKB:
Description here: https://github.com/guardian/frontend/pull/14676

Opening up this PR again, with something I forgot from https://github.com/guardian/frontend/pull/14685.

The original PR was reverted due to a build error, but that was unrelated to this change.
## Request for comment

@SiAdcock @TBonnin

PR #14734, opened 2016-10-19
Nb new header variant two tweaks

NataliaLKB:
## What does this change?

This PR does two things. First, clean up some things from the original PR #14730, and lays the groundwork for adding an opt in link to this variant.
1. Things behave better!
2. Alex and others in the building can opt into this test easily

Nope
## Request for comment

### DIFF
--- a/applications/app/controllers/OptInController.scala
+++ b/applications/app/controllers/OptInController.scala
@@ -32,6 +32,7 @@ class OptInController extends Controller {
       case "https" => https.opt(choice)
       case "hsts" => hsts.opt(choice)
       case "header" => header.opt(choice)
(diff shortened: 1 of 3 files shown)


applications/app/controllers/OptInController.scala
```diff
@@ -32,6 +32,7 @@ class OptInController extends Controller {
       case "https" => https.opt(choice)
       case "hsts" => hsts.opt(choice)
       case "header" => header.opt(choice)
+      case "headertwo" => header.opt(choice)
```
dominickendrick: should this be `headerTwo.opt(choice)` ?

dominickendrick:
👍 apart from comment !

PR #14735, opened 2016-10-19
Removing deprecated opt in/out routes

NataliaLKB:
## What does this change?

While making some changes I noticed that there were some opt in/out routes that I don't think are needed anymore. This just cleans them up.

Cleaning is always good ✨ 

Nope!
## Request for comment

@desbo @guardian/dotcom-platform

### DIFF
--- a/applications/app/controllers/OptInController.scala
+++ b/applications/app/controllers/OptInController.scala
@@ -29,14 +29,12 @@ class OptInController extends Controller {
 
   def handle(feature: String, choice: String) = Action { implicit request =>
     Cached(60)(WithoutRevalidationResult(feature match {
-      case "hsts" => hsts.opt(choice)
       case "header" => header.opt(choice)
       case "headertwo" => headerTwo.opt(choice)
       case _ => NotFound
     }))
   }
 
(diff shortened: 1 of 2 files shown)


applications/app/controllers/OptInController.scala
```diff
@@ -29,16 +29,12 @@ class OptInController extends Controller {
 
   def handle(feature: String, choice: String) = Action { implicit request =>
     Cached(60)(WithoutRevalidationResult(feature match {
-      case "https" => https.opt(choice)
       case "hsts" => hsts.opt(choice)
```
NataliaLKB: @desbo Is this opt in/out still used?

desbo: I don't think so! @JustinPinner, do you know if anyone still uses the HTTPS opt-in/out routes?

philwills: I don't believe that we need any of this any longer as everyone should always be on HTTPS.

HSTS is now also on generally, so I believe can also go, though @mchv might have an opinion.

mchv:
👍 thanks @NataliaLKB!

PR #14736, opened 2016-10-19
Refactor port logic in healthcheck and tests Suite

TBonnin:
## What does this change?

My original goal was to get rid of the `Play.current` reference into the `HealthCheck` code.
However I thought it would be a good time to refactor this weird logic where a port is specify to the Healthcheck, which is then used to override the OneServerPerSuite port so multiple suite don't overlap on the same port and the healthcheck is aware of this port on which it needs to run its internal check.

So now:
- test port logic has been removed from healthcheck 
- test port is explicitly set for each test `Suite`
- when testing healthcheck ([1 single place](https://github.com/guardian/frontend/pull/14736/files#diff-c68b31e0978004f8eac907de04add445R19)), the healthcheck port is overridden with the currently running test suite port
- => Play 2.5
- Simpler code

## Request for comment

@alexduf @jfsoul

### DIFF
--- a/article/test/CdnHealthCheckTest.scala
+++ b/article/test/CdnHealthCheckTest.scala
@@ -3,6 +3,7 @@ package test
 import controllers.HealthCheck
 import org.scalatest.{BeforeAndAfterAll, DoNotDiscover, FlatSpec, Matchers}
 import play.api.test.Helpers._
+import play.api.test.Helpers
 import org.scalatest.concurrent.ScalaFutures
 
 @DoNotDiscover class CdnHealthCheckTest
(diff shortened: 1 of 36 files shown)


common/app/conf/HealthCheck.scala
```diff
@@ -95,14 +95,7 @@ private[conf] trait HealthCheckFetcher extends ExecutionContexts with Logging {
       }
   }
 
-  protected def fetchResults(testPort: Int, healthChecks: SingleHealthCheck*): Future[Seq[HealthCheckResult]] = {
-    val defaultPort = 9000
-    val port = {
-      Play.current.mode match {
```
TBonnin: 👋

article/test/CdnHealthCheckTest.scala
```diff
@@ -14,7 +15,9 @@ import org.scalatest.concurrent.ScalaFutures
   with WithTestWsClient {
 
   "CDN health check" should "mimic the instance health check" in {
-    val controller = new HealthCheck(wsClient)
+    val testPort: Int = port
```
jfsoul: This is just for clarity or is it needed for the override to compile?

TBonnin: Necessary, otherwise the `port` val in the constructor closure shadows the test `port`

alexduf (approved):
This is great! 👍

PR #14738, opened 2016-10-19
Remove Play.isTest (deprecated in Play 2.5)

TBonnin:
## What does this change?

Remove Play.isTest (deprecated in Play 2.5). Unnecessary old code

=> Play 2.5

## Request for comment

@johnduffell


johnduffell:
sure why not! just make sure people won't get really confused if the tests are not working right 😖 👍 :play_or_pause_button:

PR #14740, opened 2016-10-19
More fixes and corrections to the email prefs page

katebee:
## What does this change?

Changes the data behind https://profile.thegulocal.com/email-prefs

Small fixes and corrections (mostly to the signupPages and exampleUrls

Bug zapping ⚡️

Nope
## Request for comment

@guardian/guardian-frontend


identity/app/model/EmailSubscriptions.scala
```diff
@@ -307,7 +308,8 @@ object EmailSubscriptions {
       description = "For all you need to know about technology in the world this week, news, analysis and comment.",
       frequency = "Every Thursday",
       listId = "1902",
-      subscribedTo = subscribedListIds.exists{ x => x == "1902" }
+      subscribedTo = subscribedListIds.exists{ x => x == "1902" },
```
gklopper: BTW `subscribedListIds.contains("1902")` is equivalent and a bit easier on the eye.

katebee: Good to know, thanks 😄

katebee: Trying the swap results in a compilation error: 
`value contains is not a member of Iterable[String]` 🤔

identity/app/model/EmailSubscriptions.scala
```diff
@@ -97,6 +97,7 @@ object EmailSubscriptions {
       frequency = "Tuesday mornings",
       listId = "3698",
       subscribedTo = subscribedListIds.exists{ x => x == "3698" },
+      signupPage = Some("https://www.theguardian.com/politics/2016/may/31/eu-referendum-morning-briefing-sign-up"),
```
joelochlann: This means dev and code environments will link to live. It would be better to have a relative path?

katebee: Not sure if we can have a relative path? Email prefs is on https://profile.theguardian.com/email-prefs so LinkTo would try and take the user to `https://profile.thegulocal.com/<custom-path>/latest/email`

katebee: I've had a look around in frontend.. and while it seems possible to change the subdomain/route linked to, dependent on the environment (CODE/PROD), it is not currently being done for identity, and may be adding more complexity than value.

I could try something hacky with /common/app/common/configuration.scala
`@if(Configuration.environment.isNonProd){ ...`

This would solve for code, but still doesn't help with linking to the correct path of localhost

joelochlann (approved):

identity/app/model/EmailSubscriptions.scala
NataliaLKB: LinkTo is only relative in lcoalhost, not in CODE and not in PROD. Or use our local env variables!  Best not to hardcode this.

PR #14742, opened 2016-10-20
Remove omniture clickstream tracking

dominickendrick:
## What does this change?

This removes general Omniture click tracking from dotcom

Help us migrate away from Omniture

Omniture

## QA

Check that internal and external and in page link interactions do not fire. Note that Pageview events still fire
## Request for comment

@guardian/dotcom-platform @mkopka

### DIFF
--- a/static/test/javascripts/spec/common/analytics/interaction-tracking.spec.js
+++ b/static/test/javascripts/spec/common/analytics/interaction-tracking.spec.js
@@ -14,21 +14,14 @@ define([
                     trackSamePageLinkClick: sinon.spy(),
                     trackExternalLinkClick: sinon.spy()
                 })
-                .mock('common/modules/analytics/omniture', {
-                    go: sinon.spy(),
-                    trackSamePageLinkClick: sinon.spy(),
-                    trackExternalLinkClick: sinon.spy()
-                })
                 .require([
                     'common/utils/mediator',
                     'common/modules/analytics/google',
-                    'common/modules/analytics/omniture',
                     'common/modules/analytics/interaction-tracking'
                 ], function () {
                     mediator = arguments[0];
                     google = arguments[1];
-                    omniture = arguments[2];
-                    interactionTracking = arguments[3];
+                    interactionTracking = arguments[2];
 
                     done();
                 });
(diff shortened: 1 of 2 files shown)


markjamesbutler:
LGTM 👍

static/test/javascripts/spec/common/analytics/interaction-tracking.spec.js
```diff
@@ -14,21 +14,14 @@ define([
                     trackSamePageLinkClick: sinon.spy(),
                     trackExternalLinkClick: sinon.spy()
                 })
-                .mock('common/modules/analytics/omniture', {
-                    go: sinon.spy(),
-                    trackSamePageLinkClick: sinon.spy(),
-                    trackExternalLinkClick: sinon.spy()
```
NataliaLKB: Does GA/Ophan still track external link clicks without omniture?

dominickendrick: This is being deprecated in GA, and it doesn't happen in omniture

NataliaLKB: Sorry Dom, I don't follow. With this PR, will the page still wait to register clicks to external websites so we don't lose that data?

I remember chatting about this with @gtrufitt a couple weeks ago. Then it was something that would have to be implemented before Omniture is taken out. It is pretty important data to capture and I just want to make sure we aren't going to lose it.

dominickendrick: @mkopka says that we rarely report on external page tracking. The comment left by Chris Birchell about re-impimenting the [500ms delay](https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/common/modules/analytics/interaction-tracking.js#L72) is untrue at the moment, as we have [explicitly turned this off in Omniture](https://github.com/guardian/frontend/blob/master/static/public/javascripts/vendor/omniture.js#L16). We will continue to report external link tracking in the same way as we do now in GA, so no data will be lost. I'm looking to implement this using the [Beacon API](https://developer.mozilla.org/en-US/docs/Web/API/Beacon_API) as a seperate PR to improve accuracy, and retain performance.

PR #14743, opened 2016-10-20
Changing redirect table location for Admin tool

dominickendrick:
## What does this change?

Currently we read redirects in `CODE` and `DEV` from `redirects-CODE` DynamoDB table in [DynamoDb.scala](https://github.com/guardian/frontend/blob/master/common/app/services/DynamoDB.scala#L22), but the admin tool is writing to `redirects-DEV` which stops us being able to test redirects in `CODE`
This PR fixes that!

@guardian/dotcom-platform


jfsoul (approved):
Good spot 👍 👀

PR #14744, opened 2016-10-20
Include surging values in Prebid

rich-nguyen:
The Rubicon SSP we use is passed information about surging content, as measured by Ophan. In the US, we use Prebid to perform an "auction" with Rubicon. I'm adding surging parameters to the Prebid system. The value is to help out the US commercial team.


rich-nguyen:
[Rubicon Pulse set up](https://trello.com/c/p5dsH96N/29-rubicon-pulse-set-up)

regiskuckaertz (approved):
Nice and tidy!

PR #14745, opened 2016-10-20
Remove wrapping div on immersives

wpf500:
It's really hard to do fullscreen (i.e. height: 100%) immersive interactives. We have to apply `height: 100%` to `html`, `body` and `figure.interactive` elements, as well as any of our own elements.

Recently a `div` has crept in which makes it impossible (has no classes to select on in most cases), and as far as I can tell `paid-content--advertisement-feature` has no effect on immersive interactives.

@regiskuckaertz do you know if that class is ever used?

We really need our immersive interactives to be a direct descendent of the `body` tag. This really needs to be a requirement if we are going to not break old interactives when we change these templates.

### DIFF
--- a/applications/app/views/fragments/interactiveBody.scala.html
+++ b/applications/app/views/fragments/interactiveBody.scala.html
@@ -18,9 +18,8 @@
 
 @body(interactive: model.Interactive, isImmersive: Boolean, isPaidContent: Boolean) = {
     @if(isImmersive) {
-        <div class="@RenderClasses(Map("paid-content--advertisement-feature" -> isPaidContent))">
-            @bodyRaw(interactive)
-        </div>
+        @* Must not be wrapped in any elements to support full screen immersives *@
+        @bodyRaw(interactive)
     } else {
         <div class="l-side-margins">
             <article id="article" class="@RenderClasses(


regiskuckaertz:
The `paid-content--advertisement-feature` class is used for styling paid content. Do @Calanthe and @lps88 have more information? Could this class be added to the interactive itself instead?

wpf500:
I understand for non-immersive content it adds a lot of styling, e.g. https://www.theguardian.com/airbnb-adventures/2016/oct/07/tips-for-hiking-with-children-how-to-turn-a-wander-into-a-wonder.

But as far as I can tell in immersive interactives, where the page is essentially empty except a thin header, it has no use, e.g. https://www.theguardian.com/inside-seat/ng-interactive/2016/sep/09/how-to-achieve-peak-performance-audio-interactive

wpf500:
@regiskuckaertz, @Calanthe, @lps88 

I've manually gone through and checked if this class is affecting any paid immersive interactive content. There are only 6 immersive interactives that I can find (the ones with displayHint immersive):
http://content.guardianapis.com/search?tag=tone%2Fadvertisement-features%2Ctype%2Finteractive&show-fields=displayHint&page-size=40&api-key=test

None of them use the `paid-content--advertisement-feature`. Unless I'm missing some content, I'm confident that this change won't break anything. What do you think?

regiskuckaertz:
:+1: I'm completely with you here. More generally, we should keep away from adding DOM elements for the sake of styling as much as possible, that `DIV` should never have been there in the first place. If it turns out that some styling is gone after you merge that, then it's back to the drawing board and figuring out a less intrusive solution.

PR #14746, opened 2016-10-20
Prominent membership engagement message test phase 1

JustinPinner:
## What does this change?

A/B test to make membership engagement messaging via the footer banner more prominent.

Hypothesis: Bolder messages will drive an uptake of membership.

No, web only.
## Screenshots

_Before:_
screen shot 2016-10-20 at 17 28 40

screen shot 2016-10-20 at 17 29 17

screen shot 2016-10-20 at 17 30 25

_After:_
screen shot 2016-10-20 at 17 22 32

screen shot 2016-10-20 at 17 23 52

screen shot 2016-10-20 at 17 25 39
## Request for comment

@rtyley @NataliaLKB @regiskuckaertz (I'll fix the merge conflicts now)


NataliaLKB:
My biggest concern is that something this intrusive may actually have lasting impact on trust with users.

After some discussing with @JustinPinner @stephanfowler @rtyley and Nicola we decided it is fine to test. However, there should possibly be some more discussion with UX before making it a feature so we can fully understand the side effects of a test like this. Especially the unintended ones.

paperboyo:
I fully support what @NataliaLKB said above. Plus, we still only serve ASCII charset of Egyptian Bold, so the pound sign is substituted from another font and looks really bad. Current coverage of Egyptian Medium could hold that message contents okay. Also, I'm not so sure there is a direct correlation between the font weight and the amount of money we''ll get - otherwise marketing typography would be a piece of (really fat) cake.

twobobswerver:
we can track feedback through usual membership feedback channels / userhelp

ScottPainterGNM:
Is that not quite alot of friction though? 

I think those are valid ways of getting feedback but that doesn't measure how many people just leave the page for example because they saw a large banner pop up for example?

rtyley:
> Plus, we still only serve ASCII charset of Egyptian Bold, so the pound sign is substituted from another font and looks really bad.

Checked with @JuliaBellis, agreed we should step back from the the Bold.

static/src/stylesheets/module/_release-message.scss
```diff
@@ -495,6 +495,166 @@ $btn-height: gs-height(1) - $gs-baseline/2; //30px
...
+    }
+    .site-message__inner {
+        padding-top: 10px;
+        padding-left: 5px;
+        padding-right: 5px;
+        padding-bottom: 5px;
+    }
+    .site-message__message--membership {
+        margin-right: 5px;
+        font-weight: bold;
```
rtyley: Ben Wuersching is very keen to keep the boldish-ness, and points out that we use "bold" `£` on Membership, where they work:

image

...but that's not actually `bold`, it's `font-weight: 900`. If we switch to that the `£` will presumably work. Ben reassures me that Alex Breuer has seen this use of emboldened text, and is fine with it.

paperboyo:
> "bold" £ on Membership, where they work

Isn't Membership loading its own fonts, though? `font-weight: 900` is the maximum and will still result in an ugly substitution on dotcom AFAIK.

Another point would be about the féerie of colours in these messages: careful and consistent use of colour should slowly let our readers be able to draw a correlation between specific colour and the type of content/message. Do we have a 'membership colour'? Why don't we use exclusively that here?

Arguments that using the whole rainbow will result in more clicks or that readers will be less annoyed having to dismiss the same banner they already dismissed again if it's in another colour - are rubbish: we might then as well pay for some topless models to promote our membership proposition. Even then: bikinis/just panties should be strictly 'membership colour' only!

JustinPinner:
I've updated the bold to weight 900, and applied a couple of tweaks to the mobile view as requested by Ben W. I'll push that in a couple of minutes.

JustinPinner:
@NataliaLKB @rtyley can one of you +1 this now?

rtyley:
> I've updated the bold to weight 900

Has that led to the `£` being correctly displayed on dotcom? According to @paperboyo _"...font-weight: 900 is the maximum and will still result in an ugly substitution on dotcom AFAIK."_

If the `£` is working then I'm OK with this going out as a test (if not, we should remove the bold/900), I believe @NataliaLKB has [already approved](https://github.com/guardian/frontend/pull/14746#issuecomment-255168927) this PR for the purpose of going out as a test, with the proviso that it undergoes further UX review before going out as a feature.

JustinPinner:
I'll take that as a :+1: then.

PR #14748, opened 2016-10-21
Increase troubleshooter fetch timeout to 5s

TBonnin:
## What does this change?

Increase troubleshooter fetch timeout to 5s

Direct fetch to the router sometimes times out with the current value. I hope increasing timeout to 5s will fix the issue 

## Request for comment

@jfsoul


jfsoul:
👍

PR #14750, opened 2016-10-21
Rewrite breakpoints detection

regiskuckaertz:
We use media query listeners to inform `getBreakpoint`/`isBreakpoint`/`hasCrossedBreakpoint`.

A few advantages over the current impl:
- relying on the browser standard APIs makes the solution more robust
- redundant computation is therefore avoided
- calls to this API now become entirely free*

Still need to update unit tests though 😅


dominickendrick:
How is the testing going ? Can we close the PR until it's nearer completion ?

regiskuckaertz:
@dominickendrick Yes, I'll raise it again when everything is ready.

PR #14752, opened 2016-10-21
Only add a badge to the first facia container that ISN'T a thrasher.

JonNorman:
#### Background / Problem

If a front is sponsored/branded, then we add a badge to the first container on that page. However, in the cases where there is a thrasher, we want to add the badge to the one below it i.e. we want to add the badge to the first _non-thrasher_ container on a sponsored front.
#### Solution

We can have multiple thrashers on a given front and so only the front fragment being rendered can determine which container should hold the badge, and it must do this by inspecting the container objects it is about to render. This requires each `FaciaContainer` knowing whether or not it is a thrasher.
#### Disclaimer

I'm not proud of this - it seems a bit gross inspecting the layout name to determine whether a `FaciaContainer` is a thrasher or not but, short of creating a new field in FAPI that explicitly specifies which container should have a branded badge on, I think this is the best/easiest solution.

@kelvin-chappell

### DIFF
--- a/applications/app/views/fragments/indexBody.scala.html
+++ b/applications/app/views/fragments/indexBody.scala.html
@@ -28,8 +28,14 @@
                 }
 
                 @defining(IndexPage.makeFront(index, Edition(request)).containers) { containers =>
-                    @containers.map { containerDefinition =>
-                        @container(containerDefinition, frontProperties = FrontProperties.fromBranding(branding))
+                    @defining(( containers filterNot (_.isThrasher) map (_.index) ).min ) { containerIndexForBadge =>
+                        @containers.map { containerDefinition =>
+
+                            @container( containerDefinition,
+                                        frontProperties = FrontProperties.fromBranding(branding),
+                                        showFrontBranding =  containerDefinition.index == containerIndexForBadge)
+                        }
+
                     }
                 }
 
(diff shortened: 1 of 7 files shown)


JonNorman:
["Sponsored by" badge appearing in thrasher on podcasts page.](https://trello.com/c/cf3rLSY4/102-sponsored-by-badge-appearing-in-thrasher-on-podcasts-page)

applications/app/views/fragments/indexBody.scala.html
```diff
@@ -28,8 +28,14 @@
...
 
                 @defining(IndexPage.makeFront(index, Edition(request)).containers) { containers =>
-                    @containers.map { containerDefinition =>
-                        @container(containerDefinition, frontProperties = FrontProperties.fromBranding(branding))
+                    @defining(( containers filterNot (_.isThrasher) map (_.index) ).min ) { containerIndexForBadge =>
+                        @containers.map { containerDefinition =>
+
+                            @container( containerDefinition,
+                                        frontProperties = FrontProperties.fromBranding(branding),
+                                        showFrontBranding =  containerDefinition.index == containerIndexForBadge)
```
kelvin-chappell: Could the evaluation of `containerIndexForBadge` be done in a support def elsewhere?
`Defining` blocks in templates are best avoided unless the value is used multiple times.
So something like `def shouldShowFrontBranding(index, containers): Boolean`.

kelvin-chappell (changes requested):
Looks like the best solution.  Just a small suggestion to clarify the template.

kelvin-chappell (approved):

JonNorman:
Yep, I reckon they probably can - doesn't seem to be anything obviously different about them at first glance...

PR #14754, opened 2016-10-24
Start moving ex-grunt tasks into __tasks__ dir

sndrs:
## What does this change?

creates a notional `__tasks__` directory, similar to the jest `__tests__` directory, and updates the runner to use it.

the change is mainly updating and moving running logic from `compile` to `runner` and renaming that to `run-task`

it sows the seeds of a tiny task runner to replace grunt and to use in place of `npm run` and possibly `make`, down the line. 

the reason for that being the more I try and corral `make` into running non-make-like tasks (all our tasks are `.PHONY`, really), the more i wonder if it was a good choice. maybe it was, but this will make it much easier to remove grunt (more PRs to come waiting on this) and if this proves too fettering, it will be easy to remove, since the logic and tasks are much clearer to understand (i hope!) and the deps on grunt plugins will be gone

no
## Request for comment

@guardian/dotcom-platform


rich-nguyen:
is that a new directory naming style? :)

sndrs:
`__dirname2k18__` 

not seen it elsewhere, but after using jest for a bit, it did actually make scanning trees much easier ('this directory is infrastructural', these are src)

tools/__tasks__/compile/javascript/index.js
```diff
@@ -1,12 +1,12 @@
-module.exports = [{
-    title: 'Compile JS',
+module.exports = {
+    description: 'Compile JS',
     task: [
         require('./clean'),
-        require('../inline-svgs/compile'),
+        'compile/inline-svgs',
```
SiAdcock: I am unclear on when you would use `'compile/inline-svgs'` instead of `require('./compile/inline-svgs')`. Are the two essentially interchangeable?

sndrs: yeah they are, but maybe that's not clear enough. i wonder if it offers too many ways of configuring?

SiAdcock: Yeah, I think it adds unnecessary API clutter. I would probably prefer to use `require` to import task objects, rather than strings to either execute tasks or terminal commands. This is just an opinion, though, not a showstopper 😄

sndrs: yeah it's a fair point. it's true that my focus so far has been getting the tasks out of grunt in a (hopefully) reasonably understandable way. i think you may be correct there though...

SiAdcock:
I'm not 100% sold on the `__tasks__` naming convention. From what I understand of Jest, the `__tests__` directories are interspersed throughout your source, so having the super-underscored directory names is helpful for distinguishing source code from non-source. Whereas `__tasks__`, well, it lives in `tools`, so is already separate from source code.

Again, non-showstopper, just another opinion. In general, I'm in favour of moving all tasks into their own directory, and further separating`compile`-related tasks into their own sub-directory 👍

PR #14759, opened 2016-10-24
Move linting and testing out of gruntfile, simply task runner api

sndrs:
## What does this change?
- moves linting and test-running out of grunt file
- simplifies test runner along lines of https://github.com/guardian/frontend/pull/14754#discussion_r84682788

nearly emptied the gruntfile

no
## Request for comment

@SiAdcock


tools/run-task
```diff
@@ -132,25 +113,22 @@ function listrify(steps, {concurrent = false, verbose = false} = {}) {
             if (Array.isArray(task)) {
                 return { title, task: () => listrify(task.map(_task => {
```
SiAdcock: I'm not sure if it makes a huge difference, or any difference, but there are a few references to `title` in this function when elsewhere the property name has changed to `description`. Might be slightly confusing

sndrs: `title` is the `listr` api, but it seemed weird to use that to describe a task in the task definition? 

`listrify` is really just the bridge between our tasks and what it expects. 

do you think it would be better to map to it more closely?

SiAdcock: Ah right, I see the purpose of `listrify` now, it looks like it's doing the right thing 😄

SiAdcock:
Looks great 👍

tools/run-task
sndrs: will add a comment

PR #14760, opened 2016-10-24
Add indication of native/non-native status to creative templates on commercial admin page.

JonNorman:
### What

Now when you go [here](https://frontend.gutools.co.uk/commercial/templates) you can easily see which templates are native and which aren't by the presence of an asterisk next to the template name.
### Why

Not groundbreaking but makes it easier to scan the list for natives, which is useful since we are currently working on a load of them.
### P.S

Also refactored a bunch of the JSON models to use macros - not all of them can be, usually because:
a) the field names of the case class do not match those in the JSON data or
b) some special transformation needs to be done to a specific field, which is a pain to do otherwise.

@guardian/commercial-dev


kelvin-chappell (approved):
👍

PR #14761, opened 2016-10-24
Fix double nav on explore template

holder-josh:
## What does this change?

This adds an extra check to make sure that we still hide the top ad banner on the explore template, even when they supply an embed code as the main media. In this circumstance isMainEmbed returns false as CAPI doesn't return an element with the relation field set to main, unlike when you supply a bootjs as the main media, so the most reliable check is whether the main field is empty.

Before:
screen shot 2016-10-24 at 16 26 42

After:
screen shot 2016-10-24 at 16 27 17

No double logo.

No

## Request for comment

@blongden73 @NataliaLKB


NataliaLKB:
👍

PR #14762, opened 2016-10-24
Interactive serviceworker route

dominickendrick:
## What does this change?

This adds a route on theguardian.com to allow service worker `javascript` files to be loaded for interactives.

This allows urls like

`https://www.theguardian.com/society/2016/oct/24/many-treatments-give-no-more-benefit-than-doing-nothing-leading-doctors`

to load a service worker from

`https://www.theguardian.com/society/2016/oct/24/many-treatments-give-no-more-benefit-than-doing-nothing-leading-doctors/interactive-service-worker.js?timestamp=1234`

This url then proxies a request to:

`https://interactive.guim.co.uk/service-workers/live|preview/society/2016/oct/24/many-treatments-give-no-more-benefit-than-doing-nothing-leading-doctors/1234/interactive-service-worker.js` depending on the environment your are viewing in.

Adds the power of service workers to interactives!

## Request for comment

@guardian/dotcom-platform @Rich-Harris @wpf500


Rich-Harris:
Excellent! After discussing with @wpf500, just one small change I would suggest – I think the filenames should always match (it's confusing if `interactive-service-worker.js` becomes `live.js`), and the live/preview distinction should happen at the top level:

```bash
# this...
https://www.theguardian.com/path/to/interactive-service-worker.js

# becomes this:
https://interactive.guim.co.uk/service-workers/live/path/to/interactive-service-worker.js

# and this...
https://preview.gutools.co.uk/path/to/interactive-service-worker.js

# becomes this:
https://interactive.guim.co.uk/service-workers/preview/path/to/interactive-service-worker.js
```

What do you reckon?

dominickendrick:
Sounds good!

applications/app/controllers/InteractiveController.scala
```diff
@@ -5,21 +5,35 @@ import common._
...
-class InteractiveController(contentApiClient: ContentApiClient) extends Controller with RendersItemResponse with Logging with ExecutionContexts {
+class InteractiveController(contentApiClient: ContentApiClient, wsClient: WSClient) extends Controller with RendersItemResponse with Logging with ExecutionContexts {
 
   def renderInteractiveJson(path: String): Action[AnyContent] = renderInteractive(path)
   def renderInteractive(path: String): Action[AnyContent] = Action.async { implicit request => renderItem(path) }
 
+  def renderInteractiveServiceworker(path: String): Action[AnyContent] = Action.async { implicit request =>
+    val stage = if (isPreview) "preview" else "live"
+    val serviceWorkerPath = s"$cdnPath/service-workers/$stage/$path/interactive-service-worker.js"
+    wsClient.url(serviceWorkerPath).get().map { response =>
```
johnduffell: what happens if it goes wrong? a 500 error? might want to log at least and handle bad status codes

applications/app/controllers/InteractiveController.scala
```diff
@@ -5,21 +5,35 @@ import common._
...
   override lazy val item = interactive
 }
 
-class InteractiveController(contentApiClient: ContentApiClient) extends Controller with RendersItemResponse with Logging with ExecutionContexts {
+class InteractiveController(contentApiClient: ContentApiClient, wsClient: WSClient) extends Controller with RendersItemResponse with Logging with ExecutionContexts {
 
   def renderInteractiveJson(path: String): Action[AnyContent] = renderInteractive(path)
   def renderInteractive(path: String): Action[AnyContent] = Action.async { implicit request => renderItem(path) }
 
+  def renderInteractiveServiceworker(path: String): Action[AnyContent] = Action.async { implicit request =>
```
johnduffell: "proxy" instead of "render"?

applications/conf/routes
```diff
@@ -89,6 +89,8 @@ GET        /$path<[\w\d-]*(/[\w\d-]*)?/(video|audio)/.*>/info.json
 GET        /$path<[\w\d-]*(/[\w\d-]*)?/(video|audio)/.*>.json                   controllers.MediaController.renderJson(path)
 GET        /$path<[\w\d-]*(/[\w\d-]*)?/(video|audio)/.*>                        controllers.MediaController.render(path)
 
+# Interactive service worker
+GET        /$path<[\w\d-]*(/[\w\d-]*)?(/[\w\d-]*)?>/interactive-service-worker.js           controllers.InteractiveController.renderInteractiveServiceworker(path)
```
johnduffell: check that *.js isn't blocked by the vcl with only a whitelist,  it used to be, but I'm not sure now

TBonnin (changes requested):

applications/app/controllers/InteractiveController.scala
```diff
@@ -5,21 +5,35 @@ import common._
...
+    val stage = if (isPreview) "preview" else "live"
+    val serviceWorkerPath = s"$cdnPath/service-workers/$stage/$path/interactive-service-worker.js"
+    wsClient.url(serviceWorkerPath).get().map { response =>
+      Cached(60) {
```
Rich-Harris: > I would cache it for far more than 1 minutes.
> Devs can use `router-prod.gu-web.net` if they want to bypass the cache

Why not just honour the cache header received from interactive.guim? That way it's just like any other part of an interactive, and control (and responsibility) rests with the interactive developer.

Otherwise, it needs to be something that all members of interactive teams have access to and instructions on how to use 😀. I can imagine a long cache time being a source of pain and confusion if we need to deploy fixes to interactives post-launch, as often happens.

In any case, from the user's perspective a short cache time on the service worker isn't really that bad, since a) requesting the service worker happens in the background while the existing registered service worker is intercepting fetch events (i.e. it needn't get in the way of other requests), and b) they tend to be small.

johnduffell: if we want to use the cache header, we should parse out the time, and generate a new header from that, to make sure we get all the right surrogate control etc.  Also might be worth putting in a floor of say 5 seconds (or just 404 if it's not present) to make sure it's not missing

TBonnin: @Rich-Harris 

My point of view is more from an infrastructure perspective:
Ideally, the cache time of any assets on the guardian website is very high, with a cache busting mechanism when assets are modified. (ie: We use content hash to differentiate each version). File is then always served from the CDN and we avoid any round trip to our servers.

Having this service-worker endpoint outside of the normal assets deployment makes thing a little bit more complicated here. 
However the best solution would be to use a similar mechanism. I believe the interactive team uses timestamp, instead of hash, but it would be best imho to store the file at `/TIMESTAMP/server-worker-interactive.js` in your S3 bucket. This would allow us to cache it forever while being able to push a new version instantly and not worry about cache time from a developer perspective.

What do you think?

applications/conf/routes
dominickendrick: Great spot ! :nerd_face:

applications/app/controllers/InteractiveController.scala
```diff
@@ -5,21 +5,49 @@ import common._
...
+
+    wsClient.url(serviceWorkerPath).get().map { response =>
+      Cached (7 days) {
+        RevalidatableResult(convertWSResponseToResult(response), response.body)
+      }
+    }
+  }
+
+  def convertWSResponseToResult(response: WSResponse): Result = {
+    val headers = response.allHeaders.map { case (header, body) =>
```
johnduffell: extremely nervous about allowing arbitrary headers through, we very nearly let through an HSTS header when we proxied discussion through theguardian.com as a temporary measure and that would have been super hard to get out of.  Obv not an issue now, but still.

applications/app/controllers/InteractiveController.scala
Rich-Harris: @TBonnin adding a timestamp to the service worker URL is a fine idea – would integrate well into our existing deployment process.

wpf500: Dom did say we should do this and I was a bit apprehensive about enforcing conventions on our deploy process, but happy to stand corrected if there is general agreement

johnduffell:
looks good now! probably not worth adding a switch with expiry as it should be dealt with soon?
your build is failing though!

applications/app/controllers/InteractiveController.scala
```diff
@@ -5,21 +5,42 @@ import common._
...
 
+  def proxyInteractiveServiceworker(path: String): Action[AnyContent] = Action.async { implicit request =>
+    val stage = if (isPreview) "preview" else "live"
+    val serviceWorkerPath = s"$cdnPath/service-workers/$stage/$path/interactive-service-worker.js"
+
+    wsClient.url(serviceWorkerPath).get().map { response =>
+      Cached (365.days) {
+        response.status match {
+          case 200 =>
+            RevalidatableResult(Ok(response.body).as("text/javascript; charset=utf8"), response.body)
```
TBonnin: Would it make sense to use `response.allHeaders("Content-Type")` rather the hardcoding it?

applications/app/controllers/InteractiveController.scala
```diff
@@ -5,21 +5,42 @@ import common._
 import contentapi.ContentApiClient
 import conf._
 import conf.switches.Switches
+import model.Cached.{WithoutRevalidationResult, RevalidatableResult}
 import model._
+import play.api.libs.iteratee.Enumerator
+import play.api.libs.ws.{WSResponse, WSClient}
```
TBonnin: I think you have some unused import here 😁

TBonnin (changes requested):

applications/app/controllers/InteractiveController.scala
johnduffell: they'll be serving gifs off it before we know it 😆

Rich-Harris:
Following on from a conversation within the interactives WG, could we change the pattern to support both `interactive-service-worker.js` *and* `interactive-worker.js`? At present there's no way to use a web worker in an interactive due to cross-origin restrictions and our CSP rules prohibiting dynamic worker creation via blobs. We could kill two birds with one stone here. Thanks!

dominickendrick:
> Following on from a conversation within the interactives WG, could we change the pattern to support both interactive-service-worker.js and interactive-worker.js? At present there's no way to use a web worker in an interactive due to cross-origin restrictions and our CSP rules prohibiting dynamic worker creation via blobs. We could kill two birds with one stone here. Thanks!

@Rich-Harris This is fine

dominickendrick:
This now supports an optional `timestamp` query parameter which allows us to cache bust, and load service workers from `/path/to/<timestamp>/interactive-service-worker.js`

applications/app/controllers/InteractiveController.scala
```diff
@@ -3,23 +3,51 @@ package controllers
...
+            val contentType = response.allHeaders("Content-Type").mkString(",")
+            RevalidatableResult(Ok(response.body).as(contentType), response.body)
+          }
+          case otherStatus => WithoutRevalidationResult(new Status(otherStatus))
+        }
+      }
+    }
+  }
+
+  def getWebWorkerPath(path: String, file: String, timestamp: Option[String]): String = {
```
TBonnin: private?

TBonnin (changes requested):

TBonnin (approved):

PR #14763, opened 2016-10-24
Use specified AWS format for date string

rjmk:
## What does this change?

How the value for the header on S3 requests is computed

Frontend can now run on (a) computers not set to English
(b) computers not set to UTC

This particularly resolves issues with fetching the pressed JSON for facia

No
## Request for comment

cc/ @guardian/dotcom-platform


rjmk:
Forgot to link in PR description: http://docs.aws.amazon.com/general/latest/gr/sigv4-date-handling.html

NataliaLKB:
Oh wow! Amazing find. Thanks for fixing this 👍

PR #14767, opened 2016-10-24
Hosted Video - onward journeys

lps88:
## What does this change?

The onward journey on Hosted Video now looks the same as it does on article pages. 

It means we have less code, and can soon test the new onward journey designs on both types of pages.
## Screenshots

image
## Request for comment

@guardian/labs-beta


commercial/app/controllers/HostedContentController.scala
```diff
@@ -85,7 +85,7 @@ class HostedContentController(contentApiClient: ContentApiClient)
           contentType match {
             case "video" =>
               val trails = HostedTrails.fromContent(itemId, trailCount = 1, results)
-              Cached(cacheDuration)(JsonComponent(hostedVideoOnward(trails.headOption)))
+              Cached(cacheDuration)(JsonComponent(hostedArticleOnward(trails, 1, 1)))
```
kelvin-chappell: Should probably rename the onward component to something like `hostedOnward` if it's generic.

commercial/app/views/hosted/hostedArticleOnward.scala.html
```diff
@@ -1,5 +1,5 @@
 @import common.commercial.hosted.HostedPage
-@(trails: Seq[HostedPage])(implicit request: RequestHeader)
+@(trails: Seq[HostedPage], defaultRows: Int = 2, maxRows: Int = 4)(implicit request: RequestHeader)
```
kelvin-chappell: I would take out the default values as there aren't that many places where it's used.  Would be clearer to have to specify them in the calling code.

kelvin-chappell (approved):
Very minor comments but otherwise 👍

PR #14768, opened 2016-10-25
Add one page test for new sonobi module

rich-nguyen:
This change adds a new module path for our sonobi integration, for a single page. This new module contains a rubicon adapter, used during the server-side auction.


kenlim (commented):
:+1:

PR #14770, opened 2016-10-25
Simplify building hosted trails

kelvin-chappell:
Gives an alternative way to build trails for the onward hosted container when all content in the campaign is to be used.

/cc @lps88 @guardian/labs-beta


lps88:
Good idea 👍

PR #14771, opened 2016-10-25
Better detection of native ads

regiskuckaertz:
Google recently changed a pivotal attribute in the detection of native ads, the `name` of the iframe. The consequence is that no message could be exchanged between native ads and the parent frame since then.

This led me to find a better way to perform the detection, using the size of the returned creative. Alas, this size is not available in the `onSlotLoad` event, only in the `onSlotRender` one, so I had to add a new property to the `Advert` structure.


janua:
:+1:

PR #14772, opened 2016-10-25
Linting testing take 2

sndrs:
## What does this change?

rerun of #14759 with the addition of piping task stdout to process.stdout in versbose mode, for teamcity

hopefully TC should recognise the tasks are running

no
## Request for comment

@SiAdcock

### DIFF
--- a/tools/run-task
+++ b/tools/run-task
@@ -104,7 +104,9 @@ const cache = [];
 const cmd = task => {
     const [binary, ...options] = task.trim().split(' ');
     if (verbose) {
-        const exec = execa(binary, options);
+        const exec = execa(binary, options, {
+            encoding: 'buffer'
+        });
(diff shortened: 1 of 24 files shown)


tools/run-task
```diff
@@ -91,39 +97,27 @@ const {dev: IS_DEV, verbose, _: TASKS} = yargs
...
-    } catch (e) {
-        return null;
-    }
-};
-
 // don't duplicate tasks
 const cache = [];
 
+// run commands with output to stdout if verbose is true
+const cmd = task => {
```
sndrs: this guy is the substantive difference compared with #14759

sndrs: this guy is the substantive difference compared with #14759

SiAdcock:
Nice catch 👍

PR #14775, opened 2016-10-25
Remove AMP live blog switches

SiAdcock:
## What does this change?

Live Blogs that use the LiveBlogPosting structured data format and live updates are now supported in AMP. We no longer need the switches that control these features. Disabling AMP will allow us to disable AMP Live Blogs, if absolutely necessary. 

Tidy up switches that are no longer needed

It only affects AMP 😄 
## Request for comment

@NataliaLKB @stephanfowler


NataliaLKB:
👍

PR #14778, opened 2016-10-25
Remove ContributionsUsa1 test

jranks123:
## What does this change?

Remove the ContributionsUsa1 test from the codebase

N/A


jfsoul (approved):
👍 

how did the test go?

jranks123:
thanks @jfsoul, it went really well actually, performed much better than expected (it's the first test we've ran like this in the US)

PR #14780, opened 2016-10-25
Hosted onward journey AB test

lps88:
## What does this change?

 Add a test for the new Onward Journey designs.

See which encourages the most clicks.
## Screenshots

image
## Request for comment

@guardian/labs-beta 
@desbo

### DIFF
--- a/common/app/conf/switches/ABTestSwitches.scala
+++ b/common/app/conf/switches/ABTestSwitches.scala
@@ -72,7 +72,7 @@ trait ABTestSwitches {
     "Show more pages from the campaign in Hosted Article/Video pages, using a carousel or popup",
     owners = Seq(Owner.withGithub("lps88")),
     safeState = Off,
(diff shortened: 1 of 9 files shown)


common/app/conf/switches/ABTestSwitches.scala
```diff
@@ -65,4 +65,14 @@ trait ABTestSwitches {
...
     exposeClientSide = true
   )
+
+  Switch(
+    ABTests,
+    "ab-hosted-onward-journey",
+    "Show more pages from the campaign in Hosted Article/Video pages, using a carousel or popup",
+    owners = Seq(Owner.withGithub("lps88")),
+    safeState = Off,
+    sellByDate = new LocalDate(2016, 12, 20),
```
kelvin-chappell: Not sure what the relationship is between switch expiring and test ending, but this should probably expire later than the end of the test.

kelvin-chappell (approved):
👍

PR #14781, opened 2016-10-26
New header: Fixing bug where on https, section link starts with `//`

NataliaLKB:
## What does this change?

There is currently a bug in the new header test where if you click on a section link instead of taking you to `www.theguardian.com/...` it takes you to `//...`, which of course doesn't exist.

Interestingly, I didn't catch this locally because locally all urls are relative and works fine. 

This change fixes this problem.

People can make the onward journeys that they expect

Nope!
## Request for comment

@dominickendrick @guardian/dotcom-platform


dominickendrick:
👍 🔗 😃

PR #14782, opened 2016-10-26
Fix: Multiple testsuites (extending OneServerPerSuite) use same test port

TBonnin:
## What does this change?

AdminTestSuite was set to specifically use port 19001 which is the `OneServerPerSuite` default testport
CachedHealthCheckTest was not overriding the test port and therefore was also using the port 19001
=> Possible conflict if the 2 test suites were run concurrently
This patch:
- changes the test port override for AdminTestSuite to another port (19015)
- Create a new test suite for common tests and override the test port to 19016. This new test suite is not absolutely necessary but would become handy if we are adding more tests which are shared and need to run with a server.

No more `org.jboss.netty.channel.ChannelException: Failed to bind to: /0.0.0.0:19001` errors when running tests

## Request for comment

@rich-nguyen @guardian/dotcom-platform

### DIFF
--- a/common/test/layout/WidthsByBreakpointTest.scala
+++ b/common/test/layout/WidthsByBreakpointTest.scala
@@ -2,8 +2,6 @@ package layout
 
 import layout.ContentWidths._
 import org.scalatest._
-import org.scalatest.concurrent.Eventually
-import test.SingleServerSuite
 
 class WidthsByBreakpointTest extends FreeSpec with ShouldMatchers {
   "ContentWidths" - {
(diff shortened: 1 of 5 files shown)


common/test/layout/WidthsByBreakpointTest.scala
```diff
@@ -2,8 +2,6 @@ package layout
 
 import layout.ContentWidths._
 import org.scalatest._
-import org.scalatest.concurrent.Eventually
-import test.SingleServerSuite
```
TBonnin: Unrelated to this PR but those imports are not used

rich-nguyen:
Great. So is anything using the default test port 19001? Can it be removed?

rich-nguyen:
👍

TBonnin:
I quickly looked and other than the `CachedHealthCheckTest` I couldn't find anything else extending `OneServerPerSuite` and not overriding the port.

rich-nguyen:
Oh i see, the trait contains a default port. ok sounds good.

PR #14787, opened 2016-10-26
Inline require calls for interactives

wpf500:
~~This is a proof of concept, the code is very hacky, but it would be great to get people's thoughts about the concept.~~

All of this only applies to the interactive content type, not interactive embeds on articles etc.

Interactive's are currently loaded in `app.js`. This means that interactives load very late (~2-3 secs on good machine/internet) and experience performance problems as they load (competing for resources with commercial components etc.). An interactive's JS is an important part of the content, so maybe we should defer loading `app.js` to later?
## What does this change?

The purpose of this PR is two-fold:
- Load interactive JS as soon as possible
- ~~Defer loading `app.js` to after interactive's have finished booting~~

I don't think interactives are ever available off platform

This would affect old interactives as we would be inlining their boot loader.

### DIFF
--- a/applications/app/views/fragments/interactiveBody.scala.html
+++ b/applications/app/views/fragments/interactiveBody.scala.html
@@ -66,13 +66,8 @@
     <script>
         @HtmlFormat.raw(common.Assets.js.curl)
 
-        [].slice.apply(document.querySelectorAll('figure.interactive')).forEach(function (el) {
-            var mainJS = el.getAttribute('data-interactive');
-            if (!mainJS) {
-                return;
-            }
-
-            require([mainJS], function (interactive) {
+        [].slice.apply(document.querySelectorAll('figure.interactive[data-interactive]')).forEach(function (el) {
+            require([el.getAttribute('data-interactive')], function (interactive) {
                 interactive.boot(el, document, window.guardian.config);
             }, function (err) {
                 console.log('Interactive failed', mainJS);
--- a/applications/app/views/fragments/interactiveBody.scala.html
+++ b/applications/app/views/fragments/interactiveBody.scala.html
@@ -62,4 +62,22 @@
         </div>
         @fragments.contentFooter(interactive, page.related, isPaidContent = isPaidContent)
     }
+
+    <script>
+        @HtmlFormat.raw(common.Assets.js.curl)
+
+        [].slice.apply(document.querySelectorAll('figure.interactive')).forEach(function (el) {
+            var mainJS = el.getAttribute('data-interactive');
+            if (!mainJS) {
+                return;
+            }
+
+            require([mainJS], function (interactive) {
+                interactive.boot(el, document, window.guardian.config);
+            }, function (err) {
+                console.log('Interactive failed', mainJS);
+                console.log('Error:', err);
+            });
+        });
+    </script>
 }
--- a/applications/app/views/fragments/interactiveBody.scala.html
+++ b/applications/app/views/fragments/interactiveBody.scala.html
@@ -63,16 +63,18 @@
         @fragments.contentFooter(interactive, page.related, isPaidContent = isPaidContent)
     }
 
-    <script>
-        @HtmlFormat.raw(common.Assets.js.curl)
+    @if(conf.switches.Switches.InlineInteractiveRequire.isSwitchedOn) {
+        <script>
+            @HtmlFormat.raw(common.Assets.js.curl)
 
-        [].slice.apply(document.querySelectorAll('figure.interactive[data-interactive]')).forEach(function (el) {
-            require([el.getAttribute('data-interactive')], function (interactive) {
-                interactive.boot(el, document, window.guardian.config);
-            }, function (err) {
-                console.log('Interactive failed', mainJS);
-                console.log('Error:', err);
+            [].slice.apply(document.querySelectorAll('figure.interactive[data-interactive]')).forEach(function (el) {
+                require([el.getAttribute('data-interactive')], function (interactive) {
+                    interactive.boot(el, document, window.guardian.config);
+                }, function (err) {
+                    console.log('Interactive failed', mainJS);
+                    console.log('Error:', err);
+                });
             });
-        });
-    </script>
+        </script>
+    }
 }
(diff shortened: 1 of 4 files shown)


static/src/javascripts/components/curl/curl-domReady.js
```diff
@@ -1,4 +1,4 @@
-(function(){/*
+window.require || (function(){/*
```
wpf500: This is a hack! It seems that loading `curl` twice makes things fail, so this just stops the `app.js` copy of curl from loading if there is already one in the page.

wpf500:
@Rich-Harris @guardian/dotcom-platform

static/src/javascripts/components/curl/curl-domReady.js
SiAdcock: An alternative might be to create a different `app.js` bundle just for interactive pages, which doesn't include `curl`. This is slightly less hacky, but maybe overkill?

wpf500: Yeah that is an option. I think it would be good to avoid a separate `app.js` if possible for simplicity, but also for things like caching etc.

applications/app/views/fragments/interactiveBody.scala.html
```diff
@@ -62,4 +67,32 @@
         </div>
         @fragments.contentFooter(interactive, page.related, isPaidContent = isPaidContent)
     }
+
+    <script>
+        require(['Promise'], function (Promise) {
```
Rich-Harris: Users without native promises are a small and dwindling minority – perhaps it should be something like this, to speed things up for the majority?

```js
function init (Promise) {
  // ...
}

window.Promise ? init(Promise) : require(['Promise'], init);
```

Rich-Harris (commented):
Thanks @wpf500, this is a huge and much-needed step in the right direction. Forgive my ignorance but would this also defer loading of `commercial.js` and `main.js` (the 'other' main.js, not the atom...), i.e. are they dependencies of `app.js`?

applications/app/views/fragments/interactiveBody.scala.html
wpf500: I had hoped that for users with native Promise this becomes a synchronous require call but then I don't really know if that's even allowed in the AMD spec :stuck_out_tongue:

wpf500: I should probably check but that but that would make sense to me.

wpf500:
`commercial.js` is a dependency of `app.js` so this would defer that too.

applications/app/views/fragments/interactiveBody.scala.html
wpf500: Hmm I think that the Promise module is always bundled in PROD so it probably wouldn't even be available, @SiAdcock do you know if this is the case?

Rich-Harris:
@wpf500 is the change in b9c0dd3641bd5c94c910d1af701d77cefb546181 intentional? Previously, interactives could return a promise from `boot(...)` which signalled they had completed their setup work and that it was safe to load `app.js`. As of b9c0dd3641bd5c94c910d1af701d77cefb546181 `app.js` would load as soon as all the interactives had been fetched and evaluated, which isn't the same thing – it solves the problem of the network request for `app.js` getting in the way of interactives, but no longer solves the significant jank those modules cause during interactives' intro sequences.

wpf500:
@Rich-Harris yeah pending a chat this morning, hoping to add something similar back in. We're keen to remove the dependency on Promises from inline code, but I've just had a thought that that would actually be quite easy, something like:
```
var p = interactive.boot(...)
p && p.then ? p.then(loaded) : loaded()
```

wpf500:
Following on from various conversations, this PR is now just about inlining `require` requests on interactive content, deferring `app.js` has been dropped. This is now just a performance improvement for existing interactives, and will not affect the new interactive atoms.

What do you think @SiAdcock @sndrs?

SiAdcock:
Good shout @wpf500. 

Is there much advantage in delaying the request for `app.js` until after the boot script has been loaded? I wonder if the `'loaded()` function is now effectively redundant... Can we not just call `@HtmlFormat.raw(loadApp().body)` at the top of the script tag?

wpf500:
Hmm yeah that's an excellent point, I should just be able to keep everything the same except inlining curl and require calls. I'll change it

wpf500:
@SiAdcock, a much simplified PR now! I've squashed all the commits to just what we've ended up with

applications/app/views/fragments/interactiveBody.scala.html
```diff
@@ -62,4 +62,22 @@
         </div>
         @fragments.contentFooter(interactive, page.related, isPaidContent = isPaidContent)
     }
+
+    <script>
+        @HtmlFormat.raw(common.Assets.js.curl)
+
+        [].slice.apply(document.querySelectorAll('figure.interactive')).forEach(function (el) {
+            var mainJS = el.getAttribute('data-interactive');
+            if (!mainJS) {
```
sndrs: would `document.querySelectorAll('figure.interactive[data-interactive]')` simplify this? remove the need to exit?

applications/app/views/fragments/interactiveBody.scala.html
```diff
@@ -62,4 +62,22 @@
...
+                return;
+            }
+
+            require([mainJS], function (interactive) {
+                interactive.boot(el, document, window.guardian.config);
+            }, function (err) {
+                console.log('Interactive failed', mainJS);
+                console.log('Error:', err);
```
sndrs: `console.error`?

wpf500: Wasn't sure if that was available on old browsers?

sndrs: yeah fair enough. there's _bound_ to be one it causes a problem in  :)

SiAdcock:
Awesome stuff, dude 👍

wpf500:
@Jholder112233 has suggested that I put this round a switch as it affects all interactives so would be nice to be able to quickly turn it off if stuff does break, so gonna do that.

wpf500:
@SiAdcock, @sndrs switch is done, ready to merge?

SiAdcock:
I like it @wpf500, but can we have the switch expire in, say a month, so it doesn't linger around forever? Presumably by then we will be confident as to whether it breaks things or not

wpf500:
Ah yup! Set it to two weeks because that should be more than enough time and its 11/11!

SiAdcock:
👍 👍 👍 👍 👍 👍 👍 👍 👍  Amaze!

PR #14789, opened 2016-10-26
Membership engagement message copy experiments

JustinPinner:
## What does this change?

Adds a multivariate test for three text versions (plus control) within the membership messaging UI.
**This is behind a switch that defaults to off**, so won't appear on prod until enabled manually.
Tests the hypothesis that more persuasive copy will improve conversions.

Find the most appealing and effective copy.

No, web only.
## Screenshots

The banner styling isn't changing in this PR, only the text that appears inside it.
## Request for comment

@rtyley @svillafe @Ap0c


JustinPinner:
cc @jranks123

svillafe:
I went through the code with @JustinPinner and it seems ok 👍 . 
We talked with @NataliaLKB and since it is a small change in an AB test for the engagement banner, it is not necessary that a member of frontend review it.

JustinPinner:
Thanks @svillafe - merging now...

PR #14790, opened 2016-10-26
Optimise Membership Engagement Banner Logic

jranks123:
## What does this change?

A small optimisation that stops an unnecessary asynchronous call to commercial features that doesn't need to be called if the user has less than 10 visits

Stops unnecessary code being ran => saves energy => saves some trees. 

No

## Request for comment

@jfsoul


jfsoul (approved):
Good catch! 👍

PR #14791, opened 2016-10-26
Update babel config

sndrs:
## What does this change?
- adds seperate prod and dev babel config 
- adds `async`/`await` for dev env (since node 6 only really needs `async`/`await`)
- updates prod env to `latest` preset (2015, 16 + 17)

i need to use `async`/`await` in the tasks :)

keeps things up to date too

no
## Request for comment

@guardian/dotcom-platform

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,14 @@
+{
+    "env": {
+        "production": {
+            "presets": [
+                "latest"
+            ]
+        },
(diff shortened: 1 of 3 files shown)


SiAdcock:
It's unlikely that the `async/await` API will change before hitting stage 4. I whole-heartedly approve 👍

PR #14797, opened 2016-10-27
Allow Outlook to use correct fallback font

joelochlann:
Outlook doesn't support web fonts and also doesn't handle fallback fonts correctly, so all "Guardian Egyptian Text" currently comes out as Times New Roman despite the fallback font of Georgia.

Now we explicitly set Georgia as the preferred font in Outlook. Since bold Georgia in the subheadings looks a bit weird, I also un-bolded the `<h2></h2>` elements.
## Outlook before:

outlook 2016 before
## Outlook after:

outlook 2016 after
## Gmail before:

gmail before
## Gmail after:

gmail after

@desbo

### DIFF
--- a/common/app/views/fragments/email/stylesheets/main.scala.html
+++ b/common/app/views/fragments/email/stylesheets/main.scala.html
@@ -28,7 +28,6 @@
     }
 
     h2 {
-        font-weight: bold;
         font-size: 24px;
         line-height: 28px;
         margin-bottom: 5px;
--- a/common/app/views/mainEmail.scala.html
+++ b/common/app/views/mainEmail.scala.html
@@ -12,6 +12,20 @@
         <link rel="canonical" href="@LinkTo(page.metadata.canonicalUrl.map(LinkTo(_)).getOrElse(CanonicalLink(request, page.metadata.webUrl)))" />
         <title>@views.support.Title(page)</title>
 
+        @*
+         * Outlook doesn't support web fonts and also doesn't handle
+         * fallback fonts correctly, so all "Guardian Egyptian Text" will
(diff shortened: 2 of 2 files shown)


common/app/views/mainEmail.scala.html
```diff
@@ -12,6 +12,20 @@
         <link rel="canonical" href="@LinkTo(page.metadata.canonicalUrl.map(LinkTo(_)).getOrElse(CanonicalLink(request, page.metadata.webUrl)))" />
         <title>@views.support.Title(page)</title>
 
+        @*
+         * Outlook doesn't support web fonts and also doesn't handle
+         * fallback fonts correctly, so all "Guardian Egyptian Text" will
+         * come out as Times New Roman despite the fallback font of Georgia.
+         * So for Outlook we explicitly set Georgia as the preferred font.
+         *@
+        <!--[if mso]>
```
sndrs: nice!

common/app/views/fragments/email/stylesheets/main.scala.html
```diff
@@ -28,7 +28,6 @@
     }
 
     h2 {
-        font-weight: bold;
```
sndrs: does this look weird where webfonts are available?

joelochlann: Good question, I did check this and I don't personally think so. I thought it would be tricky so set up a conditional style that was only bold when the webfont was available (if you have any thoughts on this let me know) and it didn't seem worth the effort given the following difference:

# before
picture 317

# after
picture 318

desbo (approved):

PR #14803, opened 2016-10-27
Return tasks for icons task, not promises

sndrs:
## What does this change?

makes the inlining of SVG icons task a return properly

it was working before but only if it won the race with beginning `grunt-sass`, because it was concluding before the promises has resolved

no
## Request for comment

@guardian/dotcom-platform


SiAdcock:
LGTM 👍

PR #14806, opened 2016-10-27
Unify Fabric background handling

regiskuckaertz:
Based on feedback coming from the DAP team and also to improve the performance of creatives, both fixed and parallax scroll effects are now handled the same way in a single module.

Also the background is promoted to its own self-contained layer, which will reduce the paint work in recent browsers.


JonNorman (approved):
🚂

PR #14813, opened 2016-10-28
No commercial.js on immersive interactives with ads disabled

wpf500:
On immersive interactives with ads disabled, `commercial.js` doesn't add anything, so lets just not load it. `enhanced.js` doesn't add anything either and seems to have some dependencies on `commercial.js`, so disable that too!

The only thing that does might be useful is the navigation, so I've moved the navigation initialisation from enhanced to standard. From what I can tell the `navigation` module doesn't require anything that standard doesn't already have, and the `newHeaderNavigation` is already in standard, so why not put both in there?

### DIFF
--- a/static/src/javascripts/boot.js
+++ b/static/src/javascripts/boot.js
@@ -32,6 +32,7 @@ define([
 
     var guardian = window.guardian;
     var config = guardian.config;
+    var page = config.page;
 
     var domReadyPromise = new Promise(function (resolve) { domReady(resolve); });
 
@@ -40,12 +41,14 @@ define([
             .then(function (boot) { boot(); });
     };
 
+    var suppressCommercial = !config.switches.commercial || (page.contentType === 'Interactive' && page.isImmersive && page.shouldHideAdverts);
+
     var bootCommercial = function () {
-        if (!config.switches.commercial) {
+        if (suppressCommercial) {
             return;
         }
 
-        if (config.page.isDev) {
+        if (page.isDev) {
             guardian.adBlockers.onDetect.push(function (isInUse) {
                 var needsMessage = isInUse && window.console && window.console.warn;
                 var message = 'Do you have an adblocker enabled? Commercial features might fail to run, or throw exceptions.';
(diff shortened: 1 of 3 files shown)


wpf500:
@stephanfowler @SiAdcock

rich-nguyen:
ccing @guardian/commercial-dev

SiAdcock:
My main concern with this approach is that excluding `enhanced.js` from Immersive Interactives would exclude more than just commercial dependencies. We would also exclude other functionality such as some tracking tools (notably our facade for Google Analytics) and notifications.

I know we probably don't need _all_ of the functionality loaded by `enhanced.js` on this page, but it is possible that we will end up excluding something we need. 

If we do need things like Google Analytics on Immersive Interactives articles, an alternative approach would be to bundle and include a stripped-down version of `enhanced.js` containing only the modules we need.

@NataliaLKB @sndrs: any thoughts on this?

wpf500:
Yeah definitely agree on tracking. I thought we'd come to the conclusion that all the tracking stuff was inline on the page anyway, so this wasn't removing that stuff

NataliaLKB:
> I know we probably don't need all of the functionality loaded by enhanced.js on this page, but it is possible that we will end up excluding something we need.

I completely agree with @SiAdcock . 

It might be best to ask @dominickendrick regarding google analytics. But from what we can tell it is definitely loaded in enhanced. Now that we are not using Omniture, it seems important to have google analytics on interactive pages

regiskuckaertz:
Why immersive articles specifically? Can this change (at least removing `commercial`) not be applied across the board?

wpf500:
This is for immersive interactives, not immersive articles (unfortunately there is a difference!). It only applies to the former because all other content types have things like related content which needs `commercial`/`enhanced`. It might be that they only need `enhanced` but not `commercial`, but that's a much bigger piece of work because currently `commercial` pulls in a lot of dependencies.

SiAdcock:
Confirmed with @dominickendrick that analytics is loaded in `enhanced`, therefore unfortunately I don't think excluding `enhanced` on immersive interactive articles is the correct approach for dealing with this issue. It's a shame that `enhanced` has become this giant bundle of hugs that gets added to every page whether we need everything or not, but we're hoping to deal with that as part of the ongoing client side re-architecture.

I'm in favour of moving `navigation` to `standard` if that helps with the jank. Also happy to discuss further

wpf500:
This is much more complicated than initially envisaged, moving to a hack (https://github.com/guardian/frontend/pull/14905) to get the desired effect in the short term (i.e. for tomorrow!)

PR #14816, opened 2016-10-28
Hosted native cards - fix the font contrast

lps88:
## What does this change?

Makes the font colour white if the brand colour is dark on Hosted By native cards.

## Screenshots

image
## Request for comment

@Calanthe


Calanthe:
👍  You could also write some unit tests ;)

PR #14820, opened 2016-10-28
Contributions uk vs us test

philwills:
## What does this change?

Adds tests to compare whether different messages are more effective in the US compared to the UK.

We hope to be able to tell if we can use the same message in the US as for the UK to ask for contributions

No

## Screenshots

image
## Request for comment

@jfsoul @jranks123 

Two tests to run simultaneously, one in the UK, one in the US to test whether different messages are more effective in the two countries.

### DIFF
--- a/static/src/javascripts/projects/common/modules/experiments/tests/contributions-countries-uk.js
+++ b/static/src/javascripts/projects/common/modules/experiments/tests/contributions-countries-uk.js
@@ -35,11 +35,11 @@ define([
 
     return function () {
 
-        this.id = 'ContributionsCountriesUK';
+        this.id = 'ContributionsCountriesUk';
         this.start = '2016-10-28';
         this.expiry = '2016-11-04';
         this.author = 'Phil Wills';
-        this.description = 'Test whether different messages perform better/worse in different countries';
+        this.description = 'Test whether different messages perform better/worse in different countries (UK)';
         this.showForSensitive = false;
         this.audience = 0.15;
         this.audienceOffset = 0.56;
--- a/static/src/javascripts/projects/common/modules/experiments/tests/contributions-countries-uk.js
+++ b/static/src/javascripts/projects/common/modules/experiments/tests/contributions-countries-uk.js
@@ -1,0 +1,143 @@
+define([
+    'bean',
+    'qwery',
+    'common/utils/$',
+    'common/utils/template',
+    'common/views/svg',
+    'common/utils/fastdom-promise',
+    'common/utils/mediator',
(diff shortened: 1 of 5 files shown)


static/src/javascripts/projects/common/modules/experiments/tests/contributions-countries-uk.js
```diff
@@ -0,0 +1,143 @@
...
+                },
+                success: completer
+            },
+            {
+                id: 'global',
+                test: function () {
+                    var component = $.create(template(contributionsEmbed, {
+                        position : 'inline',
+                        variant: 'bottom',
+                        titleCopy: 'Reporting from a global perspective is vital. But it’s also expensive. Please give to the Guardian today. ',
```
jfsoul: Noticed that we're using a backtick here but an escaped single quote in `democracy`.  Is that intentional do you think?

static/src/javascripts/projects/common/modules/experiments/tests/contributions-countries-uk.js
```diff
@@ -0,0 +1,143 @@
...
+    return function () {
+
+        this.id = 'ContributionsCountriesUk';
+        this.start = '2016-10-28';
+        this.expiry = '2016-11-04';
+        this.author = 'Phil Wills';
+        this.description = 'Test whether different messages perform better/worse in different countries (UK)';
+        this.showForSensitive = false;
+        this.audience = 0.15;
+        this.audienceOffset = 0.56;
```
jfsoul: Not sure I understand the 0.56 and 0.5 offsets.

static/src/javascripts/projects/common/modules/experiments/tests/contributions-countries-uk.js
philwills: No, I expect that's some copy-paste error, I'll update.

static/src/javascripts/projects/common/modules/experiments/tests/contributions-countries-uk.js
philwills: Sorry, should be clear that we've done much less testing in the US, so that's why we haven't already used up to 0.56 there.

jfsoul (approved):
👍 looks good!

static/src/javascripts/projects/common/modules/experiments/tests/contributions-countries-uk.js
jfsoul: Thanks for the explanation, all clear now!

PR #14823, opened 2016-10-31
Return Guardian logo to immersive template with embed code main media

holder-josh:
## What does this change?
Fixes the lack of a Guardian logo on the immersive template pages which use embed code as the main media.

Same underlying problem as https://github.com/guardian/frontend/pull/14761 fixed

> In this circumstance isMainEmbed returns false as CAPI doesn't return an element with the relation field set to main, unlike when you supply a bootjs as the main media, so the most reliable check is whether the main field is empty.


Jholder112233:
@blongden73 @NataliaLKB could you take a look please?

NataliaLKB:
lgtm 👍

PR #14821, opened 2016-10-28
Git hooks fix

sndrs:
## What does this change?
- it should reinstate the git hooks lint checks
- it adds error messages back to `make validate`

## What is the value of this and can you measure success?

should stop errors getting out
## Does this affect other platforms - Amp, Apps, etc?

no
## Screenshots

screen shot 2016-10-28 at 17 20 41
screen shot 2016-10-28 at 17 20 22
screen shot 2016-10-28 at 17 19 53
screen shot 2016-10-28 at 17 19 40
## Request for comment

@NataliaLKB @SiAdcock

PR #14824, opened 2016-10-31
--color is not universal, shoudl only be on things that support it...

sndrs:
## What does this change?

made a booboo in #14821 🙄 

eslint errors are coloured, but r.js doesn't break

no

## Screenshots

before:

screen shot 2016-10-31 at 10 33 11

after:

screen shot 2016-10-31 at 10 33 25
screen shot 2016-10-31 at 10 33 34

## Request for comment

@NataliaLKB @SiAdcock


NataliaLKB:
👍

PR #14825, opened 2016-10-31
Extend prominent membership engagement test by one day

JustinPinner:
## What does this change?
Adds one day to the expiry of the **prominent membership engagement AB test** (and switch). Based on the observed impressions rate so far, we estimate that we'll achieve statistical significance by the end of Thursday 3rd.

Don't end the test too soon!

No, web only.

## Screenshots
N/A (for this PR)

## Request for comment
@svillafe @Ap0c @rtyley @rupertbates


rupertbates (commented):
Is the 23:59:59 time difference between the 2 files correct?

JustinPinner:
It seems that tests shut down on the transition into the day they are set to expire, whereas switches expire when that day has elapsed.

rupertbates (approved):
👍

PR #14826, opened 2016-10-31
Make images links in minute email

joelochlann:
We're currently testing a Weekend Reading email in two formats:
- [minute-like](https://www.theguardian.com/membership/live/2016/oct/29/weekend-reading-heathrow-aggro-tabloids-revenge-and-good-scares/email) (generated from a live blog)
- [article-like](https://www.theguardian.com/membership/2016/oct/29/weekend-reading-heathrow-aggro-tabloids-revenge-and-good-scares/email) (generated from an article)

The difference is a bit subtle and there's not a lot to choose between them right now. I'm curious about whether making the images into links in the minute format will have an impact on the clickthrough rate relative to the article format, based on my anecdotal instinct to click on the images whenever I open the email. ([Click-To-Open-Rate](http://www.mailermailer.com/resources/metrics/2013/click-to-open-rates.rwp) is one of our email-related key results)

Making some minute-y assumptions that there's one link inside each block, I can try and generate a url for each block in the liveblog-generated email, and make the image link to that url.

@davidfurey @desbo

### DIFF
--- a/article/app/views/fragments/emailArticleBody.scala.html
+++ b/article/app/views/fragments/emailArticleBody.scala.html
@@ -81,9 +81,15 @@
                     }
 
                     case ImageBlockElement(media, data, showCredit) => {
-                        @EmailImage.bestFor(media).map { url =>
+                        @EmailImage.bestFor(media).map { imageUrl =>
                             @fullRow {
-                                <img width="580" class="full-width" src="@url" @data.get("alt").map { alt => alt="@alt" }>
+                                @block.url.fold {
+                                    <img width="580" class="full-width" src="@imageUrl" @data.get("alt").map { alt => alt="@alt" }>
+                                } { linkUrl =>
+                                    <a href="@linkUrl">
+                                        <img width="580" class="full-width" src="@imageUrl" @data.get("alt").map { alt => alt="@alt" }>
+                                    </a>
+                                }
                             }
 
                             @paddedRow {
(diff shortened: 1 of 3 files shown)


joelochlann:
@guardian/dotcom-platform

article/app/views/fragments/emailArticleBody.scala.html
```diff
@@ -81,9 +81,15 @@ <h2 class="block__title">@title</h2>
                     }
 
                     case ImageBlockElement(media, data, showCredit) => {
-                        @EmailImage.bestFor(media).map { url =>
+                        @EmailImage.bestFor(media).map { imageUrl =>
                             @fullRow {
-                                <img width="580" class="full-width" src="@url" @data.get("alt").map { alt => alt="@alt" }>
+                                @block.url.filter(_ => article.isLiveBlog).fold {
+                                    <img width="580" class="full-width" src="@imageUrl" @data.get("alt").map { alt => alt="@alt" }>
```
TBonnin: would it make sense to have an helper function in EmailHelpers for this image Html? It seems to be used several times in this template.

joelochlann:
Thanks for the suggestion, what do you think now @TBonnin? Template looks a bit cleaner though I'm not sure the helper method is that easy to understand (template syntax is nicer for that sort of thing) - maybe that's fine though

TBonnin:
@joelochlann I was actually only thinking about factoring out the `` tag not the whole logic with the link. However it is a detail. Up to you. If you keep it the way it is, I will name the parameters when you call the function though, it is not obvious what parameter is what.

joelochlann:
Yeah I think your suggestion's better @TBonnin, it was overcomplicating things putting both the `<a>` and `` markup into the helper function.

Have made that change, added a bit of protection against runtime errors when parsing the block HTML, and made it a bit more efficient by ensuring we only try and parse the HTML if it's actually a live blog (if it's not we don't expect the `url` Option to actually contain a value so there's no need to compute it).

Good to merge?

TBonnin (approved):

PR #14827, opened 2016-10-31
Weekend Reading switches: delete one, extend the other

katebee:
## What does this change?

- Deletes the expiring Weekend Reading Promo (that was competing in the snap banner slot)
- Extends the AB test running over the Weekend Reading Sign Up page that assigns users to two different email lists in ExactTarget

- Less code in frontend
- We can continue testing with the email variants
- Build doesn't break tomorrow!

Nope

## Request for comment

@guardian/guardian-frontend


NataliaLKB:
@katebee I think teamcity is failing because of this line: https://github.com/guardian/frontend/blob/749d7f58a013fe7deb7f00e1b191c484607d4f6b/static/src/stylesheets/content.scss#L119

remove that and you should be good! Everything else lgtm 👍

PR #14830, opened 2016-10-31
Refactor top sticky banner

regiskuckaertz:
I had to touch the module for some native ads stuff (those ads that have dynamic sizes), and I could not for the life of me figure out how to do it by keeping the style consistent.

So, I rewrote it. I think the logic is now much simpler to follow. I also got rid of things that are not necessary anymore:

- the `scrollTo` kung-fu because we are lazy-loading the top slot
- positioning and transitioning are now mostly done in CSS, where they belong

cc @gtrufitt 

top-above-1

top-above-2


uplne:
I think there will be different viewability tho ☯️

regiskuckaertz:
hey @upine 😀 how so?

gtrufitt:
Does this still prevent pushing the page down if you're not at the top?

regiskuckaertz:
@gtrufitt I haven't seen anything like that in my tests (because the ad is only loaded, and thus resized, when you scroll to the top), but I will run some more tests to double-check.

gtrufitt:
Ah ok, so it doesn't load if you're not at the top, that's cool 💯

regiskuckaertz:
yes, this used to be the case though, which explains why Olly had to go through this hell 🔥  in hindsight, it doesn't make sense to load _any_ ad if it is not in the viewport. Sadly, we still have to do it, because other creatives (outbrain and inline MPUs) rely on the (non-)existence of other creatives.

static/src/javascripts/projects/commercial/modules/sticky-top-banner.js
```diff
@@ -1,250 +1,152 @@
...
+    }
 
-        var diff = state.adHeight - state.previousAdHeight;
-        var adHeightHasIncreased = diff > 0;
-        if (!state.shouldTransition && adHeightHasIncreased) {
-            // If we shouldn't transition, we want to offset their scroll position
-            var pageXOffset = state.scrollCoords[0];
-            els.window.scrollTo(pageXOffset, pageYOffset + diff);
+    function onTopAdRendered(isRendered) {
+        if (isRendered) {
```
gtrufitt: do you need to return a resolved `promise` in the else to ensure the `promise.all` resolves..?

static/src/stylesheets/module/_adslot.scss
```diff
@@ -69,9 +69,17 @@
 }
 
 .sticky-top-banner-ad {
+    transform: translate3d(0, 0, 0);
+    will-change: transform;
```
gtrufitt: It might be worth adding `will-change` in the JS in the fastdom read directly before you fastdom write the CSS class for animating  - hence giving the browser time to optimise and then [removing it when done](https://dev.opera.com/articles/css-will-change-property/#remove-will-change-after-the-changes-are-done) so the browser can recover resources.

gtrufitt (approved):
LGTM 👍

static/src/stylesheets/module/_adslot.scss
regiskuckaertz: Yes, very good point.

regiskuckaertz:
Chrome handles the case where the user is not at the top better than other browsers. The lazyloading kickstarts when the slot is within 1/2 a screen from the viewport, so in FF and Opera for instance as you scroll up, you can see a subtle difference in "velocity" but not a jump. Does that count?

regiskuckaertz:
Here on Safari:

scroll-up

static/src/stylesheets/module/_adslot.scss
regiskuckaertz: Come to think of it, moving it inside the animation class is a good proxy. Since adding the class will happen before and changes in height that should be animated.

regiskuckaertz:
Alright, these turned out to be trivial changes. Half the size and reduced cyclomatic complexity make for easier maintenance.

with scroll correction:
scroll-up

and without when the user is at the top:
scroll-up-2

regiskuckaertz:
[Refactor sticky top banner module](https://trello.com/c/xQovxxbh/122-refactor-sticky-top-banner-module)

gtrufitt:
👌 👍

PR #14835, opened 2016-10-31
Make Diagnostics httpFilters consistant with the other apps

TBonnin:
## What does this change?
Change Diagnostics httpFilters to CommonFilters rather than just GZip filter

@rich-nguyen any idea why it was like that? Thx

Make Diagnostics httpFilters consistent with the other apps.

## Request for comment
@jfsoul

### DIFF
--- a/diagnostics/app/AppLoader.scala
+++ b/diagnostics/app/AppLoader.scala
@@ -3,7 +3,7 @@ import com.softwaremill.macwire._
 import common._
 import common.Logback.LogstashLifecycle
 import conf.switches.SwitchboardLifecycle
-import conf.{CachedHealthCheckLifeCycle, CommonGzipFilter}
+import conf.{CachedHealthCheckLifeCycle, CommonFilters}
 import controllers.{Assets, DiagnosticsControllers, HealthCheck}
 import model.ApplicationIdentity
 import play.api.ApplicationLoader.Context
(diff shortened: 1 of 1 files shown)


rich-nguyen:
I think that may be because it doesn't need any of the normal filters. It doesn't need the  cDN caching filters (JsonVary, SurrogateKey), it doesn't need the amp filter, obviously, we don't want to do request logging for performance and to reduce log bloat, and the backend header won't be logged anywhere.

TBonnin:
I was actually trying to look into the request logs, which currently doesn't exist. Quite annoying. I agree that adding RequestLoggingFilter might bloat the logs but I would like to verify that. 
I also don't agree with Diagnostics not needing CDN caching. Some of the supported endpoint are GET.
The only one I agree is not necessary is the amp filter.

rich-nguyen:
I think real GETs should be moved off diagnostics (I made one, it can be moved now). The other GETs act more like tracking pixels. So, that would leave the diagnostics server with a small collection of uncacheable routes. So no CDN needed.

TBonnin:
ok. closing this PR. I am gonna try to debug without those logs. Thanks @rich-nguyen

rich-nguyen:
np @TBonnin, give me a shout if you need help debugging those pesky diagnostics requests.

PR #14837, opened 2016-10-31
Guardian witness is now styled on amp

NataliaLKB:
## What does this change?

Guardian Witness articles display on AMP, but we have been hiding the button. As its a simple styling job I thought I would do a quick PR to fix this problem.

People now can click a click to contribute to guardian witness

This only affects AMP

## Screenshots
Right now it doesn't show at all, but this is what it looks like now:
image


janua:
:+1:

PR #14669, opened 2016-10-12
Optimise email images

joelochlann:
Implements some of @paperboyo's excellent suggestions in #14423 for reducing image size for email endpoints.
## What is the value of this and can you measure success?

For the [latest Weekend Reading email](https://www.theguardian.com/membership/live/2016/oct/08/weekend-reading-mays-britain-scary-clowns-and-cheese-scones/email), which is particularly image-heavy:

```
                    before  after   reduction
                    54.9KB  24.2KB  55.92%
                    33.9KB  16.4KB  51.62%
                    50.1KB  24.7KB  50.70%
                    60.7KB  27.5KB  54.70%
                    42.KB   20.2KB  51.90%
                    29.7KB  14.8KB  50.17%

all images          430.KB  217.KB  49.53%
total pageweight    467.KB  254.KB  45.61%
```
### Before

picture 312
### After

picture 311
## Request for comment

@desbo 
cc @davidfurey

PR #14840, opened 2016-10-31
Tweak email image quality

joelochlann:
I might have been a bit too aggressive with reducing image quality in #14669. I've bumped the `q` imgix param back up a little bit. It's a subtle but noticeable difference, in my opinion, and only adds 49K overall to [a typical email](https://www.theguardian.com/membership/live/2016/oct/29/weekend-reading-heathrow-aggro-tabloids-revenge-and-good-scares/email)

#### before
picture 322

#### after
picture 323


katebee (approved):
:+1:

PR #14842, opened 2016-11-01
remove Reqwest from Weekend Reading JS

katebee:
## What does this change?

- Removes the (unnecessary) inclusion of reqwest in the Weekend Reading AB JS.
- Use `toLowerCase()` for contentType checking 

- JS stops silently failing

Nope

## Request for comment

@joelochlann


janua:
:+1:

PR #14850, opened 2016-11-01
Support Vimeo elements in AMP

SiAdcock:
## What does this change?

Adds support for Vimeo embeds in AMP 🎥 ⚡️ 

Vimeo and YouTube embeds are constructed in a very similar way so I have built one consolidated cleaner for both embed types. 

Articles with Vimeo embeds will now be indexable by AMP search platforms. Users will be able to view articles containing Vimeo embeds through AMP, providing a faster mobile experience for these articles.

Yep 😄 

## Screenshots

picture 122

## Request for comment

@guardian/dotcom-platform

### DIFF
--- a/common/app/views/support/cleaner/AmpEmbedCleaner.scala
+++ b/common/app/views/support/cleaner/AmpEmbedCleaner.scala
@@ -45,7 +45,7 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
 
   object AmpExternalVideo {
     def getAmpExternalVideoByUrl(videoUrl: String) : Option[AmpExternalVideo] = {
-      val youtubePattern = """^https?:\/\/www\.youtube\.com\/watch\?v=([^#\^&\^?]+).*""".r
+      val youtubePattern = """^https?:\/\/www\.youtube\.com\/watch\?v=([^#&?]+).*""".r
       val vimeoPattern = """^https?:\/\/vimeo\.com\/(\d+).*""".r
       videoUrl match {
         case youtubePattern(videoId) => Some(YoutubeExternalVideo(videoId))
--- a/common/app/views/support/cleaner/AmpEmbedCleaner.scala
+++ b/common/app/views/support/cleaner/AmpEmbedCleaner.scala
@@ -39,15 +39,17 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
     })
   }
 
-  class AmpExternalVideo(val videoId: String, val elementType: String) {}
+  sealed abstract class AmpExternalVideo(val videoId: String, val elementType: String)
+  case class YoutubeExternalVideo(override val videoId: String) extends AmpExternalVideo(videoId, "amp-youtube")
+  case class VimeoExternalVideo(override val videoId: String) extends AmpExternalVideo(videoId, "amp-vimeo")
 
   object AmpExternalVideo {
     def getAmpExternalVideoByUrl(videoUrl: String) : Option[AmpExternalVideo] = {
-      val youtubePattern = ".*https?://www.youtube.com/watch\\?v=(.+).*".r
-      val vimeoPattern = ".*https?://vimeo.com/(\\d+).*".r
+      val youtubePattern = "^https?:\\/\\/www\\.youtube\\.com\\/watch\\?v=(.+).*".r
+      val vimeoPattern = "^https?:\\/\\/vimeo\\.com\\/(\\d+).*".r
       videoUrl match {
-        case youtubePattern(videoId) => Some(new AmpExternalVideo(videoId, "amp-youtube"))
-        case vimeoPattern(videoId) => Some(new AmpExternalVideo(videoId, "amp-vimeo"))
+        case youtubePattern(videoId) => Some(YoutubeExternalVideo(videoId))
+        case vimeoPattern(videoId) => Some(VimeoExternalVideo(videoId))
         case _ => None
       }
     }
--- a/common/app/views/support/cleaner/AmpEmbedCleaner.scala
+++ b/common/app/views/support/cleaner/AmpEmbedCleaner.scala
@@ -39,9 +39,19 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
     })
   }
 
-  object AmpYouTube {
-    def createElement(document: Document, videoId: String): Element = {
-      val video = document.createElement("amp-youtube")
+  object AmpExternalVideo {
+    def getElementTypeByUrl(videoUrl: String) : String = {
+      val youtubePattern = ".*youtube.com.*".r
+      val vimeoPattern = ".*vimeo.com.*".r
+      videoUrl match {
+        case youtubePattern() => "amp-youtube"
+        case vimeoPattern() => "amp-vimeo"
+        case _ => ""
+      }
+    }
+
+    def createElement(document: Document, videoId: String, elementType: String): Element = {
+      val video = document.createElement(elementType)
       video.attr("data-videoid", videoId)
       video.attr("width", "5")
       video.attr("height", "3")
@@ -50,37 +60,9 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
 
     def getVideoIdFromUrl(videoUrl: String): Option[String] = {
       val youtubePattern = ".*youtube.com/watch\\?v=(.+).*".r
-      videoUrl match {
-        case youtubePattern(videoId) => Some(videoId)
-        case _ => None
-      }
-    }
-
-    def clean(document: Document) = {
-      for {
-        videoElement <- document.getElementsByClass("element-video")
-        iframeElement <- videoElement.getElementsByTag("iframe")
-        videoId <- getVideoIdFromUrl(videoElement.attr("data-canonical-url"))
-      } yield {
-        val ampVideoElement = createElement(document, videoId)
-        videoElement.appendChild(ampVideoElement)
-        iframeElement.remove()
-      }
-    }
-  }
-
-  object AmpVimeo {
-    def createElement(document: Document, videoId: String): Element = {
-      val video = document.createElement("amp-vimeo")
-      video.attr("data-videoid", videoId)
-      video.attr("width", "5")
-      video.attr("height", "3")
-      video.attr("layout", "responsive")
-    }
-
-    def getVideoIdFromUrl(videoUrl: String): Option[String] = {
       val vimeoPattern = ".*vimeo.com/(\\d+).*".r
       videoUrl match {
+        case youtubePattern(videoId) => Some(videoId)
         case vimeoPattern(videoId) => Some(videoId)
         case _ => None
       }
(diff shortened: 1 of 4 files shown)


common/app/views/support/cleaner/AmpEmbedCleaner.scala
```diff
@@ -39,25 +39,47 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
...
+      video.attr("layout", "responsive")
+    }
+
+    def getVideoIdFromUrl(videoUrl: String): Option[String] = {
+      val youtubePattern = ".*youtube.com/watch\\?v=(.+).*".r
+      val vimeoPattern = ".*vimeo.com/(\\d+).*".r
+      videoUrl match {
+        case youtubePattern(videoId) => Some(videoId)
+        case vimeoPattern(videoId) => Some(videoId)
+        case _ => None
```
jfsoul: Can we combine this with the `getElementTypeByUrl` regex?  I think we should be able to do one regex and return something that encapsulates both the element name and the video ID.

tools/amp-validation/endpoints.js
```diff
@@ -27,7 +27,7 @@ module.exports = (() => {
         '/sport/2016/aug/25/katie-ledecky-first-pitch-washington-nationals-bryce-harper', // Instagram + Twitter
         '/sport/live/2016/aug/25/county-cricket-surrey-lancashire-live-blog', // Twitter
         //'/sport/blog/2016/aug/25/baseball-olympic-games-2020-tokyo-mlb-team-selection', // Vine
-        //'/stage/2016/aug/24/they-drink-it-in-the-congo-review-adam-brace', // Vimeo
+        '/stage/2016/aug/24/they-drink-it-in-the-congo-review-adam-brace', // Vimeo
```
jfsoul: 👏

common/test/views/support/cleaner/AmpEmbedCleanerTest.scala
```diff
@@ -16,6 +16,61 @@ class AmpEmbedCleanerTest extends FlatSpec with Matchers {
...
+    val youtubeUrl = "https://www.youtube.com/watch?v=foo_12-34"
+    val doc = s"""<html><body><figure class="element-video" data-canonical-url="$youtubeUrl"><iframe></iframe></figure></body></html>"""
+    val document: Document = Jsoup.parse(doc)
+    val result: Document = clean(document)
+
+    result.getElementsByTag("amp-youtube").size should be(1)
+  }
+
+  "AmpEmbedCleaner" should "not create an amp-youtube element if videoid missing" in {
+    val youtubeUrl = s"http://www.youtube.com/"
```
jfsoul: Looks like we have `http` and `https` for youtube in these tests, which makes me think: maybe we should only be accepting `https` video links?

common/test/views/support/cleaner/AmpEmbedCleanerTest.scala
```diff
@@ -16,6 +16,61 @@ class AmpEmbedCleanerTest extends FlatSpec with Matchers {
...
+  "AmpEmbedCleaner" should "be able to create an amp-youtube element and an amp-vimeo in the same document" in {
+    val youtubeUrl = "https://www.youtube.com/watch?v=foo_12-34"
+    val vimeoUrl = "https://vimeo.com/1234"
+    val doc = s"""<html><body><figure class="element-video" data-canonical-url="$youtubeUrl"><iframe></iframe></figure><figure class="element-video" data-canonical-url="$vimeoUrl"><iframe></iframe></figure></body></html>"""
+    val document: Document = Jsoup.parse(doc)
+    val result: Document = clean(document)
+
+    result.getElementsByTag("amp-youtube").size should be(1)
+    result.getElementsByTag("amp-vimeo").size should be(1)
+  }
```
jfsoul: everybody loves tests! 💯 

could add some tests for edge cases with the regex if you felt it was required (might not be).  E.g. `youtubercom.com`, `youtube.com.ru`, `fakeyoutube.com` etc.

common/app/views/support/cleaner/AmpEmbedCleaner.scala
TBonnin: Are they referring to the same element, if not I will keep them separate?

common/test/views/support/cleaner/AmpEmbedCleanerTest.scala
SiAdcock: It doesn't look like it matters. Even if the canonical URL is `http`, `amp-youtube` and `amp-vimeo` still request the `https` version of the video.

picture 123

SiAdcock: I'll create specific `http` and `https` tests since either protocol can be entered Composer.

common/app/views/support/cleaner/AmpEmbedCleaner.scala
SiAdcock: They are testing the same element. I'll try to combine them and return an instance of a class that contains the `videoId` and the `elementType`

SiAdcock:
Thanks for your suggestions, @jfsoul and @TBonnin 😄  Would you care to take one more look?

common/app/views/support/cleaner/AmpEmbedCleaner.scala
```diff
@@ -39,14 +39,16 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
...
 
+  class AmpExternalVideo(val videoId: String, val elementType: String) {}
+
   object AmpExternalVideo {
-    def getElementTypeByUrl(videoUrl: String) : String = {
-      val youtubePattern = ".*youtube.com.*".r
-      val vimeoPattern = ".*vimeo.com.*".r
+    def getAmpExternalVideoByUrl(videoUrl: String) : Option[AmpExternalVideo] = {
+      val youtubePattern = ".*https?://www.youtube.com/watch\\?v=(.+).*".r
+      val vimeoPattern = ".*https?://vimeo.com/(\\d+).*".r
```
jfsoul: Not sure this regex is quite right

- Probably start with `^` instead of `.*`
- I think there are a few `/` and `.` that need escaping
- I think the query param ? should be `\?`, it looks like it's escaped twice
- Can we capture the youtube video id with `.*v=([\d\w]+)&?.*` - otherwise we might capture other query params before or after video id.

I might be off mark on some of this as I don't have good knowledge of youtube/vimeo urls and their edge cases.  Let me know what you think.

common/app/views/support/cleaner/AmpEmbedCleaner.scala
```diff
@@ -39,14 +39,16 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
     })
   }
 
+  class AmpExternalVideo(val videoId: String, val elementType: String) {}
```
jfsoul: case class AmpExternalVideo(val videoId: String, val elementType: String)

common/app/views/support/cleaner/AmpEmbedCleaner.scala
```diff
@@ -39,14 +39,16 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
...
+    def getAmpExternalVideoByUrl(videoUrl: String) : Option[AmpExternalVideo] = {
+      val youtubePattern = ".*https?://www.youtube.com/watch\\?v=(.+).*".r
+      val vimeoPattern = ".*https?://vimeo.com/(\\d+).*".r
       videoUrl match {
-        case youtubePattern() => "amp-youtube"
-        case vimeoPattern() => "amp-vimeo"
-        case _ => ""
+        case youtubePattern(videoId) => Some(new AmpExternalVideo(videoId, "amp-youtube"))
+        case vimeoPattern(videoId) => Some(new AmpExternalVideo(videoId, "amp-vimeo"))
+        case _ => None
```
jfsoul: With the case class you can avoid the `new` here.

You could go further and introduce `case class YoutubeExternalVideo(videoId: String) extends AmpExternalVideo("amp-youtube")` or something which would require some other changes... but I think your simple solution is better in this case.

common/test/views/support/cleaner/AmpEmbedCleanerTest.scala
```diff
@@ -16,6 +16,79 @@ class AmpEmbedCleanerTest extends FlatSpec with Matchers {
     document
   }
 
+  private def documentWithVideos(videoUrls: Seq[String]): Document = {
+    val doc = "<html><body>" +
+      videoUrls.foldLeft(""){ (res: String, next: String) => res + s"""<figure class="element-video" data-canonical-url="$next"><iframe></iframe></figure>""" } +
```
TBonnin: Rather than using `foldLeft` you can simple `map` your `Seq` of urls into a `Seq` of html strings and then use `mkString` to join them

common/test/views/support/cleaner/AmpEmbedCleanerTest.scala
```diff
@@ -16,6 +16,79 @@ class AmpEmbedCleanerTest extends FlatSpec with Matchers {
     document
   }
 
+  private def documentWithVideos(videoUrls: Seq[String]): Document = {
```
TBonnin: Nitpicking: you can use `varargs` so you don't have to pass a `Seq` but multiple `Strings`
ie:
```
private def documentWithVideos(videoUrls: String*): Document
....
documentWithVideos("https://vimeo.com/")
documentWithVideos("https://www.youtube.com/watch?v=foo_12-34","https://vimeo.com/1234")

```

TBonnin (changes requested):

common/app/views/support/cleaner/AmpEmbedCleaner.scala
johnduffell: use triple double quotes to stop scala compiler looking inside strings, then you don't have to escape for the compiler's benefit
```scala
"""my \?lovely "string" of joy"""
```

SiAdcock: Cool, that's a top tip! Thanks @johnduffell and @jfsoul 👍 💎 ❤️

jfsoul (approved):
Great work! Thanks for the changes 💯

TBonnin (approved):

common/app/views/support/cleaner/AmpEmbedCleaner.scala
```diff
@@ -39,25 +39,40 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
...
-        youtube.attr("height", "3")
-        youtube.attr("layout", "responsive")
-        element.appendChild(youtube)
+  sealed abstract class AmpExternalVideo(val videoId: String, val elementType: String)
+  case class YoutubeExternalVideo(override val videoId: String) extends AmpExternalVideo(videoId, "amp-youtube")
+  case class VimeoExternalVideo(override val videoId: String) extends AmpExternalVideo(videoId, "amp-vimeo")
+
+  object AmpExternalVideo {
+    def getAmpExternalVideoByUrl(videoUrl: String) : Option[AmpExternalVideo] = {
+      val youtubePattern = """^https?:\/\/www\.youtube\.com\/watch\?v=([^#\^&\^?]+).*""".r
```
johnduffell: parser combinators are the answer! https://github.com/sirthias/parboiled2

PR #14851, opened 2016-11-01
Handle null value in headings for the minute

gtrufitt:
## What does this change?

Handles null values and prevents 500s on minute articles.

@TBonnin @jfsoul

### DIFF
--- a/common/app/views/support/MinuteCleaner.scala
+++ b/common/app/views/support/MinuteCleaner.scala
@@ -56,7 +56,7 @@ case class MinuteCleaner(article: model.Article) extends HtmlCleaner {
         if (heading.text() == "Summary" || heading.text() == "Key event") {
           heading.remove()
         } else {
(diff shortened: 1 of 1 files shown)


common/app/views/support/MinuteCleaner.scala
```diff
@@ -101,5 +101,7 @@ case class MinuteCleaner(article: model.Article) extends HtmlCleaner {
 }
 
 object regexCleaner {
-  def apply(heading: Element, regEx: String, htmlToReplace: String): String = heading.html().replaceFirst(regEx, htmlToReplace)
+  def apply(heading: Option[Element], regEx: String, htmlToReplace: String): String = heading.map { heading =>
+    heading.html().replaceFirst(regEx, htmlToReplace)
```
jfsoul: Is this clearer?
    
    heading
      .map(_.html)
      .getOrElse("")
      .replaceFirst(regEx, htmlToReplace)

gtrufitt: yes

jfsoul (approved):
lgtm apart from shortening the map statement

PR #14852, opened 2016-11-02
Update canResonablyAskForMoney to reflect the new line from central production

jranks123:
## What does this change?
Currently, if an article is marked as isSensititve, we do not show any components that ask our readers for money in any way. However, on many of these articles, the shouldHideAdverts tag is marked as false. This is because the isSensitive tag is used very freely, especially on articles about foreign countries. The shouldHideAdverts tag is applied much more carefully, only on highly sensitive articles where it would be inappropriate to show adverts. 

We think the shouldHideAdverts tag is more suitable for our ads, and we asked Central Prod and they agree. So this change reflects the new line from Central Production on when it is ok to ask our readers for money. 

We can show ads on more articles, whilst still not showing ads on clearly inappropriate articles. 

No

## Request for comment

… on when we can ask for money

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/commercial-features.js
+++ b/static/src/javascripts/projects/common/modules/commercial/commercial-features.js
@@ -124,7 +124,7 @@ define([
             switches.commercial;
 
         this.canReasonablyAskForMoney = // eg become a supporter, give a contribution
-            !(userFeatures.isPayingMember() || config.page.isSensitive || config.page.isAdvertisementFeature);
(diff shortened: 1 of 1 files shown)


dominickendrick:
👍

jranks123:
@rtyley

rtyley:
Apparently: 

    Almost all of our foreign coverage is marked as sensitive, plus other contentious issues
    such as race, religion and abortion etc. In addition, we usually turn off related content
    for legally sensitive stories or stories about child abuse or similar subjects. We would
    only block ads in the most extreme cases.

PR #14854, opened 2016-11-02
Use secure origin in youtube embed URL

kelvin-chappell:
This is another attempt at #14792 
This time assuming it will work everywhere if the origin is secure.

/cc @lps88


lps88:
👍  let's triple check it this time 😆

kelvin-chappell:
Agreed!

PR #14855, opened 2016-11-02
Hosted articles - inline video resize fix

Calanthe:
## What does this change?
There is an issue on hosted article pages when inline video elements resize themselves a few times after page load. This is caused by the styles for the `.vjs_video_3-dimensions` class, that are included by the `video.js` library https://github.com/guardian/frontend/blob/master/static/src/javascripts/components/video.js/video.js#L9321. 
screen shot 2016-11-02 at 12 37 40
Until the proper styles are loaded and parsed, those big dimensions set by the `video.js` library are applied.

I assume that those dimensions are set based on the actual video sizes. I don't know how to change aforementioned dimensions without touching Scala and/or Composer.

This is my small attempt to fix or at least slightly improve the way those video elements resize themselves.

The video elements still resize but the default dimensions are smaller so the jumping effect is less irritating.

no

## Screenshots
Before:
video-resize-bug

After:
jumping-video-fix

Please also notice that the page loading time will be much faster on PROD, because on localhost we don't have minification working.

## Request for comment
@lps88 @regiskuckaertz @akash1810


regiskuckaertz (approved):
👍

akash1810:
Nice 👍. The after GIF appears to have the default html video controls rather than videojs ones. Is this correct?

I wonder if a better, lighter, solution would be to render the poster image with a click handler which instantiates a video player?

Calanthe:
@akash1810 The GIF shows the default video controls only until the proper styles are loaded. It takes a while on localhost because styles and js files are not minified.

Do you know if there is a way to make the video smaller on the server side? So it will be loaded into the article in proper dimensions. Then in theory my fix won't be needed.

akash1810:
> Do you know if there is a way to make the video smaller on the server side?

Videos are encoded to specific dimensions that provides a compromise between quality and file size. Changing this globally or for a single video is non-trivial sadly.

Calanthe:
@akash1810 Thanks for that information!

PR #14856, opened 2016-11-02
Do not load sonobi on page skins

rich-nguyen:
The sonobi library doesn't support `enableSingleRequest`, so we should not run sonobi on pageskin pages.


rich-nguyen:
[Swisse campaign live today - Life and Style - skin and MPUs not serving](https://trello.com/c/WzikxMHq/128-swisse-campaign-live-today-life-and-style-skin-and-mpus-not-serving)

JonNorman (approved):

PR #13421, opened 2016-06-27
Twitter image overlay

katebee:
## What does this change?
- Adds a switch for applying the branded image overlay on twitter, as with the Open Graph (Facebook) overlay
- Removes the extra padding between the text and bottom of the image on the Open Graph image
- Adds a branded option for the TwitterImage profile, with the padding to account for the slight cropping difference between the twitter app and webview.
## What is the value of this and can you measure success?
- Many people will retweet a tweet with a link, without even clicking on that link
- Success will be increase in the Guardian brand on twitter, hopefully building trust and encouraging visits to the website
- Next time Snowden tweets a link to our article, we will be ready.
## Does this affect other platforms - Amp, Apps, etc?

Affects how links from the Guardian appear on twitter - the image will include a banner across the bottom.
##### Twitter image

twitter-image
##### OpenGraph image

opengraph-image
## Request for comment

@dominickendrick

PR #14860, opened 2016-11-02
Make the Twitter and Facebook overlays immortal

katebee:
## What does this change?

- Makes the expiring Overlay feature switches live on

- Switches do not expire next week
- Keeps the overlays on Twitter Cards and Open Graph image shares
- Keeps doing what we were doing

- Twitter & Facebook - keeps the banner on our images for all shares
- Overlay will also still appear in Slack, WhatsApp, Telegram and any other apps that use the Open Graph protocol http://ogp.me/

## Screenshots

Twitter overlay images here: #13421
Facebook overlay images here: #13026

#### Bonus: Slack uses Open Graph 🎉 

picture 348

## Request for comment

@dominickendrick

### DIFF
--- a/common/app/conf/switches/FeatureSwitches.scala
+++ b/common/app/conf/switches/FeatureSwitches.scala
@@ -41,7 +41,7 @@ trait FeatureSwitches {
     "If this switch is turned on, we will overlay the guardian logo along the bottom of images shared on facebook",
     owners = Seq(Owner.withGithub("dominickendrick")),
     safeState = On,
-    sellByDate = new LocalDate(2016, 11, 7),
+    sellByDate = never,
     exposeClientSide = false
   )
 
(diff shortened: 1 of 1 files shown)


dominickendrick:
👍 ⏩

SiAdcock:
@katebee Can I ask why we are making the switches immortal, rather than just deleting them altogether? Is there any benefit in being able to quickly turn these overlays on and off? Is this part of a test that may need to be backed out at some point?

katebee:
@siAdcock: Of course!
I think it is worth keeping these as switches, since it is possible Facebook or Twitter could make a drastic change to their UI, or image dimensions, which could result in a partial or ugly overlay.
E.g. Facebook switching to 5:3 image ratios (currently 19:10)

With these other platforms outside of our control, and caching or content, the ability to quickly disable this until we can correct is beneficial.

However, I would love more insight into any caveats or costs associated with feature switches.

SiAdcock:
@katebee nice one Kate. I agree, keeping this as a switch does make sense.

The cost of switches is more of a maintenance cost associated with maintaining redundant code.  We have over 120 feature switches in `frontend` and over half of them have been flagged as no longer necessary. 

Switches like this one are important and we strongly encourage their use. But we also need to be mindful of why we are using them, and whether we can tidy them away to reduce code clutter and to prevent our codebase from becoming a giant switchboard!

PR #14862, opened 2016-11-02
Video content page with media atom

gidsg:
## What does this change?
Render video content page with media atom

Required for YouTube PFP trial

No

## Screenshots
screen shot 2016-11-02 at 17 19 20

## Request for comment
CC @markjamesbutler @TBonnin


applications/app/views/fragments/mediaBody.scala.html
```diff
@@ -65,6 +66,14 @@
                                             <meta itemprop="datePublished" content="@video.trail.webPublicationDate.toString("yyyy-MM-dd")" />
                                             <meta itemprop="name" content="@video.metadata.webTitle" />
                                             <meta itemprop="uploadDate" content="@video.trail.webPublicationDate.toString("yyyy-MM-dd")" />
+
+                                            @video.mediaAtom.map { ma =>
+                                                @ma.duration.map { duration =>
+                                                    <meta itemprop="duration" content="@MediaAtom.isoDuration(duration)" />
+                                                }
+                                            }
+                                            @* TODO: Set meta thumbnail once we have better image data in model *@
```
gidsg: Right now the posterImage is only an `Option[String]` so it would not be reliable to run the standard frontend image processing on it. I have raised this with Tools team. Also the media atom model does not provide height and width so I don't have that for the metadata either. CC @clloyd

applications/app/views/fragments/mediaBody.scala.html
```diff
@@ -76,6 +85,10 @@
                                                 <meta itemprop="thumbnail" content="@SeoOptimisedContentImage.bestFor(img.images)" />
                                                 <meta itemprop="thumbnailUrl" content="@SeoOptimisedContentImage.bestFor(img.images)" />
                                             }
+                                            @video.mediaAtom.map { mediaAtom =>
+                                              @fragments.atoms.media(mediaAtom)
+                                            }
+
```
gidsg: If a video page were to include both a Video Element and a Media Atom both would currently be displayed. Though this shouldn't happen if people are concerned about this I can use an if statement to only display one or the other.

common/app/model/content.scala
```diff
@@ -620,13 +619,13 @@ object Video {
...
 
-    Video(contentOverrides, source)
+    Video(contentOverrides, source, content.media.headOption)
   }
 }
 
 final case class Video (
   override val content: Content,
-  source: Option[String] ) extends ContentType {
+  source: Option[String], mediaAtom: Option[MediaAtom] ) extends ContentType {
```
TBonnin: Let's follow the same one param per line style here.

common/app/model/content/Atoms.scala
```diff
@@ -120,6 +121,10 @@ object MediaAtom extends common.Logging {
       mimeType = mediaAsset.mimeType)
   }
 
+  def isoDuration(seconds: Long): String = {
```
TBonnin: I would add this method in the `MediaAtom` class rather than here in the object.
You will be able to do:
```
myMediaAtom.isoDuration
```
vs
```
MediaAtom.isoDuration(myMediaAtom.duration)

```

applications/app/views/fragments/mediaBody.scala.html
TBonnin: Got it. However what guarantee do I have this TODO will not still be here in 6months? 😀

TBonnin (changes requested):

common/app/model/content/Atoms.scala
```diff
@@ -27,6 +27,7 @@ final case class MediaAtom(
   assets: Seq[MediaAsset],
   title: String,
   duration: Option[Long],
+  isoDuration: Option[String],
```
TBonnin: I would actually make `isoDuration` a `def` and also move the logic in this class.

TBonnin (changes requested):

TBonnin (approved):

gidsg:
@TBonnin 
> Got it. However what guarantee do I have this TODO will not still be here in 6months? 

None!

...but I thought it was better than the alternative of this information being in my head. For now I have changed it to use the `signedArticleImage` which seems a reasonable fallback in case the work never gets done on `posterUrl`.

PR #14864, opened 2016-11-02
Use scanamo for redirects, centralise

jfsoul:
## What does this change?
Consolidates logic for inserting, removing and reading from the `redirect` dynamo tables.  There is now a single service which uses Scanamo to interact with dynamo.  

There are still a number of other dynamo interactions in frontend that still use the AWS SDK directly - I haven't switched these yet as there isn't as strong a case for it (redirects made a bare request with `wsClient`).  For these other interactions, though, I have at least moved them to use a single `DynamoDB` service for initialising the client.

This is largely an exercise in centralising logic:
- Redirect table(s) interactions happen in one place
- Dynamo client is initialised in one place (and only once)

Bonus: net reduction in LOC! 🔪 

No

## Request for comment
N.B. the diff isn't particularly helpful: `DynamoDB.scala` was renamed to `Redirects.scala`.

@TBonnin 

cc: @philwills: wondering whether there might be a better way to interact with the `redirect` table via scanamo.  Particularly thinking about the `Redirects.Redirect` case class with `Option`s.


admin/app/css_report/CssReport.scala
```diff
@@ -22,8 +22,7 @@ case class SelectorReport(selector: String, used: Int, unused: Int)
 object CssReport extends ExecutionContexts {
   private val DateFormat = "yyyy-MM-dd"
   private val TableName = "cssUsage"
-  private val dynamoDbClient = new AmazonDynamoDBAsyncClient()
-  dynamoDbClient.setRegion(Region.getRegion(Regions.EU_WEST_1))
```
jfsoul: We were doing this in a number of places - but the client is thread safe.

admin/app/jobs/R2PagePressJob.scala
```diff
@@ -154,26 +151,20 @@ class R2PagePressJob(wsClient: WSClient) extends ExecutionContexts with Logging
...
-                  case true => {
-                    PagePresses.set(urlIn, pressUrl)
+                  case true =>
+                    redirects.set(Redirects.Redirect(urlIn, archive = Some(pressUrl)))
                     log.info(s"Pressed $urlIn as $pressUrl")
                     queue.delete(notification.handle)
-                  }
-                  case _ => {
-                    log.error(s"Press failed for $pressUrl")
-                  }
```
jfsoul: Removed warnings for redundant braces in a few places here

common/app/services/DynamoDB.scala
```diff
@@ -1,64 +1,24 @@
...
-        (json \\ "destination").headOption.map(d => Redirect((d \ "S").as[String]))
-          .orElse((json \\ "archive").headOption.map(a => Archive((a \ "S").as[String])))
-      }
-    }.getOrElse(Future.successful(None))
+  private def createClient[T <: AmazonWebServiceClient](serviceClass: Class[T]): T = {
+    val client = serviceClass
+      .getConstructor(classOf[AWSCredentials])
+      .newInstance(credentials)
+    client.setRegion(region)
+    client
```
jfsoul: This might be overkill just for the 2 client types, but definitely DRY...

archive/test/ArchiveControllerTest.scala
```diff
@@ -6,16 +6,18 @@ import play.api.test.Helpers._
...
   with ConfiguredTestSuite
-  with BeforeAndAfterAll
-  with WithTestWsClient {
+  with BeforeAndAfterAll {
 
-  lazy val archiveController = new ArchiveController(new DynamoDB(wsClient))
+  lazy val archiveController = new ArchiveController(mockRedirects)
+  lazy val mockRedirects = new Redirects {
+    override def destinationFor(source: String) = Future.successful(None)
+  }
```
jfsoul: Planning to create a `WithTestDynamoClient` trait or similar which will mean this mocking isn't required, but I'll do that in another PR.  This one is already too bloated.

TBonnin (changes requested):

common/app/services/Redirects.scala
```diff
@@ -0,0 +1,54 @@
...
+
+import com.gu.scanamo.{Scanamo, ScanamoAsync}
+import com.gu.scanamo.syntax._
+import common.{ExecutionContexts, Logging}
+import conf.Configuration
+
+import scala.concurrent.Future
+
+
+object Redirects {
```
jfsoul: I think so yes, the only reason I didn't was consistency - we don't tend to use that naming convention (in `common` services anyway)

admin/app/controllers/RedirectController.scala
```diff
@@ -24,8 +24,8 @@ class RedirectController  extends Controller with Logging {
   def redirectPost() = Action { implicit request =>
 
     redirectForm.bindFromRequest().get.trim match {
-      case PageRedirect(from, "") if from.nonEmpty  => Redirects.remove(from)
-      case PageRedirect(from, to) if from.nonEmpty  => Redirects.set(from, to)
+      case PageRedirect(from, "") if from.nonEmpty  => redirects.remove(from)
```
jfsoul: I'll include this as part of next PR on the redirect tool UI

jfsoul:
Thanks for comments @TBonnin (naming is hard 😢 ).  Agree in all cases and have made changes.

@philwills thanks for the help - using `Destination` directly now - although there might be a neater way.

TBonnin (approved):

philwills:
@jfsoul That looks like a reasonable way of doing it to me. You could probably do something slightly marginally neater, but I don't think it would help with clarity much.

PR #14857, opened 2016-11-02
Remove prominent membership engagement banner AB test

JustinPinner:
## What does this change?
Remove the prominent membership engagement banner AB test as implemented across the following PRs;
https://github.com/guardian/frontend/pull/14746
https://github.com/guardian/frontend/pull/14765
https://github.com/guardian/frontend/pull/14825
 
## What is the value of this and can you measure success?
Preparing for when the test is complete.

## Does this affect other platforms - Amp, Apps, etc?
No. Web only.

## Screenshots
Nothing to see, reverts to default appearance.

## Request for comment
@svillafe @Ap0c @rtyley @rupertbates cc @NataliaLKB

PR #14868, opened 2016-11-03
Prominent membership engagement banner as a feature

JustinPinner:
## What does this change?
The UK prominent membership engagement banner test returned very promising results. That test is being decommissioned (see https://github.com/guardian/frontend/pull/14857) and will now become a feature.

This PR is a bit clunky because I've tried to structure it such that when the test removal PR is merged, the number of conflicts with this branch are minimal. I'm hoping the tricky bits will only be limited to the `membership-message.html` file.

This is behind a membership feature switch, which defaults to OFF, and has no expiry date. This allows us to deploy now and switch it on when we're ready.

Following up from the AB test we ran over the last week or so, we're confident the revised styling of the engagement banner will have an ongoing positive effect on new member sign-ups.

The feature introduces a new tracking code so we can monitor it after deployment and activation, to make sure it performs well enough to justify its ongoing existence.

No, web only.

## Screenshots
See original PR (https://github.com/guardian/frontend/pull/14746)

## Request for comment
@Ap0c @svillafe @rtyley @rupertbates 
cc @stephanfowler


Ap0c (approved):
Looks like it's working as expected, when the switch is on it shows the new prominent banner for the UK only, if it's off it goes back to the old banner 👍.

PR #14870, opened 2016-11-03
remove grunt concat

sndrs:
## What does this change?

removes grunt concat

removes another grunt plugin

no

## Screenshots

`static/target`, `static/hash` and `common/conf/assets` are identical to master, except for an extra `;` between modules in `app.js`

screen shot 2016-11-03 at 13 23 35

## Request for comment

@guardian/dotcom-platform


SiAdcock:
👍 ❤️

PR #14871, opened 2016-11-03
Add user agent and full url to GA

gtrufitt:
## What does this change?

Adds the user agent string and URL to the GA page hit.

- user-agent: Full string is required by ABC auditing and is not not available from the standard page view hits in GA
- URL: Full string is required as GA "automatically strips out all 'utm' parameters before it populates the page report (but leaves in any native non google query params like own own CMP parameters)." via @mkopka 

Yes, I have added the url but I can't add the user-agent as there's some weirdness happening with the `amp-analytics` side of things that I've queried with them so will add that later.

@guardian/dotcom-platform

### DIFF
--- a/common/app/views/fragments/amp/googleAnalytics.scala.html
+++ b/common/app/views/fragments/amp/googleAnalytics.scala.html
@@ -7,7 +7,7 @@
     <script type="application/json">
             {
               "requests": {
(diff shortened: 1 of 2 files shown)


mkopka:
big thanks

NataliaLKB (approved):
lgtm 👍 . Thanks for querying the weirdness with amp-analytics!

PR #14872, opened 2016-11-03
Extend UK message copy experiment switch beyond POTUS election

JustinPinner:
## What does this change?
The UK membership engagement copy test finishes on Monday night, and its master switch is due to expire on Tuesday night. 

We don't want to deploy anything from Monday aftenoon through to Wednesday morning due to the US election. This PR extends the switch expiry date to Thursday (night). 

The switch `safeState` now defaults to `On` so that we don't switch the test off as the change is deployed.

The test will still end as planned on Monday night, but builds won't break due to an expired switch.

Builds won't break. Or at least not because of this switch.

No, web only.

## Screenshots
N/A

## Request for comment
@Ap0c @svillafe @rtyley @rupertbates


Ap0c (approved):
LGTM

PR #14874, opened 2016-11-03
Fix for cannot dismiss bug

MatthewJWalls:
## What does this change?

Fix for cannot dismiss bug
    
1. The cannot dismiss message from breaking news is caused
    by an unhandled promise rejection bubbling up to the top
    level when local storage is not available - we now ignore
    this error.
    
2. Improvement to the fetch error messages.

Fixes another sentry crash

Nope

## Screenshots

N/A

## Request for comment

1. The cannot dismiss message from breaking news is caused
   by an unhandled promise rejection bubbling up to the top
   level when local storage is not available - we now ignore
   this error.

2. Improvement to the fetch error messages.


michaelwmcnamara:
+1

PR #14875, opened 2016-11-03
Fix membership link bug

jranks123:
## What does this change?
Apparently, hypens are ok for contributions but not for membership. the codes were being cut off at the hypen. This bug fixes the campaign codes for the membership links

We can measure the success of our test


akash1810:
don't have much context/knowledge as to why this works, but campaign codes are used for tracking so this seems harmless enough for me to 👍

PR #14876, opened 2016-11-03
Fix to fix for breaking news

MatthewJWalls:
## What does this change?

typo, weirdly only manifested on code checking


michaelwmcnamara:
+1

PR #14877, opened 2016-11-03
Do not stringify circular structures in SpaceFinder

rich-nguyen:
There are a high number of errors coming from space finder in Sentry. Removing the `stringify`, don't think it's super useful to have the full object.


janua:
:+1:

rich-nguyen:
[Investigate sentry error from commercial](https://trello.com/c/lnGcFFxX/135-investigate-sentry-error-from-commercial)

PR #14878, opened 2016-11-03
Hosted video on iOS fixed

Calanthe:
## What does this change?
This PR fixes the bug when hosted video instead of playing, was redirecting the browser to the next video from the series (on iOS, desktop bp only). That was caused by the faulty video.js library, which doesn't load video metadata properly until the whole video is loaded. Thus, the `player.duration()` value was always 0 and caused immediate redirect in here https://github.com/guardian/frontend/blob/84d69aa2edb7b5a1f89795a692f2675f4d9170e2/static/src/javascripts/projects/commercial/modules/hosted/next-video-autoplay.js#L41. This PR also fixes contrast issue with buttons on the 'next video autoplay' overlay.

iOS tablet users can play the hosted video.

no

## Screenshots
Before:
screen shot 2016-11-03 at 17 22 42

After:
screen shot 2016-11-04 at 12 12 29

## Request for comment
@lps88


lps88:
Looks good 👍

PR #14879, opened 2016-11-03
Add doc for updating test database

gidsg:
Add doc describing test database.
CC @TBonnin @jfsoul


TBonnin:
Thank you very much @gidsg 
This mechanism is in place for all external requests not just capi, we should probably make this doc more generic then. Let me know if you want me to do it.

jfsoul:
Thanks @gidsg! 👏  👍

PR #14880, opened 2016-11-04
extract some video logic into modules

akash1810:
## What does this change?
no functional change, just makes `media/main.js` slightly easier to reason about

fewer 🤕 

no

## Screenshots
n/a

## Request for comment
@gidsg @markjamesbutler


gidsg (approved):
Thanks @akash1810 makes this code a bit easier to understand 🔖 👍

PR #14882, opened 2016-11-04
Disable sticky banner when the top slot is disabled

regiskuckaertz:
Some pages don't have a top slot, like the profile pages for instance. The sticky top banner is trying to access the parent of a node that does not exist then:

```
topSlot = document.getElementById(topSlotId); // does not exist
stickyBanner = topSlot.parentNode; // ಥ_ಥ
```

To prevent this from happening, we just make sure the top slot exists.

cc @rich-nguyen


rich-nguyen:
👍  nice

PR #14887, opened 2016-11-04
adds ga tracking to off platform video player

akash1810:
## What does this change?
Adds GA tracking to off platform video player. Not sure if we want to prefix `gaEventLabel` with something to help distinguish these events as coming from the off platform video player?

GA replaces omniture.

no

## Screenshots
`cm6` is video end

screen shot 2016-11-04 at 14 17 56

## Request for comment
@gidsg

### DIFF
--- a/static/src/javascripts/bootstraps/video-embed.js
+++ b/static/src/javascripts/bootstraps/video-embed.js
@@ -83,7 +83,10 @@ define([
             var player,
                 mouseMoveIdle,
                 $el = bonzo(el).addClass('vjs'),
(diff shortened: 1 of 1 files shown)


gidsg:
> Not sure if we want to prefix gaEventLabel with something to help distinguish these events as coming from the off platform video player?

Guess this is one for @mkopka ?

akash1810:
Merging this. We weren't adding an additional identifier for Omniture, so we're not regressing on anything. We can update later if needed.

PR #14591, opened 2016-10-05
Prevent logging of crashes if adblocker on

MatthewJWalls:
## What does this change?

Prevents anybody with an adblocker enabled from sending client-side js crashes to sentry
## What is the value of this and can you measure success?

Less noise in Sentry
## Does this affect other platforms - Amp, Apps, etc?

Nope
## Screenshots

N/A
## Request for comment

N/A

PR #14888, opened 2016-11-04
Reject the promise chain if sonobi did not load

rich-nguyen:
There are still errors logged by Sentry even though they are to do with network blocking,  so this change https://github.com/guardian/frontend/pull/14591 needs further tweaks.

A filter like Ghostery does not get detected as an adblocker. This change handles the rejection without `window.onerror` @MatthewJWalls @regiskuckaertz


regiskuckaertz:
Sweet 👍

rich-nguyen:
[Reduce sonobi network entries in Sentry](https://trello.com/c/rTrgl5rf/139-reduce-sonobi-network-entries-in-sentry)

PR #14889, opened 2016-11-04
Fix membership tracking code bug

jranks123:
## What does this change?

Somehow this bug crept back into the fold. Membership doesn't like '-' in the campaign codes. This fixes it.


jfsoul (approved):
More magic that just requires implicit knowledge 😞

PR #14892, opened 2016-11-04
Remove the old stripe implementation and replace with stripe checkout.

AWare:
## What does this change?
This replaces the _quite old_ stripe implementation with a new one based on stripe checkout.

Membership and subscriptions are replacing their stripe implementations with one based on stripe checkout. (Happy to explain any details on this in person.)

No.

## Screenshots

image

## Request for comment
@SiAdcock @paulbrown1982 
NB: This should have a logo, but adding it to stripe is a one line change which I'll do with any desired changes.


static/src/javascripts/projects/membership/stripe.js
```diff
@@ -0,0 +1,128 @@
...
+        }
+
+
+        $last4.text(card.last4);
+        loading(false);
+        $type.addClass(newCardType);
+        $type.data(newCardType);
+        $parent.removeClass('is-hidden');
+        bean.off($button[0], 'click');
+        bean.on($button[0], 'click', handler($parent));
```
SiAdcock: The `handler()` function doesn't currently accept any parameters. Is this right?

static/src/javascripts/projects/membership/stripe.js
```diff
@@ -0,0 +1,128 @@
...
+                    panelLabel: 'Update',
+                    token: update(endpoint),
+                    closed: function(){
+                        loading(false);
+                    }
+
+                });
+                // Close Checkout on page navigation:
+                window.addEventListener('popstate', function() {
+                    handler.close();
```
SiAdcock: The `handler()` function doesn't have a `close` method. Should this be `checkoutHandler.close()`

static/src/javascripts/projects/membership/stripe.js
```diff
@@ -0,0 +1,128 @@
...
+
+        if (oldCardType) {
+            $type.removeClass(oldCardType);
+        }
+
+
+        $last4.text(card.last4);
+        loading(false);
+        $type.addClass(newCardType);
+        $type.data(newCardType);
```
SiAdcock: This is getting the value of a data attribute, but it's not doing anything with the value. Should this be setting?

https://github.com/ded/bonzo#api-data

SiAdcock:
Looks nice dude. I wonder if we should be using our `fastdom` wrapper for the DOM manipulations? Have a look at [this file](https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/common/modules/navigation/newHeaderNavigation.js#L28) for an idea of how you can use it

static/src/javascripts/projects/membership/stripe.js
AWare: 😱

static/src/javascripts/projects/membership/stripe.js
AWare: Yeah, I changed it to be closed in display to avoid awkwardly passing state around.

static/src/javascripts/projects/membership/stripe.js
AWare: Yes. :(

AWare:
Potentially could use fastdom, but I don't think it will help with the performance problems this page has.

static/src/javascripts/projects/membership/stripe.js
```diff
@@ -0,0 +1,139 @@
+/* global StripeCheckout */
+define([
+    'common/utils/$',
+    'bean',
+    'common/utils/ajax',
+    'common/utils/config',
+    'fastdom'
```
SiAdcock: 💯

static/src/javascripts/projects/membership/stripe.js
```diff
@@ -0,0 +1,139 @@
...
+                    description: 'Update your card details',
+                    panelLabel: 'Update',
+                    token: update(endpoint),
+                    closed: function () {
+                        loading(false);
+                    }
+
+                });
+                // Close Checkout on page navigation:
+                window.addEventListener('popstate', function () {
```
SiAdcock: Is this fired within Stripe's code?

AWare: It should be, but: https://stripe.com/docs/checkout#integration-custom
image

SiAdcock: As discussed, let's add a comment so we don't forget where this event is fired!

SiAdcock:
👍

paulbrown1982:
👍 Nice work! 🚢

Ap0c:
👍

PR #14895, opened 2016-11-04
Contributions Membership side by side test.

jranks123:
## What does this change?

Previously, we tested Contributions by itself vs Contributions with a small Membership ask and Membership with a small Contributions ask. 

We now want to test just Membership vs Membership and Contributions side by side. We also don't want to run it on articles about Brexit.

We will learn what the effect of having equally weighted Contributions and Membership asks is, against just Membership. 

## Screenshots
screenshot at nov 04 16-26-08
screenshot at nov 04 16-26-24
screenshot at nov 04 16-27-02


static/src/javascripts/projects/common/modules/experiments/ab-test-clash.js
```diff
@@ -13,8 +13,13 @@ define([
             variants: ['control', 'member-contribute']
         };
 
+        var contributionsMembershipEpicSideBySide = {
+            name: 'ContributionsMembershipEpicSideBySide',
+            variants: ['control', 'member-contribute']
```
jfsoul: does this variant name make sense for this test? (just that it's the same as the other test)

you will know better than I do...

jranks123: yes it does. Both are variants where membership and contribute are displayed. the difference is here they are both buttons rather than one buttons and one text. But member-contribute still makes sense

static/src/javascripts/projects/common/modules/experiments/tests/contributions-membership-epic-side-by-side.js
```diff
@@ -0,0 +1,133 @@
...
+        this.successMeasure = 'Impressions to number of contributions/supporter signups';
+        this.audienceCriteria = 'All users in US and UK who are not reading articles about Brexit';
+        this.dataLinkNames = '';
+        this.idealOutcome = 'One of the 2 variants proves to be a clear winner';
+        this.canRun = function () {
+            var userHasNeverContributed = !cookies.get('gu.contributions.contrib-timestamp');
+            var worksWellWithPageTemplate = (config.page.contentType === 'Article'); // may render badly on other types
+            var keywords = config.page.keywordIds.split(',');
+            var nonKeywordTagIds = config.page.nonKeywordTagIds.split(',');
+            var isNotAboutBrexit = (keywords.indexOf('politics/eu-referendum') == -1) && (nonKeywordTagIds.indexOf('tone/news') == -1);
```
jfsoul: Is this supposed to be the negation of `isAboutBrexit`? If so I think you'll want '||' instead of `&&`.

jfsoul: Also you'll want `===` instead of `==`.

static/src/javascripts/projects/common/modules/experiments/tests/contributions-membership-epic-side-by-side.js
```diff
@@ -0,0 +1,133 @@
...
+            return isNotAboutBrexit && userHasNeverContributed && commercialFeatures.canReasonablyAskForMoney && worksWellWithPageTemplate;
+        };
+
+        var membershipUrl = 'https://membership.theguardian.com/supporter?';
+        var contributeUrl = 'https://contribute.theguardian.com/?';
```
jfsoul: these hostnames might be in `guardian.page.config`

static/src/javascripts/projects/common/modules/experiments/tests/contributions-membership-epic-side-by-side.js
jranks123: Yes you are right! thanks

static/src/javascripts/projects/common/modules/experiments/tests/contributions-membership-epic-side-by-side.js
jranks123: so membership is but contribute isn't. Maybe best to keep them both like this?

jfsoul: As this is temporary, sure.  If we ever use these permanently somewhere I would advocate both being added to the config.

jfsoul (approved):

PR #14896, opened 2016-11-04
fetch end-slate on video play rather than page load

akash1810:
## What does this change?
Reduces the number of calls made on page load. The end-slate url looks like http://localhost:9000/video/end-slate/section/film.json?shortUrl=https://gu.com/p/5amt6. The shortUrl is expanded, which is slow. Removing this call on page load should mean playback starts that bit sooner.

Faster intial video playback.

No.

## Screenshots
GIF of a video I've played before, to emphasise the fact the request is done on play rather than page load.

Before
end-slate-before

After
end-slate

## Request for comment
@gidsg @markjamesbutler

### DIFF
--- a/static/src/javascripts/bootstraps/enhanced/media/main.js
+++ b/static/src/javascripts/bootstraps/enhanced/media/main.js
@@ -312,18 +312,20 @@ define([
 
     function initEndSlate(player, endSlatePath) {
         var endSlate = new Component(),
-            endState = 'vjs-has-ended';
+            endStateClass = 'vjs-has-ended';
 
         endSlate.endpoint = endSlatePath;
-        endSlate.fetch(player.el(), 'html');
 
         player.one(events.constructEventName('content:play', player), function () {
+            endSlate.fetch(player.el(), 'html');
+
             player.on('ended', function () {
-                bonzo(player.el()).addClass(endState);
+                bonzo(player.el()).addClass(endStateClass);
             });
         });
+
         player.on('playing', function () {
-            bonzo(player.el()).removeClass(endState);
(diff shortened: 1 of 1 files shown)


static/src/javascripts/bootstraps/enhanced/media/main.js
```diff
@@ -312,18 +312,20 @@ define([
 
     function initEndSlate(player, endSlatePath) {
         var endSlate = new Component(),
-            endState = 'vjs-has-ended';
+            endStateClass = 'vjs-has-ended';
```
akash1810: more explicit name

static/src/javascripts/bootstraps/enhanced/media/main.js
```diff
@@ -312,18 +312,20 @@ define([
...
 
         endSlate.endpoint = endSlatePath;
-        endSlate.fetch(player.el(), 'html');
 
         player.one(events.constructEventName('content:play', player), function () {
+            endSlate.fetch(player.el(), 'html');
```
akash1810: this is the main change

gidsg:
I think the `shortUrl` thing is maybe a bit of red herring, as there are not any additional client side redirects, but this is still a nice performance improvement for people who don't play the video. 👍 🗳

akash1810:
> not any additional client side redirects

`shortUrl` is a query string for the end-slate api - I had assumed the shortUrl was expanded server side, however looking at [the code](https://github.com/guardian/frontend/blob/master/onward/app/controllers/VideoEndSlateController.scala#L66-L73) it isn't its just used to filter out results 😄

PR #14899, opened 2016-11-04
Check test is switched on before canRun

rich-nguyen:
Urgent production fix to suppress js error on network front


static/src/javascripts/projects/common/modules/experiments/ab.js
```diff
@@ -117,7 +117,7 @@ define([
             isSensitive = config.page.shouldHideAdverts;
 
         return ((isSensitive ? test.showForSensitive : true)
-        && test.canRun() && !expired && isTestSwitchedOn(test));
```
rich-nguyen: if the `canRun` throws an exception, we need to be able to turn off the test to suppress the error.

PR #14900, opened 2016-11-04
Add guard in canRun for contributions tests

jranks123:
## What does this change?

Fixes a bug where config.page.keywordIds and config.page.nonKeywordTagIds did not exist. 

We can run the test that we planned for the weekend.


rich-nguyen:
👍

PR #14901, opened 2016-11-04
make section related calls iff section exists

akash1810:
## What does this change?
More in series and end-slate relies on page section. If the section doesn't exist then the requests 404. See console on https://www.theguardian.com/global/video/2016/nov/01/what-happened-at-the-battle-of-orgreave-video-explainer for example.

This change checks page section exists before making api calls.

Fewer 404s

No. Will very likely need to replicate on off platform embed player in a later PR.

## Screenshots
n/a

## Request for comment
@gidsg @markjamesbutler

### DIFF
--- a/common/app/model/VideoPlayer.scala
+++ b/common/app/model/VideoPlayer.scala
@@ -16,6 +16,7 @@ case class VideoPlayer(
   title: String,
   autoPlay: Boolean,
   showControlsAtStart: Boolean,
+  // TODO make Option[String] as relies on section being set
   endSlatePath: String,
   // TODO make `String` once `path` is available on main media on fronts
   path: Option[String],
--- a/static/src/javascripts/bootstraps/enhanced/media/main.js
+++ b/static/src/javascripts/bootstraps/enhanced/media/main.js
@@ -150,7 +150,6 @@ define([
         var mediaType = el.tagName.toLowerCase(),
             $el = bonzo(el).addClass('vjs'),
             mediaId = $el.attr('data-media-id'),
-            showEndSlate = $el.attr('data-show-end-slate') === 'true',
             endSlateUri = $el.attr('data-end-slate'),
             embedPath = $el.attr('data-embed-path'),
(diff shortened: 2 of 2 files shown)


common/app/model/VideoPlayer.scala
```diff
@@ -16,6 +16,7 @@ case class VideoPlayer(
   title: String,
   autoPlay: Boolean,
   showControlsAtStart: Boolean,
+  // TODO make Option[String] as relies on section being set
```
akash1810: if there isn't a section `endSlatePath` will 404 - it shouldn't be on the DOM in this case

paperboyo:
I may be entirely wrong, but maybe it should just call for latest videos from the tag instead of ending up empty? Would be https://www.theguardian.com/politics/miners-strike-1984-85+type/video for the above example.

akash1810:
@paperboyo it doesn't end up empty as the fallback behaviour is to show the most popular videos

static/src/javascripts/bootstraps/enhanced/media/main.js
```diff
@@ -399,7 +403,11 @@ define([
         mediator.on('page:media:moreinloaded', initPlayButtons);
 
         initFacia();
-        initMoreInSection();
+
+        if (config.page.section) {
```
gidsg: There are already logic checks inside `initMoreInSection()` function, so I think this extra check would make more sense to do there?

gidsg (changes requested):

static/src/javascripts/bootstraps/enhanced/media/main.js
akash1810: yeah I debated on putting it in there. My argument for putting it here was that the function name `initMoreInSection` sounds like it expects section to exist... it was a weak argument so happy to change.

akash1810: updated.

gidsg (approved):

PR #14905, opened 2016-11-07
Expose window.guardian.triggerLoadApp for US election atom

wpf500:
Short term hack as discussed with @stephanfowler, @SiAdcock and others to allow the US election interactive to delay the loading of `app.js` to reduce initial page janking.

This hack is specifically targeted at a page which contains an atom whose ID is `2016/general-election`

### DIFF
--- a/common/app/views/fragments/inlineJSBlocking.scala.html
+++ b/common/app/views/fragments/inlineJSBlocking.scala.html
@@ -85,7 +85,9 @@
         window.guardian.triggerLoadApp = function () {
             @loadApp
         };
-        if (!window.guardian.config.page.atoms || window.guardian.config.page.atoms[0] !== "2016/general-election") {
+        var page = window.guardian.config.page;
+        if (page.contentType !== 'Interactive' ||
+              !page.atoms || page.atoms[0] !== '2016/general-election') {
             window.guardian.triggerLoadApp();
         }
     } else {
--- a/common/app/views/fragments/inlineJSBlocking.scala.html
+++ b/common/app/views/fragments/inlineJSBlocking.scala.html
@@ -1,5 +1,5 @@
 @import common.InlineJs
-@import conf.switches.Switches.{AsyncCss, FontSwitch}
+@import conf.switches.Switches.{AsyncCss, FontSwitch, ElectionAppDelay}
 @import templates.inlineJS.blocking.js.{applyRenderConditions, cloudwatchBeacons, config, curlConfig, enableStylesheets, loadFonts, shouldEnhance}
 @import templates.inlineJS.blocking.polyfills.js.{classlist, details, matches, raf, setTimeout}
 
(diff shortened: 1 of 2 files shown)


wpf500:
Opps didn't mean to press Submit!

SiAdcock:
👍  for practicality!

Any idea how many articles this is going to affect? Just the one, or potentially dozens?

wpf500:
Just the one interactive

stephanfowler:
Will the '2016/general-election' atom appear on network fronts? Just checkin'

common/app/views/fragments/inlineJSBlocking.scala.html
```diff
@@ -66,18 +66,30 @@
...
-            var script = document.createElement('script');
-            script.src = '@Static("javascripts/app.js")';
-            var ref = document.getElementsByTagName('script')[0];
-            ref.parentNode.insertBefore(script, ref);
-        })(document);
+    @if(ElectionAppDelay.isSwitchedOn) {
+        window.guardian.triggerLoadApp = function () {
+            @loadApp
+        };
+        if (!window.guardian.config.page.atoms || window.guardian.config.page.atoms[0] !== "2016/general-election") {
```
stephanfowler: Can this array have length > 1 ?

wpf500:
Can atoms go on fronts?! This atom won't be on a front, but I could check for `contentType === Interactive` to be extract safe if you want?

common/app/views/fragments/inlineJSBlocking.scala.html
wpf500: There will only be one atom on this page

common/app/views/fragments/inlineJSBlocking.scala.html
```diff
@@ -66,18 +66,32 @@
...
-            var ref = document.getElementsByTagName('script')[0];
-            ref.parentNode.insertBefore(script, ref);
-        })(document);
+    @if(ElectionAppDelay.isSwitchedOn) {
+        window.guardian.triggerLoadApp = function () {
+            @loadApp
+        };
+        var page = window.guardian.config.page;
+        if (page.contentType !== 'Interactive' ||
+              !page.atoms || page.atoms[0] !== '2016/general-election') {
```
gtrufitt: Can't we check the atom server side? That way we can leave the rest of the site as-is and modify this just for the specific interactive, right?

wpf500: Yeah probably a better idea!

wpf500:
Will be hard to test this fully until its deployed, but it shouldn't impact existing articles at the very least

common/app/views/fragments/inlineJSBlocking.scala.html
gtrufitt: 💯

SiAdcock:
👍

PR #14909, opened 2016-11-07
Remove unused external discussion frontend switch

jfsoul:
## What does this change?
Remove the external discussion frontend switch which I don't think is used any more.  I think this was accidentally re-added [here](https://github.com/guardian/frontend/pull/14444/files#diff-55ff935672da59d0544844f7dcfe4601), probably through an incorrect merge conflict resolution.

Useless switch goes away.

No

## Request for comment
@nicl @crifmulholland @guardian/dotcom-platform


gtrufitt:
👍

PR #14910, opened 2016-11-07
Add New Zealand to the countries that MostPopularAgent saves in memory

janua:
## What does this change?

This updates the `GeoMostPopularAgent` to save New Zealand in memory. We have made changes in fastly that means the `NZ` region will start getting passed along.
This will mean in the `NZ` region we will see nothing as we would return `Nil`.

#### Alternative
Instead of diluting the cache in fastly with another region that doesn't have massive number, the alternative would be to change this PR to translate `NZ` to Rest of World; `row`.

The path in `ophan` for this new request is at `/api/mostread?country=nz`.

This means for `NZ` we will return most popular for the region instead of nothing.

No.

## Request for comment

@rich-nguyen

### DIFF
--- a/onward/app/feed/MostPopularAgent.scala
+++ b/onward/app/feed/MostPopularAgent.scala
@@ -44,7 +44,7 @@ class GeoMostPopularAgent(contentApiClient: ContentApiClient, ophanApi: OphanApi
   def mostPopular(country: String): Seq[RelatedContentItem] =
     ophanPopularAgent().getOrElse(country, ophanPopularAgent().getOrElse(defaultCountry, Nil))
 
-  def refresh() {
+  def refresh(): Unit = {
     log.info("Refreshing most popular for countries.")
     countries foreach update
   }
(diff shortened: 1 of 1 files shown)


gtrufitt:
> Instead of diluting the cache in fastly with another region that doesn't have massive number, the alternative would be to change this PR to translate NZ to Rest of World; row.

Why is it better to pass `Nil` than just `row` as the default in the `getOrElse`?

janua:
@gtrufitt Good question.
In this case, the `getOrElse` is around the `Option`, so it has nothing to do with country.
Another way of asking it is why don't we return `ROW` results for countries that have have no results.

I think the reason is because we never end up in this situation, we always pass along a country to this endpoint that we have. We are updating fastly and this change ensures that it again never happens.

I have updated locally the case you have mentioned. Could add it even though it will never get used? What do you think @gtrufitt

gtrufitt:
@janua Ah cool, thanks for the explanation. IMO, even though it may never get used, `row` is probably a better default than nothing. Your call 👍

janua:
@gtrufitt I have decided to do it, it's pretty straight forward and always guarantees results for unknown `X-Gu-GeoLocation`.

janua:
@gtrufitt Actually if you look here: https://github.com/guardian/frontend/blob/master/onward/app/controllers/MostPopularController.scala#L62 you can see that it already defaults the country to `row` if `X-Gu-Geolocation` doesn't exist. But this PR still covers the case of an unknown `X-Gu-GeoLocation` giving back `Nil` results.

gtrufitt:
🆒 😎

gtrufitt (approved):

PR #14911, opened 2016-11-07
Fix display of loading dots for MMA card update

AWare:
## What does this change?

Fixes small bug in the new stripe checkout flow where the old card number would be shown whilst members data api thinks about things. 

bugfix :( 

## Screenshots
na
## Request for comment
@SiAdcock

### DIFF
--- a/static/src/javascripts/projects/membership/stripe.js
+++ b/static/src/javascripts/projects/membership/stripe.js
@@ -29,51 +29,56 @@ define([
         var $button = $('.js-manage-account-change-card', $parent);
         var $updating = $('.js-updating', $parent);
 
-        /*  show/hide */
+        /*  show/hide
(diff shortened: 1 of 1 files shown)


static/src/javascripts/projects/membership/stripe.js
```diff
@@ -29,52 +29,60 @@ define([
...
         }();
 
+        //Decode and display card
         var oldCardType = $type.data('type');
         var newCardType = 'i-' + card.type.toLowerCase().replace(' ', '-');
 
-        if (oldCardType) {
-            $type.removeClass(oldCardType);
-        }
+        bean.off($button[0], 'click');
```
AWare: No, this isn't imported on any other pages apart from mma.

SiAdcock:
Nice fix / refactor 👍

AWare:
All good and tested.

PR #14912, opened 2016-11-07
Improve make install/git hooks

sndrs:
## What does this change?

- updates the way we manage git hooks:
 - only install them if we're not on TC
 - install them with the task runner
- removes the megalog success message when installing

simpler procedure for githooks and less visual cruft when installing

maybe TC, yeah (hopefully not though)

## Screenshots

screen shot 2016-11-07 at 15 26 09

## Request for comment

@guardian/dotcom-platform

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -66,7 +66,6 @@ module.exports = function (grunt) {
     /**
      * Miscellaneous tasks
      */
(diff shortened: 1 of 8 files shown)


rich-nguyen:
👍  cool

PR #14914, opened 2016-11-07
Better Meta

rjmk:
## What does this change?

By default most sharing icons are hidden and can be popped up by an ellipsis button.

All sharing icons are now the same on every page.

The dynamic share count is moved up to be on the same row as the icons.

We get to keep the share count (which we found does increase the number of shares significantly).
No yank & less of a gap between top of page and beginning of content

Yes. This affects AMP pages.
On AMP pages we just render all the icons as there is no share count to move around.

### Article left column before JS
image

### Article left column after JS
image

### Article left column share open
image

### Article single column share closed
image

### Article single column share open
image

### Article AMP
image

### Immersive gallery closed
image

### Immersive gallery open
image

## Request for comment
@zeftilldeath @stephanfowler @guardian/dotcom-platform


project/Frontend.scala
```diff
@@ -22,6 +22,7 @@ object Frontend extends Build with Prototypes {
       awsSns,
       awsSts,
       awsSqs,
+      cats,
```
rjmk: I can remove cats fairly easily -- it just helped with the way I thought about the problem. Let me know if you would rather I get rid of it

common/app/model/ShareLinks.scala
```diff
@@ -13,6 +15,19 @@ case class ShareLink (
   val userMessage: String = platform.userMessage
 }
 
+final case class ShareLinkMeta(
+  visible: Seq[ShareLink],
+  hidden: Option[NonEmptyList[ShareLink]]
```
TBonnin: What constraints are you trying to express here? I am not sure I understand

common/app/model/ShareLinks.scala
```diff
@@ -13,6 +15,19 @@ case class ShareLink (
...
+)
+
+object ShareLinkMeta {
+  def noneHidden(shares: ShareLinkMeta): ShareLinkMeta =
+    ShareLinkMeta(shares.visible ++ shares.hidden.fold(Nil: Seq[ShareLink])(_.toList), None)
+
+  def create: ((Seq[ShareLink], Option[NonEmptyList[ShareLink]])) => ShareLinkMeta =
```
TBonnin: apply?

TBonnin: also why a tuple?

project/Frontend.scala
TBonnin: I would prefer not to introduce any new dependency unless we cannot do without them.

TBonnin (changes requested):

ScottPainterGNM:
Christmas has come early!

image

common/app/model/ShareLinks.scala
rjmk: The idea is that there is a big difference between the empty case and the other cases. In the empty case we don't want to be doing anything. In the non-empty case we allow revealing arbitrarily many elements.

I'll drop this though as the `cats` dependency is going

common/app/model/ShareLinks.scala
rjmk: `apply` is a better name. 👍  A tuple because that is what we have access to later. It's a bit more verbose, but happy to manually pull apart the tuple below

project/Frontend.scala
rjmk: Will drop it

zeftilldeath:
Lovely! Gonna jump in here and do design fiddles 🌹

common/app/model/ShareLinks.scala
```diff
@@ -167,5 +168,18 @@ final case class ShareLinks(
...
+  def elementShares(elementId: String, mediaPath: Option[String]): Seq[ShareLink] = elementShareOrder.map( sharePlatform => {
+    val webUrlParams = campaignParams(sharePlatform)
+    val href = metadata.webUrl.addFragment(elementId).appendQueryParams(webUrlParams + ("page" -> s"with:$elementId"))
+    ShareLinks.create(sharePlatform, href = href, title = metadata.webTitle, mediaPath = mediaPath)
   })
+
+  val pageShares: ShareLinkMeta = {
+    pageShareOrder match {
+      case (visible, hidden) =>
+        ShareLinkMeta(sharesToLinks(visible), sharesToLinks(hidden))
```
TBonnin: I would directly pass the platform list here rather than using an intermediary tuple, that is used nowhere else
```
ShareLinkMeta(sharesToLinks(List(Facebook, Twitter, Email)), 
              sharesToLinks(List(LinkedIn, PinterestPage, GooglePlus, WhatsApp, Messenger))
)
```

rjmk: 👍

TBonnin (approved):

paperboyo:
> No yank & less of a gap between top of page and beginning of content

God bless you!

Some minor comments:
- this introduces second (and, crucially, different!) way of hiding/expanding share icons, unfortunately (third, if you count in the text sharing bubbles):

share style

- does the overlay close when one clicks outside of it (should)?
- “close” seems a bit superfluous, maybe an X?
- not sure about the shadow (and an overlay in general TBH)
- should obey media tone colour scheme

Just my 2gr, obviously!

johnduffell:
oh I'm totally seconding scott's comment! super christmas and spring clearout all in one! 🎉 
Although I didn't think the share buttons caused yank after Gareth's fixes, I think the save for later does though.  Or maybe it's a combination of the two?

rjmk:
> Although I didn't think the share buttons caused yank after Gareth's fixes, I think the save for later does though. Or maybe it's a combination of the two?

Sorry, that was a bit unclear. Gareth fixed yank. This preserves that fix while also fixing the gap to the articles

static/src/javascripts/projects/common/modules/social/hidden-share-toggle.js
```diff
@@ -0,0 +1,39 @@
...
+            });
+        });
+    }
+
+    return function hiddenShareToggle () {
+        $('.social-icon--more, .social__tray-close').each(function (toggle) {
+            bean.on(toggle, 'click', toggleDisplay);
+        });
+
+        $('.social__item--more').toggleClass('social--hidden');
```
SiAdcock: Maybe we could do this in a `fastdom.write` callback?

static/src/stylesheets/module/_social-icons.scss
```diff
@@ -0,0 +1,38 @@
+@mixin social-icon($platform) {
+    background-color: $platform;
+    &:hover {
+        background-color: darken($platform, 10%)
```
SiAdcock: Missing `;`

static/src/stylesheets/module/content/_gallery.scss
```diff
@@ -240,6 +240,15 @@
                 min-width: auto;
             }
         }
+
+        .social--expanded-top {
+            padding: 6px 0 6px 8px;
+            width: 304px; // (33 + 5) * 8
```
SiAdcock: Magic-ish numbers... Could probably use a more substantial comment or meaningful variable names

common/app/views/fragments/social.scala.html
```diff
@@ -1,24 +1,36 @@
...
+        target="_blank"
+        title="@item.text">
+            <span class='@if(item.css.matches("facebook|twitter|email")) {inline-icon__fallback button share-modal__item share-modal__item--@item.css} else { u-h }'>@item.userMessage</span>
+            @fragments.inlineSvg("share-" + item.css, "icon", List("rounded-icon", "social-icon", "social-icon--" + item.css) ++ iconModifier)
+        </a>
+    </li>
+}
+
+@SocialMore = {
+    <li class="social__secondary social__item social--hidden social__item--more"><span class="social-icon social-icon--more">&bull;&bull;&bull;</span></li>
```
gtrufitt: This should be button with hidden text as right now it's not tabble by keyboard and doesn't explain what it is via screen reader - therefore the hidden buttons are inaccessible. I would also use an SVG over the `&bull;` as it'd be strange for a screen reader to read them out.

common/app/views/fragments/social.scala.html
```diff
@@ -1,24 +1,36 @@
...
+}
+
+@SaveForLater = {
+    <li class="social--hidden social__item social__secondary"><a class="social__saveforlater-lozenge js-save-for-later saveforlater--unstyled">Save for later</a></li>
```
gtrufitt: is the `saveforlater` class consistent with a lack of hyphens elsewhere?

common/app/views/fragments/social.scala.html
```diff
@@ -1,24 +1,36 @@
...
+}
+
+@CloseTray = {
+    <li class="social__item social__secondary social--hidden"><span class="social__tray-close">Close</span></li>
```
gtrufitt: This should also be a button.

static/src/javascripts/bootstraps/enhanced/common.js
```diff
@@ -386,6 +388,7 @@ define([
                 ['c-media-listeners', mediaListener],
                 ['c-accessibility-prefs', accessibilityPrefs],
                 ['c-pinterest', modules.initPinterest],
+                ['ci-hidden-share-toggle', hiddenShareToggle],
```
gtrufitt: what's `ci`?

static/src/javascripts/projects/common/modules/social/hidden-share-toggle.js
```diff
@@ -0,0 +1,39 @@
...
+    'common/utils/$'
+], function (
+    fastdom,
+    bean,
+    $
+) {
+
+    function toggleDisplay () {
+
+        $('.social__secondary').each(function (icon){
```
gtrufitt: We try and use `js-` classes for targetting in JS.

gtrufitt: Also below targetting :)

static/src/stylesheets/module/_social.scss
```diff
@@ -64,10 +67,71 @@ category: Common
     }
 }
 
+.social-icon.social-icon--more {
```
gtrufitt: Do you need the extra specificity here or can you just use `.social-icon--more`?

gtrufitt (commented):
Nice refactor and tidy up! Good job

static/src/stylesheets/module/_social.scss
rjmk: Need the specificity. I'll add a comment

common/app/views/fragments/social.scala.html
rjmk: What about an `a`? That keeps the structure completely parallel to the other icons. We would have to use a prevent default with the event, but the button is only shown by JS anyway

common/app/views/fragments/social.scala.html
rjmk: It's hyphenated elsewhere. I find it very confusing considering the importance of the double hyphen, but will add them back in

paperboyo:
Hi, @zeftilldeath, could you screenshot your styling changes, please, for the code-illiterate?

zeftilldeath:
@paperboyo 

screen shot 2016-11-09 at 14 22 19
screen shot 2016-11-09 at 14 21 41

paperboyo:
Thanks, @zeftilldeath. Will we be measuring a possible decline in Save for later use? I know that the whole point was to get rid of too many elements in the meta, but I’m still concerned that hidden SFL may be overlooked too easily by new readers.

--- GRUMPY PERSONAL OPINION WARNING ---
I still don’t like the overlay (even if a bit less now). I would prefer for the extra bubbles to pop out alone and disappear on off-click to be consistent with comment and quote sharing. Maybe which three are constantly available could be chosen depending on per-user history of which are used? And I would leave Save for later out of it, for easier discovery and because it doesn’t have anything to do with sharing. But, yes, that would mean more clutter 😞. So: I’m not sure. Just thinking out loud. Ignore.
---

zeftilldeath:
@paperboyo Rafe and I agree with the click elsewhere to close so @rjmk will be looking into that. 🛩

paperboyo:
Incidentally: https://developers.google.com/web/updates/2016/10/navigator-share

gtrufitt:
🆗 A couple of minor changes and gotta fix that failing error which is odd as everything compiles locally but did a quick x-browser and lgtm. Will confirm with @zeftilldeath and @stephanfowler in the morning and then can push this out.

PR #14731, opened 2016-10-19
remove omniture calls from external video embed js

akash1810:
## What does this change?

Removes calls to omniture on the external embed video pages.

https://github.com/guardian/frontend/pull/14652 but scoped just to `embed.theguardian.com` pages.
## What is the value of this and can you measure success?

No more omniture.

## Does this affect other platforms - Amp, Apps, etc?

No.

## Screenshots

n/a
## Request for comment

@mkopka @guardian/guardian-frontend

PR #14916, opened 2016-11-07
Omniture - Remove page tracking and Omniture scripts. 🔥 

gtrufitt:
DNM until after US election.

## What does this change?

Removes the page tracking to Omniture and the omniture.js and related scripts. Should be the last thing that make JS calls to Omniture (clean up required for remaining pixel gifs).

Saves the 💲 by switching off Omniture.

~AMP doesn't have Omniture~. AMP has omniture removed now...

It'll affect the video-embed though @gidsg @akash1810 

## Request for comment

@guardian/dotcom-platform @dominickendrick @mkopka


common/app/views/fragments/analytics/google.scala.html
```diff
@@ -109,6 +106,7 @@
                         dimension12: referrerVars.value.path
                     });
                 }
+                window.sessionStorage.removeItem(NG_STORAGE_KEY);
```
gtrufitt: Added to remove the session storage item that was previously removed in the Omniture scripts

static/src/javascripts/projects/common/modules/analytics/analytics.js
```diff
@@ -1,254 +0,0 @@
...
-            if (navInteraction) {
-                trackNavigationInteraction(navInteraction.value);
-                window.sessionStorage.removeItem(NG_STORAGE_KEY);
-            }
-
-        } catch (e) { }
-
-        // Sponsored content
-        if(window.guardian.isModernBrowser) {
-            var contentNodes = document.querySelectorAll('[data-sponsorship]');
```
gtrufitt: @mkopka Do you know if we need to recreate this for GA? Tracking sponsored content?

dominickendrick:
I think there is some AMP Omniture stuff:

https://github.com/guardian/frontend/blob/master/common/app/views/mainAMP.scala.html#L44

dominickendrick:
screen shot 2016-11-07 at 16 59 39 🎉

dominickendrick:
👍

applications/app/views/videoEmbed.scala.html
```diff
@@ -103,7 +103,6 @@
         }
 
         document.documentElement.className = docClass;
-        @Html(common.Assets.js.omnitureJs)
```
akash1810: I removed calling omniture from this page in https://github.com/guardian/frontend/pull/14731. Forgot to remove the js from the template though. 👍

gtrufitt: Yay 👌

article/test/AnalyticsFeatureTest.scala
```diff
@@ -20,26 +20,7 @@ import conf.Configuration
...
     // Scenarios
 
-    scenario("Omniture tracks user actions") {
-
-      Given("I am on an article entitled 'Olympic opening ceremony will recreate countryside with real animals'")
-      goTo("/sport/2012/jun/12/london-2012-olympic-opening-ceremony") { browser =>
-        import browser._
-
-        Then("the Omniture webbug should record my visit")
-        val webbug = findFirst("#omnitureNoScript img")
```
gtrufitt: @mkopka @dominickendrick Hi both - this doesn't actually feel quite right to me - we don't have any `noscript` alternative to the omniture hits-secure pixel gif for GA. 

I would imagine that GA is therefore undercounting the PVs? 

This may have been a conscious decision, in which case I can leave the removal of the pageview pixel as is, otherwise we can repurpose it for GA.

Ps: [GA do provide the mechanism with utm.gif](https://developers.google.com/analytics/resources/concepts/gaConceptsTrackingOverview#gifParameters)

dominickendrick:
This can be shipped on Tuesday 15th November 2016

article/test/AnalyticsFeatureTest.scala
gtrufitt: For reference, answer to this:

"This noscript Omniture tag no longer exists for new Omniture implementations so the only reason we have it as part of dotcoms omniture implementation is down to how long ago we plugged it in.

The main issue with noscript tags we've had in the past is they only parse in a vert vanilla page view - meaning you then have a page view recorded that doesn't have the full range of custom dimensions."

PR #14917, opened 2016-11-07
Add top-traffic option to amp validation tools

jfsoul:
## What does this change?
This adds a top-traffic script to the node amp-validator.  Instead of working against a hardcoded set of endpoints this will make a request to http://api.ophan.co.uk/api/mostread and validate the top 25 articles by traffic.  This option can be used by running `npm run top-traffic` from the amp-validation directory.  The remaining functionality remains unaltered, it is still `npm start` or `node index.js`.

As non-amp-friendly articles may appear in the top 25 articles there is expected to be some errors when using this tool.  For the moment, this is purely for running locally - although a separate teamcity job could be added that has relatively permissive failure conditions (lots of test failures, 4XX/5XX statuses are a pass etc.).

We are trying to determine why some high-traffic articles/liveblogs are receiving a relatively low % of traffic via AMP in the US.  This is just a quick way of doing a spot check of the day's top content for validity.

## Request for comment

cc: @stephanfowler


tools/amp-validation/index.js
```diff
@@ -1,66 +1,4 @@
...
-    process.exit(1);
-}
-
-function buildErrorMessage(error) {
-    let msg = `line ${error.line}, col ${error.col}: ${error.message}`;
-    if (error.specUrl !== null) {
-        msg += ` (see ${error.specUrl})`;
-    }
-    return msg;
-}
```
jfsoul: This file was moved to run.js - I will mark any actual changes there as the diff is unhelpful.

tools/amp-validation/run.js
```diff
@@ -0,0 +1,69 @@
...
+function runValidator(validator, devChannel) {
+
+    return endpoint => fetchPage
+        .get(endpoint)
+        .then(res => {
+            console.log(`Checking the AMP validity (${devChannel ? 'pre-release' : 'release'}) of the page at ${isDev ? 'localhost:9000' : ''}${endpoint}, result is:`);
+            const result = validator.validateString(res);
+
+            const pass = result.status === 'PASS';
+            (pass ? console.log : console.error)(`Result for ${endpoint} was: ${result.status}`);
```
jfsoul: Changed log/error message here to include the endpoint as it's all async and the messages were appearing out of order sometimes.

tools/amp-validation/endpoints/ophan.js
```diff
@@ -0,0 +1,42 @@
...
+    });
+}
+
+function getEndpointsFromResponse(res) {
+    let body = '';
+    return new Promise((resolve, reject) => {
+        res.on('data', chunk => body += chunk)
+            .on('end', () => {
+                const endpoints = JSON.parse(body).map(hit => {
+                    return hit.url.replace(/^https?:\/\/www\.theguardian\.com/, '')
```
gtrufitt: I was going to say you could do an exact string match 'https://www.theguardian.com'  but then I looked and Ophan returns http :thinking:

jfsoul: Yeah, that puzzled me a bit - honestly I am kind of expecting this to break at some point.  The API might be able to return non-www urls - I'm not that sure.  I figure this way at least we're covering them changing to https, and removing the hostname altogether.

gtrufitt (approved):
Schweet.

PR #14918, opened 2016-11-07
Cache closed comment threads for longer.

gklopper:
## What does this change?
If comment threads are closed, they are cached for slightly more than an hour. If they are open they remain cached for 60 seconds.

I have been parsing logs of cache misses and this class of endpoints is one of the biggest in there.

We will be able to measure lower load on the discussion backend. Depending on how much lower it is, this will also equate to cpu and bandwidth savings.

No


gklopper:
cc @johnduffell

static/src/javascripts/projects/common/modules/discussion/loader.js
```diff
@@ -69,7 +69,7 @@ Loader.prototype.initTopComments = function() {
         this.gotoComment(commentId);
     });
 
-    return fetchJson('/discussion/top-comments/' + this.getDiscussionId() + '.json', {
+    return fetchJson('/discussion/top-comments/' + this.getDiscussionId() + '.json?closed=' + config.page.commentable, {
```
gklopper: There is no whitelist on api.nextgen.guardianapps.co.uk. The whitelist on www.theguardian.com is mainly to deal with campaign codes (which don't apply on this endpoint).

dominickendrick: 🎉

discussion/app/controllers/CommentsController.scala
```diff
@@ -161,4 +162,11 @@ class CommentsController(csrfConfig: CSRFConfig, val discussionApi: DiscussionAp
       }
     } recover toResult
   }
+
+  // caches "closed" comment threads for an hour.
+  // if the thread is switched on again the url changes and it cache busts itself.
+  private def cacheTime(request: RequestHeader, defaultCacheSeconds: Int) = {
+    val commentsClosed = request.getParameter("closed").contains("true")
+    if (commentsClosed && LongCacheCommentsSwitch.isSwitchedOn) 3800 else defaultCacheSeconds
```
johnduffell: ideally keep all integer cache seconds inside Cached.scala with the others, as a val with a helpful name

johnduffell:
good spot! go for it 👍 🚀

discussion/app/controllers/CommentsController.scala
```diff
@@ -161,4 +162,11 @@ class CommentsController(csrfConfig: CSRFConfig, val discussionApi: DiscussionAp
       }
     } recover toResult
   }
+
+  // caches "closed" comment threads for an hour.
+  // if the thread is switched on again the url changes and it cache busts itself.
+  private def cacheTime(request: RequestHeader) = {
+    val commentsClosed = request.getParameter("closed").contains("true")
+    if (commentsClosed && LongCacheCommentsSwitch.isSwitchedOn) CacheTime.DiskCached else CacheTime.Default
```
johnduffell: as opposed to `CacheTime.ClosedComments else CacheTime.OpenComments`?

discussion/app/controllers/CommentsController.scala
```diff
@@ -161,4 +162,11 @@ class CommentsController(csrfConfig: CSRFConfig, val discussionApi: DiscussionAp
       }
     } recover toResult
   }
+
+  // caches "closed" comment threads for an hour.
+  // if the thread is switched on again the url changes and it cache busts itself.
+  private def cacheTime(request: RequestHeader) = {
+    val commentsClosed = request.getParameter("closed").contains("true")
+    if (commentsClosed && LongCacheCommentsSwitch.isSwitchedOn) CacheTime(3800) else CacheTime(60)
```
johnduffell: 😆

PR #14919, opened 2016-11-07
Unify vendor and public js

rich-nguyen:
It's always been a bit confusing having both a `vendor` and `public` folder for js. The files are sometimes used differently, but in general the intended use is for dynamically requiring a third-party module. None of them are ever built into our js bundles. The remaining files are used to build an es5 shim, which I've reduced to a single build for any IE < 9.

I'm about to add another module to `vendor` soon, so hopefully this will be easier now.

I don't mind which location `vendor` goes in, but it was was easier to put `vendor` under `src` rather than `public`. What do you think @sndrs ?


sndrs (approved):
fyi @SiAdcock @gtrufitt

PR #14920, opened 2016-11-08
Treat 404 from members data API as valid response

philwills:
## What does this change?

Treats a 404 as a valid response from the members data API when it comes to setting the cookie that determines whether a logged in user is a paying member.

It allows us to serve membership and contributions messages to logged in users that are not members. This will be a success if we see increased rates of sign up to membership and contributions whilst avoiding showing promotional messages to those who are already paying members.

No.

## Request for comment

This is horrid to test since the members data API can't be successfully accessed from localhost or code. Any advice on a sane way to test this would be most welcome.

I can understand nervousness about integrating this change, but this bug is currently meaning a significant part of membership/contributions target audience are not seeing the messages intended for them.

cc @dominickendrick @paulbrown1982 @rtyley 

Before the refactoring to use fetch-json, any response was treated as valid, including the 404 that members data API uses to indicate that the user is not a known member. This reintroduces that behaviour, but limits it to just 404s, so 5xx statuses don't result in us setting a cookie indicating the user is not a paying member.

### DIFF
--- a/static/src/javascripts/projects/common/modules/commercial/user-features.js
+++ b/static/src/javascripts/projects/common/modules/commercial/user-features.js
@@ -1,13 +1,13 @@
 define([
     'common/utils/cookies',
     'common/utils/config',
-    'common/utils/fetch-json',
+    'common/utils/fetch',
     'common/utils/storage',
     'common/modules/identity/api'
 ], function (
     cookies,
     config,
-    fetchJson,
+    fetch,
     storage,
     identity
 ) {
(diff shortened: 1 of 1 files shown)


rtyley:
> Treats a 404 as a valid response from the members data API

I'd also be happy if the members-data-api returned an HTTP 200 and a 'not-found' json response - rather than going the full HTTP 404 - but if this is quicker, and stands less chance of breaking stuff, I'm fine with it.

paulbrown1982:
We could reduce the risk of breaking the apps by making a dotcom-only endpoint in members data API. e.g ~~~~
var userAttributeUrl = config.page.userAttributesApiUrl + '/me/features-web' }}
~~~~
But for now, I'm happy with the frontend code change too.

philwills:
I'm going to close this for now in favour of @paulbrown1982's change to the member's data API.

PR #14927, opened 2016-11-08
Fix logic of closed comments

gklopper:
## What does this change?

Fixes logic error with closed comments. Functionality is behind a switch, so not affecting PROD.

cc @johnduffell

### DIFF
--- a/discussion/app/controllers/CommentsController.scala
+++ b/discussion/app/controllers/CommentsController.scala
@@ -166,7 +166,7 @@ class CommentsController(csrfConfig: CSRFConfig, val discussionApi: DiscussionAp
   // caches "closed" comment threads for an hour.
   // if the thread is switched on again the url changes and it cache busts itself.
   private def cacheTime(request: RequestHeader) = {
-    val commentsClosed = request.getParameter("closed").contains("true")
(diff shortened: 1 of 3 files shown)


discussion/app/controllers/CommentsController.scala
```diff
@@ -166,7 +166,7 @@ class CommentsController(csrfConfig: CSRFConfig, val discussionApi: DiscussionAp
   // caches "closed" comment threads for an hour.
   // if the thread is switched on again the url changes and it cache busts itself.
   private def cacheTime(request: RequestHeader) = {
-    val commentsClosed = request.getParameter("closed").contains("true")
+    val commentsClosed = request.getParameter("commentable").contains("false")
```
johnduffell: change commentsClosed to commetable too?

johnduffell:
second time lucky! 👍

discussion/app/controllers/CommentsController.scala
gklopper: I'm leaving that one ;)

johnduffell: nice easter egg 🍳 😆

PR #14930, opened 2016-11-08
Make yarn check more robust, and check for min version

### LINKED ISSUE #14924: install yarn at minimum version
pre/post install scripts didn't run pre-v0.16.0, which we use to inject githooks

sndrs:
## What does this change?

fixes #14924 

- installs `yarn` if its missing
- updates `yarn` to minimum version if it's too low

`yarn@<0.16.0` does not run pre/post install hooks from `package.json`. We use them to inject githooks. Most people will have that installed and would need to manually update. This allows us to ensure we have a minimum version.

no

## Screenshots

screen shot 2016-11-08 at 13 42 16

## Request for comment

@gtrufitt @SiAdcock @rich-nguyen 

this uses the npm `semver` package to check the version. since `yarn` _installs_ `node_modules`, we cannot rely on that package being installed. therefore, this adds that package to `dev/` and checks it in. not sure about that...


rich-nguyen:
if `make install` can run another task, could you pop `check-yarn` into `tools/__tasks__`?

gtrufitt (approved):

sndrs:
I'll move it into `tools`, agree that make more sense than `dev`.

the thing with `__tasks__` though is that it's specifically intended for tasks  run by `run-task`, as things stand. `run-task` has the same problem of requiring deps we currently install to the top level `node_modules`.

whether or not `make` or `run-task` eventually seem the best way to get things done, i don't think having both is a good move. either the task runner should evolve into a binary of it's own or disappear, imo. if it evolves, it could then have its own deps, which seems sensible, and just install them before running anything (`yarn` when there's nothing to do is near enough instant), in which case `semver` could be one.

once grunt is all gone and the dust has settled, I will look at what we're left with. I'm not convinced a new task-runner is a good move, but there does seem to be value in the UI it provides, over `make`, to me.

but which ever seems best, plan is definitely to pick one

rich-nguyen:
That makes sense, I just misunderstood the location of the bits and bobs. As long as the `yarn install` and the yarn check are close together, it's all good. `tools` it is. 👍

rjmk:
Do we need the full power of the semver module? A solution to a similar problem is [here](https://github.com/sanctuary-js/sanctuary/blob/7d1693636f2e4468f02a4d10d81527cc9f742841/Makefile#L66). Do we think an approach like that could be robust enough?

Also, if we check in a node module, I would probably do it like

```gitignore
node_modules/*
!node_modules/semver
```

but taste varies

SiAdcock:
I agree with @rjmk that perhaps the full `semver` module is too overblown for this simple check. I'm sure there's a one-line bash script that will do it. Here's a hacky hipster functional JavaScript solution that may warm the cockles of your heart:

```
const spawn = require('child_process').spawn;
const checkVersion = spawn('yarn', ['--version']);

checkVersion.stdout.on('data', (data) => {
  const isAboveMinVersion = data
    .toString()
    .split('.')
    .map(x => parseInt(x, 10))
    .reduce((prev, current, index) => {
      if (prev) return prev;
      if (!index && current > 0) return true;
      if (index === 1 && current > 14) return true;

      return false;
    }, false);

  console.log(isAboveMinVersion);
});
```

rjmk:
Hacky bash one liner: `if ((yarn --version | cut -d . -f 1,2) > 0.15); then echo "hi"; fi`

sndrs:
what if the installed version is something the dev has been playing with, and the version is something like `1.0.0-alpha`, `1.0.0-alpha.1`, `1.0.0-0.3.7`, `1.0.0-x.7.z.92` (examples from http://semver.org)?

sndrs:
i guess @rjmk's example does allow us to specify minimum minor version, which is all we need for now

rjmk:
Yes, if I can de-break it! @SiAdcock pointed out that the comparison is not actually evaluating at the moment

rjmk:
Debroke: `echo $(yarn --version | cut -d . -f 1,2) '> 0.15' | bc` gives us the 0 or 1 we need

sndrs:
thanks everyone, this is a miles smaller PR now! i'll leave `check-node` over there in tools, though, seems the best place for it. guess this needs another ok, if that's ok?

tools/check-node-version.js
```diff
@@ -9,8 +11,8 @@ var currentVersion = Number(process.version.match(/^v(\d+\.\d+)/)[1]);
 var minVersion = Number(fs.readFileSync(path.join(__dirname, '../', '.nvmrc'), 'utf8'));
 
 if (currentVersion < minVersion) {
-    console.log("You are using Node v" + currentVersion + ".\n" +
-        "Frontend requires Node v" + minVersion + " or later.\n" +
-        "If you're using NVM, you can 'nvm use'...");
+    console.log('You are using Node v' + currentVersion + '.\n' +
+        'Frontend requires Node v' + minVersion + ' or later.\n' +
+        'If you\'re using NVM, you can \'nvm use\'...');
```
SiAdcock: Template string here? 😛

SiAdcock:
Wow, it's so small! Nice one @sndrs and @rjmk  👍

tools/check-node-version.js
sndrs: the annoying thing with template strings is needing to remove the indents in examples like this?

SiAdcock: Haha, yeah I see, that's a shame. You could extract it into a `const` at the top of the module to make the lack of indentation slightly less annoying. But meh, it's no biggie

sndrs: go on then :)

PR #14932, opened 2016-11-08
Fix structured data in articles and live blogs

SiAdcock:
## What does this change?

Structured data is broken in articles and live blogs for a number of reasons. This change fixes two things:

- Removes `coverageStartTime` from NewsArticles
- Adds correct publisher logo metadata (`height` and `width`) for AMP LiveBlogs

Structured data validation failures prevent articles from being indexed for inclusion in Google's AMP search results.  

This should hopefully get our articles AMP-ified in Google search results again ⚡️ 

## Request for comment

### DIFF
--- a/article/app/views/fragments/logo.scala.html
+++ b/article/app/views/fragments/logo.scala.html
@@ -1,0 +1,12 @@
+@(amp: Boolean = false)(implicit request: RequestHeader)
+
+@if(amp) {
+    @* AMP doesn't support sameAs *@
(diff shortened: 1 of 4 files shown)


article/app/views/fragments/logo.scala.html
```diff
@@ -0,0 +1,12 @@
+@(amp: Boolean = false)(implicit request: RequestHeader)
```
SiAdcock: No, I'll kill it 🔥

gtrufitt (approved):

PR #14903, opened 2016-11-07
Revert "Use scanamo for redirects, centralise"

jfsoul:
Reverts guardian/frontend#14864

Reverting this since redirects seem to be broken - gives me some time to figure out why.

PR #14934, opened 2016-11-08
Revert "Revert "Use scanamo for redirects, centralise""

jfsoul:
Reverts guardian/frontend#14903 and attempts to fix the reason for the original revert.

It turns out the [AWS Java SDK clients only perform credential refreshing](http://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/java-dg-roles.html) when instantiated using the default constructor, or one taking `AWSCredentialsProvider` (which makes sense).  I mistakenly used the constructor taking `AWSCredentials` instead.

I intend to merge this post-election as it'll be hard to verify the fix 100% locally.  Playing it safe, just in case.

N.B. I was overzealous in squashing the revert commits, so this double-revert is also squashed.  The interesting commit is [here](https://github.com/guardian/frontend/commit/9ba7afd79b4e77aa2284229862be683bdcc431e5)

cc: @TBonnin


dominickendrick:
👍

TBonnin (approved):
Good job. Let's merge that tomorrow

PR #14891, opened 2016-11-04
Hosted gallery Enlarged Captions mode [not to be merged yet]

lps88:
## What does this change?
New mode for the Hosted Gallery, where the captions are displayed over the top of the images. 

We will switch between the two modes based on a flag from CAPI eventually, but for now there's a keyboard shortcut (press alt-shift-C), so we can see how it looks for existing galleries.

## What is the value of this and can you measure success?
Makes the gallery more versatile (and a current prospective client happy, we hope)

## Does this affect other platforms - Amp, Apps, etc?
No 

## Screenshots
Existing mode:
image
New mode:
image

## Request for comment
@guardian/labs-beta

PR #14937, opened 2016-11-08
Hosted gallery - refactor

lps88:
## What does this change?
We dynamically resize and position lots of the gallery components to fit inside the 5:3 ratio box that the images sit in. This wraps them all in a container so that we only need to resize that.

This commit was a refactor I did as part of this https://github.com/guardian/frontend/pull/14891 which now looks unlikely to be used.

Less code/dom manipulation

## Screenshots
image
image

## Request for comment
@guardian/labs-beta


commercial/app/views/hosted/guardianHostedGallery.scala.html
```diff
@@ -38,6 +38,7 @@
         <div class="l-side-margins hosted__side">
             <section class="hosted-tone--dark hosted-gallery__gallery-section">
                 <div class="js-hosted-gallery-container hosted-gallery__gallery-container">
+                <div class="js-hosted-gallery-frame hosted-gallery__gallery-frame">
```
rich-nguyen: indent

rich-nguyen:
👍

PR #14938, opened 2016-11-08
Add epic contributions back up

jranks123:
## What does this change?

UPDATE: also fixes ab-test-clash to ensure Outbrain compliance

Tonight we want to avoid deploying unless in extreme circumstances. This pull request adds a back up test to be switched on in the case of membership failing, as currently if that happens then we would have to turn off all contributions/memebership ads on the site and we'd be left with nothing, as there are currently NO contributions-only ads live. 

(If contributions goes down, there is still plenty of membership-only ads we can put on the site)

If this test is on, the other two tests (ContributionsMembershipEpicCtaUnitedStatesTwo and ContributionsMembershipEpicCtaRestOfWorldTwo)  have to be switched off first. I am in the office tonight so i will take this responsibility. 

I realise that this functionality might be better suited to a feature switch but as a last minute safe-guard I think this suffices. 

It is worldwide, encompassing the coverage that the two aforementioned live tests running at the moment cover, and it uses the most conservative parameters for deciding whether to display: only on non-sensitive articles and just US election and US politics tagged articles. 

If membership goes down, we can still make money from contributions. 

## Screenshots
screenshot at nov 08 16-26-56

### DIFF
--- a/static/src/javascripts/projects/common/modules/experiments/tests/contributions-membership-epic-backup.js
+++ b/static/src/javascripts/projects/common/modules/experiments/tests/contributions-membership-epic-backup.js
@@ -1,0 +1,123 @@
+define([
+    'bean',
+    'qwery',
+    'common/utils/$',
+    'common/utils/template',
+    'common/views/svg',
+    'common/utils/fastdom-promise',
+    'common/utils/mediator',
+    'text!common/views/contributions-epic-equal-buttons.html',
+    'text!common/views/contributions-epic.html',
+    'common/utils/robust',
+    'inlineSvg!svgs/icon/arrow-right',
+    'common/utils/config',
+    'common/utils/cookies',
+    'common/modules/experiments/embed',
+    'common/utils/ajax',
+    'common/modules/commercial/commercial-features',
+    'lodash/arrays/intersection'
+
+], function (bean,
+             qwery,
+             $,
+             template,
+             svg,
+             fastdom,
+             mediator,
+             contributionsEpicEqualButtons,
+             contributionsEpic,
+             robust,
+             arrowRight,
+             config,
+             cookies,
+             embed,
+             ajax,
+             commercialFeatures,
+             intersection) {
+
+
+    return function () {
+
+        this.id = 'ContributionsMembershipEpicBackup';
+        this.start = '2016-11-08';
+        this.expiry = '2016-11-11';
+        this.author = 'Jonathan Rankin';
+        this.description = 'Back up contributions-only epic in case membership goes down';
+        this.showForSensitive = false;
+        this.audience = 1;
+        this.audienceOffset = 0;
+        this.successMeasure = 'Impressions to number of contributions/supporter signups';
+        this.audienceCriteria = 'All readers reading about US politics/election';
+        this.dataLinkNames = '';
+        this.idealOutcome = 'If membership goes down, we have a backup way of asking for support from our readers';
+        this.canRun = function () {
+            var whitelistedKeywordIds = ['us-news/us-elections-2016', 'us-news/us-politics'];
+            var pageKeywords = config.page.keywordIds;
+            var hasKeywordsMatch = pageKeywords && intersection(whitelistedKeywordIds, pageKeywords.split(',')).length > 0;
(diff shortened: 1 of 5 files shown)


static/src/javascripts/projects/common/modules/experiments/tests/contributions-membership-epic-backup.js
```diff
@@ -0,0 +1,127 @@
...
+
+            var userHasNeverContributed = !cookies.get('gu.contributions.contrib-timestamp');
+            var worksWellWithPageTemplate = (config.page.contentType === 'Article') && !config.page.isMinuteArticle; // may render badly on other types
+            return userHasNeverContributed && commercialFeatures.canReasonablyAskForMoney && worksWellWithPageTemplate && hasKeywordsMatch();
+        };
+
+        var contributeUrl = 'https://contribute.theguardian.com/?';
+
+
+        var message = '...we have a small favor to ask. More people are reading the Guardian than ever but far fewer are paying for it. And advertising revenues across the media are falling fast. So you can see why we need to ask for your' +
```
jfsoul: "favor" is american-english spelling, but then "advertising" is english - as this is going to all regions I assume we would default to "favour"?

static/src/javascripts/projects/common/modules/experiments/tests/contributions-membership-epic-backup.js
```diff
@@ -0,0 +1,127 @@
...
+        this.audienceCriteria = 'All readers reading about US politics/election';
+        this.dataLinkNames = '';
+        this.idealOutcome = 'If membership goes down, we have a backup way of asking for support from our readers';
+        this.canRun = function () {
+            var whitelistedKeywordIds = ['us-news/us-elections-2016', 'us-news/us-politics'];
+
+            var hasKeywordsMatch = function() {
+                var pageKeywords = config.page.keywordIds;
+                return pageKeywords && intersection(whitelistedKeywordIds, pageKeywords.split(',')).length > 0;
+            };
```
jranks123: amended

jfsoul:
So I assume this is a test because it fits the structure of our other tests very closely and was therefore much quicker to implement this way rather than as a plain feature switch?

jranks123:
@jfsoul yes that is correct. I haven't used the feature switch framework before so didn't want to risk not getting it through in time. I really think this is an important safeguard, as membership had problems during brexit, so I know it isn't the perfect way of implementing it but it works and I know this framework really well.

dominickendrick:
This is a great idea! I think it could be slightly better to implement as a feature switch, rather than a test.
I also think you should **_not_** push this up tonight, but wait until tomorrow. I understand the desire to mitigate risk during a high traffic events, but this change adds more risk right at the end of the day. @philwills agreed in person.

jranks123:
@dominickendrick ok fair enough.

static/src/javascripts/projects/common/modules/experiments/tests/contributions-membership-epic-backup.js
```diff
@@ -0,0 +1,123 @@
...
+            var worksWellWithPageTemplate = (config.page.contentType === 'Article') && !config.page.isMinuteArticle; // may render badly on other types
+            return userHasNeverContributed && commercialFeatures.canReasonablyAskForMoney && worksWellWithPageTemplate && hasKeywordsMatch;
+        };
+
+        var contributeUrl = 'https://contribute.theguardian.com/?';
+
+
+        var message = '...we have a small favour to ask. More people are reading the Guardian than ever but far fewer are paying for it. And advertising revenues across the media are falling fast. So you can see why we need to ask for your' +
+            ' help. The Guardian\'s independent, investigative journalism takes a lot of time, money and hard work to produce. But we do it because we believe our perspective matters – because it might well be your ' +
+            'perspective, too.';
```
katebee: Bit of a red herring, but I am curious about the reason for the string concatenation?

jranks123: Just because it didn't fit on my screen otherwise! Certainly not neccassery

PR #14940, opened 2016-11-08
Extend discussion switch

markjamesbutler:
## What does this change?

Extend switch to ensure no disruption for the election tonight.

None its a functional change.

No


markjamesbutler:
@nicl Are you happy with this change?

DiegoVazquezNanini:
:+1:

DiegoVazquezNanini (approved):

nicl:
👍 yes sorry would be good to extend and then we'll sort out what to do with this! Thanks.

nicl:
@markjamesbutler only comment would be that it would be helpful to extend the switch perhaps a week rather than 2 days.

PR #14942, opened 2016-11-09
Delete the Weekend Reading email AB test 

katebee:
## What does this change?

We were using the AB test to trial two formats for the Weekend Reading email. This required subscribing users to two different ListIDs in ExactTarget. This PR kills that AB test; now all users will be signed up to the same ExactTarget email list.

ListID 3743 (Minute version) is the winner! So this PR will stop us subscribing people to the Article version (3744)

**We need to co-ordinate with Editorial to ensure the correct listID persists**

- Editorial no longer have to make two versions of the Weekend Reading email
- Killing off AB test 💀

Nope


dominickendrick:
👍

PR #14943, opened 2016-11-09
Piggyback pixel advert resizing for AppNexus promo campaigns

rich-nguyen:
## What does this change?
This adds an endpoint to api.nextgen for something known in AppNexus as a piggyback pixel, a script (sic, not a pixel) that is appended to the creative served by the AppNexus tag, which is passed to the page from DFP. This piggyback-pixel-script can be passed advert dimensions through macros, and its the only party which knows the true size of the advert (DFP is not aware of the creative size, only the requested size).

Luckily, we already have a similar system in place for `sticky-top-banner` to allow Rubicon adverts to resize, so we can reuse the iframe messenger system to respond to the same `set-ad-height` message.

screen shot 2016-11-09 at 11 59 27


rich-nguyen:
[Adops would like iframe busting ad slot for AppNexus](https://trello.com/c/Fej162oN/82-adops-would-like-iframe-busting-ad-slot-for-appnexus)

regiskuckaertz (approved):
👍 

We might want to handle both the top banner and the right slot with a single module in the future, so that both Rubicon, AppNexus and DFP native ads can get delivered and resize the corresponding slots.

rich-nguyen:
@regis 3 is the magic number.

regiskuckaertz:
https://www.youtube.com/watch?v=aU4pyiB-kq0 :-)

PR #14944, opened 2016-11-09
Abstract colour out of hosted page model

kelvin-chappell:
Because there's nothing font-specific about it.


common/test/common/commercial/hosted/ColourTest.scala
```diff
@@ -0,0 +1,45 @@
...
+
+/*
+ * For colours, see http://www.colorhexa.com/<hexcode>
+ * eg. http://www.colorhexa.com/2ec869
+ */
+class ColourTest extends FlatSpec with Matchers {
+
+  "isDark" should "be true for Zootropolis green" in {
+    val zootropolisColour = Colour("#2ec869")
+    zootropolisColour shouldBe 'dark
```
lps88: this syntax is weird! Does it work using the `isDark` val?

kelvin-chappell: Yes. http://www.scalatest.org/user_guide/using_matchers#checkingBooleanPropertiesWithBe

(But if it makes the test worse, I'll get rid of it.)

lps88: Nah I think it's fine, just haven't seen it before :)

lps88:
👍

PR #14947, opened 2016-11-09
Remove variant as control was winning significantly

philwills:
## What does this change?

Drops the variant for the Epic CTA, as the control was winning by a significant margin.

We get more contributions and membership sign-ups.

No

## Screenshots
image

## Request for comment

@desbo


philwills:
We should probably stop this being a test, seeing as it only has one variant now, but I wanted to keep this as small as possible.

desbo (approved):

dominickendrick:
👍  for pragmatism !!

PR #14949, opened 2016-11-09
Hosted Gallery enhancements

lps88:
## What does this change?
- make sure links are properly formatted in the caption text (was showing escaped HTML)
- add show/hide button to the onward journey so the whole image can be viewed
- add shadow behind the counter/arrows on the right so they're visible on white
- fix image positioning on first load 

## Screenshots
image
image

## Request for comment
@guardian/labs-beta


kelvin-chappell (approved):
👍

PR #14950, opened 2016-11-10
Add outbrain compliance check for epic test

jranks123:
## What does this change?

For any ab tests that add components that sit at the end of an article, we need to update ab-test-clash with the name of the test and it's variants in order to ensure outbrain compliance. This PR does that for the current Epic test that is running on frontend. 

## Request for comment
@gtrufitt


gtrufitt:
👍 cheers

PR #14956, opened 2016-11-10
Introduce tests for crosswords

MatthewJWalls:
## What does this change?

* Separate out the crossword react component to make it accessible directly
* Remove usages of the classNames library and move it inline
    * I was having some problems compiling for test with this dependency, since it was just 5 
       lines of code I inlined it and dropped the dependency.
* Write tests

Now under test

No


static/src/javascripts/projects/common/modules/crosswords/cell.js
```diff
@@ -16,6 +14,14 @@ define([
             this.props.handleSelect(this.props.x, this.props.y);
         },
 
+        classNames : function(props) {
+            var out = [];
+            for(var f in props){
```
dominickendrick: Could this be done with:
```
return Object.keys(props)
 .filter(function(f) {props[f] === true})
 .join(' ')
```
for more functional fun ?

static/src/javascripts/projects/common/modules/crosswords/confirm-button.js
```diff
@@ -1,20 +1,27 @@
...
     var ConfirmButton = React.createClass({
+
         getInitialState: function () {
             this.timeout = this.props.timeout || 2000;
             return {
                 confirming: false
             };
         },
 
+        classNames : function(props) {
```
dominickendrick: It does feel a bit odd to have to inline these, could we put the `className` helper in `crossword` module path to see if we can fix the test compilation issues ?

dominickendrick:
👍 apart from comments - Great to see more tests !!!

static/src/javascripts/projects/common/modules/crosswords/cell.js
MatthewJWalls: themoreyouknow

MatthewJWalls:
I've cleaned up the classNames dependency and moved it as a package-level module

static/src/javascripts/projects/common/modules/crosswords/classNames.js
```diff
@@ -0,0 +1,9 @@
+define([], function () {
+
+    return function(props) {
+        return Object.keys(props)
+            .filter(function(f){ return props[f] === true; })
+            .join(' ');
+    };
+
+});
```
SiAdcock: Elegant 😍 

Since this module has no dependencies, you can drop the empty array from the first argument of the define call.  

```
define(function() { ... })
```

static/src/javascripts/projects/common/modules/crosswords/crossword.js
```diff
@@ -0,0 +1,720 @@
...
+    keycodes,
+    persistence,
+    debounce,
+    filter,
+    contains,
+    map,
+    some,
+    zip,
+    forEach,
+    find
```
SiAdcock: I wonder how many of these lodash array methods we need any more? Certainly `filter`, `map`, `some` and `forEach` are part of ES5 and supported all the way back to IE9. We include an ES5 shim to support even older browsers.

SiAdcock:
Nice, @MatthewJWalls, I'm very excited to see more client side tests! 

I noticed that a number of the tests are testing multiple (related) actions. For instance, `Can move focus with tab keys` tests that `tab` and `shift-tab` actions have the desired result. In my opinion each test should only test one action, so that they can only fail for one reason. If the aforementioned `tab keys` test failed, I wouldn't know if it was the `tab` action or the `shift tab` action that was breaking.

What do you think?

MatthewJWalls:
@SiAdcock Yeah I think the same, I'm just fantastically lazy half the time

MatthewJWalls:
now done:

* Split tests up @SiAdcock 
* Removed the redundant []
* Fixed a small styling issue I noticed I introduced on the buttons

I think this should do it? I can look at the lodash stuff separate from this, this PR has knocked around for a while now!

SiAdcock:
👍

SiAdcock:
How about we get this merged @MatthewJWalls? 🚢  🚀

SiAdcock (approved):

gtrufitt:
🙏 We're probably going to have to look at the React stuff soon to refactor out bad habits so would be good to have these!

sndrs:
shall we merge this? :

MatthewJWalls:
OK I'm just fixing some merge conflicts on this again from the javascript directory change and then I'll force push it again. Fingers crossed everything still works.

MatthewJWalls:
Ok - I'm pretty happy this is good again now. I don't want to change crosswords at the end of the day so I'll merge this tomorrow morning.

PR #14957, opened 2016-11-10
Restore right-hand side MPU in liveblogs

regiskuckaertz:
300K impressions are lost every month because this slot is missing from liveblogs since last February.


regiskuckaertz:
[Restore right MPU in liveblogs](https://trello.com/c/q7UKJGPC/155-restore-right-mpu-in-liveblogs)

JonNorman (approved):

JonNorman:
Despite having _just_ approved this - are there any screenshots that demonstrate that this doesn't mess up the styling on the live blogs? Or your irrefutable word that it doesn't? :)

regiskuckaertz:
Ha! Good reflex, Jon. By convention, `js-` prefixed classes serve as hooks for scripts, _never_ styles, so we are safe on this side. There is however another script using this class, `football.js`; fortunately, it only works when the page is a competition.

regiskuckaertz:
I will add a screenshot as soon as my compilation stops failing 😭

regiskuckaertz:
picture 2

boo-yeah!

PR #14948, opened 2016-11-09
Track sponsor logo impressions

kelvin-chappell:
This is the follow-up to #14931. 

Includes custom dimension 31 in page-view hits to hold the name of the page sponsor on all sponsored pages, and any container sponsors on a front.

Eg.
Article:
image

Front:
image

There's also some refactoring.

/cc @mkopka @dominickendrick @guardian/commercial-dev

PR #14958, opened 2016-11-10
Track sponsor logo impressions (again)

kelvin-chappell:
This is the same as #14948 but with a small bug fix.
The fix is to escape single quotes in sponsor names so that they don't break GA tracking.

### DIFF
--- a/applications/app/views/fragments/galleryBody.scala.html
+++ b/applications/app/views/fragments/galleryBody.scala.html
@@ -46,7 +46,7 @@
     }
 </div>
 <div class="l-side-margins">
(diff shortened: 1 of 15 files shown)


JonNorman (changes requested):
Could you use StringEscapeUtils here? It does what you're after, plus any other annoying characters that may pop up and we already have a dep on it.

http://www.docjar.com/docs/api/org/apache/commons/lang/StringEscapeUtils.html#escapeJava(String)

kelvin-chappell:
@JonNorman I didn't know about this.  I'll do it!  Thank you.

JonNorman (approved):

PR #14961, opened 2016-11-10
Fitter Twitter

zeftilldeath:
A small change that is tacked onto the Better Meta initiative. Calling this one Fitter Twitter, despite the fact that it's actually about twitter and email links.

Basically, the moment we go below our leftCol breakpoint, vertical space of these meta elements matters a lot more. The deeper it is, the further away the article is. 

I've got the author contact details inline with the name now, saving one line and sometimes two lines of vertical space.  

I'd say there's little to no benefit of knowing a twitter handle before being on twitter, the icon displayed alongside an author's name, communicates enough. From leftCol onwards, the styling remains unchanged. 

- I've let the 'Contact author' text remain at mobile for better clarity. 
- I've stripped away the tone from the date stamp, the date stamp is not a primary link, and that level of colour just detracts from the more important links. 

BEFORE .......................................................................................... AFTER

c

b

a

@stephanfowler @rjmk


dominickendrick:
👍 legendary !!!

sndrs:
is the hit area with just the icon large enough?

rjmk:
In the middle case (no author name), would it make sense to keep the tag?

paperboyo:
I like the above (bar the comments). Two general small things related to Better Meta initiative:
- I think clicking on a date to reveal the modification date does not work on galleries
- @superfrank has this much nicer email icon he’s been using (e.g. https://www.theguardian.com/signup/weekendreading), worth replacing our current meta email icon with his one, I think

gtrufitt (approved):

zeftilldeath:
Got the hit area as large as the other buttons now...

screen shot 2016-11-14 at 12 08 50
screen shot 2016-11-14 at 12 09 16

@paperboyo Good spot with regards to galleries date interaction, is that because galleries are different to editorial articles, or is it just something that slipped the net? 

And @superfrank could you replace all email icons with your nice new one then please?

paperboyo:
@zeftilldeath I wouldn’t know for sure, but I think this is an omission rather than a concious decision. I would expect the functionality be in parity.

@superfrank Sorry to have landed you in the pickle, but I never liked the current email icon 😜

superfrank:
I'll update the email icon as a separate PR (gotta get those PR pts up).

PR #14962, opened 2016-11-10
Make RH-MPU in liveblogs sticky again

regiskuckaertz:
I had to wrap the content of the feed in a new `DIV` to get a hook on it, hoping @gtrufitt will be happy about it? 

sticky


gtrufitt (approved):
I'm happyish. ;)

regiskuckaertz:
😆  Thanks, pal!

PR #14964, opened 2016-11-11
Block asking for contributions on Leonard Cohen pieces

philwills:
## What does this change?

Ensures we don't ask for contributions/membership on any Cohen pieces. I don't think we are, but this seems a sensible precaution.

No

## Request for comment

@jranks123

### DIFF
--- a/static/src/javascripts/projects/common/modules/experiments/tests/contributions-epic-post-election-copy-test.js
+++ b/static/src/javascripts/projects/common/modules/experiments/tests/contributions-epic-post-election-copy-test.js
@@ -48,7 +48,7 @@ define([
         this.dataLinkNames = '';
         this.idealOutcome = 'We learn to what extend using messages that chime with current events have an impact on contributor/supporter conversion';
         this.canRun = function () {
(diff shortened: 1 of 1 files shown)


jranks123:
👍

LATaylor-guardian:
Just chipping in here, but this seems like something we most certainly don't want to do for a specific subset of articles. Is there a way to do this in a more automatic way?

philwills:
@LATaylor-guardian It's what the targeting tool is supposed to fix generally, I believe, but my understanding is that that's not yet ready.

LATaylor-guardian:
@philwills Ah yes, good point, I forgot about that! Thanks.

jranks123:
I also think a better solution than included and excluded hard coded lists is in order...Maybe using the Google Sheets API?

philwills:
This is failing at the moment because @JustinPinner has an expired test.

rupertbates:
@philwills I'm about to merge a PR which will fix that

dominickendrick:
👍 Hallelujah!

PR #14966, opened 2016-11-11
Engagement banner us copy test 2

rupertbates:
## What does this change?
Extends the Membership US copy test, adds new variants and uses the new design

The new engagement banner design has been shown to be more effective so we are now rolling it out to the US.

Web only

## Request for comment

@Ap0c


Ap0c (approved):
Looks like it's working as expected 👍.

PR #14967, opened 2016-11-11
Fix inline ad slot when containing books component

Calum-Campbell:
## What does this change?
 - Fix width of slot that holds the books inline component 

no

## Screenshots
image

image

## Request for comment
@regiskuckaertz @JonNorman

### DIFF
--- a/static/src/javascripts/projects/commercial/modules/creatives/commercial-component.js
+++ b/static/src/javascripts/projects/commercial/modules/creatives/commercial-component.js
@@ -156,6 +156,7 @@ define([
      * @param {Object=} params
      */
     function CommercialComponent(adSlot, params) {
+        if(params.type == 'book') adSlot.addClass('ad-slot--books-inline');
         this.params = params || {};
         this.type = this.params.type;
         // remove type from params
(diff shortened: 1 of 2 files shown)


static/src/javascripts/projects/commercial/modules/creatives/commercial-component.js
```diff
@@ -156,6 +156,7 @@ define([
      * @param {Object=} params
      */
     function CommercialComponent(adSlot, params) {
+        if(params.type == 'book') adSlot.addClass('ad-slot--books-inline');
```
regiskuckaertz: just wrap that in a `fastdom.write` call

static/src/stylesheets/module/_adslot.scss
```diff
@@ -347,6 +347,13 @@
     }
 }
 
+.ad-slot--books-inline {
+    width: $gs-gutter * 6.5;
```
regiskuckaertz: Isn't there a `gs-span` equivalent for that?

regiskuckaertz (changes requested):

regiskuckaertz (approved):

PR #14970, opened 2016-11-11
Update 'popular' and 'most popular' to 'most viewed'

gtrufitt:
## What does this change?

Updates *most popular* to *most viewed* in labelling across the site.

Most viewed was deemed in poor taste by some readers when referring to stories that aren't semantically 'popular' but more specifically 'most viewed'. Improves the semantics of the labelling.

Yes. It updates Amp. Apps also need to do this work (@rcrphillips).

## Screenshots

image

image

## Request for comment

@guardian/dotcom-platform 

Question - Is there the concept of dictionaries that would mean we could store titles like this (i.e i18n txt files) and update in one location?


dominickendrick:
👍 you are most ~~popular~~ viewed

TBonnin:
Yeah. Looks like this could be centralized in one location

johnduffell:
indeed, the more often we change it, the more argument there is for storing it somewhere central

gtrufitt:
I [looked into i18n in Play](https://www.playframework.com/documentation/2.5.x/ScalaI18N) and started trying to implement it, but I think if it's something we want, then it might be worth doing as a seperate PR as it means messing with controllers and injection and other stuff I don't quite understand... Lets put it as a platform task.

PR #14971, opened 2016-11-11
Top-traffic content amp validation improvements

jfsoul:
## What does this change?
This PR adds some improvements to the ophan-top-traffic option of the amp validator which differentiates it more from the normal fixed-endpoints option.  Differences are:

- No longer reports the specific error details for each validation error for each failure as this is more useful for developers.
- Now checks for the presence of an `amphtml` `<link>` in the version of the page fetched from www .  If the `<link>` isn't present then we don't try and validate the page.  There is no need to apply this to the fixed endpoints, as these are hand-picked amp content.
- Logging is a bit more sensible - i.e. only suggest running `dev-build` if actually trying to fetch from localhost.  Also the pass/fail messages are easier to read now.

The top-traffic option becomes more useful as it can be used more as a *quick check* rather than a full blown validation.  It will also only attempt to validate actual amp content.

Nope

## Screenshots
image

image

## Request for comment
@gtrufitt

cc: @stephanfowler


gtrufitt (approved):
👍

PR #14972, opened 2016-11-11
Fix illegible rich links on paid articles

JonNorman:
### What does this change?
It's really good when people can read our rich-links on articles. On paid-content articles, this isn't currently possible due to the grey background:

image

These changes make it possible:
image

### Additional Notes
Originally looking into this, I thought I'd need to add a class to the rich-link based on whether the content type had branding. It turns out I didn't need to do this, but did a tiny amount of refactoring of the commercial metadata properties while I was at it. Yes, it probably shouldn't be in this PR, but the changes are minor and it _is_ Friday....

### Request for comment
@guardian/commercial-dev

### DIFF
--- a/common/app/model/meta.scala
+++ b/common/app/model/meta.scala
@@ -18,7 +18,6 @@ import org.joda.time.DateTime
 import org.scala_tools.time.Imports._
 import play.api.libs.json.{JsBoolean, JsString, JsValue}
 import play.api.mvc.RequestHeader
-import play.twirl.api.Html
 
 object Commercial {
 
@@ -31,7 +30,7 @@ object Commercial {
     )
   }
 
-  def make: model.Commercial = {
+  val empty: model.Commercial = {
     model.Commercial(
       isInappropriateForSponsorship = false,
       hasInlineMerchandise = false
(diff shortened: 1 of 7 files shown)


JonNorman:
[Illegibile rich links on paid containers](https://trello.com/c/IvVZm6Tq/149-illegibile-rich-links-on-paid-containers)

common/app/model/meta.scala
```diff
@@ -21,45 +22,26 @@ import play.twirl.api.Html
...
-  def make(metadata: MetaData, tags: Tags, apiContent: contentapi.Content): model.Commercial = {
-    make(metadata, tags, Some(apiContent))
-  }
-
-  def make(metadata: MetaData, tags: Tags): model.Commercial = {
-    make(metadata, tags, None)
-  }
-
-  def make(section: Section): model.Commercial = {
+  def make: model.Commercial = {
```
kelvin-chappell: This could be a `val` called `empty` or something, but perhaps it's not even needed.

kelvin-chappell (approved):
Just a small issue. But otherwise 👍

common/app/model/meta.scala
JonNorman: I've made it empty for now - if it can be removed then that can be done in a separate PR.

PR #14976, opened 2016-11-11
Add in a new AB test for International edition engagement banner

rupertbates:
## What does this change?
Adds in a new engagement banner test for users on the international edition.
They are now segmented into 2 groups: A control who will be shown the engagement banner after 10 visits and a second group who will be shown it from their first visit.

We hope to increase the number of membership sign ups by showing the engagement banner to a larger pool of users.
Web only
## Request for comment
@Ap0c


static/src/javascripts/projects/common/modules/commercial/membership-engagement-banner.js
```diff
@@ -187,6 +192,16 @@ define([
         return Promise.resolve();
     }
 
+    function userHasMadeEnoughVisits(edition) {
+        if (edition == 'INT') {
```
Ap0c: Is there a particular reason for using `==` instead of `===` here?

static/src/javascripts/projects/common/modules/commercial/membership-engagement-banner.js
```diff
@@ -187,6 +192,16 @@ define([
...
+            var internationalTestVariant = ab.testCanBeRun('MembershipEngagementInternationalExperiment') ? ab.getTestVariantId('MembershipEngagementInternationalExperiment') : undefined;
+            if (internationalTestVariant == '1st_article')
```
Ap0c: Should this also be `===`?

Ap0c (approved):
Minor comments about use of `==`. Apart from that, all looks good running locally 😄.

PR #14977, opened 2016-11-12
remove old .git/hooks using rimraf

SiAdcock:
## What does this change?

On a clean install of this repo `.git/hooks` is a directory, not a symlink. The `githooks` task throws (and catches) an error when it tries to unlink it, then throws an uncaught error when it tries to relink it.

screen shot 2016-11-12 at 12 43 43
_[Note: the output in this screenshot was added to demonstrate the error]_

This change fixes the problem by not assuming that `.git/hooks` will be a symlink. It uses the `rimraf` library to remove `.git/hooks` regardless of whether it is a symlink or a physical directory.

Installation no longer fails on fresh install.

No

## Screenshots

screen shot 2016-11-12 at 12 29 27

## Request for comment

@sndrs

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,6 +79,7 @@
     "postcss-reporter": "^1.4.1",
     "require-dir": "^0.3.0",
     "requirejs": "^2.3.2",
+    "rimraf": "^2.5.4",
     "sass-lint": "^1.9.1",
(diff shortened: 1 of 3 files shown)


sndrs (approved):
nice catch 🎈

PR #14980, opened 2016-11-14
Extend expiry of contributions epic

philwills:
## What does this change?

Extends the contributions epic for another few days

Continued high conversion rate for contributions and membership.

No


philwills:
I don't want to extend this further without moving this into a more permanent arrangement, rather than as an A/B test, but I think it's reasonable to keep it going for now.

dominickendrick:
I assume the switch expiry vs the test date are supposed to be different. If so 👍

philwills:
@dominickendrick One of them is inclusive, one of them is exclusive ❓ ❔ ❓

dominickendrick:
I guess it keeps you on your toes !!

PR #14981, opened 2016-11-14
Remove grunt clean

### LINKED ISSUE #14926: reopen #14866
depends on #14912

sndrs:
## What does this change?

removes grunt clean, fixes #14926

one more plugin gone

no

## Request for comment

@SiAdcock @gtrufitt @NataliaLKB @rich-nguyen


dev/teamcity/dist-npm-tc
```diff
@@ -6,5 +6,3 @@ set -o errexit
 
 echo "Install dependencies"
 make install
-
-./grunt-tc clean
```
sndrs: all individual tasks clean themselves before hand now. the last remaining clean task this ran was githooks, which are tidied and maybe installed, although explicitly not on TC now, by the githooks task

SiAdcock:
Nice, `rimraf` all the things! 👍

tools/__tasks__/compile/conf/clean.js
```diff
@@ -0,0 +1,9 @@
+const path = require('path');
+const rimraf = require('rimraf');
+
+const {root} = require('../../config').paths;
+
+module.exports = {
+    description: 'Clear template rendering artefacts',
+    task: () => rimraf.sync(path.resolve(root, 'common', 'conf', 'assets'))
```
rich-nguyen: I wish it was really obvious that this dir was the output of a copy task, an svg build task and one-half of an uglify task. The folder org in `__tasks__` gives you a much better chance of seeing the dependencies, so excellent work!

My thought just now was that every task should have a clean and a build, ideally. For example, every `index.js` has both a `task` and a `clean` key. Such ideals probably won't survive in our codebase though.

rich-nguyen:
👍

tools/__tasks__/compile/conf/clean.js
sndrs: yeah i agree with this – once this part is complete, i'm going to look at exactly what happens in the tasks, it's exposing a few oddities like that

sndrs: > My thought just now was that every task should have a clean and a build, ideally.

not sure, i think the less prescriptive it is, the better. certainly would ecnourage tasks discretely preparing their own ground. i guess the grunt set up was the result of grunt tasks being per-plugin, rather than per-task

PR #14982, opened 2016-11-14
Attach Raven to window object

SiAdcock:
## What does this change?

Currently Raven is a singleton which is configured and installed within the standard bootstrap. Any AMD bundle that requires the Raven singleton always gets the same instance.

In the near future, we will be gradually migrating the client side application to Webpack. During the migration process there will temporarily be two client side applications running side by side. They will not be able to share the same instance of the Raven singleton. 

This change adds the Raven instance to the window object, allowing the two applications to share the same instance.

Groundwork to enable migration of client side application bundler from `r.js` to Webpack

No

## Request for comment

@guardian/dotcom-platform @regiskuckaertz @rich-nguyen


static/src/javascripts/boot.js
```diff
@@ -33,6 +35,11 @@ define([
     var guardian = window.guardian;
     var config = guardian.config;
 
+    // global variable for sharing modules across bundles
+    guardian.app = {
```
regiskuckaertz: Just in case, you might want to make sure there is not an existing `app` key already.

```
guardian.app = guardian.app || {};
```

static/src/javascripts/bootstraps/enhanced/main.js
```diff
@@ -29,6 +27,8 @@ define([
     sport
 ) {
     return function () {
+        var raven = window.guardian.app.raven;
```
regiskuckaertz: don't we want to have a `raven` module similar to `config`? This to avoid touching the `window` object.

regiskuckaertz (commented):
Just minor questions.

static/src/javascripts/components/raven-js/raven.js
```diff
@@ -1410,7 +1410,14 @@ var Raven = {
...
         return lastEventId;
-    }
+    },
+
+    /*
+     * Has raven been configured?
+     *
+     * @return {boolean}
+     */
+    isSetup: isSetup
```
rich-nguyen: This is a third party script, are we going to allow modifying them? I pity the developer who will manage our dependencies in the future. Another example is here: https://github.com/guardian/frontend/commit/c6202a8926e9da0ff132de77edf56fe40e8a93bb

rich-nguyen: Maybe another developer line in the sand is needed.

rich-nguyen:
@regiskuckaertz 's suggestion of a `config` like helper seems sensible. 👍

static/src/javascripts/boot.js
sndrs: i might explicitly put it on `window.` too, just for clarity

static/src/javascripts/components/raven-js/raven.js
rich-nguyen: if it's compatible, then you could always take the new `raven-js` version, and codify the dependency. The rationale is fine, it just might mislead a developer who is wondering how we are accessing new api on an older module!

SiAdcock: Good shout. I'll experiment with the latest version (3.8.0) and if it's compatible I'll upgrade it. We're still using version 1.1.6

SiAdcock:
Thanks for your suggestions. I have made the following changes:

- upgraded to [Raven v3.8.0](https://github.com/getsentry/raven-js). This is supported by our Sentry server. Since Raven is only used on the client side in about 12 places, it was relatively easy to ensure our code is still calling it correctly
- added a Raven helper module that points to the [Raven instance on the global object](https://github.com/guardian/frontend/pull/14982/files#diff-00e692fbc2a1574721e7f591acd1777cR4). This keeps the reference to `window` out of most of our app
- boot.js [attaches an app object to the window object](https://github.com/guardian/frontend/pull/14982/files#diff-9a26962d007e34390bca3994560cd81dR39). This stores a reference to raven (and in the future will store references to other shared singleton objects such as `mediator` and Ophan). The boot script will not overwrite `app` if it already exists.

static/src/javascripts/boot.js
```diff
@@ -33,6 +35,11 @@ define([
     var guardian = window.guardian;
     var config = guardian.config;
 
+    // global variable for sharing modules across bundles
+    window.guardian.app = window.guardian.app || {
```
sndrs: it's nit picky, but i wonder if should just define these all in one place? i.e. wherever `window.guardian` is defined? not a blocker though

static/src/javascripts/projects/commercial/modules/dfp/init.js
```diff
@@ -20,10 +20,21 @@ define([
     'commercial/modules/messenger/viewport',
     'commercial/modules/messenger/click',
     'commercial/modules/messenger/background'
-], function (Promise, qwery, bonzo, raven, config, fastdom, commercialFeatures,
-             buildPageTargeting, dfpEnv, onSlotRender, onSlotLoad, PrebidService,
-             ophanTracking) {
-
+], function (
```
sndrs: ❤️

static/src/javascripts/projects/common/utils/raven.js
```diff
@@ -0,0 +1,5 @@
+/*global guardian:false */
+
+define(function () {
+    return guardian.app.raven;
```
sndrs: `window.guardian`? i kind of like the idea of being mega-explicit when we're touching the global object

sndrs:
any reason not to install `raven` as a yarn dep?

SiAdcock:
Hi @sndrs, the trouble with adding Raven with yarn is that it ends up in `node_modules`. The way we have set up our curl config, I believe all our JS needs to be in the `static` directory. Happy to discuss further if you have any ideas on how to restructure.

I have moved the Raven config from the standard bootstrap to the new Raven util as it seems to make more sense there. The Raven util attaches Raven to the global object and initialises it if it hasn't already.

sndrs:
> the trouble with adding Raven with yarn is that it ends up in node_modules. The way we have set up our curl config, I believe all our JS needs to be in the static directory.

sounds like a good reason to leave it alone for now then :)

sndrs (approved):
lgtm - great stuff

PR #14983, opened 2016-11-14
Include section sponsorships in pressed content

kelvin-chappell:
Currently only tag sponsorships are stored in the metadata of pressed pages.  So we lose any section sponsorships that have been set up.

This is a follow-up to #14973

THIS WILL BREAK ALL FRONTS until they have been re-pressed so needs to be deployed carefully!


rich-nguyen:
you could add it in a way that means you don't break the format, i suppose. `section` would be an optional string, and the new field would be `sectionSummary`. Then remove `section` at a later date safely, since it's optional.

But yes 👍

kelvin-chappell:
Yes, I think I might break it into two steps as this deployment is going to be tricky.

kelvin-chappell:
Please see #14987

PR #14984, opened 2016-11-14
Ensure we don't show the US engagement banner to users who have already contributed

rupertbates:
## What does this change?
Stops the US engagement banner (which now asks for contributions) being shown to users who have already contributed.

It is just a courtesy measure for users who have already contributed.

Web only

## Request for comment
@jranks123


jranks123 (commented):
👍

PR #14985, opened 2016-11-14
remove the sharing line from the thank you component

jranks123:
## What does this change?

The line on the thank you component asking readers to share a link did not result in many contributions or supporter sign ups so we have decided to remove it and just show the thank you message. 

Move warm fuzzinesss for our readers, without being tainted by being asked to do something. 

## Screenshots

BEFORE: 

screenshot at nov 11 15-46-53

NOW:

a


paperboyo:
❤️ We shouldn’t have to do tests to know that it was tainted...

PR #14987, opened 2016-11-14
Include section sponsorships in pressed content

kelvin-chappell:
This is a follow-up to #14973 and #14983

Currently only tag sponsorships are stored in the metadata of pressed pages. So we lose any section sponsorships that have been set up.

This PR should allow fronts to be re-pressed safely without breaking anything.
The result should be that the pressed page will hold the data but the rendered page won't use it yet.

### DIFF
--- a/common/app/model/Formats.scala
+++ b/common/app/model/Formats.scala
@@ -63,7 +63,9 @@ object MetaDataFormat {
     id: String,
     url: String,
     webUrl: String,
+    // todo remove section property in another PR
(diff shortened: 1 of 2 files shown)


kelvin-chappell:
I'm taking the thumbs up for #14983 as thumbs up for this too!

rich-nguyen:
👍 , but i think it may have helped to make `section` an `Option`, so that it is resilient to being pressed with the field missing.

PR #14989, opened 2016-11-14
Fix the standfirst font on richlinks within paid content

JonNorman:
In #14972, rich links that are within paid-content were given a bit of a makeover. However the `.rich-link__standfirst` class overrides the font of standfirsts to be serif regardless of context. As a result, I have to override the font on `rich-link__standfirst` specifically when within a `.paid-content--advertisement-feature`.


annebyrne:
👍

PR #14991, opened 2016-11-14
Engagement banner tidy up

rupertbates:
## What does this change?
Refactors the membership engagement banner to remove a whole bunch of code which is no longer needed.

Housekeeping, removes redundant code and simplifies what remains

Nope

## Request for comment
@JustinPinner @Ap0c


static/src/javascripts/projects/common/modules/commercial/membership-engagement-banner.js
```diff
@@ -1,210 +1,180 @@
...
+        };
+
+        // change messageCode to force redisplay of the message to users who already closed it.
+        // messageCode is also consumed by .../test/javascripts/spec/common/commercial/membership-engagement-banner.spec.js
+        var messageCode = 'engagement-banner-2016-11-10';
+        var notInTest = 'notintest';
+
+        var messages = {
+            UK: {
+                campaign: 'Coffee_49',
```
JustinPinner: Ideally, this would probably just be `'mem_uk_banner'` and be varied by an active test that changes things, to make it easier to spot the differences in the data. I don't know if case-sensitivity might also be an issue for data lake queries. As this code is already out, we'll have to leave it until any tests have completed.

The same would apply for the US message (below) too.

JustinPinner:
Looks sensible to me. Nice to see the complexities reduced 👍

PR #14996, opened 2016-11-15
Read section summary when rendering pressed content

kelvin-chappell:
The final step! 

As a result of #14987 we're now pressing section-level branding data but not using it during rendering.  This PR uses the data during rendering.

Facia has to be deployed before facia-press for this to work.


rich-nguyen:
👍

PR #14997, opened 2016-11-15
Upgrades facia client to 2.0.4

DiegoVazquezNanini:
@TBonnin @JonNorman @guardian/guardian-frontend 
I'm upgrading the facia-client to support a new special.


TBonnin (approved):
👍  Make sure you redeploy both facia and facia-press :)

PR #15001, opened 2016-11-15
Remove grunt shell

sndrs:
## What does this change?

- removes grunt shell
- settles `tools/ab-test-info/ab-test-info.js` into be a bit more permanent (e.g. doesn't take args, since only called in one place)

removes the craziest of all grunt plugins

no

## Screenshots

build is identical to master:

screen shot 2016-11-15 at 13 46 12

## Request for comment

@NataliaLKB @gtrufitt @SiAdcock @rich-nguyen


NataliaLKB (approved):
👍

PR #15003, opened 2016-11-15
move asset monitor to bash script in the repo

sndrs:
## What does this change?

moves the call to `grunt assetmonitor` from teamcity admin to the main repo (will update TC once merged)

nearly missed that this was being called when grepping for remaining grunt tasks

no

## Request for comment

@TBonnin

### DIFF
--- a/dev/teamcity/dist-assetmonitor-tc
+++ b/dev/teamcity/dist-assetmonitor-tc
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+set -o xtrace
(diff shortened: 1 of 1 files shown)


TBonnin:
where was it called before?

sndrs:
directly in TC build step https://teamcity.gu-web.net/admin/editBuildRunners.html?id=buildType:dotcom_master

TBonnin (approved):

PR #15013, opened 2016-11-15
Hosted thrasher - fix invalid flex values

lps88:
Invalid CSS is breaking the layout in IE11:

image

cc @guardian/labs-beta @guardian/commercial-dev


lps88:
Template is no longer being used

Calanthe:
👍

PR #15014, opened 2016-11-15
adds a placeholder attribute to the amp interactive embeds. This allows them to display in the top part of the page without breaking validation.

michaelwmcnamara:
This change adds an extra attribute to the amp-embed-cleaner for interactive-elements.
Currently, like all amp-iframes, these elements are only allowed to display least 600px or 75% of the first viewport away from the top of the page (See https://www.ampproject.org/docs/guides/iframes for full details).
 This causes validation to fail if the interactive is embedded as main-media. This was a problem for the US election liveblog and consequently this page was not surfaced as an amp-page in Google search - see:
https://trello.com/c/39OgyxMK/149-amp-omit-embeds-from-main-media

This change takes advantage of a loophole to get around this limitation. By attaching a placeholder attribute (even if left empty)  it allows the interactive to be displayed near the top of the page without failing validation.

This change should only effect Amp pages - specifically those that contain embeds near the top of the page.

The Amp test suite and embed cleaner unit tests all passed.


michaelwmcnamara:
There is a second piece to this puzzle, which is getting the dimensions for the interactive when it is rendered - hence the switch. This is a seperate issue and is currently being explored.

SiAdcock:
That's a weighty PR title! 😄

SiAdcock:
What happens if you don't supply dimensions for the interactive? Does it render a tiny `iframe` with scrollbars or something?

michaelwmcnamara:
tldr: 
The second piece of the puzzle bit is only about the election results interactive - other interactives work fine. Therefore not really relevant to the pull request (I probably shouldnt have mentioned it here).

long explanation:
As far as I can tell there are 2 types of interactives - iframes and boot.js (See here for examples: https://visuals.gutools.co.uk/feed/).
Most of the interactive embeds in articles are of the iframe type. These have a preset width of 5 and height of 1, but are resized by the amp-iframe to the correct dimensions.
The interactives of type boot.js are - as far as I can tell - intended for their own page and these pages are not supported on amp.
However in the case of the election interactive, the interactive is of the boot.js type, but it has been embedded in an article page. 

The second piece of the puzzle bit is  that I can't find any other examples of boot.js interactives embedded in articles, so I need to find out if this is an exception to the norm, or something we will see again, and if so how to handle it.

SiAdcock:
Cool thanks for the explanation @michaelwmcnamara. This change looks 👍 

There's some more detail about the different types of interactives [in the docs](https://github.com/guardian/frontend/blob/master/docs/03-dev-howtos/05-interactives.md)

PR #15018, opened 2016-11-16
Include owners of expiring switches in email

kelvin-chappell:
If only the email properties were populated, this could be quite useful.
(That can be worked out later.)


JonNorman (approved):
So this explicitly adds the expiring-switch owners' emails to the recipients lists of the emails? Sounds good to me!

PR #15024, opened 2016-11-16
Change capitalisation of scroll to comments event

jorgeazevedo:
## What does this change?

Capitalisation of scroll to comments event

We're moving to all lower case event categories, actions and labels to improve GA reporting.

It doesn't really

## Request for comment

@dominickendrick got started with this one, looks good?


dominickendrick:
👍

PR #15025, opened 2016-11-16
Make the stickiness of sticky ad stick at navigation instead of bottom of header

gtrufitt:
*Discussion-starting PR*

## What does this change?

When scrolling, the sticky ad will stick to the top of the navigation, instead of to the bottom of the header.

This improves the UX as the user doesn't have to scroll ages to use the navigation - an oft-seen complaint. Makes the site look less broken.

### Before
sticky-before

### After 

sticky-after

## Request for comment

@guardian/dotcom-platform @guardian/commercial-dev @patstirling @regiskuckaertz @ScottPainterGNM 

Let's discuss. There's ad viewability concerns - how can we test this in DFP? Can we pull in AB tests?


johnduffell:
sticky nav is back!! 🎉 😨

stephanfowler:
@johnduffell it's not a sticky nav - it's just preventing the nav from disappearing under the ad.

regiskuckaertz:
Gratifying that it was easy for you to see where and how to update the module 😎 

Changes LGTM, will see what @JamieKMorrison thinks about it.

JamieKMorrison:
We should be able to test the impact on viewability - I can add the name of the A/B test and variants to DFP and then we can report on it. The worry is that it is already a poorly performing slot, so ideally we wouldn't want to be making viewability worse. But a test is the only way to assess what the impact would be (and I totally agree that it solves an annoying UX issue).

My challenge would be whether you can come up with a way of solving the UX problem without reducing the viewability of the slot. (But definitely test this - just let me know what the test and variants are called so I can add them to DFP. And then we can figure out what will constitute statistical significance)

gtrufitt:
Added the A/B test to segment for reviewing viewability if I can get a 👍 pls

NataliaLKB:
👍

static/src/javascripts/projects/common/modules/experiments/tests/platform-sticky-ad-viewability.js
```diff
@@ -0,0 +1,31 @@
+define([], function (
+) {
+    return function () {
+        this.id = 'PlatformStickyAdViewability';
+        this.start = '2016-11-17';
+        this.expiry = '2016-11-24';
+        this.author = 'Gareth Trufitt';
+        this.description = 'Test the affect of showing the navigation on the viewability of the sticky top ad';
+        this.audience = 0.05; // (5%) - 8M ad impressions, 4% MDE at 50% baseline conversion is 8,100 - a minimum of 1% for a day for SS
```
stephanfowler: Wow. I resign. You can have my job.

sndrs:
opting in

PR #15019, opened 2016-11-16
Fix UK membership engagement message experiment

JustinPinner:
## What does this change?
Prevents the text from being eliminated for the control group. Redefines the test with new name, start date, variant names (and by extension, campaign codes) and master switch.

Also fixes the typo in the "Not got round..." message. It now reads "Not got around..." 

## What is the value of this and can you measure success?
The error rendered the data from at least one day unreliable. We'll have to restart the test and collect fresh numbers.
 
## Does this affect other platforms - Amp, Apps, etc?
No, web only.

## Screenshots
screen shot 2016-11-16 at 12 02 27
--
screen shot 2016-11-16 at 12 03 24
--
screen shot 2016-11-16 at 12 04 16
--
screen shot 2016-11-16 at 12 05 04
--
screen shot 2016-11-16 at 12 05 59

## Request for comment
@rupertbates @Ap0c @rtyley @svillafe

PR #15026, opened 2016-11-16
Assign winning text to default UK engagement banner & remove test.

JustinPinner:
## What does this change?
Uses the winning text from the AB test as the new default.

Now that we've removed the test, there is no need for https://github.com/guardian/frontend/pull/15019 so I'll close that.

During the test we noticed a +32% effect on conversions as a result of using this text.

No, web only.

## Screenshots
screen shot 2016-11-16 at 16 58 45

## Request for comment
@rupertbates @rtyley @svillafe @Ap0c


JustinPinner:
cc @JuliaBellis

rupertbates (approved):
👍

PR #15028, opened 2016-11-17
Add dimensions to hosted logo model

kelvin-chappell:
Because in amp pages logos need to have dimensions.

/cc @Calanthe


lps88:
👍

PR #15029, opened 2016-11-17
Platform meta tweaks

gtrufitt:
## What does this change?

Fixes some styling bugs with better meta:

- Interactive template fixes save for later and some spacing issues, removing some interactive overrides

image

- Gallery template fixes spacing

image

- Sets min-width as per the width of the icon

It ain't broken.

## Request for comment

@zeftilldeath @NataliaLKB @SiAdcock


dominickendrick:
👍

PR #15030, opened 2016-11-17
Fix unclear error in CSS task failure

sndrs:
## What does this change?

today I learned from @regiskuckaertz that promises can both `reject()` and `resolve()`, so some of the promises I've written were not branching as i thought they would. this makes those success/failure branches explicit. thanks @regiskuckaertz!

badly formatted sass now produces a more useful error in compilation

no

## Screenshots

screen shot 2016-11-17 at 11 42 15

## Request for comment

@regiskuckaertz @zeftilldeath

### DIFF
--- a/tools/__tasks__/compile/css/transpile-sass.js
+++ b/tools/__tasks__/compile/css/transpile-sass.js
@@ -18,16 +18,26 @@ const renderSass = (filePath, dest) => new Promise((resolve, reject) => {
         file: filePath,
         outFile: dest
     }, options), (err, result) => {
-        if (err) reject(err);
-        resolve(result.css.toString());
+        if (err) {
+            reject(err);
+        } else {
+            try {
+                resolve(result.css.toString());
+            } catch (e) {
+                reject(e);
+            }
+        }
     });
 });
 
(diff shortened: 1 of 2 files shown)


tools/__tasks__/compile/css/transpile-sass.js
```diff
@@ -18,16 +18,26 @@ const renderSass = (filePath, dest) => new Promise((resolve, reject) => {
         file: filePath,
         outFile: dest
     }, options), (err, result) => {
-        if (err) reject(err);
-        resolve(result.css.toString());
+        if (err) {
+            reject(err);
+        } else {
+            try {
```
regiskuckaertz: 👏  never too prudent

regiskuckaertz:
👍

SiAdcock:
Out of interest, why not `return reject(err)`? I think returning something meaningful from a function where possible is more idiomatic node than implicitly returning `undefined`. Also negates the need for the `else` section

sndrs:
yeah that came up on slack and i agree with you – i'd merged before that discussion, but i might change it

PR #15033, opened 2016-11-17
remove analyticsName from meta model

gidsg:
## What does this change?
Remove `analyticsName` from the Scala metadata model as it was only used by Omniture.

Note, there are still references to `pageName` within the Hosted model which I believe is probably only for Omniture. Also, there are omniture calls within the static error pages, neither of which are changed by this PR.

🔥 🔴 

No

## Request for comment
CC @mkopka @dominickendrick


rich-nguyen:
👍

rich-nguyen:
Have all js references to `analyticsName` already been removed?

rich-nguyen:
https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/common/modules/analytics/interaction-tracking.js#L68

PR #15034, opened 2016-11-17
Auto-install deps when running make tasks that need them

sndrs:
## What does this change?

- updates `make` tasks to install deps on demand
- removes the prompt to `make install` if deps change

since `yarn` takes less than a second to run if it has nothing to do, its not big overhead if we just run it before every `make` command that relies on them, and stop prompting people to install when they've changed.

in this way, `make` behaves more like SBT, and kind of treats the presence of dependencies as a dependency itself...

screen shot 2016-11-17 at 12 52 05

## Request for comment

I wonder if this is too 'magic'? maybe it's useful?

@guardian/dotcom-platform @regiskuckaertz @rich-nguyen

### DIFF
--- a/git-hooks/post-merge
+++ b/git-hooks/post-merge
@@ -1,8 +1,0 @@
-#/usr/bin/env bash
-
-# Modified version of [Sindre Sorhus's gist](https://gist.github.com/sindresorhus/7996717),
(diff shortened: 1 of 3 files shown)


NataliaLKB:
Awesome idea and work 👍

PR #15037, opened 2016-11-17
Track paid card logo impressions

kelvin-chappell:
This changes GA custom dimension 31 of a page load event to include sponsors' logos on individual cards, as well as page logos and container logos.
I don't expect this will have much effect on tracking that's already taken place, as there aren't that many containers with individual card logos at the moment.

This is a follow-up to #14958

/cc @guardian/commercial-dev @gtrufitt


rich-nguyen:
👍

PR #15039, opened 2016-11-17
Add fluid size support in Facebook IA

regiskuckaertz:
Native ads are coming through dotcom, the Android/iOS apps and AMP pages. Next in line is Facebook IA. This change allows us to:

- use the `fluid` size in the call to `defineSlot`
- specify an ad test value, useful to render specific creatives coming from DFP

... and here is a screenshot(tm)

picture 3

(it is fluid, that's why it takes up all the space)

NB: I may not get all of it, but boy this script does things the complicated way. For once, I don't get how it is possible to have more than one ad slot, huh?

cc @LATaylor-guardian

### DIFF
--- a/facia/public/facebook_ia_remote.html
+++ b/facia/public/facebook_ia_remote.html
@@ -127,7 +127,7 @@
         }
 
         function buildAdtestTargeting() {
-            var adtest = slot.getAttribute('data-adtest');
+            var adtest = slot.getAttribute('at');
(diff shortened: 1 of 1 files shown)


janua:
:+1:

regiskuckaertz:
[Native ads in Facebook IA part 1: frontend proxy](https://trello.com/c/P7s2IWZn/184-native-ads-in-facebook-ia-part-1-frontend-proxy)

PR #15041, opened 2016-11-18
Attach mediator to the window object

SiAdcock:
## What does this change?

Currently the mediator is a singleton instance of `EventEmitter`. Any AMD bundle that requires the mediator always gets the same instance.

In the near future, we will be gradually migrating the client side application to Webpack. During the migration process there will temporarily be two client side applications running side by side. They will not be able to share the same instance of the mediator singleton.

This change adds the mediator instance to the window object, allowing the two applications to share the same instance.

Groundwork to enable migration of client side application bundler from `r.js` to Webpack

No

## Request for comment

@guardian/dotcom-platform @rich-nguyen @regiskuckaertz 

## Related changes

#14982: Attach Raven to window
guardian/ophan#1845: Attach Ophan to window
#14946: Bundle standard using Webpack


NataliaLKB:
👍

PR #15042, opened 2016-11-18
Enable images in hosted pages to be rendered in amp pages

kelvin-chappell:
This will transform img elements into amp-img elements.

Use it in a play scala template like this:
```
@import model.hosted.HostedAmp.ampify
...
    { @Html(ampify(page.body)) }
...
```
ie. replace `page.body` with `ampify(page.body)` where you want it to render for amp.

/cc @Calanthe


Calanthe:
This looks ok to me but I thought that you will use the existing template https://github.com/guardian/frontend/blob/d4422b4537165424e70a898d150db4e806ba04d6/common/app/views/fragments/amp/ampImage.scala.html

kelvin-chappell:
I'll have a look.  The hosted model is quite simplified so it may need some coercion to fit it into that template.  But it would be good it use it if possible.  I'll keep this open for now while I investigate.

kelvin-chappell:
I think this solution is best for the moment.

Calanthe:
ok 👍

PR #15044, opened 2016-11-18
add world of work tag to tests

jranks123:
## What does this change?

The world of work series is proving very popular so we want to ask for contributions/supporter sign ups on these articles. This PR adds that tag. 

More contributions/supporters. 

## Request for comment

@jfsoul @NataliaLKB @gtrufitt

### DIFF
--- a/static/src/javascripts/projects/common/modules/experiments/tests/contributions-epic-limited-impressions-not-usa.js
+++ b/static/src/javascripts/projects/common/modules/experiments/tests/contributions-epic-limited-impressions-not-usa.js
@@ -53,14 +53,20 @@ define([
                 'us-news/us-politics'
             ];
 
+            var includedNonKeywordTagIds = [
+                'uk-news/series/the-new-world-of-work'
+            ];
+
             var excludedKeywordIds = ['music/leonard-cohen'];
 
             var hasKeywordsMatch = function() {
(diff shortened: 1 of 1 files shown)


static/src/javascripts/projects/common/modules/experiments/tests/contributions-epic-limited-impressions-not-usa.js
```diff
@@ -53,14 +53,20 @@ define([
...
+                'uk-news/series/the-new-world-of-work'
+            ];
+
             var excludedKeywordIds = ['music/leonard-cohen'];
 
             var hasKeywordsMatch = function() {
                 var pageKeywords = config.page.keywordIds;
-                if (typeof(pageKeywords) != 'undefined') {
+                var pageNonKeywordTagIds = config.page.nonKeywordTagIds;
+                if (typeof(pageKeywords) != 'undefined' && typeof(pageNonKeywordTagIds) != 'undefined') {
```
gtrufitt: Strict comparison here: `!==`

gtrufitt (approved):
This will skew your data I think as you're more likely to get non-regulars due to the test already having bucketed loyal users. This is why we change test ids. Worth checking if that's an issue for you here.

PR #15047, opened 2016-11-18
Fix last hosted article amp validation error

kelvin-chappell:
Removes redundant gu-atom elements and their parents from hosted body.

/cc @Calanthe


Calanthe:
👍

PR #15050, opened 2016-11-18
Fix: Add youtube to compiled bundle

markjamesbutler:
Fix to allow the youtube js to be loaded from the compiled assets.


gtrufitt:
👍

PR #15054, opened 2016-11-21
Https redirect tool updates

jfsoul:
## What does this change?
This carries on from [14393](https://github.com/guardian/frontend/pull/14393) to make sure redirects are https friendly.  Usage and instructions for the redirect tool now match again - i.e. you can request `https://www.theguardian.com/path` redirects and it will work as expected!  The protocol is no longer mandatory, but the instructions still state to use `https` version as I think it's good practice anyway.  There is also minimal feedback now in case an invalid format is used for the source of the redirect.

Finally it also simplifies the `ArchiveController` further by not expecting all archive requests to start with `www.theguardian.com` (as it's just noise) and using `LinkTo` for all redirects.  As a result custom redirects in the controller all go directly to `https`.

This PR will be accompanied by a change to router for the /404 archive route (although this should be able to go out on its own while we make the change).

The redirect instructions match the functionality and all redirects custom redirects in the `ArchiveController` now redirect to https directly.

No

## Screenshots
image
image

## Request for comment
@TBonnin


admin/app/controllers/RedirectController.scala
```diff
@@ -24,13 +24,13 @@ class RedirectController(redirects: RedirectService) extends Controller with Log
...
-    redirectForm.bindFromRequest().get.trim match {
+    val message = redirectForm.bindFromRequest().get.trim match {
       case PageRedirect(from, "") if from.nonEmpty  => redirects.remove(from)
       case PageRedirect(from, to) if from.nonEmpty  => redirects.set(PermanentRedirect(from, to))
-      case _ =>
+      case _ => "Redirect request failed, please enter a value for 'From'"
     }
 
-    SeeOther(routes.RedirectController.redirect().url)
+    Ok(views.html.redirects(redirectForm, urlMsgs = List(message)))
```
jfsoul: `urlMsgs` was already part of the template, but never used

archive/app/controllers/ArchiveController.scala
```diff
@@ -40,17 +39,17 @@ class ArchiveController(redirects: RedirectService) extends Controller with Logg
...
         // bounce these to the section
         case CombinerSectionRss(section)                => redirectTo(s"$section/rss", "combinerrss")
         case CombinerSection(section)                   => redirectTo(section, "combinersection")
         case Combiner(combiner)                         => redirectTo(combiner, "combiner")
-        case DatedSpecialIndexPage(section, rest, _)    => cachedArchiveRedirect(WithoutRevalidationResult(Redirect(s"${LinkTo(section)}$rest/all", redirectHttpStatus)))
-        case SectionSpecialIndex(section, _)            => cachedArchiveRedirect(WithoutRevalidationResult(Redirect(s"${LinkTo(section)}/all", redirectHttpStatus)))
-        case NewspaperPage(paper, date, book)           => cachedArchiveRedirect(WithoutRevalidationResult(Redirect(s"${LinkTo(paper)}/$book/$date/all", redirectHttpStatus)))
+        case DatedSpecialIndexPage(section, rest, _)    => redirectTo(section, "datedspecialindexpage", rest, "all")
+        case SectionSpecialIndex(section, _)            => redirectTo(section, "sectionspecialindex", "all")
+        case NewspaperPage(paper, date, book)           => redirectTo(paper, "newspaperpage", book, date, "all")
```
jfsoul: because of editionalisation oddities I wasn't able to do `redirectTo(s"$paper/$book/$date/all", "newspaperpage")` (for example)

archive/app/controllers/ArchiveController.scala
```diff
@@ -109,34 +108,40 @@ class ArchiveController(redirects: RedirectService) extends Controller with Logg
...
 
-  private def redirectTo(path: String, identifier: String)(implicit request: RequestHeader): Result = {
-    log.info(s"""Archive $redirectHttpStatus, "$identifier" redirect to $path""")
-    Cached(CacheTime.ArchiveRedirect)(WithoutRevalidationResult(Redirect(s"http://$path", redirectHttpStatus)))
+  private def redirectTo(path: String, identifier: String, pathSuffixes: String*)(implicit request: RequestHeader): Result = {
+    val endOfPath = if(pathSuffixes.isEmpty) "" else s"/${pathSuffixes.mkString("/")}"
+    val redirect = LinkTo(path) + endOfPath
+
+    log.info(s"""Archive $redirectHttpStatus, "$identifier" redirect to $redirect""")
+    Cached(CacheTime.ArchiveRedirect)(WithoutRevalidationResult(Redirect(redirect, redirectHttpStatus)))
```
TBonnin: Do we really need this `identifier` param which seems to only be used for logging

archive/test/ArchiveControllerTest.scala
```diff
@@ -34,167 +34,167 @@ import services.RedirectService.{ArchiveRedirect, PermanentRedirect}
...
 
   it should "not redirect a random URL that contains the word century" in {
-    val result = archiveController.lookup("www.theguardian.com/discover-culture/2014/jul/22/mid-century-textiles-then-and-now")(TestRequest())
+    val result = archiveController.lookup("/discover-culture/2014/jul/22/mid-century-textiles-then-and-now")(TestRequest())
     status(result) should be (404)
   }
 
-  it should "redirect failed combiners to the section" in {
-    val result = archiveController.lookup("www.theguardian.com/tv-and-radio/tvandradioblog+media/chris-evans")(TestRequest())
+  it should "redirect failed combiners to the (non-editionalised) section" in {
```
jfsoul: this appears to have been intended to test a non-editionalised front, but tv-and-radio recently got editionalised

common/app/filters/RequestLoggingFilter.scala
```diff
@@ -19,6 +19,7 @@ class RequestLoggingFilter extends Filter with Logging with ExecutionContexts {
           response.header.headers.get("X-Accel-Redirect") match {
             case Some(internalRedirect) => s" - internal redirect to $internalRedirect"
             case None => response.header.status match {
+              case 304 => " - 304 Not Modified"
```
jfsoul: I realised that any `304` responses were logging "redirect to [location not found]" which isn't that helpful.  There may be other exceptions where the `Location` header doesn't exist, but I think this is the main one.

common/app/services/RedirectService.scala
```diff
@@ -40,25 +40,47 @@ class RedirectService extends Logging with ExecutionContexts {
...
+      case PermanentRedirect(_, location) => pathOnly.map(PermanentRedirect(_, location))
+      case ArchiveRedirect(_, location) => pathOnly.map(ArchiveRedirect(_, location))
+    }
   }
+
+  def set(destination: Destination): String =
+    normaliseDestination(destination).map { dest =>
+      log.info(s"Setting redirect in: $tableName to: ${dest.source} -> ${dest.location}")
+      Scanamo.put(DynamoDB.syncClient)(tableName)(dest)
+      "Redirect successfully set"
```
TBonnin: Is this log really adding value?

common/app/services/RedirectService.scala
```diff
@@ -40,25 +40,47 @@ class RedirectService extends Logging with ExecutionContexts {
...
+    }.getOrElse("'From' value does not conform to correct format")
```
TBonnin: `'From'`?

archive/app/controllers/ArchiveController.scala
jfsoul: I can see that it would help to quickly diagnose which of the custom redirects in the controller was used.  Whether we ever really need that info - I'm not so sure.  It may have been useful when first adding these redirects, but I would be ok with removing now.

TBonnin: shouldn't logging the full path be enough?

common/app/services/RedirectService.scala
TBonnin: Ok. I am not sure the display logic should happen here though.

TBonnin (approved):

PR #15056, opened 2016-11-21
Sonobi 100 percent

janua:
## What does this change?

This adds a new `sonobiSwitch` to `CommercialSwitches`, turning it from a multi variate test to a switch. 

We now ignore the `X-GU-comm-hb-test` header from fastly, meaning we will serving at 100% Sonobi. This header will be removed in a separate PR for the edge cache.

Currently, if the `sonobiSwitch` is switched OFF, `Prebid` will run in the US at 100%.
`Prebid` is to be removed in a separate PR.

The other thing that seems to need another look is the performance measuring; when we are 100% Sonobi, the metrics no longer matter.

@rich-nguyen don't we already have metrics for the load time of the commercial application? I think that's probably all we need as we can't replace the live metrics with anything sensible unless we want to monitor the performance of Sonobi live. Seeing as we have that setup already we might as well do just that.

eCPM + 1

No, only web.

## Request for comment

@rich-nguyen


rich-nguyen:
Yeah let's think about how we can modularise the sonobi load in the commercial `bootstrap.js` file. That way it will appear in the commercial app performance measurements

rich-nguyen:
nice 👍

regiskuckaertz:
> Prebid is to be removed in a separate PR.

👏

regiskuckaertz:
@rich-nguyen @janua generally speaking, I find it annoying that we can't do basic diagnostics _in situ_. We should leverage `window.guardian.commercial` for that matter to let us answer some questions more easily than now. Is Sonobi enabled? What is the size of the creative in the right slot? What type of creative is running in the merchandising slot? etc.

janua:
@regiskuckaertz Great idea. Let's do that and have that the place for things like that. @rich-nguyen is currently looking at modularising the Sonobi stuff so he can probably add the `sonobiEnabled` there. Then we can add more when we start looking at and refactoring the measuring of the commercial application.

jbreckmckye:
@regiskuckaertz That's basically why I added it. Because life's too short to debug commercial.js by watching the network panel.

janua:
[Sonobi 100% in all regions](https://trello.com/c/QCzMdaS5/186-sonobi-100-in-all-regions)

PR #15057, opened 2016-11-21
Hosted article page - AMP validation

Calanthe:
## What does this change?
The hosted article page is finally passing the Amp validation.

## Request for comment
@kelvin-chappell


kelvin-chappell:
👍

PR #15060, opened 2016-11-21
Fixed background: remove layer promotion hack

regiskuckaertz:
Fixes a bug where the fixed `div` would be positioned according to its parent [which has a `transform` rule applied to it](http://meyerweb.com/eric/thoughts/2011/09/12/un-fixing-fixed-elements-with-css-transforms/).

EDIT: ha! forgot the shiny GIFs

Before:
fixed-background-2

After:
fixed-background

### DIFF
--- a/static/src/stylesheets/module/commercial/_creatives.scss
+++ b/static/src/stylesheets/module/commercial/_creatives.scss
@@ -847,9 +847,7 @@
 }
 
 .creative__background-parent {
-    transform: translate3d(0, 0, 0);
-    will-change: transform;
-    contain: strict;
+    contain: size layout style;
     position: absolute;
     top: 0;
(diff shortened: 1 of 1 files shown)


JonNorman (approved):

static/src/stylesheets/module/commercial/_creatives.scss
```diff
@@ -847,9 +847,7 @@
 }
 
 .creative__background-parent {
-    transform: translate3d(0, 0, 0);
-    will-change: transform;
-    contain: strict;
+    contain: size layout style;
```
regiskuckaertz: yes, because `paint` makes this element a containing block for its fixed positioned child, which cancels the intended effect.

PR #15061, opened 2016-11-21
Fix amp validity test fixture

kelvin-chappell:


Calanthe:
👍

jfsoul:
Do we have a test fixture file for this audi content already?  If not, make sure we include one!

PR #15062, opened 2016-11-21
Return 404 instead of 500 on no capi results round

MatthewJWalls:
## What does this change?

Crosswords to return 404 instead of 500 when no capi data found

Proper status codes

No

### DIFF
--- a/applications/app/controllers/CrosswordsController.scala
+++ b/applications/app/controllers/CrosswordsController.scala
@@ -5,7 +5,7 @@ import common.{Edition, ExecutionContexts, Logging}
 import conf.Static
 import contentapi.ContentApiClient
 import crosswords.{AccessibleCrosswordRows, CrosswordPage, CrosswordSearchPage, CrosswordSvg}
-import model.Cached.RevalidatableResult
+import model.Cached.{RevalidatableResult, WithoutRevalidationResult}
 import model._
 import org.joda.time.{DateTime, LocalDate}
 import play.api.data.Forms._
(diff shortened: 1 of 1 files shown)


MatthewJWalls:
Is this all that we need for this? I'm assuming some part of the router/fastly actually redirects to the particular 404 stub page, based on whether the status code was 404 (locally the page content is just empty if you go to a nonexistant crossword/article)

applications/app/controllers/CrosswordsController.scala
```diff
@@ -51,7 +51,7 @@ trait CrosswordController extends Controller with Logging with ExecutionContexts
 
 class CrosswordPageController(val contentApiClient: ContentApiClient) extends CrosswordController {
 
-  def noResults()(implicit request: RequestHeader) = InternalServerError("Content API query returned an error.")
+  def noResults()(implicit request: RequestHeader) = NoCache(NotFound)
```
TBonnin: You might want to cache the response, even for a very small amout of time to avoid ddos `Cached(CacheTime.NotFound)(NotFound)`

TBonnin (changes requested):

TBonnin:
Yep the router would redirect to the correct 404 page. :)

applications/app/controllers/CrosswordsController.scala
MatthewJWalls: Changed to use that!

TBonnin (approved):

PR #15031, opened 2016-11-17
Add epic usa cta test

jranks123:
## What does this change?

We think that just showing a contributions button on the epic in the USA may be more effective than both the contributions button and the become a supporter button, especially now with all the news in the USA about media organisations receiving increased levels of contributions. Before we make this call, we need to prove it. This test will inform that decision.

This also limits the ContributionsEpicLimitedImpressions test to everywhere but the USA. Once we have the results of this test, we can merge it back into one test. 

## What is the value of this and can you measure success?

We find the most profitable CTA for the USA in the epic . 

## Screenshots
screenshot at nov 17 12-01-36
screenshot at nov 17 12-02-28

PR #15063, opened 2016-11-21
Default Epic back to control message

philwills:
## What does this change?

Removes the epic contributions/subscriptions ask everywhere except the US, defaults to the control message, which was once again victorious and adds a supporter only call-to-action to the testing mix.

We aim to have a clear idea of whether asking for contributions, sign-ups to membership, or a mix of the two is the most valuable way of asking people to support the Guardian in the US.

No

## Request for comment

cc @jranks123


static/src/javascripts/projects/common/modules/experiments/tests/contributions-epic-usa-cta.js
```diff
@@ -33,32 +33,36 @@ define([
 
     return function () {
 
-        this.id = 'ContributionsEpicUsaCtaFakeNews';
+        this.id = 'ContributionsEpicUsaCta';
```
jranks123: We have already named a test with this name (https://github.com/guardian/frontend/pull/15031/), so we should give it a distinctive name so it's easy to analyse the results

static/src/javascripts/projects/common/modules/experiments/tests/contributions-epic-usa-cta.js
```diff
@@ -244,14 +205,14 @@ define([
...
                 test: function () {
-                    var ctaType = cta.justContribute;
-                    var message = messages.fake;
+                    var ctaType = cta.justSupporter;
+                    var message = messages.control;
                     var component = $.create(template(contributionsEpicEqualButtons, {
-                        linkUrl1: ctaType.url1 + message.intcmp + '_contribute',
-                        linkUrl2: ctaType.url2 + message.intcmp + '_contribute',
+                        linkUrl1: ctaType.url1 + message.intcmp + '_supporter',
+                        linkUrl2: ctaType.url2 + message.intcmp + '_supporter',
```
jranks123: message.intcmp will always be "_control" in this test, so isn't needed for the campaign code for any of the variants

jranks123: Also, the ctaType.url1 and ctaType.url2 for just supporter and just contribute can lose the "_backup" now, as they are no longer being used as backups

jranks123 (commented):
Looks good to me with comments, but get a dotcom pair of eyes as well 👍

philwills:
Thanks, @jranks123, I think I've addressed all those points, do you mind taking a second look?

katebee (approved):
Looks good to me 👀 👍

PR #15066, opened 2016-11-21
Remove Prebid (!)

janua:
## What does this change?

This removes the module `PrebidService.js` and all references.

`Prebid` is currently only used in the US edition. By the end of this week for the `US` and `CA` regions we are going to be using `Sonobi` 100% meaning `Prebid` will no longer be in use.

Better performing header bidding by using server side header bidding via Sonobi.

In terms of revenue performance, we are currently looking at the data that `Sonobi` provides over `Prebid`.

For performance, we are currently monitoring the performance of `waterfall`, `prebid` and `sonobi`, and can already see that `prebid` and `sonobi` perform equally the same.
The difference is that if we added more providers to `prebid` the performance would degrade, where with `Sonobi` it would (or should) stay the same.
Unfortunately because we are moving completely to `Sonobi` and will never add another provider to `prebid`, we will never be able to compare the performance of what `Sonobi` promises against `prebid`.

No, only web.

## Todo

- [ ] Update `header-bidding.md`

## Request for comment

@rich-nguyen @regiskuckaertz


regiskuckaertz (approved):
Happy to bid this one _adieu_

jbreckmckye:
You've deleted almost everything I did on Frontend...

Well, good riddance to Prebid. It leaked memory, couldn't handle multiple requests properly, and it exhausted the browser's HTTP connection pool. I never got around to measuring its power usage on mobile, but that can't have been great either. Just a shame it took a year to finally take the approach we all wanted in the first place...

regiskuckaertz:
@jbreckmckye hi Jimmy! Rest assured the spacefinder is there to live _forever_ 😉

janua:
[Remove Prebid](https://trello.com/c/ou592rMy/187-remove-prebid)

janua:
I am planning on merging this on Thursday once the dust has settled around Sonobi 100%.

janua:
Bye prebid 👋

PR #15067, opened 2016-11-21
Update guardian.page.host to https

markjamesbutler:
The host parameter below has not been updated since the move to https:

    guardian.page.host="http://www.theguardian.com" -> guardian.page.host="https://www.theguardian.com"

We use this parameter to set the origin parameter in the YouTube player [initialisation](https://developers.google.com/youtube/player_parameters#origin)

The actual config update is in the config file rather than this PR.


jfsoul:
I've been meaning to take a look at this 😄 

I'm not aware of (and I hope there isn't) any functionality/tests that rely on this being `http`.  I know `LinkTo` explicitly replaces `http://` with `https://`, but that will still work.

👍

PR #15071, opened 2016-11-22
Recommended for you test

davidfurey:
## What does this change?

A/B test adding a new container to the home front containing recommended content based on visitors browsing history

Demand test showed that users wanted this feature.  Showing people content that is picked specifically for them might increase visit frequency.  

No

## Screenshots

screen shot 2016-11-22 at 10 42 12

## Request for comment

@joelochlann


static/src/javascripts/projects/common/modules/experiments/tests/recommended-for-you.js
```diff
@@ -0,0 +1,157 @@
...
+
+            var request = fetch(endpoint, {
+                type: 'json',
+                method: 'post',
+                crossOrigin: true,
+                body: JSON.stringify(reqBody),
+                headers: { 'Content-Type': 'application/json' }
+            });
+
+            return request.then(response =>
```
joelochlann: I guess since we're using arrow function notation here we could use it wherever else we have a lambda in this file too?

static/src/javascripts/bootstraps/enhanced/common.js
```diff
@@ -192,10 +192,8 @@ define([
                 if (config.page.contentType !== 'Network Front') {
                     history.logSummary(config.page);
                 }
-
-                if (config.page.contentType === 'Video') {
-                    history.logHistory(config.page);
-                }
+                
+                history.logHistory(config.page);
```
joelochlann: Will this have any knock-on effect on video-related code (e.g. if it assumes that everything in the history is video)?

joelochlann: Also, it seems from glancing at `history.js` (correct me if I'm wrong) that given that the max history size is 50, this will effectively make the video history much shorter (because the vast majority of the most recent 50 will be non-video)

static/src/javascripts/projects/common/modules/experiments/tests/recommended-for-you.js
```diff
@@ -0,0 +1,157 @@
...
+    guardianLogo,
+    fetch
+) {
+    return function () {
+        this.id = 'RecommendedForYouRecommendations';
+        this.start = '2016-08-02';
+        this.expiry = '2016-12-23';
+        this.author = 'Joseph Smith';
+        this.description = 'Add a personalised container to fronts';
+        this.audience = 1;
```
joelochlann: I would set this to 0 initially so that we can deploy it live and test it by force-opting into the test without opting in any real users

common/app/conf/switches/ABTestSwitches.scala
```diff
@@ -27,6 +27,16 @@ trait ABTestSwitches {
 
   Switch(
     ABTests,
+    "ab-recommended-for-you-recommendations",
+    "Test personalised container on fronts",
+    owners = Seq(Owner.withGithub("davidfurey")),
+    safeState = On,
```
joelochlann: I would probably have `safeState = Off` here

joelochlann (commented):
Looks good!

static/src/javascripts/projects/common/modules/experiments/tests/recommended-for-you.js
joelochlann: ... as discussed, I may have jumped the gun there. If we can't use ES6 notation then I guess switch this back to ES5 rather than everything else to ES6! @guardian/dotcom-platform may have some useful info to add

joelochlann (approved):
Good stuff 👍 

I would get this deployed to prod and then distribute the URL with the opt-in hash around the team to get some eyes on it before we flip the switch and expose to real users

PR #15072, opened 2016-11-22
Fix ELBs names

TBonnin:
## What does this change?
ELBs names were hardcoded. Moving apps into the vpc leads to new elbs being created and this change broke the admin loadbalancer metrics page, troubleshooter, etc...
This patch is setting new ELB names

Working dashboards :)

## Request for comment
@jfsoul


jfsoul (approved):
hooray 🎉

PR #15073, opened 2016-11-22
Betterer meta

zeftilldeath:
Couple of tidy ups to the meta changes. 

When there was just comments or just shares showing we were sometimes getting the count not sitting flush to the edge. Gross huh!? 

before:
screen shot 2016-11-22 at 11 36 23

After:
screen shot 2016-11-22 at 12 06 29

Also gallery spacing was a bit whacky, as were the tones. 

Before:
screen shot 2016-11-22 at 12 01 34

After: 
screen shot 2016-11-22 at 12 00 42

I'll put my Nit comb away now.
article-2312124-1966b4b7000005dc-612_306x434

@gtrufitt ? X

### DIFF
--- a/static/src/stylesheets/module/content/_content.global.scss
+++ b/static/src/stylesheets/module/content/_content.global.scss
@@ -96,31 +96,9 @@
     text-align: left;
     float: left;
     min-width: 15px; // Min width is the width of the icons
-    padding-right: ($gs-gutter / 2) + 1px;
 
     @include mq($until: leftCol) {
         text-align: right;
-    }
-
-    @include mq(leftCol, wide) {
-        padding-right: 6px;
-    }
-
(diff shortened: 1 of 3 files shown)


gtrufitt (approved):
:taco:

paperboyo:
Not sure why only this one looks broken (other immersives are ok), widest BP, multiple browsers, breaks after JS kicks in: https://www.theguardian.com/books/2016/nov/26/best-books-of-2016-part-one

picture 79
(it maybe that the meta is not to blame here... but something is)

gtrufitt:
Yeah, that's one for whoever added:

```
#article p {
    margin: 0px;
}
```

to the `book-review-page-enhancer-embed` CSS that comes from `interactive.guim`

PR #15074, opened 2016-11-22
Fix history for pagination of comments in profile

nicl:
## What does this change?

Ensures browser back/forward work with pagination and tab switching on the Identity profile comments listing pages. E.g. pages like:

https://profile.theguardian.com/user/id/3024198

* better linking (linking to pages other than the first now works)
* better UX - can navigate as expected

It is difficult to measure the impact here. We could look at requests to these pages perhaps, but I'm not keen on an AB test as this PR is a bug fix more than anything.

No.

## Request for comment

The code is pretty messy here. I found it difficult to work with the existing code to be honest and using the browser history API is also complex. All advice on how to improve things welcome. I'm obviously not very familiar with things like bonzo, qwery, or the component approach that Frontend takes here. Looking at the git history this code hasn't really been touched for 2 years and I'm not sure the people who worked on it are around any more.

cc @gtrufitt

### DIFF
--- a/static/src/javascripts/projects/common/modules/discussion/activity-stream.js
+++ b/static/src/javascripts/projects/common/modules/discussion/activity-stream.js
@@ -64,7 +64,7 @@ define([
         this.removeState('loading');
 
         var activeTab = $('.tabs__tab--selected');
-        if (activeTab.data('stream-type') != streamType) {
+        if (activeTab.data('stream-type') !== streamType) {
            selectTab(streamType, this.defaultOptions.streamType);
         }
 
@@ -101,7 +101,7 @@ define([
         var from = $('.tabs__tab--selected');
         var to = $('a[data-stream-type=' + streamType + ']');
 
-        if (to.length == 0) {
+        if (to.length === 0) {
             to = $('a[data-stream-type=' + fallback + ']');
         }
 
--- a/static/src/javascripts/projects/common/modules/discussion/activity-stream.js
+++ b/static/src/javascripts/projects/common/modules/discussion/activity-stream.js
@@ -65,7 +65,7 @@ define([
 
         var activeTab = $('.tabs__tab--selected');
         if (activeTab.data('stream-type') !== streamType) {
-           selectTab(streamType, this.defaultOptions.streamType);
+           selectTab(streamType === 'comments' ? 'discussions' : streamType);
         }
 
         // update opts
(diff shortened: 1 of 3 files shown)


static/src/javascripts/projects/common/modules/discussion/activity-stream.js
```diff
@@ -59,5 +97,20 @@ define([
...
+
+        if (to.length == 0) {
+            to = $('a[data-stream-type=' + fallback + ']');
+        }
+
+        from.removeClass('tabs__tab--selected');
+        var link = $('a', from);
+        link.blur();
+
+        bonzo(to).parent().addClass('tabs__tab--selected');
```
gtrufitt: I think we could avoid doing the DOM jumping here by adding the stream type to a class on the `li`, that way you can target it directly. Minor though, don't know if the mark-up for this is somewhere horrible.

static/src/javascripts/projects/common/modules/discussion/activity-stream.js
```diff
@@ -37,15 +47,43 @@ define([
...
-            });
-            this.removeState('loading');
-        }.bind(this));
+        $.create(html).each(function (el) {
+            $el.html($(el).html()).attr({ 'class': el.className });
+        });
+        this.removeState('loading');
+
+        var activeTab = $('.tabs__tab--selected');
+        if (activeTab.data('stream-type') != streamType) {
```
gtrufitt: We try and use strict comparison everywhere `!==`

static/src/javascripts/projects/common/modules/discussion/activity-stream.js
nicl: Okay I've updated this to avoid referring to the styling class for behaviour.

static/src/javascripts/projects/common/utils/url.js
```diff
@@ -44,6 +44,23 @@ define([
                 }
             },
 
+            // equivalent to pushQueryString but uses history.replaceState to
+            // overwrite history
+            replaceQueryString: function (params) {
```
nicl: I've refactored to make the methods share the common code, which should protect a bit against errors. I'm not sure there is much to test here, as it's all pretty imperative. Perhaps that the defaults are applied correctly? But how much value is added?

gtrufitt (approved):
LGTM

PR #15076, opened 2016-11-22
Change the safeState of header bidding switch to On

janua:
The changes the `safeState` of `sonobiSwitch` to `On`.

@rich-nguyen


rich-nguyen:
why? Isnt it safer off by default?

janua:
@rich-nguyen Actually, you are right, it is safer `Off`.

PR #15080, opened 2016-11-22
Remove unnecessary wrapper for Youtube embed

lps88:
## What does this change?
The iframe for embedded Youtube videos is no longer wrapped in a div when the page is initialised.

Remove some code, and avoid some styling complications around css classes that were set on both the wrapper and iframe.

## Request for comment
@gidsg @kelvin-chappell


gidsg (approved):
👍 🌟 CC @markjamesbutler

markjamesbutler (approved):

PR #15081, opened 2016-11-22
Modularise sonobi

rich-nguyen:
I've separated `commercial/modules/dfp/init` into:

- `commercial/modules/dfp/prepare-sonobi-tag`
- `commercial/modules/dfp/prepare-googletag`

because it allows us to clearly measure the impact of one module on the other, now that sonobi is 100%. The sonobi has to finish loading before the googletag module can start. This PR makes sure our commercial timing system reflects the offset of googletag's execution caused by sonobi.

I've removed the ophan commercial advert tracking, it is not used anymore, can't be viewed, and will save us some aws running cost, especially for @obrienm 's team.

I've also renamed `commercial/modules/dfp/load` to `commercial/modules/dfp/fill-advert-slots`, for more descriptive analytics reporting.


static/src/javascripts/projects/commercial/modules/dfp/performance-logging.js
```diff
@@ -0,0 +1,132 @@
...
+        var timerStart = userTiming.getCurrentTime();
+        performanceLog.modules.push({
+            name: moduleName,
+            start: timerStart
+        });
+    }
+
+    function moduleEnd(moduleName) {
+        var timerEnd = userTiming.getCurrentTime();
+        var module = performanceLog.modules.find(function(module){
```
regiskuckaertz: `Array.prototype.find` needs a shim in older browsers. Do we provide it?

static/src/javascripts/projects/commercial/modules/dfp/ophan-tracking.js
```diff
@@ -1,205 +0,0 @@
...
-                    4: 'receive',
-                    6: 'render'
-                },
-                lifecycleIdToAdvertTiming = {
-                    3: 'dfpFetching',
-                    4: 'dfpReceived',
-                    6: 'dfpRendered'
-                };
-
-            googletag.debug_log.log = function interceptedGptDebugger(level, message, service, slot) {
```
regiskuckaertz: Good riddance 🚤

static/src/javascripts/projects/commercial/modules/dfp/prepare-googletag.js
```diff
@@ -33,64 +34,70 @@ define([
...
+                    }
+                }
 
-            if (dfpEnv.sonobiEnabled) {
-                // Just load googletag. Sonobi's wrapper will already be loaded, and googletag is already added to the window by sonobi.
-                require(['js!googletag.js']);
-                ophanTracking.addTag('sonobi');
-            } else {
-                require(['js!googletag.js']);
+                window.googletag.cmd.push = raven.wrap({deep: true}, window.googletag.cmd.push);
```
regiskuckaertz: I've always wondered whether this line made any sense at all. When `googletag.js` loads, it overwrites `googletag.cmd.push`. Anyway, that's totally unrelated 😆

static/src/javascripts/projects/commercial/modules/dfp/prepare-googletag.js
```diff
@@ -33,64 +34,70 @@ define([
...
+                    setPageTargeting,
+                    moduleCompleted,
+                    resolve
+                );
+            });
+        }
 
-            window.googletag.cmd.push = raven.wrap({deep: true}, window.googletag.cmd.push);
+        if (commercialFeatures.dfpAdvertising) {
+            return prepareSonobiTag.init().then(setupAdvertising).catch(function(){
```
regiskuckaertz: Hasn't `prepareSonobiTag` already been called at this point? (in `commercial.js`)

static/src/javascripts/projects/commercial/modules/dfp/prepare-sonobi-tag.js
```diff
@@ -0,0 +1,24 @@
...
+    'common/modules/commercial/commercial-features',
+    'commercial/modules/dfp/dfp-env'
+], function(
+    commercialFeatures,
+    dfpEnv
+){
+    var prom = null;
+
+    function setupSonobi() {
+        return new Promise.resolve(require(['js!sonobi.js']));
```
regiskuckaertz: You don't need the `new` here

regiskuckaertz: (also don't forget to import the `Promise` module)

static/src/javascripts/bootstraps/commercial.js
```diff
@@ -97,17 +100,21 @@ define([
...
         modules.forEach(function (pair) {
 
             var moduleName = pair[0];
+            var moduleInit = pair[1];
+            var hasCustomTiming = pair[2];
 
             robust.catchErrorsAndLog(moduleName, function () {
-                var modulePromise = pair[1]().then(function(){
-                    ophanTracking.moduleCheckpoint(moduleName, baseline);
+                var modulePromise = moduleInit(moduleName).then(function(){
```
rich-nguyen: The only thing I could think of was moving the name each module into the module itself, but it doesn't really help, and it might even be harder to see.

static/src/javascripts/projects/commercial/modules/dfp/prepare-googletag.js
rich-nguyen: yeah, good point, this probably doesn't do what was intended. could be another pr to think of a better way to protect our system from the cmd buffer.

static/src/javascripts/projects/commercial/modules/dfp/performance-logging.js
```diff
@@ -49,10 +49,13 @@ define([
...
     function moduleEnd(moduleName) {
         var timerEnd = userTiming.getCurrentTime();
-        var module = performanceLog.modules.find(function(module){
-            return module.name === moduleName;
-        });
-        if (module) {
+
+        var moduleIndex = performanceLog.modules.map(function (module) {
+            return module.name;
+        }).indexOf(moduleName);
```
regiskuckaertz: That's why @janua polyfill service is so awesome, you can (almost) forget about browser compat.

static/src/javascripts/projects/commercial/modules/dfp/prepare-googletag.js
rich-nguyen: Yes it is. I thought about this a bit, and decided it best to execute both the sonobi module and the googletag module in the commercial bootstrap, but maintain the dependency between the two, within the depending module. This means that the sonobi module is tracked and measured in the same way as other primary modules, whilst I can still individually measure this googletag module.

static/src/javascripts/projects/commercial/modules/dfp/prepare-sonobi-tag.js
rich-nguyen: good catch, thanks

regiskuckaertz (approved):
Great!

static/src/javascripts/projects/commercial/modules/dfp/prepare-sonobi-tag.js
```diff
@@ -0,0 +1,24 @@
+define([
+    'common/modules/commercial/commercial-features',
+    'commercial/modules/dfp/dfp-env'
+], function(
+    commercialFeatures,
+    dfpEnv
+){
+    var prom = null;
```
rich-nguyen: Cool, very tidy. I guess `require` is memoized, in a sense. what i found was i could memoize `setupSonobi`, but if i memoized `init`, the second call still executed. do you know why?

PR #15082, opened 2016-11-22
Remove unused routes in Frontend::Identity

markjamesbutler:
These routes are now served via [Identity Frontend](https://github.com/guardian/identity-frontend) the routing is done in [Fastly](https://github.com/guardian/identity-platform/blob/master/fastly/profile-theguardian-com.vcl#L29)

I am removing these pages as they are stil accessible via (https://profile-old.theguardian.com/signin) thus open to attack.


jfsoul (approved):
👍

PR #15085, opened 2016-11-23
Use lowercase names to track logo interactions

kelvin-chappell:
GA has recently been reconfigured to store event categories, actions and labels in lowercase to avoid duplicate values.

This PR sends tracking events in lowercase in the hope that it will avoid any confusion.

/cc @guardian/commercial-dev @dominickendrick @gtrufitt


gtrufitt:
👍

PR #15086, opened 2016-11-23
Fix: Update health check route

markjamesbutler:
The health check was pointing at a route which I removed #15082

Updated to new route using routes file to allow compile verification.

### DIFF
--- a/identity/app/controllers/HealthCheck.scala
+++ b/identity/app/controllers/HealthCheck.scala
@@ -4,5 +4,5 @@ import conf.{AllGoodCachedHealthCheck, NeverExpiresSingleHealthCheck}
 import play.api.libs.ws.WSClient
 
 class HealthCheck(wsClient: WSClient) extends AllGoodCachedHealthCheck(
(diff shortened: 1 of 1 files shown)


identity/app/controllers/HealthCheck.scala
```diff
@@ -4,5 +4,5 @@ import conf.{AllGoodCachedHealthCheck, NeverExpiresSingleHealthCheck}
 import play.api.libs.ws.WSClient
 
 class HealthCheck(wsClient: WSClient) extends AllGoodCachedHealthCheck(
-  NeverExpiresSingleHealthCheck("/signin")
+  NeverExpiresSingleHealthCheck(routes.ReauthenticationController.renderForm(None).url)
```
TBonnin: I don't know much about this endpoint. As long as it is accessible without being authenticated I guess it would do

markjamesbutler:
screen shot 2016-11-23 at 13 15 18

TBonnin (approved):

PR #15088, opened 2016-11-23
Move facebook remote over to commercial

regiskuckaertz:
The endpoint `/facebook-ia/remote.html` is used as a proxy to deliver ads in IA. It used to be in `facia`, but this is commercial territory so moving it over there---I'll do AMP et al in a separate PR.

I also rewrote it to:

- pass key-values to DFP
- add support for the fluid size
- make sure Krux is loaded before reading segments in local storage
- add small optimisations that make the code sensibly more readable


regiskuckaertz:
[Ads in Facebook IA: refactor and moving over to commercial](https://trello.com/c/tqwDK3C4/198-ads-in-facebook-ia-refactor-and-moving-over-to-commercial)

commercial/public/facebook_ia_remote.html
```diff
@@ -0,0 +1,126 @@
...
+        }
+    }(window))
+    </script>
+    <script async class="kxct" data-id="J_RxuxWJ" data-timing="async" data-version="1.9" src="https://cdn.krxd.net/controltag?confid=J_RxuxWJ"></script>
+    <script async src="https://www.googletagservices.com/tag/js/gpt.js"></script>
+</head>
+<body style="margin:0">
+    <div id="ad-slot-1"></div>
+    <script>
+    (function (window, localStorage, guardian, undefined) {
```
rich-nguyen: why `undefined`?

rich-nguyen:
👍  don't forget to update the router.

commercial/public/facebook_ia_remote.html
regiskuckaertz: ha! good catch. It is a trick to make `undefined` a local reference when using an IIFE. But I moved code around and this is not needed anymore, not there. Thanks :-)

PR #15091, opened 2016-11-23
Add youtube atom caption

gidsg:
## What does this change?
Adds caption equivalent to that currently shown on non-atom videos when embedded in articles and live blogs. Note, this does not link to the video page, as there may not be one in the Atom model.

Needed for feature parity for YouTube PFP trial.

No

## Screenshots
screen shot 2016-11-23 at 12 39 36

Live blog main media (this one is really low-contrast I know but not sure what if anything we should change it to?)

screen shot 2016-11-23 at 12 13 29

Article page

## Request for comment

CC @duarte @markjamesbutler

### DIFF
--- a/applications/app/views/fragments/mediaBody.scala.html
+++ b/applications/app/views/fragments/mediaBody.scala.html
@@ -1,4 +1,4 @@
-@(page: MediaPage)(implicit request: RequestHeader)
+@(page: MediaPage, displayCaption: Boolean)(implicit request: RequestHeader)
 
 @import common.LinkTo
 @import model.{Audio, AudioPlayer, Video, VideoPlayer}
@@ -89,7 +89,7 @@
                                                 <meta itemprop="thumbnail" content="@Html(video.content.signedArticleImage)" />
                                                 <meta itemprop="thumbnailUrl" content="@Html(video.content.signedArticleImage)" />
 
-                                              @fragments.atoms.media(mediaAtom)
+                                              @fragments.atoms.media(mediaAtom, displayCaption = false)
                                             }
 
                                             @video.elements.mainVideo.map { videoElement =>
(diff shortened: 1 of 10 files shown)


applications/app/views/fragments/mediaBody.scala.html
```diff
@@ -89,7 +89,7 @@
                                                 <meta itemprop="thumbnail" content="@Html(video.content.signedArticleImage)" />
                                                 <meta itemprop="thumbnailUrl" content="@Html(video.content.signedArticleImage)" />
 
-                                              @fragments.atoms.media(mediaAtom)
+                                              @fragments.atoms.media(mediaAtom, displayCaption = false)
```
TBonnin: Should this use the template param?

common/test/views/support/cleaner/AtomCleanerTest.scala
```diff
@@ -32,7 +32,7 @@ class AtomCleanerTest extends FlatSpec with Matchers with FakeRequests {
 
 
  private def clean(document: Document, atom:Option[Atoms], amp: Boolean): Document = {
-    val cleaner = AtomsCleaner(youTubeAtom, shouldFence = false, amp = amp)(TestRequest())
+    val cleaner = AtomsCleaner(youTubeAtom, amp = amp)(TestRequest())
```
TBonnin: Before your change `shouldFence` would be `false`. Now it would be `true` because of default value. Does this matter in this test?

TBonnin (changes requested):

common/test/views/support/cleaner/AtomCleanerTest.scala
gidsg: No , because `shouldFence` is only used by interactive atoms, it should not really be an `AtomsCleaner` argument I think it could be pushed down into the interactive fragment.

applications/app/views/fragments/mediaBody.scala.html
gidsg: No, I wanted to override the setting here because we don't want to show the caption on the video standalone page.

TBonnin (approved):

TBonnin:
lgtm. Check with other people who know more about youtube atom though

akash1810 (approved):
🎉

PR #15092, opened 2016-11-23
Add compile step to rename sourcemap

gtrufitt:
## What does this change?

Adds hacky fix for lack of sourcemap on concat step. Sourcemaps try and load `boot.js.map` from `app.js` but `boot.js.map` gets sent to a different hashed folder. This renames the sourcemap to `app.js.map` and changes the sourcemap location comment.

image

Sourcemaps should work better, maybe we'll get better sentry errors!?

@sndrs This feels super weird, but we've not had sourcemaps working for ages, right?

### DIFF
--- a/tools/__tasks__/compile/javascript/bundle.js
+++ b/tools/__tasks__/compile/javascript/bundle.js
@@ -12,7 +12,8 @@ const bundles = [{
     exclude: [
         'text',
         'inlineSvg'
-    ]
+    ],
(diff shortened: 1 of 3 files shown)


sndrs:
does this actually work? given `app.js` is a concatenation of `boot` and `curl`, would't the source map no longer apply?

gtrufitt:
Heh... no.

PR #15097, opened 2016-11-23
Article Refresh! 

NataliaLKB:
## What does this change?
This is testing our new article layout with the new header changes. Paired on this with @zeftilldeath but I had to squash the commits as they were getting a bit out of hand.

There is quite a lot that has changed here, but I will try to summarise below:

1. Tertiary Nav on every page
Each section of the website should "belong" to one of our top level sections such as News, Opinion, Sport, Art or Life and point to it.
image

2. Image above the headline
image

and captions are hidden by default if they are for a main image:
hiding caption

3. Share buttons above the byline
image

4. Testing slight style variations on the tertiary nav
image
image
image

**Before:**
image

**After:**
image

This will be a serverside test and we will definitely measure if it is successful before implementing for everyone

Nope :)

## Request for comment


static/src/stylesheets/layout/_ab-new-header-test-variant.scss
```diff
@@ -715,61 +715,327 @@ $navigation-horizontal-padding: $gs-gutter * 2 + $gs-gutter/2;
...
+        pointer-events: none;
+        content: '/';
+        color: #8a8a8a;
+    }
+
+    // Scrolly variant, Single line limited variant
+    .tertiary__variant-three &,
+    .tertiary__variant-five & {
+        line-height: $gs-baseline * 2.5;
+        padding: 0 .4em;
```
NataliaLKB: In this case we want the padding to be in relation to the font-size. CC @zeftilldeath

static/src/stylesheets/layout/_ab-new-header-test-variant.scss
```diff
@@ -715,61 +715,327 @@ $navigation-horizontal-padding: $gs-gutter * 2 + $gs-gutter/2;
...
+        }
+
+        .caption--main.caption--video {
+            padding-bottom: 0;
+        }
+
+        .caption--main.caption--img {
+            position: absolute;
+            background: rgba($multimedia-support-5, .8);
+            color: white;
```
SiAdcock: & line 983:
```
color: #ffffff;
```
?

NataliaLKB: Thanks for that! I fixed this and a few other css bits here: https://github.com/guardian/frontend/pull/15097/commits/b1245b1ab1de36cef8fe9d842fcd3f99bb720b89

paperboyo:
+ @stephanfowler 

Looks cool. I especially love how the triangle indicated structure within the nav! Couple of points, if I may:
1. I know that this is being considered, but I wanted to stress how the main tag disappearance would negatively impact on surfacing our structure to readers ("Climate change" in the above example). Maybe all tags should be treated the same when it comes to design and enclosed in lozenges?
2. Please be aware that there is no ratio constraints on main media element. Article would "start" quite a bit far for https://www.theguardian.com/lifeandstyle/2015/may/02/beach-body-ready-advert-backlash-hadley-freeman (admittedly, an extreme case 😆).
3. Now this makes my heart bleed... We already reduced the discoverability (and use?) of Lightbox by removing fullscreen icons from galleries. Now it looks like they are gone from main media :saddestpanda:. Plus, now the click to reveal the caption may be easily confused with the click to open the Lightbox. If this is driven solely by the designer's hate for our fullscreen icon, maybe we could just change it? Or maybe we could at least show it on hover for desktop breakpoints, pretty please? Yes, Lightbox is my wife. And my seven kids.
4. Do we have an icon for when it's an illustration instead of the photograph? Do we care?

rich-nguyen:
Have we confirmed we are ok to hide the photograph credit, as well as the caption?

SiAdcock:
Am I going crazy or is there a tiny 1px gap between the top of the secondary nav and the bottom of the little grey arrow that points to the primary nav position? Is this intentional?

1e6ef81e-b19a-11e6-957b-9752cac07e43

gtrufitt:
image

I think maybe Zef fixed this elsewhere (the Twitter icon wrapping)?

common/app/common/NewNavigation.scala
```diff
@@ -156,188 +152,48 @@ object NewNavigation {
...
-      NavLink("cricket", "/sport/cricket"),
-      NavLink("rugby union", "/sport/rugby-union"),
-      NavLink("F1", "/sport/formulaone"),
-      NavLink("tennis", "/sport/tennis"),
-      NavLink("golf", "/sport/golf"),
-      NavLink("cycling", "/sport/cycling"),
-      NavLink("boxing", "/sport/boxing"),
-      NavLink("US sports", "/sport/us-sport")
-    )
+    val uk = List(sport, football, cricket, rugbyUnion, formulaOne, tennis, golf, cycling, boxing, racing , rugbyLeague, usSports)
```
gtrufitt: This is lovely work. Tidy tidy 👌

stephanfowler:
@paperboyo wrt main tag, yes ww all agree it's important. The plan is to rethink how we do tags, badges, etc. We could combine them, make the primary one into a kicker for the headline, or for the standfirst, or put them up in the second level nav. But it's not in this PR because the PR is big already, and it's too many things to think about in one go!

stephanfowler:
@paperboyo wrt the fullscreen icon, don't forget this PR applies to mobile only. There's general agreement that fullscreen on mobile articles is somewhat pointless given that the "expanded" image is no bigger than its regular rendering.

paperboyo:
@stephanfowler 
> We could combine them

How to combine not loosing structure differentiation?

> make the primary one into a kicker for the headline, or for the standfirst

Yeah, some are quite verbose, though, e.g. https://www.theguardian.com/uk-news/2016/nov/23/philip-hammond-warned-against-brexit-and-no-one-listened-now-its-payback...

> this PR applies to mobile only

OK, was wondering that this wouldn’t play well with weightings.

> There's general agreement that fullscreen on mobile articles is somewhat pointless given that the "expanded" image is no bigger than its regular rendering.

I would say it’s only/mostly because the current Lightbox implementation does not use the Fullscreen API and changing orientation to see a bigger image is glacial. And it does not allow zoom. But yeah, I also love to disagree with the general agreement and clearly have the heart on the other side to @zef’s here 😜

NataliaLKB:
@rich-nguyen 
> Have we confirmed we are ok to hide the photograph credit, as well as the caption?

Great question! We are investigating it right now.

NataliaLKB:
@gtrufitt 
> I think maybe Zef fixed this elsewhere (the Twitter icon wrapping)?
Yes he has! But at the time of this PR that change wasn't merged in. That happened this afternoon so I will merge with master and consider it fixed.

stephanfowler:
> Have we confirmed we are ok to hide the photograph credit, as well as the caption?

Yes, it's just been approved.

common/app/common/NewNavigation.scala
```diff
@@ -535,9 +366,67 @@ object NewNavigation {
...
         item.frontId == frontId
       }
     }
   }
+
+  object SectionLinks {
+
+    var sectionLinks = List(
+
+      SectionsLink("uk", News),
```
gtrufitt: Feels like these could be represented differently to avoid the repetition but maybe as a future PR (because I don't know how off the top of my head!) :)

gtrufitt (approved):
LGTM - nice work!

common/app/common/NewNavigation.scala
NataliaLKB: Yes I agree! Certainly if this change is kept I want to sit down and discuss a nice way to avoid repetition, but still make it feel not so complicated.

PR #15100, opened 2016-11-23
Remove grunt postcss/pix-to-rem

sndrs:
## What does this change?

- removes `grunt-postcss` and `grunt-px-to-rem`
- both tasks are now carried out by `postcss`
-  sass compilation and postcss processing are performed in one pass

As a result of using @cuth's [postcss-pxtorem](https://github.com/cuth/postcss-pxtorem), the final CSS has these changes:

- `4.854166875rem` → `4.85417rem`
- `line-height:20px` → `line-height:1.25rem`
- `0` → `0rem`

The `0rem` is not the end of the world but I'll see if I can put in a PR to leave `0`s alone in `postcss-pxtorem`.

- removes more grunt plugins
- aligns prod output with dev
- consolidates CSS compilation

## Request for comment

@guardian/dotcom-platform


tools/__tasks__/compile/css/sass.js
```diff
@@ -0,0 +1,94 @@
...
+    unit_precision: 5,
+    prop_white_list: []
+};
+
+
+const renderSass = pify(sass.render);
+const saveCSS = pify(fs.writeFile);
+const getFiles = query => glob.sync(path.resolve(sassDir, query));
+
+const compile = (query, {browsers, remify = true}) => Promise.all(
```
gtrufitt: Do you think we need to get some docs/typing in place in some of these? Do you have a plan for that?

tools/__tasks__/compile/css/sass.js
```diff
@@ -0,0 +1,94 @@
...
+const remifications = {
+    replace: true,
+    root_value: 16,
+    unit_precision: 5,
+    prop_white_list: []
+};
+
+
+const renderSass = pify(sass.render);
+const saveCSS = pify(fs.writeFile);
```
gtrufitt: I think you've had a bit of a convention of writing promise wrapped `fs` methods either as `const writeFile = pify(fs.writeFile)` or `const writeFileP = pify(fs.writeFile`. I prefer the latter I think as clarity of the underlying method but promisified, as here with `saveCSS` I had to double check what it did. What do you think?

gtrufitt (approved):
Great! Did you look at how it performed against the gruntified versions (just curious if it was any quicker, probably not a whole lot)?

tools/__tasks__/compile/css/sass.js
sndrs: yeah it's true i've not been consistent, just whichever seemed right at the time. it's maybe not a bad idea to settle on one though, and I'm with you on `writeFileP`. I'll update this then look at the rest in another PR

tools/__tasks__/compile/css/sass.js
sndrs: didn't no, but seems ideal candidate for typing

sndrs:
> Did you look at how it performed against the gruntified versions (just curious if it was any quicker, probably not a whole lot)?

`14s` → `11s`

PR #15102, opened 2016-11-24
Fix masterclasses parsing failure

JonNorman:
## Background
We parse Eventbrite feeds to drive our Masterclasses and Live Events components. Currently, we expect each event to have at least one ticket that can be booked. It turns out that this was a _foolish_ assumption and it has indeed made an ass out of me.

## What does this change?
All properties derived from the `tickets` of an event are now optional.

Sneaky change - there is a logging statement that basically spams the logs with information we very very rarely care about. I've taken it out for now as it makes the logs a lot easier to grep/parse/read.

Despite the fact that we will probably never serve one of these events to the end user, we will no longer get parsing failures when parsing these events.

Nope

## Request for comment
@guardian/commercial-dev

### DIFF
--- a/commercial/app/model/merchandise/events/Eventbrite.scala
+++ b/commercial/app/model/merchandise/events/Eventbrite.scala
@@ -130,7 +130,7 @@ object Eventbrite extends ExecutionContexts with Logging {
   trait TicketHandler {
     def tickets: Seq[Ticket]
 
-    lazy val displayPriceRange: Option[String] = {
+    val displayPriceRange: Option[String] = {
 
       def determinePriceRange: String = {
 
--- a/commercial/app/model/merchandise/events/Eventbrite.scala
+++ b/commercial/app/model/merchandise/events/Eventbrite.scala
@@ -130,22 +130,32 @@ object Eventbrite extends ExecutionContexts with Logging {
   trait TicketHandler {
     def tickets: Seq[Ticket]
 
-    val displayPriceRange = {
+    lazy val displayPriceRange: Option[String] = {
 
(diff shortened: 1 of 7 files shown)


commercial/app/model/merchandise/events/Eventbrite.scala
```diff
@@ -130,22 +130,32 @@ object Eventbrite extends ExecutionContexts with Logging {
   trait TicketHandler {
     def tickets: Seq[Ticket]
 
-    val displayPriceRange = {
+    lazy val displayPriceRange: Option[String] = {
```
Calum-Campbell: Why is this lazy?

Otherwise LGTM 👍

regiskuckaertz:
Should the JSON endpoints be updated as well?

commercial/app/model/merchandise/events/Eventbrite.scala
JonNorman: Ah yes! Good spot. I originally did that in a brief spot of laziness before immediately abandoning the idea. My idea was to make it lazy and filter out all Events that have no tickets, but this was a terrible idea for many reason - forgot to delete the `lazy` though.

JonNorman:
This only affects masterclasses (which we don't have an endpoint for) and live events. The live events JSON endpoint doesn't write these fields; that the logic is handled in the commercial-templates templates, as I believe it should be.

**tl;dr** this PR only affects the server-side rendering templates

regiskuckaertz:
yahaaa 👍

JonNorman:
[Masterclass feeds is failing.](https://trello.com/c/MLieooGy/200-masterclass-feeds-is-failing)

PR #15103, opened 2016-11-24
Move third-party ads routes into commercial territory (part 1/2)

regiskuckaertz:
Okay... moving this over to commercial. Then updating routes in platform. Then removing the stuff in facia.

cc @rich-nguyen


rich-nguyen:
👍

PR #15104, opened 2016-11-24
Account deletion

mariogalic:
## What does this change?

This enables self-service account deletion for a limited kind of user that satisfies the following constraints:
 * has no saved articles, and
 * has never commented, and
 * does not have a jobs account, and
 * does not have digital subscriptions, and
 * does not have active membership

Otherwise the users is directed to contact Userhelp.

Current account deletion process is manual and time-consuming for both the user and Guardian. 
Userhelp performs around 100 manual account deletions per week (around 25% of these represent the above limited user). The current process involves at least 13 manual steps (please talk to userhelp for further details).

### Identity API
The [corresponding identity PR](https://github.com/guardian/identity/pull/601) adds deletion endpoint to Identity API. There are 3 layers of security on top of the endpoint: 
* dotcom client token, and
* secure session cookie, and 
* password.

### Dotcom
In dotcom:
* the user has to be signed in
* the user has to re-enter password in deletion form to confirm deletion
* there is anti-forgery protection on the deletion form

* New endpoint in Identity API: https://github.com/guardian/identity/pull/601

## Screenshots

If the user satisfies the above criteria for auto deletion:

account_deletion_a

otherwise the fallback to the old process:

account_deletion_b

## Request for comment

@adamnfish @markjamesbutler @andrewfindlay


identity/app/controllers/AccountDeletionController.scala
```diff
@@ -0,0 +1,162 @@
...
+          case FailedToRemoveFromAllEmailLists =>
+            logger.error(s"Failed to remove user ${request.user.user.id} from all mailing lists during account deletion.")
+            renderFormWithUnableToDeleteAccountError(boundForm)
+
+          case FailedToDeleteAccount =>
+            logger.error(s"Failed to delete account for user ${request.user.user.id}")
+            renderFormWithUnableToDeleteAccountError(boundForm)
+        }
+
+      (for {
```
mario-galic: Main business logic

identity/app/controllers/AccountDeletionController.scala
```diff
@@ -0,0 +1,162 @@
...
+    case class ManualDeletion(hasSavedArticles: Boolean, hasComment: Boolean, hasJob: Boolean, isSubscriber: Boolean, isMember: Boolean) extends DeletionType
+
+
+    /* Users can delete account themselves if they
+         - has no saved articles, and
+         - have never commented, and
+         - do not have a jobs account, and
+         - do not have active digipack subscriptions, and
+         - do not have active membership */
+    private def selectDeletionType[A](implicit request: AuthenticatedActions.AuthRequest[A]): Future[DeletionType] =
```
mario-galic: Criteria for self-service deletion

andrewfindlay:
👍  Looks great! You're going to save us a lot of work here. Thanks Mario.

identity/app/services/ExactTargetService.scala
```diff
@@ -0,0 +1,44 @@
+package services
+
+import com.exacttarget.fuelsdk.{ETResponse, ETSubscriber}
+import utils.SafeLogging
+import common.ExecutionContexts
+import conf.IdentityConfiguration
+import scala.concurrent.Future
+import scalaz.\/
```
TBonnin: Is there any way to do without scalaz? 
This is my personal opinion and I know it is already a dependency in frontend but scalaz makes the barrier of entry higher in term of contributing to the code. 
That would be fine if it was a narrow scoped project or if we were using scalaz throughout the whole codebase but we only use it in an handful of places. It is more a burden than anything else at this point I think

identity/app/controllers/AccountDeletionController.scala
```diff
@@ -0,0 +1,162 @@
...
+    def renderAccountDeletionConfirmForm = Action.async { implicit request =>
+      Future(Ok(accountDeletionConfirm(page, idRequestParser(request), idUrlBuilder)))
+    }
+
+    private def deleteAccount[A](
+        boundForm: Form[String],
+        emailPasswdAuth: EmailPassword,
+        idRequest: IdentityRequest)(implicit request: AuthenticatedActions.AuthRequest[A]) = {
+
+      sealed trait AccountDeletionFailures
```
adamnfish: This (and the `processFailures` function below) looks like unnecessary indirection to me. Why not just create the response directly? This would make heaps of line noise vanish and would also mean you have the details from the API error in scope for the log message.

identity/app/controllers/AccountDeletionController.scala
```diff
@@ -0,0 +1,162 @@
...
+            logger.error(s"Failed to delete account for user ${request.user.user.id}")
+            renderFormWithUnableToDeleteAccountError(boundForm)
+        }
+
+      (for {
+        _ <- checkUserEnteredCorrectPassword()
+        _ <- removeFromAllMalingLists()
+        _ <- unauthenticateUser()
+        _ <- deleteAccountProper()
+      } yield ()).fold(processFailures, _ => clearCookiesAndDisplaySuccessForm)
```
adamnfish: If you put the same type in Left and Right (e.g. by removing the processFailures indirection and returning the error result directly at each step) you'll have a `MergableEither` that you can just call `.merge` on instead of needing to fold.

identity/app/idapiclient/IdApi.scala
```diff
@@ -173,11 +187,33 @@ abstract class IdApi(val apiRootUrl: String, http: Http, jsonBodyParser: JsonBod
...
+    delete("user/me", Some(auth), body = Some(write(emailPassword))) map extractUnit
+
   // THIRD PARTY SIGN-IN
   def addUserToGroup(groupCode: String, auth: Auth): Future[Response[Unit]] = {
     post(urlJoin("user", "me", "group", groupCode), Some(auth)) map extractUnit
   }
 
+  // Membership & Subscription
+  def userIsMember(auth: Auth): Future[Boolean] = {
+    http.GET(
```
adamnfish: The indentation is a bit out of whack on these, makes it hard to see what is args to GET and what is processing the response.

identity/app/controllers/AccountDeletionController.scala
```diff
@@ -0,0 +1,162 @@
...
+
+    /* Users can delete account themselves if they
+         - has no saved articles, and
+         - have never commented, and
+         - do not have a jobs account, and
+         - do not have active digipack subscriptions, and
+         - do not have active membership */
+    private def selectDeletionType[A](implicit request: AuthenticatedActions.AuthRequest[A]): Future[DeletionType] =
+      for {
+        hasSavedArticles <- identityApiClient.hasNonDefaultSavedArticles(request.user.auth)
```
adamnfish: It doesn't look like there's much error handling going on here. What's the behaviour if a call fails?

identity/app/controllers/AccountDeletionController.scala
```diff
@@ -0,0 +1,162 @@
...
+        _ <- checkUserEnteredCorrectPassword()
+        _ <- removeFromAllMalingLists()
+        _ <- unauthenticateUser()
+        _ <- deleteAccountProper()
+      } yield ()).fold(processFailures, _ => clearCookiesAndDisplaySuccessForm)
+    }
+
+    sealed trait DeletionType
+    case object AutoDeletion extends DeletionType
+    case class ManualDeletion(hasSavedArticles: Boolean, hasComment: Boolean, hasJob: Boolean, isSubscriber: Boolean, isMember: Boolean) extends DeletionType
```
adamnfish: Would this be simpler as a `case class DeletionType` with all the fields and a method (or function elsewhere) that tests if it should be auto/manual? Means the logic for testing that would be in one place which is relevant if you expect the list of conditions to grow (and it sounds like that is the case).

There isn't an obvious test for this logic which is a shame given how important it is. By putting the logic in one place (and outside the controller) it'd be easy to add a unit test.

identity/app/controllers/AccountDeletionController.scala
```diff
@@ -0,0 +1,162 @@
...
+        else
+          ManualDeletion(hasSavedArticles, hasComment, hasJob, isSubscriber, isMember)
+      }
+
+    private def renderFormWithUnableToDeleteAccountError(boundForm: Form[String]) =
+      SeeOther(routes.AccountDeletionController.renderAccountDeletionForm ().url)
+        .flashing(boundForm.withGlobalError(
+          "We are experiencing technical difficulties. Your account has not been deleted. Please try again later or contact Userhelp.").toFlash)
+
+    private val cookiesToDiscard = List(
```
adamnfish: There *must* already be a logout endpoint?! Why duplicate this hard-coded list of cookie names?

identity/app/controllers/AccountDeletionController.scala
```diff
@@ -0,0 +1,162 @@
...
+        isSubscriber <- identityApiClient.userIsSubscriber(request.user.auth)
+        isMember <- identityApiClient.userIsMember(request.user.auth)
+      } yield {
+        if (!hasSavedArticles && !hasComment && !hasJob && !isSubscriber && !isMember)
+          AutoDeletion
+        else
+          ManualDeletion(hasSavedArticles, hasComment, hasJob, isSubscriber, isMember)
+      }
+
+    private def renderFormWithUnableToDeleteAccountError(boundForm: Form[String]) =
```
adamnfish: I think you can do better than this for the user. You might consider using the API response's friendly message?

identity/app/controllers/AccountDeletionController.scala
```diff
@@ -0,0 +1,162 @@
...
+import scalaz.std.scalaFuture._
+
+class AccountDeletionController(
+    api: IdApiClient,
+    idRequestParser: IdRequestParser,
+    idUrlBuilder: IdentityUrlBuilder,
+    authenticationService: AuthenticationService,
+    authenticatedActions: AuthenticatedActions,
+    val messagesApi: MessagesApi,
+    identityApiClient: IdApiClient,
```
adamnfish: Why are there two arguments of the same type, `IdApiClient`?

identity/app/idapiclient/IdApi.scala
```diff
@@ -60,6 +62,18 @@ abstract class IdApi(val apiRootUrl: String, http: Http, jsonBodyParser: JsonBod
...
 
+  def hasNonDefaultSavedArticles(auth: Auth) = {
+    def hasNoSavedArticles(savedArticles: List[SavedArticle]) = {savedArticles.isEmpty}
+
+    def hasOnlyDefaultSavedArticle(savedArticles: List[SavedArticle]) =
+      savedArticles.length == 1 && savedArticles.head.shortUrl == "/p/4ab7x"
+
+    savedArticles(auth).map(_.fold(
+      {_ => true},
+      { prefs => hasNoSavedArticles(prefs.articles) || hasOnlyDefaultSavedArticle(prefs.articles)})
```
adamnfish: :+1:

Also, best to always create (and name) boolean values the way round you intend to test them. In this case I'd recommend something like:

    hasSavedArticles && doesNotHaveOnlyDefaultSavedArticle

I'm sure you could dome up with a better name.

That said, this feels again like indirection that isn't adding much value. Can you not inline these functions? Especially true of `hasNoSavedArticles`. Just the name is as long as the implementation (`savedArticles.isEmpty`) to say nothing of the definition, signature, invocation etc.

identity/app/idapiclient/IdApi.scala
```diff
@@ -60,6 +62,18 @@ abstract class IdApi(val apiRootUrl: String, http: Http, jsonBodyParser: JsonBod
     val response = http.POST(apiUrl(apiPath), Some(updatedSavedArticles), params, headers)
     response map extractSavedArticles }
 
+  def hasNonDefaultSavedArticles(auth: Auth) = {
+    def hasNoSavedArticles(savedArticles: List[SavedArticle]) = {savedArticles.isEmpty}
+
+    def hasOnlyDefaultSavedArticle(savedArticles: List[SavedArticle]) =
+      savedArticles.length == 1 && savedArticles.head.shortUrl == "/p/4ab7x"
+
+    savedArticles(auth).map(_.fold(
```
adamnfish: This mix of indentation / (lack of) bracketing is very confusing!

identity/app/idapiclient/IdApi.scala
```diff
@@ -60,6 +62,18 @@ abstract class IdApi(val apiRootUrl: String, http: Http, jsonBodyParser: JsonBod
...
+      {_ => true},
```
adamnfish: Throwing away errors again, this always makes me nervous.

identity/app/services/ExactTargetService.scala
adamnfish: Why not use cats here for consistency with other parts of the codebase. `Either(T)` seems like a better choice. As well as consistency this means you're using the stdlib's Either, which is much more accessible.

identity/app/services/ExactTargetService.scala
```diff
@@ -0,0 +1,44 @@
...
+
+class ExactTargetService(conf: IdentityConfiguration) extends ExecutionContexts with SafeLogging {
```
adamnfish: Does this behaviour not exist in the API already? There are already pages for managing ET mailing lists so this may be redundant duplication?

As mentioned IRL, putting this kind of logic in Frontend isn't great. In particular it means the apps won't be able to use it.

identity/app/conf/IdentityConfiguration.scala
```diff
@@ -20,4 +21,17 @@ class IdentityConfiguration extends GuardianConfiguration with SafeLogging {
     }
   }
 
+  // Userhelp - 'Guardian Admin' ET Account integrated with subscriptions.dev@guardian.co.uk AppCentre
+  object exactTargetUserhelp {
```
adamnfish: Bringing this sort of behaviour into Frontend account isn't in line with the expectations of privacy and security for that AWS account. Generally applications in Frontend are considered simple transformations from JSON to HTML, whether that's from the Identity API or more commonly, the content API.

I'm sure this isn't the only example of other behaviour, but in this case you're bringing into the Frontend account the ability to *directly manage* the personal data of our users. I'd want to have a chat with the wider Frontend team about how they feel taking on the burden of maintaining the security of such a service.

adamnfish (changes requested):
This is a great feature but I do have a mild concern about what this adds to Frontend. Be good to have a chat with you and the team about these before this is released as-is.

There are also a few style / logic issues which are obviously less important, happy to chat about any of these.

identity/app/services/ExactTargetService.scala
adamnfish: Ah, just realised I'm wrong here. You're using scalaz's EitherT elsewhere so the consistency argument doesn't apply. I'd very strongly favour cats over scalaz for these use cases but if you're doing that for consistency with existing code then that'd be an argument for using scalaz.

identity/app/conf/IdentityConfiguration.scala
TBonnin: I am with @adamnfish on this. I don't think the responsibility of managing user personal data should leave in Frontend. @markjamesbutler has probably an opinion about that

TBonnin: I am with @adamnfish on this. I don't think the responsibility of managing user personal data should leave in Frontend. @markjamesbutler has probably an opinion about that

TBonnin (changes requested):

mario-galic:
Requested changes implemented, namely:

* Remove scalaz
* Remove Userhelp ExactTarget
* Invert criteria logic

@adamnfish May I provide more informative error messages in a separate PR to keep current PR as small as possible? For now I make a conservative assumption that if anything at all goes wrong with deletion criteria the user falls back to manual deletion form and Userhelp takes it from there.

PR #15106, opened 2016-11-24
Use series as the minute logo link

davidfurey:
## What does this change?

Fixes the link on "the minute" circular logo from minute/briefing tone articles.  Previously this always linked to the US minute.  Now it links to the series page.  If there is no series then it isn't a link.

Avoids confusing user experience.

No

## Screenshots
screen shot 2016-11-23 at 16 31 27

## Request for comment

@superfrank @gtrufitt @katebee


katebee (approved):
👍 cool!

PR #15108, opened 2016-11-24
remove grunt-assetmonitor

sndrs:
## What does this change?

removes grunt assetmonitor

the last grunt plugin to remove

admin, potentially

## Request for comment

@guardian/dotcom-platform, @TBonnin in particular

it's a bit hard to test, since it only runs on a successful build on teamcity...

also, i will need to point the TC [asset monitor build step](https://teamcity.gu-web.net/admin/editBuildRunners.html?id=buildType:dotcom_master) at `dev/teamcity/dist-assetmonitor-tc` before merging


SiAdcock:
LGTM 👍

PR #15110, opened 2016-11-24
Youtube: Rename wrapper

markjamesbutler:
Renaming to el following removal of wrapper in #15080 

Ideally this should have been done in #15080


gidsg (approved):
LGTM 👍 ☮️

PR #15113, opened 2016-11-24
require yarn 0.17

sndrs:
## What does this change?

require yarn 0.17

0.17 reorders the `yarn.lock` file which i appear to have inadvertently committed, having updated locally. now anyone on 0.16 will find an updated `yarn.lock` just by installing (which all make tasks now do by default).

## Request for comment

@SiAdcock


SiAdcock:
👍

PR #15115, opened 2016-11-24
AU membership engagement banner copy AB test

JustinPinner:
## What does this change?
Sets up an AB test to explore variations of the membership engagement banner copy displayed to Australia edition readers.

Hypothesis: More persuasive copy will improve membership conversions from impressions
We'll be looking at conversion rates resulting from each variant.

No, web only.

## Screenshots
**control**
screen shot 2016-11-24 at 16 11 36-au-test8-control

**fearless**
screen shot 2016-11-24 at 16 12 22-au-test8-fearless-10

**stories**
screen shot 2016-11-24 at 16 12 57-au-test8-stories-that-matter

**power**
screen shot 2016-11-24 at 16 13 35-au-test8-power-to-account

**independent**
screen shot 2016-11-24 at 16 14 15-au-test8-independent-journalism

## Request for comment
@Ap0c @svillafe @rupertbates @rtyley


Ap0c (approved):
All ran as expected for me locally 👍.

JustinPinner:
Nice. I'll get it merged. Thanks.

common/app/conf/switches/ABTestSwitches.scala
```diff
@@ -44,4 +44,15 @@ trait ABTestSwitches {
     sellByDate = new LocalDate(2016, 12, 8), // Thursday 8th December
     exposeClientSide = true
   )
+
+  Switch(
+    ABTests,
+    "ab-au-memb-engagement-msg-copy-test-8",
```
rtyley: Heck, I think this might have been wrong @JustinPinner - it looks like the value should have been `ab-au-memb-engagement-msg-copy-test8`, without the dash before the 8?

The code that matches the Scala switch with the AB test id seems to require no dashes before numerals, and if they don't match, it silently deactivates the test. So as far as I can see, the AB test on the Australia edition has not been active - only the control would have been shown. No variants will have been displayed to users.

I tried this, and as you can see, the banner message doesn't match the selected variant:

image

JustinPinner: Hm, that's odd. I'm in the fearless_10 test and see the expected messaging.

PR #15118, opened 2016-11-25
Extend expiry of contributions US test

philwills:
## What does this change?

Extends the expiry of the contributions message in the US

Continued high conversion rates

No


katebee (approved):
👍 

Curious - what is the reason for the JS having one extra day on the expiry than the switch?

philwills:
@katebee The expiry in JS requires the day to have started (https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/common/modules/experiments/ab.js#L87) where as the Scala switches require the day to have passed. I keep meaning to do a PR to make them consistent.

PR #15127, opened 2016-11-25
GA: Match the apps labelling of 'international' edition

gtrufitt:
## What does this change?

Makes the international edition use 'international' instead of 'int' when sending data to GA to match the apps.

Data integrity. 

@guardian/dotcom-platform


NataliaLKB (approved):
👍

PR #15129, opened 2016-11-25
Taste the rainbow! 🌈 tones for the email forms

katebee:
This PR is all about email forms, they live in routes like [this](https://www.theguardian.com/email/form/article/2635) and [this](https://www.theguardian.com/email/form/plain/2635). These forms are used inside iframes to create our email CTAs

Email forms are useful - they allow us to give the subscription confirmation in an iframe and avoid a whole page refresh

## What does this change?

Adds `plaintone` as a new type of email sign up form to cover gap between `article` and `plain`

Adds a option to have an accent colour (other than grey) as the border around the input and submit. Currently only `feature` has this accent:

picture 379

`plaintone` will inherit colours based on the theme/tone specified in `EmailSignupController.scala` (feature, media, comment)

🌈 Colour! 🌈

Ever wanted to have an email sign up with the fabulous colourations of the `article` version, without the pesky headline and label? Now you can!

The standalone [sign up pages](https://www.theguardian.com/signup/weekendreading) can start using these colourful iframes to prevent the new page on `submit`.

Allows a migration of email CTAs that are currently using `article` for colour, that have to run JS over the page to hide or alter the headline. (Go [here](https://www.theguardian.com/info/2016/feb/15/sign-up-to-the-media-briefing), disable JS and see what happens)

Gives the potential to retire `plaindark` (which I think is used for Documentaries) as `plaintone` could be used in it's place, with the `media` theme. Target for a future PR once usages have been checked.

Nope

###### `article` version of the email sign up (contains the headline and label)

picture 382

###### `plain` version is always brand blue

picture 381

###### new, shiny `plaintone` version will be whatever colour you specify in `EmailSignupController.scala`

picture 383

picture 380

###### example usage (bonus - this is in an interactive atom):

picture 376

## Request for comment

@shtukas @superfrank

### DIFF
--- a/common/app/controllers/EmailSignupController.scala
+++ b/common/app/controllers/EmailSignupController.scala
@@ -42,6 +42,7 @@ object ListIds {
   val greenLight = 28
   val povertyMatters = 113
   val theLongRead = 3322
+  val weekendReading = 3743
   val morningMail = 2636
   val australianPolitics = 1866
 
--- a/applications/conf/routes
+++ b/applications/conf/routes
@@ -33,7 +33,7 @@ POST           /notification/delete                                             
 
 # Email paths
 GET        /email                                                               controllers.EmailSignupController.renderPage()
-GET        /email/form/$emailType<article|footer|plain|plaindark>/:listId       controllers.EmailSignupController.renderForm(emailType: String, listId: Int)
+GET        /email/form/$emailType<article|footer|plain|plaindark|plaintone>/:listId       controllers.EmailSignupController.renderForm(emailType: String, listId: Int)
 GET        /email/:result                                                       controllers.EmailSignupController.subscriptionResult(result: String)
 POST       /email                                                               controllers.EmailSignupController.submit()
 OPTIONS    /email                                                               controllers.EmailSignupController.options()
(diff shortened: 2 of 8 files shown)


common/app/controllers/EmailSignupController.scala
```diff
@@ -42,6 +42,7 @@ object ListIds {
   val greenLight = 28
   val povertyMatters = 113
   val theLongRead = 3322
+  val weekendReading = 3743
```
TBonnin: Just curious. What do those number mean and where are they coming from?

katebee: These are the ListIDs (primary key) for the emails in ExactTarget (our email marketing software). Editorial emails are created in ExactTarget by Editorial, who then inform their contact in Digital of the listID, so that it can be added to [`EmailSubscriptions.scala`](https://github.com/guardian/frontend/blob/master/common/app/model/EmailSubscriptions.scala) and appear in the email-prefs centre.

(I would really love if these listIDs weren't also coded into `EmailSignupController.scala` and that there was one, canonical source for all email data)

katebee: With regards for what they mean, in this file the listID is used to optionally identify an Editorial email, so that a theme/tone can be added to it (comment, feature, etc.).

joelochlann:
This looks good to me :rainbow_flag: 👍 :rainbow_flag: 

Anything stopping this being merged?

katebee:
Happy to merge 🚢 Just wanted another set of eyes looking it over, cheers!

PR #15130, opened 2016-11-25
Add babel step to JS

sndrs:
## What does this change?

- adds transparent babel step to JS compilation and watching (no plugins yet)
- _doesn't_ add transpilation to testing, will come in subsequent PR

- prepares ground for adoption of es6, flow, whatever babel plugins we go for
- decouples the transpilation step from adoption of webpack

no

## Request for comment

@SiAdcock @gtrufitt


dev/bs-config.js
```diff
@@ -17,79 +17,79 @@
...
+    'open': false,
+    'browser': 'default',
+    'xip': false,
+    'hostnameSuffix': false,
+    'reloadOnRestart': true,
+    'notify': true,
+    'scrollProportionally': true,
+    'scrollThrottle': 0,
+    'reloadDelay': 0,
+    'reloadDebounce': 100,
```
SiAdcock: Literally none of these properties need quotes! 😂

sndrs: that is a very good point!

SiAdcock:
Right on!

Does the transpiled output look nice? I'm conscious that in dev mode, we'll be debugging against the transpiled code in devtools, for instance, rather than against the original ES6 source.

Also, is the `babel-watch` task performant enough? It's transpiling all the things in `static/src/javascript` as there is currently no way to distinguish between code that needs to be transpiled and ES5 code. If it's super-fast, that's cool. Otherwise, I know Membership use a `.es6` extension to identify code that needs transpilation.

sndrs:
> Does the transpiled output look nice? I'm conscious that in dev mode, we'll be debugging against the transpiled code in devtools, for instance, rather than against the original ES6 source.

currently it's ever so slightly reformatted, but barely. i plan to add sourcemap support, but maybe that should be part of this first step too...

> Also, is the babel-watch task performant enough?

it only transpiles files that change, so it's near-enough instant. during full compile/stating the watch task, the whole app takes 5-10 seconds, depending on what else your machine is doing, in my experience with this branch

i'm going to close this and re-open with #14946 merged once that's gone in to master, otherwise i'm worried this will just further-complicate that

PR #15131, opened 2016-11-28
Update default section to News for the New Nav

NataliaLKB:
## What does this change?
This change effects this:
image

In this:
image

There are some sections which don't easily belong to a top level section. For these, the default should be News.

There will always be a secondary/tertiary nav (need to get the terminology right)

Nope!

## Request for comment
@guardian/dotcom-platform

### DIFF
--- a/common/app/common/NewNavigation.scala
+++ b/common/app/common/NewNavigation.scala
@@ -422,7 +422,7 @@ object NewNavigation {
       }
 
       if (sectionList.isEmpty) {
-        ("", List())
+        ("News", News.getEditionalisedNavLinks(edition).drop(1))
       } else {
         val section = sectionList.head
         val parentSections = section.parentSection.getEditionalisedNavLinks(edition).drop(1)


jfsoul (approved):
When do we have to rename this from `NewNavigation.scala`? 😄

gtrufitt (approved):

NataliaLKB:
@jfsoul Soon! But first I need to remove `Navigation.scala`... 😓

common/app/common/NewNavigation.scala
```diff
@@ -422,7 +422,7 @@ object NewNavigation {
       }
 
       if (sectionList.isEmpty) {
-        ("", List())
+        ("News", News.getEditionalisedNavLinks(edition).drop(1))
```
TBonnin: why `drop(1)`

NataliaLKB: The first item in the list for each section is the same link as used in either News, Opinion, Sport, Arts, Life here:
image

No sense having the links twice! I may need a comment explaining this though. Thanks for asking

TBonnin: Got it

PR #15132, opened 2016-11-28
Enable ad targeting of Observer content

kelvin-chappell:
This change adds a parameter to ad calls to identify Observer content.

The ad call will have an extra parameter, `ob=t`, in Observer content, and this param will be omitted in ad calls for other content.

/cc @guardian/commercial-dev


rich-nguyen:
👍

PR #15133, opened 2016-11-28
Remove AkkaAsync object (relies on Play global object, expired in Play 2.5)

TBonnin:
## What does this change?

Remove AkkaAsync object (relies on Play global object, expired in Play 2.5)

Instead of having each switch responsible for failing if not initialized after timeout, this responsibility falls in the SwitchboardLifecycle

=> Play 2.5

## Request for comment
@alexduf @jfsoul @DiegoVazquezNanini

### DIFF
--- a/common/app/conf/switches/Switches.scala
+++ b/common/app/conf/switches/Switches.scala
@@ -4,7 +4,6 @@ import java.util.concurrent.TimeoutException
 
 import common._
 import org.joda.time.{DateTime, DateTimeZone, Days, LocalDate}
-import play.api.Play
 
 import scala.concurrent.duration._
 import scala.concurrent.{Future, Promise}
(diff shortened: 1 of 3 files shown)


common/app/conf/switches/Switches.scala
```diff
@@ -88,21 +83,21 @@ case class Switch(
     if (isSwitchedOff) {
       delegate.switchOn()
     }
-    initialized(this)
+    initialize(this)
```
TBonnin: Renaming was not required but I found a verb is better to describe an action

jfsoul (approved):
nice 👍

common/app/conf/switches/Switches.scala
alexduf: 👍

alexduf (approved):
Nice!

PR #15140, opened 2016-11-28
Extend expiry of contributions epic again

philwills:
## What does this change?
Extends the contributions epic for another week. We're intending to integrate this with the targeting tool and stop having to do this, but we need a bit more time.

More contributions/memberships

No


katebee (approved):
🚢 👍

jfsoul:
Hi @philwills, I'm going to merge this so we can get a fix out quickly - hope that's ok!

philwills:
Thanks @jfsoul and sorry for leaving it so close to the expiry before raising this.

PR #15142, opened 2016-11-28
This birds gonna fly...er

superfrank:
## What does this change?
Email banner added for The Flyer test
Naa
Naa
## Screenshots
the-flyer
## Request for comment
@joelochlann


common/app/model/EmailAddons.scala
```diff
@@ -129,6 +129,8 @@ case object KeepItInTheGround extends ArticleEmailMetadata {
 
 case object TheFlyer extends FrontEmailMetadata {
   val name = "The Flyer"
+  override val banner = Some("the-flyer.png")
+  def test(c: ContentPage) = c.item.tags.series.exists(_.id == "travel/series/the-flyer")
```
joelochlann: no need for this `test` line, the test for front emails is the same for all of them and is on line 18 (it's different from article emails where each new email published is a new article with the same series tag - there's only one front per email and the front changes over time, so the test just checks if the page title matches the name of the email)

joelochlann (changes requested):

joelochlann:
Loving the GIF

common/app/model/EmailAddons.scala
superfrank: got ya 👍  thanks.

joelochlann (approved):

PR #15143, opened 2016-11-28
CSRFAddToken and CSRFCheck object are deprecated

TBonnin:
## What does this change?
Using the class instead

Play 2.5 compliant 🎉 

## Request for comment
@alexduf @DiegoVazquezNanini @jfsoul


alexduf (approved):
🎉

PR #15144, opened 2016-11-29
Use semver to check yarn version

sndrs:
## What does this change?

uses semver to check the yarn version, rather than bash strings

`bc` doesn't work on TC. `semver` should be cross platform

no

## Request for comment

@SiAdcock


SiAdcock:
Much better than checking in `semver`! 👍

PR #15145, opened 2016-11-29
Load interactives from standard if not in Webpack test

SiAdcock:
## What does this change?

If user is not participating in the Webpack test, we should load interactives in the standard JS bootstrap. Previously, we prevented interactives from loading if the Webpack switch was on.

Fix interactives for users not participating in the Webpack test

No

## Screenshots

picture 137

## Request for comment

@gtrufitt @jfsoul


gtrufitt (approved):

jfsoul:
👍  speedy work, thanks @SiAdcock @gtrufitt

PR #15146, opened 2016-11-29
Load interactives inline in interactiveBody

SiAdcock:
## What does this change?

In Interactive articles, interactive boot scripts are loaded if user is participating in the Webpack test.

This change ensures interactive boot scripts are loaded regardless of whether the user is participating in the Webpack test. 

Fixes interactive articles for non-Webpack users.

No

## Request for comment

@gtrufitt @jfsoul @NataliaLKB


NataliaLKB (approved):
👍

PR #15148, opened 2016-11-29
Email card design

joelochlann:
Myriad design tweaks to the front email:

- no internal padding on container
- remove the ugly faux-bold
- border colour, configurable per email
- white background
- black container titles
- reduce fixed width of desktop container
- no images on small cards
- as a consequence, bottom padding on cards themselves rather than trail text (since we now need it on cards which don’t have trail text)

#### before
picture 339

#### after
picture 340

@superfrank @desbo


gtrufitt (approved):
LGTM 👍

PR #15149, opened 2016-11-29
Show onward container on hosted video and article amp pages

kelvin-chappell:
This lazy loads the onward container using an amp-list.
I've tried to make it look like the standard mobile page, but I imagine the appearance needs some more work.
They pass amp validation anyway.

Examples:
Video:
image

Article:
image

/cc @Calanthe @lps88


lps88:
I'll fix up the styling for these when I move it out of the html. Looks good 👍

PR #15151, opened 2016-11-29
Unify the deploy using Riff-Raff's the new riff-raff.yaml configuration

sihil:
## What does this change?
Adds a unified deploy artifact to allow the whole of frontend to be deployed at once.

This will simplify the deployment process considerably:
 - Fewer deploys visible in Riff-Raff (the individual applications will be grouped into one deploy)
 - No longer any need to use `goo`

No

## Screenshots
N/A

## Request for comment
It should be possible to merge this without impacting any existing processes. 

At the moment we create both the existing individual deploys and also the unified deploy. The SBT config can be simplified once the old deploys are no longer used.

The new SBT config could also do with simplification, although so far I my weak SBT macro knowledge has not been able to solve it.

**WARNING:** Once merged it will be possible to deploy both types of deploy. Doing so simultaneously could cause serious issues with the stack, potentially leaving it in an inconsistent state.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -18,5 +18,8 @@ indent_size = 2
 [package.json]
 indent_size = 2
 
+[riff-raff.yaml]
+indent_size = 2
+
 [makefile]
 indent_style = tab
--- a/project/Frontend.scala
+++ b/project/Frontend.scala
@@ -1,6 +1,7 @@
 package com.gu
 
 import com.gu.Dependencies._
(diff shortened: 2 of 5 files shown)


TBonnin (approved):
Looks sane to me. However I am not a sbt expert.

We will block `goo` deploy to avoid "double deployment" and then once we have fully tested this unifed deployment we will remove the `deploy` command in `goo`

TBonnin:
🎈 🎈 🎈 🎉  Thanks @sihil for all your work

PR #15152, opened 2016-11-29
Sentry access denied sonobi polyfill

rich-nguyen:
This a temporary fix to suppress the amount of sentry errors being sent from our commercial app. The Sonobi script uses an es6 polyfill, which can result in an "Access Denied" issue in IE11, because the native implementation has different permissions to the user-space javascript implementation. The issue is mentioned [here](https://github.com/paulmillr/es6-shim/issues/333), although we can't take their fix right now.

I'll raise this with the Sonobi team; we should not allow `Object` polyfills to be added from our third party scripts.


regiskuckaertz (approved):
😭 

How convenient would it be if libraries exposed a set of assumptions rather than directly polyfill like it's Xmas, then _we_ could take care of it the way we want to.

rich-nguyen:
yeah @regiskuckaertz , the basic concepts of target platform, and module format, are both missing from js library repositories. Partly because target platform is a difficult classifier (ideally it would just be ecmaX, but in reality its 'safari this, edge that, opera, chrome with known issue in 20, etc'. Also, the execution environment is so mutable, there are few guarantees.

PR #15153, opened 2016-11-29
Caption competition

zeftilldeath:
## What does this change?
To indicate a caption we have used a camera icon/video icon or no icon. @paperboyo once correctly pointed out how odd it is that we use a camera icon to signify an illustration credit or copyright. Instead I have standardised them all with a universal '_i_' icon. Understood as meaning information. 

Like the majority of my PRs the value is immeasurable.

Nah. But we should update the apps icon too. 

## Screenshots

Close up:
screen shot 2016-11-29 at 14 03 03

Header test variants:
screen shot 2016-11-29 at 15 11 13

Normal usage:
screen shot 2016-11-29 at 15 17 28

Media tone:
screen shot 2016-11-29 at 16 20 49

Hosted content:
screen shot 2016-11-30 at 10 50 33

Video caption:
screen shot 2016-11-29 at 16 31 19

## Request for comment
@gtrufitt I bet you'd love to look at this. Have I forgotten anywhere?


paperboyo:
There is a precedent (in the Lightbox, if that goes out, I would italicise it too):

i icon

I worry that a camera/videocamera shape may be more legible at small size, though. Is that Egyptian Semibold Italic?

IMHO:
Pluses: reduces complexity, removes odd camera-for-illustration, camera-for-composite dilemmas, harder to confuse it with open-Lightbox action when Lightbox icon not present
Minus: removes a distinction between video and image captions (could this e.g. affect screen readers?)

paperboyo:
Oh, and this is part of iconography that is also used consistently in Apps, so...

zeftilldeath:
@paperboyo Going to swap the lightbox example to italic. And yeah it's DE5.

Currently, video caption do not have an icon at all. With regards to legibility, I'd rather have an _arguably_ legible icon than one that makes no sense but it clear. 

Yep Apps will need to follow suit if this goes out.

zeftilldeath:
screen shot 2016-11-29 at 17 24 42

paperboyo:
Just because we spoke of it (something is even less right than we thought here):

what
[[URL](https://www.theguardian.com/us-news/2016/nov/30/political-correctness-how-the-right-invented-phantom-enemy-donald-trump)]

zeftilldeath:
@paperboyo As always, good find!

zeftilldeath:
screen shot 2016-12-01 at 15 00 32

static/src/images/global/information-circle-grey.svg
```diff
@@ -0,0 +1 @@
+<svg width="14" height="14" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 14 14"><path fill="#767676" d="M7 0C3.1 0 0 3.1 0 7s3.1 7 7 7 7-3.1 7-7-3.1-7-7-7zm.9 10l-.3 1c-.4.2-1.2.4-1.9.4-.5 0-.8-.1-.8-.6 0-.1 0-.3.1-.4L6.3 6h-.9l.3-.8 2.7-.3.2-.1L6.8 10h1.1zm-.2-5.8c-.6 0-.9-.3-.9-.8 0-.6.5-1 1.1-1 .6 0 .9.3.9.8 0 .6-.5 1-1.1 1z"/></svg>
```
gtrufitt: Please can you remove the `xmlns=""` from the SVGs? It borks stuff in IE8. Thx.

static/src/stylesheets/module/external/_from-content-api.scss
```diff
@@ -167,11 +167,26 @@
...
-    @include mq(leftCol) {
-        padding-bottom: $gs-baseline * 2;
+    .inline-information {
+        margin-right: $gs-gutter/10;
+        margin-top: -$gs-baseline/4;
+        width: 14px;
+        height: 14px;
+        background-color: $neutral-2;
+
+        svg {
```
gtrufitt: Should try and use a class here instead of style the element. Should be a `.inline-information__svg` class I think

gtrufitt (approved):

static/src/images/global/information-circle-grey.svg
zeftilldeath: Hey I may be wrong. But I had to leave it in here, cos this svg gets base 64 encoded and inserted as a pseudo element. PLZ HELP!

paperboyo:
Brilliant, I’m already used to it!

One more miss, though (probably broken for a long time): video embeds in a liveblog:

liveblog video embed
[[URL](https://www.theguardian.com/commentisfree/live/2016/dec/02/why-is-there-a-sleeplessness-epidemic-join-our-live-look-at-the-week)]

paperboyo:
Not sure if it's related to this change ([this?](https://github.com/guardian/frontend/blob/master/common/app/views/support/cleaner/VideoEmbedCleaner.scala#L27)), but I don't remember sharing icons being cut off like that before:

picture 83
[[URL](https://www.theguardian.com/commentisfree/2016/dec/04/boris-johnson-pluralist-migration-brexit)]

PR #15154, opened 2016-11-29
Css in AMP pages 

NataliaLKB:
## What does this change?
Trello card: https://trello.com/c/oA88V5n2/157-amp-css-should-be-minified

There are many improvements that can be made to the css in AMP pages. As a start, these are the ones I am tackling: 

- We should be able to write Sass and use our variables and mixins
- We should be linting our css to keep consistency with the rest of the code base
- It should be minified

This PR only contains the header styles as I just want to test one change at a time. More PR's to follow!

We are quickly coming close to our limit of characters allowed in the css in our amp pages. The main reason why is because the css is not minified. This solves that problem.

Also, the developer experience with creating more css styles in amp is pretty awful. This makes it a bit better at least.

This affects only amp

## Screenshots
Things don't look different yet.

## Request for comment
@guardian/dotcom-platform


NataliaLKB:
@Calanthe @lps88 If you guys have time to look into moving the glabs and hosted amp styling that would be amazing. If not, I will do it and CC you in the PRs :)

lps88:
@NataliaLKB Amazing, I was wondering if there was a better way to include these styles. I'll do the hosted ones today 👍

SiAdcock:
This is totally awesome, thanks @NataliaLKB! 🎉 😄 🎉 

I have a developer experience question that could potentially be addressed in a subsequent PR if necessary. Would the Play app automatically pick up changes to the generated CSS file, say if I was developing using `make watch`? Or would I need to force a recompile?

Calanthe:
This is nice! Thanks @NataliaLKB 

We have to remember that hosted pages rely on variables passed from CAPI (https://github.com/guardian/frontend/blob/00c0061f233f131f229bb67ada9f14cbcaff1dfb/common/app/views/fragments/amp/stylesheets/main.scala.html#L262). So some styles will have to stay in scala.html file.

gtrufitt:
> Would the Play app automatically pick up changes to the generated CSS file, say if I was developing using make watch?

We would have to do the [same as we do on the website](https://github.com/guardian/frontend/blob/a198ea43c64947464addca45750641d9fd3c39d4/common/app/views/fragments/stylesheets.scala.html#L53).

This is a great step! Was definitely thinking about the whole 50k CSS thing, we should think about our main website CSS as well and how we can promote good CSS hygiene there in a similar way. Anyhoo, I ramble.

👍

NataliaLKB:
@SiAdcock 
> I have a developer experience question that could potentially be addressed in a subsequent PR if necessary. Would the Play app automatically pick up changes to the generated CSS file, say if I was developing using make watch? Or would I need to force a recompile?

Great question! At the moment because these styles are inlined, the page has to be compiled. So you could still used make watch, but you would need to refresh the page just like when you change any scala play template. But you are right that I should look into this more in other PRS.

PR #15156, opened 2016-11-29
Switch front email style by param

joelochlann:
We're going to be testing two design variants of the front email against each other, so we need to be able to switch between them based on a URL parameter.

To achieve this, we include an override stylesheet for the "connected" design if `format=email-connected` is passed. (To get the existing design, you can pass `format=email`)

#### "Card" design, `format=email`
picture 345

#### "Connected" design, `format=email-connected`
picture 344


common/app/implicits/Requests.scala
```diff
@@ -24,7 +24,7 @@ trait Requests {
 
     lazy val isAmp: Boolean = r.getQueryString("amp").isDefined || (!r.host.isEmpty && r.host == Configuration.amp.host)
 
-    lazy val isEmail: Boolean = r.getQueryString("format").contains("email") || r.path.endsWith(EMAIL_SUFFIX)
+    lazy val isEmail: Boolean = r.getQueryString("format").exists(_.contains("email")) || r.path.endsWith(EMAIL_SUFFIX)
```
joelochlann: I'm assuming that Fastly will let `format=email-connected` through, since it looks to me like the parameter just has to *start* with `email`: https://github.com/guardian/fastly-edge-cache/blob/312b4728f0799f2677e48c249085c08faee08da2/theguardiancom/src/main/resources/varnish21/backend-nextgen.vcl#L79

joelochlann:
Anyone from @guardian/dotcom-platform available to give this a 👀 ?

facia/app/views/fragments/emailFrontBody.scala.html
```diff
@@ -19,9 +19,12 @@ <h3 class="headline">
                 <span class="kicker-separator">/</span>
             }
 
-            @if(header.isGallery) { @icon("gallery") }
-            @if(header.isAudio) { @icon("podcast") }
-            @if(header.isVideo) { @icon("video") }
+            @defining(if (request.getQueryString("format").contains("email-connected")) "-connected" else "") { suffix =>
```
SiAdcock: Can we use `request.isEmail` here?

SiAdcock:
In general, I think this is cool. If this is a test, can we also get a switch around it so we can define an explicit expiry date?

common/app/views/mainEmail.scala.html
```diff
@@ -34,6 +34,11 @@
         @if(page.metadata.isFront) {
             @fragments.email.stylesheets.front(page)
             @fragments.email.stylesheets.tones()
+
+            @* This is a temporary override to test two alternative email designs 29/11/2016 *@
+            @if(request.getQueryString("format").contains("email-connected")) {
```
joelochlann: No, `isEmail` is for both "email" and "email-connected" styles - I could add an `isEmailConnected` into `Requests.scala` but it didn't quite seem worth it for this

SiAdcock:
Real good 👍

PR #15158, opened 2016-11-29
GA: Add user timing metrics

gtrufitt:
## What does this change?

GA allows you to push 'timing' metrics to page views, this adds the current user timing metrics to a 0.1% of page views sample size. Also reduces the current site speed sample size from 1% to 0.1% which reduces the overall requests to GA.

We can see the start and parse times for our JS over time in GA and easily compare segments to help identify issues.

## Request for comment

@dominickendrick @rich-nguyen @mkopka

### DIFF
--- a/common/app/views/support/GoogleAnalyticsAccount.scala
+++ b/common/app/views/support/GoogleAnalyticsAccount.scala
@@ -4,7 +4,8 @@ import conf.Configuration.environment
 
 object GoogleAnalyticsAccount {
 
-  case class Tracker(trackingId: String, trackerName: String, samplePercentage: Int = 100)
+  // NOTE that the 'samples rates' when set to 0, seem to be 100%
+  case class Tracker(trackingId: String, trackerName: String, samplePercentage: Int = 100, siteSpeedSamplePercentage: Double = 0.1)
 
   // The "All editorial" property in the main GA account ("GNM Universal")
   val editorialProd = Tracker("UA-78705427-1", "allEditorialPropertyTracker")
--- a/static/src/javascripts/bootstraps/commercial.js
+++ b/static/src/javascripts/bootstraps/commercial.js
@@ -24,7 +24,8 @@ define([
     'commercial/modules/third-party-tags',
     'commercial/modules/paidfor-band',
     'commercial/modules/paid-containers',
-    'commercial/modules/dfp/performance-logging'
+    'commercial/modules/dfp/performance-logging',
+    'common/modules/analytics/google'
 ], function (
     Promise,
     config,
@@ -51,7 +52,8 @@ define([
     thirdPartyTags,
     paidforBand,
     paidContainers,
-    performanceLogging
+    performanceLogging,
+    ga
 ) {
     var primaryModules = [
         ['cm-thirdPartyTags', thirdPartyTags.init],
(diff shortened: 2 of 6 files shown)


penx:
https://i.ytimg.com/vi/fGl4LOAgW50/maxresdefault.jpg

static/src/javascripts/bootstraps/commercial.js
```diff
@@ -75,6 +77,9 @@ define([
         ['cm-ready', function () {
             mediator.emit('page:commercial:ready');
             userTiming.mark('commercial end');
+            robust.catchErrorsAndLog('ga-user-timing-commercial-end', function () {
```
rich-nguyen: we've got this logged in more detail on our dashboards, do we need GA measuring it too?

harrynorthover:
Still doing tracking eh?

common/app/views/support/GoogleAnalyticsAccount.scala
```diff
@@ -4,7 +4,8 @@ import conf.Configuration.environment
 
 object GoogleAnalyticsAccount {
 
-  case class Tracker(trackingId: String, trackerName: String, samplePercentage: Int = 100)
+  // NOTE that the 'samples rates' when set to 0, seem to be 100%
+  case class Tracker(trackingId: String, trackerName: String, samplePercentage: Int = 100, siteSpeedSamplePercentage: Double = 0.1)
```
regiskuckaertz: Shouldn't this be `siteSpeedSamplePercentage: Int = 10` instead?

gtrufitt: I actually do want it to be 0.1% sample for these (equals about 17k pageviews a day, enough for these metrics I think)

regiskuckaertz: oh of course, that was a dumb question 😅

static/src/javascripts/bootstraps/enhanced/main.js
```diff
@@ -170,5 +175,8 @@ define([
 
         // Mark the end of synchronous execution.
         userTiming.mark('App End');
+        robust.catchErrorsAndLog('ga-user-timing-enhanced-end', function () {
+            ga.trackPerformance('Javascript Load', 'enhancedEnd', 'Enhanced end parse time');
```
sndrs: is it more useful to record the duration than the start/end times?

gtrufitt: I wondered that, probably value in both, though I don't want to send a shed load of requests and we can probably (need to test it out for sure) use [GA's calculated metrics](https://support.google.com/analytics/answer/6121409?hl=en) to create a 'duration' metric from end - start.

static/src/javascripts/bootstraps/commercial.js
rich-nguyen: we've got per-module commercial timings, and timings over the commercial bootstrap too. I can see the value in having one place to see all timings, and I can also see a concern about sending out a lot of monitoring calls from the client. I don't mind which you choose, just highlighting it in case you decided to log commercial for our team's benefit.

gtrufitt: That's pretty nice resolution, I remember you mentioning that, would be interested to see your approach. But yeah, for now, I'll keep it in place and see how we get on (I'm not worried about number of calls to GA itself as there's a net reduction with this PR, but maybe it could make sense to hold back these calls until everything is loaded to avoid clogging the connections - will look at that separately though).

Thanks!

sndrs (approved):

PR #15161, opened 2016-11-29
Ask for front email images at 500 width

joelochlann:
Thanks to #15148 the max width of front emails is 500px, so we can save some bandwidth and increase loading speed by requesting these at 500 width rather than 580. (Article emails still have 580 max width).

There may be more optimisations we can do like this once we know sizes of the smaller images too. The Flyer, which is our first test case, only has large images.

cc @paperboyo because he loves images


katebee (approved):
👍 🎉

PR #15168, opened 2016-11-30
Remove deprecated Play.api.Crypto packages

TBonnin:
## What does this change?
Remove deprecated Play.api.Crypto packages

See https://www.playframework.com/documentation/2.5.x/CryptoMigration25
for more info about Crypto migration for Play 2.5

We only use Crypto to encrypt/decrypt cookie values (using AES
algorithm), which are relatively simple operations that can simply be
built using javax.crypto primitives
Therefore I have extracted the 2 methods (encryptAES and decryptAES) into our own Crypto object.
Given our simple use case I don't feel like adding a complex 3rd-partydependency like Kalium and Keyczar is required

Let me know what do you think

➡️  Play 2.5

## Request for comment
@alexduf @DiegoVazquezNanini @adamnfish @markjamesbutler

### DIFF
--- a/common/test/common/CryptoTest.scala
+++ b/common/test/common/CryptoTest.scala
@@ -1,0 +1,25 @@
+package common
+
+import org.scalatest.{FlatSpec, Matchers}
+
+class CryptoTest extends FlatSpec with Matchers {
+
+  val privateKey = "6FA3BE741DE87141F90EB3E67EB51976"
+  val message = "Les sanglots longs des violons de l'automne blessent mon cœur d'une langueur monotone."
+
+  "A valid encrypted string" should "be successfully decrypted" in {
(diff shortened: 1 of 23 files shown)


common/test/common/CryptoTest.scala
```diff
@@ -0,0 +1,25 @@
+package common
+
+import org.scalatest.{FlatSpec, Matchers}
+
+class CryptoTest extends FlatSpec with Matchers {
+
+  val privateKey = "6FA3BE741DE87141F90EB3E67EB51976"
+  val message = "Les sanglots longs des violons de l'automne blessent mon cœur d'une langueur monotone."
```
alexduf: I see what you did here 👍

alexduf (approved):
Nice!

I think it's the most reasonable option of all the options suggested in the Play documentation

PR #15170, opened 2016-11-30
Fix show more replies error

SiAdcock:
## What does this change?

Whenever a user expands a discussion comment thread by clicking the "Show x more replies" button, an error is thrown. This has been reported in Sentry [over 7000 times](https://sentry.io/the-guardian/client-side-prod/?query=is%3Aunresolved+hash&sort=freq&statsPeriod=14d) in the last 2 weeks.

Clicking anywhere on the page triggers the `module:clickstream:click` event. Handlers listening to this event receive a payload representing the element that was clicked, or the Boolean `false` if that element was removed from DOM. The "Show x more replies" button is [removed from the DOM](https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/common/modules/discussion/comments.js#L269) as soon as it is clicked.

 [`modules/discussion/loader`](https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/common/modules/discussion/loader.js#L283) listens to this event, but does not guard against the possibility of the payload being Boolean `false`, and hence throws an error.

I have added a guard in the loader component to ensure `clickspec` is truthy before attempting to access its properties. I have also added a comment explaining why `clickspec` might be false, as it is not obvious.

No

## Screenshots

picture 138

## Request for comment

@guardian/dotcom-platform @rich-nguyen @regiskuckaertz

### DIFF
--- a/static/src/javascripts/projects/common/modules/discussion/loader.js
+++ b/static/src/javascripts/projects/common/modules/discussion/loader.js
@@ -280,7 +280,11 @@ Loader.prototype.ready = function() {
     mediator.on('discussion:commentbox:post:success', this.removeState.bind(this, 'empty'));
 
     mediator.on('module:clickstream:click', function(clickspec) {
-        if ('hash' in clickspec.target && clickspec.target.hash === '#comments') {
+        if (
+            clickspec &&
+            'hash' in clickspec.target &&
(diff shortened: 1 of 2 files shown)


regiskuckaertz (approved):
LGTM 👍

PR #15173, opened 2016-12-01
Switch to HTTPS recommendations endpoint

davidfurey:
## What does this change?

Changes recommendations test endpoint to be https

Makes recommendations test work in production

No

## Request for comment

@katebee


katebee (approved):
👍 🔒

PR #15178, opened 2016-12-01
Opt-in controller and control group for Webpack test

SiAdcock:
## What does this change?

This change adds an OptInController and a control group for the Webpack test. 

We need to be able to test the success of Webpack in the frontend. To do so, it is convenient to set up an `OptInController`, allowing us to test Webpack in production. 

We also need a control group, whose experience will be compared with the Webpack trial group.

I have started setting up the Fastly cache splitting here: guardian/fastly-edge-cache#721 

No

## Request for comment

@guardian/dotcom-platform


NataliaLKB:
👍

PR #15179, opened 2016-12-01
Fetch hosted onward amp container from ajax domain

kelvin-chappell:
This fixes a bug in #15149 where the amp onward container isn't being fetched because the request is being made on the wrong domain.

/cc @guardian/labs-beta


kelvin-chappell:
Ignore this.

PR #15183, opened 2016-12-01
Do not attach raven to window app

rich-nguyen:
This change removes the raven attachment, I'm speculating that this may be behind the cause of the high volume of sentry errors seen since November 15. Related to this one, maybe: https://github.com/guardian/frontend/pull/14982 @SiAdcock

### DIFF
--- a/static/src/javascripts/projects/common/utils/raven.js
+++ b/static/src/javascripts/projects/common/utils/raven.js
@@ -1,21 +1,13 @@
 define([
     'raven',
+    'common/utils/config',
     'common/utils/detect'
 ], function (
     raven,
+    config,
     detect
 ) {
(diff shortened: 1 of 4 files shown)


SiAdcock:
Sounds good! 🎱 👍

static/src/javascripts/projects/common/utils/raven.js
```diff
@@ -1,21 +1,13 @@
 define([
     'raven',
+    'common/utils/config',
```
SiAdcock: Love it

rich-nguyen:
[Bug on Windows phone](https://trello.com/c/WPA5WmVP/217-bug-on-windows-phone)

PR #15184, opened 2016-12-01
Increase inline ads on the wide breakpoint

regiskuckaertz:
The spacefinder rules for inline ads work quite well on small viewports, where paragraphs are spread out vertically rather than horizontally. On desktop though, which stills accounts for the majority of our readership, these rules do not fare very well. As a result, many articles generate 0 impressions from inline MPUs. This is abysmal, inline MPUs bring a sizeable chunk of revenue.

We noticed that most of the spacefinder rules account for elements that are offset to the left-hand column, and that adding an MPU there would not create too much visual tension. In this PR, we introduce an alternative set of rules for the wide breakpoint -- essentially we get rid of most of the spacefinder rules -- and offset ads in the right-hand column (when there is nothing in it).

**Before**:
screenshot-www theguardian com-2016-12-01-16-09-33

**After**:
screenshot-localhost 9000-2016-12-01-16-16-40

We'd like to measure the impact of this change. Expectations are high in terms of revenue but we must make sure first that it doesn't disrupt user experience, or negatively affects advertising metrics (CTR, viewability et al). Therefore, we'll run this in an AB test with an X/X audience split. (Still waiting for the correct settings)


static/src/javascripts/projects/commercial/modules/article-body-adverts-wide.js
```diff
@@ -0,0 +1,241 @@
...
+            slotSelector: ' > p',
+            minAbove: isMobile ? 300 : config.page.hasShowcaseMainElement ? 900 : 700,
+            minBelow: adSizes.mpu.height,
+            selectors: {
+                ' > h2': {minAbove: isMobile ? 100 : 0, minBelow: 250},
+                ' .ad-slot': {minAbove: 500, minBelow: 500},
+                ' > :not(p):not(h2):not(.ad-slot)': {minAbove: 35, minBelow: 400}
+            },
+            filter: function(slot) {
+                if (!prevSlot || Math.abs(slot.top - prevSlot.top) - adSizes.mpu.height >= rules.selectors[' .ad-slot'].minBelow) {
```
rich-nguyen: `rules.selectors[' .ad-slot'].minBelow`?

static/src/javascripts/projects/commercial/modules/article-body-adverts-wide.js
```diff
@@ -0,0 +1,241 @@
...
+    function addArticleAds(count, rules) {
+        return spaceFiller.fillSpace(rules, insertInlineAds, {
+            waitForImages: true,
+            waitForLinks: true,
+            waitForInteractives: true
+        });
+
+        function insertInlineAds(paras) {
+            var slots = paras
+            .slice(0, Math.min(paras.length, count))
```
rich-nguyen: this whole file has non-indented chains actually, is that new style?

static/src/javascripts/projects/commercial/modules/article-body-adverts-wide.js
```diff
@@ -0,0 +1,241 @@
...
+            waitForInteractives: true
+        });
+    }
+
+    function addInlineAds() {
+        return addArticleAds(2, getRules())
+        .then(function () {
+            if (inlineSlots.length === 2) {
+                return addArticleAds(8, getLongArticleRules());
+            }
```
rich-nguyen: Is an `else` needed here? Just wondering if the intent is that when the length of `inlineSlots != 2`, should we still call `addSlots()`?

static/src/javascripts/projects/commercial/modules/article-body-adverts-wide.js
regiskuckaertz: Yep, that is the minimum space between two ad slots. This is only necessary when (1) there is an inline merch slot, or (2) when we insert MPUs during the second call ("long article ads")

rich-nguyen:
👍  ace. need to measure viewability and ad slot count, ideally.

static/src/javascripts/projects/commercial/modules/article-body-adverts-wide.js
```diff
@@ -0,0 +1,241 @@
...
+            return insertAdAtPara(paras[0], 'im', 'im');
+        }, {
+            waitForImages: true,
+            waitForLinks: true,
+            waitForInteractives: true
+        });
+    }
+
+    function addInlineAds() {
+        return addArticleAds(2, getRules())
```
regiskuckaertz: The spacing rules are different in the latter case.

rich-nguyen:
[spacefinder improvements: inlines can be located on the right](https://trello.com/c/gUOnQaum/208-spacefinder-improvements-inlines-can-be-located-on-the-right)

static/src/javascripts/projects/commercial/modules/article-body-adverts-wide.js
regiskuckaertz: Given that long article ads are more constrained, if less than 2 ad slots where added on the first call, then it is impossible that others will be added on the second.

static/src/javascripts/projects/commercial/modules/article-body-adverts-wide.js
regiskuckaertz: Ha, kind of 😅  C-like languages already have too much indentation, I feel like the `.` is enough cue to notice it is a continuation of the previous line.

PR #15185, opened 2016-12-01
Remove first-child colour differentiation until logic is added (New header)

zeftilldeath:
Title explains it all.


stephanfowler:
👍

PR #15191, opened 2016-12-02
Paid for immersive

blongden73:
Adds the ability to turn the immersive switch on for advertiser content / paid stories. Some simple styling updates to colour and typeface within the immersive header. 

**Before**
picture 246

**After**
picture 247

@lps88 @SiAdcock @NataliaLKB @Calanthe


static/src/stylesheets/module/content/_article-immersive.scss
```diff
@@ -117,8 +117,23 @@
...
     }
+
+    &.content__headline--advertisement {
+        font-family: $f-sans-serif-headline;
+        font-weight: 200;
+    }
+}
+
+.paid-for-label__immersive {
+    font-size: 1rem;
```
SiAdcock: Is this overriding anything? Is `1rem` not the default font size?

blongden73: Ooo good spot this was...but i removed the html. I'll remove this.

common/app/model/meta.scala
```diff
@@ -569,6 +569,8 @@ final case class Tags(
 
   lazy val isClimateChangeSeries = tags.exists(t => t.id =="environment/series/keep-it-in-the-ground")
   lazy val isTheMinuteArticle = tags.exists(t => t.id == "tone/minute")
+  //this is for the immersive header to access this info
+  lazy val isImmersiveAdvertisement = tags.exists( t => t.id == "tone/advertisement-features" )
```
blongden73: Ah ok - the advertisment feature tone gets added automatically when it has a sponsored tag. So I guess making it more generic would be good.

static/src/stylesheets/module/content/_content.scss
```diff
@@ -724,6 +724,15 @@
     .from-content-api {
         @include f-textSans;
     }
+    .element.element-pullquote{
+      .pullquote-paragraph {
+        font-family: $f-sans-serif-headline;
+        color: $paid-article-brand;
+      }
+    }
+    a.in-body-link--immersive {
```
blongden73: This is just to make it more specific than just .in-body-link--immersive

Calanthe: Yes, I know. But do you need to be more specific than just `.in-body-link--immersive`? What other styles do you want to override?

blongden73: .content--immersive-article .in-body-link--immersive we have this but what i want is to change the colour when its ..content--advertisement-feature .in-body-link--immersive but if I don't use a. then the first one takes priority.

Calanthe:
👍

PR #15192, opened 2016-12-02
Add 2x2 for closing slots

regiskuckaertz:
Right now 1x1 is used for several purposes, and it is problematic as the onus is on AdOps to make sure the targeting will not have unintended side-effects—like for instance delivering a page skin in an MPU slot.

We add the 2x2 size, which sole intent will be to close slots. AdOps will then update their line items to use 2x2 instead of 1x1 for that purpose, so that later on we can update the [post-rendering callback](https://github.com/guardian/frontend/blob/rk-add-2x2/static/src/javascripts/projects/commercial/modules/dfp/render-advert.js#L75).


regiskuckaertz:
[Add the 2x2 size on all slots but the pageskin](https://trello.com/c/5WseqyR9/223-add-the-2x2-size-on-all-slots-but-the-pageskin)

rich-nguyen:
👍

PR #15193, opened 2016-12-02
Add youtube atom placeholder image and play button

GHaberis:
## What does this change?

- If youtube atom has posterURL show placeholder image/overlay
- fade in button/control bar on video ready
- fade out overlay on video start
- fade in on video end

- Adds Guardian styles to initial youtube atom

- No

## Screenshots

youtube


static/src/stylesheets/head.content.scss
```diff
@@ -38,3 +38,4 @@
    ========================================================================== */
 
 @import 'module/atoms/_quiz';
+@import 'module/atoms/_youtube';
```
GHaberis: @SiAdcock @NataliaLKB - looking for your frontend guidance here! Does this seem like the correct place to import these styles?

static/src/stylesheets/module/atoms/_youtube.scss
```diff
@@ -0,0 +1,196 @@
...
+                position: absolute;
+                height: 100%;
+                width: 100%;
+                left: 0;
+            }
+
+            &::before {
+                content: '';
+                opacity: 0.1;
+                background: #333333;
```
SiAdcock: Can we use `$neutral-1`?

static/src/stylesheets/module/atoms/_youtube.scss
```diff
@@ -0,0 +1,196 @@
...
+                    text-shadow: 0 0 16px rgba(51,51,51,0.25);
+
+                    &::before {
+                        content: '';
+                        width: 2px;
+                        height: 16px;
+                        opacity: 0.5;
+                        position: absolute;
+                        left: 30px;
+                        background-color: #ffbb00;
```
SiAdcock: `$multimedia-main-2`

SiAdcock:
Love it dude. I have a couple of practical suggestions that I'll be happy to discuss:

- The stylesheet is very nested, despite all classes being very BEM-y. I wonder if we could flatten the structure, since the class names are already so specific?
- The `setDuration` function is, in my opinion, doing too much. It looks like it could be split into a function that builds the `times` array and another that updates the overlay.
- I'm not totally okay with the mutations to the `duration` argument, perhaps this could be done in a more functional way. Not a showstopper though 😄

common/app/views/fragments/atoms/youtube.scala.html
```diff
@@ -10,10 +10,27 @@
...
 
         @defining(if(!host.isEmpty) "&origin=" + host else "") { origin =>
-            <iframe id="youtube-@media.assets.head.id" width="100%" height="100%"
+            <iframe class="atom--media--youtube--iframe" id="youtube-@media.assets.head.id" width="100%" height="100%"
             src="https://www.youtube.com/embed/@media.assets.head.id?modestbranding=1&enablejsapi=1&rel=0&showinfo=0@origin" frameborder="0"
             allowfullscreen="">
             </iframe>
+
+            @media.posterUrl.map { image =>
+                <div class="atom--media--youtube--overlay">
```
NataliaLKB: This html is quite nested. Is it necessary to write it like this?

static/src/stylesheets/head.content.scss
NataliaLKB: @GHaberis where does our current video styling come from? It would be good to put it in the same place. It might work out well to have a comment explaining that it may need to be deleted once the youtube migration is complete (unless I have misunderstood and that isn't happening 😆 )

static/src/stylesheets/module/atoms/_youtube.scss
```diff
@@ -0,0 +1,196 @@
...
+}
+
+.atom--media--youtube {
+	position: relative;
+
+	.atom--media--youtube--iframe,
+	.atom--media--youtube--overlay,
+	.atom--media--youtube--placeholder {
+		height: 100%;
+        width: 100%;
```
NataliaLKB: Indentation looks a bit weird here

NataliaLKB:
@GHaberis I agree with everything @SiAdcock says. I just have a question... I have the perception that the youtube player is replacing our current player. Is that true? Is there a way to remove (or make it clear that you are going to remove) any of the hopefully-soon-will-not-be-needed css and js.

I could be wrong though, just let me know  if that is the case.

GHaberis:
@SiAdcock @NataliaLKB I've simplified the HTML and managed to remove a few levels of indentation, I've also attempted to flatten the SCSS and removed some of the nesting in here too. I've kept some nesting where it's saved having to repeat target selectors.

@NataliaLKB we're trialling the youtube player as a potential replacement of the current media player. As it's a trial we're not going to remove any of the current media player's css and js until we've completed the trial and made a call about whether we're going to use the youtube player or not...Although thinking about it, I think we'll always need the current players css/js for old articles with videos as they'll continue to use this player? This is something that needs to be clarified!

GHaberis:
@SiAdcock definitely up for revisiting setDuration if you have some recommendations!

SiAdcock:
Love the changes, dude 👍

NataliaLKB:
@GHaberis Thanks for clarifying!

common/app/views/fragments/atoms/youtube.scala.html
NataliaLKB: I also wanted to check about the classnames. Based on these it seems that `media` is a modifier on `atom` and `overlay` is a modifier of `atom--media--youtube`. Regarding the former, is there an atom class already that you are modifying? It doesn't appear so in the code. 

I would think the styling for the youtube player would be unique and should have a unique classname as well. Maybe something like `youtube-atom` and overlay, as it is a child of the youtube atom could be `youtube-atom__overlay`.

Let me know what you think. Naming is hard, but essential to keep code readable.

static/src/stylesheets/module/atoms/_youtube.scss
```diff
@@ -0,0 +1,201 @@
...
+
+	.atom--media--youtube--placeholder {
+		background-size: cover;
+        background-position: 50% 50%;
+        background-repeat: no-repeat;
+        z-index: 1;
+        text-align: center;
+	}
+
+    .atom--media--youtube--play-button {
```
NataliaLKB: Given that this is the same styling as our current play button, is it worth writing this twice? Given that it is likely we will keep the old styling, I would prefer that we have as little duplication as possible to still keep it readable.

common/app/views/fragments/atoms/youtube.scala.html
GHaberis: @NataliaLKB I followed the pattern used in the existing classname 'atom--media--youtube' which was already in use in this template, however I agree the classnames if we're following BEM are misleading and I'm happy to correct this. So if 'atom--media--youtube' is an element in it's own right then I'll rename it to 'youtube-atom' and all sub elements to use '__' eg. 'youtube-atom__overlay' as suggested!

NataliaLKB (approved):
This is great! Thanks so much for renaming things. Its so much more clear now. 

There is a deletion in the doc/README which is probably an accident. I have done that before. Once that is fixed, this PR is ready to be merged 👍

GHaberis:
@NataliaLKB thanks! Yeah the README is automatically updating when I push to the repo, from the looks of it that comment should be removed though as the path egyptian-webfont-rendering-tests/*.md doesn't exist.

NataliaLKB:
@GHaberis great thanks! 😄

paperboyo:
Hi, @GHaberis! Sorry, could you indulge the code-blind, pls? Does this poster image have to use `background-image` or could it, maybe, use Picture Element?

I’m asking because ideally (not for this PR!), for best quality, smallest filesize and consistency, it should follow the logic all other images follow: for the mediaservie’s master asset to go through the resizer to provide differently-sized, prettified assets of differing quality/file format based on displayed size by breakpoint, client’s DPR and UA.

If it definitely has to be `background-image`, we may also be able to use media queries to realise most (all?) of the above, but I wouldn’t know for sure or how much.

For an example of how dotcom uses imagery, you can take a look at any front card or in-article image’s Picture Element syntax.

cc/ @akash1810, @gidsg just because we were talking about it... and because, to my horror,  I can see that atoms provide only a single, direct media service’s crop URL as `posterUrl`, which is positively XIX century stuff 😆 (and means that, potentially, my dumbphone will be downloading a 5000px PNG at 460px wide viewport or vice versa). Even aunt CAPI emits an URL to the master asset with an `isMaster` flag... Thorough :facepalm: for me [missing it](https://github.com/guardian/content-atom/pull/33)! Or am I missing something still? Probably! 😃

GHaberis:
Hi @paperboyo! OK thanks for raising this. 

So, the main reasons behind using background image were that the initial implementation of youtube atoms implemented in hosted articles use background images for the placeholder, and this was referred to as an example of how it should behave. (eg. https://www.theguardian.com/advertiser-content/explore-canada-food-busker-in-canada/return-to-niagara).

I also wasn't sure I could guarantee the image size would adhere to the 16:9 ratio, so using background-size/background-position available to background-image meant I could easily position the image as desired. I have however since seen the grid crop functionality in action so now know it will be 16:9 (@akash1810 this is guaranteed?).

Basically it doesn't have to be background image, we can use a picture element, however that doesn't get around the issue that we're still only served a single image asset that hasn't been through the resizer! If this is something we plan to address then yes of course I can update this to follow the same behaviour as all other images.

paperboyo:
Brilliant, @GHaberis, exactly what I wanted to know.

Yes, media atom model would have to be extended to include master asset for proper image consumption on-platform. Mining mediaID from the crop URL would be daft, I think.

> I also wasn't sure I could guarantee the image size would adhere to the 16:9 ratio,

Very good point and that’s something that even legacy video route does not handle well (raised and example of bad UX that follows [here](https://github.com/guardian/media-atom-maker/issues/118)). I would say media atom should only allow for the images to be the ratio of the video file (sometimes off-16:9). And the Grid should allow to be invoked to the cropping stage from media-atom-maker and other tools at arbitrary-ratio cropping stage where there would be just one choice of crop ratio available to the user: “Native” (native to the element that Grid was invoked from). Whom should I mention here? Hmm... Maybe loosely @clloyd (because [this](https://github.com/guardian/media-atom-maker/pull/89)), @akash1810 (bacause Grid) and @blishen (just to annoy him)?

akash1810:
(late response because been on leave 😄)

> I also wasn't sure I could guarantee the image size would adhere to the 16:9 ratio

We can't guarentee the ratio as the atom models the poster image as just a simple URL. We could add some validation on the client side of media atom maker, however this is one of many clients that can create media atoms...

PR #15194, opened 2016-12-02
Twitter card metatag set to "summary_large_image" for any article

lmath:
## What does this change?
Among the twitter metatags for an article, there is a metatag "twitter:card". For any article, including liveblogs, this will have value "summary_large_image".

Previously, this was set to "summary" for liveblogs.

Articles linked on twitter will show the same type of twitter card.

This is a minor fix and won't be measurable. 

No.

## Screenshots
The twitter card metatag used for liveblogs indicated a summary card like the following:
Summary card

After this change, liveblogs, like other articles should have a metatag for "summary_large_image" so that the twitter card will appear like the following for liveblogs:
Summary large image card

## Request for comment
@katebee @joelochlann 

…liveblogs.


joelochlann:
👍  Looks good to me, but I'd get someone from @guardian/dotcom-platform to have a quick look too since this was from their backlog

gtrufitt:
LGTM, thanks! Think you need to rebase from master as failing on an expired switch and your new test might not be passing :)

katebee (approved):
👍

PR #15197, opened 2016-12-02
Centered-icon additions

zeftilldeath:
Few elements did not have centered-icon so were showing up misaligned after the class was inlined.


NataliaLKB (approved):
👍

PR #15199, opened 2016-12-02
Pass title with correct capitalisation to GA

jorgeazevedo:
## What does this change?

This patch changes the title field for GA calls.

Due to misscommunication, the title field started being lowercased for all page views about a week ago. The title field is a cornerstone in the definition of the pageview and is now inconsistent with every other platform and with historical web data.

This is a significant bug that affects our base reporting and needs to be fixed asap. The title has been tracked like this for over a week now.

The code change doesn't affect other platforms per se.

## Screenshots

Here's an example GA report using the Total Views aggregate metric that brings together pageviews from various platforms.

inconsistent_capitalisation_-_analytics_

## Request for comment

@gtrufitt @dominickendrick @mkopka

### DIFF
--- a/common/app/views/fragments/analytics/google.scala.html
+++ b/common/app/views/fragments/analytics/google.scala.html
@@ -30,8 +30,8 @@
     }
 
     function getAnalyticsTitle() {
-        var analyticsTitle = (guardian.config.page.webTitle || '').toLowerCase();
-        return (analyticsTitle === 'network front') ? getAnalyticsEdition() + ' ' + analyticsTitle : analyticsTitle;
+        var analyticsTitle = (guardian.config.page.webTitle || '');
+        return (analyticsTitle === 'Network Front') ? getAnalyticsEdition() + ' network front' : analyticsTitle;
     }
 
     var identityId = null;


common/app/views/fragments/analytics/google.scala.html
```diff
@@ -30,8 +30,8 @@
     }
 
     function getAnalyticsTitle() {
-        var analyticsTitle = (guardian.config.page.webTitle || '').toLowerCase();
-        return (analyticsTitle === 'network front') ? getAnalyticsEdition() + ' ' + analyticsTitle : analyticsTitle;
+        var analyticsTitle = (guardian.config.page.webTitle || '');
+        return (analyticsTitle === 'Network Front') ? getAnalyticsEdition() + ' network front' : analyticsTitle;
```
gtrufitt: Is 'network front' lowercased here consistent with apps then or should that be left as 'Network Front'?

gtrufitt (approved):
Thanks Jorge!

common/app/views/fragments/analytics/google.scala.html
jorgeazevedo: The title in apps is "uk network front", so this needs to be lowercased here. It's unfortunate that apps and web's backends don't produce similar titles for these things, the algorithm here ends up being quite unnatural.

gtrufitt: Ok, that makes sense 👍

jorgeazevedo:
@gtrufitt thanks for approving! Can you merge for me? I don't have write access to this repository.

gtrufitt:
Sure!

PR #15202, opened 2016-12-02
You probably can't become a supporter < 1 min

philwills:
## What does this change?
Tweaks wording for membership call to action


rich-nguyen:
👍

PR #15203, opened 2016-12-02
Paid for containers: update the layout on mobile

regiskuckaertz:
According to spec:
localhost-9000-commercial-containers iphone 6

Before:
m code dev-theguardian com-commercial-containers iphone 6

cc @kelvin-chappell @ScottPainterGNM @JonNorman


kelvin-chappell (approved):
Looks good!
Could possibly reduce the duplication of same template code but not sure there would be much benefit.
👍

regiskuckaertz:
[Paid for and Supported By Container Block design is incorrect](https://trello.com/c/CAftOZxu/174-paid-for-and-supported-by-container-block-design-is-incorrect)

PR #15204, opened 2016-12-02
Recommendations styling

joelochlann:
Fix some styling issues with #15071, detailed below, plus a JS change to fix a bug where video articles didn't have a main image.

There is still an issue with some images having a different size ratio than the container they need to fit into (you can see it in the Selasi Bake-off image below) - I'll look at that in another PR

@davidfurey @katebee @lmath 

### mobile/tablet before
picture 346
picture 347

### mobile/tablet after
picture 348
picture 349

### link behaviour before
recommended-links-before

### link behaviour after
recommended-links-after

### DIFF
--- a/static/src/javascripts/projects/common/modules/experiments/tests/recommended-for-you.js
+++ b/static/src/javascripts/projects/common/modules/experiments/tests/recommended-for-you.js
@@ -120,6 +120,8 @@ define([
                 return imageFromTemplate(item.headerImage.urlTemplate);
             } else if (item.headerVideo) {
                 return imageFromTemplate(item.headerVideo.stillImage.urlTemplate);
+            } else if (item.video) {
+                return imageFromTemplate(item.video.stillImage.urlTemplate);
             } else {
                 return null;
             }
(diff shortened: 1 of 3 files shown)


paperboyo:
Ideally, we would have full card styling for all tones - colour, iconography, trail images, save for later and comment counts, headline cleaning, standfirst, kickers. Of course, if that means that the temporary MAPI route chosen for the test would require repeating all fronts styling, this may not viable... I wouldn’t know, but I wonder if not not going through MAPI wouldn’t just be easier? One _could_ argue that the lack of any of these missing elements could have negative effect on the test outcome 😉 

Here is a gallery card in the Recommended for you container (L) and the same card from dotcom proper (R):

plain-vs -full

joelochlann:
@paperboyo agreed! The issue we have is that this is a clientside template so we don't have access to all the logic in the Scala templates for rendering all the details on facia cards. However @davidfurey is looking into refactoring this so we fetch the card HTML from an endpoint on the server which may solve this problem. Closing this PR until further notice so as not to clutter up the frontend queue

paperboyo:
@joelochlann Cool! I hardly understand the difference between client and server, but whenever I was blabbing about some obvious-stuff-that’s-gonna-happen-one-day-anyway-cause-it’s-only-logical-and-cool, like rendering search results as a front full of cards or having fronts rendered for any arbitrary CAPI query*, I was told it’s entirely impossible because-something-I-don’t-understand-thus-don’t-care-about.

Maybe @davidfurey’s work you mentioned will bring these things closer to possible? That would be super-cool!

*some work on that was done by @tsop14, @reettaphant and @piuccio during a Hackday, maybe they would have some insight?

piuccio:
@paperboyo yeah the hard bit was reusing the code to render cards / containers. It's tightly coupled to fronts data structure. Once that is made generic it should be relatively easy to render search results as fronts

PR #15206, opened 2016-12-03
Remove hard limit on number of MPUs in galleries

regiskuckaertz:


regiskuckaertz:
[Galleries - Increase the no of MPU from 2 to more (as many the page will accept based on the no of images )](https://trello.com/c/I7VOxHYp/211-galleries-increase-the-no-of-mpu-from-2-to-more-as-many-the-page-will-accept-based-on-the-no-of-images)

static/src/javascripts/projects/commercial/modules/gallery-adverts.js
```diff
@@ -64,7 +55,15 @@ define([
                 item.anchor.appendChild(item.adSlot);
             }
 
-    });
+        });
+
+    }
+
+    function getSlotNameForMobile(index) {
```
rich-nguyen: we should have this logic shared somewhere.

rich-nguyen:
👍

static/src/javascripts/projects/commercial/modules/gallery-adverts.js
regiskuckaertz: Yes! Whilst I was doing the change, I thought exactly the same. This should be in `create-slot`, and naming the slot at the calling site should be optional. Will change next 😃

PR #15207, opened 2016-12-03
Messenger: unify resize slot handling

regiskuckaertz:
Historically, we have introduced the ability to resize the top slot for Rubicon ads, in the particular case where even though DFP returned a 728x90 creative size, the Rubicon ad turned out to be 970x250. It was handled via a `window.postMessage` call that sends a message of type `set-ad-height`.

Later on, native ads had a very similar use case. We introduced the more general [`resize`](https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/commercial/modules/messenger/resize.js) event into our cross-frame protocol. This event takes care of resizing the iframe and its slot container.

Recently, AppNexus has had the very same need, although in the right-hand slot. We thus reused the `set-ad-height` message in yet another module, and it worked. But they have also asked for the same thing in the top slot. In theory, it should work, and indeed it does ... except that the `set-ad-height` does not resize the iframe (only the slot) and the AppNexus creative does not do it either. The result is we end up with a top slot that is the correct size, but the creative appears truncated.

Well, I thought that was enough forking. First, it is not good to rely on the creative to resize the iframe. In a SafeFrame context, this must be done by the parent page. Second, there was too much duplication and message handling.

In this PR, we morph the `set-ad-height` message into a `resize` one. This greatly simplifies the logic and makes everything work exactly the same.

### DIFF
--- a/static/src/javascripts/projects/commercial/modules/sticky-mpu.js
+++ b/static/src/javascripts/projects/commercial/modules/sticky-mpu.js
@@ -13,11 +13,14 @@ define([
 ) {
     var stickyElement = null;
     var rightSlotId = 'dfp-ad--right';
+    var rightSlot;
 
     function stickyMpu($adSlot) {
         if ($adSlot.data('name') !== 'right' || stickyElement) {
             return;
         }
+
+        rightSlot = $adSlot[0];
 
         var referenceElement = document.querySelector(config.page.hasShowcaseMainElement ? '.media-primary' : '.content__article-body,.js-liveblog-body-content');
         if (!referenceElement) {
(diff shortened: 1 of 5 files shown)


regiskuckaertz:
[Allow AppNexus ads to resize the top banner](https://trello.com/c/27D66Af0/178-allow-appnexus-ads-to-resize-the-top-banner)

static/src/javascripts/projects/commercial/modules/sticky-mpu.js
```diff
@@ -35,19 +37,15 @@ define([
...
-    function onAppNexusResize(specs, _, iframe) {
-        var adSlot = closest(iframe, '.js-ad-slot');
-        if (adSlot.id === rightSlotId) {
-            messenger.unregister('set-ad-height', onAppNexusResize);
-            fastdom.write(function () {
-                iframe.style.height = specs.height + 'px';
-                stickyElement.updatePosition();
-            });
+    function onResize(specs, _, iframe) {
+        if (rightSlot.contains(iframe)) {
```
rich-nguyen: where did the iframe height update go? magic!

rich-nguyen:
👍

static/src/javascripts/projects/commercial/modules/sticky-mpu.js
regiskuckaertz: In the [resize](https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/commercial/modules/messenger/resize.js) handler 😄 
Now both sticky-top and sticky-mpu can focus on making sure the stickiness works correctly. There is a bit of duplication now because Rubicon ads resize the iframe, but eventually AdOps can update the code in DFP.

PR #15208, opened 2016-12-05
Extend the membership international engagement banner test

rupertbates:
## What does this change?
Extends the membership international engagement banner test

This test is currently expired so I'm extending it until we get a proper decision what to do next

No


SiAdcock:
👍

JustinPinner (commented):
This kind of extension only works if the test hasn't expired when you do it. Otherwise, you're effectively defining a brand new test, with a new audience but using the old name. That's a viable option if the new expiry date is far enough away to give statistically significant results, but I'd still prefer to use a new test name to keep things clean.

JustinPinner:
We can extend the switch in isolation of the test if you just want frontend builds to stop breaking.

JustinPinner:
👍 We could switch the `safeState` to `off` too, but then again, we can just flick the switch in the admin tool anyway.

PR #15210, opened 2016-12-05
Support atoms in immersive main media

wpf500:
## What does this change?
Support atoms in the immersive article main media slot. ~~It uses the `MainCleaner`, in line with how the main media slot is handled in other content types (live blogs, non-immersive articles, etc.), which I think makes sense.~~

~~However, `MainCleaner` includes a number of cleaners and not just `AtomCleaner`. This solution means that all of those cleaners will now be applied, which may have been deliberately avoided in the past? A more conservative approach would be to just enable the `AtomCleaner`.~~

Now uses a new `ImmersiveMainCleaner`

## Screenshots
Before:
image

After:
image

### DIFF
--- a/static/src/stylesheets/module/atoms/_interactive.scss
+++ b/static/src/stylesheets/module/atoms/_interactive.scss
@@ -1,4 +1,9 @@
 .interactive-atom-fence {
     width: 100%;
     border: 0;
+
+    .immersive-main-media__media & {
+        // Don't allow the iframe to change its height, must be 100% in immersive main media slot
+        height: 100% !important;
+    }
 }
--- a/common/app/views/fragments/immersiveMainMedia.scala.html
+++ b/common/app/views/fragments/immersiveMainMedia.scala.html
@@ -5,8 +5,9 @@
 @import model.ContentPage
 @import views.support.RenderClasses
 @import views.support.TrailCssClasses.toneClass
+@import views.support.ImmersiveMainCleaner
 
-@(page: ContentPage)(implicit request: RequestHeader)
+@(page: ContentPage)(implicit request: RequestHeader, env: play.api.Environment)
 
 @defining((
     page.item.tags.isTheMinuteArticle,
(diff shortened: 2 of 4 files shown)


static/src/stylesheets/module/atoms/_interactive.scss
```diff
@@ -1,4 +1,8 @@
 .interactive-atom-fence {
     width: 100%;
     border: 0;
+
+    .immersive-main-media__media & {
+        height: 100% !important;
```
SiAdcock: Why is `!important` necessary here?

SiAdcock:
I can't see a problem with all the cleaners under `MainCleaner` being applied to the immersive main media content. Perhaps someone else from @guardian/dotcom-platform might have more context.

NataliaLKB:
Personally I wouldn't move code to `Common` from `Article`, which won't be used in other places other than `Article`. The reason why is that it is good to avoid unused code being compiled into every project we have. 

I think if all the rest of  the cleaners are necessary, this change is good. Otherwise, I think you should stick with just moving the atom cleaner

static/src/stylesheets/module/atoms/_interactive.scss
wpf500: Was thinking that also @NataliaLKB, just added one!

NataliaLKB: Thanks @wpf500 🌟

wpf500:
Sorry for the delay @NataliaLKB, @SiAdcock. I've made an `ImmersiveMainCleaner`, what do you think?

NataliaLKB (approved):
👍

wpf500:
Do you know why the build is failing? I don't have access to your TeamCity account.

sndrs:
@wpf500 commercial test is failing which seems unrelated. i'll re-run on TC, looks like a glitch

wpf500:
Woo thanks @sndrs, merging!

PR #15212, opened 2016-12-05
Add sonobi test url in code environment

rich-nguyen:
Adds the test url for sonobi, so that we can test the new version of the script which shouldn't throw IE11 errors.


regiskuckaertz (approved):
let's do this 💪

PR #15213, opened 2016-12-05
Caption cleanup

zeftilldeath:
Previously the caption icon wasn't showing alongside videos in live blogs. It is now. 
screen shot 2016-12-05 at 11 31 46

Also, video sharing buttons were being clipped, as far as I could see this is due to an overflow hidden. 

Before:
screen shot 2016-12-05 at 12 09 56

After:
screen shot 2016-12-05 at 12 09 45

Thank you @NataliaLKB for helping!


NataliaLKB (approved):
👍

paperboyo:
Thanks, @zeftilldeath! 🍻

PR #15214, opened 2016-12-05
Update jobs images scheme to https

mchv:
## What does this change?

Fix job images not displaying because on `http` scheme

Image displaying.

Image are not going to display as well.

## Screenshots

screen shot 2016-12-05 at 13 10 19

### DIFF
--- a/commercial/app/views/jobs/jobsBlended.scala.html
+++ b/commercial/app/views/jobs/jobsBlended.scala.html
@@ -4,7 +4,7 @@
 
 <div class="advert-blended advert-blended--jobs">
     <div class="advert-blended__title">
-        <a href="@{clickMacro}http://jobs.theguardian.com"
+        <a href="@{clickMacro}https://jobs.theguardian.com"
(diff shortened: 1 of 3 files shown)


philwills:
This only looks like it updates the test fixture. Is it supposed to contain anything else?

mchv:
@philwills completely missed the data were coming[ from a configured feed](https://github.com/guardian/frontend/blob/master/common/app/common/configuration.scala#L418)! So I have updated more urls to `https` to make my PR a bit less useless but obviously the problem will not be solved until the feed contains https links.

ScottPainterGNM:
Just noticed this but @JonNorman was looking into this and we have raised this with the merchandising team to get this feed updated. This is also an issue for some other feeds (like bookshop) so we're looking all of them.

ScottPainterGNM:
This is safe to close now @mchv as should now be fixed.

mchv:
@ScottPainterGNM thanks!

PR #15215, opened 2016-12-05
Hosted article - quotes styling

Calanthe:
## What does this change?
This PR introduces a proper styling of standard quotes (limited to one sentence) and inline quotes (usually longer that one sentence) on the hosted article pages.

Quotes look much better now.

Not yet. @blongden73 should I do the same styling in AMP?

## Screenshots
screen shot 2016-12-05 at 11 10 13

screen shot 2016-12-05 at 11 10 20

The non-red quote marks is a content issue and they will be removed in composer.

## Request for comment
@blongden73 @lps88 @kelvin-chappell


kelvin-chappell (approved):
👍

paperboyo:
These quote icons are long deprecated, dotcom contains the new, correct ones as inline SVGs, @Calanthe.

Calanthe:
@paperboyo This is the one I've got from @blongden73 
Do you remember the name of the icon being used now?

paperboyo:
This is taken from the opinion tone card: `class="inline-quote__svg inline-icon__svg"`. @blongden73: annihilate this ugly thing, nuke it, organise a black mass, stab it with a wooden stake and bury it deep.

blongden73:
@paperboyo Hey so we're using the quote mark from the icon set specifically to differentiate it from the other articles. In the same way we're using the sans.

paperboyo:
Aww, that’s awful, @blongden73. Guess, I’m not buying Honda after all 😉

> In the same way we're using the sans.
But the quote is in Egyptian, and thus off-style, in the illustration, no?

PR #15217, opened 2016-12-05
Screenshot tools and tasks

gtrufitt:
## What does this change?

This initial PR basically re-writes the existing screenshot tool (that no longer worked) and moves it into our `__task__` and `make` architecture.

Currently it only screenshots the three edition fronts, the intention being to extend this to a useful set of paths that include all content types. 

Long term - @MatthewJWalls is working on getting PR builds up into instances which we can then screenshot. We can also look at cross-browser screenshot or visual regression services and spit this into PRs. Grouping page types and focussing in on particular sections of pages may also be useful.

- Allows a simple visual check to catch oddities with changes
- Lays some of the groundwork for useful PR build screenshotting

A visual check of AMP will certainly be useful in the future.

## Screenshots

screenies

How do you screenshot screenshots.. wut

image

Bonus happy dog.

teeth

## Request for comment

@guardian/dotcom-platform


regiskuckaertz:
Oooh super interesting @gtrufitt 

We have a use case where this could come in handy, especially if we could pass an optional path as argument :-)

Might be worth discussing with @guardian/commercial-dev

michaelwmcnamara:
<iframe src="//giphy.com/embed/l0O7P6qdCa1AKJRAY" width="480" height="351" frameBorder="0" class="giphy-embed" allowFullScreen></iframe><p><a href="http://giphy.com/gifs/studio-90s-l0O7P6qdCa1AKJRAY">via GIPHY</a></p>

michaelwmcnamara:
yeah

tools/__tasks__/screenshot/screenshot.js
```diff
@@ -0,0 +1,51 @@
...
+ *  - Generate nicer viewing output
+ *  - Generate copy paste into github
+ *  - Screenshot 'components' by classname (captureSelector)
+ */
+
+const {paths, breakpoints, host, screenshotsDir, environment} = require('./config');
+
+function takeScreenshots() {
+    const taskArray =
+        paths.map((path) => {
```
sndrs: why not just return this `.map`?

sndrs: i.e. something like 
```js
task: paths.map(path => 
    Object.keys(breakpoints).map(breakpointName => {
        ...
    })
)
```

sndrs (approved):
couple of comments but not blockers - nice!

gtrufitt:
image

@SiAdcock @ScottPainterGNM used em

ScottPainterGNM:
I would be mad but this is a cause I'm very much behind

PR #15221, opened 2016-12-05
Correct the static component of the artifact generation

sihil:
## What does this change?
Fixes the naming of the static files in the unified deploy so that they are uploaded into the correct part of the S3 bucket.

The unified deploy uploads the static files into the correct location.

No


jfsoul (approved):
👍 

I looked up the `assets.path` config value for `CODE` (no surprises, it contains `/frontend-static/`)

sihil:
I suspect there is also some configuration in the CDN config for assets that I've been unable to find. Seems easier to not open this can of worms :)

jfsoul:
> I suspect there is also some configuration in the CDN config for assets that I've been unable to find. Seems easier to not open this can of worms :)

Good call!

image

PR #15222, opened 2016-12-05
Missed the dependency declaration

sihil:
## What does this change?
Whilst renaming the static deployment in #15221 I missed the dependency.


jfsoul (approved):
👍

PR #15223, opened 2016-12-05
Add module to check an A/B test against the targeting tool

desbo:
## What does this change?
The contributions team is planning to use the [targeting
tool](https://github.com/guardian/targeting) to manage the tags used in
Epic messages, so we can control where it shows up without having to
make any code changes. To help with this I've added a small module that
checks the campaigns in `guardian.config` (populated by the targeting
tool) against a given A/B test, which will give us a value to use in the
test's `canRun` criteria.

It'll allow us to make changes to the tags in Epic tests without raising PRs so there should be a bit less work for the frontend team. 
Note that this definitely won't rule out PRs for Epic tests altogether, but it's a step in that direction. 

Nope

## Request for comment
@philwills @guardian/contributions  @guardian/dotcom-platform

### DIFF
--- a/static/src/javascripts/projects/common/modules/experiments/ab-test-targeting.js
+++ b/static/src/javascripts/projects/common/modules/experiments/ab-test-targeting.js
@@ -13,10 +13,10 @@ define([
      * @param test      an A/B test object (with an 'id' property)
      * @return {Array}  campaigns associated with this test
      */
-    function campaignsFor(test) {
+    function campaignsFor(id) {
         try {
             return config.page.campaigns.filter(function (campaign) {
-                return campaign.fields && clean(campaign.fields.campaignId) === clean(test.id);
+                return campaign.fields && clean(campaign.fields.campaignId) === clean(id);
             });
         } catch (e) {
             return [];
@@ -30,7 +30,7 @@ define([
      * @return {Boolean}
      */
     function isAbTestTargeted(test) {
-        return campaignsFor(test).length > 0;
+        return campaignsFor(test.id).length > 0;
     }
 
     return {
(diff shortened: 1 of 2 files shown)


static/src/javascripts/projects/common/modules/experiments/ab-test-targeting.js
```diff
@@ -0,0 +1,39 @@
...
+    /**
+     * Get the campaigns for a test.
+     *
+     * @param test      an A/B test object (with an 'id' property)
+     * @return {Array}  campaigns associated with this test
+     */
+    function campaignsFor(test) {
+        try {
+            return config.page.campaigns.filter(function (campaign) {
+                return campaign.fields && clean(campaign.fields.campaignId) === clean(test.id);
```
gtrufitt: Feels like maybe the backend and the targeting tool model should define ID format to avoid this but ah well.

gtrufitt (approved):
👍

static/src/javascripts/projects/common/modules/experiments/ab-test-targeting.js
desbo: Yeah true, it's a user input field in the targeting tool so I'm not sure if they do any kind of processing on it, or want to. This is just a safeguard T.B.H 💂

gtrufitt: :guardswoman: Fair doos. Sounds like something to bring up with targeting so we can be sure of the format of campaign IDs because free text sounds like a world of pain

gtrufitt (approved):

gtrufitt (approved):

PR #15224, opened 2016-12-05
Serve hidden PROD fronts as email-friendly HTML

joelochlann:
This week, the Travel desk are going to be putting together their email ("The Flyer") using the new fronts tool functionality (#14236). Previously we'd been using code fronts for this purpose. We need to give them a way to create their front in the prod fronts tool, but not have it accessible via the website as a real front. So we'll tell them to make it a hidden front, and for that to work we need to `?format=email` endpoint to work for hidden fronts (which it doesn't currently).

@lmath @davidfurey @guardian/dotcom-platform

### DIFF
--- a/common/app/conf/switches/FeatureSwitches.scala
+++ b/common/app/conf/switches/FeatureSwitches.scala
@@ -487,4 +487,15 @@ trait FeatureSwitches {
     sellByDate = new LocalDate(2016, 12, 15),
     exposeClientSide = true
   )
+
+  // Owner: Joseph Smith
+  val DisplayHiddenFrontsAsEmails = Switch(
+    SwitchGroup.Feature,
+    "display-hidden-fronts-as-emails",
+    "Allows hidden fronts to be rendered as email-friendly HTML by passing ?format=email",
+    owners = Seq(Owner.withGithub("joelochlann")),
+    safeState = Off,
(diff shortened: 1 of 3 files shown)


joelochlann:
Also, I am aware that a better long-term solution may be to create a new Email front type (in addition to Training/Commercial/Editorial) in the fronts tool rather than overloading the notion of a hidden front. 

Right now people may have the expectation that a hidden front is totally inaccessible outside the Guardian - with this change, that would not exactly be the case as you could see what's in it by appending `?format=email` to the URL. If we had an Email front type, people would know to expect that email fronts are accessible via `?format=email` but not as a normal front on the website. Also, we could restrict the available layout types to make it clearer and less confusing (our plan is to have three for email: fast, medium and slow).

SiAdcock:
Nice ❤️ 👍

NataliaLKB (approved):
👍

TBonnin:
This is quite a dirty workaround that requires a lot of additional knowledge to understand why it is in the code in the first place. 
I understand this "feature" needs to happen asap but should we add unit tests for the `shouldServeFront` method? It would prevent accidental change of logic and will "document" this specific use case?

joelochlann:
@TBonnin I'll do that this afternoon

TBonnin:
Thanks @joelochlann

PR #15228, opened 2016-12-06
Tone stylesheets for emails have a background color for headlines & trail text

lmath:
## What does this change?
This is an update to the style sheet for tones used for the email format. 

This means that when displaying an email, that headlines and trail texts have an explicitly set background-color.

This improves how emails are rendered on gmail for android app.

I used litmus to check that how emails were rendered in other clients remained consistent with how they appeared previously. 

No.

## Screenshots

How the email appeared in gmail for android app before adding the background color
image

after adding background colors for headlines and trail texts:
image

## Request for comment
@joelochlann

### DIFF
--- a/common/app/views/fragments/email/stylesheets/tones.scala.html
+++ b/common/app/views/fragments/email/stylesheets/tones.scala.html
@@ -7,6 +7,7 @@
     }
     .tone-news .headline {
         color: #333333;
+        background-color: #f6f6f6;
     }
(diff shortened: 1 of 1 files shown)


joelochlann (approved):
Looks good! 💎  Thanks @lmath

lmath:
screenshot_20161207-103531

Not going to merge because when I received The Flyer and opened it with Gmail for Android, it looked fine.

PR #15229, opened 2016-12-06
Review MPUs on fronts

regiskuckaertz:
Currently, editors have full control over how many MPUs are inserted in fronts, both on desktop and on mobile; the focus is on desktop though, where they make sure the chemistry between headlines, MPUs and screen real estate is "just right."

This is to the detriment of mobile devices, where the screen ratio is very different, and there is no reason to tie MPUs to the same manual curation.

This PR addresses this by inserting MPUs after every other container on mobile, with an upper limit of 10 (the fact that this limit is never reached now is pure luck). Also a few other changes:

- `slice-adverts` will now only run on fronts and tags. Where it now runs everywhere but fails to add inline MPUs, again as a matter of pure luck—if it did, those slots would have the same IDs than those in the article, say.
- on desktop, the manual curation allows us to get rid of the redundant (and hard to follow) logic of filtering out containers that do not comply with five different criteria
- the insertion of ad slots is now completely non-blocking, which should speed up the commercial initialisation


regiskuckaertz:
[Increase MPUs on fronts on mobile](https://trello.com/c/8LNi6oEf/232-increase-mpus-on-fronts-on-mobile)

static/src/javascripts/projects/common/modules/commercial/commercial-features.js
```diff
@@ -82,6 +82,7 @@ define([
         this.sliceAdverts =
             this.dfpAdvertising &&
             !isMinuteArticle &&
+            config.page.isFront &&
```
regiskuckaertz: yes 😆  it is _pure luck_ that it just never introduced duplicate slots

regiskuckaertz: come to think of it, I can remove the test just above.

rich-nguyen:
lgtm 👍

PR #15231, opened 2016-12-06
Deal with both unified and componentised deploys

sihil:
## What does this change?
Adds logic to the deploy radiator to deal with both the existing componentised and the new unified deployments.

Since switching to the unified deploy in CODE the radiator has been misleading. This resolves that by looking at the unified deploy if it is more recent than the individual components. It should also (in theory) correctly work when we start using the unified deploy in PROD.

No.

## Screenshots
From this:
screen shot 2016-12-06 at 13 17 27

To this:
screen shot 2016-12-06 at 13 20 34

### DIFF
--- a/static/src/deploys-radiator/app/main.js
+++ b/static/src/deploys-radiator/app/main.js
@@ -37,33 +37,26 @@ const run = () => {
     const unifiedWhitelist = List([
         'all'
     ]);
-    const filterServerWhitelisted = (deploys) => deploys
-        .filter(deploy => serverWhitelist.contains(deploy.projectName))
+    const filterWhiteListed = (whitelist) => (deploys) => deploys
+        .filter(deploy => whitelist.contains(deploy.projectName))
         .toList();
-    const filterUnifiedWhitelisted = (deploys) => deploys
-        .filter(deploy => unifiedWhitelist.contains(deploy.projectName))
-        .toList();
+    const filterServer = filterWhiteListed(serverWhitelist);
+    const filterUnified = filterWhiteListed(unifiedWhitelist);
+    const timeOfMostRecent = (deploys) => getMostRecentDeploys(deploys).last().time.getTime();
     const mostRecentOf = (serverDeploys, unifiedDeploys) => {
         if (serverDeploys.isEmpty()) return unifiedDeploys;
         if (unifiedDeploys.isEmpty()) return serverDeploys;
-        const serverTime = serverDeploys.sort((a, b) => a.time.getTime() - b.time.getTime()).last().time.getTime();
-        const unifiedTime = unifiedDeploys.sort((a, b) => a.time.getTime() - b.time.getTime()).last().time.getTime();
-        if (serverTime > unifiedTime)
-            return serverDeploys;
-        else
-            return unifiedDeploys;
+        const serverTime = timeOfMostRecent(serverDeploys);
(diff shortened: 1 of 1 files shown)


static/src/deploys-radiator/app/main.js
```diff
@@ -34,14 +34,39 @@ const run = () => {
...
-    const filterWhitelisted = (deploys) => deploys
+    const unifiedWhitelist = List([
+        'all'
+    ]);
+    const filterServerWhitelisted = (deploys) => deploys
         .filter(deploy => serverWhitelist.contains(deploy.projectName))
         .toList();
+    const filterUnifiedWhitelisted = (deploys) => deploys
+        .filter(deploy => unifiedWhitelist.contains(deploy.projectName))
+        .toList();
```
jfsoul: Any reason not to combine these? (same goes for `serverTime` below) e.g.    

    const filterWhiteListed = (whitelist) => (deploys) => deploys
        .filter(...)
        .toList()

    const filterServerWhiteListed = filterWhiteListed(serverWhiteList)

static/src/deploys-radiator/app/main.js
```diff
@@ -34,14 +34,39 @@ const run = () => {
...
+    const mostRecentOf = (serverDeploys, unifiedDeploys) => {
+        if (serverDeploys.isEmpty()) return unifiedDeploys;
+        if (unifiedDeploys.isEmpty()) return serverDeploys;
+        const serverTime = serverDeploys.sort((a, b) => a.time.getTime() - b.time.getTime()).last().time.getTime();
+        const unifiedTime = unifiedDeploys.sort((a, b) => a.time.getTime() - b.time.getTime()).last().time.getTime();
+        if (serverTime > unifiedTime)
```
jfsoul: ternary might be neater

static/src/deploys-radiator/app/main.js
```diff
@@ -34,14 +34,39 @@ const run = () => {
...
         .filter(deploy => serverWhitelist.contains(deploy.projectName))
         .toList();
+    const filterUnifiedWhitelisted = (deploys) => deploys
+        .filter(deploy => unifiedWhitelist.contains(deploy.projectName))
+        .toList();
+    const mostRecentOf = (serverDeploys, unifiedDeploys) => {
+        if (serverDeploys.isEmpty()) return unifiedDeploys;
+        if (unifiedDeploys.isEmpty()) return serverDeploys;
+        const serverTime = serverDeploys.sort((a, b) => a.time.getTime() - b.time.getTime()).last().time.getTime();
+        const unifiedTime = unifiedDeploys.sort((a, b) => a.time.getTime() - b.time.getTime()).last().time.getTime();
```
jfsoul: Is this logic already in `getMostRecentDeploys` in `model-helpers.js`?

jfsoul (changes requested):
Would be good to re-use `getMostRecentDeploys` if possible. Otherwise minor style comments that don't necessarily need addressing.

static/src/deploys-radiator/app/main.js
sihil: Not quite, but I've tried using it and it seems to work as expected so the difference is possibly not important.

static/src/deploys-radiator/app/main.js
sihil: Too much time in scala :)

jfsoul (approved):
Thanks 👍

PR #15234, opened 2016-12-06
Continue css amp

NataliaLKB:
## What does this change?
This continues the progress made in https://github.com/guardian/frontend/pull/15154.

* Moves the footer styling into the static directory. This enables the use of our variables and mixins
* Does what @gtrufitt suggested https://github.com/guardian/frontend/pull/15154#issuecomment-263847642 so that `make watch` will work

Hopefully improves the developer experience with AMP

It affects AMP! 

## Request for comment
@zeftilldeath @SiAdcock @gtrufitt


common/app/views/fragments/amp/stylesheets/main.scala.html
```diff
@@ -282,3 +282,7 @@
     @Html(common.Assets.css.amp)
 
 </style>
+
+@if(play.Play.isDev() || !InlineCriticalCss.isSwitchedOn) {
```
gtrufitt: Would you ever want AMP to *not* inline the CSS? Say we decided suddenly that the website needed inline css switched off, we'd break AMP too so I think you probably just need the `isDev`?

gtrufitt (approved):

common/app/views/fragments/amp/stylesheets/main.scala.html
NataliaLKB: Actually this isn't our main mechanism for inlining css in AMP,  that is above. I think this needs a comment to explain that.  

This also won't pass validation in dev because this relies on a link tag which isn't allowed in amp. 

I am torn whether this is actually a good idea. It really does improve the developer experience, but does actually add complexity when checking if the amp pages are valid.

SiAdcock (approved):

NataliaLKB:
@gtrufitt @SiAdcock would you guys be able to take another look?

static/src/stylesheets/amp/_footer.scss
```diff
@@ -1,47 +1,62 @@
...
+    clear: both;
+}
+
+.l-footer__secondary {
+    font-size: 12px;
+    line-height: 1;
+    padding-right: $gs-gutter / 2;
+    padding-left: $gs-gutter / 2;
+
+    a {
```
gtrufitt: Why do we style elements directly here?

gtrufitt (approved):
👍🏻

PR #15240, opened 2016-12-06
Editorial email variants

katebee:
## What does this change?

Return of email AB testing! (see #14942) 💌

- Adds an AB test to trial two formats for an editorial email (patient 0 was Weekend Reading, we are now moving onto The Flyer), this requires subscribing users to two different ListIDs

- Bonus: fix the border of email input forms

Q: Why are we using the frontend AB testing framework to AB test email formats?

A: The AB function in ExactTarget re-randomises all recipients on each send, and we need users to receive their variant for several weeks. This test will ensure users are added to the corresponding email list (listId) in ExactTarget

Nope

## Request for comment

@joelochlann


static/src/javascripts/projects/common/modules/experiments/tests/editorial-email-variants.js
```diff
@@ -0,0 +1,72 @@
...
+        function updateExampleUrl(exampleUrl) {
+          return fastdomPromise.write(function () {
+            var example = $('.js-email-example')[0];
+            example.setAttribute('href', exampleUrl);
+          });
+        }
+
+        // loads the correct iframe if we can...
+        function enhanceWebView(emailListID) {
+          var insertionPoint = $('.js-email-sub__iframe')[0];
```
joelochlann: should this also be wrapped in `fastdomPromise.write()`?

katebee: Was internally debating it... does fastdomPromise wait for an opportune moment? Ideally I want the iframe updating as fast as possible.

joelochlann: it batches DOM manipulation calls together. but fair point, there's not much else happening on this page that we care about so probably fine

static/src/javascripts/projects/common/modules/experiments/tests/editorial-email-variants.js
```diff
@@ -0,0 +1,72 @@
...
+          $(insertionPoint).addClass('is-hidden');
+          var emailForm = $.create(
+                '<iframe src="/email/form/plaintone/' + emailListID +
```
joelochlann: as discussed - let's try just changing the iframe `src` attribute since we've already got an iframe on the page.

joelochlann: the following worked when hacking around in the console!

```javascript
document.querySelector('.js-email-sub__iframe').src = '/email/form/plaintone/' + emailListID
```

static/src/javascripts/projects/common/modules/experiments/tests/editorial-email-variants.js
```diff
@@ -0,0 +1,66 @@
...
+        this.description = 'Using the wonderful frontend AB testing framework to AB test emails, since the AB function in ExactTarget re-randomises all recipients on each send, and we need users to receive their variant for several weeks. This test will ensure users are added to the corresponding email list (listId) in ExactTarget';
+        this.audience = 1;
+        this.audienceOffset = 0;
+        this.successMeasure = 'We can trial two different email formats to fairly compare their CTO rates';
+        this.audienceCriteria = 'All users who visit the email sign up page';
+        this.dataLinkNames = '';
+        this.idealOutcome = 'Similar quantity of users in each list in ExactTarget';
+
+        this.canRun = function () {
+            return (config.page.webTitle.toLowerCase() === 'sign up for weekend reading');
```
katebee: Testing this on the Weekend Reading page... will be switching over to target 'the flyer' when it is published.

joelochlann (approved):
LGTM 💎

PR #15241, opened 2016-12-06
Initialise youtube atom videos in new liveblog blocks

GHaberis:
## What does this change?

- Moves youtube atom logic out of ‘bootstraps/enhanced/youtube' and put into it’s own module 'common/modules/video/youtube'
- Expose checkElemsForVideos function in new module 'common/modules/video/youtube' and call from 'common/modules/ui/autoupdate' when new liveblog block injected
- Also call checkElemsForVideos from ‘bootstraps/enhanced/youtube' when initialised.

Initialises youtube atoms in new liveblog posts

No

## Request for comment

@gidsg


static/src/javascripts/projects/common/modules/video/youtube-player.js
```diff
@@ -17,7 +17,7 @@ define([
 
     function loadYoutubeJs() {
         fastdom.write(function () {
-            if (!document.querySelector(scriptId)) {
+            if (!document.getElementById(scriptId)) {
```
GHaberis: This is a bug fix, scriptId is a string but doesn't include the '#' so this condition was always true with querySelector.

SiAdcock:
Cool stuff @GHaberis.

Do you think the name and location of the new `youtube` module expresses its intention clearly enough? Would it be clearer to call it `youtube-atom`, or perhaps keep it in the `common/modules/atoms/youtube`?

GHaberis:
@SiAdcock yeah i see your point, I would've put it in common/modules/atoms/youtube but the atom specific module's youtube-tracking and youtube-player existed already common/modules/video. Maybe all these youtube-atom specific modules should sit in common/modules/atoms?

SiAdcock:
I think `-tracking` and `-player` are also used by `hosted/youtube`...

GHaberis:
Those 2 modules are imported by hosted but they're still only used for handling a youtube-atoms (it's just hosted have their own implementation for handling these atoms). I guess there are 3 parts to the youtube-atom functionality:

- common/modules/video/youtube OR commercial/modules/hosted/youtube
- common/modules/video/youtube-player
- common/modules/video/youtube-tracking

hosted have swapped out the first part common/modules/video/youtube for their own module commercial/modules/hosted/youtube.

I definitely think I should move common/modules/video/youtube to common/modules/atoms/youtube. It just seem like if we have a directory for atoms then modules which are only used in relation to atoms should sit there? eg. - common/modules/video/youtube-player and common/modules/video/youtube-tracking.

SiAdcock:
Yeah, I tend to agree. Let's stick them all in the `atoms` directory.

GHaberis:
Ok will do, I'll make sure I update all the places which use those modules to point to the correct location!

GHaberis:
@SiAdcock OK I've moved those files and updated the modules that use them!

SiAdcock:
Love it dude 👍

PR #15244, opened 2016-12-06
Resolve stickyMpu.whenRendered when receiving a half-page

regiskuckaertz:
One learns something new every day. DMPUs do not trigger the sticky MPU module, and the `whenRendered` promise never resolves. The "offset to the right" test is waiting on that promise, so it is currently stuck when the right hand slot receives a DMPU.


JonNorman (approved):
LGTM

PR #15246, opened 2016-12-07
AWS credentials: always throw exception when credentials are not configured

TBonnin:
## What does this change?
AWS credentials: always throws exception when credentials are not configured
This is legacy code that would not throw an exception when in dev mode. However an exception is already thrown earlier in the process of loading configuration, so no need to have different logic at the credentials level

The goal of this PR and the next few ones would be to make the Configuration object totally stateless (ie: no dependency on Play conf or environment)

No

## Request for comment
@DiegoVazquezNanini @jfsoul @guardian/dotcom-platform

### DIFF
--- a/common/app/common/configuration.scala
+++ b/common/app/common/configuration.scala
@@ -527,11 +527,7 @@ class GuardianConfiguration extends Logging {
       } catch {
         case ex: AmazonClientException =>
           log.error("amazon client cross account exception", ex)
-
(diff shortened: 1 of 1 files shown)


jfsoul:
Will this mean we always have to have CMS fronts credentials in order to run frontend locally?

DiegoVazquezNanini (approved):
:+1: LGTM

TBonnin:
@jfsoul No CMS fronts credentials are only required when use

PR #15251, opened 2016-12-07
We're going to recycle this test in a few days and will redo the switch then

JustinPinner:
## What does this change?
Extends an AB test master switch. The test this controls ends today but we know we're going to recycle it within the next few days. Rather than take all the code out only to put it back in again very soon, we can just extend this switch and do the whole lot in one pass later.

Don't want to break the build. If builds don't break, success!

No.

## Screenshots
N/A

## Request for comment
@Ap0c @svillafe


Ap0c (approved):
LGTM 👍

PR #15257, opened 2016-12-07
Extend Brexit test

jranks123:
## What does this change?

Extend Brexit test until next Friday


katebee (changes requested):
😿 17th is not a weekday, so test will fail

More importantly: 

>The expiry in JS requires the day to have started [...] where as the Scala switches require the day to have passed @philwills 

It this is still the case, it would be worth giving one extra day (over `ABTestSwitches.scala` expiry) in `contributions-epic-brexit-supreme.js`.

katebee (approved):

PR #15258, opened 2016-12-07
Fix unclickable logo in perpetuity

kelvin-chappell:
I have followed @regiskuckaertz's instructions precisely.  And it seems to work!


regiskuckaertz (approved):
😋

PR #15260, opened 2016-12-07
Use window.matchMedia for breakpoint detection

regiskuckaertz:
This greatly simplifies the `getBreakpoint` and `isBreakpoint` calls, making them free. The breakpoint selection is now completely asynchronous and uses browser APIs.

cc @gtrufitt @SiAdcock @sndrs


static/src/javascripts/projects/common/utils/detect.js
```diff
@@ -71,6 +67,72 @@ define([
...
+            initMediaQueryListeners(win);
+        } else {
+            updateBreakpoints.call(win);
+            win.addEventListener('resize', debounce(updateBreakpoints, 200));
+        }
+    }
+
+    function initMediaQueryListeners(win) {
+        breakpoints
+        .forEach(function (bp, index, bps) {
```
sndrs: indent?

regiskuckaertz: Good spot! Fixed.

static/src/javascripts/projects/common/utils/detect.js
```diff
@@ -71,6 +67,72 @@ define([
...
+        while (breakpointIndex >= 0 && breakpoint.isTweakpoint) {
+            breakpointIndex -= 1;
+            breakpoint = breakpoints[breakpointIndex];
+        }
+        return breakpoint.name;
+    }
+
+    function updateBreakpoints() {
+        // The implementation for browsers that don't support window.matchMedia is simpler,
+        // but relies on (1) the resize event, (2) layout and (2) hidden generated content
```
SiAdcock: Maybe:

```
(1) ... (2) ... (3)
```

static/src/javascripts/projects/common/utils/detect.js
```diff
@@ -71,6 +67,72 @@ define([
...
+    var currentTweakpoint;
+
+    init(window);
+
+    function init(win) {
+        if ('matchMedia' in win) {
+            initMediaQueryListeners(win);
+        } else {
+            updateBreakpoints.call(win);
+            win.addEventListener('resize', debounce(updateBreakpoints, 200));
```
sndrs: we have a few of these debounced now, should we turn this into an app-wide throttled resize, like with scroll?

static/src/javascripts/projects/common/utils/detect.js
```diff
@@ -71,6 +67,72 @@ define([
...
+            updateBreakpoint(this);
+        }
+    }
+
+    function updateBreakpoint(breakpoint) {
+        if (breakpoint.isTweakpoint) {
+            currentTweakpoint = breakpoint.name;
+            currentBreakpoint = findBreakpoint(currentTweakpoint);
+        } else {
+            currentBreakpoint =
```
sndrs: this is a bit unclear i think – took me a second to work out what was happening. how do you feel about putting them on one line?

sndrs (changes requested):
lgtm other than comments (esp debouncing `resize` event)

great stuff! 🏅🌹

static/src/javascripts/projects/common/utils/detect.js
regiskuckaertz: Yep, done!

regiskuckaertz:
@sndrs ping 😅

static/src/javascripts/projects/common/utils/detect.js
sndrs: yeah it wasn't the animationframe so much as a single listener throttling once, rather than multiple listeners all throttling. i went with animation frame in the scroll listener to try and always prioritise scrolling rather than listening

> The developer needs to apply judgment as to what to do in a listener vs in an animation frame, but it should remain an informed choice

i'm not sure how i feel about this. creating that throttled event is a way of saying 'we prioritise perf over anything - you cannot have a better reason to do something, so this is all you need'. if there is a case where we *need* to block rendering the original scroll event is still there. but isn't it a fairly common pattern to abstract best practise into easily consumable apis? especially when the code has a lot of contributors who may only touch it occasionally with specific features or goals in mind. the intention was to create a frictionless way of implementing best-practise derived from a holistic, rather than case-specific, viewpoint

regiskuckaertz: > but isn't it a fairly common pattern to abstract best practise into easily consumable apis?

Yes, of course. It's just that, in this particular case, this branch is only for IE9 and Opera Mini, none of which support `matchMedia` or `requestAnimationFrame`.

But I see there already is a global event for `resize` [here](https://github.com/guardian/frontend/blob/master/static/src/javascripts/bootstraps/enhanced/common.js#L220) anyway; indeed in that case it makes sense to move that one along with `throttledScroll`. Will do, Sir.

On the _education_ value though, I kind of disagree. It's better to teach a man how to fish 😄  In that sense, I'd rather people who don't touch this very often were taught the correct _patterns_ rather than a made-up solution. In this, I think, the asynchronous buffers provide an ideal solution, both from an educational and a perf POV (unbeatable, esp. since `EventEmitter` is synchronous). I would _love_ to write a proof-of-concept.

regiskuckaertz:
@sndrs done, what do you think?

regiskuckaertz:
@sndrs 

giphy 1

gtrufitt:
@regiskuckaertz 

He is 
image for christmas

regiskuckaertz:
Ooooh cool 🏄

static/src/javascripts/bootstraps/standard/main.js
```diff
@@ -172,13 +176,23 @@ define([
...
                 window.requestIdleCallback(onScroll);
             } :
-            onScroll
+            onScroll,
+            { passive: true }
         );
 
+        // Adds a global window:resize event to mediator, which debounces events
+        // until the user has stopped resizing the window for a reasonable amount of time.
+        // CAUTION: the event listener is *passive*, which means calls to event.preventDefault
```
sndrs: might be worth adding this warning above too? (lovely to see this in there btw)

sndrs (approved):

sndrs:
a couple of other suggestions on your changes, but really nice work :) 👍

static/src/javascripts/bootstraps/standard/main.js
```diff
@@ -172,13 +176,23 @@ define([
...
+        addEventListener(window, 'scroll', userPrefs.get('use-idle-callback') && 'requestIdleCallback' in window ?
             function () {
                 window.requestIdleCallback(onScroll);
             } :
-            onScroll
+            onScroll,
+            { passive: true }
         );
 
+        // Adds a global window:resize event to mediator, which debounces events
```
regiskuckaertz: Yep, doing that!

PR #15261, opened 2016-12-08
The Australia test started a bit late and needs more time to complete

JustinPinner:
## What does this change?
The Australia membership engagement banner copy test (#8) started several days later than planned so we need to extend its lifespan to achieve statistical significance. And we don't want it to end during the holiday period, so we've pushed the end date out to January.

If we don't extend it the test will probably not have a useful outcome.

No, web only.

## Screenshots
N/A

## Request for comment
I'll try using the new review request thingy.


rupertbates (approved):
👍

PR #15262, opened 2016-12-08
Remove grunt-assetmonitor, again

sndrs:
## What does this change?

trying #15108 again...removes `grunt-assetmonitor` basically, brings it 'in-house' 

last grunt plugin to go

shouldn't do 🙏


tools/asset-monitor/cloudwatch.js
```diff
@@ -0,0 +1,102 @@
...
+            Dimensions : [
+                {
+                    Name : 'Metric',
+                    Value : 'Average Selectors'
+                }
+            ]
+        });
+    }
+
+    cloudwatch.putMetricData(params, (err, data) => {
```
SiAdcock: Could possibly `reject` here if `cloudwatch` is undefined:

```
if (!cloudwatch) reject(new Error('Cloudwatch has not been configured'));
```

SiAdcock (approved):

PR #15265, opened 2016-12-08
NewsAlerts method will access notifications directly from S3

rich-nguyen:
We are seeing issues in the Preview server around the notifications system, which throws an exception when it asks the Actor to grab the notifications. I'm temporarily replacing this with a simple S3 access. @TBonnin we can have a look soon.

There is definitely some internal issue here causing the actor system problems, will continue to investigate.


SiAdcock (approved):

rich-nguyen:
Here is preview, seemingly at the start of the issue we've been seeing. Yes, I'm aware that was my previous PR :) screen shot 2016-12-08 at 15 11 42

TBonnin:
What's strange is the error seems to have stopped today at 11am https://kibana.gu-web.net/goto/34a797da26cee7f55bc316ebc127c127

rich-nguyen:
it's really weird. And the errors only dropped when this PR was deployed. https://kibana.gu-web.net/app/kibana?#/discover?_g=(refreshInterval:(display:'1%20minute',pause:!f,section:2,value:60000),time:(from:now-4h,mode:quick,to:now))&_a=(columns:!(_source),filters:!(),index:'logstash-*',interval:auto,query:(query_string:(analyze_wildcard:!t,query:'app:preview%20AND%20resp_status:500')),sort:!('@timestamp',desc),uiState:(),vis:(aggs:!((params:(field:resp_status,orderBy:'2',size:20),schema:segment,type:terms),(id:'2',schema:metric,type:count)),type:histogram))&indexPattern=logstash-*&type=histogram

rich-nguyen:
[I borked Preview somehow](https://trello.com/c/FLYUIEto/243-i-borked-preview-somehow)

PR #15268, opened 2016-12-08
Hosted about popup unit tests

Calanthe:
## What does this change?
Some unit tests related to the about popup module.

I also had to change a few fixture id's because the same value was causing errors.


sndrs (approved):
lgtm!

PR #15270, opened 2016-12-08
Removal of omniture dashboard

scottp719:
## What does this change?

This removes all the frontend code for the omniture-dashboard page which displayed top browser and OS stats on the frontend admin page. 

This is because we said goodbye to omniture already and unless we need to replicate with GA (currently thinking not as the UI is considerably easier to use) this should be removed as now doesn't work. I have already deleted the Lambda and S3 bucket so data is no longer being served.

Goodnight sweet prince....

Less code, no longer hitting the omniture API.

Nopers


SiAdcock (approved):
LGTM

sndrs (approved):
␡

PR #15275, opened 2016-12-09
Apply test from desktop screens upward

regiskuckaertz:
The test will now include not only people on the wide breakpoint, but all those seeing a right-hand column.


kelvin-chappell (approved):
👍

PR #15276, opened 2016-12-09
Display correct quiz bucket message

SiAdcock:
## What does this change?

The quiz always displays the bucket message for the top score, regardless of how many questions were answered correctly. This change shows the bucket message relevant to the user's score

picture 144

## Request for comment

@sndrs


sndrs (approved):
✅

PR #15278, opened 2016-12-09
Update installation docs

SiAdcock:
## What does this change?

- Don't recommend [installing nvm using homebrew](https://github.com/creationix/nvm/#installation)
- Moved nvm installation instructions to section that explains what nvm is
- removed redundant `npm install` command

Makes the frontend installation process slightly smoother


sndrs (approved):
👌

PR #15279, opened 2016-12-09
Add outstream video size

regiskuckaertz:
Outstream videos (whether Teads Inread or AppNexus Oustream formats) will now be uniquely identified by a new size, `620x1`. `620` is the width of the text column, and `1` is so that we mitigate the impact of resizing the slot—since here, it is hidden anyway.

video


regiskuckaertz:
[Create new sizes for closing slots and delivering Teads et al](https://trello.com/c/5uGcepOE/26-create-new-sizes-for-closing-slots-and-delivering-teads-et-al)

static/src/javascripts/projects/common/modules/commercial/dfp/create-slot.js
```diff
@@ -9,7 +9,8 @@ define([
 ) {
     var inlineDefinition = {
         sizeMappings: {
-            mobile: compile(adSizes.outOfPage, adSizes.empty, adSizes.mpu, adSizes.fluid)
+            mobile: compile(adSizes.outOfPage, adSizes.empty, adSizes.mpu, adSizes.fluid),
+            desktop: compile(adSizes.outOfPage, adSizes.empty, adSizes.mpu, adSizes.video, adSizes.fluid)
```
JonNorman: Where is this used? Can't see it anywhere else in the diff.

JonNorman (approved):
LGTM, one quick question: in your GIF/video, there seems to be a brief expansion of an inline MPU, which then closes and the Outstream ad loads - what is going on there?

regiskuckaertz:
@JonNorman The inline ad slot initially has space to allow for an MPU, which is most of the time. However, the Teads creative is doing some ~~crazy shit~~ lovely stuff like breaking out of the iframe, which is why I used 620x1 instead of, say, 620x400. It is unlikely people will see any of this as we load the ad before it enters the viewport. That being said, I think we may rethink the layout of ad vs ad slots to get rid of that effect entirely in the future.

PR #15280, opened 2016-12-09
Inline MPUs: offset to the right on desktop

regiskuckaertz:
I forgot to add the correct layout rule for inline MPUs on desktop.

Before:
picture 7

After:
picture 8


kelvin-chappell (approved):
👍

PR #15283, opened 2016-12-09
Add missing promise polyfills

sndrs:
## What does this change?

adds missing promise polyfills

should make the code work where `Promise` is not supported 

no


regiskuckaertz (approved):
👍  I just merged changes to `slice-adverts` so you might have a conflict, sorry about that.

SiAdcock (approved):
I like it, let's move away from relying on globals!

PR #15285, opened 2016-12-12
Remove expired hosted pages from AMP validation tests

lps88:
cc @guardian/dotcom-platform


jfsoul (approved):
Thanks 👍

jfsoul:
Out of curiosity, are the other hosted articles in the tests due to expire or are they permanent?

lps88:
@jfsoul They will expire in a couple of months. We have a few pages that are permanently up as they're hardcoded in frontend, but we're planning on removing them soon. I'm going to have a chat with the team on how to manage this!

PR #15290, opened 2016-12-12
Kill the Weekend Reading static page

katebee:
## What does this change?

We now have [a fully functional interactive](https://www.theguardian.com/info/ng-interactive/2016/dec/05/sign-up-for-weekend-reading) for Weekend Reading. We no longer need the static page

🔪 💀 Kill the [static page](https://www.theguardian.com/signup/weekendreading)

🔪 💀 Deleting code from frontend

make https://www.theguardian.com/signup/weekendreading a 404 again so Central Prod can set up a redirect to the interactive


SiAdcock:
picture 145

❤️ 👍

PR #15291, opened 2016-12-12
Lint tools

sndrs:
## What does this change?

lints `tools` according to the forthcoming **Hard Lint™** rules

prep for the **Hard Lint™**

cc @SiAdcock @gtrufitt

### DIFF
--- a/tools/messages.js
+++ b/tools/messages.js
@@ -2,27 +2,23 @@ const fs = require('fs');
 
 const takeWhile = require('lodash.takewhile');
 
-function notify(message, options, type) {
-    options = options || {};
-    type = type || 'log';
-
+function notify(message, userOptions = {}, type = 'log') {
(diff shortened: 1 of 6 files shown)


tools/messages.js
```diff
@@ -67,25 +63,31 @@ switch (process.argv[2]) {
...
-      }, 'info');
-      break;
+    }
+    case 'install-steps': {
+        notify('Please run the following to complete your installation:', {
+            heading: 'Additional steps',
+        }, 'info');
+        break;
+    }
+    default: {
```
gtrufitt: Minor but could default to describe

gtrufitt (approved):
Other than minor F2F comments, LGTM!

tools/messages.js
sndrs: as discussed, not only make tasks in here

PR #15294, opened 2016-12-12
Remove temporary route for archive switchover

jfsoul:
## What does this change?
Removes a redundant route which should no longer be hit by the router now after https://github.com/guardian/platform/pull/1145 was merged.

Less redundant code.

No

## Screenshots
N/A


jfsoul:
cc: @guardian/dotcom-platform

gtrufitt (approved):

PR #15295, opened 2016-12-12
lint amp validation

sndrs:
## What does this change?

lints `amp validation` according to the forthcoming **Hard Lint™** rules

prep for the **Hard Lint™**

cc @SiAdcock @gtrufitt @jfsoul


jfsoul (approved):
Lgtm 👍 

Not sure how I feel about dangling commas, but I can get on board if that's the consensus 🚢

PR #15301, opened 2016-12-12
Fix archive healthcheck

jfsoul:
## What does this change?
Fixes healthcheck after I hastily removed an old route.

Fixes deploy of archive!

## Screenshots

@guardian/dotcom-platform


DiegoVazquezNanini (approved):
:+1:

PR #15304, opened 2016-12-12
[Do Not Merge] Remove hardcoded legacy hosted pages 🔥

lps88:
All the original campaigns are finished, so we can clean up the code base significantly!

We have a press release using Renault as an example, so we need to keep it live until the end of this week.

__Edit__: We need to keep them until the end of January, as we're repeating some UX tests on Singapore F1 campaign


kelvin-chappell (approved):
Hurray! 🔥

TBonnin (changes requested):
Should the related images in https://github.com/guardian/frontend/tree/master/static/public/images/commercial be removed as well?

NataliaLKB:
@lps88 What is happening with this PR?

lps88:
@NataliaLKB We thought we were okay to delete these as the campaigns are finished, but it turned out there was a press release and then some UX tests that needed them to stay live.

Matt Ho is going to let us know when it's okay to merge this. Last I heard was it should be around the end of this month.

NataliaLKB:
@lps88 Thanks for the update! Is it worth closing this, and reopening it when you can work on it again?

lps88:
Closing until we're able to make progress...

PR #15309, opened 2016-12-13
Extend test for another week

regiskuckaertz:
## What does this change?

We made a few changes last friday and need the test to run unchanged for a full 7-day window.


katebee:
Does the date in `its-raining-inline-ads.js` also need extending?

regiskuckaertz:
ha! it's my first AB test so I'm not yet used to it. Thank you!

katebee (approved):
👍 

:shipit:

PR #15310, opened 2016-12-13
remove expired page from test

gidsg:
## What does this change?
Remove test for a commercial page which has been removed

Allow tests to pass locally

No


lps88:
CC @kelvin-chappell

kelvin-chappell (approved):
Quicker than me!
👍  but also in #15312

gidsg:
OK I will close in favour of #15312

PR #15311, opened 2016-12-13
Refactor: wrap injected Environment into ApplicationContext instance

TBonnin:
## What does this change?
In preparation of more DI into controllers and html/js templates, we are
now injecting an ApplicationContext instance which now contains a
reference to the Environment. In future PRs  we will add the
ApplicationIdentity into this ApplicationContext which would
allow us to make it available to controllers and templates without
having to modify every files in the project

=> Play 2.5

No


jfsoul (approved):
> The table of contents is too big for display.

PR #15312, opened 2016-12-13
Remove redundant amp tests

kelvin-chappell:
All the removed tests have expired or are going to expire eventually.  
They're also already covered in `CommercialAmpValidityTest` and `ArticleAmpValidityTest`.

/cc @guardian/dotcom-platform @guardian/commercial-dev

### DIFF
--- a/commercial/test/test/CommercialAmpValidityTest.scala
+++ b/commercial/test/test/CommercialAmpValidityTest.scala
@@ -22,17 +22,16 @@ import play.api.test._
   }
 
   val hostedArticleId = "advertiser-content/audi-history-of-audi/audi-and-innovation"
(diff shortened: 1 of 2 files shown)


gidsg:
👍 🚢

jfsoul:
There is an implicit assumption at the moment with all of these tests that the content is permanent.  Even for the scala tests that use test fixtures - what do we do when the fixtures need updating due to a change in the model?  I don't have a good answer for this 😞

kelvin-chappell:
@jfsoul In most cases the fixtures will be permanent.  In the few cases where they aren't we can easily find new ones when the model is updated.  It doesn't happen very often.

PR #15315, opened 2016-12-13
Remove bold variant from, and rename, the pre-end of year contributions test

jranks123:
## What does this change?

The bolder variant performed significantly worse than the other 2 variants, so we are removing it from the test. The rename is to make sure that the segmentation of the audience still works correctly.


katebee (approved):
👍

PR #15317, opened 2016-12-13
lint dev

sndrs:
## What does this change?

lints `dev` and top-level stuff according to the forthcoming **Hard Lint™** rules

prep for the **Hard Lint™**


NataliaLKB:
👍

PR #15318, opened 2016-12-13
Remove Play.current in assets css object

TBonnin:
## What does this change?
Remove Play.current (deprecated in Play 2.5) in assets css object

-> Play 2.5


DiegoVazquezNanini (approved):
:+1: Nice job

PR #15319, opened 2016-12-13
Use IntersectionObserver to lazy load ads

regiskuckaertz:
This weekend I got frustrated scrolling through the site on my mobile, seeing ads being so slow to load. I attributed that to the fact that the `scroll` event is only triggered when I release my finger from the screen, and this can delay the DFP call by a lot.

Well, that is exactly what `IntersectionObserver` was meant for. Now that it is being developed across [all](https://bugzilla.mozilla.org/show_bug.cgi?id=1243846) [major](https://webkit.org/status/#specification-intersection-observer) [browsers](https://developer.microsoft.com/en-us/microsoft-edge/platform/status/intersectionobserver/) and there already is widespread Chrome/Opera support, we can gradually branch out all the features now relying on the `scroll` event to compute the position of elements. Also, we could do with one less global event listener.

Tested on my mobile: the difference is dramatic.


regiskuckaertz:
[Increased perceived site performance with a better lazy load method](https://trello.com/c/FHHK78SB/252-increased-perceived-site-performance-with-a-better-lazy-load-method)

paperboyo:
💎!

sndrs:
was there a reason you didn't opt to [polyfill](https://github.com/WICG/IntersectionObserver)?

static/src/javascripts/projects/commercial/modules/dfp/lazy-load.js
```diff
@@ -21,7 +21,11 @@ define([
     /* loadQueued: boolean. Set to true when a lazyload task is scheduled */
     var loadQueued = false;
 
-    var lazyLoad = throttle(function () {
+    var lazyLoad = dfpEnv.lazyLoadObserve ? onIntersect : throttle(onScroll, nbOfFrames * durationOfFrame);
```
regiskuckaertz: yes, probably. I wanted to do something along the lines of what the polyfill does with `THROTTLE_TIMEOUT` but honestly I don't know if it makes any difference. Will switch to `throttledScroll`.

regiskuckaertz:
yes, the polyfill adds a lot of event listeners, I'm not sure it's a good idea to use it at all.

joelochlann:
This is great! It reminds me of my long-dormant intention to switch the [component attention tracking in Ophan](https://github.com/guardian/ophan/pull/1515) over to use IntersectionObserver. Hopefully I'll get some time to do that in the next few weeks

regiskuckaertz:
Thanks 😃  That would be super useful. We are currently using a third-party lib to track viewability of our ads, and it's adding _loads_ of listeners: I too would like to write an alternative using `IntersectionObserver`. If I'm not mistaken, the latter would also work with an ad running in a cross-origin iframe, which none of the 3rd-party libs that I know of currently support.

sndrs (approved):
🌟

sndrs:
>  Hopefully I'll get some time to do that in the next few weeks

do it @joelochlann!

static/src/javascripts/projects/commercial/modules/dfp/enable-lazy-load.js
```diff
@@ -1,15 +1,27 @@
...
+                observer = new IntersectionObserver(lazyLoad, { rootMargin: '200px 0%' });
+                dfpEnv.adverts.forEach(function (advert) {
+                    observer.observe(advert.node);
+                });
+            } else {
+                mediator.on('window:throttledScroll', lazyLoad);
+                lazyLoad();
+            }
+        } else if (dfpEnv.lazyLoadObserve && advert) {
+            observer.observe(advert.node);
```
rich-nguyen: Why not add a separate function for observing an advert?

static/src/javascripts/projects/commercial/modules/dfp/lazy-load.js
```diff
@@ -1,69 +1,62 @@
...
-                loadAdvert(advert);
-            });
-        });
-    }, nbOfFrames * durationOfFrame);
+        lazyLoad.forEach(displayAd);
+    }
 
     function disableLazyLoad() {
         dfpEnv.lazyLoadEnabled = false;
         window.removeEventListener('scroll', lazyLoad);
```
rich-nguyen: destroy `IntersectionObserver` here?

rich-nguyen: (conditionally)

static/src/javascripts/projects/commercial/modules/dfp/lazy-load.js
```diff
@@ -1,69 +1,62 @@
...
-                var isNotHidden = rect.top + rect.left + rect.right + rect.bottom !== 0;
-                var isNotTooFarFromTop = (1 - depthOfScreen) * viewportHeight < rect.bottom;
-                var isNotTooFarFromBottom = rect.top < viewportHeight * depthOfScreen;
-                // load the ad only if it's setting within an acceptable range
-                return isNotHidden && isNotTooFarFromTop && isNotTooFarFromBottom;
-            })
-            .map(function (advert) {
-                return advert.id;
-            });
+        var lazyLoad = dfpEnv.advertsToLoad
```
rich-nguyen: this var shadows the alias to `onIntersect`, can we rename it for legibility?

static/src/javascripts/projects/commercial/modules/dfp/lazy-load.js
```diff
@@ -1,69 +1,62 @@
...
+        .filter(function (advert) {
+            var rect = advert.node.getBoundingClientRect();
+            var isNotHidden = rect.top + rect.left + rect.right + rect.bottom !== 0;
+            var isNotTooFarFromTop = (1 - depthOfScreen) * viewportHeight < rect.bottom;
```
rich-nguyen: not related to this pr, but why is this `(1 - depthOfScreen)` here?

rich-nguyen:
👍 sweet, ill keep an eye on the impressions 💰

static/src/javascripts/projects/commercial/modules/dfp/enable-lazy-load.js
regiskuckaertz: How so? If there is an incentive to measure that, then I can add a separate function, but otherwise I found this solution to have a minimal footprint. Or do you have a better solution in mind?

rich-nguyen: just for clarity really. I didn't realise the function does the setup, plus the per-advert observation, and supported individual advert observation, based on the param. I think its ok either way

static/src/javascripts/projects/commercial/modules/dfp/lazy-load.js
regiskuckaertz: Coordinates are relative to the top coordinate of the viewport. We want slots 1/2 a screen above or below the viewport to be loaded.
- Below is easy: `offsetBelow = viewportHeight * 1.5` (i.e. 150% from the top of the viewport)
- Above, the same formula, `offsetAbove = viewportHeight * -1.5` would go up "one viewport too far", so `offsetAbove = viewportHeight * (1 - 1.5)` fixes that (i.e. -50% from the top of the viewport)

regiskuckaertz:
@rich-nguyen thanks a lot for your suggestions.

giphy

PR #15320, opened 2016-12-13
Never use links as kickers in email

joelochlann:
The [kicker template I was previously using for kickers in email](https://github.com/guardian/frontend/blob/master/common/app/views/fragments/items/elements/facia_cards/kicker.scala.html) will render tag & section kickers as links. Since we're already inside an `<a></a>` element at this point, this spectacularly breaks the HTML rendering.

So now I just explicitly output a `<span></span>`.  This will allow Editorial to use tag & section kickers for convenience if they want to, rendering them as text inside the whole-card link to the article just as it would for a custom kicker.

### broken (see bottom item):
picture 353

### fixed:

picture 354

@lmath @guardian/dotcom-platform

### DIFF
--- a/facia/app/views/fragments/emailFrontBody.scala.html
+++ b/facia/app/views/fragments/emailFrontBody.scala.html
@@ -7,15 +7,16 @@
 
 @import views.support.EmailHelpers._
 @import views.support.TrailCssClasses.toneClass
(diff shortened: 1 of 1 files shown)


NataliaLKB:
👍

PR #15321, opened 2016-12-13
Hosted article pages sticky header

lps88:
## What does this change?
Make the header 'sticky' so it's always visible as you scroll through the article.

Our UX lab tests showed that the 'advertiser content' banner is the strongest indicator that this is commercial content, so we want it to be more visible.

A previous survey relating to Hosted articles showed that ~70% of users understand what Hosted pages are. We will do another survey after this change to see if this has improved.

Also implemented for AMP Hosted articles.

## Screenshots
image

CC @guardian/labs-beta


kelvin-chappell (approved):
👍

PR #15322, opened 2016-12-13
Style external links

joelochlann:
Style external links differently from Guardian links. Specifically:

- no bottom padding
- sans serif font
- border colour based on page tone (e.g. pink for flyer)

Might need some refinement by @superfrank 

## Old (card)
picture 358

## New (card)
picture 355

## Old (connected)
picture 359

## New (connected)
picture 356

### DIFF
--- a/common/app/views/fragments/email/stylesheets/connectedStyleOverride.scala.html
+++ b/common/app/views/fragments/email/stylesheets/connectedStyleOverride.scala.html
@@ -16,7 +16,7 @@
         background: #fee9ed;
     }
 
-    .tone-news {
+    .tone-news .facia-card {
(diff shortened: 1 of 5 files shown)


SiAdcock (approved):
Looks good, man. I assume there's no way around those `!important` declarations, sad times. 

👍

joelochlann:
Yeah, the CSS framework for email we use has `!important` all over the place so to override it we have no choice. I think they made that decision because of style inlining - the only way to override an inline style from a separate style declaration is with `!important`, and there are some styles which can't be inlined (like `:hover`)

joelochlann:
@superfrank is actually refining the styling a little bit for these so I'm not merging just yet

PR #15327, opened 2016-12-13
Stubbed capi at runtime

MatthewJWalls:
## What does this change?

This introduces a quick mock CAPI service which is injected in by the **article** AppLoader when the install_vars stage is set to DEVINFRA - causing all capi queries to be retrieved from the local data directory instead of going out to the real CAPI service.

This duplicates some functionality of the original http recorder thing - I'm leaving refactoring out that duplication for a different PR.

@TBonnin @jfsoul couple of questions I have about this change:

1. Is it valid for me to have this logic conditional on DEVINFRA? I'm not sure if we use that stage mode for anything in particular that I might be breaking - I'm only using DEVINFRA because the original httpRecorder in the test directory has some conditionals around that too.

2. I noticed in DEVINFRA.properties file that the image assets are coming from i.gucode.co.uk - which doesn't resolve so the images are always broken. Can I set this to something else or is there a reason it needs to be i.gucode.co.uk?

* Easier for non-guardian contributors to be able to run the app outside of our infra.
* Makes the frontend app easier to run on build/ci systems since you don't have to deal with configuration.
* Provides a way to ensure data doesn't change between regression test runs such as on the screenshot functionality we're trying to implement

No

## Screenshots

N/A

### DIFF
--- a/article/app/AppLoader.scala
+++ b/article/app/AppLoader.scala
@@ -8,7 +8,7 @@ import targeting.TargetingLifecycle
 import common.dfp.DfpAgentLifecycle
 import conf.switches.SwitchboardLifecycle
 import conf.{CachedHealthCheckLifeCycle, CommonFilters}
-import contentapi.{CapiHttpClient, ContentApiClient, HttpClient}
+import contentapi.{CapiHttpClient, ContentApiClient, HttpClient, MockHttpClient}
 import controllers.{ArticleControllers, HealthCheck}
 import dev.{DevAssetsController, DevParametersHttpRequestHandler}
 import model.ApplicationIdentity
@@ -20,6 +20,7 @@ import play.api.routing.Router
 import play.api._
 import services.{NewspaperBooksAndSectionsAutoRefresh, OphanApi}
 import router.Routes
+import conf.Configuration
 
 class AppLoader extends FrontendApplicationLoader {
   override def buildComponents(context: Context): FrontendComponents = new BuiltInComponentsFromContext(context) with AppComponents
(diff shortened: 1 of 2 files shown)


article/app/AppLoader.scala
```diff
@@ -20,14 +20,17 @@ import play.api.routing.Router
...
 
 class AppLoader extends FrontendApplicationLoader {
   override def buildComponents(context: Context): FrontendComponents = new BuiltInComponentsFromContext(context) with AppComponents
 }
 
 trait AppComponents extends FrontendComponents with ArticleControllers {
 
-  lazy val capiHttpClient: HttpClient = wire[CapiHttpClient]
+  lazy val _isDevInfra = Configuration.environment.stage.equalsIgnoreCase("DEVINFRA")
+  lazy val capiHttpClient: HttpClient = if (_isDevInfra) wire[MockHttpClient] else wire[CapiHttpClient]
```
TBonnin: I don't really like conditional in the production code. Can you explain why and how you want to use article with mock capi?

TBonnin (changes requested):

article/app/AppLoader.scala
MatthewJWalls: I want to run the actual app (as opposed to tests), without remote services, so it can be location agnostic.

we're going to run the app on an instance and point Gareth's screenshots and mike's analytics stuff to it. This is a lot easier if we don't have to worry about configuring for location of remote services (ie just reuse devinfra)

As a side effect this also makes it possible for people external to the Guardian to contribute  changes more easily, a point which was on one of the big bullet point list of nice things to do.

jfsoul:
In answer to your questions:

1. `DEVINFRA` is used for running on teamcity.
2. This is probably just really old - it points directly to an old ELB (for the now defunct in-house image resizing service).  It can be updated to `//i.guimcode.co.uk` most likely.

PR #15328, opened 2016-12-13
Use different fonts in article & front emails

joelochlann:
In #15156 we accidentally put some styles in `email/stylesheets/main.scala.html` (shared between both front & article emails) which should have gone into the front-specific CSS. We also changed the `@font-face` declarations universally, when we should've added an alternative fonts stylesheet for fronts. This caused some style problems for the article emails.

This extracts the changes which we want to apply to the front emails into the right stylesheets, restoring the common styles to how they were before #15156.

## broken
picture 366

## fixed
picture 367

@superfrank @katebee


katebee (approved):
LGTM 👍

PR #15329, opened 2016-12-13
Add transparent (for now) babel step to JS app compilation

sndrs:
## What does this change?

- reopens #15130 
- adds babel 'transpilation' step to app compilation, but as yet no actual plugins
- webpack build uses this compiled code currently, but clearly we wouldn't do this once webpack is 100%.

first step towards ES6 in frontend

no

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -8,7 +8,11 @@
         },
         "admin": {
             "presets": [
-                "latest"
+                ["env", {
+                    "targets": {
+                        "browsers": ["last 2 Chrome versions, last 1 Safari version, last 2 Firefox versions, last 2 Edge versions"]
+                    }
+                }]
             ]
         }
     }
--- a/.babelrc
+++ b/.babelrc
@@ -1,13 +1,14 @@
 {
     "env": {
-        "production": {
-            "presets": [
-                "latest"
-            ]
-        },
+        "production": {},
         "development": {
             "plugins": [
                 "transform-async-to-generator"
+            ]
+        },
+        "admin": {
+            "presets": [
+                "latest"
             ]
         }
     }
--- a/tools/__tasks__/compile/javascript/babel.js
+++ b/tools/__tasks__/compile/javascript/babel.js
@@ -1,0 +1,12 @@
+const execa = require('execa');
+
+const {src, transpiled} = require('../../config').paths;
+
+module.exports = {
+    description: 'Transpile',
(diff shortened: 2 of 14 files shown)


.babelrc
```diff
@@ -1,14 +1,15 @@
...
-        },
+        "production": {},
         "development": {
             "plugins": [
                 "transform-async-to-generator"
             ]
+        },
+        "admin": {
+            "presets": [
+                "latest"
```
gtrufitt: I know this is admin, but maybe it's worth using https://github.com/babel/babel-preset-env so we can specify browser versions (like we do with autoprefixer).

tools/__tasks__/compile/javascript/babel.js
```diff
@@ -0,0 +1,12 @@
+const execa = require('execa');
+
+const {src, transpiled} = require('../../config').paths;
+
+module.exports = {
+    description: 'Transpile',
+    task: () => execa('babel', [`${src}/javascripts`, '-d', `${transpiled}/javascripts`, '--ignore', 'bower_components/,components/,vendor/'], {
```
gtrufitt: Worth using `--out-dir` instead of `-d` to increase comprehension?

gtrufitt (approved):

gtrufitt (approved):
LGTM 👌

.babelrc
```diff
@@ -1,14 +1,19 @@
...
         "development": {
             "plugins": [
                 "transform-async-to-generator"
             ]
+        },
+        "admin": {
+            "presets": [
+                ["env", {
+                    "targets": {
+                        "browsers": ["last 2 Chrome versions, last 1 Safari version, last 2 Firefox versions, last 2 Edge versions"]
```
SiAdcock: Does this go back far enough? Does this mean that once we start using ES6 <=IE11 will go boom?

SiAdcock:
Why not use [`babel-loader`](https://github.com/babel/babel-loader) in the Webpack config, rather than loading from the `transpiled` directory? Arguably, this is what we'll be doing when we phase out `r.js`

.babelrc
sndrs: only on admin

sndrs:
> Why not use babel-loader in the Webpack config, rather than loading from the transpiled directory? Arguably, this is what we'll be doing when we phase out r.js

I didn't want to get in the way of any work you're doing there. i can update that if you like?

SiAdcock:
No, you're right @sndrs, I can do that as a separate PR 😄

sndrs:
got a better plan. closing for a second time

PR #15330, opened 2016-12-13
Handle case where media atom has no assets

gidsg:
## What does this change?
Previously if a media atom was embedded which had no assets it would throw an exception trying to read head of an empty list. Now this is a `headOption` so it will safely fall through this case.

Fix possible exception

Fix also applies to amp

## Screenshots
N/A


akash1810:
:+1:

akash1810 (approved):

PR #15331, opened 2016-12-13
Remove Play.current in assets LoadFromClasspath

TBonnin:
## What does this change?
Remove Play.current in assets LoadFromClasspath

=> Play 2.5


DiegoVazquezNanini (approved):
:+1: 
I hope it works because this is brilliant!

alexduf (commented):
Cool!

jfsoul (commented):
Nice! 👍

PR #15333, opened 2016-12-14
Enable ad-free experience for user trials

JustinPinner:
## What does this change?
Suppresses adverts for users participating in the ad-free trial, based on responses from the members data API.

We will be seeking feedback from a closed user group who will be participating in the ad-free trial. The outcome of this will inform the next phase of analysis and decision making. 

No, web only.

## Screenshots
An example video article. Signed-in, ad-free trial user on the left, unknown user on the right;
video-ad-free-trial


common/app/conf/switches/CommercialSwitches.scala
```diff
@@ -16,6 +16,16 @@ trait CommercialSwitches {
     exposeClientSide = true
   )
 
+  val AdFreeTrialSwitch = Switch(
+    SwitchGroup.Commercial,
+    "ad-free-membership-trial",
+    "Master switch for trialling ad-free membership perk",
+    owners = Seq(Owner.withGithub("JustinPinner")),
+    safeState = Off,
+    sellByDate = never,
```
katebee: How long will the trial last? (Does the switch need to be immortal?)

paulbrown1982:
Looks fine to me from a membership perspective. I think someone from the frontend team should give the thumbs up though.

paulbrown1982 (commented):
Looks fine to me from a membership perspective. I think someone from the dotcom team should give the thumbs up though.

regiskuckaertz:
Ok the code looks good, but I don't think it is complete:

- Have you tried this on an article? or a front? We have several ad slots that are injected dynamically and I don't think this change will address them.
- What about Outbrain?
- Isn't it possible to sign the `html` element with the cookie, so that the hiding is done in CSS? I haven't tested it, but it may be jarring to see slots getting closed only when `commercial` is bootstrapped, especially the top banner (cue @ScottPainterGNM). (Bonus point if this can be done on the back end) So you would have:

`<html class="cookie-adfree-on ...">`

and in CSS

`.cookie-adfree-on .ad-slot { display: none !important; }`

JustinPinner:
Thanks @regiskuckaertz, I've seen it work on fronts and articles locally, but there may well be cases that I haven't discovered yet. The network front is problematic locally as it generates other errors that prevents the commercial code from booting, so I'd like to get this at least into CODE and get a look at it there.

The backend manipulation of classes and the CSS route to block those seems like the way to go and I'll look at that and put it through as a later PR if that's ok?

regiskuckaertz:
Yes, let's see what happens on CODE. Try a long read for instance, we do insert MPUs far after DFP init has run.

regiskuckaertz (approved):

PR #15334, opened 2016-12-14
Bump targeting client

desbo:
This upgrades the version of the targeting client to support campaigns matching any tag.


dominickendrick:
👍

PR #15336, opened 2016-12-14
Add airbnb config to eslint as app-wide default

sndrs:
## What does this change?

- updates the default eslint for the whole codebase to extend airbnb config
- overrides it for the current app/tests with the current rules
  - airbnb expects es6 among others things, and that's still to come
  - once app/tests are converted, those overrides will be removed

- catching bugs
- enforcing best practice
- consistency
- `make fix` will catch and fix simple errors
  - whatever is left is (hopefully) something worth looking at
  - whatever bike-shedding style-fights we want to have are easily updated

blhspzxaqmptm


tools/amp-validation/README.md
```diff
@@ -0,0 +1,10 @@
+#amphtml-validator
+
+Utility that uses the amphtml-validator to validate AMP endpoints.
+
+Mainly for running on teamcity, the [build steps](https://teamcity.gu-web.net/admin/editBuildRunners.html?id=buildType:dotcom_AmpValidation) do the following:
+
+- `npm i` (where `cwd` is `tools/amp-validation`)
+- `npm start` (where `cwd` is `tools/amp-validation`)
+
+It has it's own npm modules.
```
SiAdcock: `its`

SiAdcock (approved):
:heart: it, dude

PR #15337, opened 2016-12-14
Stateless config part 3

TBonnin:
## What does this change?
Making GuardianConfiguration not depend on Play environment or config
Until now it was relying on Play.current which is being deprecated in Play 2.5

Reviewing commit by commit would make more sense

=> Play 2.5


jfsoul (approved):
Looks good!

PR #15339, opened 2016-12-14
Australian Epic for end of year campaign

alexduf:
## What does this change?
This adds an ask at the bottom of the article for our Australian readers.
It will be displayed to 10% of the population, and will help us determine which variant performs best
One of the variants includes a cartoon from First Dog on the Moon. I have embedded it in our `static` folder and I'm using the `config.js` to reference it. Let me know if that approach causes any issue.

We can determine what's the best way to ask for contributions and supporters in Australia, and will use the result of the test to do so.

No

## Screenshots
screenshot from 2016-12-14 14-39-48
screenshot from 2016-12-14 14-39-21
screenshot from 2016-12-14 14-38-58
screenshot from 2016-12-14 14-38-27

cc @guardian/guardian-frontend @desbo


static/src/javascripts/projects/common/modules/experiments/tests/contributions-epic-on-the-moon.js
```diff
@@ -0,0 +1,234 @@
...
+
+        function registerInsertionListener(track) {
+            mediator.on(epicInsertedEvent, track);
+        }
+
+        function registerViewListener(complete) {
+            mediator.on(epicViewedEvent, complete);
+        }
+
+        var canBeDisplayed = function() {
```
jranks123: This function is not necessary, all the logic around whether the test can run is in the canRun function on line 54.

alexduf: Good point! I'm fixing it.

alexduf:
I had another go at it, would you like to check again @jranks123 ?

regiskuckaertz:
Could @paperboyo work his magic and decrease the size of these images?

jranks123:
Looks good from my end 👍

paperboyo:
Yes, Sir! Bit lossy but should be OK.

first-dog-mb_opt2
first-dog-dt_opt3

alexduf:
Here's what I have done:

- Rebased to get the latest test flags so the build can pass
- Added logic to avoid test clash
- fix @jranks123 comments
- replaced images for lighter ones from @paperboyo 

Let me know if there's anything else I should fix

jranks123:
👍

PR #15342, opened 2016-12-15
Remove fakeOnStart workaround

TBonnin:
## What does this change?
Remove fakeOnStart workaround which is not useful anymore since we don't reference Play.current global object anymore

=> Play 2.5

### DIFF
--- a/common/app/services/DynamoDB.scala
+++ b/common/app/services/DynamoDB.scala
@@ -8,8 +8,8 @@ import conf.Configuration
 
 
 object DynamoDB {
-  private lazy val credentials = Configuration.aws.mandatoryCredentials
-  private lazy val region = Region.getRegion(Regions.fromName(Configuration.aws.region))
+  private val credentials = Configuration.aws.mandatoryCredentials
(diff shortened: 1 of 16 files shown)


jfsoul (approved):
🎉

alexduf:
This is making me happy! :)

alexduf (approved):

common/app/services/DynamoDB.scala
```diff
@@ -8,8 +8,8 @@ import conf.Configuration
 
 
 object DynamoDB {
-  private lazy val credentials = Configuration.aws.mandatoryCredentials
-  private lazy val region = Region.getRegion(Regions.fromName(Configuration.aws.region))
+  private val credentials = Configuration.aws.mandatoryCredentials
```
alexduf: You made all the properties lazy, except these two where you removed the laziness, is that voluntary?

TBonnin: It's an object so they don't need to be lazy right?

TBonnin: Reverted in 57aa80a7d3cb00dc713d7692db59c6729a5ab3cc

PR #15344, opened 2016-12-15
Extend hardcoded hosted campaign switches

kelvin-chappell:
They're all going to be removed in #15304 but this is just to keep them going in meantime.

/cc @guardian/labs-beta


common/app/conf/switches/CommercialSwitches.scala
```diff
@@ -172,7 +172,7 @@ trait CommercialSwitches {
     "Make the Chester Zoo gallery page available",
     owners = Owner.group(CommercialLabs),
     safeState = Off,
-    sellByDate = new LocalDate(2016, 12, 21),
+    sellByDate = new LocalDate(2017, 25, 1),
```
kelvin-chappell: Wow, that was stupid.  Thanks!

lps88:
👍
Eventually we will get rid of the hardcoded stuff......... 😞

PR #15345, opened 2016-12-15
Add poster only template for media atoms

gidsg:
## What does this change?
Adds a template for case where media atom has `posterUrl` but no assets. This is because editorial may wish to preview or publish an media atom before assets are available.

Fix previous exception when a media atom with no assets was embedded. Support poster only case. Adds caption icon across media atom cases.

Adds AMP template for poster only

## Screenshots
.com

screen shot 2016-12-15 at 10 36 38

AMP

screen shot 2016-12-15 at 10 36 23

### DIFF
--- a/common/app/views/fragments/atoms/mediaPosterOnly.scala.html
+++ b/common/app/views/fragments/atoms/mediaPosterOnly.scala.html
@@ -1,0 +1,19 @@
+@(media: model.content.MediaAtom)
+<figure
+itemprop="associatedMedia image"
+itemscope
+itemtype="http://schema.org/ImageObject"
+data-component="image" xmlns="http://www.w3.org/1999/html"
+class="element element-image">
+    <div class="u-responsive-ratio">
+        <picture>
+            <img itemprop="contentUrl"
+            alt="@media.title"
+            src="@media.posterUrl.get">
+        </picture>
+    </div>
+    <figcaption class="caption caption--img" itemprop="description">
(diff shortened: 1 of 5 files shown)


common/app/views/fragments/atoms/mediaPosterOnly.scala.html
```diff
@@ -0,0 +1,17 @@
+@(media: model.content.MediaAtom)
+<figure
+itemprop="associatedMedia image"
+itemscope
+itemtype="http://schema.org/ImageObject"
+data-component="image" xmlns="http://www.w3.org/1999/html"
+class="element element-image">
+    <div class="u-responsive-ratio">
+            <img
```
gidsg: I am just using a simple `` tag here as we only have a single image URL.

akash1810: there is a list of poster image assets on the atom that can be used

akash1810: is this passed through imgix?

akash1810 (approved):
👍 after rebasing to get build to pass

common/app/views/fragments/atoms/mediaPosterOnly.scala.html
gidsg: Yes, it is passed through Imgix in the building of the case class

gidsg: Thanks for letting me know about the `PosterImage` changes, I would rather merge this using the old CAPI model and pick up that work as a separate PR.

PR #15346, opened 2016-12-15
Change the message on the International membership engagement banner

rupertbates:
## What does this change?
Updates the message on the International version of the membership engagement banner

This is the message that has consistently performed best in other regions so we want to roll it out to international users as well.

Web only


JustinPinner:
Looks good to me 👍

PR #15353, opened 2016-12-15
Fix instant ad loaded twice

regiskuckaertz:
I just introduced a bug by taking adverts from the wrong array in the bootstrapping of the lazy loading process. As a result, ads that are "instantly loaded" (like the pageskin, the high-merch) are loaded twice.


rich-nguyen:
👍

rich-nguyen (approved):

PR #15355, opened 2016-12-15
Don't use display: block links in email HTML

joelochlann:
Some email clients (all Windows versions of Outlook, and Gmail on Android) had a problem with the way I was putting `display: block` on the `<a></a>` element and using it for background colour and padding (my thought was that if I could get away with it, I would save quite a bit of markup and reduce the payload).

Since I can't get away with it, I've introduced a bit more `@fullRow` table cruft and ended up with:
- `@fullRow` with `.facia-card__text` wraps the headline & standfirst, providing background colour & internal padding
- `<div class="facia-card">` wraps the image and the `.facia-card__text`, providing top border
- `<a class="facia-link>` wraps the facia-card div, providing links
- `@paddedRow` wraps the lot and provides margins between cards

I also added in some logic to set the `width` attribute on the img elements to the width of the container. I didn't think this affected anything but it does - in Outlook, of course. So this fixed some knock-on problems from my changes.

And I moved some styles out of generic `h1`, `p` etc into classes which describe the role of that element - while refactoring, having the styles tied to specific tags was a nightmare (see also: BEM).

## Outlook 2007 before
picture 370

## Outlook 2007 after
picture 368

## Android Gmail before
picture 371

## Android Gmail after
picture 369

@guardian/dotcom-platform @lmath

### DIFF
--- a/common/app/views/support/EmailHelpers.scala
+++ b/common/app/views/support/EmailHelpers.scala
@@ -53,13 +53,13 @@ object EmailHelpers {
     s"""<img src="${Static(s"images/email/icons/$name.png")}" class="icon icon-$name">"""
   }
 
-  def imgForArticle(src: String, alt: Option[String] = None) = Html {
-    s"""<img width="${EmailImage.knownWidth}" class="full-width" src="$src" ${alt.map(alt => s"""alt="$alt"""").getOrElse("")}>"""
+  private def img(src: String, width: Int, alt: Option[String] = None) = Html {
+    s"""<img width="$width" class="full-width" src="$src" ${alt.map(alt => s"""alt="$alt"""").getOrElse("")}>"""
   }
 
-  def imgForFront(src: String, alt: Option[String] = None) = Html {
-    s"""<img width="${FrontEmailImage.knownWidth}" class="full-width" src="$src" ${alt.map(alt => s"""alt="$alt"""").getOrElse("")}>"""
-  }
+  def imgForArticle(src: String, alt: Option[String] = None) = img(src, EmailImage.knownWidth, alt)
+
+  def imgForFront(src: String, alt: Option[String] = None) = img(src, FrontEmailImage.knownWidth, alt)
 
   def imgFromPressedContent(pressedContent: PressedContent) = imageUrlFromPressedContent(pressedContent).map { url =>
     imgForFront(src = url, alt = Some(pressedContent.header.headline))
--- a/common/app/views/support/EmailHelpers.scala
+++ b/common/app/views/support/EmailHelpers.scala
@@ -53,12 +53,16 @@ object EmailHelpers {
     s"""<img src="${Static(s"images/email/icons/$name.png")}" class="icon icon-$name">"""
   }
 
-  def img(src: String, alt: Option[String] = None) = Html {
-    s"""<img width="580" class="full-width" src="$src" ${alt.map(alt => s"""alt="$alt"""").getOrElse("")}>"""
+  def imgForArticle(src: String, alt: Option[String] = None) = Html {
+    s"""<img width="${EmailImage.knownWidth}" class="full-width" src="$src" ${alt.map(alt => s"""alt="$alt"""").getOrElse("")}>"""
+  }
+
+  def imgForFront(src: String, alt: Option[String] = None) = Html {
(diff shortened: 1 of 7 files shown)


common/app/views/support/Profile.scala
```diff
@@ -104,7 +104,7 @@ object Video700 extends VideoProfile(width = Some(700), height = Some(394)) // 1
 abstract class ShareImage(shouldIncludeOverlay: Boolean) extends Profile(width = Some(1200)) {
   override val heightParam = "h=630"
   override val fitParam = "fit=crop"
-  val cropParam = "crop=faces%2Centropy"
```
katebee: Is this supposed to be changing?

SiAdcock:
Tidy stuff, dude 👍

common/app/views/support/Profile.scala
jfsoul: Good spot! Please don't 😄

common/app/views/support/Profile.scala
```diff
@@ -129,6 +129,7 @@ object FacebookOpenGraphImage extends ShareImage(FacebookShareImageLogoOverlay.i
 
 object EmailImage extends Profile(width = Some(580), autoFormat = false) {
   override val qualityparam = "q=40"
+  val knownWidth = 580
```
jfsoul: How come you can't use `width`?

jfsoul (changes requested):
Watch out for the encoding on that crop param.  Without that we get a lot of errors from facebook's crawler for share images.

common/app/views/support/Profile.scala
joelochlann: 😮  whoops, must've been a bad rebase from master - thanks for spotting @katebee

common/app/views/support/Profile.scala
joelochlann: Yeah I agree it's a bit rubbish duplicating that value. Really what I want is to be able to say that `width` is not optional on this subtype - because it's an `Option` I'd need to `.getOrElse` and provide a fallback value wherever I use it, which would then duplicate the `width` value anyway. So I thought duplicating the value in one place and right next to the original was preferable. However they may be a much nicer way to do this - open to suggestions

jfsoul: I suppose you could do `knownWidth = width.get`?

I don't think doing `EmailImage.get` is too bad either since you know which subtype is used in the Helpers.  If later somebody tries to do the same thing with another subtype with `None` for width, it will still compile - but should be pretty obviously broken at runtime.

jfsoul (approved):
Your call what to do about the Helpers/widths but otherwise looks great!

common/app/views/support/EmailHelpers.scala
```diff
@@ -53,12 +53,16 @@ object EmailHelpers {
...
   }
 
-  def img(src: String, alt: Option[String] = None) = Html {
-    s"""<img width="580" class="full-width" src="$src" ${alt.map(alt => s"""alt="$alt"""").getOrElse("")}>"""
+  def imgForArticle(src: String, alt: Option[String] = None) = Html {
+    s"""<img width="${EmailImage.knownWidth}" class="full-width" src="$src" ${alt.map(alt => s"""alt="$alt"""").getOrElse("")}>"""
+  }
+
+  def imgForFront(src: String, alt: Option[String] = None) = Html {
+    s"""<img width="${FrontEmailImage.knownWidth}" class="full-width" src="$src" ${alt.map(alt => s"""alt="$alt"""").getOrElse("")}>"""
```
joelochlann: I'm not sure what you're suggesting is a higher order function (it doesn't accept a function as a parameter or return a function), but I agree that this markup could be factored out. Have done so

jfsoul: My example had the params the wrong way round, but `img(width)` was intended to return `(String, Option[String]) => Int`.

    def img(knownWidth: Int)(src: String, alt: Option[String] = None) = Html { ... }
    def imgForFront = img(FrontEmailImage.knownWidth)
    def imgForArticle = img(EmailImage.knownWidth)

joelochlann: Ahhh I see what you mean. That avoids the repetition of the parameters and default values which are only applicable to the core function. Thanks for the tip, have addressed this in #15408

PR #15340, opened 2016-12-14
Build standalone bootstraps/enhaced/youtube.js bundle - used in youtu…

GHaberis:
## What does this change?

- Fix issue with embedded youtube atoms (views/fragments/atoms/mediaEmbed.scala.html) where common/modules/atom/youtube.js is unavailable (403s when required by bootstraps/enhanced/youtube).

- bootstraps/enhanced/youtube.js is required in 2 locations bootstraps/enhanced/main.js and views/fragments/atoms/mediaEmbed.scala.html. Because it's used by main when the youtube.js asset is built it doesn't create a standalone bundle that can be used by mediaEmbed.scala.html.

## What is the value of this and can you measure success?

Embedded youtube atoms can be enhanced

## Does this affect other platforms - Amp, Apps, etc?

No

## Screenshots

N/A

## Request for comment

@sndrs @SiAdcock

PR #15357, opened 2016-12-15
Revert "Build standalone bootstraps/enhaced/youtube.js bundle - used in youtu…"

SiAdcock:
Reverts guardian/frontend#15340

This change produces a duplicate define error when using hashed bundles.


GHaberis:
go for it 👍

PR #15358, opened 2016-12-15
Restart Weback test

SiAdcock:
## What does this change?

Our Webpack test may not be entirely reliable, as the variant bucketing was implemented a day before the control bucketing. I'd like to restart the test with clean headers and a clean test name, to ensure the test is fair.

More scientific test FTW

Oh no

## Related

guardian/fastly-edge-cache#728

@guardian/dotcom-platform


jfsoul (approved):
camelCaseMagic looks good to me 👍

PR #15360, opened 2016-12-15
Code cleanup

regiskuckaertz:
Removing some unused stuff, and also renaming some classes for consistency.

(Look at each commit separately)

cc @kelvin-chappell hope I won't break anything 😆


kelvin-chappell (approved):
Nice demolition work!

PR #15364, opened 2016-12-16
Fix check to detect user in Always Ask test

desbo:
## What does this change?
Fixes a bug in detecting a user's allocation to one of the contributions tests.

Other tests were relying on this check to run, so they will run now.


alexduf (approved):
👍

PR #15370, opened 2016-12-16
Small commercial bootstrap cleanup

regiskuckaertz:
The `DisableStickyAdBannerOnMobile` switch does not make sense anymore, since there is no top banner on mobile. There's also some fine-tuning of the order of the secondary modules:

- "hosted by" modules should not trump the DFP loading process (after looking into `hosted/*` I think the 20s loading of commercial bootstrap makes a lot of sense)
- I moved out the `cm-ready` to after secondary modules have executed ... this make the second baseline measurement slightly more accurate
- finally, I now return the `loadModule` promise chain, so that later on we can improve `boot.js` to only start loading `enhanced` when the initialisation of `commercial` is done.


regiskuckaertz:
[Improve commercial loading order](https://trello.com/c/E8ojiTwb/260-improve-commercial-loading-order)

rich-nguyen:
👍  much better

rich-nguyen (approved):

PR #15373, opened 2016-12-16
Include firefox adblock detection

Calum-Campbell:
## What does this change?
 Two things : 
- The adblock_detected flag was broken on firefox due to a adblockPlus update, this updates the logic to identify when adblockerPlus is being used on FF.
- The old adblock detection was mostly using synchronous calls, this promisifys the adblock detection and makes it asychronous and updates the various parts of frontend that used the adblock detection to this new functionality. 

@guardian/commercial-dev @sndrs


static/src/javascripts/projects/common/modules/commercial/adblock-messages.js
```diff
@@ -32,18 +34,20 @@ define([
     }
 
     function noAdblockMsg() {
-        return adblockInUseSync() && notMobile() && (
+        return adblockInUse().then(function(adblockInUse){
```
regiskuckaertz: We could improve this expression and the next, by delaying the adblock test as much as possible:

```
return notMobile() && (!visitedMoreThanOnce() || !isAdblockSwitchOn() || (visitedMoreThanOnce() && isPayingMember())) ?
  adblockInUse() :
  Promise.resolve(false);
```

(notice I removed the redundant `isAdblockSwitchOn` call in the third OR clause)

I'll leave you the one in `showAdblockMsg` as a small exercise.

static/src/javascripts/projects/common/modules/commercial/donot-use-adblock.js
```diff
@@ -92,14 +92,15 @@ define([
     }
 
     function init() {
+        adblockMsg.showAdblockMsg().then(function(boolean){
```
regiskuckaertz: Ibid.

static/src/javascripts/projects/common/utils/raven.js
```diff
@@ -46,8 +46,11 @@ define([
                     }
                 }
 
-                return config.switches.enableSentryReporting &&
+                return adBlockerOn.then(function(adBlockerOn){
```
regiskuckaertz: This change is the _very_ problematic. It requires changes in raven.js, which is a third-party library.

regiskuckaertz: Perhaps we could keep it sync using the same technique than in `tech-feedback`; I think it is safe to assume the promise will be resolved before this is even executed most of the times.

regiskuckaertz (changes requested):
A few things to ponder, but overall great work pal!

regiskuckaertz (approved):

static/src/javascripts/projects/commercial/modules/third-party-tags/outbrain.js
```diff
@@ -113,8 +113,10 @@ define([
      make the call instantly when we detect adBlock in use.
      */
     function loadInstantly() {
-        return !document.getElementById('dfp-ad--merchandising-high') ||
-            detect.adblockInUseSync();
+        detect.adblockInUse.then(function(adblockInUse){
```
gtrufitt: Do you need to `return` this?

static/src/javascripts/projects/common/utils/detect.js
```diff
@@ -339,21 +339,6 @@ define([
         return window.guardian.isEnhanced;
     }
 
-    var createSacrificialAd = memoize(function () {
-        var sacrificialAd = '<div class="ad_unit" style="position: absolute; height: 10px; top: 0; left: 0; z-index: -1;">&nbsp;</div>';
-        document.body.insertAdjacentHTML('beforeend', sacrificialAd);
-        return document.body.lastChild;
-    });
-    // sync adblock detection is deprecated.
-    // it will be removed once sticky nav and omniture top up call are both removed
```
gtrufitt: Good deprecation 👌

static/test/javascripts/spec/commercial/creatives/fluid250.spec.js
```diff
@@ -18,6 +18,10 @@ define([
 
     describe('Fluid 250', function () {
 
+        afterEach(function () {
```
gtrufitt: 💯

static/test/javascripts/spec/commercial/third-party-tags/outbrain.spec.js
```diff
@@ -71,7 +71,8 @@ define([
                 identity.isUserLoggedIn = function () {
                     return false;
                 };
-                detect.adblockInUseSync = function () { return false; };
+                // detect.adblockInUseSync = function () { return false; };
```
gtrufitt: You meant to remove this?

static/test/javascripts/spec/commercial/third-party-tags/plista.spec.js
```diff
@@ -112,27 +106,33 @@ define([
...
-                sut.init();
-                expect(sut.load).not.toHaveBeenCalled();
+                sut.init().then(function () {
+                    expect(sut.load).not.toHaveBeenCalled();
+                    done();
+                });
             });
-
-            it('should not load when is preview', function () {
+            //
```
gtrufitt: Stray `//`

static/test/javascripts/spec/common/utils/detect.spec.js
```diff
@@ -68,10 +68,11 @@ define([
             expect(detect.isBreakpoint({ min: 'tablet', max: 'wide' })).toBe(true);
         });
 
-        it('should append a surrogate ad to the body', function () {
-            var html = '<div class="ad_unit" style="position: absolute; height: 10px; top: 0; left: 0; z-index: -1;">&nbsp;</div>';
-            detect.adblockInUseSync();
-            expect(document.body.lastChild.outerHTML).toEqual(html);
-        });
+        // it('should append a surrogate ad to the body', function () {
```
gtrufitt: 💥 ❗️ ❓

static/src/javascripts/projects/common/modules/commercial/donot-use-adblock.js
```diff
@@ -92,14 +92,15 @@ define([
...
     function init() {
+        adblockMsg.showAdblockMsg().then(function(adBlockInUse){
         // Show messages only if adblock is used by non paying member
-        if (adblockMsg.showAdblockMsg()) {
+        if (adBlockInUse) {
             showAdblockMessage();
             showAdblockBanner();
         }
+        });
         mediator.emit('banner-message:complete');
```
gtrufitt: Is this mediator event firing at the right time now things are async?

static/src/javascripts/projects/common/modules/onward/tech-feedback.js
```diff
@@ -116,6 +117,10 @@ define([
      * override those at the time the email is sent.
      */
     return function () {
+        detect.adblockInUse.then(function(adblockInUse){
+            adblockBeingUsed = adblockInUse;
```
gtrufitt: You probably need to make the whole chain use promises now this is async, otherwise your `adblockBeingUsed` var is always going to be `false` - the functionality below needs to be inside this `then`, right?

regiskuckaertz: > your `adblockBeingUsed` var is always going to be `false`

How so? When `detect.adblockInUse` resolves, all the chained promises across all modules will fire synchronously. There is a slight chance, if the user is fast enough (in `ms`), of a false negative.

If we turn this thing async, we'll need to fire `click` events programmatically ... alas, Chrome recently decided not to invoke the default action with "untrusted" events 😩

static/src/javascripts/projects/commercial/modules/third-party-tags/outbrain.js
regiskuckaertz: Oops, I did not see this one. Thanks @gtrufitt 😄

regiskuckaertz (changes requested):

static/src/javascripts/projects/common/modules/onward/tech-feedback.js
gtrufitt: I'm confused I think - because we're calling a promise here this is async, which means you can't guarantee that the `adblockBeingUsed` value is going to be set before its used in the below code?

Small test case, this is equivalent right (note using setTimeout to simulate a click delay)? http://jsbin.com/vunucutovi/edit?js,console

```javascript
var myProm = new Promise(function(resolve){
  return resolve('from the promise');
})

setTimeout(function(){
  var myVar = 'not from the promise';
  
  myProm.then(function(promVal){
    myVar = promVal;
    console.log('in the promise, myVar: ' + myVar)
  })
  
  console.log('after the promise, myVar: ' + myVar);
  
  setTimeout(function(){
    console.log('delay the log, myVar:' + myVar)
  }, 10)
}, 1000)
```

That logs out:

```
"after the promise, myVar: not from the promise"
"in the promise, myVar: from the promise"
"delay the log, myVar:from the promise"
```

gtrufitt: Thank you @regiskuckaertz for explaining to me in person why it's not the same because of the event listeners :dancing_men:

static/src/javascripts/projects/common/modules/commercial/adblock-messages.js
```diff
@@ -32,18 +34,15 @@ define([
     }
 
     function noAdblockMsg() {
-        return adblockInUseSync() && notMobile() && (
-                !visitedMoreThanOnce() ||
-                !isAdblockSwitchOn() ||
-                (isAdblockSwitchOn() && visitedMoreThanOnce() && isPayingMember()));
+        return notMobile() && (!visitedMoreThanOnce() || !isAdblockSwitchOn() || (visitedMoreThanOnce() && isPayingMember())) ?
```
Calum-Campbell: Yeah agree, muchos more readable.

static/test/javascripts/spec/commercial/third-party-tags/outbrain.spec.js
Calum-Campbell: Yep cheers.

static/src/javascripts/projects/commercial/modules/third-party-tags/outbrain.js
```diff
@@ -172,12 +174,14 @@ define([
...
+            loadInstantly().then(function(shouldLoadInstantly) {
+                if (shouldLoadInstantly) {
+                    return checkDependencies().then(function (widgetType) {
+                        widgetType ? module.load(widgetType) : module.load();
+                        return Promise.resolve(true);
+                    });
+                }
+            });
 
             return trackAdRender('dfp-ad--merchandising-high').then(function (isHiResLoaded) {
```
regiskuckaertz: Move that into the `loadInstantly.then` otherwise it will be called no matter what.

Calum-Campbell:
[Data lake 'ad_blocking_detected' flag and 'ads_loaded' flags are both broken](https://trello.com/c/vx8raRGP/238-data-lake-ad-blocking-detected-flag-and-ads-loaded-flags-are-both-broken)

static/src/javascripts/projects/common/modules/commercial/donot-use-adblock.js
regiskuckaertz: @Calum-Campbell buddy?

PR #15374, opened 2016-12-16
Remove unused guardian.secure in configurations

TBonnin:
## What does this change?
Remove unused guardian.secure parameter in admin and identity configurations

Cleanup 🔪


jfsoul (approved):
👍

PR #15378, opened 2016-12-16
Remove unnecessary TestSettings trait

TBonnin:
## What does this change?
It was a leftover from a time where CapiClient was a global object

🔪


jfsoul (approved):
If the tests pass 👍

PR #15380, opened 2016-12-16
More cleanup after the upgrade

DiegoVazquezNanini:
## What does this change?
Removes unused Environment imports
Replace env for AppContext


TBonnin (approved):
Thanks @DiegoVazquezNanini

PR #15381, opened 2016-12-17
Speedup commercial boot

regiskuckaertz:
Ok there are two things going on here:

1- `prepare-googletag` and `fill-advert-slots` won't slow down the startup process anymore.
2- in both module, we want to know how much time it takes DFP to empty the command queue, but we don't want this process to block the other modules.

As long as (1) every interaction with DFP is done by means of the command queue (which it is) and (2) DFP processes the command queue deterministically (which it does), all we need to care about in the startup process is that the command queue has been correctly initialized.

Paradoxically, this change will not improve the total running time of each of these modules, but we should see a significant drop in the running time of primary and secondary module batches.

### DIFF
--- a/static/src/javascripts/bootstraps/commercial.js
+++ b/static/src/javascripts/bootstraps/commercial.js
@@ -72,7 +72,7 @@ define([
     ];
 
     var secondaryModules = [
-        ['cm-fill-advert-slots', fillAdvertSlots.init],
+        ['cm-fill-advert-slots', fillAdvertSlots.init, fillAdvertSlots.customTiming],
         ['cm-paidforBand', paidforBand.init],
         ['cm-paidContainers', paidContainers.init],
         ['cm-ready', function () {
--- a/static/src/javascripts/projects/commercial/modules/dfp/prepare-googletag.js
+++ b/static/src/javascripts/projects/commercial/modules/dfp/prepare-googletag.js
@@ -54,26 +54,24 @@ define([
             // Use Custom Timing to time the googletag code without the sonobi pre-loading.
             performanceLogging.moduleStart(moduleName);
 
-            return new Promise(function(resolve) {
+            var promise = Promise.resolve(require(['js!googletag.js']));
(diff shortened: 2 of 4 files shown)


static/src/javascripts/bootstraps/commercial.js
```diff
@@ -72,7 +72,7 @@ define([
     ];
 
     var secondaryModules = [
-        ['cm-fill-advert-slots', fillAdvertSlots.init],
+        ['cm-fill-advert-slots', fillAdvertSlots.init, fillAdvertSlots.customTiming],
```
rich-nguyen: sweet

static/src/javascripts/projects/commercial/modules/dfp/prepare-googletag.js
```diff
@@ -54,38 +50,27 @@ define([
...
 
-                window.googletag.cmd.push(
-                    setListeners,
-                    setPageTargeting,
-                    moduleEnd,
-                    resolve
-                );
-            });
+            // Just load googletag. Sonobi's wrapper will already be loaded, and googletag is already added to the window by sonobi.
+            return Promise.resolve(require(['js!googletag.js']));
```
rich-nguyen: that's interesting, we consider this module resolved when the script is loaded, instead of when the buffer is flushed. I like it.

rich-nguyen:
that's cracking. 👍  🎄

rich-nguyen (approved):

gtrufitt:
Kilin' it. ❤️

static/src/javascripts/projects/commercial/modules/dfp/prepare-googletag.js
regiskuckaertz: Yes 😃  I would even resolve when the request for the script is done, so that the baseline is not impacted by random network conditions. I think it is a good practice to separate module running time from module initialisation.

PR #15382, opened 2016-12-19
[POC] Deterministic asynchronous task scheduler

regiskuckaertz:
Even though we are all religiously sealing DOM interactions between Fastdom walls, most of the front-end code still looks like this:

black-friday

... something the browser is not really fond of. Even though Web Workers have been available for years, they are still not widely used, quite understandably: with no access to DOM-related APIs, there is not much a Web Worker can do, not in the context of a site like ours. For all its attractiveness, I don't think VDOM is a solution, not with [the optimisations that browser makers are now heading towards](https://servo.org/)—ironically, the React Native team found out that the best optimisation they could come up with was ... [direct DOM manipulation](https://speakerdeck.com/vjeux/react-rally-animated-react-performance-toolbox).

That leaves us with a single thread to deal with a plethora of modules that all vie for the same opportunity. To make matters even worse, once a module gets hold of the main thread, it can keep it for as long as it desires.

Of course, we all know that and diligently leverage the opportunity that asynchronous calls give us, by delaying the execution of synchronous code until it becomes necessary. This is queueing theory 101. Still, synchronous code must run at some point, and run it will, and so this solution doesn't solve the main problem at all. If anything, it leaves all the responsibility of running the code to the vagaries of network performance: objectively, the main thread will be held for the same amount of time.

There's now an opportunity to improve that, with `requestIdleCallback` [around](https://developer.microsoft.com/en-us/microsoft-edge/platform/status/requestidlecallback/) [the](https://bugzilla.mozilla.org/show_bug.cgi?id=1198381) [corner](https://webkit.org/status/#feature-requestidlecallback). With this API, modules have now the ability to schedule work deterministically and let the browser decide when it is best to execute it. It even has room for "out-of-band modules" via the `timeout` option.

But first we need to recognize that our modules are mostly made of _discrete_ tasks that may not need to run all at once. The standard [main.js](https://github.com/guardian/frontend/blob/master/static/src/javascripts/bootstraps/standard/main.js) is a good example of that.

I think the _command queue_ pattern used by ga.js or googletag.js is a powerful abstraction that could be useful here, so I wrote this small proof-of-concept for us to play with. It is by no means complete, but I thought I'd share before having actually used it with the hope it would spark some interesting discussions.

@gtrufitt @sndrs @SiAdcock @desbo @rich-nguyen

### DIFF
--- a/static/src/javascripts/projects/common/utils/schedule.js
+++ b/static/src/javascripts/projects/common/utils/schedule.js
@@ -9,7 +9,7 @@ define(function () {
     function enqueue(tasks, options) {
         options = options || {};
 
-        var taskRunner = options.taskRunner : iife;
+        var taskRunner = options.taskRunner || iife;
         var args = options.args || [];
 
         schedule(run.bind(tasks, taskRunner, args), options.timeout | 0);
@@ -24,12 +24,12 @@ define(function () {
     }
 
     function run(taskRunner, args, deadline) {
-        while((deadline.didTimeout || deadline.timeRemaining() > 0) && tasks.length) {
+        while((deadline.didTimeout || deadline.timeRemaining() > 0) && this.length) {
             var task = this.shift();
             taskRunner(task, args);
         }
 
-        if (tasks.length) {
+        if (this.length) {
             schedule(run.bind(this, taskRunner, args));
         }
     }
--- a/static/src/javascripts/projects/common/utils/schedule.js
+++ b/static/src/javascripts/projects/common/utils/schedule.js
@@ -1,0 +1,40 @@
+define(function () {
+    var mockDeadline = { didTimeout: true };
+
+    return {
+        enqueue: enqueue,
+        schedule: schedule
+    };
+
+    function enqueue(tasks, options) {
+        options = options || {};
+
+        var taskRunner = options.taskRunner : iife;
+        var args = options.args || [];
+
+        schedule(run.bind(tasks, taskRunner, args), options.timeout | 0);
+    }
+
+    function schedule(fn, timeout) {
+        if ('requestIdleCallback' in window) {
+            window.requestIdleCallback(fn, timeout ? { timeout: timeout } : null);
+        } else {
+            fn(mockDeadline);
+        }
+    }
+
+    function run(taskRunner, args, deadline) {
+        while((deadline.didTimeout || deadline.timeRemaining() > 0) && tasks.length) {
+            var task = this.shift();
+            taskRunner(task, args);
+        }
+
+        if (tasks.length) {
+            schedule(run.bind(this, taskRunner, args));
+        }
+    }
+
+    function iife(fn, args) {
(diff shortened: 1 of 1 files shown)


sndrs:
i think this is very interesting. i tried something similar just after we had launched, and we were trying to look at what jake archibald called 'the 9am rush'. i hooked a lightweight queue up to robust, but ended up dropping it because it had almost no effect on the time line. 

that was pre-`requestIdleCallback` though, so this seems a good time to try something like this, as you say. have you tried it out at all? can you see an diff on the timeline?

sndrs:
do you think it would nice to add a timer in dev that could tell you the code you had scheduled exceeded the 10-odd ms it should complete in?

rich-nguyen:
I think a deadline-based timer sounds very cool, @sndrs.

If our application was divided into modules, and those modules did not care when they executed, but cared only for their order, then this idea of a scheduler would take the responsibility of dispatching modules through the queue, depending on the elapsed duration of the current frame. Effectively, the scheduler's role would be to allocate 16ms worth of work, every frame. It would be synchronised with v-sync, ideally, but the best proxy we have of that is currently RAF, and maybe `requestIdleCallback` in future.

It's definitely something I've thought about for `canvas` related projects, but more generally, any web or native app has to have something making temporal decisions. Maybe historically more than anything else, browsers have done that on their own, without help from the js application.

regiskuckaertz:
Yes, `requestIdleCallback` is as close as we can get to the metal. It differs from RAF in that it works like a basic OS scheduler, which is awesome.

I had tried on commercial.js to see if it worked, but haven't made any recording yet.

The timer thing-y is a great idea. I can map the ID of each callback to its running time and even string them together when a single callback does not suffice. In essence it would be similar to [`takeRecords`](https://developer.mozilla.org/en/docs/Web/API/MutationObserver#takeRecords())

sndrs:
i also wondered about being able to assign importance? we'd obviously need to guard against everything saying it was important, but, for instance, it seems less important that breaking news completes than showing ads?

sndrs:
we could then send things like error reporting to the back of the queue, for instance. batch them up and only send them when there's nothing else to do (which could also mean we could check the eval time of items in the queue IRL and hit sentry with the slugabeds?)

regiskuckaertz:
Yeah I thought about that. It seems sensible indeed, I guess a simple round-robin would be enough here. Then it's all a matter of choosing how many priority levels we need.

Side note: we are in the process of speeding up the [commercial init time](http://dashboards.gu-web.net/s3?file=dotcom-commercial/health/commercial_module_breakdown.stacking.csv#) and I was wondering if in boot.js the `bootCommercial` should not return a promise that resolves when the init is over?

regiskuckaertz:
Closing this one. I'll keep on hacking this on the side, and if anyone's interested we can have a session together? @sndrs @desbo @rich-nguyen

PR #15384, opened 2016-12-19
Prune cloudwatch beacons

rich-nguyen:
The aim here is to up our commercial confidence graph, which measures `pages with at least one ad    loaded` / `page views`. I believe that, due to lack of `navigator.sendBeacon` support, we should just use the simple `count.gif` system until we get better beacon API coverage.

There were a load of unused prod metrics, which I thought would be worth cleaning up too. If 2017, is anything like this year, we need to free up some metrics space so we can measure all the horrible things.

screen_shot_2016-12-19_at_12_17_20


TBonnin (approved):
LGTM 🔪

rich-nguyen:
I think that's done it

screen shot 2016-12-19 at 16 08 31

rich-nguyen:
[Update the commercial confidence dashboard](https://trello.com/c/N7VzBBzr/261-update-the-commercial-confidence-dashboard)

PR #15351, opened 2016-12-15
Fix the newsletters JS and stop the 403

katebee:
## What does this change?

- Adds missing file to the lists in curlConfig and webpack
- Rename `signup.js` to `newsletters.js`, since this code only runs on the newsletters page...

## What is the value of this and can you measure success?

- Allow `newsletters.js` to load on https://www.theguardian.com/email-newsletters
- Newsletters page can be enhanced

## Does this affect other platforms - Amp, Apps, etc?

Nope

-------
@gtrufitt

PR #15385, opened 2016-12-19
Add new JS file to all the bundlers... maybe (again)

katebee:
## What does this change?

- Adds missing file to the lists in curlConfig and webpack
- Rename `signup.js` to `newsletters.js`, since this code only runs on the newsletters page...

- Allow `newsletters.js` to load on https://www.theguardian.com/email-newsletters
- Newsletters page can be enhanced

Lets try #15351 again!

Nope

------------

I've tried running `dev-build` with `assets.useHashedBundles=true`... this all seems to work. 
The `newsletters.js` and `newsletters-rjs.js` are appearing in the assets map

@SiAdcock @gtrufitt


gtrufitt (approved):
Eek... Sorry this is so tough at the moment! Worth checking with the `useHashedBundles=true` after opting into webpack too if you haven't `localhost:9000/opt/in/webpack`. Other than that, it looks good (again) to me.

katebee:
@gtrufitt cheers! Will try it again

PR #15389, opened 2016-12-19
Fixing css validation issues

NataliaLKB:
## What does this change?
For a while sass validations errors were not showing or causing PRs to fail. This has meant a lot of code has gone into production which our linter isn't happy with. 

This fixes the code which doesn't abide by our linting rules which makes it so that we can start failing PRs on them again. ( @sndrs )

Better code quality and consistency

It affects amp styling, but not apps or any other off platform (only in the frontend repo)

@guardian/dotcom-platform


static/src/stylesheets/module/commercial/glabs/_hosted-gallery.scss
```diff
@@ -548,12 +548,8 @@ $progressColor: #ffffff;
     width: $progressHeight;
 
     &:before {
-        content: "";
+        content: '';
         position: absolute;
-        background-image: -o-radial-gradient(center center, ellipse cover, $neutral-1 0%, transparent 50%);
```
NataliaLKB: @lps88 I have removed these as I think auto prefixer deals with this anyway. Is that okay with you?

static/src/stylesheets/module/facia/_item.scss
```diff
@@ -157,7 +157,6 @@ $fc-item-gutter: $gs-gutter / 4;
 
     a {
         color: $neutral-1;
-        color: inherit;
```
gtrufitt: Isn't the `$neutral-1` a fallback for browsers that don't support `inherit`?

gtrufitt (approved):
1 minor query but LGTM - yay for linting!

static/src/stylesheets/module/facia/_item.scss
NataliaLKB: didn't know that! But it does make sense. I will put this line back.

PR #15390, opened 2016-12-19
Improve hosted by modules init time

regiskuckaertz:
A few small changes to the delineation of the "Hosted by" modules initialisation.

We are in a situation where `enhanced` starts loading before all our commercial modules have had the chance to set everything up. That is because the boot script [does not wait](https://github.com/guardian/frontend/blob/master/static/src/javascripts/boot.js#L64) for commercial code to get through.

We could change that now, but there are still modules that take too long to initialise. In this PR, I address the "Hosted by" modules, so that their `init` function exists asap, without losing the ability to measure the actual running time of the whole module.

cc @lps88


rich-nguyen:
looks good 👍

PR #15396, opened 2016-12-20
Added an endpoint that returns html for a content card for recommended for you

lmath:
## What does this change?
Endpoint that returns the html for a content card, with default values (of group index, card index) to produce cards like in the screenshot below, given an id for an article or video.

Being able to use the existing contentCard template to produce the html used for recommended for you container with e.g. a particular tone, or for a video.  

No.

## Screenshots
Defaults to produce html for cards with image or video like the following:
image
image

@davidfurey


dev-build/conf/routes
```diff
@@ -120,6 +120,9 @@ GET            /video/end-slate/series/*seriesId.json
 GET            /video/end-slate/section/*sectionId.json                                                                          controllers.VideoEndSlateController.renderSection(sectionId)
 GET            /embed/card/*path.json                                                                                            controllers.RichLinkController.render(path)
 GET            /embed/card/*path                                                                                                 controllers.RichLinkController.renderHtml(path)
+GET            /embed/contentcard/*path.json                                                                                     controllers.ContentCardController.render(path, group: Option[Int], card: Option[Int])
+GET            /embed/contentcard/*path                                                                                          controllers.ContentCardController.renderHtml(path)
```
davidfurey: Why not allow group and card params to be supplied on the renderHtml route?

paperboyo:
💎 
Does it allow for comment count, date and Save for later as well (if this hasn't been completely killed off)?

dev-build/conf/routes
lmath: I thought initially that only the render json route was required so I thought I'd have to go back to render html later if required. I found it useful so I should make sure it can accept group and card params too. Good point. 

I already tried briefly to replace the matcher *path with an expression that matches an id without swallowing the query parameters. I forgot to go back and figure it out!

lmath: I pushed a change to make it so that neither will accept the query params since there is currently no need for them.

lmath:
@paperboyo I don't think comment count is currently added via the contentCard template.

NataliaLKB:
@lmath the code looks good, but I am not sure what the purpose is? Actually, I am not sure what these "contentCards" are used for at all. How does it differ to Facia cards?

I just want to make sure we aren't duplicating work that has already been done.

NataliaLKB:
Actually I see that content cards are a type of facia card. That makes more sense!

common/app/layout/Sublink.scala
```diff
@@ -303,5 +310,51 @@ case class ContentCard(
...
   }
 }
+object ContentCard {
 
+  def makeFromApiContent(apiContent: contentapi.Content): Option[ContentCard] = {
+
+    apiContent.fields.map { fields =>
+      val discussionSettings = for {
+        commentable <- fields.commentable
+        isClosedForComments <- fields.commentCloseDate.map(closeDate => new DateTime(closeDate.dateTime).isBeforeNow)
```
jfsoul: is it possible to have this as another generator instead of a map?

common/app/layout/Sublink.scala
```diff
@@ -8,6 +8,13 @@ import play.api.mvc.RequestHeader
 import play.twirl.api.Html
 import views.support._
 import implicits.FaciaContentFrontendHelpers.FaciaContentFrontendHelper
+import cards.{MediaList, Standard}
+import com.gu.contentapi.client.model.v1
+import com.gu.contentapi.client.model.v1.{ContentType, ItemResponse, Tag, TagType}
+import com.gu.contentapi.client.model.{v1 => contentapi}
+import com.gu.facia.api.{utils => fapiutils}
+import com.gu.facia.api.utils
```
lmath: Yep! I will do this.

dev-build/conf/routes
```diff
@@ -120,6 +120,9 @@ GET            /video/end-slate/series/*seriesId.json
 GET            /video/end-slate/section/*sectionId.json                                                                          controllers.VideoEndSlateController.renderSection(sectionId)
 GET            /embed/card/*path.json                                                                                            controllers.RichLinkController.render(path)
 GET            /embed/card/*path                                                                                                 controllers.RichLinkController.renderHtml(path)
+GET            /embed/contentcard/*path.json                                                                                     controllers.ContentCardController.render(path, group: Option[Int], card: Option[Int])
```
NataliaLKB: Could we rename the controller `ContentCardController` to something a bit more descriptive?

The above two routes aren't descriptive either, but the controller is `RichLinkController` describes the feature which helps.

onward/app/controllers/OnwardControllers.scala
```diff
@@ -41,6 +41,7 @@ trait OnwardControllers {
   lazy val mostViewedGalleryController = wire[MostViewedGalleryController]
   lazy val videoEndSlateController = wire[VideoEndSlateController]
   lazy val richLinkController = wire[RichLinkController]
+  lazy val contentCardController = wire[ContentCardController]
```
NataliaLKB: Again renaming this to "recommendedContent" or something similar would make it easier to understand what this was for

common/app/layout/Sublink.scala
```diff
@@ -303,5 +307,51 @@ case class ContentCard(
...
+  def makeFromApiContent(apiContent: contentapi.Content): Option[ContentCard] = {
+
+    apiContent.fields.map { fields =>
+      val discussionSettings = for {
+        commentable <- fields.commentable
+        isClosedForComments <- fields.commentCloseDate.map(closeDate => new DateTime(closeDate.dateTime).isBeforeNow)
+      } yield DiscussionSettings(commentable, isClosedForComments, None)
+
+      val isVideo = apiContent.`type` == ContentType.Video
+      val faciaCardHeader = FaciaCardHeader(false, false, isVideo, false, false, None, apiContent.webTitle, EditionalisedLink(apiContent.webUrl))
```
jfsoul: do you mind using named parameters for all of these and some other vals below (and probably put them on newlines) - I think that tends to be our convention for boolean parameters at least

common/app/layout/Sublink.scala
```diff
@@ -303,5 +307,51 @@ case class ContentCard(
...
+      val unsetDisplaySettings = DisplaySettings(false, false, false, false, false)
+      val cardTypesForRecommendations = ItemClasses(mobile = MediaList, tablet = Standard, None)
+
+      def byline: Option[Byline] = {
+        val contributorApiTags = apiContent.tags.filter(tag => tag.`type` == TagType.Contributor)
+        val contributorTags = contributorApiTags.map(tag => model.Tag(TagProperties.make(tag), None, None, None))
+        fields.byline.map(byline => Byline(byline, contributorTags))
```
jfsoul: Minor style preference but how about this?

    val contributorApiTags = apiContent.tags.filter(_.`type` == TagType.Contributor)

and:

     fields.byline.map(Byline(_, contributorTags))

jfsoul: In fact a for comprehension for the whole of `byline` might even be neater?

common/app/layout/Sublink.scala
```diff
@@ -303,5 +307,51 @@ case class ContentCard(
...
+        id = Some(apiContent.id),
+        header = faciaCardHeader,
+        byline = byline,
+        displayElement = FaciaDisplayElement.fromContent(apiContent),
+        cutOut = None,
+        cardStyle = CardStyle.make(fapiutils.CardStyle.fromContent(apiContent)),
+        cardTypes = cardTypesForRecommendations,
+        sublinks = Seq(),
+        starRating = fields.starRating,
+        discussionSettings = discussionSettings.getOrElse(DiscussionSettings(false, false, None)),
```
jfsoul: More boolean params that you could name here

common/app/model/FaciaDisplayElement.scala
```diff
@@ -23,6 +24,23 @@ object FaciaDisplayElement {
...
+      sectionId <- apiContent.sectionId
+      fields <- apiContent.fields
+      shortUrl <- fields.shortUrl
+    } yield EndSlateComponents(None, sectionId, shortUrl)
+
+    val elements = Elements.make(apiContent)
+
+    maybeEndSlateComponents flatMap { endSlateComponents =>
+      elements.mainVideo match {
+        case Some(videoElement) => Option(InlineVideo(videoElement, apiContent.webTitle, endSlateComponents.toUriPath, Option(InlineImage(videoElement.images))))
```
jfsoul: I find this line a little hard to read, do you mind separating onto multiple lines?

common/test/model/FaciaDisplayElementTest.scala
```diff
@@ -0,0 +1,75 @@
...
+    apiUrl = "https://content.guardianapis.com/section/some/kinda/id",
+    elements = Some(List(
+      image("test-image-0", "main", 0, List(imageAsset("smaller picture 1", 50), imageAsset("biggest picture 1", 100)))
+    )),
+    fields = Some(ContentFields(
+      shortUrl = Some("https://gu.com/p/5ehd8")))
+  )
+
+  private def video(id: String, relation: String, caption: String, width: Int): ApiElement = {
+    val videoAsset = Asset(AssetType.Video, Some("video/webm"), Some("http://video.com/uri"), Some(AssetFields(caption=Some(caption), width=Some(width))))
```
jfsoul: space around `=`

onward/app/controllers/ContentCardController.scala
```diff
@@ -0,0 +1,55 @@
...
+      response.content flatMap { content =>
+        ContentCard.makeFromApiContent(content)
+      }
+    }
+  }
+
+
+  private def renderContent(content: ContentCard)(implicit request: RequestHeader) = {
+
+    def contentResponse: HtmlFormat.Appendable = views.html.fragments.items.facia_cards.contentCard(content, 0, 1, "all", false, false)(request)
```
jfsoul: named params please 😄

onward/test/ContentCardControllerTest.scala
```diff
@@ -1,11 +1,10 @@
...
-
-import controllers.RichLinkController
+import controllers.ContentCardController
 import org.scalatest.{BeforeAndAfterAll, DoNotDiscover, FlatSpec, Matchers}
-import play.api.test._
 import play.api.test.Helpers._
+import play.api.test._
 
-@DoNotDiscover class RichLinkControllerTest
+@DoNotDiscover class ContentCardControllerTest
```
jfsoul: I suppose this may be consolidated with the `RichLinkControllerTest` if the controllers could be deduped a bit, but the tests look great to me!

jfsoul (commented):
Really great work, I've made quite a few comments but nearly all very minor style things!

onward/app/controllers/ContentCardController.scala
```diff
@@ -0,0 +1,55 @@
...
+import contentapi.ContentApiClient
+import implicits.Requests
+import layout.ContentCard
+import model.{ApplicationContext, Cached, NoCache}
+import play.api.mvc.{Action, Controller, RequestHeader}
+import play.twirl.api.HtmlFormat
+
+import scala.concurrent.Future
+
+class ContentCardController(contentApiClient: ContentApiClient)(implicit context: ApplicationContext) extends Controller with Paging with Logging with ExecutionContexts with Requests   {
```
lmath: Good point about the duplication. I wasn't sure the best way to approach this but I've had a go in my most recent commit. Could you take a look and let me know what you think?

onward/app/controllers/RecommendedContentCardController.scala
```diff
@@ -0,0 +1,48 @@
...
+import scala.concurrent.Future
+
+class RecommendedContentCardController(contentApiClient: ContentApiClient)(implicit context: ApplicationContext) extends RenderTemplateController(contentApiClient) {
+
+  def render(path: String) = Action.async { implicit request =>
+    makeContentCardHtml(lookup(path)) map { maybeHtml =>
+      maybeHtml match {
+        case Some(html) => renderContent(html, html)
+        case None => NotFound
+      }
```
jfsoul: this can be simplified to:

    makeContentCardHtml(lookup(path)) map {
      case Some(html) => renderContent(html, html)
      case None => NotFound
    }

onward/app/controllers/RecommendedContentCardController.scala
```diff
@@ -0,0 +1,48 @@
...
+    val fields = "headline,standfirst,shortUrl,webUrl,byline,trailText,liveBloggingNow,commentCloseDate,commentable"
+    lookup(path, fields)(request)
+  }
+
+
+  private def makeContentCardHtml(response: Future[ItemResponse])(implicit request: RequestHeader) = response.map { response =>
+    response.content flatMap { content =>
+      ContentCard.fromApiContent(content) map { contentCard =>
+        contentResponse(contentCard)
+      }
```
jfsoul: I would marginally prefer a for comprehension for this but your choice!

onward/app/controllers/RecommendedContentCardController.scala
```diff
@@ -0,0 +1,48 @@
...
+    }
+  }
+
+  private def contentResponse(content: ContentCard)(implicit request: RequestHeader) = {
+    views.html.fragments.items.facia_cards.contentCard(
+    item = content,
```
jfsoul: indent the params a notch for readability?

onward/app/controllers/RichLinkController.scala
```diff
@@ -1,44 +1,32 @@
...
-    val response: Future[ItemResponse] = contentApiClient.getResponse(
-      contentApiClient.item(path, edition)
-        .showFields("headline,standfirst,shortUrl,webUrl,byline,starRating,trailText,liveBloggingNow")
-        .showTags("all")
-        .showElements("all")
-    )
-
-    response.map { response => response.content.map(Content(_))  }
+  private def lookup(path: String)(implicit request: RequestHeader) = {
+    val fields = "headline,standfirst,shortUrl,webUrl,byline,starRating,trailText,liveBloggingNow"
```
jfsoul: Might be a little ugly because of the length of the string but would it be beneficial to have `fields` in the constructor so we can define `lookup` completely in the superclass?

    ... extends RenderTemplateController(contentApiClient,
          fields = "headline,standfirst,shortUrl,webUrl,byline,trailText,liveBloggingNow,commentCloseDate,commentable")

onward/app/controllers/RenderTemplateController.scala
```diff
@@ -0,0 +1,36 @@
+package controllers
+
+import common.{Edition, ExecutionContexts, JsonComponent, Logging}
+import contentapi.ContentApiClient
+import implicits.Requests
+import model.{ApplicationContext, Cached, NoCache}
+import play.api.mvc._
+import play.twirl.api.HtmlFormat
+
+abstract class RenderTemplateController(contentApiClient: ContentApiClient)(implicit context: ApplicationContext) extends Controller with Paging with Logging with ExecutionContexts with Requests {
```
jfsoul: Not sure the name of this class tells me much about what it is.  How about `OnwardContentCardController`?

jfsoul:
Like the changes a lot! See my minor comments (mostly optional) 😄

onward/app/controllers/RenderTemplateController.scala
lmath: I'm going to take this verbatim, thanks.

onward/app/controllers/RichLinkController.scala
lmath: Hmm, yes, at present that just looks like something repeated that could be in the superclass.

I think I prefer to go the other way and make 'lookup' in the subclasses specific to the subclass. Is that better?

jfsoul: Sure, I think what you have now looks good - although I would maybe rename lookup in the subclasses to something else (I have no good suggestions).  I find overloading in a subclass a bit unclear - it makes it look more like an override that then calls `super.lookup` if you don't look closely.

lmath: Fair point. I'll rename!

jfsoul (approved):
Thanks for listening to all my nitpicks 😄 

Looks good to go 👍

PR #14425, opened 2016-09-22
Hosted autoplay

lps88:
## What does this change?

At the end of playing a hosted video, a countdown timer appears and you are automatically taken to the next video in the series.
## What is the value of this and can you measure success?

We had an AB test that indicated that the autoplay feature improved engagement. See https://github.com/guardian/frontend/pull/13325
## Screenshots

image
image

@Calanthe

PR #13325, opened 2016-06-17
Hosted page - next video autoplay

Calanthe:
# What does this change?

An AB test to check if people will autoplay the next video on the hosted page.
Variant 1:
- 10 seconds before the video is finished, there should be an 'autoplay the next video' box shown with a decrementing counter. When the video is finished, there should be a redirect to the next video page. 
  Variant 2:
- The same as variant 1 + additional box with the direct link to Renault page. 
## Screenshots

Variant 1 on desktop:
screen shot 2016-06-30 at 14 00 31

Variant 1 on mobile:
screen shot 2016-06-30 at 13 58 09

Variant 2 on desktop:
screen shot 2016-06-30 at 13 50 24

Variant 2 on mobile:
screen shot 2016-06-30 at 13 56 42

PR #15400, opened 2016-12-20
Hosted video autoplay

lps88:
## What does this change?
We tested and implemented a next video 'autoplay' component on the first Hosted By campaigns. It still works on the old hardcoded video pages, but needs to be updated to work with newer cAPI-driven ones. In this PR:

- Extract autoplay component into separate file
- Add endpoint for requesting next video in series
- Load the next video data & markup from js (similarly to how we load onward journeys)
- make sure autoplay code is initialised _after_ component is in place
- adjust button styles for campaigns with dark brand colours

NB. Autoplay here is not actually autoplaying the videos, just redirecting to the next video page in the series.

Previous AB tests (using Renault and Leffe pages) indicated that the autoplay feature improved engagement. See:
https://github.com/guardian/frontend/pull/14425
https://github.com/guardian/frontend/pull/13325

No

## Screenshots
image
image
image

CC @guardian/labs-beta


commercial/conf/routes
```diff
@@ -54,6 +54,7 @@ GET         /commercial/anx/anxresize.js                                    comm
 GET         /commercial/advertiser-content/:campaignName/:pageName          commercial.controllers.HostedContentController.renderLegacyHostedPage(campaignName, pageName)
 GET         /advertiser-content/:campaignName/:pageName                     commercial.controllers.HostedContentController.renderHostedPage(campaignName, pageName)
 GET         /advertiser-content/:campaignName/:pageName/:cType/onward.json  commercial.controllers.HostedContentController.renderOnwardComponent(campaignName, pageName, cType)
+GET         /advertiser-content/:campaignName/:pageName/autoplay.json       commercial.controllers.HostedContentController.renderAutoplayComponent(campaignName, pageName)
```
kelvin-chappell: Should be added to https://github.com/guardian/frontend/blob/master/standalone/conf/standalone.routes for preview pages as well.

kelvin-chappell (approved):
Looks good. Just an extra route to add.

PR #15401, opened 2016-12-20
Removes execution context

DiegoVazquezNanini:
## What does this change?
- Removes the use of ExecutionContext.Implicits.global making the code play 2.6 ready
- Moves all the Filters and HttpRequestHandler inside the http package.

@alexduf @TBonnin @jfsoul


common/app/model/Cached.scala
```diff
@@ -146,6 +145,8 @@ object NoCache {
 
 case class NoCache[A](action: Action[A]) extends Action[A] {
 
+  implicit val ec: ExecutionContext = ExecutionContexts.executionContext
```
alexduf: I think the `case class` should receive an implicit execution context rather than having it as an attribute. Chances are that every controller that uses `NoCache` already have an execution context somewhere.

alexduf: As chatted with Diego, this is the only sensible way to do it

TBonnin: I am curious. What is the problem with passing an implicit EC?

TBonnin:
I am not sure I follow: It looks like the entire `play.api.libs.concurrent.Execution` class has been deprecated in 2.6, so we won't be able to use `play.api.libs.concurrent.Execution.Implicits.defaultContext` 
Instead we should inject the EC from the [`ExecutionContextComponent` trait](https://github.com/guardian/frontend/blob/master/common/app/app/FrontendApplicationLoader.scala#L66-L66)

Am I missing something?

common/app/model/Cached.scala
alexduf: If I understand correctly, because it's overriding a method, it can't change the signature

DiegoVazquezNanini:
@TBonnin that's exactly right, I just find out after doing this work and maybe help us deal with the depreciation in a single place. Let's discuss it tomorrow.

TBonnin (approved):
As we said in person, this is not enough to be Play 2.6 compliant ;) but it there are improvements in this PR nonetheless. Thanks

PR #15402, opened 2016-12-20
Remove ItsRainingInlineAds test

regiskuckaertz:
The test has expired, as has the switch. I'm keeping the `-wide` version for now, as many of its features will be ported to the `-regular` insertion algorithm.


janua:
👍

PR #15404, opened 2016-12-20
Removed unused routes

markjamesbutler:
@jfsoul 

*N.B* This is the [second](https://github.com/guardian/frontend/pull/15082) attempt at removing these routes.  #15086 will ensure we do not have the same issues as before.

These routes are now served via [Identity Frontend](https://github.com/guardian/identity-frontend) the routing is done in [Fastly](https://github.com/guardian/identity-platform/blob/master/fastly/profile-theguardian-com.vcl#L29)

I am removing these pages as they are still accessible via (https://profile-old.theguardian.com/signin) thus open to attack.


jfsoul:
I seem to remember you've updated the fastly healthcheck too?

jfsoul (approved):

markjamesbutler:
@jfsoul Yes, see https://github.com/guardian/identity-platform/pull/148

PR #15408, opened 2016-12-20
Use partially applied function to avoid repeating params

joelochlann:
@jfsoul made an interesting point in #15355 about using a higher-order function to avoid some repetition. I slightly misinterpreted him, and while I did manage to avoid repeating the function body, I had to repeat function parameters (and default values) which were simply being passed through to the function which did the main work.

Using a partially applied function (as @jfsoul was suggesting) solves this. It has the interesting effect of breaking the use of named parameters, presumably because named parameters only work for methods, and for partial application Scala converts methods to functions (thanks @davidfurey for that insight)


jfsoul (approved):
Nice thanks! 👍

PR #15409, opened 2016-12-20
Extend email-related feature switches into January

joelochlann:
We're going to be running this test into 2017 so I've extended til the day after I'm back from holiday

@guardian/dotcom-platform


jfsoul (commented):
👍

PR #15411, opened 2016-12-21
Add performance timing logging to Hosted js modules

rich-nguyen:
This alters the commercial bootstrap behaviour for all commercial modules which use custom timing. These modules all share the same pattern; they use custom timing to measure their execution time because the work they do is done asynchronously (usually through the gpt command buffer), and, crucially, this asynchronous work does _not_ need to be waited for in order for execution of dependent modules to begin.

Another way to think of this is that these modules add work to the queue, and the act of adding it to the queue is the only part of the module that must be completed by the bootstrap. The queued work itself is ordered, and safe to execute outside of the bootstrap's promise handling.

So, the promise that encloses this kind of async work only serves to tell the bootstrap that the module has fully completed, and this information is *only useful in the context of hosted modules, whose pages have no MPUs, which causes the performance logging POST to be skipped.

(*only - there are some other situations we could consider in the future. Any page which has no MPUs served through DFP will not be performance logged).


static/src/javascripts/projects/commercial/modules/hosted/video.js
```diff
@@ -60,84 +60,90 @@ define([
...
+            require(['bootstraps/enhanced/media/main'], function() {
+                require(['bootstraps/enhanced/media/video-player'], function (videojs) {
+                    var $videoEl = $('.vjs-hosted__video');
+                    var $inlineVideoEl = $('video');
+                    var $youtubeIframe = $('.js-hosted-youtube-video');
+
+                    if ($youtubeIframe.length === 0 && $videoEl.length === 0) {
+                        if ($inlineVideoEl.length === 0) {
+                            // halt execution
+                            performanceLogging.moduleEnd(moduleName);
```
rich-nguyen: view this change with `?w=1` to see what i've done here.

static/src/javascripts/projects/commercial/modules/hosted/video.js
```diff
@@ -60,84 +60,90 @@ define([
...
+                            resolve();
```
regiskuckaertz: `resolve` does not stop execution, that is why there needs to be a `return` statement.

static/src/javascripts/projects/commercial/modules/hosted/onward.js
```diff
@@ -31,10 +33,12 @@ define([
...
                 })
-                .then(function () {
-                    performanceLogging.moduleEnd(moduleName);
-                });
+        } else {
+            fetchPromise = Promise.resolve();
         }
-        return Promise.resolve();
+
+        return fetchPromise.then(function () {
```
rich-nguyen: yes, hosted gallery and hosted onward both need to return a promise to bootstrap. The promise will only be used to wait until all performance `moduleEnd` calls have been made. At that point, I can report the perf data.

static/src/javascripts/projects/commercial/modules/hosted/video.js
regiskuckaertz: (although I would remove that whole test, `$videoEl.each` and `youtubeIframe.each` will handle the `length === 0` naturally)

static/src/javascripts/projects/commercial/modules/hosted/onward.js
regiskuckaertz: (in fact, the contrary, why _not_ return the `loadCssPromise` in gallery?

regiskuckaertz: oops, just saw your comment :zipper_mouth_face:

regiskuckaertz (approved):
Great! 👍

static/src/javascripts/projects/commercial/modules/hosted/onward.js
rich-nguyen: I do!

static/src/javascripts/projects/commercial/modules/hosted/video.js
rich-nguyen: 👍

static/src/javascripts/projects/commercial/modules/hosted/video.js
```diff
@@ -60,84 +60,90 @@ define([
...
-                            }
-                        }
+                    $youtubeIframe.each(function(el){
+                        hostedYoutube.init(el);
                     });
-                });
 
-                $youtubeIframe.each(function(el){
-                    hostedYoutube.init(el);
+                    performanceLogging.moduleEnd(moduleName);
```
rich-nguyen: i think that would be tidier, yes

rich-nguyen:
[Improve commercial loading order](https://trello.com/c/E8ojiTwb/260-improve-commercial-loading-order)

PR #15416, opened 2016-12-21
Next Hosted Video Autoplay - fix the test clean up

lps88:
## What does this change?

Fix the test clean up so there's no side effects.

@regiskuckaertz


static/test/javascripts/spec/commercial/hosted/next-video-autoplay.spec.js
```diff
@@ -78,13 +85,8 @@ define([
...
 
         it('should not trigger autoplay when there is no next video', function (done) {
-            fixturesConfig = {
-                id: 'next-video-autoplay',
-                fixtures: [
-                    '<div class="js-autoplay-timer" data-next-page="">10s</div>'
-                ]
-            };
-            $fixturesContainer = fixtures.render(fixturesConfig);
+            fixtures.clean(fixturesConfig.id);
```
regiskuckaertz: You don't need that line.

regiskuckaertz (approved):
Sweet

static/test/javascripts/spec/commercial/hosted/next-video-autoplay.spec.js
lps88: I need to remove the html used in the other tests (which includes '<video>' tag) otherwise the test fails! Is it cleaner to put this into a separate `describe`? (I tried this but had to duplicate most of the set up)

PR #15418, opened 2016-12-21
Add user agent to AMP google analytics

gtrufitt:
## What does this change?

Google have fixed a bug with `<amp-analytics>` which means we can now pass through the user agent as we do on platform.

The plain user-agent string is needed for our ABC audit.

@guardian/dotcom-platform


dominickendrick:
👍

rich-nguyen (approved):

PR #15421, opened 2016-12-21
Removing scalaz from frontend codebase

TBonnin:
Copy of the email I sent around a couple of days ago
```
FP libraries like cats or scalaz can be extremely useful, however they have 
a pretty steep learning curve. 
It makes totally sense to use cats or scalaz in a project with a relatively small 
scope, small team and to make use of it throughout the whole codebase.
However frontend is the opposite of this:
- It has a very large scope. In theory nothing in frontend is very complex by itself. 
The codebase is complex mostly because of the huge number of features. 
- The number of people touching the codebase is quite big (~80 in 2016 based 
on my non-scientific research ;) )
- scalaz is not used to a great extend. It is currently used in only 10 files (mostly 
for iterating with states over a collection, which could be done without too much 
difficulties in standard scala I believe)

Given the steep learning curve and the loss of efficiency when someone touching 
the code needs to learn and understand news concepts brought by those 
libraries I would lean toward removing scalaz (or any FP lib) from the frontend 
codebase. 
```

The consensus coming out of this discussion was: _Scalaz: No, Cats: great library but probably not a wise choice for the frontend codebase._

Less complexity 🔪 

Note: I have only refactored to remove scalaz but some of the classes I touched could do with some more love. Very looooong methods, lots of tuples of val, etc...
I didn't want to cram too much refactoring in the same PR though

cc @guardian/dotcom-platform


common/app/crosswords/CrosswordTreat.scala
```diff
@@ -45,9 +41,6 @@ object CrosswordGrid {
     (5, 6),
     (6, 6)
   ).map((Position.apply _).tupled))
-
-  def fromCrossword(crossword: Crossword): CrosswordGrid =
-    CrosswordGrid(crossword.entries.toList.foldMap[Set[Position]](_.allPositions.toSet))
```
TBonnin: This method is never called

common/app/layout/Column.scala
```diff
@@ -63,20 +61,35 @@ case class HtmlAndClasses(index: Int, html: Html, classes: Seq[String])
...
-            index,
-            HtmlBlob(html, classes, Column.cardStyle(column, positionInColumn).getOrElse(ItemClasses.showMore)),
-            None
-          )
-      }))
-    }._2
-
-    SliceWithCards(layout.cssClassName, columns)
+
+    @tailrec
```
TBonnin: `mapAccumL` can be easily replaced by simple recursion

dominickendrick:
👋 scalaz

common/app/layout/Column.scala
```diff
@@ -63,20 +61,35 @@ case class HtmlAndClasses(index: Int, html: Html, classes: Seq[String])
...
+    def columnsWithCards(columns: List[Column],
+                         items: Seq[HtmlAndClasses],
+                         accumulation: List[ColumnAndCards] = List.empty
```
janua: Probably want to switch this to a `Vector` because here you have to continually append to the end of the `List` which will be slower than it will be with a `Vector`.

onward/app/weather/controllers/LocationsController.scala
```diff
@@ -38,12 +37,8 @@ class LocationsController(weatherApi: WeatherApi) extends Controller with Execut
...
     log.info(s"What is my city request with headers $maybeCity $maybeRegion $maybeCountry")
 
-    tuple3(
-      maybeCity,
-      maybeRegion,
-      maybeCountry
-    ) match {
-      case Some((city, region, country)) =>
+    (maybeCity, maybeRegion, maybeCountry) match {
+      case (Some(city), Some(region), Some(country)) =>
```
janua: Scalaz... for this? 😆

onward/app/weather/models/CityResponse.scala
```diff
@@ -5,17 +5,17 @@ import common.editions.{Au, Us, Uk}
...
-import scalaz.syntax.foldable._
-
 object CityResponse {
   implicit val jsonWrites = Json.writes[CityResponse]
 
   def fromLocationResponses(locations: List[LocationResponse]) = {
     def cityAndCountry(location: LocationResponse) =
       (location.LocalizedName, location.Country.LocalizedName)
 
-    val citiesWithSameNameByCountry = locations.map({ location => Map(cityAndCountry(location) -> 1) }).suml
```
janua: This is one of those things that is really useful from the libraries; the `Monoid` instance for `Map`.
Nice workaround, I thought it might not be safe to remove until I seen the sneaky `getOrElse` in the return of the fold! 😄

janua:
Looks great! @robertberry will not be happy.
👍

JustinPinner:
:1st_place_medal:

rich-nguyen (approved):
super 👍

PR #15417, opened 2016-12-21
Refactor engagement banner and tests

rtyley:
## What does this change?

Refactors the Membership/Contributions engagement banner code to make to make the expression of the A/B tests more concise and clear.

## What is the value of this and can you measure success?

Hopefully this should lead to smaller diffs as we add and remove engagement banner tests, allowing us to iterate our tests more rapidly.

## Does this affect other platforms - Amp, Apps, etc?

No only web.

## Screenshots

image

This is kind of a rehash of PR #14244, which had a similar refactoring.

cc @JustinPinner

PR #15422, opened 2016-12-21
Extend uk membership engagement test switch again

JustinPinner:
## What does this change?
@rtyley has some changes coming through very soon that will remove this test and switch. But if the switch expires before that's out, builds will break - so I extended it by another couple of weeks.

Don't break builds with expired switches.

No.

## Screenshots
N/A


gtrufitt (approved):

JustinPinner:
Not needed anymore. https://github.com/guardian/frontend/pull/15417/files eliminates the switch.

PR #15423, opened 2016-12-21
Use same campaignId for both EOY tests

desbo:
These two files actually represent one test, split this way so we can have different audience sizes. Using the same campaign ID means we can manage the tags from one place in the targeting tool.

@alexduf


alexduf (approved):

Mullefa:
Appreciate this facilitates using the targeting tool, but it also makes analysis harder. Can we not change the tags for an ongoing test in future unless it is really necessary? And I'll do my best to use sensible tag names 😉

PR #15424, opened 2016-12-21
DEVINFRA test configuration to allow local overrides

MatthewJWalls:
## What does this change?

@TBonnin @jfsoul 

Ok,

So the preferred way of mocking out CAPI (and other services) we talked about last week is to run a little app that pretends to be CAPI as a separate process, and just configure the frontend to use http://localhost:NNNN as the capi endpoint.

There are two ways the frontend can get config currently:

1. Use DEVINFRA.properties directly if we are in stage=DEVINFRA mode, or
2. Grab the config off of S3 + apply local overrides (This will fail if S3 isn't accessible)

I still prefer not to need to screw around with S3 if possible since I'm trying to be location agnostic. So that gives me three options to achieve this:

1. Create a whole new install vars stage like PRBUILDS that loads config *only* from ~/.gu/, and use *that* instead of DEVINFRA.
2. Modify DEVINFRA.properties file in this repo to point to my preferred endpoints (I don't like because it creates an invisible dependency between my mock services and this properties file)
3. Modify config loaded from DEVINFRA.properties to be overrideable with the normal ~/.gu/frontend.conf file and I can put my capi endpoints as an override.

I've implemented number 3 under this PR since it's the smallest possible change I can make. I'm happy with going with one of the other two options if it's preferred.

Success will give frontend a way to override service locations at runtime in the absence of s3.

No

### DIFF
--- a/common/app/common/configuration.scala
+++ b/common/app/common/configuration.scala
@@ -77,7 +77,7 @@ object GuardianConfiguration extends Logging {
 
     // test mode is self contained and won't need to use anything secret
     lazy val test = ClassPathConfigurationSource("env/DEVINFRA.properties")
-    lazy val testConfig = new CM(List(test), PlayDefaultLogger).load.resolve
+    lazy val testConfig = new CM(List(userPrivate, test), PlayDefaultLogger).load.resolve
 
     val appConfig =
       if (stage == "DEVINFRA") testConfig


jfsoul:
Why would option 1 need to load config from `~/.gu` - couldn't it just have a properties file just as easily?

Option 3 certainly wins for minimal change, but I think I like option 1 because it makes everything more explicit.  What do you think?

MatthewJWalls:
@jfsoul Yeah basically it can be anything. Another file in this repo, something somewhere on your local pc, etc. I'd tend towards not putting it in the repo since I'd expect it change based on use case.

jfsoul:
Not sure I have thought about this enough to know what the use cases might be!

rich-nguyen:
What's the aim of mocking out CAPI?

TBonnin:
You actually don't want to use DEVINFRA config: CI is setup by design to not access the network and has most of the config value are nulled
You should be able to run in PROD mode (to be as close as possible from production setup) with a local override at `/.gu/frontend.conf`

MatthewJWalls:
To clarify again,

For my *specific* use case, I'm doing this because I want to run the app sandboxed and do runtime regression testing on it without having to a) screw around with making sure it has credentials for aws services b) worry about data changing underneath us, and c) ensure that any tests running against the app aren't influenced by upstream issues.

The hoped outcome is we want to get in place a system that runs pull requests, grabs Gareth's screenshots utility, runs Mike's 3rd party services like the page costing thing he's working on, and paste the results on the PR, and then see if that turns out to be valuable or not. If it's not, we'll kill it.

In *general* though I think it's a good idea because right now, a contributor outside of the guardian can't run the app, you can't do an experimental fork that lives elsewhere, you can't work on frontend without a stable internet connection, you have to constantly get janus credentials, etc. Yeah we don't actually care about solving that stuff right now, but for the low price of 1 line of code to provide a way to supply runtime configuration at runtime, you can at least support a way for people to start looking at those things. Also as a point of philosophy, having an app that has no way to work if it's not in our specific infrastructure sounds more like a design issue than a virtue to me.

rich-nguyen:
That pull request system sounds like a decent idea to try, well worth building to see how it beds in. However, it doesn't seem to follow that a sandboxed app is a requirement for such a system.

> an app that has no way to work if it's not in our specific infrastructure sounds more like a design issue

You're absolutely right, this isn't a desirable virtue. But to assume the reverse implication is true, that a sandboxable app therefore is a virtue, is a risk, without valid support.

- I don't see outside contributors as an issue, we develop in the open, that doesn't mean we can afford the dev time and effort to support outside contribution, unless real benefit can be evidenced. 
- You can't work on frontend without a stable connection. That is inconvenient, yes, but those that WFH should have a stable connection, the office wired network is adequate, and any other scenario constitutes outside of hours work, which is an entirely different conversation.
- You have to get janus credentials. That is a security choice we made, it isn't a technical issue. If that's annoying, grab your own credentials, and pop them in your config. And tell Adam why (I have done this several times).
- you can't do an experimental fork. I believe you can, but do let me know if you cannot, and we'll see if we can improve this.

In a non-philosophical view, we have tied in infrastructure assumptions in many places, because it's a live, production codebase where delivery has taken precedent. We may have gone too far (we can't deploy easily without RiffRaff, and we can't live outside of AWS, or Fastly), but I think we need to balance the dev cost with the benefit to the platform.

PR #15425, opened 2016-12-22
Remove per-app deploys

TBonnin:
## What does this change?
We have been using unified deploy for more than 10 days without much
issue.
This patch is removing the old way of deploying artifacts

Cleaning up

cc @guardian/dotcom-platform


dev/teamcity/dist-publish-assets-tc
```diff
@@ -32,20 +32,6 @@ if [[ -z $RIFF_RAFF_BUILD_BUCKET ]]; then
...
-    -e "s|<%start_time%>|$BUILD_START_DATE|" \
-    -e "s|<%revision%>|$BUILD_VCS_NUMBER|" \
-    -e "s|<%branch%>|$BUILD_VCS_BRANCH|" \
-    static/build-template.json | tee $static_folder/build.json
-
-aws s3 cp --acl bucket-owner-full-control --region=eu-west-1 --recursive $static_folder s3://$RIFF_RAFF_ARTIFACT_BUCKET/dotcom:static/$BUILD_NUMBER
-aws s3api put-object --acl bucket-owner-full-control --region=eu-west-1 --bucket $RIFF_RAFF_BUILD_BUCKET --key dotcom:static/$BUILD_NUMBER/build.json  --body $static_folder/build.json
-
 # upload static files for the build - it is critical that this is done before the main sbt-riffraff-artifact plugin runs
 aws s3 cp --acl bucket-owner-full-control --region=eu-west-1 --recursive $static_folder/packages/frontend-static s3://$RIFF_RAFF_ARTIFACT_BUCKET/dotcom:all/$BUILD_NUMBER/frontend-static
```
TBonnin: I wonder why we are not uploading static artifact using sbt. The same we do for everything else. That would allow us to get rid of this bash file. Any reason?

rich-nguyen: I think that's because there is no build dependency between the asset build and sbt, so we shouldn't make one artificially just to use an uploader. Ideally, the asset build should be separate from the server build (we just have to rethink asset maps), and then devs could build css and js and deploy static without a server deploy. That would be super fast.

rich-nguyen (approved):

dev/teamcity/dist-publish-assets-tc
TBonnin: `static` is a project like all the other app. Having the artifact definition in scala/sbt would make it easier to refactor. I live bash but touching it is quite scary.
That wouldn't change the fact that `static` is a simple bundle of files that only need to be uploaded to S3. This is already the case right now

DiegoVazquezNanini (approved):
:+1: So much simpler. Thanks @TBonnin

dev/teamcity/dist-publish-assets-tc
sihil: It's not currently a project in SBT from what I can see. We could make it one and give it an appropriate base directory and add the files to the upload.

If this script doesn't run for some reason then the artifact will probably still be 'valid' but not have any static artifacts. I'm not sure putting it into SBT would improve this though.

sihil (commented):
Nice job: -458!

Looks good to me.

PR #15426, opened 2016-12-22
Fix dev build

davidfurey:
## What does this change?

Removes an import to fix the dev build.

no


jfsoul (approved):
Thanks @davidfurey

PR #15428, opened 2016-12-22
Check for seriesName rather than webTitle when determining shouldShowAds

katebee:
## What does this change?

Checking for `webTitle` adds waaaaay too much specificity to the `shouldShowAds` check

We want the same logic to apply to all email sign up pages ([example](https://www.theguardian.com/info/ng-interactive/2016/dec/07/sign-up-for-the-flyer) and [example](https://www.theguardian.com/info/ng-interactive/2016/dec/05/sign-up-for-weekend-reading)) without adding a new case for each.

All subsequent email sign up landing pages behave the same as: https://www.theguardian.com/info/ng-interactive/2016/dec/07/sign-up-for-the-flyer

Nope

-----------

@lmath @joelochlann


lmath:
Nice! :+1:

PR #15430, opened 2016-12-22
Log missing content required to build a hosted trail

kelvin-chappell:
Just logging problem with capi responses before fixing it.


lps88 (approved):
Looks good

PR #15431, opened 2016-12-22
Fixing onward test suite

NataliaLKB:
## What does this change?
Turns out our Onward test suite hasn't been running. We are not sure exactly when they stopped, but this should ensure they run again locally and in teamcity.

Our tests test!

Not particularly

@TBonnin @guardian/dotcom-platform


TBonnin (approved):
Thank you @NataliaLKB

PR #15432, opened 2016-12-22
Update GU_ID_CSRF Play cookies to be HttpOnly

markjamesbutler:
This option is set to false in the default config.  Setting to HttpOnly to prevent JS sniffing.


katebee (approved):
👍 
and another 👍 for giving me a rabbithole on cookies 🍪

PR #15435, opened 2016-12-23
Fix confirmation message for logged in users on Newsletters page

katebee:
## What does this change?

Fix for this page: https://www.theguardian.com/email-newsletters

Signed in users are not getting the confirmation/success message upon form submission. This fixes the `subscribeToEmail` function so that the success message appears for both anon and signed in users.

#### Bonus: 
`updateFormForLoggedIn` did much more than update the form; it has therefore been renamed `updatePageForLoggedIn` 

Removing the `setAttribute` and subsequent `preventDefault`; the button does not need to be a `submit` type when we are just going to prevent the `submit` anyway...

🎄 🎅🏼 ☃️

- Better UX
- Tells users that their action worked, rather than just leaving them guessing

Nope

## Screenshots

After fix:

picture 419


NataliaLKB (approved):
Looks good!

PR #15436, opened 2016-12-23
Fix outbrian reporting bug by include geolocation checks in canRun for contributions AB tests

desbo:
## What does this change?

There was a problem where we were marking pages as non-outbrain-compliant even when they were indeed compliant. This was because the compliancy relies on the canRun function in the AB tests, but we were checking the geolocation outside of this (due to the fact that it was an asynchronous call). In order to put the geo checking in the canRun, we had to make it synchronous. 

We've added the geolocation ajax call to main.js (enhanced) which, if there is no valid geolocation in the local storage, makes the call and sets the geolocation with a 10 day expiry. When a page loads, it checks local storage for a valid geolocation. If none exists, it uses the edition as a proxy (whilst the call in main.js is updating it with the correct geolocation in the background, ready for the next pageview). 

This PR also removes some more of the rendering boilerplate from the contributions tests by providing a default test function. 

We can accurately report our outbrain usage, meaning we will get paid properly. Also, less boilerplate = easier peer reviewing of future epic tests.

no

### DIFF
--- a/static/src/javascripts/bootstraps/enhanced/main.js
+++ b/static/src/javascripts/bootstraps/enhanced/main.js
@@ -11,7 +11,8 @@ define([
     'common/modules/experiments/ab',
     './common',
     './sport',
-    'common/modules/analytics/google'
+    'common/modules/analytics/google',
+    'common/utils/geolocation'
 ], function (
     fastdom,
     bean,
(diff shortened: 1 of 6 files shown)


gtrufitt (approved):
Lgtm, nice one!

PR #15441, opened 2016-12-28
Prevent logback from outputting its own status when tests start

TBonnin:
## What does this change?
It looks like, when first running tests in Play 2.5, Logback cannot properly read the `logback.xml` configuration file. It results in Logback using the default BasicConfiguration and outputting everything to the console until Play takes over an set proper logging configuration.
Maybe this issue is caused by multiple `logback.xml` files in the classpath

This patch introduces a workaround that rename the `logback.xml` file and specify the default Logback configuration file as a `java` parameter when launching `sbt`.

I am not sure this is the most elegant solution but it has the merit to work 💃 

No more unnecessary logging in tests output

@alexduf @DiegoVazquezNanini


jfsoul (approved):

DiegoVazquezNanini (approved):
:+1:

TBonnin:
This has been reverted https://github.com/guardian/frontend/pull/15464 because it causing problem with `dev-build` 😢

PR #15442, opened 2016-12-28
Update canRun check of AB test, and make it more resilient

katebee:
## What does this change?

Previous `canRun` check was relying on the headline entered in Composer - when the headline was changed it stopped the test running.

The new check is based on the page URL, which is much less likely to change.

Fixes a test and makes it harder to break in the future.

Nope


GHaberis:
@katebee 👍

PR #15443, opened 2016-12-29
Extend RecommendedForYou and scale the images

katebee:
## What does this change?

- Extends the JavaScript expiry date of the test
- Makes the images of the recommendations facia-cards fit correctly on smaller displays

- Less red on the switches dashboard
- Recommendations facia-cards have an image overflow issue at narrow (tablet/mobile) resolutions

Nope

##### Before:

picture 420

##### After:

picture 425

---------
@lmath


lmath:
:+1:

PR #15444, opened 2016-12-29
end of year variant removed for US eoy campaign

Mullefa:
## What does this change?

Removes the End Of Year epic variant from the US end of year campaign.

The End Of Year variant's conversion rate was ~50% that of the controls. Removing the End Of Year variant will increase total contributions.

No


katebee (approved):
LGTM 👍

PR #15446, opened 2016-12-29
Enable users to update username on first comment

mariogalic:
## What does this change?

This allow users to set their username the first time they post a comment, that is, during discussion on-boarding process.

Users that register via social sign in (Google, Facebook) have their `displayNames` automatically set to their real names, thus when they comment they are unaware that the comment will be posted under their real name. Currently there is no self-service mechanism to allow users to change their username. For this reason they often contact Userhelp to request account deletion. 

Also, `username/displayName` has been removed from all registration forms to improve conversion, since usernames are used only for commenting, so there is no need to have this extra field during Membership or Subscription registration flows:
* https://github.com/guardian/identity-frontend/pull/187
* https://github.com/guardian/identity-frontend/pull/195

https://github.com/guardian/identity/pull/604

## Screenshots

image

Error path:

image

@markjamesbutler @thomaska @andrewfindlay

### DIFF
--- a/common/app/views/fragments/commentBox.scala.html
+++ b/common/app/views/fragments/commentBox.scala.html
@@ -27,7 +27,9 @@
 
         <div class="d-comment-box__onboarding-username">
             <p>Please choose your username under which you would like all your comments to show up:</p>
-            <input type="text" class="d-comment-box__onboarding-username-input" placeholder="Username: 6-20 characters, letters & numbers only, no spaces" value=""/>
+            <label class="d-comment-box__onboarding-username-label" for="usernameInput">Username: </label>
+            <span>Must have 6-20 characters, letters & numbers only, no spaces.</span>
+            <input type="text" id="usernameInput" class="d-comment-box__onboarding-username-input" placeholder="" value=""/>
             <p class="d-comment-box__onboarding-username-error-message is-hidden"></p>
         </div>
 
--- a/common/app/views/fragments/commentBox.scala.html
+++ b/common/app/views/fragments/commentBox.scala.html
@@ -24,6 +24,13 @@
     <div class="d-comment-box__onboarding">
         <h3>Welcome <span class="d-comment-box__onboarding-author"></span>, you’re about to make your first comment!</h3>
         <p>Before you post, we’d like to thank you for joining the debate - we’re glad you’ve chosen to participate and we value your opinions and experiences.</p>
+
+        <div class="d-comment-box__onboarding-username">
+            <p>Please choose your username under which you would like all your comments to show up:</p>
+            <input type="text" class="d-comment-box__onboarding-username-input" placeholder="Username: 6-20 characters, letters & numbers only, no spaces" value=""/>
+            <p class="d-comment-box__onboarding-username-error-message is-hidden"></p>
+        </div>
+
         <p>Please keep your posts respectful and abide by the <a target="_blank" href="/community-standards">community guidelines</a> - and if you spot a comment you think doesn’t adhere to the guidelines, please use the ‘Report’ link next to it to let us know.</p>
         <p>Please preview your comment below and click ‘post’ when you’re happy with it.</p>
         <div class="d-comment-box__onboarding-preview-body"></div>
--- a/common/app/views/fragments/commentBox.scala.html
+++ b/common/app/views/fragments/commentBox.scala.html
@@ -28,7 +28,7 @@
         <div class="d-comment-box__onboarding-username">
             <p>Please choose your username under which you would like all your comments to show up:</p>
             <label class="d-comment-box__onboarding-username-label" for="usernameInput">Username: </label>
-            <span>Must have 6-20 characters, letters & numbers only, no spaces.</span>
+            <span>Must be 6-20 characters, letters and/or numbers only, no spaces.</span>
             <input type="text" id="usernameInput" class="d-comment-box__onboarding-username-input" placeholder="" value=""/>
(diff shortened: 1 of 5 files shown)


paperboyo:
This is great! Just a question: some usernames contain spaces in the current system (e.g. https://profile.theguardian.com/user/id/16400749). How’s that compatible with the screenshot where the spaces are said to not be allowed?

andrewfindlay:
@paperboyo I can't answer many questions on Github, I can answer this one.

Usernames cannot contain spaces, display names can. At the moment, anyone who signs up through social sign in gets given a display name based on their Facebook/Google name which doesn't have the same restrictions as a username. If they go on to comment, their display name is used. Once this change goes through, anyone with a display name will have to change it to a username when they first comment.

paperboyo:
So no space in comment names for newly signed up users. Clear, @andrewfindlay, thanks!

common/app/views/fragments/commentBox.scala.html
```diff
@@ -24,6 +24,13 @@
     <div class="d-comment-box__onboarding">
         <h3>Welcome <span class="d-comment-box__onboarding-author"></span>, you’re about to make your first comment!</h3>
         <p>Before you post, we’d like to thank you for joining the debate - we’re glad you’ve chosen to participate and we value your opinions and experiences.</p>
+
+        <div class="d-comment-box__onboarding-username">
+            <p>Please choose your username under which you would like all your comments to show up:</p>
+            <input type="text" class="d-comment-box__onboarding-username-input" placeholder="Username: 6-20 characters, letters & numbers only, no spaces" value=""/>
```
katebee: The username requirements seem like key information; it may be better to add them as a label outside of the input area? 

Not all screen readers read placeholder text aloud:
https://www.nngroup.com/articles/form-design-placeholders/

mario-galic:
@katebee I've added a label and moved placeholder text out of the field to improve accessibility:

image

andrewfindlay:
@mario-galic I'd suggest two slight changes to the text, assuming there is space. Instead of:

Username: Must have 6-20 characters, letters & numbers only, no spaces.

I would suggest:

Username: Must be 6-20 characters, letters and/or numbers only, no spaces.

We do get the occasional user who thinks that they have to use numbers in their username when it says letters and numbers.

static/src/stylesheets/module/_discussion.scss
```diff
@@ -1530,3 +1530,34 @@ $comment-recommend-button-size: 19px;
...
+    margin-bottom: $gs-baseline;
+
+    &:focus {
+        border-color: $neutral-2;
+        outline: none;
+    }
+}
+
+.d-comment-box__onboarding-username-label {
+    font-size: 14px;
```
katebee: maybe: `@include fs-textSans(3);` instead?

https://github.com/guardian/guss

static/src/stylesheets/module/_discussion.scss
```diff
@@ -1530,3 +1530,34 @@ $comment-recommend-button-size: 19px;
...
+    display: block;
+    background-color: transparent;
+    border: 1px solid $neutral-4;
+    color: $neutral-1;
+    padding: 7px 15px;
+    border-radius: 18px;
+    margin: 0.2rem 0;
+    width: 70%;
+    margin-bottom: $gs-baseline;
+
```
katebee: Could values from the grid be used instead in padding and margin?

e.g. `padding: $gs-baseline/2 $gs-gutter;`

https://github.com/guardian/guss-grid-system

katebee:
(Someone with an eye for design please jump in and stop me...) Would it be beneficial to visually separate the user comment from the surrounding text? 

For example, adding background colour to `d-comment-box__onboarding-preview-body`:

picture 432

katebee:
If you want to make the error messages match, `identity/conf/messages.en` line 8 can be updated:

`error.usernameLength=Please enter a username between 6 and 20 characters`

static/src/javascripts/projects/common/modules/discussion/comment-box.js
```diff
@@ -517,6 +551,29 @@ CommentBox.prototype.formatComment = function(formatStyle) {
...
+ * temporarily updates HTML client-side so that user has instant feedback before the page is refreshed.
+*/
+CommentBox.prototype.refreshUsernameHtml = function() {
+    IdentityApi.reset();
+    var displayName = this.getUserData().displayName;
+    var userId = this.getUserData().id;
+
+    var menuHeaderUsername = document.querySelector(".js-profile-info");
+    var discussionHeaderUsername = document.querySelector("._author_tywwu_16");
+    var firstCommentUsername = document.querySelector('[data-comment-author-id=' + '"' + userId + '"]');
```
katebee: - Single quotes on lines 568 and 569
- Qwery (`$`) could be used rather than `document.querySelector` (and it is already being included in the AMD):

```
var menuHeaderUsername = $('.js-profile-info')[0];
var discussionHeaderUsername = $('._author_tywwu_16')[0];
var firstCommentUsername = $('[data-comment-author-id=' + '"' + userId + '"]')[0];
```
...Additionally, my `discussionHeaderUsername` is updating, but not the `firstCommentUsername`This seems to be because `firstCommentUsername = Null`. I assume this is just an issue when running locally?

katebee:
picture 436

katebee (approved):
LGTM - fix the failing test and it is GTG 💎 👍

crifmulholland:
👍  Nice work guys

PR #15447, opened 2016-12-29
us spelling of favour (favor)

Mullefa:
## What does this change?

For the US epic, uses the American spelling for favour (favor).

Improves the locality of the Guardian, although it's (probably) hard to measure the impact of this.

No


NataliaLKB (approved):
👍

PR #15449, opened 2016-12-30
Removed unnecessary css from header test

zeftilldeath:
A cheeky few lines of CSS remained from PBM (Pre better meta). I've chucked them in the bin alongside the turkey carcass and the wrapping paper*

That was happening. It doesn't now. 
screen shot 2016-12-30 at 11 43 05

@NataliaLKB Hello 👋 

* Sorry I should have recycled


NataliaLKB:
Yes you should have recycled 😞 

... 👍

PR #15454, opened 2017-01-03
Fix hosted logo link

kelvin-chappell:
This fixes a bug that has been raised on a new hosted campaign, where the link on hosted logos was wrong because they were always using the CTA link.

Eg. in https://www.theguardian.com/advertiser-content/anchor-award/anchor-reeperbahn-festival-international-music-award
logo link should be http://www.anchor-award.com/home.html
rather than https://youtu.be/Z8I1dKHNnOs

/cc @guardian/labs-beta


lps88 (approved):
Looks good, do we still have/want the behaviour that the logo should use the CTA link if it doesn't have its own set? I am remembering very old conversations around this...

kelvin-chappell:
The logo link is required in tag manager, but I'll test some of the older campaigns to check that it works as expected.

PR #15456, opened 2017-01-03
Fix the fetching of commercial feeds (again)

JonNorman:
## Background
Commercial feeds are currently [failing to be fetched](https://kibana.gu-web.net/goto/e9c9aa9a48b47e67efc7c5d9a517ff5c) with the following error:

`org.asynchttpclient.netty.NettyResponse cannot be cast to com.ning.http.client.Response`

This error arose during the migration to Play 2.5 but ultimately is a consequence of us casting the `WsResponse` to its underlying response, which is dumb, and we don't need to do this.

## What's changed?
This PR replaces the current casting method by simply getting the response body as an `akka.util.ByteString`, which we can use to simply decode the body using the encoding specified for each feed.

People can get current merchandise in their components.

No.

### DIFF
--- a/commercial/app/model/feeds/FeedReader.scala
+++ b/commercial/app/model/feeds/FeedReader.scala
@@ -1,10 +1,8 @@
 package commercial.model.feeds
 
-import java.nio.charset.Charset
 import commercial.CommercialMetrics
 import common.Logging
 import conf.switches.Switch
-import org.asynchttpclient.netty.NettyResponse
 import play.api.libs.json.{JsValue, Json}
 import play.api.libs.ws.{WSClient, WSRequest, WSResponse, WSSignatureCalculator}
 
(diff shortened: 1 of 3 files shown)


JonNorman:
[Commercial feed fetching is failing.](https://trello.com/c/4FWEOOIx/269-commercial-feed-fetching-is-failing)

commercial/app/model/feeds/FeedReader.scala
```diff
@@ -119,13 +115,12 @@ class FeedReader(wsClient: WSClient) extends Logging {
...
 }
 
-
 case class FeedRequest(feedName: String,
                        switch: conf.switches.Switch,
                        url: String,
                        parameters: Map[String, String] = Map.empty,
                        timeout: Duration = 2.seconds,
-                       responseEncoding: Option[String] = None)
+                       responseEncoding: String)
```
JonNorman: `BookFinder` is the only module that uses the `FeedRequest` and `FeedReader` classes - it can probably be cleaned up at some point but it is beyond the scope of this PR.

rich-nguyen:
👍

rich-nguyen (approved):

PR #15457, opened 2017-01-03
Better recommended for you cards

lmath:
## What does this change?
Populates recommended for you cards with the html fetched from the server instead of using the client side template that didn't have information about e.g. tone. 

Also adds opt-in prompt before recommendations are loaded.

Cards in the recommended for you container will be more consistent with how cards appear elsewhere. 

No. 

## Screenshots

opt-in prompt:

screen shot 2017-01-04 at 15 55 31

screen shot 2017-01-04 at 10 11 21

cards in recommended for you container with html from contentcard template:
image

screen shot 2017-01-04 at 10 13 29

cards in recommended for you container with html from client side template:
image


static/src/javascripts/projects/common/modules/experiments/tests/recommended-for-you.js
```diff
@@ -76,13 +76,34 @@ define([
...
         }
 
-        function getRemoteRecommendations() {
+        function getCardsHtml(items) {
+            return Promise.all(items.map(getCardHtml));
+        }
+
+        function getCardHtml(id) {
+            var endpoint = '/embed/contentcard/' + id + '.json';
+            var request = fetch(endpoint, {
```
regiskuckaertz: There's a convenience [`fetchJson`](https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/common/utils/fetch-json.js) module to handle the JSON stuff for you.

davidfurey: Great, thanks!

lmath:
@davidfurey Your changes today look good and I tried running them too.

static/src/javascripts/projects/common/modules/experiments/tests/recommended-for-you.js
```diff
@@ -112,45 +140,71 @@ define([
...
+        function insertOnBoardingSection() {
+            $recommendedForYouSection = $.create(template(recommendedForYouOptInTemplate, {
+                profileIcon: svg(profileIcon, ['rounded-icon', 'rfy-profile-icon', 'control__icon-wrapper']),
+                rightArrowIcon: svg(rightArrowIcon, ['i-right']),
+                guardianLogo: svg(guardianLogo)
+            }));
+
             return fastdom.write(function() {
                 $recommendedForYouSection.insertBefore($opinionSection);
+                $('.js-feedback-button-yes', $recommendedForYouSection[0]).each(function(el) {
```
katebee: `$recommendedForYouSection` is an Object therefore it should not need the index reference:

`$('.js-feedback-button-yes', $recommendedForYouSection).each(`

static/src/javascripts/projects/common/modules/experiments/tests/recommended-for-you.js
```diff
@@ -112,45 +140,71 @@ define([
...
+                    bean.on(el, 'click', function () {
+                        registerFeedback(true);
+                        $('.js-feedback', $recommendedForYouSection[0]).html(
```
katebee: `populateRecommendationsContainer()` immediately overwrites `.js-feedback` so this message never becomes visible to the user:

picture 438


Could lines 194 > 196 be deleted? Or does `populateRecommendationsContainer()` need to be moved?

static/src/javascripts/projects/common/modules/experiments/tests/recommended-for-you.js
```diff
@@ -112,45 +140,71 @@ define([
...
+                            '<p>Your recommendations will be ready soon.</p>'
+                        );
+                        populateRecommendationsContainer();
+                    });
+                });
+
+                $('.js-feedback-button-no', $recommendedForYouSection[0]).each(function(el) {
+                    bean.on(el, 'click', function () {
```
katebee: If we assume there should never be more than one `js-feedback-button-no`, we could remove the `.each(`:

```
bean.on($('.js-feedback-button-no', $recommendedForYouSection)[0], 'click', function () {
    registerFeedback(false);
    $recommendedForYouSection.remove();
});
```

(and the same for `.js-feedback-button-yes`)

static/src/javascripts/projects/common/modules/experiments/tests/recommended-for-you.js
davidfurey: populateRecommendationsContainer() is async so might be slow or fail, this message gives the user feedback on their action immediately.

static/src/javascripts/projects/common/modules/experiments/tests/recommended-for-you.js
davidfurey: Thanks.  There used to be two `js-feedback-button-no`, but there isn't anymore so I'll change it to this shorter format.

static/src/javascripts/projects/common/modules/experiments/tests/recommended-for-you.js
davidfurey: I may be completely confused, but I think that $recommendedForYouSection is a bonzo object, and $recommendedForYouSection[0] is the raw dom element?

katebee: Sorry 🙃 

You are correct, not confused 😝 Ignore this!

Interesting that it works as the root/context for $ without the index...

static/src/javascripts/projects/common/modules/experiments/tests/recommended-for-you.js
```diff
@@ -112,45 +140,67 @@ define([
...
+        function shouldShowRecommendations() {
+            return !!storage.local.get(cachedRecommendationsEnabledKey);
+        }
+
+        function registerFeedback(showRecommendations) {
+            storage.local.set(cachedRecommendationsEnabledKey, showRecommendations);
+        }
+
+        function insertOnBoardingSection() {
+            $recommendedForYouSection = $.create(template(recommendedForYouOptInTemplate, {
```
katebee: Could we separate some of the responsibilities of  `insertOnBoardingSection()` into their own functions? e.g.
```
function createOptInTemplate() {
    $.create(template(recommendedForYouOptInTemplate, {
        profileIcon: svg(profileIcon, ['rounded-icon', 'rfy-profile-icon', 'control__icon-wrapper']),
    rightArrowIcon: svg(rightArrowIcon, ['i-right']),
    guardianLogo: svg(guardianLogo)
    }));
}
```

static/src/javascripts/projects/common/modules/experiments/tests/recommended-for-you.js
```diff
@@ -112,45 +140,67 @@ define([
...
         function setupComponentAttentionTracking(trackingCode) {
             require(['ophan/ng'], function (ophan) {
                 ophan.trackComponentAttention(trackingCode, $recommendedForYouSection[0]);
             });
         }
 
         function insertSection(items) {
+            var $oldSection = $recommendedForYouSection;
+
             $recommendedForYouSection = $.create(template(recommendedForYouTemplate, {
```
katebee: What are the benefits of mutating `$recommendedForYouSection`, rather than defining a local var?

e.g.

```
function insertSection(items) {
    var localVarForRecommendSection = createRecommendedForYouContent(items);

    if ($recommendedForYouSection) {
        return fastdom.write(function() {
            $recommendedForYouSection.replaceWith(localVarForRecommendSection);
            setupComponentAttentionTracking('recommended-for-you_user-history');
        });
    } else {
        return fastdom.write(function() {
            localVarForRecommendSection.insertBefore($opinionSection);
            setupComponentAttentionTracking('recommended-for-you_user-history');
        });
    }
}
```

davidfurey: setupComponentAttentionTracking requires the global `$recommendedForYouSection`
Also mutating `$recommendedForYouSection` allows `insertSection` to be called multiple times without the recommendedForYou container multiplying.

static/src/javascripts/projects/common/views/experiments/recommended-for-you-opt-in.html
```diff
@@ -0,0 +1,44 @@
+<section id="recommended-for-you" class="fc-container fc-container--has-show-more js-container--toggle" data-component="recommended-for-you" data-link-name="recommended-for-you" aria-expanded="true">
+    <div class="fc-container__inner">
+        <div class="fc-container__header js-container__header ">
+            <div class="fc-container__header__title">
+                <%= profileIcon %>
+                <span class="fc-container__title__text">recommended <span class="rfy-light-blue">for&nbsp;you</span></span>
+            </div>
```
katebee: The container is missing some spacing at the top on widescreen view. Think it is due to the lack of the show/hide toggle button:

picture 441


Quick fix (if desired) is to insert this:
```
<button class="fc-container__toggle" data-link-name="Hide">
    <span class="inline-arrow inline-icon"><svg width="24" height="10" viewBox="0 0 24 10"><path d="M0 .7l10.5 8.4 1.1.9 1.1-.9L23.2.7l-.5-.7-11.1 6.3L.6 0 0 .7z"></path></svg></span>
    <span class="fc-container__toggle__text">Show</span>
</button>
```

picture 442

picture 443

katebee (approved):
LGTM 👍 

🚢 💎

PR #15458, opened 2017-01-03
Dynamic width of programmatics

regiskuckaertz:
## What does this change?

Adds the real width of the ad to the resize message being sent by Rubicon/AppNexus/et al ad networks.

There was a `width` being set on the top ad slot; knowing it would be overwritten by the message above, I used a left margin instead, which basically achieves the same purpose. Then I removed the margin for fluid slots, i.e. native ads and Fabrics.

AppNexus billboards are correctly resized vertically, but not horizontally right now, and so they appear cropped.

## Screenshots

Before:
promo size_728 width

After:
picture 2

### DIFF
--- a/static/src/javascripts/projects/commercial/modules/messenger.js
+++ b/static/src/javascripts/projects/commercial/modules/messenger.js
@@ -176,7 +176,7 @@ define([
             id: 'aaaa0000-bb11-cc22-dd33-eeeeee444444',
             type: 'resize',
             iframeId: payload.value.id,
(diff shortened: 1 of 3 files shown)


regiskuckaertz:
[LB > BB campaign not being resized properly](https://trello.com/c/piXRYl5D/267-lb-bb-campaign-not-being-resized-properly)

rich-nguyen (approved):

rich-nguyen:
great stuff 👍

PR #15190, opened 2016-12-02
Revert "upgrade to raven 3.8.0"

rich-nguyen:
Next attempt to fix windows phone errors! @SiAdcock

PR #15459, opened 2017-01-03
Fix Raven for Webpack

SiAdcock:
## What does this change?

Raven 1.1.6 is wrapped in an immediately-invoked function that passes `this` as a parameter. Under normal circumstances, `this` would be the `window` object. However, when the Raven module is bundled by Webpack, it is wrapped in another function, that is called with an [empty object as its context](https://github.com/jhnns/webpack/blob/master/lib/MainTemplate.js#L98). I don't quite understand [why Webpack does this](https://github.com/webpack/webpack/pull/102), but it's been working that way for 3 years.

Anyway, because the context is an empty object instead of `window`, Raven's test to ensure [JSON is supported in the browser](https://github.com/guardian/frontend/blob/master/static/src/javascripts/components/raven-js/raven.js#L1094) fails, so the Raven install fails.

To fix this, I'm explicitly passing the `window` object in to the immediately-invoked function. It is not an ideal fix as it involves updating vendor code. Ideally, we'd upgrade our ancient version of Raven, but when we attempted this previously, it caused [an issue in Windows phones](https://github.com/guardian/frontend/pull/15190), generating millions of Sentry reports per day.

Raven will work in Webpack bundles, so errors will be reported correctly.

No

### DIFF
--- a/static/src/javascripts/components/raven-js/raven.js
+++ b/static/src/javascripts/components/raven-js/raven.js
@@ -1835,4 +1835,4 @@ if (typeof define === 'function' && define.amd) {
     define('raven', [], function() { return Raven; });
 }
 
-})(this);
+})(window);
--- a/static/src/javascripts/projects/common/utils/raven.js
+++ b/static/src/javascripts/projects/common/utils/raven.js
@@ -11,6 +11,10 @@ define([
 
     var app = guardian.app = guardian.app || {};
 
+    if (app.raven) {
+        return app.raven;
+    }
+
     var adblockBeingUsed = false;
     detect.adblockInUse.then(function(adblockInUse){
         adblockBeingUsed = adblockInUse;


static/src/javascripts/components/raven-js/raven.js
```diff
@@ -1835,4 +1835,4 @@ if (typeof define === 'function' && define.amd) {
     define('raven', [], function() { return Raven; });
 }
 
-})(this);
+})(window);
```
regiskuckaertz: I think your version is much better. I never understood (1) why use `this` instead of the explicit `window` in the first place, it is a vestige of unsafe JavaScript, and (2) why pass the global context as an argument, instead of a proxy object, for dependency injection.

I'm curious though, do you remember what those Windows errors were exactly? Rich came across a similar problem with Sonobi and had to [patch their implementation of `Object.getOwnPropertyNames`](https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/commercial/modules/dfp/prepare-sonobi-tag.js#L29). Perhaps we could try that too with Raven?

regiskuckaertz (approved):
Nice job 👍

static/src/javascripts/components/raven-js/raven.js
sndrs: ☝️ 👍

sndrs (approved):

static/src/javascripts/components/raven-js/raven.js
SiAdcock: Yeah, it's a weird pattern!

The Windows phone error was:

>[Can't execute code from a freed script](https://sentry.io/the-guardian/client-side-prod/issues/181594029/)

We were not able to debug this issue successfully, but I'll take a look at Raven and see if a similar method call occurs.

PR #15466, opened 2017-01-04
AMP cta overflow fix

zeftilldeath:
## What does this change?
When using AMP you could occasionally see the 'Become a supporter' rotation peaking out above the header. Looks a bit messy and ruins the magic 🎩 🌟 . Couldn't use overflow hidden on the entire header because the veggie burger has to overlap. 

Neatness

Yep, only AMP

## Screenshots
Before
screen shot 2017-01-04 at 11 14 57

After
screen shot 2017-01-04 at 12 10 17

@NataliaLKB @stephanfowler 🌮


NataliaLKB:
👍

NataliaLKB (approved):

PR #15467, opened 2017-01-04
Move 3rd party JS out of static/src

sndrs:
## What does this change?

moves 3rd party JS to `static/vendor/javascripts`

in prep for converting the current JS app to es6, all this stuff was really getting in the way. it's not our source code, therefore it doesn't really seem to belong in `src` (should really be in `node_modules`, but thats for another day)

no

## Screenshots

only diff with master is the sourcemaps

fyi @rich-nguyen (because we talked around this a while ago)


docs/02-architecture/06-client-side-architecture.md
```diff
@@ -316,6 +312,10 @@ There are five projects in the Javascript architecture:
...
 - [Membership](https://github.com/guardian/frontend/tree/master/static/src/javascripts/projects/membership) - Contains the formatters, payment and stripe javascripts
 
+### Components
+
+[Here be 3rd-party JS](https://github.com/guardian/frontend/tree/master/static/vendor/javascripts/components) such as bean, bonzo and fastdom.
+
 ### Vendor
 
-Contains [vendor JS](https://github.com/guardian/frontend/tree/master/static/src/javascripts/vendor) from the likes of forsee, formstack and stripe.
+Contains [vendor JS](https://github.com/guardian/frontend/tree/master/static/vendor/javascripts) from the likes of forsee, formstack and stripe.
```
SiAdcock: What is the difference between 3rd party JS (AKA components) and vendor JS?

SiAdcock:
Is `bower.json` doing anything anymore? Or is it acting as a sort of record of all the 3rd party dependencies we use, along with their versions?

docs/02-architecture/06-client-side-architecture.md
sndrs: we currently have `vendor` which is files we've just committed, and `components`, which is files we've pulled out of `bower_components` and committed. i don't really understand why, it's historical and seems weird, but i'd rather fix that in another PR another time (by dropping bower)

tools/__tasks__/compile/javascript/rjs.config.js
```diff
@@ -7,31 +7,31 @@
...
         membership: 'projects/membership',
         commercial: 'projects/commercial',
-        bean: 'components/bean/bean',
-        bonzo: 'components/bonzo/bonzo',
-        domReady: 'components/domready/ready',
-        EventEmitter: 'components/eventEmitter/EventEmitter',
-        fastdom: 'components/fastdom/index',
-        fence: 'components/fence/fence',
-        lodash: 'components/lodash-amd',
+        bean: '../../../static/vendor/javascripts/components/bean/bean',
```
SiAdcock: Since `baseUrl` is:
```
../../../../static/src/javascripts
```
is this path equivalent to:
```
../../vendor/javascripts/components/bean/bean
```
?

SiAdcock:
I agree with your points around Bower and Components, thanks for the explanation @sndrs

tools/__tasks__/compile/javascript/rjs.config.js
sndrs: it is indeed - nice one!

rich-nguyen:
Yeah I agree with all above, the goal is to drop bower completely, in the future, and that's where we left it with my last effort, https://github.com/guardian/frontend/pull/14919 . I _think_ `vendor` contained third party files that could not be located in bower (omniture, RIP, for example), whilst `components` held all the third party files that were installed through bower, but perhaps benefitted from a path tweak, hence the use of `bower-installer`. 👍

rich-nguyen (approved):

rich-nguyen:
Also, it has been a while since we bower installed anything on team city.

PR #15469, opened 2017-01-04
Add a manual switch to disable lazy load

regiskuckaertz:
## What does this change?

Appending `?dll=1` on any page of the site will force all ads to be displayed at once (via DFP `enableSingleRequest`).

We are in the process of writing a lambda that will screengrab pages of gu.com to showcase ads in situ. This is part of an ongoing effort to relieve AdOps from repetitive, time-consuming tasks. Because ads are by default lazy-loaded though, we must either implement a convoluted way of making all the ads slot cross the viewport, or disable lazy-loading. We chose the latter.

It is not as straightforward though. Single request mode only goes so far, our lambda still has to know when ads (i.e. iframes) have been displayed. Although there is no universal way of doing that, an acceptable heuristic is to use our `alladsrendered` event. Because the lambda will run PhantomJS, we can attach a global listener to the `window` object on that side, and fire a custom even on this side.


regiskuckaertz:
[Ad Grabber: disable lazy loading on frontend](https://trello.com/c/j5OC0frO/274-ad-grabber-disable-lazy-loading-on-frontend)

JonNorman (approved):
Lovely 👍 (once the AB test switch has been fixed)

PR #15470, opened 2017-01-04
CSS refactoring for AMP

NataliaLKB:
## What does this change?
Visually, there should be no difference. The main changes are:

- Moving the amp css into sass which can be minified
- Refactoring code to make it more similar to our site

For reviewers, please look commit by commit, it will be much easier and make more sense!

## Future changes
- There are still some hosted and glab css which is still in `scala.html` files. This should be changed at some point 

- I believe more readable code and a better developer experience
- More consistency to the rest of the website

This only affects amp


static/src/stylesheets/amp/_content.scss
```diff
@@ -0,0 +1,259 @@
...
+        @include circular;
+        position: relative;
+        width: gs-span(1);
+        height: 60px; //Intentionally off grid
+        margin: $gs-baseline/2 0;
+        overflow: hidden;
+        background-color: $neutral-7;
+
+        // fixes for webkit not properly scaling/clipping the child element (still broken on some android browsers...)
+        -webkit-transform: translateZ(0);
```
SiAdcock: Will autoprefixer take care of these?

NataliaLKB: This is for legacy browser support, but potentially for AMP thats not needed. It was originally copied from here: 
https://github.com/guardian/frontend/blob/master/static/src/stylesheets/module/content/_content.scss#L455

I actually think this should be safe to delete.

SiAdcock (approved):
Really, really tidy stuff 👍

common/app/views/fragments/amp/stylesheets/main.scala.html
```diff
@@ -5,279 +5,16 @@
...
-        color: #767676;
-    }
-
-    @* temporarily hiding these until after launch *@
-    .inline-expand-image, .content__dateline-lm {
-        display: none;
-    }
-
-    @fragments.amp.stylesheets.fonts()
+    @* TODO: these two stylesheets should be refactored into SCSS *@
```
lps88: @NataliaLKB is it possible to pass in the brand colour etc to SCSS?

NataliaLKB: @lps88 without having looked too far into it, my instinct would be to add a classname based on brand colour and have css for them all. That is how we do it with all our various tones.

PR #15472, opened 2017-01-04
Remove redundant flag from clientside

kelvin-chappell:
Now we can respond to the `isInappropriateForSponsorship` capi flag on the server-side there's no need to expose it on the client-side.


regiskuckaertz (approved):
👍

PR #15474, opened 2017-01-04
Remove left margin when there's a pageskin

regiskuckaertz:
The left margin I added earlier needs to be removed when there's a page skin.

Before:
picture 4

After:
picture 3


scottp719 (approved):
👍

PR #15475, opened 2017-01-04
New admin endpoint: returns last completed deploys

TBonnin:
## What does this change?
I am gonna experiment with Prout. This patch is duplicating an endpoint in admin that returns all completed deploys made by riffraff (with their commit sha), which Prout would check to determine if a PR has been deployed or not.

If the experiment is successful, I would then cleanup all the `deploys-radiator` and `deploys-notify` endpoints which wouldn't be necessary anymore


admin/app/model/deploys/RiffRaff.scala
```diff
@@ -6,26 +6,31 @@ import model.deploys.ApiResults.{ApiError, ApiErrors, ApiResponse}
 import play.api.libs.json.{JsError, JsSuccess, Json}
 import scala.concurrent.Future
 
+case class RiffRaffDeployTags(vcsRevision: String)
+object RiffRaffDeployTags { implicit val format = Json.format[RiffRaffDeployTags] }
```
jfsoul: Not that it matters much, but could this just be a `Json.reads`?

admin/app/controllers/DeploysRadiatorController.scala
```diff
@@ -16,8 +16,12 @@ trait DeploysRadiatorController extends Controller with Logging with Requests {
   val riffRaff: RiffRaffService
   val teamcity: TeamcityService
 
-  def getDeploys(pageSize: Option[String], projectName: Option[String], stage: Option[String]) = Action.async {
-    riffRaff.getRiffRaffDeploys(pageSize, projectName, stage).map(ApiResults(_))
+  def getCompletedDeploys(pageSize: Option[String], projectName: Option[String], stage: Option[String]) = Action.async {
```
jfsoul: Is there much point in keeping `projectName` as an argument? Could hardcode to dotcom:all

jfsoul (approved):
Looks good enough to go 👍

admin/app/model/deploys/RiffRaff.scala
TBonnin: It is used by the `Json.format[RiffRaffDeploy]` below so yes we need a format

admin/app/controllers/DeploysRadiatorController.scala
TBonnin: In theory yes, I didn't want to change to much while I am still experimenting. I am happy to reconsider the `projectName` argument if/when I cleanup

PR #15476, opened 2017-01-05
extend always on test by another month

jranks123:
## What does this change?

Extends the always on test, as we wish to keep the test running for at least another month


sndrs (approved):
magical

PR #15477, opened 2017-01-05
Server-side facebook share counts

jfsoul:
## What does this change?
Adds a feature switch that fetches the share counts from a new endpoint in `applications`.  This endpoint makes a call to the most recent version of the Facebook Graph API - which requires authentication (hence being done on the server).

This way of calling the Facebook Graph API ensures we are not rate limited by IP address (sometimes a problem for the public IP of King's Place).  It also uses an explicitly versioned API endpoint which means we will be able to tell which version of the API we are using and won't get bumped without warning when the oldest version is deprecated.

I don't think so.


common/app/common/configuration.scala
```diff
@@ -338,6 +338,9 @@ class GuardianConfiguration extends Logging {
 
   object facebook {
     lazy val appId = configuration.getMandatoryStringProperty("guardian.page.fbAppId")
+    object graphApi {
+      lazy val accessToken = configuration.getMandatoryStringProperty("facebook.graphApi.accessToken")
```
TBonnin: which FB account did you created the access token with?

applications/test/helpers/FacebookGraphApiTestClient.scala
```diff
@@ -0,0 +1,29 @@
...
+
+class FacebookGraphApiTestClient(wsClient: WSClient) extends FacebookGraphApiClient(wsClient) {
+
+  val recorder = new DefaultHttpRecorder {
+    override lazy val baseDir = new File(System.getProperty("user.dir"), "data/facebook-graph-api")
+  }
+
+  override def GET[T](endpoint: Option[String], params: (String, String)*)(asResult: WSResponse => T) = {
+    val augmentedParams = super.augmentParams(params)
+    val queryString = augmentedParams.map(pair => pair._1 + "=" + URLEncoder.encode(pair._2, "UTF-8")).mkString("&")
```
TBonnin: Can you simply cal `augmentedParams(params)` without the `super`?

Also `.map( (key, value) =>...` would be more descriptive than using tuple `._1` and `._1`

applications/test/helpers/FacebookGraphApiTestClient.scala
```diff
@@ -0,0 +1,25 @@
...
+import java.net.URLEncoder
+
+import play.api.libs.ws.{WSClient, WSResponse}
+import recorder.DefaultHttpRecorder
+import services.FacebookGraphApiClient
+
+
+class FacebookGraphApiTestClient(wsClient: WSClient) extends FacebookGraphApiClient(wsClient) {
+
+  val recorder = new DefaultHttpRecorder {
```
jfsoul: Using `HttpRecorder` partly because mocking proved very fiddly

applications/app/services/FacebookGraphApi.scala
```diff
@@ -16,7 +16,7 @@ case class URLResponse(id: String, share: ShareObject)
 case class ShareObject(share_count: Int)
 
 class FacebookGraphApiClient(wsClient: WSClient) extends implicits.WSRequests with ExecutionContexts {
-  val apiRootUrl = "https://graph.facebook.com/v2.8/"
+  val apiRootUrl = s"https://graph.facebook.com/v${Configuration.facebook.graphApi.version}/"
```
TBonnin: is it really necessary?

applications/app/services/FacebookGraphApi.scala
```diff
@@ -0,0 +1,47 @@
...
+class FacebookGraphApi(facebookGraphApiClient: FacebookGraphApiClient) {
+
+  def shareCount(path: String): Future[Int] = {
+    import URLResponseDeserializer._
+
+    facebookGraphApiClient.GET[Int](
+      endpoint = None,
+      queryString = ("id", s"https://www.theguardian.com/$path")
+    ) { response =>
+      response.json.asOpt[URLResponse]
```
TBonnin: what happens if response is an error?

applications/test/helpers/FacebookGraphApiTestClient.scala
jfsoul: 👍  on the first point

...And if I'm honest I just copied play's implementation of parameter tuples to query string 🙈

applications/app/services/FacebookGraphApi.scala
jfsoul: HttpStatusException is thrown, client-side handling of this should be the same as always

applications/app/services/FacebookGraphApi.scala
jfsoul: Not really, but I like the idea of being able to bump the version without a code change

TBonnin: config changes requires code update anyway right?

TBonnin (approved):

PR #15478, opened 2017-01-05
Add Prout config file

TBonnin:
Add Prout config file
See https://github.com/guardian/prout#add-config-file for more info

## What does this change?
We are gonna try Prout so devs are notified when their changes are deployed to PROD


jfsoul (approved):
As discussed, would change to 30m overdue though

PR #15481, opened 2017-01-05
More AMP refactoring

NataliaLKB:
## What does this change?

- Adding in some important files into `head.hosted-amp`. This enabled me to delete a fair amount in the hosted styling.
- Moving some things that were in `mainAMP` into their own fragments. I find this more readable, but it might be just me!

- Removing some duplicated code
- Code is a bit nicer to read?

Just affects amp

## Screenshots
Nothing should look that different

@guardian/dotcom-platform


gtrufitt (approved):
LGTM!

PR #15484, opened 2017-01-05
Speedup inline article ads

regiskuckaertz:
Do you notice anything strange here?

picture 1

That sea of blue is the time it takes for the initialisation of the module responsible for inserting inline ads in articles. Yes, that's a whopping 2 SECONDS.

Why is this happening? Because we are blocking until all inline ads have been inserted (or not). But we don't need to! First, all inline MPUs are lazy-loaded, it's only the inline merchandising component that is [displayed immediately](https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/commercial/modules/dfp/display-lazy-ads.js#L10). Second, all the commercial bootstrap modules following this are waiting for it to end, and so we are currently delaying a whole lot of work for no reason at all.

This change will cut through all of this and return either immediately (if there is no merch component) or after the slot for the merch component has been added ... that's one DFP round-trip and a whole lot of work we don't need to wait for anymore.

NB: objectively speaking, we will lose some reporting information, because we won't measure the time it takes for the slots to be inserted anymore. I'd like to add that back later, but only if we can factor the DFP roundtrip out of the measurements @rich-nguyen 🙌


TBonnin:
That would be interesting to see the same graph after the change has been merged?

regiskuckaertz:
Will do 😃

static/src/javascripts/projects/commercial/modules/article-body-adverts.js
```diff
@@ -63,154 +101,81 @@ define([
...
 
-    return {
-        init: init,
-
-        '@@tests': {
-            waitForMerch: waitForMerch,
-            insertLongAds: insertLongAds
-        }
-    };
+    function addSlots(totalCount) {
```
rich-nguyen: what happened to the old `countAdded` check? Can't find it

regiskuckaertz: Both "short" and "long" ads are now counted in `addInlineAds`

static/src/javascripts/projects/commercial/modules/article-body-adverts.js
```diff
@@ -3,41 +3,79 @@ define([
...
-    var inlineMerchRules;
-    var longArticleRules;
+    var getSlotName;
+
+    function init() {
+        if (!commercialFeatures.articleBodyAdverts) {
+            return Promise.resolve(false);
+        }
+
+        boot();
```
rich-nguyen: maybe we could inline `boot`?

regiskuckaertz:
[Speedup inline article ads bootstrap](https://trello.com/c/SC3QjhP8/67-speedup-inline-article-ads-bootstrap)

static/src/javascripts/projects/commercial/modules/article-body-adverts.js
rich-nguyen: I mean `totalCount` looks like it could still be 0, but the old 0 test looks like it has been removed. I guess the `forEach` will still work.

rich-nguyen (approved):

rich-nguyen:
awesome, can't wait to see the graphs update. One downside of sampling data lake over a longer period of time is the sample rate means we have to wait longer than we used to! Might add some extra query complexity for that...

👍

regiskuckaertz:
Thanks! Note for later: I'd be interested if one day you make a session to explain how it all works 😃

static/src/javascripts/projects/commercial/modules/article-body-adverts.js
regiskuckaertz: oh yeah, that. I removed it because it's one of those micro-optimisation I used to do which has no impact on the result, only it makes the code less readable. An improvement would be to remove the selector altogether and use an array of slots instead (since we create them in the module). I did that in the `-wide` version, in hindsight I should have done so here as well 😅

PR #15485, opened 2017-01-05
Prepare JS for conversation to ES6

sndrs:
## What does this change?

- moves all ES5 JS to `static/src/javascripts-legacy` directory
- sets up the expectation that JS in `static/src/javascripts` is modern ES6
  - it's linted with the standard airbnb-derived lint rules
  - it's transpiled to ES5 using babel's [latest preset](https://babeljs.io/docs/plugins/preset-latest/)
  - the require JS app is built from `static/transpiled/javascripts`
  - un-bundled modules in dev are served from there too
- converts one file to ES6 (`boot-webpack.js`)
  - can check it's all working without too much danger
  - avoids having to rewrite the webpack config 
- simplifies the `gulp` watch task while we're in there
- creates 'internal' babel env for anything transpired for internal use

should enable ad hoc conversion of our JS to a modern (a couple of years ago) version

shouldn't do

## To do

1. move a more commonly used file over once this appears to be stable (maybe `boot.js`?)
2. once _that_ appears stable, enable flow checking on it
3. create script to convert individual files, and communicate about it

~#15329~ ~#15130~ ~#15483~ _4th time lucky?_


package.json
```diff
@@ -94,8 +97,9 @@
   "scripts": {
     "postinstall": "./tools/task-runner/runner githooks --verbose",
     "sass-watch": "node-sass -w ./static/src/stylesheets -o ./static/target/stylesheets --source-map=true",
-    "css-watch": "gulp --cwd ./dev watch:css",
+    "babel-watch": "BABEL_ENV=production babel ./static/src/javascripts --watch -d ./static/transpiled/javascripts --ignore .eslintrc.js >/dev/null 2>&1",
```
gtrufitt: Why do we use a `BABEL_ENV` of `production` for the watch task? I guess because it's the same as what happens on production, but there's also a `development` one which I'm not sure is used?

sndrs: yeah you're right, that dev env is not used,  i'll kill it

sndrs: hmmmm, actually watch should probably be dev, and dev updated to match prod. things like react change how they're built (debug info etc), and that would be useful in dev

tools/__tasks__/compile/javascript/babel.js
```diff
@@ -0,0 +1,12 @@
+const execa = require('execa');
+
+const { src, transpiled } = require('../../config').paths;
+
+module.exports = {
+    description: 'Transpile',
+    task: () => execa('babel', [`${src}/javascripts`, '--out-dir', `${transpiled}/javascripts`, '--ignore', 'eslintrc.js'], {
```
gtrufitt: You ignore `eslintrc.js` in both babel commands, can you not do it in the `babelrc`?

package.json
sndrs: how about 
```js
{
    "presets": [
        "latest"
    ],
    "plugins": [
        "add-module-exports",
        "transform-es2015-modules-amd"
    ],
    "env": {
        "internal": {
            "presets": [
                ["env", {
                    "targets": {
                        "browsers": ["last 2 Chrome versions", "last 1 Safari version", "last 2 Firefox versions", "last 2 Edge versions"]
                    }
                }]
            ]
        }
    }
}

```

gtrufitt: Yeah sounds good

webpack.config.js
```diff
@@ -55,4 +57,13 @@ module.exports = {
         new webpack.optimize.UglifyJsPlugin(),
         new webpack.optimize.DedupePlugin(),
     ],
+    module: {
+        loaders: [
+            {
+                test: /\.js$/,
+                exclude: /(node_modules|bower_components|javascripts-legacy|components|vendor)/,
```
gtrufitt: Probably not that simple but is this the same as `include: 'static/src/javascripts'` (or whatever the webpack incantation is)? I wonder if inclusion is safer than exclusion anyhow?

gtrufitt (approved):
Looks good! Worth an email to let people know `javascripts` is `javascripts-legacy` though. Exciting times.

SiAdcock:
Looks awesome, I'm on board dude 👍  

ES6ify All The Things Friday, anyone?

package.json
sndrs: can't - envs just append/merge defaults, so default plugins/presets are used too. will need 3 envs, with dev and prod the same :(

gtrufitt: Ah. Well, I suppose at least that's clear, as they'll probably diverge at some point. Could always have a `common` `rc` and then specific env ones that extend (or concat arrays etc) from common? Not for now though but may be a pattern that works for these configs if they get too antisocial.

webpack.config.js
sndrs: hmmm, seems like it should be doesn't it?

sndrs:
@gtrufitt yeah, will email and slack. might also hold off till monday and let the continuous dep stuff settle down, seeing as this in only way-paving stuff

gtrufitt:
@sndrs Good idea - bonus is you can be sure everything builds properly on teamcity now though before merging 🎊

webpack.config.js
sndrs: so, seems not... :( 

it's also not how babel demo the loader working: https://github.com/babel/babel-loader#usage

gtrufitt: That's odd - I had something similar set up in the [preact prototype webpack](https://github.com/guardian/frontend-future/blob/master/prototypes/preact/webpack.config.dev-ssr.js#L22)  - but no point messing about with it if it works!

SiAdcock: It's a [Webpack thing](https://webpack.github.io/docs/configuration.html#module-loaders), rather than a Babel loader thing. You can pass an array of paths you want to `include`

sndrs:
> ES6ify All The Things Friday, anyone?

close! need a few more things in place and a bit more confidence...

webpack.config.js
sndrs: i did that and got different output...?

sndrs: if you have a module in an `include` dir, that refs module elsewhere, will that still go through the loader?

SiAdcock: Classic Webpack config docs: instructions unclear, edge cases abundant 😅

sndrs: yeah it seems to run all deps of modules through the loader if you only specify includes. 

given:

```js
test: /\.js$/,
include: [
    path.resolve(__dirname, 'static', 'src', 'javascripts'),
],
loader: 'babel-loader',
```
you get this:
```js
// static/src/javascripts-legacy/projects/common/utils/raven.js 
// (dep of boot-webpack.js)
const guardian = window.guardian;
const x = "ewwerwe"
```
becomes
```js
// static/target/javascripts/boot-webpack.js
var x = "ewwerwe";
```

MatthewJWalls:
PR build results: 
* https://s3-eu-west-1.amazonaws.com/prbuilds/PR-15485/screenshots/mobile.png 
* https://s3-eu-west-1.amazonaws.com/prbuilds/PR-15485/screenshots/wide.png 
* https://s3-eu-west-1.amazonaws.com/prbuilds/PR-15485/screenshots/tablet.png 
* https://s3-eu-west-1.amazonaws.com/prbuilds/PR-15485/screenshots/desktop.png 

 -automated message

dev/gulpfile.js
```diff
@@ -1,2 +1,50 @@
...
+const sourcemaps = require('gulp-sourcemaps');
+
+const postcss = require('gulp-postcss');
+const autoprefixer = require('autoprefixer');
+const pxtorem = require('postcss-pxtorem');
+
+const STATIC = path.resolve(__dirname, '..', 'static');
+const CSS = path.resolve(STATIC, 'target', 'stylesheets');
+const JS = path.resolve(STATIC, 'src', 'javascripts-legacy');
+const TRANSPILED = path.resolve(STATIC, 'transpiled', 'javascripts');
```
rtyley: @sndrs is it possible to move the transpiled folder under `target` so that IntelliJ will ignore it?

This is driving me crazy:

image

...there seems to be a 50/50 chance the transpiled version comes first, and when I don't notice that, I end up editing the transpiled version, _not_ the one I want under source control. This gets frustrating :crying_cat_face:

sndrs:
hmmm, is that a default setting in intellij then? it's ignore by git, does intellij not observe that?

dev/gulpfile.js
sndrs: hmmm, is that a default setting in intellij then? it's ignore by git, does intellij not observe that?

PR #15486, opened 2017-01-05
Remove `insertRequire` from bundle config

SiAdcock:
## What does this change?

Removes extraneous `insertRequire` properties from the r.js bundle config. These are not necessary as curl.js [infers the execution order](https://github.com/cujojs/curl/wiki/Module-Lifecycle) of our modules by walking the dependency tree. We don't need to specify the application's entry point.

This code is redundant and potentially confusing. Removing it makes our code cleaner.

Nope


sndrs (approved):
blimey!

PR #15489, opened 2017-01-05
Prevent MPUs to be cut off in the "most popular" container

regiskuckaertz:
From time to time, we deliver an ad in the "most popular" that is being cut off because its height is more than what the allocated space allows for. That is because the ad slot is positioned absolutely: dealing with that requires amounts of JS we could really do without.

With this PR, the container adopts a flex layout on desktop, so that the ad slot is taken into account in the computation of the height of the container.

No noticeable impact when tab switching:
mostpop

... and we can now stop worrying about changes in height:
mostpop-height


regiskuckaertz:
[Ad slot causes Most Pop component to be cut off](https://trello.com/c/XUd3zeOO/35-ad-slot-causes-most-pop-component-to-be-cut-off)

gtrufitt (approved):

PR #15490, opened 2017-01-05
Conditionally prefetch app-webpack

SiAdcock:
## What does this change?

If the user is participating in the Webpack test, this change ensures the app with the Webpack-bundled standard code (`app-webpack.js`) is [prefetched and cached ready for the user's next page visit](https://developer.mozilla.org/en-US/docs/Web/HTTP/Link_prefetching_FAQ). 

It is unclear whether this is absolutely necessary as `app-webpack.js` should be downloaded and cached on every page (including the page doing the prefetching?). 

However, it is definitely not correct for the page to be prefetching `app.js` (which contains r.js-bundled standard code) when the user is participating in the Webpack test, since it will not be executed for this user.

The prefetching implementation originated in [this PR](https://github.com/guardian/frontend/pull/11258)

Nein

### DIFF
--- a/common/app/views/fragments/head.scala.html
+++ b/common/app/views/fragments/head.scala.html
@@ -1,6 +1,7 @@
 @(page: model.Page, projectName: Option[String] = None, head: Html)(implicit request: RequestHeader, context: model.ApplicationContext)
 @import model.Page.getContent
 @import play.api.Mode.Dev
+@import mvt.WebpackTest
 
 <meta charset="utf-8" />
 <!--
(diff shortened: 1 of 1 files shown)


gtrufitt (approved):
Hmm, ain't that interesting. 👍

sndrs (approved):
> It is unclear whether this is absolutely necessary as app-webpack.js should be downloaded and cached on every page (including the page doing the prefetching?).

it should just kickstart the download during pre-parsing, so it's good it's there. but it's much better that it prefetches the actual file you need!

gtrufitt:
I actually wonder whether what we want here is 'subresource`? https://www.igvita.com/posa/high-performance-networking-in-google-chrome/#prefetching

gtrufitt:
Oh nevermind `subresource` was deprecated https://groups.google.com/a/chromium.org/forum/#!msg/blink-dev/Y_2eFRh9BOs/gULYapoRBwAJ

SiAdcock:
Running some cursory analysis, it looks like the `prefetch` gets kicked off slightly _after_ the download. Although the `prefetch` does initiate earlier, sitting in a queue:

picture 156

There's like 10ms in it, so I'm not going to spend too much time worrying about it!

PR #15492, opened 2017-01-06
Do not send random Etag header

TBonnin:
This patch ensures no Etag header is being sent if not necessary
rather than sending back a random Etag header

### DIFF
--- a/common/app/model/Cached.scala
+++ b/common/app/model/Cached.scala
@@ -124,7 +124,7 @@ object Cached extends implicits.Dates {
         (etag, result)
       }
     }.getOrElse(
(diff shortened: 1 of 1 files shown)


johnduffell:
It started as an experiment, but I actually found a notable amount of requests served from fastly cache with 304 code as a result of the header, so I left it in.  Mind you, I'm not sure how many requests fall into that category now a proper etag is calculated, and it's probably not many.  I think some clients must revalidate content even if it's fresh, perhaps due to an incorrect clock.

johnduffell:
code looks good though! 👍

TBonnin:
@johnduffell Following our discussion I kept the random etag. Just changed the name ;)

johnduffell:
😢

PR #15493, opened 2017-01-06
Removing Training-preview

TBonnin:
## What does this change?
CODE Preview is now replacing Training-Preview.
It makes frontend simpler by removing one app and make the relationship
between Preview and CAPI easier to understand:
PROD preview frontend --> PROD preview CAPI
CODE preview frontend --> CDOE preview CAPI

Also there is no broad usage of training-preview anyway

Less app, less complexity


docs/03-dev-howtos/06-repressing.md
```diff
@@ -83,7 +83,7 @@ If you just ship new facia without pressing things will break until the presser
 * `SSH` into the facia presser instance and post to http://localhost:9000/press/draft/all (have patience, it takes a couple of mins) to repress all the draft fronts (`curl -v -X POST -H 'Content-Length: 0' http://localhost:9000/press/draft/all`) 
 * The fronts in prod are pressed automatically every 5 minutes, but if you like, use `frontend.gutools` to `/press` them - if you do just click `All Frequency Fronts For Press`. This will be speedy due to multiple instances.
 * Point your local machine to `facia.stage=PROD` in `frontend.conf` and make sure everything is working before you deploy for real
-* Deploy `training-preview` & `preview` and check they still work
+* Deploy `preview` and check they still work
```
gtrufitt: `it still works`

rich-nguyen:
Can you inline standalone now, the routes and all, into `preview`? And lose another project?

jfsoul (approved):

TBonnin:
@rich-nguyen Very good question. I am thinking about making this in a second PR to keep them as small as possible.

PR #15495, opened 2017-01-06
Fix hosted AMP

lps88:
## What does this change?
Fix the layout of Hosted article/video AMP Pages (side margins had gone)

Also Make sure the onward journey doesn't show when it's empty 

Just AMP

## Screenshots
Before:
image
After:
image

@guardian/labs-beta


kelvin-chappell (approved):
👍

static/src/stylesheets/amp/_hosted.scss
```diff
@@ -334,6 +334,10 @@ $badge-width: $gs-column-width + $gs-gutter;
     bottom: 0;
     width: $badge-width;
     z-index: $zindex-ads;
+    margin: 0 $gs-gutter/2;
```
NataliaLKB: You should use `@include content-gutter();` for this

NataliaLKB:
Sorry for breaking these @lps88 ! The ones I looked at seemed fine 😞 .

static/src/stylesheets/amp/_hosted.scss
lps88: I tried it, but I need a margin rather than a padding, as there is a background colour on the badge. Is there an equivalent with margins instead?

lps88:
@NataliaLKB no worries! We noticed fairly quickly :)

static/src/stylesheets/amp/_hosted.scss
NataliaLKB: Not that I am aware of, but if you find yourself using it a lot, you can make one :)

PR #15501, opened 2017-01-09
New header variant six. Secondary Nav only on fronts

NataliaLKB:
## What does this change?

Good news: variant four was successful (the data showed it performed well)

Bad news: It looks too heavy and is too big on mobile.

Conclusion:
Lets only have the secondary nav (the grey part) on fronts, where they are needed more.

Corresponding Fastly change here: https://github.com/guardian/fastly-edge-cache/pull/734

In the end this redesign was successful, but we still are tweaking to make sure we are happy to put it out to everyone. Keep your eye out for more tweaks ahead.

Nope

@guardian/dotcom-platform

### DIFF
--- a/article/app/views/fragments/articleBody.scala.html
+++ b/article/app/views/fragments/articleBody.scala.html
@@ -39,7 +39,7 @@
                 @fragments.guBand()
             }
 
-            @if(!mvt.ABNewNavVariantFour.isParticipating) {
+            @if(!mvt.ABNewNavVariantSix.isParticipating) {
                 @fragments.headTonal(article, model, isPaidContent, amp = amp)
 
                 @if(article.tags.isFeature && article.elements.hasShowcaseMainElement) {
@@ -70,7 +70,7 @@
                             @fragments.mainMedia(article, amp)
                         }
 
-                        @if(mvt.ABNewNavVariantFour.isParticipating) {
+                        @if(mvt.ABNewNavVariantSix.isParticipating) {
                             <div class="mobile-only">
                                 @fragments.headTonal(article, model, isPaidContent, amp = amp)
                             </div>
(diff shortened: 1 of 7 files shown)


jfsoul (approved):

TBonnin:
I know it is kind of late to chip in but I find troubling that the secondary menu is larger than the primary one.

NataliaLKB:
@TBonnin me too! @stephanfowler, @zeftilldeath and I talked about this last week. This is one of the considerations for this change. On fronts it isn't very disruptive, and actually might be really helpful to find where you need to go. On articles though it just takes up too much of the page. 

There was a one line variation which I really liked that didn't perform as well as this two line one. Sometimes data and personal preferences don't match.

gtrufitt:
Did we test a variant with smaller text for the sub-nav? (I hasten to add I don't want to 'design-by-comittee' here, but smaller sub-nav text seems like a standard pattern?)

NataliaLKB:
@gtrufitt it was discussed and we experimented. The main reason why we chose not to make the text smaller is that the touch area wasn't big enough.

gtrufitt:
Interesting. I guess that can be mitigated by using more padding around each label but reducing the font-size to reinforce the visual hierarchy but I know everything was tested :)

NataliaLKB:
@gtrufitt that was one of the things that was tried for sure. 

We often have to walk the line between what the data says and what our personal preferences are. I think both have a lot of value. In this case this solution isn't perfect. I don't think anyone would say it is. We are fully aware that there is still some work to do - as this redesign has spread to changing articles and maybe fronts too.

> In the end this redesign was successful, but we still are tweaking to make sure we are happy to put it out to everyone. Keep your eye out for more tweaks ahead.

stephanfowler:
@TBonnin @gtrufitt With this variant we've reached the point that combines Alex B's design/brand criteria, with the engagement data compared to the existing nav. So this is a variant to be more widely shared internally, to get a go-ahead etc.

PR #15502, opened 2017-01-09
New explore series

blongden73:
Associates a new tag to the explore series styling. And extends the test until the end of this new series. After this we should be in a position to review the current set up.  

@SiAdcock @NataliaLKB Could you take a look please.


SiAdcock (approved):
LGTM 👍

PR #15503, opened 2017-01-09
Extending recommendations a/b test

davidfurey:
## What does this change?

Extends the recommendations a/b test

We haven't run this test yet, so aren't ready for it to expire.

No


katebee:
May be worth extending the JS expiry to match: `static/src/javascripts/projects/common/modules/experiments/tests/recommended-for-you.js`

Otherwise the test will silently stop running 😓

davidfurey:
Oops, thanks @katebee

katebee (approved):
👍

PR #15504, opened 2017-01-09
Revert "Use LinkTo for redirecting to https"

jfsoul:
Reverts guardian/frontend#15496


jfsoul:
Opened this prematurely because I forgot that the /404 route I used to test this with has been removed!

PR #15506, opened 2017-01-09
Lift the first child into an option to prevent 500 if none exist

gtrufitt:
Easier to see changes with https://github.com/guardian/frontend/pull/15506/files?w=0

## What does this change?

Probably not anything as it's not possible to cause this error anymore in composer *but* fixes it historic screw ups I guess:

"Unlikely there is an existing example of this any more, as we should have fixed the one that was broken. Essentially we need to make the child array access safe in L98-101 of VideoEmbedCleaner - just in case a video is embedded in such a way that there is no child element of an "element-video"."

Bug fix, moves a Trello card to Done.

### DIFF
--- a/common/app/views/support/cleaner/VideoEmbedCleaner.scala
+++ b/common/app/views/support/cleaner/VideoEmbedCleaner.scala
@@ -106,7 +106,7 @@ case class VideoEmbedCleaner(article: Article) extends HtmlCleaner {
         }
 
         child.wrap("<div class=\"embed-video-wrapper u-responsive-ratio u-responsive-ratio--hd\"></div>")
-      };
+      }
     }
 
     cleanVideo(document)
--- a/common/app/views/support/cleaner/VideoEmbedCleaner.scala
+++ b/common/app/views/support/cleaner/VideoEmbedCleaner.scala
@@ -96,16 +96,17 @@ case class VideoEmbedCleaner(article: Article) extends HtmlCleaner {
       element.getElementsByClass("gu-video").isEmpty
     }.foreach { element: Element =>
       val canonicalUrl = element.attr("data-canonical-url")
+      element.children().lift(0).map(child => {
+        // As Facebook have declared that you have to use their video JS plugin, which in turn pulls in their whole JS API
+        // We've decided to use the canonical URL, and create the video element here rather that CAPI, as, if it changes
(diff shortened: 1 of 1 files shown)


common/app/views/support/cleaner/VideoEmbedCleaner.scala
```diff
@@ -96,16 +96,17 @@ case class VideoEmbedCleaner(article: Article) extends HtmlCleaner {
       element.getElementsByClass("gu-video").isEmpty
     }.foreach { element: Element =>
       val canonicalUrl = element.attr("data-canonical-url")
+      element.children().lift(0).map(child => {
```
jfsoul: Nice use of `lift` 

Would marginally prefer `.lift(0).map { ... }` notation for consistency though

jfsoul (approved):
Looks good, minor comment aside 👍

common/app/views/support/cleaner/VideoEmbedCleaner.scala
TBonnin: would prefer `headOption` but not a big deal

TBonnin: Actually `collectFirst` would be better :D

TBonnin (changes requested):
See my comment about `collectFirst`

gtrufitt:
@TBonnin Replaced with `headOption.foreach` as per our non-PR chat as this function has side effects https://github.com/guardian/frontend/pull/15506/files?w=1

common/app/views/support/cleaner/VideoEmbedCleaner.scala
```diff
@@ -106,7 +106,7 @@ case class VideoEmbedCleaner(article: Article) extends HtmlCleaner {
         }
 
         child.wrap("<div class=\"embed-video-wrapper u-responsive-ratio u-responsive-ratio--hd\"></div>")
-      });
+      };
```
LATaylor-guardian: Why is there a semi colon here?

gtrufitt: Because I'm a client-side developer :man_facepalming:

TBonnin (changes requested):
👍

TBonnin (approved):
👍

PR #15515, opened 2017-01-09
Webpack boot script refactor

SiAdcock:
## What does this change?

This change allows us to use the same boot script for both Webpack-bundled code and r.js-bundled code. It also changes the output target to be AMD rather than the Webpack default output target.

⬇️  Using the same boot script reduces the size of `app-webpack.js` from 49.9KB gzipped to 37.3KB, as dependencies are shared more efficiently. As a trade-off, `commercial-rjs.js` increases in size from 156KB to 167KB.

🛠  The build process for Webpack-bundled code had diverged confusingly from the build process for r.js-bundled code. This allows us to use a single boot script for all our bundles, so the Webpack build is easier to reason about, is closer aligned to the legacy application and should facilitate a smoother migration.

🚀  Using AMD as the output target appears to reduce the average evaluation time and the time spent recalculating layout as a result of `app-webpack.js`'s execution. Cursory performance tests showed a drop from 90ms to 64ms for the former, and 40ms to 1ms for the latter.

No


MatthewJWalls:
PR build results: 
* https://s3-eu-west-1.amazonaws.com/prbuilds/PR-15515/screenshots/mobile.png 
* https://s3-eu-west-1.amazonaws.com/prbuilds/PR-15515/screenshots/wide.png 
* https://s3-eu-west-1.amazonaws.com/prbuilds/PR-15515/screenshots/tablet.png 
* https://s3-eu-west-1.amazonaws.com/prbuilds/PR-15515/screenshots/desktop.png 

 -automated message

static/src/javascripts/boot.js
```diff
@@ -29,17 +29,26 @@ define([
...
-            .then(function (boot) { boot(); });
+        if (config.tests.abWebpackBundle) {
+            return new Promise(function (resolve) {
+                require(['bootstraps/standard/main'], function (boot) {
+                    boot();
+                    resolve();
+                });
+            });
+        }
+        else {
```
sndrs: don't need this else (eslint has been picking me for months on this in other parts of the codebase 😉 )

static/src/javascripts/boot.js
```diff
@@ -29,17 +29,26 @@ define([
...
     var config = guardian.config;
 
     var domReadyPromise = new Promise(function (resolve) { domReady(resolve); });
 
     var bootStandard = function () {
-        return promiseRequire(['bootstraps/standard/main'])
-            .then(function (boot) { boot(); });
+        if (config.tests.abWebpackBundle) {
+            return new Promise(function (resolve) {
+                require(['bootstraps/standard/main'], function (boot) {
```
sndrs: any reason not to use `promiseRequire`?

sndrs: ah this is a webpack `require` right? might be worth a comment explaining that? (if i've got that right!)

webpack.config.js
```diff
@@ -4,7 +4,10 @@ const webpack = require('webpack');
 const path = require('path');
 
 module.exports = {
-    entry: './static/src/javascripts/boot-webpack.js',
+    entry: [
+        './static/src/javascripts/boot.js',
+        './static/src/javascripts/bootstraps/standard/main.js',
```
sndrs: do you need to define this if you're doing the dynamic require above?

SiAdcock: I'm not 100% sure why this is necessary, but if you don't include `/bootstraps/standard/main.js` in the `entry` list, it ends up in a separate bundle. I'll do some reading around Webpack dynamic requires to see if we're missing something.

sndrs: yeah i think webpack will try to bundle what it sees as the most efficient, this will override that. i wonder if it _is_ more efficient?

SiAdcock: Webpack creates new chunk for `bootstraps/standard/main.js` when it hits the [`require()` call in boot.js](https://github.com/guardian/frontend/blob/166dcbf0c339ca1483ca15ba63fff8643bd75f7a/static/src/javascripts-legacy/boot.js#L43). Webpack treats the `require` call as a [split point](https://webpack.github.io/docs/code-splitting.html) for the application and would under normal circumstances load the chunk asynchronously.

However, since we're concatenating standard and the boot script into the same file, we don't want Webpack to treat this as a split point. Hence we bundle these modules together into the same entry chunk.

I think this makes sense for our use case.

sndrs (approved):
magic!

gtrufitt (approved):
🎊 🏅

PR #15530, opened 2017-01-11
Fix broken call to CSS.supports on Safari 9

regiskuckaertz:
It turns out the [`window.CSS` interface](https://developer.mozilla.org/en-US/docs/Web/API/CSS) is an instance on Safari 9 and thus `supports` **must** be called using `CSS.supports`


scottp719 (approved):

gtrufitt (approved):
👍

regiskuckaertz:
Expired switches 😩

jfsoul:
@regiskuckaertz - just merged to master the switch extension 😞

regiskuckaertz:
@jfsoul thank you 😃

PR #15531, opened 2017-01-11
Extend expired comment cache switch

jfsoul:
Extending switch while we work out whether it just needs removing.

@SiAdcock


SiAdcock (approved):
👍 💯

PR #15532, opened 2017-01-11
Use ISO 8601 format in AB test start & expiry, for consistent date parsing across browsers

rtyley:
## What does this change?

This change adds a test to verify that all start and expiry dates for AB tests conform to strict ISO 8601 date format, ie `yyyy-mm-dd`. It also fixes the engagement banner AB tests which weren't doing this!

We had a bug where AB tests which were _supposed_ to be expired :dagger: :skull:   were **still** being exposed to users, multiple days after expiration - this was happening for literally 50% of our audience.

On checking Google Analytics, all these users were using mobile Safari as their web browser. It turns out Safari does not parse `'2017-1-5'` as a valid date - while other browsers like Chrome do.

You can test the behaviour of your browser here https://jsfiddle.net/165zkx8o/1/

The date-parsing behaviour of browsers is extremely inconsistent, but [all browsers we care about will correctly parse ISO 8601](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse#Browser_compatibility) if we use the strict format.

image
image

Any platform that uses the client-side A/B testing framework.

cc @SiAdcock @rupertbates @JustinPinner @rich-nguyen

### DIFF
--- a/static/src/javascripts-legacy/projects/common/modules/experiments/ab.js
+++ b/static/src/javascripts-legacy/projects/common/modules/experiments/ab.js
@@ -435,6 +435,7 @@ define([
         getTestVariantId: getTestVariantId,
         setTestVariant: setTestVariant,
         getVariant: getVariant,
+        TESTS: TESTS,
 
         /**
          * check if a test can be run (i.e. is not expired and switched on)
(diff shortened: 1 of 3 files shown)


JustinPinner:
All our switch expiry dates are formatted with single digit dates. Are they all breaking?

rtyley:
> All our switch expiry dates are formatted with single digit dates. Are they all breaking?

Yup, they sure did! This change updates those single digit dates to strict double-digit format.

JustinPinner:
👍

PR #15534, opened 2017-01-11
autoplay youtube media atoms (under certain conditions)

gidsg:
## What does this change?

Autoplays youtube media atom if user:
* Is on a Video page
* Has an internal referrer 
* Not on Android or iOS *(as these OS's block autoplay)*

CC @guardian/accessibility we might want to turn this off if a user has disabled flashing elements?

Required for YouTube PFP trial feature parity

No

## Screenshots
autoplay

### DIFF
--- a/static/src/javascripts-legacy/projects/common/modules/atoms/youtube.js
+++ b/static/src/javascripts-legacy/projects/common/modules/atoms/youtube.js
@@ -79,30 +79,20 @@ define([
 
     function shouldAutoplay(){
 
-        function isMobile() {
+        function isAutoplayBlockingPlatform() {
             return detect.isIOS() || detect.isAndroid();
         }
 
         function isInternalReferrer() {
-            /**
-             * Polyfill for String.startsWith
-             * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith
-             */
-            if (!String.prototype.startsWith) {
-                String.prototype.startsWith = function(searchString, position){
-                    position = position || 0;
-                    return this.substr(position, searchString.length) === searchString;
-                };
-            }
 
             if(config.page.isDev) {
-                return document.referrer.startsWith(window.location.origin);
+                return document.referrer.indexOf(window.location.origin) === 0;
             }
             else {
-                return document.referrer.startsWith(config.page.host);
+                return document.referrer.indexOf(config.page.host) === 0;
             }
         }
-        return config.page.contentType === 'Video' && isInternalReferrer() && !isMobile();
+        return config.page.contentType === 'Video' && isInternalReferrer() && !isAutoplayBlockingPlatform();
     }
 
     function onPlayerReady(atomId, overlay, event) {
--- a/static/src/javascripts-legacy/projects/common/modules/atoms/youtube.js
+++ b/static/src/javascripts-legacy/projects/common/modules/atoms/youtube.js
@@ -3,13 +3,15 @@ define([
     'common/modules/atoms/youtube-player',
     'common/modules/atoms/youtube-tracking',
     'common/utils/$',
-    'common/utils/config'
+    'common/utils/config',
+    'common/utils/detect'
 ], function (
     fastdom,
     youtubePlayer,
     tracking,
     $,
-    config
+    config,
+    detect
 ) {
 
     var players = {};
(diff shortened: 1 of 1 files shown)


static/src/javascripts-legacy/projects/common/modules/atoms/youtube.js
```diff
@@ -71,7 +73,33 @@ define([
         }
     }
 
+
+
+    function shouldAutoplay(){
+
+        function isMobile() {
+            const isIOS = () => /(iPad|iPhone|iPod touch)/i.test(navigator.userAgent);
```
akash1810: this is a `javascripts-legacy` module - do we want to be writing ES6 here @sndrs ?

akash1810: should we put this flag on the `config` object so others can benefit from it?

static/src/javascripts-legacy/projects/common/modules/atoms/youtube.js
```diff
@@ -71,7 +73,33 @@ define([
...
+
+        function isInternalReferrer() {
+            if(config.page.isDev) {
+                return document.referrer.startsWith(window.location.origin);
+            }
+            else {
+                return document.referrer.startsWith(config.page.host);
+            }
+        }
+        return config.page.contentType == 'Video' && isInternalReferrer() && !isMobile();
```
akash1810: `===` because javascript 😄

static/src/javascripts-legacy/projects/common/modules/atoms/youtube.js
sndrs: this shouldn't be es6 no, i'm surprised it linted? it won't be transpiled anyway. 

you probably also want to use https://github.com/guardian/frontend/blob/master/static/src/javascripts-legacy/projects/common/utils/detect.js#L164

sndrs (changes requested):
arrow functions will fail on around 30% of our visitors.

static/src/javascripts-legacy/projects/common/modules/atoms/youtube.js
```diff
@@ -71,7 +73,33 @@ define([
         }
     }
 
+
+
+    function shouldAutoplay(){
+
+        function isMobile() {
```
sndrs: you probably want https://github.com/guardian/frontend/blob/master/static/src/javascripts-legacy/projects/common/utils/detect.js#L164-L170

static/src/javascripts-legacy/projects/common/modules/atoms/youtube.js
```diff
@@ -71,7 +73,33 @@ define([
...
+            const isIOS = () => /(iPad|iPhone|iPod touch)/i.test(navigator.userAgent);
+            const isAndroid = () => /Android/i.test(navigator.userAgent);
+
+            return isIOS() || isAndroid();
+        }
+
+        function isInternalReferrer() {
+            if(config.page.isDev) {
+                return document.referrer.startsWith(window.location.origin);
```
sndrs: `startsWith` does not have wide support

sndrs (changes requested):

static/src/javascripts-legacy/projects/common/modules/atoms/youtube.js
```diff
@@ -71,7 +75,41 @@ define([
...
+        function isMobile() {
+            return detect.isIOS() || detect.isAndroid();
+        }
+
+        function isInternalReferrer() {
+            /**
+             * Polyfill for String.startsWith
+             * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith
+             */
+            if (!String.prototype.startsWith) {
```
sndrs: this isn't really a good idea, I'm afraid... 

firstly, how would anyone else have any idea that it's been polyfilled in this module? if they wanted to polyfill it to, we'd be doing twice (at least).

secondly, you're changing the prototype of `String` directly, meaning that at any point before this code is evaluated, the _global_ `String` does not contain the `startsWith` method, and then after, it does. This introduces side affects to the entire environment, just so this string can be checked.

it's generally much better practise to create a method that returns the native method if exists.

however, for something that really only needs to be done once, it's not clear it even really needs polyfilling, especially when you could use the polyfill you've added directly, or maybe

```js
if(config.page.isDev) {
    return document.referrer.indexOf(window.location.origin) === 0;
}
else {
    return document.referrer.indexOf(config.page.host) === 0;
}
```

sndrs (changes requested):

static/src/javascripts-legacy/projects/common/modules/atoms/youtube.js
```diff
@@ -71,7 +73,33 @@ define([
...
 
+
+
+    function shouldAutoplay(){
+
+        function isMobile() {
+            const isIOS = () => /(iPad|iPhone|iPod touch)/i.test(navigator.userAgent);
+            const isAndroid = () => /Android/i.test(navigator.userAgent);
+
+            return isIOS() || isAndroid();
```
sndrs: is this really the check you want? isMobile doesn't really seem to equal being iOS or android does it? do you want a breakpoint maybe?

static/src/javascripts-legacy/projects/common/modules/atoms/youtube.js
```diff
@@ -71,7 +75,41 @@ define([
         }
     }
 
+
+
+    function shouldAutoplay(){
+
+        function isMobile() {
+            return detect.isIOS() || detect.isAndroid();
```
sndrs: is this really the check you want? isMobile doesn't really seem to equal being iOS or android does it? do you want a breakpoint maybe?

gidsg: Yes, this is specifically the check I want as these platforms block autoplay at the OS level.

sndrs (approved):

PR #15535, opened 2017-01-11
Hosted AMP css fix

lps88:
## What does this change?
The wrong head CSS file was being included when working locally on Hosted AMP and liveblog AMP pages. This means styling issues that were in production were not replicable locally.

These pages now have the same styles on production as on localhost.

Just AMP.

CC @NataliaLKB


NataliaLKB:
Ah that makes sense! Would you be able to replicate the same if statement? At the moment liveblogs would be affected too.

lps88:
@NataliaLKB Yes good idea, hadn't realised

NataliaLKB (approved):
Excellent! Thank you 👍

PR #15537, opened 2017-01-11
reduce opacity on images

blongden73:
reduce opacity on the image overlay for the explore template.


zeftilldeath (approved):
This is my everything <3

PR #15538, opened 2017-01-11
Send branding param in ad call

kelvin-chappell:
This should enable the targeting and exclusion of different types of branded content in a consistent way that works.
Param has already been set up in DFP with a closed set of values.


regiskuckaertz (approved):
Yepaaa! 👍

PR #15536, opened 2017-01-11
End current engagement banner tests

rtyley:
## What does this change?

Ending current engagement banner tests, using the preferred variants as the new defaults.

* https://trello.com/c/R6RNJXQy/257-engagement-banner-test-13-uk-end-test-and-revert-to-control-message
* https://trello.com/c/nLAs3Qup/258-engagement-banner-test-8-aus-end-test-and-make-fearless-10-the-new-control
* https://trello.com/c/ZRJZCTaY/259-engagement-banner-test-12-row-end-test-and-make-3rd-article-the-new-control

cc @JustinPinner

PR #15539, opened 2017-01-11
Trigger a re-display of the engagement banner

rtyley:
## What does this change?

Triggers a 're-display' of the engagement banner. People who have formerly closed the engagement banner will now see it again. They can, of course, chose to close it once again, hiding it until the next time we change the re-display code.

https://trello.com/c/lonwAy32/255-engagement-banner-redisplay

Now that https://github.com/guardian/frontend/pull/15536 has gone out, ending the current batch of engagement banner A/B tests, and choosing the most successful variants, Jesse would like us to do a re-display so we can send a lot of traffic to them.

cc @JustinPinner


JustinPinner:
That should do it! 👍

PR #15540, opened 2017-01-11
Make hosted logo tracking consistent

kelvin-chappell:
This change makes tracking of hosted logo clicks consistent with the tracking of other sponsor logo clicks.  They will be tracked in GA as `click > sponsor logo > <lowercase sponsor name>`

/cc @steppenwells


steppenwells (approved):
Looks ok to me

PR #15541, opened 2017-01-11
.gitignore updates

NataliaLKB:
## What does this change?

Deletes things I don't _think_ are needed anymore. I would really appreciate a look from @SiAdcock @jfsoul @sndrs @TBonnin @rich-nguyen to double check my changes.

- cleaner and more relevant `.gitignore`

Nope

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,15 +1,6 @@
 artifacts.zip
 /dist
 
-frontend-article.properties
-frontend-core-navigation.properties
-frontend-front.properties
-frontend-gallery.properties
-frontend-tag.properties
-frontend-section.properties
-frontend-video.properties
-frontend-applications.properties
-
 *.min.css
 *.min.css.map
 *.csv
@@ -48,12 +39,6 @@ screenshots
 common/conf/assets
 
 .eslintcache
-
-dev/gems
-dev/casperjs
-integration-tests/target/*
-
-developer.properties
 
 *-spec-runner.html
 
--- a/.gitignore
+++ b/.gitignore
@@ -29,7 +29,6 @@ log
 logs
 ivy-sbt
 
-ivy-sbt
 tmp
 old
 target
@@ -43,6 +42,7 @@ RUNNING_PID
 local.sbt
 
 node_modules
+
 screenshots
 
 common/conf/assets
@@ -66,12 +66,12 @@ static/src/stylesheets/icons/*
 
 # Eclipse #
 .project
-.classpath
 .settings/
 .metadata
 .classpath
 .springBeans
 .cache
+
 bower_components
 
 # Scala test framework #


.gitignore
```diff
@@ -1,15 +1,6 @@
 artifacts.zip
 /dist
 
-frontend-article.properties
```
NataliaLKB: This was added in 2012. I assume it isn't needed anymore.

.gitignore
```diff
@@ -42,19 +32,18 @@ RUNNING_PID
...
+/node_modules
+
+
 screenshots
 
 common/conf/assets
 
 .eslintcache
 
-dev/gems
```
NataliaLKB: Lines 52 - 56 were added in 2013. I am assuming these are not needed anymore.

.gitignore
```diff
@@ -42,19 +32,18 @@ RUNNING_PID
...
-dev/casperjs
-integration-tests/target/*
```
NataliaLKB: This is covered by `target` which exists already above

jfsoul:
Changes look good to me but why does removing the directories from .gitignore fix validation errors? I think I'm missing something.

.gitignore
```diff
@@ -42,19 +32,18 @@ RUNNING_PID
 
 local.sbt
 
-node_modules
+dev/node_modules
+tools/amp-validation/node_modules
+static/src/stylesheets/pasteup/node_modules/
+/node_modules
+
```
desbo: you could replace all of these with `**/node_modules`

sndrs: doesn't `node_modules` ignore any instance?

.gitignore
```diff
@@ -66,13 +55,11 @@ static/src/stylesheets/icons/*
 
 # Eclipse #
 .project
-.classpath
 .settings/
 .metadata
 .classpath
 .springBeans
 .cache
-bower_components
```
sndrs: we don't want to check this in do we?

.gitignore
NataliaLKB: This was added to prevent the `deploys radiator` being ignored thus validation errors.

.gitignore
NataliaLKB: We don't have anymore do it?

NataliaLKB:
@jfsoul 
> Changes look good to me but why does removing the directories from .gitignore fix validation errors? I think I'm missing something.

Because both `bower_components` and `deploys-radiator`'s subdirectories are in the `.gitignore`  they stay in your local repo even though they are not needed anymore.

Since our processes have changed to include the transpiled directory, those directories are being transpiled locally too and failing validation. It has led to a fair amount of confusion, some of which you can see on our slack channel.

NataliaLKB:
I changed the node_modules change because I think it then adds a maintenance cost every time we add a new node_modules directory. 

I will also send an email out to let people know to about this change and if random things are appearing when they pull master, it is most likely legacy things they can delete.

.gitignore
sndrs: sadly we do - the bower installation process actually copies files from your local `bower_components` to `components` to be checked in, for long-forgotten reasons. it's on my list to remove bower once we're in webpack land, but for the moment, anyone who updates the bower deps will have this folder locally...

rich-nguyen (approved):
aside from the `bower_components`, I think these all look sensible.

NataliaLKB:
Thanks @rich-nguyen . I have taken out the `bower_components`  change 👍

PR #15543, opened 2017-01-12
Add section IDs to IMR 

katebee:
## What does this change?

- Switching to use the new imrworldwide JS 
- API ID may now vary based on the `config.page.section`

Sending correct data

Nope

##### Current Params (PROD):

picture 453

##### Updated Params (DEV):

picture 451

-----------

@regiskuckaertz @kenlim


regiskuckaertz (approved):
Great job! 👍

regiskuckaertz:
[Nielsen: what are the params for?](https://trello.com/c/FAuWtJWC/8-nielsen-what-are-the-params-for)

regiskuckaertz:
I guess you can add the other pixel back 😩  Also, you can simplify a bit your config object as a `Map[String, String]` instead of `Map[String, Object]`

regiskuckaertz (approved):
sorry it took so long. rock'n'roll 👍

PR #15544, opened 2017-01-12
Extend AB test, remove check on email switch

katebee:
## What does this change?

- Keeps the editorial email variants alive for another month
- Removes a check for a switch being on
- Better code comments

- Makes sure our users can unsubscribe from all the variants of an email, even if the ABSwitch is off or gone
- No risk of annoying users with unwanted emails

Does someone in Loyalty want to become the new owner of this switch? 😢 

Nope

------

@davidfurey 
@lmath 
@joelochlann


davidfurey:
👍

PR #15545, opened 2017-01-12
Sticky right-hand slot

regiskuckaertz:
... and finally the right-hand slot is made sticky with CSS.

There is one issue in Chrome when the creative is a Sindewinder in the expanded state and the user scrolls up so the creative crosses the viewport. I expect this issue to be fixed by the time the next stable Chrome release ships (which is the one with support for `sticky`), which is around Jan 31st.


regiskuckaertz:
[Right MPU covers the Most Pop component](https://trello.com/c/K1MY2bse/69-right-mpu-covers-the-most-pop-component)

gtrufitt (approved):

PR #15546, opened 2017-01-12
Add test for opinion emails

davidfurey:
## What does this change?

A/B test for opinion email

- The new version (listId=3811) is now the default
- 10% of users will be added to another new list (listId=3814) which will be sent the old email.  This is so that we can compare fresh users to the old and new versions

Allows us to test a new version of an editorial email.

No

### DIFF
--- a/common/app/conf/switches/ABTestSwitches.scala
+++ b/common/app/conf/switches/ABTestSwitches.scala
@@ -32,6 +32,16 @@ trait ABTestSwitches {
 
   Switch(
     ABTests,
+    "ab-opinion-email-variants",
+    "Assign users to variants of opinion emails",
+    owners = Seq(Owner.withGithub("davidfurey")),
+    safeState = On,
+    sellByDate = new LocalDate(2017, 2, 8),
+    exposeClientSide = true
+  )
+
+  Switch(
+    ABTests,
     "ab-recommended-for-you-recommendations",
     "Test personalised container on fronts",
     owners = Seq(Owner.withGithub("davidfurey")),
--- a/common/app/conf/switches/ABTestSwitches.scala
+++ b/common/app/conf/switches/ABTestSwitches.scala
@@ -35,7 +35,7 @@ trait ABTestSwitches {
     "ab-opinion-email-variants",
     "Assign users to variants of opinion emails",
     owners = Seq(Owner.withGithub("davidfurey")),
-    safeState = On,
+    safeState = Off,
     sellByDate = new LocalDate(2017, 2, 8),
     exposeClientSide = true
   )
--- a/common/app/controllers/EmailSignupController.scala
+++ b/common/app/controllers/EmailSignupController.scala
@@ -36,8 +36,9 @@ object ListIds {
   val guardianTodayUs = 1493
   val guardianTodayAu = 1506
 
-  val theBestOfOpinion = 3811
+  val theBestOfOpinion = 2313
   val newBestOfOpinion = 3811
+  val controlBestOfOpinion = 3814
   val theFiver = 218
   val mediaBriefing = 217
   val greenLight = 28
--- a/common/app/controllers/EmailSignupController.scala
+++ b/common/app/controllers/EmailSignupController.scala
@@ -36,7 +36,7 @@ object ListIds {
   val guardianTodayUs = 1493
   val guardianTodayAu = 1506
 
-  val theBestOfOpinion = 2313
+  val theBestOfOpinion = 3811
   val newBestOfOpinion = 3811
   val theFiver = 218
   val mediaBriefing = 217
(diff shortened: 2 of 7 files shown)


common/app/conf/switches/ABTestSwitches.scala
```diff
@@ -32,6 +32,16 @@ trait ABTestSwitches {
 
   Switch(
     ABTests,
+    "ab-opinion-email-variants",
+    "Assign users to variants of opinion emails",
+    owners = Seq(Owner.withGithub("davidfurey")),
+    safeState = On,
```
katebee: Is this intentional?

common/app/controllers/EmailSignupController.scala
```diff
@@ -36,7 +36,7 @@ object ListIds {
   val guardianTodayUs = 1493
   val guardianTodayAu = 1506
 
-  val theBestOfOpinion = 2313
+  val theBestOfOpinion = 3811
   val newBestOfOpinion = 3811
```
katebee: Removing the check for `2313` will remove the Opinion tones from any embed using: https://www.theguardian.com/email/form/article/2313 
This may affect some composer content (they will go back to being blue).

- Do we want to keep supporting the old embed?

Mostly what this list does is set the tone used on the email form. Not all emails need to be added to it, unless you want a tone applied, or a triggered send (confirmation email).

- Line 40 is also responsible for adding tones to the new list ID (3811) Support for the 3811 iframe embed was added in #15488
- If you change the name, you will also need to change it in `emailSignUp.scala.html` (which is applying tones based on `newBestOfOpinion`)

common/app/model/EmailSubscriptions.scala
```diff
@@ -363,8 +363,8 @@ object EmailSubscriptions {
       teaser = "Get up to speed on the most interesting and provoking issues and join the debate every afternoon",
       description = "Guardian Opinion's daily email newsletter with the most shared opinion, analysis and editorial articles from the last 24 hours — sign up to read, share and join the debate every afternoon.",
       frequency = "Weekday afternoons",
-      listId = "2313",
-      subscribedTo = subscribedListIds.exists{ x => x == "2313" || x == "3811" },
```
katebee: If 3814 is the other variant, you will probably want tones applied: https://www.theguardian.com/email/form/plaintone/3814

#15488 should cover all the places that need updating

common/app/controllers/EmailSignupController.scala
katebee: Unless, `theBestOfOpinion` should be associated with the legacy (3814) listID?

common/app/model/EmailSubscriptions.scala
```diff
@@ -1,5 +1,7 @@
 package model
 
+import controllers.ListIds
```
katebee: 🙌 👏

katebee (approved):
LGTM 💎 👍

common/app/controllers/EmailSignupController.scala
```diff
@@ -38,6 +38,7 @@ object ListIds {
 
   val theBestOfOpinion = 2313
   val newBestOfOpinion = 3811
+  val controlBestOfOpinion = 3814
```
joelochlann: this is also the old email, right? (just trying to get it straight in my head. three lists, two of which go to the old email, one of which goes to the new email)

static/src/javascripts-legacy/projects/common/modules/experiments/tests/opinion-email-variants.js
```diff
@@ -0,0 +1,77 @@
...
+                var input = $('input[value="3811"]')[0];
+                var button = $('button[value="3811"]')[0];
+                input.setAttribute('value', emailListID);
+                button.setAttribute('value', emailListID);
+            });
+        }
+
+        function runTheTest(emailListID) {
+            if (config.page.contentId === OPINION_URL) {
+                enhanceWebView(emailListID);
```
joelochlann: as discussed maybe we could can this here and elsewhere to `updateSignupPage` for clarity - not a blocker though

common/app/controllers/EmailSignupController.scala
davidfurey: Yes

static/src/javascripts-legacy/projects/common/modules/identity/email-preferences.js
```diff
@@ -159,8 +159,12 @@ define([
             if (config.switches.abEditorialEmailVariants && value === 'unsubscribe-2211') {
                 buttonString += 'removeEmailSubscriptions[]=3806&';
                 buttonString += 'removeEmailSubscriptions[]=3807&';
-            } else if (value === 'unsubscribe-2313') {
-                buttonString += 'removeEmailSubscriptions[]=3811&';
+            } else if (value === 'unsubscribe-2313') { // legacy opinion listId
```
joelochlann: potentially we can remove this - as discussed

joelochlann:
Looks good to me too. 👍  Just a few comments as discussed, not blockers

PR #15548, opened 2017-01-12
Migrate gallery ad slots into Scala land

regiskuckaertz:
@JonNorman I wanted to try my hand at some Scala, but happy to pair with you for the juicier `slice-adverts`

Most of our inline ad slots are generated client-side because we use a different naming convention on mobile than on the other breakpoints. It can easily be done on the server-side though, by creating two ad slots each time which are mutually-exclusive, i.e. either one is displayed or the other depending on the breakpoint.

They can have the same name, provided they have different IDs.

Works great hehehe 😎


regiskuckaertz:
[Gallery advert slots on the server-side](https://trello.com/c/NRj6yomf/89-gallery-advert-slots-on-the-server-side)

applications/app/views/fragments/paidAdvertisementGalleryBody.scala.html
```diff
@@ -75,7 +75,11 @@
...
         <li class="gallery2__item gallery2__item--advert">
-            <div class="gallery2__img-container js-gallery-slot">
+            <div class="gallery2__img-container">
+                @defining(rowNum / adInterval) { inlineId =>
+                    @gallerySlot(inlineId)
+                    @gallerySlot(inlineId - 1, isMobile = true)
+                }
             </div>
         </li>
     }
```
kelvin-chappell: Is there any possibility of merging this template with `galleryBody` or are they very different?
I know it's outside the scope of this PR.

kelvin-chappell (approved):
👍 
Made a small comment.

applications/app/views/fragments/paidAdvertisementGalleryBody.scala.html
regiskuckaertz: I remember it was forked because the theming of "paid for" content did not play well with the new gallery layout. @blongden73 is supposed to make a round of design update to glabs templates so that they better align across platforms ... it might be a good opportunity to get rid of that cop out.

regiskuckaertz: (to answer your question, yeah I think they are different enough that merging them would produce more confusing template logic)

rich-nguyen (commented):
nice!

regiskuckaertz:
Thanks 😎  I'd like to do the same with slice adverts, but it is a bit trickier. Because we need to have a strict numbering, and not all containers have an MPU, we can't just use the `containerIndex` in `slice.scala.html`. What I'd like to do is map all containers to a pair `Container -> (Container, Int)` where the integer is the inline MPU index and is only incremented if the container contains a slice with an MPU, pass that index down to `slice.scala.html` and generate the slot ... but I don't know where to do that.

PR #15549, opened 2017-01-12
Editorial Tone clashes with special report

zeftilldeath:
This was happening. 

screen_shot_2016-11-16_at_5 13 14_pm

Now it isn't. 

@gtrufitt can you check if I've done this the right way please? 🌴


gtrufitt (approved):
💯 A+ work, top job

PR #15550, opened 2017-01-12
Only update example url on interactive page

davidfurey:
## What does this change?

Stops trying to update an non-existant example url

Fixes an exception which was being thrown

No

### DIFF
--- a/static/src/javascripts-legacy/projects/common/modules/experiments/tests/editorial-email-variants.js
+++ b/static/src/javascripts-legacy/projects/common/modules/experiments/tests/editorial-email-variants.js
@@ -58,10 +58,10 @@ define([
         function runTheTest(emailListID, exampleUrl) {
           if (config.page.contentId === FLYERURL) {
             enhanceWebView(emailListID);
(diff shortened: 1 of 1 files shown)


static/src/javascripts-legacy/projects/common/modules/experiments/tests/editorial-email-variants.js
```diff
@@ -58,10 +58,10 @@ define([
         function runTheTest(emailListID, exampleUrl) {
           if (config.page.contentId === FLYERURL) {
             enhanceWebView(emailListID);
+            updateExampleUrl(exampleUrl);
           } else {
             updateNewslettersPage(emailListID);
           }
-          updateExampleUrl(exampleUrl);
```
katebee: So, I've just realised/remembered that The Flyer does have a example on Newsletters, which is why it was outside the if-else block.

Maybe we need a check in updateExampleUrl() to see if the dom element exists...

I'll confess, I wrote this test assuming there would always be an example url

katebee (approved):
👍

PR #15551, opened 2017-01-12
Fix: Preview internal healthcheck url should not required authentication

TBonnin:
Fix: Preview internal healthcheck url should not required authentication
Url of healthcheck has been changed but the authentication exemption has not been modified at the same time. Therefore the internal healthcheck request will hit this endpoint and because it is not exempt will redirect to `/login`

This patch is making the dependency between preview healthcheck and preview authentication exemption explicit.
I also took advantage of it to do some refactoring, replacing the 2 very similar FilterExemptions objects in `admin` and `preview` by a single class that can accept additional urls to exempt 

Real healthcheck endpoint is tested, not `/login`

## Tested in CODE?
Yes

### DIFF
--- a/common/app/googleAuth/FilterExemptions.scala
+++ b/common/app/googleAuth/FilterExemptions.scala
@@ -1,17 +1,15 @@
-package http
+package googleAuth
 
 import com.gu.googleauth.FilterExemption
 
-object FilterExemptions {
+case class FilterExemptions(additionalUrls: String*) {
 
   lazy val loginExemption: FilterExemption = FilterExemption("/login")
   lazy val exemptions: Seq[FilterExemption] = List(
+    // Default
     FilterExemption("/oauth2callback"),
     FilterExemption("/assets"),
     FilterExemption("/favicon.ico"),
-    FilterExemption("/_healthcheck"),
-    FilterExemption("/2015-06-24-manifest.json"),
-    // the healthcheck url
-    FilterExemption("/world/2012/sep/11/barcelona-march-catalan-independence")
-  )
+    FilterExemption("/_healthcheck")
+  ) ++ additionalUrls.map { url => FilterExemption(url)}
 }
--- a/preview/app/http/PreviewFilters.scala
+++ b/preview/app/http/PreviewFilters.scala
@@ -3,6 +3,8 @@ package http
 import akka.stream.Materializer
 import common.ExecutionContexts
 import GoogleAuthFilters.AuthFilterWithExemptions
+import controllers.HealthCheck
+import googleAuth.FilterExemptions
 import model.ApplicationContext
 import play.api.http.HttpFilters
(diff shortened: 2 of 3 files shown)


common/app/googleAuth/FilterExemptions.scala
```diff
@@ -1,17 +1,15 @@
...
   lazy val loginExemption: FilterExemption = FilterExemption("/login")
   lazy val exemptions: Seq[FilterExemption] = List(
+    // Default
     FilterExemption("/oauth2callback"),
     FilterExemption("/assets"),
     FilterExemption("/favicon.ico"),
-    FilterExemption("/_healthcheck"),
-    FilterExemption("/2015-06-24-manifest.json"),
-    // the healthcheck url
-    FilterExemption("/world/2012/sep/11/barcelona-march-catalan-independence")
```
NataliaLKB: This may be a silly question, but I would have expected that the [new healthcheck url](https://github.com/guardian/frontend/commit/68b44f3323955572b4043b443bae4bac9db2f4c0) to have to be exempt as well. Is this not the case?

preview/app/http/PreviewFilters.scala
```diff
@@ -3,18 +3,22 @@ package http
...
 import play.api.http.HttpFilters
 import play.api.libs.crypto.CryptoConfig
 import play.api.mvc.{Filter, RequestHeader, Result}
 
 import scala.concurrent.Future
 
-class PreviewFilters(cryptoConfig: CryptoConfig)(implicit mat: Materializer, context: ApplicationContext) extends HttpFilters {
+class PreviewFilters(cryptoConfig: CryptoConfig, healthCheck: HealthCheck)(implicit mat: Materializer, context: ApplicationContext) extends HttpFilters {
 
+  private val exemptionsUrls = healthCheck.healthChecks.map(_.path) ++ Seq("/2015-06-24-manifest.json")
```
NataliaLKB: Why is `/2015-06-24-manifest.json` need to be exempt?

common/app/googleAuth/FilterExemptions.scala
NataliaLKB: Thanks for explaining @TBonnin ! All the healthcheck urls are being added here: https://github.com/guardian/frontend/blob/0295a1225ce48984dc648c6c590bec502f399717/preview/app/http/PreviewFilters.scala#L17

jfsoul (approved):
Healthchecks defined in one place 💯

preview/app/http/PreviewFilters.scala
NataliaLKB: Oh I assume this is why: https://github.com/guardian/frontend/blob/d13f8140c72bc30515c7da60d3ce9f150773dac8/common/app/views/fragments/metaData.scala.html#L43

NataliaLKB (approved):

PR #15557, opened 2017-01-13
Lint legacy-javascript harder

sndrs:
## What does this change?

- stop legacy js linting inheriting from the app-wide config, which was allowing es6 to lint (#15534)
- use remove our lodash rule, since eslint can do it now

stop es6 in legacy passing lint

no

## Screenshots

no

## Tested in CODE?

no


static/src/javascripts/.eslintrc.js
```diff
@@ -21,5 +21,11 @@ module.exports = {
         'react/no-string-refs': 'off',
         'react/prefer-stateless-function': 'off',
         'react/no-render-return-value': 'off',
+
+        // disallow modules we used to use but are retiring
+        'no-restricted-imports': ['error', {
+            paths: ['lodash'],
+            patterns: ['!lodash/*'],
```
sndrs: this requires you to import the lodash module you need, rather than the whole lib

gtrufitt (approved):

gtrufitt:
👍 other than ci failure

PR #15558, opened 2017-01-13
Fix: crosswords clues after first direction are truncated

TBonnin:
## What does this change?
Fix bug where clues after the first direction would be truncated.
ex: `1,2,3down,4` should become `1, 2, 3 down, 4` but was in fact `1, 2, 3 down` :(

## Tested in CODE?
Yes

### DIFF
--- a/common/app/model/CrosswordData.scala
+++ b/common/app/model/CrosswordData.scala
@@ -13,17 +13,25 @@ object Entry {
   implicit val positionWrites = Json.writes[CrosswordPosition]
   implicit val jsonWrites = Json.writes[Entry]
 
-  val NumberPattern = "[^a-z]+".r
-  val DirectionPattern = "[a-z]+".r
-
   def formatHumanNumber(numbers: String): Option[String] = {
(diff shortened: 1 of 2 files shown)


common/app/model/CrosswordData.scala
```diff
@@ -13,18 +13,26 @@ object Entry {
...
-    NumberPattern.findFirstIn(numbers).map { number =>
-      val spaced = number.split(",").mkString(", ")
 
-      DirectionPattern.findFirstIn(numbers) match {
-        case Some(direction) => spaced.concat(s" $direction")
-        case None => spaced
+    //Adding space between number and direction
+    //as well as after comma
+    //ex: "2,24across,16" => "2, 24 across, 16"
+    val clues = numbers.split(',').map { singleClue =>
```
jfsoul: Adding type of `clues` might help make this a bit clearer

TBonnin: Done :D

jfsoul (approved):

PR #15559, opened 2017-01-13
Updating config, which contains updated DFP API credentials.

JonNorman:
## What does this change?
I've updated the config number to v20 - this config file has updated DFP credentials in it, which have recently expired.

We can use the DFP API again, which is kind of important for, you know, commercial things.

/@guardian/commercial-dev


regiskuckaertz:
👍

PR #15562, opened 2017-01-13
Fix Membership Engagement And Breaking News Banner Overlap

Ap0c:
## What does this change?

This prevents the membership engagement banner from overlapping with the breaking news banner. The new behaviour prioritises the breaking news banner, i.e. the membership banner will not show if there is breaking news.

The breaking news banner now fires an event when it is ready, along with a boolean signifying whether it will show or not. The membership banner waits for this event before deciding whether it can show.

Source changes:
- Added a listener to the membership banner to wait for breaking news to be ready.
- Added emitters to breaking news, and passed a boolean based on whether it will show.
- Updated membership banner tests to work with the new event setup.
- Added two tests to the engagement banner spec to cover the new breaking news behaviour.
- Tweaked the existing engagement banner tests to iron out a couple of bugs.

We don't want the breaking news banner and membership engagement banner to overlap because a) it looks bad, and b) it may interfere with our engagement banner a/b tests.

No.

## Screenshots

banner-overlap

## Tested in CODE?

Testing now.

@SiAdcock @gtrufitt @JustinPinner @rupertbates @svillafe @rtyley


svillafe (approved):
LGTM

svillafe:
Tested in **DEV** and it is working correctly 👍

static/test/javascripts/spec/common/commercial/membership-engagement-banner.spec.js
```diff
@@ -69,13 +72,15 @@ define([
...
             membershipMessages.init().then(function () {
+                mediator.emit('modules:onwards:breaking-news:ready', false);
                 var message = document.querySelector('.js-site-message');
                 expect(message).toBeNull();
             }).then(done);
         }
 
         function expectMessageNotToBeVisible(done) {
             membershipMessages.init().then(function () {
+                mediator.emit('modules:onwards:breaking-news:ready', false);
```
JustinPinner: Maybe we should have test(s) where `true` is emitted from the breaking news component?

JustinPinner:
Looking good 👍 (can we have a new test or two)?

static/test/javascripts/spec/common/commercial/membership-engagement-banner.spec.js
Ap0c: Yeah that's a good idea. It would be sensible to have tests to check the following two things (just making notes for myself here):

- The banner doesn't show when breaking news shows.
- The banner does show when breaking news doesn't (all other conditions being ok).

These tests should perhaps run first, because if something goes wrong and these conditions don't hold true then some of the other tests may fail/pass for the wrong reason.

Ap0c:
@JustinPinner I've added tests for the two conditions of the breaking news banner. I've also made a couple of tweaks to the existing engagement banner tests:

1) They now check for the presence of an element within the Javascript template (`messageContent`). Previously they were checking for the presence of an element that was rendered in the test fixture, which was always present when that fixture was used.

2) I've added the fixture to the 'should not show any messages even to engaged readers' test, as previously it was always passing regardless of the state of `canDisplayMembershipEngagementBanner`. The banner would never show because the fixture HTML that it was trying to drop the message into didn't exist.

Please let me know if you think that's ok, or if I've missed anything 🐱.

JustinPinner:
Looks good 👍

SiAdcock (approved):
Awesome stuff dude 💎

Ap0c:
Banner looks to be appearing. I also briefly managed to check with some breaking news that was running just as I deployed, and the banners hid and displayed as expected.

membership-banner

PR #15563, opened 2017-01-13
Nginx setup with valid *.thegulocal.com wildcard SSL certificate

mariogalic:
## What does this change?

Uses valid Comodo `*.thegulocal.com` wildcard SSL certificate for local Nginx setup. Certificate is valid until 09.01.2019 and is stored in `s3://identity-local-ssl/` to which Dotcom has access.

To setup Dotcom Identity Frontend see [identity-platform README](https://github.com/guardian/identity-platform/pull/149)

## Screenshots

image

image


nginx/setup.sh
```diff
@@ -1,20 +1,65 @@
+#!/bin/bash
+
+# Setup Nginx for local developement
+#
+# Copies Nginx configuration and SSL certificates to Nginx home (usually /usr/local/etc/nginx)
+# Make sure you have valid AWS credentials and then run with sudo setup.sh <profile name>
+# Remember to add nginx/hosts to your /etc/hosts
+
+S3_BUCKET="s3://identity-local-ssl/"
```
TBonnin: a bucket just to store ssl certs sounds like an overkill. Can we store them in an already exisiting bucket in frontend s3?

nginx/setup.sh
```diff
@@ -1,20 +1,65 @@
...
+fi
+
+echo "Checking AWS credentials are valid"
+if [ -z "$1" ]; then
+	PROFILE=""
+else
+	PROFILE="--profile ${1}"
+fi
+
+aws ${PROFILE} s3 ls s3://identity-local-ssl/ 1>/dev/null 2>&1
```
TBonnin: why not using the $S3_BUCKET constant defined above?

TBonnin (commented):
I made a few comment. Also can we add a file in the repository [/docs](docs) folder to link to the nginx README?

nginx/setup.sh
mario-galic: This is indeed a pre-existing bucket in Identity, so I just re-used it. (Note that other projects like DAPI would need access to this cert.)

TBonnin: fine then

nginx/hosts
```diff
@@ -0,0 +1 @@
+127.0.0.1   m.thegulocal.com	                # Dotcom
```
TBonnin: also why did we choose `m.thegulocal.com` and not `gulocal.com`?

nginx/frontend.conf
```diff
@@ -36,54 +34,4 @@ server {
         proxy_set_header Host $host;
         proxy_set_header "X-Forwarded-Proto" "https";
     }
-}
```
rupertbates: Did you mean to remove all of this? I think it's still needed isn't it?

mario-galic: It's now in https://github.com/guardian/identity-platform/blob/master/nginx/identity.conf#L148

To setup identity related nginx you wourld run `identity-platform/nginx/setup.sh`

nginx/hosts
mario-galic: We already bought `thegulocal.com` domain a while ago. We have not bought `gulocal.com` yet although it is available:

nginx/frontend.conf
rupertbates: 👍

rupertbates (approved):

PR #15564, opened 2017-01-13
de-AMD webpack build

sndrs:
## What does this change?

asumes webpack will build its own thing (though it can't yet)

cleans up master while we make other pre-wp changes to accommodate it

hope not

## Tested in CODE?

nah


SiAdcock (approved):
When guardian/fastly-edge-cache#737 has gone live, should be good to go

PR #15565, opened 2017-01-13
Smooth-scroll the new nav

sndrs:
## What does this change?

makes the new nav scroll touch-style if possible

smoothness


regiskuckaertz:
segue: another small improvement to the global user experience on mobile would be to disable the 300ms delay for clicks events on links and buttons: `a, button { touch-action: manipulation; }`

NataliaLKB:
@sndrs When googling this, I found:
image

Should we be using this?

sndrs:
https://github.com/guardian/frontend/search?utf8=✓&q=-webkit-overflow-scrolling

yeah, it's fine :)

isn't that warning just saying it's only webkit (basically)?

NataliaLKB (approved):
If we are already using it, I guess it shouldn't hurt.

PR #15568, opened 2017-01-16
Show youtube atom end slate

GHaberis:
## What does this change?

When youtube atoms end we now show the "end slate"

Drive traffic to Guardian videos

No

## Screenshots

youtube

## Tested in CODE?

No


GHaberis:
@gidsg @guardian/dotcom-platform youtube atoms related PR.

gidsg:
@GHaberis as discussed IRL, would be good to update screenshot to show the lookup coming from the CAPI logic to check it is doing the right look-up.

GHaberis:
@gidsg the screenshot now reflects the actual retrieved end slate rather than example

gidsg:
@GHaberis have you tested this on fronts playables? We need to check that the end slate that gets pulled in is from the card not from the parent URL.

GHaberis:
@gidsg I've tested that it works on fronts, which it does. I'm not 100% sure how to check the end slate is derived from the card and not the parentURL? The video I tested against was a football article and when I tested against the "reetta" test front locally the end slate displayed "More Football videos"

gidsg:
LGTM 👍 but perhaps get @akash1810 to take a look, as I've worked on part of this.

common/app/model/content/Atoms.scala
```diff
@@ -86,7 +87,14 @@ object Atoms extends common.Logging {
     content.atoms.map { atoms =>
       val quizzes = extract(atoms.quizzes, atom => { Quiz.make(content.id, atom) })
 
-      val media = extract(atoms.media, atom => { MediaAtom.make(atom) })
+      val media = extract(atoms.media, atom => {
+        val endSlatePath = EndSlateComponents(
+          sectionId = content.sectionId.getOrElse(""),
+          shortUrl = content.fields.flatMap(_.shortUrl).getOrElse(""),
```
akash1810: How do media atoms that do not have a video page behave? They will not have a `shortUrl`.

gidsg: @akash1810 in this case the `endSlatePath` is set to `None` in: https://github.com/guardian/frontend/blob/14c8ef7bbc0b07374b386c5a521a3601a1fc5ebd/applications/app/controllers/MediaAtomEmbedController.scala#L22 so the class which triggers the end slate to appear will not be added.

akash1810: worth confirming this behaviour with Editorial, though shouldn't block shipping this

akash1810 (approved):

common/app/model/content/Atoms.scala
gidsg: I think this would be a new bit of work to serve an endslate on embeds as we don't know what makes sense to show. Maybe most popular or latest? But we don't really have any context to do much more than that.

PR #15569, opened 2017-01-16
Add experience dimension for GA

gtrufitt:
## What does this change?

Adds a custom dimension for 'experience' (term defined by data-team). This is a value that can be used for developer experiments and is initially reserved for use by platform for the webpack test.

Initially testing an arbitrary value to test the data is filtering through until webpack vs curl becomes an MVT test again.

It will allows us to get high-level data in GA for user timing metrics, page load time and all the other things that GA gives us. It should not be used for proving statistical significance.

Not currently, but could do in the future (technically other platforms using UK editorial property can happily push to this dimension with different data) but initially it should be reserved for the Dotcom Platform team.

## Tested in CODE?

Na.


sndrs (approved):
😘

PR #15572, opened 2017-01-16
Sticky "paid for" band not working properly

regiskuckaertz:
There were several issues:

- in this page, the band is outside of the element where we presumed it would be found, and so the sticky script never kicked off

- the sticky script has a bug where it assumes the sticky element is at the top of its parent container, which may not always be the case.


lps88 (approved):
lgtm

regiskuckaertz:
[Paid for Banner not consistently sticking on desktop pages](https://trello.com/c/Ji0WW92w/96-paid-for-banner-not-consistently-sticking-on-desktop-pages)

PR #15573, opened 2017-01-16
Disable 300ms delay on taps

regiskuckaertz:
The difference is dramatic on iOS Safari.

cc @guardian/dotcom-platform

### DIFF
--- a/static/src/stylesheets/base/_base.scss
+++ b/static/src/stylesheets/base/_base.scss
@@ -97,3 +97,12 @@ ol,
 ul {
     list-style-position: inside;
 }
+
+/* disable 300ms delay for 'clicks' on mobile */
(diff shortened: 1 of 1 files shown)


gtrufitt (approved):
Great, I personally can't think of a reason not to... 👍

rich-nguyen:
Does it affect the hard press on links? Suppose thats pressure-based more than time-based, that's the only issue I can think of. 👍

sndrs (approved):
💎

gtrufitt:
>Does it affect the hard press on links? 

Yea, my other thought was with scrolling, if you click on a link but are actually scrolling the page, does that handle it ok?

regiskuckaertz:
@rich-nguyen @gtrufitt just checked, none of these interactions are impacted

PR #15574, opened 2017-01-16
Tidy up email subscriptions

davidfurey:
## What does this change?

Refactors email subscription/newsletter data structures to avoid repeating listIds across the codebase

Reduces risk of error.

No


katebee (approved):
Hooray! This has been on my frontend wishlist 👍 💎

common/app/model/EmailNewsletters.scala
```diff
@@ -2,419 +2,417 @@ package model
 
 import controllers.ListIds
 
-case class EmailSubscriptions(subscriptions: List[EmailSubscription])
-case class EmailSubscription(
+case class EmailNewsletters(subscriptions: List[EmailNewsletter])
```
joelochlann: Do we need this case class? Could we just use `List[EmailNewsletter]` instead and avoid a level of indirection? Lines like `availableLists.subscriptions` below seem much less clear than just `availableLists`

common/app/model/EmailNewsletters.scala
```diff
@@ -2,419 +2,417 @@ package model
...
   description: String,
   frequency: String,
   listId: String,
-  subscribedTo: Boolean = false,
-  subheading: Option[String] = None,
-  tone: Option[String] = None,
-  imageFilename: Option[String] = None,
-  signupPage: Option[String] = None,
-  exampleUrl: Option[String] = None
+  subscribedTo: (Iterable[String]) => Boolean,
```
joelochlann: Why `Iterable` rather than `Seq` or `List`? (This is my own curiosity rather than a code review suggestion!)

common/app/model/EmailNewsletters.scala
```diff
@@ -2,419 +2,417 @@ package model
...
+  val newsRoundUpEmails = List(
+    EmailNewsletter(
       name = "Guardian Today",
       theme = "news",
       teaser = "The stories you need to read, delivered to your inbox each morning",
       description = "Our editors' picks for the day's top news and commentary delivered to your inbox each morning.",
       frequency = "Every day",
-      listId = "37",
-      subscribedTo = subscribedListIds.exists{ x => x == "37" },
+      listId = ListIds.guardianTodayUk.toString,
```
joelochlann: This is much better!

identity/app/views/profile/emailPrefs.scala.html
```diff
@@ -1,22 +1,22 @@
...
 @import views.support.RenderClasses
 @import views.support.`package`.Seq2zipWithRowInfo
 
 @import common.LinkTo
 
-@(page: model.Page, emailPrefsForm: Form[EmailPrefsData], formActionUrl: String, emailSubscriptions: EmailSubscriptions)(implicit request: RequestHeader, messages: play.api.i18n.Messages, context: model.ApplicationContext)
+@(page: model.Page, emailPrefsForm: Form[EmailPrefsData], formActionUrl: String, emailSubscriptions: List[String], availableLists: EmailNewsletters)(implicit request: RequestHeader, messages: play.api.i18n.Messages, context: model.ApplicationContext)
 
-@emailListCategoryList(theme: String, subscriptions: List[EmailSubscription], isActive: Boolean) = {
+@emailListCategoryList(theme: String, subscriptions: List[EmailNewsletter], isActive: Boolean) = {
```
joelochlann: Since you've very helpfully distinguished the notion of the `EmailNewsletter` itself from subscriptions (which define which newsletters a user is subscribed to), perhaps we could change this parameter name too? From `subscriptions` to `newsletters`. And similarly `subscription` to `newsletter` in the `.map` below

joelochlann (approved):
Great start to rationalising this stuff! Just a few naming things which would make things much clearer to me - see if you agree

PR #15575, opened 2017-01-16
Unstick right MPU when at the bottom of parent container

regiskuckaertz:
Before:
sticky-nope

After:
sticky-yep


regiskuckaertz:
[The sticky right slot can overlap things now](https://trello.com/c/NkjzwZDr/97-the-sticky-right-slot-can-overlap-things-now)

scottp719 (approved):
👍 release the hounds!

PR #15555, opened 2017-01-13
Opinion styling for email

joelochlann:
Some styling & extra logic necessary to render our Opinion email. Specifically:

- Bylines are pulled in and displayed below the headline for opinion-tone articles
- Render the trail text as HTML since sometimes this has markup in it
- Show trail text on all cards for the "slow" layout. (We have three layout types for email fronts: **slow**, **medium** and **fast**. So far we've only needed medium)

## before
picture 375
picture 376

## after
picture 377
picture 378

@guardian/dotcom-platform

PR #15576, opened 2017-01-16
Fix CSS typo

joelochlann:
Fixes a mistake in #15555 (which luckily would not have affected any live emails: these tone styles get overriden for "connected-style" emails and the only one being sent out in "card-style" is The Flyer, which goes out on Wednesday afternoon)

@SiAdcock


SiAdcock (approved):
Good save! ⚽️  :goal_net:

PR #15578, opened 2017-01-16
Limit items in email containers

joelochlann:
Two things for email fronts:

- Don't display empty containers
- Max six items per container - more than that simply won't display. This is because we're now using backfill for email fronts (specifically, [The Guardian Today, US Edition](https://www.theguardian.com/email/us/daily?format=email-connected)) and some of the container sizes are unwieldy. We will probably put something more sophisticated in place eventually but it depends on the needs of the desks once they start using backfilled containers in anger (we may add a "backfill max items" field to the fronts tool)

@SiAdcock

### DIFF
--- a/facia/app/views/fragments/emailFrontBody.scala.html
+++ b/facia/app/views/fragments/emailFrontBody.scala.html
@@ -106,7 +106,7 @@
     }
 }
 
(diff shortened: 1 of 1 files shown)


SiAdcock (approved):
LGTM 👍

PR #15580, opened 2017-01-16
Add data-use-beta-a4a-implementation to AMP

katebee:
## What does this change?

Adds the `data-use-beta-a4a-implementation` to the amp-ad request

This is necessary in order to enforce the prefetching workflow of A4A

However, A4A will only work on pages delivered from the AMP CDN, otherwise Doubleclick will throw `Uncaught Error: Unknown attribute for doubleclick: useBetaA4aImplementation`

picture 454

More details and discussion here: https://github.com/ampproject/amphtml/issues/6061

Example cache page (where it will work): https://cdn.ampproject.org/c/s/amp.theguardian.com/travel/2017/jan/05/spiritual-sedona-arizona-vortex-wellbeing-new-age-hotels

Yep, affecting AMP

**AMP validation passes**


katebee:
[AMP for ads beta](https://trello.com/c/CKngeCc1/66-amp-for-ads-beta)

common/app/views/support/cleaner/AmpAdCleaner.scala
```diff
@@ -88,7 +88,7 @@ case class AmpAdCleaner(edition: Edition, uri: String, article: Article) extends
 
   def adAfter(element: Element) = {
     val ampAd = <div class="amp-ad-container">
-          <amp-ad width="300" height="250" type="doubleclick" data-loading-strategy="prefer-viewability-over-views"
+          <amp-ad width="300" height="250" type="doubleclick" data-use-beta-a4a-implementation="true" data-loading-strategy="prefer-viewability-over-views"
```
NataliaLKB: Should this be here in dev? If it only serves a purpose in PROD maybe it needs an if statement around it? However, it might not hurt to just have it in always :). Also, since it throws as error, maybe it deserves a comment?

article/app/views/liveblog/liveBlogBodyContentAMP.scala.html
```diff
@@ -13,7 +13,7 @@
         @* Add advert every 5 blog posts, up to a maximum of 8 *@
         @if(blockGroup.length == 5 && index < 8) {
             <div id="amp-ad-@index" data-sort-time="1" class="block amp-ad-container amp-ad-container--live-blog">
```
NataliaLKB: I know this is bigger than this PR, but do you think at some point we can unify the two amp ad templates that seem to have come about? Seems silly to have two things so similar.

NataliaLKB (approved):
A couple questions / comments / ideas but looks good! 👍

article/app/views/liveblog/liveBlogBodyContentAMP.scala.html
katebee: @NataliaLKB how about this? 

https://github.com/guardian/frontend/pull/15580/files#diff-0bc1c927ebbe0f3e69d8b1ccbc7efef7R93

common/app/views/support/cleaner/AmpAdCleaner.scala
```diff
@@ -87,14 +88,25 @@ object AmpAdCleaner {
...
 
   def adAfter(element: Element) = {
-    val ampAd = <div class="amp-ad-container">
-          <amp-ad width="300" height="250" type="doubleclick" data-loading-strategy="prefer-viewability-over-views"
-                  json={AmpAd(article, uri, edition.id.toLowerCase()).toString()}
-                  data-slot={AmpAdDataSlot(article).toString()}>
-          </amp-ad>
+    // data-use-beta-a4a-implementation will throw an error if the page is not in the AMP CDN
+    // Lets prevent it appearing in Dev
+    if(Configuration.environment.stage.toLowerCase == "dev") {
```
katebee: There is a lot of duplication here, but it seems that conditionals can only be used to set the value of an attribute, not the presence of it. For example, this would work, but would still throw the error:

`data-use-beta-a4a-implementation={Configuration.environment.stage.toLowerCase != "dev"}`

If we aim to move the ampAd out into it's own template, it would simplify things.

PR #15581, opened 2017-01-16
Sticky: account for resized top banner

regiskuckaertz:
Because the top banner may get resized after we compute the absolute position of the element, it may happen that the sticky snaps into place too early, which creates a jarring effect. Solution: because the parent element will never stick, we use it as a proxy to account for any top banner resize ... and then all we need is the offset of the element from the top of its parent container.


janua:
👍

PR #15582, opened 2017-01-16
Use loadscript instead of conditionally requiring 3rd JS

sndrs:
## What does this change?

- moves 3rd party JS out of require config to page config
- uses our `load-script` module to load it
- updates the `load-script` module to just require a src, with optional attrs
- as per discussion with @regiskuckaertz, `xit`s the current googletag/outbrain tests, since this breaks them and regis felt they needed a deeper looking at

stops the app depending on amd loader to load 3rd party JS

no

- [x] tested in CODE

### DIFF
--- a/static/src/javascripts-legacy/projects/commercial/modules/dfp/prepare-googletag.js
+++ b/static/src/javascripts-legacy/projects/commercial/modules/dfp/prepare-googletag.js
@@ -60,16 +60,15 @@ define([
                 moduleEnd
             );
 
-            // Return a promise that resolves after the async work is done.
             // Just load googletag. Sonobi's wrapper will already be loaded, and googletag is already added to the window by sonobi.
-            return loadScript(config.libs.googletag)
-                .then(function(){
-                    return new Promise(function(resolve){
-                        window.googletag.cmd.push(
-                            resolve
-                        );
-                    });
-                });
+            loadScript(config.libs.googletag);
+
+            // Return a promise that resolves after the async work is done.
+            return new Promise(function(resolve){
+                window.googletag.cmd.push(
+                    resolve
+                );
+            });
         }
 
         if (commercialFeatures.dfpAdvertising) {
--- a/static/src/javascripts-legacy/projects/commercial/modules/dfp/prepare-googletag.js
+++ b/static/src/javascripts-legacy/projects/commercial/modules/dfp/prepare-googletag.js
@@ -61,7 +61,7 @@ define([
             );
 
             // Just load googletag. Sonobi's wrapper will already be loaded, and googletag is already added to the window by sonobi.
-            loadScript([window.guardian.config.libs.googletag]);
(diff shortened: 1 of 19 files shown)


static/src/javascripts-legacy/projects/commercial/modules/dfp/prepare-sonobi-tag.js
```diff
@@ -3,16 +3,18 @@ define([
...
     config,
     commercialFeatures,
     dfpEnv,
-    memoize
+    memoize,
+    loadScript
 ){
     var setupSonobi = memoize(function () {
-        return Promise.resolve(require(['js!sonobi.js'])).then(catchPolyfillErrors);
+        return loadScript(window.guardian.config.libs.sonobi).then(catchPolyfillErrors);
```
regiskuckaertz: Since the `config` module is loaded above, do you need to reference `window`?

static/src/javascripts-legacy/projects/commercial/modules/dfp/prepare-googletag.js
```diff
@@ -58,15 +60,16 @@ define([
...
-            require(['js!googletag.js']);
-
             // Return a promise that resolves after the async work is done.
-            return new Promise(function(resolve){
-                window.googletag.cmd.push(
-                    resolve
-                );
-            });
+            // Just load googletag. Sonobi's wrapper will already be loaded, and googletag is already added to the window by sonobi.
+            return loadScript(window.guardian.config.libs.googletag)
```
regiskuckaertz: just `config.libs.googletag`?

regiskuckaertz (approved):
👍 

I might reintroduce the `props` argument later on to experiment with [DEFER-style async loading](https://www.html5rocks.com/en/tutorials/speed/script-loading/#toc-dom-rescue). That sonobi promise I don't like :face_with_thermometer:

sndrs:
> That sonobi promise I don't like

[this one](https://github.com/guardian/frontend/pull/15582/files#diff-f5dbe5d84ef00824d18e1f747527d240R17)? how come?

sndrs:
> I might reintroduce the `props` argument later on

would you prefer me to add it back?

regiskuckaertz:
There is a hard constraint in the fact that sonobi.js **must** be evaluated before googletag.js. This promise is used in `prepare-googletag` to satisfy that constraint (hence the memoization, because it is called twice), but it is very much like doing a synchronous request. We can use the `async=false` trick instead and that is what the `props` argument is for. If adding it back is no problem, then yes please do 😃

sndrs:
@regiskuckaertz – @SiAdcock and I got the tests all sorted, and near enough restored the loading behaviours. sorry to ask, fancy another review? 😊

regiskuckaertz:
🎉  great job chaps 👍

PR #15516, opened 2017-01-09
Only serve email fronts if ?format=email

joelochlann:
Since we now have the notion of email fronts in the [facia tool](https://github.com/guardian/facia-tool/pull/220) and [facia client](https://github.com/guardian/facia-scala-client/pull/180), we can handle them differently in frontend. Because they're not designed for normal viewing on the web, they should only be visible in email-friendly format, i.e. if `format=email` is passed as a parameter. If it's not an email request, email fronts are not served.

This means I was able to get rid of the hack from #15224 and associated switch. So as regards being hidden or not, email fronts behave as other fronts (if they're hidden, they're not visible on the live site in any format).

@guardian/dotcom-platform 

and particularly @TBonnin who had some comments on #15224 - you suggested writing some tests for `shouldServeFront` to document the hacky logic. Do you still feel that's necessary now that the behaviour is saner? If so, advice on the best approach to make this testable would be appreciated, since the `ConfigAgent` is a scheduled Akka agent and it also has a static dependency on a real S3 client

PR #15587, opened 2017-01-17
Link 'Trouble viewing' to email version of email fronts

joelochlann:
The "Trouble viewing" link at the bottom of the email has always been broken for front-generated emails since it tries to link to the web front (based on the behaviour of article-generated emails which link to the normal article rather than the email-friendly one). Email fronts are only visible in email format.

This fixes the problem by using the request URI as the link href for front emails. Since we have a few different email formats & alternative ways of requesting them in play at the moment (`?format=email`, `?format=email-connected`, `/email`) this will ensure it always links to the right one. In the near future when we consolidate to a single way of requesting email format, this will still work without modification.

@davidfurey @lmath

### DIFF
--- a/common/app/views/fragments/email/footer.scala.html
+++ b/common/app/views/fragments/email/footer.scala.html
@@ -5,7 +5,7 @@
 @import common.{CanonicalLink, LinkTo}
 
 @fullRow(Seq("email__links")) {
-    <a class="email__link" href="https://profile.theguardian.com/email-prefs">Manage your emails</a> | <a class="email__link" href="%%unsub_center_url%%">Unsubscribe</a> | <a class="email__link" href="@LinkTo(page.metadata.canonicalUrl.map(LinkTo(_)).getOrElse(CanonicalLink(request, page.metadata.webUrl)))">Trouble viewing?</a>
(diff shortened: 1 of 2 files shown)


joelochlann:
Hmm, I was thinking about the fact that once we consolidate onto `?format=email` some links in existing emails would be broken (those with `?format=email-connected`). I've actually had a potentially better idea:

- Email fronts always render in email format. No `?format=email` etc needed
- Non-email fronts **can** be rendered as email with `?format=email` (there may not be a use case for this yet but seems reasonable to leave this possibility open)

This would have these advantages:
- The broken "Troubling Viewing" link would work without this extra logic, for future **and** existing emails
- The [slightly convoluted logic](https://github.com/guardian/frontend/pull/15516/files) around whether fronts can be rendered would be simplified. In fact, it wouldn't care about email fronts at all. The only logic would be: render as an email if the request is an email request or the front is an email front
- Preview in the fronts tool would work without any extra logic

What do we think?

davidfurey:
Sounds good to me

joelochlann:
Excellent, thanks @davidfurey - closing this in preference to a new PR implementing my last comment

PR #15588, opened 2017-01-17
Move Outbrain entirely on the client-side

regiskuckaertz:
This is the first part of a series of changes to allow us to run the paid content test. In hindsight, the server-side generated `DIV` did not bring anything we could not do as fast on the front-end.

I will not obsess over the performance of the third-party generation, hopefully this will go away 👼


regiskuckaertz:
[Adios Outbrain test: move Outbrain insertion to frontend](https://trello.com/c/PT5NY3Eh/101-adios-outbrain-test-move-outbrain-insertion-to-frontend)

kelvin-chappell (approved):
Is plista still going or is it just a variant of outbrain?  Otherwise it might need some kind of special treatment.

regiskuckaertz:
Still running, it is taken care of in the JS module at runtime.

PR #15589, opened 2017-01-17
resolve promise immediately if window.YT.Player already exists

GHaberis:
## What does this change?

Guards against potential race condition where window.YT.Player is defined before we define our Promise that waits for window.YT.Player to be defined before resolving. 

Reduces possibility of youtube atoms failing to initialise.

No

## Screenshots

N/A

@gidsg


GHaberis:
@gidsg I've also fixed a bug withe end slate here too (it was attempting to add the end slate if it already existed).

gidsg (approved):
Looks good, since this is defensive I think let's merge and test more on PROD. 👊 👍

PR #15594, opened 2017-01-17
Email fronts always get served as email

joelochlann:
Initially as a solution to the problem detailed in #15587 (but actually with a few other benefits too), I've decided to tweak the functionality of email fronts to be this:

- Email fronts always render in email format. No `?format=email` etc needed
- Non-email fronts **can** be rendered as email with `?format=email` (there may not be a use case for this yet but seems reasonable to leave this possibility open)

This has these advantages:
- The broken "Troubling Viewing" link (see #15587) would work without this extra logic, for future **and** existing emails
- The [slightly convoluted logic](https://github.com/guardian/frontend/pull/15516/files) around whether fronts can be rendered would be simplified. In fact, it wouldn't care about email fronts at all. The only logic is: render as an email if the request is an email request or the front is an email front
- Preview in the fronts tool would work without any extra logic

@davidfurey @lmath

### DIFF
--- a/common/app/services/ConfigAgentTrait.scala
+++ b/common/app/services/ConfigAgentTrait.scala
@@ -134,8 +134,8 @@ trait ConfigAgentTrait extends ExecutionContexts with Logging {
     getFrontPriorityFromConfig(id).contains(EmailPriority)
 
   // email fronts are only served if the email-friendly format has been specified in the request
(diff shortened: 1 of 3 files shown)


joelochlann:
Deploying now (have tested on CODE) since we'd like this fixed for tomorrow morning's Media Briefing

davidfurey:
Belated 👍

PR #15556, opened 2017-01-13
Structured data - Add recipes for google search 

mchv:
## What does this change?

The associated changes add [structured data](https://developers.google.com/search/docs/guides/intro-structured-data) markup for articles containing structured recipes.

When providing structured data markup for our online content, we make that content eligible to appear in two categories of Google Search features:

- Rich results — Structured data for things like recipes, articles, and videos can appear in Rich Cards, as either a single element or a list of items. Other kinds of structured data can enhance the appearance of your site in Search, such as with [Breadcrumbs](https://developers.google.com/search/docs/data-types/breadcrumbs), or a [Sitelinks Search Box](https://developers.google.com/search/docs/data-types/sitelinks-searchbox).

- Knowledge Graph cards — If you're the authority for certain content, Google can treat the structured data on your site as factual and import it into the [Knowledge Graph](http://www.google.com/intl/es419/insidesearch/features/search/knowledge.html), where it can power prominent answers in Search and across Google properties. Knowledge Graph cards appear for authoritative data about organizations, and events. [Movie reviews](https://developers.google.com/search/docs/data-types/reviews), and movie/music play actions, while based on ranking, can also appear in Knowledge Graph cards once they are reconciled to Knowledge Graph entities.

<p align="center">
screen shot 2017-01-13 at 11 10 50
screen shot 2017-01-13 at 11 58 33
screen shot 2017-01-13 at 11 59 01
</p>

## To do
 
The following improvements will be made in future pull requests:

* [x] Add correctly image based on google requirements
* [x] Add correctly author based on google requirements
* [x] Format correctly publication time  based on google requirements
* [x] Add `totalTime`
* [ ] Enrich our models to provide currently missing fields (`description`, `aggregateRating`)

## How many recipes are currently structured?

There is only `1` recipe that has been published on the Content API, but we have been working with [recipeasy](https://github.com/guardian/recipeasy/) to curate and structure a large number:

screen shot 2017-01-13 at 12 07 47

 

## What is the value of this and can you measure success?

By adding structured data to our archive we aim to improve our google ranking for archive content and increase our number of page views. Current metrics are:

- ranked 7.5 on average
- 50 millions page views per year

By adding the meta 

## Does this affect other platforms - Amp, Apps, etc?

Amp is concerned as well

## Screenshots

we need some improvements before showing screenshots 😃 

@stephanfowler

PR #15595, opened 2017-01-18
Structured data - Recipes - add image and improved formatting

mchv:
## What does this change?

Following [initial work on structured data ](https://github.com/guardian/frontend/pull/15556):

* Add correctly the image using our image service
* Return instructions as a `List` instead of `String`
* Improved formatting of ingredients by rendering better `Double` instance

See previous changes.

AMP is concerned as well.

## Screenshots
Soon 😄

### DIFF
--- a/common/app/views/fragments/atoms/structureddata/recipe.scala.html
+++ b/common/app/views/fragments/atoms/structureddata/recipe.scala.html
@@ -1,18 +1,20 @@
 @(recipe: model.content.RecipeAtom)(implicit request: RequestHeader)
 @import play.api.libs.json.Json
+@import views.support.{ImgSrc, GoogleStructuredData}
 
 <script type="application/ld+json">
      {
       "@@context": "http://schema.org/",
       "@@type": "Recipe",
       "name": "@recipe.data.title",
-      @for(category <- recipe.data.tags.category) {"recipeCategory": "@category",} 
-      @* TODO image and autor "image": "@http://i.guimcode.co.uk/img/media/82fbac20615e571c3783602c570cb12ffee5f000/0_2071_4912_2948/master/4912.jpg?w=300&h=300&q=55&auto=format&usm=12&fit=crop",
-      "author": {
-      "@@type": "Person",
-      "name": "scoopnana"
-     }, *@
-     @recipe.atom.contentChangeDetails.published.map { published => "datePublished": "@published",}
+      @for(category <- recipe.data.tags.category) {"recipeCategory": "@category",}
+      @recipe.data.images.headOption.map {img => "image": "@imgValue(img)",}
+      @recipe.data.credits.headOption.map { name => "author": {
+          "@@type": "Person",
+          "name": "@name"
+        },
+      } 
+      @recipe.atom.contentChangeDetails.published.map { published => "datePublished": "@published",}
      
      @* 
      TODO - We do not have this description and rating in the model yet  
@@ -41,10 +43,14 @@
      },
      *@
      "recipeIngredient": @Html(Json.stringify(Json.toJson(ingredientValues(recipe.data.ingredientsLists.flatMap(_.ingredients))))),
-     "recipeInstructions": 
-        "@recipe.data.steps.zipWithIndex.map { case(step,index) => \n@index. @step }.mkString"
+     "recipeInstructions": @Html(Json.stringify(Json.toJson(recipe.data.steps.zipWithIndex.map { case(step,index) => s"$index. $step" })))
     }
 </script>
+
+@imgValue(image: com.gu.contentatom.thrift.Image) = @{
+    val media =  model.content.MediaAtom.imageMediaMake(image, "")
+    ImgSrc.findNearestSrc(media, GoogleStructuredData).getOrElse("")
+}
 
 @formatDuration(mins: Short) = @{s"PT${mins}M"}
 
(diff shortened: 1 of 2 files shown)


mchv:
Example output 
```javascript
<script type="application/ld+json">
     {
      "@context": "http://schema.org/",
      "@type": "Recipe",
      "name": "Delia Smith’s coffee and walnut sponge cake",
      "recipeCategory": "Baking","recipeCategory": "Dessert",
       "image": "//i.guimcode.co.uk/img/media/3adaa7412ba3b6a19204cc808d30c70dd00ce035/0_1442_4809_2881/4809.jpg?w=300&amp;h=300&amp;q=55&amp;auto=format&amp;usm=12&amp;fit=max&amp;s=e7ed7a5adf83d13a2e5c05c0eb487b91",
       "author": {
          "@type": "Person",
          "name": "Delia Smith"
        },
       
       "datePublished": "ChangeRecord(1477378816000,Some(User(off-platform@guardian.co.uk,None,None)))",
      "recipeYield": "12 servings",
     "recipeIngredient": ["115 g self-raising flour ","1 teaspoon level baking powder ","115 g butter ","2  eggs ","115 g golden caster sugar ","1 tablespoon instant espresso coffee powder ","50 g walnuts ","250 g mascarpone ","1 handful instant espresso coffee powder ","1 tablespoon golden caster sugar ","1 tablespoon milk ","8  walnut halves "],
     "recipeInstructions": ["0. Lightly butter and line the bases of two 18cm x 4cm sponge tins. Preheat the oven to 170C/gas mark 3.\r\n\r\n","1. Start off by sifting the flour and baking powder into a roomy mixing bowl, holding the sieve quite high to give the flour a good airing as it goes down, then add the butter, eggs, caster sugar and coffee powder. Now, using an electric hand whisk, mix to a smooth, creamy consistency for about 1 minute. After that take a tablespoon and fold the chopped nuts into the mixture.\r\n\r\n","2. Next, divide the mixture between the two prepared tins, level off using the back of a tablespoon and bake near the centre of the oven for about 25 minutes. The sponges are cooked when you press lightly with your little finger and the centre springs back. Remove them from the oven and after about 30 seconds loosen the edges by sliding a palette knife all round then turn them out onto a wire cooling tray. Now carefully peel back the lining by gently pulling it back. Then lightly place a second cooling tray on top and just flip them both over so that the tops are facing upwards (this is to prevent them sticking to the cooling tray).\r\n\r\n","3. While the cakes are cooling, make up the filling: in a small bowl combine the mascarpone, coffee powder and caster sugar with 1 tablespoon of milk – what you need is a smooth spreadable consistency. As some mascarpones are wetter than others it’s impossible to be precise, but add a bit more milk if you think it needs it.\r\n\r\n","4. When the cakes are cold, spread half the filling over one, sandwich them together, then spread the rest over the top using a palette knife and making a swirling pattern. Then finish off by placing the walnuts in a circle near to the edge. Store in a polythene box in the fridge.\r\n"]
    }
</script>
```

shtukas:
:+1:

common/app/views/fragments/atoms/structureddata/recipe.scala.html
```diff
@@ -56,20 +62,43 @@
...
+    if(q == q.toInt) {
+        q.toInt
+    } else {
+        q match {
+            case 0.75 => "3/4"
+            case 0.5 => "1/2"
+            case 0.25 => "1/4"
+            case 0.125 => "1/8"
+            case _ => q 
+        }
```
TBonnin: What about an additional match case in order to handle the whole number case? rather than an `if`
```
...
case 0.125 => "1/8"
case qty if qty == qty.toInt => qty.toInt 
case _ => q 
...
```

common/app/views/fragments/atoms/structureddata/recipe.scala.html
```diff
@@ -56,20 +62,43 @@
...
 *@
 @ingredientValues(ingredients: Seq[com.gu.contentatom.thrift.atom.recipe.Ingredient]) = @{
     ingredients.map { ingredient => 
         val q = if (ingredient.quantity.isDefined) {
-            ingredient.quantity.get    
+            formatQuantity(ingredient.quantity.get)    
         } else if (ingredient.quantityRange.isDefined) {
-            s"${ingredient.quantityRange.get.from}-${ingredient.quantityRange.get.to}"
+            val range = ingredient.quantityRange.get
+            s"${formatQuantity(range.from)}-${formatQuantity(range.to)}"
```
TBonnin: Would the following be more concise than using `if else if else`?
```
val q = ingredient.quantity
          .map(formatQuantity)
          .orElse(ingredient.quantityRange.map(range => s"${formatQuantity(range.from)}-${formatQuantity(range.to)}" ))
          .getOrElse("")
```

TBonnin (commented):
Made tiny comments but lgtm

common/app/views/fragments/atoms/structureddata/recipe.scala.html
mchv: sounds indeed better, will update in next pull request.

paperboyo:
Image in the json example doesn't use media service's master asset. It should (for many different reasons).

mchv:
@paperboyo https://github.com/guardian/frontend/pull/15600

mchv:
`findNearestSrc` should probably marked as deprecated unless I missed why it is useful.

TBonnin:
@mchv I know this is already merged but there are few comments I didn't mention yesterday:
Putting logic in the template (here the formatting bits) should be avoided as much as possible. 
- By using reusable code block in template  you lose the benefits of type checking (everything is a more or less a string in twirl template)
- Testing is almost possible

It would make sense for example to create something like `RecipeUnitFormatter` and `RecipeQuantityFormatter` objects you can reuse and write unit tests for.

mchv:
@TBonnin I would like to discuss this in person as I disagree quite strongly on the approach. From my point of view, formatting should always be in templates and (that exactly the approach of a MVC framework such as play). I think unit testing of [each formatting property is a waste](http://rbcs-us.com/documents/Why-Most-Unit-Testing-is-Waste.pdf) but I can write few acceptance tests that ensure the response match for a given object, but still not convinced it has a lot of value..

TBonnin:
We can debate about the usefulness of every unit tests. I agree a lot of them are not required. 
The logic here is quite simple but "it is super simple, I don't need to test this" are famous last words.

However regarding writing logic in twirl template reusable blocks and losing some benefits of type checking, this PR is actually a good example: if you had been able to specify the return type of the `formatQuantity` reusable block the compiler would have warn you that it is actually returning 3 different types (Int, Double and String) and therefore its return type is Any. It might have no consequence in your case but in general I would avoid writing logic in Play templates.

My other argument regarding reusability still stands as well. If someone needs in the future to display some recipe quantity in another template, it would be more likely they know or find out about those `formatter` objects than know about reusable code blocks in another templates.

PR #15597, opened 2017-01-18
Move merchandising high over to client-side

regiskuckaertz:
Second step to get us closer to the paid content test. The high-merch component is taken care of on fronts and Outbrain is not showing up there, we just need to handle articles/videos/etc here.


katebee (approved):
👍 Cool, LGTM

PR #15600, opened 2017-01-18
Structured Data - Recipe - Use master asset for resizing or fallback to default

mchv:
## What does this change?

Usage of the largest asset for resizing following @paperboyo recommendation.
Benefits are:
- Better image quality (we resize from a non-previously resized image)
- Better caching

### DIFF
--- a/common/app/views/support/Profile.scala
+++ b/common/app/views/support/Profile.scala
@@ -200,7 +200,7 @@ object ImgSrc extends Logging with implicits.Strings {
     }
   }
 
-  private def findLargestSrc(ImageElement: ImageMedia, profile: Profile): Option[String] = {
+  def findLargestSrc(ImageElement: ImageMedia, profile: Profile): Option[String] = {
     profile.largestFor(ImageElement).flatMap(_.url).map{ largestImage =>
       ImgSrc(largestImage, profile)
     }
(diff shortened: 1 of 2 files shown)


mchv:
@paperboyo

LATaylor-guardian:
👍

common/app/views/support/Profile.scala
```diff
@@ -200,7 +200,7 @@ object ImgSrc extends Logging with implicits.Strings {
     }
   }
 
-  private def findLargestSrc(ImageElement: ImageMedia, profile: Profile): Option[String] = {
+  def findLargestSrc(ImageElement: ImageMedia, profile: Profile): Option[String] = {
```
TBonnin: Why was it private before? Would the proper function to use be `srcsetForProfile`?

mchv: @TBonnin  `findNearestSrc`is not private and is currently used in templates so I don't a reason why `findLargestSrc` should stay private.

 `srcsetForProfile ` implementation is the following:
```scala
def srcsetForProfile(profile: Profile, imageContainer: ImageMedia, hidpi: Boolean): String = {
    if(ImageServerSwitch.isSwitchedOn) {
      s"${findLargestSrc(imageContainer, profile).get} ${profile.width.get * (if (hidpi) 2 else 1)}w"
    } else {
      s"${findNearestSrc(imageContainer, profile).get} ${profile.width.get * (if (hidpi) 2 else 1)}w"
    }
  }
```

There is 2 reasons why we can't use it:

- It returns a `String` ending by `w` intended to be used as a value for `srcset`  but we need a `url`
- it fallback to an image that does not have the required ratio (`1:1`) if `ImageServerSwitch` is turned off.

paperboyo:
Great! Thanks a lot! 😃

PR #15601, opened 2017-01-18
Improve prepare googletag perf measurements

regiskuckaertz:
@rich-nguyen two small things:

1- the `init` promise resolves when the DFP lib has loaded, otherwise it has to wait for the whole command queue (which, when the lib has loaded, may contain many more tasks than the ones in here) to be emptied

2- I took the network call measurement out of the module execution time. I'm not sure you'll agree with this, but it feels like something we aren't able to improve ourselves.

### DIFF
--- a/static/src/javascripts-legacy/projects/commercial/modules/dfp/prepare-googletag.js
+++ b/static/src/javascripts-legacy/projects/commercial/modules/dfp/prepare-googletag.js
@@ -44,17 +44,21 @@ define([
 
     function init(moduleName) {
 
+        function moduleStart() {
+            // Use Custom Timing to time the googletag code without the sonobi pre-loading.
+            performanceLogging.moduleStart(moduleName);
+        }
+
         function moduleEnd() {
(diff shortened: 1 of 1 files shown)


regiskuckaertz:
On a side note, I've been thinking about performance logging. What say we slightly modify the API to:

- add a function `createLogger :: String -> Logger` that takes a module name and returns a "logger"
- define a new type `Logger` with the following interface: `start`, `stop` and `pause`

It seems to me that passing the module name to the module itself is a bit gimmicky, while passing such a logger instance would make more sense.

rich-nguyen:
Yeah, I understand why those changes make sense. Regarding the `init` change, we would now move to timing the execution of the queue, and blocking only on the enqueue, which makes total sense. Removing network call is fine, it reduces what we measure to what we can control. In other graphs and data, the network call is still embedded, so as long as that's understood, network time is still captured.

👍

rich-nguyen (approved):

PR #15603, opened 2017-01-18
Move CommercialClientSideLogging job to admin app

TBonnin:
We created admin-jobs one year ago with the goal to move all jobs from
admin into admin-jobs.
A year later we failed at doing so, only one job has been created in
admin-jobs (CommercialClientSideLogging) and none has been moved.
Since moving jobs out of admin is NOT a priority at all at the moment, I
think it makes sense to keep things simple with all the jobs in admin.
We can revisit how we manage jobs in frontend later (ie: use lambda
maybe for example)

Keep things simple and eventually remove admin-jobs which is barely doing anything at the moment

## Tested in CODE?
No

### DIFF
--- a/admin-jobs/app/AppLoader.scala
+++ b/admin-jobs/app/AppLoader.scala
@@ -14,7 +14,6 @@ import play.api.ApplicationLoader.Context
 import play.api.mvc.EssentialFilter
 import play.api.routing.Router
 import services.ConfigAgentLifecycle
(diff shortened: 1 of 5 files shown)


TBonnin:
@rich-nguyen How can I verifyit works properly once I have deployed this change? thx

jfsoul (approved):

TBonnin:
Answering my own question: I will look in the logs to make sure everything is fine

PR #15605, opened 2017-01-18
New Nav updates

NataliaLKB:
## What does this change?
This PR prepares the new nav test to go out to everyone. Other than some refactoring, The biggest difference is that there are only going to be 4-6 items shown on the subsection nav. This is ordered by popularity and so I have [split all the lists into most popular, and least popular](https://github.com/guardian/frontend/compare/nb-nav-changes?expand=1#diff-a6aa5d8644989032a408124ec83e6fd1R9).

I would recommend reviewing commit by commit 😄 

The nav is also on every page, not just fronts.

@zeftilldeath is updating the styling slightly in a separate PR.

cc @stephanfowler 

This is the subsection nav:
image

This test will go out to everyone as it was successful. We will continue to iterate and test each of the changes to make sure we are providing the best user experience to the most number of users.

Nope

## Tested in CODE?
No, but I tested a lot locally so I shouldn't need to.


static/src/stylesheets/layout/_ab-new-header-test-variant.scss
```diff
@@ -719,50 +728,45 @@ $navigation-horizontal-padding: $gs-gutter * 2 + $gs-gutter / 2;
...
+    line-height: 1.3em;
 
     &:hover,
     &:focus {
         text-decoration: none;
         color: #ffffff;
     }
 
-    .tertiary-navigation__item--current-section & {
+    &--current-section {
```
sndrs: this would make this selector tricky to grep

zeftilldeath: I will sort that.

gtrufitt (approved):
:1st_place_medal:

common/app/common/NewNavigation.scala
```diff
@@ -119,86 +70,114 @@ object NewNavigation {
...
-      NavLink("first dog on the moon", "/profile/first-dog-on-the-moon"),
-      NavLink("Katharine Murphy", "/profile/katharine-murphy"),
-      NavLink("Kristina Keneally", "/profile/kristina-keneally"),
-      NavLink("Richard Ackland", "/profile/richard-ackland"),
-      NavLink("Van Badham", "/profile/van-badham"),
-      NavLink("Lenore Taylor", "/profile/lenore-taylor"),
-      NavLink("Jason Wilson", "/profile/wilson-jason"),
-      NavLink("Brigid Delaney", "/profile/brigiddelaney"),
-      columnists
+    val au = NavLinkLists(
```
NataliaLKB: Just consistency, but looking at it again this morning I agree with you! I will change all the opinion to be multi line again.

Thanks :)

PR #15608, opened 2017-01-18
Moving breaking-news to onward and removing admin-jobs

TBonnin:
I made a lot of mistakes when creating the breaking-news feature, from the
internal design choices (using AkkaActor for instance) to the location of
the endpoint (in admin-jobs)
Ideally we would need to refactor the breaking-news feature but this
patch is only addressing the location issue.
Breaking-news has nothing to do in admin-jobs. This patch is moving it
in onward, even though it might not be the best place for the `POST /news-alerts/alert` endpoint, at least it is logical to have the `GET /news-alert` endpoint in onward (and because of the bad design choice we cannot simply break the 2 endpoints into 2 different apps)

Admin-jobs not doing anything anymore, this patch is removing it ✂️ 

Simplicity

## Tested in CODE?
Yes

cc @guardian/dotcom-platform


onward/app/authentication/AuthenticationSupport.scala
```diff
@@ -1,9 +1,12 @@
 package authentication
 
-import play.api.mvc.{Result, Results, Request, ActionBuilder}
+import play.api.mvc.{ActionBuilder, Request, Result, Results}
 
 import scala.concurrent.Future
 
+/**
+  * Created by thomas_bonnin on 18/01/2017.
+  */
```
jfsoul: ^ 😁

TBonnin: Damn it IntelliJ :fist_left:

jfsoul:
Two questions:

- Is there any config to move this time?
- Is dev-build happy? (Looks like it should be to me)

DiegoVazquezNanini:
:+1: but why are you moving the endpoints to onward? 
IMHO sounds like a weird place for the breaking news endpoints.

TBonnin:
@jfsoul 
I will remove the config block in a subsequent PR
yes dev-build is happy 😄 

@DiegoVazquezNanini 
Because displaying breaking-news is similar to an onward journey:
- Secondary content loaded asynchronously after the primary content has been display
- Non priority endpoint (ie: if article feature breaks, it is a big deal. If breaking-news feature stops working, it less of a big deal)

jfsoul (approved):

NataliaLKB (approved):

PR #15611, opened 2017-01-18
Update config (adding redis.host for admin)

TBonnin:
Follow-up on https://github.com/guardian/frontend/pull/15603. I forgot to move the `redis.host` property into admin config block :(

## Tested in CODE?
Yes

cc @rich-nguyen


rich-nguyen:
thanks for fixing, @TBonnin

PR #15612, opened 2017-01-18
Improve environment around webpack

sndrs:
## What does this change?

- update the build task to error correctly and provide feedback
- use eslint module resolving rule properly

### DIFF
--- a/tools/__tasks__/compile/javascript/webpack.js
+++ b/tools/__tasks__/compile/javascript/webpack.js
@@ -1,4 +1,34 @@
+// eslint-disable-next-line import/no-unassigned-import
+require('any-observable/register/rxjs-all');
+
+const Observable = require('any-observable');
+
+const webpack = require('webpack');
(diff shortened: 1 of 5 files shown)


tools/__tasks__/compile/javascript/webpack.js
```diff
@@ -1,4 +1,34 @@
+// eslint-disable-next-line import/no-unassigned-import
+require('any-observable/register/rxjs-all');
```
gtrufitt: Out of interest, why do we have to do this?

gtrufitt (approved):
LGTM 💙

tools/__tasks__/compile/javascript/webpack.js
sndrs: tbh, i don't really know :) i guess its an implementation of an observer, and any observer just uses it? copy and pasted!

PR #15613, opened 2017-01-19
Revert "Add section IDs to IMR "

regiskuckaertz:
Nielsen requested we revert our changes as some parameters need to be fiddled with.


katebee (approved):
👍

PR #15614, opened 2017-01-19
Effect the new business requirements for when Plista should run.

JonNorman:
## What does this change?
Previously the widget provided by third-party content-provider Plista should only be shown when the switchboard switch was toggled AND the user was in the AU, not mobile or tablet and not on world news.

The criteria have now been simplified and if the switch is on and we are in the AU edition, show Plista.

### DIFF
--- a/static/src/javascripts-legacy/projects/commercial/modules/third-party-tags.js
+++ b/static/src/javascripts-legacy/projects/commercial/modules/third-party-tags.js
@@ -60,9 +60,7 @@ define([
             .then(init);
         }
 
-        var isMobileOrTablet = ['mobile', 'tablet'].indexOf(detect.getBreakpoint(false)) >= 0;
-        var shouldIgnoreSwitch =  isMobileOrTablet || config.page.section === 'world' || config.page.edition.toLowerCase() !== 'au';
-        var shouldServePlista = config.switches.plistaForOutbrainAu && !shouldIgnoreSwitch;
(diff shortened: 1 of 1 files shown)


JonNorman:
[Outbrain/Plista switch update](https://trello.com/c/VdO5q9LT/116-outbrain-plista-switch-update)

static/src/javascripts-legacy/projects/commercial/modules/third-party-tags.js
```diff
@@ -60,9 +60,7 @@ define([
             .then(init);
         }
 
-        var isMobileOrTablet = ['mobile', 'tablet'].indexOf(detect.getBreakpoint(false)) >= 0;
-        var shouldIgnoreSwitch =  isMobileOrTablet || config.page.section === 'world' || config.page.edition.toLowerCase() !== 'au';
-        var shouldServePlista = config.switches.plistaForOutbrainAu && !shouldIgnoreSwitch;
+        var shouldServePlista = config.switches.plistaForOutbrainAu && config.page.edition.toLowerCase() === 'au';
```
rich-nguyen: could inline this now, simple logic

rich-nguyen (approved):

static/src/javascripts-legacy/projects/commercial/modules/third-party-tags.js
JonNorman: True - I think I'll leave it for now as this may all change in the near future anyway given the current Outbrain tests.

PR #15615, opened 2017-01-19
Speed up GPT & Sonobi load

regiskuckaertz:
Instead of waiting for Sonobi to have been loaded and parsed by the browser before loading the GPT library, we download them in parallel and use [the `async=false` trick](https://www.html5rocks.com/en/tutorials/speed/script-loading/#toc-dom-rescue) to force their evaluation in order.

@sndrs I slightly updated `loadScript`, as nobody is using the attributes so far. There was also a small bug in that a script could be added twice (`resolve` does not return).

Before:
picture 14

After:
picture 13


sndrs:
> `resolve` does not return

not this again 😉

static/src/javascripts-legacy/projects/common/utils/load-script.js
```diff
@@ -1,23 +1,24 @@
...
 
-    function loadScript(src, attrs) {
+    function loadScript(src, props) {
         if (typeof src !== 'string') {
-            return new Promise.reject('no src supplied');
+            return Promise.reject('no src supplied');
         }
+
+        if (document.querySelector('script[src="' + src + '"]')) {
+            return Promise.resolve();
```
sndrs: nice one

static/src/javascripts-legacy/projects/commercial/modules/dfp/prepare-sonobi-tag.js
```diff
@@ -1,21 +1,22 @@
...
+    dfpEnv
 ){
-    var setupSonobi = memoize(function () {
-        return loadScript(config.libs.sonobi).then(catchPolyfillErrors);
-    });
+
+    function setupSonobi() {
+        // Setting the async property to false will _still_ load the script in
+        // a non-blocking fashion but will ensure it is executed before googletag
+        return loadScript(config.libs.sonobi, { async: false }).then(catchPolyfillErrors);
```
sndrs: is `async` true by default?

regiskuckaertz: Yes. From the HTML5 rocks article:

> [Setting `async` to `false`] gives our scripts a mix of behaviour that can’t be achieved with plain HTML. By being explicitly not async, scripts are added to an execution queue, the same queue they’re added to in our first plain-HTML example. However, by being dynamically created, they’re executed outside of document parsing, so rendering isn’t blocked while they’re downloaded

sndrs (approved):

PR #15620, opened 2017-01-19
New footer to match the new header!

NataliaLKB:
## What does this change?
@zeftilldeath came up with a lovely design to improve our footer, and keep it consistent with the new header. 

Based on the data we looked at, we have kept the back to top button because users do use it. We have not added a menu to the bottom though, as not enough people found that useful. 

Looks nicer, and more consistent with the new header

Nope! 
...@zeftilldeath should we do the AMP footer??

## Tested in CODE?
Nope, but these changes shouldn't need to be

### DIFF
--- a/static/src/stylesheets/layout/_ab-new-header-test-variant.scss
+++ b/static/src/stylesheets/layout/_ab-new-header-test-variant.scss
@@ -740,6 +740,8 @@ $navigation-horizontal-padding: $gs-gutter * 2 + $gs-gutter / 2;
     border-bottom: $gs-baseline / 3 solid #747a81;
     max-height: $gs-baseline * 3.4;
     overflow: hidden;
+    z-index: 1;
+    position: relative;
 
     @include mq($from: mobile, $until: mobileLandscape) {
         max-height: $gs-baseline * 3.5;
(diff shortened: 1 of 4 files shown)


static/src/stylesheets/layout/_ab-new-header-test-variant.scss
```diff
@@ -901,3 +903,80 @@ $caption-button-size: 32px;
...
+    width: 100%;
+}
+.footer__back-to-top__container {
+    @include content-gutter();
+    background-color: $guardian-brand;
+
+    a:hover > &,
+    a:focus > & {
+        background-color: $guardian-brand-dark;
+    }
```
SiAdcock: Probably unnecessary for mobile

static/src/stylesheets/layout/_ab-new-header-test-variant.scss
```diff
@@ -901,3 +903,80 @@ $caption-button-size: 32px;
...
+}
+
+.back-to-top__icon {
+    position: relative;
+    float: right;
+    margin-top: -($gs-baseline / 2);
+    margin-bottom: -($gs-baseline / 2);
+    border-radius: 100%;
+    background-color: $news-main-2;
+    cursor: pointer;
```
SiAdcock: Unnecessary for mobile

SiAdcock (approved):
Couple of dinky points, otherwise 👍

static/src/stylesheets/layout/_ab-new-header-test-variant.scss
NataliaLKB: Discussed in person: Leaving in this styling in preparation for the desktop version (coming soon)

PR #15624, opened 2017-01-19
Webpack Redux

SiAdcock:
## What does this change?

- Bundle entire application using Webpack (behind a MVT switch for now 😅 )
- All Webpack chunks are now hashed, and added to the asset map 🔀 
- Upgrade to Webpack 2.2.0 🎉 

We can now conduct a more complete test of the Webpack-bundled application, without having to run the r.js-bundled app alongside it

No

## Tested in CODE?

- [x] tested on CODE

**NB:** this PR has nothing to do with [Redux](http://redux.js.org/)


tools/webpack-loaders/svg-loader/index.js
```diff
@@ -0,0 +1,17 @@
+const path = require('path');
+
+module.exports = function inlineSvgLoader(content) {
+    // eslint-disable-next-line no-unused-expressions
+    this.cacheable && this.cacheable(); // webpack-specfic weirdness
+
+    const match = content.match(/<svg([^>]+)+>([\s\S]+)<\/svg>/i);
+    const prefix = 'inline-';
+    const imageType = path.dirname(this.resourcePath).split('/').pop();
+    const fileName = path.basename(this.resourcePath).split('.svg')[0];
```
SiAdcock:
```
path.basename(this.resourcePath, '.svg');
```

https://nodejs.org/api/path.html#path_path_basename_path_ext

tools/webpack-loaders/svg-loader/index.js
```diff
@@ -7,7 +7,7 @@ module.exports = function inlineSvgLoader(content) {
     const match = content.match(/<svg([^>]+)+>([\s\S]+)<\/svg>/i);
     const prefix = 'inline-';
     const imageType = path.dirname(this.resourcePath).split('/').pop();
-    const fileName = path.basename(this.resourcePath).split('.svg')[0];
+    const fileName = path.basename(this.resourcePath, '.svg');
```
sndrs: 👍

static/src/javascripts/boot-webpack.js
```diff
@@ -1,3 +1,39 @@
-import boot from 'bootstraps/standard/main';
+import domready from 'domready';
+import raven from 'common/utils/raven';
+import bootStandard from 'bootstraps/standard/main';
+import config from 'common/utils/config';
 
-boot();
+// let webpack know where to get files from
+// eslint-disable-next-line camelcase,no-undef
+__webpack_public_path__ = `${config.page.assetsPath}javascripts/`;
```
NataliaLKB: Asked Simon in person about this. Would it be worth adding a comment explaining why this variable is in this format?

tools/__tasks__/compile/hash/index.js
```diff
@@ -12,26 +12,51 @@ const mkdirpp = pify(mkdirp);
...
-    const fileHash = hasha.fromFileSync(path.resolve(target, file), { algorithm: 'md5' });
-    return Object.assign(fileMapping, {
-        [file]: path.join(path.dirname(file), fileHash, path.basename(file)),
-    });
-}
+const standardAssetsMap = () => {
+    const assets = glob.sync('**/!(*.map|*webpack*.js)', { nodir: true, cwd: target });
+    return assets.reduce((assetMap, asset) => {
+        const assetHash = hasha.fromFileSync(path.resolve(target, asset), { algorithm: 'md5' });
+        return Object.assign(assetMap, {
```
NataliaLKB: What do you guys think about this rule: https://github.com/airbnb/javascript#whitespace--after-blocks

Personally I like it. I find code much easier to read with nice whitespace. I also read [maintainable JS](http://dl.finebook.ir/book/4a/13598.pdf) which I found really useful.

gtrufitt:
https://github.com/guardian/frontend/blob/03df2b81de80cc7ff4e0b50e4f5d048e0222fa00/common/app/views/fragments/analytics/google.scala.html#L39

You could use `window.config.tests && window.config.tests.abWebpackBundle ? 'webpack-load' : 'curl-load'` or do it on the serverside with the MVT switch.

gtrufitt (approved):
I'm not 100% on top of the architecture here, so it'd be good to have a quick explainer at some point, but lgtm. Awesome work! 💯

NataliaLKB (approved):
Amazing work guys! LGTM 👍

static/src/javascripts/boot-webpack.js
sndrs: if that wasn't clear, then absolutely - on its way

tools/__tasks__/compile/hash/index.js
sndrs: weird that's not picked up in their eslint config (which this file should be subject to)?

if you want update the lint config to catch it (esp if it can fix them automatically) be totally happy to +1 a PR

but we should be cautious about picking up style prefs in PRs – there is just no way we're all going to agree (the repo's history is littered with 'discussions' around this 😉). v up for broad, super-strict linting, but one value of being super-strict is that whatever gets through should count as ok...

that said, it seems like it should be in the config (unless its' tricky to lint?)

sndrs: maybe this? https://github.com/mgeraci/eslint-plugin-block-function-spacing

sndrs:
will merge #15648 into this before merging, if possible

tools/webpack-loaders/svg-loader/index.js
```diff
@@ -0,0 +1,17 @@
+const path = require('path');
+
+module.exports = function inlineSvgLoader(content) {
+    // eslint-disable-next-line no-unused-expressions
+    this.cacheable && this.cacheable(); // webpack-specfic weirdness
```
gtrufitt: I was reading this morning that this is the default in webpack 2, so not needed here.

SiAdcock: Can you remember where you read this? Docs say loaders are [not cacheable by default](https://webpack.js.org/api/loaders/#cacheable).

SiAdcock: Plus ça change...

gtrufitt: cool, cool...

image

PR #15625, opened 2017-01-19
Speedup Hosted video

regiskuckaertz:
We still have some work left before blocking `enhanced` until the commercial modules have run, namely the hosted by modules.

In this change, I'm just not blocking the commercial setup on any of the `require` calls necessary to load the video plugin, so that the `commercial.init` promise resolves early. This will take these operations out of the commercial load time (the `baseline` measurements), yet they still will be taken into account in the module breakdown.

### DIFF
--- a/static/src/javascripts-legacy/projects/commercial/modules/hosted/video.js
+++ b/static/src/javascripts-legacy/projects/commercial/modules/hosted/video.js
@@ -34,7 +34,7 @@ define([
     var player;
 
     function isDesktop() {
-        return contains(['desktop', 'leftCol', 'wide'], detect.getBreakpoint());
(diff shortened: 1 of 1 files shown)


lps88 (approved):
LGTM

static/src/javascripts-legacy/projects/commercial/modules/hosted/video.js
```diff
@@ -28,13 +27,12 @@ define([
     events,
     videojsOptions,
     fullscreener,
-    contains,
     loadingTmpl
 ) {
     var player;
```
akash1810: possibly out of scope of this PR, but could this `var` be declared within `setupVideo`?

akash1810 (approved):

PR #15627, opened 2017-01-19
Plaintone message should be dark, because background is light

davidfurey:
## What does this change?

Changes the colour of email signup success message from light to dark.

Improved usability

No

## Screenshots

Before:
screen shot 2017-01-19 at 16 33 27

After:
screen shot 2017-01-19 at 16 31 46

## Tested in CODE?

No

### DIFF
--- a/static/src/stylesheets/module/email/_form.scss
+++ b/static/src/stylesheets/module/email/_form.scss
@@ -334,6 +334,13 @@
     }
 }
 
+.email-sub--plaintone {
+    .email-sub__message__headline,
+    .email-sub__message__description {
+        color: $neutral-1;
+    }
+}
+
 // POST-ARTICLE
 
 .email-sub__iframe--post-article {


joelochlann (approved):

katebee:
...however, weekend Reading wants to be light :cry: :rebel:

picture 465

superfrank:
Ideally we'd pull it in based on the colour of each interactive/the background can change. So Opinion would be Orange, Weekend Reading would be feature pink. The Guardian Today (in the footer) would be the neutral grey like you've changed it to but white text.

It's also used in places like this: https://www.theguardian.com/world/ng-interactive/2016/sep/16/gun-nation-a-journey-to-the-heart-of-americas-gun-culture-video

katebee:
https://www.theguardian.com/world/ng-interactive/2016/sep/16/gun-nation-a-journey-to-the-heart-of-americas-gun-culture-video uses the `plaindark` route so wont be affected by styles nested within the `.email-sub--plaintone` selector

This will affect anything using the plaintone route (Weekend Reading, Flyer, Opinion, Composer pages)

We may be able to extend the tone-specific modes to cover the response: https://github.com/guardian/frontend/blob/65d18c35e4b782c56841ee232f4a0e3e425d1de7/static/src/stylesheets/module/email/_form.scss#L381

katebee (approved):
Since it is mostly used against a light background ([example](https://www.theguardian.com/commentisfree/2017/jan/20/the-week-in-patriarchy-your-worst-nightmare)), and the custom text doesn't appear in app, I think a default of dark font colour is the better choice 👍

PR #15631, opened 2017-01-19
Update config version to v22

TBonnin:
Update config version to v22
Removing admin-jobs block from config

## Tested in CODE?
Not yet but I will 😁


jfsoul (approved):
These config bump reviews are totally worth it 😬

TBonnin:
I mechanically added you 😁  
I am actually not expecting review of config bumping

PR #15633, opened 2017-01-19
Add section IDs to IMR, and wrap the IMR update in an AB test

katebee:
## What does this change?

IMR was updated, then it was not... Now the update is wrapped in a zero participation AB test...

Stay tuned for the next exciting instalment 💥 

Follow up to #15543

Checks if all the IMR params are correct

No

## Tested in CODE?

Not yet


regiskuckaertz (approved):
Nice 👍

PR #15634, opened 2017-01-19
Only wrap inline elements in <a>, because Outlook

joelochlann:
Outlook doesn't like `<a>` elements wrapping block-level elements. This means a lot of links don't work in Outlook. I've rejigged things so links only wrap inline elements. This means there are a few gaps on the cards which aren't clickable, but I think that's acceptable.

@lmath @guardian/dotcom-platform


lmath:
LGTM 👍

PR #15635, opened 2017-01-20
only show youtube atom duration if not on front

GHaberis:
## What does this change?

On fronts we don't want to show the bottom bar with the icon/duration

N/A

No

## Screenshots

Before...

picture 29

and after...

picture 28

## Tested in CODE?

No

@gidsg @duarte

### DIFF
--- a/static/src/javascripts-legacy/projects/common/modules/atoms/youtube.js
+++ b/static/src/javascripts-legacy/projects/common/modules/atoms/youtube.js
@@ -115,9 +115,9 @@ define([
         };
 
         if (overlay) {
-            var formattedDuration = getFormattedDuration(players[atomId].player.getDuration());
-            
-            setDuration(formattedDuration, overlay);
+            if (!config.page.isFront) {
+                showDuration(atomId, overlay);
+            }    
             
             players[atomId].overlay = overlay;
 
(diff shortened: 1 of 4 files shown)


duarte:
Looks good to me re: timestamp!

Can we get the button to go a size bigger on large cards — or should that go in a separate PR?

TBonnin:
I am curious as a reader: what lead to this decision to hide the duration?
As a reader I find it useful to know how long a video is gonna be. That's might influence me to click (or not click) on it

static/src/javascripts-legacy/projects/common/modules/atoms/youtube.js
```diff
@@ -115,9 +115,9 @@ define([
         };
 
         if (overlay) {
-            var formattedDuration = getFormattedDuration(players[atomId].player.getDuration());
-            
-            setDuration(formattedDuration, overlay);
+            if (!config.page.isFront) {
```
gidsg: We could work this out server side and pass it down as a parameter on the template.

duarte:
@TBonnin removing it from playable videos on fronts is for parity with the current (non-youtube) video setup. In small cards, and with the in-article player timestamp, the image area becomes too busy in small cards. We have alternate designs for timestamp, but they have so far tested to reduce video plays — so that still needs work

gidsg:
LGTM, but could do with another pair of eyes as I've worked on this...

akash1810 (approved):
lgtm

PR #15637, opened 2017-01-20
Speedup Hosted onward

regiskuckaertz:
Same as for hosted videos, we don't want to block the commercial bootstrap on a network call. Also moved the carousel init out of the fastdom frame to make room for other modules that need to write to the DOM.


lps88 (approved):
Nice

PR #15638, opened 2017-01-20
Visible email footers for outlook.com and office 365

lmath:
## What does this change?
Setting background-image and background-color individually in the css for email footers because just background-color being observed is better than neither. 

Users of outlook.com and outlook 365 should be able to see the manage your emails, unsubscribe, and trouble viewing? links.

No. 

## Screenshots
This is how the footer looks in e.g. Yahoo Mail:
image

The footer in outlook.com and office 365 before this change appeared transparent:
image

This change should help to ensure that if the background image fails, the background colour still appears and you can see the links:
image

## Tested in CODE?
No; not required.

@joelochlann @davidfurey

### DIFF
--- a/common/app/views/fragments/email/stylesheets/footer.scala.html
+++ b/common/app/views/fragments/email/stylesheets/footer.scala.html
@@ -4,7 +4,10 @@
 
 <style>
     .footer {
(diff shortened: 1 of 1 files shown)


joelochlann (approved):
Great stuff! 👍  🎉

PR #15639, opened 2017-01-20
The ultimate Sonobi/Googletag load speedup

regiskuckaertz:
Given that we know:

- network calls will happen
- if they succeed, sonobi will be executed before googletag
- every interaction with them is done via the googletag command queue

... then we don't need our commercial bootstrap to block on network calls.

- [ ] needs to fix failing tests


katebee (approved):
👍 ⚡️

PR #15640, opened 2017-01-20
Removing the eu referendum badge

NataliaLKB:
## What does this change?
Removes the EU referendum badge. [Trello card here](https://trello.com/c/VXDrUrLa/194-remove-the-eu-referendum-badge-from-the-brexit-eu-referendum-tag)

cc @desbo and @superfrank because you guys originally put it in I believe. Let me know if I missed anything!

Less code!

Yes, this removes it from AMP too

## Screenshots
Before:
image

After:
image

## Tested in CODE?
Nope


desbo (approved):
👍

PR #15642, opened 2017-01-20
Acquisitions A/B test wrapper

desbo:
## What does this change?
This tool prevents multiple acquisitions (formerly contributions) tests
clashing with each other by working out which one should be active. It
is not enough for us to rely on audience segments since we might
have multiple tests running against the same segments, but targeted at
different tags via the targeting tool. 

I had to move a small part of `ab.js` (the bit that works out the variant
based on the MVT ID) into its own module to use it here.

No clashing tests

No

### DIFF
--- a/static/src/javascripts-legacy/projects/common/modules/experiments/acquisition-test-selector.js
+++ b/static/src/javascripts-legacy/projects/common/modules/experiments/acquisition-test-selector.js
@@ -1,0 +1,27 @@
+define([
+    'common/modules/experiments/segment-util',
+    'common/modules/experiments/tests/contributions-epic-brexit',
+    'common/modules/experiments/tests/contributions-epic-always-ask-strategy',
+], function (
+    segmentUtil,
+    brexit,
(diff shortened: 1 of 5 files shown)


static/src/javascripts-legacy/projects/common/modules/experiments/acquisition-test-selector.js
```diff
@@ -0,0 +1,27 @@
...
+    /**
+     * acquisition tests in priority order (highest to lowest)
+     */
+    var tests = [alwaysAsk, brexit];
+
+    return {
+        getTest: function() {
+            var eligibleTests = tests.filter(function (test) {
+                var t = new test();
+                var forced = new RegExp('^#ab-' + t.id).test(window.location.hash);
```
alexduf: I think you may as well test for equality here as the regexp doesn't add that much value (as far as I understand)

alexduf (approved):
Looks good to me

PR #15644, opened 2017-01-20
Fix amp comment validation

michaelwmcnamara:
## What does this change?
This fixes https://trello.com/c/BkhPXBsH/166-amp-comment-embeds-fail-validation.

There is a fix to amp validation of comment embeds in this pull request and also a related fix to display of comment embeds in amp.
Example page containing comment embeds: standard version: https://www.theguardian.com/commentisfree/live/2016/nov/25/are-we-too-quick-to-blame-brexit-for-everything-join-our-live-look-at-the-week

amp version:
https://amp.theguardian.com/commentisfree/live/2016/nov/25/are-we-too-quick-to-blame-brexit-for-everything-join-our-live-look-at-the-week#development=1

Fix validation of amp comment embeds:
Comment embeds currently fail validation due to the avatar image on comments which displays as having an 'img' html tag, which is disallowed by amp.
The change adds a step to ampEmbedCleaner.scala to replace any elements of type 'img' with an equivalent 'amp-img' element.

Fix display of comment embeds in amp.
Currently on the standard site, comment embeds look like this:
image

However in amp, the css that handles display of comment embeds is missing, meaning they look like this:
image

This change adds a cut-down, modified version of the css for comment embeds to amp. As a result, the embeds render properly:
image

This change only impacts Amp.

Tested on local - display looks good. Validation passes. All ./sbt tests pass. All make validate-amp steps pass.

cc @guardian/dotcom-platform


jfsoul:
How come the new design has the first line indented - is that intentional?

zeftilldeath:
Yep agreed with @jfsoul, If we can clear a bit of space so that float doesn't affect the comment text then it's great.

common/app/views/support/cleaner/AmpEmbedCleaner.scala
```diff
@@ -242,6 +244,30 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
...
+          "height" -> image.attr("height"),
+          "width" -> image.attr("width"),
+          "alt" -> image.attr("alt"),
+          "layout" -> "fixed")
+        attrs.foreach {
+          case (key, value) => ampImg.attr(key, value)
+        }
+        val imageParent = image.parent()
+        image.remove()
+        imageParent.appendChild(ampImg)
```
jfsoul: Can we do `image.replaceWith(ampImg)` instead of these 3 lines?

common/test/views/support/cleaner/AmpEmbedCleanerTest.scala
```diff
@@ -30,6 +33,19 @@ class AmpEmbedCleanerTest extends FlatSpec with Matchers {
     document
   }
 
+  private def article() = {
+    val contentApiItem = contentApi()
+    val content = Content.make(contentApiItem)
+    Article.make(content)
+  }
+
+  private def contentApi() = ApiContent(
```
jfsoul: can this be a `val` instead

common/test/views/support/cleaner/AmpEmbedCleanerTest.scala
```diff
@@ -346,16 +390,18 @@ class AmpEmbedCleanerTest extends FlatSpec with Matchers {
...
-    val contentApiItem = contentApi()
-    val content = Content.make(contentApiItem)
-    Article.make(content)
+  /*
+  * Comments cleaner
+  */
+
+  "AmpEmbedCleaner" should "change the avatar img in a comment to be an amp-img" in {
+    val result = cleanDocumentWithCommentEmbed()
+    result.getElementsByTag("amp-img").size should be(1)
```
jfsoul: Is there anything else we should be checking here that are mandatory for amp? E.g. height and width attributes are present

common/test/views/support/cleaner/AmpEmbedCleanerTest.scala
```diff
@@ -64,6 +80,34 @@ class AmpEmbedCleanerTest extends FlatSpec with Matchers {
...
+          </div>
+          <div class="d2-permalink">
+          </div>
+          <div class="d2-body" itemprop="text">
+            <p>foo bar foo foo bar</p>
+          </div>
+        </div>
+      </figure>
+      </body>
+      </html>.toString()
```
jfsoul: At what point does one of these documents get big enough that it should be an external file that we load in? (I think this might be past that point 😄 )

common/test/views/support/cleaner/AmpEmbedCleanerTest.scala
```diff
@@ -64,6 +80,34 @@ class AmpEmbedCleanerTest extends FlatSpec with Matchers {
...
+    val document = Jsoup.parse(StringEscapeUtils.unescapeXml(doc))
```
jfsoul: Why do we need to unescape xml for this one?

static/src/stylesheets/amp/_d2-comments.scss
```diff
@@ -0,0 +1,55 @@
...
+    font-family: $f-serif-text;
+
+    @include mq(tablet) {
+        font-size: 1.28571428571em;
+    }
+}
+
+.d2-permalink .d2-datetime {
+    color: $neutral-2;
+    font-size: .85714285714em;
```
jfsoul: 😨

common/test/views/support/cleaner/AmpEmbedCleanerTest.scala
michaelwmcnamara: It's parsing the doc as XML, so when you do "toString()" on it, it ads a whole lot of XML code in to the string (ie '<' becomes '&lt;'). This then caused problems with Jsoup, when it was parsed.
Maybe just convert all the test data back to strings. The display is noce and readable this way, but on reflection, it is a bit weird having to convert it back from XML.

static/src/stylesheets/amp/_d2-comments.scss
michaelwmcnamara: This is a cut-down copy of the /static/src/stylesheets/module/external/_d2-comments.scss . 
Not sure why it is this way, but I've checked with Zef and we can modify this to 0.8 em instead.

common/test/views/support/cleaner/AmpEmbedCleanerTest.scala
michaelwmcnamara: Keeping the format as its lovely and readable, but will make it consistent by applying the unescapeXml to all of the test data formatted in this way.

static/src/stylesheets/amp/_d2-comments.scss
```diff
@@ -0,0 +1,50 @@
...
+    border-left: 8px solid $neutral-5;
+    clear: left;
+}
+
+.d2-avatar-image {
+    float: left;
+    margin-right: 15px;
+    margin-bottom: 12px;
+
+    @include mq(tablet) {
```
NataliaLKB: Amp doesn't go up to tablet, so this css is probably not needed.

static/src/stylesheets/amp/_d2-comments.scss
```diff
@@ -0,0 +1,50 @@
+/* Comments
+   ========================================================================== */
+
+.d2-comment-embedded {
+    background-color: #ffffff;
+    font-family: sans-serif;
+    overflow: hidden;
+    position: relative;
+    margin-bottom: 1.5em;
+    padding-top: 15px;
```
NataliaLKB: The padding and margin are not part of our grid system, any reason for using these magic numbers instead of something from https://github.com/guardian/guss-grid-system

static/src/stylesheets/amp/_d2-comments.scss
```diff
@@ -0,0 +1,50 @@
...
+}
+
+.d2-right-col {
+    position: relative;
+    top: -3px;
+    padding-right: 16px; // prevent colliding with badge
+}
+
+.d2-username {
+    font-family: $f-serif-text;
```
NataliaLKB: We can't use our normal fonts for AMP. Best to stick with a fallback. If a font is really important I would chat with @zeftilldeath

static/src/stylesheets/amp/_d2-comments.scss
```diff
@@ -0,0 +1,50 @@
...
+}
+
+.d2-permalink .d2-datetime {
+    color: $neutral-2;
+    font-size: .8em;
+    font-family: $f-sans-serif-text; // Force font when used outside of comment context
+}
+
+.d2-permalink .d2-datetime:hover {
+    border-bottom: 1px dotted #757575;
```
NataliaLKB: Why `#757575` and not something from our colour palette http://guardian.github.io/pasteup-palette/demo/

static/src/stylesheets/amp/_d2-comments.scss
michaelwmcnamara: Hi, thanks for pointing that out. I copied all this across from the settings used in the non-amp version of this css file. Not sure why the grid wasn't used originally. I can definitely update this file to use the grid settings now and can update the non-amp version of this file in a future pr.

static/src/stylesheets/amp/_d2-comments.scss
michaelwmcnamara: Excellent! the more I can rip out the better :-)

static/src/stylesheets/amp/_d2-comments.scss
NataliaLKB: I pulled this down and ran it so I could see the hover states, it doesn't look great to me because the lines don't seem to match up. This is what it looks like:

image

I would opt for removing it anyway because AMP is only really for mobile (no hover)

static/src/stylesheets/amp/_d2-comments.scss
```diff
@@ -0,0 +1,50 @@
+/* Comments
+   ========================================================================== */
+
+.d2-comment-embedded {
+    background-color: #ffffff;
```
NataliaLKB: Doesn't seem to be needed

static/src/stylesheets/amp/_d2-comments.scss
```diff
@@ -0,0 +1,50 @@
...
+    font-family: sans-serif;
+    overflow: hidden;
+    position: relative;
+    margin-bottom: 1.5em;
+    padding-top: 15px;
+    font-size: 14px;
+    padding-left: 28px; // aligns with top level comment's body
+    border-left: 8px solid $neutral-5;
+    clear: left;
```
NataliaLKB: Not sure if this or `position: relative` is needed.

static/src/stylesheets/amp/_d2-comments.scss
NataliaLKB: Best to consult @zeftilldeath if it doesn't look right with grid values, it may be belong outside the grid. A note would be good if this is the case.

For the most part there has been a slightly simpler design on AMP to keep the css down. It is completely fine for the site and amp to have different styling as long as it still reflects the brand.

NataliaLKB:
Tomorrow I am happy to work with you to figure out what else we can change with regards to reason why the build is failing (too much css! ) :)

static/src/stylesheets/amp/_d2-comments.scss
michaelwmcnamara: I see what you mean - doesnt seem to have any impact when I untick it in the debugger.
Ace - I'll strip it out :-)

michaelwmcnamara:
@NataliaLKB  Thanks - that would be a massive help!

static/src/stylesheets/amp/_d2-comments.scss
michaelwmcnamara: Thanks - thats a good point. Not very easy to hover on a phone! I'll strip this one too. :-)

static/src/stylesheets/amp/_d2-comments.scss
michaelwmcnamara: Thanks - the rest of the embed is using sans-serif. I'll strip this out & it should inherit the sans-serif font form its parent. I'll check with @zeftilldeath that it looks ok.

static/src/stylesheets/amp/_d2-comments.scss
michaelwmcnamara: Thanks, thats good to know. I've had a crack at applying the grid values. It seems to look ok, but I'll run it by @zeftilldeath to be sure.

common/test/views/support/cleaner/AmpEmbedCleanerTest.scala
michaelwmcnamara: I've reduced the size of this html document to more closely match other docs in the test file. Hopefully this eliminates the need of making it an external file, but let me know if you think otherwise.

static/src/stylesheets/amp/_d2-comments.scss
michaelwmcnamara: As per Natalia's suggestion, this now uses the grid values, making it consistent with other amp pages.

static/src/stylesheets/amp/_d2-comments.scss
michaelwmcnamara: Thanks! The background color has now been removed.

common/test/views/support/cleaner/AmpEmbedCleanerTest.scala
```diff
@@ -56,14 +86,41 @@ class AmpEmbedCleanerTest extends FlatSpec with Matchers {
...
     </html>.toString()
-    val document: Document = Jsoup.parse(StringEscapeUtils.unescapeXml(doc))
+    val document: Document = parseTestData(doc)
+    clean(document)
+  }
+
+
+  private def cleanDocumentWithCommentEmbed(className: Option[String], src: Option[String], width: Option[String], height: Option[String], alt: Option[String]): Document = {
+
+    val classString = if(className.nonEmpty){s"""class=\"${className.get}\" """}else{""}
```
NataliaLKB: Just because I don't know, what are the `"""` for?

NataliaLKB: It must be to escape the `"`. Why not just change the `"` to `'`? It would read nicer wouldn't it?

common/test/views/support/cleaner/AmpEmbedCleanerTest.scala
```diff
@@ -22,6 +25,19 @@ class AmpEmbedCleanerTest extends FlatSpec with Matchers {
   val soundcloudUrlV1 = "http://api.soundcloud.com%2Ftracks%2F1234"
   val soundcloudUrlV2 = "https://w.soundcloud.com/player/?url=https%3A//api.soundcloud.com/tracks/1234"
   val soundcloudUrlNoTrackId = "https://w.soundcloud.com/player/?url=https%3A//api.soundcloud.com/foobar"
+  val commentAvatarClass = Some("d2-avatar")
```
NataliaLKB: Why make these a Some, when you are just checking whether the string is empty or not later on? Why not just make them a string?

common/test/views/support/cleaner/AmpEmbedCleanerTest.scala
TBonnin: Triple quote are quite useful to escape characters and write multi-lines strings. 
Therefore we don't need to escape `\"`
Also when using one line `if` the convention is not to use curly brackets. ex: `if(isX) "my string" else ""

NataliaLKB (approved):
lgtm 👍

common/test/views/support/cleaner/AmpEmbedCleanerTest.scala
michaelwmcnamara: @NataliaLKB @TBonnin  Hi folks, Thanks very much for the suggestions! I've removed the triple quotes and replaced the escaped quotes with single-quotes. Also have now removed curly brackets. Let me know if this is more readable.

common/test/views/support/cleaner/AmpEmbedCleanerTest.scala
michaelwmcnamara: @NataliaLKB Thanks, thats a good point! No real need for Options here at all.
I've removed the "Some" from these in the latest commit.

PR #15646, opened 2017-01-20
remove placeholder switch - make placeholder attribute permanent

michaelwmcnamara:
## What does this change?
We were trialling the use of a placeholder attr in amp interactive embeds.
This allows the interactive to be put in the main media.
The trial was successful, so removing the switch and making this change permanent.

This allows interactive atoms to be embedded in main media without breaking amp-validation

Exclusively for Amp. No other impact.

## Tested in CODE?
Nope

Amp pages validate successfully (manual checks and tested using make validate-amp).
Amp Embed Cleaner tests all pass too.
cc @guardian/dotcom-platform

### DIFF
--- a/common/app/conf/switches/FeatureSwitches.scala
+++ b/common/app/conf/switches/FeatureSwitches.scala
@@ -487,17 +487,6 @@ trait FeatureSwitches {
     exposeClientSide = true
   )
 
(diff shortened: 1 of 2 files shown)


NataliaLKB (approved):
👍

PR #15647, opened 2017-01-20
Removing unused router supbproject

TBonnin:
## What does this change?
Removing unused router supbproject.
Router now lives in the `platform` repo

## Tested in CODE?
No


jfsoul (approved):
💯

PR #15653, opened 2017-01-23
Add webpack/curl to ga experience

sndrs:
## What does this change?

Adds webpack experience value to #15624 

we can compare webpack and curl in GA

no


gtrufitt (approved):
👍 great

PR #15657, opened 2017-01-23
Outbrain fixes

rich-nguyen:
We had an issue with the move to client side in https://github.com/guardian/frontend/pull/15588 . The old placeholder does not appear on every page, so this change explicitly avoids loading Outbrain based on the content type (article only).

A few server-side tidy ups too, with some fields and templates left unused since the move.

## Tested in CODE?
Yes

### DIFF
--- a/common/app/model/Formats.scala
+++ b/common/app/model/Formats.scala
@@ -226,7 +226,6 @@ object ContentTypeFormat {
     showByline: Boolean,
     hasStoryPackage: Boolean,
     rawOpenGraphImage: String,
-    showFooterContainers: Boolean,
     atoms: Option[Atoms])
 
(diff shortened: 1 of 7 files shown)


JonNorman (approved):
Yaaaaaaas!

JonNorman:
The failing test seems to rely on the `config.isPreview` being changed rather than the `commercialFeatures` object we now solely use. I suppose we can delete some of these tests...

rich-nguyen:
[Was the Outbrain widget running on the Trump inauguration live blog?](https://trello.com/c/Aq7qachh/129-was-the-outbrain-widget-running-on-the-trump-inauguration-live-blog)

PR #15658, opened 2017-01-23
Remove rogue <a> wrapping a <table>

joelochlann:
In #15634, I wrapped individual inline elements (images, headlines, standfirsts) with `<a>` tags rather than block-level tables. I accidentally left in place an `<a>` wrapping a table, meaning nested `<a>` tags (since that table contains a headline). Strangely this didn't seem to break Outlook (presumably the block-level `<a>` gets ignored) but I thought it best to remove it anyway!


lmath (approved):
yeah, good shout 👍

PR #15659, opened 2017-01-23
AB tests: Make the JS expiry consistent with the Switch expiry

katebee:
#### Fun fact:

> The expiry in JS requires the day to have started, where as the Scala switches require the day to have passed.

This currently means that the JavaScript of an AB test will stop running at 00:00:00 on the **DAY** of the expiry...

The Expiring Feature Switches email will tell you this is `Expiring today`, however, this test has already secretly stopped running:

picture 496

When the client currently runs the code, the value from `new Date()` is something like:
 `Tue Jan 24 2017 16:15:48`

Which would be fine, but `test.expiry` doesn't have a time, just a date, so it defaults the time to `00:00:00`. This means that the `var expired` for a test **expiring today**, would become: 
`Tue Jan 24 2017 00:00:00`

... this makes line 92 of the current code something like:

`(Tue Jan 24 2017 16:15:48) - (Tue Jan 24 2017 00:00:00)` (equals 58548041)
`58548041 > 0` (TRUE)

I'm using setHours to basically make the time of day no longer apply, so the comparisons become:
`Tues Jan 24 2017 00:00:00` > `Tues Jan 24 2017 00:00:00` (FALSE)
`Tues Jan 25 2017 00:00:00` > `Tues Jan 24 2017 00:00:00` (TRUE)

JavaScript Date/Time... yaaaaayyy~

- Ensures all tests will run until the **END** of the expiry day
- Make the expiry of the JavaScript consistent with the expiry of the AB switch
- Expiring Feature Switches email stops fibbing to you
- Test coverage for cases of 'expires today' in `getActiveTests`, `getExpiredTests` and `testCanBeRun`

Nope


katebee:
Tagging a few people in who have recently worked on AB tests... 

Comment from anyone else also welcome 😄

static/src/javascripts-legacy/projects/common/modules/experiments/ab.js
```diff
@@ -87,9 +87,12 @@ define([
     }
 
     function getActiveTests() {
-        var now = new Date();
+        var today = new Date();
+        // new Date(test.expiry) sets the expiry time to 00:00:00
+        // Using SetHours allows a test to run until the END of the expiry day
+        today.setHours(0,0,0,0);
```
rtyley: Is this setting the time forward or backward? It sets the time to the _start_ of today, ie pushes the time earlier, right?

static/src/javascripts-legacy/projects/common/modules/experiments/ab.js
```diff
@@ -87,9 +87,12 @@ define([
...
         return TESTS.filter(function (test) {
-            var expired = (now - new Date(test.expiry)) > 0;
+            var expired = today > new Date(test.expiry);
```
rtyley: I'm super-lazy, I haven't run this. But if `today` is now at the 'beginning' of the day, doesn't this mean this test will expire at the _start_ of `today` - aren't we aiming for it to expire at the end?

rtyley (commented):
I'm not certain, but I think the date logic is different to the stated aim in the PR description ("Ensures all tests will run until the END of the expiry day")

static/src/javascripts-legacy/projects/common/modules/experiments/ab.js
katebee: When the client currently runs the code, the value from `new Date()` is something like:
 `Tue Jan 24 2017 16:15:48`

Which would be fine, but `test.expiry` doesn't have a time, just a date, so it defaults the time to `00:00:00`. This means that the `var expired` for a test **expiring today**, would become: 
`Tue Jan 24 2017 00:00:00`

... this makes line 92 something like:

`(Tue Jan 24 2017 16:15:48) - (Tue Jan 24 2017 00:00:00)` (equals 58548041)
`58548041 > 0` (TRUE)

I'm using setHours to basically make the time of day no longer apply, so the comparisons become:
`Tues Jan 24 2017 00:00:00` > `Tues Jan 24 2017 00:00:00` (FALSE)
`Tues Jan 25 2017 00:00:00` > `Tues Jan 24 2017 00:00:00` (TRUE)

JavaScript Date/Time... yaaaaayyy~

rtyley: Yaaay arrg! Sounds like this code will work perfectly, thanks for explaining it to me!

I'd say is that this inequality is a little non-obvious, because we're comparing two values that both correspond to 24 hours earlier than the threshold we actually care about (ie when the test is supposed to to end!). I wonder if there is any way we could guard against someone else coming up against this code and being confused by it? I thought about:

* a test in [ab.spec.js](https://github.com/guardian/frontend/blob/ec6a89cc8/static/test/javascripts/spec/common/experiments/ab.spec.js) - but producing a good test would probably involve some kind of injectable clock (ie one you could set `now` to be an arbitrary fixed point in time)
* verbose variable naming, eg : `var startOfToday = new Date();` and `var startOfLastDayOfTest = new Date(test.expiry)`

These are just ideas, I don't want to hold you up on getting this important fix out! So :+1: in any case

katebee: Added quite a few changes and extended the fix to the other functions checking expiry 😁 

- Found a robust enough way to test this 🎉 
- Using a verbose variable name

static/src/javascripts-legacy/projects/common/modules/experiments/ab.js
```diff
@@ -105,14 +104,15 @@ define([
...
         return TESTS.filter(function (test) {
-            return (now - new Date(test.expiry)) > 0;
+            return (startOfToday > new Date(test.expiry));
         });
     }
 
     function testCanBeRun(test) {
-        var expired = (new Date() - new Date(test.expiry)) > 0,
+        var startOfToday = new Date().setHours(0,0,0,0);
+        var expired = (startOfToday > new Date(test.expiry)),
```
desbo: i would create a separate `isExpired` function and use it here and in the filter

desbo (approved):
apart from the minor comment, looks sensible to me

jfsoul:
Apologies for jumping into this PR late.  While I agree that it can be a source of confusion that the test and switch expiry are 24 hours apart despite being set to the same date, I would argue it should stay that way (or equivalent).

My concern with having the expiry at the same time (and possibly the reason it is this way in the first place) is that devs will either:

1. Wait until their both test and switch have expired before removing the switch (leading to lots of broken master builds 😭 )
2. Remove the switch early to avoid breaking the build.  For a lot of tests I don't see a particular problem with this, but for tests that are intended to run for exactly x days, it means you have to make sure your switch expiry date is at least a day after the test expiry date.

I think 2. would work, but for the sake of consistency I would suggest we recommend to everyone to set their switch expiry after the test expiry - even if they are happy for the test to end whenever their removal is deployed.  I'm not sure whether this is less confusing or not.

Let me know your thoughts, happy to be convinced either way.

cc: @guardian/dotcom-platform

katebee:
@jfsoul this discrepancy isn't by design; fix is inspired by: https://github.com/guardian/frontend/pull/15118#issuecomment-262961230

> lots of broken master builds

CI/CD means we are "Not wrong for long" Yes? 😁 Broken masters due to switches already happen, albeit rarely. Someone will generally extend the switch quickly when this does happen. 

> Wait until their both test and switch have expired before removing the switch

No-one wants to be the owner of a switch that is breaking master for everyone! Devs should be aware that their tests need deleting/updating from the daily 'expiring tests' email. 

> Remove the switch early to avoid breaking the build

Best practice would be for devs to never let a switch get to `Already expired`, and I believe this is what everyone aims for; expired switches in master are not intentional.

With no automated reminders linked to the JS expiry, I expect that the process for updating/deleting switches would remain the same, as would the incidence of broken master builds due to switches.

Sometimes, self-expiring code is the best way to ensure action is taken. 🙈 🙉 🙊 

`Expiring today` gives developers a reminder and (an assumed) grace period to make a PR. Most people seem to delete/update their tests **on the expiry date** :man_shrugging: ...I have been caught out on the last day of a test, thinking I had the day to make a PR and extend a test, when instead the JS was silently 'off' already. I'm certain that others have, and will be, caught out too.

> recommend to everyone to set their switch expiry after the test expiry

This could work and is a good idea irrespective of this PR; however, it seems counterintuitive to have two dates, particularly if they are 'off-by-one' so we would need to document and communicate why.

> for tests that are intended to run for exactly x days, it means you have to make sure your switch expiry date is at least a day after the test expiry date

Giving the expiries the same behaviour doesn't mean that they have to be the same date. I just want some consistency, and for the behaviour of ab.js to align with my expectations: sell-by-date, use-by-date, expiry-date... generally these would be all be *inclusive* of the stated date.

We use AB tests to guide many business and feature decisions; they should be as rigorous as possible; an AB test not running when expected is a pitfall.

jfsoul:
I wouldn't categorise expired switches as a rarity, we often have to chase up on these - even when they haven't yet broken the build.  The scenario we worry about is an emergency fix that needs to go out after midnight but before the switch owner has had a chance to remove it in the morning.  This is rare, but not a situation we want to find ourselves in.  We are already running this risk, so I am wary of any changes that could inadvertently lead to more expired switches.

I think the confusion stems from the assumption that switch and test expiry are the same (which is exacerbated by the switch expiry emails).  I am not against this change, but I think we have a communication problem either way.  We can either communicate that AB tests expire at the beginning of the day of expiry but not switches; or that AB test expiries should be set earlier than the switch expiry - unless you intend to end the test early.

I agree that we should strive to be rigorous with our AB tests, but I actually think it would be less rigorous if we were to expect people to end their tests "early" by default.  I think the best outcome is for tests to start and end at midnight by default, with a grace period of at least a day after that to turn the switch off.  We can of course achieve that with or without this PR, it just requires better communication.

NataliaLKB:
@katebee I believe that this is actually by design. I agree with @jfsoul that I am not sure this is the answer to the problems you stated in your description:

> Ensures all tests will run until the END of the expiry day

I would always read the expiry day as the day the switch expires, not at the end of the day the switch expired. I could be wrong, but I do believe this to be the intended meaning.

> Make the expiry of the JavaScript consistent with the expiry of the AB switch

Ideally you should actually be able to let the test expire without the build breaking. This grace period is the 24 hours between JS test expiry, and switch expiry.

> Expiring Feature Switches email stops fibbing to you

This is a problem. Clearly we are not clear enough in the email that goes out. We should probably extend it to explain that the test will expire on the day (the start of the day) and that there is a grace period to deal with the switches so you don't inconvenience other people by breaking the build. 

> Test coverage for cases of 'expires today' in getActiveTests, getExpiredTests and testCanBeRun

👍 👍 👍 👍 🎉  Yay tests!

katebee:
> the assumption that switch and test expiry are the same

Absolutely, and thinking about it, they should not be! I have written a lot of AB tests into frontend, and for the first six months I always assumed that the JS expiry was a **inclusive** date.

I still believe that an "expiry" should be at the **end** of the stated day. The expired switches problem feels like a separate process issue. However, a silent/early JS expiry is not the best way to adjust behaviour:

1. The early expiry of the JS is too quiet and insidious to remind people to delete a test
2. There is no reminder for the JS expiry - we are relying on human discipline (which is fallible) 
3. We have a process issue if expiring switches are often left to the last day

I would rather:

a. A change process, whereby the switch expiry is a week after the test expiry, then we encourage devs to deal with their switches as soon as they appear under `Expiring within a week`. Switches going yellow or red should raise alarms.
b. Renaming or better documenting the `test.expiry` so everyone is aware that the test will not run on that day.
c. If the failing switch test and broken master is an issue, is it better to delete the test? What is the purpose of the test if not to force someone to deal with expiring code?

(to be clear, I am not a fan of breaking master!)

NataliaLKB:
@jfsoul and I have been chatting and here are our thoughts

> I have written a lot of AB tests into frontend, and for the first six months I always assumed that the JS expiry was a inclusive date.

This is a great point. Completely agree that this isn't great. This should be renamed to be clearer

> a. A change process, whereby the switch expiry is a week after the test expiry, then we encourage devs to deal with their switches as soon as they appear under Expiring within a week. Switches going yellow or red should raise alarms.

I agree that switches in the yellow or red should raise alarms, but at the moment we do try and encourage people to deal with their switches as soon as they are under Expiring within a week and that doesn't always happen. But a lot of people do deal with their switches the first time its mentioned.

> b. Renaming or better documenting the test.expiry so everyone is aware that the test will not run on that day.

Both those things really need to be done! 
* `test.expiry` should definitely be renamed to be more clear. We would suggest something like `test.validUntilBeginningOf` or `test.expiryBeginningOf` (suggestions welcome)
* An email should go around after this change explaining why this has changed, and clarify any confusion.
* Docs should be changed to be more clear about this

> c. If the failing switch test and broken master is an issue, is it better to delete the test? What is the purpose of the test if not to force someone to deal with expiring code?

It is a good point that failing builds can cause problems, but we also need a way to force people to deal with expired switches. It helps stop unneeded code from lingering in the codebase. We had this problem last year which I believe is why this test was added. cc @rich-nguyen @johnduffell who had lots of discussion on this topic.

We completely agree that the process could be improved for clarity. Suggestions include:
* Changing the expiry email to be more clear that this is about switch expiry (not just ab-tests) and that they expire at the end of the day
* A line in the email should be added to also explain the the JS ab tests expiry are technically independent of the switch expiry and are usually 24 hours before.
* We should also change our process so that the switch expiry email is sent again to the dotcom platform team in the afternoon so we are more aware of who and who hasn't dealt with switches. We can then be a bit more proactive with following up with people to prevent the build breaking the next day.

katebee:
@NataliaLKB, @jfsoul 👍 on the thoughts, thank you for the discussion!

The JS expiry may also be a pitfall for people who are wanting to extend an AB test, as the test will stop running on that day. For example:

[editorial-email variants](https://github.com/guardian/frontend/blob/master/static/src/javascripts-legacy/projects/common/modules/experiments/tests/editorial-email-variants.js) - we have extended this test [before](https://github.com/guardian/frontend/pull/15544), and may need to again - @davidfurey? 

If the test doesn't get extended until the **day** of the `expiry`, then all email signups from that day will not be assigned to a variant until the `test.expiry` gets updated/extended.

If devs wait until the expiry day of their test to extend it, they may have a blip when their test stops running for a day. Not sure how much of an issue this is to everyone; this PR was intended to prevent anyone else running into the same issue I had 😄 

> It helps stop unneeded code from lingering in the codebase

💯 on expiring-code 👍 👍 👍 🎉

davidfurey:
Not a blocker for this PR, but since Jon brought up midnight - have we not been bitten by tests expiring at midnight before, would it not be better if tests had to expire during working hours?

johnduffell:
don't invite too many people to the PR of you'll never ship it! 😝 

Anyway my thoughts:

1. it's not by design - actually switches used to expire at 0:01 or something, and I sneaked a PR through ages ago to change it to the end of the day 😉 

2. Consistency is good as coded in this PR.  If we really want to have different behaviour it should not be in a way that surprises people (I'm not recommending it, but one possible way of doing that would be we report switches as expired but don't actually break the build until a switch is 1 day out of date)  Therefore I am happy to give this a 👍 

3.
> Best practice would be for devs to never let a switch get to Already expired, and I believe this is what everyone aims for; expired switches in master are not intentional.

personally I think best practice is never let switches get to expiring in the next couple of days.  If you really need to deal with it on the day, you should be setting an expiry around a week later.  Then you'll never be that person who let a switch expire on everyone.
For AB tests that means if your ab test ends on the 10th, you set the switch for 17th.

PR #15660, opened 2017-01-23
Turn off end slate for in-body YouTube atom embeds and don't show YT player on small cards

gidsg:
## What does this change?
This changes the *show endslate* and *show video* conditions to be closer to that applied to non-YouTube atoms.

However, the *show video* condition for YouTube atoms is set to false for smaller cards in order to meet the YouTube PFP minimum size requirements.

Makes end-slate behaviour consistent with non-youtube videos.
Make small cards non-playable (Required to meet minimum player size requirement for YouTube PFP.)

No

## Screenshots
Before:
screen shot 2017-01-23 at 14 04 41

After:
screen shot 2017-01-23 at 14 02 09

### DIFF
--- a/article/app/views/package.scala
+++ b/article/app/views/package.scala
@@ -29,7 +29,7 @@ object MainCleaner {
         if (amp) AmpEmbedCleaner(article) else VideoEmbedCleaner(article),
         PictureCleaner(article, amp),
         MainFigCaptionCleaner,
-        AtomsCleaner(article.content.atoms, shouldFence = true, amp)
+        AtomsCleaner(article.content.atoms, shouldFence = true, amp, mainMedia = true)
       )
   }
 }
--- a/article/app/views/package.scala
+++ b/article/app/views/package.scala
@@ -29,7 +29,7 @@ object MainCleaner {
         if (amp) AmpEmbedCleaner(article) else VideoEmbedCleaner(article),
         PictureCleaner(article, amp),
         MainFigCaptionCleaner,
-        AtomsCleaner(article.content.atoms, shouldFence = true, amp, mainMedia = true)
+        AtomsCleaner(atoms = article.content.atoms, shouldFence = true, amp = amp, mainMedia = true)
       )
   }
 }
@@ -51,7 +51,7 @@ object BodyCleaner {
       TableEmbedComplimentaryToP,
       R2VideoCleaner,
       PictureCleaner(article, amp),
-      AtomsCleaner(article.content.atoms, shouldFence = true, amp = amp),
+      AtomsCleaner(atoms = article.content.atoms, shouldFence = true, amp = amp),
       DropCaps(article.tags.isComment || article.tags.isFeature, article.isImmersive),
       ImmersiveHeaders(article.isImmersive),
       FigCaptionCleaner,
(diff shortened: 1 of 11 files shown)


common/app/views/fragments/items/facia_cards/contentCard.scala.html
```diff
@@ -88,10 +88,10 @@
                 </div>
             }
 
-            case Some(media@InlineYouTubeMediaAtom(_)) => {
+            case Some(media@InlineYouTubeMediaAtom(_)) if item.cardTypes.showYouTubeMediaAtomPlayer => {
```
GHaberis: So if showYouTubeMediaAtomPlayer is false we don't render the youtube template fragment at all? I wonder if we could still use this template but pass showYouTubeMediaAtomPlayer as an argument. We can then put the logic to check this argument in that template; if true we add the iframe and if false we add the play button and overlay (which link through to the video page). +'ve is It'd save us having a new template? -'ve is more logic in that template.

gidsg: Yes, that would be possible but we'd need to design that template, so I thought it better to do that in a subsequent PR.

GHaberis: Yeah that's fine, that's probably how I'd do it but we can make that change in the following PR.

gidsg: Let's see how much the template varies from the current playable one. It may be cleaner to make it its own fragment like `posterOnly` case.

GHaberis (commented):
@gidsg this look good to me. It'd be good for someone more Scala orientated like @TBonnin from frontend to check on this too?

article/app/views/package.scala
```diff
@@ -29,7 +29,7 @@ object MainCleaner {
         if (amp) AmpEmbedCleaner(article) else VideoEmbedCleaner(article),
         PictureCleaner(article, amp),
         MainFigCaptionCleaner,
-        AtomsCleaner(article.content.atoms, shouldFence = true, amp)
+        AtomsCleaner(article.content.atoms, shouldFence = true, amp, mainMedia = true)
```
TBonnin: Can we either have named parameters everywhere or no at all? You can put one param per line

common/app/views/fragments/atoms/youtube.scala.html
```diff
@@ -4,16 +4,16 @@
...
         "u-responsive-ratio" -> true,
         "u-responsive-ratio--hd" -> true,
         "youtube-media-atom" -> true
     ))"
-      @for(endSlate <- media.endSlatePath) {
-        data-end-slate="@endSlate"
-        }
+    @if(displayEndSlate && media.endSlatePath.isDefined) {
+      data-end-slate="@media.endSlatePath.get"
+    }
```
TBonnin: I think you can do something like this
```
@for {
   endSlate <- media.endSlatePath if(displayEndSlate)
}
```

TBonnin (changes requested):
I made 2 tiny comments but lgtm

TBonnin (approved):

PR #15661, opened 2017-01-23
Membership bundles thrasher AB test

JustinPinner:
## What does this change?
Adds an AB test to explore reader appetite for membership bundles. By default the thrasher we've defined will be invisible. If the reader is in one of the variant groups, it'll be displayed.

This will be applicable on the UK front, by UK edition readers and for desktop web only.
 
**Note:** there's no control group here as we're not trying to determine whether the feature should be added to the site, we're testing which of two initial possibilities is the best received.

It'll inform the decision-making process behind the construction of bundles, making sure we don't invest effort unless we know it's valuable.

No, web only.

## Screenshots
(updated 26 Jan, captured from CODE env)
screen shot 2017-01-26 at 12 16 33

With the test turned off
screen shot 2017-01-23 at 15 22 06

## Tested in CODE?
Yes.


static/src/javascripts-legacy/projects/common/modules/experiments/tests/membership-ab-thrasher.js
```diff
@@ -0,0 +1,99 @@
...
+        this.description = 'Test appetite for membership bundles';
+        this.showForSensitive = true;
+        this.audience = 0.05;  // 5% ~ 7,000 unique browser thrasher views per variant per day
+        this.audienceOffset = 0;
+        this.successMeasure = '';
+        this.audienceCriteria = 'People on UK network front with at least an 1140px wide display.';
+        this.dataLinkNames = '';
+        this.idealOutcome = 'One landing page will accrue a higher number of intention interactions than the other';
+
+        this.canRun = function () {
```
rtyley: Just checked with Charlotte - we want to exclude existing paying members from seeing our test ("otherwise it gets weird!") so you probably want to add `canReasonablyAskForMoney` here, like:

https://github.com/guardian/frontend/blob/43905b6b4b/static/src/javascripts-legacy/projects/common/modules/experiments/tests/membership-engagement-banner-tests.js#L40

static/src/javascripts-legacy/projects/common/modules/experiments/tests/membership-ab-thrasher.js
```diff
@@ -0,0 +1,99 @@
...
+
+        this.thrasher = function() {
+            return document.querySelector('.membership-ab-thrasher--wrapper');
+        };
+
+        this.setCopy = function() {
+            if (this.thrasher()) {
+                var mainTitleEl = document.querySelector('.membership-ab-thrasher_header .main_title');
+                if (mainTitleEl) {
+                    mainTitleEl.innerHTML = '<p>Support<br>the Guardian</p>';
```
rtyley: Is this necessary? Isn't the text already set to that value at https://github.com/guardian/thrashers/blob/76207dabe8f/embeds/membership-ab-thrasher/index.html#L4 ?

static/src/javascripts-legacy/projects/common/modules/experiments/tests/membership-ab-thrasher.js
```diff
@@ -0,0 +1,99 @@
...
+                }
+                var subTitleEl = document.querySelector('.membership-ab-thrasher_header .sub_title');
+                if (subTitleEl) {
+                    subTitleEl.innerHTML = '<p>We\'re introducing <strong>new ways</strong> to support<br>the Guardian\'s quality journalism and independent voice. Choose from digital,<br>print or a contribution today.</p>';
```
rtyley: Ditto here - could we just have this text as part of the default body at https://github.com/guardian/thrashers/blob/76207dabe8f/embeds/membership-ab-thrasher/index.html#L7 ?

static/src/javascripts-legacy/projects/common/modules/experiments/tests/membership-ab-thrasher.js
```diff
@@ -0,0 +1,99 @@
...
+            var thrasherContainer = this.thrasherContainer();
+            if (thrasherContainer) {
+                if (!thrasherContainer.classList.contains('visible')) {
+                    thrasherContainer.classList.add('visible');
+                }
+            }
+        };
+
+        this.setup = function(variant) {
+            this.setCopy();
```
rtyley: Copy doesn't vary by variant, so you can probably get rid of this.

rtyley (changes requested):

static/src/javascripts-legacy/projects/common/modules/experiments/tests/membership-ab-thrasher.js
JustinPinner: This replaces the default copy that is defined in the Thrasher and which will be shown in the event of a CSS / test / other kind of screw up, where the thrasher might possibly be seen by non-test users. So it needs to stay.

static/src/javascripts-legacy/projects/common/modules/experiments/tests/membership-ab-thrasher.js
JustinPinner: Ok

static/src/javascripts-legacy/projects/common/modules/experiments/tests/membership-ab-thrasher.js
JustinPinner: No, the default body text is there as a backstop in the event of something somewhere making the thrasher visible unexpectedly. The default text is lifted from the current supporter sign up page and the button links through to there, so the experience will still make sense if this happens.

rtyley: Oh, ok. That does sound paranoid to me, but I guess if the CSS of https://github.com/guardian/frontend/blob/e8d7496b9/static/src/stylesheets/module/facia/_containers.scss#L65 was to fail in some way the thrasher could be visible. So I guess the `setCopy()` method stays (maybe it could be called `switchFromSafetyFallbackCopy()` to indicate it's purpose), and so does the `subTitleEl` manipulation, but the `mainTitleEl` manipulation doesn't seem to be be necessary, as the replacement text is identical.

Ap0c:
Had a go at running this locally, it appears fine on the UK front and directs to the bundles page for the local version of membership. It also only shows for the 1140px breakpoint and above, as expected.

Checking it out on CODE, the behaviour is the same, except that it directs to the supporter landing page with the correct campaign code. Assuming that there is something in CODE that's preventing the link from being updated, it looks like the fallback behaviour is working as expected. If you're seeing it correctly set the link here, then it's presumably something in my setup, so no worries.

In other words, it seems to be working exactly as planned, nice work! ⛵️ 💥

static/src/javascripts-legacy/projects/common/modules/experiments/tests/membership-ab-thrasher.js
JustinPinner: Done.

rtyley (approved):
`setCopy()` could be called `switchFromSafetyFallbackCopy()` to indicate it's purpose, but apart from that looks good to me!

PR #15662, opened 2017-01-23
Gh on you tube iframe api ready fix

GHaberis:
## What does this change?

Fixes a bug we were seeing where youtube atoms were not being enhanced on pages where another non-youtube video existed which had pre-roll adverts enabled. Now we don't allow pre-roll ads if the page's content contains a youtube atom. 

Consistent loading of enhanced youtube atoms.

No

## Screenshots

N/A

## Tested in CODE?

No

@akash1810


common/app/model/content.scala
```diff
@@ -425,6 +425,15 @@ object Article {
     // `headOption` as the video could be main media or a regular embed, so just get the first video
     val videoDuration = content.elements.videos.headOption.map { v => JsNumber(v.videos.duration) }.getOrElse(JsNull)
 
+   def hasYouTubeAtom: Boolean = {
+     val hasYouTubeAtom: Option[Boolean] = for {
+       atoms <- content.atoms
+       firstMediaAtom <- atoms.media.headOption
+       firstAsset <- firstMediaAtom.assets.headOption
+     } yield firstAsset.platform == "Youtube"
```
akash1810: can we match on the `Platform.Youtube` enum rather than the string?

PR #15664, opened 2017-01-23
Avoid exceptions when the commercial switch is disabled

rich-nguyen:
Make sure the cleaner code doesn't throw an exception in the unlikely event that this switch is off.

## Tested in CODE?
Yes (but missed an error in facia).

### DIFF
--- a/common/app/views/support/HtmlCleaner.scala
+++ b/common/app/views/support/HtmlCleaner.scala
@@ -649,12 +649,14 @@ case class CommercialComponentHigh(isAdvertisementFeature: Boolean, isNetworkFro
 
       val adSlotHtml = views.html.fragments.commercial.commercialComponentHigh(isAdvertisementFeature, hasPageSkin)
 
-      val adSlot: Element = Jsoup.parseBodyFragment(adSlotHtml.toString).body().child(0)
+      val adSlot: Option[Element] = Jsoup.parseBodyFragment(adSlotHtml.toString).body().children().toList.headOption
 
-      containers.lift(containerIndex).map { case ((container, index)) => {
-          container.after(adSlot)
-          adSlot.wrap(s"""<div class="fc-container fc-container--commercial"></div>""")
-        }
+      for {
+        (container, index) <- containers.lift(containerIndex)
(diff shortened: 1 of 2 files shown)


rich-nguyen:
[Error when switching commercial off](https://trello.com/c/hZIxBhjA/122-error-when-switching-commercial-off)

common/app/views/support/HtmlCleaner.scala
```diff
@@ -649,12 +649,14 @@ case class CommercialComponentHigh(isAdvertisementFeature: Boolean, isNetworkFro
...
-      containers.lift(containerIndex).map { case ((container, index)) => {
-          container.after(adSlot)
-          adSlot.wrap(s"""<div class="fc-container fc-container--commercial"></div>""")
-        }
+      for {
+        (container, index) <- containers.lift(containerIndex)
+        slot <- adSlot
+      } {
+          container.after(slot)
+          slot.wrap(s"""<div class="fc-container fc-container--commercial"></div>""")
```
kelvin-chappell: Don't need the 's'?

common/app/views/support/HtmlCleaner.scala
```diff
@@ -649,12 +649,14 @@ case class CommercialComponentHigh(isAdvertisementFeature: Boolean, isNetworkFro
 
       val adSlotHtml = views.html.fragments.commercial.commercialComponentHigh(isAdvertisementFeature, hasPageSkin)
 
-      val adSlot: Element = Jsoup.parseBodyFragment(adSlotHtml.toString).body().child(0)
+      val adSlot: Option[Element] = Jsoup.parseBodyFragment(adSlotHtml.toString).body().children().toList.headOption
```
kelvin-chappell: Does jsoup return nulls for empty lists?

kelvin-chappell (approved):
Just a couple of points.

common/app/views/support/HtmlCleaner.scala
rich-nguyen: no, it throws an exception. :(

PR #15666, opened 2017-01-24
If outbrain container is empty outbrain isn't shown

davidfurey:
It looks like the way outbrain is included has changed, which meant that obWidgetIsShown was returning true even if the widget wasn’t shown.

## What does this change?

Should mean that email signups are shown again on certain articles

Fixes a bug

No

## Screenshots

e.g. /science/2015/mar/10/how-science-can-be-a-childrens-playground-for-serious-lessons

Before change:

screen shot 2017-01-24 at 10 28 19

After change:

screen shot 2017-01-24 at 10 26 11


rich-nguyen (approved):

static/src/javascripts-legacy/projects/common/modules/email/run-checks.js
```diff
@@ -90,7 +90,7 @@ define([
 
     function obWidgetIsShown() {
         var $outbrain = $('.js-outbrain-container');
-        return $outbrain && $outbrain.length > 0;
+        return $outbrain && $outbrain.children && $outbrain.children.length > 0;
```
katebee: Does `$outbrain.children` mean anything, if $outbrain is an array?

`$outbrain.children === undefined` ???🤔 

picture 498

katebee:
Do we need to reference an array item to make this work?
```
    function obWidgetIsShown() {
        var $outbrain = $('.js-outbrain-container')[0];
        return $outbrain && $outbrain.children && $outbrain.children.length > 0;
      }
```

PR #15667, opened 2017-01-24
New nav link updates

zeftilldeath:
Few re-orderings and word swaps in the new nav menu.

@stephanfowler @NataliaLKB


NataliaLKB (approved):
👍

PR #15668, opened 2017-01-24
Remove IMR AB test, apply the IMR update for PROD

katebee:
## What does this change?

- Remove the Neilsen AB test
- Send the new IMR script to everyone
- Update each IMR apid to make it correct for PROD ( `T` => `P` )

- Switching to use the new imrworldwide JS
- API ID can now vary based on the config.page.section
- Sending correct data to IMR

Nope


regiskuckaertz (approved):
Yip yip!

PR #15518, opened 2017-01-10
remove unused assets

alexduf:
## What does this change?
cleanup unused assets
## What is the value of this and can you measure success?
Less clutter
## Does this affect other platforms - Amp, Apps, etc?
No
## Screenshots
No!

@guardian/guardian-frontend

PR #15669, opened 2017-01-24
Updating devinfra defaults to make sure tests fail when they should

NataliaLKB:
## What does this change?
Right now the default for DEVINFRA (teamcity) is to not use hashed bundles making it different from PROD. Really this should be done anyway as you want your test environment to be as close as possible to PROD. This actually meant that [this innocent looking change](https://github.com/guardian/frontend/pull/15518) passed all the tests and went to PROD and unfortunately caused errors.

This PR also removes the same assets as https://github.com/guardian/frontend/pull/15518, but also removes references in the JS config. 

Tests should fail when things are broken

Nope

## Tested in CODE?
No


TBonnin (approved):
lgtm. Great detective world 🔍

jfsoul (approved):

MatthewJWalls (approved):
👍

PR #15670, opened 2017-01-24
Set up the Ask4 Epic deployment test and the Ask4 "earning" strategy deployment

jranks123:
## What does this change?

In this PR, we split the audience into "earning" and "learning". 

The earning segment, as 88% of the audience, will deploy the epic to everyone on all articles, with a limit of 4 impressions per month. Any additional tests added to this segment will act as a boost, so if a user uses up their impressions with the Ask4 strategy, they could see the epic again if, for example, there was a separate test running on the same segment which instructed that the epic be shown on all articles with a certain tag.

The learning segment aims to learn whether staggering out these 4 asks over the month has more of an effect than showing them in a greedy, as soon as possible way, like the current Ask4 strategy does (which is running in the earning segment). 

In order to facilitate this test, we have added the ability to add a minimum time in between impressions at the variant level for Epic tests. 

If we learn that staggering the asks is more effective, we will have a strategy that converts more people. 

No

## Tested in CODE?
No

### DIFF
--- a/static/src/javascripts-legacy/projects/common/modules/commercial/acquisitions-view-log.js
+++ b/static/src/javascripts-legacy/projects/common/modules/commercial/acquisitions-view-log.js
@@ -27,12 +27,10 @@ define([
         var now = new Date().getTime();
 
         return viewLog.filter(function (view) {
-            return (test ? view.testId === test.id : true) && view.date > (now - ms) ;
+            return (test ? view.testId === test.id : true) && view.date > (now - ms);
         }).length;
     }
-
-
-
+    
     return {
         logView: logView,
         viewsInPreviousDays: viewsInPreviousDays
--- a/static/src/javascripts-legacy/projects/common/modules/commercial/acquisitions-view-log.js
+++ b/static/src/javascripts-legacy/projects/common/modules/commercial/acquisitions-view-log.js
@@ -1,0 +1,40 @@
+define([
+    'common/utils/storage'
+], function (storage) {
+
+    var viewKey = 'gu.contributions.views';
+    var viewLog = storage.local.get(viewKey) || [];
+
+
+    var maxLogEntries = 50;
+
+    /**
+     * Log that the user has seen an Epic test so we can limit how many times they see it.
+     * The number of entries is limited to the number in maxLogEntries.
+     *
+     * @param testId
+     */
+    function logView(testId) {
+        viewLog.push({
+            date: new Date().getTime(),
+            testId: testId
+        });
+        storage.local.set(viewKey, viewLog.slice(-maxLogEntries));
+    }
+
+    function viewsInPreviousDays(days, test) {
+        var ms = days * 1000 * 60 * 60 * 24;
+        var now = new Date().getTime();
+
+        return viewLog.filter(function (view) {
+            return (test ? view.testId === test.id : true) && view.date > (now - ms) ;
+        }).length;
+    }
+
+
+
+    return {
+        logView: logView,
+        viewsInPreviousDays: viewsInPreviousDays
+    }
+});
--- a/static/src/javascripts-legacy/projects/common/modules/experiments/acquisition-test-selector.js
+++ b/static/src/javascripts-legacy/projects/common/modules/experiments/acquisition-test-selector.js
@@ -3,18 +3,20 @@ define([
     'common/modules/commercial/acquisitions-view-log',
     'common/modules/experiments/tests/contributions-epic-brexit',
     'common/modules/experiments/tests/contributions-epic-always-ask-strategy',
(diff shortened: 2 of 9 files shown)


static/src/javascripts-legacy/projects/common/modules/experiments/acquisition-test-selector.js
```diff
@@ -1,24 +1,34 @@
...
-    var tests = [alwaysAsk, brexit];
+    var tests = [alwaysAsk, brexit, askFourStagger, askFourEarning];
 
     return {
         getTest: function() {
             var eligibleTests = tests.filter(function (test) {
                 var t = new test();
                 var forced = window.location.hash.indexOf('ab-' + t.id) > -1;
+                var variant = segmentUtil.variantFor(t);
+                var acceptableViewCount =
```
desbo: i'd split this into two `vars` for clarity

static/src/javascripts-legacy/projects/common/modules/experiments/acquisition-test-selector.js
```diff
@@ -1,24 +1,34 @@
...
-    var tests = [alwaysAsk, brexit];
+    var tests = [alwaysAsk, askFourEarning, brexit, askFourStagger];
 
     return {
         getTest: function() {
             var eligibleTests = tests.filter(function (test) {
                 var t = new test();
                 var forced = window.location.hash.indexOf('ab-' + t.id) > -1;
+                var variant = segmentUtil.variantFor(t);
+                var acceptableViewCount = variant? viewLog.viewsInPreviousDays(variant.maxViews.days, t) < variant.maxViews.count : false;
```
jfsoul: missing space before `?`

static/src/javascripts-legacy/projects/common/modules/experiments/acquisition-test-selector.js
```diff
@@ -1,24 +1,34 @@
...
+                var acceptableTimeBetweenViews = variant? viewLog.viewsInPreviousDays(variant.maxViews.minDaysBetweenViews, t) === 0 : false;
 
-                return forced || (t.canRun() && segmentUtil.isInTest(t));
+
+                return forced || (t.canRun() && segmentUtil.isInTest(t) && acceptableViewCount && acceptableTimeBetweenViews);
```
jfsoul: maybe look into using isInTest in a different way because we'll end up calling `variantForId` more than once here (from `isInTest` and `segmentUtil.variantFor(t)` above)

static/src/javascripts-legacy/projects/common/modules/experiments/tests/contributions-epic-ask-four-earning.js
```diff
@@ -0,0 +1,54 @@
...
+
+    return contributionsUtilities.makeABTest({
+        id: 'ContributionsEpicAskFourEarning',
+        campaignId: 'kr1_epic_ask_four_earning',
+
+        start: '2017-01-24',
+        expiry: '2017-05-01',
+
+        author: 'Jonathan Rankin',
+        description: "This places the epic on all articles for all users, with a limit of 4 impressions in any given 30 days",
```
jfsoul: single quotes around this for consistency

static/src/javascripts-legacy/projects/common/modules/commercial/acquisitions-view-log.js
```diff
@@ -0,0 +1,40 @@
...
+        });
+        storage.local.set(viewKey, viewLog.slice(-maxLogEntries));
+    }
+
+    function viewsInPreviousDays(days, test) {
+        var ms = days * 1000 * 60 * 60 * 24;
+        var now = new Date().getTime();
+
+        return viewLog.filter(function (view) {
+            return (test ? view.testId === test.id : true) && view.date > (now - ms) ;
```
jfsoul: extra space

jfsoul (approved):
Minor comments aside, this looks good to me! Thanks for the explanations 👍

static/src/javascripts-legacy/projects/common/modules/experiments/acquisition-test-selector.js
desbo: i think this should be ok since `variantIdFor` is memoized, so we only do the cookie read etc once

jfsoul: yeah I saw that but I thought the non-memoized version will be called by `isInTest`.  Correct me if I'm wrong! 😄

desbo: oh yeah, you're right, thanks!
@jranks123 easy fix for this is to move the `memoize` call to the definition of `variantIdFor`

PR #15671, opened 2017-01-24
look up platform in atoms block

gidsg:
## What does this change?
Small refactor to look up platform of main media atom from atoms CAPI response.

Less Jsoup 

No

## Screenshots
N/A

## Tested in CODE?
No

### DIFF
--- a/common/app/implicits/FaciaContentFrontendHelpers.scala
+++ b/common/app/implicits/FaciaContentFrontendHelpers.scala
@@ -33,9 +33,8 @@ object FaciaContentFrontendHelpers {
        document <- Some(Jsoup.parse(main))
        atomContainer <- Option(document.getElementsByClass("element-atom").first())
        bodyElement <- Some(atomContainer.getElementsByTag("gu-atom"))
-       youTubeIframe <- Option(bodyElement.select("iframe[src^=https://www.youtube.com]").first())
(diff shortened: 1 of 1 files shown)


common/app/implicits/FaciaContentFrontendHelpers.scala
```diff
@@ -33,9 +33,8 @@ object FaciaContentFrontendHelpers {
        document <- Some(Jsoup.parse(main))
        atomContainer <- Option(document.getElementsByClass("element-atom").first())
        bodyElement <- Some(atomContainer.getElementsByTag("gu-atom"))
-       youTubeIframe <- Option(bodyElement.select("iframe[src^=https://www.youtube.com]").first())
        atomId <- Some(bodyElement.attr("data-atom-id"))
-       mainMediaAtom <- atoms.media.find(_.id == atomId)
+       mainMediaAtom <- atoms.media.find(ma => ma.id == atomId && ma.assets.exists(_.platform == "Youtube"))
```
akash1810: does capi provide a `Platform` enum to avoid string matching?

gidsg: Yes, but frontend uses its own representation of the CAPI model so we'd need to recreate it in Frontend. I can look into this but I think I'd rather do it in a separate PR.

akash1810 (approved):

PR #15672, opened 2017-01-24
Measure more about the module loaders

sndrs:
## What does this change?

adds the following events to ga, for both webpack and require builds

- standardBoot
- commericalRequest
- commericalBoot
- enhancedRequest
- enhancedBoot

it should allow to determine whether differences in the eval times between the bundlers, which we're currently marking, correspond to the ways in the which they work, or the actual speed in which js arrives/runs etc.

no

cc @SiAdcock

### DIFF
--- a/static/src/javascripts-legacy/boot.js
+++ b/static/src/javascripts-legacy/boot.js
@@ -20,11 +20,17 @@ Only if we detect we should run enhance.
 define([
     'Promise',
     'domReady',
-    'common/utils/raven'
+    'common/utils/raven',
+    'common/utils/user-timing',
+    'common/utils/robust',
+    'common/modules/analytics/google'
 ], function (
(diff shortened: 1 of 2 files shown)


static/src/javascripts-legacy/boot.js
```diff
@@ -57,10 +69,18 @@ define([
             });
         }
 
+        userTiming.mark('commerical request');
```
gtrufitt: Spelling mistake on commercial/commerical on all these :)

gtrufitt (approved):
Great!

static/src/javascripts-legacy/boot.js
sndrs: you've no idea how many times i've spelt that like that in my life...

PR #15673, opened 2017-01-24
Add scala linter to project

kelvin-chappell:
Just applying rules that are uncontroversially good practice, rather than stylistic or of questionable value.

First one is FuncFirstThenMap, which seems like a sensible rule to me.
Example output:
```
[error] frontend/onward/app/feed/MostPopularAgent.scala:25: [FuncFirstThenMap] Use method take first, then map.
[error]       val mostViewed = response.mostViewed.getOrElse(Nil) map { RelatedContentItem(_) } take 10
[error]                                                                                         ^
[error] frontend/onward/app/feed/MostPopularAgent.scala:154: [FuncFirstThenMap] Use method take first, then map.
[error]         val mostViewed = response.mostViewed.getOrElse(Nil) map { RelatedContentItem(_) } take 10
[error]                                                                                           ^
[error] two errors found
[error] frontend/admin/app/dfp/DataMapper.scala:12: [FuncFirstThenMap] Use method tail first, then map.
[error]     val ancestorNames = if (ancestors.isEmpty) Nil else ancestors.map(_.getName).tail
[error]                                                                                  ^
[error] frontend/admin/app/jobs/R2PagePressJob.scala:89: [FuncFirstThenMap] Use method headOption first, then map.
[error]       .headOption
[error]        ^
```

I suppose there's a question about whether these should be considered errors or warnings.
Currently they appear as errors because we consider all warnings to be errors.

/cc @guardian/dotcom-platform

### DIFF
--- a/admin/app/dfp/DataMapper.scala
+++ b/admin/app/dfp/DataMapper.scala
@@ -9,7 +9,7 @@ object DataMapper {
 
   def toGuAdUnit(dfpAdUnit: AdUnit): GuAdUnit = {
     val ancestors = toSeq(dfpAdUnit.getParentPath)
-    val ancestorNames = if (ancestors.isEmpty) Nil else ancestors.map(_.getName).tail
+    val ancestorNames = if (ancestors.isEmpty) Nil else ancestors.tail.map(_.getName)
     GuAdUnit(dfpAdUnit.getId, ancestorNames :+ dfpAdUnit.getName, dfpAdUnit.getStatus.getValue)
   }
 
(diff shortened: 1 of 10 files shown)


TBonnin (commented):
I am all up for anything that would improve the quality of the code (I am assuming that's why we would like to introduce a linter).
I have a few questions though:
Why did you choose this specific linter? Is it well maintained from a long term perspective?
Also who, how and when are we gonna introduce new rules? It won't be very useful if we introduce a linter but only ever use one single rule.

LATaylor-guardian:
For allot of our projects we use Scalariform. Not modification or configuration required and it get out of the way: https://github.com/guardian/crier/blob/master/project/plugins.sbt#L7

kelvin-chappell:
@TBonnin I've been using it for a while in my local sbt config and the rules seemed quite sensible.  Although you raise a good point that it may be solving a problem that doesn't exist in general.  I think I'll close this until there is a really good case for it.  Also it's a 0.x release.

kelvin-chappell:
@LATaylor-guardian I've seen scalariform but I got the impression it was more for code style than potential problems.

DiegoVazquezNanini:
Hi @kelvin-chappell,
If you are not going to add the lint can you please raise a PR with the FuncFirstThenMap changes?
Some aspects of the tool were controversial, but not the changes. They all look good.

kelvin-chappell:
@DiegoVazquezNanini Good point.  I won't throw the baby out with the bath water.

PR #15675, opened 2017-01-24
New header menu colour change

zeftilldeath:
## I'm Blue da ba dee da ba daa
Colour of the account area has been changed to news-main-2. Also the spacing has been tweaked slightly to add a little more breathing space between words and the top of their containers. 

screen shot 2017-01-24 at 16 51 49

## Tick
The edition switcher has been given a clear active state through the use of a TICK and BOLD TEXT.

screen shot 2017-01-24 at 16 52 14

## Outline
Now the account area is the same colour as the hamburglar, I needed to add a slight outline to the opened state of the menu button to prevent it bleeding into the background colour.

screen shot 2017-01-24 at 16 52 25

FAO the usual suspects:
@NataliaLKB @stephanfowler

### DIFF
--- a/static/src/stylesheets/layout/_ab-new-header-test-variant.scss
+++ b/static/src/stylesheets/layout/_ab-new-header-test-variant.scss
@@ -113,31 +113,12 @@ $gutter-large: 55px;
     cursor: pointer;
     bottom: -$gs-baseline / 2;
 
-    &--open {
-        z-index: 1071;
-        box-shadow: 0 0 0 1px rgba(0, 0, 0, .08);
-
-        &:before {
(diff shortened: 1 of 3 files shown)


static/src/stylesheets/layout/_ab-new-header-test-variant.scss
```diff
@@ -115,6 +115,7 @@ $gutter-large: 55px;
 
     &--open {
```
NataliaLKB: We should change this to be the full classname, not nested to make sure it is grepable

NataliaLKB (approved):
Still dislike the colour, but I will get used to it 👍

PR #15676, opened 2017-01-24
Add Renault campaign switch

kelvin-chappell:
So that we can redirect the campaign in a controlled way and bring it back if it goes wrong.

So principle is: 
1. Campaign will 404
1. We set up redirect
1. Campaign uses Composer pages


lps88 (approved):
👍

PR #15677, opened 2017-01-24
youtube iframe api ready fix (Redux)

gidsg:
## What does this change?

Fixes a bug we were seeing where youtube atoms were not being enhanced on pages where another non-youtube video existed which had pre-roll adverts enabled. Now we don't allow pre-roll ads if the page's content contains a youtube atom.

Note, we now explicitly prevent pre-roll on Fronts in the JS and don't instantiate the IMA JS whereas previously pre-roll was blocked on the DFP side.

Consistent loading of enhanced youtube atoms.
Does this affect other platforms - Amp, Apps, etc?

No
## Screenshots

N/A
## Tested in CODE?

No


akash1810 (approved):
NICE!! 👍

PR #15643, opened 2017-01-20
New header variant seven

NataliaLKB:
## What does this change?
Very little from variant six actually. The differences are:
- Improved subsection styling
- better footer styling (done in a subsequant PR https://github.com/guardian/frontend/pull/15620 )
- edition picker is in the menu
- Subscribe trapezoid is where the edition picker used to be (https://github.com/guardian/frontend/pull/15636)
- Making sure football is in the mostPopular subsection list for the homepage
- Shortening `UK news`, `US news`, `Australia News` and `world news` to `UK`, `US`, `Australia` and `world`

cc @guardian/dotcom-platform 

## What is the value of this and can you measure success?
This will hopefully be just as successful as our last variant (or more!) and we also think it looks better

## Does this affect other platforms - Amp, Apps, etc?
Nope

## Tested in CODE?
Nope

PR #15678, opened 2017-01-24
Quick new header menu fix

NataliaLKB:
## What does this change?
Some hangover bugs from #15643. The bugs includes the international editions having a noticeably smaller list, and also the us edition having repeats in the list.

I also took the time to clean up any other duplication that seemed to make it in.

Less 🐛s ! and better lists

Affects amp: fixes the bugs there too!

## Tested in CODE?
Nope


zeftilldeath (approved):
Looks
Good
To
Me

Let's
Get
This
Merged

PR #15681, opened 2017-01-25
Extending expired switch for Joe

NataliaLKB:
## What does this change?
@joelochlann this switch is breaking the build


joelochlann:
Thanks @NataliaLKB - I'm really hoping this test will be ending soon (I'm ever over-optimistic about how long it will last)

PR #15685, opened 2017-01-25
model Media Asset Platform as enum

gidsg:
## What does this change?
Models Media Asset Platform as enum

More type safety

No

## Screenshots
N/A

## Tested in CODE?
No

### DIFF
--- a/common/app/implicits/FaciaContentFrontendHelpers.scala
+++ b/common/app/implicits/FaciaContentFrontendHelpers.scala
@@ -4,7 +4,7 @@ import common.Edition
 import common.commercial.PaidContent
 import implicits.Dates._
 import model._
-import model.content.MediaAtom
+import model.content.{MediaAssetPlatform, MediaAtom}
 import model.pressed._
 import org.joda.time.DateTime
 import org.jsoup.Jsoup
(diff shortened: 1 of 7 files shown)


TBonnin (commented):
Would a `scala.Enumaration` fits your need? 
If not what does the `enumeratum` lib bring that you cannot do with a plain `sealed trait`

gidsg:
@TBonnin The main benefit of `enumeratum` for me was that it provides the implicit JSON formatter for Play. https://github.com/lloydmeta/enumeratum#playenum

TBonnin:
@gidsg I don't see where in this PR you need this implicit json formatter. Am I missing something?

TBonnin:
@gidsg Nevermind here https://github.com/guardian/frontend/pull/15685/files#diff-ed9aaed2c6c7363409f7357c602ba1e1R152 :D

akash1810 (approved):
🙌 👍

TBonnin:
Adding a dependency for a single and narrow use case seems to be the best way to end up with a huge dependencies tree. I won't block this PR, it is up to you. However I am curious to see what extra code you will need to write to have the exact same feature without using any 3rd party lib.

gidsg:
@TBonnin I think let's merge this for now and when I have a moment I will have a go at refactoring without the dependency. When I tried before with Scala Enumeration I got stuck in JSON formatting hell.

TBonnin:
@gidsg Enumeration have a lot of shortcomings. I am more talking about using a sealed trait with some extra code to handle the json formatting

gidsg:
CC @philwills

PR #15689, opened 2017-01-25
Smaller icons for smaller headlines in emails even in Outlook

lmath:
## What does this change?
In emails, when headlines have icons (e.g. the quote marks for opinion headlines), then the size of the icon should be consistent with the size of the headline. This is the case for most clients. This change should ensure this is also true for Outlook on Windows. 

This involved setting height on icon img elements. This will be ignored by most clients but will be useful for Outlook.

Emails read on Outlook on Windows will be rendered more consistently with how they appear on other clients. 

Nope.

## Screenshots
This is how the icons appear on other clients e.g. here iPhone 6:
image

This is how this email would look on Outlook on Windows before the change:
image

After the change:
image

### DIFF
--- a/facia/app/views/fragments/emailFrontBody.scala.html
+++ b/facia/app/views/fragments/emailFrontBody.scala.html
@@ -10,7 +10,7 @@
 @import views.support.TrailCssClasses.toneClassFromStyle
 
 
-@headline(card: ContentCard) = {
+@headline(card: ContentCard, largeHeadline: Boolean = false) = {
     <a @Html(card.header.url.hrefWithRel) class="facia-link">
         <h3 class="headline">
             @card.header.kicker.map { kicker =>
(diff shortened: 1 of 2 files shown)


facia/app/views/fragments/emailFrontBody.scala.html
```diff
@@ -46,9 +46,9 @@ <h4 class="trail-text">@Html(trailText)</h4>
     }
 }
 
-@headlineAndTrailWithCutout(card: ContentCard) = {
+@headlineAndTrailWithCutout(card: ContentCard, withImage: Boolean) = {
     @fullRow(Seq("facia-card__text")) {
-        @headline(card)
+        @headline(card, withImage)
```
joelochlann: Could be slightly clearer to use a named param here, i.e. `@headline(card, largeHeadline = withImage)`

facia/app/views/fragments/emailFrontBody.scala.html
```diff
@@ -10,7 +10,7 @@
 @import views.support.TrailCssClasses.toneClassFromStyle
 
 
-@headline(card: ContentCard) = {
+@headline(card: ContentCard, largeHeadline: Boolean = false) = {
```
joelochlann: maybe `isLarge` would be a better name for this param? since we already know it's a headline

joelochlann (approved):
Good stuff! Thanks for fixing. Just a few naming things

facia/app/views/fragments/emailFrontBody.scala.html
lmath: Agreed. That is better.

PR #15691, opened 2017-01-25
Add new size for outstream ads in inline slots

regiskuckaertz:
Currently, outstream video ads like Teads Inreads break out of the ad slot. This is not desirable at all, for obvious reasons, but most importantly because we can't measure viewability. In this chance, we introduce a new ad size Teads said could allow them to run _in situ_. As a bonus, they will also be pushing towards delivering SafeFrame enabled creatives.

### DIFF
--- a/static/src/javascripts-legacy/projects/common/modules/commercial/ad-sizes.js
+++ b/static/src/javascripts-legacy/projects/common/modules/commercial/ad-sizes.js
@@ -13,6 +13,7 @@ define(function () {
 
         // guardian proprietary ad sizes
         video:                  AdSize(620, 1),
+        video2:                 AdSize(620, 350),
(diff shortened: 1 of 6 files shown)


regiskuckaertz:
[Create a new size for Teads](https://trello.com/c/JDjN6pof/142-create-a-new-size-for-teads)

ScottPainterGNM:
Out of curiosity, why are we removing inline2 and 3 in this PR?

regiskuckaertz:
@ScottPainterGNM because they're all similar, I should've gotten rid of them long ago.

ScottPainterGNM:
Cool - can/have you confirmed with adops they aren't doing any specific targeting to this slot?

scottp719 (approved):
👍 Apart from one query

PR #15692, opened 2017-01-26
Fix problems found by linter

kelvin-chappell:
These are small optimisations that should produce same results more efficiently.


DiegoVazquezNanini (approved):
:+1:

TBonnin (approved):

PR #15690, opened 2017-01-25
Measure site speed at 100% of GA

sndrs:
## What does this change?

bumps site speed GA measuring to 100% 

## What is the value of this and can you measure success?

faster stats for webpack test

## Does this affect other platforms - Amp, Apps, etc?

cost of GA, must be disabled before end of the month

PR #15693, opened 2017-01-26
make it an int

sndrs:
## What does this change?

sets `siteSpeedSamplePercentage` to be an `Int`, like `samplePercentage`

opening up `siteSpeedSamplePercentage ` too 100 in #15690 has not had any noticeable effect on the stats, which we'd expect it to. the result in page source is `{'sampleRate':100,'siteSpeedSampleRate':100.0}`, so this is a stab at the problem being that `.0`


SiAdcock (approved):
:crossed_fingers:

PR #15697, opened 2017-01-26
Fixes caption padding on AMP

gidsg:
## What does this change?
Fixes caption padding for Media Atoms. 

Note this creates a new fragment for the poster image case using the `image` and `ampImage` fragments, rather than using the entirety of the `imageFigure` which was causing additional classes to be added which caused unwanted behaviour.

Consistency with other elements

Also affects AMP.

### BEFORE
In-Body
screen shot 2017-01-26 at 17 15 10
Main Media
screen shot 2017-01-26 at 17 15 33

### AFTER
screen shot 2017-01-26 at 17 26 13
screen shot 2017-01-26 at 17 25 44

#### Poster Only
screen shot 2017-01-26 at 17 34 14
screen shot 2017-01-26 at 17 34 08

## Tested in CODE?
No

### DIFF
--- a/static/src/stylesheets/module/_layout-hints.scss
+++ b/static/src/stylesheets/module/_layout-hints.scss
@@ -96,7 +96,8 @@ figure.element.element--showcase {
 }
 
 figure.element.element--supporting,
-figure.element-video {
+figure.element-video,
+figure.element-atom {
     figcaption {
         padding-top: ($gs-baseline/3)*2;
         padding-bottom: ($gs-baseline/3)*2;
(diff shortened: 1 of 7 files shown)


NataliaLKB:
> AMP only

From what I saw this was affecting the site on mobile generally, not just AMP.

NataliaLKB (approved):
lgtm 👍

gidsg:
> From what I saw this was affecting the site on mobile generally, not just AMP.

Good point @NataliaLKB you are right! It does also affect non-AMP. I have updated the description now.

static/src/stylesheets/module/_layout-hints.scss
```diff
@@ -109,11 +110,6 @@ figure.element-video {
     }
 }
 
-figcaption.caption--media-atom {
```
NataliaLKB: 👍

PR #15698, opened 2017-01-27
Fixing comment acitivity link new header menu

NataliaLKB:
## What does this change?
Currently we have a bug where the comment activity link is put on the wrong element in the new nav:

This fixes that problem.

People can easily access their comment history

Nope

## Tested in CODE?
Nope


stephanfowler:
👍

PR #15699, opened 2017-01-27
Show youtube atoms placeholder on small cards on fronts

GHaberis:
## What does this change?

Youtube atoms in small cards on fronts should not be not playable. Because of this they currently don't use the youtube template fragment. This change means non-playable cards will use the youtube template fragment, but only the poster image and play button will be rendered, the player's iframe won't be rendered.

Youtube atom poster images appear on front now for small cards

No

## Screenshots

Before
picture 32
After
picture 31

## Tested in CODE?

No

@gidsg

### DIFF
--- a/common/app/views/fragments/atoms/youtube.scala.html
+++ b/common/app/views/fragments/atoms/youtube.scala.html
@@ -4,19 +4,19 @@
 @import views.html.fragments.atoms.mediaAtomCaption
 @import com.netaporter.uri.dsl._
 
-@(media: model.content.MediaAtom, displayCaption: Boolean, embedPage: Boolean, displayDuration: Boolean = true, displayEndSlate: Boolean = true, isPlayable: Boolean = true)(implicit request: RequestHeader)
+@(media: model.content.MediaAtom, displayCaption: Boolean, embedPage: Boolean, displayDuration: Boolean = true, displayEndSlate: Boolean = true, playable: Boolean = true)(implicit request: RequestHeader)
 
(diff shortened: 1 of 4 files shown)


common/app/views/fragments/atoms/youtube.scala.html
```diff
@@ -4,31 +4,34 @@
 @import views.html.fragments.atoms.mediaAtomCaption
 @import com.netaporter.uri.dsl._
 
-@(media: model.content.MediaAtom, displayCaption: Boolean, embedPage: Boolean, displayDuration: Boolean = true, displayEndSlate: Boolean = true)(implicit request: RequestHeader)
+@(media: model.content.MediaAtom, displayCaption: Boolean, embedPage: Boolean, displayDuration: Boolean = true, displayEndSlate: Boolean = true, isPlayable: Boolean = true)(implicit request: RequestHeader)
```
gidsg: could possibly call this just `playable` as the `is` part can be inferred from the type system.

akash1810 (approved):
woop!

gidsg:
I guess we will still have this issue with the slightly unmatched image sizes? Anyway, I think changing this should be done in a different PR if it makes sense.
unnamed

akash1810:
@gidsg yep, makes sense to make it a different PR.

IIRC we render traditional videos as 5:3 on fronts which causes letterboxing 😢

gidsg:
@akash1810 we can't do that with the youtube player, it has to be 16:9 I believe, but we could do it in this 'non playable' case.

PR #15700, opened 2017-01-27
Add ability to boost fidelity of timing metrics in GA

sndrs:
## What does this change?

- adds switch-based standard GA event to track performance timings
- restores GA timing throttle to 0.1%, reverting #15690 and #15693
- removes the extra events added in #15672

Standard timing events are throttled to 0.1%. This should enable us to temporarily open performance monitoring to 100% when needed (and ok'd by the GA team).

cost of GA, potentially (hence always check with GA team before using it)


gtrufitt (approved):
Sweet as.

SiAdcock (approved):
Tidy

PR #15704, opened 2017-01-27
Fix closed comment caching (behind a switch)

jfsoul:
## What does this change?
The way that the closed comment long cache switch was implemented previously didn't work.  It relied on `config.page.commentable` which would always be true if the comment component was actually loaded - therefore we never saw any `commentable=false` requests.

This switches to using the existing "is this closed for comment?" (`getDiscussionClosed`) method so that we are now using the correct property.  There is also a read-only flag, but I'm not entirely clear on what that is used for (if at all) - so I'm only changing the cache on the closed ones.

We can test whether caching closed comment threads for longer has an effect on the number of discussion requests.

## Tested in CODE?
No


DiegoVazquezNanini (approved):
LGTM, but this needs a client side dev.

gtrufitt (approved):
LGTM

PR #15706, opened 2017-01-27
Add a hash flag to block all IAS requests

regiskuckaertz:
Alright so it appears to be working:

picture 3

... except for one URL, `https://pixel.adsafeprotected.com/jload`, for no obvious reason. Any help mucho appreciado 😄


static/src/javascripts-legacy/bootstraps/enhanced/main.js
```diff
@@ -163,10 +163,17 @@ define([
...
 
         // use a #force-sw hash fragment to force service worker registration for local dev
-        if ((window.location.protocol === 'https:' && config.page.section !== 'identity') || window.location.hash === '#force-sw') {
+        if ((window.location.protocol === 'https:' && config.page.section !== 'identity') || window.location.hash.indexOf('force-sw') > -1) {
             var navigator = window.navigator;
             if (navigator && navigator.serviceWorker) {
-                navigator.serviceWorker.register('/service-worker.js');
+                navigator.serviceWorker.register('/service-worker.js')
+                .then(function () {
+                    return navigator.serviceWorker.ready;
```
gtrufitt: Why is this inside the then of the register? Could it not be `navigation.serviceWorker.ready.then()` at the same level as the register?

gtrufitt:
Might be worth putting this behind a switch? Not that once the SW is registered that would help...

static/src/javascripts-legacy/bootstraps/enhanced/main.js
```diff
@@ -163,10 +163,17 @@ define([
         }
 
         // use a #force-sw hash fragment to force service worker registration for local dev
-        if ((window.location.protocol === 'https:' && config.page.section !== 'identity') || window.location.hash === '#force-sw') {
+        if ((window.location.protocol === 'https:' && config.page.section !== 'identity') || window.location.hash.indexOf('force-sw') > -1) {
```
gtrufitt: Does this suffer from a race condition as requests for ias will begin in commercial but this isn't registered until enhanced? I suppose that's ok as most ads are lazy loaded so we might miss sticky top but catch the others?

static/src/javascripts-legacy/bootstraps/enhanced/main.js
regiskuckaertz: Oh yeah, did not think about that.

static/src/javascripts-legacy/bootstraps/enhanced/main.js
regiskuckaertz: Yeah I guess this is an acceptable trade-off. I'm not familiar enough with SW but, once it is registered, will it kick off at once on the next page load? (as in not wait until the `register` call in enhanced)

gtrufitt: I don't think `blockIAS` will be true until we're in enhanced though? So yes, the fetch will catch it, but `blockIAS` isn't persisted for the session? Thoughts @sndrs ?

regiskuckaertz: Mind if I move the `ready` call to standard? It is just an async call so should do no harm.

sndrs: i can't think of a good reason? feels like it's not a bad move anyway to me?

regiskuckaertz: Done. I do not do the whole URL test (the promise will just never resolve if the SW is not there) but I can add it if you prefer

regiskuckaertz:
@gtrufitt yep, adding a switch

regiskuckaertz:
Also, the request I am trying to block is not even going through the Service Worker. Do you know of any reason why a network call would bypass the `fetch` event? To be precise, it is a script inserted using `document.write` that I am trying to block.

regiskuckaertz:
[IAS viewability performance issues](https://trello.com/c/3GaFqvxX/145-ias-viewability-performance-issues)

regiskuckaertz:
@rich-nguyen @gtrufitt @sndrs I don't think I can go any further. All _local_ IAS requests will be blocked. The one that I was complaining about is loaded via cross-origin iframes and is thus out-of-touch. That means, IAS (`jload`) will still be downloaded in safeframes and most programmatic creatives. I will ask Tom to setup a local test just to be sure, but otherwise I'm done.

gtrufitt (approved):
Looks good. 👍

PR #15708, opened 2017-01-27
Fix exception when fixed/video container is used with atom media video page

gidsg:
## What does this change?
The `fixed/video` container previously made an assumption that content with `type/video` tag contains a video element main media. This is not true in the case of Atom powered video pages, so when one was was added the container would throw an exception when `content.elements.mainVideo.get` was attempted.

This fix stops the exception but does leave a blank space in the container. Ideally we would filter out these unsupported items from the container, however, I thought it important to put something out to fix the exception in the first instance.

Fix possible exception

No

#### Before
screen shot 2017-01-27 at 11 40 57
#### After
screen shot 2017-01-27 at 16 29 17

## Tested in CODE?
No


akash1810:
> Ideally we would filter out these unsupported items from the container

👍 as discussed IRL - lets do this as a subsequent PR.

akash1810 (approved):

PR #15709, opened 2017-01-27
config.page property should be hasYouTubeAtom not hasYouTubeMediaAtom

GHaberis:
## What does this change?

config.page property should be hasYouTubeAtom not hasYouTubeMediaAtom, incorrectly referencing hasYouTubeMediaAtom means the ima3.js bug still occurs.

Should fix ima3.js bug

No

## Screenshots

N/A

## Tested in CODE?

No

@akash1810


akash1810 (approved):
👍

PR #15710, opened 2017-01-30
Fixing the occasional case of a double headline appearing on amp

NataliaLKB:
## What does this change?
Fixes: https://trello.com/c/8xV4abkp/199-amp-headline-standfirst-doubling-up-intermittent

If you are in the serverside test `ABNewNavVariantSeven` and on an AMP page you may occasionally see a double headline. This is because in the test the html for the headline is presented twice (can't use flexbox to reposition headline because of how the html is structured). On the site we have a lot of helper classes in css to only show certain things on mobile vs desktop, on AMP we don't have as many of those helper classes which is why both headlines show.

This is an intermittent issue because most of the pages are cached correctly. 

Fixing a 🐛 

This should only affect AMP

## Screenshots
Before:

After:

## Tested in CODE?
No

### DIFF
--- a/article/app/views/fragments/articleBody.scala.html
+++ b/article/app/views/fragments/articleBody.scala.html
@@ -39,7 +39,7 @@
                 @fragments.guBand()
             }
 
-            @if(!mvt.ABNewNavVariantSeven.isParticipating) {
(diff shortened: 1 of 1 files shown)


stephanfowler:
👍 !

PR #15712, opened 2017-01-30
Membership bundles AB test - revised variants

JustinPinner:
## What does this change?
Drop the A1 variant from the AB test defined [here](https://github.com/guardian/frontend/pull/15661)

We need to make some changes to the bundle definitions, but in the meantime we can continue with data gathering for the other variants (B1 and B2).

No, web only.

## Screenshots
N/A

## Tested in CODE?
~~Not yet. Need a build first!~~ Yes. Looks fine.

cc @svillafe @Ap0c


JustinPinner:
Closing this again. We're going to run with the three originally planned variants.

PR #15713, opened 2017-01-30
Make closed discussion long cache change permanent

jfsoul:
## What does this change?
Removes the switch controlling longer cache TTL on closed discussion threads.  I've turned this switch on and there seem to be about half the number of requests coming through for closed discussion threads.  Planning to keep an eye on this for the rest of the day, just to be extra safe and then merge tomorrow.

I would also advocate increasing this cache length even further in the future (24 hours or longer).  Leaving as is for now.

Less requests to discussion for content that doesn't change.

## Tested in CODE?
No

### DIFF
--- a/common/app/conf/switches/PerformanceSwitches.scala
+++ b/common/app/conf/switches/PerformanceSwitches.scala
@@ -56,16 +56,6 @@ trait PerformanceSwitches {
     exposeClientSide = false
   )
 
-  val LongCacheCommentsSwitch = Switch(
-    SwitchGroup.Performance,
-    "long-cache-comments-switch",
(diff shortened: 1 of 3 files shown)


TBonnin (approved):
lgtm. 
Are closed comment threads ever reopen? If yes would it make sense to have different cache keys for the same thread depending upon its closed state?

jfsoul:
@TBonnin this is the reason for the `commentsClosed` query parameter we pass back and forth from the client-side.  If it weren't for reopening of threads we could avoid that altogether and just check the CAPI response in the controller to see if it's closed.

TBonnin:
got it

DiegoVazquezNanini (approved):

PR #15714, opened 2017-01-30
Guardian today email test

davidfurey:
## What does this change?

Adds an a/b/c test for guardian today sign up messaging

Discovering which message is most effective at increasing the number of sign-ups

No

## Screenshots

Message A
screen shot 2017-01-31 at 17 03 16

Message B
screen shot 2017-01-31 at 17 03 20

Message C
screen shot 2017-01-31 at 17 03 25


static/src/javascripts-legacy/projects/common/modules/experiments/ab.js
```diff
@@ -243,9 +246,10 @@ define([
      */
     function registerCompleteEvent(complete) {
         return function initListener(test) {
-            var variantId = segmentUtil.variantIdFor(test);
 
-            if (segmentUtil.isInTest(test)) {
+            var variantId = getTestVariantId(test.id);
```
joelochlann: Might be worth checking with someone who knows the a/b testing well that this is expected behaviour (although it seems right to me)

static/src/javascripts-legacy/projects/common/modules/experiments/ab-test-clash.js
```diff
@@ -21,15 +21,31 @@ define([
...
-            ContributionsEpicAskFourStagger,
-            ContributionsEpicAskFourEarning
-        ];
-        return _testABClash(ab.isInVariant, clashingTests);
+    var nonEmailClashingTests = contributionsTests;
+
+    var clashingTests = contributionsTests.concat(emailTests);
+
+    function userIsInAClashingAbTest(tests) {
+        if (tests == null) {
```
joelochlann: `if (!tests)` is generally a more idomatic way to do this - it will check for `0, false, ''` and `null`

static/test/javascripts/spec/common/experiments/ab.spec.js
```diff
@@ -297,6 +298,9 @@ define([
             it('should fire the success function when canRun is false', function() {
                 var spy = sinon.spy();
 
+                ab.addTest(test.one);
+                ab.segmentUser();
```
joelochlann: Why did we have to add this in? Just makes me slightly worried it might be requires elsewhere due to these changes

static/src/javascripts-legacy/projects/common/modules/email/run-checks.js
```diff
@@ -30,6 +30,7 @@ define([
     Promise
 ) {
     var emailInserted = false;
+    var compliantOutbrain = false;
```
joelochlann: Seems a bit weird storing this property here - could we make this a method/property in the outbrain module and call it when we need it?

static/src/javascripts-legacy/projects/common/modules/experiments/ab-test-clash.js
joelochlann: actually an even more idiomatic way for setting default values for parameters (in ES5 anyway) would be:
```
tests = tests || clashingTests
```

joelochlann (changes requested):
Just a few questions - since this includes some other fixes/changes slightly orthognal to the test I want to be sure that they're OK.

Also just a bit hazy on how the non-compliant stuff works. We only insert email signup if the Outbrain slot is non-compliant. But isn't it the position of the email signup box that causes Outbrain to be non-compliant?? In my mind it's a chicken-and-egg scenario - but I'm pretty sure that's because I've missed something

static/test/javascripts/spec/common/experiments/ab.spec.js
davidfurey: This is just required to fix the test.  In the real world, a user will always be segmented before the success event runs.  The test previously didn't segment the user, but because of the fix (https://github.com/guardian/frontend/pull/15714#discussion-diff-98672710R250), it needs to.

static/src/javascripts-legacy/projects/common/modules/email/run-checks.js
davidfurey: If compliantOutbrain is a property of outbrain, then run-checks would require outbrain, and outbrain would require run-checks.  Are circular dependencies allowed?

static/src/javascripts-legacy/projects/common/modules/experiments/ab-test-clash.js
davidfurey: That is much nicer, thanks.

static/src/javascripts-legacy/projects/commercial/modules/third-party-tags/outbrain-codes.js
```diff
@@ -41,6 +41,8 @@ define([
     function getCode(data) {
         if (!(data.slot in outbrainCodes) || data.slot === 'defaults') {
             return outbrainCodes.defaults[getSection(data.section)][data.breakpoint === 'wide' ? 'desktop' : data.breakpoint];
+        } else if (data.slot == 'nonCompliant') {
```
davidfurey: Oops, thanks.

static/test/javascripts/spec/common/experiments/ab.spec.js
joelochlann: 👍

static/src/javascripts-legacy/projects/common/modules/experiments/ab.js
desbo: I don't think we can rely on localStorage to fire the complete event. localStorage is only updated with a test allocation if that test's `canRun` criteria is true (otherwise, it'll be set to `notintest`), but in some cases we need to register the `complete` event on a page after the test treatment, when the `canRun` criteria might be false. That's why we use the MVT ID cookie to register the completion listener: it determines which variant the user would be in without relying on `canRun`.

static/src/javascripts-legacy/projects/common/modules/email/run-checks.js
joelochlann: Ah I see. No I don't think circular dependencies are allowed

joelochlann (approved):

static/src/javascripts-legacy/projects/common/modules/experiments/ab.js
davidfurey: Discussed with @desbo and we've tested that the success event is still successfully fired even if the canRun criteria is false.

joelochlann:
👍  LGTM

static/src/javascripts-legacy/projects/common/modules/experiments/ab-test-clash.js
```diff
@@ -21,15 +21,29 @@ define([
...
+        ContributionsEpicAskFourEarning
+    ];
+
+    var emailTests = [
+        GuardianTodaySignupMessaging
+    ];
+
+    var nonEmailClashingTests = contributionsTests;
+
+    var clashingTests = contributionsTests.concat(emailTests);
```
markjamesbutler: 👍

markjamesbutler:
@davidfurey I'd echo @joelochlann's comments regarding non-compliance.  When I did as similar thing in contributions we would load the non-complaint outbrain when the contributions banner was displayed.  It seems this PR is doing the opposite.  Let's catch-up when you are back in the office.

markjamesbutler (approved):

PR #15716, opened 2017-01-30
Filter fixed/video container to only show items with Main Media Video Element

gidsg:
## What does this change?
Filter video pages without Main Video Element from appearing in `fixed/video` container.

Removes blank space when non Elements style video page is added

No

## Screenshots
Before

screen shot 2017-01-30 at 14 31 09

After
screen shot 2017-01-30 at 14 32 34

## Tested in CODE?
No

### DIFF
--- a/common/app/model/pressedContent.scala
+++ b/common/app/model/pressedContent.scala
@@ -1,5 +1,6 @@
 package model.pressed
 
+import com.gu.contentapi.client.model.v1.ElementType
 import com.gu.facia.api.utils.FaciaContentUtils
 import com.gu.facia.api.{models => fapi, utils => fapiutils}
 import com.gu.facia.client.models.{Backfill, Branded, CollectionConfigJson, Metadata}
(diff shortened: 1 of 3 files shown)


akash1810 (approved):
nice

facia/test/FaciaControllerTest.scala
```diff
@@ -140,12 +140,6 @@ import controllers.FaciaControllerImpl
     faciaController.alternativeEndpoints("uk/business/stock-markets") should be (List("business", "uk"))
   }
 
-  it should "render fronts with content that has been pre-fetched from facia-press" in {
-    // make sure you switch on facia inline-embeds switch before you press to make this pass
-    val request = FakeRequest("GET", "/inline-embeds").from(Uk)
```
gidsg: @johnduffell @rich-nguyen I deleted this test as the `/inline-embeds` front no longer exists on CODE, so I cannot generate the test data to make this test pass anymore. If this test is important to you, please publish some test data on CODE to make it pass.

johnduffell: as discussed before, I still you are responsible for deleting the test, not other people for fixing it.  I took 5 mins to take a look, and it seems you should be repressing from dev not code, and the front is there on dev as far as I can tell from searching the S3 bucket.
https://github.com/guardian/frontend/blob/3ac4896df5408db24b597ce20e58ca454d646771/docs/03-dev-howtos/06-repressing.md
Once you've fixed it, feel free to add a comment to help out the next person who has this problem!

gidsg:
I have now modelled the new field as an `Option[Boolean]` this gets rid of the issues around the pressed data tests needing to be updated and removes release complexity around adding a new mandatory field. The downside is there will be a brief period where the container will be empty (before the fronts are re-pressed populating the Option)

PR #15718, opened 2017-01-30
Remove NPE on adblock detection under phantomJS

MatthewJWalls:
## What does this change?

Small tweak to fix an NPE exception generated by the adblocker detection code, where 'adStyles.getPropertyValue('-moz-binding')' returns null and is then dereferenced.

This should remove one of the exceptions from the PRBuilds thrown-exceptions.txt file, reducing noise for devs.

no

### DIFF
--- a/common/app/templates/inlineJS/nonBlocking/detectAdblock.scala.js
+++ b/common/app/templates/inlineJS/nonBlocking/detectAdblock.scala.js
@@ -28,7 +28,9 @@ try {
                 var adStyles = window.getComputedStyle(ad);
 
                 // Due to an update to AdblockPlus we now need an extra check for adblock being used on Firefox as the sacrificial ad check does not work.
-                adBlockers.active = adStyles.getPropertyValue('display') === 'none'|| adStyles.getPropertyValue('-moz-binding').indexOf('about:')!== -1;
+                var displayProp = adStyles.getPropertyValue('display');
+                var mozBindingProp = adStyles.getPropertyValue('-moz-binding');
+                adBlockers.active = displayProp === 'none' || (mozBindingProp !== null && mozBindingProp.indexOf('about:') !== -1);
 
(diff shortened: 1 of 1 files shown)


common/app/templates/inlineJS/nonBlocking/detectAdblock.scala.js
```diff
@@ -28,7 +28,9 @@ try {
                 var adStyles = window.getComputedStyle(ad);
 
                 // Due to an update to AdblockPlus we now need an extra check for adblock being used on Firefox as the sacrificial ad check does not work.
-                adBlockers.active = adStyles.getPropertyValue('display') === 'none'|| adStyles.getPropertyValue('-moz-binding').indexOf('about:')!== -1;
+                var displayProp = adStyles.getPropertyValue('display');
+                var mozBindingProp = adStyles.getPropertyValue('-moz-binding');
+                adBlockers.active = displayProp === 'none' || (mozBindingProp !== null && mozBindingProp.indexOf('about:') !== -1);
```
katebee: Do we need the `!== null`? Would this work or is the extra specificity required?

`(mozBindingProp && mozBindingProp.indexOf('about:') !== -1)`

MatthewJWalls: I think that'd definitely work. Not sure if we have a preferred style though? There's !x, x != null, typeof(x) !== 'null' etc

gtrufitt:
image

common/app/templates/inlineJS/nonBlocking/detectAdblock.scala.js
SiAdcock: I prefer the style @katebee suggested, although we should perhaps write a lint rule to enforce this!

katebee (approved):
👍

PR #15719, opened 2017-01-30
Fix amp facebook embed validation

michaelwmcnamara:
This change means that facebook video embeds will now be cleaned and converted to amp-facebook elements.
This changes AmpEmbedCleaner and AmpEmbedCleanerTests. 
It also adds a script required by the amp-facebook element to the header.

This change will mean that pages with facebook videos embedded in the article body will now validate successfully.

This change will only impact Amp.

## Tested in CODE?
No.

Testing:
All article unit tests pass;
All amp embed cleaner tests pass;
Confirmed that pages with Facebook embeds display correctly and validate successfully;
Confirmed that pages with Youtube embeds continue to display correctly and validate successfully;
Confirmed that pages with Vimeo embeds continue to display correctly and validate successfully.

cc @guardian/dotcom-platform


common/app/views/support/cleaner/AmpEmbedCleaner.scala
```diff
@@ -66,15 +70,26 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
...
         iframeElement <- videoElement.getElementsByTag("iframe")
         ampExternalVideo <- getAmpExternalVideoByUrl(videoElement.attr("data-canonical-url"))
       } yield {
-        val ampVideoElement = createElement(document, ampExternalVideo.videoId, ampExternalVideo.elementType)
-        videoElement.appendChild(ampVideoElement)
-        iframeElement.remove()
+        val ampVideoElement = ampExternalVideo.elementType match {
+          case "amp-facebook" => createFacebookVideoElement(document, ampExternalVideo.videoId, ampExternalVideo.elementType)
+          case _ => createElement(document, ampExternalVideo.videoId, ampExternalVideo.elementType)
+        }
```
jfsoul: I would put this logic in `createElement` instead of here.

common/app/views/support/cleaner/AmpEmbedCleaner.scala
```diff
@@ -66,15 +70,26 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
       video.attr("layout", "responsive")
     }
 
+    def createFacebookVideoElement(document: Document, videoId: String, elementType: String): Element = {
+      val video = document.createElement(elementType)
+      video.attr("data-href", s"https://www.facebook.com/theguardian/videos/$videoId")
+      video.attr("data-embed-as", "video")
+      video.attr("width", "5")
+      video.attr("height", "3")
+      video.attr("layout", "responsive")
```
jfsoul: Duplication of code with `createElement` which I think we should be able to combine.

Perhaps we should have something a bit like:

`sealed abstract class AmpExternalVideo(val videoId: String, val elementType: String, val customAttributes: Map[String, String])`

...and pass the custom attributes in the case class definition.  It might need to be a bit smarter than this because videoId is needed to define the `href`.  Simple (I haven't thought this through) solution would be to replace `Map[String, String]` with `String => Map[String, String]` function.

common/test/views/support/cleaner/AmpEmbedCleanerTest.scala
```diff
@@ -173,13 +181,47 @@ class AmpEmbedCleanerTest extends FlatSpec with Matchers {
     result.getElementsByTag("amp-vimeo").size should be(0)
   }
 
-  "AmpEmbedCleaner" should "be able to create an amp-youtube element and an amp-vimeo in the same document" in {
+  "AmpEmbedCleaner" should "replace an iframe in a http Facebook video-element with an amp-facebook element" in {
```
jfsoul: Can we remove this test? Looks the same as the one below.

common/app/views/support/cleaner/AmpEmbedCleaner.scala
```diff
@@ -45,15 +45,19 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
...
+  case class FacebookExternalVideo(override val videoId: String) extends AmpExternalVideo(videoId, "amp-facebook")
+
 
 
   object AmpExternalVideo {
-    def getAmpExternalVideoByUrl(videoUrl: String) : Option[AmpExternalVideo] = {
+    def getAmpExternalVideoByUrl(videoUrl: String): Option[AmpExternalVideo] = {
       val youtubePattern = """^https?:\/\/www\.youtube\.com\/watch\?v=([^#&?]+).*""".r
       val vimeoPattern = """^https?:\/\/vimeo\.com\/(\d+).*""".r
+      val facebookPattern = """^https?:\/\/www\.facebook\.com\/theguardian\/videos\/(\d+).*\/""".r
```
jfsoul: Do we need the `.*` after the `(\d+)`?

michaelwmcnamara: Actually, no we dont. Good catch! I'll strip it out in next commit.

common/test/views/support/cleaner/AmpEmbedCleanerTest.scala
michaelwmcnamara: It was testing for http facebook embeds. Buuut looking at facebook again, I don't think its possible to embed a facebook video as http - not through our tools anyway. Stripping it out.

common/app/views/support/cleaner/AmpEmbedCleaner.scala
michaelwmcnamara: Nice idea - this is done in next commit.

common/app/views/support/cleaner/AmpEmbedCleaner.scala
michaelwmcnamara: Interesting idea! I will see what I can come up with on this one. Next commit is merging the create facebook element method into create element, but this seems like a more elegant way to do things. I will have a go at this tomorrow.

common/app/views/support/cleaner/AmpEmbedCleaner.scala
```diff
@@ -42,28 +42,35 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
...
 
-    def createElement(document: Document, videoId: String, elementType: String): Element = {
+    def createElement(document: Document, elementType: String, customAttributes: Map[String, String]): Element = {
       val video = document.createElement(elementType)
-      video.attr("data-videoid", videoId)
-      video.attr("width", "5")
-      video.attr("height", "3")
-      video.attr("layout", "responsive")
+      customAttributes.foreach{
+        case (key, value) => video.attr(key, value)
```
jfsoul: If `video.attr` was a scala method I think you could do `customAttributes.foreach((video.attr _).tupled)` which converts the tuple to method arguments.  Making it a scala method is probably less tidy than your solution so leave this as it is 😁

common/app/views/support/cleaner/AmpEmbedCleaner.scala
```diff
@@ -42,28 +42,35 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
     })
   }
 
-  sealed abstract class AmpExternalVideo(val videoId: String, val elementType: String)
-  case class YoutubeExternalVideo(override val videoId: String) extends AmpExternalVideo(videoId, "amp-youtube")
-  case class VimeoExternalVideo(override val videoId: String) extends AmpExternalVideo(videoId, "amp-vimeo")
 
+  sealed abstract class AmpExternalVideo(val videoId: String, val elementType: String, customAttributes: String => Map[String, String]){
+    def getAttributes = customAttributes(videoId) ++ Map(("width", "5"), ("height", "3"), ("layout", "responsive"))
```
jfsoul: I'd like to see this Map separated out into a `val commonAttributes = ...`

common/app/views/support/cleaner/AmpEmbedCleaner.scala
```diff
@@ -42,28 +42,35 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
...
+  }
+
+  case class YoutubeExternalVideo(override val videoId: String) extends AmpExternalVideo(videoId, "amp-youtube", customAttributes = id => Map(("data-videoid", id)))
+  case class VimeoExternalVideo(override val videoId: String) extends AmpExternalVideo(videoId, "amp-vimeo", customAttributes = id => Map(("data-videoid", id)))
+  case class FacebookExternalVideo(override val videoId: String) extends AmpExternalVideo(videoId, "amp-facebook", customAttributes = id => Map(("data-href", s"https://www.facebook.com/theguardian/videos/$id"), ("data-embed-as", "video")))
```
jfsoul: Again for these I would have `val facebookCustomAttributes` etc.

common/app/views/support/cleaner/AmpEmbedCleaner.scala
```diff
@@ -42,28 +42,35 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
...
 
+  
   object AmpExternalVideo {
```
jfsoul: General thought about this file not really related to these changes: it would be nice to put `private` on everything that's intended to be.

jfsoul:
Changes look good apart from my minor comments.

common/app/views/support/cleaner/AmpEmbedCleaner.scala
michaelwmcnamara: Good to know though - I'll do some reading on this for myself :-)

common/app/views/support/cleaner/AmpEmbedCleaner.scala
michaelwmcnamara: should be in next commit.

common/app/views/support/cleaner/AmpEmbedCleaner.scala
michaelwmcnamara: should be in next commit

common/app/views/support/cleaner/AmpEmbedCleaner.scala
michaelwmcnamara: I guess this one is best done in its own pr :-)

common/app/views/support/cleaner/AmpEmbedCleaner.scala
```diff
@@ -42,28 +42,44 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
...
+
+  def standardAttributes(videoId: String): Map[String,String] = { Map(("data-videoid", videoId)) }
+
+  def facebookAttributes(videoId: String): Map[String, String] = {
+    Map(("data-href", s"https://www.facebook.com/theguardian/videos/$videoId"), ("data-embed-as", "video"))
+  }
+
+
+
+  case class YoutubeExternalVideo(override val videoId: String) extends AmpExternalVideo(videoId, "amp-youtube", customAttributes = id => standardAttributes(id))
```
jfsoul: `case class YoutubeExternalVideo(override val videoId: String) extends AmpExternalVideo(videoId, "amp-youtube", standardAttributes(videoId))`

and

`sealed abstract class AmpExternalVideo(val videoId: String, val elementType: String, customAttributes: Map[String, String])`

...would be cleaner.  It was probably my original suggestion that led you to this, my bad!

common/app/views/support/cleaner/AmpEmbedCleaner.scala
```diff
@@ -42,28 +42,44 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
...
-  case class VimeoExternalVideo(override val videoId: String) extends AmpExternalVideo(videoId, "amp-vimeo")
+
+  sealed abstract class AmpExternalVideo(val videoId: String, val elementType: String, customAttributes: String => Map[String, String]){
+    val commonVideoAttributes = Map(("width", "5"), ("height", "3"), ("layout", "responsive"))
+    def getAttributes = customAttributes(videoId) ++ commonVideoAttributes
+  }
+
+  def standardAttributes(videoId: String): Map[String,String] = { Map(("data-videoid", videoId)) }
+
+  def facebookAttributes(videoId: String): Map[String, String] = {
```
jfsoul: don't need the curly braces here I think

common/app/views/support/cleaner/AmpEmbedCleaner.scala
```diff
@@ -42,28 +42,44 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
...
-  sealed abstract class AmpExternalVideo(val videoId: String, val elementType: String)
-  case class YoutubeExternalVideo(override val videoId: String) extends AmpExternalVideo(videoId, "amp-youtube")
-  case class VimeoExternalVideo(override val videoId: String) extends AmpExternalVideo(videoId, "amp-vimeo")
+
+  sealed abstract class AmpExternalVideo(val videoId: String, val elementType: String, customAttributes: String => Map[String, String]){
+    val commonVideoAttributes = Map(("width", "5"), ("height", "3"), ("layout", "responsive"))
+    def getAttributes = customAttributes(videoId) ++ commonVideoAttributes
+  }
+
+  def standardAttributes(videoId: String): Map[String,String] = { Map(("data-videoid", videoId)) }
```
jfsoul: Can remove these curly braces too

common/app/views/support/cleaner/AmpEmbedCleaner.scala
```diff
@@ -42,28 +42,44 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
     })
   }
 
-  sealed abstract class AmpExternalVideo(val videoId: String, val elementType: String)
-  case class YoutubeExternalVideo(override val videoId: String) extends AmpExternalVideo(videoId, "amp-youtube")
-  case class VimeoExternalVideo(override val videoId: String) extends AmpExternalVideo(videoId, "amp-vimeo")
+
+  sealed abstract class AmpExternalVideo(val videoId: String, val elementType: String, customAttributes: String => Map[String, String]){
+    val commonVideoAttributes = Map(("width", "5"), ("height", "3"), ("layout", "responsive"))
+    def getAttributes = customAttributes(videoId) ++ commonVideoAttributes
```
jfsoul: If you decide to do my other suggestion then this would become `val attributes = ` I suppose

jfsoul (approved):

PR #15720, opened 2017-01-30
Retry CSS parsing on failure

joelochlann:
We have an intermittent bug (it's occurred twice now, as far as we know, but I can't reproduce it) which seems to be caused by an exception thrown from within `.parseStyleSheet`. Thanks to `Try`, the exception is swallowed and the fallback behaviour results in lots of the styles remaining in the `<head>` when they should be inlined - meaning the email looks horrific in Gmail.

I've added in some logging if parsing fails, so we can get some extra info from the logs if it ever happens again. I've also added in some retry logic, so it will make 3 attempts to parse if parsing fails. If the failure is some strange transient kind of error, there's a slim chance that trying again might succeed and mean we at least avoid sending out an ugly email. In any case, any failed attempts will be logged.

### DIFF
--- a/common/app/common/InlineStyles.scala
+++ b/common/app/common/InlineStyles.scala
@@ -93,27 +93,7 @@ object InlineStyles {
     Html(document.toString)
   }
 
-  /**
-    * Execute some code n times, or until no exception is thrown
-    * (If n is 0, code still gets executed once)
-    */
-  def retry[T](n: Int)(r: => T)(implicit onFail: (Throwable, Int) => Unit) = {
-    def go(i: Int): Try[T] = {
-      Try(r) match {
-        case Failure(e) if i < n => {
-          onFail(e, i)
-          go(i + 1)
-        }
-        case f@Failure(e) => {
-          onFail(e, i)
-          f
-        }
-        case s@Success(v) => s
-      }
-    }
 
-    go(1)
-  }
 
   /**
     * Convert the styles in a document's <style> tags to a pair.
@@ -123,7 +103,7 @@ object InlineStyles {
     document.getElementsByTag("style").foldLeft((Seq.empty[CSSRule], Seq.empty[String])) { case ((inline, head), element) =>
       val source = new InputSource(new StringReader(element.html))
 
-      retry(3)(cssParser.parseStyleSheet(source, null, null)) { (e: Throwable, i: Int) =>
+      Retry(3)(cssParser.parseStyleSheet(source, null, null)) { (e: Throwable, i: Int) =>
         Logger.error(s"Attempt ${i} to parse stylesheet failed", e)
       } match {
         case Failure(exception) => {
--- a/common/app/common/InlineStyles.scala
+++ b/common/app/common/InlineStyles.scala
@@ -4,14 +4,15 @@ import java.io.StringReader
 
 import org.jsoup.Jsoup
 import org.jsoup.nodes.Document
-import com.steadystate.css.parser.{SACParserCSS3, CSSOMParser}
+import com.steadystate.css.parser.{CSSOMParser, SACParserCSS3}
 import org.w3c.css.sac.InputSource
-import org.w3c.dom.css.{CSSRule => W3CSSRule, CSSRuleList}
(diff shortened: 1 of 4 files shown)


common/app/common/InlineStyles.scala
```diff
@@ -93,21 +94,48 @@ object InlineStyles {
...
+    * (If n is 0, code still gets executed once)
+    */
+  def retry[T](n: Int)(r: => T)(implicit onFail: (Throwable, Int) => Unit) = {
+    def go(i: Int): Try[T] = {
+      Try(r) match {
+        case Failure(e) if i < n => {
+          onFail(e, i)
+          go(i + 1)
+        }
+        case f@Failure(e) => {
```
davidfurey: I think this would be easier to read with more whitespace `case f @ Failure(e) =>`

common/app/common/InlineStyles.scala
```diff
@@ -93,21 +94,48 @@ object InlineStyles {
...
+          onFail(e, i)
+          f
+        }
+        case s@Success(v) => s
```
davidfurey: If you aren't going to using `v` there is no point in capturing it. Could use underscore, or
`case s: Success[T] =>`

common/app/common/InlineStyles.scala
```diff
@@ -93,21 +94,48 @@ object InlineStyles {
   }
 
   /**
+    * Execute some code n times, or until no exception is thrown
```
davidfurey: I think that since this is fairly generic retry logic, it probably belongs outside of InlineStyles

common/app/common/InlineStyles.scala
```diff
@@ -93,21 +94,48 @@ object InlineStyles {
...
 
-      Try(cssParser.parseStyleSheet(source, null, null)).toOption map { sheet =>
-        val (styles, others) = seq(sheet.getCssRules).partition(isStyleRule)
-        val (inlineStyles, headStyles) = styles.flatMap(CSSRule.fromW3).flatten.partition(_.canInline)
-        val newHead = (headStyles.map(_.toString) ++ others.map(_.getCssText)).mkString("\n")
-
-        (inline ++ inlineStyles, (head :+ newHead).filter(_.nonEmpty))
-      } getOrElse {
-        (inline, head :+ element.html)
+      retry(3)(cssParser.parseStyleSheet(source, null, null)) { (e: Throwable, i: Int) =>
```
davidfurey: Throwable and Int should be inferred for you here.  Longer variable names might be more helpful than types in this case.

common/app/common/InlineStyles.scala
```diff
@@ -93,21 +94,48 @@ object InlineStyles {
...
+        Logger.error(s"Attempt ${i} to parse stylesheet failed", e)
```
davidfurey: `{}` are unnecessary here `$i` is fine

davidfurey (dismissed):

common/app/common/InlineStyles.scala
```diff
@@ -93,21 +94,48 @@ object InlineStyles {
   }
 
   /**
+    * Execute some code n times, or until no exception is thrown
+    * (If n is 0, code still gets executed once)
+    */
+  def retry[T](n: Int)(r: => T)(implicit onFail: (Throwable, Int) => Unit) = {
```
TBonnin: The `dispatch` module, already included in this project, has a Retry mechanism. I don't think you need to rewrite your own.

It is used here for instance https://github.com/guardian/frontend/blob/master/onward/app/weather/WeatherApi.scala#L58

TBonnin (dismissed):

common/app/common/InlineStyles.scala
joelochlann: Thanks @TBonnin - unfortunately that `dispatch` module is designed for asynchronous requests so it requires the use of Futures which in this case are unnecessary and misleading. Hopefully this `retry` method won't be around forever, since if this extra logging helps us diagnose and fix the underlying problem, it will be redundant.

davidfurey:
👍

PR #15721, opened 2017-01-31
Remove hardcoded legacy hosted pages 🔥 

lps88:
All the original campaigns are finished and we can clean up the codebase! (finally!)


kelvin-chappell (approved):
Hurray! 👍

PR #15724, opened 2017-01-31
set caption--main class on media atom main media

gidsg:
## What does this change?
Pass `mainMedia` attribute to `youtube` template caption, so that it inherits the corrects style.

Consistent styling across elements

No

## Screenshots
Before
screen shot 2017-01-31 at 12 43 34
After
screen shot 2017-01-31 at 12 42 43

### DIFF
--- a/common/app/views/fragments/atoms/media.scala.html
+++ b/common/app/views/fragments/atoms/media.scala.html
@@ -4,8 +4,8 @@
 @{
     media match {
             case posterOnly if media.assets.isEmpty && media.posterImage.isDefined  => views.html.fragments.atoms.posterImage(mainMedia = mainMedia, amp = amp,picture = media.posterImage.get, caption = media.title)
-            case youtube if media.assets.headOption.filter(_.platform == MediaAssetPlatform.Youtube) => if(amp) views.html.fragments.atoms.ampYoutube(media, displayCaption, mainMedia) else views.html.fragments.atoms.youtube(media, displayCaption, embedPage, displayEndSlate)
-            case genericAsset if media.assets.nonEmpty => views.html.fragments.atoms.genericMedia(media, displayCaption, amp)
+            case youtube if media.assets.headOption.filter(_.platform == MediaAssetPlatform.Youtube) => if(amp) views.html.fragments.atoms.ampYoutube(media = media, displayCaption = displayCaption, mainMedia = mainMedia) else views.html.fragments.atoms.youtube(media = media, displayCaption = displayCaption, embedPage = embedPage, displayEndSlate = displayEndSlate, mainMedia = mainMedia)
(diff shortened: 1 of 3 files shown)


common/app/views/fragments/atoms/mediaAtomCaption.scala.html
```diff
@@ -4,7 +4,6 @@
 <figcaption class="@RenderClasses(Map(
     "caption" -> true,
     "caption--img" -> true,
-    "caption--media-atom" -> true,
```
gidsg: The rules associated with this class were removed in a previous PR so it can be safely deleted.

common/app/views/fragments/atoms/media.scala.html
```diff
@@ -4,8 +4,8 @@
...
             case posterOnly if media.assets.isEmpty && media.posterImage.isDefined  => views.html.fragments.atoms.posterImage(mainMedia = mainMedia, amp = amp,picture = media.posterImage.get, caption = media.title)
-            case youtube if media.assets.headOption.filter(_.platform == MediaAssetPlatform.Youtube) => if(amp) views.html.fragments.atoms.ampYoutube(media, displayCaption, mainMedia) else views.html.fragments.atoms.youtube(media, displayCaption, embedPage, displayEndSlate)
-            case genericAsset if media.assets.nonEmpty => views.html.fragments.atoms.genericMedia(media, displayCaption, amp)
+            case youtube if media.assets.headOption.filter(_.platform == MediaAssetPlatform.Youtube) => if(amp) views.html.fragments.atoms.ampYoutube(media = media, displayCaption = displayCaption, mainMedia = mainMedia) else views.html.fragments.atoms.youtube(media = media, displayCaption = displayCaption, embedPage = embedPage, displayEndSlate = displayEndSlate, mainMedia = mainMedia)
+            case genericAsset if media.assets.nonEmpty => views.html.fragments.atoms.genericMedia(media = media, displayCaption = displayCaption, amp = amp)
             case _ =>
```
GHaberis: Do you have to set the values of the arguments media, displayCaption, mainMedia etc? eg. Could (media = media, displayCaption = displayCaption, mainMedia = mainMedia) not just be (media, displayCaption, mainMedia)?

gidsg: In the `youtube` case I do, because the `youtube` fragment contains some parameters with default values and some without. If you just use the names, then the `mainMedia` value is actually overriding a different value than you would expect from the name. In the other items it is not required but I thought it better to be consistent across the file.

gidsg: ...the other option is to not default the value of `mainMedia` in `youtube.scala.html` but then you have to explicitly set it in every usage.

GHaberis (approved):

PR #15727, opened 2017-01-31
Lint for ambiguous use of 'guardian' and global 'config'

sndrs:
## What does this change?

- sets up our own eslint config for app-specific rules
- adds lint rule to prevent ambiguous use of `guardian`
- adds our own lint rule to prevent direct use of `window.guardian.config`

ENFORCE RULES


SiAdcock (approved):
Top stuff 👍

PR #15728, opened 2017-01-31
Constrain billboards to avoid margin auto behaviour

regiskuckaertz:
Before:
picture 4

After:
picture 5


lps88:
I can't reproduce this exact problem locally, but it seems even after this fix the banner is not centered, is it supposed to look like this?

image

lps88 (approved):
Ah, if it is supposed to line up with content column this arrangement makes sense 👍

regiskuckaertz:
[Top banner positioning bug](https://trello.com/c/PBi14UCQ/161-top-banner-positioning-bug)

PR #15731, opened 2017-02-01
Amend supporter bundles thrasher AB test audience

JustinPinner:
## What does this change?
Increase audience capture size from 1.6% to 3% per variant as we're not seeing enough click-throughs to attain statistical significance in a useful timeframe.

The test will need to run for too long to arrive at a meaningful conclusion at its current rate.

We're able to tweak the participation numbers as we're sampling a random audience, so that won't invalidate the data we've already captured.

No, web only.

## Screenshots
See https://github.com/guardian/frontend/pull/15661

## Tested in CODE?
~~Not yet.~~ ~~In progress.~~ Yes.


Ap0c (approved):
🎩

PR #15733, opened 2017-02-01
Commercial modules timing makeover

regiskuckaertz:
Makes all modules responsible for tracking their running time. Currently those which don't use custom timing are measured against the baseline, which means the compounded running time of previous modules is included in their timing.

Also fixes a bug where the hosted gallery timing would start but never end.

With this change, we should see minor improvements in the timings.


static/src/javascripts-legacy/projects/commercial/modules/hosted/onward-journey-carousel.js
```diff
@@ -64,8 +68,12 @@ define([
 
     };
 
-    function init() {
-        new HostedCarousel().bindButtons();
+    function init(moduleName) {
+        if (qwery('.js-carousel-pages').length) {
+            performanceLogging.moduleStart(moduleName);
```
rich-nguyen: wrap outside `if` ? one instruction is not being logged!

rich-nguyen:
Yep, that's great. I think very soon we can remove baselines completely. The only thing I considered is that a number of the `moduleEnd` calls are located within a promise `then`, so it might be worth checking if a `catch` is needed too? Otherwise 👍

rich-nguyen (approved):

regiskuckaertz:
[Commercial modules timing makeover](https://trello.com/c/02AzjL1Z/160-commercial-modules-timing-makeover)

static/src/javascripts-legacy/projects/commercial/modules/article-body-adverts.js
```diff
@@ -46,12 +50,16 @@ define([
...
+            im.then(waitForMerch).then(addInlineAds).then(moduleEnd);
             return im;
         }
 
-        addInlineAds();
+        addInlineAds().then(moduleEnd);
         return Promise.resolve(true);
+
+        function moduleEnd() {
+            performanceLogging.moduleEnd(moduleName);
```
katebee: Q: Is there a reason for the use of `moduleEnd()` over `.bind` in this case?

Other modules are using: 
`.then(performanceLogging.moduleEnd.bind(null, moduleName))` 🤔

regiskuckaertz: Just because here I did not want to curry twice. In any case, I've been thinking about it and it remains the fact that most modules call start and end at the edges, and so we could probably rewrite them as

```
function(moduleName, module) {
  performanceLogging.moduleStart(moduleName);
  module().then(function () {
    performanceLogging.moduleEnd(moduleName);
  });
}
```

... the problem is, what about those modules that don't comply with this pattern (approx. 1/4 of them)? I'm trying to find a nice composable pattern.

static/src/javascripts-legacy/projects/commercial/modules/hosted/gallery.js
```diff
@@ -451,31 +451,35 @@ define([
...
-        });
+                return gallery;
+            })
+            .then(function (gallery) {
+                performanceLogging.moduleEnd(moduleName)
+                mediator.emit('page:hosted:gallery', gallery);
+            });
+        }
+
+        return Promise.resolve();
```
katebee: Niiiccce 🎉

static/src/javascripts-legacy/projects/commercial/modules/hosted/onward-journey-carousel.js
```diff
@@ -64,8 +68,12 @@ define([
 
     };
 
-    function init() {
-        new HostedCarousel().bindButtons();
+    function init(moduleName) {
+        if (qwery('.js-carousel-pages').length) {
```
katebee: Is it better to use `qwery` here? Or could `$` be used instead?

static/test/javascripts/spec/commercial/article-aside-adverts.spec.js
```diff
@@ -1,13 +1,16 @@
+
```
katebee: Unintentional whitespace

katebee (approved):
👍

static/src/javascripts-legacy/projects/commercial/modules/hosted/onward-journey-carousel.js
regiskuckaertz: The latter wraps the result in a `Bonzo` and I don't need that 😄

static/src/javascripts-legacy/projects/commercial/modules/hosted/onward-journey-carousel.js
regiskuckaertz: Yes, I think I made a mistake by moving everything inside the modules. There still is two different behaviours and they should be processed differently. One is:

```
start();
module().then(stop);
```

The other is

```
module(start, stop);
```

I will try to refactor by taking this into account, I think the code will be more readable that way. Perhaps this is a good user case to try [railway-oriented programming](https://fsharpforfunandprofit.com/rop/)

regiskuckaertz:
@rich-nguyen @katebee major changes since you last reviewed, you may want to have a second look.

PR #15734, opened 2017-02-01
Expand ad slot for outstream ads

regiskuckaertz:
The new teads remain inside the ad slot, we must make sure it can accommodate the creative by updating its dimensions.

Before:
picture 3

After:
picture 2


katebee (approved):
👍

PR #15741, opened 2017-02-02
Fix left margin for the top slot when page skin

regiskuckaertz:
Before:
picture 1

After:
picture 2


scottp719 (approved):
👍 When tests pass

PR #15743, opened 2017-02-02
Ad label fix

lps88:
A css class occasionally gets wrongly applied to the ad label causing this issue:
image

### DIFF
--- a/static/src/javascripts-legacy/projects/commercial/modules/dfp/render-advert.js
+++ b/static/src/javascripts-legacy/projects/commercial/modules/dfp/render-advert.js
@@ -136,12 +136,12 @@ define([
     sizeCallbacks[adSizes.merchandising] = addFluid250(['ad-slot--commercial-component']);
 
     /**
-     * @param adSlotId - DOM ID of the rendered slot
+     * @param advert - as defined in commercial/modules/dfp/Advert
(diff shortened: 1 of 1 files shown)


static/src/javascripts-legacy/projects/commercial/modules/dfp/render-advert.js
```diff
@@ -136,12 +136,12 @@ define([
...
 
     /**
-     * @param adSlotId - DOM ID of the rendered slot
+     * @param advert - as defined in commercial/modules/dfp/Advert
      * @param slotRenderEvent - GPT slotRenderEndedEvent
      * @returns {Promise} - resolves once all necessary rendering is queued up
      */
     function renderAdvert(advert, slotRenderEvent) {
-        removePlaceholders(advert.node);
+        addContentClass(advert.node);
```
regiskuckaertz: I think it would be easier for you if this function was called in the promise chain, for example: `Promise.resolve(addContentClass(advert.node)).then(applyCreativeTemplate)...`

... that way you won't need to test for the label. All you need to do is return the fastdom promise and return `adSlotNode` inside of it.

regiskuckaertz:
Great 👍  At some point we will need to refactor the whole `renderAdvert` pipeline, there are too many fastdom calls to my taste.

PR #15747, opened 2017-02-02
use smaller enumeratum-play-json dependency

gidsg:
## What does this change?
use smaller [enumeratum-play-json](https://github.com/lloydmeta/enumeratum#play-json) dependency instead of `enumeratum-play` as we are not using any of the additional play features.

12kb JAR ➡️ 3kb JAR

No

## Screenshots
N/A

## Tested in CODE?
No


TBonnin (approved):
looks fine. Thank you

PR #15753, opened 2017-02-03
Implement a simple version of the  new footer on AMP

NataliaLKB:
## What does this change?
The new footer was implemented in the new header test, but not on AMP where the new header already is. This fixes that discrepancy. I think its likely that @zeftilldeath will make some edits to this 🎨 

@guardian/dotcom-platform 

Most consistent visually

Only AMP

## Screenshots
image

## Tested in CODE?
No

### DIFF
--- a/static/src/stylesheets/amp/_footer.scss
+++ b/static/src/stylesheets/amp/_footer.scss
@@ -9,15 +9,15 @@
 }
 
 .l-footer__secondary {
-    font-size: 12px;
+    @include content-gutter();
     line-height: 1;
-    padding-right: $gs-gutter / 2;
-    padding-left: $gs-gutter / 2;
     font-family: 'Helvetica Neue', Helvetica, Arial, 'Lucida Grande', sans-serif;
 
     a {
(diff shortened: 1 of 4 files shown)


static/src/stylesheets/amp/_footer.scss
```diff
@@ -91,32 +47,68 @@ $c-primary-footer-background-side-bar: mix($guardian-brand, #ffffff, 90%);
     border-top: 1px solid $neutral-2;
 }
 
-.breadcrumb {
-    text-transform: lowercase;
+.footer__back-to-top__container {
+    @include content-gutter();
+    background-color: $guardian-brand;
+
+    a:hover > &,
```
SiAdcock: Nit: no need for `:hover` in AMP

SiAdcock (approved):
Awesome stuff 👍

static/src/stylesheets/amp/_footer.scss
NataliaLKB: Good point!! 👍

PR #15755, opened 2017-02-03
Don't show address type for restaurant reviews if part isn't present.

LATaylor-guardian:
## What does this change?
This change makes showing the address block of structured data for a restaurant review optional depending if we have any parts of the address to display.

Providing richer structure for reviews via Google rich cards.

Can effect AMP.

## Screenshots
None.

## Tested in CODE?
Tested AMP validation locally.

@mchv @stephanfowler


mchv:
👍

PR #15756, opened 2017-02-03
support multiples of the same youtube atoms on the same page

GHaberis:
## What does this change?

Fixes a bug which prevented a youtube atom being enhanced if the same atom existed elsewhere on the page (because they shared the same ID).

We can have an atom on the page twice.

No

## Screenshots

N/A

## Tested in CODE?

No

@akash1810 @gidsg

### DIFF
--- a/static/src/javascripts-legacy/projects/common/modules/atoms/youtube.js
+++ b/static/src/javascripts-legacy/projects/common/modules/atoms/youtube.js
@@ -35,7 +35,7 @@ define([
 
         killProgressTracker(atomId);
         setProgressTracker(atomId);
-        tracking.track('play', atomId);
+        tracking.track('play', getTrackingId(atomId));
 
         if (player.endSlate &&
             !player.overlay.parentNode.querySelector('.end-slate-container')) {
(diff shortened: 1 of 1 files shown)


static/src/javascripts-legacy/projects/common/modules/atoms/youtube.js
```diff
@@ -177,27 +175,32 @@ define([
...
                 youtubePlayer.init(iframe, {
                     onPlayerReady: onPlayerReady.bind(null, atomId, overlay),
                     onPlayerStateChange: onPlayerStateChange.bind(null, atomId)
-                }, youtubeId);
+                }, iframe.id);
             });
         });
     }
 
+    function getTrackingId(atomId) {
```
akash1810: could we add a comment as to why this is needed?

akash1810 (approved):
could we add an inline comment, other than that 👍

gidsg:
Am I right in understanding the fundamental issue is multiple videos having the same `iframe id`? If so perhaps we could solve this server side, say by appending a UUID to the iframe ID with a delimiting character? 

Happy for this to go out as-is though and we can think about this in a subsequent PR.

static/src/javascripts-legacy/projects/common/modules/atoms/youtube.js
GHaberis: Yep, I'll do that!

gidsg (approved):

gidsg:
...actually I think a completely random string is not viable as it would dilute the cache. Back to the drawing board!

PR #15757, opened 2017-02-03
Add JSON endpoint for inline book component

regiskuckaertz:
Needed for the upcoming native inline book.

picture 2

### DIFF
--- a/commercial/app/controllers/BookOffersController.scala
+++ b/commercial/app/controllers/BookOffersController.scala
@@ -78,6 +78,36 @@ class BookOffersController(bookFinder: BookFinder, bestsellersAgent: Bestsellers
     booksSample(specificIds, segment) map result
   }
 
+  def getBook = Action.async { implicit request =>
+    specificId map { isbn =>
+      bookFinder.findByIsbn(isbn) map {
+        _ map { book =>
+          val json = Json.toJson(book)
+          Cached(60.seconds){
+            JsonComponent(json)
+          }
+        } getOrElse {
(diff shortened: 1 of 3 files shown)


commercial/app/controllers/BookOffersController.scala
```diff
@@ -78,6 +78,36 @@ class BookOffersController(bookFinder: BookFinder, bestsellersAgent: Bestsellers
     booksSample(specificIds, segment) map result
   }
 
+  def getBook = Action.async { implicit request =>
+    specificId map { isbn =>
+      bookFinder.findByIsbn(isbn) map {
+        _ map { book =>
+          val json = Json.toJson(book)
+          Cached(60.seconds){
```
kelvin-chappell: Is the 60s specifically for this route or should it be the same as the other routes?  ie share the value

kelvin-chappell (approved):
👍

commercial/app/controllers/BookOffersController.scala
regiskuckaertz: This is the same for all endpoints I think. What is the best place to store that value? In the config?

kelvin-chappell: It looks like there's a value here: https://github.com/guardian/frontend/blob/master/commercial/app/controllers/package.scala#L12

regiskuckaertz:
[Native Inline book JSON endpoint](https://trello.com/c/2t27A7M3/170-native-inline-book-json-endpoint)

PR #15758, opened 2017-02-03
Make amp-youtube change permanent

gidsg:
## What does this change?
Removes the amp-youtube switch and uses `amp-youtube` permanently instead of `amp-iframe` 

`amp-youtube` is more performant, as `amp-iframe` rendering is deferred until later in the loading process.

AMP only

## Screenshots
N/A

## Tested in CODE?
No


NataliaLKB (approved):
🎉

PR #15763, opened 2017-02-03
Track identityId in Google Analytics

SiAdcock:
## What does this change?

Google Analytics now provides a way of [tracking a user's unique ID](https://developers.google.com/analytics/devguides/collection/analyticsjs/cookies-user-id#user_id) across sessions and devices. This has already been implemented in native apps.

We are currently using the `identityId` for logged-in users to this end, by capturing the ID in [a custom dimension](https://github.com/guardian/frontend/blob/master/common/app/views/fragments/analytics/google.scala.html#L99). This change will also capture the `identityId` in the new `userId` param

This allows us to identify unique users across multiple devices or sessions. This will give us more accurate user data that better reflects the real-world.

No

## Tested in CODE?

No


gtrufitt (approved):
🆒 🌻

PR #15764, opened 2017-02-03
Cast arguments into array

regiskuckaertz:
It turns out `Array.prototype.concat` is pretty adamant that it won't cast array-like objects into arrays. And so, `array.contat(arguments)` will give `[arr0, ..., arrN, arguments]` rather than `[arr0, ..., arrN, arg0, ..., argN]`.


rich-nguyen:
👍

rich-nguyen (approved):

gtrufitt (approved):

PR #15766, opened 2017-02-06
Fix lodash forIn enumerating over inherited properties in AB targeting

regiskuckaertz:
People in the AdOps team noticed they were not always getting back the ads they were supposed to see in a HPTO. That was due to a `toLowerCase` call on a non-String value when building the page targeting. It turns out `forIn` in lodash will enumerate over inherited properties of an object, and because of prototypal inheritance this may lead to unexpected results.

Good thing we can do without lodash for most of our needs, so I cleaned up this module quite a bit.


katebee (approved):
LGTM 👍

regiskuckaertz:
[Roadblocks not being honored.](https://trello.com/c/orzittf6/178-roadblocks-not-being-honored)

PR #15767, opened 2017-02-06
Define a new AB test for A1 vs A2 supporter bundles

JustinPinner:
## What does this change?
(see also https://github.com/guardian/membership-frontend/pull/1483)

Replaces the current A1/B1/B2 bundles test with an A1/A2 version. Having dropped to two cohorts, we're leaving the audience at 15% to gather this round of data quickly.  

We've also changed the cohort names to `control` and `variant`, primarily to keep abacus happy.

The previous test proved which overall layout was more widely appreciated, now we need to prove the demand difference between two variants (ad-free and with-ads) of the winning layout.

No, web only.

## Screenshots
There are no visible differences in frontend, except for the link params from the thrasher's 'find out more' buton, so you can refer to the screen grabs on the original test PR: https://github.com/guardian/frontend/pull/15661.

## Tested in CODE?
~~As soon as I have a PR build...~~ Yes;
screen shot 2017-02-07 at 12 23 55

### DIFF
--- a/static/src/javascripts-legacy/projects/common/modules/experiments/tests/membership-a1-a2-thrasher.js
+++ b/static/src/javascripts-legacy/projects/common/modules/experiments/tests/membership-a1-a2-thrasher.js
@@ -22,7 +22,7 @@ define([
         this.description = 'Test Supporter Bundle A1 (ad-free control) against A2 (with-ads variant)';
         this.showForSensitive = true;
         this.audience = 0.15;  // 7.5% per variant
-        this.audienceOffset = 0;
+        this.audienceOffset = 0.15; // use a new audience segment base
         this.successMeasure = '';
         this.audienceCriteria = 'People on UK network front with at least an 1140px wide display.';
         this.dataLinkNames = '';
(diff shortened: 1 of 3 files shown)


Ap0c (approved):
Minor comment, otherwise LGTM (tested locally) 🌵.

static/src/javascripts-legacy/projects/common/modules/experiments/tests/membership-a1-a2-thrasher.js
```diff
@@ -54,7 +59,7 @@ define([
             if (this.thrasher()) {
                 var linkEl = document.querySelector('.membership-ab-thrasher--wrapper .link-button');
                 if (linkEl && linkEl.getAttribute('href')) {
-                    linkEl.setAttribute('href', config.page.membershipUrl + '/bundles?INTCMP=MEMBERSHIP_AB_THRASHER_' + config.page.edition + '_' + variant.toUpperCase());
+                    linkEl.setAttribute('href', config.page.membershipUrl + '/bundles?INTCMP=MEMBERSHIP_A_THRASHER_' + config.page.edition + '_' + variant.toUpperCase());
```
JustinPinner: I suspect they bypassed the thrasher and edited the url directly, since we only saw one of those. But maybe to be double-certain a force to upper case wouldn't hurt.

PR #15776, opened 2017-02-07
Extend new header switch

NataliaLKB:
## What does this change?
Extends the new header switch until next month. We are still iterating slightly before putting it out to everyone.

Build doesn't break

Nope

## Tested in CODE?
No


stephanfowler:
👍

PR #15780, opened 2017-02-08
Turn sticky top banner into a wrapped module

regiskuckaertz:
The `onFirstRender` call was not blocking on anything, and so the manual module timing did not add any value, so I'm removing it. Besides, there is a slight chance the `trackAdRender` call did not happen before the top banner was displayed ... which would be surprising, but I don't want to take any chances.


katebee (approved):
👍

regiskuckaertz:
[Inline ads not appearing on immersives](https://trello.com/c/TljsE0KL/179-inline-ads-not-appearing-on-immersives)

PR #15782, opened 2017-02-08
AMP + New header tag tweaks

NataliaLKB:
## What does this change?
This changes two things. First, a big where we were accidentally hiding the tags on mobile outside the test.

The second thing that has changed is tweaks  to what shows up in the submeta keyword list. The tone is moved to the end of the list as it isn't as important for the user. The first tag which is the section id and is already showing up about the keyword list is dropped to avoid  the duplication.

cc @guardian/dotcom-platform 

Tweaking what is shown to be more relevant to the user

This affects AMP too

## Screenshots
Before:
image

After:
image

## Tested in CODE?
No


stephanfowler:
👍 - looks so much better!

PR #15783, opened 2017-02-08
Make acquistion test selector respect the impressions from other tests

jranks123:
## What does this change?

Previously, when deciding if an Epic test had reached it's limit, it consulted the log and counted the number of impressions from THAT SPECIFIC epic test.

This meant that if Test A had a limit of, say, 4 in a month, and Test B had a limit of 5 in a month, once they had seen the Epic in Test A 4 times, they could then see the Epic in Test B another 5 times in that month.

As a team we have decided that this is undesirable default behaviour, and if a cap has been set for a test, it should count impressions from other tests in that timeframe when tallying up the impressions. 

Less confusion over what the impression limits actually mean with respect to an individual test. If a test says 4 impressions in 30 days, it means 4 impressions in 4 days, by default (we want to keep the ability to override this for special instances).

### DIFF
--- a/static/src/javascripts-legacy/projects/common/modules/experiments/acquisition-test-selector.js
+++ b/static/src/javascripts-legacy/projects/common/modules/experiments/acquisition-test-selector.js
@@ -26,8 +26,8 @@ define([
                 var variant = segmentUtil.variantFor(t);
 
                 var hasNotReachedRateLimit = variant &&
-                    ((viewLog.viewsInPreviousDays(variant.maxViews.days, t) < variant.maxViews.count &&
-                    viewLog.viewsInPreviousDays(variant.maxViews.minDaysBetweenViews, t) === 0) ||
+                    ((viewLog.viewsInPreviousDays(variant.maxViews.days) < variant.maxViews.count &&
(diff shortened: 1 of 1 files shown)


static/src/javascripts-legacy/projects/common/modules/experiments/acquisition-test-selector.js
```diff
@@ -26,7 +26,7 @@ define([
                 var variant = segmentUtil.variantFor(t);
 
                 var hasNotReachedRateLimit = variant &&
-                    ((viewLog.viewsInPreviousDays(variant.maxViews.days, t) < variant.maxViews.count &&
+                    ((viewLog.viewsInPreviousDays(variant.maxViews.days) < variant.maxViews.count &&
                     viewLog.viewsInPreviousDays(variant.maxViews.minDaysBetweenViews, t) === 0) ||
```
desbo: think we need to remove the `t` here as well?

jranks123: ah yes, done now, good spot

desbo (approved):

PR #15784, opened 2017-02-08
Constraining amp header in footer the same as the body

NataliaLKB:
## What does this change?
We put a max-width of 600px on the body on AMP. This is adding it to the header and footer as well. 

cc @guardian/dotcom-platform 

We can't determine all the places an AMP page could show up, but as its meant for mobile and we only style for mobile this prevents it from looking broken if it does show up on a desktop/big tablet.

This only affects AMP

## Screenshots
Before:
image

After:
image

## Tested in CODE?
no


stephanfowler:
👍 lgtm

PR #15785, opened 2017-02-08
Fix issue with thin black line appearing above youtube atom overlay

GHaberis:
## What does this change?

Fixes issue in Chrome where when you use "background-size: cover", if the position of background-image isn't less than 50% then you sometimes see a 1px 'line' at the top of the overlay, setting position to 49%  instead will fix this problem.

Fixes bug

No

## Screenshots

Before...

picture 34

After...

picture 33

## Tested in CODE?

No

@akash1810

### DIFF
--- a/static/src/stylesheets/module/atoms/_youtube.scss
+++ b/static/src/stylesheets/module/atoms/_youtube.scss
@@ -75,7 +75,7 @@
 
     .youtube-media-atom__overlay {
         background-size: cover;
-        background-position: 50% 50%;
+        background-position: 49% 49%;
         background-repeat: no-repeat;
         z-index: 1;
         text-align: center;
--- a/static/src/stylesheets/module/atoms/_youtube.scss
+++ b/static/src/stylesheets/module/atoms/_youtube.scss
@@ -75,6 +75,11 @@
 
     .youtube-media-atom__overlay {
         background-size: cover;
+        /**
+         * Fixed 1px line appearing above overlay occasionally.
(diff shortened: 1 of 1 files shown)


akash1810:
Oh wow! This feels a bit magic to me. Is this a general bug in Chrome's implementation of `background-size: cover` or ours?

GHaberis:
@akash1810 it is magic, I found a handful of people reporting the same issue online although no official bug report, this solution seemed the best as it matched closely to the original rules (background-size: cover; background-position: 50% 50%) were derived from youtube's implementation.

I've done a little testing and I could remove background-position completely to remove this bug on youtube-atoms. If I do it'll default the position to 0% 0%. This means that an image which doesn't match the aspect ratio of the container would be partially 'cut off' from one side only (eg. bottom side) rather than the cut off area being equally spread (eg. between top and bottom) - this is not necessarily a problem for youtube-atoms however as the image aspect ratio always matches the container size (16:9)?

akash1810:
Interesting. Could we add a comment for future generations? Then 👍 

> this is not necessarily a problem for youtube-atoms however as the image aspect ratio always matches the container size (16:9)?

We don't enforce this rule (but it has been added to the [project](https://github.com/guardian/media-atom-maker/projects/2#card-1329735)) so we probably can't rely on this being true yet.

NataliaLKB (approved):
lgtm! But definitely agree with @akash1810 that a comment is really needed. Maybe a link to where you found the hack?

GHaberis:
I've added a little info as suggested!

PR #15786, opened 2017-02-08
Repurpose server-side share count switch

jfsoul:
## What does this change?
Repurposes the server-side facebook share count switch so that it's now a permanent feature switch for share count in general.  Share count is now always fetched server-side, but I'm leaving a switch there to turn it off in exceptional circumstances.  When off, the share count container will not be populated.

Avoids switch expiry and concludes trial period for the feature.

No

## Tested in CODE?
No


jfsoul:
cc: @guardian/dotcom-platform

NataliaLKB (approved):

PR #15788, opened 2017-02-08
Move Slice Adverts to Scala 

katebee:
## What does this change?

Previously the Slice adverts were all added by commercial JS

This deletes `slice-adverts.js` and moves the logic to a new HTML cleaner. We are using a cleaner because:

1. the slots must have unique and sequential IDs: `dfp-ad--inline1`, `dfp-ad--inline2` ...
2. we need to take into account any `fc-container--commercial` added by the `CommercialComponentHigh` cleaner

- Serverside rendering of our ad slots  🖥  💨 
- Ideally, there will be no noticeable difference in the position or number of MPUs on fronts
- Less JS and less DOM manipulation

No...

## Screenshots

In mobile view, the MPUs are inserted between `sections` of the Front. The cleaner excludes sections for various reasons, then inserts an MPU after every other eligible section:

picture 520

In desktop view, the `section` will contain `slices` which may have a MPU slot. The cleaner finds these `fc-slice__item--mpu-candidate--desktop` and adds the MPU sliceSlot:

picture 521

## Tested in CODE?

Planning to!


common/app/views/fragments/collections/popular.scala.html
```diff
@@ -51,7 +51,8 @@
         }
 
         @if(showMPU(containerDefinition)) {
-            <div class="fc-slice__popular-mpu js-fc-slice-mpu-candidate fc-slice__item--mpu-candidate fc-slice__item--no-mpu"></div>
+            <div class="fc-slice__popular-mpu js-fc-slice-mpu-candidate fc-slice__item--mpu-candidate fc-slice__item--mpu-candidate--desktop">
```
rich-nguyen: i wonder if the class `js-fc-slice-mpu-candidate` can be removed now?

common/app/views/support/HtmlCleaner.scala
```diff
@@ -630,6 +630,46 @@ object setSvgClasses {
...
+    val containers: List[Element] = document.getElementsByClass("fc-container").toList
+
+    // On mobile, we remove the first container if it is a thrasher on a Network Front
+    // and remove a container if it, or the next sibling, is a commercial container
+    // then we take every other container, up to a maximum of 10, for targeting MPU insertion
+    val containersForCommercialMPUs = containers.zipWithIndex.collect {
+      case (x, i) if !isNetworkFrontWithThrasher(x, i) && !hasAdjacentCommercialContainer(x) => x
+    }.zipWithIndex.collect {
+      case (x, i) if i % 2 == 0 => x
+    }.slice(0, 10)
```
rich-nguyen: `.take(10)` javascript alert! ;)

rich-nguyen (approved):

rich-nguyen:
Looks great! we might be able to add a scala test for it, let's try it tomorrow

common/app/views/fragments/collections/popular.scala.html
regiskuckaertz: I don't think the `--desktop` one is useful either, perhaps for clarity?

common/app/views/fragments/items/facia_cards/sliceSlot.scala.html
```diff
@@ -0,0 +1,20 @@
+@(
```
regiskuckaertz: There is a lot of redundancy between this fragment and `gallerySlot`, and in the future we might have a similar `liveblogSlot`. Only the class names change (I might do something about that in the future) if I'm not mistaken, I guess we could abstract that some more.

common/app/views/support/HtmlCleaner.scala
```diff
@@ -630,6 +630,46 @@ object setSvgClasses {
...
+    // and remove a container if it, or the next sibling, is a commercial container
+    // then we take every other container, up to a maximum of 10, for targeting MPU insertion
+    val containersForCommercialMPUs = containers.zipWithIndex.collect {
+      case (x, i) if !isNetworkFrontWithThrasher(x, i) && !hasAdjacentCommercialContainer(x) => x
+    }.zipWithIndex.collect {
+      case (x, i) if i % 2 == 0 => x
+    }.take(10)
+
+    for (container <- containersForCommercialMPUs) {
+      container.after(s"""<section class="fc-slice__item--mpu-candidate--mobile">${sliceSlot(containersForCommercialMPUs.indexOf(container), isMobile = true)}</section>""")
```
regiskuckaertz: You don't need the class here either.

regiskuckaertz:
This is awesome, great work @katebee ! There is some cleanup to do in some AMD modules (`render-advert` and `popular`) as well as in SASS.

regiskuckaertz:
(I keep on hitting the "Close and comment" button, sorry about that 😓 )

rich-nguyen:
👍  for tests!

common/app/views/fragments/collections/popular.scala.html
katebee: It seems that `js-fc-slice-mpu-candidate` is also responsible for adding the MPU into 'most viewed' when it appears at the bottom of articles... :cry:

I can simplify the other classes: 
```
<div class="fc-slice__popular-mpu js-fc-slice-mpu-candidate fc-slice__item--mpu-candidate">
```

common/app/views/support/HtmlCleaner.scala
katebee: I'm using the class as part of the testing, to ensure the sections are being inserted at the correct places.  Will change it to `fc-container__mpu--mobile` 👍

PR #15790, opened 2017-02-08
Intro text as outro text in emails

joelochlann:
Email fronts have a currently unused feature where the On-Page Description from the config tools gets inserted as text at the top. This is a bit of a hack. Ultimately we'll add functionality to the main fronts tool to allow editorial staff to edit this and choose whether it goes at the top or bottom, since the config tool is supposed to be for infrequent changes and has more restricted access control.

However, in the meantime we want to use the hack to insert some messaging of our own - a link to a feedback survey. For this we need:
- the message to be at the bottom
- the message to not escape HTML (so we can put links in, among other things)

## Screenshot
picture 404

### DIFF
--- a/common/app/views/fragments/email/stylesheets/connectedStyleOverride.scala.html
+++ b/common/app/views/fragments/email/stylesheets/connectedStyleOverride.scala.html
@@ -20,7 +20,7 @@
         padding: 0 !important;
     }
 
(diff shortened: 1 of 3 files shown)


SiAdcock (approved):
❤️  👍

PR #15798, opened 2017-02-09
SubNav tweaks so that the most specific section (or tag) is the one that is highlighted

NataliaLKB:
## What does this change?
For the new nav, we have a subnav:
image

At the moment, a context word should be highlighted for each article and front on both the subNav and the Nav 5 primary words. This is only happening for sections which is a problem, because the context word isn't accurate and actually could change unexpectedly for the user.

Current journey:
broken-subnav-journey

Proposed journey:
better-subnav-journey

Nicer user experience!

No

## Screenshots
image

## Tested in CODE?
Yes

### DIFF
--- a/common/app/common/NewNavigation.scala
+++ b/common/app/common/NewNavigation.scala
@@ -67,16 +67,6 @@ object NewNavigation {
     case "life" => Life
   }
 
-  def isCurrentSection(currentSection: String, navLink: NavLink): Boolean = {
-    val mainFront = List("uk", "au", "us", "international")
-
-    if (mainFront.contains(currentSection)) {
-      "headlines" == navLink.title
-    } else {
-      currentSection == navLink.uniqueSection
-    }
-  }
-
   case object MostPopular extends EditionalisedNavigationSection {
     val name = "news"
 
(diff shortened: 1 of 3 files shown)


NataliaLKB:
cc @guardian/dotcom-platform

common/app/common/NewNavigation.scala
```diff
@@ -454,6 +481,21 @@ object NewNavigation {
       editionalisedSubSectionLinks.exists(_.pageId == sectionId)
     }
 
+    def getSectionOrTagId(page: Page) = {
+      val tags = Navigation.getTagsFromPage(page)
+      val intersectionalKeywords = tagPages.intersect(tags.keywordIds)
```
jfsoul: what about renaming this to `commonKeywords` or something?

jfsoul (approved):
Nice work, looks good to me!

common/app/common/NewNavigation.scala
NataliaLKB: Haha yes that is a much less jargon-y name

paperboyo:
Superb! Are we planing to do anything about the repetitions?

repetition

NataliaLKB:
@paperboyo that is something that has been talked about by @stephanfowler and Alex B. But this change needs to go live before we change fronts.

Definitely something we want to do though!

paperboyo:
We should also find a way to represent third (and further) level(s) of hierarchy. Nesting is more obvious in the old nav, I'm afraid:

nav confusion

vs.

picture 96

PR #15800, opened 2017-02-09
Update JS expiry date on opinion to match ab switch expiry

davidfurey:
## What does this change?

Updates JS expiry date on opinion to match ab switch expiry

Keeps the test running longer so that we can collect more data.


katebee (approved):
👍 
May also be helpful to know, that currently: 
> the expiry in JS requires the day to have started, where as the Scala switches require the day to have passed.

PR #15801, opened 2017-02-09
Small change to AMP analytics file to allow setting of cookies

michaelwmcnamara:
Very small change to add data-credentials="include" to the amp-analytics element as per recommendations here: https://www.ampproject.org/docs/reference/components/amp-analytics#attributes

This is the first attempt to resolve problems with browser id not being set when visiting amp pages:
https://trello.com/c/vbsri2l7/209-amp-analytics-browser-id-is-being-reset-each-time-not-persisting

This change is dependent on corresponding Ophan change to allow setting of credentials:
https://github.com/guardian/ophan/pull/2012

cc @guardian/dotcom-platform


michaelwmcnamara:
Holding off merging till I can try this out on code - wasnt able to test on local due to CORS restrictions.

michaelwmcnamara:
Testing on code shows I need to set "Access-Control-Allow-Credentials" to true also.
 
Credentials flag is 'true', but the 'Access-Control-Allow-Credentials' header is ''. It must be 'true' to allow credentials. Origin 'http://m.code.dev-theguardian.com' is therefore not allowed access.

michaelwmcnamara:
This change is dependant on a change in ophan for it to work:
See: https://github.com/guardian/ophan/pull/2012

michaelwmcnamara:
Have put these changes behind a switch to make coordination with Ophan easier.

common/app/views/fragments/amp/analytics.scala.html
```diff
@@ -11,7 +12,15 @@
 <amp-pixel src="@{Configuration.debug.beaconUrl}/count/pv.gif"></amp-pixel>
 
 <amp-pixel src="//www.facebook.com/tr?id=@FBPixel.account&ev=PageView&noscript=1"></amp-pixel>
-<amp-analytics config="https://ophan.theguardian.com/amp.json"></amp-analytics>
+
+@*
```
NataliaLKB: comments?

NataliaLKB (commented):
Other than the commented out code this looks fine. Not 100% sure that this needs to be behind a switch, but probably doesn't need to be in dev. Try something like this:

`@if(context.environment.mode != Dev) { credentials="include" }`

common/app/views/fragments/amp/analytics.scala.html
michaelwmcnamara: Oops! they shouldn't be there! Thanks!

michaelwmcnamara:
Thanks @NataliaLKB! I will add the exclusion for dev mode you mentioned.
The reason for the switch is because I need a change in Ophan to happen first. Which will change the response headers for this request to use [Access-Control-Allow-Credentials = true] (https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS#Access-Control-Allow-Credentials) 
If this change becomes active before this [Ophan pull request] (https://github.com/guardian/ophan/pull/2012) is live, then the request for "https://ophan.theguardian.com/amp.json" will be blocked for all amp pages.

NataliaLKB:
@michaelmcnamara great! 

Make sure you rebase with master (that is why your PR build is failing)

common/app/conf/switches/FeatureSwitches.scala
```diff
@@ -498,6 +498,17 @@ trait FeatureSwitches {
...
 
+  val AmpOphanTrackingIncludesCredentials = Switch(
+    SwitchGroup.Feature,
+    "amp-ophan-tracking-includes-credentials",
+    "allows access to id field when page viewed from google cache so that we keep tracking consistent",
+    owners = Seq(Owner.withGithub("michaelwmcnamara")),
+    safeState = Off,
+    sellByDate = new LocalDate(2017, 3, 10),
+    exposeClientSide = true
+
```
NataliaLKB: Nit: extra whitespace?

NataliaLKB (approved):
One small nit, but everything else looks good!

PR #15802, opened 2017-02-09
Add check-mediator module to mediate between outbrain and checks in email-article/run-checks

GHaberis:
## What does this change?

**Check Mediator** 

- New check-mediator module mediates between checks and modules which will act once check has taken place.
- Check mediator module registers checks and returns Promise for each check registered
- Modules can import check-mediator and resolve/reject checks via public resolveCheck/rejectCheck methods.
- When check has taken place the promise should be resolved
- Modules can import check-mediator and wait for checks via public waitForCheck method

**Outbrain**

- Adds outbrain container **after** email signup form.
- Outbrain container imports new check mediator module
- The checks 'common/modules/email/run-checks' and  'common/modules/experiments/ab-test-clash' no longer required in oubrain.
- Outbrain waits for these check Promises to be resolved before loading.

- Makes checks in outbrain async not sync.
- We no longer have to run tests in both outbrain and email-article
- Moves 'common/modules/email/run-checks' and  'common/modules/experiments/ab-test-clash' out of outbrain and therefore standard.
- Lays down groundwork for moving outbrain into enhanced and out of standard

No

## Screenshots

N/A

## Tested in CODE?

No


static/src/javascripts-legacy/projects/commercial/modules/third-party-tags/outbrain.js
```diff
@@ -124,52 +116,21 @@ define([
...
-                    resolve();
-                }
-            });
-        }
+        var isCheckNonCompliant = function(result) {
+                return result;
+            };
+
+        return checkMediator.waitForCheck('isOutbrainNonCompliant')
+                    .then(function(resultList) {
```
regiskuckaertz: This callback should return a boolean for better separation of concerns (the `nonCompliant` id makes more sense at the call site)

static/src/javascripts-legacy/projects/commercial/modules/third-party-tags/outbrain.js
```diff
@@ -124,52 +116,21 @@ define([
...
+                        if (resultList.some(isCheckNonCompliant)) {
```
regiskuckaertz: little tip, you can use `resultList.some(Boolean)` and spare yourself the `isCheckNonCompliant` function.

regiskuckaertz:
Great! What about Plista? (Plista replaces Outbrain in Australia, although they don't have the same kind of rules, so perhaps irrelevant).

wrt moving Outbrain to enhanced, please do not do anything without approval from Jamie Morrison and Catherine Murray.

GHaberis:
@regiskuckaertz thanks for the comments, I've made some changes based on them. I'll have to look into Plista, your comment is the first I'd heard of it! I'll also confirm with Jamie/Catherine regarding moving oubrain into enhanced.

GHaberis:
@guardian/dotcom-platform would be great to get some feedback!
@davidfurey I had a few gnarly merge conflicts with some of the a/b test changes you made, I believe they're all resolved and in order but you might want to take a look and see if you're happy!

static/src/javascripts-legacy/projects/commercial/modules/third-party-tags/outbrain.js
```diff
@@ -124,61 +116,13 @@ define([
...
-
     function init() {
         if (commercialFeatures.outbrain && identityPolicy() ) {
             // if there is no merch component, load the outbrain widget right away
             return loadInstantly().then(function(shouldLoadInstantly) {
                 if (shouldLoadInstantly) {
-                    return checkDependencies().then(function (widgetType) {
-                        widgetType ? module.load(widgetType) : module.load();
+                    return checkMediator.waitForCheck('isOutbrainNonCompliant').then(function (isOutbrainNonCompliant) {
+                        isOutbrainNonCompliant ? module.load('nonCompliant') : module.load();
```
gtrufitt: I would name this param `outbrainIsNonCompliant` rather than `is..`

static/src/javascripts-legacy/projects/common/modules/email/run-checks.js
```diff
@@ -149,18 +132,21 @@ define([
...
+                        !config.page.isSensitive &&
+                        !config.page.isFront &&
+                        config.switches.emailInArticle &&
+                        !clash.userIsInAClashingAbTest(clash.nonEmailClashingTests) &&
+                        storage.session.isAvailable() &&
+                        !userHasSeenThisSession() &&
+                        !(browser === 'MSIE' && contains(['7','8','9'], version + ''));
+
+        if (!emailCanRun) {
+            checkMediator.resolveCheck('isEmailInserted', false);
```
SiAdcock: Without knowing too much about what these changes are doing, I'm a little bit 😢  about adding side effects like this to what is essentially a query function. Does this not belong in the caller, rather than in here?

GHaberis: Yes this is a good point, I'll update to remove checkMediator resolveCheck from the run-checks module and move to the caller (email-article).

static/src/javascripts-legacy/projects/common/modules/experiments/ab-test-clash.js
```diff
@@ -50,7 +52,12 @@ define([
 
     function userIsInAClashingAbTest(tests) {
         tests = tests || clashingTests;
-        return _testABClash(ab.isInVariant, tests);
+
+        var isUserInAClashingAbTest = _testABClash(ab.isInVariant, tests);
+        
+        checkMediator.resolveCheck('isUserInAClashingAbTest', isUserInAClashingAbTest);
```
SiAdcock: This function also has a side effect... Is there somewhere else this could go?

static/src/javascripts-legacy/projects/common/modules/experiments/ab-test-clash.js
```diff
@@ -50,7 +50,10 @@ define([
 
     function userIsInAClashingAbTest(tests) {
         tests = tests || clashingTests;
-        return _testABClash(ab.isInVariant, tests);
+
+        var isUserInAClashingAbTest = _testABClash(ab.isInVariant, tests);
```
SiAdcock: Could return the result of this function call without assigning it to a var

static/src/javascripts-legacy/projects/common/utils/check-dispatcher.js
```diff
@@ -0,0 +1,39 @@
...
+
+    var checksToDispatch = {
+        isUserInAClashingAbTest: function() {
+            return clash.userIsInAClashingAbTest();
+        },
+        emailCanRun: function() {
+            return emailRunChecks.allEmailCanRun();
+        },
+        listCanRun: function() {
+            return find(emailArticle.getListConfigs(), emailRunChecks.listCanRun) ? true : false;
```
SiAdcock: !!find(...) will convert a truthy / falsey value into a Boolean

static/src/javascripts-legacy/projects/common/utils/check-mediator.js
```diff
@@ -0,0 +1,146 @@
...
+            return registeredChecks[id].complete;
+        }
+
+        return Promise.reject('no deferred check with id ' + id);
+    }
+
+    /**
+     * exposed for unit testing
+    **/
+    function _testRegisterCheck(check) {
```
SiAdcock: This function is effectively identical to registerCheck. Could we dispense with it by making the checkList a parameter of init?

SiAdcock (commented):
The control flow is staring to look so much better, more testable. Registering and resolving checks feels a little spicy for me, there's a lot of binding and indirection that I think could be avoided. Perhaps this a matter of style, though!

static/src/javascripts-legacy/projects/common/modules/experiments/ab-test-clash.js
GHaberis: Yeah you're right, this variable is a hangover from when userIsInAClashingAbTest resolved the check and was used in a conditional.

static/src/javascripts-legacy/projects/common/utils/check-mediator.js
GHaberis: Yep it is, I didn't want to expose the private function registerCheck as it wasn't really part of the public interface for this module, so I created this almost identical function _testRegisterCheck as then it should be obvious to anyone that this has been exposed for testing purposes only. For the same reason I didn't want to expose checkList as this implies it can be accessed/modified by other modules.

SiAdcock: I agree that exposing `registerCheck` is not ideal. Why not accept `checkList` as a parameter of `init`?

```
function init(checkList) {
  checkList.forEach(registerCheck);
}
```
The `checkList` data structure is not used anywhere else in the module, it feels like a dependency to be injected, rather than a part of the module itself. This would also allow us to pass test data into the module, without having to expose a method just for testing.

GHaberis: Yes I did consider this, but at the time it felt messy defining the checkList var else where (in main.js) and passing it into checkMediator's init, because checkList has the potential to grow. In terms of implementation though I agree it'd be neater to pass this to checkMediator as then we can do away with the test functions.

static/src/javascripts-legacy/projects/common/utils/check-dispatcher.js
desbo: i haven't read all of this PR yet but assuming `getListConfigs` is an array, you can use `Array.some`

static/src/javascripts-legacy/projects/common/utils/check-mediator.js
SiAdcock: True that, it's not clear cut. Separating logic from data, removing state and making modules pure and testable creates additional indirection. It feels like this is part of a bigger discussion around establishing a more functional approach to structuring our application, so I won't clutter up this PR with any more high level stuff 😄

static/src/javascripts-legacy/projects/common/utils/check-mediator.js
```diff
@@ -0,0 +1,146 @@
...
+    
+    function resolveCheck(id) {
+        var argsArray = Array.prototype.slice.call(arguments, 1);
+        
+        if (registeredChecks[id]) {
+            return registeredChecks[id].resolve.apply(null, argsArray);
+        }
+    }
+
+    function rejectCheck(id) {
```
gtrufitt: There's no example of rejecting a check ID, in what situation would that happen? Would you not always resolve with a `true` or `false`?

static/src/javascripts-legacy/projects/common/utils/check-mediator.js
```diff
@@ -0,0 +1,146 @@
...
+        }.bind(this));
+
+        if (dependentCheckPromises) {
+            Promise.all(dependentCheckPromises).then(function(results) {
+
+                var hasPassed = function(result) {
+                    return result;
+                };
+
+                if (dependentChecksPassCondition.call(results, hasPassed)) {
```
gtrufitt: Could be `this.resolve(dependentChecksPassCondition.call(results, hasPassed));`

static/src/javascripts-legacy/projects/common/utils/check-mediator.js
```diff
@@ -0,0 +1,146 @@
...
+                }
+            }.bind(this));
+        }
+    }
+
+    function registerDefferedCheck(check) {
+        var dependentCheckPromises = [];
+
+        if (check.dependentChecks) {
+            check.dependentChecks.list.forEach(registerDependentCheck.bind(null, dependentCheckPromises));
```
gtrufitt: Can you use `'lodash/collections/map'` here instead of side-effecting the `dependentCheckPromises` array? If I am reading right it could then become:

```js
function registerDefferedCheck(check) {
        
        if (check.dependentChecks) {
          return new DeferredCheck(map(check.dependentChecks.list, registerDependentCheck), check.dependentChecks.passCondition);
        }

        return new DeferredCheck();
    }


    function registerDependentCheck(dependentCheck) {
        return registerCheck(dependentCheck).complete;
    }
```

static/src/javascripts-legacy/projects/common/utils/check-mediator.js
```diff
@@ -0,0 +1,146 @@
...
+) {
+
+    /**
+     * registeredChecks will store references to instances of DefferedCheck
+    **/
+    var registeredChecks = {};
+
+    /**
+     * referenced as passCondition for dependentChecks
+     * SOMECHECKSPASSED: Every dependentCheck has returned true
```
gtrufitt: Are these comments the right way around?

gtrufitt (approved):
LGTM. Much easier to reason about what depends on what. 👍

gtrufitt:
> Great! What about Plista? (Plista replaces Outbrain in Australia, although they don't have the same kind of rules, so perhaps irrelevant).

I don't think Plista has them same contract rules so there's no need for the dependency management we have with Outbrain.

> wrt moving Outbrain to enhanced, please do not do anything without approval from Jamie Morrison and Catherine Murray.

@JamieKMorrison is aware of this work, I think he has checked with Catherine. Jamie is going to track the effect on impressions that this work has so we'll need to confirm when it goes out so we can accurately measure that.

static/src/javascripts-legacy/projects/common/utils/check-mediator.js
```diff
@@ -0,0 +1,146 @@
...
+
+    function rejectCheck(id) {
+        var argsArray = Array.prototype.slice.call(arguments, 1);
+
+        if (registeredChecks[id]) {
+            return registeredChecks[id].reject.apply(null, argsArray);
+        }
+    }
+
+    function waitForCheck(id) {
```
desbo: might be better to expose `waitForOutbrain` instead of having users pass in the `id`

static/src/javascripts-legacy/projects/common/modules/email/email-article.js
```diff
@@ -285,11 +287,13 @@ define([
...
-                emailRunChecks.getUserEmailSubscriptions().then(function () {
-
+            checkMediator.waitForCheck('emailCanRun')
+                .then(function (canEmailRun) {
+                    if (canEmailRun) {
+                        return emailRunChecks.getUserEmailSubscriptions();
+                    }
+                })
+                .then(function () {
                     if (ab.isParticipating({id: 'TailorRecommendedEmail'})) {
```
regiskuckaertz: This will get called even if `canEmailRun` is `false`, is it correct?

GHaberis: No it shouldn't do that, updating this now.

static/src/javascripts-legacy/projects/common/utils/check-mediator.js
GHaberis: Yep I created and exposed this when setting the module up but you're correct it's not used as promises should be resolved with true/false, so yes this can be removed.

PR #15804, opened 2017-02-10
Fix link to foresee asset

SiAdcock:
## What does this change?

Currently, the path to `foresee-trigger.js` does not include `assets/javascript`, therefore the network returns a 404. This issue was introduced as a result of moving from using `require` to load external scripts, to using `loadScript` (#15582)

Correct path to foresee will ensure that lucky users get to see the foresee survey modal 🎉 

No

## Tested in CODE?

No

### DIFF
--- a/common/app/templates/javaScriptConfig.scala.js
+++ b/common/app/templates/javaScriptConfig.scala.js
@@ -42,7 +42,7 @@
             "timingEvents": []
         },
         "libs": {
-            "foresee": "vendor/foresee/20150703/foresee-trigger.js",
+            "foresee": "/assets/javascripts/vendor/foresee/20150703/foresee-trigger.js",
             "googletag": "@{Configuration.javascript.config("googletagJsUrl")}",
             "sonobi": "@{ if(!environment.isCode) Configuration.javascript.config("sonobiHeaderBiddingJsUrl") else "//mtrx.go.sonobi.com/morpheus.theguardian.12911_us_.js"}"
         }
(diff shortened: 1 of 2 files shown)


common/app/templates/javaScriptConfig.scala.js
```diff
@@ -42,7 +42,7 @@
             "timingEvents": []
         },
         "libs": {
-            "foresee": "vendor/foresee/20150703/foresee-trigger.js",
+            "foresee": "/assets/javascripts/vendor/foresee/20150703/foresee-trigger.js",
```
sndrs: should this be `@Static("javascripts/vendor/foresee/20150703/foresee-trigger.js")`?

sndrs:
nice one, good catch. 👍 aside from comment

stephanfowler:
What makes this resolve against https://assets.guim.co.uk - as opposed to against https://www.theguardian.com which is (part of?) the current problem?

SiAdcock:
@sndrs the foresee files are not added to the asset hash. `foresee-trigger.js` loads `foresee-surveydef.js` from [a path hardcoded in the JavaScript](https://github.com/guardian/frontend/blob/master/static/vendor/javascripts/foresee/20150703/foresee-trigger.js#L10). So while we could potentially hash `foresee-trigger.js`, the rest would have to be served unhashed. I think that would make the behaviour more confusing.

sndrs:
fair enough!

sndrs (approved):

SiAdcock:
@stephanfowler I think you're right, since `foresee-trigger` is hosted on assets.guim.co.uk, we'll need to use `Static()` to resolve it. I've made the change, accordingly.

@sndrs can you give this another quick once-over?

sndrs (approved):

PR #15791, opened 2017-02-08
Use webpack / fix sass in dev

sndrs:
## What does this change?

- fixes the `sass` compiling/loading endless loop (@superfrank, @zeftilldeath)
- centralises `css` creation (`sass`/`postcss` etc)
- uses `webpack` in `dev`
  - make `webpack` config configurable by env
- removes `gulp`
- clears out some other unused packages from `package.json`

## What is the value of this and can you measure success?

- with `webpack` in `dev` we can confidently remove `curl` in `prod` (no unexpected surprises in `prod`)
- `sass` works properly in `dev` again
- `dev` `css` and `prod` `css` are the same (no unexpected surprises in `prod`)
- provides sourcemaps for `css` and `js` in `dev` and `prod`

## Tested on CODE?
- [x] test on CODE

PR #15807, opened 2017-02-10
100% webpack (in prod)

sndrs:
## What does this change?

- removes the webpack MVT test 
- puts webpack on a switch, which is on by default

- put the whole site to webpack, but enables us to revert to requirejs if it turns out there's a fault we've overlooked
- enables a fix that means we can re-merge #15791

no

## Screenshots

no

## Tested in CODE?

- [x] test in CODE


gtrufitt (approved):
LGTM! 🆒

PR #15809, opened 2017-02-10
Increase frequency of email front refresh

joelochlann:
We have email fronts which consist entirely of backfill (e.g. https://www.theguardian.com/email/us/daily) and need to refresh in a more timely manner to make sure the email is in sync with recent changes to the front we're backfilling from.


davidfurey:
👍

davidfurey (approved):

PR #15815, opened 2017-02-13
Enable paid cards to appear in unbranded containers

kelvin-chappell:
This is just for cards in one specific test container initially.

Eg.
image


common/app/views/support/Commercial.scala
```diff
@@ -197,36 +205,66 @@ object Commercial {
...
+      edition: Edition,
+      frontId: String,
+      containerIndex: Int,
+      containerTitle: String,
+      sponsorName: String,
+      cardIndex: Int,
+      cardHeadline: String
+    ) {
+      val pipeSeparated =
+        s"$context | ${ edition.id } | $frontId | ${ containerIndex + 1 } | $containerTitle | $sponsorName | " +
```
regiskuckaertz: Just curious here: will Play take care of escaping characters like `&`, `"`?

kelvin-chappell: I'll try it and see what happens.

kelvin-chappell: Yes they go through to GA as '&' and '"' respectively and HTML is valid so I take that as a success.

PR #15816, opened 2017-02-13
Add a listener to identify the type of ad being displayed

regiskuckaertz:
This is because in particular, the inline merchandising components have slightly different layouts depending on their content. An inline event will look like this:

picture 4

... whereas an inline book will be skinnier:

picture 5

(I used the same example pre- and post-change)


regiskuckaertz:
[Native inline book](https://trello.com/c/CVE8jaJ3/150-native-inline-book)

lps88 (approved):
👍

PR #11821, opened 2016-02-03
update videoJS to 5.4

jamesgorrie:
Updating video JS so we can start looking at updating the contribs (mostyle for ads as there are some performance stuff with Google IMA).

I would have like to have update to 5.5, but there is [a problem with the autoplay logic here](https://github.com/videojs/video.js/pull/2971#issuecomment-179201603) - I have raised [an issue](https://github.com/videojs/video.js/issues/3071) and will try to get that out as soon as.

Most contribs require 5.3 - so this is good for now.

PR #15818, opened 2017-02-13
Use npm in webpack, not checked in bower packages

sndrs:
## What does this change?

- directs webpack to npm, rather than bower

- will mean require/bower can go as one, when the time is right. there's no need to use bower

no


SiAdcock (approved):
👍  Amazing

sndrs:
hit a blocker: 

currently we serve `videojs@5.4.6` and `videojs-ima@0.3.0` (both checked in), but `videojs-ima@0.3.0` specifies a dependency of `videojs@~5.3`. webpack therefore is bundling both, adding 200k to the bundle... trying to work out how to handle it

sndrs:
raised an issue https://github.com/googleads/videojs-ima/issues/333

sndrs:
tempted to downgrade our version till ima updates - seems like it might be a while, and we're running (potentially) incompatible versions as it is. 

we'll mainly miss out on https://github.com/videojs/video.js/pull/2638.

any feelings either way @SiAdcock @akash1810?

akash1810:
> tempted to downgrade our version till ima updates - seems like it might be a while

@sndrs 👍 but could we do this in a test to be sure there isn't a detrimental loss of video consumption as a result?

sndrs:
how long would we have to run it for? according to https://github.com/guardian/frontend/pull/11821, it was only to bring it inline with contribs anyway?

akash1810:
Ah, I wasn't aware of that! In that case, I'm all for downgrading.

PR #15823, opened 2017-02-14
Use returned DFP size when setting the height of the sticky banner

regiskuckaertz:
For those browsers that don't support native sticky positioning yet, the `sticky-top-banner` module kicks off. It is waiting for the ad to be rendered before setting the height of the containing `DIV`.

Recently though, it appears the value read from the CSSOM when the `onSlotRender` callback is called is not always correct, i.e. GPT has not resized the ad slot yet. So we get around that, by using the size returned by DFP instead. This also many not always be correct, but it is closer to the truth, and we have ways to deal with dynamic resizing.

For native ads, where the size is `[0, 0]`, we still read off the CSSOM though.

### DIFF
--- a/static/src/javascripts-legacy/projects/commercial/modules/dfp/on-slot-render.js
+++ b/static/src/javascripts-legacy/projects/commercial/modules/dfp/on-slot-render.js
@@ -34,6 +34,7 @@ define([
             reportEmptyResponse(advert.id, event);
             emitRenderEvents(false);
         } else {
+            advert.size = event.size;
             dfpEnv.creativeIDs.push(event.creativeId);
             renderAdvert(advert, event)
             .then(emitRenderEvents);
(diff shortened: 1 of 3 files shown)


gtrufitt (approved):
👍

static/src/javascripts-legacy/projects/commercial/modules/dfp/on-slot-render.js
```diff
@@ -34,6 +34,7 @@ define([
             reportEmptyResponse(advert.id, event);
             emitRenderEvents(false);
         } else {
+            advert.size = event.size;
```
gtrufitt: Will the event definitely always have a size?

regiskuckaertz: Yes, it is documented here: https://developers.google.com/doubleclick-gpt/reference#googletageventsslotrenderendedevent

regiskuckaertz:
[Top Banner overlapping Nav Bar](https://trello.com/c/La9k6R2P/175-top-banner-overlapping-nav-bar)

PR #15826, opened 2017-02-14
Use CSSOM for Fabrics

regiskuckaertz:
We have remnant creatives still using the old sizes (88x87) and so reading that size will certainly not provide any useful value. For these creatives, we use the traditional route of reading the dimensions of the slot.

Before:
picture 2

After:
picture 3


scottp719 (commented):
LGTM 👍

PR #15827, opened 2017-02-14
Add webpack to `make compile-dev`

sndrs:
## What does this change?

- adds webpack step to `make compile-dev`
- creates a `make compile-watch` step to prepare for watching
  - doesn't include webpack , because webpack runs the initial build in watch mode anyway 

ensures js bundles are present after running `make compile-dev`, which after removing `requirejs` from dev, they were not (only in `make watch`)

no


NataliaLKB (approved):
lgtm 👍

PR #15829, opened 2017-02-14
Use shared ad targeting

kelvin-chappell:
This change exposes a new json object on the client-side that looks like this:
```
{
  "br": "f",
  "co": "colin-marshall",
  "ct": "article",
  "edition": "uk",
  "k": "cities,seoul,south-korea,architecture,urbanisation,construction,south-korea-ferry-disaster,asia,asia-pacific,southkorea,travel,world",
  "p": "ng",
  "se": "history-cities-50-buildings",
  "tn": "resource",
  "url": "/cities/2015/may/27/seoul-sampoong-department-store-disaster-history-cities-50-buildings"
}
```

It's built up in the commercial shared library so that the page-level contextual targeting is the same for web and apps.
The only other value that's known on the server-side is the surging level, which will come in a later PR.

If nothing is broken I'll replace a lot of the current ad call parameters with these new values in another PR.


kelvin-chappell:
[Standardise ad call parameters on web and apps](https://trello.com/c/Ha3BvjDk/166-standardise-ad-call-parameters-on-web-and-apps)

PR #15831, opened 2017-02-14
Fix wrong advert.size test

regiskuckaertz:
`advert.size` now is the [`size` of a render event](https://developers.google.com/doubleclick-gpt/reference#googletag.events.SlotRenderEndedEvent_size), which can either be a string or an array of numbers.


lps88 (approved):
👍

PR #15835, opened 2017-02-14
New Header: Adding tertiary nav for some tag pages

NataliaLKB:
## What does this change?

To make the user journey feel more consistent, if we only link to a page through tertiary navigation, then when you click on that thing you should also get the tertiary nav.

cc @guardian/dotcom-platform 

Better user journey

Nope

## Screenshots
**Before:**
image

**After:**
image

## Tested in CODE?
Nope

### DIFF
--- a/common/app/common/NewNavigation.scala
+++ b/common/app/common/NewNavigation.scala
@@ -371,17 +371,17 @@ object NewNavigation {
       List(africa, middleEast, cities, globalDevelopment)
     )
 
-    val moneySubNav = NavLinkLists(List(property, pensions, savings, borrowing, careers))
+    val moneySubNav = NavLinkLists(List(money, property, pensions, savings, borrowing, careers))
 
     val footballSubNav = NavLinkLists(
       List(
         football,
-        NavLink("live scores", "/football/live"),
-        NavLink("tables", "/football/tables"),
-        NavLink("competitions", "/football/competitions"),
-        NavLink("results", "/football/results"),
-        NavLink("fixtures", "/football/fixtures"),
-        NavLink("clubs", "/football/teams")
+        NavLink("live scores", "/football/live", "football/live"),
+        NavLink("tables", "/football/tables", "football/tables"),
+        NavLink("competitions", "/football/competitions", "football/competitions"),
+        NavLink("results", "/football/results", "football/results"),
+        NavLink("fixtures", "/football/fixtures", "football/fixtures"),
+        NavLink("clubs", "/football/teams", "football/teams")
       )
     )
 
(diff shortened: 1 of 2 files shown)


stephanfowler:
Should the environment tertiary nav be in here too?

NataliaLKB:
@stephanfowler that has been dealt with now

jfsoul:
> then when you click on that thing you should also get the tertiary nav

Can you elaborate on this, I am not too clued up on tertiary nav behaviour so I need it spelled out 😄

NataliaLKB:
@jfsoul yes sorry for the quick description. 

say you are on the football front. You will see this nav:
image

This nav is actually unique to football and contains "tertiary" navigation. Before if you clicked on live scores, you would then see the sports nav like this:

image

But we don't want that to be the user journey. Instead we want the live scores page to also show the tertiary nav like this:

image


I hope that makes more sense!

common/app/common/NewNavigation.scala
```diff
@@ -462,19 +462,29 @@ object NewNavigation {
...
+      "football/competitions",
+      "football/results",
+      "football/fixtures"
+    )
+
     def getSectionOrTagId(page: Page) = {
       val tags = Navigation.getTagsFromPage(page)
       val commonKeywords = tagPages.intersect(tags.keywordIds)
-      val isTagPage = page.metadata.isFront && tagPages.contains(page.metadata.id)
+      val isTagPage = (page.metadata.isFront || frontLikePages.contains(page.metadata.id)) && tagPages.contains(page.metadata.id)
```
jfsoul: Is this already covered by all the `frontLikePages` being added to the `tagPages`?

common/app/common/NewNavigation.scala
```diff
@@ -486,15 +496,59 @@ object NewNavigation {
...
+
     def getSubSectionNavLinks(sectionId: String, edition: Edition, isFront: Boolean) = {
-      if (isFront) {
+      if (isFront || frontLikePages.contains(sectionId)) {
+
+        val id = simplifySectionId(sectionId)
 
-        if (isEditionalistedSubSection(sectionId)) {
-          val subNav = editionalisedSubSectionLinks.filter(_.pageId == sectionId).head.parentSection
+        if (isEditionalistedSubSection(id)) {
```
jfsoul: Not changed in this PR but spelling of editionalised ^

common/app/common/NewNavigation.scala
NataliaLKB: Nope. Two different checks for two different things.

This is just determining whether we need to use the section id, the page id, or something else to determine how we identify the page.

The second use is determining whether we give it a subnav list for fronts, or for articles. They differ because we don't show tertiary on articles

jfsoul (approved):
Thanks for explanations 👍

NataliaLKB:
@jfsoul haha thanks for the spell check!! Really good questions as well. I will change the spelling error in my next PR. Want to get this out now :)

PR #15836, opened 2017-02-15
Use a pre-flight ad call to initiate ad selection process

rich-nguyen:
This enables us to experiment with (allegedly) super quick header bidding solutions, so quick that they happen before the header has even been sent to the browser!

In this setup, the browser's page request triggers an ad call through the CDN. The CDN embeds a token into the html response. I'm using the ophan view id as the token, for easy query joining and reduced identifier-confusion. Once the JS application begins, the token is picked up from the page, and passed back to the ad server for ad retrieval. DFP executes after this, in the same way as a traditional header bidding solution does (that is, enriched with an external bid encoded as a key-value, which will influence the final DFP ad selection, and raise the valuation of our inventory).

This experiment is not 100% traffic, it is only active on pages chosen by the CDN which will have a pre-flight token recorded in `window.esi`.

**Tested in Code**
bazillions of times


common/app/http/Filters.scala
```diff
@@ -8,12 +8,21 @@ import common.ExecutionContexts
...
-import play.api.mvc.{EssentialFilter, Filter, RequestHeader, Result}
-import play.filters.gzip.GzipFilter
+import play.api.mvc._
+import play.filters.gzip.{GzipFilterConfig, GzipFilter}
 
 import scala.concurrent.Future
 
-class Gzipper(implicit val mat: Materializer) extends GzipFilter(shouldGzip = (_, result) => !result.header.isImage)
+class GzipperConfig() extends GzipFilterConfig {
+  val excludeFromGzip = List(
```
janua: Could we add a note here to say that this is to turn off gzipping in the response?

diagnostics/app/controllers/CommercialPreflightController.scala
```diff
@@ -0,0 +1,89 @@
...
+      host <- maybeHost
+      topUrl <- maybeTopUrl
+      userAgent <- maybeUserAgent
+    } {
+      val adHubRequest = Json.toJson(AdRequest(
+        load_id = pageViewId,
+        zone_ids = List(229, 228),
+        /*  id    sizes
+            229   970x250, 728x90
+            228   300x250
```
janua: Where do these IDs come from?

janua: (And is there a list somewhere? Would here be a place to add the mappings for reference?)

janua: Ah these are the DFP IDs!

common/app/http/Filters.scala
rich-nguyen: Yep, ok

diagnostics/app/controllers/CommercialPreflightController.scala
rich-nguyen: The sizes are ad sizes, and one or more sizes are associated with a switch zone id. These zones are defined in their switch management ui (and the integration docs).

PR #15837, opened 2017-02-15
Update error message for old flash-only videos

gidsg:
## What does this change?
Currently when a video which only has a flash encoding (typically from <~2007) is viewed, the videojs player displays the default error *No compatible source was found for this video.* This resulted in many requests to userhelp for old videos, which we are not able to provide.
This PR updates this message to make it clearer these videos are no longer available.

Less confusing message, hopefully results in fewer requests to User Help

No

#### Before

screen shot 2017-02-15 at 11 14 08

#### After
screen shot 2017-02-15 at 11 19 04

## Tested in CODE?
No


akash1810 (approved):

paperboyo:
giphy

PR #15841, opened 2017-02-15
Cleanup Wednesday #1

regiskuckaertz:
First of a series of small cleanups of the commercial code:

- remove the `data-test-id` attribute
- remove the `data-slot-target` attribute
- move `commercialFeatures.topBannerAd` into Scala-land
- remove unused `alladsrendered` event, which is also false by the way

### DIFF
--- a/common/app/views/fragments/commercial/adSlot.scala.html
+++ b/common/app/views/fragments/commercial/adSlot.scala.html
@@ -13,7 +13,6 @@
     id="dfp-ad--@optId.getOrElse(name)"
     class="js-ad-slot ad-slot ad-slot--@name @adTypes.map{ adType =>ad-slot--@adType}.mkString(" ") @optClassNames"
     data-link-name="ad slot @name"
-    data-test-id="ad-slot-@name"
     data-name="@name"
     @if(!showLabel){ data-label="false" }
     @if(!refresh){ data-refresh="false" }
--- a/common/app/views/main.scala.html
+++ b/common/app/views/main.scala.html
@@ -7,7 +7,7 @@
 
 @headerAndTopAds(showAdverts: Boolean, edition: Edition, content: Option[model.ContentType]) = {
   @if(!page.metadata.shouldHideHeaderAndTopAds) {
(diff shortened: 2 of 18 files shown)


common/app/views/main.scala.html
```diff
@@ -5,16 +5,18 @@
...
-        <div id="bannerandheader">
-            @if(showAdverts) {
-                @fragments.commercial.topBanner(page.metadata)
-            }
-            @fragments.header(page)
-        </div>
-        <div id="maincontent" tabindex="0"></div>
+@headerAndTopAds(showAdverts: Boolean, edition: Edition, content: Option[model.ContentType]) = {
+  @if(!page.metadata.shouldHideHeaderAndTopAds) {
+    @defining(showAdverts && (content.isEmpty || content.exists(content => !content.tags.isTheMinuteArticle))) { showTopSlot =>
```
regiskuckaertz: hmm, can `!content.exists(_.tags.isTheMinuteArticle)` be equivalent to the whole expression after the `&&`?

kelvin-chappell: Yes it is.

common/app/views/fragments/commercial/adSlot.scala.html
```diff
@@ -13,7 +13,6 @@
     id="dfp-ad--@optId.getOrElse(name)"
     class="js-ad-slot ad-slot ad-slot--@name @adTypes.map{ adType =>ad-slot--@adType}.mkString(" ") @optClassNames"
     data-link-name="ad slot @name"
-    data-test-id="ad-slot-@name"
```
kelvin-chappell: What was the purpose of `data-test-id`?

kelvin-chappell (approved):
Looks good to me!

common/app/views/fragments/commercial/adSlot.scala.html
regiskuckaertz: It was used by Omniture, but now is completely useless.

common/app/views/main.scala.html
regiskuckaertz: Ah, then I'll change that!

PR #15843, opened 2017-02-15
add dateModified schema.org metadata

gidsg:
## What does this change?
Adds [dateModified](https://schema.org/dateModified) schema.org metadata to articles.

Pleases 🤖  [Structured Data Testing Tool](https://search.google.com/structured-data/testing-tool):
screen shot 2017-02-15 at 15 16 20

No

## Tested in CODE?
No

### DIFF
--- a/common/app/views/fragments/meta/dateline.scala.html
+++ b/common/app/views/fragments/meta/dateline.scala.html
@@ -38,3 +38,4 @@
         <meta itemprop="coverageEndTime" content="@Format(lastModified, "yyyy-MM-dd'T'HH:mm:ssZ")">
     }
 </p>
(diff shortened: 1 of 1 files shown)


NataliaLKB (approved):
Changes look good to me! One thing to mention though is that this does affect AMP (in your description you say it doesn't ). I will assume that this validates okay because the tests pass. Probably best to check.

gidsg:
@NataliaLKB I checked this morning and this change does not affect AMP because the AMP template does not use the `dateline.scala.html` fragment.

NataliaLKB:
@gidsg apologies! I just took a look and only interactives use it I think.

PR #15845, opened 2017-02-15
Remove webpack switch 

sndrs:
## What does this change?

- removes the curl/webpack switch. we serve webapck now
- removes as much of requirejs as we can for the moment

## What _doesn't_ this change?

- doesn't address the `#wp-rjs` flags
- video embeds still using require 
- karma still depends `require` plugin

- closer to webpack-only...


static/src/javascripts-legacy/bootstraps/standard/main.js
```diff
@@ -85,53 +85,6 @@ define([
             }
         });
 
-        /*
```
sndrs: @SiAdcock does this whole delete seem right to you? i think it is in terms what the switch affected, but seems like to go?

SiAdcock: Yeah, this logic now happens in [`inlineJSNonBlocking.scala.html`](https://github.com/guardian/frontend/blob/e6c59a2dcd2e3fc4d0a3304520e8868fdc84d21d/common/app/views/fragments/inlineJSNonBlocking.scala.html#L44-L67)

applications/app/views/fragments/interactiveBody.scala.html
```diff
@@ -67,19 +66,4 @@
             @fragments.contentFooter(interactive, page.related, isPaidContent = isPaidContent)
         </div>
     }
-
-    @if(!Webpack.isSwitchedOn) {
-        <script>
```
gtrufitt: I have a question. Briefly spoke to @SiAdcock about it but it seems that [this was here to improve loading speed of interactives](https://github.com/guardian/frontend/pull/14787), but I'm not sure how much of an advantage it had over being in `jsNonBlocking` as that doesn't wait until app.js has parsed before kicking off anyway?

sndrs: no i agree - inlining it and detaching it from the app should be more than enough?

gtrufitt: To quote Mr @SiAdcock in https://github.com/guardian/frontend/pull/14946

> In addition to being remote, interactive boot scripts are also required dynamically. In order to load these scripts, I have inlined the require calls as non-blocking JavaScript. I have used curl.js to request these scripts. This is already happening in Interactive Articles; I have applied the same logic to News Articles and Liveblog Articles.

So yes, this looks like it should be all good 👌

common/app/views/fragments/inlineJSBlocking.scala.html
```diff
@@ -63,30 +58,11 @@
 
     // ************* LOAD THE MAIN APP ASYNC *************
 
-    @if(conf.Configuration.assets.useHashedBundles) {
```
gtrufitt: Do we no longer need [useHashedBundles](https://github.com/guardian/frontend/blob/531b718e2950d9ac9abd5f2dcb7221ce8576a69f/common/app/common/configuration.scala#L323-L323) then?

applications/app/views/fragments/interactiveBody.scala.html
SiAdcock: Great quote, thanks for sharing 👌 👌

common/app/views/fragments/inlineJSBlocking.scala.html
sndrs: no need in this case, the loader code is the same

gtrufitt: Yeah, I meant at all but then I looked and there's still uses dotted about

applications/app/views/fragments/interactiveBody.scala.html
sndrs: 😂 a moving treatise

static/src/javascripts-legacy/projects/common/views/svgs.js
```diff
@@ -145,10 +143,11 @@ define([
     };
 
     return function (name, classes, title) {
-        if (config.switches.webpack) {
+        // #wp-rjs - needed by karma
+        if (svgs[name].markup) { // webpack way
```
regiskuckaertz: I think this test would make better sense inside the `svg` function, just in case someone uses `svg` directly rather than `svgs`. Also, since we're going 100% webpack, is it even still necessary? Cannot we assume the `markup` will exist all the time?

SiAdcock: @regiskuckaertz it's necessary for tests at the moment

regiskuckaertz: Indeed it is. In fact you don't even need `svg` if you don't want to add classes or a title to the SVG.

SiAdcock: Only 55 calls to the `svgs()` function in the whole app. And 44 icons. It's likely a bunch of these icons aren't used and are contributing to a larger than necessary commercial chunk (54k of commercial is SVGs). Probably worth killing and loading SVGs in the modules that need them. Maybe I'll make a separate PR.

regiskuckaertz: I second that! I started a [weekly cleanup](https://github.com/guardian/frontend/pull/15841) and can take care of the commercial ones next Wednesday 😃  Also we are phasing out all the commercial templates so they will eventually all go away.

sndrs: yeah i think this could do with looking at. but fwiw, the reason those classes were added by default initially was to mimic the behaviour of the play helper, just to make sure you didn't get any nasty surprises

sndrs:
this has had a lot of changes since being opened. going to close it and reopen as a 2nd PR that covers the extra changes

PR #11975, opened 2016-02-15
Remove more old sticky code

OliverJAsh:
Extension of https://github.com/guardian/frontend/pull/11936

/cc @uplne

Revert "Merge pull request #9539 from guardian/viewability-search-form"

This reverts commit 26270f73ed122bb488106cdc34957298b2345442, reversing
changes made to afe00ae50feb3d61f9991896a0a313a582e7890a.

Issue #7991, opened 2015-01-26
Rendering of search box and results is flawed


michaelwmcnamara:
Display issues with google search on network front:

- Currently search box does not display text-box outline

![image](https://cloud.githubusercontent.com/assets/7239344/5904820/4fd003a0-a583-11e4-9a57-2cb126bc9f06.png)

- Font used for text and autocomplete is wrong (arial - sans-serif)

![image](https://cloud.githubusercontent.com/assets/7239344/5904859/94884c6e-a583-11e4-8289-a97f72686815.png)

- Size of results box does not line up with search box

![image](https://cloud.githubusercontent.com/assets/7239344/5904872/bd3499d8-a583-11e4-997b-1f6d81babd19.png)

paperboyo:
Hello,

That is a hairy one that is! More than myself, if that’s even possible ;-)

- ‘Takes up to 8s to show up, even on 8 core desktop’™

- ESC won’t dismiss the bugger

- styling/typography inconsistent with beautiful Weather search, including the ugly pixelated search icon and Google logo

- dismissing behaviour inconsistent with Weather search + the rest of the Interwebz (see https://github.com/guardian/frontend/issues/3437 and https://github.com/guardian/frontend/issues/7845 pt. 3)

- suggestions should be more structured and deduplicated:
 - contibutors (e.g. search for ‘Sarah Philips’: suggestions show multiple pages of her content, shoud show just the first one at the top; bylines in suggestions aren't updating when we update the byline in R2 - http://www.theguardian.com/profile/sarahphillips)
 - series
 - media types
 - tags
 - text search

_[hairy dream mode on]_
 - search results for a specific query should produce a new ‘search front’ much like current ‘tag fronts’ but complete with filtering, sorting and _weighting_, the query should register explicitly in the address bar and thus be shareable

_[hairy dream mode off]_

Regards
Mateusz

ScottPainterGNM:
Looking at this today it appears that some of the initial issue raised by @michaelwmcnamara are improved upon but items such as the lack of dismissal through the escape etc. and search results in the subsequent comments are still apparent. Maybe worth looking at another forum for improvement suggestions rather then the issue list but will close this off soon unless anyone protests?

paperboyo:
I won't protest but if I could be included in that forum, please - that would be amazing!

Issue #7845, opened 2015-01-16
Minor Weather widget glitches


paperboyo:
Hello,

1. The search bar does not respond to some keyboard events it should, IMHO. ESC should dismiss the search/editing mode and go back to last selected city, ~~while Return/Enter should commit (it does) and exit editing mode (it doesn't) as well.~~

2. While going through the list with UP/DOWN arrow keys there is an unneeded step after last/before first item of nothing being selected and just the entered text in the search box (yeah, I know I haven't described it well enough and also it is a minor point!). Maybe it would be a good idea not to wrap around the five suggestions, but provide a longer list when pressing the DOWN arrow key?

3. ~~The modal search box for Weather finally works like the rest of the interwebz! Clicking anywhere outside of it exits editing mode. Well done! Now if only __every other dropdown/search box of ours__ worked that well... ;-)~~ [finally fixed, only the search popup left to fix]

4. This one is probably going to be hard to describe, then hard to understand/agree with and then even harder/impossible to 'fix', so, please, stop reading now. Suppose you are from Żywiec. Suppose you are on a managed machine and cannot change the keyboard layout/IME. You type Zywiec, but it is not recognized. You are used to it. You go to google, type zywiec, it's recognised, the first result is Żywiec, you copy and paste. If only it would be super-easy to provide suggestions for Zywiec that include Żywiec (after all cities starting with Zywiec, obviously). BUT! It's great that our feed is using Unicode. Great! We should not forego that under any circumstances, the least of all to 'fix' that obscure 'issue'.

5. I can see that there is no search on mobile breakpoints. Are we querying the location data? If not, shouldn't the search for a Weather city be at least a part of future user preferences?

6. I just wonder: what is the logic behind the order of suggestions?

+7. Tooltips to smaller icons have disappeared one day. They are useful as some icons look the same even though they depict different weather conditions.

I would say, only pt. 1 ~~(and 3!)~~ is really a bug.

Regards
Mateusz

johnduffell:
sorry!
closing this as it hasn't been fixed so far and probably won't be fixed in the near future

PR #15846, opened 2017-02-15
Make search popup dismissible

SiAdcock:
## What does this change?

Currently, it is not easy to dismiss the search popup once it has been opened. The only way to dismiss it is to click the "Search" link a second time.

With this change, the user will be able to dismiss the search popup by clicking anywhere on the page (other than the search popup itself)

I have also removed the `checkResults()` call, as this method is [no longer with us](https://github.com/guardian/frontend/pull/11975). This issue has generated about 1000 errors in Sentry

This makes the search feature more accessible as it is easier to dismiss.

No

picture 185

## Tested in CODE?

Oh gosh, no

### DIFF
--- a/static/src/javascripts-legacy/projects/common/modules/navigation/search.js
+++ b/static/src/javascripts-legacy/projects/common/modules/navigation/search.js
@@ -34,8 +34,32 @@ define([
             });
 
             bean.on(document, 'click', '.js-search-toggle', function (e) {
-                searchLoader();
+                var searchToggleLink = $('.js-search-toggle');
+                var searchPopup = $('.popup--search');
+                var maybeDismissSearchPopup = function(e) {
+                    var el = $(e.target);
+                    var clickedPop = false;
 
+                    while (el.length && !clickedPop) {
+                        if (el.hasClass('popup--search')) {
+                            clickedPop = true;
+                        }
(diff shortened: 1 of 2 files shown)


static/src/javascripts-legacy/projects/common/modules/navigation/search.js
```diff
@@ -34,10 +34,33 @@ define([
             });
 
             bean.on(document, 'click', '.js-search-toggle', function (e) {
-                searchLoader();
+                var searchToggleLink = $('.js-search-toggle');
+                var searchPopup = $('.popup--search');
```
gtrufitt: Dunno if you can add a `js-` to this? We should totals lint that 🤔

NataliaLKB (approved):
Great 🎉 ! Other than Gareth's comment, this looks good to me.

paperboyo:
Wow!!! 💎! Finally (only 2 years): https://github.com/guardian/frontend/issues/7845 (pt. 1), https://github.com/guardian/frontend/issues/3437#issuecomment-37566527, https://github.com/guardian/frontend/issues/7991#issuecomment-71533847. Hugest thanks, @SiAdcock!

ScottPainterGNM:
Yessssssss!

PR #15847, opened 2017-02-15
Extend email a/b tests

davidfurey:
## What does this change?

Extends two a/b tests

Opinion - we don’t have enough data yet
Flyer - we need to work out the best way of segmenting email lists for future a/b tests


katebee (approved):
👍

PR #15848, opened 2017-02-15
Remove separate ophan request if not enhanced

sndrs:
## What does this change?

- stops trying to download ophan if we're not enhancing

- it's already bundled in standard, so this is pointless
- as well as being unnecessary, it was causing #15845 to fail


gtrufitt (approved):

SiAdcock (approved):
🎉

PR #15859, opened 2017-02-16
First attempt at creating 'tag pages' based off of queries to CAPI Recipe API.

LATaylor-guardian:
## What does this change?
This change produces a 'tag like' index page that displays individual recipes (atoms) in containers. When you click on one of the individual atoms in the container it will take you through to the original article it came from. These pages are backed by queries to Content API's Recipe API. 

This is the first kind of front that is not in essence filled with content as we know it.

In order to make this work without having to reimplement large parts of how fronts and containers are implemented I have reconstructed an IndexPage object from the recipe atoms I receive from CAPI. This allows me to pass this object into the existing templates and effectively get the rendering for free. This seems to work as a way to prove out the value of this, but will have limitations when it comes to more experimentation and design changes in this area in the future (which may or may not be worth it). In that case we have two approaches:

1. Duplicate the templates that exist and use them so that we can make the necessary changes without compromising the common templates.
2. Build a brand new way to support this.

We are likely to go the route of simply duplicating the templates but are open to preferences/suggestions.

The pages can be accessed as follows at current and are namespaced under :
 - /sdata/recipes/cuisine/:myCuisine
 - /sdata/recipes/category/:categoryValue
 - /sdata/recipes/dietary/:dietaryValue
 - /sdata/recipes/celebration/:celebrationValue

e.g /sdata/recipes/cuisine/thai or /sdata/recipes/category/salads

The general idea is that we will link to these new index pages from the current food/recipe fronts and also increase discoverability in Google for the typical types of thing users like to search for pertaining to food/recipes.

Not to my knowledge.

## Screenshots
picture 35

@jfsoul Example of page with no results.
picture 36


applications/app/controllers/StructuredDateIndexController.scala
```diff
@@ -0,0 +1,49 @@
...
+      if (response.results.nonEmpty) {
+        val page = IndexPage(
+          page = SimplePage(MetaData.make(id = "", section = None, webTitle = s"${filterValue.capitalize} Recipes")),
+          contents = response.results flatMap recipeAtomToContent,
+          tags = Tags(List.empty),
+          date = DateTime.now,
+          tzOverride = None
+        )
+
+        Ok(views.html.index(page))
```
jfsoul: Can we add caching to this?  E.g. `Cached(page)(RevalidatableResult.Ok(views.html.index(model)))`

LATaylor-guardian: Of course, was wondering about this.

applications/app/controllers/StructuredDateIndexController.scala
```diff
@@ -0,0 +1,49 @@
...
+
+      } else NotFound
```
jfsoul: Did you think about displaying an empty page rather than a 404 - I think this would be consistent with elsewhere in tag pages.

LATaylor-guardian: Am happy to take direction on whatever you do normally here.

applications/app/controllers/StructuredDateIndexController.scala
```diff
@@ -0,0 +1,49 @@
...
+import common.ExecutionContexts
+import common.`package`._
+import contentapi.ContentApiClient
+import model.{ApplicationContext, MetaData, SimplePage, Tags}
+import org.joda.time.DateTime
+import play.api.mvc.Action
+import services.{IndexPage, IndexPagePagination}
+import structureddata.AtomTransformer._
+
+class StructuredDateIndexController(val contentApiClient: ContentApiClient)(implicit val context: ApplicationContext) extends Paging with ExecutionContexts {
```
NataliaLKB: Small thing, should this be `StructuredData...` not Date?

LATaylor-guardian: Yup, will correct.

applications/app/controllers/StructuredDateIndexController.scala
```diff
@@ -0,0 +1,49 @@
...
+      case "dietary" => baseQuery.dietary(_)
+      case "celebration" => baseQuery.celebration(_)
+      case "category" => baseQuery.categories(_)
+    }
+
+    contentApiClient.getResponse(query(filterValue)) map { response =>
+
+      if (response.results.nonEmpty) {
+        val page = IndexPage(
+          page = SimplePage(MetaData.make(id = "", section = None, webTitle = s"${filterValue.capitalize} Recipes")),
```
NataliaLKB: Is the section for all the recipe pages [food & drink](https://www.theguardian.com/lifeandstyle/food-and-drink)? Why `None` if that is the case?

LATaylor-guardian: Will add.

applications/app/structureddata/AtomTransformer.scala
```diff
@@ -0,0 +1,78 @@
...
+      }
+
+    case _ => None
+  }
+
+  def atomImageToApiElement(image: Image): ApiElement =
+    ApiElement(image.mediaId, "main", ApiElementType.Image, None,
+      assets = image.assets.map { asset =>
+        ApiAsset(ApiAssetType.Image, asset.mimeType, Some(asset.file), Some(ApiAssetFields(height = asset.dimensions.map(_.height), width = asset.dimensions.map(_.width))))
+      }
```
jfsoul: I think I would find this easier to understand with using named params for everything and more newlines, but that might just be me 😁

LATaylor-guardian: Will change.

applications/app/structureddata/AtomTransformer.scala
```diff
@@ -0,0 +1,78 @@
...
+        sectionId = None,
+        sectionName = None,
+        webTitle = webTitle,
+        webUrl = s"${conf.Configuration.site.host}/$id",
+        apiUrl = "",
+        twitterHandle = None,
+        bio = None,
+        description = None,
+        emailAddress = None,
+        bylineImageUrl = None,
```
NataliaLKB: Not sure if I this is best to happen on an iteration, or if the new design doesn't include these things, but typically `twitterHandle`, `bio`, `bylineImage`, etc are all possible things to have on recipes.

applications/conf/routes
```diff
@@ -58,6 +58,9 @@ GET        /opt/$choice<in|out|delete>/:feature
 GET        /service-worker.js                                                   controllers.WebAppController.serviceWorker()
 GET        /2015-06-24-manifest.json                                            controllers.WebAppController.manifest()
 
+# Index pages for structured data
+GET        /sdata/recipes/$filterType<cuisine|dietary|celebration|category>/:filterValue       controllers.StructuredDateIndexController.render(filterType, filterValue)
```
NataliaLKB: Can you include a route in `dev-build` too?

applications/app/controllers/StructuredDateIndexController.scala
jfsoul: I would definitely go for the page with no content over a 404 if not difficult to implement

jfsoul:
I think this is a great first iteration, definitely more simple than I was envisioning (in a good way).  Couple more general comments:

- Tests would be a bonus, but I would still +1 this without them if you want to add those later.
- You may already know this but don't forget to add your new route to [the router](https://github.com/guardian/platform/blob/master/router/files/router.conf)

applications/app/controllers/StructuredDateIndexController.scala
LATaylor-guardian: @jfsoul where did you get model from?

LATaylor-guardian: no worries, got it.

jfsoul: yeah sorry, bad copy+paste - meant `page`

LATaylor-guardian:
Updated with your comments. @jfsoul @NataliaLKB What kind of approach do you take to testing generally on Frontend and what kind of approach would you take in this instance?

LATaylor-guardian:
@stephanfowler FYI.

NataliaLKB:
@LATaylor-guardian, generally I think different devs have different approaches. Certainly I think all or most controllers are tested. To see come examples, take a look [here](https://github.com/guardian/frontend/tree/master/applications/test)

jfsoul:
@NataliaLKB beat me to it 😉 .  Agree with you, testing the controller would be the norm - if you wanted to add unit tests for `AtomTransformer` it would probably be more than we would expect.  The more complex the logic in it gets the more I would want to see a test for it - seems quite simple now though.

`IndexControllerTest` a good example of how we use `testContentApiClient` to save fixed test data to the `data` directory.

common/test/services/IndexPageTest.scala
```diff
@@ -48,20 +48,16 @@ import scala.concurrent.Future
         fail("Wrong type (expected: IndexPage, real: Result)")
       case Some(page) =>
         val front = IndexPage.makeFront(page, edition)
-        front.containers.length should be(1)
+        front.containers should not be empty
```
LATaylor-guardian: @jfsoul @NataliaLKB I've refactored this test because it shouldn't fail as a result of changing the CAPI dependency version.

LATaylor-guardian:
See for route: https://github.com/guardian/platform/pull/1191

common/test/services/IndexPageTest.scala
NataliaLKB: @jfsoul and I just chatted about this. These tests aren't very good. At the moment they have to be rewritten with each data file change and that shouldn't be the case. Anyway this change is fine for now. We will look at it when we get a chance 👍

jfsoul (approved):
Thanks for making those changes - looks good to me 👍

NataliaLKB (approved):
Really great work! 👍 

Might be worth deploying to CODE first and testing it with the platform changes (you can also deploy that to CODE).

LATaylor-guardian:
@jfsoul @NataliaLKB Thanks, will test on CODE.

paperboyo:
> This is the first kind of front that is not in essence filled with content as we know it.

_Content as we know it_ is rigid, boring and makes one stumble at every turn of the journey. This, possibly, has such a bright future, I ran out of emojis...

PR #15861, opened 2017-02-16
Add admin entry point

### LINKED ISSUE #15855: webpackify admin

SiAdcock:
## What does this change?

Adds an entry point in `webpack.config.js` for admin and loads this on the admin page.

Removes another remnant of RequireJS
Fixes #15855

### DIFF
--- a/admin/app/views/admin_main.scala.html
+++ b/admin/app/views/admin_main.scala.html
@@ -90,7 +90,7 @@
             (function (document) {
                 var script = document.createElement('script');
                 var ref = document.getElementsByTagName('script')[0];
-                script.src = '@Static("javascripts/admin.js")';
+                script.src = '@Static("javascripts/admin-webpack.js")';
                 ref.parentNode.insertBefore(script, ref);
             })(document);
 
--- a/admin/app/views/admin_main.scala.html
+++ b/admin/app/views/admin_main.scala.html
@@ -84,17 +84,9 @@
         <link href='//fonts.googleapis.com/css?family=Open+Sans:800' rel='stylesheet' type='text/css'>
 
         <script type="text/javascript">
-
             @Html(config(model.AdminPage(title)).body)
-
-            (function (document) {
-                var script = document.createElement('script');
-                var ref = document.getElementsByTagName('script')[0];
-                script.src = '@Static("javascripts/admin-webpack.js")';
-                ref.parentNode.insertBefore(script, ref);
-            })(document);
-
         </script>
+        <script src="@Static("javascripts/admin-webpack.js")"></script>
         @if(loadJquery){
             <script src="//code.jquery.com/jquery.js"></script>
         }
--- a/admin/app/views/admin_main.scala.html
+++ b/admin/app/views/admin_main.scala.html
@@ -85,15 +85,15 @@
 
         <script type="text/javascript">
 
-            @Html(curlConfig().body)
-
             @Html(config(model.AdminPage(title)).body)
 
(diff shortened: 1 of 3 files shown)


webpack.config.js
```diff
@@ -7,10 +7,13 @@ const outputName = 'app-webpack';
...
-    entry: path.join(__dirname, 'static', 'src', 'javascripts', 'boot-webpack.js'),
+    entry: {
+        [outputName]: path.join(__dirname, 'static', 'src', 'javascripts', 'boot-webpack.js'),
+        admin: path.join(__dirname, 'static', 'src', 'javascripts-legacy', 'bootstraps', 'admin.js'),
+    },
     output: {
         path: path.join(__dirname, 'static', 'target', 'javascripts'),
-        filename: `${env === 'dev' ? '' : '[chunkhash]/'}${outputName}.js`,
+        filename: `${env === 'dev' ? '' : '[chunkhash]/'}[name].js`,
         chunkFilename: `${env === 'dev' ? '' : '[chunkhash]/'}${outputName}.chunk-[id].js`,
```
sndrs: can this be 
```js
chunkFilename: `${env === 'dev' ? '' : '[chunkhash]/'}[name].chunk-[id].js\`
``` 
too?

and then `[outputName]` can just be `'app-webpack'`?

tools/__tasks__/compile/hash/index.js
```diff
@@ -19,7 +19,7 @@ module.exports = {
         {
             description: 'Hash assets',
             task: () => {
-                const webpackRegex = /app-webpack/;
+                const webpackRegex = /[app-webpack|admin]/;
```
sndrs: feels like `admin` might be a bit too permissive for a regex? will this mean anything with that in the filename will be matched?

webpack.config.js
SiAdcock: Yeah, I initially tried that, but it doesn't work out quite like that! 

`[name]` in the context of `chunkFilename` refers to the name given to the chunk, an optional [third parameter passed to `require.ensure`](https://webpack.js.org/guides/code-splitting-require/#components/sidebar/sidebar.jsx). If no name is passed, [`[name]` becomes the `[id]`](https://webpack.js.org/configuration/output/#output-chunkfilename). So we end up with chunks called `0.chunk-0.js`, and so on.

I thought: why doesn't `[name]` just take the name of the parent entry chunk? It's because a chunk could have multiple entry chunks.

Let me know if you think of a preferable way of doing this

tools/__tasks__/compile/hash/index.js
SiAdcock: True facts, man. We could go with `admin-webpack` instead?

webpack.config.js
sndrs: no that's cool then. once require is gone we can look at doing them properly, naming the chunks. ill a thing to the project

admin/app/views/admin_main.scala.html
```diff
@@ -85,15 +85,15 @@
...
         <script type="text/javascript">
 
-            @Html(curlConfig().body)
-
             @Html(config(model.AdminPage(title)).body)
 
-            window.curl.paths['bootstraps/admin'] = '@Static("javascripts/bootstraps/admin.js")';
-
-            @Html(common.Assets.js.curl);
+            (function (document) {
```
sndrs: this could just be a script tag if we wanted, only admin. nbd though, no blocker.

sndrs (approved):
great stuff

PR #15862, opened 2017-02-16
Don't show play button on non-playable non-media cards

gidsg:
## What does this change?
This PR stops the play button appearing on small non-playable non-video page cards, as this could give a false impression they will autoplay when clicked, when in fact only video content type pages are eligible for autoplay.

More consistent user experience regarding expected behaviour of play button

No

#### Before
screen shot 2017-02-16 at 16 16 04

#### After

screen shot 2017-02-16 at 16 15 35

## Tested in CODE?
No


duarte:
Sounds good to me

GHaberis (approved):
Looks good

PR #15867, opened 2017-02-17
Fix end slate display on Media Atom video pages

gidsg:
## What does this change?
The End Slate was not displaying when `displayEndSlate=true` parameter was set because were checking the wrong parameter within the template

Restore previous end slate behaviour

No

## Screenshots
screen shot 2017-02-17 at 12 30 49


GHaberis (approved):
love it 👍

PR #15872, opened 2017-02-17
Enable Sonobi when #sonobi

regiskuckaertz:
So the guys at Sonobi can try to reproduce the issue.


static/src/javascripts-legacy/projects/commercial/modules/dfp/dfp-env.js
```diff
@@ -16,7 +16,7 @@ define([
         adSlotSelector: '.js-ad-slot',
 
         /* sonobiEnabled: boolean. Set to true if sonobi real-time-bidding is enabled*/
-        sonobiEnabled: config.switches.sonobiHeaderBidding,
```
rich-nguyen: || ?

rich-nguyen (approved):
thanks! 👍

PR #15873, opened 2017-02-17
Fix for sticky ads

lps88:
## What does this change?
There isn't enough space in the right column for the MPU to be sticky on some shorter articles eg. [here](https://www.theguardian.com/stage/2017/feb/16/miranda-hart-west-end-debut-annie-musical-miss-hannigan) and [here](https://www.theguardian.com/business/2017/feb/15/lingerie-brand-agent-provocateur-could-be-heading-for-administration). 

This disables the sticky behaviour in those articles so Most Pop container doesn't crash into the bottom of the article:

image


regiskuckaertz (approved):
Good 👍

PR #15878, opened 2017-02-20
Remove requirejs

sndrs:
## What does this change?

_this is a redux of #15845, which was maybe getting a bit hard to follow and had significant changes..._

- removes requirejs from frontend
- removes the webpack switch (we cannot turn off webpack)
- leaves `curl` for interactives

one module bundler

fingers crossed, no

## Tested in CODE?

- [x] tested on code


regiskuckaertz:
giphy

static/src/javascripts-legacy/bootstraps/enhanced/media/video-player.js
```diff
@@ -10,9 +10,7 @@ define([
     // In rjs this module is shimmed in the config
     // https://github.com/guardian/frontend/blob/webpack-redux/tools/__tasks__/compile/javascript/rjs.config.js#L42
     // We can move this up to the define() when we go 100% WP
```
SiAdcock: Can we action this comment now, or will there be a forthcoming PR to deal with comments like this?

sndrs: as discussed irl, we'll do this when cleaning out all the `#wp-rjs`s

gtrufitt (approved):
👍 💥

PR #15879, opened 2017-02-20
Zero-dimension slots and IntersectionObserver

regiskuckaertz:
The IO spec has been recently changed to account for [zero-dimension elements](https://github.com/WICG/IntersectionObserver/issues/69), resulting in the addition of a new `isIntersecting` property. We need to take it into account, otherwise newest versions of Chrome will display all slots at once.

Before:
picture 4

After:
picture 3


katebee:
Trying this on Chrome 55.0.2883.95 (64-bit), this expected?

picture 535

katebee (approved):
👍

PR #15881, opened 2017-02-20
Add server-side ad targeting to pressed pages

kelvin-chappell:
This uses the commercial shared library to ensure fronts have similar targeting across web and apps.

Like #15829 the ad targeting isn't actually being used yet.  This will just expose it on the client-side for comparison.


kelvin-chappell:
[Standardise ad call parameters on web and apps](https://trello.com/c/Ha3BvjDk/166-standardise-ad-call-parameters-on-web-and-apps)

regiskuckaertz (approved):
Yahaa!

PR #15883, opened 2017-02-20
Bump version of commercial-shared dependency

kelvin-chappell:
To get change in 'about this' link on logos.
It now defaults to https://www.theguardian.com/info/2016/jan/25/content-funding.


kelvin-chappell:
[Fix 'about this' link](https://trello.com/c/oDuQaNhp/202-fix-about-this-link)

PR #15885, opened 2017-02-20
Extend recommendations expiry date

davidfurey:
## What does this change?

Extends the recommendations expiry date by two weeks because we don’t have enough data yet


katebee (approved):
:+1: LGTM

the AB test JS will now run until the end of the expiry day, so it may not be necessary to have the switch expire one day earlier, but it wont hurt 👍

PR #15887, opened 2017-02-21
Moving the config to version 24

DiegoVazquezNanini:
Push the config to version 24
@jfsoul


jfsoul (approved):
For future reference, these changes are changes to hostnames and related config to allow https on `code`.

PR #15889, opened 2017-02-21
Bugfix for liveblog splash updates

MatthewJWalls:
## What does this change?

Bugfix for https://trello.com/c/Q5mTqXho/162-choosing-show-updates-with-liveblog-splash-on-fronts-no-longer-displays-a-rolling-update

Two fixes to this code:

1. We must request at least a few blocks from CAPI or else it isn't recognised as being a live blog, so adding Some(Canonical) will request a standard five blocks.
2. The body of the live blog update must be plaintext, not html, so use the new summary field from CAPI.

When you see a liveblog on a front, we will now correctly load in the second-to-most-recent update summary into the liveblog box, then it will auto-scroll to the most recent update. Updates the user has already seen will not be shown.

No


sndrs (approved):
lgtm!

sndrs:
want to merge this matt?

PR #15892, opened 2017-02-21
Fix the ophan epic tracking

jranks123:
## What does this change?

We were not sending the full campaign code to ophan, meaning that we couldn't tell which variant we were tracking. This PR fixes this.


desbo (approved):
Looks good but maybe we should hold off on releasing it, since it'll change the campaign code of running tests.

sndrs:
is it possible to merge this yet?

sndrs:
an news? 2 weeks old and approved too...

PR #15893, opened 2017-02-21
Clear left-hand side of inline ads in articles

regiskuckaertz:
Slightly improve the layout of articles when a rich link and an ad sit in the same area.

Before:
picture 10

After:
picture 9


regiskuckaertz:
[Spacefinder clashes with rich media links in this article](https://trello.com/c/ajReovBx/198-spacefinder-clashes-with-rich-media-links-in-this-article)

scottp719 (approved):
👍

PR #15896, opened 2017-02-21
check youtube atom is main video when checking should autoplay

GHaberis:
## What does this change?

- Fixes this bug: All playable youtube atoms on a page were auto playing if the parent page was a Video page, the referrer was internal and they weren't being viewed on an auto play blocking platform. Only the main video youtube atom should auto play.  
- I've added an additional check to see whether the atom is within a figure with the class "main video".

- Stops multiple videos playing

No

## Screenshots

N/A

## Tested in CODE?

No

@gidsg @akash1810

### DIFF
--- a/static/src/javascripts-legacy/projects/common/modules/atoms/youtube.js
+++ b/static/src/javascripts-legacy/projects/common/modules/atoms/youtube.js
@@ -5,7 +5,8 @@ define([
     'common/modules/component',
     'common/utils/$',
     'common/utils/config',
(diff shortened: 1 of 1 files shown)


gidsg (approved):
⏯ 👍  Looks good to me. If this turns out to be flaky we could add a more specific data attribute but I think it's nice to work with what we have already in the DOM (this fragment is only used by the video canonical page so should be pretty stable.)

akash1810 (approved):

PR #15901, opened 2017-02-21
docs: fix broken link pointing at nginx config

gu-stav:
## What does this change?

Fixes a broken link in the documentation.

A better documentation.


jfsoul (approved):
Thanks!

PR #15902, opened 2017-02-21
Don't transpile in compilation

sndrs:
## What does this change?

removes the transpilation step in compile

no longer needed for compilation, only in karma 

no


SiAdcock (approved):
Excellent 👍

PR #15907, opened 2017-02-21
Move the MostPop MPU to Scala ✨🌟

katebee:
## What does this change?

Let our MPUs break free from their cage! No not really, but we can at least drag them higher in the document tree, and delete a bit of JS.

Part 2 of https://github.com/guardian/frontend/pull/15788

##### Long version:
There is a 'most viewed' container at the bottom of articles. This is inserted by the enhanced JS, as part of [trails/onward](https://github.com/guardian/frontend/blob/master/static/src/javascripts-legacy/bootstraps/enhanced/trail.js). The 'most viewed' container is home to the `dfp-ad--mostpop` MPU, which currently relies on `popular.js` to insert and render it. 

N.B. The 'most viewed' container on fronts/facia is not touched by `popular.js`, however, it does share use of the `popular.scala.html` template.

##### Short version:

- Delete some JS, replace with HTML and some CSS
- Use the mostPopPlaceholder to include the MPU in the template, and reserve space for the injected content
- Bonus: Fix layout issue affecting the 'most viewed' container in Firefox 

- Less YANK on the page, less JS, and less DOM manipulation 🖥 💨
- Stop relying on the enhanced JS to add `dfp-ad--mostpop` into article pages, bring the responsibility home to commercial JS, where it belongs 🏡 🤗 
- Make 'most viewed' use the space properly in Firefox

Nope

###### BEFORE:

picture 557

###### Firefox AFTER:

picture 556

_____________

###### BEFORE:

picture 559

###### AFTER:

picture 558

## Tested in CODE?

- [x] test in CODE
- [x] & test in Chrome
- [x] & test in Safari
- [x] & test in Firefox


common/app/views/fragments/mostPopularPlaceholder.scala.html
```diff
@@ -13,9 +15,22 @@ <h2 class="fc-container__header__title">
...
-
+            <div class="fc-container__body fc-container--rolled-up-hide popular-trails">
+                <div class="popular-trails__content js-popular-trails">
+                </div>
+                <div class="popular-trails__mpu">
+                    <div class="fc-slice__popular-mpu js-fc-slice-mpu-candidate fc-slice__item--mpu-candidate">
+                        @adSlot(
+                        "mostpop",
+                        Seq("container-inline"),
+                        Map("mobile" -> Seq("1,1", "2,2", "300,250", "fluid"), "desktop" -> Seq("1,1", "2,2", "300,250", "620,1", "620,350", "fluid")),
```
regiskuckaertz: You can remove the `desktop` sizes, will never be used.

common/app/views/fragments/mostPopularPlaceholder.scala.html
```diff
@@ -13,9 +15,22 @@ <h2 class="fc-container__header__title">
                     <a href="@LinkTo(url)" data-link-name="Most viewed @section">most viewed</a>
                 </h2>
             </div>
-            <div class="fc-container__body fc-container--rolled-up-hide js-popular-trails">
-
+            <div class="fc-container__body fc-container--rolled-up-hide popular-trails">
+                <div class="popular-trails__content js-popular-trails">
+                </div>
+                <div class="popular-trails__mpu">
+                    <div class="fc-slice__popular-mpu js-fc-slice-mpu-candidate fc-slice__item--mpu-candidate">
```
regiskuckaertz: Cannot the `js-` class go away?

static/src/javascripts-legacy/projects/common/modules/onward/popular.js
```diff
@@ -42,22 +41,7 @@ define([
         this.fetch(qwery('.js-popular-trails'), 'html');
     };
 
-    MostPopular.prototype.mobileMaximumSlotsReached = function () {
-        return (detect.getBreakpoint() === 'mobile' && $('.ad-slot--inline').length > 1);
```
regiskuckaertz: Nice to see this go! I believe the module signature can also be reduced.

common/app/views/fragments/mostPopularPlaceholder.scala.html
katebee: Hmm, depends. I was keeping it so unfilled slots could be closed, for example:

The Children's book section doesn't show ads, so at a tablet breakpoint it has a large empty space where the MPU would be:

picture 553

If we keep `js-fc-slice-mpu-candidate`, then the `fc-slice__item--no-mpu` class will be added by `render-advert.js`, I can use this to close the slot:

picture 552

Or would it be better to leave it open?

regiskuckaertz: We close slots when `dfpAdvertising` is false. Does that mean the enclosing element has dimensions?

katebee: `fc-slice__item--mpu-candidate` has a min-height, but only in tablet? 🤔

Origin story: https://github.com/guardian/frontend/pull/6736/files#diff-aab6219b8b34a505d876d268a7c2f735R38

It doesn't seem to have any negative consequences if I delete those lines... ¯\\\_(ツ)\_\/¯

regiskuckaertz: Yeah, probably a residue of the time it was positioned absolutely 💩

common/app/views/fragments/collections/popular.scala.html
```diff
@@ -49,16 +50,13 @@
                 </ul>
             </div>
         }
-
-        @if(showMPU(containerDefinition)) {
+        @if(showMPU(containerDefinition) && isFront) {
             <div class="fc-slice__popular-mpu js-fc-slice-mpu-candidate fc-slice__item--mpu-candidate">
```
regiskuckaertz: This one here too?

regiskuckaertz (approved):
Perhaps interesting to write Scala tests? Otherwise, if your test are conclusive, good riddance 👍

YetiFace:
Does anyone have an idea when this'll go live? As a Firefox user the single column Most Viewed section has been standing out since it first broke.

PR #15909, opened 2017-02-21
amp: fix position of cta button if amp-list is collapsed

gu-stav:
## What does this change?

Overwrites the default cta position of an element within an `amp-list`.

Temporarily bugfix, till the AMP issue is fixed.

Amp

### DIFF
--- a/static/src/stylesheets/amp/_buttons.scss
+++ b/static/src/stylesheets/amp/_buttons.scss
@@ -16,12 +16,20 @@
 }
 
 .cta--show-more {
-    position: absolute;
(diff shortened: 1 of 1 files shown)


NataliaLKB (approved):
lgtm 👍 .

Have you checked AMP validates?

gustavpursche:
@NataliaLKB Yes, I did. The AMP team already replied - they are probably going to reflect the CSS overwrite in their documentation, so this way isn't even a hack. Maybe I can just write it a bit cleaner, after they changed the docs.

PR #15910, opened 2017-02-21
Bye bye bower

sndrs:
## What does this change?

removes bower*

one package manager

## Tested in CODE?

- [x] test on CODE

*_it's still used by karma, but we're freezing that at the same time as sunsetting it. bower and requirejs will both truly be gone when karma goes_


SiAdcock (approved):
Solid de-bowering m9 👍

PR #15911, opened 2017-02-21
Make svg-loader more Webpack-like

SiAdcock:
## What does this change?

Update references to the `inlineSvg` loader to the more Webpack-friendly `svg-loader`. This means we don't need the loader alias in the Webpack config.

Rids ourselves of one more vestige of RequireJS.

No

## Tested in CODE?

- [ ] Test in CODE


sndrs (approved):
gert lush

SiAdcock:
In a bit sad that I couldn't do the same for the raw-loader. Sadly, this breaks the tests as RequireJS expects its plugins' filename to be the same as the plugin name(?!) Since the text plugin (RequireJS equivalent of raw-loader) is installed via NPM, we have no control over the plugin's filename.

I guess we could rename it or symlink it on post install. Or check it in somewhere. All these options feel slightly icky.

sndrs:
can we alias it in karma's require config?

SiAdcock:
@sndrs No, I tried and RequireJS couldn't find the plugin. When I renamed the plugin from `node_modules/text/text.js` to `node_modules/text/raw-loader.js` it worked, which was interesting.

regiskuckaertz:
You remove 66K lines and you're jealous of my weekly cleanup?! 😆 

👍

SiAdcock:
Oh, wait, that ain't right! Thanks for pointing that out 😄

PR #15912, opened 2017-02-22
Remove epic one line v2 test

jranks123:
## What does this change?

The test reached significance, so we are pulling the test


katebee (approved):
LGTM 👍

PR #15913, opened 2017-02-22
Reverse polarity of outbrain compliance test

davidfurey:
## What does this change?

There might be a race condition between email signup boxes loading and outbrain loading, which would in some cases result in non-compliant outbrain calls when we could have made a compliant call.  This change should mean that if a race condition happens, we err on the side of assuming that outbrain is compliant.

More compliant outbrain calls, measured from outbrain daily report


joelochlann (approved):
Looks good 👍

static/src/javascripts-legacy/projects/common/modules/email/run-checks.js
```diff
@@ -159,7 +159,7 @@ define([
             !clash.userIsInAClashingAbTest(clash.nonEmailClashingTests) &&
             storage.session.isAvailable() &&
             !userHasSeenThisSession() &&
-            !compliantObWidgetIsShown() &&
+            nonCompliantObWidgetIsShown() &&
```
GHaberis: Just a small thing...the function call nonCompliantObWidgetIsShown is kind of unnecessary now, you could just use the variable nonCompliantOutbrain in this conditional check. eg. !userHasSeenThisSession() && nonCompliantOutbrain. Then you can remove the function nonCompliantObWidgetIsShown as this is the only place it's called from.

GHaberis (approved):
Looks good, makes sense that only adding email signups if outbrain has set nonCompliantOutbrain to true will remove occurrences of invalidating compliant outbrain containers.

PR #15914, opened 2017-02-22
Add ISBN to inline merchandising slot targeting when available

regiskuckaertz:
The inline book component may need this value specified in composer. This is to ease the process of setting up inline components for the merchandising team, who doesn't have to manually specify the ISBN corresponding to each book-related article.

(It is not elegant but I have a refactoring of `createSlot` and `defineSlot` in the pipeline, so enough for now)

Tested on [this article](https://www.theguardian.com/books/2017/feb/22/utopia-for-realists-and-how-we-can-get-there-by-rutger-bregman-review?adtest=native-inline-book).

picture 5


JonNorman (approved):
Lovely stuff

PR #15916, opened 2017-02-22
Server side test: Always use 5x3 image for youtube atom content cards

gidsg:
## What does this change?
Adds a server side opt-in test which when enabled means youtube atom playable content cards on fronts will always use a 5x3 trail image. 

Note the fastly changes required for an opt-in link will follow in a subsequent PR.

This matches the current behaviour of playables for non-youtube media but has some downsides as described below.

### Pros
* Editorial can show a different image on fronts than on articles
* Better matches other cards in the grid

### Cons
* Noticeable flash when we enhance from YouTube holding image to Guardian one
* Letterboxing on video player on fronts

Running this as a opt-in test will allow editorial to evaluate the pros and cons of this behaviour whilst using it in production.

No

### Before

16 x 9

#### After (when switch enabled)

5 by 3

## Tested in CODE?
No

### DIFF
--- a/common/app/mvt/MultiVariateTesting.scala
+++ b/common/app/mvt/MultiVariateTesting.scala
@@ -50,6 +50,17 @@ object CommercialClientLoggingVariant extends TestDefinition(
   }
 }
 
+object YouTubePosterOverride extends TestDefinition(
+  name = "youtube-poster-override",
+  description = "Users in the test will always see the trail image on YouTube atom content cards instead of the poster image",
+  owners = Seq(Owner.withGithub("gidsg")),
+  sellByDate = new LocalDate(2017, 4, 1)
+  ) {
+  def canRun(implicit request: RequestHeader): Boolean = {
+    request.headers.get("X-GU-youtube-poster-override").contains("true")
+  }
+}
+
 trait ServerSideABTests {
   val tests: Seq[TestDefinition]
 
@@ -65,7 +76,8 @@ object ActiveTests extends ServerSideABTests {
   val tests: Seq[TestDefinition] = List(
     ABNewNavVariantSeven,
     ABNewNavControl,
-    CommercialClientLoggingVariant
+    CommercialClientLoggingVariant,
+    YouTubePosterOverride
   )
 }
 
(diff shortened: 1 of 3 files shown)


common/app/mvt/MultiVariateTesting.scala
```diff
@@ -50,6 +50,17 @@ object CommercialClientLoggingVariant extends TestDefinition(
   }
 }
 
+object YouTubePosterOverride extends TestDefinition(
+  name = "youtube-poster-override",
+  description = "Users in the test will always see the trail image on YouTube atom content cards instead of the poster image",
+  owners = Seq(Owner.withGithub("gidsg")),
+  sellByDate = new LocalDate(2017, 4, 1)
```
akash1810: build is failing because 1 Apr is a Saturday

akash1810 (commented):
other than build failing, looks good. Will wait for vcl change before approving

duarte:
I discussed with @gidsg earlier — I would really like to look into only showing the youtube video when the enhancement is ready…

GHaberis:
@duarte this is something we've looked into before, unfortunately we don't always enhance (eg. on iPad), this means we can't hide the youtube player by default as then there'd be no way of showing it if enhanced hasn't run.

duarte:
@GHaberis yes! Gids explained it to me, and I get it — but thought perhaps it was worth another go as this thing of the YT interface popping up before ours feels slightly disappointing…

Perhaps wait for either enhancement or a set timeout. Alternatively we could check for iOS 10 — this could be an easy way of confirming the user is on a reasonably fast iPad. iOS 10 only supports iPad Air and above (this is roughly equivalent to iPhone 5S specs, so likely to support the enhancement?).

In any case, I don't mean to hold up this PR with this, just something that I think would be a good idea in the future…

akash1810:
> Alternatively we could check for iOS 10

the "no fronts enhance on iPad" rule is a blanket one, the enhanced bundle isn't even on the page iirc. To do this change and check would be non-trivial.

> just something that I think would be a good idea in the future…

Definitely a good idea for the future

akash1810 (approved):

PR #15920, opened 2017-02-22
Cleanup Wednesday #3

rich-nguyen:
Memoize the ad unit but defer the memoize until first defineSlot() call.

**Tested in CODE**
yes


regiskuckaertz (approved):
oohoo great 👍

PR #15921, opened 2017-02-22
De-duplicate collections/containers in emails

davidfurey:
## What does this change?

Removes duplicate articles from containers on email fronts

Simplifies curation of email fronts

No

### DIFF
--- a/common/app/layout/CollectionEmail.scala
+++ b/common/app/layout/CollectionEmail.scala
@@ -1,0 +1,52 @@
+package layout
+
+import model.PressedPage
+import model.facia.PressedCollection
+import model.pressed.{CollectionConfig, PressedContent}
+
+import PartialFunction.condOpt
+
+object EmailContainer {
+
+  def fromPressedCollections(pressedCollections: List[PressedCollection]): List[EmailContainer] = {
+    val (_, reversedContainers) = pressedCollections.foldLeft((List.empty[EditionalisedLink], List.empty[EmailContainer])) {
+      case ((alreadySeen, emailContainers), pressedCollection) =>
+        val cards = cardsForCollection(pressedCollection, alreadySeen)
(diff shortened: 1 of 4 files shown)


common/app/layout/CollectionEmail.scala
```diff
@@ -0,0 +1,52 @@
...
+        (newUrls ::: alreadySeen, emailContainer :: emailContainers)
+    }
+    reversedContainers.reverse.filter(_.cards.nonEmpty)
+  }
+
+  private def cardsForCollection(collection: PressedCollection, seen: List[EditionalisedLink]): List[ContentCard] = {
+    collection
+      .curatedPlusBackfillDeduplicated
+      .flatMap(contentCard(_, collection.config))
+      .filterNot({ content => seen.contains(content.header.url) })
```
joelochlann: might be my Scala knowledge at fault but are the `{}` necessary here?

common/app/layout/CollectionEmail.scala
```diff
@@ -0,0 +1,52 @@
...
+
+import model.PressedPage
+import model.facia.PressedCollection
+import model.pressed.{CollectionConfig, PressedContent}
+
+import PartialFunction.condOpt
+
+object EmailContainer {
+
+  def fromPressedCollections(pressedCollections: List[PressedCollection]): List[EmailContainer] = {
```
joelochlann: For me personally, I would have to squint at these functions for quite a while before realising that there was de-duplication across containers going on (though to be fair it's clear from reading the tests). Could we tweak the naming somehow or add a comment clarifying this?

joelochlann (approved):
Generally looks good, I'm just concerned that it wouldn't be immediately obvious when reading the code that this de-duplication is happening, which might trip someone up in future. If we can find an easy way to clarify this, then I'm happy 👍

PR #15922, opened 2017-02-22
Expanded sharing at small break points

zeftilldeath:
On really small devices you see a slither of timestamp when you expand the menu.
That's not sexy and it certainly isn't cricket.

Before
screen shot 2017-02-22 at 16 54 28

After
screen shot 2017-02-22 at 16 54 34


SiAdcock:
article-2484111-00622fe700000258-147_634x436

SiAdcock (approved):

PR #15924, opened 2017-02-22
Extending a/b tests

davidfurey:
## What does this change?

Extends two email a/b tests

Flyer - still to work out how future a/b tests will work so extending
this one for now

Opinion - more data needed to make decision

no

### DIFF
--- a/common/app/conf/switches/ABTestSwitches.scala
+++ b/common/app/conf/switches/ABTestSwitches.scala
@@ -56,7 +56,7 @@ trait ABTestSwitches {
     "Assign users to variants of our editorial emails",
     owners = Seq(Owner.withGithub("davidfurey")),
     safeState = Off,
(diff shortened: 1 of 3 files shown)


SiAdcock (approved):

katebee (approved):
👍

katebee:
Aside thought: when these tests end, any JS on the relevant email interactive will have to be updated/deleted, so that app users also stop being bucketed.

davidfurey:
Yes, good to remember the apps.  There is quite a bit of tidying up to do when the test ends, merging the lists, removing the magic around unsubscribing and updating the interactives. Maybe I should have extended the tests for longer!

rich-nguyen:
@davidfurey I don't think the deploy to PROD succeeded here. It hung. This will be deployed with my changes.

davidfurey:
@rich-nguyen thanks!

PR #15929, opened 2017-02-23
add manual refresh to the service worker

sndrs:
## What does this change?

add a counter in comment we can use to manually force a refresh of the sw on its next check

we're seeing assets being requested multiplied times by SW-enabled browsers. 

forcing the SW to update _seems_ to fix it, but we cannot do it manually currently. 

this is a hopeful stab in fixing that issue!


jfsoul (approved):
Worth a go!

PR #15931, opened 2017-02-23
Improve lazy load of containers on fronts

regiskuckaertz:
The `js-front-bottom` sentinel worked while all containers were all sitting at the same coordinate, but now that ad slots are inserted on the server side, it is not the case anymore. Because of that, the distance detection test is screwed; the end result is that people scroll down and see a wall of ads, and containers only get displayed too late (sometimes never, if the user does not bother to scroll within the test distance of `js-front-bottom`.

This changes addresses the fact that each container sits at a different location, and will only display those within a certain range (both above and below) the viewport.

Before:
containers-before

After:
containers


regiskuckaertz:
[Improve containers lazy loading on fronts](https://trello.com/c/yQ2hz967/214-improve-containers-lazy-loading-on-fronts)

paperboyo:
💎

static/src/javascripts-legacy/projects/facia/modules/ui/lazy-load-containers.js
```diff
@@ -16,30 +14,55 @@ define([
...
-                            $container;
-
-                        if (scrollBottom > bottomOffset - distanceBeforeLoad) {
-                            $container = bonzo(containers.shift());
-
-                            fastdom.write(function () {
-                                $container.removeClass('fc-container--lazy-load');
-                            });
-                        }
+                        var cs = split(containers, withinRange);
```
rich-nguyen: `cs` isn't too clear, i can see you're crying out for destructuring! how about just `splitResult` for now?

static/src/javascripts-legacy/projects/facia/modules/ui/lazy-load-containers.js
```diff
@@ -16,30 +14,55 @@ define([
...
         mediator.on('window:throttledScroll', lazyLoad);
         lazyLoad();
     };
+
+    // split :: [Element] -> (Element -> Bool) -> ([Element], [Element])
+    // invariant: { inA all fn ^ outA all (complement fn) | (inA, outA) = split xs fn }
+    // Split an array of elements into two arrays, one where element comply, one where they don't
+    function split(arr, testFn) {
+        var i = 0, ii = arr.length;
+        var result = [[], []];
```
rich-nguyen: could be an object with two friendly-named keys? `{partitionTrue:[], partitionFalse:[]}`

SiAdcock:
js-front-bottom. Man I nearly spit coffee all over the screen

static/src/javascripts-legacy/projects/facia/modules/ui/lazy-load-containers.js
```diff
@@ -16,30 +14,55 @@ define([
...
             }, 500);
 
         mediator.on('window:throttledScroll', lazyLoad);
         lazyLoad();
     };
+
+    // split :: [Element] -> (Element -> Bool) -> ([Element], [Element])
+    // invariant: { inA all fn ^ outA all (complement fn) | (inA, outA) = split xs fn }
+    // Split an array of elements into two arrays, one where element comply, one where they don't
+    function split(arr, testFn) {
```
rich-nguyen: its called `partition` in scala, could use that name for consistency and dev-friendliness

static/src/javascripts-legacy/projects/facia/modules/ui/lazy-load-containers.js
rich-nguyen: my example looks ugly, actually. `[[],[]]` is cleaner. ignore!

rich-nguyen (approved):

rich-nguyen:
amaze 👍

static/src/javascripts-legacy/projects/facia/modules/ui/lazy-load-containers.js
desbo: i guess `[[], []]` would be closer to the tuple you get from `partition` but i think in js an i'd find an object easier to work with, maybe just with `true` and `false` as keys.

rich-nguyen: that was my initial and second thoughts, in the other order! Maybe you got the balance right, just by suggesting cleaner key names. Let's hope @regiskuckaertz takes the `true` and `false` option.

static/src/javascripts-legacy/projects/facia/modules/ui/lazy-load-containers.js
sndrs: or even something mega explicit, like `containersInRange`?

static/src/javascripts-legacy/projects/facia/modules/ui/lazy-load-containers.js
sndrs: this feels to me like an in situ `reduce` at line 23 would be a more natural way to express this in js? something like:

```js
var containersInRange = containers.reduce(function(result, container) {
    result[inRange(container) ? 'in' : 'out'].push(container);
    return result;
}, { in: [], out: [] });
```
feels a lot more native and quickly understandable?

maybe it's a taste thing, and I don't want to meander off into the bike-shed so ignore if you think I am, but i'm not sure that abstracting it like this – and naming it after a scala method – really _is_ developer friendly. the context here _is_ js...

static/src/javascripts-legacy/projects/facia/modules/ui/lazy-load-containers.js
```diff
@@ -16,30 +14,55 @@ define([
...
+                        });
                     });
                 }
             }, 500);
 
         mediator.on('window:throttledScroll', lazyLoad);
         lazyLoad();
     };
+
+    // split :: [Element] -> (Element -> Bool) -> ([Element], [Element])
```
sndrs: i've not seen this doc syntax before, it took me reading the english to understand it :) what is it?

static/src/javascripts-legacy/projects/facia/modules/ui/lazy-load-containers.js
rich-nguyen: I think if we can keep the context down to just syntax where possible, and not semantics, then our codebase and discussions will be more productive. For example, in your example above, I don't need to check your specific implementation of `reduce`, because I assume it's the same as every other language. You didn't (can't, in this case) use a `fold`, and that conveys meaning to me.

I'll leave the separate question of in-situ or module-scope function for @regiskuckaertz , certainly feels like a matter of taste.

regiskuckaertz: Hi guys! I think for catamorphisms like this one, a reducer is indeed more idiomatic, at least as we try to write more functional expressions. Besides, it felt like too much to write a utility function like that.

static/src/javascripts-legacy/projects/facia/modules/ui/lazy-load-containers.js
regiskuckaertz: @rich-nguyen You're reading my mind. It feels to me destructuring as a weak form of pattern matching is the best feature they added to the language in a long while.

regiskuckaertz:
@sndrs @rich-nguyen @SiAdcock there it is. I also removed the sentinel as it is not useful anymore. Everyone happy with this?

sndrs (approved):
👍 from me

PR #15936, opened 2017-02-23
tests: replace two more legacy rjs text-loader occurences

gu-stav:
## What does this change?

Fixes regression with two forgotten `text!` require-js references.

Tests succeed.

No.

## Tested in CODE?

No.


jfsoul (approved):
looks good thanks!

PR #15937, opened 2017-02-23
Provide Ophan URL for interactives

SiAdcock:
## What does this change?

We have bundled Ophan's Tracker JS into our application, but interactives do not have access to our application.

This change provides the path to the hosted version of Ophan's Tracker JS in the curl config.

Provides interactives with access to Ophan's Tracker JS.

No

## Tested in CODE?

No


jfsoul (approved):
👍

PR #15938, opened 2017-02-23
Support super sticky banner in CSS

regiskuckaertz:
An upcoming campaign is the only instance of a super sticky banner, that goes all the way down the page.

In that situation:

- it must not be wrapped with the header, otherwise it will just disappear
- the paid for band must not be sticky


scottp719 (approved):
Release the banner!

PR #15939, opened 2017-02-23
More consistent facia cards in the recommended for you container

lmath:
## What does this change?
Cards in the recommended for you container have some inconsistencies e.g. bylines when they don't appear elsewhere on the site and with unwanted information (Name + e.g. Scotland editor, for the Guardian, in New York), missing quote marks on opinion cards, etc. 

I added an endpoint in [15396](https://github.com/guardian/frontend/pull/15396/) to get the HTML for the cards including some additional style information, but this was pretty basic and since then I've found that there is some existing code that converts a response from capi to the model needed for facia cards, and that it is more sophisticated. Simply using this existing code for the conversion from capi response to facia card html resolves most of the inconsistencies observed when using the endpoint added in [15396](https://github.com/guardian/frontend/pull/15396/). 

Consistency. 

No. 

## Screenshots
Before:
image image

With these changes:
imageimage

## Tested in CODE?
Yes.


superfrank:
Maybe you might know the answer @joelochlann but any reason there's no padding between the cards on mobile? As atm: theylooklikethis

paperboyo:
💎 but also ☝️

common/app/layout/Sublink.scala
```diff
@@ -329,77 +330,21 @@ object ContentCard {
 
   def fromApiContent(apiContent: contentapi.Content): Option[ContentCard] = {
 
-    apiContent.fields.map { fields =>
+    val cardTypesForRecommendations = ItemClasses(mobile = MediaList, tablet = Standard, None)
```
davidfurey: desktop defaults to None, so isn't needed.  If you want to keep it, my preference is that named parameters should be all or nothing.

`ItemClasses(mobile = MediaList, tablet = Standard, desktop = None)`
or
`ItemClasses(mobile = MediaList, tablet = Standard)`

common/app/layout/Sublink.scala
```diff
@@ -329,77 +330,21 @@ object ContentCard {
...
-        useShortByline = false,
-        group = ""
-      )
+    FaciaCard.fromTrail(
+      faciaContent = FaciaContentConvert.contentToFaciaContent(apiContent),
+      config = CollectionConfig.empty,
+      cardTypes = cardTypesForRecommendations,
+      showSeriesAndBlogKickers = false
+    ) match {
+      case content: ContentCard => Some(content)
```
davidfurey: There is a slightly nicer pattern for this:
```
PartialFunction.condOpt(FaciaCard.fromTrail(etc)) {
   case content: ContentCard => content
}
```

davidfurey:
Apart from the minor style changes: 👍

common/app/layout/Sublink.scala
lmath: TIL! Thanks.

davidfurey (approved):

PR #15943, opened 2017-02-23
Unstick paid for band when super sticky

regiskuckaertz:
It turns out the position of the band in the DOM changes from page type to page type. A more robust approach is then to sign the `body` element with a class and use that as an anchor.


rich-nguyen (approved):
ship it

PR #15948, opened 2017-02-24
Include ophan-tracker-js in tests

SiAdcock:
## What does this change?

Ophan's tracker JS is needed in our Karma tests, specifically as a result of #15947

Karma includes Ophan, RequireJS knows where to find it, tests pass, everyone happy 🌈 

No

## Tested in CODE?

No


desbo (approved):
thank you!

PR #15952, opened 2017-02-24
Video main media caption

zeftilldeath:
Currently, within the new-header test we're getting some horrible banding due to the placement of captions under video-main-media. If this becomes permanent we'll need to support youtube videos too. 

## This will not do
screen shot 2017-02-24 at 15 57 30

## This will do
screen shot 2017-02-24 at 15 55 47

When a video is main media AND has a caption AND there is a feature tone, you get this horrible white banding under the image to make room for the caption. Instead with a lot of help from @NataliaLKB I have used the caption toggle just like images that are main media.

### DIFF
--- a/article/app/views/fragments/mainMedia.scala.html
+++ b/article/app/views/fragments/mainMedia.scala.html
@@ -42,10 +42,22 @@
 
 
                     @mainVideo.videos.caption.map { caption =>
-                        <figcaption data-link-name="Video caption link" class="caption caption--main caption--video" itemprop="description">
-                            @fragments.inlineSvg("information", "icon", List("rounded-icon", "centered-icon"))
-                            @Html(caption)
-                        </figcaption>
+                        @if(mvt.ABNewNavVariantSeven.isParticipating) {
+                            <input type="checkbox" id="show-caption" class="mobile-only u-h">
+
+                            <label class="mobile-only reveal-caption reveal-caption--img" for="show-caption">
+                                @fragments.inlineSvg("information", "icon", List("reveal-caption-icon", "centered-icon"))
+                            </label>
+
+                            <figcaption data-link-name="Video caption link" class="caption caption--main caption--video" itemprop="description">
+                                @Html(caption)
+                            </figcaption>
+                        } else {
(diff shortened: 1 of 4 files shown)


article/app/views/fragments/mainMedia.scala.html
```diff
@@ -42,10 +42,24 @@
 
 
                     @mainVideo.videos.caption.map { caption =>
-                        <figcaption data-link-name="Video caption link" class="caption caption--main caption--video" itemprop="description">
-                            @fragments.inlineSvg("information", "icon", List("rounded-icon", "centered-icon"))
-                            @Html(caption)
-                        </figcaption>
+                        @if(mvt.ABNewNavVariantSeven.isParticipating) {
```
gidsg: To clarify, is this solely an issue for people in this A/B test? If so would have been helpful to put this in the description.

static/src/javascripts-legacy/projects/common/modules/video/events.js
```diff
@@ -275,12 +275,17 @@ define([
...
+        $('.js-hide-on-play').hide();
+    }
+
     // These events are so that other libraries (e.g. Ophan) can hook into events without
     // needing to know about videojs
     function bindGlobalEvents(player) {
         player.on('playing', function () {
             kruxTracking(player, 'videoPlaying');
             bean.fire(document.body, 'videoPlaying');
+            hideCaption();
```
gidsg: Are you doing this all the time or only when the video plays

gidsg:
It would be worth testing this same scenario with YouTube atoms, as I think they would need to be updated separately, though this can be done in a separate PR.

gidsg:
Would be helpful for me to talk with @zeftilldeath or @NataliaLKB to provide a bit more context if possible.

NataliaLKB:
@gidsg Best to chat to @zeftilldeath about this! I am in NY until the 14th. But after the 14th I am free to chat 😄

gidsg:
Talked through IRL with @zeftilldeath looks good to me 👍 🎥

gidsg (approved):

PR #15954, opened 2017-02-24
Liveblog: prevent share-numbers container from changing its height

gu-stav:
## What does this change?

Prevents share-number container next to liveblog from changing its height, while fetching the counts.

Less visual changes, happier users.

No.

## Screenshots

No.

## Tested in CODE?

No.


SiAdcock (approved):
💎 💎 💎

PR #15955, opened 2017-02-24
engagement banner copy tests

Mullefa:
@guardian/contributions 

## What does this change?

Tests 3 new copies for the membership engagement banner. For each variant, any reference to the cost of membership will be given in local currency (as defined by the reader's geo location). This should improve consistency between the banner and the membership website. As we are testing the copy, the colour of the banner (dark blue) has been kept consistent for this test.

To find a copy with a higher conversion rate than the control. Success can be measured by conversion rate.

No

### Control (Pounds)

control_pounds

### Control (Euros)

control_euros

### Control (Canadian Dollars)

control_ca_dollars

### Control (US Dollars)

control_us_dollars

### Weekly (Pounds)

weekly_pounds

### Social Proof (Pounds)

social_proof_pounds

### Opportunity (Pounds)

opportunity_pounds

## Tested in CODE?

No. Tested locally.

### DIFF
--- a/static/src/javascripts-legacy/projects/common/modules/experiments/tests/membership-engagement-banner-tests.js
+++ b/static/src/javascripts-legacy/projects/common/modules/experiments/tests/membership-engagement-banner-tests.js
@@ -171,8 +171,8 @@ define([
 
     var CopyTest = function() {
         this.id = 'MembershipEngagementBannerCopyTest';
-        this.start = '2017-02-22'; // TODO
-        this.expiry = '2017-03-10'; // TODO
+        this.start = '2017-02-27';
+        this.expiry = '2017-03-13';
         this.author = 'Guy Dawson';
         this.description = 'Test different copy for the engagement banner.';
         this.audience = 0.5;
--- a/static/src/javascripts-legacy/projects/common/modules/experiments/tests/membership-engagement-banner-tests.js
+++ b/static/src/javascripts-legacy/projects/common/modules/experiments/tests/membership-engagement-banner-tests.js
@@ -2,6 +2,7 @@ define([
     'bean',
     'qwery',
     'common/utils/config',
+    'common/utils/geolocation',
     'common/utils/storage',
     'common/utils/template',
     'commercial/modules/commercial-features',
@@ -11,6 +12,7 @@ define([
     bean,
     qwery,
     config,
+    geolocation,
     storage,
     template,
     commercialFeatures,
@@ -54,7 +56,6 @@ define([
 
     };
 
-
     EditionTest.prototype.addMessageVariant = function (variantId, variantParams) {
         this.variants.push({
             id: variantId,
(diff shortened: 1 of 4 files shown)


Mullefa:
Pull request ready for review.

@guardian/membership-and-subscriptions 
@guardian/contributions

static/src/javascripts-legacy/projects/common/utils/geolocation.js
```diff
@@ -51,14 +51,98 @@ define([
...
-        getSync: function() {
-            var geolocationFromStorage = storage.local.get(storageKey);
-            return geolocationFromStorage ? geolocationFromStorage : editionToGeolocation(config.page.edition)
-        },
+    // Returns one of { GB, US, AU, CA, EU, INT }
+    // These are the different 'regions' we accept when taking payment.
+    // See https://membership.theguardian.com/uk/supporter# for more context.
+    function getSupporterPaymentRegion() {
+        var location = getSync();
+        for (var i = 0; i < regionCountryCodes.length; i++) {
```
alexduf: I think you could use
```javascript
if (regionCountryCodes.indexOf(location) > -1) {
  return location;
}
if (europCountryCodes.indexOf(location) > -1) {
  return 'EU';
}
```

alexduf (approved):
My only comment is a nitpick, feel free to ignore :)

static/src/javascripts-legacy/projects/common/utils/geolocation.js
Mullefa: but I want to use a proper `for` loop 😉

static/src/javascripts-legacy/projects/common/modules/experiments/tests/membership-engagement-banner-tests.js
```diff
@@ -80,8 +81,177 @@ define([
...
+    function isNotInUSOrAU() {
+        var countryCode = geolocation.getSync();
+        return countryCode !== 'US' && countryCode !== 'AU'
+    }
+
+    // Prices taken from https://membership.theguardian.com/<region>/supporter
+    var monthlySupporterCost = {
+        GB:  '£5',
+        US:  'US$6.99',
+        AU:  'AU$10',
```
svillafe: Since we are not showing the test in US and AU, it is better to delete this currency, isn't it?

Mullefa: Hey, had a bit more of a think regarding this since we spoke. Since the test isn't running in the US or AUS, `getSupporterPaymentRegion()` should never return `US` or `AU`, but in the instance it does, I think it makes sense to use the local cost. When the test is released, I'll use the `pageviews` table to determine whether this is happening at all. Is that ok?

rtyley:
I don't know if you've already seen this but you might find the data at https://membership.theguardian.com/currencies useful - says what currency Membership uses for every 2-character ISO country code.

PR #15957, opened 2017-02-24
Amp Cleaner - Allow cleaning of elements embedded as element-embed

michaelwmcnamara:
Certain frequently used page elements, including youtube, vimeo and googleMaps, can be embedded in composer using embed code rather than a url. When this happens they are created inside a figure container of class "element-embed" instead of "element-<embed-type>".
Our current cleaner removes any embeds of this type (with the exception of soundcloud embeds).
This change is to allow the embedding of other embed types frequently added in this way.

Embed types that will now display on amp even if in an "element-embed" figure are:
- soundcloud 
- audioboom (the other commonly used audio embed)
- external videos (eg: youtube, facebook and vimeo)
- instagram
- googlemap

This passes all tests and passes the make validate-amp test
## Tested in CODE?
Nope

As mentioned above, this is working, but I suspect there are more elegant ways to achieve this and am keen for suggestions.
cc @guardian/dotcom-platform

### DIFF
--- a/common/app/views/support/cleaner/AmpEmbedCleaner.scala
+++ b/common/app/views/support/cleaner/AmpEmbedCleaner.scala
@@ -107,7 +107,6 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
     }
 
     def getTrackIdFromUrl(soundcloudUrl: String): Option[String] = {
-      //val pattern = ".*api.soundcloud.com/tracks/(\\d+).*".r
       URLDecoder.decode(soundcloudUrl,"UTF-8") match {
         case soundCloudPattern(trackId) => {
           Some(trackId)}
--- a/common/app/views/support/cleaner/AmpEmbedCleaner.scala
+++ b/common/app/views/support/cleaner/AmpEmbedCleaner.scala
@@ -124,15 +124,15 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
 
 
   object AmpAudioElements {
-    def createAmpIframeElement(document: Document, src: String, width: String, height: String, frameborder: String): Element = {
+    def createAmpIframeElement(document: Document, iframe: Element): Element = {
       val ampIframe = document.createElement("amp-iframe")
       val attrs = Map(
-        "width" -> width,
-        "height" -> height,
+        "width" -> iframe.attr("width"),
+        "height" -> iframe.attr("height"),
         "layout" -> "responsive",
         "sandbox" -> "allow-scripts allow-same-origin allow-popups",
-        "frameborder" -> frameborder,
-        "src" -> src)
+        "frameborder" -> iframe.attr("frameborder"),
+        "src" -> iframe.attr("src"))
       attrs.foreach {
         case (key, value) => ampIframe.attr(key, value)
       }
@@ -152,11 +152,7 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
         } else {
           // if iframe is not a SoundCloud Element, replace it with amp-iframe
           if (validIframe(iframeElement)) {
-            val src = iframeElement.attr("src")
-            val width = iframeElement.attr("width")
-            val height = iframeElement.attr("height")
-            val frameBorder = iframeElement.attr("frameborder")
-            iframeElement.replaceWith(createAmpIframeElement(document, src, width, height, frameBorder))
+            iframeElement.replaceWith(createAmpIframeElement(document, iframeElement))
           } else {
             iframeElement.remove()
           }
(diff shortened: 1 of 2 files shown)


jfsoul:
I would say that each type of embed needs to inherit from the same parent class or trait so that you can iterate through a sequence of them and end up with something like (pseudocode):

    embedType.foreach {
      if (embedType.pattern matches href) {
         embedType.createElement(document)
         exit foreach
      }
    }

In practice it may be more complicated that this but there should be a consistent interface for all embedTypes that defines a pattern, and a way to create an element or change the document in the desired way.

common/app/views/support/cleaner/AmpEmbedCleaner.scala
```diff
@@ -104,9 +107,9 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
     }
 
     def getTrackIdFromUrl(soundcloudUrl: String): Option[String] = {
-      val pattern = ".*api.soundcloud.com/tracks/(\\d+).*".r
+      //val pattern = ".*api.soundcloud.com/tracks/(\\d+).*".r
```
DiegoVazquezNanini: Delete this comment?

common/app/views/support/cleaner/AmpEmbedCleaner.scala
```diff
@@ -281,7 +299,7 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
 
 
   def cleanAmpEmbed(document: Document) = {
-    document.getElementsByClass("element-embed")
+   /* document.getElementsByClass("element-embed")
```
DiegoVazquezNanini: Maybe remove this commented block?

common/app/views/support/cleaner/AmpEmbedCleaner.scala
michaelwmcnamara: Hi, this should be removed now.

common/app/views/support/cleaner/AmpEmbedCleaner.scala
michaelwmcnamara: This should also be gone now.

michaelwmcnamara:
@DiegoVazquezNanini @jfsoul 
Hi fellas, 
The issues you raised for this PR should now be fixed.
Cheers

Mike

common/app/views/support/cleaner/AmpEmbedCleaner.scala
```diff
@@ -280,19 +293,91 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
...
-      .foreach(_.getElementsByTag("iframe").foreach {
-        //check for soundcloud embeds and remove any others
-        iframeElement: Element =>
-          val soundcloudElement = AmpSoundcloud.getSoundCloudElement(document, iframeElement)
-          if (soundcloudElement.nonEmpty) {
-            iframeElement.replaceWith(soundcloudElement.get)
-          } else
-            iframeElement.remove()
-      })
+  sealed abstract class AmpElementEmbed(val ampElementName: String, matchingUrl: String, pattern: Regex, x: (Document, Element) => Option[Element]){
```
jfsoul: rename `x`

common/app/views/support/cleaner/AmpEmbedCleaner.scala
```diff
@@ -280,19 +293,91 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
...
+    val isAMatch = pattern.findFirstIn(matchingUrl).isDefined
+    def createInstance(doc: Document, element: Element): Option[Element] = x.apply(doc,element)
+  }
+
+  val audioBoomPattern = """^https?:\/\/audioboom\.com\/""".r
+  val instagramPattern = """^\/\/platform\.instagram\/.com\/*""".r
+  val googleMapsPattern = """^https?:\/\/www\.google\.com\/maps\/embed*""".r
+
+  case class GoogleMapsEmbed(document: Document, iframe: Element) extends  AmpElementEmbed("amp-iframe", getAttrib(iframe, "src"), googleMapsPattern, (d: Document, e: Element) => Some(AmpMaps.createMapElement(d, e)))
```
jfsoul: maybe don't specify the `Document` and `Element` types for every one of these

common/app/views/support/cleaner/AmpEmbedCleaner.scala
```diff
@@ -280,19 +293,91 @@ case class AmpEmbedCleaner(article: Article) extends HtmlCleaner {
...
+    iframe.getElementsByTag("a").map(_.attr("href")).headOption.getOrElse("")
+  }
+
+  def getAudioBoomElement(doc: Document, element: Element): Option[Element] = {
+    if(AmpAudioElements.validIframe(element))
+      Some(AmpAudioElements.createAmpIframeElement(doc, element))
+    else
+      None
   }
+  def getInstagramElement(doc: Document, element: Element): Option[Element] = {
```
jfsoul: not used?

jfsoul:
As discussed f2f with @michaelwmcnamara we think this cleaner has reached a tipping point where we need a refactor.  There's too much going on in one place to make sense of any more, so a bit of simplification and splitting out is needed.  Discussed possible strategies for this f2f.

michaelwmcnamara:
Following talks with @jfsoul its been decided to implement the changes described above in a new PR. Closing this one.

PR #15958, opened 2017-02-27
Use prettier for main app js formatting

sndrs:
## What does this change?

- adds [`prettier`](https://github.com/prettier/prettier) to the main app eslint config
- this supplants any formatting-only rules from whichever eslint configs we want to use i.e. it's the law about formatting
- update: defaults to double quotes, like in scala (less context switching)

- very nice styling of code that accounts for readability as well as clarity (i.e. line length etc)
- if this is accepted, I'll add it non-legacy JS too


desbo:
haven't seen prettier before, but i think this is a great idea

gtrufitt (approved):
`bikeshedding--` 👍

PR #15971, opened 2017-02-28
Remove React alias

SiAdcock:
## What does this change?

We no longer need to alias `react` as `react/addons` in the Webpack config. We can do this inline in the module definition.

Less legacy config from the RequireJS times.

No

## Tested in CODE?

No

### DIFF
--- a/static/src/javascripts-legacy/bootstraps/enhanced/accessibility.js
+++ b/static/src/javascripts-legacy/bootstraps/enhanced/accessibility.js
@@ -1,5 +1,5 @@
 define([
-    'react',
+    'react/addons',
     'common/modules/accessibility/main'
 ], function (
     React,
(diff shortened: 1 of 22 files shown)


sndrs:
if we don't alias it, and someone imports `react` without realising we're bundling `react/addons` already, do we end with both being bundled?

SiAdcock:
@sndrs No. `ReactWithAddons` imports `React` anyway, and Webpack takes care of the deduping.

Some of the benefits of `react/addons` over plain old `react` are quite esoteric, some of them are now deprecated, and it's questionable whether we need to use it in any of our modules. I think it's beneficial to give module writers the choice of one or the other depending on whether they need the addons, rather than forcing them to always use `react/addons`

sndrs:
> No. ReactWithAddons imports React anyway, and Webpack takes care of the deduping.

in that case great 👍 totally agree about making choices, was only worried people could accidentally bundle react twice without knowing.

> it's questionable whether we need to use it in any of our modules

yeah we should have a look at that. also, see how easy it would be to upgrade to latest/jsx it in es6 conversion

sndrs (approved):
to misquote @gtrufitt: `config--` ❤️

PR #15972, opened 2017-02-28
Enhance async CSS loading performance

gu-stav:
## What does this change?

Changes the way we async load CSS files. Before we always loaded all `<link rel="stylesheet" />` tags with a JS function. Now we rely on `rel="preload"` and polyfill the attribute behavoir for browsers which don't understand it yet.

It's supported currently on Chrome and Android and is under development in [FF and Safari](https://developer.microsoft.com/en-us/microsoft-edge/platform/status/preload/).

Bonus: async CSS loading works in FF at all now (through the polyfill).

Although the new JS loader looks much bigger, it's actually only an increase of ~100 Bytes after being gzipped.

Improved loading times, reduce the initial paint flickering on articles with async CSS, which then leads to a better UX.

No.

### Chrome

We should see a better performance.

** before (with Queueing) **

screen shot 2017-02-28 at 10 59 33

** after (without Queueing) **

screen shot 2017-02-28 at 10 58 15

### Firefox

We should see about the same performance.

** before **

screen shot 2017-02-28 at 12 24 07

** after **

screen shot 2017-02-28 at 12 18 52

### Android 4.4

We should see about the same performance.

** before **

screen shot 2017-02-28 at 12 28 03

** after **

screen shot 2017-02-28 at 12 34 03

### Android 7

We should see a better performance.

** before **

screen shot 2017-02-28 at 12 39 45

** after **

screen shot 2017-02-28 at 12 36 05

## Tested in CODE?

No.

### DIFF
--- a/common/app/templates/inlineJS/blocking/enableStylesheets.scala.js
+++ b/common/app/templates/inlineJS/blocking/enableStylesheets.scala.js
@@ -25,7 +25,7 @@
         if (before) {
             ref = before;
         } else {
-            var refs = (doc.body || doc.getElementsByTagName( 'head' )[ 0 ]).childNodes;
+            var refs = (doc.body || doc.getElementsByTagName('head')[ 0 ]).childNodes;
             ref = refs[refs.length - 1];
         }
 
@@ -36,7 +36,7 @@
         ss.media = 'only x';
 
         // wait until body is defined before injecting link. This ensures a non-blocking load in IE11.
-        function ready( cb ) {
+        function ready(cb) {
             if (doc.body) {
                 return cb();
             }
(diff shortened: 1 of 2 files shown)


common/app/templates/inlineJS/blocking/enableStylesheets.scala.js
```diff
@@ -1,38 +1,126 @@
...
+            ss.addEventListener('load', loadCB);
         }
+
+        ss.onloadcssdefined = onloadcssdefined;
+        onloadcssdefined(loadCB);
+
+        return ss;
+    };
+
+    preloadSpported = function() {
```
SiAdcock: Will the implicit `var` attach this to the `window`? Also is the typo intentional? 😉

gustavpursche: 👍 Good catch. It's a typo.

gustavpursche:
@sndrs @SiAdcock @gtrufitt I think it's ready for review now. I didn't see any problems in any browsers. I tried to bring back `requestAnimationFrame` at some point, but I've always seen a big flash again, so I guess that's the motivation behind the decision of choosing `setInterval` ...

SiAdcock (approved):
Let's do it. Would be good to collect similar metrics to the above from production after it's merged, just to make sure!

PR #15974, opened 2017-02-28
New header feedback request

zeftilldeath:
Popped a little prompt for feedback into the bottom of the expanded menu.

screen shot 2017-02-28 at 11 06 23


common/app/views/fragments/nav/newHeaderMenu.scala.html
```diff
@@ -130,5 +130,10 @@
                 </a>
             </li>
         </ul>
+
+        <div class="new-header-feedback">
+            @fragments.inlineSvg("information-circle", "icon", List("main-menu__icon", "main-menu__icon--information"))
```
NataliaLKB: Would the class `rounded-icon` work here?

NataliaLKB (approved):
One questions/comment, but looks good! 👍

common/app/views/fragments/nav/newHeaderMenu.scala.html
zeftilldeath: It would make centering the icons a lot more complicated sadly. As the icon layout in the menu used the span and the svg.

NataliaLKB: Fair enough!

PR #15975, opened 2017-02-28
Add research pixels to all Fabric creatives

regiskuckaertz:
This changes addresses a need for advertisers to be able to supply a research pixel in addition to an impression pixel. Also I removed the pixels from the markup, because a JS-only solution works just as well.

### DIFF
--- a/static/src/javascripts-legacy/projects/commercial/modules/creatives/add-tracking-pixel.js
+++ b/static/src/javascripts-legacy/projects/commercial/modules/creatives/add-tracking-pixel.js
@@ -1,14 +1,9 @@
-define([
-    'common/utils/template',
-    'raw-loader!commercial/views/creatives/tracking-pixel.html'
-], function (
-    template,
-    trackingPixelStr
-) {
-    var trackingPixelTpl = template(trackingPixelStr);
-    function addTrackingPixel($adSlot, url) {
-        $adSlot.before(trackingPixelTpl({ url: encodeURI(url) }));
-    }
+define(function () {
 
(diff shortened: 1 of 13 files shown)


regiskuckaertz:
[Fabric tracking pixels](https://trello.com/c/pYqdJ8E5/220-fabric-tracking-pixels)

static/src/javascripts-legacy/projects/commercial/modules/creatives/add-tracking-pixel.js
```diff
@@ -1,14 +1,9 @@
...
-) {
-    var trackingPixelTpl = template(trackingPixelStr);
-    function addTrackingPixel($adSlot, url) {
-        $adSlot.before(trackingPixelTpl({ url: encodeURI(url) }));
-    }
+define(function () {
 
     return addTrackingPixel;
+
+    function addTrackingPixel(url) {
```
JonNorman: For my own understanding: before this function inserted the pixel before the provided `$adSlot`, now it returns an `Image` - where is this actually added to the DOM?

regiskuckaertz: It is not added to the DOM, because that is not necessary. The image will be loaded as soon as the `src` is assigned 😄

JonNorman: Muy excelente

JonNorman (approved):

PR #15977, opened 2017-02-28
Remove US campaign minute email signup and associated logic

joelochlann:
The email signup box at the bottom of a minute-style article was a special case which applied only the US campaign minute signup. This signup box is defunct so I've removed and the associated logic for `end-article` vs `post-article` mod classes. This doesn't affect the appearance of any active signup boxes.

This is a precursor to a restyle which will have all the signup boxes in post-article position, giving it a more "related content" look and allowing us to recommend emails from different sections without it seeming to jarring (e.g. recommending a film email at the end of a football article because we think the user likes film).

@lindseydew @davidfurey

### DIFF
--- a/static/src/javascripts-legacy/projects/common/modules/email/email-article.js
+++ b/static/src/javascripts-legacy/projects/common/modules/email/email-article.js
@@ -42,25 +42,7 @@ define([
     var insertBottomOfArticle = function ($iframeEl) {
             $iframeEl.appendTo('.js-article__body');
         },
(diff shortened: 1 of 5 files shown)


lindseydew:
Nice! Looks good 👍

joelochlann:
Actually also @gtrufitt since he did the original work on this I think

gtrufitt (approved):
Really great this is getting some love and tidied up, thanks Joe! I've seen some of the stuff that Frank has done and it looks great. 

Let me know if you want to catch up around some of the approaches and hindsight I have around this!

davidfurey:
👍

PR #15978, opened 2017-02-28
Update magic ES6 script to allocate work by email address

SiAdcock:
## What does this change?

The script previously allocated modules to developers using their GitHub username. This created some inconsistent behaviour as `config.user.name` ought to contain the user's full name (but doesn't have to). 

I have updated the script to use email addresses instead as they are more reliable (although admittedly this could behave inconsistently too!).

If this doesn't work for individual users, they should be directed to manually update their email address in [`es5to6-files-per-user.js`](https://github.com/guardian/frontend/blob/master/tools/es5to6-files-per-user.js)

Magic ES6 script will work (kinda) 🎩  🐇 

No


tools/es5to6.json
```diff
@@ -1,5 +1,138 @@
...
+    "projects/commercial/modules/dfp/display-lazy-ads.js",
+    "projects/commercial/modules/dfp/empty-advert.js",
+    "projects/commercial/modules/dfp/enable-lazy-load.js",
+    "projects/commercial/modules/dfp/fill-advert-slots.js",
+    "projects/commercial/modules/dfp/get-advert-by-id.js",
+    "projects/commercial/modules/dfp/get-adverts.js",
+    "projects/commercial/modules/dfp/get-creative-ids.js",
+    "projects/commercial/modules/dfp/lazy-load.js",
+    "projects/commercial/modules/dfp/load-advert.js",
+    "bootstraps/enhanced/crosswords.js",
```
regiskuckaertz: 😨  that will be fun

tools/es5to6-files-per-user.js
```diff
@@ -18,37 +18,37 @@ const filesPerHuman = (filesToCheck, humans) =>
...
-    'rich-nguyen',
-    'katebee',
-    'Calum-Campbell',
-    'ScottPainterGNM',
+    'francis.carr@guardian.co.uk',
+    'jon.norman@guardian.co.uk',
+    'regis.kuckaertz@guardian.co.uk',
+    'rich.nguyen@guardian.co.uk',
+    'kate.whalen@guardian.co.uk',
+    'calum.campbell@guardian.co.uk',
```
regiskuckaertz: @Calum-Campbell  is not part of the commercial team anymore, but @lps88 is. Also I'm not sure @ScottPainterGNM will be comfortable handling these. Scott, happy to pair with you if you want.

SiAdcock: Thanks @regiskuckaertz. Perhaps we can move Calum to misc. I'm going to send a mass email around to gather recruits!

SiAdcock:
Closing this as using `user.email` is no more reliable than `user.name`. I've sent around an email to gather the necessary data

PR #15979, opened 2017-02-28
Make sure email forms appear correctly for Sleeve Notes sign up

lmath:
We are going to run a test where new subscribers to sleeve notes get either the new
or old version of the email, and so we need to add two new ids for old/new versions of the email for these new subscribers and indicate that they should appear with a 'review' tone. 

Also, if a user unsubscribes from any Sleeve Notes email, we need to ensure that they are unsubscribed from any of the three lists (after the test, when all users are merged into one list, we can remove this unsubscribe hack). 

And so finally, the sign up form should appear correctly regardless of whether the user gets subscribed to the new or old version during the test. 

## What does this change?
There are going to be three lists of sleevenotes subscribers during the test, and this adds the two new ids for new subscribers needed for the test. It also ensures that 

Calling https://www.theguardian.com/email/form/plaintone/<id> for any of the Sleeve Notes ids provides a form that appears correctly (as below) rather than defaulting to blue. 

It's required to run an A/B test on new subscribers to Sleeve Notes. 

No.

## Screenshots
Current sign up form, which will be consistent when new subscribers are subscribed to a new list used in the test too: 
 
image

Without this change, the sign up form for the new list ids would appear like:
image

## Tested in CODE?
No.


davidfurey:
👍

PR #15980, opened 2017-02-28
Use a slot ready wait before displaying ads

rich-nguyen:
This enables the switch test to complete some async work before the googletag `display` call is made. Third parties may need to update the slot with additional targeting, this permits that to happen, whilst allowing the standard Advert lifecycle to continue with no delay.

## Tested in CODE?
Yes


static/src/javascripts-legacy/projects/commercial/modules/dfp/load-advert.js
```diff
@@ -5,8 +5,14 @@ define([
     return loadAdvert;
 
     function loadAdvert(advert) {
-        Advert.startLoading(advert);
-        window.googletag.display(advert.id);
-        dfpEnv.firstAdDisplayed = true;
+        advert.whenSlotReady
```
regiskuckaertz: Are you not renaming this to `displayAdvert`?

static/src/javascripts-legacy/projects/commercial/modules/dfp/prepare-switch-tag.js
```diff
@@ -74,19 +74,25 @@ define([
...
             var adUnitIds = findAdUnitIds(sizeMapping.size);
 
             adUnitIds.forEach(function(adUnitId) {
                 if (adUnitId){
-                    __switch_zero.units.push({
-                        dfpDivId: dfpDivId,
-                        switchAdUnitId: adUnitId
-                    });
+                    promises.push(new Promise(function(resolve){
+                        __switch_zero.units.push({
```
regiskuckaertz: I wonder if an exception will be swallowed here, should it happen? I think inside a resolver function, you have to explicitly call `reject` so that the error can be caught later on.

regiskuckaertz (approved):
Brill 💥

static/src/javascripts-legacy/projects/commercial/modules/dfp/prepare-switch-tag.js
```diff
@@ -74,19 +74,25 @@ define([
     function pushAdUnit(dfpDivId, sizeMapping) {
 
         var __switch_zero = window.__switch_zero;
+        var promises = [];
 
         if (__switch_zero) {
             var adUnitIds = findAdUnitIds(sizeMapping.size);
 
             adUnitIds.forEach(function(adUnitId) {
```
regiskuckaertz: Just an idea here: you could return a `map` of `adUnitId`s to `Promise`s objects and just return `Promise.resolve()` in the other case.

lps88 (approved):
Nice, can we merge this soon? It'll be useful for my test! 👍

static/src/javascripts-legacy/projects/commercial/modules/dfp/load-advert.js
rich-nguyen: I wanted to, but some of the metrics have been sent as `loadStart` and `loadEnd` for a while now, which made me think twice. Also, our `Advert` object has `rendering` vars, and it's not going to be obvious what the difference is between "display" and "render". The real answer lies in Googletag's implementation, where render occurs within display, and display really means "load and render".

static/src/javascripts-legacy/projects/commercial/modules/dfp/prepare-switch-tag.js
rich-nguyen: I thought an exception thrown in the executor will cause the promise to be rejected. Otherwise, every promise we write would need a try-catch and an explicit call to `reject`.

rich-nguyen:
@lps88 I love it when that happens! Hopefully will merge today.

static/src/javascripts-legacy/projects/commercial/modules/dfp/prepare-switch-tag.js
rich-nguyen: Yes, however I didn't want to burden the caller with unpacking the return type. Also, there's currently no reason for any other module to care about Switch-specific adunit ids.

PR #15981, opened 2017-02-28
Es6 script rejig

sndrs:
## What does this change?

previous version turned out to be full of bugs! this is a total rework that hopefully also offer better feedback when something goes wrong

- fewer bugs
- hopefully, better feedback when something goes wrong

### todo
- [x] update the json once a file has been copied


SiAdcock (approved):
Looks much more robust 💎

PR #15986, opened 2017-02-28
Add viewability tracking ability to Fabrics

regiskuckaertz:
All the custom template now have a `DIV` like so:

```
<div id="viewabilityTracker"><!-- [%Viewabilitytracker%] --></div>
```

If AdOps supply a viewability tracking script, it will be extracted during the `applyCreativeTemplate` phase and passed to the template. Each template then adds the script after the creative, leaving it as is or replace the string `INSERT_UNIQUE_ID` with the id of the creative ... just in case the viewability script needs that piece of data.

(I also threw up in my mouth 😭 )


regiskuckaertz:
[Fabric viewability trackers](https://trello.com/c/2h4jH9a7/223-fabric-viewability-trackers)

katebee (approved):
💎 👍

PR #15987, opened 2017-02-28
test run of module conversion to es6

sndrs:
## What does this change?

- converts `projects/common/utils/$.js` to es6
- lints unnecessary lodash array methods

see the conversion script performs, and whether es6 env is setup correctly

no

## Tested in CODE?

- [x] test on CODE


static/src/javascripts/.eslintrc.js
```diff
@@ -31,7 +31,12 @@ module.exports = {
         'no-restricted-imports': [
             'error',
             {
-                paths: ['lodash'],
+                paths: [
+                    'lodash',
+                    'lodash/collections/forEach',
+                    'lodash/collections/map',
+                    'lodash/collections/reduce',
+                ],
```
gustavpursche: What do you think about adding 

- [ ] `collections/filter`
- [ ] `collections/some`
- [ ] `objects/assign`
- [ ] `objects/values`
- [ ] `objects/merge`
- [ ] `objects/keys`
- [ ] `collections/every`
- [ ] `collections/contains`
- [ ] `objects/isArray`
- [ ] `arrays/indexOf`

This would allow the removal a huge part of lodash ...

static/src/javascripts/projects/common/utils/$.js
```diff
@@ -0,0 +1,33 @@
...
+        return el.parentNode;
+    }
+    return $.ancestor(el.parentNode, c);
+};
+
+$.forEachElement = (selector, fn) => {
+    const els = qwery(selector);
+    els.forEach(fn);
+    return els;
+};
```
gustavpursche: Wouldn't it be better that way:

```js
$.forEachElement = (selector, fn) => {
    const els = qwery(selector);
    return els.map(fn);
};
```

static/src/javascripts/.eslintrc.js
sndrs: 💯 these came off the top of my head, but totally. will update

static/src/javascripts/projects/common/utils/$.js
sndrs: ha indeed it would! so even converting the simplest modules we musn't be lazy...good one

gustavpursche: Does it return a NodeList instead?

sndrs: ooh yeah you're correct, i misread the docs, thought it returned an enhanced array-like object, but doesn't look like it all. will test it in the morning, but looks good!

desbo: simplification looks good but I'd stick with `forEach` rather than `map` – you wouldn't expect a return value from a function called `forEachElement` so there's no need to build an array.

sndrs: it did already though?

desbo: true, but none of the current uses of it do anything with the returned value. could be a good optimisation ([jsperf](https://jsperf.com/foreach-map-speed/1)) unless there's a reason i'm missing for returning the nodes

gustavpursche: I just checked, how often `forEachElement` is used. Turns out: 7 times, so it's not as critical as I thought.

1. What do you think about taking it out of `utils/$` and move it to it's own utility, to clearify, it's not part of `bonzo`?

2. I'm not very opinionated about whether to use `map` for `forEach` in this case. I'm asking myself, why it exists anywy, since there is [bonzo().each()](https://github.com/ded/bonzo#api-each).

3. 🐎 Let's not fall into a micro optimization trap for now - just make `forEachElement` work - my fault.

sndrs: it does throw up an interesting question with regards to how much we want to revisit code as we convert it. i guess that would have to be up to the individual, since it's their PR and their responsibility to deal with any errors it introduces etc. 

there's some good points in this conversation though, but i wonder if they would be better addressed as separate PRs?

gustavpursche: > there's some good points in this conversation though, but i wonder if they would be better addressed as separate PRs?

👍 

Does this work though?

```js
const els = qwery(selector);
els.forEach(fn);
return els;
```

I'd give you my approval then, so we can move on :)

SiAdcock (approved):

NataliaLKB (approved):

sndrs:
closing while we look at utils in more depth (sorry everyone!)

PR #15994, opened 2017-03-01
Relax criteria to show gu today signup box

davidfurey:
## What does this change?

Removes the requirement for the referrer to be a network to display the GU Today email signup box

So that we get more impressions for our a/b test


joelochlann (approved):
Looks good

PR #15996, opened 2017-03-01
Add test for Sleeve Notes emails

lmath:
## What does this change?
A/B test for the new Sleevenotes email

- Current subscribers will continue to receive the same version (listId=39).
- For new subscribers, the new version of Sleeve Notes is the default (listId=3834). 
- 30% of new subscribers will be in a control group subscribed to the old email (listId=3835).

Then we can compare old and new versions of the email across just new subscribers. 

Allows us to test a new version of the Sleeve Notes email.

No. 
## Tested in CODE?
No.


lmath:
@davidfurey has done some work to make it simpler to create files like sleeve-notes-email-variants.js so I'm going to try and apply that here.

static/src/javascripts-legacy/projects/common/modules/experiments/tests/sleeve-notes-email-variants.js
```diff
@@ -0,0 +1,80 @@
...
+        function enhanceWebView(emailListID) {
+            var emailForm = $('.js-email-sub__iframe')[0];
+            emailForm.setAttribute('src', 'https://www.theguardian.com/email/form/plaintone/' + emailListID);
+        }
+
+        // Runs the test on https://www.theguardian.com/email-newsletters
+        function updateNewslettersPage(emailListID) {
+
+            return fastdomPromise.write(function () {
+                var input = $('input[value=${newSleevenotesListId.toString()}]')[0];
```
joelochlann: Does this actually work!? I think you need backticks for variable substitution in a string. And even with the backticks, it's an ES6 feature, so there's no support in IE. We'll be able to use this and other ES6 features once we're bundling all our modules with webpack but we're still not quite there yet (@SiAdcock can confirm this!)

static/src/javascripts-legacy/projects/common/modules/experiments/tests/sleeve-notes-email-variants.js
```diff
@@ -0,0 +1,80 @@
...
+    return function () {
+        this.id = 'SleevenotesEmailVariants';
+        this.start = '2017-03-03';
+        this.expiry = '2017-03-31';
+        this.author = 'Leigh-Anne Mathieson';
+        this.description = 'Using the wonderful frontend AB testing framework to AB test emails, since the AB ' +
+            'function in ExactTarget re-randomises all recipients on each send, and we need users to receive their ' +
+            'variant for several weeks. This test will ensure users are added to the corresponding email list ' +
+            '(listId) in ExactTarget';
+        this.audience = 0.6;
```
joelochlann: I would start of with an audience size of 0. Then you can deploy live, opt into the test and manually test in production, then deploy a change setting the audience size to the correct amount

joelochlann (changes requested):

common/app/model/EmailNewsletters.scala
```diff
@@ -227,10 +227,10 @@ object EmailNewsletters {
     teaser = "Everything you need to know from the Guardian's music site, squeezed into one handy email",
     description = "Everything you need to know from the Guardian's music site, squeezed into one handy email.",
     frequency = "Every Friday",
-    listId = 39,
-    aliases = List(3834, 3835),
+    listId = 3834,
```
davidfurey: It might make sense to switch the primary listId based on the AB test switch, so that the new list ID is only enabled when the test is enabled.

davidfurey (commented):
Seems fine, but is it not possible to share some of the code sleeve-notes-email-variants.js across other email a/b tests?

static/src/javascripts-legacy/projects/common/modules/experiments/tests/sleeve-notes-email-variants.js
lmath: Ok, noted. Thanks.

static/src/javascripts-legacy/projects/common/modules/experiments/tests/email-variant-example.js
```diff
@@ -0,0 +1,22 @@
+define([
```
davidfurey: I think you can delete this file

lmath: Oh, right. Agreed. I will do that.

static/src/javascripts-legacy/projects/common/modules/experiments/tests/generic-email-variants.js
```diff
@@ -0,0 +1,68 @@
...
+], function (
+    bean,
+    fastdom,
+    qwery,
+    $,
+    config,
+    fastdomPromise
+) {
+    return function (props) {
+        this.id = props['id'];
```
joelochlann: Using square bracket notation with string literals for accessing object properties is stylistically a bit weird (https://github.com/airbnb/javascript#properties) - best to stick with dot notation unless you're using a variable as the key

joelochlann (approved):

PR #15997, opened 2017-03-01
Expiring switches email changes

jfsoul:
## What does this change?
We have some concerns about the clarity/usefulness of the expiring switches email.  This came up after @katebee kindly amended the AB test expiry time to match the switch expiry time (now both 23:59 GMT).  This PR includes a few small tweaks to the email that we decided might improve it:

- Send the email once more in the afternoon to the dotcom platform team only.  This way we can see if there are any switches we need to follow up about.
- Make it more explicit that the switch expires at 23:59
- Include a short note about AB test expiry:

> Note on AB tests: AB tests do not necessarily expire at the same time as their switch - they may have been configured to expire earlier.  Please do not rely only on this email to determine when your test will end.

I will also amend the AB test docs and the name of the `expiry` property in another PR.

Less chance of confusion about expiry of tests and switches - easier to follow up in the afternoon on switches that haven't been addressed.

No

## Screenshots
(Updated AB test message slightly after taking screenshot)
image

## Tested in CODE?
No


admin/app/model/AdminLifecycle.scala
```diff
@@ -89,13 +89,15 @@ class AdminLifecycle(appLifecycle: ApplicationLifecycle,
       matchDayRecorder.record()
     }
 
-    if (environment.isProd) {
```
jfsoul: This check was unnecessary since the email job itself relies on an `email: Option[String]` which is only configured in `PROD`

admin/app/model/AdminLifecycle.scala
```diff
@@ -89,13 +89,15 @@ class AdminLifecycle(appLifecycle: ApplicationLifecycle,
...
-      val londonTime = TimeZone.getTimeZone("Europe/London")
-      jobs.scheduleWeekdayJob("ExpiringSwitchesEmailJob", 48, 8, londonTime) {
-        log.info("Starting ExpiringSwitchesEmailJob")
-        ExpiringSwitchesEmailJob(emailService).run()
-      }
+    val londonTime = TimeZone.getTimeZone("Europe/London")
```
DiegoVazquezNanini: @jfsoul Do you know why we force the admin to use London timezone?

jfsoul: Not exactly - I suppose so we didn't rely on the `java.util.TimeZone.getDefault()`.

`./sbt` also explicitly sets the timezone to Sydney so helpful in `DEV` too...

admin/app/model/AdminLifecycle.scala
```diff
@@ -89,13 +89,15 @@ class AdminLifecycle(appLifecycle: ApplicationLifecycle,
...
+    jobs.scheduleWeekdayJob("ExpiringSwitchesEmailJob", 48, 8, londonTime) {
```
NataliaLKB: Why 8:48? Such a random time!

NataliaLKB (approved):
Thanks for doing this Jon!

PR #15999, opened 2017-03-01
ES6ify defer-to-analytics.js

SiAdcock:
## What does this change?

Converts the defer-to-analytics module to ES6

#thatwaseasy

Part of the gradual migration towards ES6

No

## Tested in CODE?

No


static/src/javascripts/projects/common/utils/defer-to-analytics.js
```diff
@@ -0,0 +1,19 @@
...
+    if (analyticsReady) {
+        afterAnalytics();
+    } else {
+        mediator.on('analytics:ready', () => {
+            afterAnalytics();
+        });
+    }
+}
+
+export default deferToAnalytics; // define
```
sndrs: don't think we need that comment?

sndrs (approved):

PR #16002, opened 2017-03-01
Sudoku: remove orphaned JS application

gu-stav:
## What does this change?

Removes the orphaned Sudoku JS application. There is still a lot Scala code left, which we're going to check and remove in a separate PR.

Less code to maintain and transpile.

No.


DiegoVazquezNanini (commented):
LGTM

SiAdcock (approved):
🔥

NataliaLKB (approved):
Haha so funny that we have this. Definitely make sure to remove the Scala code too! 👍 💯

PR #16005, opened 2017-03-01
ES6: transform common/modules/ui/autoupdate.js into ES6

gu-stav:
## What does this change?

Transforms `projects/common/modules/ui/autoupdate.js` into an ES6 module

ES6!

No.


SiAdcock (approved):
LGTM

PR #16006, opened 2017-03-01
Move article email signup to post-article position

joelochlann:
## What does this change?
Moves the email signup box which appears in articles from the end of the article to after the article, and restyles it so it looks more like related content.

Using tailor, we're about to start inserting email signup boxes which are different to the main article, e.g. sign up to Film Today at the bottom of a news article, based on the user's reading history. This looks jarring with the existing style and we hope that this restyle will make it look less confusing

No

## Screenshots
Will need @superfrank to do a design review and set the colours correctly for all tones other than `media` (which I've set myself)

cc @lindseydew 

### Before
picture 469

### Desktop
picture 465
### Tablet
picture 466
### Wide mobile
picture 467
### Mobile
picture 468

## Tested in CODE?
No


davidfurey (approved):
Looks fine to me.  Have you checked that Iona is okay with the old headlines not being displayed?

lindseydew:
Looks AWESOME 👍 Thank you so much @joelochlann

superfrank:
Absolutely nailed it from the looks of it. I'll jump in and make some changes to the background colours for the other emails, but then all good to go from a design perspective 👍

static/src/javascripts-legacy/projects/common/modules/email/email-article.js
```diff
@@ -127,6 +159,10 @@ define([
                 listId: '38',
                 listName: 'greenLight',
                 campaignCode: 'green_light_article_bottom',
+                displayName: {
+                    normalText: 'film',
```
lindseydew: Ooh just noticed the display name isn't the right one in this case

joelochlann: :eye: Good spot!

joelochlann:
Thanks @superfrank for some super colours. Here they are in all their glory:

screencapture-localhost-3000-science-2017-mar-06-how-a-game-of-draughts-helped-unravel-the-brains-reward-system-2017-brain-prize-1488820659028

PR #16008, opened 2017-03-01
add `--has-items` class to end-slate fragment

akash1810:
## What does this change?
Currently if there are no items available in the endslate (e.g. https://api.nextgen.guardianapps.co.uk/video/end-slate/section/beyond-the-lights.json?shortUrl=https://gu.com/p/62k36) we're still displaying an empty `div` which looks strange (e.g. https://www.theguardian.com/beyond-the-lights/video/2017/feb/28/las-vegas-nature-parks-trails-desert-canyon).

This change will add the class `end-slate-container--has-items` to the end-slate `div` if there are items and we can then `display: block` if we have the class else we `display: none`.

Ideally we'd just set [`data-show-end-slate`](https://github.com/guardian/frontend/blob/master/common/app/views/fragments/media/video.scala.html#L42) to `false` if there are no items in the end-slate but alas always set it to `true`. Not entirely sure why, I think its because it relies on a (blocking) CAPI request so we set it to `true` and make the request client-side? Hiding the end-slate this way is cheaper than answering this question.

Better UX.

No.

## Screenshots
Before
image

After
image

### DIFF
--- a/onward/app/views/fragments/videoEndSlate.scala.html
+++ b/onward/app/views/fragments/videoEndSlate.scala.html
@@ -1,9 +1,15 @@
 @(videos: Seq[model.Video], componentId: String, heading: String)(implicit request: RequestHeader)
 
 @import views.support.Seq2zipWithRowInfo
+@import views.support.RenderClasses
 
-
-<div class="end-slate-container end-slate-container--video" data-component="video-end-slate-@componentId">
+<div class="@RenderClasses(
+    Map(
+        "end-slate-container--has-items" -> videos.nonEmpty
+    ),
+    "end-slate-container",
+    "end-slate-container--video"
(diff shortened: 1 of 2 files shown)


akash1810:
WIPing because I'm having problems running frontend locally (`http://localhost:9000/assets/javascripts/app-webpack.js ` returns a 500). Will chat with @sndrs on how to fix this as I've nuked everything but no luck so far.

kelvin-chappell (approved):
Looks good to me.  Thanks for fixing!

gidsg:
@akash1810 you could deploy to CODE to test this if that's easier.

akash1810:
@gidsg removing all target folders works! Thanks.

Fully tested in DEV now. Merging.

PR #16009, opened 2017-03-02
Enable jest on main js tests

sndrs:
## What does this change?

- runs `static/src/javascripts/**/*.spec.js` tests with jest (i.e. not the current tests, only new ones added/ported over)

- get jest up and running
- be able to add tests for converted modules


tools/__tasks__/compile/javascript/babel.js
```diff
@@ -7,7 +7,13 @@ module.exports = {
     task: () =>
         execa(
             'babel',
-            [`${src}/javascripts`, '--out-dir', `${transpiled}/javascripts`],
+            [
+                `${src}/javascripts`,
+                '--out-dir',
+                `${transpiled}/javascripts`,
+                '--ignore',
+                '**/*.spec.js',
```
gtrufitt: Does this work ok? I was adding the same thing in my branch yesterday and my `.spec.js` files still ended up in there. Just tested it on this branch and get the same thing:

- Added `test.spec.js` to javascripts > projects > common > modules > commercial
- Ran `make test` and my jest test passes happily enough but then in the legacy test run I got:

```js
SyntaxError: Unexpected token ')'
  at static/transpiled/javascripts/projects/common/modules/test.spec.js:3
```

Is this babel still whacking it over or is there a copy task somewhere?

gtrufitt:
Added the basic setup file for if a JS file is looking for `window.guardian.config` but LGTM!

gtrufitt (approved):

tools/__tasks__/compile/javascript/babel.js
sndrs: hmmm, i didn't think it was – will double check, ta

sndrs: it's working for me – i've added a fake test to make sure it's not a weird env thing, want to pull latest and try? will see how TC goes too

sndrs: ah got it - i didn't try in the location you suggested, will fix

gtrufitt: Doh, yeah, sorry, just pushed one nested :D

sndrs: > is there a copy task somewhere

bingo – added 2nd nested spec, fixed the copy ignore glob

sndrs: oops, we're lapping each other :)

.eslintrc.js
```diff
@@ -4,6 +4,7 @@ module.exports = {
         commonjs: true,
         es6: true,
         node: true,
+        jest: true,
```
SiAdcock: Is this only needed in https://github.com/guardian/frontend/blob/master/static/src/javascripts/.eslintrc.js?

sndrs: it'll be needed wherever we have tests – i guess this is sort of prefiguring a world where all our tests sit alongside the module they test i.e. not just in the app?

PR #16010, opened 2017-03-02
update babel env settings

sndrs:
## What does this change?

- removes AMD output for normal transpilation, since webpack 2 understands es6 modules
- adds a temporary `karma` env for the transpilation step required for legacy karma tests

this is how we'd want it, once everything we don't want has gone

### todo

have a think about what envs we want to build for in prod. but that's for another time

## Check on code
- [x] checked on CODE


gtrufitt (approved):
👍

PR #16001, opened 2017-03-01
New recipe design

tsop14:
## What does this change?
This PR adds logic to display articles-with-structure-recipe-data using a _beautiful_ new design but does **not** implement it; currently `&& false` is included in relevant if statements. So, it changes nothing from a user's perspective. This is so the code can be reviewed and a subsequent PR will turn on the feature for a given number of users as part of an AB test. 

- Adds view fragments `recipeArticleBody.scala.html` and `recipeArticle.scala.html`.
- Sets `isRecipeArticle` to true if an article has associated recipe atoms, i.e. has structured recipe data.
- Displays article with new view if `isRecipeArticle` is true.
- Uses slim nav bar if `isRecipeArticle` is true.

The new view is based on @blongden73's design. 
- Large image on LHS
- Original article (hidden after the first few paragraphs)
- Each recipe atom. 

Adds javascript to skip between recipes and update the image on the LHS to match the recipe in view. Without js enabled the main image remains in view.

## What is the value of this and can you measure success?
To create more engaging recipes using structured data. An AB test will be run to compare user interaction with old v new design. It will also allow users to click a recipe from the new recipe fronts and be taken directly to that recipe (rather than simply the top of the article which it is included in).

## Does this affect other platforms - Amp, Apps, etc?
No.

## Screenshots
desktop:
recipe
mobile:
mobile

moving from article with structured recipes to article without:
only-structured-recipes

## Tested in CODE?
No

PR #16012, opened 2017-03-02
Removes default NewsArticle schema type on AMP articles

tsop14:
## What does this change?
Google now [supports schema types](https://github.com/ampproject/amphtml/tree/master/examples/metadata-examples) in their amp carousel; it no longer needs to default to 'NewsArticle'. https://www.ampproject.org/docs/guides/discovery

More accurate description of content. 

It only effects AMP. 

## Tested in CODE?
No

Spotted by @NataliaLKB on separate [PR](https://github.com/guardian/frontend/pull/16001) which copied this code.

### DIFF
--- a/article/app/views/fragments/articleBody.scala.html
+++ b/article/app/views/fragments/articleBody.scala.html
@@ -1,4 +1,3 @@
-@import model.ArticleSchemas
 @(model: ArticlePage, amp: Boolean = false)(implicit request: RequestHeader, context: _root_.model.ApplicationContext)
 
 @import common.LinkTo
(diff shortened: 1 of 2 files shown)


NataliaLKB (approved):
Thanks so much for doing this @tsop14 !!! 💯 

Before merging, would you be able to check that a recipe (no atom obviously) and a review page validate on AMP?

Just to make sure 100% that this is okay.

tsop14:
> Before merging, would you be able to check that a recipe (no atom obviously) and a review page validate on AMP?

An article with recipes (with or without recipe atoms) is type `NewsArticle` so will be the same. 
A review page passes validation on https://ampbench.appspot.com but all `make validate-amp` tests are failing (on master and this branch).

@NataliaLKB

NataliaLKB:
@tsop14 great! Thanks for testing. 

If you look into our [AMP docs](https://github.com/guardian/frontend/blob/master/docs/03-dev-howtos/17-working-with-amp.md) at the bottom we have this:
image

[Here is the file](https://github.com/guardian/frontend/blob/master/common/app/views/fragments/amp/customStyles.scala.html#L31)

I haven't gotten around to doing what I would like to do with this. I want to make sure that if we run the tests, it ignores this particular error. It is on my list, but until it is implemented best to just do as the comment suggests and comment out the link tag while you run the tests.

Sorry about that 😞

NataliaLKB:
Great work though! Definitely good to merge

PR #16013, opened 2017-03-02
removed white outline

superfrank:
## What does this change?
Removes a white outline from the CTA
## Screenshots
This is what it currently looks like:
image

## Tested in CODE?

@joelochlann


katebee (approved):
👍

PR #16015, opened 2017-03-02
Media Atom Refactor: consolidate parameters into enum

gidsg:
## What does this change?
Consolidates three  presentational boolean parameters which cannot co-occur:
* embedPage
* mainMedia
* immersiveMainMedia

into an enum `MediaWrapper`

Code should be cleaner to read as fewer parameters are passed to media template and by extension in places they are called such as Cleaners and templates

This is a refactor. So existing behaviour on AMP should continue.

## Screenshots
N/A

## Tested in CODE?
No

### DIFF
--- a/common/app/model/content/Atoms.scala
+++ b/common/app/model/content/Atoms.scala
@@ -9,6 +9,7 @@ import org.joda.time.{DateTime, DateTimeZone, Duration}
 import play.api.libs.json.{JsError, JsSuccess, Json}
 import quiz._
 import enumeratum._
+import model.content.MediaAssetPlatform.findValues
 
 final case class Atoms(
   quizzes: Seq[Quiz],
@@ -51,6 +52,16 @@ object MediaAssetPlatform extends Enum[MediaAssetPlatform] with PlayJsonEnum[Med
   case object Dailymotion extends MediaAssetPlatform
   case object Mainstream extends MediaAssetPlatform
   case object Url extends MediaAssetPlatform
+}
+
+sealed trait MediaWrapper extends EnumEntry
+
+object MediaWrapper extends Enum[MediaWrapper] with PlayJsonEnum[MediaWrapper] {
+  val values = findValues
+
+  case object MainMedia extends MediaWrapper
+  case object ImmersiveMainMedia extends MediaWrapper
+  case object Embed extends MediaWrapper
 }
 
 final case class MediaAsset(
--- a/common/app/model/content/Atoms.scala
+++ b/common/app/model/content/Atoms.scala
@@ -61,7 +61,7 @@ object MediaWrapper extends Enum[MediaWrapper] with PlayJsonEnum[MediaWrapper] {
 
   case object MainMedia extends MediaWrapper
   case object ImmersiveMainMedia extends MediaWrapper
-  case object Embed extends MediaWrapper
+  case object EmbedPage extends MediaWrapper
 }
 
 final case class MediaAsset(
--- a/common/app/views/fragments/atoms/ampYoutube.scala.html
+++ b/common/app/views/fragments/atoms/ampYoutube.scala.html
@@ -1,4 +1,5 @@
-@(media: model.content.MediaAtom,  displayCaption: Boolean, mainMedia: Boolean)(implicit request: RequestHeader)
+@import model.content.MediaWrapper
+@(media: model.content.MediaAtom,  displayCaption: Boolean, mediaWrapper: Option[MediaWrapper])(implicit request: RequestHeader)
 @import views.html.fragments.atoms.mediaAtomCaption
     <amp-youtube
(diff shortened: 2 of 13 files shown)


GHaberis (commented):
I feel like this makes sense, but it's pushing my comprehension of Scala, It probably needs someone more Scala orientated to stamp it! @TBonnin would you mind taking a look?

akash1810 (approved):
looks good

common/app/views/fragments/atoms/ampYoutube.scala.html
```diff
@@ -1,4 +1,5 @@
-@(media: model.content.MediaAtom,  displayCaption: Boolean, mainMedia: Boolean)(implicit request: RequestHeader)
+@import model.content.MediaWrapper
+@(media: model.content.MediaAtom,  displayCaption: Boolean, mediaWrapper: Option[MediaWrapper])(implicit request: RequestHeader)
```
TBonnin: We both use full class path and import in this 2 lines? (ie: `model.content.MediaAtom` and `import model.content.MediaWrapper`)
Can we stick to a single one? I vote for the full class style but it is up to you

common/app/views/fragments/atoms/atom.scala.html
```diff
@@ -1,11 +1,12 @@
-@(model: _root_.model.content.Atom, shouldFence: Boolean, amp: Boolean, mainMedia: Boolean, immersiveMainMedia: Boolean)(implicit request: RequestHeader, context: _root_.model.ApplicationContext)
+@import model.content.MediaWrapper
+@(model: _root_.model.content.Atom, shouldFence: Boolean, amp: Boolean, mediaWrapper: Option[MediaWrapper])(implicit request: RequestHeader, context: _root_.model.ApplicationContext)
```
TBonnin: Same comment about `import` vs using full class path

common/app/views/fragments/atoms/media.scala.html
```diff
@@ -1,11 +1,14 @@
 @import model.content.MediaAssetPlatform
-@(media: _root_.model.content.MediaAtom, amp: Boolean = false, displayCaption: Boolean, embedPage: Boolean = false, displayEndSlate: Boolean = false, mainMedia: Boolean = false, immersiveMainMedia: Boolean = false)(implicit request: RequestHeader, context: model.ApplicationContext)
+@import model.content.MediaWrapper
+@(media: _root_.model.content.MediaAtom, amp: Boolean = false, displayCaption: Boolean, displayEndSlate: Boolean = false, mediaWrapper: Option[MediaWrapper])(implicit request: RequestHeader, context: model.ApplicationContext)
```
TBonnin: Same here even though it was already importing `MediaAssetPlatform`

common/app/views/fragments/atoms/ampYoutube.scala.html
TBonnin: Looking at other files we are far from being consistent so feel free to ignore my comment I guess

common/app/views/fragments/atoms/youtube.scala.html
```diff
@@ -86,6 +88,6 @@
         </div>
 }
 
-@if(displayCaption & !immersiveMainMedia) {
-    @mediaAtomCaption(media.title, mainMedia)
+@if(displayCaption & !mediaWrapper.contains(ImmersiveMainMedia)) {
```
TBonnin: I know it was already here but I curious about this single `&`

common/app/model/content/Atoms.scala
```diff
@@ -53,6 +54,16 @@ object MediaAssetPlatform extends Enum[MediaAssetPlatform] with PlayJsonEnum[Med
...
 }
 
+sealed trait MediaWrapper extends EnumEntry
+
+object MediaWrapper extends Enum[MediaWrapper] with PlayJsonEnum[MediaWrapper] {
+  val values = findValues
+
+  case object MainMedia extends MediaWrapper
+  case object ImmersiveMainMedia extends MediaWrapper
+  case object EmbedPage extends MediaWrapper
```
TBonnin: Would it make sense to have an additional `None` value (ex: NoMediaWrapper). So you don't have to carry Option[MediaWrapper] all over the place, as well as simplifying comparison?

TBonnin:
@gidsg looks like I am a bit late 😢

gidsg:
Thanks @TBonnin I can pick this feedback up in a follow-up PR.

common/app/model/content/Atoms.scala
gidsg: Unfortunately, as the fronts are burned in on S3 it is not easy to add mandatory fields to models that are used by Facia.

TBonnin: Got it. Forgot about that

PR #16016, opened 2017-03-02
Move `utils` to `lib`

SiAdcock:
## What does this change?

Moves all of the JavaScript utils into a new `lib` directory at the root of `javascripts` and `javascripts-legacy`. 

This is the first change as a result of an audit of our directory structure. Burying these useful libraries under `projects/common/utils` made them harder to find. It was also unclear where they were supposed to be used. Moving them up to the root of the `javascripts` folders will emphasise the intention that they may be used anywhere.

The contents of the lib directory will be the first to receive Flow type annotations.

No

## Tested in CODE?

- [ ] Test in CODE


gu-stav (approved):
🐳 💨 LGTM.

gustavpursche:
@SiAdcock You might want to update the ES6 responsibility list as part of this PR aswell?

sndrs (approved):
♻️

PR #16017, opened 2017-03-02
New nav - newsletters? I don't think so. 

zeftilldeath:
Newsletters has been removed from the expanded menu of the new nav. Cos @stephanfowler told me to.


NataliaLKB (approved):
🌟

joelochlann:
😭  Why don't we want the newsletters?

NataliaLKB:
@joelochlann from what I understand its coming back, but it came from Monica so she should know why.

PR #16020, opened 2017-03-02
Tailor email test

lindseydew:
## What does this change?
Adds a test to see whether presenting email lists based on their browsing habits has a higher sign up rate than email lists based on the page. Depends on https://github.com/guardian/frontend/pull/16006

Experiment into whether we can curate a user experience based on their history. Success can be measured by whether there are more signups in the case of tailor recommendations

No
## Screenshots
No visual changes
## Tested in CODE?
No


joelochlann:
👍  Looks good! #16006 is merged so should be good to ship

PR #16021, opened 2017-03-02
Wait for commercial end

regiskuckaertz:
The `enhanced` bootstrap is loaded in parallel with the sonobi and google libraries:

picture 10

That's because all the commercial stuff is done via promises. The commercial bootstrap returns a promise that resolves when all the commercial modules have been initiated, which takes half a second on average:

untitled

All the heavy work is pushed into the google command queue which is itself asynchronous (at least until GPT has been evaluated). In this PR, I'm just waiting on the global promise, it gives the commercial stuff some room to breathe -- not too much though, because we never wait on anything.

### DIFF
--- a/static/src/javascripts-legacy/bootstraps/commercial.js
+++ b/static/src/javascripts-legacy/bootstraps/commercial.js
@@ -149,6 +149,10 @@ define([
                     // There are no MPUs on hosted pages, so no slot render events, and therefore no reporting would be done.
                     Promise.all(customTimingModules).then(performanceLogging.reportTrackingData);
                 }
+            })
+            .catch(function () {
+                // Just in case something goes wrong, we don't want it to
+                // prevent enhanced from loading
             });
         }
     };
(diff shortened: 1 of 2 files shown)


SiAdcock (approved):
I believe that the intention behind this is to ensure that `commercial` is initialised before the request for `enhanced` is made, because `commercial` ought to be prioritised ahead of `enhanced`.

I am generally on board with this. Without a clear strategy for prioritising JavaScript execution, I'm unsure what the implications will be for introducing or inadvertently reducing jank / yank as different resources compete for use of the main thread. It's probably a bigger question for discussion outside this PR.

regiskuckaertz:
It is hard to reason about network calls prioritization without a good model of our runtime. Blocking on the promise is a way of controlling back-pressure here, it should give a slight edge on the ads loading process.

static/src/javascripts-legacy/bootstraps/commercial.js
```diff
@@ -149,6 +149,10 @@ define([
                     // There are no MPUs on hosted pages, so no slot render events, and therefore no reporting would be done.
                     Promise.all(customTimingModules).then(performanceLogging.reportTrackingData);
                 }
+            })
+            .catch(function () {
+                // Just in case something goes wrong, we don't want it to
```
rich-nguyen: `report-error`?

regiskuckaertz: oh yeah, great idea!

PR #16024, opened 2017-03-02
ES6: Make conversion script less error prone

gu-stav:
## What does this change?

- The conversion branch is always created from `master`
- Only the ES5 and ES6 files are commited - nothing else, which allows to have unstaged files.

Easier conversion.

### DIFF
--- a/tools/es5to6.js
+++ b/tools/es5to6.js
@@ -1,5 +1,6 @@
 const path = require('path');
 
+const git = require('simple-git')();
 const gitUser = require('git-user-name');
 const chalk = require('chalk');
 const execa = require('execa');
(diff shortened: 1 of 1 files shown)


tools/es5to6.js
```diff
@@ -17,91 +16,70 @@ process.on('uncaughtException', handleError);
...
-                        })
-                        .then(() => {
-                            console.log(chalk.green('done'));
-                        })
-                        .catch(e => {
-                            console.log(chalk.red(e.stack));
-                            megalog.error(
-                                `\`${step}\` did not complete.
+const steps = {
+    'Create a branch for the conversion': `git checkout -b "${branchName}" master || git checkout -b "${branchName}-${unique}" master`,
```
sndrs: will this get origin? local master could be out of date?

sndrs: does this mean someone could end up with a lot of unused branches after running this a few times?

tools/es5to6.js
```diff
@@ -17,91 +16,70 @@ process.on('uncaughtException', handleError);
...
+    'Move the legacy module to the new location': `mkdir -p ${es6Module}; mv ${es5Module} $_; node ./tools/es5to6-remove-module.js ${moduleId}`,
+    'Commit the move': `git add ${es6Module} ${es5Module}; git commit -m "move ${moduleId} from legacy to standard JS"`,
```
sndrs: should this include the json too?

tools/es5to6.js
```diff
@@ -17,91 +16,70 @@ process.on('uncaughtException', handleError);
...
 if (!userModules || !userModules.length) {
     console.log(chalk.green('⭐️  You have no more modules to convert! ⭐️'));
     process.exit();
 }
 
-git
-    .status((err, status) => {
-        if (
-            status.current !== 'master' ||
-            status.files.length > 0 ||
```
sndrs: without this check, what happens when we switch the user's branch and they have unstaged changes that conflict with master?

gustavpursche: You are right - I'll put the checks back. It's more annoying but much safer.

tools/es5to6.js
gustavpursche: No it doesn't. Should we do a `git fetch` as part of this first step?

gustavpursche: Yes, that would happen. Do you think that is a problem? Doing it the right way (checking for branch and print error message in case it's a duplicate) would be a bit of work, which I want to avoid ...

tools/es5to6.js
gustavpursche: 👍

PR #16025, opened 2017-03-02
corrects url field in review data

tsop14:
## What does this change?
Corrects bug in film review structured data, replaces string interpolation with magic @ character. 

Now: `"url" : "https://www.theguardian.com/music/2016/aug/25/gary-numan-android-in-la-la-land-review-electronic-music-pioneer-in-sparky-doc"`

Before: 
`"url" : s"https://www.theguardian.com/${articleId}"`

Correct data. Fixes one error on google data validation: 

screen shot 2017-03-02 at 15 38 00

It might make the articles appear on AMP and search results on mobile. Although there are other validation errors, the same errors appear on articles e.g. https://www.theguardian.com/film/2010/feb/11/takeshis-film-review which do show up.

## Tested in CODE?
Only locally. 

NB Do not think will fix missing stars on google search, which could be caused by a missing `itemReview` within the article - which should be created by [this code](https://github.com/guardian/frontend/blob/master/common/app/views/fragments/headTonal.scala.html#L71).

screen shot 2017-03-02 at 15 54 31

@LATaylor-guardian

### DIFF
--- a/common/app/views/fragments/atoms/structureddata/filmReview.scala.html
+++ b/common/app/views/fragments/atoms/structureddata/filmReview.scala.html
@@ -12,7 +12,7 @@
   "@@context":"http://schema.org",
   "@@type":"Review",
   "author": @fragments.atoms.structureddata.contributor(review.data.reviewer, contributors),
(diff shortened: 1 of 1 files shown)


LATaylor-guardian:
👍

PR #16028, opened 2017-03-03
add preload attribute to none for audio tag

akash1810:
prevents audio source from downloading on page load, instead it'll load on play i.e. after showing intent.

NB we set this attribute to `metadata` in [videojs](https://github.com/guardian/frontend/blob/master/static/src/javascripts-legacy/projects/common/modules/video/videojs-options.js#L21) however that only takes effect once the js has executed. `metadata` [will fetch the audio length](https://developer.mozilla.org/en/docs/Web/HTML/Element/audio).

Smaller page on load.

No.

## Screenshots
Before (page load, filtering network requests to `.mp3`)
image

After (page load, filtering network requests to `.mp3`)
image

## Tested in CODE?
No.


gtrufitt (approved):
💯

PR #16031, opened 2017-03-03
Npm: update `sass-mq` to a valid version

gu-stav:
## What does this change?

Updates `sass-mq` dependency. The listed version `3.0.0` never was published on npm and so throw an error, when running certain npm/ yarn commands.

According to the [changelog](https://github.com/sass-mq/sass-mq/releases) there shouldn't be any changes at all.

Edit: Turns out, the linter didn't run since end of January. I'll raise another issue, that we need a check, if our linters are actually running during the validate step :)

Proper toolchain.

No.


gustavpursche:
@sndrs Adding you as reviewer, because you initialised the `package.json` change for `sass-mq` - just in case it was intentional.

SiAdcock (approved):
Looks good, although `yarn.lock` seems to have gone a bit mental and updated hundreds of dependencies! Otherwise 👌

gustavpursche:
@SiAdcock Thanks for pointing this out. Should be fixed now - could you have a look again, if this looks right to you?

SiAdcock:
Much healthier! @gustavpursche 💎 👍 💎

sndrs:
ha – nice one!

PR #16035, opened 2017-03-03
Adding Flowtype to the new world

gtrufitt:
## What does this change?

Makes flowtype checking available to the ES6 JS.  

Typing our JS, in particular our libs/utils, ensures a higher level of safety and documentation that we didn't previously have.

## Tested in CODE?
Not yet.

- [x] Need a global promise so we can use webpack imports as current require fails flow
- [x] Ensure building works correctly
- [x] Add flowTyped libraries? Maybe a different PR


SiAdcock:
Regarding the first checkbox, we will also need to define a type for `require.ensure` [(like this one)](https://github.com/facebook/flow/issues/3217), which is used to specify split points.

gtrufitt:
> Regarding the first checkbox, we will also need to define a type for require.ensure (like this one), which is used to specify split points.

[Yup](https://github.com/guardian/frontend/pull/16035/commits/52f131131f34b26f10bd127953cebcda55c035e5) 👍

SiAdcock:
I wonder what the advantage of using `import` would be, over Webpack's `require.ensure`. Both define split points, but `require.ensure` also lets us name the split point.

Once we start using `require.ensure`, it would be worth defining a [`libdefs` folder](https://flowtype.org/docs/third-party.html#writing-custom-library-definition-files), to define Flow type declarations for third party libraries that have not been added to `flow-typed`.

gtrufitt:
> It would be worth defining a libdefs folder, to define Flow type declarations for third party libraries that have not been added to flow-typed.

Yup, agreed, I added `flow-typed` in a [different branch](https://github.com/guardian/frontend/tree/Platform_add-flow-typed) to avoid making this one huge, so I can move those declarations into the `flow-typed` folder as part of that branch? Or maybe I should create the folder in this branch and move them here...

SiAdcock:
I think it would be better to keep `flow-typed` definitions separate from definitions we have written ourselves manually, so we know where the definitions came from. So maybe this branch?

static/src/javascripts/lib/defer-to-analytics.js
```diff
@@ -6,7 +7,7 @@ mediator.on('analytics:ready', () => {
     analyticsReady = true;
 });
 
-function deferToAnalytics(afterAnalytics) {
+function deferToAnalytics(afterAnalytics: () => void) {
```
SiAdcock: I'm not sure how far we're intending to go with the Flow, but should we specify a return type of `void` for the `deferToAnalytics` function itself?

sndrs (approved):
type A+

sndrs:
sorry mate #23382 

can't wait to see what replaces typescript...

PR #16036, opened 2017-03-03
Include lodash find in check-dispatcher

SiAdcock:
## What does this change?

It looks like the lodash `find` method was not included in `check-dispatcher` when it was created in #15802. The behaviour was reverting to the non-standard [`window.find`](https://developer.mozilla.org/en-US/docs/Web/API/Window/find) method, which does something completely different to lodash's `find` and doesn't exist in some browsers, throwing a `ReferenceError`.

Check now does the correct thing. 

No

## Tested in CODE?

No

### DIFF
--- a/static/src/javascripts-legacy/lib/check-dispatcher.js
+++ b/static/src/javascripts-legacy/lib/check-dispatcher.js
@@ -1,10 +1,12 @@
 define([
+    'lodash/collections/find',
     'lib/config',
     'lib/check-mediator',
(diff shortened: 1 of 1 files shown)


sndrs:
can we regex it, avoiding the dependency?

sndrs:
sorry, misunderstood this completely!

can we use `indexOf` to avoid it?

sndrs (approved):
deleted a load of comments, changed my mind entirely :)

do it

PR #16039, opened 2017-03-03
Remove old participation work

gtrufitt:
## What does this change?

Removes some old files that aren't used.

Tidy tidy.


SiAdcock (approved):
🎉

PR #16040, opened 2017-03-03
Remove dead drop-while utility

gu-stav:
## What does this change?

Removes dead code from `lib/`.

Less code.


SiAdcock (approved):
👋

PR #16041, opened 2017-03-03
Add commercial endpoints for getting Preview Urls for creatives

JonNorman:
### Background
We are developing a small tool for the Ad Ops team that will ultimately provide a way of screen grabbing all ads from a given campaign as they appear on various sections of theguardian.com. To force a creative to show on a given page a Preview URL can be generated by DFP.

### What does this PR do?
This PR creates a couple of "ad grabber" API routes that can be used to get:

a) All the line items for a given DFP order
b) A list of preview URLs for each creative in a line item, provided a section of the website

This leverages all of the caching and fetching that is set up in frontend/admin.

@guardian/commercial-dev

### DIFF
--- a/admin/app/controllers/admin/CommercialController.scala
+++ b/admin/app/controllers/admin/CommercialController.scala
@@ -111,9 +111,9 @@ class CommercialController(implicit context: ApplicationContext) extends Control
     val previewUrls: Seq[String] =
       (for {
         lineItemId <- Try(lineitemId.toLong).toOption
-        section <- validSections.find(_ == section)
+        validSection <- validSections.find(_ == section)
       } yield {
-          DfpApi.getCreativeIds(lineItemId) flatMap (DfpApi.getPreviewUrl(lineItemId, _, s"https://theguardian.com/$section"))
+          DfpApi.getCreativeIds(lineItemId) flatMap (DfpApi.getPreviewUrl(lineItemId, _, s"https://theguardian.com/$validSection"))
       }) getOrElse Nil
 
     Cached(5.minutes) {
--- a/admin/app/controllers/admin/CommercialController.scala
+++ b/admin/app/controllers/admin/CommercialController.scala
@@ -1,14 +1,17 @@
 package controllers.admin
 
 import common.dfp.{GuCreativeTemplate, GuLineItem}
-import common.{ExecutionContexts, Logging}
+import common.{ExecutionContexts, JsonComponent, Logging}
 import conf.Configuration
-import dfp.{OrderAgent, CreativeTemplateAgent, DfpApi, DfpDataExtractor, AdvertiserAgent}
+import dfp.{AdvertiserAgent, CreativeTemplateAgent, DfpApi, DfpDataExtractor, OrderAgent}
 import model._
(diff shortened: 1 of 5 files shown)


JonNorman:
[Ad Grabber Backend Functionality](https://trello.com/c/NI2a1JjD/236-ad-grabber-backend-functionality)

admin/app/controllers/admin/CommercialController.scala
```diff
@@ -74,13 +77,14 @@ class CommercialController(implicit context: ApplicationContext) extends Control
 
     val (hasNumericTestValue, hasStringTestValue) =
       lineItemsByAdTest partition { case (testValue, _) =>
-      def isNumber(s: String) = s forall Character.isDigit
+        def isNumber(s: String) = s forall Character.isDigit
```
kelvin-chappell: This will be true for an empty string.  I don't know if that matters.  But I see that this is just a reformatting anyway.

admin/app/controllers/admin/CommercialController.scala
```diff
@@ -92,6 +96,31 @@ class CommercialController(implicit context: ApplicationContext) extends Control
...
+    }
+  }
+
+  def getCreativesListing(lineitemId: String, section: String) = Action { implicit request: RequestHeader =>
+
+    val validSections: List[String] = List("uk", "lifeandstyle", "sport", "science")
+
+    val previewUrls: Seq[String] =
+      (for {
+        lineItemId <- Try(lineitemId.toLong).toOption
```
kelvin-chappell: Could `Try(lineitemId.toLong).toOption` just be `Option(lineitemId)`?  I'm not what the exception could be except for an NPE.

admin/app/controllers/admin/CommercialController.scala
```diff
@@ -92,6 +96,31 @@ class CommercialController(implicit context: ApplicationContext) extends Control
...
+        section <- validSections.find(_ == section)
```
kelvin-chappell: Seem to be using the same name twice in scope.

kelvin-chappell (approved):
👍  Just a couple of comments but nothing important

admin/app/controllers/admin/CommercialController.scala
JonNorman: Good point - it looks like this is okay because:
a) the values used are the ad test values, which should never be empty and
b) it is just used to group the ad tests together when rendering the [ad test admin page](https://frontend.gutools.co.uk/commercial/adtests); worst case scenario the ad test with no name will be in the numeric section rather than the non-numeric.

admin/app/controllers/admin/CommercialController.scala
JonNorman: It can throw a `NumberFormatException` and since `lineitemId` is supplied as part of the route, this is the only/most sensible place to do the conversion/validation.

admin/app/controllers/admin/CommercialController.scala
JonNorman: Ah yeah, I'll change that!

PR #16044, opened 2017-03-03
ES6: remove custom `lib/to-array`

gu-stav:
## What does this change?

Removes custom `lib/to-array` implementation and replaces it with lodash functions. This allows an easier ES6 transformation later on.

Less code, easier ES6 conversion.

### DIFF
--- a/tools/es5to6.json
+++ b/tools/es5to6.json
@@ -159,14 +159,12 @@
     "lib/client-rects.js",
     "lib/closest.js",
     "lib/config.js",
-    "lib/contains.js",
     "lib/cookies.js"
   ],
   "NataliaLKB": [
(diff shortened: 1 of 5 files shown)


sndrs (approved):

tools/es5to6.json
```diff
@@ -159,14 +159,12 @@
     "lib/client-rects.js",
     "lib/closest.js",
     "lib/config.js",
-    "lib/contains.js",
```
sndrs: looks like you've got some unmarked changes from another PR branch in this. not a problem in this case, but generally speaking it's easier to review PRs if you can hold off on ones that depend on others till they're merged

PR #16046, opened 2017-03-06
Disallow some deps in es6 world

sndrs:
## What does this change?

disallows importing methods/object etc that are natively available (or already polyfilled)

simpler code, smaller bundles


SiAdcock (approved):
🚀

PR #16047, opened 2017-03-06
Use import in webpack

sndrs:
## What does this change?

- disallows amd require code split points in webpack build
- transpiles `import` to `require` for the karma AMD build

hopefully help #16035

## Tested in CODE?

- [ ] tested


gtrufitt (approved):

SiAdcock (approved):
😍

static/src/javascripts/boot-webpack.js
```diff
@@ -11,6 +11,14 @@ import capturePerfTimings from 'lib/capture-perf-timings';
 // eslint-disable-next-line camelcase,no-undef
 __webpack_public_path__ = `${config.page.assetsPath}javascripts/`;
 
+// Selectively attempt to pollute the global namespace with babel polfills.
+// The polyfills babel provides from core-js will be browser natives where
+// available anyway, so this shouldn't muck about with UAs that don't need it.
+if (!window.Promise) {
+    // used by webpack `import`
+    window.Promise = Promise;
```
regiskuckaertz: I don't see the polyfill as a local variable, does it mean this is redundant? i.e. `Promise` is _already_ being overwritten?

static/src/javascripts/boot-webpack.js
```diff
@@ -31,8 +39,7 @@ domready(() => {
         }
 
         userTiming.mark('commercial request');
-        require(
-            ['bootstraps/commercial'],
+        import('bootstraps/commercial').then(
             raven.wrap({ tags: { feature: 'commercial' } }, commercial => {
```
regiskuckaertz: Just thinking out loud here. Does `raven.wrap` return whatever the enclosed function expression returns? If yes, then you can reduce to (simplifying):

```
import(...)
.then(raven.wrap(commercial => commercial.init())
.then(() => isEnhanced ? import(...) : () => ())
.then(bootEnhanced => bootEnhanced())
```

static/src/javascripts/boot-webpack.js
sndrs: babel will provide it when it sees `Promise`

sndrs:
we're going to use `require.ensure`, closing this for now, to reappear later

PR #16048, opened 2017-03-06
Bundle ophan explicitly

sndrs:
## What does this change?

stops pretending ophan is a code split point, and explicitly bundles it as a dependency

ophan is bundled by webpack already, so we can reduce the instances of split points dramatically = simpler

## Tested in CODE?

- [x] tested


gu-stav (approved):
LGTM 🎆

SiAdcock (approved):
Tidy

PR #16052, opened 2017-03-06
Remove unused JS modules

gtrufitt:
## What does this change?

Removes the following modules that aren't imported into the application anywhere:

- content-websocket
- sticky-social
- socially-referred-burners
- social-most-popular
- track-ad-load ( @regiskuckaertz this isn't used elsewhere is it?)
- take-while
- filterPairs
- date-formats
- subsciber-number-form
- history-containers 

Less files to convert.

No.

## Tested in CODE?

No.


regiskuckaertz:
@gtrufitt nope 👍  don't you remove them from the `es5toes6` map too?

gtrufitt:
> don't you remove them from the es5toes6 map too?
Yup, will do now

gu-stav (approved):
🚀 🔥 LGTM

sndrs (approved):
🔥 🔥

sndrs:
> don't you remove them from the es5toes6 map too?

we can regenerate that when we're ready to open it up. i guess a lot things may change before then

PR #16053, opened 2017-03-06
Remove router module

gtrufitt:
## What does this change?

There used to be frontend routing, now there isn't. Removing the module.

I'll re-balance `es5to6.json` once I've merged a couple of others I've found!

Tidy.


gu-stav (approved):
Nice. Had the same in another open branch :)

sndrs (approved):
🔥

PR #16054, opened 2017-03-06
Extended Switch For Membership Bundles Thrasher

Ap0c:
## What does this change?

Extends the switch for the membership bundles thrasher by two weeks.

This prevents the switch from expiring; we don't want to strip the thrasher code out yet, because we may be doing further tests with it.

## Tested in CODE?

Not yet.


JustinPinner (approved):
👍 🚢 🏁

Ap0c:
Deploy looks ok in Kibana, the switch is off and has 16 days left in the switchboard.

PR #16056, opened 2017-03-06
Load the alternative merchandising component instantly

regiskuckaertz:
The alternative slot that is inserted for people in the Oubtrain_less_ A/B test variant needs to be loaded instantly to be 100% reflective of the current Outbrain_ful_ situation. We add an argument to `addSlot` in order to force instant display and make sure the flag is never set by accident elsewhere.

### DIFF
--- a/static/src/javascripts-legacy/projects/commercial/modules/article-body-adverts.js
+++ b/static/src/javascripts-legacy/projects/commercial/modules/article-body-adverts.js
@@ -174,7 +174,9 @@ define([
     }
 
     function addSlots(totalCount) {
-        qwery('.ad-slot--inline').forEach(addSlot);
+        qwery('.ad-slot--inline').forEach(function (slot) {
+            addSlot(slot);
+        });
         return totalCount;
     }
 });
--- a/static/src/javascripts-legacy/projects/commercial/modules/dfp/add-slot.js
+++ b/static/src/javascripts-legacy/projects/commercial/modules/dfp/add-slot.js
@@ -9,27 +9,27 @@ define([
 ], function (mediator, dfpEnv, Advert, queueAdvert, loadAdvert, enableLazyLoad, performanceLogging) {
     return addSlot;
 
-    function addSlot(adSlot) {
+    function addSlot(adSlot, forceDisplay) {
         adSlot = adSlot instanceof HTMLElement ? adSlot : adSlot[0];
 
         if (dfpEnv.firstAdDisplayed && !(adSlot.id in dfpEnv.advertIds)) { // dynamically add ad slot
             // this is horrible, but if we do this before the initial ads have loaded things go awry
             if (dfpEnv.firstAdRendered) {
-                displayAd(adSlot);
+                displayAd(adSlot, forceDisplay);
             } else {
                 mediator.once('modules:commercial:dfp:rendered', function () {
-                    displayAd(adSlot);
+                    displayAd(adSlot, forceDisplay);
                 });
             }
         }
     }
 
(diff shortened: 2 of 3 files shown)


static/src/javascripts-legacy/projects/commercial/modules/article-body-adverts.js
```diff
@@ -174,7 +174,9 @@ define([
     }
 
     function addSlots(totalCount) {
-        qwery('.ad-slot--inline').forEach(addSlot);
+        qwery('.ad-slot--inline').forEach(function (slot) {
```
regiskuckaertz: Reminder: the signature of an array iteration callback is `function(element, index, originalArray)`

regiskuckaertz:
[Load lucky merch component instantly](https://trello.com/c/cOIG1aKX/242-load-lucky-merch-component-instantly)

static/src/javascripts-legacy/projects/commercial/modules/dfp/add-slot.js
```diff
@@ -9,27 +9,27 @@ define([
 ], function (mediator, dfpEnv, Advert, queueAdvert, loadAdvert, enableLazyLoad, performanceLogging) {
     return addSlot;
 
-    function addSlot(adSlot) {
+    function addSlot(adSlot, forceDisplay) {
         adSlot = adSlot instanceof HTMLElement ? adSlot : adSlot[0];
 
         if (dfpEnv.firstAdDisplayed && !(adSlot.id in dfpEnv.advertIds)) { // dynamically add ad slot
             // this is horrible, but if we do this before the initial ads have loaded things go awry
             if (dfpEnv.firstAdRendered) {
```
rich-nguyen: maybe `firstAdRendered` and the `modules:commercial:dfp:rendered` could be replaced by a single promise?

rich-nguyen (approved):

static/src/javascripts-legacy/projects/commercial/modules/article-body-adverts.js
JonNorman: Nice - so this prevents us calling `adSlot` with the index of the array and thus forcing all but the first ad? Nice catch!

static/src/javascripts-legacy/projects/commercial/modules/high-merch.js
```diff
@@ -67,7 +67,9 @@ define([
             if (args) {
                 fastdom.write(function () {
                     args[1].parentNode.insertBefore(args[0], args[1].nextSibling);
-                    addSlot(args[0]);
+                })
+                .then(function () {
+                    addSlot(args[0], true);
```
JonNorman: Why is this now out of the fastdom write?

JonNorman (approved):

static/src/javascripts-legacy/projects/commercial/modules/dfp/add-slot.js
regiskuckaertz: I wanted to test whether that assumption still held actually, I don't think it does. It was there before I started working here and I never really questioned this code. Yet it feels like a hack. According to [the GPT reference](https://developers.google.com/doubleclick-gpt/reference#googletag.display), we're doing everything right.

static/src/javascripts-legacy/projects/commercial/modules/high-merch.js
regiskuckaertz: 1- Yes 😄 
2- I'm taking the habit of moving everything that is not touching the DOM out `fastdom` frames.

PR #16057, opened 2017-03-06
Remove live/notification-bar JS module

gu-stav:
## What does this change?

Removes `common/modules/live/notification-bar.js` module. It's included, but never gets called.

Less code.


gtrufitt (approved):
🙆‍♂️

PR #16058, opened 2017-03-06
Add fast-images email style

davidfurey:
## What does this change?

Like fast layout, but includes images for all cards.  This commit also swaps the image to the left hand side (for fast and fast-images)

Might increase CTO rate for opinion email.  Measured using exact target

No

## Screenshots

Before:

screen shot 2017-03-06 at 15 28 16

After:

screen shot 2017-03-06 at 15 28 06


joelochlann (approved):

facia/app/layout/EmailLayouts.scala
```diff
@@ -0,0 +1,49 @@
+package layout
+
+case class EmailCardStyle(
```
joelochlann: 💎  This is much nicer

joelochlann:
👍  Great stuff! Happy for this to be merged

PR #16059, opened 2017-03-06
Remove legacy gallery layout for paid galleries

regiskuckaertz:
Katie asked if we could move paid for galleries over to the latest design.

Not sure whether the `isPaidContent(page)` boolean should be moved over to where `showInRelated` is defined. Opinion?

- [x] ask Design if further visual changes are needed
- [x] get approval from Katie
- [x] get approval from all the other people (logos, etc.)

Before:
screenshot-www theguardian com-2017-03-06-16-04-32

After:
screenshot-localhost-9000-2017-03-06-16-05-57


regiskuckaertz:
[Move paid for galleries to new design](https://trello.com/c/BJKjkRJr/244-move-paid-for-galleries-to-new-design)

NataliaLKB:
Yay! Thanks @regiskuckaertz for doing something which was on my list 🙈 .

Is there styling that can be deleted as well? It might have changed, but I remember there being quite a lot of paid for styling for the paid for galleries.

kelvin-chappell (approved):
Looks good to me.  Nice to get rid of special case parallel templates!  Assuming design are happy. 👍 

I think `isPaidContent` should be kept separate as `showInRelated` comes through direct from capi and is a checkbox in Composer.  So it might be confusing if it became a complex property in frontend.

regiskuckaertz:
@kelvin-chappell nice! I'll leave it as-is, then.

@NataliaLKB oh yeah, unused styles! I will hunt them down 🏇

ScottPainterGNM:
@blongden73 would be good to check the styling, I'm not sure whether we have to keep the grey background to differentiate this from editorial content. I feel like we do but would be good to get confirmation.

blongden73:
@ScottPainterGNM @regiskuckaertz Hey so is this design a paid for article or a supported by article? If it's paid for by we should make sure that the colour palette is taken from the old gallery - the grey and teal (I know sorry). But until we have a full scale re-think of the design for paid for by that colour palette needs to exist on those articles.

ScottPainterGNM:
This is paid for @blongden73 so agreed, we must stick with teal and grey. However it is still tonnes better with the new design!

regiskuckaertz:
@blongden73 Yes, paid for. What say I change the background color to be a darker shade of the "paid for" grey? The teal appears in the sticky band, as with the previous design, and so assuming it is enough for now.

blongden73:
@regiskuckaertz That sounds good. Any yellow also needs to be teal.

regiskuckaertz:
@blongden73 how about this?

picture 2

ScottPainterGNM:
If we do go this dark we might need to add the logic in for secondary high contrast logos that we use on the editorial gallery pages.

blongden73:
@regiskuckaertz The background of the article looks good. You could keep the black overlay on the image the same as it was I think. Not sure about the grey. The icons and the series title should defiantly be teal though.

regiskuckaertz:
This?

picture 3

blongden73:
@regiskuckaertz oh sorry not the share icons the camera icon, and the series title above the headline.

regiskuckaertz:
@blongden73 what think you of this:

picture 2

regiskuckaertz:
(you can see dark lines between images and captions, do I keep them as is? also, the background of inline ads is now transparent to pick up the general background color)

regiskuckaertz:
ping @blongden73

blongden73:
@regiskuckaertz The only thing that might be worth doing is making the the rules above the images and in between the image $neutral-6 (#ececec) but apart from that 👍

ScottPainterGNM:
I think this might need to use the same logo logic as the editorial galleries to account for contrast issues. @kelvin-chappell knows about this but checking for the secondary high contrast logo if there is one would stop issues like the screenshot above.

regiskuckaertz:
@ScottPainterGNM I think that is already the case [here](https://github.com/guardian/frontend/blob/master/common/app/views/fragments/commercial/badge.scala.html#L8). I wonder if it is just that dark versions are not supplied at all for some logos?

regiskuckaertz:
@blongden73 like this:

picture 5

ScottPainterGNM:
@regiskuckaertz ah yes so there is! Maybe we just note that in any comms that we ask Katie to share or whoever so that people check and are aware. I guess if they were used to just grey maybe they won't know there is an issue as of yet.

ScottPainterGNM:
Looks like Katie has got this approved so lets 🚢 !

PR #16061, opened 2017-03-06
Sleeve notes test: put all new subscribers in the test

lmath:
## What does this change?
In [PR 15996](https://github.com/guardian/frontend/pull/15996), I set up an a/b test so that new subscribers could either be assigned to the receive the new email or be allocated into the control group.

15996 set the audience percentage at 0%. Until this PR is merged, any new subscribers to Sleeve notes will only be able to be added to the existing email list with the old Sleeve notes email. 

This change sets the audience percentages so that once merged, if the corresponding ab test switches are on:
- 70% of new subscribers receive the new Sleeve notes email
- The remaining 30% of new subscribers are in the control group and receive the old email 

Additionally:
- Adds the url of the new sign up page 

Allows us to test a new version of the Sleeve notes email.

No.

## Tested in CODE?
No.

### DIFF
--- a/common/app/model/EmailNewsletters.scala
+++ b/common/app/model/EmailNewsletters.scala
@@ -230,7 +230,7 @@ object EmailNewsletters {
     listId = 39,
     aliases = List(3834, 3835),
     tone = Some("review"),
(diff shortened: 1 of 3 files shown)


davidfurey:
👍  as long as it is off on the switchboard

PR #16062, opened 2017-03-06
Fixing overflow and layout of paidFor labels

katebee:
## What does this change?

css ph'nglui mglw'nafh Cthulhu R'lyeh wgah'nagl fhtagn 🐙

Besides the css, lets wrap some text in a DIV so it can bully other elements for space, and have class

Desired behaviour:

- Badge label (`Paid for by`) should always be above, and centred over, the badge logo
- Badge logo should never be wider than the card
- Reserve space for the badge label, and badge logo, so that the standfirst text doesn't collide when the label uses the full height
- Not breaking layout anywhere else, in the great game of css-whack-a-mole
- Headache free IE styling?

## BEFORE:

picture 584

picture 585

picture 587

picture 589

## AFTER:

picture 581

picture 586

picture 588

picture 582

## Tested in CODE?

DONE ✅

### DIFF
--- a/static/src/stylesheets/module/commercial/_advert.scss
+++ b/static/src/stylesheets/module/commercial/_advert.scss
@@ -126,9 +126,9 @@
     display: flex;
     align-items: stretch;
 
-    > * {
-        width: 100%;
-    }
+    // > * {
+    //     width: 100%;
+    // }
 }
 
 /* [1] Prevents IE11 bug where the image container's height would go ballooney */
--- a/static/src/stylesheets/module/commercial/_advert.scss
+++ b/static/src/stylesheets/module/commercial/_advert.scss
@@ -125,10 +125,6 @@
 .advert-container {
     display: flex;
     align-items: stretch;
-
-    // > * {
-    //     width: 100%;
-    // }
 }
 
 /* [1] Prevents IE11 bug where the image container's height would go ballooney */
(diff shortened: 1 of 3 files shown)


static/src/stylesheets/module/commercial/_advert.scss
```diff
@@ -126,9 +126,9 @@
     display: flex;
     align-items: stretch;
 
-    > * {
-        width: 100%;
-    }
+    // > * {
+    //     width: 100%;
+    // }
```
katebee: This seems to be contributing to layout difficulties on the paidFor containers, removing it hasn't broken anything AFAIK... _yet_

🔥⁉️

regiskuckaertz: hmm, yeah presumably you are referring to the layout of the badge? If yes, then fair enough. Perhaps have a quick look on IE11 if it has any effect.

regiskuckaertz (approved):
Woohoo after IE tests (do you have a BrowserStack account? If no, just ask Scott)

katebee:
Ha, good call @regiskuckaertz. 

I ran into another weird height issue in IE11, `badge__links` were messing up the layout, fixed it.

regiskuckaertz:
Smashing 😃  :1st_place_medal:

PR #16063, opened 2017-03-07
Fix double high-merch lucky display call

regiskuckaertz:
Instantly loading a slot that is added after the initial commercial load never happened before, so I hacked a quick solution. This fix makes sure the slot does not appear in the `advertsToLoad` queue, for those browsers relying on the scroll listener.

Later, we'll review the `queueAdvert` function, I think it is doing too much.


katebee (approved):
👍

PR #16064, opened 2017-03-07
Extended recommended for you test

davidfurey:
## What does this change?

Extended the recommended for you a/b test and increases from 2% to 5%

The visit frequency of people who have opted in to the test increased.  This change will give us more data, and will also show us how the server side copes with more load.

No

### DIFF
--- a/static/src/javascripts-legacy/projects/common/modules/experiments/tests/recommended-for-you.js
+++ b/static/src/javascripts-legacy/projects/common/modules/experiments/tests/recommended-for-you.js
@@ -38,12 +38,12 @@ define([
     return function () {
         this.id = 'RecommendedForYouRecommendations';
         this.start = '2017-01-17';
-        this.expiry = '2017-03-08';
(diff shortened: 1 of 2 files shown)


static/src/javascripts-legacy/projects/common/modules/experiments/tests/recommended-for-you.js
```diff
@@ -38,12 +38,12 @@ define([
     return function () {
         this.id = 'RecommendedForYouRecommendations';
         this.start = '2017-01-17';
-        this.expiry = '2017-03-08';
+        this.expiry = '2017-04-05';
```
katebee: Not an issue, but it is no longer necessary to have the expiries off-by-one:

Expiries (Scala switch, JS test) now have the same behaviour; the JavaScript test will be able to run until 23:59 on 2017-04-05

katebee (approved):
👍 LGTM

PR #16065, opened 2017-03-07
Remove Guardian Today a/b/c test

davidfurey:
## What does this change?

Removes a test.  All of the new messages performed better than the existing message, but there was no clear winner between the three new messages. 

Removes redundant code


joelochlann (approved):
Looks good!

PR #16067, opened 2017-03-07
Facia: update trendind-topics to the new tags style

gu-stav:
## What does this change?

Updates trending-topics tags on facia pages to the new style used in articles.

More consistent design.

No.

## Screenshots

** Before **

screen shot 2017-03-07 at 11 32 46

screen shot 2017-03-07 at 11 33 11

** After **

screen shot 2017-03-07 at 11 32 35

screen shot 2017-03-07 at 11 32 58

## Tested in CODE?

No.


SiAdcock (approved):
👍 👍 👍

gustavpursche:
@zeftilldeath Although Simon already approved this one, I'd like to hear yur opinion on the combination with the "More ..." Button and the increased space below the element.

zeftilldeath:
Increased space below the element looks great. With regards to the more button, I don't feel like our lozenge button style works inline with text links like that. I feel just the word 'more...' after the final trailing slash will be sufficient.

zeftilldeath:
Also, while we're looking at this bit. Could we standardise this section with the rest of the home page? A blue border at the top edge to edge and blue lowercase heading. No reason to have a new style here (no idea why it was like that originally)

screen shot 2017-03-07 at 16 30 31

paperboyo:
Isn't this huge "Topics" text some regression? I don't remember it being like that until very recently. Apologies if I just name-called someone's design a regression 😜

gustavpursche:
@paperboyo Interesting - thanks for pointing this out. 👍  I'll have a chat with Zef about this.

gustavpursche:
@zeftilldeath 

> I feel just the word 'more...' after the final trailing slash will be sufficient.

I'll try it out, but my feeling is somehow, that it would be confusing, if it just looks like the other links, since it's a different functionality.

> Also, while we're looking at this bit. Could we standardise this section with the rest of the home page? A blue border at the top edge to edge and blue lowercase heading. No reason to have a new style here (no idea why it was like that originally)

Will have a look 👍

gustavpursche:
After a f2f chat we decided to remove the previously hidden tags at all.

paperboyo:
> After a f2f chat we decided to remove the previously hidden tags at all

The ones behind More..? 😱 As I understand, they dynamically indicate the composition of tags on a front. We have so little of this kind of structure discoverability mechanisms... it seems a pity to reduce them even more. Just my 2 cents, obviously...

SiAdcock (approved):

PR #16068, opened 2017-03-07
Move Outbrain / email checks to module root

SiAdcock:
## What does this change?

The intention of the top level `lib` directory is to house self-contained, reusable, composable modules that have no dependencies on other parts of our application. While the `check-dispatcher` and `check-mediator` modules could be made more generic, I have moved them into the module root to signify that they are intended to be generic.

This prepares the way for the ES6 conversion, testing and Flow typing of the `lib` directory.

No

## Tested in CODE?

No


gtrufitt:
I don't think email is the right module for this, it's independent of any particular module (just so happens that the first module to use it is email).

GHaberis (approved):
👍

GHaberis:
Ok, yes the reasoning behind moving it to email was that it's only used to run email checks right now, it should theoretically be available to any module though, so perhaps common/modules would be a suitable place for these files to live.

SiAdcock:
@GHaberis @gtrufitt I have moved modules into the module root. I have also moved the test spec to mirror this

gtrufitt (approved):

PR #16069, opened 2017-03-07
Remove fastdom-idle

SiAdcock:
## What does this change?

Remove another unused module in `lib`.

Less code to maintain. This module was also breaking the rule of no application dependencies in `lib` modules.

No

## Tested in CODE?

No


gu-stav (approved):
Boom! 🎆

gtrufitt:
As discussed f2f, ideally this would be *the* way we do fastdom but there's probably a refactoring task to ⬆️ upgrade fastdom or wrap all fastdom calls like this?

SiAdcock:
I've added this as a Performance-related task to the [Client Side Defeat in Detail Document](https://docs.google.com/spreadsheets/d/1uUms84HMp2M1ju4uVUbFRImd8D7fxgpQxCkWmf76av8/edit#gid=0)

sndrs (approved):
should have done something with this, nice one

PR #16070, opened 2017-03-07
Remove _gallerythumbs.scss

gu-stav:
## What does this change?

Removes `content/_gallerythumbs.scss`, since there are no other references to the module.

Less code, slimmer builds.

No.


katebee (approved):
:+1: and hooray for less css 🔥

PR #16073, opened 2017-03-07
Convert modules/sport/football/match-info to ES6

gu-stav:
## What does this change?

Converts another module into ES6.

ES6.

No.


static/src/javascripts/projects/common/modules/sport/football/match-info.js
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * @param {Object} match
+ * @param {string} whosCalling (url)
+ */
+
+class MatchInfo {
+    constructor(match, whosCalling) {
+        const urlBase = '/football/api/match-nav';
+        const pageSlug = encodeURIComponent(whosCalling);
+        let matchSlug = match.id;
```
regiskuckaertz: `const matchSlug = match.id || [match.date].concat(match.teams).join('/');`

regiskuckaertz:
yay 🎉

Although I would rather you removed this module entirely, it is completely pointless and dangles in memory for nothing.

regiskuckaertz:
(sorry I keep on hitting the 'close' button by mistake 😅 )

gustavpursche:
@regiskuckaertz Addressed your coding style comment now - could you review it again? :) I'm completly on your side about the removal. Let's discuss this soon after the ES6 conversion!

regiskuckaertz (approved):
Great! Happy to discuss further topics with you afterwards.

static/src/javascripts/projects/common/modules/sport/football/match-info.js
```diff
@@ -0,0 +1,29 @@
...
+ * @param {Object} match
+ * @param {string} whosCalling (url)
+ */
+
+class MatchInfo {
+    constructor(match, whosCalling) {
+        const urlBase = '/football/api/match-nav';
+        const pageSlug = encodeURIComponent(whosCalling);
+        const matchSlug = [match.date].concat(match.teams).join('/');
+        const endpoint = match.id || matchSlug;
```
regiskuckaertz: I don't think it matters here, just keep in mind that `A; B;` is operationally different from `A || B`. In the latter case, `B` is not evaluated if `A` is truthy, which could matter to avoid divisions by zero or error cases.

sndrs:
did you mean to commit the addition of the merge conflict?

static/src/javascripts/projects/common/modules/sport/football/match-info.js
gustavpursche: You are right. I wanted to avoid the ugly formatting, because it doesn't fit on a single line (now it does, thanks to a shorter variable name). Thanks 👍

SiAdcock:
If this module is potentially deletable, can we add a magical `#?` comment saying so, so we don't forget about it?

PR #16074, opened 2017-03-07
Component: improve readability of render() and catch errors

gu-stav:
## What does this change?

Adds check around `document.getElementById('tmpl-' + this.templateName)` to avoid TypeError when accessing the array.

Less JS errors.

No.


SiAdcock (approved):
Tidy, robust and clear(er)! 👍

PR #16076, opened 2017-03-07
Break one-word email names in the appropriate place

joelochlann:
Fixes a bug from #16006.

# before
picture 481

# after
picture 484
picture 485

@superfrank @gidsg


superfrank:
Can we tighten the line spacing up, so if it does break it becomes the same px size as the font please?

joelochlann:
Did what you suggested @superfrank. It now looks like this:
picture 488

gu-stav (approved):
🌷 Looks good!

PR #16077, opened 2017-03-07
Require.ensure

sndrs:
## What does this change?

updates webpack code split points to use `require.ensure`

- flow conformance (hopefully) for #16035
- human-friendly bundle names

## Tested in CODE?

- [x] tested


static/src/javascripts/boot.js
```diff
@@ -0,0 +1,71 @@
...
+            });
+        }
+
+        userTiming.mark('commercial request');
+        require.ensure(
+            [],
+            require => {
+                raven.context({ tags: { feature: 'commercial' } }, () => {
+                    userTiming.mark('commercial boot');
+                    require('bootstraps/commercial').init().then(() => {
```
gtrufitt: This is getting a little pyramid-y because of the `require.ensure` then `require` stuff, do you think it could do with extracting the functions so the `then`s are clearer?

gtrufitt (approved):
Great 👍

static/src/javascripts/boot.js
sndrs: yeah - i wondered how the code splitting would work then? currently they have to be nested to so that the enhanced bundle doesn't include deps common with commercial

sndrs: double checked - webpack must be looking for strings? whatever it is, it doesn't understand that you're calling it's code splitting points out of line... :(

gtrufitt: boo :/ Fair enough, not a major, just a nitpick :D

PR #16081, opened 2017-03-07
Tailor surveys

MahanaC:
I've made the following changes to @oilnam's work:
- Tailor now specifies the survey question and ID, and we use this to display the survey (a question was hardcoded previously)
- Tailor now says that for a given user, we shouldn't show a specific survey again until a specified date. We store this in the cookie and if that data hasn't passed we pass tailor a list of surveys not to show
- Ophan tracking now includes the ID of the survey - both in the lazy component when the survey is served, and in the interaction when the survey is answered
- We pass edition to tailor, which is used as a proxy for location for targeting. This defaults to ROW.
- For testing/debugging purposes, if you set surveyToShow to contain a survey ID in local storage, this will force that survey to be shown.

A sample response from tailor: `https://tailor.guardianapis.com/suggestions?browserId=wcKHZX-illSiSNxQq45rHcCg&surveyToShow=1`.

@stephanfowler @SiAdcock @oilnam @lindseydew @mike-ruane

### DIFF
--- a/static/src/javascripts-legacy/projects/common/modules/experiments/tests/tailor-survey.js
+++ b/static/src/javascripts-legacy/projects/common/modules/experiments/tests/tailor-survey.js
@@ -13,33 +13,33 @@ define([
     'raw-loader!common/views/experiments/tailor-survey.html',
     'lib/fetch-json',
     'lodash/collections/forEach',
-    'ophan/ng'
-], function (
-    bean,
-    bonzo,
-    fastdom,
-    Promise,
-    debounce,
-    config,
-    cookies,
-    storage,
-    mediator,
-    fastdomPromise,
-    privateBrowsing,
-    quickSurvey,
-    fetchJson,
-    forEach,
-    ophan
-) {
+    'ophan/ng',
+    'lib/template'
+], function (bean,
+             bonzo,
+             fastdom,
+             Promise,
+             debounce,
+             config,
+             cookies,
+             storage,
+             mediator,
+             fastdomPromise,
+             privateBrowsing,
+             tailorSurvey,
+             fetchJson,
+             forEach,
+             ophan,
+             template) {
     return function () {
         this.id = 'TailorSurvey';
-        this.start = '2017-01-25';
-        this.expiry = '2017-03-31';
-        this.author = 'Manlio';
+        this.start = '2017-03-07';
+        this.expiry = '2017-04-30';
+        this.author = 'Manlio & Mahana';
         this.description = 'Testing Tailor surveys';
         this.audience = 0.01;
         this.audienceOffset = 0.7;
-        this.successMeasure = 'We can show a survey on Frontend to regular users only (as decided by Tailor)';
+        this.successMeasure = 'We can show a survey on Frontend as decided by Tailor';
         this.audienceCriteria = 'All users';
         this.dataLinkNames = 'Tailor survey';
         this.idealOutcome = '';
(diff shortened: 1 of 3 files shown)


static/src/javascripts-legacy/projects/common/modules/experiments/tests/tailor-survey.js
```diff
@@ -13,33 +13,33 @@ define([
...
+             forEach,
+             ophan,
+             template) {
     return function () {
         this.id = 'TailorSurvey';
-        this.start = '2017-01-25';
-        this.expiry = '2017-03-31';
-        this.author = 'Manlio';
+        this.start = '2017-03-07';
+        this.expiry = '2017-04-30';
```
katebee: You will also want to update the expiry in:
`common/app/conf/switches/ABTestSwitches.scala`

N.B. it is better to pick a weekday 🙃  ...otherwise a test will fail

MahanaC: Thanks @katebee! Have done that :)

static/src/javascripts-legacy/projects/common/modules/experiments/tests/tailor-survey.js
```diff
@@ -48,31 +48,122 @@ define([
...
+            // If we want to force tailor to show a particular survey we can set an attribute in local storage to have
+            // key = 'surveyToShow', and value = the survey id. Tailor will then override other logic for display, and
+            // look for a survey with this ID to return. This is useful as we can easily see how a particular survey
+            // would be rendered, without actually putting it live. If this parameter is empty or not specified, tailor
+            // behaves as usual.
+            var surveyToShow = localStorage.getItem('surveyToShow');
 
-        function callTailor(bwid) {
-            var endpoint = 'https://tailor.guardianapis.com/suggestions?browserId=' + bwid;
+            var endpoint = 'https://tailor.guardianapis.com/suggestions?browserId=' + bwid + '&edition=' + config.page.edition +
```
gidsg: I would suggest adding `tailorHost` to the Javascript config as you are using it a couple of places in Frontend now. i.e. https://github.com/guardian/frontend/blob/master/docs/03-dev-howtos/13-Update-configuration-in-s3.md

gidsg: This formatting is inconsistent, I would suggest one parameter per line, here you add 2 in one line.

gidsg (changes requested):

static/src/javascripts-legacy/projects/common/modules/experiments/tests/tailor-survey.js
MahanaC: Thanks, have done that.

static/src/javascripts-legacy/projects/common/modules/experiments/tests/tailor-survey.js
```diff
@@ -13,33 +13,33 @@ define([
...
+             cookies,
+             storage,
+             mediator,
+             fastdomPromise,
+             privateBrowsing,
+             tailorSurvey,
+             fetchJson,
+             forEach,
+             ophan,
+             template) {
```
SiAdcock: Nit: call me nostalgic, can we keep the old spacing? 😄

MahanaC: That look better? :wink:

static/src/javascripts-legacy/projects/common/modules/experiments/tests/tailor-survey.js
```diff
@@ -48,31 +50,124 @@ define([
...
+        }
+        // Given a response from tailor, we see if the response has a survey suggestion, and if so return the first
+        // survey suggestion (there should only ever be one, but just in case).
+        function getSurveySuggestionToShow(response) {
+            if (response.suggestions) {
+                var surveySuggestions = response.suggestions.filter(function (suggestion) {
+                    return suggestion.class == 'SurveySuggestion';
+                });
+
+                if (surveySuggestions) {
```
SiAdcock: `filter` always returns an array, so `surveySuggestions` will always be truthy, even if the array is empty. It would be better to check `surveySuggestions.length`

static/src/javascripts-legacy/projects/common/modules/experiments/tests/tailor-survey.js
```diff
@@ -48,31 +50,124 @@ define([
...
+            return surveysWeCannotShow.map(function (idAndDate) {
+                return idAndDate.split('=')[0]
+            }).toString();
+        }
+
+        // Getting simple json from tailor's reponse to be passed to the html template
+        function getJsonFromSurvey(surveySuggestionToShow) {
+            return {
+                "question" : surveySuggestionToShow.data.survey.question,
+                "id" : surveySuggestionToShow.data.survey.surveyId
```
SiAdcock: Nit: We don't need quotes around these property names

static/src/javascripts-legacy/projects/common/modules/experiments/tests/tailor-survey.js
```diff
@@ -48,31 +50,124 @@ define([
...
+
+            var isAfterToday = function (cookieValue) {
+                var date = cookieValue.split('=')[1];
+                return new Date(date).valueOf() > new Date().valueOf();
+            };
+
+            var surveysWeCannotShow = values.filter(isAfterToday);
+
+            return surveysWeCannotShow.map(function (idAndDate) {
+                return idAndDate.split('=')[0]
```
SiAdcock: Nit: missing a semicolon

static/src/javascripts-legacy/projects/common/modules/experiments/tests/tailor-survey.js
```diff
@@ -48,31 +50,124 @@ define([
...
+                // first time we show any survey
+                cookies.add('GU_TAILOR_SURVEY', newCookieValue, 365);
+            }
+        }
+        // Given a response from tailor, we see if the response has a survey suggestion, and if so return the first
+        // survey suggestion (there should only ever be one, but just in case).
+        function getSurveySuggestionToShow(response) {
+            if (response.suggestions) {
+                var surveySuggestions = response.suggestions.filter(function (suggestion) {
+                    return suggestion.class == 'SurveySuggestion';
```
SiAdcock: Nit: prefer `===` instead of `==`

static/src/javascripts-legacy/projects/common/modules/experiments/tests/tailor-survey.js
```diff
@@ -96,30 +191,30 @@ define([
...
+                            recordOphanAbEvent(answer, surveyId);
+
+                            mediator.emit('tailor:survey:clicked');
+                            fastdom.write(function () {
+                                disableRadioButtons('fi-survey__button');
+                                surveyFadeOut();
+                                thankyouFadeIn();
+                            });
+                        }
+                    });
```
SiAdcock: Nit: editor went a bit over-eager with the tabs here

static/src/javascripts-legacy/projects/common/modules/experiments/tests/tailor-survey.js
```diff
@@ -48,31 +50,124 @@ define([
...
                 method: 'get'
             });
         }
 
+        // Every time we show a survey to a user, we cannot show it again to that suer for a specified number of days.
+        // We store 'surveyId=dayShowAgain' in the cookie, and pass any surveys that cannot currently be shown in the
+        // call to tailor.
+        function storeSurveyShowedInCookie(surveySuggestionToShow) {
+            var id = surveySuggestionToShow.data.survey.surveyId;
+            var dayCanShowAgain = surveySuggestionToShow.data.dayCanShowAgain;
```
SiAdcock: Can we pass `surveySuggestionToShow.data` straight into this function, rather than `surveySuggestionToShow`?

MahanaC:
@SiAdcock Thanks for comments, have made those changes. Look ok now?

static/src/javascripts-legacy/projects/common/modules/experiments/tests/tailor-survey.js
MahanaC: Thanks, didn't know about this. I might do this as a separate PR if that's ok, and make sure we do this in all places it's used.

SiAdcock (approved):
👍

PR #16086, opened 2017-03-07
Move utils tests to lib

SiAdcock:
## What does this change?

Our JS `projects/common/utils` directory was moved to `lib` in #16016. This change similarly moves the tests, to mirror the code under test.

Consistent structure across code and tests.

No

## Tested in CODE?

No


sndrs (approved):
mega

PR #16087, opened 2017-03-07
Add sizes options to createSlot

regiskuckaertz:
Will come in handy when we move out of using `data-` attributes, but right now I need it for an upcoming AB test.

PR #16088, opened 2017-03-07
Offset ads in the right-hand column in articles

regiskuckaertz:
Reinstate the previous test with the following difference: `inline2` and up can now receive a DMPU (hence [this PR](https://github.com/guardian/frontend/pull/16087)).


static/src/javascripts-legacy/projects/common/modules/experiments/tests/increase-inline-ads.js
```diff
@@ -0,0 +1,30 @@
...
+        this.audience = .05;
+        this.audienceOffset = 0;
+        this.successMeasure = 'Our inventory has increased';
+        this.audienceCriteria = '';
+        this.dataLinkNames = '';
+        this.idealOutcome = 'Viewability and user engagement has not been impacted';
+        this.showForSensitive = true;
+
+        this.canRun = function () {
+          return detect.isBreakpoint({ min: 'desktop' });
```
regiskuckaertz: I _think_ I have to disable the test in immersive articles, will do that.

regiskuckaertz:
[Inventory OKR: Inline ads into right hand column on desktop tests](https://trello.com/c/fyEtT96v/174-inventory-okr-inline-ads-into-right-hand-column-on-desktop-tests)

katebee (approved):
👍 💎

PR #16089, opened 2017-03-07
Es6 lib/$

sndrs:
## What does this change?

- converts `lib/$` to es6
- adds a test for it
- rationalised `babelrc` a little

- trying out conversion script
- getting jest running properly

## Tested in CODE?

- [x] tested


SiAdcock (approved):
Love it man. Them tests tho 👌 👌 👌

static/src/javascripts/lib/$.js
```diff
@@ -0,0 +1,33 @@
...
+    },
+});
+
+function $(selector, context) {
+    return bonzo(qwery(selector, context));
+}
+
+$.create = s => bonzo(bonzo.create(s));
+
+$.ancestor = (el, className) => {
```
regiskuckaertz: `Element.closest`? I don't think this is used anywhere though

gustavpursche: `lib/closest`, since `Element.closest` still has bad support :)

sndrs:
fair few changes just gone in with typing stuff, would you mind looking over it again?

static/src/javascripts/lib/$.js
```diff
@@ -0,0 +1,34 @@
...
+    if (el === null || el.nodeName.toLowerCase() === 'html') {
+        return false;
+    }
+    if (!el.parentNode || bonzo(el.parentNode).hasClass(className)) {
+        return el.parentNode;
+    }
+    return $.ancestor(el.parentNode, className);
+};
+
+$.forEachElement = (selector: string, fn: Function): Array<number> => {
```
regiskuckaertz: Isn't the return type `Array<Element>`?

regiskuckaertz:
Nice 👍

static/src/javascripts/lib/$.spec.js
```diff
@@ -0,0 +1,48 @@
...
+    test(
+        'applies a function to an array of elements described by a selector',
+        () => {
+            $.forEachElement('.parent > *', el => el.classList.add('red'));
+            expect($('.child').hasClass('red')).toBe(true);
+            expect($('.sibling').hasClass('red')).toBe(true);
+        }
+    );
+
+    test("can find an element's ancestor", () => {
```
gtrufitt: Why are these double quoteys here, thought we linted them?

sndrs: `t's`, prettier converts them

static/src/javascripts/lib/$.js
sndrs: er...yeah, looks like it! will check

PR #16091, opened 2017-03-08
Load enhanced if commercial switch is off

SiAdcock:
## What does this change?

Currently, if the Commercial Switch is off, the Enhanced chunk is never loaded. 

This change adds a new split point that creates an Enhanced chunk containing all dependencies that are also part of the Commercial chunk. The `enhanced-no-commercial` chunk is loaded if the Commercial Switch is off.

Enhanced JavaScript is loaded regardless of the state of the Commercial Switch 

No

## Tested in CODE?

No

### DIFF
--- a/static/src/javascripts/boot.js
+++ b/static/src/javascripts/boot.js
@@ -75,5 +75,21 @@ domready(() => {
             },
             'commercial'
         );
+    } else if (window.guardian.isEnhanced) {
+        userTiming.mark('enhanced request');
+        require.ensure(
+            [],
+            require => {
+                userTiming.mark('enhanced boot');
+                require('bootstraps/enhanced/main')();
+
+                if (document.readyState === 'complete') {
+                    capturePerfTimings();
+                } else {
+                    window.addEventListener('load', capturePerfTimings);
+                }
+            },
+            'enhanced-no-commercial'
+        );
     }
 });


static/src/javascripts/boot.js
```diff
@@ -75,5 +75,21 @@ domready(() => {
             },
             'commercial'
         );
+    } else if (window.guardian.isEnhanced) {
```
regiskuckaertz: hmmm, I think you have to chain them somehow:
```
bootCommercial().then(bootEnhanced)
function bootCommercial() {
  return config.switches.commercial ? Promise.resolve() : commercial.init();
}
```

SiAdcock: Would it not be better to prevent commercial from being requested if the switch is off? It saves the client from having to download and parse ~1MB of JavaScript

regiskuckaertz: Exactly! I was simplifying:

```
bootCommercial = () => config.switches.commercial ? Promise.resolve() : loadCommercial()
loadCommercial = () => new Promise(resolve => require.ensure(...))
```

regiskuckaertz: ooooh dammit, I didn't you did not get rid of the `then` part above. All is fine then.

regiskuckaertz (approved):
There's potential for abstraction, but can be done later.

sndrs:
missed this but ❤️

PR #16092, opened 2017-03-08
Photo essay styling

blongden73:
Styling for photoessay display hint. 

screencapture-localhost-3000-global-2017-feb-07-photo-essay-testing-1488973218909

@NataliaLKB @SiAdcock @desbo @gtrufitt - Could you take a look please.


common/app/model/content.scala
```diff
@@ -79,7 +79,8 @@ final case class Content(
     metadata.contentType.toLowerCase == "gallery" && !metadata.commercial.exists(_.isPaidContent)
   }
   lazy val isExplore = ExploreTemplateSwitch.isSwitchedOn && tags.isExploreSeries
-  lazy val isImmersive = fields.displayHint.contains("immersive") || isImmersiveGallery || tags.isTheMinuteArticle || isExplore
+  lazy val isImmersive = fields.displayHint.contains("immersive") || isImmersiveGallery || tags.isTheMinuteArticle || isExplore || isPhotoEssay
+  lazy val isPhotoEssay = fields.displayHint.contains("photoEssay")
```
SiAdcock: Can we move this up a line? Makes the logic slightly more understandable

static/src/stylesheets/module/content/_photo-essay.scss
```diff
@@ -0,0 +1,93 @@
...
+
+        //important because it is important elsewhere
+        .pullquote-cite {
+            color: $neutral-2 !important;
+        }
+
+        &.element--halfWidth {
+          width: 48%;
+          float: left;
+          margin-right: 20px;
```
SiAdcock:
```css
margin-right: $gs-gutter;
```

static/src/stylesheets/module/content/_photo-essay.scss
```diff
@@ -0,0 +1,93 @@
...
+          border-bottom: none;
+          margin-top: 6px;
```
SiAdcock:
```css
margin-top: $gs-baseline / 2;
```

static/src/stylesheets/module/content/_photo-essay.scss
```diff
@@ -0,0 +1,93 @@
...
+            top: 0px;
+            bottom: 0px;
+            z-index: 300;
+            position: absolute;
+            background-color: white;
+          }
+
+          & + p {
+            clear: both;
+            padding-top: 6px;
```
SiAdcock:
```css
padding-top: $gs-baseline / 2;
```

static/src/stylesheets/module/content/_photo-essay.scss
```diff
@@ -0,0 +1,93 @@
...
+          .inline-expand-image {
+            display: none;
+          }
+
+          &:before{
+            content: '';
+            display: block;
+            width: 2px;
+            top: 0px;
+            bottom: 0px;
```
SiAdcock: Can we use `0` instead of `0px`?

static/src/stylesheets/head.content.scss
```diff
@@ -25,6 +25,7 @@
 @import 'module/content/_recipe-article';
 @import 'module/content/_article-minute';
 @import 'module/content/_gallery.head.scss';
+@import 'module/content/_photo-essay';
```
SiAdcock: The contents of `head.content.scss` are included on every page. I think there's a case for inlining these styles to keep the CSS weight and parse time down for pages that don't need them

static/src/stylesheets/module/content/_photo-essay.scss
```diff
@@ -0,0 +1,93 @@
+.content--article {
+    .element-image--photo-esssay {
+        padding-bottom: 0;
+        margin-bottom: 0.5rem;
```
gustavpursche: Would it be possible to work with `$gs-baseline` here?

static/src/stylesheets/module/content/_photo-essay.scss
```diff
@@ -0,0 +1,93 @@
...
+
+        &.fig--no-caption.fig--has-shares {
+            padding-bottom: 0px !important;
+        }
+
+        .block-share {
+            display: none !important;
```
gustavpursche: If there is a a way to avoid these `!important`s we should do so. If you need any help with this, let me know!

static/src/stylesheets/module/content/_photo-essay.scss
```diff
@@ -0,0 +1,93 @@
...
+        }
+        .caption {
+            max-width: 100%;
+            color: $news-main-1;
+        }
+        .article__fullscreen {
+            background-color: transparent;
+            border:1px solid rgba($neutral-6,0.5);
```
gustavpursche: Does this pass the linter? I guess there should be a space between `: 1px` and after the comma.

static/src/stylesheets/module/content/_photo-essay.scss
```diff
@@ -0,0 +1,93 @@
...
+        }
+
+        &:hover .article__fullscreen {
+            background-color: white;
```
gustavpursche: Please use `#ffffff` instead - shouldn't even pass the linter :)

static/src/stylesheets/module/content/_photo-essay.scss
```diff
@@ -0,0 +1,93 @@
...
+
+          &:before{
+            content: '';
+            display: block;
+            width: 2px;
+            top: 0px;
+            bottom: 0px;
+            z-index: 300;
+            position: absolute;
+            background-color: white;
```
gustavpursche: Same here - Hex value instead of name.

static/src/stylesheets/module/content/_photo-essay.scss
```diff
@@ -0,0 +1,93 @@
...
+            clear: both;
+            padding-top: 6px;
+          }
+
+        }
+
+    }
+
+
+    .element-pullquote--photoEssay {
```
gustavpursche: I think we should not use camelCase here. Would it be possible to rewrite it into `.element-pullquote--photo-essay`?

gtrufitt:
image

Caught this?

static/src/stylesheets/module/content/_photo-essay.scss
katebee: We have the `is-hidden` class in `_helpers.scss`; could adding the class to the element work instead?
```
.is-hidden,
[hidden] {
    display: none !important;
}
```

gustavpursche: Sounds good to me!

paperboyo:
If that is to replace the clunky bespoke “tool” which is hard to work with and traps great imagery in small low res frames without native Lightbox, only 🍻 are not enough! 🍻:beers: also aren’t.

Some loose thoughts:

1. Should we use @zef’s new info icon-styled caption anywhere here, like e.g. main media or chapter images (at least on mobile)?

2. I do hope images from a grid of four (visible in the middle, above) take part in an ad hoc article-wide Lightbox.

3. Can this grid of four be weighted other than inline (it may work better as showcase or, maybe, even immersive)? More generally, how is this grid of four put together in Composer, hopefully not through any ugly hack? What happens to individual pic’s captions (can they be visible in-article or only in a Lightbox)?

4. Does it always have this turquoise tone or can it inherit others too?

5. Recently, we (@zef again) have changed how Lightbox CTAs are displayed on pics within galleries (they are hidden by default and only show up on hover). I would argue this style would fit photo essay well.

6. I hope that photo essey cards on fronts finally will have proper media tone treatment. Which brings quite a radical idea: maybe the whole essay itself shoul be laid out in media tone (as a plus, one wouldn’t have to bother with tone inheritance over newsy turquoise...)

Me having bazyllion of numbered questions = a clear sign of 💎!

static/src/stylesheets/module/content/_photo-essay.scss
blongden73: yeah it passed. I was suprised that some of this did to be honest went back in and tidied it up after but there were some indenting that it didnt pick up on either.

static/src/stylesheets/module/content/_photo-essay.scss
blongden73: The first !important is because it already gets given !important from somewhere else. The second one I think I can get rid of.

blongden73:
@paperboyo 

1. yes they use Zef's caption style. 

2. yes they do. 

3. There is a new weighting in composer called "half-width" which currently only gets styled for photoessay templates like this. They just take up half the width of the main column. I thought about making them wider but it causes all kinds of alignment issues. 

4. The photo essay uses the blue throughout. Currently it just takes this colour - we could look at making it tonal but it feels to me like it needs to just have one consistent style and "tone" for now. 

5. Yes this would probably be a good call - currently they are styled white to make them more recessive. 

6. Not sure about this - as otherwise you could argue that a photoessay is just a gallery which its not - it will always have a different pacing to a gallery - using more text and pullquotes within it - its definatly more an article than a gallery so I think making the background black wouldn't be right for ease of reading etc. 

thanks for the feedback!! :-)

static/src/stylesheets/module/content/_photo-essay.scss
katebee: > these get given !important else where so they need to have !important here

😱 ... I suppose this is where I ask how far down the rabbit hole do we go? Is it possible to remove some of the specificity (or `!important`) from elsewhere, so that we don't have to use `!important` to override?

Unless it is `!importants` all the way down... 🐢🐢🐢🐢🐢🐢🐢🐢

paperboyo:
@blongden73 

1-5 👍
6. My thinking went that photo essay was styled like that because as an interactive it was always outside our tone styling types. Much like [this](https://www.theguardian.com/profile/simone-lia) or [that](https://www.theguardian.com/profile/first-dog-on-the-moon) which should both really be media tone and aren't because they are hacks. I thought that making photo essay a first class citizen is an opportunity to rectify that. But you may have a point that styling it as a gallery could impede legibility (although, I'm not sure). Anyway, just a thought and, obviously, entirely not my call!

Thanks!

static/src/stylesheets/inline/article-photo-essay.scss
```diff
@@ -0,0 +1,156 @@
...
+      }
+    }
+
+    blockquote.photo-essay-block-quote {
+      font-family: $f-serif-headline;
+      font-style: normal;
+      margin: 0 0 $gs-baseline * 2;
+
+      &:not(.quoted) {
+        background-color: #f6f6f6;
```
gustavpursche: Could you choose a gray, which is already defined here http://guardian.github.io/pasteup-palette/demo/? We're trying not to add new colors, since the palette is already rich :)

static/src/stylesheets/inline/article-photo-essay.scss
```diff
@@ -0,0 +1,156 @@
...
+        padding: $gs-baseline / 1.5  $gs-baseline $gs-baseline;
+      }
+
+      h2 {
+        color: $news-main-1;
+        font-weight: 900;
+        font-size: 1.8rem;
+        line-height: 1.8rem;
```
gustavpursche: I was told, we're always writing font-sizes in `px`, since they are converted automatically. In this case you could also write `line-height: 1` I guess.

static/src/stylesheets/inline/article-photo-essay.scss
```diff
@@ -0,0 +1,156 @@
...
+        &:hover {
+          border-bottom: 0px solid $news-main-1;
+          background-color: $news-main-1;
+          color: #ffffff;
+        }
+      }
+
+      &.quoted {
+        color: $news-main-1;
+        font-size: 1.2rem;
```
gustavpursche: Same here - please use `px` instead.

static/src/stylesheets/inline/article-photo-essay.scss
```diff
@@ -0,0 +1,156 @@
...
+        padding-bottom: 0;
+        margin-bottom: $gs-baseline;
+
+        //importants here to overide previously set !important
+        &.fig--no-caption.fig--has-shares {
+            padding-bottom: 0px !important;
+        }
+
+        .block-share {
+            display: none !important;
```
gustavpursche: Did the `u-hidden` not work?

static/src/stylesheets/inline/article-photo-essay.scss
```diff
@@ -0,0 +1,156 @@
...
+            max-width: 100%;
+            color: $news-main-1;
+
+            .inline-information.reveal-caption-icon {
+              background-color: $news-main-1;
+            }
+        }
+        .article__fullscreen {
+            background-color: transparent;
+            border: 1px solid rgba($neutral-6,0.5);
```
gustavpursche: Spaces after commas missing. Does this pass the sass-linter?

static/src/stylesheets/inline/article-photo-essay.scss
```diff
@@ -0,0 +1,156 @@
...
+            width: 48%;
+            float: left;
+            margin-right: $gs-gutter;
+            border-bottom: none;
+            margin-top: $gs-baseline / 2;
+        }
+
+        .inline-quote {
+          svg {
+            fill:$news-main-2;
```
gustavpursche: Space missing after ":".

static/src/stylesheets/inline/article-photo-essay.scss
```diff
@@ -0,0 +1,156 @@
...
+            opacity: 1;
+        }
+
+        .section-title {
+            &:before {
+                background-color: $news-main-2;
+                display: block;
+                content: '';
+                width: gs-span(2);
+                height: 0.5rem;
```
gustavpursche: If possible, please try to use `$gs-baseline` here.

gu-stav (commented):
Cool! Just a few minor things ... 🚀

static/src/stylesheets/inline/article-photo-essay.scss
blongden73: Yeah the is-hidden gets overwritten by an !important else where.

static/src/stylesheets/inline/article-photo-essay.scss
blongden73: Yup.

SiAdcock (approved):
Other than Gustav's comments, LGTM nice work 👍

static/src/stylesheets/inline/article-photo-essay.scss
blongden73: @gustavpursche I've just removed the block-share div in the cleaner instead.

PR #16093, opened 2017-03-08
Add flow typed libraries

gtrufitt:
## What does this change?

Adds the `flow-typed` libraries.

Typing for libraries.


sndrs (approved):
👀 27k lines!

gtrufitt:
Productivity 📈

rich-nguyen:
why did these go into the repo, rather than be package-installed? And at the root of the project?

gtrufitt:
@rich-nguyen Re "why did these go into the repo, rather than be package-installed?" https://github.com/flowtype/flow-typed#why-do-i-need-to-commit-the-libdefs-that-flow-typed-installs-for-my-project

> Libdefs in flow-typed are tagged at both Flow and library version when they are installed, but libdefs themselves can improve over time. For example, they may have a bug or there may be an improvement to their accuracy or completeness.

> When a libdef is improved or updated in flow-typed, there's some chance that the change could introduce new Flow errors into your project. As good as it is to find new issues, we also want to make sure that Flow-errors in your project are consistent and predictable over time.

And re: "And at the root of the project?" - Yup, I'm going to move these and the Gu specific ones further down.

rich-nguyen:
Oh. Sounds like a job for sub-versioning. As much as I disliked shrinkwrap, I never felt like file-copying was a better solution.

gtrufitt:
Yeah, very true, certainly sounds like something sem-ver and a lockfile solves.

It's been discussed in the flowtyped issues (https://github.com/flowtype/flow-typed/issues/413) but I don't really buy the reasoning of "sometimes a dev might just need to hack something and move forward". I'll find out if there's been any kind of thoughts on it since then.

PR #16094, opened 2017-03-08
Epic regulars test

jranks123:
## What does this change?

A new epic message test with messages that allude to the fact that a lot of people read our news regularly for free.

We have used tailor to see if the user is a regular and if they are, we append _regular to the campaign code (_non_regular if not). We are testing these messages on both regulars and non-regulars, so we can see if they impact either differently from the other. In order to do this, we had to do a little rejigging of contributions utilities

We will be analysing the response from regulars and non-regulars separately, in order to learn whether different messaging works for users of differing regularity 

## Screenshots
Control message
screenshot at mar 08 13-30-57

Fairness message strong
fairness_strong

Reliance message
reliance

Fairness message mild
screenshot at mar 08 12-00-35

Alternate hook
alternate_hook


Mullefa:
@guardian/contributions

gu-stav (approved):
Looks great 🚀

static/src/javascripts-legacy/projects/common/modules/tailor/tailor.js
```diff
@@ -9,8 +9,22 @@ define([
...
 
+    function getSuggestions(browserId) {
+        return fetchJson('https://tailor.guardianapis.com/suggestions?browserId=' + browserId, {
+            method: 'get'
+        });
+    }
+
+    function getRegularStatus(browserID) {
+        return getSuggestions(browserID).then(function(res) {
+            return res.userDataForClient.regular;
```
desbo: just as defensive thing it could be worth using `try`/`catch` here and returning false if we don't have the right response from tailor

desbo (approved):

PR #16096, opened 2017-03-08
Revert "Video main media caption"

zeftilldeath:
Reverts guardian/frontend#15952

Something seems to be keeping the video controls from auto-hiding when the video plays. Going to revert then look into why.


akash1810 (approved):
😢

PR #16099, opened 2017-03-08
Update lint references to validate and move flow into javascript validation

gtrufitt:
## What does this change?

As the title. Also, this makes flow fail in `post-push` hook.

Integrated flowing.


gu-stav (approved):
Looking forward to all fights with `pre-push`! 🌷

SiAdcock (approved):

PR #16103, opened 2017-03-08
Remove lib/QueueAsync

gu-stav:
## What does this change?

I wanted to convert another module into ES6, but it turned out, it would be easier to remove it at all.

Less code & less cognitive load.

No.


static/src/javascripts-legacy/projects/common/modules/article/space-filler.js
```diff
@@ -32,7 +28,7 @@ define([
      */
     SpaceFiller.prototype.fillSpace = function (rules, writer, options) {
         var write = (options && options.domWriter) || fastdom.write;
-        return queue.add(insertNextContent);
+        return this.queue.then(insertNextContent).catch(onError);
```
regiskuckaertz: You will also have to assign the result to `this.queue`

regiskuckaertz (changes requested):
`then` and `catch` return new promises, so you need to keep track of that.

static/src/javascripts-legacy/projects/common/modules/article/space-filler.js
gustavpursche: You are right - I didn't think about that the promises should run in the right order. 👍

sndrs (approved):

regiskuckaertz (approved):

regiskuckaertz:
Ace 👍

PR #16104, opened 2017-03-08
Remove lib/template

gu-stav:
## What does this change?

Removes `lib/template` which is only a wrapper around `lodash/utilities/template`.

Less cognitive load when reading code.


gtrufitt:
lol

gtrufitt (approved):

sndrs:
one of the values of that lib file was that it abstracted tempting away from the implementation. for instance, it didn't used to be a lodash method, if i recall. not quite a 👎 but it does mean if we decided on a different templating method, we'd have to refactor away from lodash, rather than just make the next implementation fit a signature. it's not clear to me how the cognitive load differs with this.

PR #16109, opened 2017-03-09
Rendering for tailor survey

MahanaC:
Some more changes to tailor's survey:
- CSS modifications to make sure the styling looks right, e.g. on [https://www.theguardian.com/business/2017/mar/08/nike-launches-hijab-for-female-muslim-athletes](url)
screen shot 2017-03-09 at 13 14 53
now becomes
screen shot 2017-03-09 at 14 57 48

- Change to use space filler to find location for survey, as the current method sometimes doesn't look right (I've played around with the rules and these seem ok, but if anyone has further suggestions that would be great), e.g. on [https://www.theguardian.com/business/2017/mar/08/nike-launches-hijab-for-female-muslim-athletes](url)
screen shot 2017-03-09 at 13 10 40
now becomes 
screen shot 2017-03-09 at 13 12 55

- Tracking changes to follow @JPodmore's new naming convention

@lindseydew @oilnam @GHaberis @stephanfowler @SiAdcock @mike-ruane

### DIFF
--- a/static/src/javascripts-legacy/projects/common/modules/experiments/tests/tailor-survey.js
+++ b/static/src/javascripts-legacy/projects/common/modules/experiments/tests/tailor-survey.js
@@ -135,8 +135,8 @@ define([
         }
 
         // Rules to use when finding a space for the survey
-        function getSpacefinderRules() {
-            return {
+        var spacefinderRules =
+             {
                 bodySelector: '.js-article__body',
                 slotSelector: ' > p',
                 minAbove: 0,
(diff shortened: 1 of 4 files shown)


static/src/javascripts-legacy/projects/common/modules/experiments/tests/tailor-survey.js
```diff
@@ -134,6 +134,37 @@ define([
             };
         }
 
+        // Rules to use when finding a space for the survey
+        function getSpacefinderRules() {
```
sndrs: does this need to be a function? could just be a variable?

MahanaC:
Thanks @sndrs have made that change - does it look ok to merge now? If so does someone mind merging it, as I don't have write access? Thanks

SiAdcock (approved):

PR #16113, opened 2017-03-10
Devtool: A/B Test Variant Opt-in

SiAdcock:
## What does this change?

Currently, the way to opt in to A/B test variants is to add post-`#` "parameters" to gu.com URLs

picture 197
 
These are often unwieldy, difficult to discover, even harder to remember and a nightmare to type, especially on mobile devices.

We used to have a [Chrome extension](https://chrome.google.com/webstore/detail/guardian-ab-tests/nehbenedinjacnhlkjbdneedibcagmno), but this only worked on one browser on desktop devices, and is now broken since the migration to Webpack.

This change allow us to opt into A/B test variants using a user interface.

**Desktop**

picture 198

**Mobile**

localhost-3000-environment-2016-jun-11-humble-daisy-brings-smile-face-country-diary iphone 6 plus

You can access the tool by adding `#devtools` to the end of any gu.com URL. No devtool assets are loaded unless this hash is present, so our users will never have to download / parse unnecessary assets. 

Conceptually, this was designed with extension in mind, so we can potentially add more devtools as we think of them.

- Cross browser ✅ 
- Cross device ✅ 
- Extensible
- Easy to use
- Easy to remove

No

## Request review

The code could use a tidy-up and the visual design would definitely benefit from the awesome skills of our world-class design team! It would also need some documentation.

Hot on the heels of Hack Day, what is the feeling about getting this into use? Any thoughts or opinions would be appreciated. Since it's Friday, don't forget to enable @kenlim's Beer Guard™️ 

@sndrs @gtrufitt @GHaberis @paperboyo @jfsoul @NataliaLKB @zeftilldeath @gustavpursche @michaelmcnamara @stephanfowler @TBonnin @DiegoVazquezNanini 

## Tested in CODE?

- [x] Test in CODE


static/src/javascripts-legacy/bootstraps/enhanced/main.js
```diff
@@ -188,6 +188,12 @@ define([
             }
         });
 
+        if (window.location.hash.includes('devtools')) {
```
regiskuckaertz: Is `includes` polyfilled?

SiAdcock: Good spot, probably not!

mchv:
Great work @SiAdcock ! I think it will be useful to add the link to #devtools in teleporter so it is even more discoverable.

static/src/javascripts-legacy/bootstraps/enhanced/main.js
sndrs: looks like it? https://github.com/zloirock/core-js/blob/master/fn/string/includes.js

SiAdcock: Where does that get included?

gtrufitt: https://babeljs.io/docs/usage/polyfill/

> Babel includes a polyfill that includes a custom regenerator runtime and core-js.

SiAdcock: Ah, I didn't think we were passing `javascripts-legacy` through Babel?

gtrufitt: Ah oh.

gtrufitt: Convert `main.js` to `/javascripts` 😁

SiAdcock:
```js
// TODO
```

gtrufitt: The quote I put was wrong btw, we don't polyfill like that, we do it with https://www.npmjs.com/package/babel-plugin-transform-runtime which states 

// instance methods such as "foobar".includes("foo") will not work since that would require modification of existing builtins

So it wouldn't work anyway? :man_shrugging:

sndrs: sorry, misinformation https://babeljs.io/docs/usage/polyfill/

sndrs (approved):
SHIP IT!

PR #16114, opened 2017-03-10
removing include credentials switch

michaelwmcnamara:
This removes the "AmpOphanTrackingIncludesCredentials" switch and
makes the functionality behind this switch the default behaviour.

Amp tests pass
Make validate-amp passes

@guardian/dotcom-platform


common/app/conf/switches/FeatureSwitches.scala
```diff
@@ -487,6 +487,8 @@ trait FeatureSwitches {
     exposeClientSide = true
   )
 
+
+  // Owner: Joseph Smith
   val AmpOphanTrackingIncludesCredentials = Switch(
```
SiAdcock: Can we delete the switch entirely?

michaelwmcnamara: @SiAdcock  Thanks for that - I think I messed up the conflict resolution when merging from master. It should be gone now.

SiAdcock (approved):

PR #16115, opened 2017-03-13
adds switch for recipes

tsop14:
## What does this change?

- Adds a switch for recipe articles: `ArticleWithStructuredRecipe`.
- Sets default value to `false`. 
- Uses switch to set value of `isRecipeArticle`, which determines if new template design is used and sets slim header.
- Removes `false` if statements.

Easier to turn on / off new design. 

No.

## Tested in CODE?
No

@TBonnin @NataliaLKB 

Thanks @blongden73 for help!

### DIFF
--- a/common/app/conf/switches/FeatureSwitches.scala
+++ b/common/app/conf/switches/FeatureSwitches.scala
@@ -504,7 +504,7 @@ trait FeatureSwitches {
     "changes design of articles with strucutred recipe data",
     owners = Seq(Owner.withGithub("tsop14"), Owner.withGithub("blongden73")),
     safeState = Off,
-    sellByDate = new LocalDate(2017, 12, 1),
+    sellByDate = new LocalDate(2017, 4, 4),
     exposeClientSide = true
 
   )
--- a/common/app/conf/switches/FeatureSwitches.scala
+++ b/common/app/conf/switches/FeatureSwitches.scala
@@ -497,4 +497,15 @@ trait FeatureSwitches {
     exposeClientSide = true
 
   )
+
+  val ArticleWithStructuredRecipe = Switch(
+    SwitchGroup.Feature,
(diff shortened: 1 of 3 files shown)


common/app/conf/switches/FeatureSwitches.scala
```diff
@@ -497,4 +497,15 @@ trait FeatureSwitches {
...
 
   )
+
+  val ArticleWithStructuredRecipe = Switch(
+    SwitchGroup.Feature,
+    "is-article-with-structured-recipe-data",
+    "changes design of articles with strucutred recipe data",
+    owners = Seq(Owner.withGithub("tsop14"), Owner.withGithub("blongden73")),
+    safeState = Off,
+    sellByDate = new LocalDate(2017, 12, 1),
```
TBonnin: That's pretty far away. When do you think the feature would be actually ready and tested?

TBonnin (approved):
lgtm

common/app/conf/switches/FeatureSwitches.scala
tsop14: Think it should be ready and tested starting this week. What is the convention for `sellByDate`? Would 2 months be more sensible?

TBonnin: In case of a feature switch, the convention is to pick a reasonable date where we expect the feature to be fully ready (or removed if tests are unsuccessful).
The idea behind a feature switch is to give us more control in order to enable/disable a feature while it is behind developed and tested. The `sellByDate` is here to ensure we don't sit on a half-baked feature for too long and force us to reevaluate it.

In your case I would set the date to 2-3 weeks. That would give you plenty of time to test and finalize this feature while keeping the lifetime of the switch as short as possible. 

Does it make sense?

PR #16116, opened 2017-03-13
Share recipe formatting functions

tsop14:
## What does this change?

- Moves recipe schema JSON-LD formatting logic from template to `RecipeAtom` object.
- Shares logic with the `recipeArticleBody` view. 
- Starts instructions at 1.
- Uses empty string when unit Option is `None`.

Less code, safer code (properly compiled as no longer in a template file), shared code. ✌️

No

## Tested in CODE?
No

@mchv


LATaylor-guardian:
Looks okay to me 👍

mchv:
👍

PR #16117, opened 2017-03-13
Use all testCanRun checks in aquisitions-test-selector

desbo:
## What does this change?
Currently the selector module ignores a test's switch, so disabled tests
can still be passed into ab.js. This refactor moves the canRun checks
from ab.js into their own module, which is used by both ab.js and the
selector, so the same criteria (including switch checks) is applied.

### DIFF
--- a/static/src/javascripts-legacy/projects/common/modules/experiments/acquisition-test-selector.js
+++ b/static/src/javascripts-legacy/projects/common/modules/experiments/acquisition-test-selector.js
@@ -10,7 +10,7 @@ define([
     'common/modules/experiments/tests/acquisitions-epic-design-variations'
 ], function (
     segmentUtil,
-    canRunChecks,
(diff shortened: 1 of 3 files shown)


static/src/javascripts-legacy/projects/common/modules/experiments/acquisition-test-selector.js
```diff
@@ -9,6 +10,7 @@ define([
     'common/modules/experiments/tests/acquisitions-epic-design-variations'
 ], function (
     segmentUtil,
+    canRunChecks,
```
dominickendrick: Can this be consistent with `testCanRunChecks` used in `ab.js`

dominickendrick (approved):
Apart from comment

Mullefa:
@guardian/contributions

PR #16118, opened 2017-03-14
Await upvote unset before reporting error

SiAdcock:
## What does this change?

There is currently a race condition in our recommendation tests. `unsetClicked` adds the `js-recommend-comment` class to the `target` asynchronously using `fastdomPromise`. If `reportError` throws an error before `unsetClicked` has resolved, [one of the recommendation tests](https://github.com/guardian/frontend/blob/master/static/test/javascripts-legacy/spec/common/discussion/recommendation.spec.js#L101) fails, as it falls into the `catch` callback before the class has been added. 

This is causing builds to [fail intermittently on CI](https://teamcity.gu-web.net/viewLog.html?buildId=31561&buildTypeId=dotcom_frontend&tab=buildLog)

Allowing `unsetClicked` to resolve before throwing an error makes the code more deterministic.

No

## Tested in CODE?

No

### DIFF
--- a/static/src/javascripts-legacy/projects/common/modules/discussion/upvote.js
+++ b/static/src/javascripts-legacy/projects/common/modules/discussion/upvote.js
@@ -29,9 +29,10 @@ define([
                 return setRecommended(target);
             })
             .catch(function (ex) {
-                unsetClicked(target);
-                reportError(ex, {
-                    feature: 'comments-recommend'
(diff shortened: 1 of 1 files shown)


gtrufitt (approved):
Good huntin'!

regiskuckaertz:
Argh, nice catch! It's making my build fail 💨

gustavpursche:
👍

PR #16119, opened 2017-03-14
Es6 lib closest

gtrufitt:
## What does this change?

ES6ifes, flows and jests the closest polyfill.

Some questions:

- Should this even be a util or should it be handled as a seperate babel polyfill, did we decide on the approach?
- If not, the whole `delete Element.prototype.closest` to force it into the polyfill branch feels a bit grim to me - an alternative might be [separating out the polyfill and exposing it under a test func](https://github.com/guardian/frontend/commit/d77ffefaaf5ed51050718f5166b89db1971a5bc5#diff-e2f97c5b2e1d2dffc0c326c71fdf748fR29) or creating a seperate module for the polyfill bit completely. 

Thoughts welcomed.

Thanks!


static/src/javascripts/lib/closest.spec.js
```diff
@@ -0,0 +1,60 @@
...
+            </div>
+        </div>`;
+    }
+
+    nestedChild = document.querySelector('.child');
+});
+
+afterAll(() => {
+    // Flow moans that I'm assigning something to a prototype
+    // $FlowFixMe
```
gtrufitt: Flow says 

```js
static/src/javascripts/lib/closest.spec.js:15
 15:     Element.prototype.closest = closestProto;
                           ^^^^^^^ property `closest`. Covariant property `closest` incompatible with contravariant use in
 15:     Element.prototype.closest = closestProto;
         ^^^^^^^^^^^^^^^^^^^^^^^^^ assignment of property `closest`
```

Can anyone explain what I might need to do? I read some stuff on covariance and cotravariant but my brain can't quite apply it to this...

SiAdcock: This is possibly a bug in Flow? https://github.com/facebook/flow/issues/2207

SiAdcock: Oh wait no, this is the opposite problem! I guess Flow considers `closest` to be covariant (why? Is it defined this way somewhere?). And [you can't write to covariant properties](https://flowtype.org/blog/2016/10/04/Property-Variance.html#property-variance)

gustavpursche:
> Should this even be a util or should it be handled as a seperate babel polyfill, did we decide on the approach?

I think we didn't. We could remove the util completly for now and use the Polyfill MDN suggests: https://developer.mozilla.org/en-US/docs/Web/API/Element/closest#Polyfill In `inlineJSBlocking.scala.html` we already have a few of those - IMHO that's the place where it could live then until we find a better way for all of them.

What do you think about that?

static/src/javascripts/lib/closest.spec.js
gtrufitt: Will try!

gtrufitt:
> use the Polyfill MDN suggests: https://developer.mozilla.org/en-US/docs/Web/API/Element/closest#Polyfill In inlineJSBlocking.scala.html we already have a few of those - IMHO that's the place where it could live then until we find a better way for all of them.

I think I agree but also slightly put off by the fact it's only used in a couple of files and then only in ones with a specific content type, so the extra page weight is kinda annoying (though only a tiny amount).

sndrs:
i'd be v cautious about adding more polyfills to the head of each page. only ones we need for render-clocking code should be there really...

gustavpursche:
@sndrs @gtrufitt I'm also not a big fan of the blocking polyfill concept, but that's the place where we have stuff like this at the moment. All I'm saying is, that I'd like to keep them (`classList, requestAnimationFrame, Element.matches, ...`) together for now and then think about, how we can build them into our [babel] polyfill-pipeline :)

sndrs:
i don't think babel will polyfill dom methods? so maybe we need a module that just pulls in poyfills we manually want to provide, like this one? they could be from npm or our own etc.

we could require that at the start of boot, and then we can inline only polyfills which we actually need before the main JS arrives in the head. 

effectively, i suppose, a manual polyfill.io.

in which case, maybe it would be a stop gap till we have that working reliably?

SiAdcock:
> so maybe we need a module that just pulls in poyfills we manually want to provide, like this one? they could be from npm or our own etc.

This would be my preferred option. If we install them with yarn, we won't even need to test them,  and then we can even close this PR (after we've refactored the 9 places that use `lib/closest.js`)!

gtrufitt:
Yeah, this sounds like a nice pattern for now in-between the module specific import and the inline-blocking. I'll look at refactoring, thanks all.

sndrs:
we could so something approaching this with 
```
import 'babel-polyfill';
```
then using https://github.com/babel/babel-preset-env#usebuiltins, but it's not selective - it gives you only what need, but to create a fully compliant env, regardless of what you use, for the target you specify.

our current method (using `'babel-plugin-transform-runtime'` doesn't let you use instance methods, because it explicitly won't touch the prototype (so no `"qwerty".includes('qwe')`)

doing that with a target of IE9 and nothing else creates a bundle of ~22kb gzipped. chrome 57 is only 4k though...

sndrs:
in fact, `["last 5 chrome versions",  "last 1 safari versions"]` is 4.55 kB gzipped. earlier safs, edge, ff jumps to 15kb. wonder what's bulking that up?

sndrs:
those 2 alone get us to ~65% of traffic

static/src/javascripts/lib/closest.js
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+function closest(element: ?Element, selectors: string): ?Node {
+    if (!element) return null;
+
+    if ('closest' in Element.prototype) {
+        return element.closest(selectors) || null;
+    }
+
+    let el = element.parentElement;
```
regiskuckaertz: warning: `Element.closest` checks the element itself first.

SiAdcock: Well that would technically be the closest!

gtrufitt: Oh yeahhh.. I'll probably close this for an alternative polyfill method anyway.

regiskuckaertz: Yeah, you have to think in pure logic, where the closest element of an element is the element itself. My initial polyfill had a bug where I used `parentNode` instead of `parentElement`, but I see you are using it, so good (although no need to test for `nodeType` as `Element.parentElement ≡ Element.parentNode ^ Element.parentNode.nodeType = 1`).

gtrufitt: I think I had to test for nodeType because otherwise it threw on `document.matches` when it goes all the way up the tree.

gtrufitt:
Closing this, will open a branch that replaces closest for native ready for when we're polyfill.io-ing it.

PR #16120, opened 2017-03-14
ES6: Convert `lib/report-error`

gu-stav:
## What does this change?

Converts `lib/report-error` into ES6.

ES6!


static/src/javascripts/lib/report-error.spec.js
```diff
@@ -0,0 +1,22 @@
...
+        expect(() => {
+            reportError(error, meta, false);
+        }).not.toThrowError(error);
+    });
+
+    test('Throws an error', () => {
+        expect(() => {
+            reportError(error, meta);
+        }).toThrowError(error);
+    });
```
gustavpursche: @SiAdcock @gtrufitt Should we test here aswell, whether `raven.captureException(err, { tags });` is called? If so, how would you do this?

SiAdcock: [Mock](http://facebook.github.io/jest/docs/jest-object.html#jestmockmodulename-factory-options) Raven, [spy](http://facebook.github.io/jest/docs/mock-function-api.html) on `captureException`:
```js
jest.mock('lib/fetch', () => {
  captureException: jest.fn()
});
```
Import the mocked version of Raven in the spec:
```js
import fakeRaven from 'lib/raven';
```
[Test the spy is called](http://facebook.github.io/jest/docs/expect.html#tohavebeencalledwitharg1-arg2-):
```js
expect(fakeRave.captureException).toHaveBeenCalledWith(error, meta);
```

sndrs (approved):
lgtm! ¯\_(ツ)_/¯

static/src/javascripts/lib/report-error.js
```diff
@@ -0,0 +1,32 @@
...
+};
+
+function reportError(
+    err: ReportedError,
+    tags: Object,
+    shouldThrow: ?boolean = true
+): void {
+    raven.captureException(err, { tags });
+
+    if (shouldThrow === true) {
```
SiAdcock: Is there a need to be so explicit here? Since `shouldThrow` has to be a boolean, the only possible truthy value it could take is `true`

static/src/javascripts/lib/report-error.js
```diff
@@ -0,0 +1,32 @@
...
+import raven from 'lib/raven';
+
+type ReportedError = Error & {
+    reported?: boolean,
+};
+
+function reportError(
+    err: ReportedError,
+    tags: Object,
+    shouldThrow: ?boolean = true
```
SiAdcock: As discussed, `shouldThrow?` would be a better way of indicating that the parameter is optional. `?boolean` indicates that the type of the argument is nullable, which is not strictly true

gustavpursche:
@SiAdcock Made the requested changes. LGTY now? :)

SiAdcock (approved):
🚢

PR #16121, opened 2017-03-14
Remove football-specific right slot rules

regiskuckaertz:
## What does this change?

It removes the specific length an article should have in the football section. This rule was probably there because of match reports, where there needs more space for the match stats component. The right slot is [completely disabled](https://github.com/guardian/frontend/blob/master/static/src/javascripts-legacy/projects/commercial/modules/commercial-features.js#L82) on those pages, so this test does not make sense anymore.

The DMPU unit has more value, as does the Portrait size (300x1050) in the US.

Nope

## Screenshots

Before:
picture 3

After:
picture 4

White-space in the bottom-right shows we make space for larger units.

### DIFF
--- a/static/src/javascripts-legacy/projects/commercial/modules/article-aside-adverts.js
+++ b/static/src/javascripts-legacy/projects/commercial/modules/article-aside-adverts.js
@@ -18,7 +18,6 @@ define([
     commercialFeatures
 ) {
     var minArticleHeight = 1300;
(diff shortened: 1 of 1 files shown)


regiskuckaertz:
[DMPU in the football section](https://trello.com/c/UVsRnxHd/256-dmpu-in-the-football-section)

katebee (approved):
👍 Woohoo 💎

PR #16122, opened 2017-03-14
Force into AB test before common init

SiAdcock:
## What does this change?

Previously, the enhanced bootstrap was forcing users into AB tests (using [`segmentUser`](https://github.com/guardian/frontend/blob/master/static/src/javascripts-legacy/projects/common/modules/experiments/ab.js#L370)) after most of the page JavaScript had been initialised. This meant that when opting into a particular variant using the URL hash approach, the developer had to refresh the page a second time in order to see the result of opting in.

Running `segmentUser` before common module initialisation ensures that the developer is opted into the variant for this session, and does not have to reload the page.

A smoother developer experience for opting into AB test variants.

No

## Tested in CODE?

No


gtrufitt (approved):
Lgtm.

PR #16127, opened 2017-03-14
Show Survey Overlay with Tailor

GHaberis:
## What does this change?

Show the Tailor survey overlay message If the TailorSurveyOverlay switch is ON and a user matches the following criteria:

- has a browser ID
- viewing a non-sensitive/non-commercial article page
- hasn't seen the Tailor survey overlay message in the past 100 days
- Tailor end point responds with true

- If forceShow is set in localStorage we always show survey (for testing purposes).

Allows us to collect survey responses from users

No

## Screenshots

Default message

tailor_survey_overlay

Australian message

tailor_survey_overlay_au

## Tested in CODE?

No


static/src/javascripts-legacy/projects/common/modules/tailor/tailor-survey-overlay.js
```diff
@@ -0,0 +1,115 @@
...
+        ophan.record({
+            component: 'tailor-survey-overlay',
+            value: 'impression'
+        });
+
+        cookies.add('GU_TAILOR_SURVEY_OVERLAY', 1, 100); // do not show this survey to the user for the next 100 days
+    }
+
+    function init() {
+        var bwid = cookies.get('bwid') || 'Qb6TGfeA6wRVyIkMkm7z9hMg';
```
MahanaC: Does this return the RHS if there is no cookie set? Not sure if this is what we want

static/src/javascripts-legacy/projects/common/modules/tailor/tailor-survey-overlay.js
```diff
@@ -0,0 +1,115 @@
...
+        } else {
+            return EDITION_CONFIG['default'][key];
+        }
+    }
+
+    function onSurveyAdded() {
+        ophan.record({
+            component: 'tailor-survey-overlay',
+            value: 'impression'
+        });
```
MahanaC: I'm not sure if this is sufficient to get it tracked as a lazy component, I think you might need 

`mediator.emit('register:begin', componentName);`
`... add component...;`
`mediator.emit('register:end', componentName);`

I think you might also need a `data-component` in the html, e.g. https://github.com/guardian/frontend/blob/master/static/src/javascripts-legacy/projects/common/views/experiments/tailor-survey.html

MahanaC (commented):
Looking good!

static/src/javascripts-legacy/projects/common/modules/tailor/tailor-survey-overlay.js
GHaberis: Thanks for spotting, I've removed this now. this was a dummy id.

static/src/javascripts-legacy/projects/common/modules/tailor/tailor-survey-overlay.js
GHaberis: Yeah this is something I've been trying to find examples of how this has been done elsewhere. These mediator calls will eventually just record 'register:begin' and 'register:end' with ophan in the same way I've directly recorded 'impression'. It doesn't feel quite right calling 2 events to register a single event (element added), but as this has been used elsewhere in this way I'm happy to do this here for now. I've set the data-component attribute in the JS.

MahanaC:
👍 looks great!!

static/src/javascripts-legacy/bootstraps/enhanced/main.js
```diff
@@ -191,6 +193,11 @@ define([
         // initialise email/outbrain check dispatcher
         checkDispatcher.init();
 
+        // initialise tailor overlay survey
+        if (config.switches.tailorSurveyOverlay) {
+            tailorSurveyOverlay.init();
```
gtrufitt: We try and bootstrap everything with the bootstrapContext, which will wrap your module in error reporting. I **think** as you have an init function you should just be able to do `bootstrapContext('tailorSurveyOverlay', tailorSurveyOverlay);`

static/src/stylesheets/module/tailor/_tailor-survey-overlay.scss
```diff
@@ -0,0 +1,128 @@
...
+	h1,
+	p {
+		color: $neutral-6;
+		font-size: 22px;
+		line-height: 25px;
+		margin: 12px 8px;
+		font-family: $f-serif-headline;
+	}
+
+	h1 {
```
gtrufitt: We should always try and avoid styling on elements and add classes to them

static/src/stylesheets/module/tailor/_tailor-survey-overlay.scss
```diff
@@ -0,0 +1,128 @@
...
+	top: 0;
+	left: 0;
+	width: 100%;
+	height: 100%;
+	z-index: 1050;
+	position: fixed;
+	display: block;
+}
+
+.tailor-survey-overlay .tailor-survey-background {
```
gtrufitt: You probably want to use BEM syntax here `tailor-survey__overlay` to define this as a sub-element of your `tailor-survey` block

static/src/stylesheets/module/tailor/_tailor-survey-overlay.scss
gtrufitt: The font-weight and family can probably be bought in using the font mixins in `stylesheets/pasteup/typography/_src.scss` -  `@fs-headline`

static/src/stylesheets/module/tailor/_tailor-survey-overlay.scss
```diff
@@ -0,0 +1,128 @@
...
+
+	    &:hover {
+	    	border-color: #ffffff;
+
+			.inline-cross svg {
+				fill: #ffffff;
+			}
+	    }
+
+	    &:focus {
```
gtrufitt: Could you add a focus style, maybe even just duplicate of hover for keyboard users?

static/src/javascripts-legacy/projects/common/modules/tailor/tailor-survey-overlay.js
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    function init() {
+        var bwid = cookies.get('bwid');
+        var forceShow = storage.local.get('gu.tailorSurvey.forceShow') || false;
+
+        if (bwid && shouldCallTailor(forceShow)) {
+            tailor.getSurvey(bwid, config.page.edition, forceShow)
+                .then(handleResponse)
+                .then(onSurveyAdded);
```
gtrufitt: if `showSurvey` in `handleResponse` is false, then you don't want to emit the mediator event in `onSurveyAdded`?

static/src/stylesheets/module/tailor/_tailor-survey-overlay.scss
```diff
@@ -0,0 +1,128 @@
+.tailor-survey-overlay,
+.tailor-survey-background {
+	top: 0;
+	left: 0;
+	width: 100%;
+	height: 100%;
+	z-index: 1050;
```
gtrufitt: You should be able to use `_vars` z-index vars - `$zindex-overlay` here

gtrufitt (approved):

PR #16128, opened 2017-03-14
New Header: cartoon should be in opinion

NataliaLKB:
## What does this change?
This does two things to solve the problem that cartoon doesn't fit within opinion.

First, the page id was wrong! Not sure how I missed that, but it is changed now.

The second, the blue part of the new header (news, opinion, sport, life, arts) was using slightly different logic to find the parent "pillar" than the subnav leading to inconsistencies. Those should be fixed now.

More consistent and more expected behaviour 🐛 🐛 

Nope!

## Screenshots
**Before:**
image
cartoon should show in opinion, and show the opinion subnav. This doesn't seem to be happening which is a problem.

image
This is an issue because women isn't actually a section, but just a keyword or tag page. We treat it like a section though which is why this problem is highlighted.

**After:**
image

image

## Tested in CODE?
Nope

cc @stephanfowler @guardian/dotcom-platform

### DIFF
--- a/common/app/views/fragments/newHeader.scala.html
+++ b/common/app/views/fragments/newHeader.scala.html
@@ -19,7 +19,7 @@
                 ) { editionId =>
                     <a class="header__supporter-cta"
                         data-link-name="nav2 : supporter-cta"
-                        data-edition=@{editionId}
+                        data-edition="@{editionId}"
                         href="@{Configuration.id.membershipUrl}/@{editionId}/supporter?INTCMP=mem_@{editionId}_web_newheader_trapezoid">
 
                         Become a <span>Supporter</span>
--- a/common/app/views/fragments/newHeader.scala.html
+++ b/common/app/views/fragments/newHeader.scala.html
@@ -4,48 +4,50 @@
 @import common.{LinkTo, NewNavigation, Edition}
 
 <header class="new-header" role="banner">
-    @defining(
-        NewNavigation.SectionLinks.getTopLevelSection(page.metadata.sectionId)
-    ) { case (currentTopLevelSection) =>
-        <div class="new-header__inner gs-container">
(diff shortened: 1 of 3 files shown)


common/app/views/fragments/newHeader.scala.html
```diff
@@ -4,48 +4,50 @@
...
-            @defining(
-                Edition(request).id.toLowerCase()
-            ) { editionId =>
-                <a class="header__supporter-cta"
-                    data-link-name="nav2 : supporter-cta"
-                    data-edition=@{editionId}
-                    href="@{Configuration.id.membershipUrl}/@{editionId}/supporter?INTCMP=mem_@{editionId}_web_newheader_trapezoid">
-
-                    Become a <span>Supporter</span>
+    @defining(NewNavigation.SubSectionLinks.getSectionOrTagId(page)) { id =>
```
NataliaLKB: The diff is really confusing, but I only added this line and the closing }.

common/app/views/fragments/newHeader.scala.html
```diff
@@ -4,48 +4,50 @@
...
-                    <a class="new-header__nav__link @if(link.title == currentTopLevelSection) {section-indicator}"
-                        href="@LinkTo(link.url)"
-                        data-link-name="nav2 : primary : @link.title">
-                            @link.title
+                @defining(
+                    Edition(request).id.toLowerCase()
+                ) { editionId =>
+                    <a class="header__supporter-cta"
+                        data-link-name="nav2 : supporter-cta"
+                        data-edition=@{editionId}
```
SiAdcock: Not sure how important, but do we need quotes around this?

NataliaLKB: It works, but I think it should have quotes. Thanks you! Fixed now :)

SiAdcock (approved):
LGTM 👍

PR #16129, opened 2017-03-14
Enforce committers to fix sass-lint warnings

gu-stav:
## What does this change?

Displays output of sass-lint warnings during `make validate-sass` and aborts the `pre-push` hook, in case of warnings, to enforce a better SASS style.

We are linting our sass-files properly (again) and improving the quality of our sass code.

No.


sndrs (approved):
👍 for code, let's see if anyone has a strong feeling against it?

PR #16130, opened 2017-03-14
ES6 - lib/config

gtrufitt:
## What does this change?

ESifies config.

Depends on #16176


static/src/javascripts/lib/config.js
```diff
@@ -0,0 +1,44 @@
+// @flow
+
+import pad from 'lib/pad';
+// eslint-disable-next-line guardian-frontend/global-config
```
sndrs: maybe a note saying this is the only place it should be used and why?

static/src/javascripts/lib/config.js
```diff
@@ -0,0 +1,44 @@
...
+            return null;
+        },
+
+        // returns 2014/apr/22
+        dateFromSlug() {
+            const s = this.page.pageId.match(/\d{4}\/\w{3}\/\d{2}/);
+            return s ? s[0] : null;
+        },
+
+        isMedia: ['Video', 'Audio'].indexOf(config.page.contentType) > -1,
```
sndrs: maybe a little `.some()` in here :) nbd

static/src/javascripts/lib/config.js
```diff
@@ -0,0 +1,44 @@
+// @flow
+
+import pad from 'lib/pad';
```
sndrs: babel will pollyfill `padStart` if you want that?

static/src/javascripts/lib/config.js
```diff
@@ -0,0 +1,44 @@
...
+// eslint-disable-next-line guardian-frontend/global-config
+const config = window.guardian.config;
```
sndrs: feel like we should check the objects we're merging here will not overwrite each other... maybe that's out of the scope of this pr?

static/src/javascripts/lib/config.js
sndrs: ☹️ no it won't

> NOTE: Instance methods such as "foobar".includes("foo") will not work since that would require modification of existing builtins (Use babel-polyfill for that).

https://babeljs.io/docs/plugins/transform-runtime/

gtrufitt: Yeah, just read that :/ I guess we need to be clear about what is polyfilled and what isn't - ideally everything (maybe instance methods get polyfilled in the global scope everything else with this)

static/src/javascripts/lib/config.js
```diff
@@ -0,0 +1,44 @@
...
+
+export default Object.assign(
+    {
+        hasTone(name) {
+            return (this.page.tones || '').indexOf(name) > -1;
```
regiskuckaertz: general code style question: for simple expressions like this one, could we opt for the arrow syntax? i.e.
```
hasTone: name => (this.page.tones || '').indexOf(name) > -1
```
feels a bit more readable.

sndrs: this would done automatically by `fix` if it wasn't an object property, so good case for it

gtrufitt: 👍

static/src/javascripts/lib/config.js
gtrufitt: Hmm, you mean where `Object.assign(config, allTheNewMethodsHere)` happens, make sure that we're not overriding properties from config?

sndrs (approved):
lgtm!

PR #16133, opened 2017-03-14
Adding cartoon front, as well as the tag page to the new nav logic

NataliaLKB:
## What does this change?
I made this change: https://github.com/guardian/frontend/pull/16128/commits/c90a29228d79a7fb0507d06131d2d736b5bc03c4

But it turns out cartoons locally (run by applications) has the page id of `type/cartoons`, but in PROD has the page id of `cartoons/archive`. This adds that page id back so the right Nav should show up on that page.

But when I run Facia and go to the cartoon page, I get nothing. Why is PROD and DEV different? 😕 

Maybe this will work now?

Nope

## Tested in CODE?
Nope

### DIFF
--- a/common/app/common/NewNavigation.scala
+++ b/common/app/common/NewNavigation.scala
@@ -289,6 +289,7 @@ object NewNavigation {
       SectionsLink("australia-news/indigenous-australians", indigenousAustralia, News),
 
       SectionsLink("commentisfree", opinion, Opinion),
+      SectionsLink("cartoons/archive", cartoons, Opinion),
       SectionsLink("type/cartoon", cartoons, Opinion),
       SectionsLink("index/contributors", columnists, Opinion),
       SectionsLink("commentisfree/series/comment-is-free-weekly", inMyOpinion, Opinion),
@@ -491,7 +492,8 @@ object NewNavigation {
       "football/competitions",
       "football/results",
       "football/fixtures",
-      "type/cartoon"
+      "type/cartoon",
+      "cartoons/archive"
     )
 
     def getSectionOrTagId(page: Page) = {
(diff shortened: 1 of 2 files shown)


SiAdcock (approved):
Nice! Worth testing on CODE (on the off-chance that it has the the same config as PROD)?

paperboyo:
Hi, so this is a little rabbit hole this is... I think this is an old philosophy (category) error resulting in a practical error being then patched all over the place differently... Maybe better to finally fix it properly than to apply another patch (I've tried long time ago, but failed).

Cartoons used to be a `type`, now they are a `tone`:
gu.com/type/cartoon (containing new cartoons) redirects to gu.com/cartoons/archive, but then its own All today's stories link (gu.com/cartoons/archive/all) points to a page when the latest item is from 11 April 2015 (the last one containing both `id: "type/cartoon", type: "type"` & `id: "tone/cartoons",type: "tone"` not just the latter).

IOW, all cartoons have both `id`s, but the new ones have only a `tone` `id`.

Oh and to add insult to injury, some cartoons are created by bespoke tools and are of `type/interactive`. Luckily, they also have `tone/cartoons`. There is this one [pesky Australian](https://www.theguardian.com/profile/first-dog-on-the-moon) who is of `type/article` and of `tone/comment`, so even though he features on gu.com/cartoons/archive, he can't ever be included in gu.com/tone/cartoons, because a piece can't be of two `tone`s or `type`s (I think). He is very, very funny, though...

NataliaLKB:
@paperboyo we have so many things like that on the website 😞 . That is very confusing! Having all these inconsistencies definitely adds complexity, but it is very difficult to fix. Certainly can't do it to hold up getting the New Header out, but its good to be aware of.

PR #16134, opened 2017-03-14
Use Json-Ld format for live blog structured data instead of microdata [do not merge]

MatthewJWalls:
## What does this change?

This replaces the microdata format for Live Blogs with Json-Ld.

This has been done minimally as possible by simply removing the itemscope/itemtype tags off of the core article html element, and instead we prefix the element with the json-ld content. 

Everything inside of the article element still retains all of it's original microdata properties. This was left unchanged because template reuse makes a "big bang" approach of changing all out structured data at once very difficult. The structured data/AMP validators seems fine with this.

note: since we are essentially duplicating the content as metadata in json, this will bloat the page size more than microdata did!

This change is a requirement for us to support the upcoming Google indexing API. It's likely that we'll want to convert everything else to json-ld at some point as well.

Affects AMP


stephanfowler:
@guardian/dotcom-platform - could someone review this today please? Keen to see this released.

article/app/views/liveblog/liveBlogBodyContentAMP.scala.html
```diff
@@ -28,6 +29,11 @@
         @if(model.currentPage.currentPage.isArchivePage) {
             @fragments.liveBlogNavigation(article.content.metadata.id, model.currentPage, amp = true)
         }
+
+            <script type="application/ld+json">
```
NataliaLKB: Why wouldn't this go into https://github.com/guardian/frontend/blob/master/common/app/views/fragments/metaData.scala.html instead of in the template for the live blog?

article/app/views/liveblog/liveBlogBodyContentAMP.scala.html
```diff
@@ -28,6 +29,11 @@
...
+            @Html(LiveBlogPosting(model).toString())
```
gtrufitt: This is AMP only? Does it need to be on the main liveblog template too?

article/app/views/support/structuredData/BlogPosting.scala
```diff
@@ -0,0 +1,59 @@
...
+        case None => Organisation()
+      }
+
+    }
+
+    Json.obj(
+      "@type" -> "BlogPosting",
+      "headline" -> block.title.getOrElse[String](blog.article.trail.headline),
+      "author" -> blockAuthor(blog, block),
+      "publisher" -> Organisation(),
```
gtrufitt: Can this be at the level above instead of repeating in each block?

article/app/views/liveblog/liveBlogBodyContentAMP.scala.html
MatthewJWalls: It is not AMP only, so I think I made a mistake here. I originally thought it would be, but that was before I saw how we'd implemented the microdata. It looks like I missed the template @NataliaLKB mentioned above so I should probably move it there. Will double check!

gtrufitt (commented):
As far as the move from metadata -> json_ld goes, this lgtm bar minor comments. I'll let someone more scala-y comment on structure (it did lgtm too though).

article/app/views/liveblog/liveBlogBodyContentAMP.scala.html
MatthewJWalls: I think you're most definitely right, I've put it in the wrong place. Will double check and look at fixing shortly!

article/app/views/support/structuredData/BlogPosting.scala
MatthewJWalls: Unfortunately in order for the validator to pass, the publisher field has to be duplicated in each block (As each update of a live-blog can technically be authored by different authors/organisations).

gtrufitt: :man_facepalming: 

So just reading about it, I think you can maybe do this (tested it in the validator with their example and it seems to link it, (but there's so many errors in the original example...), like this:

```
"publisher": {
      	"@id": "http://awesometechblog.com/2015/03/09/acme_spring_forward/#publisher"
      },
```
and use that as the id of a different `json+ld` block:
```
 "@id": "http://awesometechblog.com/2015/03/09/acme_spring_forward/#publisher",
```

Full example - I only added publisher to the first block:

```json
<script type="application/ld+json">
{
    "@id": "http://awesometechblog.com/2015/03/09/acme_spring_forward/#publisher",
    "@context": "http://schema.org",
    "@type": "Organization",
    "name": "The Guardian"
    }
</script>
<script type="application/ld+json">
{
  "@context": "http://schema.org",
  "@type": "LiveBlogPosting",
  "url": "http://awesometechblog.com/2015/03/09/acme_spring_forward",
  "about": {
    "@type": "Event",
    "startDate": "2015-03-09T13:00:00-07:00",
    "name": "Acme Spring Forward Event",
    "location": {
      "@type": "EventVenue",
      "name": "Jane Gardner",
      "address" :{
        "@type": "PostalAddress",
        "streetAddress": "701 Mission St",
        "addressLocality": "San Francisco",
        "addressRegion": "CA",
        "postalCode": "94103",
        "addressCountry": "US"
      }
    }
  },
  "coverageStartTime": "2015-03-09T11:30:00-07:00",
  "coverageEndTime": "2015-03-09T16:00:00-07:00",
  "headline": "Acme Spring Forward Event Live Blog",
  "description": "Welcome to Foo's live coverage of the Acme Spring Forward …",
  "liveBlogUpdate": [
    {
      "@type": "BlogPosting",
      "headline": "Coming this April, HBO NOW will be available exclusively in the U.S. on Acme TV and the App Store.",
      "url": "http://awesometechblog.com/2015/03/09/acme_spring_forward#post3",
      "datePublished": "2015-03-09T13:08:00-07:00",
      "publisher": {
      	"@id": "http://awesometechblog.com/2015/03/09/acme_spring_forward/#publisher"
      },
      "articleBody": {
        "@value": "It's $14.99 a month. <br> And for a limited time, only $9.99",
        "@type": "rdf:HTML"
      }
    },
    {
      "@type": "BlogPosting",
      "headline": "iPhone is growing at nearly twice the rate of the rest of the smartphone market.",
      "datePublished": "2015-03-09T13:13:00-07:00",
"image": "http://images.acme.com/live/2015-mar-event/images/573cb_xlarge_2x.jpg"
    },
    {
      "@type": "BlogPosting",
      "headline": "See the new flagship Acme Retail Store in West Lake, China.",
      "url": "http://awesometechblog.com/2015/03/09/acme_spring_forward#post1",
      "datePublished": "2015-03-09T13:17:00-07:00",
      "video": {
        "@type":"VideoObject",
"thumbnail": "http://images.acme.com/live/2015-mar-event/images/908d2e_large_2x.jpg"
      }
    }
  ]
}
</script>
```

gtrufitt:
Are we able to strip whitespace out of this to save some bytes?

article/app/views/liveblog/liveBlogBodyContentAMP.scala.html
MatthewJWalls: @NataliaLKB @gtrufitt The coding for this turned out to be a bit messy - since Metadata is in common, it doesn't have access to the data structures used to decide the range of blocks to present on the page that are coded into the article module, and I don't really want to duplicate the code if I can help it

MatthewJWalls: Talked to Jon and Diego and the consensus is to extract it all into common for now.

MatthewJWalls:
We're going to have to revisit this after this PR is closed because extracting the relevant classes into Metadata/Head has made a bit of a mess. The structured data does belongs here, but it's fractured a lot of the article-specific logic to now be scattershot across two places.

Pushed a change for Gareth's id linking, plus the move into metadata

PR #16135, opened 2017-03-14
Avoid the side swipe

gtrufitt:
## What does this change?

Byline image positioning was forcing the body wider than viewport, causing side swiping possible on mobile.

NO SIDE SWIPES.

No

### Before

SIDE SWIPE

screen shot 2017-03-14 at 17 32 31

### After

NO SIDE SWIPE

screen shot 2017-03-14 at 17 32 36

## Tested in CODE?

No sir.

### DIFF
--- a/static/src/stylesheets/module/content/_content.scss
+++ b/static/src/stylesheets/module/content/_content.scss
@@ -263,6 +263,8 @@
 }
 
 .content__head--byline-pic {
+    overflow: hidden;
+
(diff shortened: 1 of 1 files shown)


zeftilldeath (approved):
Looks good. Still don't know how this has only just started happening?

gtrufitt:
Maybe it didn't? It's only on pages with byline in the new header I think

PR #16138, opened 2017-03-15
Explicitely add right slot

regiskuckaertz:
Adding the right-hand slot is totally asynchronous, and so depending on the host environment and network conditions, we may run into a race condition where the right-hand slot is added _after_ `fillAdvertSlots` gets executed and is thus left ignored. To solve this, we explicitly call `addSlot`, just like we do in `articleBodyAdverts`.


regiskuckaertz:
[Avoid right slot race condition](https://trello.com/c/GFcKLlyP/259-avoid-right-slot-race-condition)

katebee (approved):
👍 💎

PR #16141, opened 2017-03-15
Support storyquestions atom

LATaylor-guardian:
## What does this change?
This change will render story question atoms when present on articles.

This allows journalists to place questions they think may want answering by readers on articles and allows readers to upvote which of the questions they most want answering. This data will be fed into Ophan and allow journalists to make informed decisions about what content they commission off of the back of this information.

Amp (I have excluded the atom from showing on AMP so that it doesn't fail the validation as it requires js. We will implement this on AMP at a later date).

## Screenshots
picture 42

Thanks @blongden73 for doing this!

## Tested in CODE?
No.


LATaylor-guardian:
@NataliaLKB May you review please?

NataliaLKB:
Does this conflict with the outbrain restrictions at all?

common/app/views/fragments/atoms/atom.scala.html
```diff
@@ -1,13 +1,14 @@
...
 @import _root_.model.ShareLinkMeta
-@import _root_.model.content.{InteractiveAtom, MediaAtom, Quiz}
+@import _root_.model.content.{InteractiveAtom, MediaAtom, Quiz, StoryQuestionsAtom}
 
 @{
     model match {
         case quiz: Quiz => views.html.fragments.atoms.quiz(quiz, maybeResults = None, showResults = false, ShareLinkMeta(Nil, Nil))
         case media: MediaAtom => views.html.fragments.atoms.media(media = media, amp = amp, displayCaption = true, displayEndSlate = mediaWrapper.contains(MediaWrapper.MainMedia), mediaWrapper = mediaWrapper)
         case interactive: InteractiveAtom => views.html.fragments.atoms.interactive(interactive, shouldFence)
+        case storyquestions: StoryQuestionsAtom => views.html.fragments.atoms.storyquestions(storyquestions, isAmp = amp)
```
NataliaLKB: I think you want `amp = amp` here

common/app/views/fragments/atoms/storyquestions.scala.html
```diff
@@ -0,0 +1,26 @@
+@(storyquestions: model.content.StoryQuestionsAtom, isAmp: Boolean)(implicit request: RequestHeader)
+
+@if(!isAmp) {
+    <div class="submeta user__question">
```
NataliaLKB: How does this look with the submeta already in the page? Would you be able to add a screenshot?

LATaylor-guardian:
@NataliaLKB What outbrain restrictions? 😆

common/app/views/fragments/atoms/storyquestions.scala.html
```diff
@@ -0,0 +1,26 @@
...
+    <span class="js-storyquestion-atom-id" style="display: none" id="@storyquestions.id"></span>
```
NataliaLKB: Could you use the `is-hidden` class instead https://github.com/guardian/frontend/blob/master/static/src/stylesheets/base/_helpers.scss#L63

common/app/views/fragments/atoms/storyquestions.scala.html
LATaylor-guardian: picture 43

common/app/views/fragments/atoms/storyquestions.scala.html
```diff
@@ -0,0 +1,26 @@
...
+    @for(questions <- storyquestions.data.editorialQuestions) {
+        @for(qs <- questions) {
+            <p>
+                <div class="user__question-section">
+                  <h3>@qs.questionSetTitle</h3>
+                </div>
+                @for(question <- qs.questions) {
+                    <p class="user__questions-text--wrapper">
+                        <span id="js-question-text-@question.questionId" class="user__question-text"><span>@question.questionText</span></span>
+                        <a href="#" id="@question.questionId" class="user__question-upvote js-ask-question-link">Ask this <span class="tick">@fragments.inlineSvg("tick", "icon")</span></a>
```
NataliaLKB: Nit: for readability, could you have what is inside this link on a separate line. What about: 

```
<a href="#" id="@question.questionId" class="user__question-upvote js-ask-question-link">
    Ask this 
    @fragments.inlineSvg("tick", "icon")
</a>
```

Notice I removed the span as well because  the icon will already have a span with the class `tick` why next two of them?

static/src/stylesheets/module/atoms/_storyquestions.scss
```diff
@@ -0,0 +1,139 @@
+.user__question {
+    padding-bottom: $gs-baseline;
+    overflow: auto;
+}
+.user__question-title {
+    font-weight: 900;
+    font-size: 22px;
```
NataliaLKB: A font mixin should be used here instead. Use one from here: https://github.com/guardian/frontend/blob/master/static/src/stylesheets/pasteup/typography/_src.scss

static/src/stylesheets/module/atoms/_storyquestions.scss
```diff
@@ -0,0 +1,139 @@
...
+    line-height: 24px;
+    color: $news-main-1;
+    font-family: $f-serif-headline;
+
+    @include mq($from: tablet) {
+        font-size: 26px;
+        line-height: 28px;
+    }
+
+    span {
```
NataliaLKB: Does this span have a class? Best to use the class and not nest instead.

static/src/stylesheets/module/atoms/_storyquestions.scss
```diff
@@ -0,0 +1,139 @@
...
+            color: #ffffff;
+
+            span {
+                background-color: $news-main-1;
+            }
+        }
+        .tick {
+            background-color: $news-main-1;
+
+            svg {
```
NataliaLKB: Best to add a more specific class to the svg. It will be added to the span and then the svg will be given the same class with `__svg` after it. For example, if the span has the class `user__tick-icon` or something similar, then the class on the svg would be `user__tick-icon__svg`. It is good to avoid nesting in css

static/src/stylesheets/module/atoms/_storyquestions.scss
```diff
@@ -0,0 +1,139 @@
...
+    }
+}
+
+.user__question-section {
+    border-top: 1px dotted $neutral-6;
+    margin-top: $gs-baseline * 2;
+
+    h3 {
+        color: $neutral-2;
+        font-family: $f-sans-serif-text;
```
NataliaLKB: Use the font mixin anytime you reference a font

common/app/views/fragments/atoms/storyquestions.scala.html
NataliaLKB: Thanks!

NataliaLKB:
> What outbrain restrictions?

Good question! So we have a very strict contract with outbrain and if we don't follow the rules we don't get money from those page views. Putting containers above outbrain is an example of not following the rules. Best to talk to someone from @guardian/commercial-dev. Maybe @ScottPainterGNM ?

LATaylor-guardian:
@NataliaLKB this is part of the content as it's an atom and is not a new container so I don't believe its a problem.

NataliaLKB:
It is definitely worth double checking. Things like the email signup (even less container looking) is something that conflicted in the past. I just want to make sure that we won't have an unexpected loss of money first before we merge this.

regiskuckaertz:
Hello - Natalia's right, _anything_ that pushes the Outbrain container further from the bottom of the article body is considered a breach of contract. In such cases, we have "non-compliant" Outbrain impressions, OB gives us less money for those, but we first need to estimate how many impressions we are talking about and the corresponding revenue loss. I would not merge without the approval of Catherine Murray in the AdOps team, which is ultimately the one deciding those matters.

LATaylor-guardian:
@regiskuckaertz @NataliaLKB Thanks, we're speaking to Catherine Murray now about this.

common/app/views/fragments/atoms/storyquestions.scala.html
```diff
@@ -0,0 +1,33 @@
...
+              <h3>@qs.questionSetTitle</h3>
+            </div>
+            @for(question <- qs.questions) {
+                <p class="user__quesions-text--wrapper">
+                    <span id="js-question-text-@question.questionId" class="user__question-text">
+                      <span class="user__questions-text--inner">
+                        @question.questionText
+                      </span>
+                    </span>
+                    <a href="#" id="@question.questionId" class="user__question-upvote js-question-upvote-link">
```
gustavpursche: A link without an `href` should be a `<button>` :)

common/app/views/fragments/atoms/storyquestions.scala.html
```diff
@@ -0,0 +1,33 @@
...
+
+    <h1 class="user__question-title">Finished this article but want to know more? <span class="user__question-title--secondary">Let us know what questions you want answered.</span></h1>
+@for(questions <- storyquestions.data.editorialQuestions) {
+    @for(qs <- questions) {
+        <p>
+            <div class="user__question-section">
+              <h3>@qs.questionSetTitle</h3>
+            </div>
+            @for(question <- qs.questions) {
+                <p class="user__quesions-text--wrapper">
```
gustavpursche: > The P element represents a paragraph. It cannot contain block-level elements (including P itself). [source](https://www.w3.org/TR/html401/struct/text.html#h-9.3.1)

NataliaLKB: Should be a `div` in this case

common/app/views/fragments/atoms/storyquestions.scala.html
```diff
@@ -0,0 +1,33 @@
+@(storyquestions: model.content.StoryQuestionsAtom, isAmp: Boolean)(implicit request: RequestHeader)
+
+@if(!isAmp) {
+    <div class="submeta user__question">
+    <span class="js-storyquestion-atom-id is-hidden" id="@storyquestions.id"></span>
+
+    <h1 class="user__question-title">Finished this article but want to know more? <span class="user__question-title--secondary">Let us know what questions you want answered.</span></h1>
```
gustavpursche: IMHO this should never be a `<h1>`, but only the article-headline, but I'm not sure how we deal with atoms and headline structure to be honest.

project/Dependencies.scala
```diff
@@ -32,7 +32,7 @@ object Dependencies {
   val commonsLang = "commons-lang" % "commons-lang" % "2.4"
   val commonsIo = "commons-io" % "commons-io" % "2.4"
   val cssParser = "net.sourceforge.cssparser" % "cssparser" % "0.9.21"
-  val contentApiClient = "com.gu" %% "content-api-client" % "11.3"
+  val contentApiClient = "com.gu" %% "content-api-client" % "11.8"
```
NataliaLKB: Have you checked a front loads fine running locally?

gustavpursche:
@LATaylor-guardian Just a few thoughts on the layout:

- I wouldn't make the sub-headline so big. It's more of an explaining text, so maybe it could follow the paragraph styling from the article.
- Why is there a gray background behind the questions? It reduces the contrast and makes it harder to read.
- I'd make the question itself clickable. They are already blue and look somehow like a link to me. Alternative could be to not make them blue, but black-ish.
- I guess it would be easier, to move the `ask-this` button to the left, so it's easier to deal with the case of multiple lines? It would, maybe, make the borders obsolete, since thei are closer to the question.
- A nice effect for the "Voted state" could be: Only remove "Ask this" and make the "check" green.

Just wanted to get this out of my head :)

🌷

LATaylor-guardian:
Thanks for your feedback @gustavpursche. I didn't do any of the design work on this so am simply happy for whomever is the design authority (and Its @blongden73 in this case) to make those decisions.

blongden73:
@gustavpursche Thanks for your thoughts :-)

There are a couple of things with this. The headline style needs to feel like its not part of the article. I think keeping the size of the heading the same would make it feel like it was still within the article - esspecially as this is to sit above the social media buttons and tags etc. It need to be big enough especially at first to catch peoples eyes. 

If you are looking at just from the screenshots in the PR - it does make it harder to read but actually in the page it's fine. The thinking behind it is to highlight them to make it feel like these are questions that have been written - it adds a bit of visual interest it also makes them feel more important than if they were just given a treatment that blended into the style of the article.

If the ask this sat on the left you end up with either the ask this sitting in different places if it hangs of the end of the line. Or if it sits before the line starts its says ask this before you've read the question. 

I think you're right about the voted state having the tick remain would be better! 

:-)

mchv:
Thanks @regiskuckaertz and @NataliaLKB! Following your useful comments we have asked nath. and chris w. to discuss this with the appropriate people and ensure to have the green light before it can get merged. We will keep you updated

static/src/stylesheets/module/atoms/_storyquestions.scss
```diff
@@ -0,0 +1,139 @@
...
+    // font-size: 18px;
+    float: left;
+    color: $news-main-1;
+    padding-left: $gs-baseline / 4;
+    position: relative;
+    margin-bottom: $gs-baseline;
+
+    @include mq($from: tablet) {
+        max-width: 80%;
+        @include fs-headline(3);
```
NataliaLKB: This shouldn't pass the css linter. Mixins should go before declarations (at the top of a block)

blongden73: @NataliaLKB oooo my bad - yeah i got and error and fixed this maybe didnt do the second commit.

LATaylor-guardian:
@NataliaLKB @regiskuckaertz The decision has been made to call non compliant on this. Moving forward are there any specific changes that need to be made to this PR in light of this? Many thanks.

regiskuckaertz:
@LATaylor-guardian that seems like the best move going forward. You will have to flick the 'non-compliant' switch on, which @GHaberis knows all about as he's refactoring that right now in #16208

GHaberis:
Hi @LATaylor-guardian, the checks which determine whether outbrain is compliant/non-compliant are defined in the check-mediator module and dispatched by the check-dispatcher module. A new check that returns whether this new component is present will need adding to these modules. As @regiskuckaertz mentioned I'm about to merge a PR which touches this logic so I'd hold until then.  I'll let you know when this is done, and am happy to assist.

LATaylor-guardian:
Okay thanks @GHaberis, yes please let me know :)

LATaylor-guardian:
@GHaberis Thank you very much for your help. Please see our changes for making outbrain non compliant. @NataliaLKB How is this looking now, are we good to go?

common/app/views/fragments/atoms/storyquestions.scala.html
```diff
@@ -0,0 +1,33 @@
+@(storyquestions: model.content.StoryQuestionsAtom, isAmp: Boolean)(implicit request: RequestHeader)
+
+@if(!isAmp) {
+    <div class="submeta user__question">
+        <span class="js-storyquestion-atom-id is-hidden" id="@storyquestions.id"></span>
+        <h2 class="user__question-title">Finished this article but want to know more? <span class="user__question-title--secondary">Let us know what questions you want answered.</span></h2>
+        @for(questions <- storyquestions.data.editorialQuestions) {
+            @for(qs <- questions) {
+                <p>
```
NataliaLKB: Nit: indentation is weird. Could we straighten this out?

LATaylor-guardian: Done.

static/src/stylesheets/module/atoms/_storyquestions.scss~a50954f50b9fb4759c3c14b1246094d52299d760
```diff
@@ -0,0 +1,137 @@
+.user__question {
```
NataliaLKB: What is this file? I don't think this should be committed?

LATaylor-guardian: Don't know how that got there - deleted.

NataliaLKB (approved):

common/app/model/Formats.scala
```diff
@@ -198,9 +198,31 @@ object ContentTypeFormat {
   implicit val genericThriftAtomFormat = GenericThriftAtomFormat
   implicit val recipeThriftAtomFormat = RecipeThriftAtomFormat
   implicit val reviewThriftAtomFormat = ReviewThriftAtomFormat
+  implicit val storyquestionsThriftAtomFormat = StoryquestionsThriftAtomFormat
   implicit val recipeAtomFormat = Json.format[RecipeAtom]
   implicit val reviewAtomFormat = Json.format[ReviewAtom]
-  implicit val atomsFormat = Json.format[Atoms]
+  implicit val storyquestionsAtomFormat = Json.format[StoryQuestionsAtom]
```
NataliaLKB: cc @TBonnin to get your thoughts. This change would mean we don't have to repress fronts just to add a field for an atom.

I am happy with this change, but we have added it to our team's goals to understand and potentially come up with a better architecture for atoms in frontend, as adding it as part of facia seems a bit hacky.

common/app/model/Formats.scala
```diff
@@ -423,6 +445,11 @@ object ReviewThriftAtomFormat extends Format[com.gu.contentatom.thrift.atom.revi
   def writes(review: com.gu.contentatom.thrift.atom.review.ReviewAtom) = JsObject(Seq.empty)
 }
 
+object StoryquestionsThriftAtomFormat extends Format[com.gu.contentatom.thrift.atom.storyquestions.StoryQuestionsAtom] {
+  def reads(json: JsValue) = JsError("Converting from Json is not supported by intent!")
+  def writes(review: com.gu.contentatom.thrift.atom.storyquestions.StoryQuestionsAtom) = JsObject(Seq.empty)
```
TBonnin: Is that the correct parameter name? 😉

common/app/model/Formats.scala
```diff
@@ -198,9 +198,31 @@ object ContentTypeFormat {
...
+  implicit val reads: Reads[Atoms] = (
+    (__ \ "quizzes").read[Seq[Quiz]].orElse(Reads.pure(Nil)) and
+      (__ \ "media").read[Seq[MediaAtom]].orElse(Reads.pure(Nil)) and
+      (__ \ "interactives").read[Seq[InteractiveAtom]].orElse(Reads.pure(Nil)) and
+      (__ \ "recipes").read[Seq[RecipeAtom]].orElse(Reads.pure(Nil)) and
+      (__ \ "reviews").read[Seq[ReviewAtom]].orElse(Reads.pure(Nil)) and
+      (__ \ "storyquestions").read[Seq[StoryQuestionsAtom]].orElse(Reads.pure(Nil))
+    )(Atoms.apply _)
+
+
```
TBonnin: Would it make more sense to make the `Atoms` fields optional? 
That's the representation we are trying to model here?

common/app/model/Formats.scala
LATaylor-guardian: Will fix now.

common/app/model/Formats.scala
LATaylor-guardian: I don't mind - if you feel strongly about it then we can refactor in a later PR but I'd prefer to get this out as I need to move on.

TBonnin: Yes. merge it. I don't feel strongly about it neither. I would like to explore making thos `Atoms` field optional in the future. I might miss the reason why it wouldn't make sense

LATaylor-guardian: Great thanks! I think they are Seq's because an Option[Seq] feels odd when you can represent the None case as an empty Seq. It also makes working with them from there on in a little nicer.

TBonnin: Agreed. In term of data, None would mean the field is not present, Seq.empty means field is present but doesn't contain any atom. (I give you that from the app perspective this is the same)
I don't think models should necessary match how they are represented in json. I am just throwing an idea (and its associated drawbacks) that you summarise well: Having to define and maintain our own `Reads` vs handling optional fields when working with `Atoms`.

Thanks for your work anyway. It looks like we can improve the way we support Atoms in general in the frontend codebase.

LATaylor-guardian: Yeah sure, happy to help with this in any way I can.

PR #16145, opened 2017-03-15
Reading time demand test

lmath:
## What does this change?
We are going to run an A/B test where 1.45% of users visiting the network front will be invited to pick an option for how much time they have to read, and we will suggest something they can read in that timeframe. 

Test just to see if users are interested in factoring reading time when determining what to read. We will just count how many users click a time when they see the thrasher. 

If enough users click to tell us how much time they have, we will know to try more sophisticated ways to use reading time to suggest content to users. 

## Screenshots
image
image

image
image

image
image

## Tested in CODE?

This was heavily inspired by the securedrop thrasher (https://github.com/guardian/thrashers/pull/103/files) which looks quite similar. If there's some obvious way for me to do this with less code duplication, let me know!


static/src/javascripts-legacy/projects/common/modules/experiments/tests/reading-time.js
```diff
@@ -78,37 +75,24 @@ define([
...
         }
 
-        function replaceWithLink(suggestionHtml){
-            $('.cta').empty();
-            $('.cta').html(suggestionHtml);
-            $('.selections').empty();
-        }
+        function ctaReplacer(suggestion, completeFunction) {
+            return function () {
+                $('.cta').empty();
```
davidfurey: I think this might be redundant because of the following line

static/src/javascripts-legacy/projects/common/modules/experiments/tests/reading-time.js
```diff
@@ -0,0 +1,100 @@
...
+
+        this.canRun = function () {
+            $sectionBelowReadingTime = $('#opinion');
+            return config.page.contentType === 'Network Front' && $sectionBelowReadingTime.length;
+        };
+
+        this.variants = [
+            {
+                id: 'static-links',
+                success: function(complete) {
```
davidfurey: I think this could just be `success: insertReadingTimeSelection`

davidfurey (commented):
Looks good to me, but I'll let someone with some html/css knowledge +1

static/src/stylesheets/module/experiments/_reading-time.scss
```diff
@@ -0,0 +1,141 @@
+#reading-time.fc-container--thrasher {
```
joelochlann: Generally we only want to use class selectors for styling, to avoid extra complexity around specificity: https://developer.mozilla.org/en/docs/Web/CSS/Specificity

static/src/stylesheets/module/experiments/_reading-time.scss
```diff
@@ -0,0 +1,141 @@
...
+        }
+
+        .selections {
+            padding-top: 10px;
+
+            .clock {
+                svg {
+                    @include mq(mobileLandscape) {
+                        transform-origin: left;
+                        transform: scale(0.5);
```
joelochlann: I would just set a width here rather than using `transform`

static/src/stylesheets/module/experiments/_reading-time.scss
```diff
@@ -0,0 +1,141 @@
...
+                left: 20px;
+            }
+        }
+        &__roundel {
+            float: right;
+            margin-top: 5px;
+            svg {
+                @include mq(mobileLandscape) {
+                    transform-origin: top right;
+                    transform: scale(0.65);
```
joelochlann: again, you should probably use `width` here

static/src/stylesheets/module/experiments/_reading-time.scss
```diff
@@ -0,0 +1,141 @@
...
+            display: inline-block;
+            padding-bottom: 15px;
+        }
+
+        .btn {
+            display: inline-block;
+            vertical-align: top;
+            position: relative;
+            top: 1px;
+            line-height: 30px;
```
joelochlann: This might be OK for a test, but generally you would try and use variables & multiples of variables for all of these - `$gs-baseline` for vertical height, `gs-span` for horizontal widths. This is for design consistency and comprehensibility of the code. Also for font-related stuff there are some typography helpers. Have a look in the `pasteup` directory for all of this

static/src/javascripts-legacy/projects/common/modules/experiments/tests/reading-time.js
```diff
@@ -0,0 +1,100 @@
...
+            var suggestion30 = 'If you have thirty minutes, you might enjoy our latest long read  <a href="https://www.theguardian.com/news/series/the-long-read/latest">'+arrow+'</a>';
+
+            bean.on($('.js-feedback-button-5', section)[0], 'click', ctaReplacer(suggestion5, completeFunction));
+            bean.on($('.js-feedback-button-15', section)[0], 'click', ctaReplacer(suggestion15, completeFunction));
+            bean.on($('.js-feedback-button-30', section)[0], 'click', ctaReplacer(suggestion30, completeFunction));
+        }
+
+        function ctaReplacer(suggestion, completeFunction) {
+            return function () {
+                $('.cta').empty();
```
joelochlann: `.cta` seems a bit dangerous as it's not unlikely there's something else with this class. I would namespace it with `reading-time__`

ditto `.selections`

joelochlann: no need to call `.empty()`, `.html()` will replace the contents

static/src/javascripts-legacy/projects/common/modules/experiments/tests/reading-time.js
```diff
@@ -0,0 +1,100 @@
...
+            }));
+            registerTimeSelectionHandlers($readingTimeSelection, completeFunction);
+
+            return fastdom.write(function() {
+                $readingTimeSelection.insertBefore($sectionBelowReadingTime);
+            });
+        }
+
+        function registerTimeSelectionHandlers(section, completeFunction) {
+            var arrow = '<div class="btn"><svg width="24" height="24" viewBox="240 334 24 24" xmlns="http://www.w3.org/2000/svg"><g fill="none" fill-rule="evenodd" transform="translate(240 334)"><path fill="#F6F6F6" d="M4 12.68h13.3l-5.22 6.15.68.67L20 12.34v-.68L12.76 4.5l-.68.67 5.2 6.15H4"/></path></g></svg></div>';
```
joelochlann: Can you not put all this stuff in the `reading-time` template?

joelochlann (changes requested):

static/src/javascripts-legacy/projects/common/modules/experiments/tests/reading-time.js
joelochlann: ... maybe not, since it happens on click!

superfrank:
Heya, looking good/seems like most of it is in place. Just a few little tweaks to spacing etc:

- I'd check the placement of the G/match it to the securedrop placement, it doesn't look like it's to the right enough to me. 
- I'd check the vertical padding of the buttons when they stack - should be spaced a $gs-baseline apart, see here for help on consistent padding/gutters: https://github.com/guardian/guss-grid-system
- The paragraph seems to be spaced quite far from the header - again i'd make it a $gs-baseline and then a $row-height below.
- If the clocks within the CTA can sit consistently to the left - a bit tighter to the left of the lozenge.

mobile screenshot attached for reference (let me know if you'd like me to talk anything through in person):

image

static/src/stylesheets/module/experiments/_reading-time.scss
lmath: Thanks Joe. I'm going to try and keep this in mind from the start next time.

lmath:
Thanks @superfrank. I think I've adopted your feedback in the commits after your comment.

static/src/javascripts-legacy/projects/common/modules/experiments/ab.js
```diff
@@ -55,7 +57,9 @@ define([
         new TailorSurvey(),
         SleevenotesNewEmailVariant,
         SleevenotesLegacyEmailVariant,
-        new IncreaseInlineAds()
+        new IncreaseInlineAds(),
+        SleevenotesLegacyEmailVariant,
```
joelochlann: Whoops! think this is a mistake

joelochlann (changes requested):

static/src/javascripts-legacy/projects/common/modules/experiments/tests/reading-time.js
```diff
@@ -0,0 +1,97 @@
...
+            var suggestion30 = 'If you have thirty minutes, you might enjoy our latest long read  <a href="https://www.theguardian.com/news/series/the-long-read/latest">'+arrow+'</a>';
+
+            bean.on($('.js-feedback-button-5', section)[0], 'click', ctaReplacer(suggestion5, completeFunction));
+            bean.on($('.js-feedback-button-15', section)[0], 'click', ctaReplacer(suggestion15, completeFunction));
+            bean.on($('.js-feedback-button-30', section)[0], 'click', ctaReplacer(suggestion30, completeFunction));
+        }
+
+        function ctaReplacer(suggestion, completeFunction) {
+            return function () {
+                $('.js-feedback-text-reading-time__cta').html(suggestion);
```
joelochlann: Actually this should probably be wrapped in `fastdom.write` as well. @sndrs can probably confirm that

joelochlann (changes requested):

static/src/javascripts-legacy/projects/common/modules/experiments/tests/reading-time.js
sndrs: yeah good catch @joelochlann – @lmath it's best if you can use https://github.com/guardian/frontend/blob/master/static/src/javascripts-legacy/lib/fastdom-promise.js when doing anything to the dom that will trigger layout, in this instance  a `write` as joe says. give me shout if it's not clear

lmath: Thanks @joelochlann and @sndrs. I'll keep that in mind. Is this right? https://github.com/guardian/frontend/pull/16145/files#diff-0a4033ccb9178b6dedf5917057e6aa34R87

sndrs: yep - looks good!

joelochlann (approved):
:champagne:

PR #16146, opened 2017-03-15
Spacing tweaks and improved trailing slash mixin

zeftilldeath:
## What does this change?
On the new header sub-menu it looks aesthetically better if words don't appear directly beneath the veggie burger. Currently to achieve that there is a right-margin: 

screen shot 2017-03-14 at 16 54 32

The problem with that, is when there are more links like 'clubs' in this example they wrap, despite there being room on the second row for them to be show.

To sort this out I've changed all links with trailing slashes to be inline-block with the font-size: 0; fix which allows for correct alignment when they wrap. They are no longer floats.

To further standardise all elements with trailing slashes I have made the menu unordered lists just like page tags and improved the logic of the trailing slash mixin to incorporate that.

Yep, updated amp to use this logic also.

## Screenshots
This illustrates the tap targets and the pink line illustrates the spacer than knocks the links onto a new line.

screen shot 2017-03-15 at 14 07 25

screen shot 2017-03-15 at 14 28 56

### DIFF
--- a/common/app/views/fragments/amp/header.scala.html
+++ b/common/app/views/fragments/amp/header.scala.html
@@ -16,17 +16,17 @@
         </a>
 
         <nav class="header__nav">
-            <ul class="header__nav__primary-links">
-            @NewNavigation.PrimaryLinks.map { link =>
-                <li class="header__nav__link-item">
-                    <a class="header__nav__link"
-                        href="@LinkTo(link.url)"
-                        data-link-name="amp : nav : @link.title">
-                            @link.title
-                    </a>
-                </li>
-            }
-        </ul>
+            <ul class="header__nav-primary-links">
+                @NewNavigation.PrimaryLinks.map { link =>
+                    <li class="header__nav-link-item">
+                        <a class="header__nav-link"
+                            href="@LinkTo(link.url)"
+                            data-link-name="amp : nav : @link.title">
+                                @link.title
+                        </a>
+                    </li>
+                }
+            </ul>
 
             <button class="header__nav__menu-button" on='tap:sidebar1.toggle'
                 data-link-name="amp : nav : toggle-veggie-burger">
--- a/common/app/views/fragments/amp/header.scala.html
+++ b/common/app/views/fragments/amp/header.scala.html
@@ -16,13 +16,17 @@
         </a>
 
         <nav class="header__nav">
+            <ul class="header__nav__primary-links">
             @NewNavigation.PrimaryLinks.map { link =>
-                <a class="header__nav__link"
-                    href="@LinkTo(link.url)"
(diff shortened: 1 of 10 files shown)


common/app/views/fragments/amp/header.scala.html
```diff
@@ -16,13 +16,17 @@ <h1 class="u-h">The Guardian</h1>
         </a>
 
         <nav class="header__nav">
+            <ul class="header__nav__primary-links">
             @NewNavigation.PrimaryLinks.map { link =>
```
NataliaLKB: Nit: the spacing is off here. make sure the opening and closing `ul` tags match!

common/app/views/fragments/nav/subSectionNav.scala.html
```diff
@@ -3,18 +3,19 @@
...
-            <a class="@RenderClasses(Map(
-                    "tertiary-navigation__link" -> true,
-                    "tertiary-navigation__link--current-section" -> (id == link.uniqueSection)
-                ))"
-                href="@LinkTo(link.url)" data-link-name="nav2 : tertiary : @{ if(link.longTitle.isEmpty) link.title else link.longTitle}">
-                    @link.title
-            </a>
+            <li class="tertiary-navigation__link-item">
+                <a class="@RenderClasses(Map(
+                        "tertiary-navigation__link" -> true,
```
NataliaLKB: separate to this PR maybe, but we really should change `tertiary` to `subnav` to keep names consistent. My fault!

common/app/views/fragments/amp/header.scala.html
```diff
@@ -16,13 +16,17 @@ <h1 class="u-h">The Guardian</h1>
...
-                <a class="header__nav__link"
-                    href="@LinkTo(link.url)"
-                    data-link-name="amp : nav : @link.title">
-                        @link.title
-                </a>
+                <li class="header__nav__link-item">
```
gustavpursche: In BEM there is only one level of nesting allowed :) So this should better be `header__nav-link-item`. Applies to a few cases following aswell.

gu-stav (commented):
Just a little BEM nit - rest looks great. 🌷

common/app/views/fragments/amp/header.scala.html
zeftilldeath: Thanks

common/app/views/fragments/amp/header.scala.html
zeftilldeath: Cheers, have changed them, please check to see I understood correctly.

NataliaLKB (approved):
lgtm 👍 !!

PR #16147, opened 2017-03-15
Check for getEntriesByName() before calling it

gu-stav:
## What does this change?

Adds a check around `window.performance.getEntriesByName()` before calling it. I saw this while investigating scroll issues on iPhones on IOS9.

Less JS errors, happier Sentry.

No.

## Screenshots

No.

## Tested in CODE?

No.


NataliaLKB (approved):
Awesome 👍

PR #16148, opened 2017-03-15
Design variations tweaks

desbo:
## What does this change?
* extend test run time 
* restyle highlighting
  * before 
screen shot 2017-03-15 at 16 26 41

  * after
screen shot 2017-03-15 at 16 26 17

### DIFF
--- a/static/src/stylesheets/module/experiments/_acquisitions-epic-design-variations.scss
+++ b/static/src/stylesheets/module/experiments/_acquisitions-epic-design-variations.scss
@@ -68,7 +68,9 @@
 
 .contributions__paragraph--highlight {
     background-color: #ffe399;
-    padding: 4px;
+    padding: 4px 2px;
+    margin-left: -2px;
+    margin-right: -2px;
 }
 
 .contributions__epic--paypal {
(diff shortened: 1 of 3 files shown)


static/src/stylesheets/module/experiments/_acquisitions-epic-design-variations.scss
```diff
@@ -67,7 +67,8 @@
 // ============
 
 .contributions__paragraph--highlight {
-    background-color: transparentize($multimedia-main-2, .6);
+    background-color: #ffe399;
+    padding: 4px;
```
dominickendrick: With my eagle DesignEye™ I think there is no left or right padding on the highlighting as it adds extra space.

Mullefa (approved):
👍

static/src/stylesheets/module/experiments/_acquisitions-epic-design-variations.scss
desbo: that's true but it also looked a bit odd with the highlight exactly on the start of the characters, I have instead added 2px horizontal padding and adjusted the margins to keep the spacing the same as it would be without the highlight

dominickendrick (approved):

PR #16149, opened 2017-03-15
Paid comments (0%) AB test

JustinPinner:
## What does this change?
We're introducing a test for paid commenting. Before we do though, we need to get a look at how it would play out for the reader, and have a chance to make it as smooth and hassle-free as possible.

This is a zero percent test so only visible when specifically (manually) opted in.

We want to make sure that commenting still works, and that the diversion introduced by the test is as light-touch as possible. And we want to have it assessed internally before it goes live to users on the site.

No, it's web only.

## Screenshots
paidcommentingendtoend1

## Tested in CODE?
YES!

cc @nicl @Ap0c @svillafe


nicl:
👍 happy with this from the Discussion side. I think the main thing is to go over it with @crifmulholland once it's on PROD and get any of his thoughts on the UI side (but I don't anticipate any issues).

JustinPinner:
Retested all these changes on CODE after @nicl's review. Looking reasonable. We'll apply styling & copy tweaks through a later PR.

PR #16154, opened 2017-03-16
Remove AB test events

SiAdcock:
## What does this change?

The `events` property is no longer needed, used or documented on AB Tests. It was introduced in #1698, and then removed from the docs in #3065. No existing AB Test has an `events` property, so the `getActiveTestsEventIsApplicableTo` and `isEventApplicableToAnActiveTest` always return falsey values.

- Less code to maintain
- Clickstream no longer dependent on AB tests
- Less head scratching, more time to enjoy life

No

## Tested in CODE?

No


regiskuckaertz (approved):
#dumpeverything

giphy

PR #16156, opened 2017-03-16
ES6 lib/fetch-json

SiAdcock:
## What does this change?

- ES6ifies and Flowifies `fetch-json`
- Jestifies `fetch-json.spec`
- Mocks `fetch-json` in `breaking-news.spec`. This change is not strictly necessary for this PR, and was part of a misguided effort to stub `fetch-json` in all tests that currently depend on it. I'll be happy to revert this change if deemed necessary.

✨  Shiny code ✨ 

No

## Tested in CODE?

No


static/src/javascripts/lib/fetch-json.js
```diff
@@ -0,0 +1,39 @@
+// @flow
+import config from 'lib/config';
+import fetch from 'lib/fetch';
+
+function json(input: string, init: any = {}) {
+    // #? we should use Object.assign for this assignment, but this currently breaks
+    // Karma tests that depend on fetch-json.js and have not been stubbed
+    const options = init;
```
regiskuckaertz: Have you tried `options = { ...init }` ?

SiAdcock: This transpiles down in the same way as using `Object.assign`

static/src/javascripts/lib/fetch-json.js
```diff
@@ -0,0 +1,39 @@
+// @flow
+import config from 'lib/config';
+import fetch from 'lib/fetch';
+
+function json(input: string, init: any = {}) {
```
gustavpursche: Is it really `:any` or can we assume `Object` instead?

static/src/javascripts/lib/fetch-json.js
```diff
@@ -0,0 +1,39 @@
...
+    // #? we should use Object.assign for this assignment, but this currently breaks
+    // Karma tests that depend on fetch-json.js and have not been stubbed
+    const options = init;
+    let path = input;
+
+    if (!input.match('^(https?:)?//')) {
```
gustavpursche: What about `input.startsWith()`?

package.json
```diff
@@ -8,6 +8,7 @@
     "babel-runtime": "^6.20.0",
     "bean": "~1.0.14",
     "bonzo": "~2.0.0",
+    "chance": "^1.0.6",
```
sndrs: devdep?

static/src/javascripts/lib/fetch-json.js
sndrs: `URL` instead of `string`?

static/src/javascripts/lib/fetch-json.js
sndrs: is this different from 
```js
function json(input: string, options: any = {})
```
and no 
```js
const options = init;
```
?

static/test/javascripts-legacy/spec/common/onward/breaking-news.spec.js
```diff
@@ -40,37 +37,35 @@ define([
         }
 
         function mockBreakingNewsWith(collections) {
+            fetchJson.and.callFake(function() {
```
sndrs: lovely

static/src/javascripts/lib/fetch-json.js
sndrs: no, sorry ignore. that's a `URL` object

sndrs:
we don't check that the input is a well formed URI, is that ok do you reckon?

static/src/javascripts/lib/fetch-json.js
SiAdcock: Functionally equivalent yes. I have structured it like this for 2 reasons:

1. [mutating the `init` object](https://github.com/guardian/frontend/blob/60bce285302cd92a4e845acafe7905dec0bf5827/static/src/javascripts/lib/fetch-json.js#L13) directly causes an ESLint error
2. Reassigning in this way acts as a reminder to do it properly using `Object.assign` later

gu-stav (commented):
Looks great! Just 2 small comments.

static/src/javascripts/lib/fetch-json.js
regiskuckaertz: Is it worth defining a [`RequestInit` dictionary](https://fetch.spec.whatwg.org/#requestinit) type here? It seems to be [already done](http://www.saltycrane.com/blog/2016/06/flow-type-cheat-sheet/#lib/dom.js), although under the erroneous name `RequestOptions`.

static/src/javascripts/lib/fetch-json.js
sndrs: woah – nopolyfills yet :)

SiAdcock:
@sndrs I think it should be the responsibility of the calling function to provide a valid URI. `fetch-json`, like `fetch`, like `reqwest` should try to do whatever it's told by the caller. What do you think?

regiskuckaertz:
@sndrs @SiAdcock why not copy the official [fetch signature](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch)?

static/src/javascripts/lib/fetch-json.js
SiAdcock: The object passed into this function is used to create another object that gets passed to `reqwest`. As such it doesn't have the same structure as `RequestInit` or `RequestOptions` but is something entirely bespoke for our application. 

I'll probably go with `Object` for now, since whatever type we define here will not be reusable.

SiAdcock:
@regiskuckaertz that would require us to refactor this function and `lib/fetch` which uses the `init` object as a seed for another object that gets created and passed to `reqwest`, which in turn bears only a passing resemblance to the official fetch signature. We would also have to refactor everything that calls `lib/fetch` and `lib/fetch-json`. 

It feels like the time for this refactor would be more appropriate when we ditch `reqwest` and start using official `fetch`

regiskuckaertz:
I see, fair enough. Note that the thing you mention is a simple morphism `coerce(o : RequestOptions) : ReqwestOptions`, I don't think Flow would complain about that.

> We would also have to refactor everything that calls lib/fetch and lib/fetch-json.

Wow 😮  If that, then every time you guys use Flow, all the calling sites need to be updated too. It reminds me of a Carl Sagan quote about an apple pie and the universe.

SiAdcock:
Okay, for some reason I thought we'd need to refactor everything that calls `lib/fetch-json` but those fears were misplaced. The signature of `lib/fetch-json` was not as different to official fetch as I thought, and Flow appears to be happy as you suggested, @regiskuckaertz. Apologies for taking a while to get there 😊

regiskuckaertz:
@SiAdcock 🎩  tip, Monsieur

gu-stav (commented):
🌷  Looks great.

gu-stav (approved):
Looks great 🌷

PR #16157, opened 2017-03-16
Fix immersive media atom template by using named parameter for mediaWrapper

gidsg:
## What does this change?
Previously we were passing unnamed parameters in the YouTube template. This meant the `mediaWrapper` wasn't being correctly set I think because the compiler thought it to be a different parameter.

Fixes immersive template

No

### Before
screen shot 2017-03-16 at 10 33 37
### After
screen shot 2017-03-16 at 10 35 50

## Tested in CODE?
No

CC @duarte


duarte:
Thanks Gids!

akash1810 (approved):

PR #16159, opened 2017-03-16
Add commercialFeatures.thirdPartyTags check to check-mediator/dispatcher

GHaberis:
## What does this change?

**Add thirdPartyTagsDisabled check to check-mediator/dispatcher**
- If thirdPartyTagsDisabled (commercialFeatures.thirdPartyTags is false) the hasHighPriorityAdLoaded/hasLowPriorityAdLoaded checks immediately return false as no ads will be loaded. This now means we don't needlessly run the trackAdRender methods in check-dispatcher on pages on where thirdPartyTagsDisabled is true (eg. identity).

**Update email sign-up check**
- Previously the check emailCanRun in check-mediator would resolve true if either isUserInEmailAbTestAndCanEmailBeAdded and isOutbrainMerchandiseCompliantOrBlockedByAds were true. It should have but didn't also check to see whether the following checks were true: 'emailCanRunPreCheck', 'listCanRun' and 'emailInArticleOutbrainEnabled'.
- I've updated email-article.js to first check whether emailCanRun ('emailCanRunPreCheck', 'listCanRun' and 'emailInArticleOutbrainEnabled' are all true) and if so then check whether the  'emailCanRunPostCheck' is also true.
- 'emailCanRunPostCheck'  checks to see whether the user 'isUserInEmailAbTest', 'isOutbrainMerchandiseCompliantOrBlockedByAds' or 'thirdPartyTagsDisabled' are true.
- if emailCanRun and emailCanRunPostCheck both pass then we can add an email sign-up form!

Prevents running needless checks on pages where thirdPartyTagsDisabled.

No

## Screenshots

N/A

## Tested in CODE?

No

### DIFF
--- a/static/src/javascripts-legacy/projects/common/modules/email/email-article.js
+++ b/static/src/javascripts-legacy/projects/common/modules/email/email-article.js
@@ -283,22 +283,25 @@ define([
 
     return {
         init: function () {
-            checkMediator.waitForCheck('emailCanRun').then(function (canEmailRun) {
-                if (canEmailRun) {
-                    emailRunChecks.getUserEmailSubscriptions().then(function () {
-                        if (ab.isParticipating({id: 'TailorRecommendedEmail'})) {
-                            switch (ab.getTestVariantId('TailorRecommendedEmail')) {
-                                case 'tailor-recommended': tailorInTest(); break;
-                                case 'control': tailorControl(); break;
-                                case 'tailor-random': tailorRandom(); break;
-                                default: addListToPage(find(listConfigs, emailRunChecks.listCanRun)); break;
-                            }
-                        } else {
-                            addListToPage(find(listConfigs, emailRunChecks.listCanRun));
-
+            checkMediator.waitForCheck('emailCanRun').then(function (emailCanRun) {
+                if (emailCanRun) {
+                    checkMediator.waitForCheck('emailCanRunPostCheck').then(function (emailCanRunPostCheck) {
+                        if (emailCanRunPostCheck) {
+                            emailRunChecks.getUserEmailSubscriptions().then(function () {
+                                if (ab.isParticipating({id: 'TailorRecommendedEmail'})) {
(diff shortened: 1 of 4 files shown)


GHaberis:
@davidfurey you may want to look at the **Update email sign-up check** change outlined above. I don't think there will be any email-signups currently showing when they shouldn't as subsequent logic in email-article prevents this.

GHaberis:
@regiskuckaertz just tagging you as i'm pulling in some commercial checks related to outbrain in the check-dispatcher.js (thirdPartyTagsDisabled, hasHighPriorityAdLoaded, hasLowPriorityAdLoaded).

davidfurey (approved):
Looks fine to me from an email point of view

static/src/javascripts-legacy/projects/common/modules/email/email-article.js
```diff
@@ -283,22 +283,25 @@ define([
...
-                                default: addListToPage(find(listConfigs, emailRunChecks.listCanRun)); break;
-                            }
-                        } else {
-                            addListToPage(find(listConfigs, emailRunChecks.listCanRun));
-
+            checkMediator.waitForCheck('emailCanRun').then(function (emailCanRun) {
+                if (emailCanRun) {
+                    checkMediator.waitForCheck('emailCanRunPostCheck').then(function (emailCanRunPostCheck) {
+                        if (emailCanRunPostCheck) {
+                            emailRunChecks.getUserEmailSubscriptions().then(function () {
```
regiskuckaertz: It's purely a matter of style, you might find this slightly more readable:
```
checkMediator.waitForCheck('emailCanRun')
.then(emailCanRun => emailCanRun && checkMediator.waitForCheck('emailCanRunPostCheck'))
.then(emailCanRunPostCheck => emailCanRunPostCheck && emailRunChecks.getUserEmailSubscriptions())
.then(subscriptions => {
  ...
})
.catch(error => robust.log('c-email', error));
```

regiskuckaertz:
Great work 👍

PR #16161, opened 2017-03-16
Add checks around window.performance.timing

gu-stav:
## What does this change?

Checks for existance before calling `window.performance.timing`.

Less JS errors in Safari.

No.

## Screenshots

No.


SiAdcock:
`navType` and `redirectCount` are [required fields](https://dashboard.ophan.co.uk/docs/thrift/event.html#Struct_WebPerformanceData) and need to be integer values. If either is omitted, or is a non-integer value, I believe the request will be ignored by Thrift. I think therefore it would be better not to make the call to Ophan at all, if `supportsNavigation` is false.

This structure is effectively[ a duplicate of a request made with Ophan's Tracker JS](https://github.com/guardian/ophan/blob/master/tracker-js/assets/coffee/ophan/perf.coffee) (which is installed in our application). If we want to reduce Sentry errors, we should probably make the same change in Ophan as well.

gustavpursche:
@SiAdcock Col, thanks. Didn't know about the documentation. Unfortunately I don't have access to the ophan-repo. Is this on purpose?

Could you give the code a new check? :)

SiAdcock (approved):

SiAdcock:
Out of interest do you have a link to an error that has been logged in Sentry? I'd like to determine how / whether this is also affecting Ophan Tracker JS to see if it might be having an impact on web performance data.

PR #16162, opened 2017-03-16
Add polyfill.io AB test

gu-stav:
## What does this change?

Adds a test for polyfill.io. We are trying to figure out, whether we can use polyfill.io as service to polyfill browser functions, dependent on your User-Agent. On the other hand, we want to know, how relieable the service is.

ES6 conversion, smaller builds, seperation of concerns, we don't have to maintain tests for our own polyfills anymore.

No.

## Screenshots

No.

## Tested in CODE?

No.


common/app/views/fragments/inlineJSBlocking.scala.html
```diff
@@ -6,29 +6,32 @@
...
 <script id="gu">
+    @if(!mvt.PolyfillIO.isParticipating) {
+        // ************* POLYFILLS *************
+        // Mostly handled by es5 shim above. Those which aren't are polyfilled ASAP.
 
-    // ************* POLYFILLS *************
-    // Mostly handled by es5 shim above. Those which aren't are polyfilled ASAP.
+        // It's faster to pass arguments in setTimeout than to use an anon function, but IE <10 can't do that.
+        // Used by RAF polyfill.
+        @Html(setTimeout().body)
```
sndrs: this was used in css loading, is not needed any more?

common/app/views/fragments/inlineJSBlocking.scala.html
```diff
@@ -6,29 +6,32 @@
...
 
-    // It's faster to pass arguments in setTimeout than to use an anon function, but IE <10 can't do that.
-    // Used by RAF polyfill.
-    @Html(setTimeout().body)
+        // Pollyfill requestAnimationFrame
+        @InlineJs(raf().body, "raf.js")
```
sndrs: this is used in nonblocking js, which may run before pf response comes back?

sndrs (changes requested):

common/app/views/fragments/inlineJSBlocking.scala.html
sndrs: >Used by RAF polyfill.

doh!

common/app/views/fragments/inlineJSBlocking.scala.html
gustavpursche: We checked it and couldn't find a call in the blocking code ... Would be great, if you could have a second look though :)

gtrufitt: Oops, my fault, forgot to check non blocking! You're right, it's used in there

gtrufitt: Would we ever see JS using these polyfills in the body? Could we move them to above the things that use them in `inlineJSNonBlocking` instead of blocking?

gtrufitt: Oh no, never mind, that's stupid, the main app needs them 🙄

common/app/views/fragments/inlineJSBlocking.scala.html
```diff
@@ -6,13 +6,14 @@
 
 @(page: model.Page)(implicit request: RequestHeader, context: model.ApplicationContext)
 
-<!--[if lt IE 9]>
-    <script src="@Static("javascripts/es5-html5.js")"></script>
-<![endif]-->
+@if(!mvt.PolyfillIO.isParticipating) {
+    <!--[if lt IE 9]>
+        <script src="@Static("javascripts/es5-html5.js")"></script>
```
sndrs: This cc is such an edge case but is the HTML patched by pfio too?

common/app/views/fragments/inlineJSBlocking.scala.html
```diff
@@ -23,12 +24,14 @@
     // Pollyfill requestAnimationFrame
     @InlineJs(raf().body, "raf.js")
 
+    @if(!mvt.PolyfillIO.isParticipating) {
+        // Polyfill Element.matches
+        @InlineJs(matches().body, "matches.js")
```
sndrs: Is this used?

sndrs (commented):
Couple of comments

common/app/views/fragments/inlineJSBlocking.scala.html
gustavpursche: Yes, e.g. https://github.com/guardian/frontend/blob/master/static/src/javascripts-legacy/projects/commercial/modules/messenger/get-stylesheet.js#L25

common/app/views/fragments/inlineJSBlocking.scala.html
gustavpursche: Ah, good catch. I didn't know `html5shiv` is bundled into this. So, we have to keep it for now.

But maybe interesting for your branch then:

```js
// bundle-shims.js

                require.resolve('es5-shim'),
                require.resolve('html5shiv'),
                path.resolve(
                    path.dirname(require.resolve('JSON2')),
                    'json2.js'
                ),
```

We could remove `require.resolve('es5-shim')` and `path.dirname(require.resolve('JSON2'))` then and add `JSON3` to pf.io aswell.

common/app/views/fragments/inlineJSBlocking.scala.html
sndrs: does it need to be inlined though? not sure it's used by any inlined js?

sndrs (approved):
query about matches, but given it's already like this, can leave it for now if you like – your call :)

PR #16167, opened 2017-03-16
Move huge images out of SVGs

regiskuckaertz:
The size of the commercial JS is in large part due to inlined assets, notably the adblock coins:

picture 1
picture 2
picture 3
picture 4

Let's move those suckers into static PNGs instead (thanks @SiAdcock for the insight of using `javascriptConfig`). People from the @guardian/membership-and-subscriptions team, so you're aware.


SiAdcock (approved):
Real nice, man ❤️

regiskuckaertz:
Why are tests failing? 😭

philwills:
@regiskuckaertz think those images need to go in `static/public`, not `static/src`

regiskuckaertz:
Thanks @philwills 😃

PR #16168, opened 2017-03-17
More email a/b tests

davidfurey:
## What does this change?

Adds a/b tests for fashion, film, books and long read emails

(NB- these tests will need to be updated once we have interactive signup pages)

Allows us to check whether the new designs for these emails have an effect on CTR and CTO rates.

No


joelochlann (approved):
Looks good! 👍

PR #16176, opened 2017-03-17
Add polyfill.io support

sndrs:
## What does this change?

- adds polyfill.io to the main JS app
- provides a checked in fallback should that fail to load within 3 secs
- wraps it in a switch incase pf.io is known to be down (fallback immediately)

plus:
- removes the es5/html5 shims
- manually polyfills some es7 features that polyfill.io doesn't include
- removes some inlined polyfills that didn't need to be there
- splits webpack config into common, dev and prod

should mean full modern env _and_ lighter bundles for (practically) everyone

## todo

- [x] add switch to kill pf.io request
- [x] auto-update fallback
- [x] check watch mode
- [x] get confirmation from polyfill.io that we're ok to go 100%


SiAdcock:
I possibly missed / forgot a conversation, but why are no longer including the ES5 shiv, while still including the HTML5 shiv? Why not remove both?

SiAdcock:
Ah, ignore me, I guess polyfill.io takes care of ES5 polyfills too 👍

static/src/javascripts/boot.js
```diff
@@ -13,85 +20,89 @@ import capturePerfTimings from 'lib/capture-perf-timings';
...
-// 1. Promise – https://webpack.js.org/guides/code-splitting-import/#promise-polyfill
-if (!window.Promise) {
-    window.Promise = Promise;
-}
-
-domready(() => {
-    // 1. boot standard, always
-    userTiming.mark('standard boot');
-    bootStandard();
+// kick of the app
```
SiAdcock: *off 😄

static/src/javascripts/polyfill.io
```diff
@@ -0,0 +1 @@
+https://polyfill.guim.co.uk/v2/polyfill.min.js?features=es6,es7,default-3.6&flags=gated&callback=guardianPolyfilled
```
phamann: Interested to know why you're serving this off a new hostname and not `https://assets.guim.co.uk/` with your other assets? This will incur connection and full roundtrip latency penalty.

sndrs: yeah that's a very good point – i'm not sure. will need to check with @janua

static/src/javascripts/boot.js
```diff
@@ -1,5 +1,12 @@
 // @flow
 
+// es7 polyfills not provided by pollyfill.io
+import 'core-js/modules/es7.object.values';
+import 'core-js/modules/es7.object.entries';
+import 'core-js/modules/es7.object.get-own-property-descriptors';
+import 'core-js/modules/es7.string.pad-start';
+import 'core-js/modules/es7.string.pad-end';
```
JakeChampion: string pad-end and pad-start will be in the next release 🎉  - https://github.com/Financial-Times/polyfill-service/pull/1117

static/src/javascripts/boot.js
```diff
@@ -1,5 +1,12 @@
 // @flow
 
+// es7 polyfills not provided by pollyfill.io
+import 'core-js/modules/es7.object.values';
```
JakeChampion: This should hopefully be in the next release -- https://github.com/Financial-Times/polyfill-service/pull/1082

static/src/javascripts/boot.js
```diff
@@ -1,5 +1,12 @@
...
+import 'core-js/modules/es7.object.entries';
```
JakeChampion: We have this polyfill :-) -- https://github.com/Financial-Times/polyfill-service/tree/a1bd243398089d126151ad5a0000c31437ed68af/polyfills/Object/entries

static/src/javascripts/boot.js
sndrs: aw that's our @regiskuckaertz – great!

static/src/javascripts/boot.js
sndrs: oops must have missed this, cheers @JakeChampion

JakeChampion:
@SiAdcock polyfill.io also serves the html5shiv to only the browsers which require it -- https://polyfill.io/v2/docs/features/#_html5_elements

static/src/javascripts/boot.js
JakeChampion: I updated the PR to ensure padEnd and padStart are not enumerable. Will run the test suite and have it in the next release :-)

static/src/javascripts/polyfill.io
sndrs: with fastly's idiot proof UI, i have managed to do this https://assets.guim.co.uk/polyfill.io/v2/polyfill.min.js

cheers for the prompt 👍

sndrs:
ok - think this is ready to de-WIP - thanks for your help @phamann and @JakeChampion!

.babelrc
```diff
@@ -1,44 +1,46 @@
 {
+    "plugins": [
+        "babel-plugin-transform-flow-strip-types"
```
sndrs: just do this all the time

static/vendor/javascripts/polyfillio.fallback.js
```diff
@@ -0,0 +1,5482 @@
...
+ * - Symbol.replace, License: MIT (required by "es6")
+ * - Symbol.search, License: MIT (required by "es6")
+ * - Symbol.split, License: MIT (required by "es6")
+ * - Symbol.toPrimitive, License: MIT (required by "es6")
+ * - Symbol.unscopables, License: MIT (required by "es6")
+ * - WeakMap, License: https://github.com/webcomponents/webcomponentsjs/blob/master/LICENSE.md (required by "es6")
+ * - WeakSet, License: https://github.com/webcomponents/webcomponentsjs/blob/master/LICENSE.md (required by "es6")
+ * - Array.prototype.includes, License: CC0 (required by "es7") */
+
+(function(undefined) {
```
JakeChampion: How come the fallback is unminified? It would be ~22KB minified instead of ~44KB unminified.

sndrs: it's minified during build, left it like this so we can see what's in it if we need to

gtrufitt (approved):
Awesome 💯

sndrs:
will be shipped @1100hrs on monday by @SiAdcock

janua:
@sndrs Missed this as was on holidays! Great job! 👏

static/src/javascripts/polyfill.io
janua: @phamann There was no major reason for doing this other than having a clean slate as a separate service in Fastly where we could monitor the affect of having ESIs. If I did it on `assets.guim.co.uk` I would have had to turn on ESIs there and would have found it harder to see the efficiency of the cache with all the other assets mixed in.

PR #16177, opened 2017-03-20
Replace uses of closest module with native

gtrufitt:
## What does this change?
Removes the uses of the closest module (polyfill) for using the native version which will be polyfilled on the prototype by polyfill.io.

Depends on #16176 

Removes an unneeded module.

No.

## Tested in CODE?
No.


regiskuckaertz (approved):
Sweet 👍

gu-stav (approved):
🎉 Looks good!

PR #16180, opened 2017-03-20
Paid containers - fix logo sizing

lps88:
…and tidy up the CSS a bit so that styles apply to SVG elements themselves rather than their wrappers.

Before:
image

After:
image


regiskuckaertz (approved):
Cool 👍

PR #16182, opened 2017-03-20
Convert `lib/robust` to ES6

gu-stav:
## What does this change?

Converts `lib/robust` and the corresponding specs to ES6. It also removes the `makeBlocks()` method, because of the duplication with `catchErrorsAndLogAll()`.

ES6!

No.

## Screenshots

No.

## Tested in CODE?

No.


static/src/javascripts/lib/robust.js
```diff
@@ -0,0 +1,53 @@
...
+function catchErrorsAndLog(name: string, fn: Function, reporter?: Function) {
+    const error = catchErrors(fn);
+
+    if (error) {
+        log(name, error, reporter);
+    }
+}
+
+function catchErrorsAndLogAll(modules: Array<any>) {
+    modules.forEach(pair => catchErrorsAndLog(pair[0], pair[1]));
```
regiskuckaertz: Just out of curiosity: have you decided on a list of language constructs you want/want not in the code? This one is a use case where destructuring would shine.

gustavpursche: I think we didn't finalize it, but we should do so soon. How would write the line if you could use destructuring?

regiskuckaertz: I was thinking sth like this: `modules.forEach(([name, fn]) => catchErrorsAndLog(name, fn))`

sndrs: i don't think we can say "you must destructure" but we can say "be as explicit for the next person as possible". in this case, it's hard for me see what this is a pair of? destructuring could make it clearer, so something like:
```js
modules.forEach(module => {
    const [whateverThisIs, andWhateverThisis] = module;
    catchErrorsAndLog(whateverThisIs, andWhateverThisis);
})
```
might make it clearer? realise that's not how it was in the original

is that what you meant @reg

regiskuckaertz: Yes exactly. To take an example, many promise libraries offered convenience methods to spread arrays and objects and spare you the trouble of manual destructuring. But now that we have it for free, we can just do it ourselves. I find it especially useful when we use arrays/objects as tuples rather than as collections.

sndrs: oop - crossed in time - but much nicer!

static/src/javascripts/lib/robust.js
```diff
@@ -0,0 +1,52 @@
...
+    try {
+        fn();
+    } catch (e) {
+        error = e;
+    }
+
+    return error;
+}
+
+function log(name: string, error: Error, customReporter?: Function): void {
```
SiAdcock: We could save some cruft by specifying a default argument
```
customReporter?: Function = reportError
```

static/src/javascripts/lib/robust.js
```diff
@@ -0,0 +1,52 @@
...
+
+    if (customReporter) {
+        reporter = customReporter;
+    }
+
+    if (window.console && window.console.warn) {
+        window.console.warn('Caught error.', error.stack);
+    }
+
+    reporter(error, { module: name }, false);
```
SiAdcock: If `name` was declared as `module` in the function parameter list, we could use the sweeter sugar:
```js
reporter(error, { module }, false);
```

static/src/javascripts/lib/robust.js
```diff
@@ -0,0 +1,52 @@
...
+}
+
+function catchErrorsAndLog(name: string, fn: Function, reporter?: Function) {
```
SiAdcock: Can we declare the return type as `void`?

static/src/javascripts/lib/robust.js
```diff
@@ -0,0 +1,52 @@
...
+    const error = catchErrors(fn);
+
+    if (error) {
+        log(name, error, reporter);
+    }
+}
+
+function catchErrorsAndLogAll(modules: Array<any>) {
```
SiAdcock: As above, return type is `void`

static/src/javascripts/lib/robust.js
```diff
@@ -0,0 +1,52 @@
...
+    return modules.map(pair => {
+        const [name, fn] = pair;
+        catchErrorsAndLog(name, fn);
+        return undefined;
```
SiAdcock: Are we intentionally building an array full of `undefined`? Should this be `forEach` instead of `map`?

static/src/javascripts/lib/robust.spec.js
```diff
@@ -0,0 +1,72 @@
...
+        expect(window.console.warn).toHaveBeenCalledTimes(1);
+
+        // reset window.console.warn
+        window.console.warn = origConsoleWarn;
+    });
+
+    test('catchErrorsAndLog() - default reporter', () => {
+        reportError.mockClear();
+        robust.catchErrorsAndLog('test', noError);
+        expect(reportError).not.toHaveBeenCalledWith();
```
SiAdcock: Are we testing to ensure that `reportError` has not been called at all? If so, I think `toHaveBeenCalled()` would work better

static/src/javascripts/lib/robust.js
gustavpursche: Nice one.

static/src/javascripts/lib/robust.spec.js
gustavpursche: Cool. Copy-Paste error.

static/src/javascripts/lib/robust.js
gustavpursche: Unfortunately we have to stick with `map` here, since this is called within `forEach` loops, which rely on returning an array. I tried, if I can pass the output of the actual function through, but they are somehow inconsistent ...

Do you have a better idea?

gustavpursche:
@SiAdcock We've got a 💚 build! Could you review this (again) and then, finally, let's get this merged!

static/src/javascripts/lib/robust.js
```diff
@@ -0,0 +1,53 @@
...
+    if (error) {
+        log(name, error, reporter);
+    }
+};
+
+const catchErrorsAndLogAll = (modules: Array<any>): Array<any> =>
+    modules.map(pair => {
+        const [name, fn] = pair;
+        catchErrorsAndLog(name, fn);
+        return undefined;
```
SiAdcock: I'm still not convinced this will work correctly. In the [existing implementation](https://github.com/guardian/frontend/blob/d99ec442f7bcbe0d1ade712c757cf83f6fc4f8c9/static/src/javascripts-legacy/lib/robust.js#L46-L48), `makeBlocks` maps over the `modules` array and for each module returns a new function that when executed will call `catchErrorsAndLog` with the context bound to `this`, with the first two parameters applied.

This function is executed in the `forEach` loop in [`enhanced/common`](https://github.com/guardian/frontend/blob/5805db95dc687111731d1111cb6c1abdde3559e1/static/src/javascripts-legacy/bootstraps/enhanced/common.js#L373) and [`enhanced/facia`](https://github.com/guardian/frontend/blob/5805db95dc687111731d1111cb6c1abdde3559e1/static/src/javascripts-legacy/bootstraps/enhanced/facia.js#L116)

Instead of returning `undefined` here, I believe we need to return `catchErrorsAndLog.bind(this, name, fn)`

What do you think?

SiAdcock (changes requested):
See comment under `catchErrorsAndLogAll`

gustavpursche:
@SiAdcock Thanks for the review - you were right: this implementation differed from the previous one.

Before the `catchErrorsAndLogAll()` you mentioned, were `makeBlocks()` which had the bound function as return value. I missed this, when converting them to `catchErrorsAndLogAll()`. Instead of changing the return value of `catchErrorsAndLogAll()` I updated the calls in `common` and `facia` to make all `catchErrorsAndLogAll()` equal.

Good catch! Thank you 🌹

SiAdcock (approved):

PR #16183, opened 2017-03-20
Remove tailor survey overlay

GHaberis:
## What does this change?

Removes the tailor survey overlay from frontend
- Remove switch
- Don't initialise from enhanced/main.js
- Remove JS/SCSS
- Remove getSurvey call from tailor.js
- Remove 'contains' module from enhanced/main as eslint warned this wasn't used

We have enough survey responses so no longer need to recruit anymore readers to fill it out.

No

## Screenshots

N/A

## Tested in CODE?

No


MahanaC (approved):
Looks great thanks!!

PR #16184, opened 2017-03-20
Update polyfill.io hosts

gu-stav:
## What does this change?

Updates the hosts used to fetch polyfill.io.

Shorter response times, due to less DNS, TLS roundtrips

No.

## Screenshots

No.

## Tested in CODE?

No.

### DIFF
--- a/common/app/views/fragments/head.scala.html
+++ b/common/app/views/fragments/head.scala.html
@@ -26,7 +26,7 @@ http://developers.theguardian.com/join-the-team.html
 
 @if(mvt.PolyfillIO.isParticipating) {
     <link rel="prefetch"
-          href="https://assets.guim.co.uk/polyfill-io/v2/polyfill.min.js?features=es6,Element.prototype.closest,Element.prototype.matches">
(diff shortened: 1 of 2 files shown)


common/app/views/fragments/head.scala.html
```diff
@@ -26,7 +26,7 @@
 
 @if(mvt.PolyfillIO.isParticipating) {
     <link rel="prefetch"
-          href="https://polyfill.guim.co.uk/v2/polyfill.min.js?features=es6,Element.prototype.closest,Element.prototype.matches">
+          href="https://assets.guim.co.uk/polyfill-io/v2/polyfill.min.js?features=es6,Element.prototype.closest,Element.prototype.matches">
```
jfsoul: would be nice to get the host from config even if this is a test

jfsoul: also it looks like the regex we're using is not escaping the full stop in `polyfill.io` - allowing `polyfill-io` to work.  Probably should avoid this and stick to `polyfill.io`

gustavpursche: Yeah, it's only for 2 days and after that @sndrs has a much better solution :)

jfsoul (approved):

sndrs (approved):

PR #16186, opened 2017-03-20
Cleanup Wednesday #7

regiskuckaertz:
I know it's early, but this is the _I'm sick of `commercial` being the biggest bundle_ edition\*.

- Adios manual templates (the last of the merchandising components): multiple, inline & single
- Bye bye Fabric Expandable Video v1
- **Thou shalt not include `svgs` anymore**: because we can cherry-pick the only 6 icons we need
- Some unused SVG icons
- Some small refactoring

- [x] Warning email sent
- [x] Deactivate templates in DFP

\* even if I managed to get from 150 to 89 KB minified/gzipped


JonNorman (approved):
Tasty :+1:

PR #16189, opened 2017-03-20
Subnav bug fix + tweaks

NataliaLKB:
## What does this change?
This pull request handles most of the bugs mentioned in [this google doc]( https://docs.google.com/document/d/1vCcuObMzneYm2-5JNpsHAoG7MOcVI1kNuvo3yRG9H4M/edit) as well as adds the slim header option to the new nav

🚫 🐛 🐛 

Nope

## Screenshots
**Before:**
This is showing the wrong subnav
image

**After:**
This is the right one
image

**Slim Nav**
image

## Tested in CODE?
Nope

cc @guardian/dotcom-platform

### DIFF
--- a/common/app/common/NewNavigation.scala
+++ b/common/app/common/NewNavigation.scala
@@ -309,7 +309,6 @@ object NewNavigation {
       SectionsLink("commentisfree/series/comment-is-free-weekly", inMyOpinion, Opinion),
       SectionsLink("profile/editorial", theGuardianView, Opinion),
 
-
       SectionsLink("sport", sport, Sport),
       SectionsLink("football", football, Sport),
       SectionsLink("sport/rugby-union", rugbyUnion, Sport),
@@ -518,11 +517,11 @@ object NewNavigation {
       if (page.metadata.sectionId == "commentisfree") {
         page.metadata.sectionId
       } else if (isTagPage) {
-        page.metadata.id
+        simplifySectionId(page.metadata.id)
       } else if (isArticleInTagPageSection) {
-        commonKeywords.head
+        simplifySectionId(commonKeywords.head)
       } else {
-        page.metadata.sectionId
+        simplifySectionId(page.metadata.sectionId)
       }
     }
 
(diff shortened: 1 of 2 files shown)


common/app/common/NewNavigation.scala
```diff
@@ -498,19 +498,19 @@ object NewNavigation {
 
     def getSectionOrTagId(page: Page) = {
       val tags = Navigation.getTagsFromPage(page)
-      val commonKeywords = tagPages.intersect(tags.keywordIds)
+      val commonKeywords = tags.keywordIds.intersect(tagPages)
```
jfsoul: As discussed, (being fussy) it would be nice if we could sort this explicitly and not rely on the implementation of `intersect`

NataliaLKB: Not fussy! Its a really good point. I have made the change. I spent some time looking, but I couldn't find a simpler way to do it than using `sortWith` like you showed me.

TBonnin: Does the `intersect` modify the order of the initial `List`? I don't think so but I might be wrong.

If you still want to make it explicit you can do it like this, it is shorter 😁  `tags.keywordIds.intersect(tagPages).sortBy(tags.keywordIds.indexOf)`

common/app/common/NewNavigation.scala
```diff
@@ -509,19 +509,19 @@ object NewNavigation {
 
     def getSectionOrTagId(page: Page) = {
       val tags = Navigation.getTagsFromPage(page)
-      val commonKeywords = tagPages.intersect(tags.keywordIds)
+      val commonKeywords = tags.keywordIds.intersect(tagPages).sortWith(tags.keywordIds.indexOf(_) < tags.keywordIds.indexOf(_))
```
DiegoVazquezNanini: I think we should add some tests to capture what should be the expected behaviour of this lines

jfsoul (approved):

common/app/common/NewNavigation.scala
NataliaLKB: @DiegoVazquezNanini I completely agree, but I will do this tomorrow, and get this change out today - its a blocker on some coms to go out to editorial. 

A lot of this logic needs tests, but up until yesterday there was still the chance of throwing all of this away (we still needed approval once the test was successful). Now I really want to go through and write some tests and rewrite things, and document it all.

PR #16194, opened 2017-03-21
Fix indexedDB InvalidStateError occurring in FF

gu-stav:
## What does this change?

As part of my "a sentry error a day" I looked into https://sentry.io/the-guardian/client-side-prod/issues/228307682/.

This Pr should not change the indended behavoir, but improve the code readability. It also adds a `try{} catch` around the indexedDB open call.

Less sentry errors.

No.

## Screenshots

No.

## Tested in CODE?

No.

### DIFF
--- a/static/src/javascripts-legacy/lib/private-browsing.js
+++ b/static/src/javascripts-legacy/lib/private-browsing.js
@@ -4,14 +4,13 @@ define([
     Promise
 ) {
     var browserCheck = new Promise(function (resolve) {
-        var db;
         var on = function () {
             resolve(true);
         };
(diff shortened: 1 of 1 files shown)


katebee:
Is Browsercheck still being used anywhere? 🤔 Although we left it as a util after deleting an AB test, I'm not sure if anything else has used it since...

If not, should we just 🔥🔥 this, and someone can resurrect it if they need it? :woman_shrugging:

gustavpursche:
@katebee After a chat to @oilnam we agreed that we can remove it from the last module which is using it. Great idea 🌷

PR #16197, opened 2017-03-21
Remove `lib/private-browsing`

gu-stav:
## What does this change?

Removes `lib/private-browsing`. After I tried [to fix the JS errors the module throws in FF](https://github.com/guardian/frontend/pull/16194) @katebee raised the question whether we still need this module. Turns out, only in one place from which @oilnam agreed to remove it.

So: let's 🔥 it.

Less code, less Sentry errors, less ES6 conversion, less "OMG this will be hard to test" code.

No.

## Screenshots

No.

## Tested in CODE?

No.


katebee (approved):
🔥 🔥 Farewell JS

💎 👍

oilnam:
👌  to 🔪 it 🙈

PR #16198, opened 2017-03-21
stop applying CSS to fc-container--thrasher

lmath:
## What does this change?
In my [reading time test PR](https://github.com/guardian/frontend/pull/16145), I've erroneously applied a style to the quite general class fc-container--thrasher, and this is affecting thrashers elsewhere when it shouldn't be. 

Removing a CSS bug I introduced. 

## Tested in CODE?

@guardian/dotcom-platform


sndrs (commented):
maybe `fc-container--thrasher--reading-time`?

joelochlann (approved):
Thanks @sndrs, this is a quick fix for a bug in production though (and a one-week test) so we'll leave that classname as is

PR #16200, opened 2017-03-21
Paid cards in editorial containers

lps88:
## What does this change?
Add an extra term to `data-link-name` for clarity in reporting. Also make the badge unclickable (should not link to client site here)

## Screenshots
image


regiskuckaertz (approved):
👍

PR #16201, opened 2017-03-21
Styling on new recipe design

blongden73:
Sorry for the masssssive PR. Mostly just styling tweaks to the new recipe article which looks something like this...

picture 340
picture 341

@NataliaLKB @SiAdcock @gustavpursche - Could you take a look please.

### DIFF
--- a/article/app/views/fragments/recipeArticleBody.scala.html
+++ b/article/app/views/fragments/recipeArticleBody.scala.html
@@ -107,15 +107,17 @@
 
 @recipePage(recipe: model.content.RecipeAtom, i: Int) = {
     <div data-recipe-index="recipe__@i" id="@recipe.atom.id" class="js-recipe__article--structured recipe__article--structured" itemtype="http://schema.org/Recipe">
+        <div class="recipe__article--structured-image mobile-only">
         @RecipeAtom.picture(recipe).orElse(maybeMainImage).map { image =>
             @fragments.image(
                 picture = image,
-                classes = Seq("recipe__image-inline, mobile-only"),
+                classes = Seq("recipe__image-inline mobile-only"),
                 imageAltText = "",
                 widths = MainMedia.inline,
                 isImmersiveMainMedia = true
             )
         }
+        </div>
         @for(credit <- recipe.data.credits){<h2 class="recipe__article--structured-byline">@recipe.data.credits</h2>}
         <div class="js-recipe__article--structured-headline recipe__article--structured-headline">
             <h1 itemprop="name">@recipe.data.title</h1>
@@ -158,6 +160,3 @@
         </div>
     </div>
 }
-
-
-
--- a/article/app/views/fragments/recipeArticleBody.scala.html
+++ b/article/app/views/fragments/recipeArticleBody.scala.html
@@ -107,6 +107,15 @@
 
 @recipePage(recipe: model.content.RecipeAtom, i: Int) = {
     <div data-recipe-index="recipe__@i" id="@recipe.atom.id" class="js-recipe__article--structured recipe__article--structured" itemtype="http://schema.org/Recipe">
+        @RecipeAtom.picture(recipe).orElse(maybeMainImage).map { image =>
+            @fragments.image(
+                picture = image,
+                classes = Seq("recipe__image-inline, mobile-only"),
+                imageAltText = "",
+                widths = MainMedia.inline,
+                isImmersiveMainMedia = true
+            )
+        }
         @for(credit <- recipe.data.credits){<h2 class="recipe__article--structured-byline">@recipe.data.credits</h2>}
         <div class="js-recipe__article--structured-headline recipe__article--structured-headline">
             <h1 itemprop="name">@recipe.data.title</h1>
--- a/article/app/views/fragments/recipeArticleBody.scala.html
+++ b/article/app/views/fragments/recipeArticleBody.scala.html
@@ -28,7 +28,7 @@
             <a href="https://www.theguardian.com/tone/recipes?CMP=recipe-article">
                 <div class="recipe__label"><h2>recipes</h2></div>
             </a>
-            <div class="recipe__image__wrapper recipe__image__wrapper--is-displayed">
+            <div class="js-recipe__image__wrapper recipe__image__wrapper recipe__image__wrapper--is-displayed">
                 @maybeMainImage.map{ image =>
                     @fragments.image(
                         picture = image,
@@ -54,7 +54,7 @@
                 </div>
             }
         </div>
-        <div class="recipe__gutter-wrapper">
+        <div class="js-recipe__gutter-wrapper recipe__gutter-wrapper">
             @defining(articleModel.item.content.tags) { tags =>
                 @if(tags.hasLargeContributorImage && tags.contributors.length == 1) {
                     <a href=@LinkTo{/@tags.contributors.map(_.id)}><div class="recipe__article--byline-image">@fragments.meta.bylineImage(tags)</div></a>
@@ -103,7 +103,6 @@
 </article>
 
 }
-
 
 @recipePage(recipe: model.content.RecipeAtom, i: Int) = {
     <div data-recipe-index="recipe__@i" id="@recipe.atom.id" class="js-recipe__article--structured recipe__article--structured" itemtype="http://schema.org/Recipe">
--- a/article/app/views/fragments/recipeArticleBody.scala.html
+++ b/article/app/views/fragments/recipeArticleBody.scala.html
@@ -25,7 +25,7 @@
 
     <div class="recipes__main">
         <div class="recipe__images-wrapper">
-            <a href="https://www.theguardian.com/tone/recipes?CMP=recipe-article">
+            <a href=@LinkTo("/tone/recipes?CMP=recipearticle")>
                 <div class="recipe__label"><h2>recipes</h2></div>
             </a>
             <div class="js-recipe__image__wrapper recipe__image__wrapper recipe__image__wrapper--is-displayed">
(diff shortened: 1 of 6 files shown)


static/src/stylesheets/module/content/_recipe-article.scss
```diff
@@ -1,9 +1,4 @@
 $grey: #dbdbdb;
```
gustavpursche: What about killing this too and replacing it by `$neutral-3` (`neutral-2` is a bit brighter, `neutral-4` is a bit darker).

static/src/stylesheets/module/content/_recipe-article.scss
```diff
@@ -103,34 +101,122 @@ $padding-wide: 100px;
...
 
 .recipe__label {
+    @include f-headline;
     position: absolute;
     z-index: 300;
-    background-color: $purple;
-    height: 90px;
-    width: 160px;
+    background-color: $features-main-2;
+    height: 65px;
```
gustavpursche: `$gs-baseline*5.5` ?

static/src/stylesheets/module/content/_recipe-article.scss
```diff
@@ -103,34 +101,122 @@ $padding-wide: 100px;
...
+    width: 110px;
     border-radius: 50%;
     text-align: center;
-    padding-top: 70px;
+    padding-top: 45px;
```
gustavpursche: `$gs-gutter*2.5` ?

common/app/views/support/HtmlCleaner.scala
```diff
@@ -522,7 +533,7 @@ case class ImmersiveHeaders(isImmersive: Boolean) extends HtmlCleaner {
   }
 }
 
-case class DropCaps(isFeature: Boolean, isImmersive: Boolean) extends HtmlCleaner {
+case class DropCaps(isFeature: Boolean, isImmersive: Boolean, isRecipeArticle: Boolean) extends HtmlCleaner {
```
tsop14: This is called in several other places without a third argument. Could fix error by adding default value: `case class DropCaps(isFeature: Boolean, isImmersive: Boolean, isRecipeArticle: Boolean = false) extends HtmlCleaner {`

static/src/stylesheets/module/content/_recipe-article.scss
```diff
@@ -103,34 +101,122 @@ $padding-wide: 100px;
...
+
+.recipe__gutter-wrapper {
+    display: none;
+
+    @include mq($from: desktop) {
+        display: flex;
+        flex-basis: 5%;
+        position: sticky;
+        top: 0;
+        height: 100vh;
```
gustavpursche: Should we be a bit more careful here, since we say we'd support older browsers? I'm not 100% sure ...

static/src/stylesheets/module/content/_recipe-article.scss
```diff
@@ -245,150 +326,114 @@ $padding-wide: 100px;
...
         &:nth-child(2) {
-            margin-left: 12px;
+            margin-left: calc(#{$gs-gutter / 5} * 2);
         }
     }
 }
 
 .recipe__article--structured-image {
     img {
+        object-fit: cover;
```
gustavpursche: Is it a problem that this will fail all Edge/ IE?

static/src/stylesheets/module/content/_recipe-article.scss
```diff
@@ -398,59 +443,92 @@ $padding-wide: 100px;
...
     }
 }
 
 .ad-wrapper {
-    background-color: $grey-two;
-    margin-left: -20px;
-    margin-top: 20px;
-    padding-top: 6px;
-    padding-bottom: 6px;
+    background-color: #f6f6f6;
```
gustavpursche: `$neutral-6` ?

gu-stav (changes requested):

static/src/stylesheets/module/content/_recipe-article.scss
blongden73: Oops...thats shouldn't be there!!

article/app/views/fragments/recipeArticleBody.scala.html
```diff
@@ -25,10 +25,10 @@
 
     <div class="recipes__main">
         <div class="recipe__images-wrapper">
-            <div class="recipe__label">
-                <h2>food &amp; drink</h2>
-            </div>
-            <div class="recipe__image__wrapper recipe__image__wrapper--is-displayed">
+            <a href="https://www.theguardian.com/tone/recipes?CMP=recipe-article">
```
SiAdcock: Can we use `/tone/recipes?CMP=recipe-article` here?

static/src/stylesheets/module/content/_recipe-article.scss
blongden73: @gustavpursche $neutral-6 is #ececec which is too dark as this is the background colour for the ad container it needs to be lighter. Would light($neutral-6, 10%) be acceptable?

article/app/views/fragments/recipeArticleBody.scala.html
blongden73: @SiAdcock This is something which doesn't exist in the data model anywhere so i dont think we can do it dynamically. Which part is troubling for you?

article/app/views/fragments/recipeArticleBody.scala.html
```diff
@@ -1,14 +1,14 @@
 @import layout.ContentWidths.MainMedia
```
NataliaLKB: You are still using immersive [media widths](https://github.com/guardian/frontend/blob/master/common/app/layout/WidthsByBreakpoint.scala#L185). These are needlessly big on mobile. Would you be able to change this soon?

article/app/views/fragments/recipeArticleBody.scala.html
NataliaLKB: @blongden73 given this is on www.theguardian.com, it would be better to do this:
```
@LinkTo("/tone/recipes?CMP=recipe-article")
```

Just like @SiAdcock was saying. That way on code, it will point to the code url and in dev it will be relative

article/app/views/fragments/recipeArticleBody.scala.html
```diff
@@ -93,9 +107,20 @@
...
-
 @recipePage(recipe: model.content.RecipeAtom, i: Int) = {
     <div data-recipe-index="recipe__@i" id="@recipe.atom.id" class="js-recipe__article--structured recipe__article--structured" itemtype="http://schema.org/Recipe">
+        <div class="recipe__article--structured-image mobile-only">
+        @RecipeAtom.picture(recipe).orElse(maybeMainImage).map { image =>
+            @fragments.image(
+                picture = image,
+                classes = Seq("recipe__image-inline mobile-only"),
+                imageAltText = "",
+                widths = MainMedia.inline,
```
NataliaLKB: Are [inline](https://github.com/guardian/frontend/blob/master/common/app/layout/WidthsByBreakpoint.scala#L152) the right widths for this? Might have to create your own if they are not.

static/src/stylesheets/module/content/_recipe-article.scss
NataliaLKB: I agree. Normally we would always have a fallback for 100vh, just in case.

static/src/stylesheets/module/content/_recipe-article.scss
```diff
@@ -103,34 +101,122 @@ $padding-wide: 100px;
...
+    background-color: $grey;
+    overflow: hidden;
+    top: $gs-baseline;
+    margin: auto;
+    left: 0;
+    right: 0;
+}
+
+.recipe__article--byline-image .byline-img .byline-img__img {
+    width: 150%;
```
NataliaLKB: Why 150% here?

static/src/stylesheets/module/content/_recipe-article.scss
NataliaLKB: Isn't this `$neutral-8` ?

NataliaLKB:
Some more comments/questions:
* What does this look like without JS? It doesn't have to look great, but the content does need to be readable
* Old browser support: that should be a blocker for going out to 100%, but might take some thinking about what a sensible fallback should be.
* There is quite a lot of css, is this being loaded on every article page? Could we only load it on recipe pages?

static/src/stylesheets/module/content/_recipe-article.scss
blongden73: @NataliaLKB We do 110% else where for the features template - I just wanted a slightly closer crop on the images as the byline image is quite small. 

https://github.com/guardian/frontend/blob/43bcdf3cd10d5a0839d843cd6a23042d1312e10d/static/src/stylesheets/module/content/_content.scss#L484

blongden73: I can add a comment to explain what its doing but its basically just centering the image in the circle.

NataliaLKB: Fair enough!

article/app/views/fragments/recipeArticleBody.scala.html
NataliaLKB: @blongden73 and I discussed, this is fine 😄

article/app/views/fragments/recipeArticleBody.scala.html
NataliaLKB: @blongden73 and I discussed, this is fine 😄

tsop14:
> What does this look like without JS? It doesn't have to look great, but the content does need to be readable

Without JS: clicking read more button takes to content at the bottom of the page. The LHS image will not update as you scroll.

When the JS runs it removes the original article content from the bottom and puts in a hidden div before the structured recipes - read more button updated to work as a hide/show. Thanks @gtrufitt for the advice, hope it's OK! 

no-js

static/src/stylesheets/module/content/_recipe-article.scss
tsop14: Added `min-height: 600px`

NataliaLKB:
Hey @blongden73 @tsop14 this branch looks like it needs a rebase with master.

Regarding my points...
* Thanks so much regarding the no js changes, it looks just fine I think.
* We agreed offline to make loading the css conditionally be a different PR

How about old browser support?

gu-stav (approved):

static/src/javascripts-legacy/bootstraps/enhanced/recipe-article.js
```diff
@@ -2,36 +2,47 @@ define([
...
 
         if (nextRecipe && nextRecipe.length > 0) {
             nextButton.removeClass('top');
-            nextRecipeTitle.html(nextRecipe.html());
+            nextRecipeTitle.html(nextRecipeHeadline.html());
             nextRecipeKicker.html('Next recipe ');
         } else {
             nextButton.addClass('top');
-            nextRecipeTitle.html(firstRecipe.html());
+            nextRecipeTitle.html(firstRecipeHeadline.html());
```
gustavpursche: Just a nit: this is redundant and could be moved out of the condition.

gustavpursche:
@blongden73 @tsop14 Looks good to me now! Thanks for working so hard on it 🌷 As we have spoken: let's work on better fallbacks for older browsers if the test is successful and this goes out to 100% of our audience.

PR #16205, opened 2017-03-21
Up the target audience to 25pc

oilnam:
It turned out that targeting 1pc of our audience isn't working, because we don't have enough top users in such a small segment, so I'm upping the targeting to 25pc 🙈 

Please note that this doesn't actually mean the surveys will be showed to 25pc of the audience, as the final decision wether to show a survey or not will be taken by Tailor.

@SiAdcock


SiAdcock (approved):
Nice 👍

oilnam:
@MahanaC FYI

PR #16206, opened 2017-03-21
Make up for paid container missing impressions

regiskuckaertz:
The paid container replacing the OB container is not getting enough impressions according to Rob. I have a hunch this is because OB may run even if there is no high merch slot in some cases. I've refactored a little so that the alt slot is inserted also _if_ the high merch slot is disabled but the outbrain flag is on.

hint: easier to look at each commit separately.


regiskuckaertz:
[Paid container vs Outbrain test: try to fix the lower number of impressions](https://trello.com/c/NaZogT3a/270-paid-container-vs-outbrain-test-try-to-fix-the-lower-number-of-impressions)

JonNorman (approved):

lps88 (approved):

PR #16209, opened 2017-03-21
Remove `lib/time`

gu-stav:
## What does this change?

Removes `lib/time`, due to lack of usage.

Less code, less cognitive load. 

No.

## Screenshots

No.

## Tested in CODE?

No.


sndrs (approved):
Ha!

PR #16210, opened 2017-03-21
Remove inline-source-map

SiAdcock:
## What does this change?

This change ensures sourcemaps will always be generated using `source-map` rather than [the now undocumented](https://webpack.js.org/configuration/devtool/) `inline-source-map`.

`inline-source-map` does not work in Firefox at all, making it very awkward to debug. `source-map` does produce usable sourcemaps in Firefox Developmer Edition, making it possible to add breakpoints in the debugger. However, errors in the console are still not mapped to the pristine source.

There is some very active investigation happening in Webpack-land around browser support for source maps (webpack/webpack#3165) so we may be able to start experimenting with the newer devtools soon.

No

## Tested in CODE?

No

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -5,7 +5,7 @@ const webpack = require('webpack');
 const bundlePrefix = 'graun.';
 
 module.exports = ({ env = 'dev', plugins = [] } = {}) => ({
-    devtool: env === 'dev' ? 'inline-source-map' : 'source-map',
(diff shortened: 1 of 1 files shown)


gu-stav (approved):
In case this works: sweet! 🍪 ☕️

sndrs:
we need babel 7 alpha

sndrs (approved):
good digging

PR #16211, opened 2017-03-21
Ensure advert exists before checking conditions

SiAdcock:
## What does this change?

Ensure an `advert` exists before checking its size.

Removes a whole bunch of odd error messages that have been plaguing our builds and causing confusion

No

## Screenshots

picture 203

## Tested in CODE?

No


gtrufitt (approved):
💡 :3rd_place_medal:

